PR #3726, opened 2024-04-17
fix: remove ttl option from metadata region

waynexia:

TTL is not applicable to metric engine metadata region

## Checklist

### DIFF
--- a/src/metric-engine/src/engine/create.rs
+++ b/src/metric-engine/src/engine/create.rs
@@ -381,6 +381,10 @@ impl MetricEngineInner {
         // concat region dir
         let metadata_region_dir = join_dir(&request.region_dir, METADATA_REGION_SUBDIR);
 
+        // remove TTL option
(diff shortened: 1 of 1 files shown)


src/metric-engine/src/engine/create.rs
```diff
@@ -381,6 +381,10 @@ impl MetricEngineInner {
         // concat region dir
         let metadata_region_dir = join_dir(&request.region_dir, METADATA_REGION_SUBDIR);
 
+        // remove TTL option
+        let mut options = request.options.clone();
+        options.remove("ttl");
```
fengjiachun, line 386: Perhaps it's better to throw an error message than silently delete?

waynexia: This is not an error case. TTL still applies for data regions.

fengjiachun (approved):
LGTM

MichaelScofield (approved):

PR #3727, opened 2024-04-17
refactor(meta): Ensure all moving values remain unchanged between two transactions

WenyXu:

1. Introduce `move_values` in `TombstoneManager`
2. Ensure all moving values remain unchanged between two transactions
3. Make `delete_table_metadata` and `restore_table_metadata` to be Idempotent.

## Checklist

### DIFF
--- a/src/common/meta/src/key.rs
+++ b/src/common/meta/src/key.rs
@@ -1332,14 +1332,17 @@ mod tests {
             .delete_table_metadata(table_id, &table_name, table_route_value)
             .await
             .unwrap();
-
+        // Should be ignored.
+        table_metadata_manager
+            .delete_table_metadata(table_id, &table_name, table_route_value)
+            .await
+            .unwrap();
         assert!(table_metadata_manager
             .table_info_manager()
             .get(table_id)
             .await
             .unwrap()
             .is_none());
-
         assert!(table_metadata_manager
             .table_route_manager()
             .table_route_storage()
@@ -1347,7 +1350,6 @@ mod tests {
             .await
             .unwrap()
             .is_none());
-
         assert!(table_metadata_manager
             .datanode_table_manager()
             .tables(datanode_id)
@@ -1362,7 +1364,6 @@ mod tests {
             .await
             .unwrap();
         assert!(table_info.is_none());
-
         let table_route = table_metadata_manager
             .table_route_manager()
             .table_route_storage()
@@ -1832,6 +1833,13 @@ mod tests {
             .delete_table_metadata(table_id, &table_name, &table_route_value)
             .await
             .unwrap();
+        table_metadata_manager
+            .restore_table_metadata(table_id, &table_name, &table_route_value)
+            .await
+            .unwrap();
+        let kvs = mem_kv.dump();
+        assert_eq!(kvs, expected_result);
+        // Should be ignored.
         table_metadata_manager
             .restore_table_metadata(table_id, &table_name, &table_route_value)
             .await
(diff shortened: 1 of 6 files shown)


tisonkun (commented):
What is the purpose of this refactor?

WenyXu:
> What is the purpose of this refactor?

For moving value to tombstone values operation, it ensures all values remain unchanged between two transactions instead of a single value in the previous version.

src/common/meta/src/key.rs
```diff
@@ -666,7 +659,7 @@ impl TableMetadataManager {
         let operations = self
             .table_metadata_keys(table_id, table_name, table_route_value)?
             .into_iter()
-            .map(|key| TxnOp::Delete(key.into_bytes()))
+            .map(TxnOp::Delete)
```
fengjiachun, line 662: Better to use [batch_delete](https://github.com/GreptimeTeam/greptimedb/blob/09b3c7029b7d7ac8abd1cebb613c9a00ea7c66bb/src/common/meta/src/kv_backend.rs#L65), it has no keys count limit.

src/common/meta/src/key/tombstone.rs
```diff
@@ -29,307 +29,158 @@ pub(crate) struct TombstoneManager {
...
-        impl FnMut(&'_ mut TxnOpGetResponseSet) -> Option<Vec<u8>>,
-    ) {
-        TxnOpGetResponseSet::build_get_op(to_tombstone(self.0))
+        src_key: Vec<u8>,
+        value: Vec<u8>,
+        dest_key: Vec<u8>,
+    ) -> (Txn, impl FnMut(&mut TxnOpGetResponseSet) -> Option<Vec<u8>>) {
+        let txn = Txn::new()
+            .when(vec![
+                Compare::with_not_exist_value(dest_key.clone(), CompareOp::Equal),
```
fengjiachun, line 63: Do not need compare `dest_key`

src/common/meta/src/key/tombstone.rs
```diff
@@ -29,307 +29,158 @@ pub(crate) struct TombstoneManager {
...
-    /// Returns a new normal key.
-    pub(crate) fn new<T: Into<Vec<u8>>>(key: T) -> Self {
-        Self {
-            bytes: key.into(),
-            atomic: false,
+        error::MoveValuesSnafu {
+            err_msg: format!(
+                "keys: {:?}, dest keys: {:?}",
+                keys.iter().map(|key| String::from_utf8_lossy(key)),
+                dest_keys.iter().map(|key| String::from_utf8_lossy(key))
```
fengjiachun, line 121: Do not need to give `dest_keys` since we can obtain them from the `keys`.

fengjiachun (approved):
LGTM

src/common/meta/src/rpc/store.rs
```diff
@@ -509,6 +509,13 @@ impl BatchDeleteRequest {
         }
     }
 
+    #[inline]
+    /// Sets `keys`.
```
tisonkun, line 513: This should violate the latest style guide to hold comment above attributions.

And IIRC @killme2008 suggests that we never use `#[inline]` but let the compiler decide it.

src/common/meta/src/rpc/store.rs
```diff
@@ -509,7 +509,6 @@ impl BatchDeleteRequest {
         }
     }
 
-    #[inline]
```
fengjiachun, line 512: remove all `inline` in this file.

src/common/meta/src/key/tombstone.rs
```diff
@@ -29,307 +29,154 @@ pub(crate) struct TombstoneManager {
...
+                    );
+                    (txn, (key.clone(), filter))
+                })
+                .unzip();
+            let mut resp = self.kv_backend.txn(Txn::merge_all(txns)).await?;
+            if resp.succeeded {
+                return Ok(());
+            }
+            let mut set = TxnOpGetResponseSet::from(&mut resp.responses);
+            // Updates results.
```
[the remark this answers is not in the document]
WenyXu: No, it only retrieves the latest values and starts a new transaction.

MichaelScofield (approved):

PR #3740, opened 2024-04-18
build: update toolchain to nightly-2024-04-18

waynexia:

- Allows `clippy.readonly_write_lock = "allow"`. We need to check it afterward.
- Remove some unused structs

## Checklist


src/table/src/predicate.rs
```diff
@@ -573,6 +573,7 @@ mod tests {
         let file = std::fs::OpenOptions::new()
             .write(true)
             .create(true)
+            .truncate(true)
```
tisonkun, line 576: Seems unrelated change?

src/query/src/tests/time_range_filter_test.rs
```diff
@@ -58,7 +58,7 @@ struct DataSourceWrapper {
 
 impl DataSource for DataSourceWrapper {
     fn get_stream(&self, request: ScanRequest) -> Result<SendableRecordBatchStream, BoxedError> {
-        *self.filter.write().unwrap() = request.filters.clone();
+        self.filter.write().unwrap().clone_from(&request.filters);
```
tisonkun, line 61: I guess this comes from clippy, but why `clone_from`/`clone_into` preferred?

src/common/time/src/timestamp.rs
```diff
@@ -381,7 +381,7 @@ impl Timestamp {
 
     pub fn from_chrono_datetime(ndt: NaiveDateTime) -> Option<Self> {
         let sec = ndt.and_utc().timestamp();
-        let nsec = ndt.timestamp_subsec_nanos();
+        let nsec = ndt.and_utc().timestamp_subsec_nanos();
```
tisonkun, line 384: This seems to change the semantic. Why do we need it?

src/common/procedure/src/local/rwlock.rs
```diff
@@ -18,6 +18,8 @@ use std::sync::{Arc, Mutex};
 
 use tokio::sync::{OwnedRwLockReadGuard, OwnedRwLockWriteGuard, RwLock};
 
+// Override warnings that those fields are not accessed.
+#[allow(dead_code)]
```
tisonkun, line 22: Is here a `TODO` to drop this suppression?

src/common/meta/src/state_store.rs
```diff
@@ -79,6 +79,8 @@ fn decode_kv(kv: KeyValue) -> Result<(String, Vec<u8>)> {
     Ok((key, value))
 }
 
+// Override warnings that those fields are not accessed.
+#[allow(dead_code)]
```
tisonkun, line 83: Is here a TODO to drop this suppression?

src/common/meta/src/sequence.rs
```diff
@@ -164,15 +164,15 @@ impl Inner {
 
             let req = CompareAndPutRequest {
                 key: key.to_vec(),
-                expect,
+                expect: expect.clone(),
```
tisonkun, line 167: This seems to introduce an extra clone?

tisonkun (commented):
Rest LGTM. Great patch!

src/common/meta/src/sequence.rs
waynexia: I don't know why it didn't need before

src/common/procedure/src/local/rwlock.rs
waynexia: This doesn't look like a TODO. How would you like to proceed?

src/common/time/src/timestamp.rs
waynexia: it's deprecated

src/mito2/src/cache/cache_size.rs
```diff
@@ -137,6 +137,6 @@ mod tests {
     fn test_parquet_meta_size() {
         let metadata = parquet_meta();
 
-        assert_eq!(956, parquet_meta_size(&metadata));
+        assert_eq!(964, parquet_meta_size(&metadata));
```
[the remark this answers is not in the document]
waynexia: Maybe some indirect dep changes its implement? IDK

src/table/src/predicate.rs
waynexia: We need to truncate here.

waynexia: https://rust-lang.github.io/rust-clippy/master/index.html#/suspicious_open_options

src/common/meta/src/sequence.rs
tisonkun: This is because it's a retry loop. If the request failed, we always update `expect` so NLL analysis should know here is no need for cloning.

If it can't compile, either the `clone_from` clippy has an edge case, or it's a compiler regression.

src/table/src/predicate.rs
tisonkun: Got.

src/mito2/src/cache/cache_size.rs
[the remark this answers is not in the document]
tisonkun: Just wonder if it would be a breaking data format change ...

src/common/procedure/src/local/rwlock.rs
tisonkun: Pushed b7a57b38f85953d07d4085cf1b3d43986a224e6e to resolve this.

src/common/meta/src/state_store.rs
tisonkun: Pushed b7a57b38f85953d07d4085cf1b3d43986a224e6e to resolve this.

src/common/meta/src/state_store.rs
```diff
@@ -79,19 +80,21 @@ fn decode_kv(kv: KeyValue) -> Result<(String, Vec<u8>)> {
...
 }
 
-fn split_value(value: &[u8], max_value_size: Option<usize>) -> SplitValue<'_> {
+fn split_value(value: Vec<u8>, max_value_size: Option<usize>) -> SplitValue {
     if let Some(max_value_size) = max_value_size {
         if value.len() <= max_value_size {
             SplitValue::Single(value)
         } else {
-            SplitValue::Multiple(value.chunks(max_value_size).collect::<Vec<_>>())
+            let mut values = vec![];
```
tisonkun, line 93: This also reduce a clone. Since previously:

```rust
    pub fn with_value(mut self, value: impl Into<Vec<u8>>) -> Self {
        self.value = value.into();
        self
    }
```

would clone the inner value chunk.

src/common/meta/src/sequence.rs
tisonkun: Resolve in 9a8bcac12b6f1e68c899ddcf88187d216262228e.

src/mito2/src/cache/cache_size.rs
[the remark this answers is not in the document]
tisonkun: Resolved as we have parquet read/write test.

tisonkun (approved):
LGTM.

Cargo.toml
```diff
@@ -70,7 +70,10 @@ license = "Apache-2.0"
 clippy.print_stdout = "warn"
 clippy.print_stderr = "warn"
 clippy.implicit_clone = "warn"
+clippy.readonly_write_lock = "allow"
```
tisonkun, line 73: This should be a todo that we later revisit if locks can improved. And we should suppress the warning in place where it's not the case.

We can create an issue for it.

tisonkun:
cc @MichaelScofield @evenyag @v0y4g3r @zhongzc PTAL.

MichaelScofield (approved):

src/common/meta/src/state_store.rs
```diff
@@ -79,17 +80,21 @@ fn decode_kv(kv: KeyValue) -> Result<(String, Vec<u8>)> {
...
+            for chunk in value.into_iter().chunks(max_value_size).into_iter() {
```
evenyag, line 94: @WenyXu Will this introduce additional allocations? The original `SplitValue` holds `Vec<&[u8]>`.

tisonkun: No. This call into_iter. Actually the previous code has an extra alloc. See LibraryLoader.

tisonkun: We move the bytes here. Previously it uses a ref but we can actually take the ownership. This calls delegate to `itertools` `IntoChunks`, not the slice's `chunks` that may clone.

Issue #2144, opened 2023-08-10
Using SecretString to store senstive values in `copy` stmt


MichaelScofield:
### What problem does the new feature solve?

`copy` stmt can carry senstive data like S3 API key or secret. This makes leaking easily.

### What does the feature do?

Right now the senstive values are all stored in plain strings, and we have to write special redacting codes for printing them. It's better we use [`SecretString`](https://docs.rs/secrecy/latest/secrecy/type.SecretString.html) for them.

A follow up of https://github.com/GreptimeTeam/greptimedb/pull/3744. 

### Implementation challenges

tisonkun:
Current code -

* https://github.com/GreptimeTeam/greptimedb/pull/2141
* https://github.com/GreptimeTeam/greptimedb/pull/3080

A thoroughly solution should be, supporting reconstruct query from statements, and redact options of copy stmt on demand.

See:

```rust
match self.query_statement(stmt, query_ctx.clone()).await {
    Ok(output) => {
        let output_result =
            query_interceptor.post_execute(output, query_ctx.clone());
        results.push(output_result);
    }
    Err(e) => {
        let redacted = sql::util::redact_sql_secrets(query.as_ref());
        error!(e; "Failed to execute query: {redacted}");

        results.push(Err(e));
        break;
    }
}
```

We currently print all the query if any of the statement failed to execute.

MichaelScofield:
We can make the sensetive data stored in their own fields, rather then the `OptionMap`.

tisonkun:
Then the code to access an option value is somehow awkward. If we insist in trying `SecretString`, I may suggest:

```rust
pub struct OptionMap {
    map: HashMap<String, String>,
    secrets: HashMap<String, String>,
}
```

And do the merge read within the encapsulation of `OptionMap`.

MichaelScofield:
I'm ok with moving all the senstive data to a "secrets" map like that.

PR #1336, opened 2023-04-06
feat: show create table

killme2008:
I hereby agree to the terms of the [GreptimeDB CLA](https://gist.github.com/xtang/6378857777706e568c1949c7578592cc)

## What's changed and what's your intention?

Impl `show create table [table]` statement.

```sql

mysql> show create table monitor;
+---------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table   | Create Table                                                                                                                                                                                                     |
+---------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| monitor | CREATE TABLE IF NOT EXISTS monitor (
  host STRING NULL,
  ts TIMESTAMP NOT NULL,
  cpu DOUBLE NULL DEFAULT 0,
  memory DOUBLE NULL,
  TIME INDEX (ts),
  PRIMARY KEY (host)
)
ENGINE=mito
WITH(
  regions = 1
) |
+---------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.01 sec)

```

Main changes:

* impl `Display` for create table statements.
* Create `CreateTable`  statement from table info.
* Impl show create table handlers.
* Fixed column comment lost.

## Checklist

- [ ]  I have written the necessary rustdoc comments.
- [x]  I have added the necessary unit tests and integration tests.

## Refer to a related PR or issue link (optional)

Close #535

PR #3744, opened 2024-04-19
feat: impl Display for Statement

### LINKED ISSUE #3646: impl Display for Statement

API improvement

sqlparser's Statement implements Display, we need to implement Display for our Statement, named `sql::statement::Statement` also.

See 

```rust
match self.query_statement(stmt, query_ctx.clone()).await {
    Ok(output) => {
        let output_result =
            query_interceptor.post_execute(output, query_ctx.clone());
        results.push(output_result);
    }
    Err(e) => {
        let redacted = sql::util::redact_sql_secrets(query.as_ref());
        error!(e; "Failed to execute query: {redacted}");

        results.push(Err(e));
        break;
    }
}
```

We currently print all the query if any of the statement failed to execute. We should write instead:

```rust
match self.query_statement(stmt, query_ctx.clone()).await {
    Ok(output) => {
        let output_result =
            query_interceptor.post_execute(output, query_ctx.clone());
        results.push(output_result);
    }
    Err(e) => {
        error!(e; "Failed to execute query: {stmt}");

        results.push(Err(e));
        break;
    }
}
```

with `stmt` probably redacted. And formatting structural `stmt` would be more reliable to redact fields, instead of depending on brittle regexps.

### Implementation challenges

_No response_

irenjj:

close #3646

Implement Display for Statement.

## Checklist


irenjj:
hi @tisonkun , I've finished the main logic. Could you please help me check if there are any issues? Afterwards, I will add tests based on this.

src/sql/src/statements/drop.rs
```diff
@@ -65,4 +84,21 @@ impl DropDatabase {
...
+            ""
+        }
+    }
+}
+
+impl Display for DropDatabase {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let if_exists = self.format_if_exists();
+        let name = self.name();
+        write!(f, r#"DROP DATABASE {if_exists} {name}"#)
```
tisonkun, line 102: Ditto.

src/sql/src/statements/set_variables.rs
```diff
@@ -21,3 +23,25 @@ pub struct SetVariables {
...
+    pub fn variable(&self) -> &ObjectName {
+        &self.variable
+    }
+
+    pub fn format_value(&self) -> String {
+        self.value
+            .iter()
+            .map(|expr| format!("{}", expr))
+            .collect::<Vec<_>>()
+            .join(", ")
```
tisonkun, line 37: I guess our set value should be always one element. But this is out of the scope of this PR. Just to comment it clearly.

src/sql/src/statements/show.rs
```diff
@@ -51,6 +51,48 @@ pub struct ShowColumns {
...
+    }
+}
+
+impl Display for ShowColumns {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let kind = self.kind();
+        let table = self.format_table();
+        let database = self.format_database();
+        let full = self.format_full();
+        write!(f, r#"SHOW {full} {table} {database} {kind}"#)
```
tisonkun, line 92: Similarly extra space issue.

src/sql/src/statements/show.rs
```diff
@@ -74,18 +155,78 @@ pub struct ShowTables {
...
+    }
+
+    #[inline]
+    fn format_full(&self) -> &str {
+        if self.full {
+            "FULL"
+        } else {
+            ""
+        }
+    }
```
tisonkun, line 180: Let's try to all inline the function manually, instead of use an extra function. In this way, we can handle whitespace more precise also.

Like sqlparser-rs:

```rust
            Statement::SetTimeZone { local, value } => {
                f.write_str("SET ")?;
                if *local {
                    f.write_str("LOCAL ")?;
                }
                write!(f, "TIME ZONE {value}")
            }
```

src/sql/src/statements/tql.rs
tisonkun, line 1: @waynexia please review if this unparse logic correct for tql.

src/sql/src/statements/copy.rs
```diff
@@ -43,6 +120,36 @@ pub struct CopyDatabaseArgument {
     pub location: String,
 }
 
+impl CopyDatabaseArgument {
+    #[inline]
+    fn format_with(&self) -> String {
+        if self.with.map.is_empty() {
+            String::default()
+        } else {
+            let options = format_sorted_hashmap!(&self.with.map);
```
tisonkun, line 129: No all options should be unparsed as is. This is the original motivation of https://github.com/GreptimeTeam/greptimedb/issues/3646. We'd like to redact sensitive options like `access_key_id` and `secret_access_key`.

Although we may do it in a follow-up PR, this patch should be open to that change instead of making the wrong assumption.

src/sql/src/statements/create.rs
```diff
@@ -229,6 +265,68 @@ pub struct CreateExternalTable {
...
+    fn format_if_not_exists(&self) -> &str {
+        if self.if_not_exists {
+            "IF NOT EXISTS"
+        } else {
+            ""
+        }
+    }
+
+    #[inline]
+    fn format_options(&self) -> String {
```
tisonkun, line 296: ditto for all these options. We may make a util for formatting or redacting options (map sensitive_key=value to sensitive_key="***") however.

tisonkun (commented):
Thank you! This patch goes in the good direction. Comments inline.

src/sql/src/statements/copy.rs
```diff
@@ -27,6 +28,7 @@ macro_rules! format_sorted_hashmap {
         let mut result = String::new();
         for key in sorted_keys {
             if let Some(val) = hashmap.get(key) {
+                redact_sql_secrets(val);
```
tisonkun, line 31: This is not how it works. You may instead check if key equals (case-insensitively) to `access_key_id` or `secret_access_key`.

Take a look at `redact_sql_secrets` and `SQL_SECRET_PATTERNS` so you may know how the previous logic actually works.

src/sql/src/statements/tql.rs
```diff
@@ -55,6 +124,41 @@ pub struct TqlAnalyze {
...
+        let step = &self.step;
+        let lookback = self.format_lookback();
+        let query = &self.query;
+        let is_verbose = self.format_is_verbose();
+        write!(
+            f,
+            "TQL ANALYZE {is_verbose} ({start}, {end}, {step}{lookback}) {query}"
+        )
+    }
+}
```
tisonkun, line 160: Seems can still have the extra whitespace issue. Let's inline the code directly.

src/sql/src/statements/tql.rs
```diff
@@ -43,6 +77,41 @@ pub struct TqlExplain {
...
+        let step = &self.step;
+        let lookback = self.format_lookback();
+        let query = &self.query;
+        let is_verbose = self.format_is_verbose();
+        write!(
+            f,
+            "TQL EXPLAIN {is_verbose} ({start}, {end}, {step}{lookback}) {query}"
+        )
+    }
+}
```
tisonkun, line 113: Ditto

src/sql/src/statements/show.rs
```diff
@@ -74,18 +149,72 @@ pub struct ShowTables {
...
+        &self.variable
+    }
+}
+
+impl Display for ShowVariables {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let variable = self.variable();
+        write!(f, r#"SHOW VARIABLES {variable}"#)
+    }
+}
```
tisonkun, line 216: Let's try to inline everywhere unless it's too complex to extract to a method. Such an indirection doesn't help much but introduce extra understanding burden.

src/sql/src/statements/show.rs
```diff
@@ -74,18 +149,72 @@ pub struct ShowTables {
...
+        &self.table_name
+    }
+}
+
+impl Display for ShowCreateTable {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let table_name = self.table_name();
+        write!(f, r#"SHOW CREATE TABLE {table_name}"#)
+    }
+}
```
tisonkun, line 197: ditto

src/sql/src/statements/show.rs
```diff
@@ -74,18 +149,72 @@ pub struct ShowTables {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("SHOW")?;
+        if self.full {
+            f.write_str(" FULL")?;
+        }
+        let database = self.format_database();
+        let kind = self.kind();
+        write!(f, r#" TABLES {database} {kind}"#)
+    }
+}
```
tisonkun, line 178: ditto - perhaps extra whitespace.

tisonkun (changes requested):
Let's try to inline everywhere unless it's too complex to extract to a method. Such an indirection doesn't help much but introduce extra understanding burden.

src/sql/src/statements/show.rs
```diff
@@ -53,19 +53,15 @@ pub struct ShowColumns {
...
         }
-
-        let table = format!("IN {}", &self.table);
-        let database = match &self.database {
-            Some(database) => format!("IN {}", database),
-            None => String::default(),
-        };
-
-        let kind = &self.kind;
-        write!(f, r#" COLUMNS {table} {database} {kind}"#)
```
tisonkun: @irenjj I pushed a commit to showcase my suggestions here.

In the previous pattern, ` {database} ` can still have an extra whitespace if database is empty.

And why I'd prefer this fallthrough way to the construct and then fill the string template?

This is because formatting (display a struct) is a linear job. If you write code like:

```rust
        let if_not_exists = self.format_if_not_exists();
        let name = &self.name;
        let columns = format_list_indent!(self.columns);
        let constraints = self.format_constraints();
        let partitions = self.format_partitions();
        let engine = &self.engine;
        let options = self.format_options();
        let maybe_external = if self.engine == FILE_ENGINE {
            "EXTERNAL "
        } else {
            ""
        };
        write!(
            f,
            r#"CREATE {maybe_external}TABLE {if_not_exists} {name} (
{columns},
{constraints}
)
{partitions}ENGINE={engine}
{options}"#
        )
```

You may jump up and down to check the template is correct, and every part is correct. But we're just printing chars, so we can `write!` on the fly.

However, I noticed that the original implementation you may refer to https://github.com/GreptimeTeam/greptimedb/pull/1336 by @killme2008. So we still have some existing code to migrate. Then I'm OK as long as the result is correct (no extra whitespace) and we gradually move to the flavor I proposed here:

```rust
impl Display for ShowTables {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "SHOW ")?;
        if self.full {
            write!(f, "FULL ")?;
        }
        write!(f, "TABLES ")?;
        if let Some(database) = &self.database {
            write!(f, "IN {database} ")?;
        }
        write!(f, "{}", &self.kind)
    }
}
```

tisonkun (commented):
I'll give this patch another round tomorrow, and I suggest you revise it once more with this comment https://github.com/GreptimeTeam/greptimedb/pull/3744/commits/b24d334036e43de5128962ef51b188f7a67d5f7d#r1576504197 in consideration.

src/sql/src/statements/statement.rs
```diff
@@ -89,6 +91,37 @@ pub enum Statement {
...
+            Statement::ShowCreateTable(s) => s.fmt(f),
+            Statement::DescribeTable(s) => s.fmt(f),
+            Statement::Explain(s) => s.inner.fmt(f),
+            Statement::Copy(s) => s.fmt(f),
+            Statement::Tql(s) => s.fmt(f),
+            Statement::TruncateTable(s) => s.fmt(f),
+            Statement::SetVariables(s) => s.fmt(f),
+            Statement::ShowVariables(s) => s.fmt(f),
+            Statement::ShowCharset(s) => s.fmt(f),
+            Statement::ShowCollation(s) => s.fmt(f),
```
tisonkun, line 120: For these two statements, only format the `ShowKind` seems incorrect. We should format like:

```
SHOW CHARSET <kind>;
SHOW COLLATION <kind>;
```

instead.

tisonkun: Do we have some tests cover these cases? It seems we should not have such issues if test coverage is enough.

src/sql/src/statements/statement.rs
```diff
@@ -89,6 +91,37 @@ pub enum Statement {
...
+            Statement::DropDatabase(s) => s.fmt(f),
+            Statement::CreateDatabase(s) => s.fmt(f),
+            Statement::Alter(s) => s.fmt(f),
+            Statement::ShowDatabases(s) => s.fmt(f),
+            Statement::ShowTables(s) => s.fmt(f),
+            Statement::ShowColumns(s) => s.fmt(f),
+            Statement::ShowIndex(s) => s.fmt(f),
+            Statement::ShowCreateTable(s) => s.fmt(f),
+            Statement::DescribeTable(s) => s.fmt(f),
+            Statement::Explain(s) => s.inner.fmt(f),
```
tisonkun, line 113:
```suggestion
            Statement::Explain(s) => s.fmt(f),
```

Explain implement Display already. Although it's `write!(f, "{}", self.inner)` inside so the same of this line physically, we'd better call the implementation on our Explain so that further changes underneath don't break this line.

src/sql/src/statements/show.rs
tisonkun: Seems only

* CreateExternalTable
* CreateTable

doesn't follow this pattern. But we can leave it to a follow up. This patch is already too large.

tisonkun:
I push a commit to resolve most of the comments above. Now the remaining part is:

* create.rs
* copy.rs

I'll give another round tomorrow.

src/sql/src/statements/statement.rs
tisonkun: Resolved.

tisonkun (approved):

tisonkun:
cc @killme2008 @MichaelScofield in the follow-up we'd change the SQL logging logics with these `Display` implementations:

```rust
                        Err(e) => {
                            let redacted = sql::util::redact_sql_secrets(query.as_ref());
                            if e.status_code().should_log_error() {
                                error!(e; "Failed to execute query: {redacted}");
                            } else {
                                debug!("Failed to execute query: {redacted}, {e}");
                            }

                            results.push(Err(e));
                            break;
                        }
```

to

```
                        Err(e) => {
                            if e.status_code().should_log_error() {
                                error!(e; "Failed to execute query: {stmt}");
                            } else {
                                debug!("Failed to execute query: {stmt}, {e}");
                            }

                            results.push(Err(e));
                            break;
```

see https://github.com/GreptimeTeam/greptimedb/issues/3646 for more details.

src/sql/src/statements/create.rs
```diff
@@ -166,36 +130,32 @@ impl Display for Partitions {
...
 impl Display for CreateTable {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        let if_not_exists = self.format_if_not_exists();
-        let name = &self.name;
-        let columns = format_list_indent!(self.columns);
-        let constraints = self.format_constraints();
-        let partitions = self.format_partitions();
-        let engine = &self.engine;
-        let options = self.format_options();
-        let maybe_external = if self.engine == FILE_ENGINE {
```
tisonkun, line 185: This should be a historical hack. Now `CREATE EXTERNAL` always construct a `CreateExternalTable` struct.

src/sql/src/statements/create.rs
```diff
@@ -166,36 +130,32 @@ impl Display for Partitions {
...
+        writeln!(f, "{},", format_list_indent!(self.columns))?;
+        writeln!(f, "{}", format_table_constraint(&self.constraints))?;
+        writeln!(f, ")")?;
+        if let Some(partitions) = &self.partitions {
+            writeln!(f, "{partitions}")?;
+        }
+        writeln!(f, "ENGINE={}", &self.engine)?;
+        if !self.options.is_empty() {
+            writeln!(f, "WITH (")?;
+            writeln!(f, "{}", format_list_indent!(self.options))?;
```
tisonkun, line 155: IIUC we can further make `pub options: Vec<SqlOption>,` to `pub options: OptionMap,` as in `CreateExternalTable`.

src/sql/src/statements/create.rs
tisonkun: No. We even have a test to guard this manner. So revert it. But I suppose here has some potential risks that we should validate the engine on parsing instead.

src/sql/src/statements.rs
```diff
@@ -58,6 +59,26 @@ use crate::error::{
     SerializeColumnDefaultConstraintSnafu, TimestampOverflowSnafu, UnsupportedDefaultValueSnafu,
 };
 
+const REDACTED_OPTIONS: [&str; 2] = ["access_key_id", "secret_access_key"];
+
+fn redact_and_sort_options(options: &OptionMap) -> Vec<String> {
```
killme2008, line 64: Please adds a description for this function.

src/sql/src/statements.rs
```diff
@@ -58,6 +59,26 @@ use crate::error::{
...
+    let mut result = vec![];
```
killme2008, line 65:
```suggestion
    let mut result = Vec::with_capacity(options.len());
```

src/sql/src/statements.rs
```diff
@@ -58,6 +59,26 @@ use crate::error::{
...
+    let keys = options.as_ref().keys().sorted();
+    for key in keys {
+        if let Some(val) = options.get(key) {
+            let redacted = REDACTED_OPTIONS
+                .iter()
+                .any(|opt| opt.eq_ignore_ascii_case(key));
+            if redacted {
+                result.push(format!("{key} = ******"));
```
killme2008, line 73:
```suggestion
                result.push(format!("{key} = '******'"));
```

src/sql/src/statements.rs
```diff
@@ -58,6 +59,26 @@ use crate::error::{
...
+            } else {
+                result.push(format!("{key} = {val}"));
```
killme2008, line 75: Looks better to add a quote for values:
```suggestion
                result.push(format!("{key} = '{val}'"));
```

src/sql/src/statements/describe.rs
```diff
@@ -32,6 +34,13 @@ impl DescribeTable {
     }
 }
 
+impl Display for DescribeTable {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let name = self.name();
+        write!(f, r#"DESCRIBE TABLE {name}"#)
```
killme2008, line 40: I think it's better to always quote the database names, table names and column names with `"`.

src/sql/src/statements/tql.rs
```diff
@@ -55,6 +97,23 @@ pub struct TqlAnalyze {
...
+        if self.is_verbose {
+            write!(f, "VERBOSE ")?;
+        }
+        write!(f, "(")?;
+        write!(f, "{}, ", &self.start)?;
+        write!(f, "{}, ", &self.end)?;
+        write!(f, "{}", &self.step)?;
+        if let Some(lookback) = &self.lookback {
+            write!(f, ", {}", lookback)?;
+        }
```
killme2008, line 112: Looks like this code block appears three times in this file, could we extract it with a macro?

killme2008 (commented):
Fantastic effort and dedication! Many thanks! I've included some suggestions for your consideration. Please review them. @irenjj

src/sql/src/statements.rs
```diff
@@ -58,6 +59,26 @@ use crate::error::{
     SerializeColumnDefaultConstraintSnafu, TimestampOverflowSnafu, UnsupportedDefaultValueSnafu,
 };
 
+const REDACTED_OPTIONS: [&str; 2] = ["access_key_id", "secret_access_key"];
+
+fn redact_and_sort_options(options: &OptionMap) -> Vec<String> {
+    let mut result = vec![];
+    let keys = options.as_ref().keys().sorted();
```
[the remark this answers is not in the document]
tisonkun: 1. It follows the previous implementation for `CreateTable`.
2. I suppose that it helps in test stability. That is, when we check the format result, it's stable.

src/sql/src/statements/copy.rs
```diff
@@ -12,29 +12,90 @@
...
+                write!(f, "{} TO {}", &args.table_name, &args.location)?;
+                (&args.with, &args.connection)
+            }
+            CopyTable::From(args) => {
+                write!(f, "{} FROM {}", &args.table_name, &args.location)?;
+                (&args.with, &args.connection)
+            }
+        };
+        if !with.map.is_empty() {
+            let options = redact_and_sort_options(with);
```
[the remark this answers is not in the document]
tisonkun: This is good for a follow-up :D

src/sql/src/statements/describe.rs
tisonkun: The underneath Ident impls:

```rust
impl fmt::Display for Ident {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        match self.quote_style {
            Some(q) if q == '"' || q == '\'' || q == '`' => {
                let escaped = value::escape_quoted_string(&self.value, q);
                write!(f, "{q}{escaped}{q}")
            }
            Some('[') => write!(f, "[{}]", self.value),
            None => f.write_str(&self.value),
            _ => panic!("unexpected quote style"),
        }
    }
}
```

src/sql/src/statements/copy.rs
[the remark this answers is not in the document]
MichaelScofield: Updated the issue https://github.com/GreptimeTeam/greptimedb/issues/2144

killme2008 (approved):
LGTM

PR #3701, opened 2024-04-14
feat: Modify Column Type

KKould:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)
https://github.com/GreptimeTeam/greptimedb/issues/3517
https://github.com/GreptimeTeam/greptime-proto/pull/146

## What's changed and what's your intention?

Support: `ALTER TABLE <table_name> ALTER COLUMN <column_name> TYPE <data_type>;`
ref: https://www.postgresql.org/docs/current/sql-altertable.html

Tips 1: sql parser does not support `Modify Column` temporarily, I have created a PR: https://github.com/sqlparser-rs/sqlparser-rs/pull/1216

Tips 2: I temporarily use my own fork of `greptime-proto`, and then replace it after merging it at https://github.com/GreptimeTeam/greptime-proto/pull/146

```shell
public=> CREATE TABLE test(i INTEGER, j TIMESTAMP TIME INDEX);
OK 0
public=> ALTER TABLE test ALTER COLUMN I TYPE STRING;
OK 0
public=> DESCRIBE t2;
+--------+----------------------+-----+------+---------+---------------+
| Column | Type                 | Key | Null | Default | Semantic Type |
+--------+----------------------+-----+------+---------+---------------+
| i      | String               |     | YES  |         | FIELD         |
| j      | TimestampMillisecond | PRI | NO   |         | TIMESTAMP     |
+--------+----------------------+-----+------+---------+---------------+
```

## Checklist

- [x]  I have written the necessary rustdoc comments.
- [x]  I have added the necessary unit tests and integration tests.
- [x]  This PR does not require documentation updates.

Issue #3517, opened 2024-03-14
Support altering column's data type


MichaelScofield:
### What problem does the new feature solve?

Sometimes user just want to modify the column's data type.

### What does the feature do?

Make column's data type changeable via "alter column" stmt. For example, in mysql:
```sql
ALTER TABLE t1 MODIFY b INT
```
(see https://dev.mysql.com/doc/refman/8.0/en/alter-table.html). Or in postgresql:
```sql
ALTER TABLE t1 ALTER COLUMN b TYPE varchar(80)
```
(see https://www.postgresql.org/docs/current/sql-altertable.html).

We might want to go with mysql's "alter column" syntax for the conciseness.

One critical consideration is how to deal with un-convertable data. For example, when you modify the column from string to int, how to deal with the un-convertable data of "a"? In mysql, this results in an error:
```text
mysql> select * from foo;
+------+
| s    |
+------+
| a    |
| 1    |
+------+
2 rows in set (0.01 sec)

mysql> alter table foo modify s int;
ERROR 1366 (HY000): Incorrect integer value: 'a' for column 's' at row 1
mysql> show create table foo;
+-------+----------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                       |
+-------+----------------------------------------------------------------------------------------------------+
| foo   | CREATE TABLE `foo` (
  `s` text
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci |
+-------+----------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```
In postgresql, it's error, too:
```text
postgres=# select * from foo;
 s
---
 a
 1
(2 rows)

postgres=# ALTER TABLE foo ALTER COLUMN s TYPE int;
ERROR:  column "s" cannot be cast automatically to type integer
HINT:  You might need to specify "USING s::integer".
postgres=# ALTER TABLE foo ALTER COLUMN s TYPE int using s::integer;
ERROR:  invalid input syntax for type integer: "a"
postgres=# \d foo
               Table "public.foo"
 Column | Type | Collation | Nullable | Default
--------+------+-----------+----------+---------
 s      | text |           |          |
```
(Interestingly, postgresql has a "hint" to let user specify the data convention method. We might want to do the same.)

So it's clear that we'd better follow their steps and return some errors complaining some data are just not convertable to the target type.

### Implementation challenges

[...]

PR #3745, opened 2024-04-19
feat: support different types for `CompatReader`

KKould:

https://github.com/GreptimeTeam/greptimedb/issues/3517

Support CompatReader to be compatible with different types of reads (for modify column type)

Whether the type allows conversion will be verified in the next pr: (reference: 
`ModifyColumn::validate` https://github.com/GreptimeTeam/greptimedb/pull/3701/files#diff-839e23ed322c90eacc42fdddeb1f7108335e6434eba0a7a1c6dbc42052a8b9e9)

Tips: Please closely [64FD8R166](https://github.com/GreptimeTeam/greptimedb/pull/3745/files#diff-f6ecbd6cf3bbc3d92c1993aeed2c4af70186b8014af7a3a14798ac52a1664fd8R166) in the `cast` in this line of code, the `safe` of the `CastOptions` it used by default is true, so it is guaranteed that the type is allowed to be transformed and specific When the value cannot be converted, set the value to null, e.g. type: `String` -> `Integer`, value: `"1"` -> `1`, `"hello"` -> `null`

## Checklist


src/mito2/src/read/compat.rs
```diff
@@ -85,7 +86,7 @@ pub(crate) fn has_same_columns(left: &RegionMetadata, right: &RegionMetadata) ->
     }
 
     for (left_col, right_col) in left.column_metadatas.iter().zip(&right.column_metadatas) {
-        if left_col.column_id != right_col.column_id {
+        if left_col != right_col {
```
waynexia, line 89: Only comparing what you need can be a little faster. And the `ColumnMetadata` contains much information we don't care about here like name, metadata, semantic type etc. Can you make a util method to only compare whether we need cast types on read?

KKould:
> We can also set up a simple rule.
> - fill NULL if the column is nullable
> - fill default if the column is not null

https://github.com/GreptimeTeam/greptimedb/issues/3517#issuecomment-2063190801

`cast` can only be converted to null but cannot set default. Do you have any suggestions on this?

src/mito2/src/read/compat.rs
```diff
@@ -239,9 +254,10 @@ fn may_compat_fields(
     mapper: &ProjectionMapper,
     actual: &RegionMetadata,
 ) -> Result<Option<CompatFields>> {
+    let expect = mapper.metadata();
     let expect_fields = mapper.batch_fields();
     let actual_fields = Batch::projected_fields(actual, mapper.column_ids());
-    if expect_fields == actual_fields {
+    if expect.schema == actual.schema {
```
evenyag, line 260: If we only read a subset of fields (projection) then this check will also return false. We can returns the column id and data type in `batch_fields()` and `projected_fields()` above.

src/mito2/src/read/compat.rs
```diff
@@ -156,7 +157,21 @@ impl CompatFields {
...
             .iter()
             .map(|index_or_default| match index_or_default {
-                IndexOrDefault::Index(index) => batch.fields()[*index].clone(),
+                IndexOrDefault::Index { pos, cast_type } => {
+                    let old_column = &batch.fields()[*pos];
+
+                    let data = if let Some(ty) = cast_type {
+                        // Safety: We ensure type can be converted and the new batch should be valid.
+                        // Tips: `safe` must be true in `CastOptions`, which will replace the specific value with null when it cannot be converted.
+                        old_column.data.cast(ty).unwrap()
```
evenyag, line 166: We need test cases to ensure updating arrow won't break this. We could test some behavior in `VectorOp::cast()`'s test cases.

src/mito2/src/read/compat.rs
```diff
@@ -255,8 +271,21 @@ fn may_compat_fields(
         .iter()
         .map(|column_id| {
             if let Some(index) = source_field_index.get(column_id) {
+                // TODO(kould): more Modify case, e.g. column name
```
evenyag, line 274: We use column id internally so modifying the column name should not affect this.

src/mito2/src/read/compat.rs
```diff
@@ -255,8 +271,21 @@ fn may_compat_fields(
...
+                let mut modify_type = None;
+
+                // Safety: check on [`TableMeta::modify_columns`].
+                let expect_column = expect.column_by_id(*column_id).unwrap();
+                let actual_column = actual.column_by_id(*column_id).unwrap();
+
+                if expect_column != actual_column {
+                    modify_type = Some(expect_column.column_schema.data_type.clone())
+                }
```
evenyag, line 283: Compare the whole metadata here may introduce the same issue mentioned in https://github.com/GreptimeTeam/greptimedb/pull/3745#discussion_r1572049239

We can reuse the column type if `batch_fields()` and `projected_fields()` can return it. Then we don't need to call `column_by_id` to acquire the metadata again.

evenyag:
> `cast` can only be converted to null but cannot set default. Do you have any suggestions on this?

Maybe we could only allow modifying the data type when the column is nullable or casting the type won't be fallible.

evenyag (commented):
Sorry for the slow review. Don't have enough time to review this last week.

KKould:
> > `cast` can only be converted to null but cannot set default. Do you have any suggestions on this?
> 
> Maybe we could only allow modifying the data type when the column is nullable or casting the type won't be fallible.

I added this check in pr: https://github.com/GreptimeTeam/greptimedb/pull/3757 https://github.com/GreptimeTeam/greptimedb/pull/3757/commits/9003afc9876665dfb71f2eb1e482b4cda7579480

src/mito2/src/read/compat.rs
```diff
@@ -248,15 +263,23 @@ fn may_compat_fields(
...
     let index_or_defaults = expect_fields
         .iter()
-        .map(|column_id| {
-            if let Some(index) = source_field_index.get(column_id) {
+        .map(|(column_id, expect_data_type)| {
+            if let Some((index, actual_data_type)) = source_field_index.get(column_id) {
+                let mut modify_type = None;
+
+                if expect_data_type != *actual_data_type {
+                    modify_type = Some(expect_data_type.clone())
```
evenyag, line 276:
```suggestion
                let mut cast_type = None;

                if expect_data_type != *actual_data_type {
                    cast_type = Some(expect_data_type.clone())
```

src/mito2/src/read/compat.rs
```diff
@@ -248,15 +263,23 @@ fn may_compat_fields(
...
+                }
                 // Source has this field.
-                Ok(IndexOrDefault::Index(*index))
+                Ok(IndexOrDefault::Index {
+                    pos: *index,
+                    cast_type: modify_type,
```
evenyag, line 281:
```suggestion
                    cast_type,
```

evenyag (commented):
Mostly LGTM

evenyag (approved):
LGTM

v0y4g3r (approved):
LGTM!

PR #3723, opened 2024-04-16
build(deps): Update RustPython to 0.3.1

youknowone:
#3063 draft

The code part looks done, but I believe @discord9 wants to change the dependency to the GC version.

PR #3749, opened 2024-04-19
refactor!: drop RustPython support

### LINKED ISSUE #3063: Upgrade rustpython

Tech debt reduction

Our rustpython version looks too old, maybe we need to upgrade it.

### Implementation challenges

_No response_

tisonkun:

This supersedes and thus closes https://github.com/GreptimeTeam/greptimedb/pull/3723.

This closes #3063.

Drop RustPython backend for scripting and always build with PyO3. This is an attempt to simplify our scripting functions.

This should involve a lot of docs changes, but we haven't decided yet. Open this patch as a preview.

## Checklist


.github/workflows/develop.yml
```diff
@@ -149,6 +149,10 @@ jobs:
         run: |
           sudo apt update && sudo apt install -y libfuzzer-14-dev
           cargo install cargo-fuzz
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.8'
```
tisonkun, line 154: @davidhewitt @messense

We are trying to ship the Python script function with PyO3 primarily here. And since statically link to libpython can be complicated, I'm investigating the way to dynamically link to libpython.

So far, it works well and we can see the CI all passed. But this requires a `libpython3.8.so` can be linked in the system.

Thus, I have two questions:

1. How can I make this binary work well with all the libpython >= 3.8?
2. How can I disable the related functionalities if libpython3.so is not found?

tisonkun: cc @discord9 any thoughts here?

discord9: > cc @discord9 any thoughts here?

Maybe we can use `bindgen`  to create a shared library(call it `libpython_finder` maybe?) with the same symbols as limited python abi, and forward those calls to actual `libpython3.*` at runtime with `libloading`, we would also need to export extra functions to check if `libpython` is found, and config pyo3 to use our shared library, the shared library would not need much effect to mantain after we are done developing it as limited api doesn't change that much.

discord9 (approved):
Rest LGTM

tisonkun:
Closed as undetermined yet. This patch can be picked up since it touches only the Python script code path.

Before we make the decision to lift PyO3 as the default backend and distribute it in every artifact, we should resolve the issue described in https://github.com/GreptimeTeam/greptimedb/issues/3777, namely how to support multiple Python versions, or at least disable when Python environment not found, instead of panic.

davidhewitt:
Sorry for the late reply, I am somewhat behind on OSS so please expect some delay from me at the moment.

> Thus, I have two questions:
> 
> 1. How can I make this binary work well with all the libpython >= 3.8?
> 2. How can I disable the related functionalities if libpython3.so is not found?

In principle you can link to `libpython3.so` of any version of 3.8 or higher using the `abi3-py38` feature. But my previous experience is that distribution packagers are frequently omitting this file - e.g. https://github.com/PyO3/pyo3/issues/2901#issuecomment-1400205303

Probably you are safer to avoid depending on the system Python and distribute the Python version you want to link with. This means either you:
- statically link (the easiest way I know to do this is with `PyOxidizer`)
- dynamically link and ship the linked libpython along with your binary. I believe you can use a `rpath` to ensure that the linked libpython is preferred for loading (but this may affect your other dependencies too).

tisonkun:
@davidhewitt Thanks a lot! According to your comment, here are my two questions:

> libpython3.so of any version of 3.8 or higher using the abi3-py38 feature

How does it work? We currently add the `abi3-py37` feature, but it still failed to load shared lib as in https://github.com/GreptimeTeam/greptimedb/actions/runs/8751166135/job/24016840142:

```
./bins/greptime: error while loading shared libraries: libpython3.8.so.1.0: cannot open shared object file: No such file or directory
```

You can see the commits above to fight with this issue.

> dynamically link and ship the linked libpython along with your binary. I believe you can use a rpath to ensure that the linked libpython is preferred for loading (but this may affect your other dependencies too).

It still has two challenges:

1. Is it possible to opt-out such Python scripting support on missing dylib? Currently, it panics on binary starts up.
2. Either static linking or dynamic linking with bundled dylib, we may meet challenge to ship 3rdparty dependencies like numpy. So it's still desired to leverage the Python installation on user environment. But GreptimeDB wants a way to be compatible with as many as Python versions and gracefully disable the related features if non dylib found.

PR #3765, opened 2024-04-22
ci: secrets.GITHUB_TOKEN not exist on self-hosted runner

tisonkun:

```
Error: C:\a\greptimedb\greptimedb\./.github/actions/build-windows-artifacts\action.yml (Line: 30, Col: 21): Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.GITHUB_TOKEN
```

This default secrets doesn't exist on self-hosted runner. The TOKEN is mainly for avoid rate limitation, but that may not be quite an issue for self-hosted daily cronjob. Remove firstly to unlock the workflow. We can revisit this topic if the rate limit comes back to be an issue.

## Checklist

### DIFF
--- a/.github/actions/build-windows-artifacts/action.yml
+++ b/.github/actions/build-windows-artifacts/action.yml
@@ -26,8 +26,6 @@ runs:
   using: composite
   steps:
     - uses: arduino/setup-protoc@v3
-      with:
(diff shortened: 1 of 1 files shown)


killme2008 (approved):
LGTM

evenyag (approved):

PR #3769, opened 2024-04-22
feat(flow): render reduce

discord9:

add `render_reduce` to render module, which render reduce operator into executable dataflow.
Also added related tests for each case

## Checklist

### DIFF
--- a/src/flow/src/compute/render/reduce.rs
+++ b/src/flow/src/compute/render/reduce.rs
@@ -1,0 +1,957 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::BTreeMap;
+use std::ops::Range;
+
+use datatypes::data_type::ConcreteDataType;
+use datatypes::value::{ListValue, Value};
+use hydroflow::scheduled::graph_ext::GraphExt;
+use hydroflow::scheduled::port::{PortCtx, SEND};
+use itertools::Itertools;
+use snafu::{ensure, OptionExt, ResultExt};
+
+use crate::adapter::error::{Error, PlanSnafu};
+use crate::compute::render::{Context, SubgraphArg};
+use crate::compute::state::Scheduler;
+use crate::compute::types::{Arranged, Collection, CollectionBundle, ErrCollector, Toff};
+use crate::expr::error::{DataTypeSnafu, InternalSnafu};
+use crate::expr::{AggregateExpr, EvalError, ScalarExpr};
+use crate::plan::{AccumulablePlan, AggrWithIndex, KeyValPlan, Plan, ReducePlan};
+use crate::repr::{self, DiffRow, KeyValDiffRow, Row};
+use crate::utils::{ArrangeHandler, ArrangeReader, ArrangeWriter};
+
+impl<'referred, 'df> Context<'referred, 'df> {
+    /// render `Plan::Reduce` into executable dataflow
+    // There is a false positive in using `Vec<ScalarExpr>` as key due to `Value` have `bytes` variant
+    #[allow(clippy::mutable_key_type)]
+    pub fn render_reduce(
+        &mut self,
+        input: Box<Plan>,
+        key_val_plan: KeyValPlan,
+        reduce_plan: ReducePlan,
+    ) -> Result<CollectionBundle, Error> {
+        let input = self.render_plan(*input)?;
+        // first assembly key&val that's ((Row, Row), tick, diff)
+        // Then stream kvs through a reduce operator
+
+        // the output is concat from key and val
+        let output_key_arity = key_val_plan.key_plan.output_arity();
+
+        // TODO: config global expire time from self
+        let arrange_handler = self.compute_state.new_arrange(None);
+        // reduce need full arrangement to be able to query all keys
+        let arrange_handler_inner =
+            arrange_handler
+                .clone_full_arrange()
+                .with_context(|| PlanSnafu {
+                    reason: "No write is expected at this point",
+                })?;
+
+        let distinct_input = self.add_accum_distinct_input_arrange(&reduce_plan);
+
+        let reduce_arrange = ReduceArrange {
+            output_arrange: arrange_handler_inner,
+            distinct_input,
+        };
+
+        let now = self.compute_state.current_time_ref();
+
+        let err_collector = self.err_collector.clone();
+
+        // TODO(discord9): better way to schedule future run
+        let scheduler = self.compute_state.get_scheduler();
+        let scheduler_inner = scheduler.clone();
+
+        let (out_send_port, out_recv_port) = self.df.make_edge::<_, Toff>("reduce");
+
+        let subgraph = self.df.add_subgraph_in_out(
+            "reduce",
+            input.collection.into_inner(),
+            out_send_port,
+            move |_ctx, recv, send| {
+                // mfp only need to passively receive updates from recvs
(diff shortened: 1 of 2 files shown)


killme2008:
@zhongzc @waynexia Please take a look.

src/flow/src/compute/render/reduce.rs
```diff
@@ -0,0 +1,957 @@
...
+        let accums = accums.inner;
+
+        // deser accums from offsets
+        let accum_ranges = {
+            let res = err_collector
+                .run(|| from_val_to_slice_idx(accums.first().cloned(), full_aggrs.len()));
+            if let Some(res) = res {
+                res
+            } else {
+                // ignore for whatever reason
```
killme2008, line 391: I'm not certain whether we need to add debug logs for this specific corner case. It seems like it could be challenging to troubleshoot runtime flow tasks if we encounter issues. Additionally, it appears that the flow crate lacks sufficient logs.

src/flow/src/compute/render/reduce.rs
```diff
@@ -0,0 +1,957 @@
...
+}
+
+#[derive(Debug)]
+struct AccumOutput {
+    accum: BTreeMap<usize, Vec<Value>>,
+    output: BTreeMap<usize, Value>,
+}
+
+impl AccumOutput {
+    fn new() -> Self {
```
killme2008, line 491: Looks like missing a construct such as `with_capacity(uszie)` when we known the exact capacity.

src/flow/src/compute/render/reduce.rs
```diff
@@ -0,0 +1,957 @@
...
+                })?;
+
+        let distinct_input = self.add_accum_distinct_input_arrange(&reduce_plan);
+
+        let reduce_arrange = ReduceArrange {
+            output_arrange: arrange_handler_inner,
+            distinct_input,
+        };
+
+        let now = self.compute_state.current_time_ref();
```
killme2008, line 69: Is the time or tick is monotonic?

src/flow/src/compute/render/reduce.rs
```diff
@@ -0,0 +1,957 @@
...
+
+        let err_collector = self.err_collector.clone();
+
+        // TODO(discord9): better way to schedule future run
+        let scheduler = self.compute_state.get_scheduler();
+        let scheduler_inner = scheduler.clone();
+
+        let (out_send_port, out_recv_port) = self.df.make_edge::<_, Toff>("reduce");
```
killme2008, line 77: Maybe extract a constant `const REDUCE: &str = "reduce"; `  for the edge.

src/flow/src/compute/render/reduce.rs
```diff
@@ -0,0 +1,957 @@
...
+    SubgraphArg {
+        now,
+        err_collector,
+        scheduler,
+        send,
+    }: SubgraphArg,
+) {
+    let mut row_buf = Row::empty();
+    let key_val = data.into_iter().filter_map(|(row, sys_time, diff)| {
+        // error is collected and then ignored
```
killme2008, line 198: Could ignoring the errors make it difficult to address certain runtime issues?

src/flow/src/compute/render/reduce.rs
```diff
@@ -0,0 +1,957 @@
...
+        // it is assumed that value is in order of insertion
+        let vals = key2vals.entry(key).or_default();
+        vals.push((val, diff));
+    }
+
+    let mut all_updates = Vec::with_capacity(key2vals.len());
+    let mut all_outputs = Vec::with_capacity(key2vals.len());
+
+    // lock the arrange for write
+    let mut arrange = arrange.write();
```
killme2008, line 367: The lock scope includes the remaining function body, is it expected?

killme2008 (commented):
I'm not entirely clear on the logic, but I do have a few minor comments for your consideration. Thank you. Overall, it seems impressive to me. @discord9

src/flow/src/compute/render/reduce.rs
```diff
@@ -0,0 +1,957 @@
...
+        // the output is concat from key and val
+        let output_key_arity = key_val_plan.key_plan.output_arity();
+
+        // TODO: config global expire time from self
+        let arrange_handler = self.compute_state.new_arrange(None);
+        // reduce need full arrangement to be able to query all keys
+        let arrange_handler_inner =
+            arrange_handler
+                .clone_full_arrange()
+                .with_context(|| PlanSnafu {
```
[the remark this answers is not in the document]
discord9: also changed a few unnecessary `with_context` to `context`

src/flow/src/compute/render/reduce.rs
discord9: > I'm not certain whether we need to add debug logs for this specific corner case. It seems like it could be challenging to troubleshoot runtime flow tasks if we encounter issues. Additionally, it appears that the flow crate lacks sufficient logs.

All errors generated in flow's eval process is collected in `err_collector`, and the logging of `err_collector` should be sufficient?(although not yet impl), as err_collector is now per-dataflow, might add per-subgraph location later

src/flow/src/compute/render/reduce.rs
discord9: > Is the time or tick is monotonic?

This is system time(or processing time) which is assumed to be montonic, contrast to event time(which can be arbitrary)

src/flow/src/compute/render/reduce.rs
killme2008: Got it. That is enough.

src/flow/src/compute/render/reduce.rs
discord9: > Could ignoring the errors make it difficult to address certain runtime issues?

All `EvalError` are collected in `err_collector` and should be send to be logging, but the impl of logging it is still WIP

src/flow/src/compute/render/reduce.rs
discord9: > The lock scope includes the remaining function body, is it expected?

Yes, it's there to prevent potential race condition, might consider use key level locl later

src/flow/src/compute/render/reduce.rs
discord9: > Looks like missing a construct such as `with_capacity(uszie)` when we known the exact capacity.

Since BTreeMap lack a `with_capacity` method(see https://users.rust-lang.org/t/btreemap-with-capacity/39770/3) I'm not sure it would be necessary, although  maybe I should use `Vec<Option<>>` instead...

killme2008 (approved):
LGTM from my perspective.

src/flow/src/compute/render/reduce.rs
```diff
@@ -0,0 +1,978 @@
...
+    send.give(ret);
+}
+
+/// eval accumulable reduce plan by eval aggregate function and reduce the result
+///
+/// TODO(discord9): eval distinct by adding distinct input arrangement
+///
+/// invariant: it'is assumed `kv`'s time is always <= now,
+/// since it's from a Collection Bundle, where future inserts are stored in arrange
+fn reduce_accum_subgraph(
```
zhongzc, line 350: Can you describe what the row of result looks like here?

src/flow/src/compute/render/reduce.rs
```diff
@@ -0,0 +1,1057 @@
...
+            // split value_diffs into columns
+            (0..row_len)
+                .map(|i| {
+                    value_diffs
+                        .iter()
+                        .map(|(row, diff)| (row.get(i).cloned().unwrap(), *diff))
+                        .collect_vec()
+                })
+                .collect_vec()
+        };
```
zhongzc, line 398: I think we can extract this section into a function and then make some necessary checks (ensure equal number of columns in all rows)

src/flow/src/compute/render/reduce.rs
```diff
@@ -247,28 +243,26 @@ fn eval_distinct_core(
...
-                    None
-                } else {
-                    // does not exist, insert it and emit
-                    inner_map.insert(k.clone(), (v.clone(), t, d));
-                    Some(((k, v), t, d))
-                }
-            });
+            if let Some(((k, v), t, d)) = new_key_val.clone() {
+                // update the inner_map, so later upadtes can be checked against it
+                inner_map.insert(k.clone(), (v.clone(), t, d));
```
zhongzc, line 264:
```suggestion
                inner_map.insert(k, (v, t, d));
```

zhongzc (approved):
Generally LGTM

PR #3771, opened 2024-04-22
fix: the `dropping_regions` guards should be dropped on procedure done

MichaelScofield:

The [failed sqlness test](https://github.com/GreptimeTeam/greptimedb-enterprise/actions/runs/8778950389/job/24086386812?pr=43) shows that there might be some data race on dropping tables: 

image

After digging into the codes, I think the problem is caused by the implementation that the procedure returned does not mean it's dropped. The logs suggest the same:

image

This PR fix it.

## Checklist

### DIFF
--- a/src/common/meta/src/ddl/drop_table.rs
+++ b/src/common/meta/src/ddl/drop_table.rs
@@ -46,7 +46,7 @@ pub struct DropTableProcedure {
     /// The serializable data.
     pub data: DropTableData,
     /// The guards of opening regions.
-    pub dropping_regions: Vec<OperatingRegionGuard>,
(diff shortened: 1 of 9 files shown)


src/common/meta/src/ddl/drop_table.rs
```diff
@@ -163,6 +163,8 @@ impl DropTableProcedure {
         self.executor
             .on_delete_metadata_tombstone(&self.context, table_route_value)
             .await?;
+
+        self.dropping_regions.clear();
```
MichaelScofield, line 167: core change

WenyXu (approved):

fengjiachun (approved):
LGTM

PR #3772, opened 2024-04-22
feat: add validate method to CreateExpr

### LINKED ISSUE #3759: Unexpected corrupted table info
> BTW, I found the database would crash once I create a metric table with 2 columns with the same name:
> 
> ```sql
> mysql> CREATE TABLE `phy`(ts timestamp time index, val double) engine=metric with ("physical_metric_table" = "");
> 
> mysql> CREATE TABLE t1(ts timestamp time index, val double, host text, host string) engine=metric with ("on_physical_tab
> le" = "phy");
> ERROR 1815 (HY000): Internal error: 1003
> 
> mysql> show tables;
> ERROR 1815 (HY000): Internal error: 1003
> ```
> 
> Is it the intended behavior?

Good catch! It's a bug🥲

_Originally posted by @WenyXu in https://github.com/GreptimeTeam/greptimedb/issues/3741#issuecomment-2067610661_

### LINKED ISSUE #3760: Validation needs to be added while creating the logical table
> > Sorry, I missed something. Could you provide an example to explain the difference?
> 
> As for the physical table, here I add some constraints for val:
> 
> ```sql
> CREATE TABLE phy (ts timestamp time index, val double not null default 0) engine=metric with ("physical_metric_table" = ""); 
> ```
> 
> Currently we can create a logical table with the `val` column but without constraints (or add more constraints), and also change the data type of it:
> 
> ```sql
> CREATE TABLE phy (ts timestamp time index, val int, host string, primary key (host)) engine=metric with ("on_physical_table" = "phy"); 
> ```
> 
> I wonder if its appropriate to create sql like this for fuzz test.

I got you. It's a bug; it will throw an error if users try to insert values into a logical table. Some validation needs to be added while creating the logical table(It will be tracked in another issue). 

Reproduce steps:
1. Create tables
```sql
CREATE TABLE `phy`(ts timestamp time index, val double) engine=metric with ("physical_metric_table" = "");

CREATE TABLE t1 (ts timestamp time index, val int, host string, primary key (host)) engine=metric with ("on_physical_table" = "phy");

CREATE TABLE t2 (ts timestamp time index, val string, host string, primary key (host)) engine=metric with ("on_physical_table" = "phy");
```
4. Insert values
```
mysql> insert t2 (host,ts,val) values ("hi",1702433146000,"10");
ERROR 1815 (HY000): Invalid request to region 4398046511104(1024, 0), reason: column val expect type Float64(Float64Type), given: STRING(12)

mysql> insert t1 (host,ts,val) values ("hi",1702433146000,1);
ERROR 1815 (HY000): Invalid request to region 4398046511104(1024, 0), reason: column val expect type Float64(Float64Type), given: INT32(3)
```

_Originally posted by @WenyXu in https://github.com/GreptimeTeam/greptimedb/issues/3741#issuecomment-2067616844_

waynexia:

Closes #3759
Closes #3760

Add various validation to `CreateExpr` and `RegionCreateRequest`.

This patch enforces the constraint that both logical and physical tables can only have one field column. We are going to support multi-field later, but before it's fully implemented we'll limit this behavior for correctness.

## Checklist


waynexia:
Going to fix https://github.com/GreptimeTeam/greptimedb/issues/3760 in this PR

src/metric-engine/src/engine/create.rs
```diff
@@ -323,6 +335,39 @@ impl MetricEngineInner {
             ConflictRegionOptionSnafu {}
         );
 
+        // check if only one field column is declared, and all tag columns are string
+        let mut field_col: Option<&ColumnMetadata> = None;
+        for col in &request.column_metadatas {
+            if col.semantic_type == SemanticType::Field {
```
killme2008, line 341: What about refactoring this `if else` by `match` block?

src/metric-engine/src/engine/create.rs
```diff
@@ -323,6 +335,39 @@ impl MetricEngineInner {
...
+                    ColumnTypeMismatchSnafu {
+                        expect: ConcreteDataType::string_datatype(),
+                        actual: col.column_schema.data_type.clone(),
+                    }
+                );
+            }
+        }
+        let field_col = field_col.context(NoFieldColumnSnafu)?;
+
+        // make sure the field column is float64 type
```
killme2008, line 362: Looks like we don't need to add this restriction. The all numeric type must be allowed.

src/metric-engine/src/error.rs
```diff
@@ -182,7 +196,10 @@ impl ErrorExt for Error {
             | MissingRegionOption { .. }
             | ConflictRegionOption { .. }
             | ColumnTypeMismatch { .. }
-            | PhysicalRegionBusy { .. } => StatusCode::InvalidArguments,
+            | PhysicalRegionBusy { .. }
```
killme2008, line 199: The status code of `PhysicalRegionBusy` must be `RegionBusy`.

https://greptimedb.rs/common_error/status_code/enum.StatusCode.html#variant.RegionBusy

src/operator/src/expr_factory.rs
```diff
@@ -214,9 +214,72 @@ pub fn create_to_expr(create: &CreateTable, query_ctx: QueryContextRef) -> Resul
...
     };
+
+    validate_create_expr(&expr)?;
     Ok(expr)
 }
 
+/// Validate the [`CreateTableExpr`] request.
+pub fn validate_create_expr(create: &CreateTableExpr) -> Result<()> {
+    // construct column list
+    let mut column_to_indices = HashMap::new();
```
killme2008, line 225:
```suggestion
    let mut column_to_indices = HashMap::new(create.column_defs.len());
```

tests/cases/standalone/common/create/create_metric_table.sql
```diff
@@ -38,18 +50,10 @@ DROP TABLE phy2;
...
-  `AutEM` INT,
-  esse DOUBLE,
-  `Tempore` BOOLEAN,
-  `reruM` BOOLEAN,
-  `eRrOR` BOOLEAN NULL,
-  `cOMmodi` BOOLEAN,
-  `PERfERENdIS` DOUBLE,
-  `eSt` FLOAT DEFAULT 0.70978713,
-  PRIMARY KEY(`cOMmodi`, `PERfERENdIS`, esse)
+  `REPREHenDERIt` double DEFAULT 0.70978713,
```
killme2008, line 53: What's the reason to change this fuzz test?

killme2008:
I created an issue to track the limiations on metric engine and PromQL https://github.com/GreptimeTeam/greptimedb/issues/3780

src/metric-engine/src/error.rs
waynexia: "busy" here means there are still some logical regions using it. Like dropping a physical region when its logical regions are not dropped first. Hence invalid argument

src/metric-engine/src/engine/create.rs
waynexia: The key point here is every logical table's field column should have the same type.

WenyXu (approved):

PR #3773, opened 2024-04-22
fix: do not remove deletion markers when window time range overlaps

v0y4g3r:

This PR disables deletion marker removal when some window's time range overlaps with another window, so that deletion markers will not be removed, which may cause deleted rows appear again.

## Checklist

### DIFF
--- a/src/mito2/src/compaction/twcs.rs
+++ b/src/mito2/src/compaction/twcs.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, HashMap};
 use std::fmt::{Debug, Formatter};
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


src/mito2/src/compaction/twcs.rs
```diff
@@ -195,24 +201,88 @@ impl Picker for TwcsPicker {
...
+    }
+
+    let mut windows = windows.into_values().collect::<Vec<_>>();
+    windows.sort_unstable_by(|l, r| l.start.cmp(&r.start).then(l.end.cmp(&r.end).reverse()));
+
+    let mut current_range: Option<(Timestamp, Timestamp)> =
+        Some((windows[0].start.unwrap(), windows[0].end.unwrap()));
+
+    for idx in 1..windows.len() {
+        match current_range {
```
evenyag, line 260: Looks like that current_range won't be None

src/mito2/src/compaction/twcs.rs
```diff
@@ -195,24 +201,88 @@ impl Picker for TwcsPicker {
...
 }
 
+#[derive(Default)]
+struct Window {
+    start: Option<Timestamp>,
+    end: Option<Timestamp>,
+    files: Vec<FileHandle>,
+    time_window: i64,
+    overlapping: bool,
+}
```
evenyag, line 211: nitpick: We can define a constructor `Window::new(file: FileHandle)` and remove `default()` so files are not empty
```suggestion
struct Window {
    start: Timestamp,
    end: Timestamp,
    files: Vec<FileHandle>,
    time_window: i64,
    overlapping: bool,
}
```

src/mito2/src/compaction/twcs.rs
```diff
@@ -195,24 +201,88 @@ impl Picker for TwcsPicker {
...
+                current_range = Some(windows[idx - 1].range());
+                continue;
+            }
+            Some(current) => {
+                let next_range = windows[idx].range();
+                if overlaps(&current, &next_range) {
+                    windows[idx - 1].overlapping = true;
+                    windows[idx].overlapping = true;
+                }
+                current_range = Some((current.0.min(next_range.0), current.1.max(next_range.1)));
```
evenyag, line 271: I guess we should set the range to None or the next range if `overlaps()` returns false. Do unit tests cover this?

src/mito2/src/compaction/twcs.rs
v0y4g3r: fixed in 2b4a00cf0f

evenyag (approved):

MichaelScofield (approved):

Issue #2532, opened 2023-10-07
The script must have the same name as the function


WenyXu:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Datanode

### What happened?

Mentioned in https://docs.greptime.com/user-guide/python-scripts/data-types

**Notes:** the function name is `query_numbers`

// file query_numbers.py
```python
from greptime import PyDataFrame, col
@copr(returns = ["value"])
def query_numbers() -> vector[f64]:
    df = PyDataFrame.from_sql("select number from numbers")
    return df.filter(col('number') <= 5).collect()[0]
```

curl --data-binary "@query_numbers.py" -XPOST "http://localhost:4000/v1/scripts?name=query_number&db=public"

Execute sql `select query_number();`

Output: 
```
ERROR 1815 (HY000): Error during planning: Invalid function 'query_number'.
Did you mean 'encode'?
```

However, it works via http.
```bash
curl  -XPOST \
 "http://localhost:4000/v1/run-script?name=query_number&db=public"
{"code":0,"output":[{"records":{"schema":{"column_schemas":[{"name":"value","data_type":"Float64"}]},"rows":[[0.0],[1.0],[2.0],[3.0],[4.0],[5.0]]}}],"execution_time_ms":16}%   
```

If you change `?name=query_number` to `?name=query_numbers` , everything works.

### What operating system did you use?

Ubuntu 23.04

### How can we reproduce the bug?

Mentioned in https://docs.greptime.com/user-guide/python-scripts/data-types

**Notes:** the function name is `query_numbers`

// file query_numbers.py
```python
from greptime import PyDataFrame, col
@copr(returns = ["value"])
def query_numbers() -> vector[f64]:
    df = PyDataFrame.from_sql("select number from numbers")
    return df.filter(col('number') <= 5).collect()[0]
```

curl --data-binary "@query_numbers.py" -XPOST "http://localhost:4000/v1/scripts?name=query_number&db=public"

Execute sql `select query_number();`

killme2008:
It's not a bug too. 

If you want to call script by HTTP API, the name is the script name you pass when saving the scripts.
If you want to call the function in SQL, you must use the function name in script.
Sorry for the misunderstood, I will clarify it in the documentation.

tisonkun:
Some more subtle manner:

1. If the session closed and a new session start, you should execute via HTTP or register a new script to trigger the scripts function load for SQL.

[...]

Issue #2434, opened 2023-09-18
Invalidate compiled script cache when it is updated


killme2008:
### What type of enhancement is this?

User experience, API

### What does the enhancement do?

We have to invalidate other frontend's compiled python script cache when the script is updated in cluster mode.

We need an event broadcast mechanism to do it.

MichaelScofield:
It can be done by calling `Mailbox::broadcast` to submit a "script invalidation request" to all the frontend nodes. There's an example to start with, see struct `InvalidateCache`.

xxxuuu:
Hi, I would like to try this one. Could you assign it to me?

WenyXu:
> Hi, I would like to try this one. Could you assign it to me?

Have fun🚀

xxxuuu:
> > Hi, I would like to try this one. Could you assign it to me?
> 
> Have fun🚀

Hi @WenyXu, I plan to communicate via `Frontend--notify-->Metasrv--broadcast-->Frontend` When I implemented it, I found that `mailbox::send` in `Frontend` requires an `InstructionReply`, and it seems that `Instruction` is always sent by `Metasrv` as heartbeat response. If I make `Frontend` notify `Metasrv`, it seems to break this semantics and design. Should I do it this way? Cloud you give me any suggestions?

WenyXu:
> > > Hi, I would like to try this one. Could you assign it to me?
> > 
> > 
> > Have fun🚀
> 
> Hi @WenyXu, I plan to communicate via `Frontend--notify-->Metasrv--broadcast-->Frontend`. When I implemented it, I found that `mailbox::send` in `Frontend` requires an `InstructionReply`, and it seems that `Instruction` is always sent by `Metasrv` as heartbeat response. If I make `Frontend` notify `Metasrv`, it seems to break this semantics and design. Should I do it this way? Cloud you give me any suggestions?

PTAL @waynexia

WenyXu:
> > > > Hi, I would like to try this one. Could you assign it to me?
> > > 
> > > 
> > > Have fun🚀
> > 
> > 
[...]

PR #3774, opened 2024-04-22
fix: dynamically get python udf from scripts table

xxxuuu:

https://github.com/GreptimeTeam/greptimedb/issues/2434
https://github.com/GreptimeTeam/greptimedb/issues/2532

Add a `version` column to the scripts table and retrieve the latest script from the database before each script execution. If it is newer than the one in the local cache, recompile it.

This solves the problem where a script updated on one frontend in a cluster with multiple frontends does not take effect on the other frontends(because the script compilation cache cannot be updated, or the script is not registered).

**BREAKING CHANGE:**
Since the scripts are now obtained through scripts table, when executing via UDF in SQL, the script name must be used instead of the function name. This fixes the issue mentioned in https://github.com/GreptimeTeam/greptimedb/issues/2532. However, in cases where the script name and function name are different, if the program is using the function name, it will no longer work.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8842,6 +8842,7 @@ dependencies = [
  "common-error",
  "common-function",
  "common-macro",
+ "common-meta",
  "common-query",
  "common-recordbatch",
  "common-runtime",
--- a/src/cmd/src/frontend.rs
+++ b/src/cmd/src/frontend.rs
@@ -275,11 +275,12 @@ impl StartCommand {
             cached_meta_backend.clone(),
             catalog_manager,
             Arc::new(DatanodeClients::default()),
-            meta_client,
+            meta_client.clone(),
         )
         .with_plugin(plugins.clone())
         .with_cache_invalidator(multi_cache_invalidator)
         .with_heartbeat_task(heartbeat_task)
+        .with_meta_client(meta_client.clone())
         .try_build()
         .await
         .context(StartFrontendSnafu)?;
--- a/src/common/function/src/function_registry.rs
+++ b/src/common/function/src/function_registry.rs
@@ -17,6 +17,7 @@ use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
 
 use once_cell::sync::Lazy;
+use session::context::QueryContextRef;
 
 use crate::function::FunctionRef;
 use crate::scalars::aggregate::{AggregateFunctionMetaRef, AggregateFunctions};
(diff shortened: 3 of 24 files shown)


tisonkun:
Thanks for your contribution!

You may take a look at https://github.com/GreptimeTeam/greptimedb/issues/3777. I suppose we may implement PyUDF in a more structural and proven way, said PL/Python, instead of hacky interpreting decorators and making new tribal rules.

For example, instead of making version names, we should use statement like:

```sql
CREATE FUNCTION pymax (a integer, b integer)
  RETURNS integer
AS $$
  if (a is None) or (b is None):
    return None
  if a > b:
    return a
  return b
$$ LANGUAGE plpython3u;
```

And then a function named `pymax` is always stored among the cluster, and no the confusing "name of file" vs. "name of function" vs. "name of name in args" things.

tisonkun:
If you have other proposal, we can discuss in #3777. Generally it's good to discuss before implementing code, to avoid going in the wrong direction in the first place.

xxxuuu:
> Thanks for your contribution!
> 
> You may take a look at #3777. I suppose we may implement PyUDF in a more structural and proven way, said PL/Python, instead of hacky interpreting decorators and making new tribal rules.
> 
> For example, instead of making version names, we should use statement like:
> 
> ```sql
> CREATE FUNCTION pymax (a integer, b integer)
>   RETURNS integer
> AS $$
>   if (a is None) or (b is None):
>     return None
>   if a > b:
>     return a
>   return b
> $$ LANGUAGE plpython3u;
> ```
> 
> And then a function named `pymax` is always stored among the cluster, and no the confusing "name of file" vs. "name of function" vs. "name of name in args" things.

Yes, this is a better approach, and I've also seen some other DBMSs doing it similar way. I will consider close this PR and then reevaluate/discuss this feature.

tisonkun:
@xxxuuu Thanks for your reply. I'm open to respond in #3777 and support any effort to implement PyUDF in a cleaner `CREATE FUNCTION` way. If you're interested in this feature, feel free to drop a comment there so that we can describe more details.

xxxuuu:
> I found that we can easily solve the original issue by banning "update scripts" but instead add a delete script API. Thus the script cache can be invalidated on deleted and otherwise it's always the corrected compiled cached ones.

Scripts created on one frontend are not registered on another frontend, and there is a similar issue with deletion. We still need to solve this issue.

tisonkun:
> Scripts created on one frontend are not registered on another frontend, and there is a similar issue with deletion. We still need to solve this issue.

This is quite surprising. IIUC the script table should be identical with other tables, and thus it should be visible to every frontend instances.

xxxuuu:
> This is quite surprising. IIUC the script table should not be different with other normal tables, and thus it should be visible to every frontend instances.

scripts table is visible, but it is not accessed when executed through SQL. Therefore, the script remains unregistered. This PR includes the modifications for this issue.

PR #3775, opened 2024-04-23
fix: operating region guards should be dropped when procedure is done

MichaelScofield:

Same as #3771, there're data race in creating table procedure.

Other two places that use this operating region guard are drop database procedure and region migration procedure. However, due to the different way of how their procedures are implemented, they are not affected. An interesting side effect.

## Checklist


WenyXu (approved):

fengjiachun (approved):
LGTM

PR #3784, opened 2024-04-23
build: only build amd64 for CentOS

evenyag:

https://github.com/briansmith/ring/issues/1728

https://github.com/GreptimeTeam/greptimedb/actions/runs/8778537946/job/24085145007

While building dev builder image on CentOS we got the following error
```
error: failed to run custom build command for `ring v0.17.8`

cargo:warning=include/ring-core/arm_arch.h:82:2: error: #error "ARM assembler must define __ARM_ARCH"
```

Version 0.17 of ring doesn't support CentOS arm64. Since we don't provide arm64 binary for CentOS, I removed the arm64 image for CentOS in this PR.

`BUILDX_MULTI_PLATFORM_BUILD` now has three possible value
- all: builds images for both amd64 and arm64 platform
- amd64: builds amd64 image only
- false: builds image for current platform

## Checklist

### DIFF
--- a/.github/actions/build-dev-builder-images/action.yml
+++ b/.github/actions/build-dev-builder-images/action.yml
@@ -22,15 +22,15 @@ inputs:
   build-dev-builder-ubuntu:
     description: Build dev-builder-ubuntu image
     required: false
-    default: 'true'
+    default: "true"
   build-dev-builder-centos:
     description: Build dev-builder-centos image
     required: false
-    default: 'true'
+    default: "true"
   build-dev-builder-android:
     description: Build dev-builder-android image
     required: false
-    default: 'true'
+    default: "true"
 runs:
   using: composite
   steps:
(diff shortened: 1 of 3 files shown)


killme2008 (approved):
LGTM

zyy17 (approved):

waynexia (approved):

tisonkun:
To be clear, once ring supports Linux arm64 again, we'd revert this change? Or we just no longer provide arm64 image.

evenyag:
> To be clear, once ring supports Linux arm64 again, we'd revert this change? Or we just no longer provide arm64 image.

ring supports arm64, but not on CentOS 7 arm64. We also have other issues while compiling on CentOS 7 arm64 so we only provide binary for CentOS 7 x64.

tisonkun:
Make sense. CentOS 7 will be end of life at June 30, 2024. I suppose we provide such an image for certain user cases, and thus if no users need CentOS 7 arm64 explicitly, we can provide only CentOS 7 x64.

evenyag:
> I suppose we provide such an image for certain user cases, and thus if no users need CentOS 7 arm64 explicitly, we can provide only CentOS 7 x64.

It only builds an image for compiling the database (dev-builder), not for serving the database. We didn't use that image in the release action as we only used the x64 dev-builder image.

So this PR fixes an unexpected behavior.

PR #3789, opened 2024-04-24
chore: always obsolete wal to avoid discontinuous log entries

v0y4g3r:

This changes the behavior during datanode startup in that it will always obsolete wal log entries to `flushed_entry_id` in manifest so that if it can tolerate inserts when the entry id span in logstore < `flushed_entry_id`.

## Checklist

### DIFF
--- a/src/mito2/src/region/opener.rs
+++ b/src/mito2/src/region/opener.rs
@@ -441,13 +441,11 @@ pub(crate) async fn replay_memtable<S: LogStore>(
     // data in the WAL.
     let mut last_entry_id = flushed_entry_id;
     let replay_from_entry_id = flushed_entry_id + 1;
-    let mut stale_entry_found = false;
 
     let mut wal_stream = wal.scan(region_id, replay_from_entry_id, wal_options)?;
     while let Some(res) = wal_stream.next().await {
         let (entry_id, entry) = res?;
         if entry_id <= flushed_entry_id {
-            stale_entry_found = true;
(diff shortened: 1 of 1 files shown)


waynexia (approved):

src/mito2/src/region/opener.rs
```diff
@@ -476,11 +474,8 @@ pub(crate) async fn replay_memtable<S: LogStore>(
         region_write_ctx.write_memtable();
     }
 
-    if allow_stale_entries && stale_entry_found {
-        wal.obsolete(region_id, flushed_entry_id, wal_options)
-            .await?;
-        info!("Force obsolete WAL entries, region id: {}, flushed entry id: {}, last entry id read: {}", region_id, flushed_entry_id, last_entry_id);
-    }
+    wal.obsolete(region_id, flushed_entry_id, wal_options)
```
evenyag, line 477: Maybe we can only do this when the last_id of the log store is not equal to `flushed_entry_id`.

evenyag (approved):
Mostly LGTM

src/mito2/src/region/opener.rs
v0y4g3r: > Maybe we can only do this when the last_id of the log store is not equal to `flushed_entry_id`.

`LogStore` does not have a `span` API that returns current min/max entry id (only raft-engine provide that functionality).

killme2008: I think we should keep the info log.

```rust
info!("Force obsolete WAL entries, region id: {}, flushed entry id: {}, last entry id read: {}", region_id, flushed_entry_id, last_entry_id);
```

PR #3792, opened 2024-04-24
refactor!: unify sql options into OptionMap

tisonkun:

This is a follow-up to #3646 

## Checklist


src/cmd/src/cli/export.rs
```diff
@@ -493,7 +493,7 @@ mod tests {
 
 ENGINE=mito
 WITH(
-  regions = 1
+  regions = '1'
```
tisonkun, line 496: We anyway pass string underneath. For historical reason, we support pass option in number for `CreateTable`.

v0y4g3r (approved):
LGTM

src/cmd/src/cli/export.rs
evenyag: Is this a breaking change?

src/sql/src/util.rs
```diff
@@ -47,29 +48,16 @@ pub fn format_raw_object_name(name: &ObjectName) -> String {
...
-        let _ = map.insert(name.value.to_lowercase().clone(), value_str);
-    }
-    map
+pub fn parse_option_string(option: SqlOption) -> Result<(String, String)> {
+    let (key, value) = (option.name, option.value);
+    let v = match value {
+        Expr::Value(Value::SingleQuotedString(v)) | Expr::Value(Value::DoubleQuotedString(v)) => v,
+        Expr::Identifier(v) => v.value,
+        Expr::Value(Value::Number(v, l)) => format!("{}{long}", v, long = if l { "L" } else { "" }),
+        value => return InvalidTableOptionValueSnafu { key, value }.fail(),
```
tisonkun, line 57: Previously we silently ignore unknown value, this can introduce quite subtle failure to debug. We can always fail on unknown option value.

src/sql/src/util.rs
```diff
@@ -47,29 +48,16 @@ pub fn format_raw_object_name(name: &ObjectName) -> String {
...
-        };
-        let _ = map.insert(name.value.to_lowercase().clone(), value_str);
-    }
-    map
+pub fn parse_option_string(option: SqlOption) -> Result<(String, String)> {
+    let (key, value) = (option.name, option.value);
+    let v = match value {
+        Expr::Value(Value::SingleQuotedString(v)) | Expr::Value(Value::DoubleQuotedString(v)) => v,
+        Expr::Identifier(v) => v.value,
+        Expr::Value(Value::Number(v, l)) => format!("{}{long}", v, long = if l { "L" } else { "" }),
```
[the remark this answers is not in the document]
tisonkun: This keeps the previous manner:

```rust
pub fn to_lowercase_options_map(opts: &[SqlOption]) -> HashMap<String, String> {
    let mut map = HashMap::with_capacity(opts.len());
    for SqlOption { name, value } in opts {
        let value_str = match value {
            Expr::Value(Value::SingleQuotedString(s))
            | Expr::Value(Value::DoubleQuotedString(s)) => s.clone(),
            Expr::Identifier(i) => i.value.clone(),
            _ => value.to_string(),
        };
        let _ = map.insert(name.value.to_lowercase().clone(), value_str);
    }
    map
}
```

`value.to_string()` will act as this line underneath.

But you're right that we may always drop the suffix since it won't help how we parse numbers.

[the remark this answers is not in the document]
tisonkun:
```suggestion
        Expr::Value(Value::Number(v, _)) => format!("{v}"),
```

src/table/src/requests.rs
```diff
@@ -53,7 +53,6 @@ pub fn validate_table_option(key: &str) -> bool {
         // common keys:
         WRITE_BUFFER_SIZE_KEY,
         TTL_KEY,
-        REGIONS_KEY,
```
tisonkun, line 56: This change make `regions = 1` an illegal option and thus a breaking change.

@killme2008 @waynexia please confirm if we should allow `regions = 1` but leave it unused, or we can break and throw an `InvalidTableOption` error.

killme2008: The `regions` option is for future usage, and we must support multiple regions for standalone mode.

killme2008: I think we can just ignore it currently but doesn't need to break it.

tisonkun: OK. Then I'll add back this option in validation so such an option can be passed but with no effect.

killme2008: No hurry, looks like we don't need it at all even in standalone mode.

PR #3793, opened 2024-04-24
fix: post process result on query full column name of prom labels API

waynexia:

`/labels` without matcher is handled in another way. This patch fixes the incorrect behavior of two cases.

## Checklist

### DIFF
--- a/src/servers/src/http/prometheus.rs
+++ b/src/servers/src/http/prometheus.rs
@@ -21,7 +21,6 @@ use catalog::CatalogManagerRef;
 use common_catalog::parse_catalog_and_schema_from_db_string;
 use common_error::ext::ErrorExt;
 use common_error::status_code::StatusCode;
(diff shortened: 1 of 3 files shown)


killme2008 (approved):
LGTM

yuanbohan (approved):
LGTM

src/servers/src/http/prometheus.rs
```diff
@@ -398,6 +401,10 @@ async fn get_all_column_names(
         }
     }
 
+    let _ = labels.insert(METRIC_NAME.to_string());
+    let _ = labels.remove(GREPTIME_TIMESTAMP);
```
yuanbohan, line 405: BTW, all field columns SHOULD BE filtered from the response, and the GREPTIME_VALUE is just a special field column

yuanbohan (approved):

PR #3794, opened 2024-04-24
fix!: use the right precision

fengjiachun:

Use the right precision.

Before: The data written  with Influxdb line protocol before was stored as type `Millisecond`, resulting in loss of precision.
After: The data written with influxes line protocol and OTLP protocol will be stored as type "Nanosecond"

Note that this is a breaking change because the data storage type has changed from `Millisecond` to `Nanosecond`. For existing tables, new writes will result in an error. One solution is to delete the data and re-import it.

Others:
1. Remove `writer.rs` and `line_writer.rs` since we do not need them any more(There were for column protocol)
2. Add more unit test

## Checklist


src/servers/src/influxdb.rs
```diff
@@ -309,9 +309,9 @@ monitor2,host=host4 cpu=66.3,memory=1029 1663840496400340003";
         }
     }
 
-    fn extract_ts_millis_value(value: &ValueData) -> i64 {
+    fn extract_ts_nanos_value(value: &ValueData) -> i64 {
         match value {
-            ValueData::TimestampMillisecondValue(v) => *v,
+            ValueData::TimestampNanosecondValue(v) => *v,
             _ => panic!(),
```
tisonkun, line 315: unreachable?

src/servers/src/influxdb.rs
```diff
@@ -88,7 +88,7 @@ impl TryFrom<InfluxdbRequest> for RowInsertRequests {
 
             // timestamp
             let precision = unwrap_or_default_precision(value.precision);
```
etolbakov, line 90: it's not strictly related to the changes, but a side observation - 
does it make sense to move `precision` calculation out of the `for line in &lines {` WDYT?

fengjiachun: Good catch!

yuanbohan (approved):

src/common/grpc/src/precision.rs
```diff
@@ -0,0 +1,139 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fmt::Display;
+
+use common_time::timestamp::TimeUnit;
+
+use crate::Error;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
killme2008, line 21: Please add some descriptions.

killme2008 (approved):
LGTM

PR #3796, opened 2024-04-24
feat: support `ALTER TABLE ... MODIFY COLUMN ... ...`

KKould:

https://github.com/GreptimeTeam/greptimedb/issues/3517
https://github.com/GreptimeTeam/greptime-proto/pull/149

Support: `alter table xxx modify column xxx xxx`
https://dev.mysql.com/doc/refman/8.0/en/alter-table.html

```shell
public=> CREATE TABLE test(id INTEGER PRIMARY KEY, i INTEGER NULL, j TIMESTAMP TIME INDEX, k BOOLEAN);
OK 0
public=> ALTER TABLE test MODIFY COLUMN I STRING;
OK 0
public=> DESCRIBE test;
 Column |         Type         | Key | Null | Default | Semantic Type 
--------+----------------------+-----+------+---------+---------------
 id     | Int32                | PRI | YES  |         | TAG
 i      | String               |     | YES  |         | FIELD
 j      | TimestampMillisecond | PRI | NO   |         | TIMESTAMP
 k      | Boolean              |     | YES  |         | FIELD
(4 rows)


```

## Checklist


src/common/meta/src/ddl/alter_table/region_request.rs
```diff
@@ -244,15 +227,99 @@ mod tests {
...
+                                location_type: LocationType::After as i32,
+                                after_column_name: "my_tag2".to_string(),
+                            }),
+                        }]
+                    }
+                ))
+            );
+        }
+        {
+            let change_column_type_task = AlterTableTask {
```
evenyag, line 281: Could we move this to a new test?

src/store-api/src/region_request.rs
```diff
@@ -612,6 +621,23 @@ impl ChangeColumnType {
...
+            change_column_type.target_type(),
+            change_column_type.target_type_extension,
+        )
+        .into();
+
+        Ok(ChangeColumnType {
+            column_name: change_column_type.column_name,
+            target_type,
+        })
+    }
```
evenyag, line 638: It doesn't return any error. Could we implement `From` instead?

Cargo.toml
```diff
@@ -115,7 +115,7 @@ etcd-client = { git = "https://github.com/MichaelScofield/etcd-client.git", rev
 fst = "0.4.7"
 futures = "0.3"
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "73ac0207ab71dfea48f30259ffdb611501b5ecb8" }
+greptime-proto = { git = "https://github.com/KKould/greptime-proto.git", rev = "d7c348fda27a593f34c19793652d591dd8af1106" }
```
evenyag, line 118: Wait for https://github.com/GreptimeTeam/greptime-proto/pull/149

evenyag (commented):
Mostly looks good.

killme2008:
Good job! But why not use the common SQL syntax?

```sql
ALTER TABLE table_name
MODIFY COLUMN column_name datatype;
```

It would be better that the user doesn't need to learn a new syntax.

evenyag:
> But why not use the common SQL syntax?

This is the syntax PostgreSQL currently uses.
https://www.postgresql.org/docs/current/sql-altertable.html

```sql
ALTER TABLE distributors
    ALTER COLUMN address TYPE varchar(80),
    ALTER COLUMN name TYPE varchar(100);
```

killme2008:
> > But why not use the common SQL syntax?
> 
> This is the syntax PostgreSQL currently uses. https://www.postgresql.org/docs/current/sql-altertable.html
> 
> ```sql
> ALTER TABLE distributors
>     ALTER COLUMN address TYPE varchar(80),
>     ALTER COLUMN name TYPE varchar(100);
> ```

Yes, but MySQL, Oracle, and SQLServer are using `modify column`.

KKould:
tips: sqlparser-rs already supports modify column: https://github.com/sqlparser-rs/sqlparser-rs/pull/1216

MichaelScofield:
Create an issue [here](https://github.com/GreptimeTeam/docs/issues/915) for reminding us the docs after this whole feature is done.

src/sql/src/parsers/alter_parser.rs
```diff
@@ -73,6 +73,26 @@ impl<'a> ParserContext<'a> {
                     parser.peek_token()
                 )));
             }
+        } else if parser
+            .expect_keywords(&[Keyword::ALTER, Keyword::COLUMN])
```
MichaelScofield, line 77: +1 for "modify column" for the conciseness

KKould: Even if you want to use `ModifyColumn`, you need to upgrade sqlparser-rs to the latest version first. There is no `Keyword::Modify` now.

WenyXu:
Hi @KKould There are some conflicts

WenyXu:
> tips: sqlparser-rs already supports modify column: [sqlparser-rs/sqlparser-rs#1216](https://github.com/sqlparser-rs/sqlparser-rs/pull/1216)

Is it possible to upgrade the `SQL parser-rs` to the latest?

KKould:
I created PR: https://github.com/GreptimeTeam/sqlparser-rs/pull/11 to support `modify column`

src/common/meta/src/ddl/alter_table/region_request.rs
```diff
@@ -91,6 +92,23 @@ fn create_proto_alter_kind(
...
+                    target_type: change_column_type.target_type,
+                    target_type_extension: change_column_type.target_type_extension.clone(),
+                })
+                .collect();
+
+            Ok(Some(alter_request::Kind::ChangeColumnTypes(
+                ChangeColumnTypes {
+                    change_column_types,
+                },
+            )))
```
evenyag, line 110: Could we pass `x` directly? They should have the same type now.
```suggestion
        Kind::ChangeColumnTypes(x) => {
            Ok(Some(alter_request::Kind::ChangeColumnTypes(x)))
```

Cargo.toml
```diff
@@ -115,7 +115,7 @@ etcd-client = { git = "https://github.com/MichaelScofield/etcd-client.git", rev
 fst = "0.4.7"
 futures = "0.3"
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "783682fabc38c57b5b9d46bdcfeebe2496e85bbb" }
+greptime-proto = { git = "https://github.com/KKould/greptime-proto.git", rev = "9022a4eb94b25696c20ec816834d39c305b46122" }
```
evenyag, line 118: Remember to use the official proto.

src/store-api/src/region_request.rs
```diff
@@ -15,6 +15,7 @@
 use std::collections::HashMap;
 use std::fmt::{self};
```
evenyag, line 16:
```suggestion
use std::fmt;
```

src/sql/src/parsers/alter_parser.rs
evenyag: Maybe we could change the syntax or support `modify` in another PR.

src/store-api/src/region_request.rs
KKould: eagle eyes

evenyag (approved):

evenyag:
If the CI failure in https://github.com/GreptimeTeam/sqlparser-rs/pull/11 is easy to fix. We could implement the final syntax in this PR. @KKould Could you please take a look?

KKould:
> If the CI failure in [GreptimeTeam/sqlparser-rs#11](https://github.com/GreptimeTeam/sqlparser-rs/pull/11) is easy to fix. We could implement the final syntax in this PR. @KKould Could you please take a look?

I checked before and it seems that these errors are caused by other PRs, not this PR.

Taylor-lagrange:
The pr should be open on [v0.44.x](https://github.com/GreptimeTeam/sqlparser-rs/tree/v0.44.x) instead of the main branch, because the sqlparser code will be synchronized with datafusion' sqlparser code regularly. However, I still recommend not to change the keywords and use parser directly to identify the token to avoid conflict when sync upstream sqlparser later.

src/sql/src/parsers/alter_parser.rs
```diff
@@ -73,6 +73,15 @@ impl<'a> ParserContext<'a> {
                     parser.peek_token()
                 )));
             }
+        } else if ParserContext::inner_consume_token(parser, "MODIFY") {
```
evenyag, line 76: Why do we need `inner_consume_token()` here? Could we use `parser.consume_token()`?
```suggestion
        } else if parser.consume_token("MODIFY") {
```

src/sql/src/parsers/alter_parser.rs
```diff
@@ -253,14 +262,60 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_parse_alter_change_column_type() {
+        let sql_1 = "ALTER TABLE my_metric_1 MODIFY COLUMN a STRING";
```
evenyag, line 267: Will the parser return an error if we also modify the column constraint? `ALTER TABLE my_metric_1 MODIFY COLUMN a STRING NOT NULL`

KKould: I thought this would just ignore and not return an error

src/sql/src/parsers/alter_parser.rs
KKould: I used parser.consume_token() at first, but it was not compatible with both upper and lower case

evenyag: Sorry, I mean `self.consume_token()`. There is no need to keep the `&mut self.parser` in L33 which prevent us using `&mut self` later.
```suggestion
        } else if self.consume_token("MODIFY") {
```

src/sql/src/parser.rs
```diff
@@ -197,7 +202,12 @@ impl<'a> ParserContext<'a> {
 
     #[inline]
     pub(crate) fn peek_token_as_string(&self) -> String {
-        self.parser.peek_token().to_string()
+        Self::inner_peek_token_as_string(&self.parser)
+    }
+
+    #[inline]
+    pub(crate) fn inner_peek_token_as_string(parser: &Parser<'a>) -> String {
+        parser.peek_token().to_string()
```
evenyag, line 210: Prefer to remove `inner_peek_token_as_string`

src/sql/src/parser.rs
```diff
@@ -187,8 +187,13 @@ impl<'a> ParserContext<'a> {
...
     pub fn consume_token(&mut self, expected: &str) -> bool {
-        if self.peek_token_as_string().to_uppercase() == *expected.to_uppercase() {
-            let _ = self.parser.next_token();
+        Self::inner_consume_token(&mut self.parser, expected)
+    }
+
+    #[inline]
+    pub(crate) fn inner_consume_token(parser: &mut Parser<'a>, expected: &str) -> bool {
+        if Self::inner_peek_token_as_string(parser).to_uppercase() == *expected.to_uppercase() {
+            let _ = parser.next_token();
```
evenyag, line 196: We can just call `consume_token()`.

src/sql/src/parser.rs
```diff
@@ -187,8 +187,13 @@ impl<'a> ParserContext<'a> {
...
 
     pub fn consume_token(&mut self, expected: &str) -> bool {
-        if self.peek_token_as_string().to_uppercase() == *expected.to_uppercase() {
-            let _ = self.parser.next_token();
+        Self::inner_consume_token(&mut self.parser, expected)
+    }
+
+    #[inline]
+    pub(crate) fn inner_consume_token(parser: &mut Parser<'a>, expected: &str) -> bool {
+        if Self::inner_peek_token_as_string(parser).to_uppercase() == *expected.to_uppercase() {
```
evenyag, line 195: We can call `eq_ignore_ascii_case` instead of allocating two new string using `to_uppercase()`
https://doc.rust-lang.org/stable/std/string/struct.String.html#method.eq_ignore_ascii_case

src/sql/src/parsers/alter_parser.rs
KKould: lol, I also tried it, but because parser is a mutable reference of self.parser, self cannot be used here.

KKould: I'm troubled by this, and I don't want to get an `inner_consume_token`

src/operator/src/expr_factory.rs
```diff
@@ -17,9 +17,9 @@ use std::collections::{HashMap, HashSet};
 use api::helper::ColumnDataTypeWrapper;
 use api::v1::alter_expr::Kind;
 use api::v1::{
-    AddColumn, AddColumns, AlterExpr, Column, ColumnDataType, ColumnDataTypeExtension,
-    CreateFlowTaskExpr, CreateTableExpr, DropColumn, DropColumns, RenameTable, SemanticType,
-    TableName,
+    AddColumn, AddColumns, AlterExpr, ChangeColumnType, ChangeColumnTypes, Column, ColumnDataType,
+    ColumnDataTypeExtension, CreateFlowTaskExpr, CreateTableExpr, DropColumn, DropColumns, RenameTable,
+    SemanticType, TableName,
```
waynexia, line 22:
```suggestion
    ColumnDataTypeExtension, CreateFlowTaskExpr, CreateTableExpr, DropColumn, DropColumns,
    RenameTable, SemanticType, TableName,
```

MichaelScofield (approved):

PR #3802, opened 2024-04-25
fix: prune row groups correctly for columns with the same name

evenyag:

This PR fixes an issue that the parquet reader might prune row groups unexpectedly if two different columns are using the same name. This is possible if we drop a column and then add a column with the same name.

This PR passes the metadata of the region to the `RowGroupPruningStats` so it can get the correct column id to read.

For example, if we have the following table:
```sql
CREATE TABLE test(i TIMESTAMP TIME INDEX, j INTEGER, k INTEGER NOT NULL);
INSERT INTO test(i, j, k) VALUES (1, 11, 5), (2, 12, 5);
SELECT FLUSH_TABLE('test');
ALTER TABLE test DROP COLUMN j;
ALTER TABLE test ADD COLUMN j INTEGER DEFAULT 0;
INSERT INTO test(i, j, k) VALUES (3, 0, 6);
INSERT INTO test VALUES (4, 7, 0);
```

Before this PR:
```bash
SELECT * FROM test order by i;
+-------------------------+---+---+
| i                       | k | j |
+-------------------------+---+---+
| 1970-01-01T00:00:00.001 | 5 | 0 |
| 1970-01-01T00:00:00.002 | 5 | 0 |
| 1970-01-01T00:00:00.003 | 6 | 0 |
| 1970-01-01T00:00:00.004 | 7 | 0 |
+-------------------------+---+---+

SELECT * FROM test WHERE j = 0 order by i;
+-------------------------+---+---+
| i                       | k | j |
+-------------------------+---+---+
| 1970-01-01T00:00:00.003 | 6 | 0 |
| 1970-01-01T00:00:00.004 | 7 | 0 |
+-------------------------+---+---+
```

After this PR:
```bash
SELECT * FROM test WHERE j = 0 order by i;
+-------------------------+---+---+
| i                       | k | j |
+-------------------------+---+---+
| 1970-01-01T00:00:00.001 | 5 | 0 |
| 1970-01-01T00:00:00.002 | 5 | 0 |
| 1970-01-01T00:00:00.003 | 6 | 0 |
| 1970-01-01T00:00:00.004 | 7 | 0 |
+-------------------------+---+---+
```

## Checklist


fengjiachun (approved):
LGTM

waynexia (approved):
LGTM

PR #3804, opened 2024-04-26
refactor: use secrecy SerectString to hold secrets option

### LINKED ISSUE #2144: Using SecretString to store senstive values in `copy` stmt

tisonkun:

This closes #2144.

## Checklist

### DIFF
--- a/src/sql/src/statements/option_map.rs
+++ b/src/sql/src/statements/option_map.rs
@@ -78,15 +78,11 @@ impl OptionMap {
 
     pub fn kv_pairs(&self) -> Vec<String> {
         let mut result = Vec::with_capacity(self.options.len() + self.secrets.len());
-        for key in self.options.keys() {
-            if let Some(val) = self.options.get(key) {
-                result.push(format!("{key} = '{}'", val.escape_default()));
-            }
+        for (k, v) in self.options.iter() {
+            result.push(format!("{k} = '{}'", v.escape_default()));
         }
-        for key in self.secrets.keys() {
-            if self.secrets.contains_key(key) {
-                result.push(format!("{key} = '******'"));
-            }
+        for (k, _) in self.secrets.iter() {
+            result.push(format!("{k} = '******'"));
         }
         result
     }


src/sql/src/statements/copy.rs
```diff
@@ -152,7 +152,7 @@ mod tests {
             Statement::Copy(copy) => {
                 let new_sql = format!("{}", copy);
                 assert_eq!(
-                    r#"COPY tbl FROM s3://my-bucket/data.parquet WITH (format = 'parquet', pattern = '.*parquet.*') CONNECTION (region = 'us-west-2', secret_access_key = '******')"#,
+                    r#"COPY tbl FROM s3://my-bucket/data.parquet WITH (format = 'parquet', pattern = '.*parquet.*') CONNECTION (region = 'us-west-2', secret_access_key = Secret([REDACTED alloc::string::String]))"#,
```
tisonkun, line 155: Actually I don't find this is better than `'******'`, perhaps we can still display `SecretString` as `'******'`?

src/sql/src/statements/option_map.rs
```diff
@@ -12,52 +12,144 @@
...
+        );
+        map
+    }
+
+    pub fn into_map(self) -> HashMap<String, String> {
+        let mut map = self.options;
+        map.extend(
+            self.secrets
+                .into_iter()
+                .map(|(k, v)| (k, v.expose_secret().to_string())),
```
tisonkun, line 74: Extra clone in `to_string`. The `secrecy` upstream doesn't export a method for something like `into_string`.

They're experiencing a rewrite so I'm unsure if they can even accept patches:

https://github.com/iqlusioninc/crates/tree/main/secrecy

src/sql/src/statements/option_map.rs
```diff
@@ -12,52 +12,144 @@
...
+    pub fn kv_pairs(&self) -> Vec<String> {
+        let mut result = Vec::with_capacity(self.options.len() + self.secrets.len());
+        for key in self.options.keys().sorted() {
+            if let Some(val) = self.options.get(key) {
+                result.push(format!("{key} = '{}'", val.escape_default()));
+            }
+        }
+        for key in self.secrets.keys().sorted() {
+            if let Some(val) = self.secrets.get(key) {
+                result.push(format!("{key} = {val:?}"));
```
tisonkun, line 88:
```suggestion
                result.push(format!("{key} = '******'"));
```

?

src/sql/src/statements/option_map.rs
```diff
@@ -12,52 +12,144 @@
...
-    fn from_iter<I: IntoIterator<Item = (String, String)>>(iter: I) -> Self {
-        Self {
-            map: iter.into_iter().collect(),
+impl VisitMut for OptionMap {
+    fn visit<V: VisitorMut>(&mut self, visitor: &mut V) -> ControlFlow<V::Break> {
+        for (_, v) in self.options.iter_mut() {
+            v.visit(visitor)?;
+        }
+        for (_, v) in self.secrets.iter_mut() {
+            let mut v = v.expose_secret().to_string();
```
tisonkun, line 149: Extra clone. IIRC new nightly version of `secrecy` has an `expose_secret_mut` method for ease this case.

src/operator/src/expr_factory.rs
```diff
@@ -189,7 +189,8 @@ pub fn create_to_expr(create: &CreateTable, query_ctx: QueryContextRef) -> Resul
 
     let time_index = find_time_index(&create.constraints)?;
     let table_options = HashMap::from(
-        &TableOptions::try_from(create.options.as_ref()).context(UnrecognizedTableOptionSnafu)?,
+        &TableOptions::try_from(&create.options.clone().into_map())
```
tisonkun, line 192: Extra clone. This is subtle because it's awkward to convert between `HashMap<&String, &String>` and `&HashMap<String, String>`.

Maybe we can duplicate the `try_from` impl for `HashMap<&String, &String>`, but I don't like this idea actually.

src/sql/src/statements/option_map.rs
tisonkun: Resolved.

src/sql/src/statements/option_map.rs
tisonkun: This is unavoidable because the SecretBox implements `Drop` in a customized way. I suppose it won't introduce too much overhead.

src/sql/src/statements/option_map.rs
tisonkun: Resolved.

src/sql/src/statements/option_map.rs
```diff
@@ -12,52 +12,142 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.options.is_empty() && self.secrets.is_empty()
+    }
+
+    pub fn len(&self) -> usize {
+        self.options.len() + self.secrets.len()
+    }
+
+    pub fn to_str_map(&self) -> HashMap<&str, &str> {
+        let mut map = HashMap::new();
```
MichaelScofield, line 59:
```suggestion
        let mut map = HashMap::with_capacity(self.len());
```

src/sql/src/statements/option_map.rs
```diff
@@ -12,52 +12,142 @@
...
+        map.extend(
+            self.secrets
+                .iter()
+                .map(|(k, v)| (k.as_str(), v.expose_secret().as_str())),
+        );
+        map
+    }
+
+    pub fn into_map(self) -> HashMap<String, String> {
+        let mut map = self.options;
```
MichaelScofield, line 70:
```suggestion
        let mut map = HashMap::with_capacity(self.len());
        map.extend(self.options);
```

src/sql/src/statements/option_map.rs
```diff
@@ -12,52 +12,142 @@
...
+            self.secrets
+                .into_iter()
+                .map(|(k, v)| (k, v.expose_secret().to_string())),
+        );
+        map
+    }
+
+    pub fn kv_pairs(&self) -> Vec<String> {
+        let mut result = Vec::with_capacity(self.options.len() + self.secrets.len());
+        for key in self.options.keys().sorted() {
```
MichaelScofield, line 81: Is it necessary to sort the options?

tisonkun: This is asked and the same reason as https://github.com/GreptimeTeam/greptimedb/pull/3744#discussion_r1577214798.

Perhaps we can use `BTreeMap` or other stable iterable collections and avoid this sort. But it doesn't affect too much IMO.

tisonkun:
@MichaelScofield Comments addressed.

src/sql/src/statements/option_map.rs
MichaelScofield: If it's for testing purpose, you can sort and compare the final `Vec<String>`, in testing codes, outside this method.

tisonkun: Many of our testing code compare the format result. I'd try to implement `Display` for `OptionMap` but still keep the sort manner. Generally we don't have more than 100 options and this overhead is minor.

tisonkun: Use BTreeMap for a quick "fix".

MichaelScofield (approved):

src/sql/src/statements/option_map.rs
```diff
@@ -12,52 +12,139 @@
...
+        map
+    }
+
+    pub fn kv_pairs(&self) -> Vec<String> {
+        let mut result = Vec::with_capacity(self.options.len() + self.secrets.len());
+        for key in self.options.keys() {
+            if let Some(val) = self.options.get(key) {
+                result.push(format!("{key} = '{}'", val.escape_default()));
+            }
+        }
```
fengjiachun, line 85: Why not use kv iter?
```
        for (key, val) in self.options.iter() {
            result.push(format!("{key} = '{}'", val.escape_default()));
        }
```

src/sql/src/statements/option_map.rs
```diff
@@ -12,52 +12,139 @@
...
+        for key in self.secrets.keys() {
+            if self.secrets.contains_key(key) {
```
fengjiachun, line 87: Question: why we need to check contains?

src/sql/src/statements/option_map.rs
tisonkun: Previously we sort keys, but now we use `BTreeMap` that has stable iteration. I'll update this.

src/sql/src/statements/option_map.rs
tisonkun: Correct. I'll update this/

tisonkun:
@fengjiachun Resolved.

fengjiachun (approved):
LGTM

PR #3808, opened 2024-04-26
feat: replay all regions in a batched manner

niebayes:

When using Kafka WAL, the read amplification of region replay can become exceedingly high as multiple regions are assigned to a topic, requiring numerous WAL entries to be pulled repeatedly. This PR aims to remove the read amplification by performing region replay in a batched manner.

Specifically, each RegionOpenRequest now includes an optional WalReader field. This field serves as a mutable reference to the WalReader instance. If the WalReader is Some, region replay is bypassed, and essential information for replaying the region is stored within the WalReader.

Once the region server completes handling all RegionOpenRequests, the WalReader contains the necessary information for replaying all regions. This information enables replaying all regions in a batched manner without introducing any read amplification.

On the other hand, to support replaying regions by topic rather than by region, we have added the group_by_namespace interface to the LogStore. This interface will perform a group by operation on the input namespaces according to certain rules. For Kafka WAL, this interface will group namespaces by topic.

Once we obtain the streamlined namespaces, we then fetch logs from multiple namespaces, aka. topics,  in parallel. We parse data from each log, then insert it into a RegionPutRequest, which is then processed by the region server. Considering that replayed data should not be written to the WAL again, we have added an Option<EntryId> to the RegionPutRequest. If it is Some, it indicates that this is replayed data. If it is None, it indicates that this is newly written data.

This PR is still in its draft version. I will add necessary tests and refine the codes.

## Checklist


tisonkun:
Closed as stale and conflict. Further pick-up should redo this patch.

PR #3812, opened 2024-04-26
refactor: remove substrait ser/de for region query in standalone

### LINKED ISSUE #3520: Remove the unnecessary ser/de by substrait in standalone mode

Performance

Currently the standalone use the same query engine just like distributed, which rewrites the logical plan and ser/de it by substrait. This is not required for standalone, and can bring some performance penalty.

### Implementation challenges

_No response_

MichaelScofield:

close #3520

This PR makes the logical plan in region query request not always be serialized to and deserialized from substrait bytes, which is not necessary in standalone mode.

## Checklist


src/common/meta/src/node_manager.rs
```diff
@@ -16,13 +16,27 @@ use std::sync::Arc;
...
+pub struct QueryRequest {
+    /// The header of this request. Often to store some context of the query. None means all to defaults.
+    pub header: Option<RegionRequestHeader>,
+
+    /// The id of the region to be queried.
+    pub region_id: RegionId,
+
+    /// The form of the query: a logical plan.
+    pub plan: LogicalPlan,
+}
```
evenyag, line 38: Putting the `QueryRequest` in the `common-meta` is weird. What about moving it to `common-query`?

src/query/src/dist_plan/planner.rs
```diff
@@ -229,24 +215,3 @@ impl TreeNodeVisitor for TableNameExtractor {
         }
     }
 }
-
-struct TableNameRewriter;
```
evenyag, line 233: Why we don't need to rewrite the table name?

src/common/meta/src/node_manager.rs
MichaelScofield: Can't do it because `RegionId` is in `store-api`, and `store-api` depends on `common-query`.

src/query/src/dist_plan/planner.rs
MichaelScofield: It seems it's not needed at the first place. Substrait is able to serde the fully qualified table name correctly. Rewriting table name like this could make physical plan failed at the optimization stage because the schema is not match before and after.

src/common/meta/src/node_manager.rs
evenyag: Do we need to keep `region_id` in the `QueryRequest`? We can pass `region_id` to `handle_read()` like `handle_request()`.
https://github.com/GreptimeTeam/greptimedb/blob/a6a702de4e51e91d4066d5a8e916b2e99116316a/src/datanode/src/region_server.rs#L117-L121

MichaelScofield: I think it's better to lessen the args of a function. The `QueryRequest` is directly used in `trait Datanode` defined in the same file, if `QueryRequest` is to be not placed in common-meta crate, so is that trait. I think we can do it in the future, to better organize the codes.

evenyag (approved):

killme2008:
@waynexia Please take a look.

waynexia:
I'm going to look into this after v0.8 release.

MichaelScofield:
@waynexia PTAL

src/common/meta/src/node_manager.rs
waynexia: moved to `common-query`

waynexia (approved):

src/common/meta/Cargo.toml
```diff
@@ -23,18 +23,20 @@ common-catalog.workspace = true
 common-error.workspace = true
 common-grpc-expr.workspace = true
 common-macro.workspace = true
-common-procedure.workspace = true
 common-procedure-test.workspace = true
+common-procedure.workspace = true
```
evenyag, line 27:
```suggestion
common-procedure.workspace = true
common-procedure-test.workspace = true
```

src/common/meta/Cargo.toml
```diff
@@ -23,18 +23,20 @@ common-catalog.workspace = true
...
 common-time.workspace = true
 common-wal.workspace = true
 datafusion-common.workspace = true
+datafusion-expr.workspace = true
 datatypes.workspace = true
 derive_builder.workspace = true
 etcd-client.workspace = true
-futures.workspace = true
 futures-util.workspace = true
+futures.workspace = true
```
evenyag, line 39:
```suggestion
futures.workspace = true
futures-util.workspace = true
```

PR #3817, opened 2024-04-26
feat: improve error message for typo in IF NOT EXISTS

### LINKED ISSUE #3471: Misleading error message for SQL typo

User experience

```sql
CREATE TABLE IF NOT EXIST t(`Hits` STRING NULL, ts TIMESTAMP(3) NOT NULL, TIME INDEX ("ts"));
```

throws:

```
ERROR 1815 (HY000): Missing time index constraint
```

But it's actually because `IF NOT EXIST` should be `IF NOT EXISTS`.

### Implementation challenges

_No response_

tisonkun:

This closes #3471.

Result:

```
public=> CREATE TABLE IF NOT EXIST t();
ERROR:  Unexpected token while parsing SQL statement: CREATE TABLE IF NOT EXIST t();, expected: 'EXISTS', found: EXIST: sql parser error: Expected EXISTS, found: EXIST at Line: 1, Column 21
public=> CREATE TABLE IF NOT t();
ERROR:  Unexpected token while parsing SQL statement: CREATE TABLE IF NOT t();, expected: 'EXISTS', found: t: sql parser error: Expected EXISTS, found: t at Line: 1, Column 21
public=> CREATE TABLE IF EXISTS t();
ERROR:  SQL statement is not supported: CREATE TABLE IF EXISTS t();, keyword: EXISTS
public=> CREATE TABLE IF EXIST t();
ERROR:  Missing time index constraint
public=> CREATE TABLE t();
ERROR:  Missing time index constraint
public=> CREATE TABLE t(ts TIMESTAMP TIME INDEX);
OK 0
public=> CREATE TABLE IF();
ERROR:  Missing time index constraint
public=> CREATE TABLE IF(ts TIMESTAMP TIME INDEX);
OK 0
public=> DROP TABLE t;
OK 0
public=> DROP TABLE IF;
OK 0
```

## Checklist


killme2008 (commented):
Could you add some sqlness cases?

tisonkun:
@killme2008 added. The sample queries above are added as a new sqlness case `create_if_not_exists.sql`

src/sql/src/parsers/create_parser.rs
```diff
@@ -144,6 +137,35 @@ impl<'a> ParserContext<'a> {
...
+                .context(UnexpectedSnafu {
+                    sql: self.sql,
+                    expected: "EXISTS",
+                    actual: self.peek_token_as_string(),
+                });
+        }
+
+        if self.parser.parse_keywords(&[Keyword::IF, Keyword::EXISTS]) {
+            return UnsupportedSnafu {
+                sql: self.sql,
```
evenyag, line 160: non-blocking: Maybe we should remove the sql from the error message.

evenyag (approved):

src/sql/src/parsers/create_parser.rs
tisonkun: IIRC this error goes to client and we don't log it. So it should be fine to keep as is.

We may investigate this manner later and not couple it with this patch :D It's anyway not a regression.

waynexia (approved):

PR #3819, opened 2024-04-27
feat: limit total rows copied in `COPY TABLE FROM` stmt

### LINKED ISSUE #3714: Limiting total rows copied in `COPY TABLE FROM` statement

As pointed out by [this article](https://duckdb.org/2024/03/26/42-parquet-a-zip-bomb-for-the-big-data-age.html), a 42 KiB parquet file may contain hundreds of trillions of values, just like what zip bomb does. If GreptimeDB does not limit the total rows copied in single `COPY TABLE FROM` statement, these parquet bombs may immediately overload the backend storage.

### What does the feature do?

Limit total row imported by single `COPY TABLE FROM` statement.

The execution of copy statements will build a record batch stream on file and insert the batches yielded to mito engine. 

https://github.com/GreptimeTeam/greptimedb/blob/3acd5bfad0ab5f0786ce9206e9cb8dd1a7cc5890/src/operator/src/statement/copy_table_from.rs#L402-L437

We can add a config option to limit the total rows read from the record batch stream and terminate the execution once  it exceeds threshold.

irenjj:

close #3714

limit total rows copied in `COPY TABLE FROM`

## Checklist


src/operator/src/statement/copy_table_from.rs
```diff
@@ -427,6 +432,10 @@ impl StatementExecutor {
                     rows_inserted += rows;
                     insert_cost += cost;
                 }
+
+                if rows_inserted >= max_insert_rows {
```
irenjj, line 436: hi @v0y4g3r , should we immediately stop when rows_inserted reaches max_insert_rows?

v0y4g3r: > hi @v0y4g3r , should we immediately stop when rows_inserted reaches max_insert_rows?

Maybe we should break the outer loop and terminate the execution of whole `COPY` statement.

6156eb7d770bb3532bf4aeb5ac911bef1c847f77
tisonkun, line 1: Seems an unrelated file. Please remove it.

src/operator/src/statement/copy_table_from.rs
```diff
@@ -377,7 +377,17 @@ impl StatementExecutor {
 
         let mut rows_inserted = 0;
         let mut insert_cost = 0;
-        for (compat_schema, file_schema_projection, projected_table_schema, file_metadata) in files
+        let max_insert_rows = req
+            .with
+            .get("max_insert_rows")
+            .and_then(|val| val.parse::<usize>().ok())
+            .unwrap_or(1000);
```
tisonkun, line 384: I'd prefer make this default a constsant.

src/operator/src/statement/copy_table_from.rs
tisonkun: I suggest instead we extract a function to return `(rows_inserted, insert_cost)` and return here. break with label is generally hard to interpret and the logic here is large enough to have a dedicated function.

tisonkun (commented):
Generally LGTM. A few style suggestions.

WenyXu:
Could you add a sqlness test? Here are some examples: 
https://github.com/GreptimeTeam/greptimedb/tree/d11b1fa389c54381c46fa17894b05fc21536815b/tests/cases/standalone/common/copy

tests/cases/standalone/common/copy/copy_from_fs_parquet.sql
```diff
@@ -22,10 +22,18 @@ Copy with_pattern FROM '/tmp/demo/export/parquet/' WITH (PATTERN = 'demo.*');
 
 select * from with_pattern order by ts;
 
+CREATE TABLE with_limit_rows(host string, cpu double, memory double, ts timestamp time index);
+
+Copy with_limit_rows FROM '/tmp/demo/export/parquet/' WITH (MAX_INSERT_ROWS = 10);
```
WenyXu, line 27: Let's add another case to ensure it works as expected.
```sql
Copy with_limit_rows FROM '/tmp/demo/export/parquet/' WITH (MAX_INSERT_ROWS = 1);
```

irenjj: For small datasets, batch_insert may not accurately limit the number of inserted rows. Perhaps we can create another Parquet file separately so that we can limit the maximum number of inserted rows in the case of small datasets.

src/operator/src/statement/copy_table_from.rs
```diff
@@ -377,6 +378,11 @@ impl StatementExecutor {
 
         let mut rows_inserted = 0;
         let mut insert_cost = 0;
+        let max_insert_rows = req
+            .with
+            .get("max_insert_rows")
```
v0y4g3r, line 383: Maybe make this option name a constant.

v0y4g3r (approved):
LGTM

v0y4g3r:
@irenjj  Just forgot to mention that the docs for `copy` statement also needs to be updated.

waynexia:
Are you willing to take the follow-up task to make this limit configurable? @irenjj

irenjj:
> @irenjj Just forgot to mention that the docs for `copy` statement also needs to be updated.

Sure, Could you help review it?

irenjj:
> Are you willing to take the follow-up task to make this limit configurable? @irenjj

I'd like to, Could you please give me a little hint?

waynexia:
> > Are you willing to take the follow-up task to make this limit configurable? @irenjj
> 
> I'd like to, Could you please give me a little hint?

Filed https://github.com/GreptimeTeam/greptimedb/issues/3889

PR #3823, opened 2024-04-28
feat: support evaluate expr in range query param

Taylor-lagrange:

Support use expr can be evaluate at logical plan stage (e.g. `INTERVAL '2' day - INTERVAL '1' day`)

Support case like:

```sql
SELECT
    ts, 
    min(val) RANGE (now() - (now() - INTERVAL '2' day + INTERVAL '1' day)) 
FROM 
    host
ALIGN
    (now() - (now() - INTERVAL '2' day + INTERVAL '1' day)) 
TO 
    (now() - (now() + INTERVAL '1' hour)) 
by (1) 
ORDER BY ts;
```

- support parse expr in pr  https://github.com/GreptimeTeam/sqlparser-rs/pull/12
- support use `ExprSimplifier` to evaluate expr in range query plan rewrite stage

## Checklist

### DIFF
--- a/src/query/src/range_select/plan_rewrite.rs
+++ b/src/query/src/range_select/plan_rewrite.rs
@@ -112,14 +112,14 @@ fn parse_expr_to_string(args: &[Expr], i: usize) -> DFResult<String> {
 /// Parse a duraion expr:
 /// 1. duration string (e.g. `'1h'`)
 /// 2. Interval expr (e.g. `INTERVAL '1 year 3 hours 20 minutes'`)
-/// 3. An expr can be evaluated at the logical plan stage (e.g. `INTERVAL '2' day - INTERVAL '1' day`)
+/// 3. An interval expr can be evaluated at the logical plan stage (e.g. `INTERVAL '2' day - INTERVAL '1' day`)
 fn parse_duration_expr(args: &[Expr], i: usize) -> DFResult<Duration> {
     match args.get(i) {
         Some(Expr::Literal(ScalarValue::Utf8(Some(str)))) => {
             parse_duration(str).map_err(DataFusionError::Plan)
         }
         Some(expr) => {
-            let ms = evaluate_expr_to_millisecond(args, i)?;
+            let ms = evaluate_expr_to_millisecond(args, i, true)?;
             if ms <= 0 {
                 return Err(dispose_parse_error(Some(expr)));
             }
@@ -131,13 +131,18 @@ fn parse_duration_expr(args: &[Expr], i: usize) -> DFResult<Duration> {
 
 /// Evaluate a time calculation expr, case like:
 /// 1. `INTERVAL '1' day + INTERVAL '1 year 2 hours 3 minutes'`
-/// 2. `now() - INTERVAL '1' day`
+/// 2. `now() - INTERVAL '1' day` (when `interval_only==false`)
 ///
 /// Output a millisecond timestamp
-fn evaluate_expr_to_millisecond(args: &[Expr], i: usize) -> DFResult<i64> {
+///
+/// if `interval_only==true`, only accept expr with all interval type (case 2 will return a error)
+fn evaluate_expr_to_millisecond(args: &[Expr], i: usize, interval_only: bool) -> DFResult<i64> {
     let Some(expr) = args.get(i) else {
         return Err(dispose_parse_error(None));
     };
+    if interval_only && !interval_only_in_expr(expr) {
+        return Err(dispose_parse_error(Some(expr)));
+    }
     let execution_props = ExecutionProps::new().with_query_execution_start_time(Utc::now());
     let info = SimplifyContext::new(&execution_props).with_schema(Arc::new(DFSchema::empty()));
     let interval_to_ms =
@@ -183,7 +188,7 @@ fn evaluate_expr_to_millisecond(args: &[Expr], i: usize) -> DFResult<i64> {
 /// 4. leave empty (as Default Option): align to unix epoch 0 (timezone aware)
 fn parse_align_to(args: &[Expr], i: usize, timezone: Option<&Timezone>) -> DFResult<i64> {
     let Ok(s) = parse_str_expr(args, i) else {
-        return evaluate_expr_to_millisecond(args, i);
+        return evaluate_expr_to_millisecond(args, i, false);
     };
     let upper = s.to_uppercase();
     match upper.as_str() {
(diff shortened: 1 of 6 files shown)


src/query/src/range_select/plan_rewrite.rs
```diff
@@ -108,34 +112,77 @@ fn parse_expr_to_string(args: &[Expr], i: usize) -> DFResult<String> {
...
+        expr.display_name().unwrap_or_default()
+    )))
 }
 
 /// Parse the `align to` clause and return a UTC timestamp with unit of millisecond,
 /// which is used as the basis for dividing time slot during the align operation.
 /// 1. NOW: align to current execute time
 /// 2. Timestamp string: align to specific timestamp
-/// 3. leave empty (as Default Option): align to unix epoch 0 (timezone aware)
+/// 3. A expr can be evaluate at logical plan stage (e.g. `now() - INTERVAL '1' day`)
```
waynexia, line 180:
```suggestion
/// 3. An expr can be evaluated at logical plan stage (e.g. `now() - INTERVAL '1' day`)
```

src/query/src/range_select/plan_rewrite.rs
```diff
@@ -108,34 +112,77 @@ fn parse_expr_to_string(args: &[Expr], i: usize) -> DFResult<String> {
...
+            Ok(Duration::from_millis(ms as u64))
+        }
+        None => Err(dispose_parse_error(None)),
+    }
+}
+
+/// Evaluate a time calculation expr, case like:
+/// 1. `INTERVAL '1' day + INTERVAL '1 year 2 hours 3 minutes'`
+/// 2. `now() - INTERVAL '1' day`
+/// 3. A expr can be evaluate at logical plan stage (e.g. `now() - INTERVAL '1' day`)
```
waynexia, line 135:
```suggestion
/// 3. An expr can be evaluated at logical plan stage (e.g. `now() - INTERVAL '1' day`)
```

src/query/src/range_select/plan_rewrite.rs
```diff
@@ -108,34 +112,77 @@ fn parse_expr_to_string(args: &[Expr], i: usize) -> DFResult<String> {
...
+        Expr::Literal(ScalarValue::IntervalDayTime(interval)) => {
+            interval.map(|v| interval_to_ms(Interval::from_i64(v)))
         }
-        other => Err(dispose_parse_error(other)),
+        Expr::Literal(ScalarValue::IntervalMonthDayNano(interval)) => {
+            interval.map(|v| interval_to_ms(Interval::from_i128(v)))
+        }
+        _ => None,
     }
+    .ok_or(DataFusionError::Plan(format!(
```
evenyag, line 170:
```suggestion
    .ok_or_else(|| DataFusionError::Plan(format!(
```

src/query/src/range_select/plan_rewrite.rs
```diff
@@ -108,34 +112,77 @@ fn parse_expr_to_string(args: &[Expr], i: usize) -> DFResult<String> {
...
+        "{} is not a expr can be evaluate and use in range query",
+        expr.display_name().unwrap_or_default()
+    )))
 }
 
 /// Parse the `align to` clause and return a UTC timestamp with unit of millisecond,
```
evenyag, line 176: out-of-topic: Will it be a problem if we want to process timestamps in other resolutions? e.g. nanosecond

src/query/src/range_select/plan_rewrite.rs
```diff
@@ -108,34 +112,77 @@ fn parse_expr_to_string(args: &[Expr], i: usize) -> DFResult<String> {
 /// Parse a duraion expr:
 /// 1. duration string (e.g. `'1h'`)
 /// 2. Interval expr (e.g. `INTERVAL '1 year 3 hours 20 minutes'`)
+/// 3. An expr can be evaluated at the logical plan stage (e.g. `INTERVAL '2' day - INTERVAL '1' day`)
```
evenyag, line 115: Seems we could use `NOW()` as an interval. This is a little confusing. Should we restrict the expression output must be interval or duration type?

src/query/src/range_select/plan_rewrite.rs
Taylor-lagrange: Nanoseconds are generally not processed. The original design goal should be to provide functions similar to PromQL range queries, so the syntax for specifying time durations also comes from PromQL, which the highest precision is milliseconds. In the execution of range plan, if the underlying time is nanoseconds, it will also be converted to milliseconds for calculation. There may indeed be potential resolution issues

evenyag: Does our document cover this point?

src/query/src/range_select/plan_rewrite.rs
Taylor-lagrange: My original intention is to introduce some computing capabilities in duration param, case like `SELECT ts, min(val) RANGE (INTERVAL '2' day - INTERVAL '1' day) FROM host .....`, but duration string here(e.g. `'1h'`)  is parsed separately. The calculation capability can only be introduced on the interval type（like `INTERVAL '2' day - INTERVAL '1' day`）, but duration lacks this calculation capability (like `'2d' - '1d'`).

src/query/src/range_select/plan_rewrite.rs
Taylor-lagrange: Not show in document. The calculation is compatible with nanosecond timestamp data, but the underlying calculation uses milliseconds as the unit.

Taylor-lagrange: Some parameters of the range query only support to millisecond precision, and the other part can specify a timestamp with any precision, but all the underlying calculation is based on millisecond precision.

evenyag: What if users would like to align to a microsecond or nanosecond timestamp? We allow users to pass timestamps in any resolution in SQL but the output may not as what users expect. It would be confusing without a document.

Taylor-lagrange: I will add some explanation to document

src/query/src/range_select/plan_rewrite.rs
Taylor-lagrange: Now the duration expr only support interval calculation

evenyag (approved):

killme2008:
@waynexia Please take a look.

waynexia (approved):
🚀

PR #3826, opened 2024-04-28
fix: wrong handler implementation of prometheus remote write

waynexia:

To be honest, this part is awful. Four almost identical functions with only two booleans different.

I tried to refactor them with const generic but `axum` doesn't support function handler with const generic yet https://github.com/tokio-rs/axum/pull/1265. Using macro is another option but given this part is not likely to change and macro is also not perfect to my perspective, I leave it as is and only fix the wrong param.

## Checklist

### DIFF
--- a/src/servers/src/http/prom_store.rs
+++ b/src/servers/src/http/prom_store.rs
@@ -34,7 +34,7 @@ use session::context::QueryContextRef;
 use snafu::prelude::*;
 
 use super::header::{write_cost_header_map, GREPTIME_DB_HEADER_METRICS};
-use crate::error::{self, Result, UnexpectedPhysicalTableSnafu};
+use crate::error::{self, Result};
 use crate::prom_store::{snappy_decompress, zstd_decompress};
 use crate::proto::PromWriteRequest;
 use crate::query_handler::{PromStoreProtocolHandlerRef, PromStoreResponse};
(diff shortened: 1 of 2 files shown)


src/servers/src/http/prom_store.rs
```diff
@@ -257,29 +237,6 @@ pub async fn remote_read(
     handler.read(request, query_ctx).await
 }
 
-async fn decode_remote_write_request_to_row_inserts(
```
waynexia, line 260: This seems identical to `decode_remote_write_request`

src/servers/src/http/prom_store.rs
```diff
@@ -72,67 +72,44 @@ impl Default for RemoteWriteQuery {
...
-    let _timer = crate::metrics::METRIC_HTTP_PROM_STORE_WRITE_ELAPSED
-        .with_label_values(&[db.as_str()])
-        .start_timer();
-
-    let is_zstd = content_encoding.contains(VM_ENCODING);
-    let (request, samples) = decode_remote_write_request(is_zstd, body, false).await?;
-    // reject if physical table is specified when metric engine is disabled
-    if params.physical_table.is_some() {
-        return UnexpectedPhysicalTableSnafu {}.fail();
-    }
```
waynexia, line 127: This logic is removed, we should not report error for unrecognized header

tisonkun (approved):
Thank you!

evenyag (approved):

Issue #2931, opened 2023-12-14
information_schema improvements


killme2008:
### What type of enhancement is this?

API improvement, Configuration, User experience, Other

### What does the enhancement do?

Information Schema provides an ANSI-standard way of viewing system metadata. But we only have two internal tables: `tables` and `columns`, that's not enough for compatibility or such as query functions in #2921 .

Tidb has a great doc that describes its information schema: https://docs.pingcap.com/tidb/stable/information-schema

## Tasks

So I created this tracking issue to add more tables into `information_schema` and welcome all developers who are interested in GreptimeDB to contribute:

* [ ] [CHARACTER_SETS](https://docs.pingcap.com/tidb/stable/information-schema-character-sets), we only supports `utf-8`, so we can just return one row.
* [ ] [CHECK_CONSTRAINTS](https://docs.pingcap.com/tidb/stable/information-schema-check-constraints), we don't support check constrais.
* [ ] [COLLATIONS](https://docs.pingcap.com/tidb/stable/information-schema-collations) and [COLLATION_CHARACTER_SET_APPLICABILITY](https://docs.pingcap.com/tidb/stable/information-schema-collation-character-set-applicability), we don't support them yet
* [ ] `COLUMN_PRIVILEGES`, not implemented.
* [ ] `COLUMN_STATISTICS`, not implemented.
* [ ] [ENGINES](https://docs.pingcap.com/tidb/stable/information-schema-engines), we have `mito`, `file` engines etc.
* [ ] `EVENTS` not implemented.
* [ ] `FILES` not implemented.
* [ ] `GLOBAL_STATUS` not implemented.
* [ ] `GLOBAL_VARIABLES` not implemented. An idea is that we keep all configurations in this table.
* [ ] [KEY_COLUMN_USAGE](https://docs.pingcap.com/tidb/stable/information-schema-key-column-usage), need to implement it.
* [ ] `OPTIMIZER_TRACE` not implemented.
* [ ] `PARAMETERS`  not implemented.
* [ ] [PARTITIONS](https://docs.pingcap.com/tidb/stable/information-schema-partitions), we have the partitions info.
* [ ] `PROFILING` not implemented.
* [ ] `REFERENTIAL_CONSTRAINTS`, we don't support it, returns zero rows instead.
* [ ] `ROUTINES` not implemented.
* [ ] [SCHEMATA](https://docs.pingcap.com/tidb/stable/information-schema-schemata), need to implement it
[...]

PR #3832, opened 2024-04-29
feat: adds information_schema cluster_info table

### LINKED ISSUE #1768: Cluster management interface in Dashboard

lack of control panels of cluster

### What does the feature do?

create a cluster management interface in Metasrv, to let users control their cluster in Dashboard

PS: cluster status feature is tracked in issue https://github.com/GreptimeTeam/greptimedb/issues/1131

### Implementation challenges

How Dashboard interacts with cluster? Sending http requests to Frontend or Metasrv?

### LINKED ISSUE #3141: Retrieves cluster metadata via GreptimeDB Cli

User experience

Now, the meta-srv has several HTTP APIs that easily retrieve cluster metadata via HTTP:

e.g.,
```
curl localhost:3002/admin/leader
curl localhost:3002/admin/route?table_id=1024
```

The corresponding HTTP handlers can be found at:
https://github.com/GreptimeTeam/greptimedb/blob/312e8e824ec24448f6cdf449919d0bfbe49d37ea/src/meta-srv/src/service/admin.rs#L37

Let's add the ability to retrieve and print the cluster metadata to CLI.

### Implementation challenges

_No response_

killme2008:

#2931 

Close #1768 and close #3141 

First, rename the `greptime_region_peers` to `region_peers`. It is more appropriate.

Second, adds `cluster_info` table to `information_schema`, it provides the information about the current topology of the cluster.

It depends on https://github.com/GreptimeTeam/greptime-proto/pull/160

```sql
mysql> DESC TABLE CLUSTER_INFO;
+-------------+----------------------+-----+------+---------+---------------+
| Column      | Type                 | Key | Null | Default | Semantic Type |
+-------------+----------------------+-----+------+---------+---------------+
| peer_id     | Int64                |     | NO   |         | FIELD         |
| peer_type   | String               |     | NO   |         | FIELD         |
| peer_addr   | String               |     | YES  |         | FIELD         |
| version     | String               |     | NO   |         | FIELD         |
| git_commit  | String               |     | NO   |         | FIELD         |
| start_time  | TimestampMillisecond |     | YES  |         | FIELD         |
| uptime      | String               |     | YES  |         | FIELD         |
| active_time | String               |     | YES  |         | FIELD         |
+-------------+----------------------+-----+------+---------+---------------+
```

 - `peer_id`: the peer server id.
 -  `peer_type`: the peer type, such as `datanode`, `frontend`, `metasrv` etc.
 - `peer_addr`: the peer gRPC address.
 - `version`: the build package version of the peer.
 - `git_commit`: the build git commit hash of the peer.
 - `start_time`: the starting time of the peer.
 - `uptime`: the uptime of the peer.
- `active_time`: the time since the last activity of the peer.
- 
## For example

In standalone mode:

```sql
mysql> USE INFORMATION_SCHEMA;

mysql> SELECT * FROM CLUSTER_INFO;
+---------+------------+-----------+---------+------------+-------------------------+--------+-------------+
| peer_id | peer_type  | peer_addr | version | git_commit | start_time              | uptime | active_time |
+---------+------------+-----------+---------+------------+-------------------------+--------+-------------+
| 0       | STANDALONE |           | 0.7.2   | 86ab3d9    | 2024-04-30T06:40:02.074 | 18ms   |             |
+---------+------------+-----------+---------+------------+-------------------------+--------+-------------+


```

In standalone mode, the `peer_addr` is always empty and `peer_id` is always `0`.

In distributed mode:

```sql
+---------+-----------+----------------+---------+------------+-------------------------+----------+-------------+
| peer_id | peer_type | peer_addr      | version | git_commit | start_time              | uptime   | active_time |
+---------+-----------+----------------+---------+------------+-------------------------+----------+-------------+
| 1       | DATANODE  | 127.0.0.1:4101 | 0.7.2   | 86ab3d9    | 2024-04-30T06:40:04.791 | 4s 478ms | 1s 467ms    |
| 2       | DATANODE  | 127.0.0.1:4102 | 0.7.2   | 86ab3d9    | 2024-04-30T06:40:06.098 | 3s 171ms | 162ms       |
| 3       | DATANODE  | 127.0.0.1:4103 | 0.7.2   | 86ab3d9    | 2024-04-30T06:40:07.425 | 1s 844ms | 1s 839ms    |
| -1      | FRONTEND  | 127.0.0.1:4001 | 0.7.2   | 86ab3d9    | 2024-04-30T06:40:08.815 | 454ms    | 47ms        |
| 0       | METASRV   | 127.0.0.1:3002 | unknown | unknown    |                         |          |             |
+---------+-----------+----------------+---------+------------+-------------------------+----------+-------------+

```

It will list all the nodes' info in cluster. The `peer_id` in frontends are always `-1`.

## Checklist


Cargo.toml
```diff
@@ -116,7 +116,7 @@ etcd-client = { git = "https://github.com/MichaelScofield/etcd-client.git", rev
 fst = "0.4.7"
 futures = "0.3"
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "aba235025ac5643c12bfdcefd656af11ad58ea8e" }
```
killme2008, line 119: Will be udpated after https://github.com/GreptimeTeam/greptime-proto/pull/160 merged.

src/meta-client/src/client.rs
```diff
@@ -264,18 +264,30 @@ impl ClusterInfo for MetaClient {
 
         let mut nodes = if get_metasrv_nodes {
             let last_activity_ts = -1; // Metasrv does not provide this information.
+
+            // TODO(dennis): Get Metasrv node info
```
killme2008, line 268: Looks like we can't retrieve the metasrv node info in `meta-client`. @fengjiachun

sunng87 (commented):
There is no `peer-id` concept in frontend (and metasrv maybe). Introducing unique id will bring in overall operation and tooling complexity, which we would like to avoid. We can keep it all `0` and it requires explanations from docs.

Another idea is to change `peer_addr` to `hostname`:

- The port number in peer_addr is implemented as grpc service port. However. it makes little sense when we listing those grpc ports from frontend/datanode/metasrv together because they serve different purpose.
- In kubernetes and some other modern environment, hostname offers better readability than IP addresses. Also IP address may change after pod rebuild.

And I wonder if we have sufficient information to include a new field like `state` or `health`.

src/common/meta/src/cluster.rs
```diff
@@ -100,6 +106,19 @@ pub enum NodeStatus {
     Datanode(DatanodeStatus),
     Frontend(FrontendStatus),
     Metasrv(MetasrvStatus),
+    Standalone,
```
sunng87, line 109: consider include these status as a json field in cluster_info.

src/frontend/src/heartbeat.rs
```diff
@@ -30,28 +30,40 @@ use tokio::time::{Duration, Instant};
...
+        opts: &FrontendOptions,
         meta_client: Arc<MetaClient>,
         heartbeat_opts: HeartbeatOptions,
         resp_handler_executor: HeartbeatResponseHandlerExecutorRef,
     ) -> Self {
         HeartbeatTask {
+            // FIXME(dennis): the node_id in FrontendOptions is a string
+            // so we use 0 instead currently.
+            node_id: 0,
+            // We use datanode's start time millis as the node's epoch.
```
sunng87, line 60: this comment is at wrong position

killme2008:
> There is no `peer-id` concept in frontend (and metasrv maybe). Introducing unique id will bring in overall operation and tooling complexity, which we would like to avoid. We can keep it all `0` and it requires explanations from docs.
> 
> Another idea is to change `peer_addr` to `hostname`:
> 
> * The port number in peer_addr is implemented as grpc service port. However. it makes little sense when we listing those grpc ports from frontend/datanode/metasrv together because they serve different purpose.
> * In kubernetes and some other modern environment, hostname offers better readability than IP addresses. Also IP address may change after pod rebuild.
> 
> And I wonder if we have sufficient information to include a new field like `state` or `health`.

1. Agree, frontends don't need the `peer_id` at all, but in datanodes it makes sense, so I like to set all the `peer_id` in frontends to be `-1`.
2. Of course, we have a `last_active_ts` in `NodeInfo`, and we can use it to determine if a peer is alive, but it looks like we don't have it for Metasrv @fengjiachun 
3. Disgree. because in some cases(not k8s env), users may deploy some nodes in the same pod or host, and the `hostname` can't distinguish the peers.

src/common/meta/src/cluster.rs
killme2008: It will be in another table such as `cluster_stats` etc. The `cluster_info` only provides the topology info.

src/meta-client/src/client.rs
fengjiachun: I'm going to add it in next PR.

fengjiachun:
> Of course, we have a last_active_ts in NodeInfo, and we can use it to determine if a peer is alive, but it looks like we don't have it for Metasrv @fengjiachun

Metasrv cannot have a `last_active_ts` since there is no heartbeat from followers to leader.
However, due to the internal impl mechanism of metasrv, if a metasrv node is disconnected for more than a certain period of time, we will no longer be able to see it through the `cluster info` list, which means it will be automatically removed.
That is to say: when you see it, that means it's healthy.

src/meta-client/src/client.rs
killme2008: Ok

killme2008:
@fengjiachun @sunng87 @MichaelScofield Please take a look, thank you.

killme2008:
> > Of course, we have a last_active_ts in NodeInfo, and we can use it to determine if a peer is alive, but it looks like we don't have it for Metasrv @fengjiachun
> 
> Metasrv cannot have a `last_active_ts` since there is no heartbeat from followers to leader. However, due to the internal impl mechanism of metasrv, if a metasrv node is disconnected for more than a certain period of time, we will no longer be able to see it through the `cluster info` list, which means it will be automatically removed. That is to say: when you see it, that means it's healthy.

I added an `active_time` column to represent the time since the last activity of the peer.

src/catalog/src/information_schema/runtime_metrics.rs
```diff
@@ -170,18 +184,19 @@ impl InformationSchemaMetricsBuilder {
                     .join(", "),
                 // Safety: always has a sample
                 ts.samples[0].value,
+                // The peer column is always `None` for standalone
+                None,
+                "standalone",
```
fengjiachun, line 189:
```suggestion
                "STANDALONE",
```
Because here's `STANDALONE`, I think we should unify the naming. https://github.com/GreptimeTeam/greptimedb/blob/97babb0457ec605681d6157fd59c0e694bdc3002/src/common/meta/src/cluster.rs#L119

src/catalog/src/information_schema/utils.rs
```diff
@@ -0,0 +1,53 @@
...
+    let meta_client = match catalog_manager
+        .as_any()
+        .downcast_ref::<KvBackendCatalogManager>()
+    {
+        None => None,
+        Some(manager) => manager.meta_client(),
+    };
+
+    Ok(meta_client)
+}
```
fengjiachun, line 53: Nonblocking:  I'm not sure if this is a good practice. I thought you would pass `Option(ClusterInfo)` when constructing `InformationSchemaProvider`.

https://github.com/GreptimeTeam/greptimedb/blob/76fb99aa1810bebe1b08b00e3eab41c9e2e4ece9/src/catalog/src/kvbackend/manager.rs#L123

src/datanode/src/heartbeat.rs
```diff
@@ -254,6 +256,12 @@ impl HeartbeatTask {
                             region_stats,
                             duration_since_epoch,
                             node_epoch,
+                            info: Some(NodeInfo {
+                                version: build_info.version.to_string(),
+                                git_commit: build_info.commit_short.to_string(),
+                                // The start timestamp is the same as node_epoch currently.
+                                start_time_ms: node_epoch,
```
fengjiachun, line 263: Is this field necessary? We already have `node_epoch`.

src/meta-srv/src/service/cluster.rs
```diff
@@ -128,6 +128,7 @@ impl cluster_server::Cluster for Metasrv {
 
 impl Metasrv {
     pub fn is_leader(&self) -> bool {
-        self.election().map(|x| x.is_leader()).unwrap_or(false)
+        // Returns true when using memory backend.
```
fengjiachun, line 131:
```suggestion
        // Returns true when there is no `election`, indicating the presence of only one `Metasrv` node, which is the leader.
```

src/catalog/src/information_schema/utils.rs
killme2008: It will be used in other places, such as fetching all regions or nodes statistics info, etc.

fengjiachun (approved):
Almost LGTM

tisonkun (approved):
Thank you!

Issue #2806, opened 2023-11-23
Support parallel scan in mito engine


evenyag:
### What type of enhancement is this?

Performance

### What does the enhancement do?

Currently, the Mito engine only supports single-threaded scanning. We can consider using parallel scanning to improve speed when dealing with larger amounts of data.

### Implementation challenges

There are several approaches to parallel scanning:
- Parallel scanning of each file: If there is only one file to scan, parallelization is not possible.
- Parallel scanning of row groups: It is important to note that, in order to ensure the final result is sorted, it is necessary to scan the row groups used by the MergeReader. Implementing this approach can be more complex.

We also need to figure out a way to control the parallelism of the query so spawning a task for each file might not be the best solution (We might need some experiments).

### Steps
- [x] Scan files in parallel https://github.com/GreptimeTeam/greptimedb/pull/2852
- [x] Scan row groups in parallel if the table is append only
  - [x] Build multiple ranges to scan https://github.com/GreptimeTeam/greptimedb/pull/3841
  - [x] Scan multiple ranges https://github.com/GreptimeTeam/greptimedb/pull/3992/
- [x] Scan row groups in parallel for non append only table https://github.com/GreptimeTeam/greptimedb/pull/4060
- [x] Fix the pruning issue for columns with the same name but different column ids https://github.com/GreptimeTeam/greptimedb/pull/3802
- [x] https://github.com/GreptimeTeam/greptimedb/issues/3886

evenyag:
I'm going to reopen this issue as file-level parallelism is not enough if the number of parquet files is less than the parallelism. We still need to implement a more fine-grained parallel scan strategy, such as row group level parallel scan. However, parallel scanning row groups might not be able to maintain a sorted order of the SST file. As a result, we can implement this parallel strategy in append only tables that use `UnorderedScan`.

I did some experiments on this [branch](https://github.com/evenyag/greptimedb/blob/poc/bench-parallel/). I also found that we must support multiple output partitioning to maximize computations.

PR #3841, opened 2024-04-30
feat: Parquet reader builder supports building multiple ranges to read

evenyag:

This PR defines the `FileRange` struct for parquet files and implements a method `build_file_ranges()` to build ranges from a parquet file. A `FileRange` contains a range of rows to read from a parquet file. We can read different ranges in parallel later. Now a `FileRange` is a row group exactly.

```rust
pub struct FileRange {}

impl FileRange {
    pub(crate) async fn reader(&self) -> Result<RowGroupReader> {}
}
```

To reuse code, this PR
- Refactors the `ParquetReader`. Now it adds a `RowGroupReader` to read `Batches` from a row group. The `ParquetReader` invokes the `RowGroupReader` to read the parquet file.
- Defines a struct `FileRangeContext` for all ranges of the same parquet file. This PR also moves the `precise_filter()` method to the context as the context contains all inputs the method needs.

Now the builder uses a method `build_reader_input()` to construct the `FileRangeContext` and row groups to read. Both `build()` and `build_file_ranges` can reuse this method.

```rust
impl ParquetReaderBuilder {
    pub async fn build(&self) -> Result<ParquetReader> {
        let (context, row_groups) = self.build_reader_input().await?;
        ParquetReader::new(context, row_groups).await
    }

    pub async fn build_file_ranges(&self, file_ranges: &mut Vec<FileRange>) -> Result<()> {}

    async fn build_reader_input(&self) -> Result<(FileRangeContextRef, RowGroupMap)> {}
}
```

This PR also fixes some remaining issues and improves the `ReadFormat` helper struct.
- It passes the projection while constructing the struct so it could
  - compute the `projection_indices` in advance
  - compute the `field_id_to_projected_index` in advance. Then `convert_record_batch()` doesn't require `&mut self`. This is necessary for the `FileRangeContext` as we have to share the `ReadFormat`
- It wraps the `SimpleFilterEvaluator` into a `SimpleFilterContext`. The context gets the column info in advance,  from the expected region metadata. This ensures the context can use the correct column id to find the column in the `ReadFormat`

## Checklist

### DIFF
--- a/src/mito2/src/sst/parquet/reader.rs
+++ b/src/mito2/src/sst/parquet/reader.rs
@@ -18,11 +18,14 @@ use std::collections::{BTreeMap, VecDeque};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
+use api::v1::SemanticType;
 use async_trait::async_trait;
 use common_recordbatch::filter::SimpleFilterEvaluator;
 use common_telemetry::{debug, warn};
 use common_time::range::TimestampRange;
+use datafusion_expr::Expr;
 use datatypes::arrow::record_batch::RecordBatch;
+use datatypes::data_type::ConcreteDataType;
 use itertools::Itertools;
 use object_store::ObjectStore;
 use parquet::arrow::arrow_reader::{ParquetRecordBatchReader, RowSelection};
@@ -155,20 +158,21 @@ impl ParquetReaderBuilder {
         let parquet_meta = self.read_parquet_metadata(&file_path, file_size).await?;
         // Decodes region metadata.
         let key_value_meta = parquet_meta.file_metadata().key_value_metadata();
-        let region_meta = Self::get_region_metadata(&file_path, key_value_meta)?;
-        // Always list all column ids to read.
+        // Gets the metadata stored in the SST.
+        let region_meta = Arc::new(Self::get_region_metadata(&file_path, key_value_meta)?);
+        // Lists all column ids to read, we always use the expected metadata if possible.
         let column_ids = self.projection.clone().unwrap_or_else(|| {
-            let metadata = self
+            let expected_meta = self
                 .expected_metadata
-                .as_deref()
+                .as_ref()
                 .unwrap_or_else(|| &region_meta);
-            metadata
+            expected_meta
                 .column_metadatas
                 .iter()
                 .map(|col| col.column_id)
                 .collect()
         });
-        let read_format = ReadFormat::new(Arc::new(region_meta), &column_ids);
+        let read_format = ReadFormat::new(region_meta.clone(), &column_ids);
 
         // Computes the projection mask.
         let parquet_schema_desc = parquet_meta.file_metadata().schema_descr();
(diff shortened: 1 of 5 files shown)


src/mito2/src/sst/parquet/reader.rs
```diff
@@ -147,10 +146,29 @@ impl ParquetReaderBuilder {
...
     /// This needs to perform IO operation.
     pub async fn build(&self) -> Result<ParquetReader> {
+        let (context, row_groups) = self.build_reader_input().await?;
+        ParquetReader::new(context, row_groups).await
+    }
+
+    /// Builds [FileRange]s to read and pushes them to `file_ranges`.
+    #[allow(dead_code)]
+    pub async fn build_file_ranges(&self, file_ranges: &mut Vec<FileRange>) -> Result<()> {
+        let (context, row_groups) = self.build_reader_input().await?;
```
v0y4g3r, line 160: nit: maybe reserve for extra space

```
        file_ranges.reserve_exact(row_groups.len());
```

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -159,18 +177,24 @@ impl ParquetReaderBuilder {
...
         let parquet_schema_desc = parquet_meta.file_metadata().schema_descr();
-        let projection_mask = if let Some(column_ids) = self.projection.as_ref() {
-            let indices = read_format.projection_indices(column_ids.iter().copied());
-            // Now we assumes we don't have nested schemas.
-            ProjectionMask::roots(parquet_schema_desc, indices)
-        } else {
-            ProjectionMask::all()
-        };
+        let indices = read_format.projection_indices();
+        // Now we assumes we don't have nested schemas.
```
v0y4g3r, line 196: Add a todo here since it will soon be a json type

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -159,18 +177,24 @@ impl ParquetReaderBuilder {
...
+        // Lists all column ids to read, we always use the expected metadata if possible.
+        let column_ids = self.projection.clone().unwrap_or_else(|| {
+            let expected_meta = self.expected_metadata.as_ref().unwrap_or(&region_meta);
+            expected_meta
+                .column_metadatas
+                .iter()
+                .map(|col| col.column_id)
+                .collect()
+        });
+        let read_format = ReadFormat::new(region_meta.clone(), &column_ids);
```
v0y4g3r, line 191: Hopefully we can eliminate a clone
```suggestion
        let read_format = if let Some(column_ids) = &self.projection {
            ReadFormat::new(region_meta.clone(), column_ids)
        } else {
            let expected_meta = self.expected_metadata.as_ref().unwrap_or(&region_meta);
            let column_ids = expected_meta
                .column_metadatas
                .iter()
                .map(|col| col.column_id)
                .collect::<Vec<_>>();
            ReadFormat::new(region_meta.clone(), &column_ids)
        };
```

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -513,166 +534,293 @@ impl RowGroupReaderBuilder {
...
+                return Ok(None);
             };
+            self.metrics.num_record_batches += 1;
 
-            // Sets current reader to this reader.
-            self.current_reader = Some(row_group_reader);
-            return Ok(Some(record_batch));
+            self.context
+                .read_format()
+                .convert_record_batch(&record_batch, &mut self.batches)?;
```
v0y4g3r, line 794: Can we concatenate the pruned batches into one batch, so that we can use an `Option` to hold the resulting batch and avoid yielding many fragmented batches?

evenyag: Concatenating batches might be slower than returning those batches directly.

fengjiachun (approved):
LGTM

v0y4g3r (approved):
LGTM

PR #3845, opened 2024-04-30
perf: optimize `count(*)`

waynexia:

TL;DR: 5x

Before

```sql
MySQL [(none)]> select count(*) from phy;
+------------+
| COUNT(*)   |
+------------+
| 2582870657 |
+------------+
1 row in set (4 min 38.347 sec)
```

After

```sql
MySQL [(none)]> select count(*) from phy;
+------------+
| COUNT(*)   |
+------------+
| 2582870657 |
+------------+
1 row in set (50.387 sec)
```

This patch adds a new optimizer rule that converts `count(*)` to `count(<TIME INDEX>)`. This optimization is based on the fact that our underlying storage engine scans faster on time index column than primary key column. Reading time index column does not need decoding phase like in primary key column. This rule is extended from the one from datafusion and overrides it.

This patch also changes some logic in range plan, making it able to handle aggr expr with alias.

## Checklist


tisonkun (approved):

evenyag (approved):
LGTM

PR #3846, opened 2024-04-30
fix: retrieve all info instead of checking on demand

WenyXu:

`RegionLeaseKeeper` should retrieve all operating regions in one operation instead of checking on demand to avoid inconsistencies caused by on-demand checking.

## Checklist

### DIFF
--- a/src/common/meta/src/region_keeper.rs
+++ b/src/common/meta/src/region_keeper.rs
@@ -83,8 +83,8 @@ impl MemoryRegionKeeper {
         inner.contains(&(datanode_id, region_id))
     }
 
-    /// Filters out all operating regions from `region_ids` and returns operating regions.
-    pub fn filter_operating_regions(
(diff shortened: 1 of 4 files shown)


MichaelScofield (approved):

src/common/meta/src/region_keeper.rs
```diff
@@ -83,16 +83,18 @@ impl MemoryRegionKeeper {
         inner.contains(&(datanode_id, region_id))
     }
 
-    /// Returns a set of filtered out regions that are opening.
-    pub fn filter_opening_regions(
+    /// Filters out all operating regions from `region_ids` and returns operating regions.
+    pub fn filter_operating_regions(
```
fengjiachun, line 87: Non-blocking: How about rename it as `extract_operating_regions`? Because I found that this method seems to have done the opposite of what it originally did after being modified.

fengjiachun (approved):

PR #3847, opened 2024-04-30
fix: count_wildcard_to_time_index_rule doesn't handle table reference properly

waynexia:

Fix regressions from https://github.com/GreptimeTeam/greptimedb/pull/3845

Use `TableReference` to avoid unexpected canonicalization. And check if the time index column is valid at the plan to use it.

## Checklist


killme2008 (approved):
LGTM. I need a coffee, you are welcome :D.

PR #2911, opened 2023-12-12
refactor: make instance started separately

MichaelScofield:
I hereby agree to the terms of the [GreptimeDB CLA](https://gist.github.com/xtang/6378857777706e568c1949c7578592cc)

## What's changed and what's your intention?

to support further integrated into other binaries

## Checklist

- [ ]  I have written the necessary rustdoc comments.
- [ ]  I have added the necessary unit tests and integration tests.

PR #3850, opened 2024-05-01
refactor: add the main cli entry point struct `Command{}` and simplify `main()`

zyy17:

The parsing entry of the current command line is a bit messy and I think we need a cleaner and more unified entry. 

- Add the `Command{}` struct: Unify all the cli options, including subcommand and global options. When we load the cli option, use `Command::parse()`;

- Make the `main()` cleaner by using the following sub-functions:

  - Add `setup_human_panic()`;
  - Add `start()` to load the options and start the service;

- Rename `CliOptions` to `GlobalOptions`;

Reference: [`uv`](https://github.com/astral-sh/uv/blob/main/crates/uv/src/cli.rs#L21-L43) has a clean entry point.

## Checklist

### DIFF
--- a/src/cmd/src/lib.rs
+++ b/src/cmd/src/lib.rs
@@ -15,7 +15,6 @@
 #![feature(assert_matches, let_chains)]
 
 use async_trait::async_trait;
-use clap::arg;
 use common_telemetry::{error, info};
 
 pub mod cli;
@@ -77,15 +76,6 @@ pub fn log_versions(version_string: &str, app_version: &str) {
     info!("GreptimeDB version: {}", version_string);
 
     log_env_flags();
-}
-
-pub fn greptimedb_cli() -> clap::Command {
-    let cmd = clap::Command::new("greptimedb").subcommand_required(true);
-
-    #[cfg(feature = "tokio-console")]
-    let cmd = cmd.arg(arg!(--"tokio-console-addr"[TOKIO_CONSOLE_ADDR]));
-
-    cmd.args([arg!(--"log-dir"[LOG_DIR]), arg!(--"log-level"[LOG_LEVEL])])
 }
 
 fn log_env_flags() {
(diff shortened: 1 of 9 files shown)


tisonkun (approved):
Good refactor! Thank you.

src/cmd/src/lib.rs
```diff
@@ -79,15 +78,6 @@ pub fn log_versions(version_string: &str, app_version: &str) {
...
 }
 
-pub fn greptimedb_cli() -> clap::Command {
-    let cmd = clap::Command::new("greptimedb").subcommand_required(true);
-
-    #[cfg(feature = "tokio-console")]
-    let cmd = cmd.arg(arg!(--"tokio-console-addr"[TOKIO_CONSOLE_ADDR]));
-
-    cmd.args([arg!(--"log-dir"[LOG_DIR]), arg!(--"log-level"[LOG_LEVEL])])
-}
```
evenyag, line 89: We switched to operating `Command` directly in https://github.com/GreptimeTeam/greptimedb/pull/2911

Does this prevent us from extending arguments or using the command in other binaries? @MichaelScofield Please take a look.

zyy17: IMO, some original args(`--log-level`, `--log-dir`, `--tokio-console-addr`) should be global. The PR will take those args global explicitly. We can still extend the args by adding the new fields into the related struct.

MichaelScofield: My intention in #2911 was to create `clap::Command` dynamically, to use different sub commands. So is it possible to make the new `Command` owned a dyn `SubCommand`?

zyy17: What's the scenario of #2911, or how can I test the cli from a user's perspective? I think this PR is equivalent based on #2911.

MichaelScofield: See the codes here https://github.com/GreptimeTeam/greptimedb-enterprise/blob/develop/projects/dynamo/src/bin/greptime.rs

MichaelScofield: ok to me

MichaelScofield (approved):

PR #3853, opened 2024-05-02
refactor: add `etcd_connect_options` for metasrv

zyy17:

The original `create_etcd_client()` uses `None` for [`ConnectOptions`](https://docs.rs/etcd-client/latest/etcd_client/struct.ConnectOptions.html) and can't configure some important arguments for connecting etcd.

In this PR, I created the `EtcdConnectOptions`, a subset of `ConnectOptions`, in the `MetasrvOptions`. I also added the covert function to let `EtcdConnectOptions` convert `ConnectOptions` easily.

btw, it's convenient to add TLS configurations in the future in the same structure(I will add it in the next PR). 

## Checklist

### DIFF
--- a/src/meta-srv/src/metasrv.rs
+++ b/src/meta-srv/src/metasrv.rs
@@ -109,6 +109,10 @@ pub struct MetasrvOptions {
     /// limit the number of operations in a txn because an infinitely large txn could
     /// potentially block other operations.
     pub max_txn_ops: usize,
+
+    /// The etcd connect options.
+    /// Most of the options are from `etcd_client::ConnectOptions`.
+    pub etcd_connect_options: EtcdConnectOptions,
 }
 
 impl MetasrvOptions {
@@ -146,6 +150,7 @@ impl Default for MetasrvOptions {
             export_metrics: ExportMetricsOption::default(),
             store_key_prefix: String::new(),
             max_txn_ops: 128,
+            etcd_connect_options: EtcdConnectOptions::default(),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/meta-srv/src/metasrv.rs
```diff
@@ -109,6 +109,10 @@ pub struct MetasrvOptions {
     /// limit the number of operations in a txn because an infinitely large txn could
     /// potentially block other operations.
     pub max_txn_ops: usize,
+
+    /// The etcd connect options.
+    /// Most of the options are from `etcd_client::ConnectOptions`.
+    pub etcd_connect_options: EtcdConnectOptions,
```
tisonkun, line 115: This option itself can be `Option`? Or we define a `etcd` subsection for metasrv. Extending this option looks like we leak/couple the metasrv with etcd implementation.

zyy17: It seems reasonable. It also reminds me that there is more than one option for metasrv backend storage(file system, memory, etcd, ...). I think we need the `BackendStorageProvider` and etcd is one of the providers. 

For example:

```toml
[storage]
type = "etcd"
addr = "127.0.0.1"
timeout = "5s"
```

What do you think? @fengjiachun

fengjiachun: > It seems reasonable. It also reminds me that there is more than one option for metasrv backend storage in the future(file system, memory, etcd, ...). I think we need the `BackendStorageProvider` and etcd is one of the providers.
> 
> For example:
> 
> ```toml
> [storage]
> type = "etcd"
> addr = "127.0.0.1:2379"
> timeout = "5s"
> ```
> 
> What do you think? @fengjiachun

+1

zyy17:
Let me refactor again for adding the `BackendStorageProvider`.

zyy17:
Close it temporarily. I will re-open it when the PR is ready for review.

PR #3858, opened 2024-05-06
feat(flow): mirror insert req to flow node

discord9:

- Add a path in `do_request` of `Inserter` to allow copy source table insert requests and send to flow node
- Also added `TableFlowManager` field in `Inserter` to allow query which insert request should go which flownode, but since we only have standalone flownode for now, they are mostly blanket impl
- Some minor refactor, including `Flownode` `handle_request` change to `handle_requests` for batching request, and add XXXRef for Flow related managers.

## Checklist


src/operator/src/insert.rs
```diff
@@ -231,16 +260,37 @@ impl Inserter {
     async fn group_requests_by_peer(
         &self,
         requests: RegionInsertRequests,
-    ) -> Result<HashMap<Peer, RegionInsertRequests>> {
-        let mut inserts: HashMap<Peer, RegionInsertRequests> = HashMap::new();
+    ) -> Result<HashMap<PeerTyped, RegionInsertRequests>> {
+        let mut inserts: HashMap<PeerTyped, RegionInsertRequests> = HashMap::new();
 
         for req in requests.requests {
```
zhongzc, line 266: Can we group requests in `table_id` for flow-grouping and in `region_id` for datanode-grouping first to reduce calling to `nodes` and `find_region_leader`?

src/common/meta/src/key/flow/table_flow.rs
```diff
@@ -188,7 +190,9 @@ impl TableFlowManager {
     }
 
     /// Retrieves all [TableFlowKey]s of the specified `table_id`.
-    pub fn nodes(&self, table_id: TableId) -> BoxStream<'static, Result<TableFlowKey>> {
+    ///
+    /// TODO(discord9): add cache for it since range request does not support cache.
```
fengjiachun, line 194: Seems we can cache all metadata of flow, like what we have done with table metadata. cc @WenyXu

zhongzc (approved):

src/operator/src/insert.rs
```diff
@@ -228,19 +257,88 @@ impl Inserter {
...
+        for req in &requests.requests {
+            match src_table_reqs.get_mut(&RegionId::from_u64(req.region_id).table_id()) {
+                Some(Some((_peers, reqs))) => reqs.requests.push(req.clone()),
+                // already know this is not source table
+                Some(None) => continue,
+                _ => {
+                    let table_id = RegionId::from_u64(req.region_id).table_id();
+                    let peers = self
+                        .table_flow_manager
+                        .flows(table_id)
```
fengjiachun, line 277: Trying to obtain flows within a for loop will be very slow. 🤔

src/common/meta/src/key/flow/table_flow.rs
fengjiachun: Oh, now we can only cache it on the metasrv leader. It seems that the frontend needs this cache now.

src/operator/src/insert.rs
fengjiachun: Perhaps we can add cache in the next PR to solve this performance bottleneck.

fengjiachun (approved):

PR #3859, opened 2024-05-06
fix: register regions during procedure recovery

WenyXu:

Ensures all operating regions are registered during procedure recovery.

## Checklist


src/common/meta/src/ddl/drop_database.rs
```diff
@@ -105,7 +116,13 @@ impl DropDatabaseProcedure {
                 tables: None,
             },
             state,
-        })
+        };
+        procedure
```
fengjiachun, line 120: Perhaps we should let the framework call `on_recover` (after executing `from_json`).

WenyXu:
@fengjiachun @MichaelScofield PTAL

fengjiachun (approved):
LGTM

MichaelScofield (approved):

PR #3862, opened 2024-05-06
refactor: move `Plugins` to `plugins.rs` and use rwlock

zyy17:

It's a minor refactoring for `Plugins` to make it clean.

In this PR:

1. Move `Plugins` from `lib.rs` to the new mod file `plugins.rs`;
2. Add some simple and necessary unit tests for `Plugins`;
3. Use `RwLock` instead of `Mutex` to improve performance;

## Checklist

### DIFF
--- a/src/common/base/src/plugins.rs
+++ b/src/common/base/src/plugins.rs
@@ -13,40 +13,35 @@
 // limitations under the License.
 
 use std::any::Any;
-use std::sync::{Arc, Mutex, MutexGuard};
+use std::sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard};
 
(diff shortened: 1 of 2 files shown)


src/common/base/src/plugins.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    pub fn insert<T: 'static + Send + Sync>(&self, value: T) {
+        let _ = self.lock().insert(value);
+    }
+
+    pub fn get<T: 'static + Send + Sync + Clone>(&self) -> Option<T> {
+        let binding = self.lock();
+        binding.get::<T>().cloned()
+    }
```
shuiyisong, line 43: Now that we look at it, can we change it to a read-write lock for better performance? cc @paomian

zyy17: I already modify the plugins for using rwlock. PTAL.

sunng87 (approved):

tisonkun (approved):

PR #3864, opened 2024-05-06
fix: return metric name instead of query in Prometheus /series API

waynexia:

The previous implement returns the whole query instead of metric name in result. This patch fixes this behavior

image

## Checklist

### DIFF
--- a/tests-integration/tests/http.rs
+++ b/tests-integration/tests/http.rs
@@ -508,10 +508,7 @@ pub async fn test_prom_http_api(store_type: StorageType) {
         .collect::<BTreeMap<String, String>>();
     let expected = BTreeMap::from([
         ("__name__".to_string(), "demo".to_string()),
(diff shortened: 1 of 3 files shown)


killme2008 (approved):
LGTM, but could you assert it in the test?

yuanbohan:
BTW, the `field` and `timestamp` columns SHOULD BE filtered from the response.

tests-integration/tests/http.rs
```diff
@@ -492,7 +492,7 @@ pub async fn test_prom_http_api(store_type: StorageType) {
 
     // series
     let res = client
-        .get("/v1/prometheus/api/v1/series?match[]=demo&start=0&end=0")
+        .get("/v1/prometheus/api/v1/series?match[]=demo{}&start=0&end=0")
```
waynexia, line 495: Test to verify this change @killme2008

yuanbohan (approved):

PR #3875, opened 2024-05-07
refactor(logstore): remove Entry::namemspace

v0y4g3r:

- Remove `Entry::namespace` method and related asscoiated type.
- Fix the estimated size logic to computing the encoded length of entries.

## Checklist

### DIFF
--- a/src/log-store/src/kafka.rs
+++ b/src/log-store/src/kafka.rs
@@ -57,7 +57,6 @@ pub struct EntryImpl {
 
 impl Entry for EntryImpl {
     type Error = Error;
-    type Namespace = NamespaceImpl;
 
     fn data(&self) -> &[u8] {
         &self.data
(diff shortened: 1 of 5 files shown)


evenyag (approved):

tisonkun (commented):
This is `Entry::namespace()`?

tisonkun:
And `create_namespace`, `delete_namespace`, `list_namespaces` is unused also.

I suppose if we'd want to use the namespace encapsulation, we should use the Namespace type/trait, instead of `pub ns: NamespaceImpl`.

tisonkun (approved):
LGTM. Although I'd prefer to declare our plan to the Namespace abstraction first and avoid directly access `ns` field if we'd hold this abstraction.

BTW, `NoopLogStore` is never used even in tests. We can remove it in a following PR since it is totally useless (the associated tests is noop actually).

Issue #3869, opened 2024-05-07
Column types doesn't match schema types


WenyXu:
Currently, I am testing the deletion of a row and have encountered a peculiar issue：

- fuzz test case:
```sql
CREATE TABLE `esT`(
  `eT` TIMESTAMP(3) TIME INDEX,
  `eAque` BOOLEAN,
  `DoLOruM` INT,
  `repudiAndae` FLOAT,
  `ULLaM` BOOLEAN,
  `COnSECTeTuR` SMALLINT DEFAULT -31852,
  `DOLOrIBUS` FLOAT NOT NULL,
  `QUiS` SMALLINT NULL,
  `consEquatuR` BOOLEAN NOT NULL,
  `vERO` BOOLEAN,
  PRIMARY KEY(`repudiAndae`, `ULLaM`, `DoLOruM`)
) ENGINE = mito;
```

- DDL rows:

```sql
INSERT INTO
  `esT` (
    `consEquatuR`,
    `eAque`,
    `eT`,
    `repudiAndae`,
    `DOLOrIBUS`
  )
VALUES
  (
    false,
    false,
    '+234049-06-04 01:11:41.163+0000',
    0.2339946,
    0.97377783
  ),
  (
    false,
    true,
    '-19578-12-20 11:45:59.875+0000',
    NULL,
    0.3535998
  );
  
  DELETE FROM
  `esT`
WHERE
  `DOLOrIBUS` = 0.73806196
  AND `consEquatuR` = false;
```

- error: 

```txt
Failed to execute query: DELETE FROM `esT` WHERE `DOLOrIBUS` = 0.73806196 AND `consEquatuR` = false err=0: Table operation error, at src/frontend/src/instance.rs:302:14
1: Failed to execute logical plan, at /Users/hanfox/Desktop/code/contribute/greptimedb/src/operator/src/statement.rs:269:14
2: Failed to create RecordBatch, at src/query/src/datafusion.rs:134:31
3: , at src/common/recordbatch/src/adapter.rs:254:55
4: External(0: , at src/common/recordbatch/src/adapter.rs:254:55
1: External(0: , at src/common/recordbatch/src/adapter.rs:254:55
1: External(External(0: Fail to create datafusion record batch, at /Users/hanfox/Desktop/code/contribute/greptimedb/src/common/recordbatch/src/recordbatch.rs:51:14
1: InvalidArgumentError("column types must match schema types, expected Float32 but found Int32 at column index 3")))))
```

[...]

PR #3876, opened 2024-05-07
fix: add data type to vector cache key

evenyag:

- fixes https://github.com/GreptimeTeam/greptimedb/issues/3869

This PR adds data type to the cache key for repeated vectors. It fixes the issue that different data types use the same null vector in the cache. When the reader gets a null vector from the cache, the data type of the vector might not be the expected type and result in the following error: `column types must match schema types`.

## Checklist


v0y4g3r (approved):
LGTM

fengjiachun (approved):
LGTM

PR #3880, opened 2024-05-08
feat: support querying field column names in Prometheus HTTP API

waynexia:

The extended special matcher `__field__` is used to specify which field to query in multi-field scenario. This patch adds support for querying the possible values of `__field__`

It works like 
image

## Checklist

### DIFF
--- a/src/servers/src/http/prometheus.rs
+++ b/src/servers/src/http/prometheus.rs
@@ -49,7 +49,7 @@ use crate::error::{
     UnexpectedResultSnafu,
 };
 use crate::http::header::collect_plan_metrics;
-use crate::prom_store::METRIC_NAME_LABEL;
+use crate::prom_store::{FIELD_NAME_LABEL, METRIC_NAME_LABEL};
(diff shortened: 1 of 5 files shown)


yuanbohan (approved):

src/servers/src/http/prometheus.rs
```diff
@@ -690,6 +690,41 @@ pub async fn label_values_query(
...
         table_names.sort_unstable();
         return PrometheusJsonResponse::success(PrometheusResponse::LabelValues(table_names));
+    } else if label_name == FIELD_NAME_LABEL {
+        let mut field_columns = HashSet::new();
+        let mut matches = params.matches.0;
+        // query all tables if no matcher is provided
+        if matches.is_empty() {
+            matches = match handler
+                .catalog_manager()
+                .table_names(&catalog, &schema)
```
fengjiachun, line 700: We have `tables` API for fetching all tables info : https://h6tbhdl56ey8.cn-hangzhou.alicloud.cloud.greptime.cn/playground/

PR #3901, opened 2024-05-10
feat: limiting the size of query results to Dashboard

### LINKED ISSUE #3716: Limiting the size of query results to Dashboard

User experience

When the query results is huge, Dashboard could be easily crashed. For example, when user accidently `select * from` a table with millions of rows. Dashboard is mainly for humans, returning too many rows is of no use. We can limit the returned rows to 1000, iff the query comes from Dashboard.

### Implementation challenges

_No response_

realtaobo:

close #3716

Limit the returned rows to 1000, iff the query comes from Dashboard.

Adding a new `source enum` to the `SqlQuery` to identify if it's a request from DashBorad or not, and then handle the flow inside the `src/servers/src/http/handler.rs:72`.  In addition, considering that the dashboard only supports `greptimedb_v1` format, only the output of `greptimedb_v1` format outputs is processed this PR.

## Checklist

### DIFF
--- a/src/servers/src/http.rs
+++ b/src/servers/src/http.rs
@@ -336,6 +336,20 @@ impl Display for Epoch {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RequestSource {
+    Dashboard,
+}
+
+impl RequestSource {
+    pub fn parse(s: &str) -> Option<Self> {
+        match s {
+            "dashboard" => Some(RequestSource::Dashboard),
+            _ => None,
+        }
+    }
+}
+
 #[derive(Serialize, Deserialize, Debug, JsonSchema)]
 pub enum HttpResponse {
     Arrow(ArrowResponse),
--- a/src/servers/src/http.rs
+++ b/src/servers/src/http.rs
@@ -15,7 +15,6 @@
 use std::collections::HashMap;
 use std::fmt::Display;
 use std::net::SocketAddr;
-use std::str::FromStr;
 use std::sync::Mutex as StdMutex;
 use std::time::Duration;
 
@@ -348,15 +347,6 @@ impl Display for Epoch {
 pub enum RequestSource {
     #[strum(ascii_case_insensitive)]
     Dashboard,
-}
-
-impl RequestSource {
-    pub fn parse(s: &str) -> Option<Self> {
-        match RequestSource::from_str(s) {
-            Ok(v) => Some(v),
-            _ => None,
-        }
-    }
 }
 
 #[derive(Serialize, Deserialize, Debug, JsonSchema)]
--- a/src/servers/src/http.rs
+++ b/src/servers/src/http.rs
@@ -15,6 +15,7 @@
 use std::collections::HashMap;
 use std::fmt::Display;
 use std::net::SocketAddr;
+use std::str::FromStr;
 use std::sync::Mutex as StdMutex;
 use std::time::Duration;
 
@@ -41,6 +42,7 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use snafu::{ensure, ResultExt};
+use strum::EnumString;
 use tokio::sync::oneshot::{self, Sender};
 use tokio::sync::Mutex;
 use tower::timeout::TimeoutLayer;
@@ -190,6 +192,10 @@ impl From<SchemaRef> for OutputSchema {
 pub struct HttpRecordsOutput {
     schema: OutputSchema,
     rows: Vec<Vec<Value>>,
+    // total_rows is equal to rows.len() in most cases,
+    // the Dashboard query result may be truncated, so we need to return the total_rows.
+    #[serde(default)]
+    total_rows: usize,
 
     // plan level execution metrics
     #[serde(skip_serializing_if = "HashMap::is_empty")]
@@ -224,6 +230,7 @@ impl HttpRecordsOutput {
             Ok(HttpRecordsOutput {
                 schema: OutputSchema::from(schema),
                 rows: vec![],
+                total_rows: 0,
                 metrics: Default::default(),
             })
         } else {
@@ -244,6 +251,7 @@ impl HttpRecordsOutput {
 
             Ok(HttpRecordsOutput {
                 schema: OutputSchema::from(schema),
+                total_rows: rows.len(),
                 rows,
                 metrics: Default::default(),
             })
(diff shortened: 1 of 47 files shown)


src/servers/src/http/handler.rs
```diff
@@ -117,17 +123,49 @@ pub async fn sql(
...
+        resp = from_dashboard(resp, query_ctx).await;
+    }
     resp.with_execution_time(start.elapsed().as_millis() as u64)
 }
 
+pub async fn from_dashboard(resp: HttpResponse, ctx: QueryContextRef) -> HttpResponse {
+    let HttpResponse::GreptimedbV1(response) = resp else {
+        return resp;
+    };
+    let GreptimedbV1Response {
```
MichaelScofield, line 144: I think we should add a field of `total_rows: u32` in `GreptimedbV1Response`, to inform the Dashboard there might be some rows "limit"ed.

src/servers/src/http/handler.rs
```diff
@@ -93,12 +95,16 @@ pub async fn sql(
         .or(form_params.epoch)
         .map(|s| s.to_lowercase())
         .map(|s| Epoch::parse(s.as_str()).unwrap_or(Epoch::Millisecond));
+    let source = query_params
+        .source
+        .map(|s| RequestSource::parse(s.to_lowercase().as_str()))
+        .unwrap_or(None);
```
evenyag, line 101:
```suggestion
        .source
        .and_then(|s| RequestSource::parse(s.to_lowercase().as_str()));
```

src/servers/src/http/handler.rs
```diff
@@ -117,17 +123,49 @@ pub async fn sql(
...
+
+    let mut outputs = Vec::with_capacity(output.len());
+    for query_result in output {
+        let GreptimeQueryOutput::Records(mut records) = query_result else {
+            outputs.push(query_result);
+            continue;
+        };
+        if records.rows.len() > query_limit {
+            records.rows.resize(query_limit, vec![]);
+        }
```
evenyag, line 159: We should truncate the result instead of resizing it.
```suggestion
        records.rows.truncate(query_limit);
```

src/servers/src/http.rs
```diff
@@ -336,6 +336,20 @@ impl Display for Epoch {
...
+}
+
+impl RequestSource {
+    pub fn parse(s: &str) -> Option<Self> {
+        match s {
+            "dashboard" => Some(RequestSource::Dashboard),
+            _ => None,
+        }
+    }
+}
```
evenyag, line 351: Consider using strum `EnumString` and the `#[strum(ascii_case_insensitive)]` attribute. Then we could get rid of the `to_lower_case()` call.
https://docs.rs/strum/latest/strum/derive.EnumString.html

src/servers/src/http/handler.rs
```diff
@@ -62,6 +62,8 @@ pub struct SqlQuery {
     // specified time precision. Maybe greptimedb format can support this
     // param too.
     pub epoch: Option<String>,
+    // request source, ["dashboard"]
```
zyy17, line 65: question: how to identify the requests are from dashboard?

src/servers/src/http/handler.rs
realtaobo: To be precise, it should be added to `HttpRecordsOutput`.  It corresponds to the execution result of each statement.
https://github.com/GreptimeTeam/greptimedb/blob/d0820bb26dfd6962c7a2f92980d90ee25841280c/src/servers/src/http.rs#L190-L192

src/servers/src/http/handler.rs
realtaobo: Obviously, we must modify the dashboard request body, add a new arg `source`. Maybe https://github.com/GreptimeTeam/greptimedb/issues/3716#issuecomment-2105590075 more reasonable？

zyy17: I thought there was an identifier in the request body before.

If there is no identifier in the original request(which means we need to modify the dashboard anyway), my advice of adding the parameters for limiting query results may be reasonable. 

The implementation may be more straightforward than the current PR. For example:

- Add `limit` parameters:

  ```
  curl -X POST \
    -H 'Authorization: Basic {{authentication}}' \
    -H 'Content-Type: application/x-www-form-urlencoded' \
    -d 'sql={{SQL-statement}}' \
    http://{{API-host}}/v1/sql?db={{db-name}}&limit=1000
  ```

- Parse the parameter and handle the query(most of the logic are same as the current PR).

@realtaobo @MichaelScofield

realtaobo: > If there is no identifier in the original request(which means we need to modify the dashboard anyway), my advice of adding the parameters for limiting query results may be reasonable.

Sure, this is a design decision. It hinges on whether future projects will leverage the request source for additional tasks, like data statistics. Both approaches are valid, so the optimal choice depends on project's development roadmap.

src/servers/src/http.rs
```diff
@@ -336,6 +344,21 @@ impl Display for Epoch {
...
+
+impl RequestSource {
+    pub fn parse(s: &str) -> Option<Self> {
+        match RequestSource::from_str(s) {
+            Ok(v) => Some(v),
+            _ => None,
+        }
+    }
+}
```
evenyag, line 360:
```suggestion
```

src/servers/src/http/handler.rs
```diff
@@ -93,12 +95,15 @@ pub async fn sql(
         .or(form_params.epoch)
         .map(|s| s.to_lowercase())
         .map(|s| Epoch::parse(s.as_str()).unwrap_or(Epoch::Millisecond));
+    let source = query_params
+        .source
+        .and_then(|s| RequestSource::parse(s.as_str()));
```
evenyag, line 100:
```suggestion
        .and_then(|s| RequestSource::from_str(s.as_str()).ok());
```

evenyag (approved):
We don't need to define the `parse()` method which is equivalent to `from_str().ok()`.

src/servers/src/http/handler.rs
evenyag: A `limit` parameter is reasonable. The dashboard could also add an identifier in its request as it might be useful.

MichaelScofield: Ok, add a "limit" parameter is fine by me. However, I suggest adding a "total_rows" field in the response. Otherwise user could be misguided.

realtaobo: > I suggest adding a "total_rows" field in the response. Otherwise user could be misguided.

@MichaelScofield  "total_rows" field has been added. I will add a new parameter limit later.

tisonkun (changes requested):
If we go forward in the `limit` param way, I don't think we should add the RequestSource option that would introduce tech debt.

src/servers/src/http.rs
```diff
@@ -336,6 +343,12 @@ impl Display for Epoch {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, EnumString)]
+pub enum RequestSource {
+    #[strum(ascii_case_insensitive)]
+    Dashboard,
```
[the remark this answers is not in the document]
realtaobo: make sense, I will remove it later.

realtaobo:
@MichaelScofield @zyy17 @tisonkun @evenyag PTAL: refactor PR by `limit` param.

tisonkun (approved):
LGTM. Thanks for your contribution!

Issue #3889, opened 2024-05-09
Condionally limit total rows in COPY FROM


waynexia:
### What problem does the new feature solve?

Follow up of https://github.com/GreptimeTeam/greptimedb/pull/3819/files

After https://github.com/GreptimeTeam/greptimedb/pull/3819/files, a `COPY FROM` can only copy a **hardcoded** number of rows, and terminate the `COPY FROM` task without any hint.

### What does the feature do?

Change the limit behavior to
- do not limit by default
- make the limit rows configurable

As for the way to config it, we come up with two possible approaches:
- Add a `LIMIT` segment in `COPY FROM` statement, like 
```sql
COPY [<db>.]<table_name>
FROM { '<path>/[<filename>]' }
[ [ WITH ]
 (
   [ FORMAT =  { parquet } ]
   [ PATTERN = '<regex_pattern>' ]
 )
]
[LIMIT NUM]
```

- Add an option for it:
```sql
COPY [<db>.]<table_name>
FROM { '<path>/[<filename>]' }
[ [ WITH ]
 (
   [ FORMAT =  { parquet } ]
   [ PATTERN = '<regex_pattern>' ]
   [ LIMIT = NUM ]
 )
]
```

We would personally prefer the first way as it is more natural to write.

PR #3910, opened 2024-05-10
feat: limit total rows copied in COPY TABLE FROM with LIMIT segment

irenjj:

close: #3889

limit total rows copied in COPY TABLE FROM with LIMIT segment

## Checklist

### DIFF
--- a/src/operator/src/statement/copy_table_from.rs
+++ b/src/operator/src/statement/copy_table_from.rs
@@ -378,12 +378,21 @@ impl StatementExecutor {
 
         let mut rows_inserted = 0;
         let mut insert_cost = 0;
-        let (max_insert_rows, is_limit) = if let Some(num) = req.limit {
-            (num as usize, true)
+        let max_insert_rows = if let Some(num) = req.limit {
+            Some(num as usize)
         } else {
             match req.with.get(MAX_INSERT_ROWS) {
-                Some(num) => (num.parse::<usize>().ok().unwrap(), true),
-                None => (0, false),
+                Some(num) => match num.parse::<usize>() {
+                    Ok(num) => Some(num),
+                    Err(_) => {
+                        return Err(error::InvalidCopyParameterSnafu {
+                            key: MAX_INSERT_ROWS,
+                            value: num,
+                        }
+                        .build());
+                    }
+                },
+                None => None,
             }
         };
         for (compat_schema, file_schema_projection, projected_table_schema, file_metadata) in files
@@ -437,8 +446,10 @@ impl StatementExecutor {
                     insert_cost += cost;
                 }
 
-                if is_limit && rows_inserted >= max_insert_rows {
-                    return Ok(gen_insert_output(rows_inserted, insert_cost));
+                if let Some(max_insert_rows) = max_insert_rows {
+                    if rows_inserted >= max_insert_rows {
+                        return Ok(gen_insert_output(rows_inserted, insert_cost));
+                    }
                 }
             }
 
--- a/src/operator/src/statement/copy_table_from.rs
+++ b/src/operator/src/statement/copy_table_from.rs
@@ -379,11 +379,14 @@ impl StatementExecutor {
 
         let mut rows_inserted = 0;
         let mut insert_cost = 0;
-        let max_insert_rows = req
-            .with
-            .get(MAX_INSERT_ROWS)
-            .and_then(|val| val.parse::<usize>().ok())
-            .unwrap_or(DEFAULT_MAX_INSERT_ROWS);
+        let max_insert_rows = if let Some(num) = req.limit {
+            num as usize
+        } else {
+            req.with
+                .get(MAX_INSERT_ROWS)
+                .and_then(|val| val.parse::<usize>().ok())
+                .unwrap_or(DEFAULT_MAX_INSERT_ROWS)
+        };
         for (compat_schema, file_schema_projection, projected_table_schema, file_metadata) in files
         {
             let mut stream = self
--- a/src/operator/src/statement/copy_table_from.rs
+++ b/src/operator/src/statement/copy_table_from.rs
@@ -377,7 +377,7 @@ impl StatementExecutor {
 
         let mut rows_inserted = 0;
         let mut insert_cost = 0;
-        let max_insert_rows = req.limit;
+        let max_insert_rows = req.limit.map(|n| n as usize);
         for (compat_schema, file_schema_projection, projected_table_schema, file_metadata) in files
         {
             let mut stream = self
@@ -430,7 +430,7 @@ impl StatementExecutor {
                 }
 
                 if let Some(max_insert_rows) = max_insert_rows {
-                    if rows_inserted >= max_insert_rows as usize {
+                    if rows_inserted >= max_insert_rows {
                         return Ok(gen_insert_output(rows_inserted, insert_cost));
                     }
                 }
--- a/src/operator/src/statement/copy_table_from.rs
+++ b/src/operator/src/statement/copy_table_from.rs
@@ -53,7 +53,6 @@ use crate::statement::StatementExecutor;
 const DEFAULT_BATCH_SIZE: usize = 8192;
 const DEFAULT_READ_BUFFER: usize = 256 * 1024;
 const MAX_INSERT_ROWS: &str = "max_insert_rows";
-const DEFAULT_MAX_INSERT_ROWS: usize = 1000;
 
 enum FileMetadata {
     Parquet {
@@ -379,13 +378,13 @@ impl StatementExecutor {
 
         let mut rows_inserted = 0;
         let mut insert_cost = 0;
-        let max_insert_rows = if let Some(num) = req.limit {
-            num as usize
+        let (max_insert_rows, is_limit) = if let Some(num) = req.limit {
+            (num as usize, true)
         } else {
-            req.with
-                .get(MAX_INSERT_ROWS)
-                .and_then(|val| val.parse::<usize>().ok())
-                .unwrap_or(DEFAULT_MAX_INSERT_ROWS)
+            match req.with.get(MAX_INSERT_ROWS) {
+                Some(num) => (num.parse::<usize>().ok().unwrap(), true),
+                None => (0, false),
+            }
         };
         for (compat_schema, file_schema_projection, projected_table_schema, file_metadata) in files
         {
@@ -438,7 +437,7 @@ impl StatementExecutor {
                     insert_cost += cost;
                 }
 
-                if rows_inserted >= max_insert_rows {
+                if is_limit && rows_inserted >= max_insert_rows {
                     return Ok(gen_insert_output(rows_inserted, insert_cost));
                 }
             }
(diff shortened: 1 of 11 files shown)


src/operator/src/statement/copy_table_from.rs
```diff
@@ -379,11 +379,14 @@ impl StatementExecutor {
...
-            .and_then(|val| val.parse::<usize>().ok())
-            .unwrap_or(DEFAULT_MAX_INSERT_ROWS);
+        let max_insert_rows = if let Some(num) = req.limit {
+            num as usize
+        } else {
+            req.with
+                .get(MAX_INSERT_ROWS)
+                .and_then(|val| val.parse::<usize>().ok())
+                .unwrap_or(DEFAULT_MAX_INSERT_ROWS)
+        };
```
evenyag, line 389: Is there a way to disable the limit?

irenjj: Maybe we can add `LIMIT 0` to disable it.

evenyag: @waynexia What do you think?

waynexia: Maybe we should not limit if no `LIMIT` is specified

>Change the limit behavior to
>- do not limit by default
>- make the limit rows configurable

evenyag: > Maybe we should not limit if no LIMIT is specified

Makes sense.

src/operator/src/statement/copy_table_from.rs
```diff
@@ -379,13 +378,13 @@ impl StatementExecutor {
...
-            num as usize
+        let (max_insert_rows, is_limit) = if let Some(num) = req.limit {
+            (num as usize, true)
         } else {
-            req.with
-                .get(MAX_INSERT_ROWS)
-                .and_then(|val| val.parse::<usize>().ok())
-                .unwrap_or(DEFAULT_MAX_INSERT_ROWS)
+            match req.with.get(MAX_INSERT_ROWS) {
+                Some(num) => (num.parse::<usize>().ok().unwrap(), true),
```
evenyag, line 385: Did we check that `num` is a valid numeric string?

src/operator/src/statement/copy_table_from.rs
```diff
@@ -379,11 +378,23 @@ impl StatementExecutor {
...
+                Some(num) => match num.parse::<usize>() {
+                    Ok(num) => Some(num),
+                    Err(_) => {
+                        return Err(error::InvalidCopyParameterSnafu {
+                            key: MAX_INSERT_ROWS,
+                            value: num,
+                        }
+                        .build());
+                    }
+                },
```
evenyag, line 394: We could further simplify the code
```suggestion
                Some(num) => Some(num.parse::<usize>()
                        .context(error::InvalidCopyParameterSnafu {
                            key: MAX_INSERT_ROWS,
                            value: num,
                        })?),
```

src/operator/src/statement/copy_table_from.rs
```diff
@@ -379,11 +378,23 @@ impl StatementExecutor {
...
-            .and_then(|val| val.parse::<usize>().ok())
-            .unwrap_or(DEFAULT_MAX_INSERT_ROWS);
+        let max_insert_rows = if let Some(num) = req.limit {
+            Some(num as usize)
+        } else {
+            match req.with.get(MAX_INSERT_ROWS) {
+                Some(num) => match num.parse::<usize>() {
+                    Ok(num) => Some(num),
+                    Err(_) => {
+                        return Err(error::InvalidCopyParameterSnafu {
```
evenyag, line 388: nitpick: Maybe we could add a sqlness test that the max_insert_rows option is invalid.

evenyag (approved):
Generally LGTM

tests/cases/standalone/common/copy/copy_from_fs_parquet.sql
```diff
@@ -40,6 +40,12 @@ Copy with_limit_rows FROM '/tmp/demo/export/parquet_files/' WITH (MAX_INSERT_ROW
 
 select count(*) from with_limit_rows;
 
+CREATE TABLE with_limit_rows_segment(host string, cpu double, memory double, ts timestamp time index);
+
+Copy with_limit_rows FROM '/tmp/demo/export/parquet_files/' LIMIT 2;
```
waynexia, line 45:
```suggestion
Copy with_limit_rows_segment FROM '/tmp/demo/export/parquet_files/' LIMIT 2;
```

src/operator/src/statement/copy_table_from.rs
evenyag: You could use [ok()](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.ok) to convert the result into an option.

src/operator/src/statement/copy_table_from.rs
evenyag: I see we also support the following syntax:
```
Copy with_limit_rows FROM '/tmp/demo/export/parquet_files/' WITH (MAX_INSERT_ROWS = 2);
```

But if we already support the limit syntax, I don't see the necessity to provide two ways to set this limit.

tests/cases/standalone/common/copy/copy_from_fs_parquet.sql
```diff
@@ -40,6 +40,14 @@ Copy with_limit_rows FROM '/tmp/demo/export/parquet_files/' WITH (MAX_INSERT_ROW
 
 select count(*) from with_limit_rows;
 
+CREATE TABLE with_limit_rows_segment(host string, cpu double, memory double, ts timestamp time index);
+
+Copy with_limit_rows_segment FROM '/tmp/demo/export/parquet_files/' LIMIT 2;
+
+select count(*) from with_limit_rows_segment;
+
+Copy with_limit_rows_segment FROM '/tmp/demo/export/parquet_files/' LIMIT -1;
```
evenyag, line 49: We should test providing an invalid option to the with clause. But I think we don't need to support this option once the limit syntax is implemented.
```
Copy with_limit_rows FROM '/tmp/demo/export/parquet_files/' WITH (MAX_INSERT_ROWS = 2);
```

src/operator/src/statement/copy_table_from.rs
```diff
@@ -379,11 +378,23 @@ impl StatementExecutor {
...
+                            key: MAX_INSERT_ROWS,
+                            value: num,
+                        }
+                        .build());
+                    }
+                },
+                None => None,
+            }
```
evenyag, line 396:
```suggestion
        let max_insert_rows = req.limit.map(|n| n as usize);
```

waynexia (approved):

evenyag (commented):
We could only support the `limit` syntax.

src/sql/src/parsers/copy_parser.rs
```diff
@@ -56,7 +56,7 @@ impl<'a> ParserContext<'a> {
             })?;
 
         let req = if self.parser.parse_keyword(Keyword::TO) {
-            let (with, connection, location) = self.parse_copy_parameters()?;
```
tisonkun, line 59: Perhaps we should fail if COPY DATABASE with a LIMIT param? This seems an invalid input.

src/sql/src/error.rs
```diff
@@ -149,6 +149,14 @@ pub enum Error {
         location: Location,
     },
 
+    #[snafu(display("Unrecognized database option key: {}, value: {}", key, value))]
+    InvalidDatabaseOptionValue {
+        key: Ident,
+        value: u64,
+        #[snafu(implicit)]
+        location: Location,
+    },
```
evenyag, line 158: Could we reuse the `InvalidSql` variant? Since `limit` isn't a database option now. The error message might be `limit xxx is not supported`

src/sql/src/parsers/copy_parser.rs
```diff
@@ -56,7 +56,14 @@ impl<'a> ParserContext<'a> {
...
+            let (with, connection, location, limit) = self.parse_copy_parameters()?;
+            if limit.is_some() {
+                return error::InvalidSqlSnafu {
+                    msg: "limit xxx is not supported",
```
killme2008, line 62:
```suggestion
                    msg: "limit is not supported",
```

src/sql/src/parsers/copy_parser.rs
```diff
@@ -68,7 +75,14 @@ impl<'a> ParserContext<'a> {
             self.parser
                 .expect_keyword(Keyword::FROM)
                 .context(error::SyntaxSnafu)?;
-            let (with, connection, location) = self.parse_copy_parameters()?;
+            let (with, connection, location, limit) = self.parse_copy_parameters()?;
+            if limit.is_some() {
+                return error::InvalidSqlSnafu {
+                    msg: "limit xxx is not supported",
```
killme2008, line 81:
```suggestion
                    msg: "limit is not supported",
```

tests/cases/standalone/common/copy/copy_database_from_fs_parquet.result
```diff
@@ -52,6 +52,14 @@ SELECT * FROM demo ORDER BY ts;
 | host1 | 66.6 | 1024.0 | 2022-06-15T07:02:37 |
 +-------+------+--------+---------------------+
 
+DELETE FROM demo;
+
+Affected Rows: 1
+
+COPY DATABASE public FROM '/tmp/demo/export/parquet_range/' LIMIT 2;
+
+Error: 2000(InvalidSyntax), Invalid SQL, error: limit xxx is not supported
```
killme2008, line 61:
```suggestion
Error: 2000(InvalidSyntax), Invalid SQL, error: limit is not supported
```

killme2008 (commented):
Thank you. A few suggestions about the error info, please take a look.

src/sql/src/parsers/copy_parser.rs
```diff
@@ -58,9 +58,8 @@ impl<'a> ParserContext<'a> {
         let req = if self.parser.parse_keyword(Keyword::TO) {
             let (with, connection, location, limit) = self.parse_copy_parameters()?;
             if let Some(num) = limit {
-                return error::InvalidDatabaseOptionValueSnafu {
-                    key: "LIMIT",
-                    value: num,
+                return error::InvalidSqlSnafu {
+                    msg: "limit xxx is not supported",
```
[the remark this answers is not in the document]
evenyag: Well, I think current error message is enough.

tests/cases/standalone/common/copy/copy_from_fs_parquet.result
```diff
@@ -93,22 +93,26 @@ select count(*) from without_limit_rows;
...
 
 +----------+
 | COUNT(*) |
 +----------+
 | 2        |
 +----------+
 
+Copy with_limit_rows_segment FROM '/tmp/demo/export/parquet_files/' LIMIT -1;
+
+Error: 2000(InvalidSyntax), Unexpected token while parsing SQL statement: Copy with_limit_rows_segment FROM '/tmp/demo/export/parquet_files/' LIMIT -1;, expected: 'maximum rows', found: 1: sql parser error: Expected literal int, found: - at Line: 1, Column 75
```
evenyag, line 114: As you can see, the error comes from the sql parser itself as it can't parse `-1`, not from our codes.

PR #3922, opened 2024-05-11
feat: introduce `TableRouteCache` to `PartitionRuleManager`

WenyXu:

Introduce `TableRouteCache` to `PartitionRuleManager` to reduce overhead of deserializing cached bytes into `TableRoute`.

## Checklist


src/common/meta/src/cache/container.rs
```diff
@@ -101,9 +101,15 @@ where
...
     pub async fn get(&self, key: K) -> Result<Option<V>> {
+        metrics::CACHE_CONTAINER_CACHE_GET
+            .with_label_values(&[&self.name])
+            .inc();
         let moved_init = self.initializer.clone();
         let moved_key = key;
         let init = async move {
+            metrics::CACHE_CONTAINER_CACHE_MISS
+                .with_label_values(&[&self.name])
+                .inc();
```
fengjiachun, line 112: Perhaps what we need is a timer to track the time spent, as "time" may require more attention.

src/partition/src/error.rs
```diff
@@ -29,6 +29,13 @@ use crate::expr::PartitionExpr;
 #[snafu(visibility(pub))]
 #[stack_trace_debug]
 pub enum Error {
+    #[snafu(display("Failed to find table route for table id {}", table_id))]
```
fengjiachun, line 32:
```suggestion
    #[snafu(display("Failed to find table route: {}", table_id))]
```

src/common/meta/src/cache/container.rs
```diff
@@ -101,9 +101,15 @@ where
 {
     /// Returns a _clone_ of the value corresponding to the key.
     pub async fn get(&self, key: K) -> Result<Option<V>> {
```
fengjiachun, line 103: Can we also provide a `batch_get`? But not in this PR. We have already implemented this in Cached KvBackend: https://github.com/GreptimeTeam/greptimedb/blob/89da42dbc12687157e00dbd4efae75ba154381c1/src/catalog/src/kvbackend/client.rs#L170

fengjiachun (approved):

src/common/meta/src/cache/container.rs
WenyXu: It's a little tricky, we need to employ a lock to ensure consistency.

waynexia (commented):
The changes looks good to me. But I'm not sure about other underlying logics (like who should use the cache but not, or who should invalid the cache but not). PTAL @fengjiachun

fengjiachun:
> The changes looks good to me. But I'm not sure about other underlying logics (like who should use the cache but not, or who should invalid the cache but not). PTAL @fengjiachun

I'm o_O ing

fengjiachun (approved):

Issue #3944, opened 2024-05-15
ParquetReader does not respect time range provided


v0y4g3r:
### What type of enhancement is this?

Performance

### What does the enhancement do?

`ParquetReaderBuilder` provides a `time_range` option to filter the timestamps of rows to read. But currently it is not used anywhere when reading parquet files. We need to respect this option which will boost the scan performance when an exact time range is provided.

https://github.com/GreptimeTeam/greptimedb/blob/5a0629eaa0309f10bbcf11a0772c02e271fc1853/src/mito2/src/sst/parquet/reader.rs#L69

### Implementation challenges

There's a workaround: we can transform the time range into predicates and inserts them into `ParquetReaderBuilder::predicate`. These predicates will be applied to the batches read from parquet files.

But this workaround does not skip pages and row groups, which may waste extra IO overhead. We need to use the time range to prune row groups and data pages according the the file metadata.

PR #3947, opened 2024-05-15
feat: respect time range when building parquet reader

v0y4g3r:

- #3944

This PR converts the desired time range into `SimpleFilterContext` while building parquet readers so that the reader can filter out those rows outside the desired time range.

## Checklist

### DIFF
--- a/src/mito2/src/read/scan_region.rs
+++ b/src/mito2/src/read/scan_region.rs
@@ -18,7 +18,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use common_recordbatch::SendableRecordBatchStream;
-use common_telemetry::{debug, error, info, warn};
+use common_telemetry::{debug, error, warn};
 use common_time::range::TimestampRange;
 use store_api::region_engine::{RegionScannerRef, SinglePartitionScanner};
 use store_api::storage::ScanRequest;
@@ -278,7 +278,6 @@ impl ScanRegion {
         );
 
         let index_applier = self.build_index_applier();
-        info!("Request filters: {:?}", self.request.filters);
         let predicate = Predicate::new(self.request.filters.clone());
         // The mapper always computes projected column ids as the schema of SSTs may change.
         let mapper = match &self.request.projection {
--- a/src/mito2/src/read/scan_region.rs
+++ b/src/mito2/src/read/scan_region.rs
@@ -18,7 +18,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use common_recordbatch::SendableRecordBatchStream;
-use common_telemetry::{debug, error, warn};
+use common_telemetry::{debug, error, info, warn};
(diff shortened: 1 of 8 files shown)


src/mito2/src/read/scan_region.rs
```diff
@@ -269,6 +269,7 @@ impl ScanRegion {
         );
 
         let index_applier = self.build_index_applier();
+        info!("Request filters: {:?}", self.request.filters);
```
evenyag, line 272: Do we need this log? Maybe we can change it to `debug!()`.

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -225,7 +229,7 @@ impl ParquetReaderBuilder {
 
         metrics.build_cost = start.elapsed();
 
-        let filters = if let Some(predicate) = &self.predicate {
+        let mut filters = if let Some(predicate) = &self.predicate {
```
evenyag, line 232: Is it possible to reserve the capacity for `filters`?

src/query/src/tests/time_range_filter_test.rs
```diff
@@ -114,9 +114,9 @@ struct TimeRangeTester {
 impl TimeRangeTester {
     async fn check(&self, sql: &str, expect: TimestampRange) {
         let _ = exec_selection(self.engine.clone(), sql).await;
-        let filters = self.get_filters();
+        let mut filters = self.get_filters();
```
evenyag, line 117: Should it be `take_filters()`?

src/mito2/src/sst/parquet/reader.rs
v0y4g3r: > Is it possible to reserve the capacity for `filters`?

Cannot determine the num of time range filters here.

src/table/src/predicate.rs
```diff
@@ -131,250 +131,249 @@ impl Predicate {
...
 }
 
-// tests for `TimeRangePredicateBuilder` locates in src/query/tests/time_range_filter_test.rs
+// tests for `build_time_range_predicate` locates in src/query/tests/time_range_filter_test.rs
 // since it requires query engine to convert sql to filters.
-/// `TimeRangePredicateBuilder` extracts time range from logical exprs to facilitate fast
+/// `build_time_range_predicate` extracts time range from logical exprs to facilitate fast
 /// time range pruning.
-pub struct TimeRangePredicateBuilder<'a> {
+pub fn build_time_range_predicate<'a>(
```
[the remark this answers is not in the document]
v0y4g3r: > Could we port more tests for this function? I remember that we had many tests for it before.

Most cases are already here: https://github.com/GreptimeTeam/greptimedb/commit/43aefc5d74dfa73b7819cae77b7eb546d8534a41

waynexia (approved):

evenyag (approved):

Issue #3886, opened 2024-05-08
Partitioned scan for RegionEngine


evenyag:
### What type of enhancement is this?

API improvement

### What does the enhancement do?

The `RegionEngine` trait provides a `handle_query()` method to scan a region and returns a stream of `RecordBatch`.
https://github.com/GreptimeTeam/greptimedb/blob/d99746385b8da491d31a953b1d26a348cf00bc73/src/store-api/src/region_engine.rs#L136-L140

This method is easy to use but has some limitations:
- The output concurrency is always 1
- The engine can't return more information about the query to callers

To maximize parallelism in https://github.com/GreptimeTeam/greptimedb/issues/2806, the engine should provide a way to return multiple streams to scan different partitions of a region concurrently.

### Implementation challenges

This issue proposes to add a new method to the region engine which supports partitioned scan. The method returns a trait object that can create a stream according to a partition index.
```rust
pub struct ScannerProperties {
    // Properties of the scanner
    // e.g. number of partitions, range of partitions
}

pub trait RegionScanner {
    fn properties(&self) -> &ScannerProperties;

    fn scan_partition(&self, partition: usize) -> Result<SendableRecordBatchStream, BoxedError>;
}

pub type RegionScannerRef = Arc<dyn RegionScanner>;

pub trait RegionEngine {
    async fn handle_partitioned_query(
        &self,
        region_id: RegionId,
        request: ScanRequest,
    ) -> Result<RegionScannerRef, BoxedError>;
}
```

We could then use the scanner to implement a `PhysicalPlan` and let the query engine process multiple partitions. We might need to refactor the `StreamScanAdapter` as it assumes there is only one partition.
https://github.com/GreptimeTeam/greptimedb/blob/a6a702de4e51e91d4066d5a8e916b2e99116316a/src/table/src/table/scan.rs#L103-L119

PR #3948, opened 2024-05-15
feat: Adds `RegionScanner` trait

evenyag:

- https://github.com/GreptimeTeam/greptimedb/issues/3886

This PR adds the `RegionScanner` trait mentioned in https://github.com/GreptimeTeam/greptimedb/issues/3886

```rust
pub trait RegionScanner: Debug + DisplayAs + Send + Sync {
    fn properties(&self) -> &ScannerProperties;
    fn schema(&self) -> SchemaRef;
    fn scan_partition(&self, partition: usize) -> Result<SendableRecordBatchStream, BoxedError>;
}
```

It allows us to split rows to scan into multiple partitions and increase scan parallelism. The `RegionEngine` has a new method `handle_partitioned_query()` to get the `RegionScanner`. It removes the old `handle_query()` method from the trait and only exposes it in the `MitoEngine`.

Since most region engines don't support partitioning, this PR adds a `SinglePartitionScanner` to turn a stream into a region scanner.

The execution plan returned by the dummy table provider now becomes the `ReadFromRegion` plan. This plan adapts a `RegionScanner` to the `ExecutionPlan` trait.

## Checklist

### DIFF
--- a/src/datanode/src/tests.rs
+++ b/src/datanode/src/tests.rs
@@ -23,7 +23,6 @@ use common_function::function::FunctionRef;
 use common_function::scalars::aggregate::AggregateFunctionMetaRef;
 use common_query::prelude::ScalarUdf;
 use common_query::Output;
-use common_recordbatch::SendableRecordBatchStream;
 use common_runtime::Runtime;
 use query::dataframe::DataFrame;
 use query::plan::LogicalPlan;
@@ -187,14 +186,6 @@ impl RegionEngine for MockRegionEngine {
 
         let _ = self.sender.send((region_id, request)).await;
         Ok(RegionResponse::new(0))
-    }
-
-    async fn handle_query(
-        &self,
-        _region_id: RegionId,
-        _request: ScanRequest,
-    ) -> Result<SendableRecordBatchStream, BoxedError> {
-        unimplemented!()
     }
 
     async fn handle_partitioned_query(
--- a/src/datanode/src/tests.rs
+++ b/src/datanode/src/tests.rs
@@ -202,7 +202,7 @@ impl RegionEngine for MockRegionEngine {
         _region_id: RegionId,
         _request: ScanRequest,
     ) -> Result<RegionScannerRef, BoxedError> {
-        todo!()
+        unimplemented!()
     }
 
     async fn get_metadata(&self, _region_id: RegionId) -> Result<RegionMetadataRef, BoxedError> {
--- a/src/datanode/src/tests.rs
+++ b/src/datanode/src/tests.rs
@@ -188,7 +188,7 @@ impl RegionEngine for MockRegionEngine {
         Ok(RegionResponse::new(0))
     }
 
-    async fn handle_partitioned_query(
+    async fn handle_query(
         &self,
         _region_id: RegionId,
         _request: ScanRequest,
--- a/src/datanode/src/tests.rs
+++ b/src/datanode/src/tests.rs
@@ -32,7 +32,7 @@ use query::query_engine::DescribeResult;
 use query::{QueryEngine, QueryEngineContext};
 use session::context::QueryContextRef;
 use store_api::metadata::RegionMetadataRef;
-use store_api::region_engine::{RegionEngine, RegionRole, SetReadonlyResponse};
+use store_api::region_engine::{RegionEngine, RegionRole, RegionScannerRef, SetReadonlyResponse};
 use store_api::region_request::{AffectedRows, RegionRequest};
 use store_api::storage::{RegionId, ScanRequest};
 use table::TableRef;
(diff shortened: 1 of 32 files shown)


src/query/src/dummy_catalog.rs
```diff
@@ -168,12 +168,12 @@ impl TableProvider for DummyTableProvider {
...
 
-        let stream = self
+        let scanner = self
             .engine
-            .handle_query(self.region_id, request)
+            .handle_partitioned_query(self.region_id, request)
             .await
             .map_err(|e| DataFusionError::External(Box::new(e)))?;
-        Ok(Arc::new(StreamScanAdapter::new(stream)))
+        Ok(Arc::new(ReadFromRegion::new(scanner)))
```
waynexia, line 176: The struct name changes from a noun to a verb, do want to think it a better name?

src/store-api/src/region_engine.rs
```diff
@@ -172,3 +235,52 @@ pub trait RegionEngine: Send + Sync {
 }
 
 pub type RegionEngineRef = Arc<dyn RegionEngine>;
+
+/// A [RegionScanner] that only scans a single partition.
+pub struct SinglePartitionScanner {
+    stream: std::sync::Mutex<Option<SendableRecordBatchStream>>,
```
waynexia, line 241:
```suggestion
    stream: Mutex<Option<SendableRecordBatchStream>>,
```

src/store-api/src/region_engine.rs
```diff
@@ -120,6 +124,65 @@ impl From<PbRegionRole> for RegionRole {
...
+    }
+
+    /// Converts to datafusion's [Partitioning].
+    pub fn to_df_partitioning(&self) -> Partitioning {
+        match self {
+            ScannerPartitioning::Unknown(num_partitions) => {
+                Partitioning::UnknownPartitioning(*num_partitions)
+            }
+        }
+    }
```
waynexia, line 149: We suggest not to provide such a method. Leaving `ScannerPartitioning` internal (in this project) only. DF's `Partitioning` is 
- for generic data partition, especially intermediate data
- hard to add new variants
- not as expressive as we want

And the `ScannerPartitioning` is likely to be used in our optimizer rule, which also bypasses the `Partitioning` interface

src/store-api/src/region_engine.rs
```diff
@@ -120,6 +124,65 @@ impl From<PbRegionRole> for RegionRole {
...
+
+    /// Returns properties of partitions to scan.
+    pub fn partitioning(&self) -> &ScannerPartitioning {
+        &self.partitioning
+    }
+}
+
+/// A scanner that provides a way to scan the region concurrently.
+/// The scanner splits the region into partitions so that each partition can be scanned concurrently.
+pub trait RegionScanner: Debug + DisplayAs + Send + Sync {
```
waynexia, line 173: Also document this trait is beneath `ExecutionPlan`

src/datanode/src/tests.rs
```diff
@@ -189,11 +188,11 @@ impl RegionEngine for MockRegionEngine {
         Ok(RegionResponse::new(0))
     }
 
-    async fn handle_query(
+    async fn handle_partitioned_query(
         &self,
         _region_id: RegionId,
         _request: ScanRequest,
-    ) -> Result<SendableRecordBatchStream, BoxedError> {
+    ) -> Result<RegionScannerRef, BoxedError> {
```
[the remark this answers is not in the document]
evenyag: Yes, `handle_partitioned_query()` is not a good name. I need a name for the previous `handle_query()` method that returns a stream. Do you have any suggestions?

src/store-api/src/region_engine.rs
evenyag: We always map our partitions to `unknown` partitioning in DF. The main purpose of this method is to tell df that we can return multiple streams. I can change this to `partition_count()`.

src/query/src/dummy_catalog.rs
evenyag: Naming is hard. Do you have suggestions?

src/datanode/src/tests.rs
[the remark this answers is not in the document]
waynexia: 👀why not keep it as is, `handle_query()`

src/store-api/src/region_engine.rs
evenyag: Removed in https://github.com/GreptimeTeam/greptimedb/pull/3948/commits/537f4b5f47934bf77b47bf499b81100f12f7b42d

waynexia: >We always map our partitions to unknown partitioning in DF.

Makes a little more sense to me. Let's keep it for now and review the usage one month later

evenyag: Already removed. Now only one place uses it.

src/datanode/src/tests.rs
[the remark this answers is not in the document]
evenyag: Need a name for `MitoEngine::handle_query() -> Result<SendableRecordBatchStream>` to avoid confusion.

v0y4g3r (approved):
LGTM

waynexia (approved):

PR #3950, opened 2024-05-15
feat: ignore internal columns in `SHOW CREATE TABLE`

WenyXu:

Ignore internal columns in `SHOW CREATE TABLE`. But, we still print the internal columns in `information_schema`(it's by design).

## Checklist


src/query/src/sql/show_create_table.rs
```diff
@@ -126,6 +135,10 @@ fn create_table_constraints(
     constraints
 }
 
+fn is_internal_column(name: &str) -> bool {
```
killme2008, line 138: It's better to move this function into `store_api::metric_engine_consts`, it can be used in some places.

killme2008 (approved):
Restly LGTM

waynexia (approved):

Issue #3729, opened 2024-04-17
Implement VIEW


waynexia:
### What problem does the new feature solve?

The VIEW statement in SQL allows users to store complex queries in the database as virtual tables. This provides several benefits:

1. Simplifies complex queries: Views can encapsulate complex logic and querying requirements into a single easily accessible object. This can simplify the querying process for users by abstracting away the complexity of the underlying data model.
2. Data security: Views allow users to restrict access to certain rows or columns of a table, providing an additional layer of security. Users can only query the data that is exposed through the view, limiting their access to sensitive information.
3. Reusability: Views can be reused across multiple queries, reports, and applications. This promotes code reusability, reduces redundancy, and makes it easier to maintain and update the underlying database schema.
4. Performance optimization: Views can improve planning performance by caching the input plan.
5. Data abstraction: Views can hide the underlying data model complexity from users, providing a simplified and consistent interface to the database. This can be particularly useful in situations where users need to interact with a complex database schema.

### What does the feature do?

Implement `VIEW` like https://www.postgresql.org/docs/current/sql-createview.html

waynexia:
Going to try it within one week

waynexia:
> Point 4 is not for VIEW, at least the one linked. It's for MATERIALIZE VIEW. They're different concept and quite different in implementation.

Updated

killme2008:
Tracking issues:

* [x] create view #3807 
* [x] query view #3952 
* [ ] view columns
* [ ] drop view
* [ ] show create view
* [ ] show views and `information_schema.views`

PR #3952, opened 2024-05-15
feat: querying from view works

killme2008:

#3729 

Make querying view works.

Main changes:

1. Adds `ViewInfoCache` for caching `ViewValueInfo`. @WenyXu 
2. Destroy view metadata in `DropDatabaseCursor `.
3. Extract and rewrite all table names in view logical plan. Saves the table names into view info.
4. Add `SubstraitPlanDecoder` trait and implement it in the query crate, which decodes the logical plan from bytes, takes care of the session state, etc.
5. Resolve view in `DfTableSourceProvider`. It queries the view info and tries to decode it, then wraps the logical plan into `ViewTable`.
6. Some refactoring:  move the PromQL planner to the query crate, and remove some unused errors, etc.

Todo:
1. Supports column names(alias) for view.
2. Save the optimized logical plan for view.  There are some serialization issues when saving the optimized plan. @waynexia 

## Checklist


Cargo.toml
```diff
@@ -120,7 +120,7 @@ etcd-client = { git = "https://github.com/MichaelScofield/etcd-client.git", rev
 fst = "0.4.7"
 futures = "0.3"
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "a11db14b8502f55ca5348917fd18e6fcf140f55e" }
+greptime-proto = { git = "https://github.com/killme2008/greptime-proto.git", rev = "9aa33273cf5ab449e687e9aa2bb10af3067984f3" }
```
killme2008, line 123: Depends on https://github.com/GreptimeTeam/greptime-proto/pull/172

src/catalog/src/table_source.rs
```diff
@@ -94,15 +107,47 @@ impl DfTableSourceProvider {
...
+            let catalog_list = Arc::new(DummyCatalogList::new(self.catalog_manager.clone()));
+            let logical_plan = self
+                .plan_decoder
+                .decode(Bytes::from(view_info.view_info.clone()), catalog_list, true)
+                .await
+                .context(DecodePlanSnafu {
+                    name: &table.table_info().name,
+                })?;
+
+            Arc::new(ViewTable::try_new(logical_plan, None).context(DatafusionSnafu)?)
```
killme2008, line 136: Use the datafusion `ViewTable`. as the table provider. https://docs.rs/datafusion/latest/datafusion/datasource/view/struct.ViewTable.html

src/common/meta/src/ddl/drop_database/cursor.rs
```diff
@@ -122,6 +160,20 @@ impl State for DropDatabaseCursor {
...
+                let table_info_value = ddl_ctx
+                    .table_metadata_manager
+                    .table_info_manager()
+                    .get(table_id)
+                    .await?
+                    .with_context(|| TableInfoNotFoundSnafu {
+                        table: format_full_table_name(&ctx.catalog, &ctx.schema, &table_name),
+                    })?;
+
+                if table_info_value.table_info.table_type == TableType::View {
```
killme2008, line 173: Not so elegant. May refactor it when implementing `DROP VIEW`.

zhongzc (commented):
I checked, basic LGTM

src/cache/src/lib.rs
```diff
@@ -82,11 +84,22 @@ pub fn build_fundamental_cache_registry(kv_backend: KvBackendRef) -> CacheRegist
...
+    // Builds the view info cache
+    let cache = CacheBuilder::new(DEFAULT_CACHE_MAX_CAPACITY)
+        .time_to_live(DEFAULT_CACHE_TTL)
+        .time_to_idle(DEFAULT_CACHE_TTI)
+        .build();
+    let view_info_cache = Arc::new(new_view_info_cache(
+        VIEW_INFO_CACHE_NAME.to_string(),
+        cache,
+        kv_backend.clone(),
+    ));
```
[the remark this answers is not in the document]
killme2008: It's called in here https://github.com/GreptimeTeam/greptimedb/blob/9e1af79637fff334a13682882635ae7631ec1380/src/operator/src/statement/ddl.rs#L520

fengjiachun:
@WenyXu Please take a look at the cache section

WenyXu:
> @WenyXu Could please take a look at the cache section?

The cache part, LGTM

killme2008:
@waynexia Please take a look.

src/cache/src/lib.rs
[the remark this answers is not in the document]
fengjiachun: This is not enough; it will only invalidate the local cache. A broadcast needs to be sent to the cluster (this can be done in the procedure).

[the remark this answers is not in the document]
killme2008: @WenyXu  Is that right? I thought the `cache_invalidator` would broadcast the invalidating message.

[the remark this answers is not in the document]
WenyXu: > @WenyXu Is that right? I thought the `cache_invalidator` would broadcast the invalidating message.

No, `cache_invalidator ` won't broadcast the invalidation messages. The broadcast should be done via Metasrv.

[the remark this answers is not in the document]
killme2008: @WenyXu Ok, i'll fix it.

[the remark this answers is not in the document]
killme2008: Fixed by https://github.com/GreptimeTeam/greptimedb/pull/3952/commits/74f47487873c6d14a2610e4f2820993333d23985

fengjiachun (approved):
Rest LGTM, @waynexia PTAL

src/query/src/error.rs
```diff
@@ -56,20 +56,6 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("Catalog not found: {}", catalog))]
-    CatalogNotFound {
```
waynexia, line 60: Are those error not used before, or removed in this patch?

src/query/src/query_engine/context.rs
```diff
@@ -58,6 +61,14 @@ impl QueryEngineContext {
         ))
     }
 
+    /// Creates a `[LogicalPlan]` decoder
```
waynexia, line 64:
```suggestion
    /// Creates a [`LogicalPlan`] decoder
```

src/query/src/query_engine/default_serializer.rs
```diff
@@ -0,0 +1,171 @@
...
+use prost::Message;
+use session::context::QueryContextRef;
+use snafu::ResultExt;
+use substrait::extension_serializer::ExtensionSerializer;
+use substrait::{DFLogicalSubstraitConvertor, SubstraitPlan};
+
+use crate::dist_plan::MergeScanLogicalPlan;
+use crate::error::DataFusionSnafu;
+
+/// Extended `[substrait::extension_serializer::ExtensionSerializer]` but supports `[MergeScanLogicalPlan]` serialization.
```
waynexia, line 39:
```suggestion
/// Extended [`substrait::extension_serializer::ExtensionSerializer`] but supports [`MergeScanLogicalPlan`] serialization.
```

src/query/src/error.rs
killme2008: They were previously used but got deprecated during refactoring, and it was an oversight not to remove them. I have a python script to find unused errors, so try to remove some at this patch.

src/query/src/query_engine/default_serializer.rs
```diff
@@ -0,0 +1,171 @@
...
+use datafusion::execution::registry::SerializerRegistry;
+use datafusion::execution::FunctionRegistry;
+use datafusion::logical_expr::LogicalPlan;
+use datafusion_expr::UserDefinedLogicalNode;
+use greptime_proto::substrait_extension::MergeScan as PbMergeScan;
+use prost::Message;
+use session::context::QueryContextRef;
+use snafu::ResultExt;
+use substrait::extension_serializer::ExtensionSerializer;
+use substrait::{DFLogicalSubstraitConvertor, SubstraitPlan};
```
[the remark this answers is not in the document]
killme2008: That's the question I want to ask. It's named `substrait ` but not `common-substrait` in 

https://github.com/GreptimeTeam/greptimedb/blob/4aa756c8960407e425da52f2891dc1563ade3175/Cargo.toml#L234

Do you want me to change it in this patch?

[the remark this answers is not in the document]
waynexia: Got it. Just to make sure it's not the "upstream substrait".

src/common/query/src/logical_plan.rs
```diff
@@ -68,6 +71,21 @@ pub fn create_aggregate_function(
     )
 }
 
+/// The datafusion `[LogicalPlan]` decoder.
+#[async_trait::async_trait]
+pub trait SubstraitPlanDecoder {
+    /// Decode the `[LogicalPlan]` from bytes with the `[CatalogProviderList]`.
```
waynexia, line 77:
```suggestion
    /// Decode the [`LogicalPlan`] from bytes with the [`CatalogProviderList`].
```

src/common/query/src/logical_plan.rs
```diff
@@ -68,6 +71,21 @@ pub fn create_aggregate_function(
...
+    /// When `optimize` is true, it will do the optimization for decoded plan.
+    async fn decode(
+        &self,
+        message: bytes::Bytes,
+        catalog_list: Arc<dyn CatalogProviderList>,
+        optimize: bool,
```
waynexia, line 83: Why a decoder needs to optimize the plan? We prefer to remove this param and let others call optimizer by themselves

waynexia (commented):
we should not depend on `test-integration`. where is it used?

src/common/meta/src/lib.rs
```diff
@@ -40,7 +40,7 @@ pub mod region_keeper;
 pub mod rpc;
 pub mod sequence;
 pub mod state_store;
-pub mod table_name;
+pub use table::table_name;
```
waynexia, line 43: Why re-export this

waynexia (approved):
Others LGTM

src/common/query/src/logical_plan.rs
killme2008: Agree, let me fix it.

src/common/meta/src/lib.rs
killme2008: Because the `table_name` was in `common-meta` before, and I moved it to `table` crate in this patch. Re-export it to avoid breaking too many codes.

killme2008: But let's break.

killme2008:
> we should not depend on `test-integration`. where is it used?

Sorry, i don't get the point. This patch doesn't change the dependencies in `test-integration` or make other crates depend on it.

src/common/query/src/logical_plan.rs
killme2008: I found the reason why I did it this way. The reason is that we need to decode the plan in `DfTableSourceProvider`, which is in the `catalog` crate. 

However, we can't call the optimizer in the `catalog` crate because the `SessionState` is only in the `query` crate. If we let the `query` catalog depend on the `catalog`, there will be a cyclic dependency issue.

killme2008: I leave a `TODO` here. If you have a great suggestion, I am open to refactoring it.

PR #3954, opened 2024-05-16
feat(flow): flow node manager

discord9:

Build flownode in standalone mode, allow flow create

- Added flow node mangager, which impl flownode trait
- Also impl grpc service trait for it

## Checklist


src/flow/src/adapter.rs
```diff
@@ -14,19 +14,678 @@
...
+            None,
+            false,
+            self.plugins.clone(),
+        );
+        let query_engine = query_engine_factory.query_engine();
+
+        let (tx, rx) = oneshot::channel();
+
+        let _handle = std::thread::spawn(move || {
+            let node_id = Some(1);
```
zhongzc, line 134: constant?

src/flow/src/adapter.rs
```diff
@@ -14,19 +14,678 @@
...
+            let (flow_node_manager, mut worker) =
+                FlownodeManager::new_with_worker(node_id, query_engine, self.table_meta.clone());
+            let _ = tx.send(flow_node_manager);
+            info!("Flow Worker started in new thread");
+            worker.run();
```
zhongzc, line 139: Consider stopping the worker gracefully

src/flow/src/adapter.rs
```diff
@@ -14,19 +14,678 @@
...
+        });
+        let man = rx.await.unwrap();
+        info!("Flow Node Manager started");
+        man
+    }
+}
+
+/// This function will create a new thread for flow worker and return a handle to the flow node manager
+pub fn start_flow_node_with_one_worker(
```
zhongzc, line 148: unused function?

src/flow/src/adapter.rs
```diff
@@ -14,19 +14,678 @@
...
+    }
+}
+
+/// FlowTickManager is a manager for flow tick, which trakc flow execution progress
+///
+/// TODO(discord9): better way to do it, and not expose flow tick even to other flow to avoid
+/// TSO coord mess
+#[derive(Clone)]
+pub struct FlowTickManager {
+    anchor: Anchor,
```
zhongzc, line 671: `minstant::Anchor` is not suitable for long-term timing, it is recommended to use SystemTime instead

src/cmd/src/standalone.rs
```diff
@@ -482,6 +496,12 @@ impl StartCommand {
         .await
         .context(StartFrontendSnafu)?;
 
+        // flow server need to be able to use frontend to write insert requests back
+        flownode
+            .set_frontend_invoker(Box::new(frontend.clone()))
+            .await;
+        let _handle = flownode.clone().run_background();
```
zhongzc, line 503: Consider adding `start` and `stop` to flownode to unify startup and graceful shutdown in `Instance::start` and `Instance::stop`

src/flow/src/adapter.rs
```diff
@@ -14,19 +14,678 @@
...
+    pub rpc_addr: String,
+}
+
+/// Flownode Builder
+pub struct FlownodeBuilder {
+    opts: FlownodeOptions,
+    plugins: Plugins,
+    kv_backend: Option<KvBackendRef>,
+    table_meta: TableMetadataManagerRef,
+    catalog_manager: Arc<KvBackendCatalogManager>,
```
fengjiachun, line 91: It's better to avoid using `KvBackendCatalogManager` directly since it's just an implementation of the catalog manager. We should use `CatalogManagerRef` instead.

src/flow/src/adapter.rs
fengjiachun: Why always be `1`? please add some comments or a TODO if needed.

src/flow/src/adapter.rs
```diff
@@ -14,19 +14,668 @@
...
+        for (table_name, reqs) in all_reqs {
+            if reqs.is_empty() {
+                continue;
+            }
+            let (catalog, schema) = (table_name[0].clone(), table_name[1].clone());
+            let ctx = Arc::new(QueryContext::with(&catalog, &schema));
+            // TODO(discord9): instead of auto build table from request schema, actually build table
+            // before `create flow` to be able to assign pk and ts etc.
+            let (primary_keys, schema) = if let Some(table_id) = self
+                .table_info_source
```
fengjiachun, line 276: Can we use `CatalogManager` directly? It is good at caching for `Table`. cc @WenyXu

discord9: > Question: can we use `CatalogManager` directly? It is good at caching for `Table`. cc @WenyXu

There is some need to query table from given `TableId` (i.e. when query source table's schema when accepting `RegionInsertRequest` wit h`TableId` from frontend) but CatalogManager doesn't have similiar methods?

zhongzc (approved):

src/cmd/src/standalone.rs
discord9: > Consider adding `start` and `stop` to flownode to unify startup and graceful shutdown in `Instance::start` and `Instance::stop`

Added `TODO`

PR #3957, opened 2024-05-16
feat: Add TLS support for gRPC service

### LINKED ISSUE #3336: TLS for gRPC service

Configuration

Add TLS support for our gRPC services (default port: 4001)

### Implementation challenges

_No response_

realtaobo:

close #3336 

- Add grpc service tls configure
- use `tonic::transport::ServerTlsConfig`

## Checklist


src/servers/src/grpc.rs
```diff
@@ -64,13 +68,15 @@ pub struct GrpcServerConfig {
     pub max_recv_message_size: usize,
     // Max gRPC sending(encoding) message size
     pub max_send_message_size: usize,
+    pub tls: TlsOption,
```
realtaobo, line 71: Why `serve_state` and `handlers` require `Mutex`, and `tls_config` also require `Mutex`?

sunng87:
It's great to see this! Thank you @realtaobo!

Before we get deep, I want to make sure that actually there are two types of grpc communications in GreptimeDB cluster: 

- the client-server grpc: by which our sdk talks to the frontends.
- the intra-cluster grpc: by which frontends/datanodes/metasrvs talks to each other.

If we are going to implement both, we need to make sure that they may or must use different tls configuration including certificates/keys/modes. We will need separated configuration items for that.

I will recommend you to do them in separated patches. Perhaps with this one, we focus on the client-server grpc first.

src/servers/src/grpc.rs
sunng87: The comment says it's mutable.

sunng87: for `tls_config` we have a feature to watch tls certficates and keys for changes and do live reloading without restart the server. I think for grpc we need to keep this behaviour.

realtaobo: > we have a feature to watch tls certficates and keys for changes and do live reloading without restart the server

@sunng87 Hey, I'm a bit confused. Currently, `Frontend` sets `grpc TLS configuration` when the service starts up.
https://github.com/GreptimeTeam/greptimedb/blob/7da92eb9eba63ce3f5757b86fdd151a2602cced0/src/servers/src/grpc.rs#L175-L179
 However, if we observe changes to the config after the service has started, it seems there's no way to reload the configuration without restarting the grpc service?

realtaobo:
> I will recommend you to do them in separated patches. Perhaps with this one, we focus on the client-server grpc first.

Yes, you're right, and PR becoming clearer. Let me revert the changes of `DataNode`.  And the intra-cluster grpc tls will be impl in next patches.

src/servers/src/grpc.rs
sunng87: I haven't look into how to configure TLS for tonic. But for mysql/postgres servers, we configure it using `ReloadableTlsServerConfig` (you can search for its definition and usage). 

For a TLS server, you don't have to restart the server to make the new certificates/keys live. It's mostly because how API designed. Actually the tls configuration for server is a stateless thing. We just need to make sure we use latest version of configuration when accepting new connection.

sunng87: By checking the builder pattern of tonic api, I think it will be difficult to get the auto reload feature work for tonic based gRPC. 

If we cannot have that reload feature supported, we will need to raise an error when user configured `watch: true` in `TlsOption` for gRPC.

sunng87:
Please let me know when the patch is ready for review @realtaobo , thank you!

realtaobo:
@sunng87 PTAL

src/servers/src/grpc/builder.rs
```diff
@@ -157,11 +163,33 @@ impl GrpcServerBuilder {
         &mut self.routes_builder
     }
 
+    pub fn with_tls_config(mut self, tls_option: TlsOption) -> Result<Self> {
+        // tonic does not support watching for tls config changes
+        // so we don't support it either for now
+        if tls_option.watch {
+            return Err(Error::NotSupported {
+                feat: "grpc tls watch".to_string(),
```
sunng87, line 171:
```suggestion
                feat: "Certificates watch and reloading is not supported at the moment".to_string(),
```

tests-integration/tests/http.rs
```diff
@@ -744,6 +744,12 @@ runtime_size = 8
 max_recv_message_size = "512MiB"
 max_send_message_size = "512MiB"
 
+[grpc.tls]
+mode = "disable"
+cert_path = ""
+key_path = ""
+watch = false
```
sunng87, line 751: just remove empty/default entries when mode is "disable"

src/client/src/client.rs
```diff
@@ -87,6 +87,17 @@ impl Client {
...
+    pub fn with_tls_and_urls<U, A>(urls: A, client_tls: ClientTlsOption) -> Result<Self>
+    where
+        U: AsRef<str>,
+        A: AsRef<[U]>,
+    {
+        let channel_config = ChannelConfig::default().client_tls_config(client_tls);
+        let channel_manager = ChannelManager::with_tls_config(channel_config)
+            .context(error::CreateTlsChannelSnafu)?;
+        Ok(Self::with_manager_and_urls(channel_manager, urls))
+    }
```
sunng87, line 99: I think this part is still about intra-cluster communication, right? We should remove it for now.

tests-integration/tests/grpc.rs
```diff
@@ -129,6 +136,7 @@ pub async fn test_grpc_message_size_ok(store_type: StorageType) {
     let config = GrpcServerConfig {
         max_recv_message_size: 1024,
         max_send_message_size: 1024,
+        tls: TlsOption::default(),
```
sunng87, line 139: I think we can use `..Default::default()` here to save effort if we are adding more members in future.

sunng87:
It mostly looks good to me. If you have time, you may want to look into tonic to see if there is a chance to make certificate reloadable without restarting the server.

tests-integration/tests/http.rs
realtaobo: It seems we can't remove it. Because it's a  serialize result from `toml-rs/toml` crate serialize.
https://github.com/GreptimeTeam/greptimedb/blob/418090b4647e5e2d9bc248f940f387070c8479a9/src/common/config/src/config.rs#L88-L91

src/datanode/src/service.rs
```diff
@@ -95,6 +96,7 @@ impl<'a> DatanodeServiceBuilder<'a> {
         let config = GrpcServerConfig {
             max_recv_message_size: opts.rpc_max_recv_message_size.as_bytes() as usize,
             max_send_message_size: opts.rpc_max_send_message_size.as_bytes() as usize,
+            tls: TlsOption::default(),
```
zyy17, line 99: nit: There is some tech debt between `FrontendOptions` and `DatanodeOptions`. It's better to unify by using `GrpcOptions`. Maybe it can be refactored in the next PR.

src/frontend/src/service_config/grpc.rs
```diff
@@ -26,6 +27,8 @@ pub struct GrpcOptions {
     pub max_recv_message_size: ReadableSize,
     // Max gRPC sending(encoding) message size
     pub max_send_message_size: ReadableSize,
+    #[serde(default = "Default::default")]
```
zyy17, line 30: Can you add the configuration in `config/frontend.example.toml` and `standalone.example.toml` and execute `make config-docs`?

src/cmd/src/frontend.rs
```diff
@@ -216,6 +216,7 @@ impl StartCommand {
 
         if let Some(addr) = &self.rpc_addr {
             opts.grpc.addr.clone_from(addr);
+            opts.grpc.tls = tls_opts.clone();
```
zyy17, line 219: nit: using the same certificate for multiple services is not the best security practice. However, it's acceptable for this PR to keep it simple; we can refactor it in the future.

realtaobo: Okay. For now, it seems  that`grpc`,`mysql`,`pg` are all using the same TLS configuration directly. We could refactor this: reading the TLS configuration from  config file firstly, or retaining the original logic.

realtaobo:
There seem to be there things to do:
- [ ] Support the intra-cluster grpc communications with TLS
- [ ] Refactored `DatanodeOptions`,  that is replace some grpc config(`max_recv_message_size`, etc.) with `GrpcOptions`
- [ ] It seems that grpc,mysql,pg are all using the same TLS configuration directly. We could refactor these with: Reading the TLS configuration from config file firstly.
--- 
I believe these tasks should be deferred to the future PRs, and we should keep the PR simple for now. @sunng87 @zyy17 How do you think so?

zyy17:
> There seem to be there things to do:
> 
> * [ ]  Support the intra-cluster grpc communications with TLS, from [comment](https://github.com/GreptimeTeam/greptimedb/pull/3957#issuecomment-2118549288)
> * [ ]  Refactored `DatanodeOptions`,  that is replace some grpc config(`max_recv_message_size`, etc.) with `GrpcOptions`, from [comment](https://github.com/GreptimeTeam/greptimedb/pull/3957#discussion_r1610909164)
> * [ ]  It seems that grpc,mysql,pg are all using the same TLS configuration directly. We could refactor these with: Reading the TLS configuration from config file firstly, from [comment](https://github.com/GreptimeTeam/greptimedb/pull/3957#discussion_r1610909164)
> 
> I believe these tasks should be deferred to the future PRs, and we should keep the PR simple for now. @sunng87 @zyy17 How do you think so?

I think so. Thank you for your summary! 😊

zyy17 (approved):
LGTM

sunng87 (approved):

PR #3958, opened 2024-05-16
test: gracefully shutdown postgres client in sql tests

evenyag:

Some tests of our postgres server failed on macOS
```bash
integration_sql_file_test::test_postgres_parameter_inference ---
thread 'tokio-runtime-worker' panicked at tests-integration/tests/sql.rs:773:26:
called `Result::unwrap()` on an `Err` value: Error { kind: Io, cause: Some(Os { code: 54, kind: ConnectionReset, message: "Connection reset by peer" }) }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
libunwind: stepWithCompactEncoding - invalid compact unwind encoding
```

After adding some logs, I found that the test finished and the server was shut down. We unwrap the result from the connection in tests so the test is panic if the server is shut down.
```rust
connection.await.unwrap();
```

This PR drops the client and then waits for the connection before shutting down the server.

The macOS action with this patch:
https://github.com/GreptimeTeam/greptimedb/actions/runs/9108898568

## Checklist

### DIFF
--- a/tests-integration/tests/sql.rs
+++ b/tests-integration/tests/sql.rs
@@ -431,8 +431,10 @@ pub async fn test_postgres_bytea(store_type: StorageType) {
     let (client, connection) = tokio_postgres::connect(&format!("postgres://{addr}/public"), NoTls)
         .await
         .unwrap();
+    let (tx, rx) = tokio::sync::oneshot::channel();
     tokio::spawn(async move {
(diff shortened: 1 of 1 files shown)


zyy17 (approved):

fengjiachun (approved):
LGTM

PR #3961, opened 2024-05-16
feat: support compression on gRPC server

shuiyisong:

This pr mainly supports gRPC server compression encoding options(`gzip`, `zstd`). Now gRPC server is allowed to receive requests with such encoding.

In this pr we also enable zstd compression encoding by default between frontend and datanode.

## Checklist

### DIFF
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -29,7 +29,7 @@ rpc_max_recv_message_size = "512MB"
 ## The maximum send message size for gRPC server.
 rpc_max_send_message_size = "512MB"
 
-## Supported compression encoding for gRPC server, e.g: gzip, zstd.
+## Supported compression encoding for gRPC server. Currently, `gzip` and `zstd` are supported.
 rpc_accept_compressed = ["gzip"]
 
 ## Enable telemetry to collect anonymous usage data.
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -29,7 +29,7 @@ rpc_max_recv_message_size = "512MB"
 ## The maximum send message size for gRPC server.
 rpc_max_send_message_size = "512MB"
 
-## Supported compression encoding for gRPC server, e.g: gzip, zstd
+## Supported compression encoding for gRPC server, e.g: gzip, zstd.
 rpc_accept_compressed = []
 
 ## Enable telemetry to collect anonymous usage data.
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -29,9 +29,6 @@ rpc_max_recv_message_size = "512MB"
 ## The maximum send message size for gRPC server.
 rpc_max_send_message_size = "512MB"
 
-## Supported compression encoding for gRPC server. Currently, `gzip` and `zstd` are supported.
-rpc_accept_compressed = ["gzip"]
-
 ## Enable telemetry to collect anonymous usage data.
 enable_telemetry = true
 
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -29,8 +29,8 @@ rpc_max_recv_message_size = "512MB"
 ## The maximum send message size for gRPC server.
 rpc_max_send_message_size = "512MB"
 
-## Enable gzip compression for gRPC
-enable_gzip_compression = false
+## Supported compression encoding for gRPC server, e.g: gzip, zstd
+rpc_accept_compressed = []
 
 ## Enable telemetry to collect anonymous usage data.
 enable_telemetry = true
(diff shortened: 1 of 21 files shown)


src/common/grpc/src/channel_manager.rs
```diff
@@ -261,6 +262,8 @@ pub struct ChannelConfig {
     pub max_recv_message_size: ReadableSize,
     // Max gRPC sending(encoding) message size
     pub max_send_message_size: ReadableSize,
+    // Enable gzip compression for gRPC
```
fengjiachun, line 265: Maybe:
```suggestion
    // Supported compression encoding for gRPC, e.g: gzip, zstd.
```
If I understand this code correctly.

src/servers/src/grpc.rs
```diff
@@ -64,13 +66,33 @@ pub struct GrpcServerConfig {
...
+    // Enable gzip compression for gRPC
+    pub accept_compressed: Vec<CompressionEncoding>,
+}
+
+pub fn parse_grpc_compression_encoding(
+    encodings: &Vec<String>,
+) -> Result<Vec<CompressionEncoding>> {
+    let mut result = Vec::with_capacity(encodings.len());
+    for encoding in encodings {
+        let encoding = match encoding.as_str() {
```
fengjiachun, line 78: Is it necessary to convert `encoding` to lowercase before matching?

fengjiachun (approved):

src/servers/src/grpc.rs
```diff
@@ -64,13 +66,33 @@ pub struct GrpcServerConfig {
...
+            "gzip" => CompressionEncoding::Gzip,
+            "zstd" => CompressionEncoding::Zstd,
+            _ => {
+                return UnsupportedGrpcCompressionEncodingSnafu { encoding }.fail();
```
waynexia, line 82: Maybe we should maintains a table for all supported methods in docs

src/servers/src/grpc/builder.rs
```diff
@@ -41,15 +41,21 @@ use crate::query_handler::OpenTelemetryProtocolHandlerRef;
...
-                .max_encoding_message_size(max_send_message_size),
-        )
+        let mut service_builder = $service
+            .max_decoding_message_size(max_recv_message_size)
+            .max_encoding_message_size(max_send_message_size);
+
+        for encoding in $accept_compressed {
+            service_builder = service_builder
+                .accept_compressed(*encoding)
+                .send_compressed(*encoding);
```
waynexia, line 55: Should this behavior use another config entry?

waynexia (commented):
🚀

config/datanode.example.toml
```diff
@@ -29,6 +29,9 @@ rpc_max_recv_message_size = "512MB"
 ## The maximum send message size for gRPC server.
 rpc_max_send_message_size = "512MB"
 
+## Supported compression encoding for gRPC server. Currently, `gzip` and `zstd` are supported.
+rpc_accept_compressed = ["gzip"]
```
sunng87, line 33: as discussed in slack, I think it's not necessary to configure this, especially for accepting client data

src/common/grpc/src/channel_manager.rs
```diff
@@ -381,6 +385,16 @@ impl ChannelConfig {
         self.client_tls = Some(client_tls_option);
         self
     }
+
+    /// Enable compression for gRPC
+    ///
+    /// Disabled by default.
+    pub fn accept_compressed(self, compression: Vec<CompressionEncoding>) -> Self {
```
tisonkun, line 392: I don't find the logic a config option can manipulate this option.

Also, is it possible that we enable compression by default? It seems there is no downside for doing so.

shuiyisong: It's confirmed that `channel_manager` and `client` is only used internally, we can remove those configs now. Should we also enable gzip/zstd compression encoding by default internally(between frontend/datanode/metasrv)? @sunng87 @waynexia @fengjiachun 

> Also, is it possible that we enable compression by default? It seems there is no downside for doing so.

Now the compression support is auto-enabled on our server(namely frontend), but it depends on the client request to determine whether to use compression or not(the compression encoding is done on client side).

tisonkun: > Now the compression support is auto-enabled on our server(namely frontend), but it depends on the client request to determine whether to use compression or not(the compression encoding is done on client side).

Of course. We can "allow" the support. Whether or not it's compressed is based on the request - this is normal :D

tisonkun (approved):

waynexia (approved):

PR #3962, opened 2024-05-16
feat: remove timeout in the channel between frontend and datanode

waynexia:

Don't timeout on gRPC connection from frontend to datanode. For some queries are expected to take a long while.

## Checklist

### DIFF
--- a/src/cmd/src/frontend.rs
+++ b/src/cmd/src/frontend.rs
@@ -24,6 +24,7 @@ use catalog::kvbackend::{CachedMetaKvBackendBuilder, KvBackendCatalogManager, Me
 use clap::Parser;
 use client::client_manager::DatanodeClients;
 use common_config::Configurable;
+use common_grpc::channel_manager::ChannelConfig;
 use common_meta::cache::{CacheRegistryBuilder, LayeredCacheRegistryBuilder};
 use common_meta::heartbeat::handler::parse_mailbox_message::ParseMailboxMessageHandler;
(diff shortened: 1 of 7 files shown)


src/cmd/src/frontend.rs
```diff
@@ -318,11 +319,19 @@ impl StartCommand {
             Arc::new(executor),
         );
 
+        // frontend to datanode need not timeout.
+        // Some queries are expected to take long time.
+        let channel_config = ChannelConfig {
+            timeout: None,
+            ..Default::default()
+        };
```
waynexia, line 327: Clippy makes me write this

fengjiachun, line 327: Why not put these in DatanodeClients#default ?

fengjiachun (approved):

killme2008:
I'm not sure if this behavior is expected. If the query request from the client has already been canceled or aborted, it doesn't make sense for the requests between the frontend and the datanode to continue processing. 
Personally, I think it would be better if this parameter could be configured. By default, it can be set to a relatively large value, leaving the choice ultimately up to the user.
Of course, it would be even better to properly implement request abort and cancel functionalities.

waynexia:
The cancellation should always be controlled by frontend protocol layer or the client. But not timeout-ed by the channel between datanode and frontend.

One related enhancement is we don't propagate this cancellation for now. I filed a issue for this https://github.com/GreptimeTeam/greptimedb/issues/3964

evenyag (approved):

PR #3966, opened 2024-05-16
feat: skip read-only region when trying to flush on region full

waynexia:

Read-only regions cannot be flushed. Skip them on trying to pick one when engine is full

## Checklist

### DIFF
--- a/src/mito2/src/worker/handle_flush.rs
+++ b/src/mito2/src/worker/handle_flush.rs
@@ -85,8 +85,8 @@ impl<S> RegionWorkerLoop<S> {
         let mut max_mem_region = None;
 
         for region in &regions {
-            if self.flush_scheduler.is_flush_requested(region.region_id) {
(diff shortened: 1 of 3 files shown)


fengjiachun (approved):
LGTM

src/mito2/src/worker/handle_flush.rs
```diff
@@ -85,8 +85,8 @@ impl<S> RegionWorkerLoop<S> {
         let mut max_mem_region = None;
 
         for region in &regions {
-            if self.flush_scheduler.is_flush_requested(region.region_id) {
-                // Already flushing.
+            if self.flush_scheduler.is_flush_requested(region.region_id) || !region.is_writable() {
```
evenyag, line 88: We should also check it in `flush_periodically()`.

evenyag (approved):

PR #3981, opened 2024-05-18
refactor:  make the command entry cleaner

zyy17:

**NOTE**: The PR is decoupled with draft PR https://github.com/GreptimeTeam/greptimedb/pull/3941, and it's just a simple refactoring based on the original implementation.

- **Remove** `Options{}` and some related methods, and all the building process of options can be in their app(for example, `datanode.rs`/`metasrv.rs`/...);

- **Remove** `SubCommand.build()` and `SubCommand.load_options()` which are unnecessary;

- **Move** `common_telemetry::init_global_logging` into `<app>:build()` which can be easy to receive the arguments;

- **Move** `start_app()` into `App` and rename `run()`;

- Make `common_telemetry::init_global_logging` init once, or it will fail in integration test (what an ugly function...);

## Checklist

### DIFF
--- a/src/cmd/src/bin/greptime.rs
+++ b/src/cmd/src/bin/greptime.rs
@@ -19,7 +19,7 @@ use std::fmt;
 use clap::{Parser, Subcommand};
 use cmd::error::Result;
 use cmd::options::{GlobalOptions, Options};
-use cmd::{cli, datanode, frontend, log_versions, metasrv, standalone, start_app, App};
+use cmd::{cli, datanode, frontend, log_versions, metasrv, standalone, App};
 use common_version::{short_version, version};
 
 #[derive(Parser)]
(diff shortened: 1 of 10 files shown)


src/cmd/src/bin/greptime.rs
```diff
@@ -56,87 +54,50 @@ enum SubCommand {
...
+            cmd.build(cmd.load_options(&cli.global_options)?)
+                .await?
+                .run()
+                .await
+        }
+        SubCommand::Cli(cmd) => {
+            cmd.build(cmd.load_options(&cli.global_options)?)
+                .await?
+                .run()
+                .await
```
MichaelScofield, line 98: You can use macro to remove repetition.

MichaelScofield (approved):

evenyag (approved):
LGTM

src/cmd/src/bin/greptime.rs
evenyag: The logic inside the match arm is simple. Not very necessary to write a macro for it.

PR #3992, opened 2024-05-20
feat: Implements row group level parallel unordered scanner

evenyag:

This PR changes the `UnorderedScan` to read parquet row groups in parallel to improve scan performance. It also implements the `RegionScanner` trait for `UnorderedScan`.
- Adds `build_parts` method to `ScanInput` to collect all `FileRanges` (row groups) and memtables
  - ranges and memtables are distributed by parallelism
- `UnorderedScan` can return `parallelism` streams
- Adds `CompatBatch` to adapt batch schema.
  - We might remove the `CompatReader` in the future as `CompatBatch` doesn't require implementing the `BatchReader` trait.

It also defines a `ScanPartBuilder` trait to allow different parallel scan strategies.
```rust
pub(crate) trait ScanPartBuilder {
    fn set_parallelism(&mut self, parallelism: usize);

    fn append_file_ranges(&mut self, file_ranges: impl Iterator<Item = FileRange>);

    fn build_parts(self, memtables: &[MemtableRef]) -> Vec<ScanPart>;
}
```

## Checklist

### DIFF
--- a/src/mito2/src/read/scan_region.rs
+++ b/src/mito2/src/read/scan_region.rs
@@ -602,11 +602,6 @@ impl ScanInput {
                     }
                 }
             };
-            common_telemetry::debug!(
-                "Build parts for file {}, row_groups: {}",
-                file.file_id(),
-                row_groups.len()
-            );
             if !compat::has_same_columns(
                 self.mapper.metadata(),
                 file_range_ctx.read_format().metadata(),
@@ -631,11 +626,7 @@ impl ScanInput {
 
         READ_SST_COUNT.observe(self.files.len() as f64);
 
-        if !self.enable_parallel_scan() {
-            common_telemetry::debug!(
-                "Parallel scan is disabled, {}",
-                self.parallelism.parallelism
-            );
+        if !self.parallelism.allow_parallel_scan() {
             // Returns a single part.
             let part = ScanPart {
                 memtables: self.memtables.clone(),
@@ -651,7 +642,7 @@ impl ScanInput {
         let mems_per_part = ((self.memtables.len() + parallelism - 1) / parallelism).max(1);
         let ranges_per_part = ((file_ranges.len() + parallelism - 1) / parallelism).max(1);
         common_telemetry::debug!(
-            "Parallel scan is enabled, {} ({}, {}), mems_per_part: {}, ranges_per_part: {}",
+            "Parallel scan is enabled, parallelism: {}, {} memtables, {} file_ranges, mems_per_part: {}, ranges_per_part: {}",
             self.parallelism.parallelism,
             self.memtables.len(),
             file_ranges.len(),
@@ -708,11 +699,6 @@ impl ScanInput {
                 }
             }
         });
-    }
-
-    /// Returns whether to scan in parallel.
-    pub(crate) fn enable_parallel_scan(&self) -> bool {
-        self.parallelism.allow_parallel_scan() && (self.files.len() + self.memtables.len()) > 1
     }
 }
 
--- a/src/mito2/src/read/scan_region.rs
+++ b/src/mito2/src/read/scan_region.rs
@@ -34,7 +34,7 @@ use crate::cache::CacheManagerRef;
 use crate::error::Result;
 use crate::memtable::MemtableRef;
 use crate::metrics::READ_SST_COUNT;
-use crate::read::compat::CompatReader;
+use crate::read::compat::{CompatBatch, CompatReader};
 use crate::read::projection::ProjectionMapper;
 use crate::read::seq_scan::SeqScan;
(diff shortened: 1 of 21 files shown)


src/mito2/src/read/unordered_scan.rs
```diff
@@ -14,148 +14,109 @@
...
+            for mut stream in streams {
+                while let Some(rb) = stream.next().await {
+                    yield rb;
+                }
             }
-
-            metrics.total_cost = query_start.elapsed();
-            Self::observe_metrics_on_finish(&metrics);
-            debug!("Unordered scan in parallel finished, region_id: {}, metrics: {:?}", mapper.metadata().region_id, metrics);
         };
```
waynexia, line 101: Why don't we return a vec of streams? It seems this is not really "paralleled"

src/mito2/src/sst/file.rs
```diff
@@ -175,9 +175,15 @@ impl FileHandle {
         self.inner.compacting.store(compacting, Ordering::Relaxed);
     }
 
+    /// Returns a copy of the [FileMeta].
     pub fn meta(&self) -> FileMeta {
         self.inner.meta.clone()
     }
```
waynexia, line 181: nit: this seems unnecessary as others can clone themselves on `meta_ref`

src/mito2/src/read/scan_region.rs
```diff
@@ -220,11 +227,9 @@ impl ScanRegion {
     }
 
     /// Unordered scan.
-    pub(crate) fn unordered_scan(self) -> Result<UnorderedScan> {
+    pub(crate) async fn unordered_scan(self) -> Result<UnorderedScan> {
         let input = self.scan_input(true)?;
-        let scan = UnorderedScan::new(input);
-
-        Ok(scan)
+        UnorderedScan::new(input).await
```
waynexia, line 232: Can you postpond reading parquet meta to `UnorderedScan::build_stream` like how `SeqScan` does? To avoid I/O in the plan phase

src/mito2/src/read/scan_region.rs
```diff
@@ -620,3 +687,43 @@ impl ScanInput {
...
+            f,
+            "ScanPart({} memtables, {} file ranges)",
+            self.memtables.len(),
+            self.file_ranges.len()
+        )
+    }
+}
+
+/// Strategy to build [ScanPart] for parallel scan.
+pub(crate) trait ScanPartBuilder {
```
waynexia, line 714: 😣 why `ScanPart` is a struct but its builder is a trait

src/mito2/src/read/scan_region.rs
```diff
@@ -572,6 +576,69 @@ impl ScanInput {
         Ok(sources)
     }
 
+    /// Builds and returns parts to read.
```
waynexia, line 579: Does the returned vec have the same length with `partition`?

src/mito2/src/read/scan_region.rs
evenyag: The SeqScanner needs to create parts with a different strategy.......

src/mito2/src/read/unordered_scan.rs
evenyag: Because some callers like your metric engine can't process a vec of streams, so we need a method to join all streams together.
https://github.com/GreptimeTeam/greptimedb/blob/a58256d4d3f031470dd765a8cbc849dbbba6863a/src/metric-engine/src/engine/read.rs#L54-L67

src/mito2/src/read/scan_region.rs
evenyag: It's hard to distribute row groups to partitions without reading the metadata. Since we might read different numbers of row groups from different files. I'll also modify the `SeqScan` in the next PR......

We will cache the metadata after we read it so we won't do I/O next time.

But I'll find some ways to address this later.

src/mito2/src/read/scan_region.rs
evenyag: Refactored in https://github.com/GreptimeTeam/greptimedb/pull/3992/commits/78163d791b8f6c327700850f5e107ef6ed49093d

Now it only prune files.

src/mito2/src/read/scan_region.rs
evenyag: Refactor in https://github.com/GreptimeTeam/greptimedb/pull/3992/commits/78163d791b8f6c327700850f5e107ef6ed49093d

Now this trait is replaced by a collector trait that only collects file ranges. A trait is still necessary because I need to use another container in SeqScan later.

fengjiachun (approved):
LGTM

waynexia:
🛫

PR #3994, opened 2024-05-21
ci: change the image name of nightly build

zyy17:

Change the image name of nightly build for avoiding conflict with the release images.

## Checklist

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -66,6 +66,10 @@ env:
 
   NIGHTLY_RELEASE_PREFIX: nightly
 
+  # Use the different image name to avoid conflict with the release images.
+  # The DockerHub image will be greptime/greptimedb-nightly.
+  IMAGE_NAME: greptimedb-nightly
+
 permissions:
   issues: write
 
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-build.yml
```diff
@@ -66,6 +66,10 @@ env:
 
   NIGHTLY_RELEASE_PREFIX: nightly
 
+  # Use the different image name to avoid conflict with the release images.
+  # The DockerHub image will be greptime/greptimedb-nightly.
+  IMAGE_NAME: greptimedb-nightly
```
tisonkun, line 71:
```suggestion
  IMAGE_NAME: greptimedb-weekly
```

Maybe we can distinguish two nightly image as "weekly" and "daily" according to their natural property?

evenyag (approved):

.github/workflows/nightly-build.yml
zyy17: Make sense. It should use `daily` in this PR. Let me change it.

zyy17: @tisonkun I just thought about it more carefully. I think we can:

- `nightly-build.yaml` -> use `greptime/greptimedb-nightly:nightly-xxx` because it's nightly, actually;
- scheduled weekly release -> use `greptime/greptimedb:v0.2.0-weekly-20230313`;

It's more precise and easy to classify. Of cause, we also should add the documents to describe our CD strategy.

tisonkun (approved):

PR #4005, opened 2024-05-21
fix: can't print log because the tracing guard is dropped

zyy17:

The bug is caused by https://github.com/GreptimeTeam/greptimedb/pull/3981.

It's a fast fix for adding `guard` in `Instance` to avoid dropping.

I think the elegant way to resolve the problem is to refactor the `init_global_logging()`, split it into multiple functions, and initialize it at the beginning of `main()`.

**We should release the new version of db after the PR is merged**. Fortunately, it's not in the v0.8.0.

## Checklist

### DIFF
--- a/src/cmd/src/cli.rs
+++ b/src/cmd/src/cli.rs
@@ -51,7 +51,6 @@ pub struct Instance {
     tool: Box<dyn Tool>,
 
     // Keep the logging guard to prevent the worker from being dropped.
-    #[allow(dead_code)]
     _guard: Vec<WorkerGuard>,
 }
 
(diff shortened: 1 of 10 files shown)


src/cmd/src/cli.rs
```diff
@@ -48,11 +49,15 @@ pub trait Tool: Send + Sync {
 
 pub struct Instance {
     tool: Box<dyn Tool>,
+
+    // Keep the logging guard to prevent the worker from being dropped.
+    #[allow(dead_code)]
+    guard: Vec<WorkerGuard>,
```
tisonkun, line 55:
```suggestion
    _guard: Vec<WorkerGuard>,
```

Idiom.

WenyXu:
Let's run `make fmt`

evenyag (approved):

waynexia (approved):

PR #4009, opened 2024-05-21
feat: add fallback logic for vmagent sending wrong content type

### LINKED ISSUE #3929: vmagent could send remote-write requests in Snappy compression after the VM protocol handshake
### What type of bug is this?

Unexpected error

Write Protocols

### Minimal reproduce step

As commented in #3641:

> vmagent could send remote-write requests in Snappy compression after the VM protocol handshake. 

This occurs when:
1. `vmagent` sends data in `Snappy`.
2. `vmagent` buffers data on disk when the remote-write target is down for upgrading or other reasons.
3. `vmagent` restarts after the remote-write target is up and performs the handshake.
4. `vmagent` sends data in `zstd`.

In the fourth step, `vmagent` actually needs to send the buffered (on-disk) data (in `Snappy`) before sending new data.

I mentioned this issue in https://jiekun.dev/posts/vmagent-data-structures (思考题2). VictoriaMetrics will handle such situations with fallback logic. It will try `Snappy` when `zstd` encounters an error.

I'm not familiar with Rust, so I haven't read the code and I'm not sure if this is also implemented in GreptimeDB. This could potentially cause `vmagent` to receive an error response and retry. I don't know if this is a necessary feature for GreptimeDB since it's hard to reproduce in production :)

---
Edit: It seems the related source codes is:
https://github.com/GreptimeTeam/greptimedb/blob/6ab3aeb14221cb7164845061a9ed56a7fcf82b33/src/servers/src/http/prom_store.rs#L250

It could be fixed like this (generated by ChatGPT):
```rust
let buf = match is_zstd {
    true => {
        match zstd_decompress(&body[..]) {
            Ok(result) => Bytes::from(result),
            Err(_) => snappy_decompress(&body[..]).map_err(|_| "Both decompression methods failed")?,
        }
    }
    false => {
        match snappy_decompress(&body[..]) {
            Ok(result) => Bytes::from(result),
            Err(_) => zstd_decompress(&body[..]).map_err(|_| "Both decompression methods failed")?,
        }
    }
};
```

Ubuntu 22.04 x64

v0.7.2

### Notes
As @zyy17 said this is not a bug of GreptimeDB. It's actually caused by `vmagent` who failed to send compressed data that match the protocol flag in HTTP header. I believe we could have a simple discussion here to see if GreptimeDB should have fallback logic for those edge case.

sunng87:

Fixes #3929 

Added fallback logic to handle situation when vmagent send wrong content type of data.

## Checklist

### DIFF
--- a/src/servers/src/http/prom_store.rs
+++ b/src/servers/src/http/prom_store.rs
@@ -239,9 +239,9 @@ pub async fn remote_read(
 
 fn try_decompress(is_zstd: bool, body: &[u8]) -> Result<Bytes> {
     Ok(Bytes::from(if is_zstd {
-        zstd_decompress(&body[..])?
+        zstd_decompress(body)?
     } else {
-        snappy_decompress(&body[..])?
+        snappy_decompress(body)?
     }))
 }
 
--- a/src/servers/src/http/prom_store.rs
+++ b/src/servers/src/http/prom_store.rs
@@ -257,7 +257,7 @@ async fn decode_remote_write_request(
 
     // due to vmagent's limitation, there is a chance that vmagent is
     // sending content type wrong so we have to apply a fallback with decoding
-    // the content is the other method.
+    // the content in another method.
     //
     // see https://github.com/VictoriaMetrics/VictoriaMetrics/issues/5301
     // see https://github.com/GreptimeTeam/greptimedb/issues/3929
--- a/src/servers/src/http/prom_store.rs
+++ b/src/servers/src/http/prom_store.rs
@@ -237,6 +237,14 @@ pub async fn remote_read(
     handler.read(request, query_ctx).await
 }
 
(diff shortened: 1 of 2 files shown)


src/servers/src/http/prom_store.rs
```diff
@@ -247,11 +255,18 @@ async fn decode_remote_write_request(
         .await
         .context(error::HyperSnafu)?;
 
-    let buf = Bytes::from(if is_zstd {
-        zstd_decompress(&body[..])?
+    // due to vmagent's limitation, there is a chance that vmagent is
+    // sending content type wrong so we have to apply a fallback with decoding
+    // the content is the other method.
```
evenyag, line 260:
```suggestion
    // the content in another method.
```

evenyag (approved):

zyy17 (approved):
LGTM

PR #4018, opened 2024-05-23
fix(flow): correctness bugs

discord9:

fix some correctness bugs

- fix reorder of insert requests columns
- fix choose the correct scalar function type
- fix combine of avg&tumble cause transform errors
- fix tumble ts < window start offset by one 

## Checklist

### DIFF
--- a/src/flow/src/adapter.rs
+++ b/src/flow/src/adapter.rs
@@ -319,7 +319,7 @@ impl FlownodeManager {
                 );
                 // TODO(discord9): bugged so we can't infer time index from flow plan, so we have to manually set one
                 let ts_col = ColumnSchema::new(
-                    "GREPTIME_TIMESTAMP",
+                    GREPTIME_TIMESTAMP,
                     ConcreteDataType::timestamp_millisecond_datatype(),
                     true,
                 )
--- a/src/flow/src/adapter.rs
+++ b/src/flow/src/adapter.rs
@@ -66,6 +66,7 @@ use error::Error;
 
 pub const PER_REQ_MAX_ROW_CNT: usize = 8192;
 
+// TODO: replace this with `GREPTIME_TIMESTAMP` before v0.9
 pub const AUTO_CREATED_PLACEHOLDER_TS_COL: &str = "__ts_placeholder";
 
 pub const UPDATE_AT_TS_COL: &str = "update_at";
--- a/src/flow/src/adapter.rs
+++ b/src/flow/src/adapter.rs
@@ -26,7 +26,6 @@ use common_base::Plugins;
 use common_error::ext::BoxedError;
 use common_frontend::handler::FrontendInvoker;
 use common_meta::key::TableMetadataManagerRef;
-use common_query::prelude::GREPTIME_TIMESTAMP;
 use common_runtime::JoinHandle;
 use common_telemetry::{debug, info};
 use datatypes::schema::ColumnSchema;
@@ -66,6 +65,10 @@ mod table_source;
 use error::Error;
 
 pub const PER_REQ_MAX_ROW_CNT: usize = 8192;
+
+pub const AUTO_CREATED_PLACEHOLDER_TS_COL: &str = "__ts_placeholder";
+
+pub const UPDATE_AT_TS_COL: &str = "update_at";
 
 // TODO: refactor common types for flow to a separate module
 /// FlowId is a unique identifier for a flow task
@@ -282,7 +285,7 @@ impl FlownodeManager {
                 let schema = meta.schema.column_schemas;
                 let is_auto_create = schema
                     .last()
-                    .map(|s| s.name == GREPTIME_TIMESTAMP)
+                    .map(|s| s.name == AUTO_CREATED_PLACEHOLDER_TS_COL)
                     .unwrap_or(false);
                 (primary_keys, schema, is_auto_create)
             } else {
@@ -314,13 +317,13 @@ impl FlownodeManager {
                     })
                     .unwrap_or_default();
                 let update_at = ColumnSchema::new(
-                    "update_at",
+                    UPDATE_AT_TS_COL,
                     ConcreteDataType::timestamp_millisecond_datatype(),
                     true,
                 );
                 // TODO(discord9): bugged so we can't infer time index from flow plan, so we have to manually set one
                 let ts_col = ColumnSchema::new(
-                    GREPTIME_TIMESTAMP,
+                    AUTO_CREATED_PLACEHOLDER_TS_COL,
                     ConcreteDataType::timestamp_millisecond_datatype(),
                     true,
                 )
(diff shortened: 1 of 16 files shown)


src/flow/src/adapter/flownode_impl.rs
```diff
@@ -100,12 +104,57 @@ impl Flownode for FlownodeManager {
...
+                if !fetch_order.iter().enumerate().all(|(i, &v)| i == v) {
+                    debug!("Reordering columns: {:?}", fetch_order)
+                }
+                fetch_order
+            };
+
             let rows: Vec<DiffRow> = rows_proto
                 .into_iter()
                 .map(repr::Row::from)
+                .map(|r| {
```
zhongzc, line 151: Combine them?

src/flow/src/adapter/node_context.rs
```diff
@@ -297,7 +303,7 @@ impl FlownodeContext {
             .get_by_name(table_name)
             .map(|(_, gid)| gid)
             .unwrap();
-        self.schema.insert(gid, schema);
+        self.schema.insert(gid, schema.into_named(vec![]));
```
zhongzc, line 306: We don't need to pass a name here?

src/flow/src/adapter/table_source.rs
```diff
@@ -123,25 +122,34 @@ impl TableSource {
...
+        let (col_names, column_types): (Vec<_>, Vec<_>) = raw_schema
             .column_schemas
+            .clone()
             .into_iter()
-            .map(|col| ColumnType {
-                nullable: col.is_nullable(),
-                scalar_type: col.data_type,
+            .map(|col| {
+                (
+                    col.name.clone(),
```
zhongzc, line 131: I guess the clone can be eliminated
```suggestion
                    col.name,
```

src/flow/src/transform/expr.rs
```diff
@@ -102,7 +102,7 @@ impl TypedExpr {
 
         match arg_len {
             // because variadic function can also have 1 arguments, we need to check if it's a variadic function first
```
zhongzc, line 104: Comments need to be updated

src/flow/src/transform/expr.rs
```diff
@@ -125,7 +125,13 @@ impl TypedExpr {
                 Ok(TypedExpr::new(arg.call_unary(func), ret_type))
             }
             // because variadic function can also have 2 arguments, we need to check if it's a variadic function first
```
zhongzc, line 127: ditto

src/flow/src/utils.rs
```diff
@@ -95,17 +96,33 @@ impl KeyExpiryManager {
             return Ok(None);
         };
 
+        self.event_ts_to_key
+            .entry(event_ts)
+            .or_default()
+            .insert(row.clone());
+
+        self.get_expire_duration(now, row)
```
zhongzc, line 104:
```suggestion
        if let Some(expire_time) = self.compute_expiration_timestamp(now) {
            if expire_time > event_ts {
                // return how much time it's expired
                return Ok(Some(expire_time - event_ts));
            }
        }
        
        Ok(None)
```

zhongzc (approved):

src/flow/src/compute/render/reduce.rs
```diff
@@ -397,6 +401,24 @@ fn reduce_accum_subgraph(
     // TODO(discord9): consider key-based lock
     let mut arrange = arrange.write();
     for (key, value_diffs) in key_to_vals {
+        if let Some(expire_man) = &arrange.get_expire_state() {
+            let mut is_expired = false;
+            err_collector.run(|| {
+                if let Some(expired) = expire_man.get_expire_duration(now, &key)? {
+                    is_expired = true;
+                    DataAlreadyExpiredSnafu {
```
discord9, line 409: Should late data arrive be collected as a error? Or separately be collected only by a distribution of time or just max time?

killme2008: Expired data should be ignored. Maybe a warning log is better.

src/flow/src/adapter.rs
```diff
@@ -281,7 +282,7 @@ impl FlownodeManager {
                 let schema = meta.schema.column_schemas;
                 let is_auto_create = schema
                     .last()
-                    .map(|s| s.name == "__ts_placeholder")
+                    .map(|s| s.name == GREPTIME_TIMESTAMP)
```
killme2008, line 285: Will it introduce a compibility issue?

discord9: > Will it introduce a compibility issue?

Indeed, and may cause other issues(like when user manually create a table with last col of name `GREPTIME_TIMESTAMP`, I should still instead use a column name that user usually will not  use like `__ts_placeholder`?

discord9: > > Will it introduce a compibility issue?
> 
> Indeed, and may cause other issues(like when user manually create a table with last col of name `GREPTIME_TIMESTAMP`, I should still instead use a column name that user usually will not use like `__ts_placeholder`?

now change back to prevent breaking change, will change in 0.9 when flow can correctly auto deduce table schema

src/flow/src/expr/error.rs
```diff
@@ -100,4 +100,11 @@ pub enum EvalError {
         #[snafu(implicit)]
         location: Location,
     },
+
+    #[snafu(display("Incoming data already expired by {} ms", expired_by))]
+    DataAlreadyExpired {
```
waynexia, line 105: I just realize that this error enum doesn't map to `StatusCode`?

src/flow/src/repr/relation.rs
```diff
@@ -262,6 +262,19 @@ impl RelationType {
...
+        RelationDesc { typ: self, names }
+    }
+
+    /// Return relation describe without column names
+    pub fn into_unnamed(self) -> RelationDesc {
+        RelationDesc {
+            typ: self,
+            names: vec![],
+        }
+    }
```
waynexia, line 277: Are those methods used somewhere? or it's for later usages

src/flow/src/transform/expr.rs
```diff
@@ -124,8 +123,13 @@ impl TypedExpr {
 
                 Ok(TypedExpr::new(arg.call_unary(func), ret_type))
             }
-            // because variadic function can also have 2 arguments, we need to check if it's a variadic function first
-            2 if VariadicFunc::from_str_and_types(fn_name, &arg_types).is_err() => {
+            2 if BinaryFunc::from_str_expr_and_type(
```
waynexia, line 126: Maybe add an unit test for function sig matching

src/flow/src/adapter.rs
```diff
@@ -66,6 +66,10 @@ use error::Error;
 
 pub const PER_REQ_MAX_ROW_CNT: usize = 8192;
 
+pub const AUTO_CREATED_PLACEHOLDER_TS_COL: &str = "__ts_placeholder";
```
waynexia, line 69:
```suggestion
// TODO: replace this with `GREPTIME_TIMESTAMP` before v0.9
pub const AUTO_CREATED_PLACEHOLDER_TS_COL: &str = "__ts_placeholder";
```

waynexia (approved):
Others LGTM

PR #4019, opened 2024-05-23
feat: enable tcp keepalive for http server

MichaelScofield:

In our customer's env, we've observed a lot of "dangling" http connections. They are all in the “established” state, even the client side pods were already destroyed long before. These dangling connections are harmful, somehow they consumed a lot of memory.

For the origin of the dangling connections, my guessing is that the client side pods are not gracefully shutdown, or the bad network. They are both common situations in cloud env, so I decide to add the keepalive option for http server: if the connections are idle for 1 hour, close it actively in the server side.

I'm a little hesitate to expose the keepalive option in the config file. For http connection, if not http2, rebuilding is very common. Any decent http connection pool can do that (there are keepalive options in themselves, too). So I think for the sake of simplicity, the keepalive option is hardcoded, and make it one hour long so to give the connections long enough time to say they are alive.

After adding the keepalive option, our customer says the memory issue is gone.

This PR also modify some codes to make greptimedb able to be integrated into other projects.

## Checklist

### DIFF
--- a/tests-integration/src/test_util.rs
+++ b/tests-integration/src/test_util.rs
@@ -25,9 +25,11 @@ use common_base::secrets::ExposeSecret;
 use common_config::Configurable;
 use common_meta::key::catalog_name::CatalogNameKey;
 use common_meta::key::schema_name::SchemaNameKey;
+use common_query::OutputData;
 use common_runtime::Builder as RuntimeBuilder;
 use common_telemetry::warn;
 use common_test_util::ports;
+use common_test_util::recordbatch::{check_output_stream, ExpectedOutput};
 use common_test_util::temp_dir::{create_temp_dir, TempDir};
 use common_wal::config::DatanodeWalConfig;
 use datanode::config::{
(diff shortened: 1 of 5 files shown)


sunng87 (approved):
LGTM. I thought this is not required for modern network stack. 

And do we need same configuration for gRPC, mysql and postgres?

tests-integration/src/test_util.rs
```diff
@@ -685,3 +688,19 @@ where
...
     test(endpoints).await
 }
+
+pub async fn execute_and_check_output(db: &Database, sql: &str, expected: ExpectedOutput<'_>) {
+    let output = db.sql(sql).await.unwrap();
+    let output = output.data;
+
+    match (&output, expected) {
+        (OutputData::AffectedRows(x), ExpectedOutput::AffectedRows(y)) => {
+            assert_eq!(*x, y, "actual: \n{}", x)
```
evenyag, line 698: Do we need to print the expected value?

evenyag (approved):

MichaelScofield:
> LGTM. I thought this is not required for modern network stack.
> 
> And do we need same configuration for gRPC, mysql and postgres?

Thought about that. We can enable http server keepalive like this is because the protocol used here(http 1.1) is simple enough. For the other three, there might be their specific mechanism for this kind of purpose. For example, grpc has "ping" based keepalive made into its protocol layer(not tcp layer). And jdbc seems to have its own keepalive parameter as well. So I don't want to all simply blindly enable the underlying tcp keepalive for them.

PR #4052, opened 2024-05-27
feat: open region in background

WenyXu:

## Checklist


src/mito2/src/worker/handle_open.rs
```diff
@@ -71,18 +86,33 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
         .skip_wal_replay(request.skip_wal_replay)
-        .parse_options(request.options)?
-        .cache(Some(self.cache_manager.clone()))
-        .open(&self.config, &self.wal)
-        .await?;
-
-        info!("Region {} is opened", region_id);
+        .parse_options(request.options)
+        {
+            Ok(opener) => opener.cache(Some(self.cache_manager.clone())),
```
evenyag, line 91: You could move `.cache(Some(self.cache_manager.clone()))` in front of `parse_options()`.

src/mito2/src/worker/handle_open.rs
```diff
@@ -71,18 +86,33 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+        common_runtime::spawn_bg(async move {
+            match opener.open(&config, &wal).await {
+                Ok(region) => {
+                    info!("Region {} is opened", region_id);
+                    REGION_COUNT.inc();
 
-        // Insert the MitoRegion into the RegionMap.
-        self.regions.insert_region(Arc::new(region));
+                    // Insert the MitoRegion into the RegionMap.
+                    regions.insert_region(Arc::new(region));
```
evenyag, line 108: Looks like we can open the same region multiple times.

src/mito2/src/worker/handle_open.rs
```diff
@@ -66,6 +66,10 @@ impl<S: LogStore> RegionWorkerLoop<S> {
         request: RegionOpenRequest,
         sender: OptionOutputTx,
     ) {
+        if self.opening_regions.is_region_exists(region_id) {
+            sender.send(Ok(0));
+            return;
+        };
```
evenyag, line 72: Maybe we should not return ok directly. If the opener fails in the background, then the region isn't opened but we return ok.

One solution is adding the sender to the `opening_regions` map and notifying the result later.

src/mito2/src/worker/handle_open.rs
```diff
@@ -59,10 +59,34 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
         }
 
+        Ok(())
+    }
+
+    pub(crate) async fn handle_open_request(
+        &mut self,
+        region_id: RegionId,
+        request: RegionOpenRequest,
+        sender: OptionOutputTx,
```
fengjiachun, line 69: I wonder if it could have a more meaningful name?

fengjiachun (approved):
Rest LGTM

src/mito2/src/request.rs
```diff
@@ -384,11 +384,11 @@ pub(crate) fn validate_proto_value(
 
 /// Oneshot output result sender.
 #[derive(Debug)]
-pub(crate) struct OutputTx(Sender<Result<AffectedRows>>);
+pub(crate) struct OutputTx(Sender<Result<AffectedRows, Error>>);
 
 impl OutputTx {
     /// Creates a new output sender.
-    pub(crate) fn new(sender: Sender<Result<AffectedRows>>) -> OutputTx {
+    pub(crate) fn new(sender: Sender<Result<AffectedRows, Error>>) -> OutputTx {
```
evenyag, line 391:
```suggestion
    pub(crate) fn new(sender: Sender<Result<AffectedRows>>) -> OutputTx {
```

src/mito2/src/worker/handle_open.rs
```diff
@@ -71,18 +95,43 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+            Ok(opener) => opener,
+            Err(err) => {
+                sender.send(Err(err));
+                return;
+            }
+        };
 
-        REGION_COUNT.inc();
+        let regions = self.regions.clone();
+        let wal: crate::wal::Wal<S> = self.wal.clone();
```
evenyag, line 109: Do we need this annotation?
```suggestion
        let wal = self.wal.clone();
```

src/mito2/src/worker/handle_open.rs
```diff
@@ -59,10 +59,34 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+            .opening_regions
+            .wait_for_opening_region(region_id, sender)
+        else {
+            return;
+        };
+
+        if self.regions.is_region_exists(region_id) {
+            sender.send(Ok(0));
+            return;
+        }
```
evenyag, line 81: prefer to check existence first.
```suggestion
        if self.regions.is_region_exists(region_id) {
            sender.send(Ok(0));
            return;
        }

        let Some(sender) = self
            .opening_regions
            .wait_for_opening_region(region_id, sender)
        else {
            return;
        };
```

src/mito2/src/worker/handle_open.rs
```diff
@@ -71,18 +95,43 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+                    info!("Region {} is opened", region_id);
+                    REGION_COUNT.inc();
 
-        // Insert the MitoRegion into the RegionMap.
-        self.regions.insert_region(Arc::new(region));
+                    // Insert the Region into the RegionMap.
+                    regions.insert_region(Arc::new(region));
 
-        Ok(0)
+                    let senders = opening_regions.remove_sender(region_id);
```
evenyag, line 122: Be careful. It's easy to have concurrent issues if we check these two maps in a different order. Maybe we should insert the region into the `regions` map while holding the `opening_regions` lock.

evenyag (approved):
Rest LGTM

PR #3820, opened 2024-04-27
refactor: move database client to test

dimbtp:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

Closes #3798

## What's changed and what's your intention?

- replace `Database` client in `cli export`
- move `Database` to `tests-integration`

## Checklist

- [ ]  I have written the necessary rustdoc comments.
- [ ]  I have added the necessary unit tests and integration tests.
- [x]  This PR does not require documentation updates.

PR #4057, opened 2024-05-27
refactor: move database client to tests/common

sunng87:

Resolves dependencies issue introduced in #3820

This patch extract `database.rs` to standalone module to avoid introducing whole dependency tree of tests-integration.

Also it removes dependency from `benchmarks` which is obsolete.

## Checklist


src/meta-srv/Cargo.toml
```diff
@@ -61,7 +61,7 @@ url = "2.3"
 
 [dev-dependencies]
 chrono.workspace = true
-client = { workspace = true, features = ["testing"] }
```
sunng87, line 64: another question about this: because we already have `testing` feature in `client`, is it really necessary to move `database.rs` out of `client`?

sunng87: and we do have only one testing function in client for which we can't remove `testing` from `client` at the moment.

tisonkun: I agree we can move this struct to `client` crate behind the `testing` feature.

IIRC previously it's in the `client` crate without `testing` feature in control.

sunng87: I think we can close this one, revert #3820, and create a new patch to feature-gate `database.rs`

tisonkun: Yeah. This is possible. Would you follow these changes?

tisonkun:
I'm manually reverting this. The related files changed a bit .

PR #4063, opened 2024-05-28
feat: set global runtime size by config file

MichaelScofield:

Originally the size of all global runtimes are hardcoded to 8, which is not flexible enough. For example, edit region requests are executed in the global background runtime. Our customer may emit a large volume of that requests, expecting them to be ran fast, only to find the runtime is saturated.

This PR makes the size of global runtimes configurable, by env var. The reason I choose env var instead of config file entries is because the parameters are not related to any kinds of greptimedb components. Whether it's standalone, frontend or datanode, doesn't matter, all need these parameters. So I don't want to set it in config files. Besides, env vars are good in cloud env, especially when greptimedb is started by helm chart.

## Checklist

### DIFF
--- a/src/cmd/src/options.rs
+++ b/src/cmd/src/options.rs
@@ -13,6 +13,9 @@
 // limitations under the License.
 
 use clap::Parser;
+use common_config::Configurable;
+use common_runtime::global::RuntimeOptions;
+use serde::{Deserialize, Serialize};
 
 #[derive(Parser, Default, Debug, Clone)]
 pub struct GlobalOptions {
@@ -28,4 +31,23 @@ pub struct GlobalOptions {
     #[clap(long, value_name = "TOKIO_CONSOLE_ADDR")]
     #[arg(global = true)]
     pub tokio_console_addr: Option<String>,
+}
+
+// TODO(LFC): Move logging and tracing options into global options, like the runtime options.
+/// All the options of GreptimeDB.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq)]
+#[serde(default)]
+pub struct GreptimeOptions<T> {
+    /// The runtime options.
+    pub runtime: RuntimeOptions,
+
+    /// The options of each component (like Datanode or Standalone) of GreptimeDB.
+    #[serde(flatten)]
+    pub component: T,
+}
+
+impl<T: Configurable> Configurable for GreptimeOptions<T> {
+    fn env_list_keys() -> Option<&'static [&'static str]> {
+        T::env_list_keys()
+    }
 }
--- a/src/common/runtime/Cargo.toml
+++ b/src/common/runtime/Cargo.toml
@@ -13,13 +13,15 @@ common-error.workspace = true
 common-macro.workspace = true
 common-telemetry.workspace = true
 lazy_static.workspace = true
+num_cpus.workspace = true
 once_cell.workspace = true
 paste.workspace = true
 prometheus.workspace = true
+serde.workspace = true
 snafu.workspace = true
 tokio.workspace = true
 tokio-metrics = "0.3"
-tokio-metrics-collector = "0.2"
+tokio-metrics-collector = { git = "https://github.com/MichaelScofield/tokio-metrics-collector.git", rev = "89d692d5753d28564a7aac73c6ac5aba22243ba0" }
 tokio-util.workspace = true
 
 [dev-dependencies]
--- a/src/common/runtime/src/global.rs
+++ b/src/common/runtime/src/global.rs
@@ -40,11 +40,11 @@ pub struct RuntimeOptions {
 
 impl Default for RuntimeOptions {
     fn default() -> Self {
-        let two_fold_cpus = num_cpus::get() * 2;
+        let cpus = num_cpus::get();
         Self {
-            read_rt_size: two_fold_cpus,
-            write_rt_size: two_fold_cpus,
-            bg_rt_size: two_fold_cpus,
+            read_rt_size: cpus,
+            write_rt_size: cpus,
+            bg_rt_size: cpus,
         }
     }
 }
--- a/src/common/runtime/src/global.rs
+++ b/src/common/runtime/src/global.rs
@@ -19,12 +19,35 @@ use std::sync::{Mutex, Once};
 use common_telemetry::info;
 use once_cell::sync::Lazy;
 use paste::paste;
+use serde::{Deserialize, Serialize};
 
 use crate::{Builder, JoinHandle, Runtime};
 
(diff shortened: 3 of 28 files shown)


MichaelScofield:
This PR also makes some changes to let greptimedb codes able to be embedded by other projects.

evenyag:
I think we should expose them to the config file

killme2008:
> I think we should expose them to the config file

Agree, like `[runtime]` section.

src/common/runtime/Cargo.toml
```diff
@@ -13,13 +13,14 @@ common-error.workspace = true
...
 lazy_static.workspace = true
+num_cpus.workspace = true
 once_cell.workspace = true
 paste.workspace = true
 prometheus.workspace = true
 snafu.workspace = true
 tokio.workspace = true
 tokio-metrics = "0.3"
-tokio-metrics-collector = "0.2"
+tokio-metrics-collector = { git = "https://github.com/MichaelScofield/tokio-metrics-collector.git", rev = "89d692d5753d28564a7aac73c6ac5aba22243ba0" }
```
tisonkun, line 23: What is this fork for?

src/common/runtime/src/global.rs
```diff
@@ -75,13 +71,34 @@ impl GlobalRuntimes {
     define_spawn!(bg);
 
     fn new(read: Option<Runtime>, write: Option<Runtime>, background: Option<Runtime>) -> Self {
+        fn env_rt_size(name: &str) -> usize {
+            std::env::var(name)
+                .ok()
+                .and_then(|v| v.parse().ok())
```
tisonkun, line 77: I suppose a malformed config deserves a warning or even we should panic.

It's subtle that user configures `GLOBAL_READ_RT_SIZE=8o` or other unintentional typos but we silently ignore it.

tisonkun (changes requested):
For https://github.com/GreptimeTeam/greptimedb/pull/4063/files#r1617995047.

src/common/runtime/Cargo.toml
MichaelScofield: I was fixing a typo in the original release.

MichaelScofield:
@evenyag @killme2008 @tisonkun PTAL

src/cmd/src/options.rs
```diff
@@ -29,3 +32,22 @@ pub struct GlobalOptions {
...
     pub tokio_console_addr: Option<String>,
 }
+
+// TODO(LFC): Move logging and tracing options into global options, like the runtime options.
+/// All the options of GreptimeDB.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq)]
+#[serde(default)]
+pub struct GreptimeOptions<T> {
+    /// The runtime options.
+    pub runtime: RuntimeOptions,
```
evenyag, line 42: Why not put the runtime options to the `GlobalOptions`?

src/common/runtime/src/global.rs
```diff
@@ -19,13 +19,36 @@ use std::sync::{Mutex, Once};
 use common_telemetry::info;
 use once_cell::sync::Lazy;
 use paste::paste;
+use serde::{Deserialize, Serialize};
 
 use crate::{Builder, JoinHandle, Runtime};
 
 const READ_WORKERS: usize = 8;
 const WRITE_WORKERS: usize = 8;
 const BG_WORKERS: usize = 8;
```
evenyag, line 28: Do we still need them?

src/common/runtime/src/global.rs
```diff
@@ -19,13 +19,36 @@ use std::sync::{Mutex, Once};
...
+}
+
+impl Default for RuntimeOptions {
+    fn default() -> Self {
+        let two_fold_cpus = num_cpus::get() * 2;
+        Self {
+            read_rt_size: two_fold_cpus,
+            write_rt_size: two_fold_cpus,
+            bg_rt_size: two_fold_cpus,
+        }
```
evenyag, line 48: Using 2x num_cpus as runtime size may be too large. I prefer to use num_cpus or a smaller size.

evenyag (commented):
@zyy17 This PR contains some changes to the Configurable trait and options. Could you please also take a look at this?

src/cmd/src/options.rs
MichaelScofield: Because `GlobalOptions` will be parsed by clap. We have to add some `#[clap]` annotations to `RuntimeOptions` fields which is not good considering it's in common-runtime crate.

src/common/runtime/src/global.rs
MichaelScofield: Right now yes, although I think the global runtime init can be refactor to a more clear way. But that doesn't have to be in this PR.

src/common/runtime/src/global.rs
MichaelScofield: I think 2x cpus are better, each core can deal with more work, improve general throughput, just like intel's hyper-threading. But I'm OK with setting them to cpu num. Let's hear other people say.

evenyag: `num_cpus` already counts logical core num. Note that we have 3 runtimes so we already use 3x cpu num.

tisonkun (approved):

src/cmd/src/options.rs
evenyag: I think we could move runtime options to `StandalonOptions`/`DatanodeOptions`/`FrontendOptions`. The `GreptimeOptions` also relies on `serde(flatten)` which sometimes doesn't work well. So I prefer not introducing this struct.
https://github.com/search?q=repo%3Aserde-rs%2Fserde+flatten&type=issues

MichaelScofield: There are already four global options repeated in `*Options`: `LoggingOptions`, `enable_telemetry`, `ExportMetricsOption` and `TracingOptions`. I ready don't like `RuntimeOptions` to be the fifth, and more to come in the future.

`serde(flatten)` is useful, and it's already used inside our codes. There are issues about it, true, and that's why I write the tests to read our example toml files into each `*Options`, to make sure it works as expected.

Putting the global options together makes our code more concise. If this PR is ok to you, I'm planning to refactor other four global options like this.

evenyag: > There are issues about it, true, and that's why I write the tests to read our example toml files into each *Options

Note that the example file contains all options but `serde(flatten)` might break `serde(default)`. I have tested some cases manually and haven't found any issues.

evenyag (approved):
I have manually tested some cases. This PR should be ready for merge, although I think we should introduce the `GreptimeOptions` in another PR.

cc @zyy17

evenyag:
The CI is failed.

Issue #3887, opened 2024-05-08
the timezone parameter in jdbc url is not used


MichaelScofield:
### What type of bug is this?

User Experience

### What subsystems are affected?

Frontend

### Minimal reproduce step

create table: 

```sql
create table test(ts timestamp time index, a DateTime);
```

then insert data:

```sql
insert into test values(1, '2024-05-08 08:00:00 +0800');
```

then execute:

```java
String sql = "select * from test";

try (Connection conn = DriverManager.getConnection("jdbc:mysql://127.0.0.1:4002?serverTimezone=Asia/Shanghai")) {
    Statement statement = conn.createStatement();
    ResultSet resultSet = statement.executeQuery(sql);
    while (resultSet.next()) {
        System.out.println(resultSet.getObject(2));
    }
}
```

### What did you expect to see?

the right (+8 timezone) date time: "2024-05-08 08:00:00"

### What did you see instead?

utc date time: "2024-05-08T00:00"

### What operating system did you use?

macos

### What version of GreptimeDB did you use?

main

killme2008:
From the MySQL connector doc: https://dev.mysql.com/doc/connector-j/en/connector-j-connp-props-datetime-types-processing.html

> This option itself does not set MySQL server session variable 'time_zone' to the given value. To do that the 'forceConnectionTimeZoneToSession' connection option must be set to "true".

So the correct URL is `jdbc:mysql://127.0.0.1:4002?serverTimezone=Asia/Shanghai&forceConnectionTimeZoneToSession`.

The connection option 'serverTimezone' is an alias of `connectionTimeZone ` but may be deprecated in the future.

After setting as above, the JDBC driver sends `SET SESSION time_zone='Asia/Shanghai'` to the server, but unfortunately, we don't process it properly.

killme2008:
According to https://dev.mysql.com/doc/refman/8.3/en/set-variable.html

> To assign a value to a session system variable, precede the variable name by the SESSION or LOCAL keyword, by the @@SESSION., @@LOCAL., or @@ qualifier, or by no keyword or no modifier at all:
> 
> SET SESSION sql_mode = 'TRADITIONAL';
> SET LOCAL sql_mode = 'TRADITIONAL';
> SET @@SESSION.sql_mode = 'TRADITIONAL';
> SET @@LOCAL.sql_mode = 'TRADITIONAL';
> SET @@sql_mode = 'TRADITIONAL';
[...]

PR #4064, opened 2024-05-28
fix: set local or session time_zone not work

killme2008:

See https://github.com/GreptimeTeam/greptimedb/issues/3887

The doc PR https://github.com/GreptimeTeam/docs/pull/988

Fix `SET SESSION time_zone='Asia/Shanghai'` or `SET LOCAL time_zone='Asia/Shanghai'`  not work.

## Checklist

### DIFF
--- a/src/sql/src/parsers/set_var_parser.rs
+++ b/src/sql/src/parsers/set_var_parser.rs
@@ -29,11 +29,9 @@ impl<'a> ParserContext<'a> {
             SpStatement::SetVariable {
                 variable,
                 value,
-                local,
                 hivevar,
-            } if !local && !hivevar => {
-                Ok(Statement::SetVariables(SetVariables { variable, value }))
-            }
+                ..
+            } if !hivevar => Ok(Statement::SetVariables(SetVariables { variable, value })),
             unexp => error::UnsupportedSnafu {
                 sql: self.sql.to_string(),
                 keyword: unexp.to_string(),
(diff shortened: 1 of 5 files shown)


sunng87 (approved):

tisonkun (commented):
I don't find logics to distinguish local variables and global variables on planning. Is it intended?

src/sql/src/parsers/set_var_parser.rs
tisonkun, line 1: sqlparser distinguish PostgreSQL-specific statements `SET TIME ZONE <value>` as `SetTimeZone { local: bool, value: Expr }`. Perhaps we can handle it here as well.

tisonkun (approved):
I noticed that "local" is the default manner for setting session variable.

While global variable starts with `@@`, we don't support any of them now:

```rust
Statement::SetVariables(set_var) => {
let var_name = set_var.variable.to_string().to_uppercase();
match var_name.as_str() {
// known vars
    _ => {
        return NotSupportedSnafu {
            feat: format!("Unsupported set variable {}", var_name),
        }
        .fail()
    }
}}
```

https://dev.mysql.com/doc/refman/8.0/en/set-variable.html

src/sql/src/parsers/set_var_parser.rs
tisonkun: This can be a follow-up. But I won't push this PR to the merge queue so that @killme2008 can decide if it would be included in this patch.

killme2008:
> I noticed that "local" is the default manner for setting session variable.
> 
> While global variable starts with `@@`, we don't support any of them now:
> 
> ```rust
> Statement::SetVariables(set_var) => {
> let var_name = set_var.variable.to_string().to_uppercase();
> match var_name.as_str() {
> // known vars
>     _ => {
>         return NotSupportedSnafu {
>             feat: format!("Unsupported set variable {}", var_name),
>         }
>         .fail()
>     }
> }}
> ```
> 
> https://dev.mysql.com/doc/refman/8.0/en/set-variable.html

We don't have a plan to set global variables. This PR is trying to fix that `set local {var}` and `set session {var}` are not working, they are the same as `set {var}`.

src/sql/src/parsers/set_var_parser.rs
killme2008: Fixed via https://github.com/GreptimeTeam/greptimedb/pull/4064/commits/43e173d1465fbe69eec0a66b99d2337ac0afdab6

PR #4068, opened 2024-05-29
feat: avoid some cloning when mirror requests to flownode

fengjiachun:

1. Do not spawn to check `source table` since there already has cache for this.
2. Avoid to cloning all requests to mirror.

## Checklist

### DIFF
--- a/src/operator/src/insert.rs
+++ b/src/operator/src/insert.rs
@@ -15,7 +15,6 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use api::region::RegionResponse;
 use api::v1::alter_expr::Kind;
 use api::v1::region::{InsertRequests as RegionInsertRequests, RegionRequestHeader};
 use api::v1::{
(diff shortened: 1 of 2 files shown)


src/operator/src/insert.rs
```diff
@@ -238,8 +202,34 @@ impl Inserter {
...
+                    .context(JoinTaskSnafu)
+                {
+                    warn!(err; "Failed to insert data into flownode");
+                }
+            }
+            Err(err) => warn!(err; "Failed to mirror request to flownode"),
+        }
+
+        let write_tasks = self
+            .group_requests_by_peer(requests)
```
fengjiachun, line 232: Avoid cloning the request

src/operator/src/insert.rs
```diff
@@ -294,7 +286,7 @@ impl Inserter {
 
                     if !peers.is_empty() {
                         let mut reqs = RegionInsertRequests::default();
-                        reqs.requests.push(req);
+                        reqs.requests.push(req.clone());
```
fengjiachun, line 289: Only clone those that are necessary

src/operator/src/insert.rs
```diff
@@ -310,14 +302,26 @@ impl Inserter {
...
                 inserts
-                    .entry(flownode.clone())
+                    .entry(peers[0].clone())
                     .or_default()
                     .requests
-                    .extend(reqs.requests.clone());
+                    .extend(reqs.requests);
+                continue;
+            } else {
+                // TODO(discord9): need to split requests to multiple flownodes
```
fengjiachun, line 314: cc @discord9

src/operator/src/insert.rs
```diff
@@ -238,8 +202,34 @@ impl Inserter {
...
 
-        let tasks = self
-            .group_requests_by_peer(requests.clone())
+        // Mirror requests for source table to flownode
+        match self.mirror_flow_node_requests(&requests).await {
+            Ok(flow_requests) => {
+                let node_manager = self.node_manager.clone();
+                let flow_tasks = flow_requests.into_iter().map(|(peer, inserts)| {
+                    let node_manager = node_manager.clone();
+                    common_runtime::spawn_write(async move {
```
waynexia, line 211: how about `spawn_bg` to avoid impact front-ground write

fengjiachun: Looks good.

src/operator/src/insert.rs
discord9: make sense since we are only at standalone mode now

src/operator/src/insert.rs
```diff
@@ -238,8 +202,34 @@ impl Inserter {
             ..Default::default()
         });
 
-        let tasks = self
-            .group_requests_by_peer(requests.clone())
+        // Mirror requests for source table to flownode
+        match self.mirror_flow_node_requests(&requests).await {
```
killme2008, line 206: What about adding a metric here?

killme2008 (approved):
Good job!

discord9 (approved):

src/operator/src/insert.rs
fengjiachun: 1. Added mirror rows count metric,
2. We already have metric in the cache manager

PR #4070, opened 2024-05-29
fix: avoid acquiring lock during reading stats

WenyXu:

This PR uses `Arc<Atomicu64>` to store the shared value instead of `RwLock`, to avoid acquiring lock during reading value.

The datanode will report the stats (like manifest usage) via heartbeat. Before sending the heartbeat to the metasrv, it collects all stats from the region server. While collecting `disk_usage` will invoke the `manifest_usage` method on the `manifest_manager`, this operation is required to acquire the read lock of the `manifest_manager`. 

```rust
       // read manifest_usage
       let manifest_usage = self
            .manifest_ctx
            .manifest_manager
            .read()
            .await
            .manifest_usage();
``` 

```rust
      let mut manager = self.manifest_manager.write().await;
      // update manifest
      // ..
```
However, acquiring a read lock will be blocked by the updating manifest operations, and **the wait time may be as long as tens of seconds**. This makes the datanode fail to send heartbeats in time, **then all region leases on the datanodes will expire.**

## Checklist

### DIFF
--- a/src/mito2/src/region.rs
+++ b/src/mito2/src/region.rs
@@ -20,7 +20,7 @@ pub(crate) mod version;
 
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
-use std::sync::atomic::{AtomicI64, Ordering};
+use std::sync::atomic::{AtomicI64, AtomicU64, Ordering};
 use std::sync::{Arc, RwLock};
 
 use common_telemetry::{error, info, warn};
@@ -106,6 +106,8 @@ pub(crate) struct MitoRegion {
     time_provider: TimeProviderRef,
     /// Memtable builder for the region.
     pub(crate) memtable_builder: MemtableBuilderRef,
+    /// Region stats
+    stats: Stats,
 }
 
 pub(crate) type MitoRegionRef = Arc<MitoRegion>;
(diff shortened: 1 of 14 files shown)


killme2008:
Good catch!

src/mito2/src/region.rs
```diff
@@ -526,6 +522,18 @@ impl OpeningRegions {
 
 pub(crate) type OpeningRegionsRef = Arc<OpeningRegions>;
 
+/// Single region stats.
+#[derive(Default, Debug, Clone)]
+pub(crate) struct Stats {
+    total_manifest_size: Arc<AtomicU64>,
```
killme2008, line 528: Looks like the total data size for the single region is missing.

WenyXu: The `Stats` stores the single region statistics.

killme2008: Yep, it only contains the manifest size, can it include the data size too?

WenyXu: I got you, we can introduce it in the future. Let's fix the bug first.

src/mito2/src/region.rs
```diff
@@ -526,6 +522,18 @@ impl OpeningRegions {
 
 pub(crate) type OpeningRegionsRef = Arc<OpeningRegions>;
 
+/// Single region stats.
+#[derive(Default, Debug, Clone)]
+pub(crate) struct Stats {
```
evenyag, line 527: Maybe we can rename it to ManifestStats?

waynexia (approved):

evenyag (approved):

src/mito2/src/region.rs
fengjiachun: +1

fengjiachun (approved):
LGTM

PR #4073, opened 2024-05-30
feat(flow): make write path faster with shared lock

discord9:

make flow's insert request faster to handle

Previously everytime a insert request comes in a exclusive lock on flow node's ctx is acquired, which is unnecessary, hence now it will only acquire a shared read lock, and send rows into flow node

## Checklist

### DIFF
--- a/src/flow/src/adapter/node_context.rs
+++ b/src/flow/src/adapter/node_context.rs
@@ -147,7 +147,7 @@ impl FlownodeContext {
     /// flush all sender's buf
     ///
     /// return numbers being sent
-    pub async fn flush_all_sender(&mut self) -> Result<usize, Error> {
+    pub async fn flush_all_sender(&self) -> Result<usize, Error> {
         let mut sum = 0;
         for sender in self.source_sender.values() {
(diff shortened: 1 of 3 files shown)


src/flow/src/adapter/node_context.rs
```diff
@@ -128,30 +133,35 @@ impl FlownodeContext {
...
-    pub fn flush_all_sender(&mut self) -> Result<usize, Error> {
-        self.source_sender
-            .iter_mut()
-            .map(|(_table_id, src_sender)| src_sender.try_send_all())
-            .try_fold(0, |acc, x| x.map(|x| x + acc))
+    pub async fn flush_all_sender(&self) -> Result<usize, Error> {
+        let mut sum = 0;
+        for sender in self.source_sender.values() {
+            sender.try_send_all().await.map(|x| sum += x)?;
+        }
```
zhongzc, line 154:
```suggestion
        for sender in self.source_sender.values() {
            sender.try_send_all().await.inspect(|x| sum += x)?;
        }
```

zhongzc (approved):

waynexia (approved):

PR #4076, opened 2024-05-30
test: add fuzz tests for column data type alteration

### LINKED ISSUE #3973: Add fuzz tests for column data type alteration

Tech debt reduction

We support to alter column data type in #3517, Let's add fuzz tests for this feature. 

See also: 
- #3972
- #3174 
- #3517

### Implementation challenges

_No response_

realtaobo:

#3973 

Just add fuzz test for `Alter Table <table_name> MODIFY COLUMN <column_name> <column_type>`

## Checklist


WenyXu:
Would you like to add the target to CI cfg
https://github.com/GreptimeTeam/greptimedb/blob/f024054ed34f722533110fa43468e1079edf2750/.github/workflows/develop.yml#L144
https://github.com/GreptimeTeam/greptimedb/blob/f024054ed34f722533110fa43468e1079edf2750/.github/workflows/develop.yml#L286

realtaobo:
@WenyXu Sure.
BTW, it seems not work to read env var `GT_MYSQL_ADDR` form .env config file.
https://github.com/GreptimeTeam/greptimedb/blob/f024054ed34f722533110fa43468e1079edf2750/tests-fuzz/src/utils.rs#L38-L41
If I don't set Environment variables，I got `"GT_MYSQL_ADDR is empty, ignores test"`, even though I have a .env file int the root.  How should I solve this problem?
```sh
ProductName:            macOS
ProductVersion:         14.5
BuildVersion:           23F79
```
**Exec cmd:** `cargo fuzz run fuzz_alter_table --fuzz-dir tests-fuzz -D -s none -- -runs=1`

WenyXu:
> @WenyXu It seems unnecessary. I added it to `fuzz_alter_table` as a part of alter table. What do you think？

Cool, I got you

> BTW, it seems not work to read env var GT_MYSQL_ADDR form .env config file.

Interesting, it works on my machine.

> If I don't set Environment variables，I got "GT_MYSQL_ADDR is empty, ignores test", even though I have a .env file int the root. How should I solve this problem?

What does the directory structure and .env file look like? (Don't submit any secrets)

realtaobo:
> What does the directory structure and .env file look like? (Don't submit any secrets)

The content is the same as file [.env.example](https://github.com/GreptimeTeam/greptimedb/blob/main/.env.example), only `GT_MYSQL_ADDR` is changed to:
```sh
GT_MYSQL_ADDR=127.0.0.1:4002
```

WenyXu:
> The content is the same as file [.env.example](https://github.com/GreptimeTeam/greptimedb/blob/main/.env.example), only `GT_MYSQL_ADDR` is changed to:

I don't have any idea about this👀

realtaobo:
> I don't have any idea about this👀

Got it, a little problem, maybe I should restart my machine 🐶

tests-fuzz/src/ir.rs
```diff
@@ -355,6 +355,20 @@ pub fn droppable_columns(columns: &[Column]) -> Vec<&Column> {
...
+/// Returns columns that can use the alter table modify command
+pub fn modifiable_columns(columns: &[Column]) -> Vec<&Column> {
+    columns
+        .iter()
+        .filter(|column| {
+            !column.options.iter().any(|option| {
+                option == &ColumnOption::PrimaryKey
+                    || option == &ColumnOption::TimeIndex
+                    || option == &ColumnOption::NotNull
+            })
```
WenyXu, line 367: The `Nullable` and `TimeIndex` columns should also be modifiable. cc @evenyag

tests-fuzz/src/context.rs
```diff
@@ -129,6 +129,15 @@ impl TableContext {
                 self.name = new_table_name;
                 Ok(self)
             }
+            AlterTableOperation::ModifyDataType { column } => {
+                for iter in self.columns.iter_mut() {
+                    if iter.name != column.name {
+                        continue;
+                    }
+                    iter.column_type = column.column_type.clone();
+                }
```
WenyXu, line 138: How about using `position`

WenyXu:
We can't to close the #3973 now. We need to insert some values before the alteration and validate the inserted values after the alteration. However, it could be done in another PR.

tests-fuzz/src/ir.rs
realtaobo: It should not be supported at present.
https://github.com/GreptimeTeam/greptimedb/blob/4e5dd1ebb04d910bdb88611a8f824529701606b8/src/table/src/metadata.rs#L487-L494

WenyXu: I see

WenyXu (approved):

evenyag (approved):

PR #4080, opened 2024-05-31
refactor!: remove the tableid in ddl response since tableids is enough

fengjiachun:

As the title said, we do not need `table id` any more, `table ids` is enough, the PR of proto is: https://github.com/GreptimeTeam/greptime-proto/pull/175

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -120,7 +120,7 @@ etcd-client = { git = "https://github.com/MichaelScofield/etcd-client.git", rev 
 fst = "0.4.7"
 futures = "0.3"
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "6e0f5e684875207ba460e08a5d21ab7f775bba9e" }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -120,7 +120,7 @@ etcd-client = { git = "https://github.com/MichaelScofield/etcd-client.git", rev
 fst = "0.4.7"
 futures = "0.3"
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/killme2008/greptime-proto.git", rev = "a15a54a714fe117d7e9f7635e149c4eecac773fa" }
+greptime-proto = { git = "https://github.com/killme2008/greptime-proto.git", rev = "6e0f5e684875207ba460e08a5d21ab7f775bba9e" }
```
WenyXu, line 123: Should be https://github.com/GreptimeTeam/greptime-proto?

WenyXu (approved):

Cargo.toml
fengjiachun: Good catch!

evenyag (approved):

PR #4084, opened 2024-05-31
fix(flow): mfp operator missing rows

discord9:

fix a bug that mfp operator will missing rows if run multiple tims in same tick

- fix a bug that mfp operator will missing rows if run multiple tims in same tick
- add related tests

## Checklist

### DIFF
--- a/src/flow/src/compute/render/map.rs
+++ b/src/flow/src/compute/render/map.rs
@@ -125,9 +125,9 @@ fn mfp_subgraph(
                 false
             }
         });
-        let future_udpates = all_updates;
+        let future_updates = all_updates;
 
-        arrange.write().apply_updates(now, future_udpates)?;
+        arrange.write().apply_updates(now, future_updates)?;
         Ok(())
     };
     err_collector.run(run_mfp);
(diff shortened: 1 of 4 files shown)


killme2008:
Can we test it?

discord9:
> Can we test it?

Unit test are added, but since this bug is not quite yet suitable to be tested using sqlness/tests-intergation crate because it require above certain qps to be triggered, is considering modify tsbs or add a extra script/repo to test it

src/flow/src/compute/render/map.rs
```diff
@@ -115,7 +115,13 @@ fn mfp_subgraph(
 ) {
     let run_mfp = || {
         let all_updates = eval_mfp_core(input, mfp_plan, now, err_collector);
-        arrange.write().apply_updates(now, all_updates)?;
+        // if run mfp in the same tick for multiple times, we need to set last_compaction_time to now-1
+        // so that current rows doesn't get compacted(into nothing) and get missed, that's because mfp's arrangement is future only
+        let mut arrange = arrange.write();
+        if arrange.last_compaction_time() >= Some(now - 1) {
```
zhongzc, line 121: So hacky... Why it's unable for mfp to run multiple times in the same tick? Can we choose a better time to do compaction?

discord9:
flow doesn't work real well with sqlness since it's kind of async-ish, and can't update to dest table fast enough before sqlness deem test are done, currently working on that, meanwhile I'm using a python script to locally test flow's functionality

zhongzc (approved):

WenyXu (approved):

PR #4088, opened 2024-06-02
refactor!: unify `FrontendOptions` and `DatanodeOptions` by using `GrpcOptions`

realtaobo:

As mentioned in the comment at https://github.com/GreptimeTeam/greptimedb/pull/3957#issuecomment-2126260152, try to unify `FrontendOptions` and `DatanodeOptions` by using `GrpcOptions`.

- move `GrpcOptions` to servers/src/grpc
- refactor `DatanodeOptions`
- wrapping arguments like `rpc_addr` by Option. e.g. rpc_addr: Option<String>, set them to None by default and print a warning if people set it(https://github.com/GreptimeTeam/greptimedb/pull/4088#discussion_r1627424905). The final result is as follows：
```bash
 *  Executing task: cargo test --package cmd --lib -- datanode::tests::test_deprecated_cli_options --exact --show-output 
...
running 1 test
2024-06-17T15:10:58.675193Z  INFO common_telemetry::logging: logs dir = /tmp/__unittest_logs
2024-06-17T15:10:58.682675Z  WARN cmd::datanode: Use the deprecated attribute `DatanodeOptions.rpc_addr`, please use `grpc.addr` instead.
2024-06-17T15:10:58.682740Z  WARN cmd::datanode: Use the deprecated attribute `DatanodeOptions.rpc_hostname`, please use `grpc.hostname` instead.
test datanode::tests::test_depreacted_cli_options ... ok
```

## Checklist


src/datanode/src/config.rs
```diff
@@ -218,13 +216,8 @@ pub struct DatanodeOptions {
     pub node_id: Option<u64>,
     pub require_lease_before_startup: bool,
     pub init_regions_in_background: bool,
-    pub rpc_addr: String,
```
zyy17, line 221: It's an incompatible change. I'm not sure it's an acceptable refactor. If not, we can use some `#serde[]` annotations to make it compatible, for example, use `#[serde(flatten)]` and `#[serde(rename = "name")]`.

@evenyag @waynexia

zyy17:
Most of the changes are look good to me, except:

1. It's an **breaking changes** and it's worth for it? Or we should make it compatible?
2. The PR failed the fuzz tests;

realtaobo:
> Most of the changes are look good to me, except:
> 
>     1. It's an **breaking changes** and it's worth for it? Or we should make it compatible?
> 
>     2. The PR failed the fuzz tests;

Yes, this change  seems affect cluster communication, so Distributed fuzz tests failed. I haven't found a solution yet. 😂

src/datanode/src/config.rs
```diff
@@ -218,13 +216,8 @@ pub struct DatanodeOptions {
...
     pub rpc_hostname: Option<String>,
```
evenyag, line 219: The rpc_hostname is not a part of RPC options. Should we move it to the `GrpcOptions`?

src/datanode/src/config.rs
evenyag: A breaking change may be better than `serde(flatten)`. But we must modify the PR title to mark it as breaking.

@shuiyisong @sunng87 What do you think?

src/datanode/src/config.rs
realtaobo: I think so, but only the datanode use it. I don't know if I should refactor it.

evenyag: It should be fine. We can document this in the code.

shuiyisong: It looks like `rpc_hostname` is only used to report to metasrv through heartbeat. What's the usage of this field, can we remove it? @fengjiachun @sunng87

sunng87: We cannot. It's a fatal feature because metasrv can only get the IP address from heartbeat. Also datanode has no idea of the hostname that it will use for intra-cluster communication without specified.

src/datanode/src/config.rs
sunng87: I agress. But instead of a hard breaking change, can we add a deprecation warning for some time to smooth the upgrade?

src/datanode/src/config.rs
fengjiachun: We can't remove `rpc_home`, but I think we can move it to grpcOptions.

src/datanode/src/config.rs
evenyag: If we want to deprecate it, one way is wrapping arguments like `rpc_addr` by `Option`. e.g. `rpc_addr: Option<String>`. We can use it to initialize the grpc option. We set them to None by default and print a warning if people set it.

evenyag:
Since many people may already set the old `rpc_addr`, we might still keep these options but mark them as deprecated.

realtaobo:
> Since many people may already set the old `rpc_addr`, we might still keep these options but notify users that they are deprecated.

Yes, I'm aware of this, but still exist some other questions for fuzz tests within distributed mode. I'll ping again when everything is ready.

evenyag:
> > Since many people may already set the old `rpc_addr`, we might still keep these options but notify users that they are deprecated.
> 
> Yes, I'm aware of this, but still exist some other questions for fuzz tests within distributed mode. I'll ping again when everything is ready.

I guess some of the fuzz tests use the old rpc options.

realtaobo:
Hi, all, please take a look when you have some free time.
>1. It's an **breaking changes** and it's worth for it? Or we should make it compatible?

This PR moves `GrpcOptions` to `servers/src/grpc` and refactors the gRPC-related configurations of `DatanodeOptions`. Additionally, I have tagged `rpc_addr`, `rpc_runtime_size`, `rpc_hostname`, and `rpc_max_recv/send_message_size` as deprecated fields, setting them to None by default and printing a warning if they are set (**The execution result example can be found in the description**). Moreover, if the old RPC options are set, we read them first for compatibility.

>2. The PR failed the fuzz tests;

It has been resolved. I guess the distributed fuzz tests failed previously because they were not compatible with the greptimedb cluster version used by CI.

src/servers/src/grpc.rs
```diff
@@ -50,6 +52,39 @@ use crate::tls::TlsOption;
 
 type TonicResult<T> = std::result::Result<T, Status>;
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct GrpcOptions {
+    pub addr: String,
+    pub hostname: String,
+    // Max gRPC receiving(decoding) message size
+    pub max_recv_message_size: ReadableSize,
+    // Max gRPC sending(encoding) message size
```
evenyag, line 61:
```suggestion
    /// Max gRPC receiving(decoding) message size
    pub max_recv_message_size: ReadableSize,
    /// Max gRPC sending(encoding) message size
```

evenyag (approved):
Looks good.

killme2008 (approved):
Great work! Thanks a lot.

PR #4092, opened 2024-06-03
feat: support gRPC cancellation

### LINKED ISSUE #3964: Pass cancellation from client to datanode

Other

When a client connection is cancelled, either passively by frontend or actively by the client, we should pass this cancellation to cancel all on-going tasks. Both frontend and datanode should handle this cancellation.

### Implementation challenges

_No response_

fengjiachun:

This PR enables gRPC server cancellation support in `RegionServer` and `GreptimeService`.

Closes #3964 

## Checklist

### DIFF
--- a/src/servers/src/grpc/cancellation.rs
+++ b/src/servers/src/grpc/cancellation.rs
@@ -42,3 +42,49 @@ where
 
     select_task.await.unwrap()
 }
+
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use tokio::sync::mpsc;
+    use tokio::time;
+    use tonic::Response;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_request_completes_first() {
+        let request = async { Ok(Response::new("Request Completed")) };
+
+        let cancellation = async {
+            time::sleep(Duration::from_secs(1)).await;
+            Ok(Response::new("Cancelled"))
+        };
+
+        let result = with_cancellation_handler(request, cancellation).await;
+        assert_eq!(result.unwrap().into_inner(), "Request Completed");
+    }
+
+    #[tokio::test]
+    async fn test_cancellation_when_dropped() {
+        let (tx, mut rx) = mpsc::channel(2);
+        let tx_cloned = tx.clone();
+        let request = async move {
+            time::sleep(Duration::from_secs(1)).await;
+            tx_cloned.send("Request Completed").await.unwrap();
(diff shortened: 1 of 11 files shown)


src/servers/src/grpc/greptime_handler.rs
```diff
@@ -98,12 +91,28 @@ impl GreptimeRequestHandler {
...
+
+        match &self.runtime {
+            Some(runtime) => {
+                // Executes requests in another runtime to
+                // 1. prevent the execution from being cancelled unexpected by Tonic runtime;
+                //   - Refer to our blog for the rational behind it:
+                //     https://www.greptime.com/blogs/2023-01-12-hidden-control-flow.html
+                //   - Obtaining a `JoinHandle` to get the panic message (if there's any).
+                //     From its docs, `JoinHandle` is cancel safe. The task keeps running even it's handle been dropped.
+                // 2. avoid the handler blocks the gRPC runtime incidentally.
```
waynexia, line 104: cancellation handler is still an "external cancel", and may violate (1) which brings unexpected (from the task's perspective) cancel

src/servers/src/grpc/cancellation.rs
```diff
@@ -0,0 +1,90 @@
...
+    let token = CancellationToken::new();
+    // Will call token.cancel() when the future is dropped, such as when the client cancels the request
+    let _drop_guard = token.clone().drop_guard();
+    let select_task = tokio::spawn(async move {
+        // Can select on token cancellation on any cancellable future while handling the request,
+        // allowing for custom cleanup code or monitoring
+        select! {
+            res = request => res,
+            _ = token.cancelled() => cancellation.await,
+        }
```
waynexia, line 40: `select` does cancel other branches, but its [docs](https://docs.rs/tokio/latest/tokio/macro.select.html#cancellation-safety) also says

>When using select! in a loop to receive messages from multiple sources, you should make sure that the receive call is cancellation safe to avoid losing messages.

fengjiachun: Do you mean the `request` is not cancellation safe?

waynexia (approved):
My 5th time forgetting about the relationship of `Datanode`, `RegionInvoker` and `RegionServer`

fengjiachun:
> My 5th time forgetting about the relationship of `Datanode`, `RegionInvoker` and `RegionServer`

Me too, cc @WenyXu

Issue #2662, opened 2023-10-27
S3 EntityTooSmall error in flush/compaction


evenyag:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Datanode

### What happened?

S3 returns `EntityTooSmall` while closing the writer in flush/compaction job.

### What operating system did you use?

Unrelated

### Relevant log output and stack trace

```bash
2023-10-26T00:20:40.133578Z ERROR mito2::compaction: Region 5738076307456(1336, 0) failed to flush, cancel all pending tasks err=0: Failed to write to buffer, at greptimedb/src/mito2/src/sst/stream_writer.rs:103:14
2023-10-26T00:20:40.133445Z ERROR mito2::worker::handle_compaction: Failed to compact region: 5738076307456(1336, 0) err=0: Failed to write to buffer, at greptimedb/src/mito2/src/sst/stream_writer.rs:103:14
2023-10-26T00:20:40.133288Z ERROR mito2::compaction::twcs: Failed to compact region, region id: 5738076307456(1336, 0) err=0: Failed to write to buffer, at greptimedb/src/mito2/src/sst/stream_writer.rs:103:14
2023-10-26T00:20:40.133225Z ERROR mito2::compaction::twcs: Failed to compact region: 5738076307456(1336, 0) err=0: Failed to write to buffer, at greptimedb/src/mito2/src/sst/stream_writer.rs:103:14
[...]

PR #4098, opened 2024-06-03
chore: revert opendal 0.46

WenyXu:

#2662

Revert opendal 0.46

After upgrading to opendal 0.46,  it incurs the `EntityTooSmall` error during closing the writer in the flush/compaction job.  
The opendal's `FuturesAsyncWriter` doesn't work well with our `BufferedWriter`🥲. Maybe we should use the `chunk` method provided by opendal instead of implementing a `BufferedWriter` .

i.e.
```rust
                store
                    .writer_with(&path)
                    .chunk(DEFAULT_WRITE_BUFFER_SIZE)
                    .concurrent(concurrency)
                    .await
```

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,7 +83,7 @@ dependencies = [
  "axum",
  "bytes",
  "cfg-if",
- "http 0.2.12",
+ "http",
  "indexmap 1.9.3",
  "schemars",
  "serde",
@@ -764,9 +764,9 @@ dependencies = [
  "bytes",
  "futures-util",
  "headers",
- "http 0.2.12",
- "http-body 0.4.6",
- "hyper 0.14.28",
+ "http",
+ "http-body",
+ "hyper",
  "itoa",
  "matchit",
  "memchr",
@@ -794,8 +794,8 @@ dependencies = [
  "async-trait",
  "bytes",
  "futures-util",
- "http 0.2.12",
- "http-body 0.4.6",
+ "http",
+ "http-body",
  "mime",
  "rustversion",
  "tower-layer",
@@ -1853,7 +1853,7 @@ dependencies = [
  "common-telemetry",
  "common-test-util",
  "common-version",
- "hyper 0.14.28",
+ "hyper",
  "reqwest",
  "serde",
  "tempfile",
@@ -1962,7 +1962,7 @@ dependencies = [
  "futures-util",
  "hex",
  "humantime-serde",
- "hyper 0.14.28",
+ "hyper",
  "itertools 0.10.5",
  "lazy_static",
  "moka",
@@ -3533,6 +3533,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a357d28ed41a50f9c765dbfe56cbc04a64e53e5fc58ba79fbc34c10ef3df831f"
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b45de904aa0b010bce2ab45264d0631681847fa7b6f2eaa7dab7619943bc4f59"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
 name = "endian-type"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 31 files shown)


tisonkun (commented):
Could you elaborate what is the issue introduced by OpenDAL 0.46?

tisonkun:
Generally we fix issues instead of reverting. Unless there is an emergency to release or it's the dependency's fault that the new version contains bugs.

A PR without a description to revert commits looks confusing.

fengjiachun:
We need to explain why we're doing this. @WenyXu

WenyXu:
> Generally we fix issues instead of reverting. Unless there is an emergency to release or it's the dependency's fault that the new version contains bugs.
> 
> A PR without a description to revert commits looks confusing.

Yes, I will try to remove the `BufferedWriter ` tomorrow. 🥲 cc @v0y4g3r

Xuanwo:
> Revert opendal 0.46
> 
> After upgrading to opendal 0.46, it incurs the `EntityTooSmall` error during closing the writer in the flush/compaction job. The opendal's `FuturesAsyncWriter` doesn't work well with our `BufferedWriter`🥲. Maybe we should use the `chunk` method provided by opendal instead of implementing a `BufferedWriter` .

Hi, seems better to adapt new API instead of reverting?

Xuanwo:
> The opendal's `FuturesAsyncWriter` doesn't work well with our `BufferedWriter`

Would you like to raise a bug for this? `BufferedWriter` should work just fine like before with FuturesAsyncWriter. What's buf size do you set for your `BufferedWriter`?

WenyXu:
> > > The opendal's `FuturesAsyncWriter` doesn't work well with our `BufferedWriter`
> > 
> > 
> > Would you like to raise a bug for this? `BufferedWriter` should work just fine like before with FuturesAsyncWriter. What's buf size do you set for your `BufferedWriter`?
> 
> I think the key problem here is that `FuturesAsyncWriter ` will send a small part of the data to the underlying writer directly instead of being controlled by `BufferedWriter.` I created a PR to fix it #4100

Oops, I reviewed the code; the chunk size will be set automatically via the service's capability attributes. There must be something else I don't know🥹

Xuanwo:
> Oops, I reviewed the code; the chunk size will be set automatically via the service's capability attributes. 

Only set while chunk has been set. Write will go directly if chunk is None.

WenyXu:
> > Oops, I reviewed the code; the chunk size will be set automatically via the service's capability attributes.
> 
> Only set while chunk has been set. Write will go directly if chunk is None.

Oh, I see. So, the key problem is that we didn't set the chunk size.🥲

Xuanwo:
> Oh, I see. So, the key problem is that we didn't set the chunk size.🥲

Yep, it's quite easy to be used in wrong. 

I'm considering change the behavior here: we can make the min_write_size as the default chunk size.

If users has chunk set, we will respect users input and pick up max(chunk, min_write_size). All requests will be the same chunk size.

If users don't have chunk set, we will use min_write_size as chunk size. But we won't really perform the chunk operation: all write that larger than min_write_size will be sent directly.

Would you like to start a pr for this?

killme2008:
Should we release version 0.8.2? 
This is why I mentioned in Slack regarding Pull Request #1244. Testing this fundamental library upgrade extensively in our production environment is crucial.

tisonkun:
> Testing this fundamental library upgrade extensively in our production environment is crucial.

Another topic is that we're currently mixing up bug fixes and feature development in patch releases as we are in the 0.x series.

Once we release 1.0, perhaps we should use multi branches and only backport bug fixes to patch releases. The feature releases (minor or major) would have a release train model so that developers know clear the timeline and when to include changes and when feature freeze and test.

cc @fengjiachun

killme2008:
> > Testing this fundamental library upgrade extensively in our production environment is crucial.
> 
> Another topic is that we're currently mixing up bug fixes and feature development in patch releases as we are in the 0.x series.
> 
> Once we release 1.0, perhaps we should use multi branches and only backport bug fixes to patch releases. The feature releases (minor or major) would have a release train model so that developers know clear the timeline and when to include changes and when feature freeze and test.
> 
> cc @fengjiachun

Agree, I think we must set up the release rules after 1.0

WenyXu:
> If users don't have chunk set, we will use min_write_size as chunk size. But we won't really perform the chunk operation: all write that larger than min_write_size will be sent directly.

It seems we can't reuse the `ChunkedWriter` directly.

Xuanwo:
> It seems we can't reuse the `ChunkedWriter` directly.

Sorry for mixing the context. I'm talking about sending PRs to opendal for changing the behavior.

WenyXu:
fixed #4100

PR #4100, opened 2024-06-03
fix: fix EntityTooSmall issue

WenyXu:

We rely on the `BufferedWriter` to avoid writing too small entities to S3 directly. However, the `into_futures_async_write` will break this feature(it introduces another buffer in the underlying writer). And consider simplifying the `BufferredWriter` in the future.

1. fix S3 returns `EntityTooSmall` error issues
2. Add a unit test for this case.

## Checklist


v0y4g3r (approved):
LGTM

v0y4g3r:
Maybe elaborate the reason why the entity will be 256KB if `chunk` parameter is not provided in the pr detail.

.github/workflows/develop.yml
```diff
@@ -528,6 +531,11 @@ jobs:
           GT_S3_ACCESS_KEY_ID: ${{ secrets.AWS_CI_TEST_ACCESS_KEY_ID }}
           GT_S3_ACCESS_KEY: ${{ secrets.AWS_CI_TEST_SECRET_ACCESS_KEY }}
           GT_S3_REGION: ${{ vars.AWS_CI_TEST_BUCKET_REGION }}
+          GT_MINIO_BUCKET: greptime
```
evenyag, line 534: Why not reuse the S3 test bucket?

src/mito2/src/sst/parquet/writer.rs
```diff
@@ -208,3 +208,79 @@ impl SourceStats {
...
+    use datatypes::arrow::array::{ArrayRef, Int64Array, RecordBatch};
+    use datatypes::arrow::datatypes::{DataType, Field, Schema};
+    use object_store::services::S3;
+    use object_store::ObjectStore;
+    use rand::{thread_rng, Rng};
+
+    use crate::sst::DEFAULT_WRITE_CONCURRENCY;
+
+    #[tokio::test]
+    async fn test_parquet_writer() {
```
evenyag, line 228: We should put this test in the common-datasource crate.

.github/workflows/develop.yml
```diff
@@ -518,6 +518,9 @@ jobs:
       - name: Setup kafka server
         working-directory: tests-integration/fixtures/kafka
         run: docker compose -f docker-compose-standalone.yml up -d --wait
+      - name: Setup minio
+        working-directory: tests-integration/fixtures/minio
+        run: docker compose -f docker-compose-standalone.yml up -d --wait
```
evenyag, line 523: - Is it a good idea to set up a minio in CI? Since we already have a S3 bucket for testing. When should we use minio?
- Should we modify our S3 tests to use minio?

.github/workflows/develop.yml
evenyag: We should discover the issue once the test for the main branch fails. But exposing a bug earlier is not a bad thing.

.github/workflows/develop.yml
WenyXu: > When should we use minio?
When we need to test common behaviors in unit tests.

>  Should we modify our S3 tests to use minio?
No, It's fine to continue using S3 in integration tests

evenyag (approved):

PR #4102, opened 2024-06-04
fix: recover memtable options when opening physical regions

v0y4g3r:

During creating physical regions for metric engine, index and memtable options are set manually:
https://github.com/GreptimeTeam/greptimedb/blob/df01ac05a1ec5f7c3c4630eca7b58e8f0b39f8fe/src/metric-engine/src/engine/create.rs#L482-L485

But when open physical regions, only index options are set:

https://github.com/GreptimeTeam/greptimedb/blob/d1838fb28de054c5f4c82ea07683f28acd4871db/src/metric-engine/src/engine/open.rs#L83

### Memory usage

image

## Checklist

### DIFF
--- a/src/metric-engine/src/engine/open.rs
+++ b/src/metric-engine/src/engine/open.rs
@@ -26,7 +26,7 @@ use store_api::region_request::{AffectedRows, RegionOpenRequest, RegionRequest};
 use store_api::storage::RegionId;
 
 use super::MetricEngineInner;
-use crate::engine::options::set_index_options_for_data_region;
+use crate::engine::options::{set_index_options_for_data_region, set_memtable_options_for_data_region};
 use crate::error::{OpenMitoRegionSnafu, Result};
 use crate::metrics::{LOGICAL_REGION_COUNT, PHYSICAL_REGION_COUNT};
 use crate::utils;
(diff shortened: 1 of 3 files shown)


waynexia (approved):
Core tech ⚡

WenyXu (approved):

src/metric-engine/src/engine/open.rs
```diff
@@ -81,6 +83,7 @@ impl MetricEngineInner {
 
         let mut data_region_options = request.options;
         set_index_options_for_data_region(&mut data_region_options);
+        set_memtable_options_for_data_region(&mut data_region_options);
```
evenyag, line 86: Maybe we should merge `set_index_options_for_data_region()` and `set_memtable_options_for_data_region()` into one function.

fengjiachun (approved):
LGTM

src/metric-engine/src/engine/open.rs
v0y4g3r: > Maybe we should merge `set_index_options_for_data_region()` and `set_memtable_options_for_data_region()` into one function.

Fixed in 69de0e869

PR #4105, opened 2024-06-04
chore: run fuzz tests with kafka remote wal

WenyXu:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

1. Run fuzz tests with Minio
2. Run fuzz tests with Remote WAL

## Checklist

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.

PR #4112, opened 2024-06-06
refactor: simplify parquet writer

v0y4g3r:

This PR simplifies SST writing process by removing the `BufferedWriter`.

The `BufferedWriter` came to solve the problem that arrow's paruqet writer requires `std::io::Write` while opendal only provides async S3 writer that implements `tokio::io::AsyncWrite`.  Now arrow provides `AsyncArrowWriter` so we can remove those structs.

## Performance

Writing an parquet file with size 1.2GB,
- before: 11143163us
- after: 10507815us

A slight improvement of 6%.

## Checklist

### DIFF
--- a/src/common/datasource/src/file_format/parquet.rs
+++ b/src/common/datasource/src/file_format/parquet.rs
@@ -20,26 +20,28 @@ use arrow_schema::{Schema, SchemaRef};
 use async_trait::async_trait;
 use datafusion::datasource::physical_plan::{FileMeta, ParquetFileReaderFactory};
 use datafusion::error::Result as DatafusionResult;
+use datafusion::parquet::arrow::{ArrowWriter, parquet_to_arrow_schema};
 use datafusion::parquet::arrow::async_reader::AsyncFileReader;
-use datafusion::parquet::arrow::{parquet_to_arrow_schema, ArrowWriter};
 use datafusion::parquet::errors::{ParquetError, Result as ParquetResult};
 use datafusion::parquet::file::metadata::ParquetMetaData;
 use datafusion::parquet::format::FileMetaData;
 use datafusion::physical_plan::metrics::ExecutionPlanMetricsSet;
 use datafusion::physical_plan::SendableRecordBatchStream;
+use futures::{StreamExt,};
 use futures::future::BoxFuture;
-use futures::StreamExt;
-use object_store::{FuturesAsyncReader, ObjectStore};
+use parquet::arrow::AsyncArrowWriter;
 use parquet::basic::{Compression, ZstdLevel};
 use parquet::file::properties::WriterProperties;
 use snafu::ResultExt;
 use tokio_util::compat::{Compat, FuturesAsyncReadCompatExt, FuturesAsyncWriteCompatExt};
 
+use object_store::{FuturesAsyncReader, ObjectStore};
+
 use crate::buffered_writer::{ArrowWriterCloser, DfRecordBatchEncoder, LazyBufferedWriter};
-use crate::error::{self, Result};
+use crate::DEFAULT_WRITE_BUFFER_SIZE;
+use crate::error::{self, Result, WriteObjectSnafu, WriteParquetSnafu};
 use crate::file_format::FileFormat;
 use crate::share_buffer::SharedBuffer;
-use crate::DEFAULT_WRITE_BUFFER_SIZE;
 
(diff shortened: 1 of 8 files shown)


src/mito2/src/sst/parquet/writer.rs
```diff
@@ -208,3 +224,55 @@ impl SourceStats {
...
+            size: size.clone(),
+        }
+    }
+}
+
+impl<W> AsyncWrite for SizeAwareWriter<W>
+where
+    W: AsyncWrite + Unpin,
+{
+    fn poll_write(
```
waynexia, line 248: A new manual poll impl 🤣 (not saying this is bad)

src/mito2/src/sst/parquet/writer.rs
```diff
@@ -14,52 +14,51 @@
...
+    writer_factory: F,
     /// Region metadata of the source and the target SST.
     metadata: RegionMetadataRef,
-    object_store: ObjectStore,
     indexer: Indexer,
+    bytes_written: Arc<Mutex<usize>>,
 }
 
-impl ParquetWriter {
+impl<W: AsyncWrite + Send + Unpin, Fut: Future<Output = Result<W>>, F: FnMut() -> Fut>
```
waynexia, line 51: style: prefer tailing `where` bound for long type params

src/mito2/src/sst/parquet/writer.rs
```diff
@@ -14,52 +14,51 @@
...
-    /// SST output file path.
-    file_path: String,
+pub struct ParquetWriter<W, F> {
+    writer: Option<AsyncArrowWriter<SizeAwareWriter<W>>>,
+    writer_factory: F,
     /// Region metadata of the source and the target SST.
     metadata: RegionMetadataRef,
-    object_store: ObjectStore,
     indexer: Indexer,
+    bytes_written: Arc<Mutex<usize>>,
```
waynexia, line 48: Would an `AtomicUsize` be more appropriate?

waynexia (commented):
Great refactor overall. One thing I concert is now the parquet writer accepts a closure as "factory". Is it possible to remove this indirection and pass something like a future directly?

src/common/datasource/src/file_format/parquet.rs
```diff
@@ -174,123 +174,47 @@ impl ArrowWriterCloser for ArrowWriter<SharedBuffer> {
...
-        path.to_string(),
-        store,
-        schema,
-        Some(write_props),
-        threshold,
-        concurrency,
-    )
-    .await?;
+    let inner_writer = store
+        .writer_with(path)
```
WenyXu, line 191: We must invoke the `chunk` method to avoid the `EntityTooSmall` error from S3. BTW, we can utilize the tests introduced in #4105, which ensure there are no `EntityTooSmall ` errors returned from the MinIO during the flush and compaction operations.

v0y4g3r: > We must invoke the `chunk` method to avoid the `EntityTooSmall` errors from S3. BTW, we can utilize the tests introduced in #4105, which ensure there are no `EntityTooSmall ` errors returned from the MinIO during the flush and compaction operations.

fixed in 1a2bd1020

WenyXu (approved):

src/mito2/src/access_layer.rs
```diff
@@ -147,9 +148,16 @@ impl AccessLayer {
             }
             .build();
             let mut writer = ParquetWriter::new(
-                file_path,
+                || async {
+                    self.object_store
+                        .writer_with(&file_path)
+                        .chunk(DEFAULT_WRITE_BUFFER_SIZE.as_bytes() as usize)
+                        .concurrent(DEFAULT_WRITE_CONCURRENCY)
+                        .await
```
evenyag, line 156: Is it possible to wrap a helper function to create the writer? We repeat this many times.

src/mito2/src/sst/parquet.rs
```diff
@@ -125,9 +126,15 @@ mod tests {
         };
 
         let mut writer = ParquetWriter::new(
-            file_path,
+            || async {
+                Ok(object_store
```
evenyag, line 130: We should use the same function to build the writer/or the async writer.

evenyag (approved):
LGTM

waynexia (approved):

Issue #3456, opened 2024-03-07
Writing manifest files outside of the write thread


evenyag:
### What type of enhancement is this?

Performance

### What does the enhancement do?

As mentioned in https://github.com/GreptimeTeam/greptimedb/issues/3447#issuecomment-1982284075, we write manifest files inside the region worker thread (the write thread) and is likely to block the thread for a long time if there are many flush jobs.

We should move this operation to the background thread if possible and this should reduce the block time of the worker.

### Implementation challenges

Once we write manifests outside of the worker thread, we must find a way to serialize the following operations:
- checks whether a region is active (not been closed, dropped)
- writes the manifest file
- gets current region version
- apply edits to the region version

### Steps
- [x] https://github.com/GreptimeTeam/greptimedb/pull/3689
- [x] Spawns a task to write the manifest https://github.com/GreptimeTeam/greptimedb/pull/3709

evenyag:
If we update the manifest in the background, we should ensure the background flush and compaction don't corrupt the manifest. For example, a flush job updates the manifest after a truncate action is written to the manifest. Then we might see data still exist after truncation.

Fortunately, flush and compaction don't break most operations:
- ALTER
  - We always flush all memtables before altering the flush job is already finished
  - Compaction doesn't interrupt existing alteration as they only modify existing files
- DROP
  - We write the drop marker so changing the manifest doesn't affect it
- TRUNCATE
  - The only operation we need to consider is truncating the region
  - We can detect whether a region is truncated and whether a `RegionEdit` is created before truncation. Then we can ignore these edits.
- SET READONLY
  - We can set and check the state of the region inside the manifest lock
  - We don't update the manifest if the region is read-only

evenyag:
There is a remaining issue: If a region is editing the manifest, then sending another `RegionEdit` to the region will get a `RegionState` error. A more user-friendly way would be queueing the request and execute it after current edit request is done.

PR #4114, opened 2024-06-06
fix: Update region `Version` in the worker loop

evenyag:

- https://github.com/GreptimeTeam/greptimedb/issues/3456

This PR sends a notification to the worker loop after the manifest manager stores the update action. The worker will apply the update to the region `Version`.

It ensures the version is updated in the worker loop. If we update the version in another thread, then we might not have a consistent view of the version in the worker loop.

## Checklist


v0y4g3r (approved):
LGTM, exquisite design!

WenyXu (approved):

PR #4119, opened 2024-06-07
fix: executes pending ddls if region memtable is empty while scheduling next flush

evenyag:

This PR fixes an issue: If a region schedules itself to flush again but its memtable is empty, then it will return success directly without clearing the `FlushStatus` and executing pending DDL requests.

So an alter request may be blocked for a long time in this situation:
1. the region triggers a flush
2. the region wants to flush again, it adds the flush task to the pending flush list
3. the region receives an alter request and it adds the requests to the pending DDL list
4. the region doesn't receive any write request
5. the flush job is finished and the worker schedules the same region to flush
6. the region checks the memtable and marks the pending flush request as success immediately because the memtable is empty, but lefts the alter request in the list

There is a debug assertion in the scheduler to detect a similar case. This PR also adds two tests to cover this case in different level.

## Checklist

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -1,26 +1,26 @@
 on:
   merge_group:
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [ opened, synchronize, reopened, ready_for_review ]
     paths-ignore:
-      - "docs/**"
-      - "config/**"
-      - "**.md"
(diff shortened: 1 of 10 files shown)


killme2008 (approved):
LGTM

sunng87:
Fixes #4120

sunng87 (approved):

PR #4121, opened 2024-06-08
feat(pipeline): gsub prosessor

yuanbohan:

implement [Elastic Gsub processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/gsub-processor.html) 

## Checklist


shuiyisong:
Wait for #4014 get merged first, then we can add some test.

shuiyisong (approved):

src/pipeline/src/etl/processor/gsub.rs
```diff
@@ -0,0 +1,229 @@
...
+    IGNORE_MISSING_NAME,
+};
+use crate::etl::value::{Array, Map, Value};
+
+pub(crate) const PROCESSOR_GSUB: &str = "gsub";
+
+const REPLACEMENT_NAME: &str = "replacement";
+const PATTERN_NAME: &str = "pattern";
+
+/// only support string value, and array string value
```
killme2008, line 29:
```suggestion
/// A processor to replace all matches of a pattern in string by a replacement, only support string value, and array string value
```

src/pipeline/src/etl/processor/gsub.rs
```diff
@@ -0,0 +1,229 @@
...
+        Ok(())
+    }
+
+    fn with_replacement(&mut self, replacement: impl Into<String>) {
+        self.replacement = Some(replacement.into());
+    }
+
+    fn check(self) -> Result<Self, String> {
+        if self.pattern.is_none() {
+            return Err("pattern is required".to_string());
```
killme2008, line 58: Maybe we should replace all these Errors with snafu. It can't be done with this PR, but we must refactor them in the future.

src/pipeline/src/etl/processor/gsub.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        if self.replacement.is_none() {
+            return Err("replacement is required".to_string());
+        }
+
+        Ok(self)
+    }
+
+    fn process_string_field(&self, val: &str, field: &Field) -> Result<Map, String> {
+        let replacement = self.replacement.as_ref().unwrap();
```
killme2008, line 69: Do we need to call `check` before `unwrap` the replacement? Or do we have an API contract here?

src/pipeline/src/etl/processor/gsub.rs
```diff
@@ -0,0 +1,229 @@
...
+    fn process_array_field(&self, arr: &Array, field: &Field) -> Result<Map, String> {
+        let key = match field.target_field {
+            Some(ref target_field) => target_field,
+            None => field.get_field(),
+        };
+
+        let re = self.pattern.as_ref().unwrap();
+        let replacement = self.replacement.as_ref().unwrap();
+
+        let mut result = Array::default();
```
killme2008, line 95: Not sure current implementation. But looks like the Array can be replaced with Arrow's Array, and use vector execution to process these transforms. It's good for performance.

killme2008 (approved):
Almost LGTM

src/pipeline/src/etl/processor/gsub.rs
shuiyisong: https://github.com/GreptimeTeam/greptimedb/blob/69f5eca96b70a844276b42fb388c9ba36f7953b2/src/pipeline/src/etl/processor/gsub.rs#L147

On constructing the processor, it will check the `replacement` is set in the end. The `Option` here is more for the `default`. But we can use something like `OnceCell` to replace the `Option` in later refactor.

src/pipeline/src/etl/processor/gsub.rs
shuiyisong: Currently pipeline use its own `Value` enum(including an `Array(Array)` to hold array of `Value`). We will change its value type in next major refactor to boost performance.

PR #4123, opened 2024-06-08
feat(pipeline): transform support on_failure

yuanbohan:

when transform failed, the `on_failure` will try to prevent the failure ot ETL. And the valid value for on_failure is: `default` or `ignore`.

- default: set to default value
- ignore: set to None (the same result of no such field to be set)

The following will set the `number` field to default (which is 0 in the yaml) if its value can not be coerced to int32

```yaml
trnsform:
  field: number
  type: int32
  default: 0
  on_failure: default
```

## Checklist


shuiyisong (approved):

src/pipeline/src/etl/transform/mod.rs
```diff
@@ -38,6 +39,34 @@ pub trait Transformer: std::fmt::Display + Sized + Send + Sync + 'static {
     fn transform(&self, val: crate::etl::value::Value) -> Result<Self::Output, String>;
 }
 
+#[derive(Debug, Clone, Default)]
```
killme2008, line 42: I'd like to add comments for this enum and its variants.

src/pipeline/src/etl/transform/mod.rs
```diff
@@ -38,6 +39,34 @@ pub trait Transformer: std::fmt::Display + Sized + Send + Sync + 'static {
...
+    #[default]
+    Ignore,
+    Default,
+}
+
+impl std::str::FromStr for OnFailure {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
```
killme2008, line 53: to_lowercase?

killme2008 (approved):
LGTM. Good job!

src/pipeline/src/etl/transform/mod.rs
shuiyisong: The config is yml-style. It will complain at compile(insertion) time if user gives a wrong key name. I think we are good for now.

PR #4132, opened 2024-06-11
fix: explicitly set config instead of using changable default in tests

### LINKED ISSUE #4126: Workflow run 'Release' failed
@GreptimeTeam/db-approver
New failure: https://github.com/GreptimeTeam/greptimedb/actions/runs/9440573664

MichaelScofield:

fix #4126 

fix the test failure in #4126 

## Checklist

### DIFF
--- a/src/cmd/tests/load_config_test.rs
+++ b/src/cmd/tests/load_config_test.rs
@@ -16,6 +16,7 @@ use std::time::Duration;
 
 use cmd::options::GreptimeOptions;
 use cmd::standalone::StandaloneOptions;
+use common_base::readable_size::ReadableSize;
 use common_config::Configurable;
 use common_runtime::global::RuntimeOptions;
 use common_telemetry::logging::LoggingOptions;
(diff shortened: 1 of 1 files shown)


tisonkun (approved):

tisonkun:
Does this mean that the default value is improper and we should revisit them later? Or just the one for this scenario need tunning.

src/cmd/tests/load_config_test.rs
```diff
@@ -70,6 +71,11 @@ fn test_load_datanode_example_config() {
                 num_workers: 8,
                 auto_flush_interval: Duration::from_secs(3600),
                 scan_parallelism: 0,
```
evenyag, line 73: Also set `max_background_jobs` explicitly? We might adjust it later.

evenyag (approved):
LGTM

evenyag:
> Does this mean that the default value is improper and we should revisit them later?

No. These default options are auto-tunable. They depend on the total memory and cores of the host machine running the test. We can't assert the value of the default option with a fixed value in the example config file.

PR #4133, opened 2024-06-12
refactor: make region manifest checkpoint ran in background

MichaelScofield:

Region manifest checkpoint currently requires many object store reads and writes. The checkpoint is performed synchronously in the region manifest update. When the object store is slow, it could easily cause the region manifest update method executes too long. This PR makes the checkpoint ran in the background (performed asynchronously) to solve the problem.

This PR also makes some changes to how the checkpoint is done. Originally the current checkpoint is the last checkpoint merged with all the incremental files. This process has to scan the object store. However, since the checkpoint is actually a snapshot of the current region manifest, which is already present in the memory, stick to the old ways seems unnecessary. So this PR directly use the region manifest snapshot to make current checkpoint, and delete the obsolete incremental files.

## Checklist

### DIFF
--- a/src/mito2/src/manifest/checkpointer.rs
+++ b/src/mito2/src/manifest/checkpointer.rs
@@ -28,31 +28,81 @@ use crate::metrics::MANIFEST_OP_ELAPSED;
 /// [`Checkpointer`] is responsible for doing checkpoint for a region, in an asynchronous way.
 #[derive(Debug)]
 pub(crate) struct Checkpointer {
-    region_id: RegionId,
     manifest_options: RegionManifestOptions,
-    manifest_store: Arc<ManifestObjectStore>,
-    last_checkpoint_version: Arc<AtomicU64>,
     is_doing_checkpoint: Arc<AtomicBool>,
+    inner: Arc<Inner>,
+}
+
+#[derive(Debug)]
+struct Inner {
+    region_id: RegionId,
+    manifest_store: ManifestObjectStore,
+    last_checkpoint_version: AtomicU64,
+}
+
+impl Inner {
+    async fn do_checkpoint(&self, checkpoint: RegionCheckpoint) {
+        let _t = MANIFEST_OP_ELAPSED
+            .with_label_values(&["checkpoint"])
+            .start_timer();
+
+        let region_id = self.region_id();
+        let version = checkpoint.last_version();
+        let checkpoint = match checkpoint.encode() {
+            Ok(checkpoint) => checkpoint,
+            Err(e) => {
+                error!(e; "Failed to encode checkpoint {:?}", checkpoint);
+                return;
+            }
+        };
+        if let Err(e) = self
+            .manifest_store
+            .save_checkpoint(version, &checkpoint)
+            .await
+        {
+            error!(e; "Failed to save checkpoint for region {}", region_id);
+            return;
+        }
+
+        if let Err(e) = self.manifest_store.delete_until(version, true).await {
+            error!(e; "Failed to delete manifest actions until version {} for region {}", version, region_id);
+            return;
+        }
+
+        self.last_checkpoint_version
+            .store(version, Ordering::Relaxed);
+
+        info!(
+            "Checkpoint for region {} success, version: {}",
+            region_id, version
+        );
+    }
+
+    fn region_id(&self) -> RegionId {
+        self.region_id
+    }
 }
 
 impl Checkpointer {
     pub(crate) fn new(
         region_id: RegionId,
         manifest_options: RegionManifestOptions,
-        manifest_store: Arc<ManifestObjectStore>,
+        manifest_store: ManifestObjectStore,
         last_checkpoint_version: ManifestVersion,
     ) -> Self {
         Self {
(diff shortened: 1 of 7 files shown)


src/mito2/src/manifest/manager.rs
```diff
@@ -670,6 +593,6 @@ mod test {
 
         // get manifest size again
         let manifest_size = manager.manifest_usage();
-        assert_eq!(manifest_size, 1312);
+        assert_eq!(manifest_size, 1173);
```
MichaelScofield, line 596: This test changes because the old checkpoint way is one less incremental file than the current region manifest. Originally, the incremental files are scanned by `self.store.scan(start_version, end_version).await`, left close right open range. And the `env_version` is `self.last_version`, which is already in current manifest. 

Same to the other tests below.

src/mito2/src/manifest/checkpointer.rs
```diff
@@ -0,0 +1,150 @@
...
+        };
+        self.do_checkpoint(checkpoint);
+    }
+
+    fn do_checkpoint(&self, checkpoint: RegionCheckpoint) {
+        let is_doing_checkpoint = self.is_doing_checkpoint.clone();
+        is_doing_checkpoint.store(true, Ordering::Relaxed);
+        let guard = scopeguard::guard(is_doing_checkpoint, |x| {
+            x.store(false, Ordering::Relaxed);
+        });
```
evenyag, line 105: The guard is quite easy to implement. Do we need to introduce a new crate dependency for this? cc @tisonkun

MichaelScofield: We already `scopeguard` in our deps, search it in the cargo.lock and you'll see. Besides, `scopeguard` is quite popular, it's top 100 crates, no harm to introduce it.

src/mito2/src/manifest/checkpointer.rs
```diff
@@ -0,0 +1,150 @@
...
+use crate::metrics::MANIFEST_OP_ELAPSED;
+
+/// [`Checkpointer`] is responsible for doing checkpoint for a region, in an asynchronous way.
+#[derive(Debug)]
+pub(crate) struct Checkpointer {
+    region_id: RegionId,
+    manifest_options: RegionManifestOptions,
+    manifest_store: Arc<ManifestObjectStore>,
+    last_checkpoint_version: Arc<AtomicU64>,
+    is_doing_checkpoint: Arc<AtomicBool>,
```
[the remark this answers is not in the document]
MichaelScofield: Sorry, I'm still fond of `is_doing_checkpoint`, for its clearness.

MichaelScofield:
@evenyag @WenyXu @killme2008 PTAL

src/mito2/src/manifest/checkpointer.rs
```diff
@@ -0,0 +1,171 @@
...
+use crate::manifest::action::{RegionCheckpoint, RegionManifest};
+use crate::manifest::manager::RegionManifestOptions;
+use crate::manifest::storage::ManifestObjectStore;
+use crate::metrics::MANIFEST_OP_ELAPSED;
+
+/// [`Checkpointer`] is responsible for doing checkpoint for a region, in an asynchronous way.
+#[derive(Debug)]
+pub(crate) struct Checkpointer {
+    manifest_options: RegionManifestOptions,
+    is_doing_checkpoint: Arc<AtomicBool>,
```
evenyag, line 32: We should put `is_doing_checkpoint` into the inner struct which avoids allocating an `Arc` for it.

src/mito2/src/manifest/checkpointer.rs
```diff
@@ -0,0 +1,171 @@
...
+        is_doing_checkpoint.store(true, Ordering::Relaxed);
+        let guard = scopeguard::guard(is_doing_checkpoint, |x| {
+            x.store(false, Ordering::Relaxed);
+        });
+
+        let inner = self.inner.clone();
+        common_runtime::spawn_bg(async move {
+            let _guard = guard;
+
+            inner.do_checkpoint(checkpoint).await;
```
evenyag, line 163: We move the scope guard into the async block eventually, so there is no need to put it in an individual `Arc`.
```suggestion
        let inner = scopeguard::guard(self.inner.clone(), |x| {
            x.is_doing_checkpoint.store(false, Ordering::Relaxed);
        });

        common_runtime::spawn_bg(async move {
            inner.is_doing_checkpoint.store(true, Ordering::Relaxed);
            inner.do_checkpoint(checkpoint).await;
```

killme2008 (approved):
LGTM. It will help a lot when writing.

src/mito2/src/manifest/checkpointer.rs
MichaelScofield: If setting the `is_doing_checkpoint` inside the async task, we can't be sure when it's been set. The next `maybe_do_checkpoint` could be called only to find it can do checkpoint again, which is not good.

MichaelScofield:
@evenyag PTAL

evenyag (approved):

PR #4146, opened 2024-06-14
refactor: make FileRange a trait

v0y4g3r:

This PR changes the `FileRange` from a struct to a trait so that we can adapt in-memory parquet bytes to `SeqScan` or `UnorderedScan` and reuse the merge logic.

## Checklist


src/mito2/src/read/unordered_scan.rs
```diff
@@ -253,14 +253,14 @@ async fn maybe_init_parts(
 /// is no output ordering guarantee of each partition.
 #[derive(Default)]
 struct UnorderedDistributor {
-    file_ranges: Vec<FileRange>,
+    file_ranges: Vec<Box<dyn FileRange>>,
```
evenyag, line 256: Making the file range a trait may not be a good idea. We may create many trait objects because we will create a file range for each row group.

PR #4162, opened 2024-06-18
feat: Implement memtable range

evenyag:

This PR implements `MemtableRange` to represent a partition (range) of a memtable. This allows iterating the memtable concurrently. The memtable range keeps an `IterBuilder` to build an iterator to read that range.

The `Memtable` trait has a new method `ranges()` to return the list of ranges to read. Both `time_series` and `partition_tree` memtables return just 1 range.

The `ScanPart` now keeps a list of `MemtableRange` instead of `Memtable`.

It also refactors the `FileRange` and extracts a struct `RangeBase` so we can reuse it in other places. The `RangeBase` contains fields to prune batches from a file.

## Checklist

### DIFF
--- a/src/mito2/src/memtable.rs
+++ b/src/mito2/src/memtable.rs
@@ -297,18 +297,18 @@ pub trait IterBuilder: Send + Sync {
 pub type BoxedIterBuilder = Box<dyn IterBuilder>;
 
 /// Context shared by ranges of the same memtable.
-pub(crate) struct MemRangeContext {
+pub struct MemRangeContext {
     /// Id of the memtable.
     id: MemtableId,
     /// Iterator builder.
     builder: BoxedIterBuilder,
 }
 
-pub(crate) type MemRangeContextRef = Arc<MemRangeContext>;
+pub type MemRangeContextRef = Arc<MemRangeContext>;
 
 impl MemRangeContext {
     /// Creates a new [MemRangeContext].
-    pub(crate) fn new(id: MemtableId, builder: BoxedIterBuilder) -> Self {
+    pub fn new(id: MemtableId, builder: BoxedIterBuilder) -> Self {
         Self { id, builder }
     }
 }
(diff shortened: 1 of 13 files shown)


v0y4g3r:
Is it possible to completely deprecate `Memtable::iter` with `Memtable::range`? I think range is more generic and compatible with current `RegionScanner` design. In the long term, memtable should be considered as an in-memory file part just like those in disks.

src/mito2/src/memtable.rs
```diff
@@ -278,6 +285,57 @@ impl MemtableBuilderProvider {
...
+impl MemtableRangeContext {
+    /// Creates a new [MemtableRangeContext].
+    pub fn new(id: MemtableId, builder: BoxedIterBuilder) -> Self {
+        Self { id, builder }
+    }
+}
+
+/// A range in the memtable.
+#[derive(Clone)]
+pub struct MemtableRange {
```
v0y4g3r, line 316: Looks like `MemtableRange` is simply a wrapper for `BoxedIterBuilder`?

src/mito2/src/memtable.rs
```diff
@@ -110,6 +110,13 @@ pub trait Memtable: Send + Sync + fmt::Debug {
         predicate: Option<Predicate>,
     ) -> Result<BoxedBatchIterator>;
 
+    /// Returns the ranges in the memtable.
+    fn ranges(
+        &self,
+        projection: Option<&[ColumnId]>,
+        predicate: Option<Predicate>,
+    ) -> Vec<MemtableRange>;
```
v0y4g3r, line 118: Maybe return a result
```suggestion
    ) -> Result<Vec<MemtableRange>>;
```

src/mito2/src/memtable.rs
evenyag: Yes. I'm not sure what we need to put into the range. We may need a part index/id to locate a part in the memtable.

src/mito2/src/memtable.rs
evenyag: Should we postpone this until some memtable implementations need to return results?

v0y4g3r (approved):
LGTM

evenyag:
> Is it possible to completely deprecate `Memtable::iter` with `Memtable::range`?

It should be possible. Do we need to do it in this PR? Maybe we can deprecate `iter()` after `ranges()` is stable.

killme2008 (approved):
LGTM

PR #4167, opened 2024-06-18
chore: enhance add pipeline http api return data

paomian:

Making "add pipeline" http API return more info about added pipeline.

## Checklist

### DIFF
--- a/src/pipeline/src/manager/table.rs
+++ b/src/pipeline/src/manager/table.rs
@@ -57,6 +57,11 @@ pub type PipelineVersion = Option<TimestampNanosecond>;
 
 pub type PipelineTableRef = Arc<PipelineTable>;
 
+pub type PipelineRef = Arc<Pipeline<GreptimeTransformer>>;
+
+/// Pipeline info. A tuple of timestamp and pipeline reference.
+pub type PipelineInfo = (Timestamp, PipelineRef);
+
 pub const PIPELINE_TABLE_NAME: &str = "pipelines";
 
 pub const PIPELINE_TABLE_PIPELINE_NAME_COLUMN_NAME: &str = "name";
@@ -311,7 +316,7 @@ impl PipelineTable {
         name: &str,
         content_type: &str,
         pipeline: &str,
-    ) -> Result<Arc<Pipeline<GreptimeTransformer>>> {
+    ) -> Result<PipelineInfo> {
         let compiled_pipeline = Arc::new(Self::compile_pipeline(pipeline)?);
         // we will use the version in the future
         let version = self
@@ -329,7 +334,7 @@ impl PipelineTable {
             );
         }
 
-        Ok(compiled_pipeline)
+        Ok((version, compiled_pipeline))
     }
 
     async fn find_pipeline_by_name(
--- a/src/servers/src/http/event.rs
+++ b/src/servers/src/http/event.rs
@@ -25,7 +25,7 @@ use axum::{async_trait, BoxError, Extension, TypedHeader};
 use common_telemetry::{error, warn};
 use common_time::{Timestamp, Timezone};
 use datatypes::timestamp::TimestampNanosecond;
-use http::HeaderMap;
+use http::{HeaderMap, HeaderValue};
 use mime_guess::mime;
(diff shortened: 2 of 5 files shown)


killme2008 (approved):
LGTM

src/pipeline/src/manager/table.rs
```diff
@@ -57,6 +57,11 @@ pub type PipelineVersion = Option<TimestampNanosecond>;
 
 pub type PipelineTableRef = Arc<PipelineTable>;
 
+pub type PipelineRef = Arc<Pipeline<GreptimeTransformer>>;
+
+/// Pipeline info. A tuple of timestamp and pipeline reference.
+pub type PipelineInfo = (Timestamp, PipelineRef);
```
shuiyisong, line 63: `PipelineRef` is not actually used, Can we just return `PipelineVersion`?

src/servers/src/http/event.rs
```diff
@@ -123,10 +124,28 @@ pub async fn add_pipeline(
...
         .await;
 
-    result.map(|_| "ok".to_string()).map_err(|e| {
-        error!(e; "failed to insert pipeline");
-        e
-    })
+    result
+        .map(|pipeline| {
+            let mut headers = HeaderMap::new();
+            headers.append(CONTENT_TYPE, "application/json".parse().unwrap());
```
shuiyisong, line 130: nitpick: 
```rust
            // safe unwrap here
            let json_header =
                HeaderValue::from_str(mime_guess::mime::APPLICATION_JSON.as_ref()).unwrap();

            let mut headers = HeaderMap::new();
            headers.append(CONTENT_TYPE, json_header);
```
Avoid hard-coded header value and parse.

src/servers/src/http/event.rs
```diff
@@ -123,10 +124,28 @@ pub async fn add_pipeline(
...
+            // Safety check: unwrap is safe here because we have checked the format of the timestamp
+            let version = pipeline
+                .0
+                .as_formatted_string(
+                    "%Y-%m-%d %H:%M:%S%.fZ",
+                    // Safety check: unwrap is safe here because we have checked the format of the timezone
+                    Some(Timezone::from_tz_string("UTC").unwrap()).as_ref(),
```
shuiyisong, line 137: How about use UNIX_MILLS format(like '1718768595191') as version. It's kinda strange to me using a formatted date string(like '2024-06-19 03:43:15.191') a version number.

killme2008: I think it's like the `last-modified-since`.

shuiyisong: The string is later used as a version parameter to find target pipeline(`pipeline_name` + `version`, `latest` if `version` is `None`), rather then a display-only information.

https://github.com/GreptimeTeam/greptimedb/blob/01e3a24cf77d748e4e4cfdf4155d6e2e2a3aabd6/src/servers/src/http/event.rs#L187-L198

src/pipeline/src/manager/table.rs
paomian: A detailed description of the pipeline should be returned in the future.

src/servers/src/http/event.rs
paomian: I considered the unix timestamp format, but for the user I wanted him to make it simple for the user to understand the content. unix timestamps are pretty standard and simple, and I think both would work. But for user readability I chose the formatted time string.

shuiyisong: One issue here is that date string is formatted with timezone offset. For example if I were in +0800 and I got result string back in UTC, it would be confusing.

paomian: Do you mean to use the same formatter for parse and format?

paomian: The version is `2024-06-19 09:24:55.788204038Z` . 
The "Z" at the end means it's in UTC time. If I use timestamp I need to convert it to even know the exact time.

killme2008: Agree with string format, it's more readable.

shuiyisong: Ok, let's keep it.

shuiyisong (approved):

PR #4178, opened 2024-06-20
perf: optimize RecordBatch to HttpOutput conversion

waynexia:

Optimize `HttpRecordsOutput::try_new()` method. Reduce the consumption of converting 409600 rows from ~400ms to ~120ms. Saves ~70% CPU.

This patch also adds a microbenchmark for that method.

Key optimizations are:
- Change the algorithm to do column-row conversion
- Change the underlying type of `Value::String` from `Bytes` to `String`
- Avoid all unnecessary `Clone`s

Now the entire conversion method only has one `clone` that takes data from the data vector, and one write that `writes` the serialized content to the result vector.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1877,7 +1877,7 @@ dependencies = [
  "common-runtime",
  "common-telemetry",
  "common-time",
- "criterion",
+ "criterion 0.4.0",
  "dashmap",
  "datatypes",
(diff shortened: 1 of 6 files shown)


fengjiachun (approved):
LGTM

evenyag (approved):

PR #4181, opened 2024-06-20
feat: introduce the interface of `RemoteJobScheduler`

zyy17:

### `RemoteJobScheduler`

For the storage disaggregated system, we can always offload the CPU-intensive and IO-intensive tasks(for example, compaction and index) to the remote workers. For the above scenario, the PR introduces the abstraction.

`RemoteJobScheduler` is a trait that defines the API for scheduling remote jobs. Its implementation is in **GreptimeDB Enterprise**. 

```rust
/// RemoteJobScheduler is a trait that defines the API to schedule remote jobs.
#[async_trait::async_trait]
pub trait RemoteJobScheduler: Send + Sync + 'static {
    /// Sends a job to the scheduler and returns a unique identifier for the job.
    async fn schedule(&self, job: RemoteJob, notifier: Arc<dyn Notifier>) -> Result<JobId>;
}

/// Notifier is used to notify the mito engine when a remote job is completed.
#[async_trait::async_trait]
pub trait Notifier: Send + Sync + 'static {
    /// Notify the mito engine that a remote job is completed.
    async fn notify(&self, result: RemoteJobResult);
}
```

The PR modify `schedule_compaction_request()` to support remote compaction:

- If the compact request specifies the `remote_compaction` in `region_options` and the `RemoteJobScheduler` is initialized, the Mito will execute remote compaction;
- If the remote compaction fails, fall back to local compaction;

### Other changes

- Add the `async` keyword for all the compaction-related functions because the `schedule_compaction_request` needs to be `async`;

- Use `Option` type for `senders` in `CompactionFinished` because we don't need it in remote compaction scenario; 

- Add `remote_compaction` in compaction options;

### TODOs

- [x] Inject `RemoteJobScheduler` from the plugin system;
- [ ] ~~Design the API to fetch the Jobs from the scheduler. When the datanode restarts, it can rebuild the context of the remote job~~;
- [ ] ~~Add the unit tests for the `RemoteJobScheduler`~~;

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced support for remote compaction scheduling.
  - Added `Plugins` functionality to enhance compaction scheduling and handling.
  - Enabled asynchronous operations for scheduling and finishing compaction tasks.
  - Integrated remote job scheduling mechanics for compaction.

- **Enhancements**
  - Updated `MitoEngine` and `EngineInner` constructors to accept `plugins` parameters.
  - Improved error handling with new `RemoteJobScheduler` and other error variants.
  - Added checks for `remote_compaction` in compaction options.

- **Bug Fixes**
  - Improved test framework to handle new plugin arguments and asynchronous calls.


zyy17:
I make some stupid(`git reflog` save me again) git operations and suddenly can't reopen the original PR #4124. I had no choice but to create a new PR. I'm sorry for the confusion.

src/mito2/src/compaction.rs
```diff
@@ -299,20 +276,86 @@ impl CompactionScheduler {
...
             }
-            return None;
+            self.region_status.remove(&region_id);
+            return Ok(());
         };
 
-        let task = CompactionTaskImpl {
+        // If specified to run compaction remotely, we schedule the compaction job remotely.
+        // It will fall back to local compaction if there is no remote job scheduler.
+        if current_version.options.compaction.remote_compaction() {
```
sunng87, line 285: I think we can add a step to evaluate the compaction job to see if it's suitable to run locally or remotely. The default implementation can be noop. This allows us add more strategy in future to control the behaviour.

src/mito2/src/schedule/remote_job_scheduler.rs
```diff
@@ -0,0 +1,137 @@
...
+/// Notifier is used to notify the mito engine when a remote job is completed.
+#[async_trait::async_trait]
+pub trait Notifier: Send + Sync + 'static {
+    /// Notify the mito engine that a remote job is completed.
+    async fn notify(&self, result: RemoteJobResult);
+}
+
+/// JobId is a unique identifier for a remote job and allocated by the scheduler.
+#[derive(Default, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct JobId(u64);
```
sunng87, line 46: Can we add more context information so we can track the job better:

1. catalog/schema/table information
2. job sender (the datanode id)

sunng87:
Mostly LGTM. @fengjiachun @v0y4g3r @evenyag PTAL and add comments from your perspective

src/mito2/src/compaction.rs
zyy17: I agree with that. The current implementation decides whether to use remote compaction based on the region options specified with the table creation. I think we can put more strategy in the future.

src/mito2/src/schedule/remote_job_scheduler.rs
zyy17: sounds reasonable.

src/mito2/src/request.rs
```diff
@@ -693,7 +693,7 @@ pub(crate) struct CompactionFinished {
     /// Region id.
     pub(crate) region_id: RegionId,
     /// Compaction result senders.
-    pub(crate) senders: Vec<OutputTx>,
+    pub(crate) senders: Option<Vec<OutputTx>>,
```
evenyag, line 696: Why do we need to wrap the `Vec` by `Option`? We can use an empty vec if there is no sender.

src/mito2/src/compaction.rs
```diff
@@ -300,20 +277,86 @@ impl CompactionScheduler {
...
+                    info!(
+                        "Scheduled remote compaction job {} for region {}",
+                        job_id.as_u64(),
+                        region_id
+                    );
+
+                    // Return 0 to the client as we are going to schedule the compaction job remotely.
+                    for waiter in waiters {
+                        waiter.send(Ok(0));
+                    }
```
evenyag, line 313: When waiters are provided, we expect to send the compaction result after the compaction job is done. We should send the result after the remote job is finished.

src/mito2/src/compaction.rs
```diff
@@ -300,20 +277,86 @@ impl CompactionScheduler {
...
+
+                    // Finish the remote compaction job.
+                    return Ok(());
+                } else {
+                    error!("Failed to schedule remote compaction job for region {}, fallback to local compaction", region_id)
+                }
+            } else {
+                info!(
```
evenyag, line 321: Maybe change to debug log.

src/mito2/src/region/options.rs
```diff
@@ -163,6 +172,7 @@ impl Default for TwcsOptions {
             max_active_window_files: 4,
             max_inactive_window_files: 1,
             time_window: None,
+            remote_compaction: false,
```
evenyag, line 175: Also update the tests in this file to see whether we can deserialize the `remote_compaction` flag correctly.

src/mito2/src/schedule/remote_job_scheduler.rs
evenyag, line 46: Maybe we can use uuid.

src/mito2/src/schedule/remote_job_scheduler.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/// CompactionJobResult is the result of a compaction job.
+#[allow(dead_code)]
+pub struct CompactionJobResult {
+    pub job_id: JobId,
+    pub region_id: RegionId,
+    pub start_time: Instant,
+    pub region_edit: Option<RegionEdit>,
+    pub err: Option<Error>,
```
evenyag, line 89: Can we use `Result<RegionEdit>`? We don't expect both `region_edit` and `err` to be None.

src/mito2/src/request.rs
zyy17: oops, you're right. Let me change that.

src/mito2/src/compaction.rs
zyy17: I remove the above `waiter.send(Ok(0))`. Instead, I add the `waiters` field in `DefaultNotifier`. Since the `waiter` can't be clonable(beacause of `Oneshot`), I have to use `std::mem::take(&mut waiters)`.

src/mito2/src/schedule/remote_job_scheduler.rs
evenyag: > After some thought, I think it needs to keep the above metadata in mito if we want to add it in the Job

I prefer not to add them to mito directly.

`region_dir` should contain the catalog/schema name and table id. We also have region id in the compaction request which already tracks a region.

For other metadata like the datanode id, you could attach them to the RPC request in the implementation of the `RemoteJobScheduler`.

evenyag: Could we use the `Uuid` struct directly? `String` isn't copyable but the `Uuid` struct is.

src/mito2/src/schedule/remote_job_scheduler.rs
```diff
@@ -0,0 +1,136 @@
...
+/// ```
+#[async_trait::async_trait]
+pub trait RemoteJobScheduler: Send + Sync + 'static {
+    /// Sends a job to the scheduler and returns a UUID for the job.
+    async fn schedule(&self, job: RemoteJob, notifier: Arc<dyn Notifier>) -> Result<String>;
+}
+
+/// Notifier is used to notify the mito engine when a remote job is completed.
+#[async_trait::async_trait]
+pub trait Notifier: Send + Sync + 'static {
```
evenyag, line 58: Does the notifier need to be `Sync`?

src/mito2/src/schedule/remote_job_scheduler.rs
evenyag: The `ProcedureId` is an example
https://github.com/GreptimeTeam/greptimedb/blob/8702066967bd3a68896b55cdee6ccde3eeca8812/src/common/procedure/src/procedure.rs#L264-L266

src/mito2/src/schedule/remote_job_scheduler.rs
```diff
@@ -0,0 +1,136 @@
...
+///   CompactionScheduler->>Plugins: Handle plugins
+///   CompactionScheduler->>RemoteJobScheduler: schedule(CompactionJob)
+///   RemoteJobScheduler-->>CompactionScheduler: Returns Job UUID
+///   CompactionScheduler-->>MitoEngine: Task scheduled with Job UUID
+///   MitoEngine-->>User: Compaction task scheduled
+/// ```
+#[async_trait::async_trait]
+pub trait RemoteJobScheduler: Send + Sync + 'static {
+    /// Sends a job to the scheduler and returns a UUID for the job.
+    async fn schedule(&self, job: RemoteJob, notifier: Arc<dyn Notifier>) -> Result<String>;
```
evenyag, line 53: Do we need to use `Arc` instead of `Box` for the notifier? Since `notify()` requires `&mut self` but you are hard `&mut self` from `Arc`.

I'd recommend changing it to `Box<dyn Notifier>`.

src/mito2/src/compaction.rs
```diff
@@ -302,20 +279,81 @@ impl CompactionScheduler {
...
+
+                let result = remote_job_scheduler
+                    .schedule(
+                        RemoteJob::CompactionJob(remote_compaction_job),
+                        Arc::new(DefaultNotifier {
+                            request_sender: request_sender.clone(),
+                            waiters: std::mem::take(&mut waiters),
+                        }),
+                    )
+                    .await;
```
evenyag, line 304: If the remote job scheduler returns an error, we will lose all `waiters`.

src/mito2/src/compaction.rs
evenyag: Got it.

src/mito2/src/schedule/remote_job_scheduler.rs
zyy17: TIL, you're right.

src/mito2/src/compaction.rs
zyy17: If we follow the [original implementation](https://github.com/GreptimeTeam/greptimedb/blob/main/src/mito2/src/compaction/task.rs#L146-L173), maybe I need to add the following code:

```rust
impl DefaultNotifier {
    fn on_failure(&mut self, err: Arc<Error>, region_id: RegionId) {
        COMPACTION_FAILURE_COUNT.inc();
        for waiter in self.waiters.drain(..) {
            waiter.send(Err(err.clone()).context(CompactRegionSnafu {
                region_id,
            }));
        }
    }
}

#[async_trait::async_trait]
impl Notifier for DefaultNotifier {
    async fn notify(&mut self, result: RemoteJobResult) {
        match result {
            RemoteJobResult::CompactionJobResult(result) => {
                let notify = {
                    match result.region_edit {
                        Ok(edit) => BackgroundNotify::CompactionFinished(CompactionFinished {
                            region_id: result.region_id,
                            senders: std::mem::take(&mut self.waiters),
                            start_time: result.start_time,
                            edit,
                        }),
                        Err(err) => {
                            error!("Compaction failed for region {}: {:?}", result.region_id, err);
                            let err = Arc::new(err);
                            self.on_failure(err.clone(), result.region_id);
                            BackgroundNotify::CompactionFailed(CompactionFailed {
                                region_id: result.region_id,
                                err,
                            })
                        }
                    }
                };

                if let Err(e) = self
                    .request_sender
                    .send(WorkerRequest::Background {
                        region_id: result.region_id,
                        notify,
                    })
                    .await
                {
                    error!(
                        "Failed to notify compaction job status for region {}, error: {:?}",
                        result.region_id, e
                    );
                }
            }
        }
    }
}
```

Should we need to add the new type of error or we can put it in the new PR?

src/mito2/src/schedule/remote_job_scheduler.rs
zyy17: @evenyag you're right, I will implement it in the plugin.

src/mito2/src/compaction.rs
zyy17: The new PR has the following refactoring:

1. Add the `waiters` in `RemoteJobScheduler` error;
2. Move `waiters` from `DefaultNotifier` to `Job,` and it will be convenient to pass between multiple objects;

src/mito2/src/error.rs
```diff
@@ -754,6 +755,23 @@ pub enum Error {
...
     },
 
+    #[snafu(display("Internal error occurred in remote job scheduler: {}", reason))]
+    RemoteJobScheduler {
+        #[snafu(implicit)]
+        location: Location,
+        reason: String,
+        // Keep the waiters in the error so that we can notify them when fallback to the local compaction.
+        waiters: Vec<OutputTx>,
+    },
```
evenyag, line 765: Maybe we should define a dedicated error struct for this. This avoids the remote scheduler returns other error variants.
https://docs.rs/snafu/latest/snafu/index.html#struct-style

src/mito2/src/compaction.rs
```diff
@@ -302,20 +279,88 @@ impl CompactionScheduler {
...
+                        Box::new(DefaultNotifier {
+                            request_sender: request_sender.clone(),
+                        }),
+                    )
+                    .await;
+
+                match result {
+                    Ok(job_id) => {
+                        info!(
+                            "Scheduled remote compaction job {:?} for region {}",
```
evenyag, line 309: If we implement display for the job id, then we can use the display trait to print it
```suggestion
                            "Scheduled remote compaction job {} for region {}",
```

evenyag:
@coderabbitai pause

src/mito2/src/error.rs
zyy17: I add the error `RemoteJobSchedulerError` that used for remote job scheduler.

evenyag (approved):

sunng87 (approved):

fengjiachun (approved):
LGTM

PR #4197, opened 2024-06-24
feat: support inserting into binary value through string

CookiePieWw:

Introduce a similar behavior from MySQL to GreptimeDB:
```SQL
# In MySQL
mysql> create table test(a int, b blob);
mysql> insert into test values(1, "123");
mysql> select * from test;
+------+------------+
| a    | b          |
+------+------------+
|    1 | 0x313233   |
+------+------------+
1 row in set (0.00 sec)
```

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced support for a new `Binary` data type.

- **Bug Fixes**
  - Enhanced data type handling by properly accounting for binary values in string conversions.

- **Tests**
  - Added comprehensive test cases to ensure the correct handling and stringification of the new `Binary` data type.

### DIFF
--- a/src/datatypes/src/data_type.rs
+++ b/src/datatypes/src/data_type.rs
@@ -161,6 +161,7 @@ impl ConcreteDataType {
                 | ConcreteDataType::Interval(_)
                 | ConcreteDataType::Duration(_)
                 | ConcreteDataType::Decimal128(_)
+                | ConcreteDataType::Binary(_)
         )
     }
 
@@ -717,6 +718,7 @@ mod tests {
         assert!(!ConcreteDataType::int32_datatype().is_stringifiable());
         assert!(!ConcreteDataType::float32_datatype().is_stringifiable());
         assert!(ConcreteDataType::string_datatype().is_stringifiable());
+        assert!(ConcreteDataType::binary_datatype().is_stringifiable());
         assert!(ConcreteDataType::date_datatype().is_stringifiable());
         assert!(ConcreteDataType::datetime_datatype().is_stringifiable());
         assert!(ConcreteDataType::timestamp_second_datatype().is_stringifiable());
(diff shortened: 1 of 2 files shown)


CookiePieWw:
I happened to this while working for the tests of `PREPARE` stmts, there's a TODO for this, but a little different? I think it means something like `b'abcd'` instead of plain string `'abcd'`.
https://github.com/GreptimeTeam/greptimedb/blob/ea7c17089f2ded9ec0db215b42856a700645c287/src/sql/src/statements.rs#L251

src/datatypes/src/data_type.rs
```diff
@@ -161,6 +161,7 @@ impl ConcreteDataType {
                 | ConcreteDataType::Interval(_)
                 | ConcreteDataType::Duration(_)
                 | ConcreteDataType::Decimal128(_)
+                | ConcreteDataType::Binary(_)
```
WenyXu, line 164: What if the data in binary isn't encoded in UTF-8?

CookiePieWw: Ah sry, I didn't think about this. We can convert strings into binary values, but not the same in turn.

Here `parse_string_to_value` checks the conversion from other types into strings for converting strings into other types,  maybe we need to change this?
https://github.com/GreptimeTeam/greptimedb/blob/77904adaaf9c970fc7487f62480e7fa737090baf/src/sql/src/statements.rs#L61-L68

Also, I noticed that MySQL displays binary values like `0x313233` without caring about its encoding. I wonder if its only about displaying or we could take it as an option for the conversion between strings and blobs.

evenyag: `is_stringifiable()` is different from `is_string()` so this change should be safe. Maybe we should document that the method is used to indicate that the type could be parsed from a string literal.

evenyag (approved):

WenyXu (approved):

Issue #4186, opened 2024-06-21
Incorrect peer_addr in cluster_info meta table


sunng87:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Distributed Cluster

### Minimal reproduce step

select * from information_schema.cluster_info

### What did you expect to see?

This `peer_addr` field has different implementation for different peer types.

- datanode is using `--rpc-hostname` parameter, which uses hostname of the node. This is preferred for displaying host address.
- metasrv seems to use its grpc connection's local address here, which need be corrected.
- frontend seems use its grpc servers' listening address here, which need be corrected.

Also we should not show grpc port here because on these three node types, grpc serves different purpose and its meaningless to list them together.

### What did you see instead?

```
+---------+-----------+-------------------------------------------------------------------+---------+------------+----------------------------+-------------------+-------------+
| peer_id | peer_type | peer_addr                                                         | version | git_commit | start_time                 | uptime            | active_time |
+---------+-----------+-------------------------------------------------------------------+---------+------------+----------------------------+-------------------+-------------+
|      -1 | METASRV   | 10.244.0.25:3002                                                  | 0.8.1   | unknown    | 2024-06-20 13:41:04.339000 | 12h 22m 10s 857ms | NULL        |
|       0 | DATANODE  | greptimedb-datanode-0.greptimedb-datanode.greptimedb-cluster:4001 | 0.8.1   | unknown    | 2024-06-20 13:43:25.491000 | 12h 19m 49s 705ms | 2s 598ms    |
|       1 | DATANODE  | greptimedb-datanode-1.greptimedb-datanode.greptimedb-cluster:4001 | 0.8.1   | unknown    | 2024-06-20 13:42:52.340000 | 12h 20m 22s 856ms | 2s 764ms    |
|       2 | DATANODE  | greptimedb-datanode-2.greptimedb-datanode.greptimedb-cluster:4001 | 0.8.1   | unknown    | 2024-06-20 13:42:19.742000 | 12h 20m 55s 454ms | 2s 83ms     |
|      -1 | FRONTEND  | 0.0.0.0:4001                                                      | 0.8.1   | unknown    | 2024-06-20 13:43:27.032000 | 12h 19m 48s 164ms | 14s 947ms   |
+---------+-----------+-------------------------------------------------------------------+---------+------------+----------------------------+-------------------+-------------+
5 rows in set (0.006 sec)
```

### What operating system did you use?

any

[...]

PR #4199, opened 2024-06-24
fix: wrong frontend registration address

killme2008:

#4186 

Fixed the wrong peer address that frontend registers.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a new configuration parameter `grpc.hostname` to improve clarity and understanding of its usage for external connections.
  
- **Improvements**
  - Updated configuration files to provide clarification on the `hostname` field for gRPC servers, enhancing documentation for end-users.

- **Refactor**
  - Renamed `server_addr` to `peer_addr` in heartbeat components for consistency.
  - Improved address resolution logic for better handling of hostname and port.

### DIFF
--- a/src/frontend/src/heartbeat.rs
+++ b/src/frontend/src/heartbeat.rs
@@ -22,6 +22,7 @@ use common_meta::heartbeat::mailbox::{HeartbeatMailbox, MailboxRef, OutgoingMess
 use common_meta::heartbeat::utils::outgoing_message_to_mailbox_message;
 use common_telemetry::{debug, error, info};
 use meta_client::client::{HeartbeatSender, HeartbeatStream, MetaClient};
+use servers::addrs;
 use servers::heartbeat_options::HeartbeatOptions;
 use snafu::ResultExt;
 use tokio::sync::mpsc;
@@ -37,7 +38,7 @@ pub mod handler;
 /// The frontend heartbeat task which sending `[HeartbeatRequest]` to Metasrv periodically in background.
 #[derive(Clone)]
 pub struct HeartbeatTask {
-    server_addr: String,
+    peer_addr: String,
     meta_client: Arc<MetaClient>,
     report_interval: u64,
     retry_interval: u64,
@@ -53,7 +54,7 @@ impl HeartbeatTask {
         resp_handler_executor: HeartbeatResponseHandlerExecutorRef,
     ) -> Self {
         HeartbeatTask {
-            server_addr: opts.grpc.addr.clone(),
+            peer_addr: addrs::resolve_addr(&opts.grpc.addr, Some(&opts.grpc.hostname)),
             meta_client,
             report_interval: heartbeat_opts.interval.as_millis() as u64,
             retry_interval: heartbeat_opts.retry_interval.as_millis() as u64,
@@ -129,7 +130,7 @@ impl HeartbeatTask {
         let self_peer = Some(Peer {
             // The peer id doesn't make sense for frontend, so we just set it 0.
             id: 0,
-            addr: self.server_addr.clone(),
+            addr: self.peer_addr.clone(),
         });
 
         common_runtime::spawn_hb(async move {
--- a/src/servers/src/addrs.rs
+++ b/src/servers/src/addrs.rs
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use common_telemetry::warn;
+
 /// Resolves hostname:port address for meta registration.
 /// If `hostname_addr` is present, prefer to use it, `bind_addr` otherwise.
-///
 pub fn resolve_addr(bind_addr: &str, hostname_addr: Option<&str>) -> String {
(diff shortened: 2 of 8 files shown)


sunng87 (commented):
We should remove the port part when displaying `peer_addr` in cluster_info table, because those port serves different purpose, it's meaningless to list them in the same context.

killme2008:
> We should remove the port part when displaying `peer_addr` in cluster_info table, because those port serves different purpose, it's meaningless to list them in the same context.

As I said in https://github.com/GreptimeTeam/greptimedb/issues/4186#issuecomment-2187329797

I think using gRPC is good currently, and the gRPC is the main communication between nodes, it's part of cluster-info.

src/frontend/src/heartbeat.rs
```diff
@@ -53,7 +54,7 @@ impl HeartbeatTask {
         resp_handler_executor: HeartbeatResponseHandlerExecutorRef,
     ) -> Self {
         HeartbeatTask {
-            server_addr: opts.grpc.addr.clone(),
+            peer_addr: addrs::resolve_addr(&opts.grpc.addr, Some(&opts.grpc.hostname)),
```
fengjiachun, line 57: In the example configuration file of datanode, there is a hostname https://github.com/GreptimeTeam/greptimedb/blob/22d12683b47e36aa349ca6d6ffe619444763f669/config/datanode.example.toml#L47
But it's not included in the frontend's configuration file. Can we add it?

fengjiachun:
Flow's heartbeat task also need to do the same thing: https://github.com/GreptimeTeam/greptimedb/blob/a44fe627ce6a16b703756e4ffb7e48559062b079/src/flow/src/heartbeat.rs#L38

src/frontend/src/heartbeat.rs
killme2008: Fixed via https://github.com/GreptimeTeam/greptimedb/pull/4199/commits/d059e915dbc2ba6ec2f66592abf8e63b15c070a0

sunng87:
@zyy17 In our operator, we will need to pass hostname argument to frontend after this patch merged.

src/servers/src/addrs.rs
```diff
@@ -0,0 +1,54 @@
...
+            "tomcat:3002",
+            super::resolve_addr("127.0.0.1:3001", Some("tomcat:3002"))
+        );
+
+        assert_eq!(
+            "127.0.0.1:3001",
+            super::resolve_addr("127.0.0.1:3001", None)
+        );
+    }
+}
```
sunng87, line 54: I suggest to move this file to some common module like base. This module has nothing to do with server.

src/servers/src/addrs.rs
```diff
@@ -0,0 +1,54 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Resolves hostname:port address for meta registration.
+/// If `hostname_addr` is present, prefer to use it, `bind_addr` otherwise.
+///
```
v0y4g3r, line 17:
```suggestion
```

src/servers/src/addrs.rs
```diff
@@ -0,0 +1,54 @@
...
+            if hostname_addr.contains(':') {
+                hostname_addr.to_string()
+            } else {
+                // otherwise, resolve port from bind_addr
+                // should be safe to unwrap here because bind_addr is already validated
+                let port = bind_addr.split(':').nth(1).unwrap();
+                format!("{hostname_addr}:{port}")
+            }
+        }
+        None => bind_addr.to_string(),
```
v0y4g3r, line 31: Should we print a warning when fallback to `bind_addr`?

src/servers/src/addrs.rs
killme2008: The server is not just a server but has many helpers for running HTTP, gRPC servers, etc. This is another help method, so I think `server` is more proper than `base`.

killme2008:
@fengjiachun @sunng87 @v0y4g3r Please take another look.

fengjiachun:
> Flow's heartbeat task also need to do the same thing:
> 
> https://github.com/GreptimeTeam/greptimedb/blob/a44fe627ce6a16b703756e4ffb7e48559062b079/src/flow/src/heartbeat.rs#L38

Another heartbeat task, it is for flow engine.  @killme2008

killme2008:
> > Flow's heartbeat task also need to do the same thing:
> > https://github.com/GreptimeTeam/greptimedb/blob/a44fe627ce6a16b703756e4ffb7e48559062b079/src/flow/src/heartbeat.rs#L38
> 
> Another heartbeat task, it is for flow engine. @killme2008

Fixed

fengjiachun (approved):
LGTM

Issue #4193, opened 2024-06-24
[Tracking Issue] Support put directory and compression in Puffin


zhongzc:
### What problem does the new feature solve?

In order to support log indexing, we need to let Puffin provide the ability to put directories and compress.

### What does the feature do?

- [x] Support footer lz4 compression
- [x] Introduce puffin manager trait
- [x] Introduce writer for puffin manager
- [x] Introduce cache manager
- [x] Introduce reader for puffin manager

PR #4203, opened 2024-06-25
feat(puffin): implement CachedPuffinWriter

zhongzc:

#4193

Mainly implement `put_blob` and `put_dir`.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced support for compression codecs in various operations, enabling enhanced data handling capabilities.
  - Added cache management functionality to efficiently handle puffin files.

- **Bug Fixes**
  - Improved error handling for file operations, unsupported compression, and duplicate blob writes.

- **Improvements**
  - Standardized byte count handling by changing types from `usize` to `u64`, enhancing consistency.
  - Enhanced the `Blob` struct by introducing a `compression_codec` field and renaming `data` to `compressed_data`.

- **Dependency Updates**
  - Added `async-compression` and `async-walkdir` as dependencies.
  - Moved `uuid` from development to regular dependencies.


src/puffin/src/file_format/writer.rs
```diff
@@ -30,7 +31,10 @@ pub struct Blob<R> {
     pub blob_type: String,
 
     /// The data of the blob
-    pub data: R,
+    pub compressed_data: R,
+
+    /// The codec used to compress the blob.
+    pub compression_codec: Option<CompressionCodec>,
```
waynexia, line 37: How about define `None` also as a "uncompressed" branch of `CompressionCodec`, like `CompressionType`

src/puffin/src/puffin_manager/cached_puffin_manager/writer.rs
```diff
@@ -0,0 +1,193 @@
...
+    blob_keys: HashSet<String>,
+}
+
+#[async_trait]
+impl<CR, W> PuffinWriter for CachedPuffinWriter<CR, W>
+where
+    CR: AsyncRead + AsyncSeek,
+    W: AsyncWrite + Unpin + Send,
+{
+    async fn put_blob<R>(&mut self, key: &str, raw_data: R, options: PutOptions) -> Result<u64>
```
WenyXu, line 57: It seems we should put blob into `cache_manager` too.

waynexia (approved):
Mostly LGTM

src/puffin/src/file_format/writer.rs
zhongzc: I'm quite happy now. It doesn't bring much profit but there are certain modification costs.

src/puffin/src/puffin_manager/cached_puffin_manager/writer.rs
zhongzc: The reason for the difference in behavior between `blob` and `dir` here is:

1. When the user prepares a `dir` for writing, `dir` can be repeatedly read, and it is assumed that the user has relinquished ownership of `dir`. At this point, the cache can take advantage of this by directly moving the `dir` to the cache directory.
2. However, for `blob`, the reader provided by the user is not necessarily a file in the filesystem. This reader cannot be repeatedly read. If it must be implemented, the process would involve writing to the cache first, then reading back from the cache to write into puffin. But handling the potential issues with interruptions during this process is cumbersome and undesirable.

Here, I will treat the handling of `dir` as a special case that takes full advantage of the user's `dir` characteristics, rather than deliberately aligning the behavior of handling `blob` with that of handling `dir`.

WenyXu (approved):

PR #4211, opened 2024-06-25
feat(puffin): implement MokaCacheManager

zhongzc:

#4193

as @coderabbitai said

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced `MokaCacheManager` for advanced caching using the `moka` library.
  - Enhanced error management with new error variants for creation, renaming, and removal failures.

- **Refactor**
  - Updated `PuffinReader` trait to return `Result<Self::Dir>` and introduced the `DirGuard` trait.
  - Modified `CachedPuffinWriter` struct and methods to include new generic types for cache management.

- **Bug Fixes**
  - Improved cache management and error handling in the `moka_cache_manager` module.


src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
```diff
@@ -0,0 +1,475 @@
...
+
+impl MokaCacheManager {
+    fn encode_cache_key(puffin_file_name: &str, key: &str) -> String {
+        let key1 = Sha256::digest(puffin_file_name);
+        let key2 = Sha256::digest(key);
+
+        let mut hasher = Sha256::new();
+        hasher.update(key1);
+        hasher.update(key2);
+        let hash = hasher.finalize();
```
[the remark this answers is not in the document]
zhongzc: I want to try to avoid conflicts like this: (file_0, key_0) != (file_1, key_1), file_0+key_0 == file_1+key_1

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
```diff
@@ -0,0 +1,475 @@
...
+            Ok(fs::File::open(file_path).await.context(OpenSnafu)?.compat())
+        }
+    }
+
+    async fn get_dir<'a>(
+        &self,
+        puffin_file_name: &str,
+        key: &str,
+        init_fn: Box<dyn InitDirFn + Send + 'a>,
+    ) -> Result<PathBuf> {
```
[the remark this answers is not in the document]
zhongzc: Yes we do, that was one of the challenges and I forgot about it after dealing with atomicity.

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
```diff
@@ -0,0 +1,475 @@
...
+}
+
+impl MokaCacheManager {
+    #[allow(unused)]
+    pub async fn new(root: PathBuf, max_size: u64) -> Result<Self> {
+        let cache_root_clone = root.clone();
+
+        let cache = Cache::builder()
+            .max_capacity(max_size)
+            .weigher(|_: &String, size: &u64| *size as u32)
```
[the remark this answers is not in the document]
zhongzc: I will change to `(*size).try_into().unwrap_or(u32::MAX)`. Hope to work as much as possible

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
```diff
@@ -0,0 +1,927 @@
...
+pub struct MokaCacheManager {
+    /// The base directory of the cache.
+    base_dir: PathBuf,
+
+    /// The cache maintaining the cache key to the size of the file or directory.
+    cache: Cache<String, CacheValue>,
+
+    /// The unreleased directories that are waiting to be released.
+    /// The value is the count of references to the directory.
+    unreleased_dirs: Arc<Mutex<HashMap<String, usize>>>,
```
WenyXu, line 57: How about wrapping it with an `Arc`
```suggestion
    unreleased_dirs: Arc<Mutex<HashMap<String, Arc<()>>>>,
```

evenyag, line 57: Is it possible to implement the `CacheValue` as a reference counted handle to a file or a dir? It might be similar to the [`FileHandle`](https://github.com/GreptimeTeam/greptimedb/blob/352cc9dddeeca9c57e4eb253934a6788a576969a/src/mito2/src/sst/file.rs#L139-L143) in mito.

One benefit is that if we `get` a cache value from the cache and keep holding the cache value, we can ensure the file/dir won't be removed.

We can move the file/dir path to the unreleased queue in `CacheValue::drop()` and release them later. What's more, since we don't track the count in `unreleased_dirs`, we can change the mutex to a sync one.

This also avoids the retry loop in `get_blob` as we can hold the returned `CacheValue`.

zhongzc: Unlike `FileHandle` in mito, here the cache may perform multiple `get_blob`s on the same key. Consider a corner case:
1. The moka cache has already evicted a certain key, but the `FileHandle` for this key is still being held.
2. At this point, a new `get_blob` call arrives, and the file is initialized and overwritten.
3. The `FileHandle` held in step 1 is then dropped, which deletes the newly initialized file.
4. The new `get_blob` attempts to read the file and encounters a `NotFound` error.

Possible solutions to this problem are:
1. Have the `FileHandle` hold the file descriptor (fd), but this could lead to the potential problem of "opening too many files".
2. Have the cache track unreleased files, similar to how `unreleased_dirs` are handled. The way I handle `get_blob` now, the intention is essentially to obtain the fd. If that's achieved, it is considered successful. The cache is secondary and only needs to handle cleanup during eviction. Therefore, there's no need to track unreleased files.

zhongzc: The handling approach for directories is with the intention of needing a `Mutex` to avoid potential unintended deletions. As mentioned in the previous comment, a reentrant `get_dir` will inevitably need to track unreleased directories.

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
```diff
@@ -0,0 +1,941 @@
...
+        init_fn: Box<dyn InitDirFn + Send + Sync + 'a>,
+    ) -> Result<Self::Dir> {
+        let cache_key = Self::encode_cache_key(puffin_file_name, key);
+
+        let v = self
+            .cache
+            .try_get_with(cache_key.clone(), async {
+                if let Some(v) = self.recycle_bin.get(&cache_key).await {
+                    return Ok(v);
+                }
```
waynexia, line 161: is it necessary to remove hit entries from recycle bin?

zhongzc: amazing catch, remove is right

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
```diff
@@ -0,0 +1,941 @@
...
+            .context(CacheGetSnafu)
+    }
+}
+
+impl MokaCacheManager {
+    fn encode_cache_key(puffin_file_name: &str, key: &str) -> String {
+        let mut hasher = Sha256::new();
+        hasher.update(puffin_file_name);
+        hasher.update(key);
+        hasher.update(puffin_file_name);
```
evenyag, line 220: Why do we update the puffin file name twice?

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
```diff
@@ -0,0 +1,941 @@
...
+    }
+
+    async fn write_blob(
+        target_path: &PathBuf,
+        init_fn: &(dyn InitBlobFn + Send + Sync + '_),
+    ) -> Result<u64> {
+        // To guarantee the atomicity of writing the file, we need to write
+        // the file to a temporary file first...
+        let tmp_path = target_path.with_extension(TMP_EXTENSION);
+        let _ = fs::remove_file(&tmp_path).await;
```
evenyag, line 233: Do we need to check the error kind and log the error if it is not `NotFound`?

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
```diff
@@ -0,0 +1,941 @@
...
+    }
+
+    async fn write_dir(
+        target_path: &PathBuf,
+        init_fn: &(dyn InitDirFn + Send + Sync + '_),
+    ) -> Result<u64> {
+        // To guarantee the atomicity of writing the directory, we need to write
+        // the directory to a temporary directory first...
+        let tmp_base = target_path.with_extension(TMP_EXTENSION);
+        let _ = fs::remove_dir_all(&tmp_base).await;
```
evenyag, line 257: ditto. Maybe we should log the error at least.

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
```diff
@@ -0,0 +1,941 @@
...
+                    break;
+                }
+            }
+        }
+    }
+}
+
+impl Drop for MokaCacheManager {
+    fn drop(&mut self) {
+        let _ = self.delete_queue.try_send(DeleteTask::Terminate);
```
evenyag, line 393: Maybe we don't need this task as the receiver will return `None` once the sender is dropped.

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
```diff
@@ -0,0 +1,941 @@
...
+
+        while let Some(entry) = wd.next().await {
+            let entry = entry.context(WalkDirSnafu)?;
+            size += entry.metadata().await.context(MetadataSnafu)?.len();
+        }
+
+        Ok(size)
+    }
+
+    async fn delete_routine(mut receiver: Receiver<DeleteTask>) {
```
evenyag, line 353: Add a log after the `delete_routine` quits.

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
zhongzc: Here is the initiative to let `delete_routine` exit, otherwise the `*Guard` in the cache will delete the file when the `MokaCacheManager` drops

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
zhongzc: It can avoid conflicts caused by different `(puffin_file_name, key)` but the same after concatenation.

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
evenyag: Got it. Could you please document this in the comment?

src/puffin/src/puffin_manager/cache_manager/moka_cache_manager.rs
evenyag: Could you please document this in the comment?

evenyag (approved):

waynexia:
@coderabbitai pause

PR #4212, opened 2024-06-25
refactor: expose `DatanodeBuilder::build_object_store_manager()` and `MitoConfig::sanitize()`

zyy17:

Some minor refactoring:

1. Expose `DatanodeBuilder::build_object_store_manager()`

   The scenario is the following code:

   ```rust
   ...
   let object_storage_manager =
            datanode::datanode::DatanodeBuilder::build_object_store_manager(&storage_config)
                .await
                .context(BuildObjectStoreManagerSnafu)?;
   ...
   let compaction_region = open_compaction_region(
                    &open_region_request,
                    &self.mito_config,
                    object_storage_manager.clone(),
                )
   ...
   ```

2. Expose `MitoConfig::sanitize()` to initialize Mito config outside the crate;

~~2. Set default value for `intermediate_path` instead of empty value(it will failed to [create a access layer](https://github.com/GreptimeTeam/greptimedb/blob/main/src/mito2/src/compaction/compactor.rs#L98-L101) if using `open_compaction_region()` with default Mito config) (cc @zhongzc);~~

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Set a default data home directory for file storage, enhancing ease of configuration.
  - Updated intermediate path configurations to use the default data home directory across various components.

- **Documentation**
  - Revised configuration documentation to reflect changes in default paths.

- **Configuration**
  - Updated example configuration files to include the new default paths for intermediate storage.

### DIFF
--- a/src/mito2/src/config.rs
+++ b/src/mito2/src/config.rs
@@ -151,7 +151,7 @@ impl MitoConfig {
     /// Sanitize incorrect configurations.
     ///
     /// Returns an error if there is a configuration that unable to sanitize.
-    pub(crate) fn sanitize(&mut self, data_home: &str) -> Result<()> {
+    pub fn sanitize(&mut self, data_home: &str) -> Result<()> {
         // Use default value if `num_workers` is 0.
         if self.num_workers == 0 {
             self.num_workers = divide_num_cpus(2);
--- a/src/datanode/src/datanode.rs
+++ b/src/datanode/src/datanode.rs
@@ -52,7 +52,7 @@ use store_api::storage::RegionId;
 use tokio::fs;
 use tokio::sync::Notify;
 
-use crate::config::{DatanodeOptions, RegionEngineConfig};
+use crate::config::{DatanodeOptions, RegionEngineConfig, StorageConfig};
 use crate::error::{
     self, BuildMitoEngineSnafu, CreateDirSnafu, GetMetadataSnafu, MissingKvBackendSnafu,
     MissingNodeIdSnafu, OpenLogStoreSnafu, Result, RuntimeResourceSnafu, ShutdownInstanceSnafu,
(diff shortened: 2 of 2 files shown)


src/mito2/src/config.rs
```diff
@@ -298,7 +298,7 @@ impl Default for InvertedIndexConfig {
             apply_on_query: Mode::Auto,
             write_buffer_size: ReadableSize::mb(8),
             mem_threshold_on_create: Some(ReadableSize::mb(64)),
-            intermediate_path: String::new(),
+            intermediate_path: join_dir(common_config::DEFAULT_DATA_HOME, "index_intermediate"),
```
evenyag, line 301: Does it mean that we change the `intermediate_path` from `data_home/index_intermediate` to a fixed path `/tmp/greptimedb/index_intermediate`? This might be unexpected.

We should call the `MitoConfig::sanitize()` method first. It will set the intermediate path correctly.

zyy17: Maybe it's ok in `default()` ? Or should we need to sanitize the intermediate path in the `default()`? If not, we need to write the following code to make sure the config is correct:

```rust
let mut cfg = Mito::default();
cfg.sanitize("<data_home>")
``` 

Of course, the `sanitize()` needs to be exposed too.

zyy17: If we're ok with that, I can revert the above modifications and expose `sanitize()`.

evenyag: We can provide a method in `InvertedIndexConfig` to get the `intermediate_path` or return a default if it is not set. If we set the path in `default()`, then we can't distinguish whether to set the path in `sanitize()` as the path is not empty.

zyy17: It seems less elegant. The root cause is that `data_home` is not part of the Mito config. My suggestion is to make `sanitize()` public. It will be easy to initialize Mito outside the crate.

zhongzc: Why dont we expose `sanitize()`?

zyy17: Feel good to me.

zyy17:
@evenyag @zhongzc PTAL

evenyag (approved):

WenyXu (approved):
LGTM

Issue #3354, opened 2024-02-22
Extensions to SHOW Statements for information_schema tables


killme2008:
### What type of enhancement is this?

API improvement, User experience

### What does the enhancement do?

MySQL supports `SHOW XXXX from [table]` or `SHOW FULL XXX from [table]` to query information_schema, for example:

```sql
SHOW FULL TABLES WHERE Table_type != 'VIEW'

SHOW TABLE STATUS

SHOW PROCEDURE STATUS where DB='public';

SHOW INDEX FROM `monitor`;

SHOW FULL COLUMNS from `monitor`;

```

We have to support these syntax sugars if we want to improve the compatibility of MySQL protocol.

Tasks:

* [x] Refactor `show tables` and `show databases` executor, make them querying from `information_schema` and support `where` clause such as `SHOW FULL TABLES WHERE Table_type != 'VIEW'`. https://github.com/GreptimeTeam/greptimedb/pull/3423
* [x] `show columns` https://github.com/GreptimeTeam/greptimedb/pull/3577
* [x] `show table status` https://github.com/GreptimeTeam/greptimedb/pull/4303
* [x] `show index` https://github.com/GreptimeTeam/greptimedb/pull/3577
* [x] `SHOW CHARACTER SET` https://github.com/GreptimeTeam/greptimedb/pull/3753
* [x] `SHOW COLLATION` https://github.com/GreptimeTeam/greptimedb/pull/3753
* [ ] Improve `show variables`.

killme2008:
The related MySQL documents:

* https://dev.mysql.com/doc/refman/8.0/en/show.html
* https://dev.mysql.com/doc/refman/8.0/en/extended-show.html

I'll implement the extensions to the show statement for information_schema in 0.7

killme2008:
I'll implement the below show statements:

```sql
SHOW CHARACTER SET
SHOW COLLATION
SHOW COLUMNS
SHOW DATABASES
SHOW FUNCTION STATUS
SHOW INDEX
SHOW OPEN TABLES
SHOW PROCEDURE STATUS
SHOW STATUS
SHOW TABLE STATUS
SHOW TABLES
SHOW TRIGGERS
SHOW VARIABLES
```

MichaelScofield:
Navicat will submit "show status" and "show table status" upon connection.

killme2008:
> Navicat will submit "show status" and "show table status" upon connection.

`show status` looks working but returns empty:

```sql
[...]

PR #2022, opened 2023-07-24
refactor: query context

shuiyisong:
I hereby agree to the terms of the [GreptimeDB CLA](https://gist.github.com/xtang/6378857777706e568c1949c7578592cc)

## What's changed and what's your intention?

This pr mainly refactors `QueryContext`
1. change builder to use `derive_builder`
2. make `catalog` and `schema` within query context immutable
  - stateless clients like grpc and http should be able to change it on demand before sending request
  - stateful clients like mysql and pg should use `Session` to hold `catalog` and `schema`, and construct immutable query context on receiving next query
3. mark `arc()` as test only. Query context should be constructed with catalog and schema explicitly to avoid potential bug

Discussion
1. now that we actually use session to hold mutable `catalog` and `schema`, should we remove `Statement::Use` since its basically unreachable?
  - on mysql it calls `on_init` rather than treat `use` statement as a normal query
  - we require schema parameter on connecting with pg protocol. it would be strange to allow `use` statement within pg session; or we can process it before it goes into `SqlHandler` like mysql does

## Checklist

- [ ]  I have written the necessary rustdoc comments.
- [ ]  I have added the necessary unit tests and integration tests.

PR #4218, opened 2024-06-26
feat: dbeaver mysql compatibility, use statement and information_schema.tables

sunng87:

image

This patch is to resolve compatibility issues with dbeaver client:

- it tries to access some field like `data_size` in information_schema.tables
- it uses `use` statement instead of modern `use` command which most other clients use

This patch fills these two gaps and make dbeaver client works for most cases. There is another statement to be supported: `SHOW TABLE STATUS FROM public LIKE 'system_metrics'` but it's not a blocking issue for our scenario. `SHOW TABLE STATUS` is tracked in #3354 

This patch brings back `USE` statement support in #2022 as we finally found a real use-case for it. And it also makes `current_schema` modifiable and propagate to `Session`.

The data fields of `information_schema.TABLES` in this patch are all for placeholder. I haven't fill real data into these fields. TODO item is left to future enhancement.

## Future work

I'm going refactor `Session` and `QueryContext` to use `Arc<RwLock<>>` for all modifiable fields, like time-zone, schema. This makes the change automatically propagate from `QueryContext` to `Session`. The idea of `QueryContext::update_session` is smart, but it's going to be difficult to maintain in future.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced new columns (`data_length`, `max_data_length`, `index_length`, `avg_row_length`) in the `information_schema` tables for enhanced data insight.
  - Enhanced query result structure with additional relevant columns for better data representation.

- **Improvements**
  - Implemented custom display formatting for `TableOptions` to improve readability.
  
- **Tests**
  - Added tests to verify the string conversion of `TableOptions` ensuring accurate display.
  
- **Bug Fixes**
  - Corrected column arrangement in query results to ensure consistency and correctness.


killme2008:
> I'm going refactor Session and QueryContext to use Arc<RwLock<>> for all modifiable fields, like time-zone, schema. This > makes the change automatically propagate from QueryContext to Session. The idea of QueryContext::update_session is > smart, but it's going to be difficult to maintain in future.

It may hurt performance, must be careful or do some benchmark.

killme2008:
There are many columns in `tables`, see https://dev.mysql.com/doc/refman/8.4/en/information-schema-tables-table.html

I believe we should consider including all placeholders at this time.

sunng87:
> It may hurt performance, must be careful or do some benchmark.

Should be ok because these data are all associated with a single database connection. Actually there is no concurrent access at all. Using locks is just to bypass rust compiler check.

killme2008 (approved):
LGTM

killme2008:
@evenyag @fengjiachun Please take a look.

src/catalog/src/information_schema/tables.rs
```diff
@@ -213,9 +287,37 @@ impl InformationSchemaTablesBuilder {
...
+        self.auto_increment.push(Some(0));
+        self.row_format.push(Some("Fixed"));
+        self.table_collation.push(None);
+        self.update_time.push(None);
+        self.check_time.push(None);
+
+        self.version.push(Some(11));
+        self.table_comment.push(table_info.desc.as_deref());
+        self.create_options
+            .push(Some(format!("{:?}", table_info.meta.options).as_ref()));
```
evenyag, line 310: Maybe serialize it to JSON format?

src/catalog/src/information_schema/tables.rs
```diff
@@ -213,9 +287,37 @@ impl InformationSchemaTablesBuilder {
...
+        self.checksum.push(Some(0));
+        self.table_rows.push(Some(0));
+        self.data_free.push(Some(0));
+        self.auto_increment.push(Some(0));
+        self.row_format.push(Some("Fixed"));
+        self.table_collation.push(None);
+        self.update_time.push(None);
+        self.check_time.push(None);
+
+        self.version.push(Some(11));
```
evenyag, line 307: non-blocking: Any reason to use 11 as the version?

src/servers/src/mysql/federated.rs
```diff
@@ -249,15 +246,12 @@ fn check_show_variables(query: &str) -> Option<Output> {
...
-fn check_others(query: &str, query_ctx: QueryContextRef) -> Option<Output> {
+fn check_others(query: &str, _query_ctx: QueryContextRef) -> Option<Output> {
     if OTHER_NOT_SUPPORTED_STMT.is_match(query.as_bytes()) {
         return Some(Output::new_with_record_batches(RecordBatches::empty()));
     }
 
-    let recordbatches = if SELECT_DATABASE_PATTERN.is_match(query) {
-        let schema = query_ctx.current_schema();
-        Some(select_function("database()", schema))
-    } else if SELECT_TIME_DIFF_FUNC_PATTERN.is_match(query) {
```
evenyag, line 260: Why do we remove this?

src/catalog/src/information_schema/tables.rs
killme2008: It's better in the format of `key=value` and joined with `,`, just like them in SQL.

src/servers/src/mysql/federated.rs
sunng87: We already have the `database()` as a built in function

src/catalog/src/information_schema/tables.rs
evenyag: Let's document this in the comment

src/catalog/src/information_schema/tables.rs
sunng87: The mysql way is joined with ` `, which I just updated.

src/catalog/src/information_schema/tables.rs
sunng87: I've add comments to explain this.

src/table/src/requests.rs
```diff
@@ -128,6 +129,25 @@ impl TableOptions {
...
+            key_vals.push(format!("{}={}", TTL_KEY, humantime::Duration::from(ttl)));
+        }
+
+        for (k, v) in &self.extra_options {
+            key_vals.push(format!("{}={}", k, v));
+        }
+
+        write!(f, "{}", key_vals.join(" "))
+    }
+}
```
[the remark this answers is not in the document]
sunng87: I believe my original one is better

sunng87:
@evenyag just updated, please check again

src/table/src/requests.rs
```diff
@@ -128,6 +129,25 @@ impl TableOptions {
     }
 }
 
+impl fmt::Display for TableOptions {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut key_vals = vec![];
+        if let Some(size) = self.write_buffer_size {
+            key_vals.push(format!("{}={}", WRITE_BUFFER_SIZE_KEY, size));
```
evenyag, line 136: Should we quote key values by `""`?
e.g. `"write_buffer_size"="128.0MiB" "ttl"="16m 40s"`

evenyag (approved):

src/table/src/requests.rs
killme2008: Looks like It's better to convert it into `OptionMap`, call its function `kv_pairs`, and join the result. This function will obscure information such as access keys.

https://github.com/GreptimeTeam/greptimedb/blob/be29e48a60345599373128c8cd6632f4f407165c/src/sql/src/statements/option_map.rs#L79

sunng87: It seems mysql/mariadb is writing it without quote

```
MariaDB [(none)]> select create_options from information_schema.tables limit 5;
+--------------------------------------------+
| create_options                             |
+--------------------------------------------+
| max_rows=10380835156842741 transactional=0 |
| max_rows=10720                             |
| max_rows=34952                             |
| max_rows=8587869680497929 transactional=0  |
| max_rows=51150                             |
+--------------------------------------------+
```

killme2008: I don't think we need to keep it compatible with MySQL or Mariadb in this column.

PR #4224, opened 2024-06-27
build(deps): Upgrade OpenDAL to 0.47

tisonkun:

Upgrade OpenDAL to 0.47.1.

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Improved the efficiency and flexibility of cache handling by making `LruCacheLayer` and `LruCacheAccess` generic over a type that implements `Access`.
  - Renamed `read_cache_stat` to `cache_stat` for consistency.
  - Updated various method signatures to enhance performance and reliability.


src/object-store/src/layers/lru_cache/read_cache.rs
```diff
@@ -51,9 +51,9 @@ fn can_cache(path: &str) -> bool {
     !path.ends_with("_last_checkpoint")
 }
 
-/// Generate an unique cache key for the read path and range.
-fn read_cache_key(path: &str, range: BytesRange) -> String {
-    format!("{:x}.cache-{}", md5::compute(path), range.to_header())
+/// Generate a unique cache key for the read path and range.
+fn read_cache_key(path: &str) -> String {
+    format!("{:x}.cache", md5::compute(path))
 }
```
tisonkun, line 57: @Xuanwo supposes that we don't need to cache ranges but always read the whole object.

I'll comment on other diff below.

src/object-store/src/layers/lru_cache/read_cache.rs
```diff
@@ -275,7 +262,7 @@ impl<I: Access> Read for ReadCacheReader<I> {
                     }
                     Err(_) => {
                         OBJECT_STORE_LRU_CACHE_MISS.inc();
-                        self.inner_reader.read_at(offset, limit).await
+                        self.inner_reader.read().await
```
tisonkun, line 265: I guess we need `read` rather than `read_all` here. Because we are in the implementation of `read`. cc @Xuanwo please confirm.

src/object-store/src/layers/lru_cache/read_cache.rs
```diff
@@ -211,11 +200,11 @@ pub struct ReadCacheReader<I: Access> {
 
 impl<I: Access> ReadCacheReader<I> {
     /// TODO: we can return the Buffer directly to avoid another read from cache.
-    async fn read_remote(&self, offset: u64, limit: usize) -> Result<ReadResult> {
+    async fn read_remote(&mut self) -> Result<ReadResult> {
         OBJECT_STORE_LRU_CACHE_MISS.inc();
 
-        let buf = self.inner_reader.read_at(offset, limit).await?;
-        let result = self.try_write_cache(buf, offset).await;
+        let buf = self.inner_reader.read_all().await?;
```
tisonkun, line 206: So, instead of chunked read, we `read_all` the object here. It may cause some regressions like reading a large object can be slow, worse concurrency and worse cache friendly.

@Xuanwo told me that https://github.com/apache/opendal/issues/4672 is for this case. So I leave this to him for any further improvement.

@WenyXu You may test if we suffer some significant regression here.

src/object-store/src/layers/lru_cache/read_cache.rs
Xuanwo: Hi, this understanding is not true. I will give it a fix.

src/object-store/src/layers/lru_cache/read_cache.rs
Xuanwo: Should be addressed in https://github.com/tisonkun/greptimedb/pull/7

Xuanwo:
The experience of implementing caching in OpenDAL layer is not cool. I plan to make it easier in the future.

src/object-store/src/layers/lru_cache/read_cache.rs
tisonkun: Thank you!

Xuanwo:
`coderabbitai` doesn't look good to me, more like spam.

src/object-store/src/layers/lru_cache/read_cache.rs
WenyXu: > @WenyXu You may test if we suffer some significant regression here.

The original design is cache the whole file. I think we may remove this cache layer ASAP. cc @evenyag

Xuanwo: > The original design is cache the whole file. I think we may remove this cache layer ASAP. cc @evenyag

Hi, I didn't get the enough context here. Our current implemention should be exact same with 0.6 version.

killme2008: > > @WenyXu You may test if we suffer some significant regression here.
> 
> The original design is cache the whole file. I think we may remove this cache layer ASAP. cc @evenyag

The original design is to cache the chunk.

WenyXu: > Hi, I didn't get the enough context here. Our current implemention should be exact same with 0.6 version.

> The original design is to cache the chunk.

Sorry, I misremembered. The original design is chunk caching.

src/object-store/src/layers/lru_cache.rs
```diff
@@ -68,14 +68,14 @@ impl<I: Access> Layer<I> for LruCacheLayer {
...
     inner: I,
-    read_cache: ReadCache,
+    read_cache: ReadCache<C>,
 }
 
-impl<I: Access> LayeredAccess for LruCacheAccess<I> {
+impl<I: Access, C: Access> LayeredAccess for LruCacheAccess<I, C> {
     type Inner = I;
-    type Reader = Arc<dyn ReadDyn>;
+    type Reader = Box<dyn ReadDyn>;
```
killme2008, line 78: The same question, why change it from `Arc` to `Box`?

src/object-store/src/layers/lru_cache/read_cache.rs
```diff
@@ -51,23 +51,36 @@ fn can_cache(path: &str) -> bool {
     !path.ends_with("_last_checkpoint")
 }
 
-/// Generate an unique cache key for the read path and range.
-fn read_cache_key(path: &str, range: BytesRange) -> String {
-    format!("{:x}.cache-{}", md5::compute(path), range.to_header())
+/// Generate a unique cache key for the read path and range.
+fn read_cache_key(path: &str, args: &OpRead) -> String {
```
killme2008, line 55: So the `range` is removed?

src/object-store/src/layers/lru_cache/read_cache.rs
```diff
@@ -149,73 +162,110 @@ impl ReadCache {
...
-    mem_cache: Cache<String, ReadResult>,
-}
+    async fn try_write_cache<I>(&self, mut reader: I::Reader, read_key: &str) -> Result<usize>
+    where
+        I: Access,
+    {
+        let (_, mut writer) = self.file_cache.write(read_key, OpWrite::new()).await?;
+        let mut total = 0;
+        loop {
+            let bytes = reader.read().await?;
```
killme2008, line 241: What about using `read_all` to avoid this loop?

killme2008: Of course, it's better if there is a function that can copy from `reader` to `writer`.

src/object-store/src/layers/lru_cache/read_cache.rs
tisonkun: I think it's contained within `OpRead`. See `args.range().to_header()` below.

src/object-store/src/layers/lru_cache/read_cache.rs
tisonkun: I think `read_all` may cause extra memory consumption and this implementation flush out the bytes ASAP. But ideally we can have a streaming solution as you suggested..

killme2008: Make sense.

Xuanwo: > Of course, it's better if there is a function that can copy from `reader` to `writer`, just like https://doc.rust-lang.org/std/io/fn.copy.html

Perhaps, but not to the extent of `oio`. I'm considering how to enhance the UX within Layer. I wish we can have a way to implement cache without touch `opendal::raw`.

WenyXu (approved):
LGTM

evenyag (approved):

src/object-store/src/layers/lru_cache.rs
tisonkun: And we don't actually clone this point in the GreptimeDB codebase. It relies on the OpenDAL codebase to manage out this pointer. Perhaps we should use its exported type. Let me try.

tisonkun: Done in c6e9f1293e25fcb55c1a87816ce7facda5e595e1

PR #4225, opened 2024-06-28
refactor: add interceptor after Influxdb lines are converted to grpc row insert

MichaelScofield:

... so that other projects could modify the row insert according to their specific intention.

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Improved the initialization sequence for frontend services, simplifying the method calls and argument passing for better performance and maintainability.

These changes enhance the backend structure without affecting user-facing features directly.


killme2008:
@MichaelScofield It's better to use an option to control this behavior and it's disabled by default.

MichaelScofield:
@killme2008 The current behavior is unaffected, new influxdb tables are still nanosecond time index. This PR simply brings compatibility back, so I don't think there's a need for an option.

killme2008:
> @killme2008 The current behavior is unaffected, new influxdb tables are still nanosecond time index. This PR simply brings compatibility back, so I don't think there's a need for an option.

Alignment and checking typically impact performance, often unnecessary.

In other words, the user may create a table with a millisecond time index, and ingestion with influxdb, which we want to prohibit.

killme2008:
We already discussed this before and have an agreement.  Maybe you can discuss it directly with @fengjiachun . I have to sleep.

MichaelScofield:
@killme2008 I have added an option for that.

src/servers/src/influxdb.rs
```diff
@@ -86,12 +93,34 @@ impl TryFrom<InfluxdbRequest> for RowInsertRequests {
             });
             row_writer::write_fields(table_data, fields, &mut one_row)?;
 
-            // timestamp
-            row_writer::write_ts_to_nanos(
+            // Align to the table's time index unit if needed.
+            let target_unit = if auto_align_precision {
+                catalog_manager
```
fengjiachun, line 98: I'm unsure if introducing the catalog here is a good idea.

src/servers/src/influxdb.rs
```diff
@@ -132,7 +162,11 @@ monitor2,host=host4 cpu=66.3,memory=1029 1663840496400340003";
             lines: lines.to_string(),
         };
 
-        let requests: RowInsertRequests = influxdb_req.try_into().unwrap();
+        let catalog_manager = MemoryCatalogManager::new() as _;
```
killme2008, line 165: It's better to add a test to cover `auto_align_precision=true`. You can use a mock catalog manager like `MemoryCatalogManager`.

tests-integration/src/cluster.rs
```diff
@@ -376,14 +377,22 @@ impl GreptimeDbClusterBuilder {
             Arc::new(InvalidateTableCacheHandler::new(cache_registry.clone())),
         ]);
 
+        let options = FrontendOptions {
+            influxdb: InfluxdbOptions {
+                enable: true,
```
killme2008, line 382: It's better to cover the old behavior too.

src/servers/src/influxdb.rs
killme2008: Agree. But I don't have a better solution. Maybe adds some handler to breaking the strong coupling, just like:

```rust
pub struct AlignPrecisionHandler {
   caatlog_manager: CatalogManagerRef,
}
impl AlignPrecisionHandler {
   pub async fn get_target_time_unit(query_ctx: &QueryContextRef, table_name: &str) -> Result<TimeUnit> {
        ....
   }
}

....
to_row_insert_requests(AlignPrecisionHandler{catalog_manager}, ...);


```

tests-integration/src/tests.rs
```diff
@@ -15,7 +15,7 @@
 mod instance_kafka_wal_test;
 mod instance_test;
 mod promql_test;
-mod test_util;
+pub mod test_util;
```
killme2008, line 18:
```suggestion
pub(crate) mod test_util;
```

killme2008 (approved):
LGTM. A better way.

tisonkun (approved):

PR #4232, opened 2024-06-29
refactor: use rwlock for modifiable session data

sunng87:

Follow up for #4218 . Need a rebase after #4218 merged.

Some of the properties, like schema and timezone, are modifiable within a session. 

Our original design is to generate a read-only `QueryContext`, then pass it with all the execution process path. This works well before we have statements that modify some of the session data, like `SET time_zone` and `USE`.

After that, we introduce a new method called `update_session` to give back modified properties to session, after query execution. See https://github.com/GreptimeTeam/greptimedb/blob/b739c9fd10ba486726f6ee30d260792081eafd2f/src/servers/src/mysql/handler.rs#L107 . The implementation of `update_session` has to check each modifiable field and update the session. This is because `ArcSwap` data structure is not clonable. So we cannot share reference between the parent `Session` and the child `QueryContext`.

This main point of this patch is to change those modifiable fields from `ArcSwap` to `Arc<RwLock>` so that we can share it between parent and child. Note that because `Session` is associated with a signle connection, so actually there is no race condition with reading and updating those modifiable fields.  

This patch also include a feature flag fix for orc-rust to exclude it datafusion dependency.

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Improved thread safety by replacing `ArcSwap` with `Arc<RwLock>` for fields in `Session` struct, enhancing concurrent access management.

- **Bug Fixes**
  - Simplified user handling by removing unnecessary wrappers, improving code clarity and conciseness.
  - Adjusted method calls to improve timezone and schema handling in query contexts.

- **Chores**
  - Updated dependencies to include `async` feature for `orc-rust`, enhancing asynchronous capabilities.

### DIFF
--- a/src/auth/src/permission.rs
+++ b/src/auth/src/permission.rs
@@ -42,7 +42,7 @@ pub enum PermissionResp {
 pub trait PermissionChecker: Send + Sync {
     fn check_permission(
         &self,
-        user_info: Option<UserInfoRef>,
+        user_info: UserInfoRef,
         req: PermissionReq,
     ) -> Result<PermissionResp>;
 }
@@ -50,7 +50,7 @@ pub trait PermissionChecker: Send + Sync {
 impl PermissionChecker for Option<&PermissionCheckerRef> {
     fn check_permission(
         &self,
-        user_info: Option<UserInfoRef>,
+        user_info: UserInfoRef,
         req: PermissionReq,
     ) -> Result<PermissionResp> {
         match self {
--- a/src/auth/tests/mod.rs
+++ b/src/auth/tests/mod.rs
@@ -27,7 +27,7 @@ struct DummyPermissionChecker;
 impl PermissionChecker for DummyPermissionChecker {
     fn check_permission(
         &self,
-        _user_info: Option<UserInfoRef>,
+        _user_info: UserInfoRef,
         req: PermissionReq,
     ) -> auth::error::Result<PermissionResp> {
         match req {
@@ -45,13 +45,13 @@ fn test_permission_checker() {
     let checker: PermissionCheckerRef = Arc::new(DummyPermissionChecker);
 
     let grpc_result = checker.check_permission(
-        None,
+        auth::userinfo_by_name(None),
         PermissionReq::GrpcRequest(&Request::Query(Default::default())),
     );
     assert_matches!(grpc_result, Ok(PermissionResp::Allow));
 
     let sql_result = checker.check_permission(
-        None,
+        auth::userinfo_by_name(None),
         PermissionReq::SqlStatement(&Statement::ShowDatabases(ShowDatabases::new(
             ShowKind::All,
             false,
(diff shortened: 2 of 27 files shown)


src/auth/tests/mod.rs
```diff
@@ -45,20 +45,21 @@ fn test_permission_checker() {
     let checker: PermissionCheckerRef = Arc::new(DummyPermissionChecker);
 
     let grpc_result = checker.check_permission(
-        None,
+        auth::userinfo_by_name(None),
```
sunng87, line 48: @shuiyisong In this patch, I changed empty user from `None::Option<UserInfoRef>` to `auth::userinfo_by_name(None)`. Just to confirm it won't break some of our assumptions.

src/session/src/lib.rs
```diff
@@ -32,11 +31,11 @@ use crate::context::{Channel, ConnInfo, QueryContextRef};
...
 pub struct Session {
-    catalog: ArcSwap<String>,
-    schema: ArcSwap<String>,
-    user_info: ArcSwap<UserInfoRef>,
+    catalog: Arc<RwLock<String>>,
+    schema: Arc<RwLock<String>>,
+    user_info: Arc<RwLock<UserInfoRef>>,
     conn_info: ConnInfo,
-    timezone: ArcSwap<Timezone>,
+    timezone: Arc<RwLock<Timezone>>,
```
evenyag, line 38: Consider moving mutable fields to a new struct. e.g. `MutableSession`.
```rust
pub struct Session {
    mutable: Arc<RwLock<MutableSession>>.
}
```

src/session/src/context.rs
```diff
@@ -38,11 +37,10 @@ pub type ConnInfoRef = Arc<ConnInfo>;
...
     current_catalog: String,
-    #[builder(setter(custom))]
-    current_schema: ArcSwap<String>,
-    current_user: ArcSwap<Option<UserInfoRef>>,
-    #[builder(setter(custom))]
-    timezone: ArcSwap<Timezone>,
+    // we use Arc<RwLock>> for modifiable fields
+    current_schema: Arc<RwLock<String>>,
+    current_user: Arc<RwLock<UserInfoRef>>,
+    timezone: Arc<RwLock<Timezone>>,
```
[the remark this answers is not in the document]
sunng87: Yes, good idea. That will the code more clear.

src/auth/src/permission.rs
```diff
@@ -42,15 +42,15 @@ pub enum PermissionResp {
 pub trait PermissionChecker: Send + Sync {
     fn check_permission(
         &self,
-        user_info: Option<UserInfoRef>,
+        user_info: UserInfoRef,
```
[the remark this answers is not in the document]
sunng87: I just updated this interface to keep it consistent with other ones that involves `UserInfo`. Perhaps we can address this in future PR.

evenyag (approved):

src/common/datasource/Cargo.toml
```diff
@@ -31,7 +31,9 @@ derive_builder.workspace = true
 futures.workspace = true
 lazy_static.workspace = true
 object-store.workspace = true
-orc-rust = { git = "https://github.com/datafusion-contrib/datafusion-orc.git", rev = "502217315726314c4008808fe169764529640599" }
+orc-rust = { git = "https://github.com/datafusion-contrib/datafusion-orc.git", rev = "502217315726314c4008808fe169764529640599", default-features = false, features = [
+    "async",
+] }
```
sunng87, line 36: @WenyXu Here I removed additional datafusion dependency

shuiyisong (approved):
LGTM

PR #4272, opened 2024-07-03
fix: prepare inserting with column defaults not work, #4244

### LINKED ISSUE #4244: Failed to support insert data without specify the `ts` column when using `?` placeholder
### What type of bug is this?

Unexpected error

Query Engine

## Test code

I'm using the Go and `github.com/go-sql-driver/mysql` to connect greptimedb:

```go
package main

import (
	"context"
	"database/sql"

	"github.com/go-sql-driver/mysql"
)

func main() {
	if err := testDB(); err != nil {
		panic(err)
	}
}

func testDB() error {
	cfg := mysql.Config{
		Net:                  "tcp",
		Addr:                 "localhost:4002",
		AllowNativePasswords: true,
	}
	ctx := context.Background()

	db, err := sql.Open("mysql", cfg.FormatDSN())
	if err != nil {
		return err
	}

	conn, err := db.Conn(ctx)
	if err != nil {
		return err
	}

	createStandaloneTableSQL := `CREATE TABLE test_table(
							ts TIMESTAMP DEFAULT current_timestamp(),
							n INT,
							row_id INT,
							PRIMARY KEY(n),
							TIME INDEX (ts)
						  )
						  engine=mito`

	_, err = db.Exec(createStandaloneTableSQL)
	if err != nil {
		return err
	}

	insertDataSQL := `INSERT INTO test_table(n, row_id) VALUES (?, ?);`
	for i := 0; i < 10; i++ {
		_, err = conn.ExecContext(ctx, insertDataSQL, i, i)
		if err != nil {
			return err
		}
	}

	return nil
}

```

## Test steps

1. Run the greptimedb

```
docker run -p 127.0.0.1:4000-4003:4000-4003 \
-v "$(pwd)/greptimedb:/tmp/greptimedb" \
--name greptime --rm \
greptime-registry.cn-hangzhou.cr.aliyuncs.com/greptime/greptimedb:v0.8.2 standalone start \
--http-addr 0.0.0.0:4000 \
--rpc-addr 0.0.0.0:4001 \
--mysql-addr 0.0.0.0:4002 \
--postgres-addr 0.0.0.0:4003
```

2. Run the test code

```
go run test-placeholder.go
panic: Error 1815: Invalid request to region 4398046511104(1024, 0), reason: column ts is not null but input has null
...
```

After multiple tests, the above test failed when the db release version > v0.7.2. **It works well in v0.7.2**.

3. Specify the `ts` column **can work**

```go
package main

import (
	"context"
	"database/sql"
	"time"

	"github.com/go-sql-driver/mysql"
)

func main() {
	if err := testDB(); err != nil {
		panic(err)
	}
}

func testDB() error {
	cfg := mysql.Config{
		Net:                  "tcp",
		Addr:                 "localhost:4002",
		AllowNativePasswords: true,
	}
	ctx := context.Background()

	db, err := sql.Open("mysql", cfg.FormatDSN())
	if err != nil {
		return err
	}

	conn, err := db.Conn(ctx)
	if err != nil {
		return err
	}

	createStandaloneTableSQL := `CREATE TABLE test_table(
							ts TIMESTAMP DEFAULT current_timestamp(),
							n INT,
							row_id INT,
							PRIMARY KEY(n),
							TIME INDEX (ts)
						  )
						  engine=mito`

	_, err = db.Exec(createStandaloneTableSQL)
	if err != nil {
		return err
	}

	insertDataSQL := `INSERT INTO test_table(n, row_id, ts) VALUES (?, ?, ?);`
	for i := 0; i < 10; i++ {
		now := time.Now()
		timestampInMillisecond := now.Unix()*1000 + int64(now.Nanosecond())/1e6
		_, err = conn.ExecContext(ctx, insertDataSQL, i, i, timestampInMillisecond)
		if err != nil {
			return err
		}
	}

	return nil
}
```

4. If I don't use the `?` placeholder, **it works**

```go
package main

import (
	"context"
	"database/sql"
	"fmt"

	"github.com/go-sql-driver/mysql"
)

func main() {
	if err := testDB(); err != nil {
		panic(err)
	}
}

func testDB() error {
	cfg := mysql.Config{
		Net:                  "tcp",
		Addr:                 "localhost:4002",
		AllowNativePasswords: true,
	}
	ctx := context.Background()

	db, err := sql.Open("mysql", cfg.FormatDSN())
	if err != nil {
		return err
	}

	conn, err := db.Conn(ctx)
	if err != nil {
		return err
	}

	createStandaloneTableSQL := `CREATE TABLE test_table(
							ts TIMESTAMP DEFAULT current_timestamp(),
							n INT,
							row_id INT,
							PRIMARY KEY(n),
							TIME INDEX (ts)
						  )
						  engine=mito`

	_, err = db.Exec(createStandaloneTableSQL)
	if err != nil {
		return err
	}

	for i := 0; i < 10; i++ {
		insertDataSQL := fmt.Sprintf("INSERT INTO test_table(n, row_id) VALUES (%d, %d)", i, i)
		_, err = conn.ExecContext(ctx, insertDataSQL)
		if err != nil {
			return err
		}
	}

	return nil
}
```

The test code should work.

### What did you see instead?

The test code failed.

macOS

v0.8.2

_No response_

killme2008:

Close #4244 , cc @zyy17 

* Impl Datafusion `TableProvider#get_column_default` for adapters, which returns the column default value if available.
* Adds `column_defaults` field into `Table`, which is collected from `column_schemas` when building table.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced default values for columns in table creation and MySQL CRUD operations.
  - Added support for column defaults in the schema and table structures.

- **Bug Fixes**
  - Corrected typo in comments to enhance clarity (does not affect functionality).

- **Tests**
  - Added tests for column defaults to ensure proper functionality and coverage.

- **Chores**
  - Updated dependencies in `Cargo.toml` files to include new workspace requirements.


src/datatypes/src/schema.rs
```diff
@@ -43,6 +44,8 @@ pub struct Schema {
     /// Timestamp key column is the column holds the timestamp and forms part of
     /// the primary key. None means there is no timestamp key column.
     timestamp_index: Option<usize>,
+    /// Columns default [`Expr`]
+    column_defaults: HashMap<String, Expr>,
```
killme2008, line 48: Just put this field into `Schema` to avoid building it every time. @evenyag .

tests-integration/tests/sql.rs
```diff
@@ -234,6 +234,23 @@ pub async fn test_mysql_crud(store_type: StorageType) {
         .unwrap();
     assert_eq!(rows.len(), 0);
 
+    // test prepare with default columns
+    sqlx::query("insert into demo(i) values(?)")
```
killme2008, line 238: Before this patch, this test case will fail.

src/datatypes/src/schema.rs
evenyag: We want to keep the schema struct simple and we usually don't have to read `column_defaults`. Also, a schema may not depend on an `Expr`.

Datafusion only computes the default value once and stores it in the projection plan.
https://github.com/apache/datafusion/blob/fe66daaa81738dc60afc297a79897faaaa127724/datafusion/sql/src/statement.rs#L1457-L1480

killme2008: Yes, but we don't have a plan cache yet, so the computation will be repeated every time, it's not necessary at all.

killme2008: But looks like we need to create it every time because the `current_timestamp` must return the current time.

evenyag: I don't think schema is a proper place to cache this. It also adds the whole `datafusion-expr` to the dependencies of a basic data type crate.

We can cache the default value in the `Table`.
https://github.com/GreptimeTeam/greptimedb/blob/03c933c0069994270ba4084523df70867954142a/src/table/src/table.rs#L35-L40
We have a cache for the `Table`.
https://github.com/GreptimeTeam/greptimedb/blob/03c933c0069994270ba4084523df70867954142a/src/catalog/src/kvbackend/table_cache.rs#L29-L30

What do you think? @waynexia

tests-integration/tests/sql.rs
```diff
@@ -234,6 +234,23 @@ pub async fn test_mysql_crud(store_type: StorageType) {
...
+        .execute(&pool)
+        .await
+        .unwrap();
+    let rows = sqlx::query("select * from demo")
+        .fetch_all(&pool)
+        .await
+        .unwrap();
+    assert_eq!(rows.len(), 1);
+
+    for row in rows.iter() {
```
evenyag, line 249:
```suggestion
    for row in rows {
```

src/datatypes/src/schema.rs
killme2008: We should not cache at all, because the `create_default` always returns the current timestamp when it's called. It's not expected.

killme2008: The issue is that `get_column_default` returns a reference like `Option<&Expr>`. Consequently, I need to compute the `column_defaults` each time I create the adapters, even if `get_column_default` won't be used at all.

I don't know if there is a better way @waynexia . Maybe we can proposal the Datafusion to change this function signature.

evenyag: > We should not cache it at all, because the create_default always returns the current timestamp when it's called. If we cache it, the default value will not be changed after that. It's not expected.

Maybe we should document this while creating the adapter.

evenyag (approved):

src/datatypes/src/schema.rs
waynexia: >The issue is that get_column_default returns a reference like Option<&Expr>. Consequently, I need to compute the column_defaults each time I create the adapters, even if get_column_default won't be used at all.

For situations that the default value is a function like `current_timestamp`, maybe we should return that function in `Option<&Expr>` instead of literal value? Like `Expr::ScalarFunction(current_timestamp)` instead of `Expr::Literal(12345)`

killme2008: Fixed.

killme2008:
@evenyag @waynexia PTAL

Remove the `get_column_default` in `DummyTableProvider`, looks like it's unnecessary.

Cache the `column_defaults` in `Table` and use `Expr::ScalarFunction` for `now`, and `current_timestamp` default constraints.

src/table/src/table.rs
```diff
@@ -30,13 +38,29 @@ mod metrics;
...
+    //// The [`Expr`] to call UDF function `now()`.
+    static ref NOW_EXPR: Expr = {
+         let ctx = SessionContext::new();
+
+        let now_udf = ctx.udf("now").expect("now UDF not found");
+
+        Expr::ScalarFunction(ScalarFunction {
+            func: now_udf,
+            args: vec![],
+        })
```
evenyag, line 51:
```suggestion
        let ctx = SessionContext::new();

        let now_udf = ctx.udf("now").expect("now UDF not found");

        Expr::ScalarFunction(ScalarFunction {
            func: now_udf,
            args: vec![],
        })
```

killme2008: It seems odd that `cargo fmt` does not work for this situation.

src/table/src/table.rs
```diff
@@ -30,13 +38,29 @@ mod metrics;
 pub mod numbers;
 pub mod scan;
 
+lazy_static! {
+    //// The [`Expr`] to call UDF function `now()`.
```
waynexia, line 42:
```suggestion
    /// The [`Expr`] to call UDF function `now()`.
```

src/table/src/table.rs
```diff
@@ -113,3 +143,84 @@ impl Table {
...
+        }
+
+        ColumnDefaultConstraint::Function(name)
+            if matches!(
+                name.as_str(),
+                CURRENT_TIMESTAMP | CURRENT_TIMESTAMP_FN | NOW_FN
+            ) =>
+        {
+            Some(Expr::Cast(Cast {
+                expr: Box::new(NOW_EXPR.clone()),
```
waynexia, line 178: Looks like we only need a convenient builder for `now()` UDF but not a lazy static?

waynexia (approved):
Others LGTM

src/table/src/table.rs
killme2008: We just want to reuse it, and try to avoid creating it by `SessionContext` repeatedly.

PR #2861, opened 2023-12-04
fix: replace opendal `PrometheusLayer`

Taylor-lagrange:
I hereby agree to the terms of the [GreptimeDB CLA](https://gist.github.com/xtang/6378857777706e568c1949c7578592cc)

## What's changed and what's your intention?

Use self implementation of `PrometheusMetricsLayer` to replace `opendal`'s `PrometheusLayer` to solve the crash caused by repeated registration of metrics under multiple `object_store`s ( solve by use  `lazy_static`), and add the `opendal_` prefix to the metric.

## Checklist

- [x]  I have written the necessary rustdoc comments.
- [x]  I have added the necessary unit tests and integration tests.

PR #4277, opened 2024-07-04
feat: add path prefix label to obejct storage metrics

sunng87:

This patch add a path label to object store related prometheus metrics, so that we can observe which path has most writes/reads.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced metrics tracking with more granular insights by including path-specific metrics in monitoring.
- **Improvements**
  - Improved function for extracting parent paths from object storage operations, enhancing path management.


evenyag:
The official `PrometheusLayer` already supports this, I think we should switch to it
https://docs.rs/opendal/latest/opendal/layers/struct.PrometheusLayer.html#method.enable_path_label

sunng87:
According to the patch comment, we forked it to avoid a panic with metrics registration. https://github.com/GreptimeTeam/greptimedb/pull/2861

Checked upstream it doesn't seem to be addressed after our fork. Also our solution, using `lazy_static`, is not suitable to upstream because it's not suitable for a library.

evenyag:
> According to the patch comment, we forked it to avoid a panic with metrics registration. #2861
> 
> Checked upstream it doesn't seem to be addressed after our fork. Also our solution, using `lazy_static`, is not suitable to upstream because it's not suitable for a library.

The layer itself could be cloned. So we can use lazy_static to define a global `PrometheusLayer` that uses the default registry.

src/object-store/src/util.rs
```diff
@@ -123,6 +123,24 @@ pub fn normalize_path(path: &str) -> String {
     p
 }
 
+// This logical tries to extract parent path from the object storage operation
+// the function also relies on assumption that the region path is built from
+// pattern `<data|index>/catalog/schema/table_id/....`
+//
+// this implementation is also safe when the path is not constructed in a
+// way that we assumed.
+pub fn extract_parent_path(path: &str) -> String {
```
evenyag, line 132: We can support the path level like the official layer.
https://docs.rs/opendal/latest/opendal/layers/struct.PrometheusLayer.html#method.enable_path_label

evenyag:
> > According to the patch comment, we forked it to avoid a panic with metrics registration. #2861
> > Checked upstream it doesn't seem to be addressed after our fork. Also our solution, using `lazy_static`, is not suitable to upstream because it's not suitable for a library.
> 
> ~The layer itself could be cloned. So we can use lazy_static to init a global `PrometheusLayer` that uses the default registry. Then we can clone the global layer and reuse it.~
> 
> Yes, I realized that the layer will create a `PrometheusMetrics` in each call to the `Layer::layer()` implementation. The `PrometheusMetrics` always registers metrics to the provided registry. However, the registry doesn't allow registering the same metric multiple times.
> 
> So we can't create multiple `Operator` with the same prometheus layer. https://github.com/apache/opendal/blob/174bda53f79123cd114d2409189423a0a4cf6bf3/core/src/layers/prometheus.rs#L193-L210

I created an issue for this. https://github.com/apache/opendal/issues/4854

Maybe @Xuanwo can give us some suggestions.

sunng87:
Actually I found the Prometheus library has a few runtime panics which I think can be avoided by design. For example, if you provide labels doesn't match the declaration, it ends up with a runtime panics, which is very dangerous for some corner cases.

src/object-store/src/util.rs
sunng87: Just changed my mind... I changed this implementation to return full path when there is no more than 3 levels. But I'm not going to make it fully configurable for levels...

Because this scenario is highly associated our path naming pattern. Having a constant number 3 for each setup can be more confusing than this centralized one.

Xuanwo:
> Actually I found the Prometheus library has a few runtime panics which I think can be avoided by design. For example, if you provide labels doesn't match the declaration, it ends up with a runtime panics, which is very dangerous for some corner cases.

Would you like to submit an issue to upstream and link back here?

evenyag:
> Actually I found the Prometheus library has a few runtime panics which I think can be avoided by design. For example, if you provide labels doesn't match the declaration, it ends up with a runtime panics, which is very dangerous for some corner cases.

It provides a method `get_metric_with_label_values()` that returns an error instead.
https://docs.rs/prometheus/latest/prometheus/core/struct.MetricVec.html#method.get_metric_with_label_values

evenyag:
> Thanks! I'm also thinking allow users to set metrics or reuse them in someway. I will give it look.

@Xuanwo I can submit a patch for a workaround by adding a method to set the `metrics`.
```rust
impl PrometheusLayer {
    fn metrics(mut self, metrics: Arc<PrometheusMetrics>) -> Self {
        self.metrics = Some(metrics);
        self
    }
}
```

But this may not be elegant enough.

src/object-store/src/util.rs
evenyag: This implementation should be more memory efficient, could we port it?
https://github.com/apache/opendal/blob/174bda53f79123cd114d2409189423a0a4cf6bf3/core/src/layers/prometheus.rs#L880-L889

WenyXu:
The CI should be fixed in main branch🥲

src/object-store/src/util.rs
sunng87: Yes this one is way better. It's simply substring. I lost my mind when seeking the correct root path.

evenyag (approved):

src/object-store/src/util.rs
```diff
@@ -123,6 +123,20 @@ pub fn normalize_path(path: &str) -> String {
...
+// the function also relies on assumption that the region path is built from
+// pattern `<data|index>/catalog/schema/table_id/....`
+//
+// this implementation tries to extract at most 3 levels of parent path
+pub(crate) fn extract_parent_path(path: &str) -> &str {
+    // split the path into `catalog`, `schema` and others
+    path.char_indices()
+        .filter(|&(_, c)| c == '/')
+        // we get the data/catalog/schema from path, split at the 3rd /
+        .nth(2)
```
zyy17, line 136: The `2` means it should always maintain the layout:

```
<data_home>/<catalog>/<schema>/...
```

And the `region_dir` is `<catalog>/<schema>/...`. Actually, the `<data_home>` is not always one directory, maybe it can be:

```
data/<cluster>/<catalog>/<schema>/...
```

For this scenario, the magic number `2` will not work.

zyy17: The possible elegant solution may:

1. Make the extract function as the field in `PrometheusMetricsLayer` and be initialized by the caller;
2. The caller should always know how to cut the directory since it always knows how to organize the data;

sunng87: This is what I thought to be. But I tested by myself that `<data_home>` seems not included in this path, this is already a subpath when operating the storage. I placed a `dbg!()` call in read and it shows:

```
[src/object-store/src/layers/prometheus.rs:157:9] path = "data/greptime/public/1024/1024_0000000000/manifest/00000000000000000000.json"
```

The `data_home` is `/tmp/greptimedb`. 

Correct me if I missed something.

evenyag: We set the data home as the root path of the OpenDAL operator which is not included in the path. But I'm not sure about the cluster mode.

zyy17: You're right. I didn't read the specific logic of `new_s3_object_store()`. The [root](https://github.com/GreptimeTeam/greptimedb/blob/main/src/datanode/src/store/s3.rs#L35) is already applied for all the operations. It's OK for the original implementation.

zyy17 (approved):

PR #4300, opened 2024-07-05
feat: support `text/plain` format for log ingestion

shuiyisong:

This pr 
 - adds support for `text/plain` and `text/plain; charset=utf-8` on log ingestion
 - refactors pipeline find and delete api using `dataframe` api

now you can use plain text logs to ingest log data, for example:
```
url="http://localhost:4000/v1/events/logs?table=$table_name&pipeline_name=$pipeline_name"
curl -X POST "$url" -H "Content-Type: text/plain" \
 -d '
2024-05-25 20:16:37.217 hello
2024-05-25 20:16:37.218 hello world
'
```

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a new error type, `DataFrame`, for more detailed error reporting related to data frames.
  - Added `TimestampNanosecondValue` support to handle more precise timestamp data in tests.
  - Added plaintext ingestion test for HTTP pipelines.

- **Refactor**
  - Improved data frame preparation and filtering logic in the `PipelineTable` module.
  - Updated content type processing methods in HTTP event handling.

- **Bug Fixes**
  - Enhanced error handling to include new data frame-related errors.

### DIFF
--- a/src/pipeline/src/etl/processor.rs
+++ b/src/pipeline/src/etl/processor.rs
@@ -93,6 +93,19 @@ pub trait Processor: std::fmt::Debug + Send + Sync + 'static {
                 Value::Map(map) => {
                     values.push(self.exec_map(map)?);
                 }
+                Value::String(_) => {
+                    let fields = self.fields();
+                    if fields.len() != 1 {
+                        return Err(format!(
+                            "{} processor: expected fields length 1 when processing line input, but got {}",
+                            self.kind(),
+                            fields.len()
+                        ));
+                    }
+                    let field = fields.first().unwrap();
+
+                    values.push(self.exec_field(&val, field).map(Value::Map)?);
+                }
                 _ if self.ignore_processor_array_failure() => {
                     warn!("expected a map, but got {val}")
                 }
--- a/src/servers/src/http/event.rs
+++ b/src/servers/src/http/event.rs
@@ -254,14 +254,13 @@ pub async fn log_ingester(
             Deserializer::from_str(&payload).into_iter(),
             ignore_errors,
         )?,
-        ct if ct == ContentType::text() || ct == ContentType::text_utf8() => {
-            let a = payload
+        ct if ct == ContentType::text() || ct == ContentType::text_utf8() => Value::Array(
+            payload
                 .lines()
                 .filter(|line| !line.is_empty())
                 .map(|line| json!({ "line": line }))
-                .collect::<Vec<Value>>();
-            Value::Array(a)
-        }
+                .collect::<Vec<Value>>(),
+        ),
         _ => UnsupportedContentTypeSnafu { content_type }.fail()?,
     };
 
--- a/src/servers/src/http/event.rs
+++ b/src/servers/src/http/event.rs
@@ -25,13 +25,12 @@ use axum::http::{Request, StatusCode};
 use axum::response::{IntoResponse, Response};
 use axum::{async_trait, BoxError, Extension, TypedHeader};
 use common_telemetry::{error, warn};
-use mime_guess::mime;
 use pipeline::error::{CastTypeSnafu, PipelineTransformSnafu};
 use pipeline::util::to_pipeline_version;
 use pipeline::{PipelineVersion, Value as PipelineValue};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use serde_json::{Deserializer, Value};
+use serde_json::{json, Deserializer, Value};
 use session::context::QueryContextRef;
 use snafu::{ensure, OptionExt, ResultExt};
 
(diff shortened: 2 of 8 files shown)


src/servers/src/http/event.rs
```diff
@@ -250,13 +249,18 @@ pub async fn log_ingester(
...
             Deserializer::from_str(&payload).into_iter(),
             ignore_errors,
         )?,
-        // add more content type support
+        ct if ct == ContentType::text() || ct == ContentType::text_utf8() => Value::Array(
+            payload
+                .lines()
+                .filter(|line| !line.is_empty())
+                .map(|line| json!({ "line": line }))
+                .collect::<Vec<Value>>(),
```
sunng87, line 262: Can we modify our pipeline the support line based plain text? Here we transform it to json and later in pipeline we immediately extract this line field, which seems a little waste of compute.

If we want to stick with JSON or JavaScript terms, we can use `this` as field name to reference the whole line.

paomian: It may take some time to get the pipeline to support plain text. It seems to me that a headerless csv would be better, and then we could use index to refer to field as well.

shuiyisong: Input of pipeline engine is first converted into [`pipeline value`](https://github.com/GreptimeTeam/greptimedb/blob/849e0b924984727cacfd086a9db5bcb236e17d14/src/pipeline/src/etl/value.rs#L29). Instead of convert plain text into json object, we can directly convert them into `pipeline value`.

@paomian `csv` is yet another format we should support in the future. But plain text is also needed and widely used.

shuiyisong: @sunng87 @paomian PTAL

src/servers/src/http/event.rs
```diff
@@ -271,18 +263,47 @@ pub async fn log_ingester(
...
+                .context(PipelineSnafu)?
+        }
+        ct if ct == ContentType::text() || ct == ContentType::text_utf8() => {
+            let arr = payload
+                .lines()
+                .filter(|line| !line.is_empty())
+                .map(|line| {
+                    let mut map = HashMap::new();
+                    map.insert("line".to_string(), PipelineValue::String(line.to_string()));
+                    PipelineValue::Map(map.into())
```
sunng87, line 289: Can we use some like `PipelineValue::TextLines(line)` here? Adding this map can have same overhead as `json!`

src/servers/src/http/event.rs
sunng87: I'll suggest to add a new variant `TextLines` or just use `String` for pipeline_value. And we can use `this` as field name when referencing this field in pipeline yaml.

src/servers/src/http/event.rs
sunng87: Just noticed we have `PipelineValue::String`. But I think we don't need to use internal value type here. Perhaps we can create another enum say `PipelineDataSource` which has variants like `Json`, `Text`, `CSV` and even `DLT` in future.

src/servers/src/http/event.rs
shuiyisong: By allowing pipeline input as `array(string)`, we can eliminate the map here. Please review this commit https://github.com/GreptimeTeam/greptimedb/pull/4300/commits/db0e1ec8e07f5ea211da4de51b33af02d88f45c0

src/servers/src/http/event.rs
paomian: pipeline does not currently support parsing plain text. This is because pipelines currently need to support keys to select what to parse.

src/pipeline/src/etl/processor.rs
```diff
@@ -93,6 +93,19 @@ pub trait Processor: std::fmt::Debug + Send + Sync + 'static {
                 Value::Map(map) => {
                     values.push(self.exec_map(map)?);
                 }
+                Value::String(_) => {
```
paomian, line 96: I think the design is a bit confusing

src/servers/src/http/event.rs
shuiyisong: I think what you suggest here is a input datasource trait for pipeline engine, that allows custom value extraction using field name in pipeline processing. Since we can have multiple processors along the line, the datatype before and after data is processed should be same and consistent, so that processors can be arranged in arbitrary order. 

Simply add an enum for indicating input datatype at entry level doesn't seem to benefit a lot since we still need to convert data into some compatible format during processor line. Maybe this is a topic for refactoring the whole pipeline datatype system that we can discuss later.

paomian:
I'm in the process of refactoring `pipeline processors`.
I don't think the incoming data needs to be the same type as the processors intermediate state data.
So I'm going to change the signature of the pipeline processors function.
I think this pr can be merged first if @sunng87  agrees. I'll make the changes later.

sunng87:
I agree. We can merge this first and discuss for next step.

sunng87 (approved):

src/servers/src/http/event.rs
shuiyisong: @paomian See discussion here. The `Value::String(_)` pattern is used so that `array(string)` can be accepted as input, that we don't need to convert the plain text into a map first.

paomian (approved):

Issue #4246, opened 2024-07-02
Tracking Issue: support full-text index


zhongzc:
### What problem does the new feature solve?

Support full-text index.

### What does the feature do?

* [x] Creator for full-text index
* [x] Integrate with sst writer
* [x] CREATE statement support
* [ ] push down `matches` and apply the index

killme2008:
So , there is only one remaining task that integrates the index with the `match` function? @zhongzc

zhongzc:
> So , there is only one remaining task that integrates the index with the `match` function? @zhongzc

Yes, I will send the PR soon.

PR #4302, opened 2024-07-05
feat(fulltext_index): integrate full-text indexer with sst writer

zhongzc:

#4246 

* Add config for full-text index
* Integrate full-text index with sst writer
* TODO: tests til search available

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced new configuration options for full-text indexing in Mito engine, including settings for index creation, query application, and memory thresholds.

- **Bug Fixes**
  - Enhanced error handling capabilities for full-text index operations.

- **Documentation**
  - Updated configuration files with new full-text indexing options.

- **Tests**
  - Added new test cases to validate full-text indexing configurations and operations.

- **Chores**
  - Refined internal data structures and methods to support full-text indexing functionality.

### DIFF
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -434,6 +434,27 @@ mem_threshold_on_create = "64M"
 ## Deprecated, use `region_engine.mito.index.aux_path` instead.
 intermediate_path = ""
 
+## The options for full-text index in Mito engine.
+[region_engine.mito.fulltext_index]
+
+## Whether to create the index on flush.
+## - `auto`: automatically
+## - `disable`: never
+create_on_flush = "auto"
+
+## Whether to create the index on compaction.
+## - `auto`: automatically
+## - `disable`: never
+create_on_compaction = "auto"
+
+## Whether to apply the index on query
+## - `auto`: automatically
+## - `disable`: never
+apply_on_query = "auto"
+
+## Memory threshold for index creation.
+mem_threshold_on_create = "64M"
+
 [region_engine.mito.memtable]
 ## Memtable type.
 ## - `time_series`: time-series memtable
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -413,17 +413,17 @@ staging_size = "2GB"
 [region_engine.mito.inverted_index]
 
 ## Whether to create the index on flush.
-## - `auto`: automatically
+## - `auto`: automatically (default)
 ## - `disable`: never
 create_on_flush = "auto"
 
 ## Whether to create the index on compaction.
-## - `auto`: automatically
+## - `auto`: automatically (default)
 ## - `disable`: never
 create_on_compaction = "auto"
 
 ## Whether to apply the index on query
-## - `auto`: automatically
+## - `auto`: automatically (default)
 ## - `disable`: never
 apply_on_query = "auto"
 
@@ -438,17 +438,17 @@ intermediate_path = ""
 [region_engine.mito.fulltext_index]
 
 ## Whether to create the index on flush.
-## - `auto`: automatically
+## - `auto`: automatically (default)
 ## - `disable`: never
 create_on_flush = "auto"
 
 ## Whether to create the index on compaction.
-## - `auto`: automatically
+## - `auto`: automatically (default)
 ## - `disable`: never
 create_on_compaction = "auto"
 
 ## Whether to apply the index on query
-## - `auto`: automatically
+## - `auto`: automatically (default)
 ## - `disable`: never
 apply_on_query = "auto"
 
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -457,7 +457,7 @@ apply_on_query = "auto"
 ## Memory threshold for index creation.
 ## - `auto`: automatically determine the threshold based on the system memory size (default)
 ## - `unlimited`: no memory limit
-## - `[size]`: e.g. `64MB`, fixed memory threshold
+## - `[size]` e.g. `64MB`: fixed memory threshold
 mem_threshold_on_create = "auto"
 
 [region_engine.mito.memtable]
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -428,8 +428,10 @@ create_on_compaction = "auto"
 apply_on_query = "auto"
 
 ## Memory threshold for performing an external sort during index creation.
-## Setting to empty will disable external sorting, forcing all sorting operations to happen in memory.
-mem_threshold_on_create = "64M"
+## - `auto`: automatically determine the threshold based on the system memory size (default)
+## - `unlimited`: no memory limit
+## - `[size]` e.g. `64MB`: fixed memory threshold
+mem_threshold_on_create = "auto"
 
 ## Deprecated, use `region_engine.mito.index.aux_path` instead.
 intermediate_path = ""
@@ -453,7 +455,10 @@ create_on_compaction = "auto"
 apply_on_query = "auto"
 
 ## Memory threshold for index creation.
-mem_threshold_on_create = "64M"
+## - `auto`: automatically determine the threshold based on the system memory size (default)
+## - `unlimited`: no memory limit
+## - `[size]`: e.g. `64MB`, fixed memory threshold
+mem_threshold_on_create = "auto"
 
 [region_engine.mito.memtable]
 ## Memtable type.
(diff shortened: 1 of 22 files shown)


zhongzc:
@coderabbitai pause

fengjiachun (approved):
LGTM

config/datanode.example.toml
```diff
@@ -434,6 +434,27 @@ mem_threshold_on_create = "64M"
 ## Deprecated, use `region_engine.mito.index.aux_path` instead.
 intermediate_path = ""
 
+## The options for full-text index in Mito engine.
+[region_engine.mito.fulltext_index]
+
+## Whether to create the index on flush.
```
killme2008, line 440: Please add the default value of these options to the doc.

config/standalone.example.toml
```diff
@@ -457,6 +457,27 @@ mem_threshold_on_create = "64M"
...
+## - `auto`: automatically
+## - `disable`: never
+create_on_compaction = "auto"
+
+## Whether to apply the index on query
+## - `auto`: automatically
+## - `disable`: never
+apply_on_query = "auto"
+
+## Memory threshold for index creation.
```
killme2008, line 478: Is it too small? Maybe we can set the default value according to the system memory, just like 

https://github.com/GreptimeTeam/greptimedb/blob/09fff24ac4149a22f530a0e25df60c1aab0f5f73/src/mito2/src/config.rs#L145

Then the default value can be `auto`.

killme2008: Other index memory or parallelism configurations can be tweaked this way too.

src/mito2/src/sst/index/fulltext_index/creator.rs
```diff
@@ -0,0 +1,293 @@
...
+        for column in &metadata.column_metadatas {
+            let options =
+                column
+                    .column_schema
+                    .fulltext_options()
+                    .context(FulltextOptionsSnafu {
+                        column_name: &column.column_schema.name,
+                    })?;
+
+            // Relax the type constraint here as many types can be casted to string.
```
killme2008, line 71: I think we can constrain the type to be string, other types such as DateTime or integers, even though they can be cast into string, but the full text on for them is useless.

src/mito2/src/sst/index/fulltext_index/creator.rs
```diff
@@ -0,0 +1,293 @@
...
+        Ok(Self {
+            creators,
+            aborted: false,
+            stats: Statistics::default(),
+        })
+    }
+
+    /// Updates the index with the given batch.
+    pub async fn update(&mut self, batch: &Batch) -> Result<()> {
+        ensure!(!self.aborted, OperateAbortedIndexSnafu);
```
killme2008, line 113: What about adding some metrics? About the timing and data size etc.

src/mito2/src/sst/index/fulltext_index/creator.rs
```diff
@@ -0,0 +1,293 @@
...
+        Ok(())
+    }
+
+    async fn do_finish(&mut self, puffin_writer: &mut SstPuffinWriter) -> Result<()> {
+        let mut guard = self.stats.record_finish();
+
+        let mut written_bytes = 0;
+        for creator in self.creators.values_mut() {
+            written_bytes += creator.finish(puffin_writer).await?;
+        }
```
killme2008, line 193: Prefer the iterator with reduce and sum, but it's ok.

src/mito2/src/sst/index/fulltext_index/creator.rs
```diff
@@ -0,0 +1,293 @@
...
+
+        guard.inc_byte_count(written_bytes);
+        Ok(())
+    }
+
+    async fn do_abort(&mut self) -> Result<()> {
+        let _guard = self.stats.record_cleanup();
+
```
killme2008, line 201: Do we need to add a check?

```rust
if self.aborted {
     return Ok(());
}
```

src/mito2/src/sst/index/fulltext_index/creator.rs
```diff
@@ -0,0 +1,293 @@
...
+                    self.inner
+                        .push_text(text)
+                        .await
+                        .context(FulltextPushTextSnafu)?;
+                }
+            }
+            _ => {
+                for _ in 0..batch.num_rows() {
+                    self.inner
+                        .push_text("")
```
killme2008, line 255: Not quite sure, in such a case,do we still need to index the empty string?

src/mito2/src/sst/index/indexer/finish.rs
```diff
@@ -36,10 +38,43 @@ impl Indexer {
...
+        }
+
         output.file_size = self.do_finish_puffin_writer(writer).await;
         output
     }
 
+    async fn build_puffin_writer(&mut self) -> Option<SstPuffinWriter> {
+        let Some(puffin_manager) = self.puffin_manager.take() else {
+            return None;
+        };
```
killme2008, line 56:
```suggestion
        let puffin_manager) = self.puffin_manager.take();
```

src/mito2/src/sst/index.rs
```diff
@@ -117,31 +143,28 @@ pub(crate) struct IndexerBuilder<'a> {
...
-        if indexer.inverted_indexer.is_none() {
-            indexer.abort().await;
-            return Indexer::default();
-        }
-
-        indexer.puffin_writer = self.build_puffin_writer().await;
-        if indexer.puffin_writer.is_none() {
-            indexer.abort().await;
+        indexer.fulltext_indexer = self.build_fulltext_indexer().await;
+        if indexer.inverted_indexer.is_none() && indexer.fulltext_indexer.is_none() {
```
killme2008, line 163: Does the `indexer` need to abort before returning just like before?

src/mito2/src/sst/index/fulltext_index/creator.rs
zhongzc: Clause with async and &mut ref is annoying. The following code is closed but still doesn't work due to ref of `puffin_writer`.

```rust
let written_bytes = stream::iter(self.creators.values_mut())
    .then(|creator| async move { creator.finish(puffin_writer).await })
    .try_fold(0, |acc, x| async move { Ok(acc + x) })
    .await?;
```

src/mito2/src/sst/index/fulltext_index/creator.rs
zhongzc: In do_update, `stats` will record the metrics and flush when dropped. However, you reminded me that it is possible to distinguish between the metrics of the inverted index and the full-text index.

src/mito2/src/sst/index/fulltext_index/creator.rs
zhongzc: My idea is that specific constraints can be implemented by the frontend, and by relaxing constraints in the mito, many things can function.

src/mito2/src/sst/index/fulltext_index/creator.rs
zhongzc: Agreed, although this `abort` doesn't do much, it reduces the mental burden for the code reader, so I will add it.

src/mito2/src/sst/index.rs
zhongzc: Agreed, although this `abort` doesn't do much, it reduces the mental burden for the code reader, so I will add it.

src/mito2/src/sst/index/fulltext_index/creator.rs
zhongzc: If the `text` column in a certain batch is all `None`, we still need to `push("")` to ensure the row numbers in the index match.

src/mito2/src/sst/index/fulltext_index/creator.rs
zhongzc: https://github.com/GreptimeTeam/greptimedb/blob/21b38311d777cf931e5ffe55653577f5448f30ec/src/mito2/src/sst/index/fulltext_index/creator.rs#L153-L157

`abort` has checked it.

config/standalone.example.toml
zhongzc: good idea!

zhongzc: updated

src/mito2/src/sst/index/fulltext_index/creator.rs
killme2008: @zhongzc Got it, could you add a comment for it? Other people may have the same question.

killme2008 (approved):
LGTM

Issue #4304, opened 2024-07-06
Sqlness test with remote WAL fails


killme2008:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Distributed Cluster

### Minimal reproduce step

https://github.com/GreptimeTeam/greptimedb/actions/runs/9814869931/job/27104613264?pr=4303

There is an error in `global-bg-worker` runtime:

```
thread 'global-bg-worker' has overflowed its stack
fatal runtime error: stack overflow
```

### What did you expect to see?

Run successfully

### What did you see instead?

Error

### What operating system did you use?

Not relevant

### What version of GreptimeDB did you use?

Not relevant

killme2008:
Looks weird, only failed with remote WAL. I run it locally, and all is good.

killme2008:
After reverting `lz4` compression, it works

https://github.com/GreptimeTeam/greptimedb/pull/4303/commits/180dda13faad3713d2597ab9d93235a90ecb9dcd

Looks like it's introduced by #4294

WenyXu:
Another failure: https://github.com/GreptimeTeam/greptimedb/actions/runs/9833718316/job/27144618061

WenyXu:
fixed in #4317

PR #4294, opened 2024-07-05
feat(remote wal): set default compresion to LZ4

WenyXu:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

Switch the default compression algorithm  to LZ4 to reduce WAL size

|               | RaftEngine       | Kafka, NoCompression | Kafka, LZ4       |
|---------------|------------------|----------------------|------------------|
| WAL Size      | 14GiB            | 38GiB                | 15GiB            |
| Replay (mean) | 57s              | 53s                  | 54s              |
| Insertion     | 970-980k point/s | 960-970k point/s     | 960-970k point/s |

## Checklist

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.

## Summary by CodeRabbit

- **New Features**
  - Updated compression method from `NoCompression` to `Lz4` across all configurations for improved data handling efficiency.
  
- **Improvements**
  - Enhanced data compression in TopicManager and various configuration settings, leading to better performance and reduced storage requirements.

PR #4303, opened 2024-07-06
feat: impl show table status

killme2008:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

#4017 #3354 

## What's changed and what's your intention?

* Impl `show table status [IN database] [LIKE name | WHERE ..]`.

```sql
mysql> show table status\G;
*************************** 1. row ***************************
           Name: numbers
         Engine: test_engine
        Version: 11
     Row_format: Fixed
           Rows: 0
 Avg_row_length: 0
    Data_length: 0
Max_data_length: 0
   Index_length: 0
      Data_free: 0
 Auto_increment: 0
    Create_time: 2024-07-05 23:20:47.992000
    Update_time: NULL
     Check_time: NULL
      Collation: NULL
       Checksum: 0
 Create_options:
        Comment: NULL
1 row in set (0.20 sec)
```

* Fixed the `ShowKind` display, the `All` should not display, it's just a placeholder.

## Checklist

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.

## Summary by CodeRabbit

- **New Features**
  - Introduced `SHOW TABLE STATUS` command to retrieve information about table statuses within the database.

- **Enhancements**
  - Improved display logic for various `SHOW` commands to provide more consistent and clearer output.

- **Configuration Changes**
  - Changed default Kafka compression setting from `Lz4` to `NoCompression` for both Datanode and Standalone configurations.

- **Tests**
  - Added new test cases for `SHOW TABLES LIKE`, `SHOW FULL TABLES`, and `SHOW TABLE STATUS` commands.

PR #4312, opened 2024-07-08
revert: set default compresion to LZ4

WenyXu:

Due to #4304
Reverts GreptimeTeam/greptimedb#4294

## Checklist

## Summary by CodeRabbit

- **Configuration Updates**
  - Updated Kafka compression settings from `Lz4` to `NoCompression` for topic management and various configurations.

These changes aim to standardize compression settings across the application, potentially improving consistency and performance.

### DIFF
--- a/src/common/meta/src/wal_options_allocator/kafka/topic_manager.rs
+++ b/src/common/meta/src/wal_options_allocator/kafka/topic_manager.rs
@@ -173,7 +173,7 @@ impl TopicManager {
                     timestamp: chrono::Utc::now(),
                     headers: Default::default(),
                 }],
-                Compression::Lz4,
+                Compression::NoCompression,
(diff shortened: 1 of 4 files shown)


tisonkun:
Somehow we may investigate how to do the compression and avoid stack overflow ... Any ideas why it causes stack overflow?

evenyag:
I think we should find the root cause.

WenyXu:
> Somehow we may investigate how to do the compression and avoid stack overflow ... Any ideas why it causes stack overflow?

> I think we should find the root cause.

Yes, I will investigate it ASAP. However, If we can't find the root cause before releasing the new version, we'd better to revert the PR?

killme2008:
Already done in #4303

PR #4314, opened 2024-07-08
fix: test_fulltext_intm_path

zhongzc:

#4311

normalize before regex

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Improved file path handling and regex matching to be consistent across operating systems, ensuring better compatibility and reliability.


v0y4g3r (approved):
LGTM

src/mito2/src/sst/index/intermediate.rs
```diff
@@ -190,18 +190,16 @@ mod tests {
...
+            (aux_path.replace('\\', "/"), p.replace('\\', "/"))
         } else {
-            let p = fulltext_path.to_string_lossy().to_string();
-            let r = Regex::new(&format!(
-                "{aux_path}/{INTERMEDIATE_DIR}/0/{sst_file_id}/fulltext-0-\\w{{8}}-\\w{{4}}-\\w{{4}}-\\w{{4}}-\\w{{12}}",
+            (aux_path, p)
+        };
+
+        let r = Regex::new(&format!(
+                "{normalized_aux_path}/{INTERMEDIATE_DIR}/0/{sst_file_id}/fulltext-0-\\w{{8}}-\\w{{4}}-\\w{{4}}-\\w{{4}}-\\w{{12}}",
```
evenyag, line 201: Maybe we should use `Path` to iterate it and check dir names? We only need to check the last 4 directories.
https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.iter

We also don't need to worry about special characters in the `normalized_aux_path`.

zhongzc: updated.

evenyag (approved):

PR #4315, opened 2024-07-08
chore: disable TraceLayer on_failure log

paomian:

* Disable the TraceLayer on_failure log because the logged message is not helpful.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Improved error handling with more detailed error messages and conditional error logging.

- **Refactor**
  - Updated internal error management to use a new function for logging errors based on conditions.

### DIFF
--- a/src/servers/src/error.rs
+++ b/src/servers/src/error.rs
@@ -25,7 +25,7 @@ use common_error::define_into_tonic_status;
 use common_error::ext::{BoxedError, ErrorExt};
 use common_error::status_code::StatusCode;
 use common_macro::stack_trace_debug;
-use common_telemetry::{debug, error};
+use common_telemetry::{error, warn};
 use datatypes::prelude::ConcreteDataType;
 use headers::ContentType;
 use query::parser::PromQuery;
@@ -737,8 +737,17 @@ impl From<std::io::Error> for Error {
     }
 }
 
+fn log_error_if_necessary(error: &Error) {
+    if error.status_code().should_log_error() {
+        error!(error; "Failed to handle HTTP request ");
+    } else {
+        warn!(error; "Failed to handle HTTP request ");
+    }
+}
+
 impl IntoResponse for Error {
     fn into_response(self) -> Response {
+        use pipeline::error::Error as PipelineError;
         let error_msg = self.output_msg();
         let status = match self {
             Error::InfluxdbLineProtocol { .. }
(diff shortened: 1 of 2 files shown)


paomian:
@evenyag I didn't end up modifying the logging level here, we can control the printing of debug logs here with RUST_LOG at startup.
https://github.com/GreptimeTeam/greptimedb/blob/59afa70311ead38aa531f908ad3bcb1e93972143/src/servers/src/error.rs#L759

sunng87 (approved):

src/servers/src/error.rs
```diff
@@ -752,15 +766,13 @@ impl IntoResponse for Error {
...
-
-                HttpStatusCode::INTERNAL_SERVER_ERROR
-            }
+            Error::Pipeline { ref source, .. } => match source {
+                PipelineError::CompilePipeline { .. }
+                | PipelineError::InvalidPipelineVersion { .. }
+                | PipelineError::PipelineNotFound { .. } => HttpStatusCode::BAD_REQUEST,
+                _ => log_error_and_return_internal_server_error_http_status(&self),
+            },
+            _ => log_error_and_return_internal_server_error_http_status(&self),
```
sunng87, line 775: I think it's better to avoid side-effect here because in some case we may forget to use this log function and just return the INTERNAL_SERVER_ERROR here.

We can simply call the `log_error_if_necessary` after this pattern matching.

src/servers/src/error.rs
```diff
@@ -752,15 +766,13 @@ impl IntoResponse for Error {
             | Error::InvalidPromRemoteRequest { .. }
             | Error::InvalidQuery { .. }
             | Error::TimePrecision { .. } => HttpStatusCode::BAD_REQUEST,
```
evenyag, line 768: Do we need to log a trace log in this case? cc @sunng87

waynexia (approved):

src/servers/src/error.rs
```diff
@@ -737,8 +737,22 @@ impl From<std::io::Error> for Error {
...
 
+fn log_error_if_necessary(error: &Error) {
+    if error.status_code().should_log_error() {
+        error!(error; "Failed to handle HTTP request: ");
+    } else {
+        debug!("Failed to handle HTTP request: {error:?}");
+    }
+}
+
+fn log_error_and_return_internal_server_error_http_status(error: &Error) -> HttpStatusCode {
```
evenyag, line 748:
```suggestion
fn log_error_and_return_internal_server_error(error: &Error) -> HttpStatusCode {
```

src/servers/src/error.rs
```diff
@@ -737,8 +737,22 @@ impl From<std::io::Error> for Error {
     }
 }
 
+fn log_error_if_necessary(error: &Error) {
+    if error.status_code().should_log_error() {
+        error!(error; "Failed to handle HTTP request: ");
+    } else {
+        debug!("Failed to handle HTTP request: {error:?}");
```
evenyag, line 744: Can we change it to info? We still need to know what the error is.
```suggestion
        info!("Failed to handle HTTP request: {error:?}");
```

src/servers/src/error.rs
evenyag: Agree.

evenyag:
> @evenyag I didn't end up modifying the logging level here, we can control the printing of debug logs here with RUST_LOG at startup.
> 
> https://github.com/GreptimeTeam/greptimedb/blob/59afa70311ead38aa531f908ad3bcb1e93972143/src/servers/src/error.rs#L759

I'm not sure if it is the proper level.

paomian:
> I'm not sure if it is the proper level.

After my testing, adding `RUST_LOG=info,servers::error=debug` before the startup parameter makes the debug level logs printed here. The user can decide whether to turn it on or off as needed.

```
2024-07-08T08:27:09.389642Z DEBUG servers::error: Failed to handle HTTP request: 0: Pipeline management api error, at src/frontend/src/instance/log_handler.rs:68:14
1: Failed to parse pipeline: transform cannot be empty, at src/pipeline/src/manager/table.rs:205:61
```

src/servers/src/error.rs
sunng87: I think it's ok to ignore them from log because we will return the error message to user for this type. For internal errors, we only give user a error code and `Internal Error` message, so it's necessary to log detail by our own.

sunng87 (approved):

evenyag (approved):

PR #4323, opened 2024-07-09
feat: flownode use `Inserter` to write to database

discord9:

flownode now use `Inserter`/`Deleter` to write to database, hence removing need to specify frontend addr and such

- add `RemoteFrontendInvoker` and it's builder to build a frontend invoker
- call inovker to set it for flownode

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced error handling and context management in standalone mode with new invoker setup.
  - Improved frontend invocation setup through `build_frontend_invoker`.

- **Refactor**
  - Reorganized frontend server connection and initialization processes.
  - Updated several object instantiations to use cloned parameters for improved stability.

- **Bug Fixes**
  - Addressed issues in the `FlowWorkerManager` to handle shutdown signals more effectively.

- **Documentation**
  - Revised method and struct definitions to clarify new frontend invoker and error handling procedures.


src/flow/src/server.rs
```diff
@@ -303,3 +323,116 @@ impl FlownodeBuilder {
         Ok(man)
     }
 }
+
+pub struct RemoteFrondendInvoker {
```
sunng87, line 327: typo, and also this is `Remote` here is not quite accurate.

fengjiachun: +1 Using `FrontendInvoker` is OK

src/flow/src/server.rs
```diff
@@ -303,3 +323,116 @@ impl FlownodeBuilder {
...
+            table_route_cache,
+        ));
+
+        let invoker = RemoteFrondendInvoker::new(inserter, deleter, statement_executor);
+        Ok(invoker)
+    }
+}
+
+#[async_trait::async_trait]
+impl FrontendInvoker for RemoteFrondendInvoker {
```
fengjiachun, line 414: Seems we do not need FrontendInvoker any more, can we just use `inserter` and `deleter`?

src/flow/src/server.rs
fengjiachun: I think we don't need `FrontendInvoke` anymore; using just `inserter` and `deleter` is sufficient. Another layer of abstraction isn't necessary.

fengjiachun: Or `FrontendInvoker` can be a struct:

```
struct FrontendInvoker {
    inserter: InserterRef,
    deleter: DeleterRef,
}
```

src/flow/src/server.rs
discord9: > Seems we do not need FrontendInvoker any more, can we just use `inserter` and `deleter`?

Refactor to use a `FrontendInvoker` for clarity of code， since invoker need many things to be constructed

src/flow/src/adapter.rs
```diff
@@ -120,10 +119,10 @@ pub struct FlowWorkerManager {
     /// which is `!Send` so a handle is used
     pub worker_handles: Vec<Mutex<WorkerHandle>>,
     /// The query engine that will be used to parse the query and convert it to a dataflow plan
-    query_engine: Arc<dyn QueryEngine>,
+    pub query_engine: Arc<dyn QueryEngine>,
     /// Getting table name and table schema from table info manager
     table_info_source: TableSource,
-    frontend_invoker: RwLock<Option<Box<dyn FrontendInvoker + Send + Sync>>>,
+    frontend_invoker: RwLock<Option<FrontendInvoker>>,
```
fengjiachun, line 125: We should be able to avoid using locks. When constructing FlowWorkerManager, passing in frontend_invoker should suffice.

discord9: There is a depend when building standalone:
```
flownode -> node_manager -> ddl_task_executor -> FrontendInvoker
```
if not use mutex, then there will be:
```
FrontendInvoker -> flownode
```
which will be cyclic

> We should be able to avoid using locks. When constructing FlowWorkerManager, passing in frontend_invoker should suffice.

fengjiachun (approved):
Rest LGTM

WenyXu (approved):

Issue #4339, opened 2024-07-10
Adds information_schema.views


killme2008:
### What type of enhancement is this?

API improvement, Configuration, User experience

### What does the enhancement do?

Part of #2931 #3729 

Implements `information_schema.views` to provide information about views in databases.

Refer to https://dev.mysql.com/doc/refman/8.4/en/information-schema-views-table.html

### Implementation challenges

The implement detail notes:

1. Refer to `information_schema.tables` for example to implement the `views`:
    - https://github.com/GreptimeTeam/greptimedb/blob/main/src/catalog/src/information_schema/tables.rs
2. Register it to `information_schema`:
    - https://github.com/GreptimeTeam/greptimedb/blob/da0c8402619134d455242bd5c16e16a3b3955994/src/catalog/src/information_schema.rs#L211
3. Adds unit tests and sqlness tests https://github.com/GreptimeTeam/greptimedb/tree/main/tests/cases/standalone/common/information_schema

lyang24:
Hi, I would like to work on this. I assume the view info could be retrieved by `ViewInfoManager`, is there a collection of viewIDs that i can iterate to get view info?

killme2008:
@lyang24 Great, thanks a lot.

There is a field `definition` which is the view definition:

https://github.com/GreptimeTeam/greptimedb/blob/da0c8402619134d455242bd5c16e16a3b3955994/src/common/meta/src/key/view_info.rs#L94

Retrieving the `view_info` by `ViewInfoCache`:

https://github.com/GreptimeTeam/greptimedb/blob/da0c8402619134d455242bd5c16e16a3b3955994/src/common/meta/src/cache/table/view_info.rs#L30

`ViewInfoCache` can be retrieved by:

https://github.com/GreptimeTeam/greptimedb/blob/da0c8402619134d455242bd5c16e16a3b3955994/src/catalog/src/kvbackend/manager.rs#L103

killme2008:
One way to obtain the `ViewInfoCache` is by incorporating a function similar to the following:

https://github.com/GreptimeTeam/greptimedb/blob/da0c8402619134d455242bd5c16e16a3b3955994/src/catalog/src/information_schema/utils.rs#L57

During runtime, the `catalog_manager` is consistently `KvBackendCatalogManager`.

killme2008:
Another example to retrieve view info in `catalog` crate:

[...]

PR #4342, opened 2024-07-10
feat: adding information_schema.views table

lyang24:

https://github.com/GreptimeTeam/greptimedb/issues/4339
adding information_schema.views table to display views.

This content is generated by iterate table information from catalog manager and with filter `table_type = VIEW`, and retrieve view definition from table id. A very simple sqlness test is added as well.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced support for `VIEWS` in the information schema, allowing users to manage and query views within the database.
  - Added ability to create, query, and drop views using SQL statements.

- **Enhancements**
  - Expanded `SHOW` queries to include the `views` table and relevant metadata.
  - Updated information schema tables with new columns to support view-related metadata.

- **Tests**
  - Added new test scenarios to validate the creation, querying, and dropping of views.


src/catalog/src/information_schema/views.rs
```diff
@@ -0,0 +1,288 @@
+// Copyright 2023 Greptime Team
```
lyang24, line 1: update `greptimedb/licenserc.toml` to 2024?

killme2008: No, just keep it, we'll update it in batch.

src/catalog/src/information_schema/views.rs
```diff
@@ -0,0 +1,288 @@
...
+pub const TABLE_NAME: &str = "table_name";
+pub const VIEW_DEFINITION: &str = "view_definition";
+pub const CHECK_OPTION: &str = "check_option";
+pub const IS_UPDATABLE: &str = "is_updateble";
+pub const DEFINER: &str = "definer";
+pub const SECURITY_TYPE: &str = "security_type";
+pub const CHARACTER_SET_CLIENT: &str = "character_set_client";
+pub const COLLATION_CONNECTION: &str = "collation_connection";
+
+pub(super) struct InformationSchemaViews {
```
killme2008, line 56: Adds a description to describe this table, like:
```
/// The `information_schema.views` to provides information about views in databases.
```

src/catalog/src/information_schema/views.rs
```diff
@@ -0,0 +1,288 @@
...
+            definer: StringVectorBuilder::with_capacity(INIT_CAPACITY),
+            security_type: StringVectorBuilder::with_capacity(INIT_CAPACITY),
+            character_set_client: StringVectorBuilder::with_capacity(INIT_CAPACITY),
+            collation_connection: StringVectorBuilder::with_capacity(INIT_CAPACITY),
+        }
+    }
+
+    /// Construct the `information_schema.views` virtual table
+    async fn make_views(&mut self, request: Option<ScanRequest>) -> Result<RecordBatch> {
+        let catalog_name = self.catalog_name.clone();
```
killme2008, line 184: Avoid unnecessary cloning; use `catalog_name` instead of `&catalog_name` since it's always referenced below. Clone only when essential:

```suggestion
        let catalog_name = &self.catalog_name;
```

src/catalog/src/information_schema/views.rs
```diff
@@ -0,0 +1,288 @@
...
+            let mut stream = catalog_manager.tables(&catalog_name, &schema_name);
+
+            while let Some(table) = stream.try_next().await? {
+                let table_info = table.table_info();
+                if table_info.table_type == TableType::View {
+                    let backend_catalog_manager = catalog_manager
+                        .as_any()
+                        .downcast_ref::<KvBackendCatalogManager>()
+                        .context(CastManagerSnafu)?;
+                    let view_info = backend_catalog_manager
```
killme2008, line 201: What about extracting the `view_info_cache` out of the loop:
```rust
let view_info_cache = backend_catalog_manager
    .view_info_cache()?;

while let Some(table) = stream.try_next().await? {
      let view_info = view_info_cache.get(table_info.ident.table_id)
                              .await
                              ....
}
```

tests/cases/standalone/common/information_schema/views.sql
```diff
@@ -0,0 +1,13 @@
...
+create table t (ts timestamp time index, val int);
+
+create view myview as select * from t where val > 5;
+
+select table_catalog, table_schema, table_name, view_definition from information_schema.views;
+
+-- FIXME(dennis): DROP VIEW not supported yet--
+-- DROP VIEW myview --
+
+drop table t;
```
killme2008, line 13: Needs a blank line at the end of file.

tests/cases/standalone/common/information_schema/views.sql
```diff
@@ -0,0 +1,13 @@
+--- test information_schema.views ----
+USE public;
```
killme2008, line 2: Drop view is implemented #4231, but it is not ready to merge.
So I think you can create an independent database for this test case to avoid name confusing, just like 

https://github.com/GreptimeTeam/greptimedb/blob/da1ea253ba49fdaf0d7c9e7dd2bee46ab15b8115/tests/cases/standalone/common/view/create.sql#L5

Just remember to drop it at lat and use `public` back.

tests/cases/standalone/common/information_schema/views.sql
```diff
@@ -0,0 +1,13 @@
...
+
+create table t (ts timestamp time index, val int);
+
+create view myview as select * from t where val > 5;
```
killme2008, line 6: Creating at least two views ensures that everything is correct.

src/catalog/src/information_schema/views.rs
lyang24: I think the clone here is necessary because its borrowed as a mutable reference with add_view() call

src/catalog/src/information_schema/views.rs
lyang24: good catch fixed

tests/cases/standalone/common/information_schema/views.sql
lyang24: fixed

killme2008 (approved):
LGTM

killme2008:
@fengjiachun @sunng87 Please take a look.

fengjiachun (approved):
LGTM

fengjiachun:
@lyang24 Thank you!

PR #4389, opened 2024-07-18
refactor: add `&mut Plugins` argument in plugins setup api and remove unnecessary mut

zyy17:

**NOTE**: The PR will **break** the `setup_<role>_plugins()` API.

1. Add the `&mut Plugins` argument in the plugins setup API instead of creating `Plugins` inside since we need to add frontend and datanode plugins in standalone mode in one `Plugins`;

2. Remove unnecessary `mut` for `*Options`
   
   I think it's not a good practice to implicitly modify the `*Options` in the plugin setup API. We can modify the `*Options` from layered configuration, for example, environment variables, config file, etc;

## Checklist

### DIFF
--- a/src/plugins/src/datanode.rs
+++ b/src/plugins/src/datanode.rs
@@ -16,8 +16,13 @@ use common_base::Plugins;
 use datanode::config::DatanodeOptions;
 use datanode::error::Result;
 
-pub async fn setup_datanode_plugins(_opts: &mut DatanodeOptions) -> Result<Plugins> {
-    Ok(Plugins::new())
+#[allow(unused_variables)]
+#[allow(unused_mut)]
+pub async fn setup_datanode_plugins(
+    plugins: &mut Plugins,
+    dn_opts: &DatanodeOptions,
+) -> Result<()> {
+    Ok(())
 }
 
 pub async fn start_datanode_plugins(_plugins: Plugins) -> Result<()> {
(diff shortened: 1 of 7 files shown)


src/plugins/src/datanode.rs
```diff
@@ -16,10 +16,16 @@ use common_base::Plugins;
 use datanode::config::DatanodeOptions;
 use datanode::error::Result;
 
-pub async fn setup_datanode_plugins(_opts: &mut DatanodeOptions) -> Result<Plugins> {
-    Ok(Plugins::new())
+#[allow(unused_variables)]
+#[allow(unused_mut)]
+pub async fn setup_datanode_plugins(
+    plugins: &mut Plugins,
+    dn_opts: &DatanodeOptions,
```
shuiyisong, line 23: Can we use `_plugins` to avoid `#[allow(unused_variables)]`, or it's a deliberate choice?

shuiyisong (approved):
Rest LGTM

sunng87 (approved):
LGTM if we use `_` for unused variables

sunng87:
by the way, this doesn't have to be a breaking change. Plugin is not our public API.

src/plugins/src/datanode.rs
zyy17: Actually, I did it on purpose because we don't change the signature when we override the plugins crate.

PR #4390, opened 2024-07-18
perf: Optimizing pipeline performance

paomian:

**There are over a thousand lines of code that are test data and benchmarks can be ignored.**
`src/pipeline/benches/data.log` is test data. and `src/pipeline/benches/processor.rs` is benchmark.

* Optimizing the performance of type conversions
* Use Arc<String> clone instead of String clone
* Passing Intermediate State Using Variable References
* add an bench
* Optimize hash get overconsumption by changing the pipeline's intermediate state to vec
* Both `timeindex` and `timestamp` are supported to specify the timeindex in the transform.

## Original pipeline workflow
pipeline_original

## Current pipeline workflow

pipeline_current

In the original method, there is a lot of data allocation and drop, and a lot of hash get operations. By converting the data into a vector, hash get operations are avoided. And the same batch of data share a common intermediate state. After testing, the performance of the same pipeline processing the same data is improved by more than 30%.

## Checklist


src/servers/src/http/event.rs
```diff
@@ -311,20 +313,41 @@ async fn ingest_logs_inner(
...
+                    .observe(transform_timer.elapsed().as_secs_f64());
+                PipelineTransformSnafu { reason }.build()
+            })
+            .context(PipelineSnafu)?;
+        results.push(r);
+        pipeline.reset_intermediate_state(&mut intermediate_state);
+    }
+
+    METRIC_HTTP_LOGS_TRANSFORM_ELAPSED
+        .with_label_values(&[db.as_str(), METRIC_FAILURE_VALUE])
```
shuiyisong, line 344:
```suggestion
        .with_label_values(&[db.as_str(), METRIC_SUCCESS_VALUE])
```

src/pipeline/src/etl/transform/index.rs
```diff
@@ -13,20 +13,22 @@
 // limitations under the License.
 
 const INDEX_TIMESTAMP: &str = "timestamp";
+const INDEX_TIMEINDEX: &str = "timeindex";
```
shuiyisong, line 16: @killme2008 Would you prefer `index: timeindex` or just `index: time`?

src/servers/src/http/event.rs
```diff
@@ -311,20 +313,41 @@ async fn ingest_logs_inner(
...
-                .observe(transform_timer.elapsed().as_secs_f64());
-        })
-        .map_err(|reason| {
-            METRIC_HTTP_LOGS_TRANSFORM_ELAPSED
-                .with_label_values(&[db.as_str(), METRIC_FAILURE_VALUE])
-                .observe(transform_timer.elapsed().as_secs_f64());
-            PipelineTransformSnafu { reason }.build()
-        })
-        .context(PipelineSnafu)?;
+    let mut intermediate_state = pipeline.init_intermediate_state().clone();
```
zhongzc, line 316:
```suggestion
    let mut intermediate_state = pipeline.init_intermediate_state();
```

src/pipeline/src/etl/processor.rs
```diff
@@ -55,110 +54,254 @@ const SEPARATOR_NAME: &str = "separator";
...
+    Gsub(GsubProcessor),
+    Join(JoinProcessor),
+    Letter(LetterProcessor),
+    Regex(RegexProcessor),
+    UrlEncoding(UrlEncodingProcessor),
+}
+
+impl Clone for ProcessorKind {
+    fn clone(&self) -> Self {
+        todo!()
```
shuiyisong, line 112: What does this `todo` do, is this a left-out to be deleted?

src/pipeline/src/etl.rs
```diff
@@ -97,13 +220,110 @@ impl<T> Pipeline<T>
...
+
+    pub fn exec_mut(&self, val: &mut Vec<value::Value>) -> Result<T::VecOutput, String> {
+        for processor in self.processors.iter() {
+            processor.exec_mut(val)?;
+        }
+
+        self.transformer.transform_mut(val)
+    }
+
+    pub fn preprepase(&self, val: serde_json::Value, result: &mut [Value]) -> Result<(), String> {
```
zhongzc, line 263: Too much `pre`? Is `prepare` enough?

src/pipeline/src/etl.rs
```diff
@@ -97,13 +220,110 @@ impl<T> Pipeline<T>
...
         for processor in self.processors.iter() {
-            val = processor.exec(val)?;
+            v = processor.exec_map(v)?;
         }
+        Ok(v)
+    }
+
+    pub fn exec(&self, mut val: value::Value) -> Result<T::Output, String> {
+        let result = match val {
+            value::Value::Map(ref mut map) => {
```
zhongzc, line 233: Non-blocking: style

```suggestion
        let result = match &mut val {
            value::Value::Map(map) => {
```

src/pipeline/src/etl.rs
zhongzc: Add some unittests and doc comments to let everyone see what the expected behavior is?

shuiyisong (commented):
Please add some comments for the purpose of the newly added fields and methods, like 
```
processors.output_keys();
processors.required_keys();
processors.required_original_keys();
transforms.required_keys();
```

src/pipeline/src/etl/processor.rs
```diff
@@ -55,110 +54,254 @@ const SEPARATOR_NAME: &str = "separator";
...
+    }
+}
+
+impl Processor for ProcessorKind {
```
sunng87, line 116: there is a `enum_dispatch` macro can simply this

src/pipeline/src/etl.rs
paomian: Add a simple test to show how it works.

src/pipeline/src/etl/field.rs
```diff
@@ -66,12 +69,42 @@ impl std::ops::Deref for Fields {
...
+        InputFieldInfo {
+            name: field.into(),
+            index: 0,
+        }
+    }
+}
+
 #[derive(Debug, Default, Clone)]
 pub struct Field {
-    pub field: String,
```
sunng87, line 71: We need more comments for this struct.

src/pipeline/src/etl/processor/cmcd.rs
```diff
@@ -205,15 +260,46 @@ impl crate::etl::processor::Processor for CMCDProcessor {
             )),
         }
     }
+
+    fn exec_mut(&self, val: &mut Vec<Value>) -> Result<(), String> {
+        for field in self.fields.iter() {
+            match val.get(field.input_field.index) {
+                Some(Value::String(v)) => {
+                    // TODO(qtang): Let this method use the intermediate state collection directly.
+                    let map = self.process_field(v, field)?;
```
sunng87, line 269: next step will be refactoring `process_field` to support in-place modification

src/pipeline/src/etl.rs
```diff
@@ -32,6 +36,60 @@ pub enum Content {
...
+                    .position(|r| *r == *k)
+                    .ok_or(format!(
+                    "output field {k} is not found in intermediate keys: {final_intermediate_keys:?} when set processor keys index"
+                ))?;
+                *v = index;
+            }
+        }
+    }
+    Ok(())
+}
```
sunng87, line 65: Do you think it can be easier to understand if we maintain an `input_field_key_index_mapping` and `output_field_key_index_mapping` in `Processor`?

src/pipeline/src/etl.rs
```diff
@@ -97,23 +223,178 @@ impl<T> Pipeline<T>
...
+        "#;
+        let input_value: serde_json::Value = serde_json::from_str(input_value_str).unwrap();
+
+        let pipeline_yaml = r#"
+---
+description: Pipeline for Apache Tomcat
+
+processors:
+  - csv:
+      field: my_field, my_field,field1, field2
```
zhongzc, line 362: What a `field`... I feel the same way as you. @paomian

paomian: In the next step. We can add a new yaml field `target_fields` to specify the target_fields field.

src/pipeline/src/etl.rs
paomian: Yes, my plan is to do it in the next step.

killme2008:
I think we must deliver this PR in v0.9.1, so please take a look @sunng87 @shuiyisong @zhongzc

src/pipeline/benches/processor.rs
```diff
@@ -0,0 +1,262 @@
...
+    group.bench_function("processor map", |b| {
+        b.iter(|| processor_map(black_box(&pipeline), black_box(input_value.clone())))
+    });
+    group.bench_function("processor mut", |b| {
+        b.iter(|| processor_mut(black_box(&pipeline), black_box(input_value.clone())))
+    });
+    group.finish();
+}
+
+criterion_group!(benches, criterion_benchmark);
```
killme2008, line 261: It's better to add some descriptions about what you benchmark.

src/pipeline/src/etl/transform/index.rs
killme2008: `index: time` looks good.

src/pipeline/src/etl.rs
```diff
@@ -32,6 +36,60 @@ pub enum Content {
     Yaml(String),
 }
 
+fn set_processor_keys_index(
```
killme2008, line 39: Adds some comments to these functions. How the key and index are mapping like that?

src/pipeline/src/etl/processor.rs
```diff
@@ -149,16 +156,47 @@ impl TryFrom<&Vec<yaml_rust::Yaml>> for Processors {
...
+
+            for key in &processor_required_keys {
+                if !all_output_keys.contains(key) {
+                    all_required_original_keys.insert(key.clone());
+                }
+            }
+
+            all_required_keys.extend(processor_required_keys);
+
+            let processor_output_keys = processor.output_keys().into_iter();
```
killme2008, line 180: Do we need to update `all_out_keys` before line 174? Which is checking the key is not in `all_output_keys` before adding it to `all_required_original_keys`.

src/pipeline/src/etl/processor.rs
```diff
@@ -53,94 +53,101 @@ const SEPARATOR_NAME: &str = "separator";
...
+    Letter(LetterProcessor),
+    Regex(RegexProcessor),
+    UrlEncoding(UrlEncodingProcessor),
+}
 
-        Ok(Value::Array(Array { values }))
-    }
+#[derive(Debug, Default)]
+pub struct Processors {
+    pub processors: Vec<ProcessorKind>,
```
killme2008, line 117: Better to give comments for all these fields

src/pipeline/src/etl.rs
```diff
@@ -56,24 +114,92 @@ where
...
+
+            for key_name in ordered_intermediate_keys.iter() {
+                if required_keys.contains(key_name) {
+                    final_intermediate_keys.push(key_name.clone());
+                } else {
+                    intermediate_keys_exclude_original.push(key_name.clone());
+                }
+            }
+
+            final_intermediate_keys.extend(intermediate_keys_exclude_original);
```
killme2008, line 169: So this logic just wants to split the ordered_intermediate_keys by checking if it is in `required_keys`? If so, maybe use a group_by and concat is more clear.

src/pipeline/src/etl/processor.rs
```diff
@@ -53,94 +53,101 @@ const SEPARATOR_NAME: &str = "separator";
...
 
-    /// default behavior does nothing and returns the input value
-    fn exec_field(&self, val: &Value, field: &Field) -> Result<Map, String> {
-        Ok(Map::one(field.get_field(), val.clone()))
-    }
+    /// processor all output keys
+    /// if a processor has multiple output keys, it should return all of them
+    fn output_keys(&self) -> HashSet<String>;
+
+    fn exec_field(&self, val: &Value, field: &Field) -> Result<Map, String>;
```
killme2008, line 71: Adds comments for all functions in the trait.

src/pipeline/src/etl/processor/letter.rs
```diff
@@ -144,6 +155,43 @@ impl Processor for LetterProcessor {
             )),
         }
     }
+
+    fn exec_mut(&self, val: &mut Vec<Value>) -> Result<(), String> {
+        for field in self.fields.iter() {
+            let index = field.input_field.index;
+            match val.get(index) {
+                Some(Value::String(s)) => {
+                    // TODO(qtang): Let this method use the intermediate state collection directly.
```
killme2008, line 164: Why not do it right now?

src/pipeline/src/etl/processor/regex.rs
```diff
@@ -203,6 +237,48 @@ impl Processor for RegexProcessor {
             )),
         }
     }
+
+    fn exec_mut(&self, val: &mut Vec<Value>) -> Result<(), String> {
+        for field in self.fields.iter() {
```
killme2008, line 242: The code block appears to be similar across different processors. Perhaps we can utilize a macro to eliminate the duplications.

src/pipeline/src/etl/transform/index.rs
killme2008: Can we deprecate the `time` type too? Use `timestamp` type instead.

src/pipeline/tests/pipeline.rs
```diff
@@ -488,7 +488,7 @@ transform:
     type: string
   - field: ts
     type: time
```
killme2008, line 490: Let's change the type too.

src/servers/src/http/event.rs
```diff
@@ -188,43 +188,40 @@ pub async fn delete_pipeline(
...
-                    warn!("invalid item in array: {:?}", item);
-                    Ok(Value::Array(acc_array))
                 }
+                Ok(acc_array)
+            } else if !ignore_error {
+                item.map(|x| vec![x]).context(ParseJsonSnafu)
+            } else {
+                warn!("invalid item in array: {:?}", item);
+                Ok(acc_array)
             }
```
zhongzc, line 223: use match?

```rust
match item {
  Ok(_) => { todo!() }
  Err(_) if !ignore_error => { todo!() }
  _ => { todo!() }
}

File: src/pipeline/benches/processor.rs [hunk 22]
```diff
@@ -0,0 +1,262 @@
...
+use pipeline::{parse, Array, Content, GreptimeTransformer, Pipeline, Value as PipelineValue};
+use serde_json::{Deserializer, Value};
+
+fn processor_map(
+    pipeline: &Pipeline<GreptimeTransformer>,
+    input_values: Vec<Value>,
+) -> impl IntoIterator<Item = greptime_proto::v1::Rows> {
+    let pipeline_data = input_values
+        .into_iter()
+        .map(|v| PipelineValue::try_from(v).unwrap())
```
zhongzc, line 25: Need to describe why `unwrap` is safe here

src/pipeline/benches/processor.rs
```diff
@@ -0,0 +1,262 @@
...
+    let mut result = Vec::with_capacity(input_values.len());
+
+    for v in input_values {
+        pipeline.prepare(v, &mut payload)?;
+        let r = pipeline.exec_mut(&mut payload)?;
+        result.push(r);
+        pipeline.reset_intermediate_state(&mut payload);
+    }
+
+    Ok::<Vec<greptime_proto::v1::Row>, String>(result)
```
zhongzc, line 47:
```suggestion
    Some(result)
```

src/pipeline/benches/processor.rs
zhongzc: Ignore it, this is a bench, let's relax it

src/pipeline/src/etl/processor.rs
paomian: There is no need, we only care about the processor `all_out_keys` before the current processor.

src/pipeline/src/etl.rs
paomian: Ensure that all user-supplied data required by the processor (except for process output data) fields are at the top of the list and keep the fields in order.

src/pipeline/src/etl/processor/letter.rs
paomian: This is because the original design remodeling required extensive modifications. The data structure of each processor needed to be modified. All things considered, it is ready to be implemented in the next phase.

src/pipeline/src/etl.rs
```diff
@@ -97,23 +223,178 @@ impl<T> Pipeline<T>
 where
     T: Transformer,
 {
-    pub fn exec(&self, val: value::Value) -> Result<T::Output, String> {
-        let mut val = val;
+    fn exec_map<'a>(&self, map: &'a mut Map) -> Result<&'a mut Map, String> {
```
sunng87, line 226: this function already accepts aa mutable reference to map, so it doesn't have to return it. Just use `Result<(), String>` to make it clear.

src/pipeline/src/etl/processor/regex.rs
sunng87: My suggestion is to avoid macro when it's possible. Macro can make the code difficult to understand.

src/pipeline/src/etl.rs
```diff
@@ -97,23 +223,178 @@ impl<T> Pipeline<T>
...
+        self.transformer.transform_mut(val)
+    }
+
+    pub fn prepare(&self, val: serde_json::Value, result: &mut [Value]) -> Result<(), String> {
+        match val {
+            serde_json::Value::Object(map) => {
+                let mut index = 0;
+
+                // because of the key in the required_keys is ordered
+                for (payload_key, payload_value) in map.into_iter() {
```
zhongzc, line 272: Is the iterated `payload_key` in ascending order?

src/pipeline/src/etl.rs
zhongzc, line 169: looks equivalent to

```suggestion
            let final_intermediate_keys = ordered_intermediate_keys.clone();
```

paomian: Here is to keep a specific order.

src/pipeline/src/etl.rs
paomian: Yep, `map` is a `BtreeMap`.

zhongzc: If I understand correctly, is this correct:
```rust
let mut from = 0;
for (payload_key, payload_value) in map {
    if from >= self.intermediate_keys.len() {
        break;
    }

    if let Some(pos) = self.intermediate_keys[from..]
        .iter()
        .position(|k| k == &payload_key)
    {
        result[pos] = payload_value.try_into()?;
        from = pos + 1;
    }
}
```

paomian: Yes. But this map is from user. Some keys in `intermediate_keys` is processor's output key. so we select all keys from map which is from user.

paomian:
Incompatible changes added
1. Delete `date`, `epoch` processor
2. add `timestamp` process, which is a combination of date and epoch. It can parse time strings as timestamps or treat numbers as timestamps.
3. remove `time` and `epoch` from transfrom and replace them with `timestmap`.
4. add `time` index type

src/pipeline/src/etl/processor/regex.rs
killme2008: However, these function implementations appear identical. If you want to change one place, you have to change almost everywhere. Maybe not macro but some way to reduce the duplicated code.

src/pipeline/src/etl/processor/dissect.rs
```diff
@@ -543,6 +542,27 @@ impl DissectProcessor {
...
+                                .is_some_and(|x| *x == StartModifier::NamedSkip)
+                        {
+                            fields
+                                .output_fields_index_mapping
+                                .insert(name.to_string(), 0);
+                        }
+                    }
+                }
+            }
+        }
```
fengjiachun, line 564: Nonblocking, code style: Too many nested layers, can it be modified like this:
```suggestion
        let output_keys = self
            .patterns
            .iter()
            .flat_map(|pattern| pattern.iter())
            .filter_map(|part| {
                if let Part::Name(name) = part {
                    if !name.is_empty()
                        && !name
                            .start_modifier
                            .as_ref()
                            .is_some_and(|x| *x == StartModifier::NamedSkip)
                    {
                        Some(name.to_string())
                    } else {
                        None
                    }
                } else {
                    None
                }
            })
            .collect::<HashSet<_>>();

        for field in self.fields.iter_mut() {
            for key in &output_keys {
                field.output_fields_index_mapping.insert(key.clone(), 0);
            }
        }
```

In addition, this modification also avoids multiple repeated iterations of the `pattern`.

src/pipeline/src/etl.rs
zhongzc: Can you simplify the implementation according to my suggestion?

killme2008:
What's the progress of this PR? @shuiyisong @paomian

src/pipeline/src/etl.rs
shuiyisong: `self.intermediate_keys` contains the needed user input keys(which is `payload_key` here) and its index in vec, and `intermediate_keys` is also corresponded to `intermediate_state`(which is `result: &mut [Value]` here).

How about getting index using `self.intermediate_keys` and set `payload_value` to result using the index?

BTW: `self.intermediate_keys` can just be `BTreeMap<String, usize>` to indicate mapping between key and index, and the index can be used to set and retrieve value from `intermediate_state`

paomian:
> What's the progress of this PR? @shuiyisong @paomian

There are a few details that need to be changed. Prepare for careful testing.

src/pipeline/src/etl.rs
paomian: @zhongzc after test. `self.intermediate_keys[from..].iter().position(|k| k == &payload_key)` get the pos is start from `from` not start from intermediate_keys first item. so it can't be rewrite like your suggestion.

paomian: @shuiyisong `required_keys` is keys which needed from user input. `intermediate_keys` are `required_keys` append `intermediate_state`. The reason of not use BtreeMap because it will always search from start. The operations performed on each row of user data are kept as efficient as possible.

shuiyisong (approved):
Since the pr passed all the test, I assume it keeps compatible with our api. Let's give it a go.
Nitpick: The naming of the variables can be improved, let's discuss later.

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -32,54 +32,71 @@ const DEFAULT_GREPTIME_TIMESTAMP_COLUMN: &str = "greptime_timestamp";
...
                     Some(val) => coerce_value(val, transform)?,
-                    None if transform.get_default().is_some() => {
-                        coerce_value(transform.get_default().unwrap(), transform)?
+                    None => {
+                        let default = transform.get_default();
+                        if default.is_some() {
+                            coerce_value(default.unwrap(), transform)?
+                        } else {
+                            None
+                        }
```
zhongzc, line 76: Avoid unwrap, use match

zhongzc (approved):

src/pipeline/src/etl/processor/regex.rs
paomian: Next phase here will be another design to ensure code readability and reusability.

Issue #4399, opened 2024-07-19
Auto inference standalone/distributed mode from configuration


sunng87:
### What type of enhancement is this?

Configuration

### What does the enhancement do?

Currently we require user to configure `mode` to `distributed` or `standalone` on frontend (or datanode). Even if we have metasrv address configured for frontend, it still assumes itself as `standalone` when this `mode` set to `standalone`. However, no other feature is affected except the `cluster_info` meta table reports wrong role.

The suggested solution is to remove this `mode` configuration item and auto inference its mode based on `metasrv` configuration.

PR #4401, opened 2024-07-19
refactor!: Remove `Mode` from `FrontendOptions`

zyy17:

Resolve https://github.com/GreptimeTeam/greptimedb/issues/4399.

Remove `Mode` from `FrontendOptions` and determine whether it is Standalone or Distributed mode based on whether the Meta Client is configured.

## Checklist

### DIFF
--- a/src/frontend/src/server.rs
+++ b/src/frontend/src/server.rs
@@ -143,7 +143,15 @@ where
         };
 
         let user_provider = self.plugins.get::<UserProviderRef>();
-        let runtime = match opts.mode {
+
+        // Determine whether it is Standalone or Distributed mode based on whether the meta client is configured.
+        let mode = if opts.meta_client.is_none() {
+            Mode::Standalone
+        } else {
+            Mode::Distributed
+        };
+
+        let runtime = match mode {
             Mode::Standalone => Some(builder.runtime().clone()),
             _ => None,
         };
--- a/config/config.md
+++ b/config/config.md
@@ -166,7 +166,6 @@
 
 | Key | Type | Default | Descriptions |
 | --- | -----| ------- | ----------- |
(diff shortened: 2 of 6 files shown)


sunng87:
Just realized that if user doesn't edit their config toml to remove this `mode`, will we throw an error for this "unknown" config item?

If so, this can be a breaking change. Perhaps we can keep it for a minor release and just change the behaviour

zyy17:
> Just realized that if user doesn't edit their config toml to remove this `mode`, will we throw an error for this "unknown" config item?
> 
> If so, this can be a breaking change. Perhaps we can keep it for a minor release and just change the behaviour

Make sense. Let's keep it.

tisonkun:
serde typically ignores unknown fields. Do we assert unknown fields to failure anywhere?

zyy17:
> serde typically ignores unknown fields. Do we assert unknown fields to failure anywhere?

You are right. @sunng87, the `Mode` on the front end is not always necessary for our released version. When the user uses the current version, they don't need to modify their original config file and don't throw unrecoverable errors.

src/frontend/src/server.rs
```diff
@@ -143,7 +143,17 @@ where
...
-        let runtime = match opts.mode {
+
+        // Determine whether it is Standalone or Distributed mode based on whether the meta client is configured.
+        let mode = {
+            if opts.meta_client.is_none() {
+                Mode::Standalone
+            } else {
+                Mode::Distributed
+            }
+        };
```
evenyag, line 154: non-blocking: We don't need the outer brackets.
```suggestion
        let mode = if opts.meta_client.is_none() {
            Mode::Standalone
        } else {
            Mode::Distributed
        };
```

evenyag (approved):

fengjiachun (approved):
LGTM

PR #4405, opened 2024-07-22
feat: support setting time range in Copy From statement

### LINKED ISSUE #3511: Support setting time range in `Copy From` statement

User experience

As title. 

Currently there's `timestamp_range: Option<TimestampRange>` field in the `struct CopyTableRequest`, but it's not set. We can set it from the SQL parser first, then check if it works.

The `copy from` in postgresql supports "where condition" (https://www.postgresql.org/docs/current/sql-copy.html), maybe we can do that, too. For this issue, we can extract the time range from the "where condition", and leave more general filtering in the future.

### Implementation challenges

_No response_

realtaobo:

Resolve #3511 

Reuse the time range specified in the WITH clause and ensure the data can be filtered based on the 'ts' column.  Hence, `Copy table x from datasource with (start_time='2022-06-15 07:02:37', end_time='2022-06-15 07:02:38');` statement should work properly.
All work based [filter_record_batch](https://docs.rs/datafusion/latest/datafusion/common/arrow/compute/fn.filter_record_batch.html) in file:https://github.com/GreptimeTeam/greptimedb/blob/2ae2a6674e8a780cb93723de9b749b65ea28562e/src/common/recordbatch/src/adapter.rs#L94-L95

## Checklist

### DIFF
--- a/src/common/query/src/logical_plan/expr.rs
+++ b/src/common/query/src/logical_plan/expr.rs
@@ -18,6 +18,33 @@ use common_time::Timestamp;
 use datafusion_common::{Column, ScalarValue};
 use datafusion_expr::expr::Expr;
 use datafusion_expr::{and, binary_expr, Operator};
+use datatypes::data_type::DataType;
+use datatypes::schema::ColumnSchema;
+use datatypes::value::Value;
+
+pub fn build_same_type_ts_filter(
+    ts_schema: &ColumnSchema,
+    time_range: Option<TimestampRange>,
+) -> Option<Expr> {
+    let ts_type = ts_schema.data_type.clone();
+    let time_range = time_range?;
+    let start = time_range
+        .start()
+        .and_then(|start| ts_type.try_cast(Value::Timestamp(start)));
+    let end = time_range
+        .end()
+        .and_then(|end| ts_type.try_cast(Value::Timestamp(end)));
+
+    let time_range = match (start, end) {
+        (Some(Value::Timestamp(start)), Some(Value::Timestamp(end))) => {
+            TimestampRange::new(start, end)
+        }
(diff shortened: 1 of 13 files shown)


src/common/query/src/logical_plan/expr.rs
```diff
@@ -18,6 +18,33 @@ use common_time::Timestamp;
 use datafusion_common::{Column, ScalarValue};
 use datafusion_expr::expr::Expr;
 use datafusion_expr::{and, binary_expr, Operator};
+use datatypes::data_type::DataType;
+use datatypes::schema::ColumnSchema;
+use datatypes::value::Value;
+
+pub fn build_same_type_ts_filter(
```
killme2008, line 25: Could you add a comment about this function? Thank you. It's better to have a comment for every public function.

src/common/recordbatch/src/filter.rs
```diff
@@ -144,13 +151,42 @@ impl SimpleFilterEvaluator {
...
+pub fn batch_filter(
+    batch: &RecordBatch,
+    predicate: &Arc<dyn PhysicalExpr>,
+) -> DfResult<RecordBatch> {
+    predicate
+        .evaluate(batch)
+        .and_then(|v| v.into_array(batch.num_rows()))
+        .and_then(|array| {
+            let filter_array = match as_boolean_array(&array) {
+                Ok(boolean_array) => Ok(boolean_array.to_owned()),
```
killme2008, line 164: Prefer to use `clone` to indicate the operation explicitly.
```suggestion
                Ok(boolean_array) => Ok(boolean_array.clone()),
```

killme2008 (approved):
Almost LGTM. Please take a look @fengjiachun @WenyXu

WenyXu (approved):
Nice done🚀

PR #4410, opened 2024-07-22
chore: add more metrics about parquet and cache

MichaelScofield:

as title

useful for diagnosing slow queries

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a cache eviction metric for improved monitoring of cache performance.
	- Added a histogram metric to measure the time taken for loading Parquet metadata.

- **Improvements**
	- Enhanced observability of cache eviction reasons for better diagnostics.
	- Implemented performance tracking in the MetadataLoader for metadata loading efficiency.

### DIFF
--- a/src/mito2/src/cache.rs
+++ b/src/mito2/src/cache.rs
@@ -27,6 +27,7 @@ use std::sync::Arc;
 
 use datatypes::value::Value;
 use datatypes::vectors::VectorRef;
+use moka::notification::RemovalCause;
 use moka::sync::Cache;
 use parquet::column::page::Page;
 use parquet::file::metadata::ParquetMetaData;
@@ -266,6 +267,15 @@ impl CacheManagerBuilder {
 
     /// Builds the [CacheManager].
     pub fn build(self) -> CacheManager {
+        fn to_str(cause: RemovalCause) -> &'static str {
+            match cause {
+                RemovalCause::Expired => "expired",
+                RemovalCause::Explicit => "explicit",
+                RemovalCause::Replaced => "replaced",
+                RemovalCause::Size => "size",
+            }
+        }
+
         let sst_meta_cache = (self.sst_meta_cache_size != 0).then(|| {
             Cache::builder()
                 .max_capacity(self.sst_meta_cache_size)
@@ -276,7 +286,7 @@ impl CacheManagerBuilder {
                         .with_label_values(&[SST_META_TYPE])
                         .sub(size.into());
                     CACHE_EVICTION
-                        .with_label_values(&[SST_META_TYPE, &format!("{:?}", cause)])
+                        .with_label_values(&[SST_META_TYPE, to_str(cause)])
                         .inc();
                 })
                 .build()
@@ -291,7 +301,7 @@ impl CacheManagerBuilder {
                         .with_label_values(&[VECTOR_TYPE])
                         .sub(size.into());
                     CACHE_EVICTION
-                        .with_label_values(&[VECTOR_TYPE, &format!("{:?}", cause)])
+                        .with_label_values(&[VECTOR_TYPE, to_str(cause)])
                         .inc();
                 })
                 .build()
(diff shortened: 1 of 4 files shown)


src/mito2/src/metrics.rs
```diff
@@ -190,6 +190,12 @@ lazy_static! {
         "mito upload bytes total",
     )
     .unwrap();
+    /// Cache eviction counter, labeled with cache type and eviction reason.
+    pub static ref CACHE_EVICTION: IntGaugeVec = register_int_gauge_vec!(
+        "greptime_mito_cache_eviction",
+        "mito cache eviction",
+        &[TYPE_LABEL, "cause"]
```
v0y4g3r, line 197: Make `"cause"` a constant like `TYPE_LABEL`

src/mito2/src/cache.rs
```diff
@@ -270,33 +270,42 @@ impl CacheManagerBuilder {
             Cache::builder()
                 .max_capacity(self.sst_meta_cache_size)
                 .weigher(meta_cache_weight)
-                .eviction_listener(|k, v, _cause| {
+                .eviction_listener(|k, v, cause| {
```
v0y4g3r, line 273: Also check `cause.was_evicted()` : https://docs.rs/moka/latest/moka/notification/enum.RemovalCause.html#method.was_evicted

MichaelScofield: `was_evicted` simply checks the cause's variants, I think it's not necessary to be included in the original metrics, Grafana must has some way to derive a new "was evicted" metric from it.

MichaelScofield:
@v0y4g3r PTAL

src/mito2/src/metrics.rs
```diff
@@ -316,4 +323,12 @@ lazy_static! {
...
         &["op"]
     ).unwrap();
+
+    // Parquet related metrics:
+
+    /// Elapsed time of loading parquet metadata.
+    pub static ref PARQUET_METADATA_LOAD_ELAPSED: Histogram = register_histogram!(
+        "greptime_parquet_metadata_load_elapsed",
+        "parquet metadata load elapsed",
+    ).unwrap();
```
evenyag, line 333: You can reuse the `READ_STAGE_ELAPSED` metric and add a stage `load_parquet_meta`/`load_parquet_metadata`.

MichaelScofield: I think it's good to have some metrics all related to parquet read, this can be their first one, doesn't have to be a stage of the query.

evenyag: > I think it's good to have some metrics all related to parquet read

Then we should add a histogram vec instead of a dedicated metric for each operation. Maintaining too many metrics is a burden. You might need to modify the dashboard panel every time we add a new metric.

We also have to set the bucket size of the histogram as the default bucket size is not suitable.

MichaelScofield:
@evenyag PTAL

src/mito2/src/metrics.rs
```diff
@@ -190,6 +191,12 @@ lazy_static! {
         "mito upload bytes total",
     )
     .unwrap();
+    /// Cache eviction counter, labeled with cache type and eviction reason.
+    pub static ref CACHE_EVICTION: IntGaugeVec = register_int_gauge_vec!(
+        "greptime_mito_cache_eviction",
+        "mito cache eviction",
+        &[TYPE_LABEL, CACHE_EVICTION_CAUSE]
+    ).unwrap();
```
evenyag, line 199: I realized that we should use `CounterVec` here.

MichaelScofield:
@evenyag PTAL

src/mito2/src/cache.rs
```diff
@@ -270,33 +270,42 @@ impl CacheManagerBuilder {
...
                     let size = meta_cache_weight(&k, &v);
                     CACHE_BYTES
                         .with_label_values(&[SST_META_TYPE])
                         .sub(size.into());
+                    CACHE_EVICTION
+                        .with_label_values(&[SST_META_TYPE, &format!("{:?}", cause)])
```
evenyag, line 279: nitpick: Maybe we can implement a function to match the cause and return a `&'static str` to avoid allocating a string. It also improves the readability of the label value.

evenyag (approved):

v0y4g3r (approved):
LGTM

PR #4423, opened 2024-07-24
fix(fulltext-index): clean up 0-value timer

zhongzc:

Metric `greptime_index_create_elapsed_bucket` reports non-zero values even though no index needs to be built.

image

This patch fixes it.

## Checklist


killme2008:
Could you describe the bug? Is it fatal?

zhongzc:
> Could you describe the bug? Is it fatal?

Description updated. Not fatal.

waynexia (approved):

Issue #3560, opened 2024-03-22
Compatible with Postgres system catalog


tisonkun:
### What type of enhancement is this?

User experience

### What does the enhancement do?

Since we want to support users to access GreptimeDB with psql, we'd better be more compatible when possible.

Currently, when you run `psql -l` against GreptimeDB, it would fail with:

```
psql: ERROR:  Failed to plan SQL: Error during planning: Table not found: greptime.pg_catalog.pg_database
```

### Implementation challenges

This should be similar to MySQL's information

tisonkun:
Not quite easy. Seems related to some internal functions also:

```
Failed to execute query: SELECT d.datname as "Name",
       pg_catalog.pg_get_userbyid(d.datdba) as "Owner",
       pg_catalog.pg_encoding_to_char(d.encoding) as "Encoding",
       pg_catalog.array_to_string(d.datacl, '\n') AS "Access privileges"
FROM pg_catalog.pg_database d
ORDER BY 1; err=0: Table operation error, at src/frontend/src/instance.rs:272:14
1: Failed to plan statement, at /Users/tison/GreptimeWorkspace/greptimedb/src/operator/src/statement.rs:253:14
2: Failed to plan SQL, at src/query/src/planner.rs:90:14
...
```

tisonkun:
And this is the query for "show tables" (`\dt`):

```
Failed to execute query: SELECT n.nspname as "Schema",
  c.relname as "Name",
  CASE c.relkind WHEN 'r' THEN 'table' WHEN 'v' THEN 'view' WHEN 'm' THEN 'materialized view' WHEN 'i' THEN 'index' WHEN 'S' THEN 'sequence' WHEN 's' THEN 'special' WHEN 't' THEN 'TOAST table' WHEN 'f' THEN 'foreign table' WHEN 'p' THEN 'partitioned table' WHEN 'I' THEN 'partitioned index' END as "Type",
  pg_catalog.pg_get_userbyid(c.relowner) as "Owner"
FROM pg_catalog.pg_class c
     LEFT JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
WHERE c.relkind IN ('r','p','')
      AND n.nspname <> 'pg_catalog'
      AND n.nspname !~ '^pg_toast'
      AND n.nspname <> 'information_schema'
  AND pg_catalog.pg_table_is_visible(c.oid)
[...]

PR #4428, opened 2024-07-24
feat: support pg_namespace, pg_class and related psql command

J0HN50N133:

#3560 

- Summarize your change
  Add `pg_catalog.pg_namespace` and `pg_catalog.pg_class`. Now we can use `\d`, `\dt`, `\dv` and all 
  related psql commands.
- Describe any limitations of the current code:
  1. Column `oid` in `pg_catalog.pg_namespace` need `u32` values to identify `namespace`(i.e. schema in greptime), but there is no numeric schema id. So we just use `schema_name` as identifier. This workaround may introduce potential incompatibility with the existed pg eco. 
  2. `pg_catalog.pg_get_userbyid` always return empty string, since we don't have `UserId` to get `UserInfo`.
  for details see: https://github.com/GreptimeTeam/greptimedb/issues/3560#issuecomment-2242835478

## Checklist


J0HN50N133:
@tisonkun can you take a look?

src/catalog/src/system_schema/pg_catalog.rs
```diff
@@ -93,6 +97,11 @@ impl PGCatalogProvider {
         for name in MEMORY_TABLES.iter() {
             tables.insert(name.to_string(), self.build_table(name).expect(name));
         }
+        tables.insert(
+            PG_NAMESPACE.to_string(),
+            self.build_table(PG_NAMESPACE).unwrap(),
+        );
+        tables.insert(PG_CLASS.to_string(), self.build_table(PG_CLASS).unwrap());
```
sunng87, line 104: try to avoid `unwrap` here. If it's really safe to use unwrap, leave a comment for the background

src/catalog/src/system_schema/information_schema.rs
```diff
@@ -37,14 +35,19 @@ use common_recordbatch::SendableRecordBatchStream;
...
 use table::TableRef;
 pub use table_names::*;
 use views::InformationSchemaViews;
 
 use self::columns::InformationSchemaColumns;
+pub(crate) use crate::system_schema::predicate::Predicates;
+// TODO(J0HN150N133): re-export to provide compatibility for legacy `information_schema` code.
+pub(crate) mod utils {
+    pub use crate::system_schema::utils::*;
+}
```
sunng87, line 49: Yes we don't need compatibility at this level just refactor it

src/common/function/src/system/pg_catalog/pg_get_userbyid.rs
```diff
@@ -0,0 +1,72 @@
...
+            Volatility::Immutable,
+        )
+    }
+
+    fn eval(&self, _func_ctx: FunctionContext, columns: &[VectorRef]) -> Result<VectorRef> {
+        with_match_primitive_type_id!(columns[0].data_type().logical_type_id(), |$T| {
+            let col = scalar_unary_op::<<$T as LogicalPrimitiveType>::Native, String, _>(&columns[0], pg_get_user_by_id, &mut EvalContext::default())?;
+            Ok(Arc::new(col))
+        }, {
+            unreachable!()
```
sunng87, line 61: make sure this unreachable, otherwise return an error

src/common/function/src/system/pg_catalog/pg_get_userbyid.rs
```diff
@@ -0,0 +1,72 @@
...
+        })
+    }
+}
+
+fn pg_get_user_by_id<I>(table_oid: Option<I>, _ctx: &mut EvalContext) -> Option<String>
+where
+    I: AsPrimitive<u32>,
+{
+    // There is no user authentication in greptime open source core, so we always return ""
+    table_oid.map(|_| "".to_string())
```
sunng87, line 71: There is user concept in open source, see QueryContext which we can access from func_ctx

J0HN50N133: I know. But the actual meaning is the owner of the table or the view. And we need a way to get their info with a numeric id.

src/common/function/src/system/pg_catalog/pg_get_userbyid.rs
J0HN50N133: pretty sure. Since incompatible type will not pass type check. See other function in the codebase: https://github.com/J0HN50N133/greptimedb/blob/f7d251ab5657de5d11858b190ddbbc15a8632b2d/src/common/function/src/scalars/math/pow.rs#L53

src/catalog/src/system_schema/pg_catalog.rs
J0HN50N133: Yeah. Should use `expect` instead.PG_NAMESPACE

sunng87: Oh wait, expect is actually same with unwrap and it's just carrying a text message for the panic. We need to avoid panic in worst case.

J0HN50N133: Actually it's safe to use `unwarp` or `expect` here. If the constants passed into `build_table` have been handled by 'system_table`, it bounds to return a `SystemTableRef`. Since the function is not able to be called outside and the match logic is carefully checked, I think it's ok to unwrap here. Option is just used to pass type check here(since table_name is not modeling as a type). Even panic, it will just happen is dev stage. To cleanup the code, I would use `enum` and `strum` to statically trace the table supported to avoid `Option` here. This can be done by another pr.

sunng87: Sounds good. Please leave a comment and TODO here to other developers to understand this.

J0HN50N133:
@tisonkun Let's push forward this pr to avoid extra burden to keep track with main branch.

killme2008:
Great work! I'll give a review today. Thank you.

killme2008:
> Column oid in pg_catalog.pg_namespace need u32 values to identify namespace(i.e. schema in greptime), but there is no numeric schema id. So we just use schema_name as identifier. This workaround may introduce potential incompatibility with the existed pg eco

A question or proposition: could we use a hash value of the schema name(if it is a perfect hash function, AKA no conflicts)?

src/catalog/src/system_schema/pg_catalog/pg_class.rs
```diff
@@ -0,0 +1,248 @@
...
+                    .await
+                    .map(|x| x.into_df_record_batch())
+                    .map_err(Into::into)
+            }),
+        ))
+    }
+}
+
+/// Builds the `pg_catalog.pg_class` table row by row
+/// `relowner` is always the [`DUMMY_OWNER_ID`] cuz we don't have user
```
killme2008, line 152: Could you change this comment to `TODO`? We will design the user system for GreptimeDB.

killme2008 (approved):
Thanks a lot. LGTM.

tisonkun (approved):

tisonkun:
Thanks for your contribution! Merging ...

killme2008:
> > Column oid in pg_catalog.pg_namespace need u32 values to identify namespace(i.e. schema in greptime), but there is no numeric schema id. So we just use schema_name as identifier. This workaround may introduce potential incompatibility with the existed pg eco
> 
> A question or proposition: could we use a hash value of the schema name(if it is a perfect hash function, AKA no conflicts)?

@J0HN50N133 Could you consider this suggestion? Using `schema_name` as schema id is a potential issue, the data type is totaly different.

killme2008:
> > > > Column oid in pg_catalog.pg_namespace need u32 values to identify namespace(i.e. schema in greptime), but there is no numeric schema id. So we just use schema_name as identifier. This workaround may introduce potential incompatibility with the existed pg eco
> > > 
> > > 
> > > A question or proposition: could we use a hash value of the schema name(if it is a perfect hash function, AKA no conflicts)?
> > 
> > 
> > @J0HN50N133 Could you consider this suggestion? Using `schema_name` as schema id is a potential issue, the data type is totally different.
> 
> @killme2008 A perfect hash is hard to find, but I have a workaround. If we can make sure that the numeric id is only used in `catalog` module, we could use hash and handle the collision inside pg_catalog. Here's the detail:
> 
> 1. we hash a schema name to get a u32 id.
> 2. If no collision's found, that's good, put this id into an `id_set`
> 3. If collision's found, rehash the u32 id until no collision.
>    But this method may cause the id change along with schema creation. Suppose there is a schema `a`.  At first, `hash(a)` is good, no collision's found. Then we create another schema `b`, which hash value collide with `a` and is hashed before `a`. Then  `a`'s id will be a different value.
> 
> Or we can create a new metadata manager to maintain the schema id.

I think it can work currently. The best way is to add a schema id in metadata, let's do the refactor in the future. @fengjiachun 
In fact, there will not be so many schemas in actual deployments except the cloud.

J0HN50N133:
> > > > > Column oid in pg_catalog.pg_namespace need u32 values to identify namespace(i.e. schema in greptime), but there is no numeric schema id. So we just use schema_name as identifier. This workaround may introduce potential incompatibility with the existed pg ecopg_catalog.pg_namespace 表中的列 oid 需要 u32 值来标识命名空间（即 Greptime 中的 schema），但目前没有数字形式的 schema ID。因此，我们暂时使用 schema_name 作为标识符。这种权宜之计可能会引入与现有 PostgreSQL 生态的不兼容性。
> > > > 
> > > > 
> > > > A question or proposition: could we use a hash value of the schema name(if it is a perfect hash function, AKA no conflicts)?一个问题或提议：我们能否使用模式名称的哈希值（如果它是一个完美的哈希函数，即无冲突）？
> > > 
> > > 
> > > @J0HN50N133 Could you consider this suggestion? Using `schema_name` as schema id is a potential issue, the data type is totally different.您是否考虑过这个建议？使用 `schema_name` 作为模式 ID 可能存在问题，因为数据类型完全不同。
> > 
> > 
> > @killme2008 A perfect hash is hard to find, but I have a workaround. If we can make sure that the numeric id is only used in `catalog` module, we could use hash and handle the collision inside pg_catalog. Here's the detail:完美哈希难以寻觅，但我有一套变通方案。若能确保数值 ID 仅在 `catalog` 模块内使用，我们便可采用哈希并在 pg_catalog 内部处理冲突。以下是详细方案：
> > 
> > 1. we hash a schema name to get a u32 id.我们通过哈希一个模式名称来获取一个 u32 ID。
> > 2. If no collision's found, that's good, put this id into an `id_set`如果没有发现冲突，那就好，将此 ID 放入 `id_set`
> > 3. If collision's found, rehash the u32 id until no collision.如果发现冲突，重新哈希 u32 ID 直到无冲突。
> >    But this method may cause the id change along with schema creation. Suppose there is a schema `a`.  At first, `hash(a)` is good, no collision's found. Then we create another schema `b`, which hash value collide with `a` and is hashed before `a`. Then  `a`'s id will be a different value.但这种方法可能导致 ID 随着模式创建而改变。假设存在一个模式 `a` 。起初， `hash(a)` 是良好的，未发现冲突。随后，我们创建另一个模式 `b` ，其哈希值与 `a` 发生冲突，并且在 `a` 之前进行哈希处理。那么 `a` 的 ID 将会是一个不同的值。
> > 
> > Or we can create a new metadata manager to maintain the schema id.或者，我们可以创建一个新的元数据管理器来维护模式 ID。
> 
> I think it can work currently. The best way is to add a schema id in metadata, let's do the refactor in the future. 我认为目前它可以正常工作。最佳方案是在元数据中添加一个模式 ID，我们将来再进行重构。@fengjiachun In fact, there will not be so many schemas in actual deployments except the cloud.实际上，除了云端，实际部署中并不会存在如此多的架构。

@killme2008 Got it. I'll implement the workaround tomorrow.

PR #4436, opened 2024-07-25
feat: remove dedicated runtime for grpc, mysql and pg protocols 

waynexia:

Refine runtime related usages.

Major changes:
- remove dedicated runtime for gRPC, MySQL and Postgres
- replace read, write, bg runtimes with the new unified one `global_runtime`
- add a new `compact_runtime` for compaction only to limit its max CPU usage. This is a simple workaround at present.

## Checklist


src/common/runtime/src/global.rs
```diff
@@ -95,48 +90,38 @@ macro_rules! define_spawn {
...
-                .unwrap_or_else(|| create_runtime("global-read", "read-worker", READ_WORKERS)),
-            write_runtime: write
-                .unwrap_or_else(|| create_runtime("global-write", "write-worker", WRITE_WORKERS)),
-            bg_runtime: background
-                .unwrap_or_else(|| create_runtime("global-bg", "bg-worker", BG_WORKERS)),
+            global_runtime: global
+                .unwrap_or_else(|| create_runtime("global", "global-worker", GLOBAL_WORKERS)),
+            compact_runtime: compact
+                .unwrap_or_else(|| create_runtime("compact", "compact-worker", COMPACT_WORKERS)),
             hb_runtime: heartbeat
```
zhongzc, line 103: It turns out that `hb` means `heartbeat`. How about making it clear?

src/servers/tests/mysql/mysql_server_test.rs
```diff
@@ -45,14 +46,13 @@ struct MysqlOpts<'a> {
...
-    let query_handler = create_testing_sql_query_handler(table);
-    let io_runtime = Arc::new(
-        RuntimeBuilder::default()
-            .worker_threads(4)
-            .thread_name("mysql-io-handlers")
-            .build()
-            .unwrap(),
-    );
+    let query_handler: Arc<dyn SqlQueryHandler<Error = servers::error::Error> + Send + Sync> =
+        create_testing_sql_query_handler(table);
```
zhongzc, line 50:
```suggestion
let query_handler = create_testing_sql_query_handler(table);
```

zhongzc (approved):

config/datanode.example.toml
```diff
@@ -73,11 +73,9 @@ watch = false
 ## The runtime options.
 [runtime]
 ## The number of threads to execute the runtime for global read operations.
```
killme2008, line 75: Don't forget to update the configuration comments

src/common/runtime/src/global.rs
```diff
@@ -95,48 +90,38 @@ macro_rules! define_spawn {
...
         Self {
-            read_runtime: read
-                .unwrap_or_else(|| create_runtime("global-read", "read-worker", READ_WORKERS)),
-            write_runtime: write
-                .unwrap_or_else(|| create_runtime("global-write", "write-worker", WRITE_WORKERS)),
-            bg_runtime: background
-                .unwrap_or_else(|| create_runtime("global-bg", "bg-worker", BG_WORKERS)),
+            global_runtime: global
+                .unwrap_or_else(|| create_runtime("global", "global-worker", GLOBAL_WORKERS)),
+            compact_runtime: compact
```
sunng87, line 101: if we are adding something like async index, will we reuse this runtime?

sunng87:
Actually this is a breaking change. It changes how we configure and tune the server. We'd better including this in next minor release instead of patch release.

waynexia:
> Actually this is a breaking change. It changes how we configure and tune the server. We'd better including this in next minor release instead of patch release.

I've tested this change won't fail the server start progress. Only the old config won't take effect so I didn't categorize it as breaking change.

config/datanode.example.toml
waynexia: It's updated

src/common/runtime/src/global.rs
waynexia: I changed the runtime name. The thread name has char limits

src/common/runtime/src/global.rs
waynexia: Yes, this global one is for all operations. Two exceptions are compaction and heartbeat (not included at present

evenyag (approved):

src/datanode/src/alive_keeper.rs
```diff
@@ -187,7 +187,7 @@ impl RegionAliveKeeper {
             let running = self.started.clone();
 
             // Watches changes
-            common_runtime::spawn_bg(async move {
+            common_runtime::spawn_global(async move {
```
evenyag, line 190:
```suggestion
            common_runtime::spawn_hb(async move {
```

src/datanode/src/alive_keeper.rs
```diff
@@ -286,7 +286,7 @@ impl CountdownTaskHandle {
             region_id,
             rx,
         };
-        let handler = common_runtime::spawn_bg(async move {
+        let handler = common_runtime::spawn_global(async move {
```
evenyag, line 289:
```suggestion
        let handler = common_runtime::spawn_hb(async move {
```

evenyag (commented):
Do we need to run the alive keeper in the heartbeat runtime? The region alive keeper may be related to the heartbeat task.

evenyag:
Still timeout
https://github.com/GreptimeTeam/greptimedb/actions/runs/10146099115/job/28053577040?pr=4436

WenyXu:
> Still timeout https://github.com/GreptimeTeam/greptimedb/actions/runs/10146099115/job/28053577040?pr=4436

I updated the runtime config for tests. By default, each pod will be allocated up to 1 core CPU (which will lead the default global_rt_size to be 1). However, it's problematic to set the global runtime size to 1 (e.g., deadlock).

evenyag:
> I updated the runtime config for tests. By default, each pod will be allocated up to 1 core CPU (which will lead the default global_rt_size to be 1). However, it's problematic to set the global runtime size to 1 (e.g., deadlock).

I found the table is flushing:
```
2024-07-29T14:29:49.385378397Z stdout F 2024-07-29T14:29:49.385189Z  INFO operator::request: Handle table manual flush request: FlushTableRequest { catalog_name: "greptime", schema_name: "public", table_name: "quod" }
```

The admin function may be the cause. We try to block inside the async runtime and hangs the whole runtime.
https://github.com/GreptimeTeam/greptimedb/blob/53fc14a50b192f628f4ba6cfc759de9e1d063105/src/common/macro/src/admin_fn.rs#L230-L248

src/common/macro/src/admin_fn.rs
```diff
@@ -228,7 +228,7 @@ fn build_struct(
                         .create_mutable_vector(rows_num);
 
                     if columns_num == 0 {
-                        let result = common_runtime::block_on_read(async move {
```
evenyag, line 231: The current implementation of the admin fn may hang the runtime.

tisonkun (approved):

PR #4448, opened 2024-07-29
feat: track channels with query context and w/rcu

sunng87:

This patch adds a new field `channel` in query context so the execution path now knows which source the query comes from. It also add this channel to meter, by which we can add statistics for w/rcus of each channel.

Upstream changes that has to be merged before this:

- https://github.com/GreptimeTeam/greptime-proto/pull/183
- https://github.com/GreptimeTeam/greptime-meter/pull/24

**Note that** this patch changes QueryContext extension type in axum, from `QueryContextRef` to plain `QueryContext` so that we can modify it in http handlers, to set `channel` field based on different handler.

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,12 +119,12 @@ etcd-client = { version = "0.13" }
 fst = "0.4.7"
 futures = "0.3"
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "e52fa3e63c77545618328781e11178eb2cfc45c5" }
+greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "c437b55725b7f5224fe9d46db21072b4a682ee4b" }
 humantime = "2.1"
 humantime-serde = "1.1"
 itertools = "0.10"
 lazy_static = "1.4"
-meter-core = { git = "https://github.com/GreptimeTeam/greptime-meter.git", rev = "488844b4702916cc6f7d04898fb0b837afce4ca7" }
+meter-core = { git = "https://github.com/GreptimeTeam/greptime-meter.git", rev = "049171eb16cb4249d8099751a0c46750d1fe88e7" }
 mockall = "0.11.4"
 moka = "0.12"
 notify = "6.1"
@@ -238,7 +238,7 @@ table = { path = "src/table" }
 
 [workspace.dependencies.meter-macros]
 git = "https://github.com/GreptimeTeam/greptime-meter.git"
-rev = "488844b4702916cc6f7d04898fb0b837afce4ca7"
+rev = "049171eb16cb4249d8099751a0c46750d1fe88e7"
 
 [profile.release]
 debug = 1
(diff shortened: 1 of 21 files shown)


shuiyisong:
Please resolve conflicts first _(:з」∠)_

src/servers/src/http/authorize.rs
```diff
@@ -68,7 +66,7 @@ pub async fn inner_auth<B>(
         .current_schema(schema.clone())
         .timezone(timezone);
 
-    let query_ctx = Arc::new(query_ctx_builder.build());
+    let query_ctx = query_ctx_builder.build();
```
shuiyisong, line 69: Should we add `Channel::Http` as build default value here?

Cargo.toml
```diff
@@ -119,12 +119,13 @@ etcd-client = { version = "0.13" }
...
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "7ca323090b3ae8faf2c15036b7f41b7c5225cf5f" }
+greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "e52fa3e63c77545618328781e11178eb2cfc45c5" }
 humantime = "2.1"
 humantime-serde = "1.1"
 itertools = "0.10"
 lazy_static = "1.4"
-meter-core = { git = "https://github.com/GreptimeTeam/greptime-meter.git", rev = "80b72716dcde47ec4161478416a5c6c21343364d" }
+meter-core = { git = "https://github.com/GreptimeTeam/greptime-meter.git", rev = "488844b4702916cc6f7d04898fb0b837afce4ca7" }
+meter-macros = { git = "https://github.com/GreptimeTeam/greptime-meter.git", rev = "488844b4702916cc6f7d04898fb0b837afce4ca7" }
```
shuiyisong, line 128: `meter-macros` is placed deliberately in a different section  because we want to use different features in other version. I'm not sure place the dependency here is a good option.

shuiyisong (approved):
Rest LGTM

src/servers/src/http/authorize.rs
sunng87: Because http as a transport channel is shared by a few protocols, so we will specify the channel in each particular handler

Cargo.toml
sunng87: I see. Let me revert this.

src/session/src/context.rs
```diff
@@ -289,17 +306,45 @@ impl ConnInfo {
...
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Default, Clone, Copy)]
+#[repr(u8)]
 pub enum Channel {
-    Mysql,
-    Postgres,
+    #[default]
+    Other = 0,
```
MichaelScofield, line 313: I think "Unknown" may be a better name.

MichaelScofield (approved):

src/session/src/context.rs
sunng87: Sounds good. `Unknown` is mostly for a (de)serialization issue.

Issue #4475, opened 2024-07-31
Region stall never recover


v0y4g3r:
### What type of bug is this?

Locking issue, Performance issue

### What subsystems are affected?

Storage Engine

### Minimal reproduce step

Ingesting large amount of data to partitioned tables with multiple regions.

### What did you expect to see?

Data ingestion is expected to recover when flush finished.

### What did you see instead?

Region write stalls forever and it can be observed from `greptime_mito_write_stall_total` gauge.

### What operating system did you use?

NA

### What version of GreptimeDB did you use?

0.9.0

PR #4476, opened 2024-07-31
fix: notify flush receiver after write buffer is released

evenyag:

fixes https://github.com/GreptimeTeam/greptimedb/issues/4475

This PR notifies all workers once a write buffer (memtable) is released so workers can handle stalled requests.

It also adds worker id to some logs and metrics so we can know the relationship between regions and workers.

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a notification system for memory management in the Write Buffer Manager, improving responsiveness.
	- Enhanced metrics tracking with REGION_COUNT now providing per-worker granularity.

- **Bug Fixes**
	- Improved logging to include worker IDs, aiding in better traceability during operations.

- **Refactor**
	- Shifted from global metrics to instance-based tracking for region counts, enhancing modularity and maintainability.

- **Tests**
	- Added a new test to validate the notifier functionality within the Write Buffer Manager.

### DIFF
--- a/src/mito2/src/flush.rs
+++ b/src/mito2/src/flush.rs
@@ -23,7 +23,7 @@ use smallvec::SmallVec;
 use snafu::ResultExt;
 use store_api::storage::RegionId;
 use strum::IntoStaticStr;
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, watch};
 
 use crate::access_layer::{AccessLayerRef, OperationType, SstWriteRequest};
 use crate::cache::CacheManagerRef;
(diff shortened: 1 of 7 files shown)


evenyag:
@coderabbitai summary

src/mito2/src/flush.rs
```diff
@@ -88,6 +88,9 @@ pub struct WriteBufferManagerImpl {
     memory_used: AtomicUsize,
     /// Memory that hasn't been scheduled to free (e.g. used by mutable memtables).
     memory_active: AtomicUsize,
+    /// Optional notifier.
+    /// The manager can wake up the worker once we free the write buffer.
+    notifier: Option<watch::Sender<()>>,
```
waynexia, line 93: When would this optional notifier be set and consumed?

waynexia (approved):
This way is kind of error-prone and doesn't smell good

src/mito2/src/flush.rs
evenyag: The worker group will set it.

evenyag:
> This way is kind of error-prone and doesn't smell good

The write buffer limit is global so typically we don't have a perfect way.

WenyXu (approved):

PR #4491, opened 2024-08-02
fix: partition column with mixed quoted and unquoted idents

waynexia:

Queries like this cannot be parsed correctly before:

```sql
CREATE TABLE monitor (
  `host_id`    INT,
  idc        STRING,
  ts         TIMESTAMP,
  cpu        DOUBLE DEFAULT 0,
  memory     DOUBLE,
  TIME INDEX (ts),
  PRIMARY KEY (host),
)
PARTITION ON COLUMNS(IdC, host_id) (
  idc <= 'hz' AND host_id < 1000,
  idc > 'hz' AND idc <= 'sh' AND host_id < 2000,
  idc > 'sh' AND host_id < 3000,
  idc > 'sh' AND host_id >= 3000,
)";
```

Because the quote style is also included in the equality comparison which is not correct and necessary.

## Checklist

### DIFF
--- a/src/sql/src/parsers/create_parser.rs
+++ b/src/sql/src/parsers/create_parser.rs
@@ -950,7 +950,7 @@ fn ensure_exprs_are_binary(exprs: &[Expr], columns: &[&Column]) -> Result<()> {
             ensure_one_expr(right, columns)?;
         } else {
             return error::InvalidSqlSnafu {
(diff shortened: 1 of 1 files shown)


evenyag (approved):

v0y4g3r (approved):
LGTM

PR #4494, opened 2024-08-02
chore: bump `shadow-rs` version to set the path to find the correct git repo

MichaelScofield:

In our another project with "greptimedb" as its submodule, the git info fetched in building is wrong. This is because the git repo finding ends within the submodule, which should have been the outside repo that owns the submodule.

## Checklist

### DIFF
--- a/src/common/version/build.rs
+++ b/src/common/version/build.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::env;
+
 use build_data::{format_timestamp, get_source_time};
 
 fn main() -> shadow_rs::SdResult<()> {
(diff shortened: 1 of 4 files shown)


src/common/version/build.rs
```diff
@@ -25,5 +27,9 @@ fn main() -> shadow_rs::SdResult<()> {
         }
     );
     build_data::set_BUILD_TIMESTAMP();
-    shadow_rs::new()
+
+    let src_path = env::var("CARGO_WORKSPACE_DIR").or_else(|_| env::var("CARGO_MANIFEST_DIR"))?;
+    let out_path = env::var("OUT_DIR")?;
+    let _ = shadow_rs::Shadow::build_with(src_path, out_path, Default::default())?;
+    Ok(())
```
tisonkun, line 34: Just to ensure that this requires version bump?

The commit message looks misleading .. We primarily want to fix this issue, while version bump is part of the requirement ..

MichaelScofield: The `build_with` function is made pub in new `shadow-rs` version, see https://github.com/baoyachi/shadow-rs/pull/169. The fix is actually in our another repo, not here.

tisonkun (approved):

sunng87 (approved):

PR #4495, opened 2024-08-03
fix: sample configs

killme2008:

comment out all sample configurations with default values.

## Checklist

### DIFF
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -241,11 +241,11 @@ type = "File"
 ## Cache configuration for object storage such as 'S3' etc.
 ## The local file cache directory.
 ## +toml2docs:none-default
-cache_path = "/path/local_cache"
+## cache_path = "/path/local_cache"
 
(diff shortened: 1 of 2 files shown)


killme2008:
If we don't want to comment out these parameters, it's better to support `auto` as the default value, just like

https://github.com/GreptimeTeam/greptimedb/blob/fe1cfbf2b3984eaaefd688e056ce5824139d4cf2/config/datanode.example.toml#L459

evenyag (approved):

zyy17 (approved):

zyy17:
@killme2008 

After some thought, I think it's not proper to comment those fileds. We should list all the available options in the example toml to generate the docs. If there is not a default value, we can use:

```
## +toml2docs:none-default
...
```

Or can we create another directory to store some config samples for development?

killme2008:
I have closed this issue. The current sample configurations are acceptable, though not ideal. A detailed configuration guide will be included in the user documentation.

Issue #4497, opened 2024-08-03
Supports auto as values for all numeric configurations with default settings


killme2008:
### What type of enhancement is this?

Configuration

### What does the enhancement do?

The numeric configurations like 

https://github.com/GreptimeTeam/greptimedb/blob/fe1cfbf2b3984eaaefd688e056ce5824139d4cf2/config/datanode.example.toml#L351

It has a default value that depends on system memory, but we hard-coded it in the sample configuration, it's not expected.

I create a PR #4495  to comment out all of them. But a better way is to make them support `auto` as the default value just like

https://github.com/GreptimeTeam/greptimedb/blob/fe1cfbf2b3984eaaefd688e056ce5824139d4cf2/config/datanode.example.toml#L432

PR #4513, opened 2024-08-05
feat: supports auto as default value for some mito configs

killme2008:

#4497 

Main changes:
* Adds a macro `define_mem_size_enum` to define an enum type, which supports `auto`, `unlimited`, and `ReadableSize`. When it's `auto`, its value is determined by system memory size, factor, and maximum size.
* Use `define_mem_size_enum` to define `global_buffer_size`, `page_cache_size` etc.
* Update sample configurations.
* Fixed `/config` HTTP API returning the wrong value in standalone mode.
* When the default value is `auto`, make `/config` to return the actual value.
* `information_schema` is read-only currently, updated `export_metrics.self_import.db` to `greptime_metrics`.
* Adds `http` configs to the datanode sample config file.

## Checklist


src/cmd/src/standalone.rs
```diff
@@ -561,7 +568,7 @@ impl StartCommand {
 
         let (tx, _rx) = broadcast::channel(1);
 
-        let servers = Services::new(fe_opts, Arc::new(frontend.clone()), plugins)
+        let servers = Services::new(opts, Arc::new(frontend.clone()), plugins)
```
zyy17, line 571: non-blocking: We also can remove the clone of `fe_opts` in L543.

src/mito2/src/config.rs
```diff
@@ -48,6 +49,30 @@ const PAGE_CACHE_SIZE_FACTOR: u64 = 8;
 /// Use `1/INDEX_CREATE_MEM_THRESHOLD_FACTOR` of OS memory size as mem threshold for creating index
 const INDEX_CREATE_MEM_THRESHOLD_FACTOR: u64 = 16;
 
+define_mem_size_enum!(
+    /// Global write buffer size threshold to trigger flush.
+    GlobalWriteBufferSize, GLOBAL_WRITE_BUFFER_SIZE_FACTOR, ReadableSize::gb(1), Some(ReadableSize::gb(1)));
```
evenyag, line 54: Do we need to support unlimited size for all options? Also, if we decouple the example config from the default value, we can use the `auto` value by not specifying this option in the TOML. Then we can still use `ReadableSize`.

Also, there is an issue with the mem size enum: We can't precompute the size while the enum is `Auto`. We need to compute the size each time we call `as_bytes()`.

We may also have some options that we want to disable by default and can't get the default value automatically.

Maybe a better way is to support our requirements in https://github.com/zyy17/toml2docs

evenyag (commented):
The "auto" option may not be the best solution. We might need to find a way to support the example config more properly.

zyy17 (approved):
LGTM. Nice refactoring.

config/frontend.example.toml
```diff
@@ -128,6 +113,21 @@ enable = true
 ## Whether to store the data from Prometheus remote write in metric engine.
 with_metric_engine = true
 
+## The runtime options.
+[runtime]
+## The number of threads to execute the runtime for global read operations.
+global_rt_size = 8
+## The number of threads to execute the runtime for global write operations.
+compact_rt_size = 4
```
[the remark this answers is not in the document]
zyy17: Yeap, it's one of the solution that I'm plan to do.

zyy17:
> The "auto" option may not be the best solution. We might need to find a way to support the example config more properly.

Why is it not a good solution? IMO, taking `auto` as default is natural.

src/mito2/src/config.rs
killme2008: I thought we could remove the `unlimited`, what do you think? @zhongzc .

> We can't precompute the size while the enum is Auto. We need to compute the size each time we call as_bytes().

Yes, but I think that is good. We don't call it many times.

evenyag:
> > The "auto" option may not be the best solution. We might need to find a way to support the example config more properly.
> 
> Why is it not a good solution? IMO, taking `auto` as default is natural.

Mentioned in https://github.com/GreptimeTeam/greptimedb/pull/4513#discussion_r1705703011

We also have to define a new type for each size in this way, just because we need a different default value. @zyy17

src/mito2/src/config.rs
evenyag: We need to define a new type for each size just because we need a different default value. At least we can define a function for each default value and provide a method for `get_or_default()`.
```rust
enum AutoReadableSize {}

let size = AutoReadableSize::Auto;
let x = size.get_or_default(default_global_write_buffer_size);
```

Issue #4230, opened 2024-06-28
[Tracking issue] Add json datatype


CookiePieWw:
### What type of enhancement is this?

Other

### What does the enhancement do?

Support to store and query json data as an independent datatype.

### Tasks

- [x] Build a benchmark to choose the representation of json inside memory and SST files.
   - see https://github.com/CookiePieWw/json-format-in-parquet-benchmark for more information.
- [x] RFC https://github.com/GreptimeTeam/greptimedb/pull/4515
- [x] Add json datatype and other related support.
   - https://github.com/GreptimeTeam/greptimedb/pull/4619
   - Currently the insertions of json type through "PREPARE" and "EXECUTE" stmt are not available.
- [x] Add UDFs for querying json.
   - [x] Add casts between jsonb value and other types, e.g. `parse_jsonb("{\"a\": \"b\"}")`
      - [x] string to json: see `to_json` in #4619 
   - [x] Add functions to query inside json through paths, e.g. `get_by_path_int(".a.b", json)`
      https://github.com/GreptimeTeam/greptimedb/pull/4720
   - [x] Add functions to check if the given json is of specified json type.
      https://github.com/GreptimeTeam/greptimedb/pull/4726

WenyXu:
Future work will be tracked in issue #4805.

PR #4515, opened 2024-08-06
docs: json datatype rfc

CookiePieWw:

#4230

RFC of json datatype. 

## Checklist


docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,141 @@
...
+   Struct(
+       Field("a", Int32),
+   )
+   =
+   Struct(
+       Field("a", BinaryArray), // JSONB
+   )
+   ```
+
+Like schema inference, schema changes are performed automatically without manual configuration.
```
killme2008, line 131: The current schema altering in GreptimeDB is really expensive, do we have a plan to improve it in such a case? cc @WenyXu

killme2008 (approved):
Looks cool to me!

fengjiachun:
LGTM, @WenyXu PTAL

docs/rfcs/2024-08-06-json-datatype.md
fengjiachun: I think the schema of JSON does not need to be in the table schema. This way, we only need to manage the JSON schema on the datanode without having to execute the alter table procedure. The JSON data itself carries schema information, so if you want to view the schema, directly viewing the data might also be a good way.

WenyXu: > I think the schema of JSON does not need to be in the table schema. This way, we only need to manage the JSON schema on the datanode without having to execute the alter table procedure. The JSON data itself carries schema information, so if you want to view the schema, directly viewing the data might also be a good way.

I agree with you.

WenyXu: Maybe we can alter the schema while building the vector in `VectorBuilder`. 
i.e.,
```rust
let mut builder = JsonBuilder::default();
builder.push(json1);
builder.push(json2); // The schema of JSON could be changed at this time.
```

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,141 @@
...
+    Field("a", Utf8),
+    Field("b", Utf8),
+    Field("c", Struct(Field("d", Float64))),
+)
+```
+
+Dynamic schema inference helps achieve compression in some scenarios. See [benchmark](https://github.com/CookiePieWw/json-format-in-parquet-benchmark/) for more information.
+
+## Schema Change
+The schema must remain consistent for a column within a table. When inserting data with different schemas, schema changes may occur. There are two types of schema changes:
```
evenyag, line 78: Will json schema change trigger a flush?

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,141 @@
...
+       Field("a", BinaryArray), // JSONB
+   )
+   ```
+
+Like schema inference, schema changes are performed automatically without manual configuration.
+
+# Drawbacks
+
+1. This datatype is best suited for data with similar schemas. Varying schemas can lead to frequent schema changes and fallback to JSONB.
+2. Schema inference and change bring additional writing overhead in favor of better compression rate.
```
evenyag, line 136: Did we have a benchmark to show the performance impact?

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,141 @@
...
+
+   Compatible fields can be altered to the widest type, similar to integral promotion in C:
+   ```Rust
+   Struct(
+       Field("a", Int16),
+   )
+   +
+   Struct(
+       Field("a", Int32),
+   )
```
evenyag, line 109: What if Int16 + String? Does it become a String or a JSONB?

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,141 @@
...
+   =
+   Struct(
+       Field("a", Utf8),
+       Field("e", Int32)
+   )
+   ```
+
+2. Field Modification
+
+   Compatible fields can be altered to the widest type, similar to integral promotion in C:
```
evenyag, line 101: I wonder if it brings additional complexity to the storage engine.

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,141 @@
...
+   Struct(
+       Field("a", Int32),
+   )
+   =
+   Struct(
+       Field("a", Int32),
+   )
+   ```
+
+   Non-compatible fields will fallback to a binary array to store the JSONB encoding:
```
evenyag, line 116: Maybe we can support JSONB first to see whether we have unexpected problems implementing features like projection, and functions for JSON.

docs/rfcs/2024-08-06-json-datatype.md
CookiePieWw: Yes, it should trigger a flush.

docs/rfcs/2024-08-06-json-datatype.md
CookiePieWw: Schema inference takes about 20% of time as parsing a json string. For schema change, I'd like to assume it rarely happens and only results in a flush and a metadata change if we make the schema of json self-explanatory instead of being in the table schema.

As for the compression rate, results are presented [here](https://github.com/CookiePieWw/json-format-in-parquet-benchmark). About 80% smaller than jsonb at most in some cases.

docs/rfcs/2024-08-06-json-datatype.md
CookiePieWw: It becomes a JSONB. I think `integers` and `float to double` are compatible, others are not.

docs/rfcs/2024-08-06-json-datatype.md
CookiePieWw: I think so. At least there's some additional logic about sst compaction.

docs/rfcs/2024-08-06-json-datatype.md
CookiePieWw: We can provide jsonb for common users, and this datatype for those who cares about compression rate? cc @WenyXu

WenyXu: > Yes, I think we can provide the JSONB for users.

docs/rfcs/2024-08-06-json-datatype.md
WenyXu: > Yes, it should trigger a flush.

I think there is **no actual schema change**. The actual type (parquet column schema) is inferred from the data in the memtable and the column schema is only known in the parquet files.  i.e., 

// t1, field a, parquet column schema: object { a: string }
// t2, field a, parquet column schema: object { a: string, b: int  }
// t3, field a, parquet column schema: object { a: object { c: string }  }

For the upper components, mito, or frontend component. They only know the field `a` is a JSON datatype column.

CookiePieWw: Ah I see, so the schemas can keep different in the memtable, and when flushing into parquet files, we merge different schemas. It's more like an implicit compaction instead of an explicit change.

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,141 @@
...
+Tracking Issue: https://github.com/GreptimeTeam/greptimedb/issues/4230
+Date: 2024-8-6
+Author: "Yuhan Wang <profsyb@gmail.com>"
+---
+
+# Summary
+This RFC proposes a method for storing and querying JSON data in the database.
+
+# Motivation
+JSON is widely used across various scenarios. Direct support for writing and querying JSON can significantly enhance the database's flexibility.
```
waynexia, line 12: From a real scenario, ingesting JSON data can invoke a PIPELINE to preprocess JSON data into table structure data which should have better performance and usability. And given we already have the auto-alter mechanism, the flexibility also looks good

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,141 @@
...
+
+## User Interface
+The feature introduces a new data type for the database, similar to the common JSON type. Data is written as JSON strings and can be queried using functions.
+
+For example:
+```SQL
+CREATE TABLE IF NOT EXISTS test (
+    ts TIMESTAMP TIME INDEX,
+    a INT,
+    b JSON
```
waynexia, line 24: I'm not sure if the new `JSON` type is good. It erases the detailed type info and makes everything in this field dynamic. Is it possible to represent this in schema with a clear `Struct` type?

We can still invoke those JSON-related functions on `String` or `Bytes` fields. But not include a new opaque `JSON` type. If we have some optimization to do for `JSON` column, defining this as a column constraint also looks viable. E.g.:
```sql
CREATE TABLE IF NOT EXISTS test (
    ts TIMESTAMP TIME INDEX,
    b BYTES CHECK (is_json)
```

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,141 @@
...
+);
+
+SELECT json_get(b, 'name') FROM test;
++---------------------+
+| b.name              |
++---------------------+
+| jHl2oDDnPc1i2OzlP5Y |
++---------------------+
+
+SELECT json_get_by_paths(b, 'attributes', 'event_attributes') + 1 FROM test;
```
waynexia, line 44: This looks like a downside to the current proposed implementation. We have to access fields by function. But if we leverage a more "native" type like `Struct` it would be much easier to use like 

```sql
select b.attributes.event_attributes from test
```

And `Struct` is more friendly to parquet

docs/rfcs/2024-08-06-json-datatype.md
WenyXu: > Is it possible to represent this in schema with a clear Struct 

However, the question here is why not let users use PIPELINE to extract all fields from the nested struct? I think the 
dynamic is key for the `JSON` type.

docs/rfcs/2024-08-06-json-datatype.md
WenyXu: The JSON can provide an option for users who don't want to write PIPELINE

docs/rfcs/2024-08-06-json-datatype.md
WenyXu: I think we can provide the syntactic sugar for `JSON`?

e.g., 
```sql
select (b?.attributes?.event_attributes || default_value) from test
```

Can be treats as `get_path_or_default("b?.attributes?.event_attributes", default_value)`

waynexia:
I came up with two scenarios of how JSON would be used by end users:
- Data source uses JSON to encode their data, but the schema to JSON is relatively fixed. Like OTLP protocol or some HTTP header/response. In this case the JSON itself is usually defined with a clear schema.
- Use us to store an arbitrary JSON object. Neither the user nor user of our user knows the schema.

For the first case, we can suggest users decompose the JSON object into table columns. Either by themselves (predefined table schema) or by GreptimeDB (gRPC auto alter or through pipeline). So from the DB side, it's just a wide table with several atomic types. And is not distinguishable from other tables. Advantages like high compression ratio or SQL interface are available out-of-box.

The second case on the contrary, requires a fully dynamic ability to JSON data. We'd prefer to handle the "JSON part" on read (like decoding and/or field accessing) to minimize the affection to write performance. This choice is based on the hypothesis that data without known schema is unlikely to be queried frequently.

Reflecting to the implementation phase, the need & no-need to are:
- Need a mechanism to decompose JSON and trigger ALTER TABLE if necessary. 
  - This operation needs to happen on frontend or before frontend. The Pipeline is a good choice.
- Need a bunch of util functions to handle (undecomposed, the second case) JSON data
- Don't need to be aware of the concrete JSON schema in the storage engine layer. JSON data is either decomposed for the first case or stored as an opaque field for the second case
- Don't need to introduce STRUCT type as it's not a hard requirement. I.e., it can bring a better experience in expressing the schema of decomposed JSON data, but we can still use this without STRUCT type.

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,141 @@
...
+  "c": { "d": 48.28648 }
+}
+```
+This will be parsed at runtime and stored as a corresponding `Struct` type in Arrow:
+```Rust
+Struct(
+    Field("a", Utf8),
+    Field("b", Utf8),
+    Field("c", Struct(Field("d", Float64))),
+)
```
waynexia, line 72: We have to know the concrete type before seeing actual data. Otherwise this seems not quite viable

CookiePieWw:
Thanks for all suggestions! I've changed the rfc to a relatively simple jsonb plan. It's not a detailed plan. I think I can first make some efforts in the direction and then get back and refine this rfc.

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,104 @@
...
+        "attributes": { "event_attributes": 48.28667 }
+    }'
+);
+
+SELECT json_get(b, 'name') FROM test;
++---------------------+
+| b.name              |
++---------------------+
+| jHl2oDDnPc1i2OzlP5Y |
++---------------------+
```
WenyXu, line 42:
```suggestion
SELECT as_string(json_get(b, 'name')) FROM test;
+---------------------+
| b.name              |
+---------------------+
| jHl2oDDnPc1i2OzlP5Y |
+---------------------+
```

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,104 @@
...
+
+SELECT json_get_by_paths_int(b, 'attributes', 'event_attributes') + 1 FROM test;
++-------------------------------+
+| b.attributes.event_attributes |
++-------------------------------+
+|                      49.28667 |
++-------------------------------+
```
WenyXu, line 49:
```suggestion
SELECT as_double(json_get_by_paths(b, 'attributes', 'event_attributes')) + 1 FROM test;
+-------------------------------+
| b.attributes.event_attributes |
+-------------------------------+
|                      49.28667 |
+-------------------------------+
```

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,104 @@
...
+);
+
+SELECT json_get(b, 'name') FROM test;
++---------------------+
+| b.name              |
++---------------------+
+| jHl2oDDnPc1i2OzlP5Y |
++---------------------+
+
+SELECT json_get_by_paths_int(b, 'attributes', 'event_attributes') + 1 FROM test;
```
evenyag, line 44: The return value has double type.

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,104 @@
...
+         JSON Strings ┌────────────┐ JSONB Data ┌────────────┐
+ client ------------->│  Frontend  │----------->│  Datanode  │--> Storage
+                      └────────────┘            └────────────┘
+```
+
+The data of JSON type is represented by `Binary` data type in arrow. There are 2 types of JSON queries: get json elements through keys and compute over json elements.
+
+For the former, the query engine performs queries directly over binary data. We provide functions like `json_get` and `json_get_by_paths` to extract json elements through keys.
+
+For the latter, users need to manually specify the data type of the json elements for computing. We provide functions like `as_int` and `as_double` to decode the binary data into data with specified data type for further computation.
```
evenyag, line 75: Can we use CAST or a `cast` function? How does other databases do?

discord9, line 75: is functions like `json_agg` needed?

docs/rfcs/2024-08-06-json-datatype.md
CookiePieWw: Since the binary json will be presented as strings, we don't need to cast here if there's no computation over strings?

WenyXu: Make sense.

docs/rfcs/2024-08-06-json-datatype.md
CookiePieWw: Postgresql uses a unified cast function like:
```SQL
> CREATE TABLE test (s JSONB);
> INSERT INTO test VALUES('{"a":1}');
> SELECT (s->'a')::int FROM test;
 int4 
------
    1
(1 row
```
MySQL as well.

While databend uses different functions for different types as the rfc.  https://docs.databend.com/sql/sql-functions/semi-structured-functions/as-type

Making use of current cast function is more consistent, but it needs some hack of `arrow_cast` in datafusion. Different functions are just different udfs by ourselves. Which one do you prefer @WenyXu? For me the former seems better for users.

CookiePieWw: I'm not sure if `json_agg` is an urgent need, but I think I can add some of these json-related functions after the basic functions (extract through keys and casts) are ready for use :)

WenyXu: I’m okay with both implementation

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,104 @@
...
+Queries without computation, decoding in frontend:
+                          Decode                     Query
+         JSON Strings ┌────────────┐ JSONB Data ┌──────────────┐
+ client <-------------│  Frontend  │<-----------│  Datafusion  │<-- Storage
+                      └────────────┘            └──────────────┘
+
+Queries with computation, decoding in datafusion:
+                                                                         Query
+         Data of Specified Type ┌────────────┐ Data of Certain Type ┌──────────────┐
+ client <-----------------------│  Frontend  │<---------------------│  Datafusion  │<-- Storage
```
waynexia, line 94:
```
 client <-----------------------│  Frontend  │<---------------------│  Datafusion  │<-- Storage
                                                                                     ^
                                                                                     What's the type here
```

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,104 @@
...
+
+## User Interface
+The feature introduces a new data type for the database, similar to the common JSON type. Data is written as JSON strings and can be queried using functions.
+
+For example:
+```SQL
+CREATE TABLE IF NOT EXISTS test (
+    ts TIMESTAMP TIME INDEX,
+    a INT,
+    b JSON
```
waynexia, line 24: What type should we use? `JSON` or `JSONB` or ` BINARY`

docs/rfcs/2024-08-06-json-datatype.md
CookiePieWw: It's a binary array here. We keep it as binary array before casting for computation.

docs/rfcs/2024-08-06-json-datatype.md
CookiePieWw: I'd like to assume it's about the name of the type? The name `BINARY` is used already. I think if we're not going to add other json types in the future, `JSON` would be a good choice.

evenyag (approved):

waynexia:
For those "util functions", we can use https://github.com/datafusion-contrib/datafusion-functions-json which provides an out-of-box implementation. As well as the binary representation, JSONA.

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,104 @@
...
+For the former, the query engine performs queries directly over binary data. We provide functions like `json_get` and `json_get_by_paths` to extract JSON elements through keys.
+
+For the latter, users need to manually specify the data type of the JSON elements for computing. Users can use `CAST` to convert the JSON elements to the specified data type. Computation without explicit conversion will result in an error.
+
+Queries of `JSON` goes through following steps:
+
+1. Client sends query to frontend, and frontend sends it to datafusion, which is the query engine of GreptimeDB.
+2. Datafusion performs query over binray data of JSONB format, and returns binary data to frontend.
+3. If no computation is needed, frontend directly decodes the binary data to JSON strings and return it to clients.
+4. If computation is needed, the binary data is decoded and converted to the specified data type to perform computation. There's no need for further decoding in the frontend.
```
waynexia, line 82: Not very important but... there are too many errors in this paragraph

WenyXu:
> For those "util functions", we can use https://github.com/datafusion-contrib/datafusion-functions-json which provides an out-of-box implementation. As well as the binary representation, JSONA.

> set of functions for querying JSON strings in DataFusion

This crate just provides functions for querying JSON strings.

docs/rfcs/2024-08-06-json-datatype.md
WenyXu: I think we could start by writing an MVP or doing a POC for this RFC. What do you think, @CookiePieWw? In my opinion, these details can be worked out during implementation.

CookiePieWw: > What do you think, @CookiePieWw?

Sure I'd like to :)

> there are too many errors in this paragraph

TBH I'm not very familiar about the internal interactions of these components. I'd like to get back to this document after having a clearer view of them. Hope I could make it better :)

CookiePieWw:
I happens to trigger some auto-review-request because of misoperation QAQ, sry for that.

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,120 @@
...
+This RFC proposes a method for storing and querying JSON data in the database.
+
+# Motivation
+JSON is widely used across various scenarios. Direct support for writing and querying JSON can significantly enhance the database's flexibility.
+
+# Details
+
+## Storage and Query
+
+The type system of GreptimeDB is based on the types of arrow/datafusion, each type has a corresponding physical type from arrow/datafusion. Thus, the json type is built on top of the `Binary` type, utilizing current implementation of both `Value` and `Vector` of it. JSON type performs the same as Binary type inside the storage layer and query engine.
```
WenyXu, line 18: prefer to use `datatype::value::Value`, `datatype::vectors::BinaryVector`

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,120 @@
...
+
+This also brings 2 problems: insertion and query interface.
+
+## Insertion
+
+User commonly write JSON data as strings. Thus we need to make conversion between string and binary data. There are 2 ways to do this:
+
+1. MySQL and PostgreSQL servers provide auto-conversion between string and JSON data. When a string is inserted into a JSON column, the server will try to parse the string as JSON data and convert it to binary data of JSON type. The non-JSON string will be rejected.
```
WenyXu, line 26:
```suggestion
1. MySQL and PostgreSQL servers provide auto-conversion between string and JSON data. When a string is inserted into a JSON column, the server will try to parse the string as JSON data and convert it to binary data of JSON type(JSONB). The non-JSON string will be rejected.
```

WenyXu: Can you explain the difference between SQL insertion and inserting through the prepared statement? You can draw a simple graph to explain the internal details.

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,120 @@
...
+    parse_json('{
+        "name": "jHl2oDDnPc1i2OzlP5Y",
+        "timestamp": "2024-07-25T04:33:11.369386Z",
+        "attributes": { "event_attributes": 48.28667 }
+    }')
+);
+```
+Are both valid.
+
+For former the conversion is done by the server, while for the latter the conversion is done by the query engine.
```
WenyXu, line 60: > done by the query engine.

It's a little weird.

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Query Interface
+
+Correspondingly, users prefer to display JSON data as strings. Thus we need to make conversion between binary data and string data. There are alsol 2 ways to do this: auto-conversions on MySQL and PostgreSQL servers, and function `json_to_string`.
```
WenyXu, line 64:
```suggestion
Correspondingly, users prefer to display JSON data as strings. Thus we need to make conversion between binary data and string data. There are also 2 ways to do this: auto-conversions on MySQL and PostgreSQL servers, and function `json_to_string`.
```

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,197 @@
...
+ Client <----------------------│  Server  │<------------------│ Query Engine │<----------------- Storage
+                               └──────────┘                   └──────────────┘
+(Server identifies JSON type and performs auto-conversion based on column metadata)
+
+Query through json_to_string function:
+                                                               Scan & Decode
+        String(Serialized JSON)┌──────────┐Arrow Binary(JSONB)┌──────────────┐Arrow Binary(JSONB)
+ Client <----------------------│  Server  │<------------------│ Query Engine │<----- Storage
+                               └──────────┘                   └──────────────┘
+                                              (Conversion is performed by UDF inside Query Engine)
```
WenyXu, line 145:
```suggestion
                                                               Scan & Decode
                String         ┌──────────┐String(Serialized JSON)┌──────────────┐Arrow Binary(JSONB)
 Client <----------------------│  Server  │<----------------------│ Query Engine │<----- Storage
                               └──────────┘                       └──────────────┘
                                              (Conversion is performed by UDF inside Query Engine)
```

docs/rfcs/2024-08-06-json-datatype.md
```diff
@@ -0,0 +1,197 @@
...
+        "attributes": { "event_attributes": 48.28667 }
+    }'
+);
+
+SELECT json_get_int(b, 'name') FROM test;
++---------------------+
+| b.name              |
++---------------------+
+| jHl2oDDnPc1i2OzlP5Y |
++---------------------+
```
waynexia, line 177: Is this an integer?

WenyXu: I think the SQL should be🤪
```rust
SELECT json_get_string(b, 'name') FROM test;
```

PR #4521, opened 2024-08-07
fix: larger stack size in debug mode

discord9:

set larger stack size for debug mode, so it doesn't overflow for deep recursive functions. After setting this, `cargo sqlness` doesn't overflow in windows(it can relably reproduce overflow before this fix)

https://github.com/GreptimeTeam/greptimedb/actions/runs/10261340360/job/28388920598

## Checklist

### DIFF
--- a/src/cmd/src/bin/greptime.rs
+++ b/src/cmd/src/bin/greptime.rs
@@ -72,17 +72,13 @@ fn main() -> Result<()> {
         .name("main_spawn".to_string())
         .stack_size(8 * 1024 * 1024)
         .spawn(|| {
-            let body = async {
-                setup_human_panic();
-                start(Command::parse()).await
-            };
             {
                 tokio::runtime::Builder::new_multi_thread()
                     .thread_stack_size(8 * 1024 * 1024)
                     .enable_all()
                     .build()
                     .expect("Failed building the Runtime")
-                    .block_on(body)
+                    .block_on(main_body())
             }
         })
         .context(cmd::error::SpawnThreadSnafu)?
@@ -93,6 +89,10 @@ fn main() -> Result<()> {
 #[cfg(not(debug_assertions))]
 #[tokio::main]
 async fn main() -> Result<()> {
+    main_body().await
+}
+
+async fn main_body() -> Result<()> {
     setup_human_panic();
     start(Command::parse()).await
 }
--- a/src/cmd/src/bin/greptime.rs
+++ b/src/cmd/src/bin/greptime.rs
@@ -15,11 +15,10 @@
 #![doc = include_str!("../../../../README.md")]
 
 use clap::{Parser, Subcommand};
-use cmd::error::{Result, SpawnThreadSnafu};
+use cmd::error::Result;
 use cmd::options::GlobalOptions;
 use cmd::{cli, datanode, flownode, frontend, metasrv, standalone, App};
 use common_version::version;
-use snafu::ResultExt;
 
 #[derive(Parser)]
 #[command(name = "greptime", author, version, long_version = version(), about)]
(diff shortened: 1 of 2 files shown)


src/cmd/src/bin/greptime.rs
```diff
@@ -62,10 +62,36 @@ enum SubCommand {
 #[global_allocator]
 static ALLOC: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 
-#[tokio::main]
-async fn main() -> Result<()> {
-    setup_human_panic();
-    start(Command::parse()).await
+fn main() -> Result<()> {
```
evenyag, line 65: Don't need to return `Result` now.

src/cmd/src/bin/greptime.rs
```diff
@@ -62,10 +62,36 @@ enum SubCommand {
...
+    // Set the stack size to 8MB for the thread so it wouldn't crash on large stack usage in debug mode
+    // see https://github.com/GreptimeTeam/greptimedb/pull/4317
+    // and https://github.com/rust-lang/rust/issues/34283
+    let builder = std::thread::Builder::new().name("main_larger_stack".to_string());
```
evenyag, line 69: Can we use `main` for the thread name?

discord9: > Can we use `main` for the thread name?

Will cause confusion since the old main function is also named `main` thread

src/cmd/src/bin/greptime.rs
```diff
@@ -62,10 +63,36 @@ enum SubCommand {
...
+            let body = async {
+                setup_human_panic();
+                start(Command::parse()).await
+            };
+            #[allow(clippy::expect_used, clippy::diverging_sub_expression)]
+            {
+                let mut builder = tokio::runtime::Builder::new_multi_thread();
+
+                #[cfg(debug_assertions)]
+                let builder = builder.thread_stack_size(8 * 1024 * 1024);
```
evenyag, line 84: If we set the stack size of the runtime, do we need to set the stack size for the main thread?

discord9: > If we set the stack size of the runtime, do we need to set the stack size for the main thread?

still need to. overflow happen in main thread before this fix, meaning tokio is running a large stack using main thread, builder for runtime only set stack size for newly spawned thread not the thread runtime is `block_on` anyway, so the main problem is setting larger stack size for main thread(but rust's main thread seems to be default to be 2MB), setting stack size for runtime is just a precaution

discord9: > > If we set the stack size of the runtime, do we need to set the stack size for the main thread?
> 
> still need to. overflow happen in main thread before this fix, meaning tokio is running a large stack using main thread, builder for runtime only set stack size for newly spawned thread not the thread runtime is `block_on` anyway, so the main problem is setting larger stack size for main thread(but rust's main thread seems to be default to be 2MB), setting stack size for runtime is just a precaution(previous it's 2MB by default though, maybe setting to 4 MB is more reasonable?)

src/cmd/src/bin/greptime.rs
```diff
@@ -62,6 +62,35 @@ enum SubCommand {
...
+    // Set the stack size to 8MB for the thread so it wouldn't overflow on large stack usage in debug mode
+    // see https://github.com/GreptimeTeam/greptimedb/pull/4317
+    // and https://github.com/rust-lang/rust/issues/34283
+    std::thread::Builder::new()
+        .name("main_spawn".to_string())
+        .stack_size(8 * 1024 * 1024)
+        .spawn(|| {
+            let body = async {
+                setup_human_panic();
+                start(Command::parse()).await
```
evenyag, line 77: non-blocking: Maybe we can wrap the body into a function.
```
async fn main_body() -> Result {
}
```

src/cmd/src/error.rs
```diff
@@ -305,6 +305,12 @@ pub enum Error {
         error: std::io::Error,
     },
 
+    #[snafu(display("Failed to spawn thread"))]
+    SpawnThread {
+        #[snafu(source)]
+        error: std::io::Error,
+    },
```
evenyag, line 312: Do we need to add `#[cfg(debug_assertions)]` for it?

evenyag (approved):
Others LGTM

src/cmd/src/error.rs
discord9: > Do we need to add `#[cfg(debug_assertions)]` for it?

Other function in cmd crate might need it anyway and cfg in enum and match arm is ugly so

fengjiachun (approved):
LGTM

PR #4525, opened 2024-08-07
fix: rollback only if dropping the metric physical table fails

WenyXu:

Rollback metadata only if dropping the metric physical table fails

## Checklist

### DIFF
--- a/src/common/meta/src/ddl/drop_table/metadata.rs
+++ b/src/common/meta/src/ddl/drop_table/metadata.rs
@@ -12,8 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use common_catalog::format_full_table_name;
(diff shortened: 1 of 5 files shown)


src/common/meta/src/ddl/drop_table/metadata.rs
```diff
@@ -29,6 +33,28 @@ impl DropTableProcedure {
         self.data.physical_region_routes = physical_table_route_value.region_routes;
         self.data.physical_table_id = Some(physical_table_id);
 
+        let table_info_value = self
```
fengjiachun, line 36: If there is a logical table (task.table_id != physical_table_id), can we avoid to do this check?

fengjiachun (approved):
LGTM

evenyag (approved):

PR #4531, opened 2024-08-08
chore: make mysql server version changable

MichaelScofield:

because our customer might report some security issues just by the mysql server version. trying to make them happy. now the default returned 8.4.2 is the latest mysql lts version

## Checklist

### DIFF
--- a/src/servers/src/mysql/handler.rs
+++ b/src/servers/src/mysql/handler.rs
@@ -221,6 +221,10 @@ impl MysqlInstanceShim {
 impl<W: AsyncWrite + Send + Sync + Unpin> AsyncMysqlShim<W> for MysqlInstanceShim {
     type Error = error::Error;
 
+    fn version(&self) -> String {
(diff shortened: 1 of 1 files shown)


sunng87 (approved):

src/servers/src/mysql/handler.rs
```diff
@@ -221,6 +221,10 @@ impl MysqlInstanceShim {
 impl<W: AsyncWrite + Send + Sync + Unpin> AsyncMysqlShim<W> for MysqlInstanceShim {
     type Error = error::Error;
 
+    fn version(&self) -> String {
+        std::env::var("GREPTIMEDB_MYSQL_SERVER_VERSION").unwrap_or_else(|_| "8.4.2".to_string())
```
evenyag, line 225: non-blocking: Maybe we should document this env in our docs.

evenyag (approved):

src/servers/src/mysql/handler.rs
MichaelScofield: Well, this is the "dark side" of greptimedb, let's not create more confusing to our users, just keep it deep inside our codes.

PR #4546, opened 2024-08-12
chore: Helper function to convert `Vec<Value>` to VectorRef

discord9:

added a helper function to convert `Vec<Value>` to `VectorRef`, and it's test

## Checklist

### DIFF
--- a/src/datatypes/src/duration.rs
+++ b/src/datatypes/src/duration.rs
@@ -111,6 +111,24 @@ macro_rules! define_duration_with_unit {
                     val.0.value()
                 }
             }
+
+            impl TryFrom<Value> for Option<[<Duration $unit>]> {
+                type Error = $crate::error::Error;
+
+                #[inline]
+                fn try_from(from: Value) -> std::result::Result<Self, Self::Error> {
+                    match from {
+                        Value::Duration(v) if v.unit() == TimeUnit::$unit => {
+                            Ok(Some([<Duration $unit>](v)))
+                        },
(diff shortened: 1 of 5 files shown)


src/datatypes/src/vectors/helper.rs
```diff
@@ -367,6 +371,185 @@ impl Helper {
         })
     }
 
+    /// Try to cast an vec of values into vector, fail if type is not the same across all values.
+    pub fn try_from_row_into_vector(row: Vec<Value>, dt: &ConcreteDataType) -> Result<VectorRef> {
```
evenyag, line 375: I remember that the `MutableVector` trait has a `try_push_value_ref()` method. Can we use the mutable vector to build the `VectorRef` from `Vec<Value>`?

src/datatypes/src/duration.rs
```diff
@@ -111,6 +111,24 @@ macro_rules! define_duration_with_unit {
                     val.0.value()
                 }
             }
+
+            impl TryFrom<Value> for Option<[<Duration $unit>]> {
```
evenyag, line 115: Do we still need these try from?

evenyag (approved):
LGTM

src/datatypes/src/duration.rs
discord9: > Do we still need these try from?

Not really, but since other variant seems also `impl TryFrom<Value>` so for the sake of completeness still keep them?

WenyXu (approved):

PR #4564, opened 2024-08-14
perf: set simple filter on primary key columns to exact filter

waynexia:

Simple filter on primary key columns are precisely evaluated and need not to be evaluated again on Filter plan.

This patch has resulted in good performance improvement on my current test case. Before:
image

After:

image

## Checklist

### DIFF
--- a/src/query/src/dummy_catalog.rs
+++ b/src/query/src/dummy_catalog.rs
@@ -182,7 +182,7 @@ impl TableProvider for DummyTableProvider {
         let supported = filters
             .iter()
             .map(|e| {
-                // Simple filter on primary key columns are percisely evaluated.
+                // Simple filter on primary key columns are precisely evaluated.
                 if let Some(simple_filter) = SimpleFilterEvaluator::try_new(e) {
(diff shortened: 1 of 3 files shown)


src/query/src/dummy_catalog.rs
```diff
@@ -177,7 +179,27 @@ impl TableProvider for DummyTableProvider {
...
+        let supported = filters
+            .iter()
+            .map(|e| {
+                // Simple filter on primary key columns are precisely evaluated.
+                if let Some(simple_filter) = SimpleFilterEvaluator::try_new(e) {
+                    if self
+                        .metadata
+                        .column_by_name(simple_filter.column_name())
+                        .and_then(|c| (c.semantic_type == SemanticType::Tag).then_some(()))
+                        .is_some()
```
evenyag, line 191: Looks like we can also use `map()` with `unwrap_or(false)`. But `then_some()` is fine.

evenyag (approved):

fengjiachun (approved):

PR #4569, opened 2024-08-15
feat: able to handle concurrent region edit requests

MichaelScofield:

Currently the region edit requests can only be handled one by one. This is because once a region is processing the request, it's set to the "Editing" state, which rejects other concurrent edit requests (require "Writable" state). 

This PR adds a "region edit queue" for temporarily storing the concurrent region edit requests. When concurrent requests come in, and the worker loop finds the region is in "Editing" state, it stores the requests in the queue. Once the current edit request is done, the worker loop checks the queue for a remaining edit request and handles it.

## Checklist

### DIFF
--- a/src/mito2/src/engine/edit_region_test.rs
+++ b/src/mito2/src/engine/edit_region_test.rs
@@ -18,8 +18,7 @@ use object_store::ObjectStore;
 use store_api::region_engine::RegionEngine;
 use store_api::region_request::RegionRequest;
 use store_api::storage::RegionId;
-use tokio::sync::Semaphore;
-use tokio::task::JoinSet;
+use tokio::sync::Barrier;
 
 use crate::config::MitoConfig;
 use crate::engine::MitoEngine;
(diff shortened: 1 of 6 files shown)


src/mito2/src/engine/edit_region_test.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    let mut join_set = JoinSet::new();
+    // This semaphore is used to coordinate the tasks, making them started at roughly the same time.
+    let s = Arc::new(Semaphore::new(0));
+    for task in tasks {
+        join_set.spawn({
+            let s = s.clone();
+            let engine = engine.clone();
+            async move {
+                let _permit = s.acquire().await.unwrap();
```
evenyag, line 111: You can use [Barrier](https://docs.rs/tokio/latest/tokio/sync/struct.Barrier.html)

src/mito2/src/engine/edit_region_test.rs
```diff
@@ -0,0 +1,123 @@
...
+                task.edit_region(engine).await;
+            }
+        });
+    }
+    s.add_permits(tasks_count);
+    while join_set.join_next().await.is_some() {}
```
evenyag, line 117: Can we use `join_all()`?

MichaelScofield:
@evenyag PTAL

evenyag (approved):

sunng87 (approved):

PR #4580, opened 2024-08-17
fix: table resolving logic related to pg_catalog

### LINKED ISSUE #4543: Make pg_catalog schema visible only when client connect from Postgres protocol

API improvement

This was a comment from the initial PR of pg_catalog feature #4332 . This `pg_catalog` is designed to work with postgresql client tools like psql to get meta information of the database. It should only be visible when client connects from postgresql protocol. We should not see it from our mysql clients, and http dashbaord.

### Implementation challenges

Check the `QueryContext` to see current whether current client is from postgresql port or not.

J0HN50N133:
refer to
https://github.com/GreptimeTeam/greptimedb/issues/3560#issuecomment-2287794348 and #4543

close #4543 and https://github.com/GreptimeTeam/greptimedb/issues/3560#issuecomment-2287794348.

This is a quiet huge change. Now query to catalog_manager have to bring a catalog_protocol arguments. So that, the catalog_manager can decide the table resolve logic according to the protocol.

## Checklist


J0HN50N133:
@sunng87 could you check this pr. It a quiet big change and I'm not sure is it worth to do for such small feature? Or we could find a way that's affect less modules? Also, I wonder is it possible to specify protocol to pg in sqlness. The modification makes the sqlness test of pg_catalog unusable.

sunng87:
Thank you so much for your effort. 

We do have protocol information in `QueryContext`'s channel field. Instead of adding new protocol type, I would suggest to reuse this `channel` enum. Another suggestion is to pass the whole `QueryContextRef` to catalog methods so we have more context information for future extensibility.

I'm not 100% sure if there is always a `QueryContext` when calling catalog methods, perhaps we can use `Option<QueryContextRef>` for catalog methods when it's missing.

src/catalog/src/kvbackend/manager.rs
```diff
@@ -320,12 +345,21 @@ struct SystemCatalog {
...
+            // information_schema visible for mysql protocol and other
+            CatalogProtocol::MySQL => {
+                vec![INFORMATION_SCHEMA_NAME.to_string()]
+            }
+            CatalogProtocol::Other => {
+                vec![
+                    PG_CATALOG_NAME.to_string(),
+                    INFORMATION_SCHEMA_NAME.to_string(),
+                ]
+            }
```
sunng87, line 361: `information_schema` is needed for both postgres and mysql. This logic can be changed to appending `pg_catalog` when it's calling from postgres

src/catalog/src/kvbackend/manager.rs
```diff
@@ -339,8 +373,14 @@ impl SystemCatalog {
...
-    fn schema_exists(&self, schema: &str) -> bool {
-        schema == INFORMATION_SCHEMA_NAME || schema == PG_CATALOG_NAME
+    fn schema_exists(&self, schema: &str, catalog_protocol: CatalogProtocol) -> bool {
+        match catalog_protocol {
+            CatalogProtocol::PostgreSQL => schema == PG_CATALOG_NAME,
+            CatalogProtocol::MySQL => schema == INFORMATION_SCHEMA_NAME,
+            CatalogProtocol::Other => {
+                schema == PG_CATALOG_NAME || schema == INFORMATION_SCHEMA_NAME
+            }
+        }
```
sunng87, line 383: same as above

J0HN50N133:
> Thank you so much for your effort.
> 
> We do have protocol information in `QueryContext`'s channel field. Instead of adding new protocol type, I would suggest to reuse this `channel` enum. Another suggestion is to pass the whole `QueryContextRef` to catalog methods so we have more context information for future extensibility.
> 
> I'm not 100% sure if there is always a `QueryContext` when calling catalog methods, perhaps we can use `Option<QueryContextRef>` for catalog methods when it's missing.

I'm pretty sure that there is no QueryContext in some call sites of Catalog's function. But it's a good idea to use Option<QueryContext>. I'll modify the implementation.

src/catalog/src/kvbackend/manager.rs
J0HN50N133: We can do even harder. Since `pg_catalog` is only visible for pg protocol, we could only build it when we need to start a pg frontend.

sunng87: Sounds reasonable. Can we get postgres options when building system catalog? But since the memory footprint should be small if we don't query it, I think it's acceptable to build it unconditionally.

J0HN50N133: I'will figure it out. If we could, why not?

J0HN50N133:
@sunng87 I've modified the implementation , re-using the `Channel` type. But I got some problems with sqlness test. Since `pg_catalog` is only visible while using pg protocol, it's not able to test the behaviour of pg_catalog via sqlness. Of course I could test them by integration test, but it's somehow unnessary and inconvinent. To handle this, I suggest to control the visibility of `pg_catalog` with a private session variable named `__PG_CATALOG_VISIBLE`. It's value is false by default, and we set it to true when creating a pg connection. Then we could do `SET __PG_CATALOG_VISIBLE = true` in sqlness.

sunng87:
Sorry for being late. I have been travelling these days. The first approach sounds good to me. We also have some MySQL specific scenario to cover.

Another approach comes to my mind is to provide a http parameter to override the `Channel` to postgres or mysql. We can also add a sqlness directive to tell sqlness runner to use this parameter when sending SQL commands.

But I think it can be implemented in a separated patch.

J0HN50N133:
Can we just push forward this pr and keep the the pg specific behaviors untested for a while? I will try to improve sqlness in another patch.

sunng87:
I think so. Current tests should guarantee that it will not break existed functionalities.

src/catalog/src/lib.rs
```diff
@@ -44,26 +46,52 @@ pub trait CatalogManager: Send + Sync {
 
     async fn catalog_names(&self) -> Result<Vec<String>>;
 
-    async fn schema_names(&self, catalog: &str) -> Result<Vec<String>>;
+    async fn schema_names(&self, catalog: &str, channel: Channel) -> Result<Vec<String>>;
```
sunng87, line 49: Can we use `Option<QueryContextRef>` for all `CatalogManager` functions? I think it makes more sense to provide whole context instead of just `channel`. It make behaviour of `CatalogManager` more extensible.

src/catalog/src/system_schema/information_schema/region_peers.rs
```diff
@@ -176,9 +177,9 @@ impl InformationSchemaRegionPeersBuilder {
 
         let predicates = Predicates::from_scan_request(&request);
 
-        for schema_name in catalog_manager.schema_names(&catalog_name).await? {
+        for schema_name in catalog_manager.schema_names(&catalog_name, Mysql).await? {
```
sunng87, line 180: I think we should try to avoid hard code `Mysql` here. That's also why I suggest using `Option<QueryContextRef>` for this.

J0HN50N133: I've tried use Option< QueryContext> though. We also need to query the CatalogManager in `pg_catalog` module. Same, we don't have QueryContext, so we have to pass `None`. There ambiguity comes. We don't know what to do when we get none(do pg resolving or mysql resolving)?

sunng87: My idea is we don't need to add default `query_context` when it's not from a query path. I'm afraid it will introduce more confusion with future maintenance. When the query context is `None`, we just fallback to default behaviour.

At the moment, there are only default behaviour and postgres behaviour:

- default: return information_schema
- postgres: return pg_catalog and information_schema, when table name equals `pg_class` (and maybe more like `pg_namespace`), return `<catalog>.pg_catalog.pg_class` table

J0HN50N133: The point is we also don't have `query_context` whiling query catalog_manager inside pg_catalog. It's not correct to use the default behaviour under this scenario.

sunng87: oh, I got your point. When we are calling catalog_manager from pg_catalog, is there a chance we have `QueryContext` in the context that we can pass all the way to catalog_manager methods. If I understand correctly, the call must be initiated from a postgres query.

J0HN50N133: Sorry for replying late. Unfortunately, it may hard to have `QueryContext` when calling catalog_manager methods. We only have `ScanRequest` when constructing a SystemTable to data stream. Check:
https://github.com/GreptimeTeam/greptimedb/blob/8eda36bfe3661c6355ef7ef71c2e6529e2166b53/src/catalog/src/system_schema.rs#L106

J0HN50N133:
@sunng87 This pr is now ready for reviewed.

src/catalog/src/system_schema/information_schema.rs
```diff
@@ -299,3 +302,12 @@ where
...
 }
+
+/// Provide query context to call the [`CatalogManager`]'s method.
+static MYSQL_QUERY_CTX: LazyLock<QueryContext> = LazyLock::new(|| {
+    QueryContext::with_channel(DEFAULT_CATALOG_NAME, DEFAULT_SCHEMA_NAME, Channel::Mysql)
+});
+
+fn query_ctx() -> Option<&'static QueryContext> {
+    Some(&MYSQL_QUERY_CTX)
+}
```
sunng87, line 313: I think the only tricky point is here. I still suggest to use `None` query context for `information_schema`'s builder where there is no real query context available. 

Different from `pg_catalog`, which is dedicated to postgres, `information_schema` is a global concept that available to all protocols like mysql, postgresql and others we offer. So I think it's safe to use `None` and it's default behaviour here. 

Using `Mysql` channel may introduce misunderstanding in future. It's hard to distinguish whether it's real mysql or placeholder.

sunng87 (commented):
Thank you for the effort. It looks good to me mostly except the one I left comment.

src/catalog/src/system_schema/information_schema.rs
J0HN50N133: fair enough. done.

sunng87:
@J0HN50N133 seems we still have a test to fix.

@killme2008 @fengjiachun @WenyXu @evenyag PTAL when you got time.

sunng87 (approved):

tisonkun (approved):

PR #4604, opened 2024-08-22
fix: build in asahi linux

fengys1996:

I recently had some problems compiling greptimedb on an [asahi linux](https://asahilinux.org/) platform.

The error reported is as follows:
```
error: linker `aarch64-linux-gnu-gcc` not found
  |
  = note: No such file or directory (os error 2)
```

After investigation, found that linker was specified in [cargo's configuration]( https://github.com/GreptimeTeam/greptimedb/blob/main/.cargo/config.toml#L1-L2).

If not cross-compiling, it seems that we don't need to specify a linker.

## Checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,2 @@
-[target.aarch64-unknown-linux-gnu]
-linker = "aarch64-linux-gnu-gcc"
-
 [alias]
 sqlness = "run --bin sqlness-runner --"


sunng87:
@zyy17 could you confirm if this setting is still in use nowadays? I saw it's from early days.

zyy17:
Although it's not being used now, I think we should keep this configuration. Maybe we need it in the future.

evenyag:
What's the purpose of setting this linker?

fengys1996:
> What's the purpose of setting this linker?

I guess it might be used when cross-compiling.

tisonkun:
> If not cross-compiling, it seems that we don't need to specify a linker.

It's possible used by Andorid platform where it is ARM arch. Please check if it's the case.

For GCC linker on Asahi Linux, you should be able to install it manually.

PR #4605, opened 2024-08-22
fix: incremental compilation always compile the common-version crate 

fengys1996:

I found a problem:
Each incremental build compiles the common-version crate and its dependent crates.

This causes each incremental build to take a very long time and puts pressure on my limited disk space.

## Checklist

### DIFF
--- a/src/common/version/build.rs
+++ b/src/common/version/build.rs
@@ -19,7 +19,7 @@ use build_data::{format_timestamp, get_source_time};
 use shadow_rs::{CARGO_METADATA, CARGO_TREE};
 
 fn main() -> shadow_rs::SdResult<()> {
-    println!("cargo:rerun-if-changed=.git/refs/heads");
+    println!("cargo:rerun-if-changed=../../../.git/refs/heads");
     println!(
         "cargo:rustc-env=SOURCE_TIMESTAMP={}",
         if let Ok(t) = get_source_time() {


src/common/version/build.rs
```diff
@@ -19,7 +19,7 @@ use build_data::{format_timestamp, get_source_time};
 use shadow_rs::{CARGO_METADATA, CARGO_TREE};
 
 fn main() -> shadow_rs::SdResult<()> {
-    println!("cargo:rerun-if-changed=.git/refs/heads");
+    println!("cargo:rerun-if-changed=../../../.git/refs/heads");
```
tisonkun, line 22: Why this is a fix?

tisonkun: Found https://github.com/rust-lang/cargo/issues/4213

tisonkun (approved):

MichaelScofield (approved):

src/common/version/build.rs
fengys1996: I'm not very familiar with the gix library and not sure if it will work with git submodules.

It's best to keep it simple. I can improve it once I'm sure it works.

tisonkun: Record a code snippet for handling submodule:

```rust
fn main() {
    let mut current = std::path::PathBuf::new().join(env!("CARGO_MANIFEST_DIR"));

    while let Ok((dir, _)) = gix_discover::upwards(&current) {
        let dir = match dir {
            gix_discover::repository::Path::LinkedWorkTree { work_dir, .. } => {
                current = work_dir.parent().unwrap().to_path_buf();
                continue;
            }
            gix_discover::repository::Path::WorkTree(dir) => dir,
            gix_discover::repository::Path::Repository(dir) => dir,
        };

        let git_refs_heads = dir.join(".git/refs/heads");
        println!("cargo::rerun-if-changed={}", git_refs_heads.display());
        break;
    }
}
```

@Byron hopefully this is the correct way to "discover the git repository path that is not a submodule".

Byron: I just re-read the [API docs](https://github.com/Byron/gitoxide/blob/e9295dc7a98eaedddb6941daa39b80412164360d/gix-discover/src/repository.rs#L6-L13) and could only see that it *can* be a submodule, but it could also be another worktree checkout.

`gix-discover` can only give an idea of what it probably is, and to know for sure one would have to open the repository and perform more checks.

However, with modern Git repositories, the `git_dir` of `LinkedWorkTree` should look like `…/modules/<name>`, so one could additionally try to find the `modules` subdirectory to be reasonably sure it's a submodule worktree you are looking at.

PR #4629, opened 2024-08-27
fix: fallback to window size in manifest

evenyag:

This PR uses the compaction time window as the partition bucket if the region doesn't provide the time window in the compaction options.

Otherwise, the region may always infer the time bucket after re-opening the region.

## Checklist


fengjiachun (approved):
LGTM

waynexia (approved):

PR #4635, opened 2024-08-28
feat: initialize partition range from ScanInput

waynexia:

Initialize `PartitionRange`s from `ScanInput` for planner. It only contains a rough time range and number of rows at present.

The next step is to modify the corresponding logic about initializing `StreamContext`.

This patch also fixes the problem that `RegionScanExec` always generates only 1 partition.

## Checklist

### DIFF
--- a/src/mito2/src/read/scan_region.rs
+++ b/src/mito2/src/read/scan_region.rs
@@ -25,7 +25,7 @@ use common_time::range::TimestampRange;
 use common_time::Timestamp;
 use datafusion::physical_plan::DisplayFormatType;
 use smallvec::SmallVec;
-use store_api::region_engine::RegionScannerRef;
+use store_api::region_engine::{PartitionRange, RegionScannerRef};
 use store_api::storage::{ScanRequest, TimeSeriesRowSelector};
 use table::predicate::{build_time_range_predicate, Predicate};
 use tokio::sync::{mpsc, Mutex, Semaphore};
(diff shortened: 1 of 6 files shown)


src/mito2/src/read/seq_scan.rs
```diff
@@ -67,10 +67,11 @@ impl SeqScan {
     /// Creates a new [SeqScan].
     pub(crate) fn new(input: ScanInput) -> Self {
         let parallelism = input.parallelism.parallelism.max(1);
-        let properties = ScannerProperties::default()
+        let mut properties = ScannerProperties::default()
```
evenyag, line 70: Do we need to also modify unordered_scan?

evenyag (approved):

src/mito2/src/read/scan_region.rs
```diff
@@ -705,6 +705,37 @@ impl ScanInput {
...
     }
+
+    /// Retrieves [`PartitionRange`] from memtable and files
+    pub(crate) fn partition_ranges(&self) -> Vec<PartitionRange> {
+        let mut id = 0;
+        let mut container = Vec::with_capacity(self.memtables.len() + self.files.len());
+
+        for memtable in &self.memtables {
+            let range = PartitionRange {
+                // TODO: filter out empty memtables in the future.
```
fengjiachun, line 716:
```suggestion
                // TODO(ruihang): filter out empty memtables in the future.
```

😄

fengjiachun (approved):
LGTM

Issue #4480, opened 2024-08-01
Add debug level traces in mito engine


WenyXu:
### What type of enhancement is this?

Other

### What does the enhancement do?

GreptimeDB exports all collected spans using the gRPC-based OTLP protocol ([see also](https://docs.greptime.com/user-guide/operations/tracing#tutorial-use-jaeger-to-trace-greptimedb)). 

However, current traces are still not good enough for performance debugging, consider adding more detailed spans in the mito engine. e.g., The time reading bytes from S3, The time building reader, etc.

Michaelin007:
Hi @WenyXu , I would like to work on this. Can you provide more insight into the problem?

WenyXu:
> Hi @WenyXu , I would like to work on this. Can you provide more insight into the problem?

Hey, I suggest starting with our [tracing tutorials](https://docs.greptime.com/user-guide/operations/tracing/#tutorial-use-jaeger-to-trace-greptimedb). Run a query and check out the traces—you'll probably notice they’re a bit too coarse. For example, we don't really see how much time is spent on things like downloading files or building readers. This issue is about improving that by adding more detailed, debug-level traces in the Mito engine, especially for queries.

PR #4643, opened 2024-08-29
feat: add more spans to mito engine

lyang24:

https://github.com/GreptimeTeam/greptimedb/issues/4480

This commit adds more spans into mito engine to give better insights on debugging slow queires.
For reads i enhanced spans on reader building time.

## Checklist

### DIFF
--- a/src/mito2/src/read/scan_region.rs
+++ b/src/mito2/src/read/scan_region.rs
@@ -20,7 +20,7 @@ use std::time::{Duration, Instant};
 
 use common_error::ext::BoxedError;
 use common_recordbatch::SendableRecordBatchStream;
-use common_telemetry::{debug, error, warn};
+use common_telemetry::{debug, error, tracing, warn};
 use common_time::range::TimestampRange;
 use common_time::Timestamp;
 use datafusion::physical_plan::DisplayFormatType;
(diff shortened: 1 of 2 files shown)


MichaelScofield (approved):

WenyXu:
How about setting the tracing level to Debug? i.e., `#[instrument(level = Level::DEBUG, ...)]`

lyang24:
> How about setting the tracing level to Debug? i.e., `#[instrument(level = Level::DEBUG, ...)]`

done

discord9:
Calls that occur very frequently and are typically not time-consuming, such as `write_to_wal` or `next_batch`, can be set to a tracing level of TRACE to avoid excessive verbosity during debug mode, or those traces could be completely removed to prevent pref regression(Removing might be unnecessary since tracing is essentially a comparison to an integer when not enabled, and branch prediction could minimize the time cost significantly, might only see 0.5%~1% regression in speed)

lyang24:
> Calls that occur very frequently and are typically not time-consuming, such as `write_to_wal` or `next_batch`, can be set to a tracing level of TRACE to avoid excessive verbosity during debug mode, or those traces could be completely removed to prevent pref regression(Removing might be unnecessary since tracing is essentially a comparison to an integer when not enabled, and branch prediction could minimize the time cost significantly, might only see 0.5%~1% regression in speed)

thanks for the feedback i agree that adding spans on frequently called methods might not be very helpful and it could be verbose and leads to potential perf regression so i removed them. Please let me know if there are places in mito2 engine that should be traced that i missed in this pr.

WenyXu (approved):

PR #4646, opened 2024-08-30
fix: correct otlp endpoint formatting

WenyXu:

Correct otlp endpoint formatting

## Checklist

### DIFF
--- a/src/common/telemetry/src/logging.rs
+++ b/src/common/telemetry/src/logging.rs
@@ -234,7 +234,13 @@ pub fn init_global_logging(
                     opentelemetry_otlp::new_exporter().tonic().with_endpoint(
                         opts.otlp_endpoint
                             .as_ref()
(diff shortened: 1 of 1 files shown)


waynexia (approved):

src/common/telemetry/src/logging.rs
```diff
@@ -234,7 +234,13 @@ pub fn init_global_logging(
                     opentelemetry_otlp::new_exporter().tonic().with_endpoint(
                         opts.otlp_endpoint
                             .as_ref()
-                            .map(|e| format!("http://{}", e))
+                            .map(|e| {
```
zyy17, line 237: question: Do we have `https` scenarios?

WenyXu: I think we should leave the choice to the user

fengjiachun (approved):
LGTM

PR #4648, opened 2024-08-30
feat: schedule compaction when adding sst files by editing region

MichaelScofield:

as title

## Checklist

### DIFF
--- a/src/mito2/src/engine/edit_region_test.rs
+++ b/src/mito2/src/engine/edit_region_test.rs
@@ -15,6 +15,7 @@
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
+use common_time::util::current_time_millis;
 use object_store::ObjectStore;
 use store_api::region_engine::RegionEngine;
 use store_api::region_request::RegionRequest;
@@ -22,6 +23,7 @@ use store_api::storage::RegionId;
 use tokio::sync::{oneshot, Barrier};
 
 use crate::config::MitoConfig;
+use crate::engine::flush_test::MockTimeProvider;
 use crate::engine::listener::EventListener;
 use crate::engine::MitoEngine;
 use crate::manifest::action::RegionEdit;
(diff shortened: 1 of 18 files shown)


src/mito2/src/worker/handle_manifest.rs
```diff
@@ -165,6 +169,26 @@ impl<S> RegionWorkerLoop<S> {
                 self.handle_region_edit(request).await;
             }
         }
+
+        if need_compaction {
```
killme2008, line 173: I think it should be a request option, it's not a common practice to schedule compaction every time when editing regions.

MichaelScofield: After discussed with @evenyag , we decide to add a config value to determine the time interval of two successive compactions. To align with the old behavior, the default value of this config is 0: no restrictions on how compactions are scheduled. In this editing region scene, people can adjust this config accordingly, to not trigger too frequent compactions.

src/mito2/src/engine/edit_region_test.rs
```diff
@@ -15,20 +15,102 @@
...
+        compaction_time_window: None,
+        flushed_entry_id: None,
+        flushed_sequence: None,
+    };
+    engine
+        .edit_region(region.region_id, new_edit())
+        .await
+        .unwrap();
+    // Makes some time for the worker loop to process the request.
+    tokio::time::sleep(Duration::from_secs(1)).await;
```
evenyag, line 93: Do we need this sleep?

src/mito2/src/region/opener.rs
```diff
@@ -66,13 +66,15 @@ pub(crate) struct RegionOpener {
...
-    time_provider: Option<TimeProviderRef>,
+    time_provider: TimeProviderRef,
     stats: ManifestStats,
     wal_entry_reader: Option<Box<dyn WalEntryReader>>,
 }
 
 impl RegionOpener {
     /// Returns a new opener.
+    // TODO(LFC): Reduce the number of arguments.
+    #[allow(clippy::too_many_arguments)]
```
evenyag, line 77: Maybe we can leave it as a good first issue.

src/mito2/src/worker/handle_manifest.rs
```diff
@@ -165,6 +169,36 @@ impl<S> RegionWorkerLoop<S> {
...
+            self.schedule_compaction(&region).await;
+        }
+    }
+
+    /// Schedule compaction for the region if necessary.
+    pub(crate) async fn schedule_compaction(&mut self, region: &MitoRegionRef) {
```
evenyag, line 179: Maybe we should put this method in `handle_compaction.rs`.

src/mito2/src/engine/edit_region_test.rs
MichaelScofield: Indeed not needed, removed.

src/mito2/src/region/opener.rs
MichaelScofield: It requires some degree of understanding of the mito engine codes (to do the refactoring), I'm afraid it's not suitable for a "first" issue.

killme2008 (approved):

PR #4653, opened 2024-08-30
feat: change log level dynamically

discord9:

add a http route to dynamic change log level, using the same syntax like `RUST_LOG`, i.e:
```bash
> curl --data "trace;flow=debug" 127.0.0.1:4000/debug/log_level;
Log Level changed from Some("info") to "trace"%
```
will change current log level to `trace;flow=debug` and appiled to logs came after this

Impl-wise, when init global logging, still will first check env `RUST_LOG` and etc, only difference are using `reload::Layer` to allow change log level dynamically.

perf-wise if not contended, logging with reloadable tracing layer is just fast path of acquire Read lock on a `RwLock` which is just one relaxed atomic operation and a few bit op

## Checklist

### DIFF
--- a/src/servers/src/http.rs
+++ b/src/servers/src/http.rs
@@ -710,7 +710,7 @@ impl HttpServer {
                     )),
             )
             .nest(
-                &format!("/{HTTP_API_VERSION}/admin"),
+                "/debug",
                 Router::new()
                     // handler for changing log level dynamically
                     .route(
--- a/src/servers/src/http.rs
+++ b/src/servers/src/http.rs
@@ -74,6 +74,7 @@ use crate::query_handler::{
 use crate::server::Server;
 
 pub mod authorize;
+pub mod dyn_log;
 pub mod event;
 pub mod handler;
 pub mod header;
@@ -707,6 +708,15 @@ impl HttpServer {
                         AuthState::new(self.user_provider.clone()),
                         authorize::check_http_auth,
                     )),
+            )
+            .nest(
+                &format!("/{HTTP_API_VERSION}/admin"),
+                Router::new()
+                    // handler for changing log level dynamically
+                    .route(
+                        "/log_level",
+                        routing::get(dyn_log::dyn_log_handler).post(dyn_log::dyn_log_handler),
+                    ),
             )
             // Handlers for debug, we don't expect a timeout.
             .nest(
--- a/src/servers/src/http/dyn_log.rs
+++ b/src/servers/src/http/dyn_log.rs
@@ -1,0 +1,59 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use axum::http::StatusCode;
+use axum::response::IntoResponse;
+use common_telemetry::info;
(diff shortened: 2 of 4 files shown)


src/servers/src/http.rs
```diff
@@ -708,10 +709,24 @@ impl HttpServer {
...
+                    ),
+            )
             // Handlers for debug, we don't expect a timeout.
             .nest(
                 &format!("/{HTTP_API_VERSION}/prof"),
                 Router::new()
+                    // handler for changing log level dynamically
+                    .route(
+                        "/log_level",
+                        routing::get(dyn_log::dyn_log_handler).post(dyn_log::dyn_log_handler),
```
evenyag, line 728: Maybe we should move it to `/debug/log_level`

src/servers/src/http.rs
```diff
@@ -708,10 +709,24 @@ impl HttpServer {
...
             )
+            .nest(
+                &format!("/{HTTP_API_VERSION}/admin"),
+                Router::new()
+                    // handler for changing log level dynamically
+                    .route(
+                        "/log_level",
+                        routing::get(dyn_log::dyn_log_handler).post(dyn_log::dyn_log_handler),
+                    ),
+            )
```
shuiyisong, line 720:
```suggestion
```

Let's not add an `/admin` router here. Putting it under `/debug/log_level` would be enough.

src/servers/src/http.rs
```diff
@@ -708,10 +709,24 @@ impl HttpServer {
...
             // Handlers for debug, we don't expect a timeout.
             .nest(
                 &format!("/{HTTP_API_VERSION}/prof"),
```
shuiyisong, line 723:
```suggestion
                &format!("/prof"),
```

We can agree that `/{HTTP_API_VERSION}` is for user protocol interfaces, but `/prof` and `/debug` are not. So let's remove the version prefix. cc @evenyag

discord9:
Pending #4657 for updating `tracing_subscriber` crate

discord9:
> Pending #4657 for updating `tracing_subscriber` crate

edited: now ready for review

src/servers/src/http/dyn_log.rs
```diff
@@ -0,0 +1,54 @@
...
+            InternalSnafu {
+                err_msg: format!("Fail to modify filter: {e:?}"),
+            }
+            .build()
+        })?;
+    let change_note = format!(
+        "Log Level changed from {:?} to {:?}",
+        old_filter.map(|f| f.to_string()),
+        new_filter.to_string()
+    );
```
evenyag, line 51:
```suggestion
    let change_note = format!(
        "Log Level changed from {} to {}",
        old_filter.map(|f| f.to_string()).unwrap_or_default(),
        new_filter
    );
```

zyy17 (approved):

src/servers/src/http.rs
```diff
@@ -708,6 +709,15 @@ impl HttpServer {
                         authorize::check_http_auth,
                     )),
             )
+            .nest(
```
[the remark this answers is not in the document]
discord9: > We already have a `/admin` route, why add a new route?

1. `/admin` only exist for metasrv apparently, adding it to all role could be messy?
2. @evenyag suggest add a `debug` route which seems more appropriately: https://github.com/GreptimeTeam/greptimedb/pull/4653#discussion_r1739598842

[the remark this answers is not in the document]
killme2008: Ok, but it must be under the`/v1/` path, just like `/v1/prof/mem`.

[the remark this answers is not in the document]
discord9: > Ok, but it must be under the`/v1/` path, just like `/v1/prof/mem`.

@shuiyisong also suggest reserve `v1` for protocol message only, see: https://github.com/GreptimeTeam/greptimedb/pull/4653#discussion_r1740279857
> We can agree that /{HTTP_API_VERSION} is for user protocol interfaces, but /prof and /debug are not. So let's remove the version prefix. cc

oh and I forget to remove `v1` for `prof/mem`, might also remove memprof port later in separate PR since it's currently broken for some platform(and recommend samply in docs)

PR #4654, opened 2024-08-30
feat: export import database

fengjiachun:

1. export `create database` SQL
2. create the database before importing
3. fix a bug (unset `db` parameter) for HTTP SQL server

## Checklist

### DIFF
--- a/src/cmd/src/cli/export.rs
+++ b/src/cmd/src/cli/export.rs
@@ -130,6 +130,7 @@ impl Export {
             return Ok(db_names);
         };
 
+        // Check if the schema exists
         db_names
             .into_iter()
             .find(|db_name| db_name.to_lowercase() == schema.to_lowercase())
--- a/src/cmd/src/cli/export.rs
+++ b/src/cmd/src/cli/export.rs
@@ -29,7 +29,7 @@ use tracing_appender::non_blocking::WorkerGuard;
 
 use crate::cli::database::DatabaseClient;
 use crate::cli::{database, Instance, Tool};
-use crate::error::{EmptyResultSnafu, Error, FileIoSnafu, Result};
+use crate::error::{EmptyResultSnafu, Error, FileIoSnafu, Result, SchemaNotFoundSnafu};
 
 type TableReference = (String, String, String);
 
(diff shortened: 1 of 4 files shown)


src/servers/src/http/handler.rs
```diff
@@ -76,6 +77,11 @@ pub async fn sql(
 ) -> HttpResponse {
     let start = Instant::now();
     let sql_handler = &state.sql_handler;
+    if let Some(db) = &query_params.db.or(form_params.db) {
+        let (catalog, schema) = parse_catalog_and_schema_from_db_string(db);
+        query_ctx.set_current_catalog(&catalog);
+        query_ctx.set_current_schema(&schema);
+    }
```
fengjiachun, line 84: Bug fix

WenyXu (approved):
Rest LGTM

src/cmd/src/cli/export.rs
```diff
@@ -120,18 +120,34 @@ pub struct Export {
...
+
+        // Check if the schema exists
+        db_names
+            .into_iter()
+            .find(|db_name| db_name.to_lowercase() == schema.to_lowercase())
+            .map(|name| vec![name])
+            .context(SchemaNotFoundSnafu {
+                catalog: &self.catalog,
+                schema,
+            })
```
MichaelScofield, line 141: Nit:
try this
```rust
use itertools::Itertools;

let db_names = self.all_db_names().await?;
let db_names = db_names
    .cartesian_product(self.schema.iter())
    .find_map(|(x, y)| if x.to_lowercase() == y.to_lowercase() {
        Some(x)
    } else {
        None
    })
    .collect();
ensure!(!db_names.is_empty(), SchemaNotFoundSnafu {
    catalog: &self.catalog,
    schema,
};

File: src/cmd/src/error.rs [hunk 3]
```diff
@@ -415,6 +423,7 @@ impl ErrorExt for Error {
             }
             Error::MetaClientInit { source, .. } => source.status_code(),
             Error::TonicTransport { .. } => StatusCode::Internal,
+            Error::SchemaNotFound { .. } => StatusCode::Unexpected,
```
waynexia, line 426: We have a `DatabaseNotFound = 4004` for this case

```suggestion
            Error::SchemaNotFound { .. } => StatusCode::DatabaseNotFound,
```

waynexia (approved):
Others LGTM

src/cmd/src/cli/export.rs
```diff
@@ -232,19 +253,57 @@ impl Export {
...
+                .await
+                .context(FileIoSnafu)?;
+            let file = db_dir.join("create_database.sql");
+            let mut file = File::create(file).await.context(FileIoSnafu)?;
+            match self
+                .show_create("DATABASE", &self.catalog, &schema, None)
+                .await
+            {
+                Err(e) => {
+                    error!(e; r#"Failed to export database schema "{}"."{}""#, self.catalog, schema)
```
MichaelScofield, line 293: Is it better to change the behavior to "fail-fast" here? If something goes wrong for showing one database, the successive databases' "show" might be in vain: user has to execute the export again for the integrity of data.

src/cmd/src/cli/export.rs
fengjiachun: Thank you for the suggestion, but my idea is to keep it simple since performance is not a major concern here.

src/cmd/src/cli/export.rs
fengjiachun: Good catch, I am going to make all `show create database` and `show create table` to be fail-fast,

MichaelScofield (approved):

Issue #4696, opened 2024-09-08
Panic "index out of bounds" while inserting a lot of data


akhilman:
### What type of bug is this?

Crash

### What subsystems are affected?

Standalone mode

### Minimal reproduce step

I am having an issue while inserting a lot of data (~100M rows) in to database. Crash seems to appear at stable points but this point changes when I change data order and insert batch size.
I am unable to reproduce this issue with synthetic data.

I am inserting data from python using `mysql-connector-python`.

### What did you expect to see?

All data in database.

### What did you see instead?

Mysql client exits with error.

### What operating system did you use?

ArchLinux x64

### What version of GreptimeDB did you use?

0.9.3 from docker.io

### Relevant log output and stack trace

```bash
name = "GreptimeDB"
operating_system = "Ubuntu 22.4.0 [64-bit]"
crate_version = "0.9.3"
explanation = """
Panic occurred in file 'library/core/src/unicode/unicode_data.rs' at line 80
"""
cause = "index out of bounds: the len is 54 but the index is 54"
method = "Panic"
backtrace = """

   0: 0x5d099067d4f3 - core::panicking::panic_fmt::h088f9f84f0369284
   1: 0x5d099067d707 - core::panicking::panic_bounds_check::h8d8570f0c291a3bc
   2: 0x5d0990689ff1 - core::unicode::unicode_data::alphabetic::lookup::ha3afd12d2ef9c31c
   3: 0x5d09967d45c2 - <sqlparser::dialect::mysql::MySqlDialect as sqlparser::dialect::Dialect>::is_identifier_start::hb7160ac9833879dc
   4: 0x5d099677d845 - sqlparser::tokenizer::Tokenizer::next_token::h96a647e5f1c84413
   5: 0x5d099677851b - sqlparser::tokenizer::Tokenizer::tokenize_with_location_into_buf::h6c37888545515fb1
   6: 0x5d09967782fc - sqlparser::tokenizer::Tokenizer::tokenize_with_location::h5dcc8f918a10dccf
   7: 0x5d09965c9a78 - sqlparser::parser::Parser::try_with_sql::h0d1d0257366c368d
   8: 0x5d099656f139 - sql::parser::ParserContext::create_with_dialect::h13e0ced9455ecc62
   9: 0x5d09924230c7 - <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll::h9c609bd0d0a5476c
[...]

PR #4694, opened 2024-09-06
fix: unconditional statistics

waynexia:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

Closes #4692

## What's changed and what's your intention?

#4545 gives an incorrect impl of `statistics` method, which should consider both projection and filter of the underlying format reader when generating statistics.

## Checklist

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.

PR #4698, opened 2024-09-09
fix: use `is_ascii_alphabetic`

discord9:

[#4694](https://github.com/GreptimeTeam/greptimedb/issues/4696)

use `is_ascii_alphabetic` instead of `is_alphabetic`  which checks for **Other** langauages' alphabet too(i.e: `assert!('京'.is_alphabetic());`

## Checklist

### DIFF
--- a/src/sql/src/dialect.rs
+++ b/src/sql/src/dialect.rs
@@ -20,11 +20,11 @@ pub struct GreptimeDbDialect {}
 
 impl Dialect for GreptimeDbDialect {
     fn is_identifier_start(&self, ch: char) -> bool {
-        ch.is_alphabetic() || ch == '_' || ch == '#' || ch == '@'
+        ch.is_ascii_alphabetic() || ch == '_' || ch == '#' || ch == '@'
     }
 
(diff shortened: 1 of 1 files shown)


discord9:
After though look into unicode standard(https://www.unicode.org/Public/UCD/latest/ucd/DerivedCoreProperties.txt) and mysql standard(https://dev.mysql.com/doc/refman/8.0/en/identifiers.html) and psql standard(https://www.postgresql.org/docs/current/sql-syntax-lexical.html), I found:
- unicode `Char` with `Alphabetic` property have too many of them(10052 in BMP(Basic Multilingual Plane) and 101539 if include Han&Hangul)
- mysql just support ascii char & the entire range of BMP
- psql support a-z, but also letters with diacritical marks and non-Latin letters

with all those complication, I think we better stick to `is_alphabetic` and correct upstream `sqlparser`'s usage of `is_alphabetic` instead?

discord9:
Rethinking this `is_alphabetic` is better than just allow all chars in BMP, so we can avoid column names like `，，，` while accepting `一二三` so it's more useful

PR #4699, opened 2024-09-09
feat: protect datanode with concurrency limit.

lyang24:

Adding query parallelism limit in region server to protect datanode from query overload.

## Checklist

### DIFF
--- a/config/config.md
+++ b/config/config.md
@@ -17,6 +17,7 @@
 | `default_timezone` | String | Unset | The default timezone of the server. |
 | `init_regions_in_background` | Bool | `false` | Initialize all regions in the background during the startup.<br/>By default, it provides services after all regions have been initialized. |
 | `init_regions_parallelism` | Integer | `16` | Parallelism of initializing regions. |
+| `max_concurrent_queries` | Integer | `0` | The maximum current queries allowed to be executed. Zero means unlimited. |
 | `runtime` | -- | -- | The runtime options. |
 | `runtime.global_rt_size` | Integer | `8` | The number of threads to execute the runtime for global read operations. |
 | `runtime.compact_rt_size` | Integer | `4` | The number of threads to execute the runtime for global write operations. |
@@ -335,6 +336,7 @@
 | `init_regions_in_background` | Bool | `false` | Initialize all regions in the background during the startup.<br/>By default, it provides services after all regions have been initialized. |
 | `enable_telemetry` | Bool | `true` | Enable telemetry to collect anonymous usage data. |
 | `init_regions_parallelism` | Integer | `16` | Parallelism of initializing regions. |
+| `max_concurrent_queries` | Integer | `0` | The maximum current queries allowed to be executed. Zero means unlimited. |
 | `rpc_addr` | String | Unset | Deprecated, use `grpc.addr` instead. |
 | `rpc_hostname` | String | Unset | Deprecated, use `grpc.hostname` instead. |
 | `rpc_runtime_size` | Integer | Unset | Deprecated, use `grpc.runtime_size` instead. |
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -19,6 +19,9 @@ enable_telemetry = true
 ## Parallelism of initializing regions.
 init_regions_parallelism = 16
 
+## The maximum current queries allowed to be executed. Zero means unlimited.
+max_concurrent_queries = 0
+
 ## Deprecated, use `grpc.addr` instead.
 ## @toml2docs:none-default
 rpc_addr = "127.0.0.1:3001"
--- a/src/datanode/src/config.rs
+++ b/src/datanode/src/config.rs
@@ -305,6 +305,7 @@ pub struct DatanodeOptions {
     pub meta_client: Option<MetaClientOptions>,
     pub wal: DatanodeWalConfig,
     pub storage: StorageConfig,
+    pub max_concurrent_queries: usize,
     /// Options for different store engines.
     pub region_engine: Vec<RegionEngineConfig>,
     pub logging: LoggingOptions,
@@ -339,6 +340,7 @@ impl Default for DatanodeOptions {
             meta_client: None,
             wal: DatanodeWalConfig::default(),
             storage: StorageConfig::default(),
+            max_concurrent_queries: 0,
             region_engine: vec![
                 RegionEngineConfig::Mito(MitoConfig::default()),
                 RegionEngineConfig::File(FileEngineConfig::default()),
--- a/src/datanode/src/datanode.rs
+++ b/src/datanode/src/datanode.rs
@@ -314,7 +314,7 @@ impl DatanodeBuilder {
         &self,
         event_listener: RegionServerEventListenerRef,
     ) -> Result<RegionServer> {
-        let opts = &self.opts;
+        let opts: &DatanodeOptions = &self.opts;
 
         let query_engine_factory = QueryEngineFactory::new_with_plugins(
             // query engine in datanode only executes plan with resolved table source.
@@ -334,6 +334,9 @@ impl DatanodeBuilder {
             common_runtime::global_runtime(),
             event_listener,
             table_provider_factory,
+            opts.max_concurrent_queries,
+            //TODO: revaluate the hardcoded timeout on the next version of datanode concurrency limiter.
+            Duration::from_millis(100),
         );
 
         let object_store_manager = Self::build_object_store_manager(&opts.storage).await?;
--- a/src/datanode/src/region_server.rs
+++ b/src/datanode/src/region_server.rs
@@ -16,6 +16,7 @@ use std::collections::HashMap;
 use std::fmt::Debug;
 use std::ops::Deref;
 use std::sync::{Arc, RwLock};
+use std::time::Duration;
 
 use api::region::RegionResponse;
 use api::v1::region::{region_request, RegionResponse as RegionResponseV1};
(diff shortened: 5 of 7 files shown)


src/datanode/src/region_server.rs
```diff
@@ -140,6 +144,15 @@ impl RegionServer {
         region_id: RegionId,
         request: RegionRequest,
     ) -> Result<RegionResponse> {
+        if let Some(semaphore) = &self.inner.concurrency_limit {
+            let permit = semaphore
+                .acquire()
+                .await
+                .context(OverConcurrencyLimitSnafu)?;
```
sunng87, line 151: we can add a configurable timeout here to make it more usable, code like

```
select!{
  semaphore.acquire(), timeout(...)
}
```

lyang24: is the timeout act like a lock timeout for acquiring permit from semaphore?

sunng87: Yes. Comparing to plain `acquire`, this allows request to fail without waiting forever.

src/datanode/src/config.rs
```diff
@@ -305,6 +305,8 @@ pub struct DatanodeOptions {
     pub meta_client: Option<MetaClientOptions>,
     pub wal: DatanodeWalConfig,
     pub storage: StorageConfig,
+    pub parallelism: Option<usize>,
+    pub semaphore_timeout_ms: u64,
```
sunng87, line 309: I thikn semaphore is an implementation related concept, so we better not to expose this name in configuration and error output. 

Clickhouse has a configuration item called `max_concurrent_queries` can be a good name for `parallelism`.

Something like `concurrent_query_limiter_timeout` can be good, but you might have better names for this.

sunng87 (commented):
Remember to update config docs when we finalize the names

src/datanode/src/config.rs
lyang24: thanks for the suggestion I have updated the naming and configurations docs

sunng87 (approved):
LGTM

config/config.md
```diff
@@ -335,6 +335,8 @@
 | `init_regions_in_background` | Bool | `false` | Initialize all regions in the background during the startup.<br/>By default, it provides services after all regions have been initialized. |
 | `enable_telemetry` | Bool | `true` | Enable telemetry to collect anonymous usage data. |
 | `init_regions_parallelism` | Integer | `16` | Parallelism of initializing regions. |
+| `max_concurrent_queries` | Integer | `0` | The maximum current queries allowed to be executed. Zero means unlimited. |
+| `concurrent_query_limiter_timeout` | String | `25ms` | The timeout of acquiring concurrency permits. |
```
waynexia, line 339: Do we need this option? Despite the implementation details (`acquire` vs. `try_acquire`), it seems like users would hardly adjust this option. And exposing such an option would also limit us from implementing a better behavior in the future, like "only pending 30 awaiting requests in the queue, for at most 100ms".

We can hide such logic as our system's internal details. WDYT @sunng87 @evenyag

src/datanode/src/region_server.rs
```diff
@@ -450,6 +465,33 @@ struct RegionServerInner {
...
+        if max_concurrent_queries == 0 {
+            return None;
+        }
+        Some(RegionServerParallelism {
+            semaphore: Semaphore::new(max_concurrent_queries),
+            timeout,
+        })
+    }
+
+    pub async fn acquire(&'a self) -> Result<SemaphorePermit<'a>> {
```
evenyag, line 489: Looks like we can remove the lifetime parameter
```suggestion
impl RegionServerParallelism {
    pub fn from_opts(max_concurrent_queries: usize, timeout: Duration) -> Option<Self> {
        if max_concurrent_queries == 0 {
            return None;
        }
        Some(RegionServerParallelism {
            semaphore: Semaphore::new(max_concurrent_queries),
            timeout,
        })
    }

    pub async fn acquire(&self) -> Result<SemaphorePermit> {
```

config/config.md
lyang24: I agree remove this option will give better flexibility for future. I think `try_acquire` is a special case that is equivalent to 0 timeout i would prefer more flexibility in this case. I removed the `concurrent_query_limiter_timeout` in DataNode options and hardcoded 25ms wait with a TODO. In future versions i think we could update the `max_concurrent_queries` and `concurrent_query_limiter_timeout` dynamically via sql such as `SET max_concurrent_queries = 8;`

src/datanode/src/region_server.rs
```diff
@@ -140,6 +152,12 @@ impl RegionServer {
         region_id: RegionId,
         request: RegionRequest,
     ) -> Result<RegionResponse> {
+        if let Some(p) = &self.inner.parallelism {
+            let permit = p.acquire().await?;
+            let res = self.inner.handle_request(region_id, request).await;
+            drop(permit);
+            return res;
+        }
```
v0y4g3r, line 160: Looks like this semaphore also applies to `RegionPutRequest` whilst the purpose is to limit concurrent queries. We need to distinguish different request types.

src/datanode/src/datanode.rs
```diff
@@ -334,6 +334,9 @@ impl DatanodeBuilder {
             common_runtime::global_runtime(),
             event_listener,
             table_provider_factory,
+            opts.max_concurrent_queries,
+            //TODO: revaluate the hardcoded timeout on the next version of datanode concurrency limiter.
+            Duration::from_millis(25),
```
v0y4g3r, line 339: 25ms seems to be too short. Maybe we need to hold those throttled queries instead of fail them very quickly.

src/datanode/src/region_server.rs
lyang24: should i put the rate limiting on `handle_read` and `handle_remote_read` that does datafusion queries? seems like `handle_requests` method is more of writes and region ddl changes.

src/datanode/src/datanode.rs
lyang24: I can increase the threshold. By hold do you mean to allow queries to wait for the semaphore until the request to datanode timeout?

src/datanode/src/region_server.rs
```diff
@@ -167,6 +179,11 @@ impl RegionServer {
         &self,
         request: api::v1::region::QueryRequest,
     ) -> Result<SendableRecordBatchStream> {
+        let permit = if let Some(p) = &self.inner.parallelism {
+            Some(p.acquire().await?)
+        } else {
+            None
+        };
```
evenyag, line 186: The permit is dropped once this method returns. We don't need to drop it manually.
```suggestion
        let _permit = if let Some(p) = &self.inner.parallelism {
            Some(p.acquire().await?)
        } else {
            None
        };
```

src/datanode/src/region_server.rs
```diff
@@ -189,17 +206,27 @@ impl RegionServer {
...
+        res
     }
 
     #[tracing::instrument(skip_all)]
     pub async fn handle_read(&self, request: QueryRequest) -> Result<SendableRecordBatchStream> {
+        let permit = if let Some(p) = &self.inner.parallelism {
+            Some(p.acquire().await?)
+        } else {
+            None
+        };
```
evenyag, line 229:
```suggestion
        let _permit = if let Some(p) = &self.inner.parallelism {
            Some(p.acquire().await?)
        } else {
            None
        };
```

src/datanode/src/region_server.rs
```diff
@@ -189,17 +206,27 @@ impl RegionServer {
             .await
             .context(DecodeLogicalPlanSnafu)?;
 
-        self.inner
+        let res = self
+            .inner
```
evenyag, line 210:
```suggestion
        self
            .inner
```

src/datanode/src/region_server.rs
```diff
@@ -189,17 +206,27 @@ impl RegionServer {
...
                 header: request.header,
                 region_id,
                 plan,
             })
-            .await
+            .await;
+        if let Some(_permit) = permit {
+            drop(_permit)
+        };
+        res
```
evenyag, line 220:
```suggestion
            .await
```

src/datanode/src/region_server.rs
```diff
@@ -228,9 +255,14 @@ impl RegionServer {
             .context(DataFusionSnafu)?
             .data;
 
-        self.inner
+        let res = self
+            .inner
```
evenyag, line 259:
```suggestion
        self
            .inner
```

src/datanode/src/region_server.rs
```diff
@@ -228,9 +255,14 @@ impl RegionServer {
...
             .handle_read(QueryRequest { plan, ..request })
-            .await
+            .await;
+        if let Some(_permit) = permit {
+            drop(_permit)
+        };
+        res
```
evenyag, line 265:
```suggestion
            .await
```

config/datanode.example.toml
```diff
@@ -19,6 +19,9 @@ enable_telemetry = true
 ## Parallelism of initializing regions.
 init_regions_parallelism = 16
 
+## The maximum current queries allowed to be executed. Zero means unlimited.
+max_concurrent_queries = 0
```
evenyag, line 23: Please also update the example in `standalone.example.toml`
https://github.com/GreptimeTeam/greptimedb/blob/5c64f0ce0950e653ac2238c61b37b32a489810c7/config/standalone.example.toml#L15-L17

evenyag (approved):
Others LGTM

config/datanode.example.toml
lyang24: TYFR this is done

sunng87 (approved):

config/datanode.example.toml
v0y4g3r: @evenyag Please check if this comment is resolved so we can proceed.

v0y4g3r (approved):
LGTM

PR #4709, opened 2024-09-10
fix: pipeline dissert error is returned directly to the user, instead of printing a warn log

paomian:

* pipeline dissert error is returned directly to the user, instead of printing a warn log

## Checklist


shuiyisong (approved):
Can we add a test to match this case?

Rest LGTM

paomian:
I will add some test for this case.

shuiyisong (approved):

sunng87 (approved):

Issue #4708, opened 2024-09-10
Comment out the parameters in the sample configuration that are linked to the system's CPU and memory.


killme2008:
### What type of enhancement is this?

Configuration

### What does the enhancement do?

Comment out the parameters in the sample configurations that are linked to the system's CPU and memory. By default, they should not be enabled, meaning the default configuration will be used. Users can optionally uncomment and modify them.

https://github.com/GreptimeTeam/greptimedb/tree/main/config

PR #4711, opened 2024-09-10
docs: use docs comment prefix and bump toml2docs version

zyy17:

Resolve https://github.com/GreptimeTeam/greptimedb/issues/4708.

Bump `toml2docs` to `v0.1.3` for following changes:

1. Use the special prefix `#+` to comment on some fields in the example because we want to decide at runtime(But we still need it to generate docs). For those fields, we mark its default value as `None`;

2. Change comment metadata from `#+toml2docs:none-default` to `@toml2docs:none-default` for new syntax;

## Checklist


config/standalone.example.toml
evenyag, line 1: We should also remove the example values for the `[runtime]` section.

config/standalone.example.toml
```diff
@@ -447,26 +447,30 @@ max_background_jobs = 4
 auto_flush_interval = "1h"
 
 ## Global write buffer size for all regions. If not set, it's default to 1/8 of OS memory with a max limitation of 1GB.
-global_write_buffer_size = "1GB"
+## @toml2docs:none-default
+#+ global_write_buffer_size = "1GB"
```
evenyag, line 451: Also remove the default `num_workers = 8` from the mito config.

config/standalone.example.toml
zyy17: done

config/config.md
```diff
@@ -116,12 +116,12 @@
 | `region_engine.mito.compress_manifest` | Bool | `false` | Whether to compress manifest and checkpoint file by gzip (default false). |
 | `region_engine.mito.max_background_jobs` | Integer | `4` | Max number of running background jobs |
 | `region_engine.mito.auto_flush_interval` | String | `1h` | Interval to auto flush a region if it has not flushed yet. |
-| `region_engine.mito.global_write_buffer_size` | String | `1GB` | Global write buffer size for all regions. If not set, it's default to 1/8 of OS memory with a max limitation of 1GB. |
-| `region_engine.mito.global_write_buffer_reject_size` | String | `2GB` | Global write buffer size threshold to reject write requests. If not set, it's default to 2 times of `global_write_buffer_size` |
-| `region_engine.mito.sst_meta_cache_size` | String | `128MB` | Cache size for SST metadata. Setting it to 0 to disable the cache.<br/>If not set, it's default to 1/32 of OS memory with a max limitation of 128MB. |
+| `region_engine.mito.global_write_buffer_size` | String | `None` | Global write buffer size for all regions. If not set, it's default to 1/8 of OS memory with a max limitation of 1GB. |
```
killme2008, line 119: `None` is the value of Rust language, it doesn't make sense for users. So I think `Not set` or something else would be better.

evenyag (approved):

config/config.md
sunng87: Can we use `auto` here?

killme2008 (approved):
LGTM

config/config.md
killme2008: @sunng87 I've tried it before, however, there were some issues on which we couldn't reach an agreement, so they were temporarily shelved.

PR #4716, opened 2024-09-10
fix: support append-only physical table

waynexia:

support append-only physical table

## Checklist


evenyag (approved):

src/metric-engine/src/engine/create.rs
```diff
@@ -428,7 +429,8 @@ impl MetricEngineInner {
 
         // remove TTL option
```
fengjiachun, line 430:
```suggestion
        // remove TTL and APPEND_MODE option
```

WenyXu (approved):

fengjiachun (approved):

Issue #4450, opened 2024-07-29
Add the slow operation logging


zyy17:
### What type of enhancement is this?

Other

### What does the enhancement do?

For better troubleshooting, we should capture the slow operation(for example, query, compaction, etc.) into a standalone log file by adding another logging layer.

PR #4719, opened 2024-09-10
feat: add `StatementStatistics` for slow query logging implementation

zyy17:

Resolve https://github.com/GreptimeTeam/greptimedb/issues/4450

The PR introduces the new struct `StatementStatistics` for implementing statistics of query statements. For now, it's used for slow query logging.

## Checklist

### DIFF
--- a/config/config.md
+++ b/config/config.md
@@ -166,6 +166,7 @@
 | `logging.slow_query` | -- | -- | The slow query log options. |
 | `logging.slow_query.enable` | Bool | `false` | Whether to enable slow query log. |
 | `logging.slow_query.threshold` | String | Unset | The threshold of slow query. |
+| `logging.slow_query.sample_rate` | Float | Unset | The sampling rate of slow query log. |
 | `export_metrics` | -- | -- | The datanode can export its metrics and send to Prometheus compatible service (e.g. send to `greptimedb` itself) from remote-write API.<br/>This is only used for `greptimedb` to export its own metrics internally. It's different from prometheus scrape. |
 | `export_metrics.enable` | Bool | `false` | whether enable export metrics. |
 | `export_metrics.write_interval` | String | `30s` | The interval of export metrics. |
@@ -255,6 +256,7 @@
 | `logging.slow_query` | -- | -- | The slow query log options. |
 | `logging.slow_query.enable` | Bool | `false` | Whether to enable slow query log. |
 | `logging.slow_query.threshold` | String | Unset | The threshold of slow query. |
+| `logging.slow_query.sample_rate` | Float | Unset | The sampling rate of slow query log. |
 | `export_metrics` | -- | -- | The datanode can export its metrics and send to Prometheus compatible service (e.g. send to `greptimedb` itself) from remote-write API.<br/>This is only used for `greptimedb` to export its own metrics internally. It's different from prometheus scrape. |
 | `export_metrics.enable` | Bool | `false` | whether enable export metrics. |
 | `export_metrics.write_interval` | String | `30s` | The interval of export metrics. |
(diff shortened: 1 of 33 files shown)


sunng87:
Is there any chance we can put this at instance level, say `SqlQueryHandler` and `PrometheusHandler`? so we don't need to configure it for each protocol.

zyy17:
> Is there any chance we can put this at instance level, say `SqlQueryHandler` and `PrometheusHandler`? so we don't need to configure it for each protocol.

Good idea. Let me try to do that.

config/config.md
```diff
@@ -162,6 +162,9 @@
 | `logging.log_format` | String | `text` | The log format. Can be `text`/`json`. |
 | `logging.tracing_sample_ratio` | -- | -- | The percentage of tracing will be sampled and exported.<br/>Valid range `[0, 1]`, 1 means all traces are sampled, 0 means all traces are not sampled, the default value is 1.<br/>ratio > 1 are treated as 1. Fractions < 0 are treated as 0 |
 | `logging.tracing_sample_ratio.default_ratio` | Float | `1.0` | -- |
+| `logging.slow_query` | -- | -- | The slow query log options. |
+| `logging.slow_query.enable` | Bool | `false` | Whether to enable slow query log. |
+| `logging.slow_query.threshold` | String | Unset | The threshold of slow query. |
```
killme2008, line 167: Except the threshold, maybe we need a sample rate too. Just like mongodb 

https://www.mongodb.com/zh-cn/docs/cloud-manager/performance-advisor/slow-query-threshold/

src/common/telemetry/src/logging.rs
```diff
@@ -235,6 +252,42 @@ pub fn init_global_logging(
             None
         };
 
+        let slow_query_logging_layer = if !opts.dir.is_empty() && opts.slow_query.enable {
+            let rolling_appender =
+                RollingFileAppender::new(Rotation::HOURLY, &opts.dir, "greptimedb-slow");
```
killme2008, line 257: Looks better
```suggestion
                RollingFileAppender::new(Rotation::HOURLY, &opts.dir, "greptimedb-slow-queries");
```

src/servers/src/http/handler.rs
```diff
@@ -138,7 +138,19 @@ pub async fn sql(
     if let Some(limit) = query_params.limit {
         resp = resp.with_limit(limit);
     }
-    resp.with_execution_time(start.elapsed().as_millis() as u64)
+
+    let cost = start.elapsed();
+    if let Some(threshold) = state.slow_query_threshold {
```
killme2008, line 143: What about move it into 

https://github.com/GreptimeTeam/greptimedb/blob/36b1bafbf00152cc89c8d4362ace432bbb1932a3/src/operator/src/statement.rs#L111

?

Then you can intercept all the queries.

killme2008 (commented):
Good job! Another suggestion is that maybe we can record all the slow queries into a table like `information_schema.slow_queries` like that. Which is a append-only tables with TTL setting.

zyy17:
> Good job! Another suggestion is that maybe we can record all the slow queries into a table like `information_schema.slow_queries` like that. Which is a append-only tables with TTL setting.



> Good job! Another suggestion is that maybe we can record all the slow queries into a table like `information_schema.slow_queries` like that. Which is a append-only tables with TTL setting.

I can take it as the next PR. Let me learn how to implement it.

config/config.md
zyy17: Maybe it's not a necessary configuration? The current implementation is very cheap: log the slow query into file and collect by 3rd tools.

src/servers/src/http/handler.rs
zyy17: Seems a good idea. Let me have a try.

config/config.md
killme2008: The sample ratio determines whether to print all slow queries, defaulting to 1 (100%). If all queries are slow, printing them all could harm database performance.

killme2008 (approved):
LGTM

v0y4g3r:
Is it possible that we may print sensitive data into slow log file? Like `copy from/to` statements usually take a long time, while bearing AK/SK info.

fengjiachun (approved):
LGTM

zyy17:
> Is it possible that we may print sensitive data into slow log file? Like `copy from/to` statements usually take a long time, while bearing AK/SK info.

You're right. Let me check it.

zyy17:
> Is it possible that we may print sensitive data into slow log file? Like `copy from/to` statements usually take a long time, while bearing AK/SK info.

We already use https://github.com/GreptimeTeam/greptimedb/blob/main/src/sql/src/statements/option_map.rs#L27 for sensitive data. If we use `Display`, it will print `****` for senstive data.

PR #4721, opened 2024-09-11
feat: improve support for postgres extended protocol

sunng87:

This patch improves postgres extended query subprotocol:

- add support for empty query
- add support for statements from `fixtures` module
- improved situation that server failed to inference parameter types (for some temporary tables like information_schema)
- residual issue: it's still impossible to get a valid return value for statement describe request of temporary table.

## Checklist

### DIFF
--- a/src/servers/src/postgres/handler.rs
+++ b/src/servers/src/postgres/handler.rs
@@ -190,16 +190,8 @@ impl QueryParser for DefaultQueryParser {
         crate::metrics::METRIC_POSTGRES_PREPARED_COUNT.inc();
         let query_ctx = self.session.new_query_context();
 
-        if sql.is_empty() {
-            return Ok(SqlPlan {
-                query: sql.to_owned(),
-                plan: None,
-                schema: None,
-            });
-        }
-
-        // test against predefined rules first
-        if fixtures::matches(sql) {
+        // do not parse if query is empty or matches rules
+        if sql.is_empty() || fixtures::matches(sql) {
             return Ok(SqlPlan {
                 query: sql.to_owned(),
                 plan: None,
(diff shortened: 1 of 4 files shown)


src/servers/src/postgres/handler.rs
```diff
@@ -244,6 +260,16 @@ impl ExtendedQueryHandler for PostgresServerHandler {
 
         let sql_plan = &portal.statement.statement;
 
+        if sql_plan.query.is_empty() {
+            // early return if query is empty
+            return Ok(Response::EmptyQuery);
+        }
+
+        if let Some(mut resps) = fixtures::process(&sql_plan.query, query_ctx.clone()) {
```
shuiyisong, line 268: Since fixtures are rare cases, should we use `matches` first to short-circuit if it's normal SQL?

sunng87: I think `process` is just `matches` and execute, it makes no difference if it doesn't match. Otherwise we might need to match twice for those actually in our fixture whitelist.

src/servers/src/postgres/types.rs
```diff
@@ -279,298 +279,345 @@ pub(super) fn parameters_to_scalar_values(
...
-                        ));
+                if let Some(server_type) = &server_type {
+                    match server_type {
+                        ConcreteDataType::Binary(_) => {
+                            ScalarValue::Binary(data.map(|d| jsonb::Value::from(d).to_vec()))
+                        }
+                        _ => {
+                            return Err(invalid_parameter_error(
+                                "invalid_parameter_type",
+                                Some(&format!(
```
v0y4g3r, line 612: nitpick: we can change the `detail` argument in `invalid_parameter_error` from `Option<&str>` to `Option<String>` to avoid a clone.

v0y4g3r (approved):
Rest LGTM

shuiyisong (approved):
LGTM

PR #4725, opened 2024-09-11
fix: determine region role by using is_readonly

WenyXu:

determine region role by using `is_readonly`

## Checklist


waynexia (approved):

MichaelScofield:
So region state and region role are interwind: if a region is set to "read only", then it must be a "follower". Is it expected from the aspect of design? @fengjiachun

v0y4g3r:
It's quite subtle that a region is follower if and only if it's read-only 😢

Issue #3686, opened 2024-04-10
Supports JSON format


WenyXu:
### What problem does the new feature solve?

Supports inserting/querying JSON data.

### What does the feature do?

Support for JSON data format will be divided into several stages:

1. Add necessary data type support, including Lists and Nested data
2. Supports inserting JSON data
3. Supports querying JSON data
4. Supports Schemaless JSON

### Implementation challenges

See also: 
- https://github.com/apache/arrow-datafusion/issues/7845
- https://github.com/mongodb/bson-rust
- https://github.com/datafuselabs/jsonb

CookiePieWw:
I'd like to get involved in this. Also, I plan to apply for the corresponding project in OSPP as well. Here's a rough plan.

1. Add necessary data type support, including Lists and Nested data

   The List data type has been supported already. As for Array, I assume it's corresponding to the FixedSizeList in arrow. I wonder if the Nested data is corresponding to the Schema or other types in arrow. If so, I think the type support part can be finished following the implementation of List.

2. Supports inserting JSON data

   As for json itself, since apache arrow has not supported json yet, we can store it in String (as a plain string of the json) or Bytes (as for jsonb).

3. Supports querying JSON data

   This part requires some UDFs in datafusion for querying string datatypes. We can pick up the functions supported by Postgres to see what we can do.

4. Supports Schemaless JSON

  As described above, the Schemaless JSON seems to be json data type itself, and we can call the json with specific schema 'object' or something else?

Since I'm not familiar with the storage and query part of greptimedb, the plan above may be naive. Hope I can make a better proposal next month. Looking forward for your advice :)

WenyXu:
As mentioned in https://github.com/apache/datafusion/issues/7845#issuecomment-2068061465, I was greatly inspired by [JSONA](https://github.com/apache/datafusion/issues/7845#issuecomment-1779821351) and proposed a JSONA variant that may leverage the high compression rates of columnar storage formats. BTW, If we decide to implement our own JSON  storage implementation, It's definitely an excellent opportunity to evaluate various storage implementations of JSON in the OLAP scenario.


## A naive proposal of JSONA variant 

For JSON `[false, 10, {"k":"v"}, null]` can be stored as the following struct.
```
Struct {
    Nodes: StartArray, FALSE, Number, StartObject, Key, String, EndObject, NULL, EndArray
    Offsets: [NULL, NULL, 0, NULL, 0, 0, NULL,…]
    Keys: ["k"]
[...]

PR #4726, opened 2024-09-11
feat: add respective `json_is` UDFs for JSON type

CookiePieWw:

#3686 #4515 #4230

As title. Add following `json_is` UDFs: `json_is_null`, `json_is_array`, `json_is_object`, `json_is_int`, `json_is_float`, `json_is_string`, `json_is_bool`. Also rename `to_json` to `parse_json` and add some sqlness tests.

## Checklist


tests/cases/standalone/common/function/json/json_get.result
```diff
@@ -196,3 +196,66 @@ DROP TABLE jsons;
...
+
+SELECT * FROM jsons WHERE json_get_string(j, 'a.b.c') = 'foo';
+
++--------------------------------------------------------------------------------------+-------------------------+
+| j                                                                                    | ts                      |
++--------------------------------------------------------------------------------------+-------------------------+
+| 40000001100000015000001d614000000110000001500000106240000001100000011000000363666f6f | 1970-01-01T00:00:00.003 |
++--------------------------------------------------------------------------------------+-------------------------+
+
+SELECT * FROM jsons WHERE json_get_bool(j, 'a.b.c') = true;
```
WenyXu, line 245: let's use `json_to_string`

src/common/function/src/scalars/json/json_is.rs
```diff
@@ -0,0 +1,214 @@
...
+    array,
+    "Checks if the input JSONB is a JSON array"
+);
+json_is!(
+    JsonIsObject,
+    object,
+    "Checks if the input JSONB is a JSON object"
+);
+
+mod tests {
```
WenyXu, line 133:
```suggestion
#[cfg(test)]
mod tests {
```

WenyXu (approved):
Great! 🚀 LGTM

sunng87:
Do you think it will be a good idea to have `jq` as a built-in function? I think it can cover most JSON operations so we don't have to implement and document them one-by-one

CookiePieWw:
> Do you think it will be a good idea to have `jq` as a built-in function? I think it can cover most JSON operations so we don't have to implement and document them one-by-one

Yes, `jq` is quite powerful, but it returns an json type. The problem is that when we use UDFs, the binary data of json type are processed through datafusion and then the type info are erased since datafusion has no json type, then the servers would return plain binary data to clients.

We may have 2 choice here:
- let users use `json_to_string` function whenever using json type.
- or let users specify the type they want here.

BTW, databend provide `jq` function as well as `is_<type>` function and many other functions. Maybe it's for flexibility?

paomian:
In my opinion, parse_json should be called by default when not performing any get path operation on a json column, rather than having to call this function manually for every query. Refer to https://www.postgresql.org/docs/current/datatype-json.html

rest LGTM

WenyXu:
> In my opinion, parse_json should be called by default when not performing any get path operation on a json column, rather than having to call this function manually for every query. Refer to https://www.postgresql.org/docs/current/datatype-json.html
> 
> But I think this can be optimized later without blocking this pr merge.
> 
> rest LGTM

Both ways work(i.e., `INSERT INTO table VALUES (parse_json('{ "a": 1}'));`, `INSERT INTO table VALUES ('{ "a": 1}');`)

paomian (approved):

PR #4727, opened 2024-09-12
feat: list/array/timezone support for postgres output

sunng87:

Implement array(postgres) type for list(greptime) output.
Also included a geohash_neighbours function that returns list.

## Checklist


src/common/function/src/scalars/geo/geohash.rs
```diff
@@ -130,6 +132,144 @@ impl Function for GeohashFunction {
...
+    }
+
+    fn return_type(&self, _input_types: &[ConcreteDataType]) -> Result<ConcreteDataType> {
+        Ok(ConcreteDataType::list_datatype(
+            ConcreteDataType::string_datatype(),
+        ))
+    }
+
+    fn signature(&self) -> Signature {
+        let mut signatures = Vec::new();
```
killme2008, line 159: You can use 
```rust
helper::one_of_sigs2(vec![ 
            ConcreteDataType::float32_datatype(),
            ConcreteDataType::float64_datatype()],
            [
               ConcreteDataType::int8_datatype(),
                ConcreteDataType::int16_datatype(),
                ConcreteDataType::int32_datatype(),
                ConcreteDataType::int64_datatype(),
                ConcreteDataType::uint8_datatype(),
                ConcreteDataType::uint16_datatype(),
                ConcreteDataType::uint32_datatype(),
                ConcreteDataType::uint64_datatype(),
             ])
```
instead.

https://github.com/GreptimeTeam/greptimedb/blob/36b1bafbf00152cc89c8d4362ace432bbb1932a3/src/common/function/src/helper.rs#L23

src/common/function/src/scalars/geo/geohash.rs
```diff
@@ -130,6 +132,144 @@ impl Function for GeohashFunction {
...
+                        .into_iter()
+                        .map(Value::from)
+                        .collect(),
+                        ConcreteDataType::string_datatype(),
+                    ))
+                }
+                _ => None,
+            };
+
+            if let Some(list_value) = result {
```
killme2008, line 260: Seems easier:
```rust
result.push(result.map(|list_value| list_value.as_scalar_ref()));
```

sunng87: this is broken due to lifetime of list_value

src/common/function/src/scalars/geo/geohash.rs
sunng87: the helper only works with 2-argument signature. In this case there is 3, especially the frist two are in the same type. That's a little bit different from general use-case

src/common/function/src/scalars/geo/geohash.rs
```diff
@@ -130,6 +132,144 @@ impl Function for GeohashFunction {
...
+
+        let size = lat_vec.len();
+        let mut results =
+            ListVectorBuilder::with_type_capacity(ConcreteDataType::string_datatype(), size);
+
+        for i in 0..size {
+            let lat = lat_vec.get(i).as_f64_lossy();
+            let lon = lon_vec.get(i).as_f64_lossy();
+            let r = match resolution_vec.get(i) {
+                Value::Int8(v) => v as usize,
```
v0y4g3r, line 210: It's not always safe to cast a signed integer to usize since it underflows to large numbers instead of just panic where the signed number is negative.

src/servers/src/postgres/types.rs
```diff
@@ -58,6 +59,317 @@ pub(super) fn schema_to_pg(origin: &Schema, field_formats: &Format) -> Result<Ve
         .collect::<Result<Vec<FieldInfo>>>()
 }
 
+fn encode_array(
```
v0y4g3r, line 62: Can we use some macro to simplify this?

src/servers/Cargo.toml
```diff
@@ -68,7 +68,7 @@ openmetrics-parser = "0.4"
 opensrv-mysql = "0.7.0"
 opentelemetry-proto.workspace = true
 parking_lot = "0.12"
-pgwire = "0.20"
+pgwire = "0.20.2"
```
v0y4g3r, line 71: Cargo will ignore the patch version you specified and automatically update to 0.20.2 in Cargo.lock when you run `cargo update pgwire`. Please see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

v0y4g3r (changes requested):

src/common/function/src/scalars/geo/geohash.rs
sunng87: Just added range check before conversion

src/servers/Cargo.toml
sunng87: Here I need to specify this changeset only works for pgwire >= 0.20.2 because the patch version ships key fixes for this feature. This will make cargo.lock reproducible.

src/servers/src/postgres/types.rs
sunng87: It's hard to find a common abstraction for these match branches. Using macro may fragment the code and sacrifice readability quite a bit.

src/servers/Cargo.toml
v0y4g3r: > Here I need to specify this changeset only works for pgwire >= 0.20.2 because the patch version ships key fixes for this feature. This will make cargo.lock reproducible.

But this change does not actually pin the patch version. To do this you need:
```
pgwire = "=0.20.2"
```

sunng87: No, this will pin pgwire to 0.20.2 forever, which can be a resolver issue when there is a version conflict. All we need is 0.20.2 or later.

v0y4g3r (approved):
LGTM

killme2008 (approved):
LGTM

Issue #4679, opened 2024-09-04
Customize channel information for sqlness tests


sunng87:
### What type of enhancement is this?

Other

### What does the enhancement do?

Our sqlness tests are using http api for sql test. Now with changes like #4580 different protocol/channel may behaviour different. It would be helpful to be able to test different channel via sqlness tests.

J0HN50N133:
May I try?

sunng87:
@J0HN50N133 Yes this is for you. I just can't assign it to you without you participating this thread. :laughing:

PR #4729, opened 2024-09-16
feat: customize channel information for sqlness tests #4679

J0HN50N133:

#4679

## Checklist


tests/runner/src/protocol_interceptor.rs
```diff
@@ -0,0 +1,47 @@
...
+use sqlness::interceptor::{Interceptor, InterceptorFactory, InterceptorRef};
+use sqlness::SqlnessError;
+
+pub const PROTOCOL_KEY: &str = "protocol";
+
+pub const POSTGRES: &str = "postgres";
+pub const MYSQL: &str = "mysql";
+
+pub const PREFIX: &str = "PROTOCOL";
+pub struct ProtocolInterceptor {
```
evenyag, line 24:
```suggestion
pub const PROTOCOL_KEY: &str = "protocol";
pub const POSTGRES: &str = "postgres";
pub const MYSQL: &str = "mysql";
pub const PREFIX: &str = "PROTOCOL";

pub struct ProtocolInterceptor {
```

tests/runner/src/protocol_interceptor.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+pub struct ProtocolInterceptorFactory;
+impl Interceptor for ProtocolInterceptor {
+    fn before_execute(&self, _: &mut Vec<String>, context: &mut sqlness::QueryContext) {
+        context
+            .context
+            .insert(PROTOCOL_KEY.to_string(), self.protocol.clone());
+    }
+}
+impl InterceptorFactory for ProtocolInterceptorFactory {
```
evenyag, line 35:
```suggestion
}

impl Interceptor for ProtocolInterceptor {
    fn before_execute(&self, _: &mut Vec<String>, context: &mut sqlness::QueryContext) {
        context
            .context
            .insert(PROTOCOL_KEY.to_string(), self.protocol.clone());
    }
}

pub struct ProtocolInterceptorFactory;

impl InterceptorFactory for ProtocolInterceptorFactory {
```

tests/runner/src/protocol_interceptor.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn try_new(&self, ctx: &str) -> Result<InterceptorRef, SqlnessError> {
+        match ctx.to_lowercase().as_str() {
+            POSTGRES | MYSQL => Ok(Box::new(ProtocolInterceptor {
+                protocol: ctx.to_lowercase().to_string(),
+            })),
```
evenyag, line 40:
```suggestion
        let protocol = ctx.to_lowercase();
        match protocol.as_str() {
            POSTGRES | MYSQL => Ok(Box::new(ProtocolInterceptor {
                protocol,
            })),
```

tests/runner/src/env.rs
```diff
@@ -635,3 +739,117 @@ impl Display for ResultDisplayer {
...
+        let result_displayer = ResultDisplayer {
+            result: Ok(Output::new_with_record_batches(recordbatches)),
+        };
+        write!(f, "{}", result_displayer)?;
+
+        Ok(())
+    }
+}
+
+struct MysqlFormatter {
```
evenyag, line 806: Do we have sqlness tests that use the MySQL formatter? We might need to add a test for this.

tests/runner/src/env.rs
```diff
@@ -635,3 +739,117 @@ impl Display for ResultDisplayer {
...
+        }
+
+        // convert to string vectors
+        let mut columns: Vec<StringVectorBuilder> = (0..schema.num_columns())
+            .map(|_| StringVectorBuilder::with_capacity(schema.num_columns()))
+            .collect();
+        for row in self.rows.iter().skip(1) {
+            if let PgRow::Row(row) = row {
+                for (i, column) in columns.iter_mut().enumerate().take(schema.num_columns()) {
+                    column.push(row.get(i));
```
evenyag, line 785: Is it possible that the output column doesn't have the string type?

tests/runner/src/env.rs
```diff
@@ -635,3 +739,117 @@ impl Display for ResultDisplayer {
...
+                .collect(),
+        ));
+
+        // convert to string vectors
+        let mut columns: Vec<StringVectorBuilder> = (0..schema.num_columns())
+            .map(|_| StringVectorBuilder::with_capacity(schema.num_columns()))
+            .collect();
+        for row in self.rows.iter() {
+            for (i, name) in names.iter().enumerate() {
+                columns[i].push(row.get::<String, &str>(name).as_deref());
```
evenyag, line 837: Do we need to deal with other types?

tests/runner/src/env.rs
J0HN50N133: No. Pg's simple query protocol return [every retried data in text](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-SIMPLE-QUERY:~:text=In%20simple%20Query%20mode%2C%20the%20format%20of%20retrieved%20values%20is%20always%20text%2C%20except%20when%20the%20given%20command%20is%20a%20FETCH%20from%20a%20cursor%20declared%20with%20the%20BINARY%20option.%20In%20that%20case%2C%20the%20retrieved%20values%20are%20in%20binary%20format.%20The%20format%20codes%20given%20in%20the%20RowDescription%20message%20tell%20which%20format%20is%20being%20used.).

tests/runner/src/env.rs
J0HN50N133: I need some help from @sunng87. He've mentioned there're also some mysql specific behavior https://github.com/GreptimeTeam/greptimedb/pull/4580#issuecomment-2312991800

tests/runner/src/env.rs
J0HN50N133: No, similar to pg's [situation](https://docs.rs/mysql/25.0.1/mysql/prelude/trait.Queryable.html#tymethod.query_iter).

J0HN50N133:
For some reason, the connection to PostgreSQL is closed in the CI environment. Does anyone have any idea?

tests/runner/src/env.rs
sunng87: you can use either of 

- `show @@tx_isolation`
- `show @@version_comment`

to test mysql specific interface

tests/runner/src/env.rs
```diff
@@ -520,6 +605,25 @@ impl Database for GreptimeDB {
     }
 }
 
+#[async_trait]
+impl Database for GreptimeDB {
+    async fn query(&self, ctx: QueryContext, query: String) -> Box<dyn Display> {
+        if ctx.context.contains_key("restart") && self.env.server_addrs.server_addr.is_none() {
+            self.env.restart_server(self).await;
```
sunng87, line 612: @J0HN50N133 note that the server restarts here. you will need to recreate the connections for mysql/postgres

if you want to keep it simple, you can create mysql/postgres connection for each `query`

tests/cases/standalone/common/mysql.result
```diff
@@ -0,0 +1,21 @@
+-- SQLNESS PROTOCOL MYSQL
+SHOW @@tx_isolation;
```
sunng87, line 2: sorry, it was my mistake the statement should `select @@tx_isolation` and `select @@version_comment`

sunng87 (approved):

tests/cases/standalone/common/mysql.result
J0HN50N133: It's ok. I've replace the statement

evenyag:
@coderabbitai summary

waynexia:
merge `main` to update the lock file

tests/runner/src/main.rs
```diff
@@ -30,6 +33,23 @@ enum Wal {
...
+
+    /// Address of the postgres server. Must be set if server_addr is set.
+    #[clap(short, long, requires = "server_addr")]
+    pg_server_addr: Option<String>,
+
+    /// Address of the mysql server. Must be set if server_addr is set.
+    #[clap(short, long, requires = "server_addr")]
+    mysql_server_addr: Option<String>,
+}
+
```
waynexia, line 52: Which would be used by default without a protocol interceptor?

tests/runner/Cargo.toml
```diff
@@ -15,8 +15,11 @@ common-error.workspace = true
 common-query.workspace = true
 common-recordbatch.workspace = true
 common-time.workspace = true
+datatypes = { workspace = true }
+mysql = { version = "23.0.1" }
 serde.workspace = true
 serde_json.workspace = true
+tokio-postgres = { workspace = true }
```
waynexia, line 22: We are using `sqlx` in the integration test, is it possible to use it here as well?

J0HN50N133: With sqlx text we have to handle many decode logic to convert value into text. I'm not sure whether it's as convenient as this two crates. Since `sqlness` somehow independent to the core part,  I think it's ok to introduce a new dependency?

J0HN50N133:
@waynexia can we push forward this pr?

tests/runner/src/main.rs
evenyag: I guess the default is the grpc server. @J0HN50N133 Could you confirm this?

tests/runner/Cargo.toml
evenyag: > We are using `sqlx` in the integration test, is it possible to use it here as well?

We can merge this PR first and create another issue to discuss it. @waynexia @J0HN50N133

evenyag (approved):
I think we can move forward and create an issue to see if we can switch to sqlx.

@J0HN50N133 Please confirm the remaining comments and resolve the conflicts.

killme2008:
@J0HN50N133 Thank you for your hard work and patience.

waynexia:
> @waynexia can we push forward this pr?

Sorry for the delay. The idea behind preferring `sqlx` is it's wider accepted based on stats in crates.io, and thus is more aligned with end users. But this is not a hard requirement after all 😉

PR #4734, opened 2024-09-18
chore: add log ingest interceptor

shuiyisong:

This pr mainly adds `LogIngestInterceptor` to intercept execution before the pipeline and ingestion.

## Checklist

### DIFF
--- a/src/servers/src/interceptor.rs
+++ b/src/servers/src/interceptor.rs
@@ -407,19 +407,19 @@ pub trait LogIngestInterceptor {
     /// Called before pipeline execution.
     fn pre_pipeline(
         &self,
-        _values: &[Value],
+        values: Vec<Value>,
         _query_ctx: QueryContextRef,
-    ) -> Result<(), Self::Error> {
-        Ok(())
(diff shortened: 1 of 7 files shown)


src/servers/src/interceptor.rs
```diff
@@ -397,3 +398,61 @@ impl<E: ErrorExt> PromStoreProtocolInterceptor for Option<PromStoreProtocolInter
...
+    }
+
+    /// Called before insertion.
+    fn pre_ingest(
+        &self,
+        _request: &RowInsertRequests,
+        _query_ctx: QueryContextRef,
+    ) -> Result<(), Self::Error> {
+        Ok(())
+    }
```
sunng87, line 423: Not sure if it's fully possible but if we can take values and request, then return them from interceptor, it can be great. This will allow us to modify the payload of request.

fengys1996 (approved):

sunng87 (approved):

sunng87:
In the next patch, we can merge validator into interceptor

PR #4737, opened 2024-09-18
refactor(tables): improve `tables` performance

v0y4g3r:

This patch tries to solve some user's complaint about the lag while switching database in mysql cli by enabling concurrent tables iteration in information_schema to accelerate table rehash.

It also introduces some refactor regarding reading table information in kvbackend.

For a database with 100k tables, it cuts `show tables` cost time from 430ms to 200ms.
```
# From
$ time mysql -h 127.0.0.1 -P 4002 -e "show tables" > /dev/null 2>&1
mysql -h 127.0.0.1 -P 4002 -e "show tables" > /dev/null 2>&1  0.01s user 0.00s system 7% cpu 0.432 total

# To
$ time mysql -h 127.0.0.1 -P 4002 -e "show tables" > /dev/null 2>&1
mysql -h 127.0.0.1 -P 4002 -e "show tables" > /dev/null 2>&1  0.01s user 0.00s system 7% cpu 0.201 total
```
## Checklist

### DIFF
--- a/src/catalog/src/kvbackend/manager.rs
+++ b/src/catalog/src/kvbackend/manager.rs
@@ -42,6 +42,7 @@ use table::dist_table::DistTable;
 use table::table::numbers::{NumbersTable, NUMBERS_TABLE_NAME};
 use table::table_name::TableName;
 use table::TableRef;
+use tokio_stream::wrappers::UnboundedReceiverStream;
 
 use crate::error::{
     CacheNotFoundSnafu, GetTableCacheSnafu, InvalidTableInfoInCatalogSnafu, ListCatalogsSnafu,
@@ -182,18 +183,15 @@ impl CatalogManager for KvBackendCatalogManager {
         let stream = self
             .table_metadata_manager
             .table_name_manager()
-            .tables(catalog, schema);
-        let mut tables = stream
+            .tables(catalog, schema)
+            .map_ok(|(table_name, _)| table_name)
             .try_collect::<Vec<_>>()
             .await
             .map_err(BoxedError::new)
-            .context(ListTablesSnafu { catalog, schema })?
-            .into_iter()
-            .map(|(k, _)| k)
-            .collect::<Vec<_>>();
-        tables.extend_from_slice(&self.system_catalog.table_names(schema, query_ctx));
+            .context(ListTablesSnafu { catalog, schema })?;
 
-        Ok(tables.into_iter().collect())
+        tables.extend(self.system_catalog.table_names(schema, query_ctx));
+        Ok(tables)
     }
 
     async fn catalog_exists(&self, catalog: &str) -> Result<bool> {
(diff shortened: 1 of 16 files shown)


src/catalog/src/kvbackend/manager.rs
```diff
@@ -303,36 +301,70 @@ impl CatalogManager for KvBackendCatalogManager {
...
+                        }
+                    }
+                }));
+                if task_batch.len() == CONCURRENCY {
+                    let _ = futures::future::join_all(std::mem::take(&mut task_batch)).await;
                 }
             }
+            if !task_batch.is_empty() {
+                let _ = futures::future::join_all(std::mem::take(&mut task_batch)).await;
+            }
```
WenyXu, line 364: How about using [Semaphore](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#)?

src/catalog/src/kvbackend/manager.rs
```diff
@@ -303,36 +301,70 @@ impl CatalogManager for KvBackendCatalogManager {
...
-        let table_id_stream = self
-            .table_metadata_manager
-            .table_name_manager()
-            .tables(catalog, schema)
-            .map_ok(|(_, v)| v.table_id());
         const BATCH_SIZE: usize = 128;
-        let user_tables = try_stream!({
+        const CONCURRENCY: usize = 8;
+
+        let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
```
WenyXu, line 307: I think the bounded channel should also work well

WenyXu (commented):
My elegant state machine 🥲

src/catalog/src/kvbackend/manager.rs
```diff
@@ -303,36 +301,70 @@ impl CatalogManager for KvBackendCatalogManager {
...
+                    Ok(table_ids) => table_ids,
+                    Err(e) => {
+                        let _ = tx.send(Err(e));
+                        return;
+                    }
+                };
+
+                let metadata_manager = metadata_manager.clone();
+                let tx = tx.clone();
+                task_batch.push(tokio::spawn(async move {
```
MichaelScofield, line 338: Why not use one of our common runtimes to spawn the tasks? It makes the async tasks better governed and monitored in the future.

src/catalog/src/kvbackend/manager.rs
v0y4g3r: > I think the bounded channel should also work well

We don't explicitly set the bound here since the concurrency if stream if harnessed.

src/catalog/src/kvbackend/manager.rs
v0y4g3r: > How about using [Semaphore](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#)?

fixed in 1d55a5f3cdd0e3d6bea62cf5948a91bafb8e7c65

src/catalog/src/kvbackend/manager.rs
v0y4g3r: > I think the bounded channel should also work well

fixed in 1d55a5f3cdd0e3d6bea62cf5948a91bafb8e7c65

src/catalog/src/kvbackend/manager.rs
v0y4g3r: > Why not use one of our common runtimes to spawn the tasks? It makes the async tasks better governed and monitored in the future.

fixed in [24ff86b](https://github.com/GreptimeTeam/greptimedb/pull/4737/commits/24ff86bfb5809f4ef6ac5b56ac2c0f727762f62e)

v0y4g3r:
@WenyXu @MichaelScofield PTAL

MichaelScofield (approved):

WenyXu (approved):

PR #4744, opened 2024-09-19
chore: add json write

paomian:

* If the pipeline parameter is empty, the data is written directly.

## Checklist

### DIFF
--- a/src/pipeline/src/etl/error.rs
+++ b/src/pipeline/src/etl/error.rs
@@ -40,7 +40,7 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("{processor} processor: missing field: {field}"))]
+    #[snafu(display("Processor {processor}: missing field: {field}"))]
     ProcessorMissingField {
         processor: String,
         field: String,
@@ -48,7 +48,7 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("{processor} processor: expect string value, but got {v:?}"))]
+    #[snafu(display("Processor {processor}: expect string value, but got {v:?}"))]
     ProcessorExpectString {
         processor: String,
         v: crate::etl::Value,
@@ -56,7 +56,7 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("{processor} processor: unsupported value {val}"))]
+    #[snafu(display("Processor {processor}: unsupported value {val}"))]
     ProcessorUnsupportedValue {
         processor: &'static str,
         val: String,
@@ -64,13 +64,13 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("processor key must be a string"))]
+    #[snafu(display("Processor key must be a string"))]
     ProcessorKeyMustBeString {
         #[snafu(implicit)]
         location: Location,
     },
 
-    #[snafu(display("{kind} processor: failed to parse {value}"))]
+    #[snafu(display("Processor {kind}: failed to parse {value}"))]
     ProcessorFailedToParseString {
         kind: String,
         value: String,
@@ -78,13 +78,13 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("processor must have a string key"))]
+    #[snafu(display("Processor must have a string key"))]
     ProcessorMustHaveStringKey {
         #[snafu(implicit)]
         location: Location,
     },
 
-    #[snafu(display("unsupported {processor} processor"))]
+    #[snafu(display("Unsupported {processor} processor"))]
     UnsupportedProcessor {
         processor: String,
         #[snafu(implicit)]
@@ -108,7 +108,7 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("failed to parse {key} as int: {value}"))]
+    #[snafu(display("Failed to parse {key} as int: {value}"))]
     FailedToParseIntKey {
         key: String,
         value: String,
@@ -126,7 +126,7 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
-    #[snafu(display("failed to parse {key} as float: {value}"))]
+    #[snafu(display("Failed to parse {key} as float: {value}"))]
     FailedToParseFloatKey {
         key: String,
         value: String,
@@ -136,7 +136,7 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("{kind} processor.{key} not found in intermediate keys"))]
+    #[snafu(display("Processor {kind}: {key} not found in intermediate keys"))]
     IntermediateKeyIndex {
         kind: String,
         key: String,
@@ -144,41 +144,41 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("{k} missing value in {s}"))]
+    #[snafu(display("Cmcd {k} missing value in {s}"))]
     CmcdMissingValue {
         k: String,
         s: String,
         #[snafu(implicit)]
         location: Location,
     },
-    #[snafu(display("{part} missing key in {s}"))]
+    #[snafu(display("Part: {part} missing key in {s}"))]
     CmcdMissingKey {
         part: String,
         s: String,
         #[snafu(implicit)]
         location: Location,
     },
-    #[snafu(display("key must be a string, but got {k:?}"))]
+    #[snafu(display("Key must be a string, but got {k:?}"))]
     KeyMustBeString {
         k: yaml_rust::Yaml,
         #[snafu(implicit)]
         location: Location,
     },
 
-    #[snafu(display("csv read error"))]
+    #[snafu(display("Csv read error"))]
     CsvRead {
         #[snafu(implicit)]
         location: Location,
         #[snafu(source)]
         error: csv::Error,
     },
-    #[snafu(display("expected at least one record from csv format, but got none"))]
+    #[snafu(display("Expected at least one record from csv format, but got none"))]
     CsvNoRecord {
         #[snafu(implicit)]
         location: Location,
     },
 
-    #[snafu(display("'{separator}' must be a single character, but got '{value}'"))]
+    #[snafu(display("Separator '{separator}' must be a single character, but got '{value}'"))]
     CsvSeparatorName {
         separator: &'static str,
         value: String,
@@ -186,7 +186,7 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("'{quote}' must be a single character, but got '{value}'"))]
+    #[snafu(display("Quote '{quote}' must be a single character, but got '{value}'"))]
     CsvQuoteName {
         quote: &'static str,
         value: String,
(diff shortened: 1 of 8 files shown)


src/servers/src/http/event.rs
```diff
@@ -410,7 +408,7 @@ fn extract_pipeline_value_by_content_type(
 
 async fn ingest_logs_inner(
     state: LogHandlerRef,
-    pipeline_name: String,
+    pipeline_name: Option<String>,
```
shuiyisong, line 411: Using just `Option` to switch modes seems like a weak constraint, which can be unconsciously triggered by mistake. I was thinking of adding a bool flag to switch to 'auto-create' JSON mode by force, but that seems too heavy.
What do you think, @sunng87

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -191,3 +197,214 @@ impl Transformer for GreptimeTransformer {
...
+) {
+    if let Some(index) = index {
+        let api_value = GreptimeValue {
+            value_data: Some(value_data),
+        };
+        let value_column_data_type = proto_value_type(&api_value);
+        // safety unwrap is fine here because index is always valid
+        let schema_column_data_type = schema.get(index).unwrap().datatype();
+        if value_column_data_type.is_some_and(|t| t != schema_column_data_type) {
+            row[index] = GreptimeValue { value_data: None };
```
shuiyisong, line 216: How about throwing an error here?

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -191,3 +197,214 @@ impl Transformer for GreptimeTransformer {
...
+fn json_value_to_row(schemas: &mut Vec<ColumnSchema>, map: Map<String, serde_json::Value>) -> Row {
+    let mut row: Vec<GreptimeValue> = Vec::with_capacity(schemas.len());
+    for _ in 0..schemas.len() {
+        row.push(GreptimeValue { value_data: None });
+    }
+    for (key, value) in map {
+        if key == DEFAULT_GREPTIME_TIMESTAMP_COLUMN {
+            continue;
+        }
+        let index = schemas.iter().position(|x| x.column_name == key);
```
shuiyisong, line 238: nitpick: We could have a map<key, index> to speed up finding the index.

shuiyisong (commented):
We should add some integration tests if time permits.

src/pipeline/src/etl/transform/transformer/greptime.rs
paomian: I'm not sure which is better.

src/pipeline/src/etl/transform/transformer.rs
```diff
@@ -13,3 +13,4 @@
 // limitations under the License.
 
 pub mod greptime;
+pub use greptime::identify_pipeline;
```
sunng87, line 16: identity

src/servers/src/http/event.rs
sunng87: How about a enum:

```
PipelineChoice{
  UserDefined(String)
  Auto
```

paomian: The question here is whether to add a parameter to specify the use of auto pipeline or to treat it as if it were used without the pipeline_name. or some other way to specify auto pipeline.

shuiyisong: On second thought, I think we could use `None` to indicate direct log write. This might usually happen in debugging or quick demo phases.

Do we need to validate the `Content-Type` to be JSON only in direct mode? If not, it would be the whole line as a column field and a `greptime_timestamp` as a timestamp index, which seems ok to me.

src/pipeline/src/etl/transform/transformer/greptime.rs
killme2008: I think raising an error is better when the column data type is changed.

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -194,3 +201,225 @@ impl Transformer for GreptimeTransformer {
         &mut self.transforms
     }
 }
+
+#[derive(Debug, Default)]
+struct SchemaInfo {
+    schema: Vec<ColumnSchema>,
+    index: HashMap<String, usize>,
```
killme2008, line 208: Please describe the key and value in this map.

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -194,3 +201,225 @@ impl Transformer for GreptimeTransformer {
...
+        row.push(api_value);
+    }
+}
+
+fn json_value_to_row(schema_info: &mut SchemaInfo, map: Map<String, serde_json::Value>) -> Row {
+    let mut row: Vec<GreptimeValue> = Vec::with_capacity(schema_info.schema.len());
+    for _ in 0..schema_info.schema.len() {
+        row.push(GreptimeValue { value_data: None });
+    }
+    for (key, value) in map {
```
killme2008, line 246: The key is the column name, why not just use `column_name` instead?

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -194,3 +201,225 @@ impl Transformer for GreptimeTransformer {
...
+                        semantic_type: SemanticType::Field as i32,
+                        datatype_extension: None,
+                        options: None,
+                    },
+                    &mut row,
+                    schema_info,
+                );
+            }
+            serde_json::Value::Number(n) => {
+                if n.is_i64() {
```
killme2008, line 286: Do we need to do the type conversion here? For example, i64 -> f64 etc.

killme2008: The user writes the f64 at first time, but then the value is an integer, I think we should take care of such cases.

src/servers/src/http/event.rs
killme2008: I think we don't need to change the pipeline type to `Option`. The user must specify the pipeline name even when using the `identify` pipeline. We can ensure that the `identity` pipeline is the internal pipeline, and maybe we will add more internal pipelines in the future. Just checking if the pipeline name is reserved when the user creates new pipelines.

src/pipeline/src/etl/transform/transformer/greptime.rs
paomian: I don't think auto-conversion is the way to go here. If the types don't match, it'll probably report an error. Seems like that'd be a better option.

src/servers/src/http/event.rs
paomian: I also favor the idea of defining some of the internal pipeline names. I've tentatively set all pipelines starting with `greptime_` to be for our internal use, and prohibited users from creating pipeline names starting with `greptime_`.

shuiyisong (approved):

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -194,3 +201,298 @@ impl Transformer for GreptimeTransformer {
...
+        if value_column_data_type.is_some_and(|t| t != schema_column_data_type) {
+            IdentifyPipelineColumnTypeMismatchSnafu {
+                column: column_schema.column_name,
+                original: schema_column_data_type.as_str_name(),
+                now: value_column_data_type
+                    .map(|t| t.as_str_name())
+                    .unwrap_or_else(|| "null"),
+            }
+            .fail()
+        } else {
```
shuiyisong, line 239: We could use `unwrap()` here since it's already inside `is_some_and()`.

src/pipeline/src/etl/error.rs
```diff
@@ -537,6 +537,15 @@ pub enum Error {
...
     },
+
+    #[snafu(display("Column type mismatch. column: {column}, original: {original}, now: {now}"))]
+    IdentifyPipelineColumnTypeMismatch {
+        column: String,
+        original: String,
+        now: String,
+        #[snafu(implicit)]
+        location: Location,
+    },
```
WenyXu, line 548: nit: 
```suggestion
    #[snafu(display("Column datatype mismatch. For column: {column}, expected datatype: {expected}, actual datatype: {actual}"))]
    IdentifyPipelineColumnTypeMismatch {
        column: String,
        expected: String,
        actual: String,
        #[snafu(implicit)]
        location: Location,
    },
```

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -120,6 +126,7 @@ impl Transformer for GreptimeTransformer {
             if let Some(idx) = transform.index {
                 if idx == Index::Time {
                     match transform.real_fields.len() {
+                        //safety unwrap is fine here because we have checked the length of real_fields
```
WenyXu, line 129: nit: styling
```suggestion
                        //Safety: unwrap is fine here because we have checked the length of real_fields
```

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -194,3 +201,297 @@ impl Transformer for GreptimeTransformer {
...
+    value_data: ValueData,
+    column_schema: ColumnSchema,
+    row: &mut Vec<GreptimeValue>,
+    schema_info: &mut SchemaInfo,
+) -> Result<()> {
+    if let Some(index) = index {
+        let api_value = GreptimeValue {
+            value_data: Some(value_data),
+        };
+        // safety unwrap is fine here because api_value is always valid
```
WenyXu, line 227: ditto

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -194,3 +201,297 @@ impl Transformer for GreptimeTransformer {
...
+        let value_column_data_type = proto_value_type(&api_value).unwrap();
+        // safety unwrap is fine here because index is always valid
```
WenyXu, line 229: ditto

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -194,3 +201,297 @@ impl Transformer for GreptimeTransformer {
...
+            SemanticType::Field as i32,
+        )
+    } else if n.is_f64() {
+        (
+            ValueData::F64Value(n.as_f64().unwrap()),
+            ColumnDataType::Float64 as i32,
+            SemanticType::Field as i32,
+        )
+    } else {
+        unreachable!("unexpected number type");
```
WenyXu, line 280: I prefer to throw an error? or add panic note in function comment

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -194,3 +201,297 @@ impl Transformer for GreptimeTransformer {
...
+                    &mut row,
+                    schema_info,
+                )?;
+            }
+        }
+    }
+    Ok(Row { values: row })
+}
+
+pub fn identity_pipeline(array: Vec<serde_json::Value>) -> Result<Rows> {
```
WenyXu, line 369: let's add some comments for the public function.

WenyXu (approved):

PR #4746, opened 2024-09-19
feat: flush other workers if still need flush

evenyag:

This PR checks all workers to see if we need to flush them after a flush job is finished. An inactive worker may keep a large write buffer but never flush it. Then other workers may flush very frequently since the global write buffer size is still high.

## Checklist


v0y4g3r (approved):
LGTM

waynexia (approved):

PR #4755, opened 2024-09-22
feat: add API to write OpenTelemetry logs to GreptimeDB

paomian:

* add API to write OpenTelemetry logs to GreptimeDB

## Checklist


src/servers/src/http/otlp.rs
```diff
@@ -93,20 +87,139 @@ pub async fn traces(
...
+}
+
+fn pipeline_header_error(
+    header: &HeaderValue,
+) -> StdResult<String, (http::StatusCode, &'static str)> {
+    match header.to_str() {
+        Ok(s) => Ok(s.to_string()),
+        Err(_) => Err((
+            StatusCode::BAD_REQUEST,
+            "`X-Pipeline-Name` or `X-Pipeline-Version` header is not string type.",
```
sunng87, line 110: to_str doesn't mean the value is not string. http headers are always string. we can return the nested error message.

Again, don't reinvent error handling in this module.

src/servers/src/otlp/logs.rs
```diff
@@ -0,0 +1,543 @@
...
+    map.insert("ResourceAttributes".to_string(), resource_attr);
+    map.insert("ScopeSchemaUrl".to_string(), scope_schema_url);
+    map.insert("ScopeName".to_string(), scope_name);
+    map.insert("ScopeVersion".to_string(), scope_version);
+    map.insert("ScopeAttributes".to_string(), scope_attrs);
+    map.insert("LogAttributes".to_string(), log_attrs);
+    PipelineValue::Map(Map { values: map })
+}
+
+fn build_identity_schema() -> Vec<ColumnSchema> {
```
sunng87, line 194: by identity pipeline, I prefer it's an auto generated pipeline and schema based on input data. Here we already manually defined a schema for otel logs. So we better not to name it and assume it the *identity* pipeline

paomian: I've considered ways to automatically generate a pipeline. But it still has to convert the `ExportLogsServiceRequest` into a `PipelineValue` before it can be processed, so for performance reasons I skipped that step and just converted the ExportLogsServiceRequest into a Rows

src/servers/src/otlp/logs.rs
```diff
@@ -0,0 +1,543 @@
...
+            "log_attributes",
+            ColumnDataType::Binary,
+            SemanticType::Field,
+            Some(ColumnDataTypeExtension {
+                type_ext: Some(TypeExt::JsonType(JsonTypeExtension::JsonBinary.into())),
+            }),
+            None,
+        ),
+        (
+            "timestamp",
```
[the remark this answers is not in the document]
paomian: This is the otlp field, and I think it's better to leave it as is. Refer to https://opentelemetry.io/docs/specs/otel/logs/data-model/#log-and-event-record-definition

src/servers/src/otlp/logs.rs
```diff
@@ -0,0 +1,543 @@
...
+            None,
+            None,
+        ),
+        (
+            "resource_schema_url",
+            ColumnDataType::String,
+            SemanticType::Field,
+            None,
+            None,
+        ),
```
shuiyisong, line 232: I don't think we need to keep these `schema_url`s.

tests-integration/tests/http.rs
```diff
@@ -1490,6 +1492,41 @@ pub async fn test_otlp_traces(store_type: StorageType) {
...
+
+    // TODO(qtang): we show convert jsonb to json string in http sql API
+    let expected = r#"[["","",[64,0,0,0],"https://opentelemetry.io/schemas/1.0.0/scopeLogs","https://opentelemetry.io/schemas/1.0.0/resourceLogs",[64,0,0,1,16,0,0,13,16,0,0,19,114,101,115,111,117,114,99,101,95,97,116,116,114,114,101,115,111,117,114,99,101,45,97,116,116,114,45,118,97,108,45,49],[64,0,0,2,16,0,0,8,16,0,0,3,16,0,0,4,16,0,0,3,99,117,115,116,111,109,101,114,101,110,118,97,99,109,101,100,101,118],1581452773000000789,1581452773000000789,"30","30",1,"Info",9,"null"],["","",[64,0,0,0],"https://opentelemetry.io/schemas/1.0.0/scopeLogs","https://opentelemetry.io/schemas/1.0.0/resourceLogs",[64,0,0,1,16,0,0,13,16,0,0,19,114,101,115,111,117,114,99,101,95,97,116,116,114,114,101,115,111,117,114,99,101,45,97,116,116,114,45,118,97,108,45,49],[64,0,0,2,16,0,0,3,16,0,0,12,16,0,0,6,32,0,0,2,97,112,112,105,110,115,116,97,110,99,101,95,110,117,109,115,101,114,118,101,114,64,1],1581452773000009875,1581452773000009875,"3038303430323031303030303030303030303030303030303030303030303030","30313032303430383030303030303030",1,"Info",9,"null"]]"#;
+    validate_data(&client, "select * from logs;", expected).await;
```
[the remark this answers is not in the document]
shuiyisong: The expected value seems to lack all the JSON field values, like those attribute maps. Also, the log body seems not appear in the result.

src/pipeline/src/etl.rs
```diff
@@ -205,6 +205,34 @@ where
...
+                        result[search_from + pos] = payload_value;
+                        search_from += pos;
+                    }
+                }
+            }
+            Value::String(_) => {
+                result[0] = val;
+            }
+            _ => {
+                return Err("expect object".to_string());
```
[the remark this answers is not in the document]
paomian: In the next release, I will modify the error type. At present, the error types in the Pipeline crate are of the String type.

src/servers/src/otlp/logs.rs
```diff
@@ -0,0 +1,543 @@
...
+                log.time_unix_nano as i64,
+            )),
+        },
+        GreptimeValue {
+            value_data: Some(ValueData::TimestampNanosecondValue(
+                log.observed_time_unix_nano as i64,
+            )),
+        },
+        GreptimeValue {
+            value_data: Some(ValueData::StringValue(bytes_to_hex_string(&log.trace_id))),
```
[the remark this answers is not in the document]
paomian: I think we should store anomalies. We just need to store the data, verifying if it's legit I don't think we need to do.

src/servers/src/http/otlp.rs
```diff
@@ -34,13 +43,18 @@ use super::header::{write_cost_header_map, CONTENT_TYPE_PROTOBUF};
 use crate::error::{self, Result};
 use crate::query_handler::OpenTelemetryProtocolHandlerRef;
 
+const OTLP_GREPTIME_PIPELINE_NAME_HEADER_NAME: &str = "x-greptime-pipeline-name";
+const OTLP_GREPTIME_PIPELINE_VERSION_HEADER_NAME: &str = "x-greptime-pipeline-version";
+const OTLP_GREPTIME_TABLE_NAME_HEADER_NAME: &str = "x-greptime-table-name";
```
sunng87, line 48: I suggest to rename all these three headers to add log, say

- `x-greptime-log-pipeline-name`
- `x-greptime-log-pipeline-version`
- `x-greptime-log-table-name`

Because in otel collectors and api/sdk, the headers may be configured for all of metrics, logs and traces. We have to be able to know the table name is for logs instead of traces.

sunng87 (approved):

src/pipeline/src/etl/value.rs
```diff
@@ -110,8 +113,8 @@ impl Value {
                 _ => Ok(Value::Timestamp(Timestamp::Nanosecond(0))),
             },
 
-            "array" => Ok(Value::Array(Array::default())),
-            "map" => Ok(Value::Map(Map::default())),
+            // We only consider object and array to be json types. and use Map to represent json
+            "json" => Ok(Value::Map(Map::default())),
```
shuiyisong, line 117: Feels strange here. Maybe add a TODO to change it later in a separate PR.

src/pipeline/src/etl/value.rs
```diff
@@ -331,3 +333,79 @@ impl TryFrom<&yaml_rust::Yaml> for Value {
...
+                    let val: JsonbValue = v.into();
+                    map.insert(k.to_string(), val);
+                }
+                JsonbValue::Object(map)
+            }
+        }
+    }
+}
+
+impl<'a> From<Value> for JsonbValue<'a> {
```
shuiyisong, line 375: `impl<'a> From<&Value> for JsonbValue<'a> {`
`impl<'a> From<Value> for JsonbValue<'a> {`

These two `From` methods look extremely similar. Can we merge them into one?

src/pipeline/src/etl.rs
```diff
@@ -210,6 +210,37 @@ where
         self.transformer.transform_mut(val)
     }
 
+    pub fn prepare_pipeline_value(&self, val: Value, result: &mut [Value]) -> Result<()> {
```
shuiyisong, line 213: Is the original `pub fn prepare(&self, val: serde_json::Value, result: &mut [Value]) -> Result<()> {`  deprecated?

src/servers/src/otlp/logs.rs
shuiyisong: I would prefer using a different name other than `identity`, maybe something like `pre_defined` or `otlp_logs`, so the semantics of `identity` actually remains same within the project.

src/servers/src/otlp/logs.rs
```diff
@@ -0,0 +1,516 @@
...
+            ColumnDataType::String,
+            SemanticType::Field,
+            None,
+            Some(ColumnOptions {
+                options: StdHashMap::from([(
+                    "fulltext".to_string(),
+                    r#"{"enable":true}"#.to_string(),
+                )]),
+            }),
+        ),
```
shuiyisong, line 296: If I recall correctly there is a performance cost for opening full-text index. Should we add a config or parameter for controlling this behavior?

src/servers/src/http.rs
```diff
@@ -239,14 +240,21 @@ impl HttpRecordsOutput {
...
 
             let mut finished_row_cursor = 0;
             for recordbatch in recordbatches {
-                for col in recordbatch.columns() {
+                for (col_idx, col) in recordbatch.columns().iter().enumerate() {
+                    // safety here: schemas length is equal to the number of columns in the recordbatch
+                    let schema = &schemas[col_idx];
                     for row_idx in 0..recordbatch.num_rows() {
-                        let value = Value::try_from(col.get_ref(row_idx)).context(ToJsonSnafu)?;
+                        let column_pair = ColumnPair {
```
shuiyisong, line 253: The whole purpose of `ColumnPair` seems only to help convert JSON data. Maybe we can just use a helper function instead of introducing `ColumnPair` and use `From`.

src/servers/src/http/otlp.rs
```diff
@@ -34,13 +43,18 @@ use super::header::{write_cost_header_map, CONTENT_TYPE_PROTOBUF};
 use crate::error::{self, Result};
 use crate::query_handler::OpenTelemetryProtocolHandlerRef;
 
+const OTLP_GREPTIME_LOG_PIPELINE_NAME_HEADER_NAME: &str = "x-greptime-log-pipeline-name";
+const OTLP_GREPTIME_LOG_PIPELINE_VERSION_HEADER_NAME: &str = "x-greptime-log-pipeline-version";
+const OTLP_GREPTIME_LOG_TABLE_NAME_HEADER_NAME: &str = "x-greptime-log-table-name";
```
shuiyisong, line 48: Suggesting move these to `src/servers/src/http/header.rs` and removing the `OTLP` prefix. We may reuse them in other scenarios like Loki log write ingestion.

shuiyisong (approved):
Rest LGTM

src/servers/src/otlp/logs.rs
```diff
@@ -0,0 +1,516 @@
...
+
+/// Normalize otlp instrumentation, metric and attribute names
+///
+/// <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-name-syntax>
+/// - since the name are case-insensitive, we transform them to lowercase for
+/// - better sql usability
+/// - replace `.` and `-` with `_`
+fn normalize_otlp_name(name: &str) -> String {
+    name.to_lowercase().replace(|c| c == '.' || c == '-', "_")
+}
```
shuiyisong, line 42: If we normalize `.` and `-` to `_`, can we transform them back to the original format when needed?

src/frontend/src/instance/otlp.rs
```diff
@@ -92,4 +94,31 @@ impl OpenTelemetryProtocolHandler for Instance {
...
+            .check_permission(ctx.current_user(), PermissionReq::Otlp)
+            .context(AuthSnafu)?;
+
+        let interceptor_ref = self
+            .plugins
+            .get::<OpenTelemetryProtocolInterceptorRef<servers::error::Error>>();
+        interceptor_ref.pre_execute(ctx.clone())?;
+        let (requests, rows) = otlp::logs::to_grpc_insert_requests(request, pipeline, table_name)?;
+
+        self.handle_row_inserts(requests, ctx)
```
shuiyisong, line 118: Should use `handle_log_inserts()` instead of `handle_row_inserts()`

src/pipeline/src/etl.rs
paomian: `prepare` is transform serde_json::Value
this method is transform pipeline Value

src/servers/src/otlp/logs.rs
paomian: We can turn it off by default and change it via alter option.

src/servers/src/otlp/logs.rs
paomian: This is the same as metric and trace. I think it can be removed with json support, but metrics and traces should be kept compatible. I'm not going to change the metrics and traces yet.

src/pipeline/src/etl/value.rs
paomian: There's no good way to avoid it here.

src/servers/src/otlp/logs.rs
paomian: I made a mistake, the json of the trace data doesn't do this conversion strategy. We don't need to add this conversion.

PR #4763, opened 2024-09-24
refactor: Change the error type in the pipeline crate from String to Error

paomian:

* Change the error type in the pipeline crate from String to Error

## Checklist

### DIFF
--- a/src/pipeline/src/etl/processor.rs
+++ b/src/pipeline/src/etl/processor.rs
@@ -39,7 +39,12 @@ use regex::{RegexProcessor, RegexProcessorBuilder};
 use timestamp::{TimestampProcessor, TimestampProcessorBuilder};
 use urlencoding::{UrlEncodingProcessor, UrlEncodingProcessorBuilder};
 
+use super::error::{
+    FaildParseFieldFromStringSnafu, FieldMustBeTypeSnafu, ProcessorKeyMustBeStringSnafu,
+    ProcessorMustBeMapSnafu, ProcessorMustHaveStringKeySnafu, UnsupportedProcessorSnafu,
+};
 use super::field::{Field, Fields};
+use crate::etl::error::{Error, Result};
 use crate::etl::value::Value;
 
 const FIELD_NAME: &str = "field";
@@ -70,7 +75,7 @@ pub trait Processor: std::fmt::Debug + Send + Sync + 'static {
     fn ignore_missing(&self) -> bool;
 
     /// Execute the processor on a vector which be preprocessed by the pipeline
-    fn exec_mut(&self, val: &mut Vec<Value>) -> Result<(), String>;
+    fn exec_mut(&self, val: &mut Vec<Value>) -> Result<()>;
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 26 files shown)


shuiyisong (approved):
All errors that come from the pipeline crate are `map_err`ed into `PipelineTransformSnafu`. We can change that too to wrap the original error.

Rest LGTM

src/pipeline/src/etl/processor.rs
```diff
@@ -264,58 +267,86 @@ fn parse_processor(doc: &yaml_rust::Yaml) -> Result<ProcessorBuilders, String> {
...
 
     Ok(processor)
 }
 
-pub(crate) fn yaml_string(v: &yaml_rust::Yaml, field: &str) -> Result<String, String> {
-    v.as_str()
-        .map(|s| s.to_string())
-        .ok_or(format!("'{field}' must be a string"))
+pub(crate) fn yaml_string(v: &yaml_rust::Yaml, field: &str) -> Result<String> {
+    v.as_str().map(|s| s.to_string()).ok_or(
```
sunng87, line 277: not sure why clippy is not jumping in here. But I think we should use `ok_or_else` for lazy-evaluation

src/pipeline/src/etl/processor.rs
```diff
@@ -264,58 +267,86 @@ fn parse_processor(doc: &yaml_rust::Yaml) -> Result<ProcessorBuilders, String> {
...
 where
     T: std::str::FromStr,
-    T::Err: ToString,
+    T::Err: std::error::Error + Send + Sync + 'static,
 {
     yaml_strings(v, field).and_then(|v| {
         v.into_iter()
-            .map(|s| s.parse::<T>().map_err(|e| e.to_string()))
+            .map(|s| {
+                s.parse::<T>().map_err(|e| {
```
sunng87, line 334: I think there is a `context` method for this type of usage

```
s.parse::<T>().context(FailedParseFieldFromStringSnafu{ field: field.to_string() })
```

src/pipeline/src/etl/processor/cmcd.rs
```diff
@@ -201,28 +206,51 @@ impl Default for CmcdOutputInfo {
...
 }
 
 /// function to resolve CMCD_KEY_BR | CMCD_KEY_BL | CMCD_KEY_D | CMCD_KEY_DL | CMCD_KEY_MTP | CMCD_KEY_RTP | CMCD_KEY_TB
-fn br_tb(s: &str, k: &str, v: Option<&str>) -> Result<Value, String> {
-    let v = v.ok_or(format!("{k} missing value in {s}"))?;
-    let val: i64 = v
-        .parse()
-        .map_err(|_| format!("failed to parse {v} as i64"))?;
+fn br_tb(s: &str, k: &str, v: Option<&str>) -> Result<Value> {
+    let v = v.ok_or(
```
sunng87, line 215: after import `snafu::OptionExt`, it also possible to do 

v.context(CmcdMissingValueSnafu {... })

src/pipeline/src/etl/processor.rs
paomian: We can't do that, because some errors from other third-party libraries can't be used as a source of Errors because of `stack_trace_debug`。

sunng87 (approved):

PR #4750, opened 2024-09-20
fix: Release CI & make rustls use `ring`

discord9:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

use new dev-build image for release CI

Major changes are:
- use new `dev-build` image built from newest Dockerfile
- which install gcc-10 to avoid a compiler bug in gcc-9 warning by `aws-lc-sys`
- update `rustls` to `0.23`
- use forked repo of `tokio-rustls`, `hyper-rustls` and `rustls` to set `ring` as default libcrypto and preventing building `aws-lc-sys` which have numerous building problem
- and install `cargo-binstall` from binary to speed up building `dev-build` image(from 40 minutes to 20 mintues)

## Checklist

- [x] wait for [Release CI](https://github.com/GreptimeTeam/greptimedb/actions/runs/10989910424) to pass on this branch
- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.

PR #4767, opened 2024-09-25
chore: make sure aws-lc-sys wouldn't be built

discord9:

https://github.com/GreptimeTeam/greptimedb/pull/4750

Make sure `aws-lc-sys` wouldn't accidentially be built.

After update rustls to `0.23`, it switch it's crypto lib to `aws-lc`, however, it's wrapper crate `aws-lc-sys` have many building problems on different platform due to chore with gcc version and etc, and currently it will fail building on our CentOS image, and possibly on andorid, so add a check to avoid bring in `aws-lc-sys` depend before they solve most of the building problem. Have already test it locally by removing patches for `rustls`, `hyper-rustls` and `tokio-ruslts` and this check will detect `aws-lc-sys` dependency 

## Alternative
Switch back to `rustls` 0.22, however that might not be feasible, since some of our dependenices might have move to higher version of `rustls` without knowing this change of default on crypto library and it's building problems. 

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced GitHub Actions workflow to check for successful builds of the `aws-lc-sys` crate.
	- Added informative output regarding potential compilation issues with `aws-lc-sys` on older GCC versions, suggesting alternatives.

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -269,6 +269,9 @@ jobs:
       - name: Install cargo-gc-bin
         shell: bash
         run: cargo install cargo-gc-bin
+      - name: Check aws-lc-sys will not build
+        shell: bash
+        run: cargo tree -i aws-lc-sys -e features | grep aws-lc-sys && echo "Found aws-lc-sys, due to it having problem compiling on older gcc and etc. for now, please replace it with ring for now until it improve it's building experience" && exit 1
       - name: Build greptime bianry
         shell: bash
         # `cargo gc` will invoke `cargo build` with specified args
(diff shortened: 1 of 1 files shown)


zyy17 (approved):

discord9:
@coderabbitai review

.github/workflows/develop.yml
```diff
@@ -269,6 +269,13 @@ jobs:
       - name: Install cargo-gc-bin
         shell: bash
         run: cargo install cargo-gc-bin
+      - name: Check aws-lc-sys will not build
+        shell: bash
+        run: |
+             if cargo tree -i aws-lc-sys -e features | grep -q aws-lc-sys; then
```
waynexia, line 275: Is cargo tree available here?

discord9: > Is cargo tree available here?

cargo tree is now a cargo command, so if cargo is installed it should be supported

waynexia (approved):

PR #4775, opened 2024-09-26
feat: add a new status code for "external" errors

MichaelScofield:

This "external" status code can represent errors originated from system that is outside of greptimedb, useful for authentication.

## Checklist

### DIFF
--- a/src/auth/src/error.rs
+++ b/src/auth/src/error.rs
@@ -89,7 +89,7 @@ impl ErrorExt for Error {
             Error::FileWatch { .. } => StatusCode::InvalidArguments,
             Error::InternalState { .. } => StatusCode::Unexpected,
             Error::Io { .. } => StatusCode::StorageUnavailable,
-            Error::AuthBackend { .. } => StatusCode::Internal,
+            Error::AuthBackend { .. } => StatusCode::External,
 
             Error::UserNotFound { .. } => StatusCode::UserNotFound,
             Error::UnsupportedPasswordType { .. } => StatusCode::UnsupportedPasswordType,
--- a/src/auth/src/error.rs
+++ b/src/auth/src/error.rs
@@ -89,7 +89,7 @@ impl ErrorExt for Error {
             Error::FileWatch { .. } => StatusCode::InvalidArguments,
             Error::InternalState { .. } => StatusCode::Unexpected,
             Error::Io { .. } => StatusCode::StorageUnavailable,
-            Error::AuthBackend { .. } => StatusCode::External,
+            Error::AuthBackend { source, .. } => source.status_code(),
 
             Error::UserNotFound { .. } => StatusCode::UserNotFound,
             Error::UnsupportedPasswordType { .. } => StatusCode::UnsupportedPasswordType,
--- a/src/auth/src/lib.rs
+++ b/src/auth/src/lib.rs
@@ -15,7 +15,7 @@
 mod common;
 pub mod error;
 mod permission;
-mod user_info;
+pub mod user_info;
 mod user_provider;
 
 #[cfg(feature = "testing")]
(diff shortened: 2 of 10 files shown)


sunng87 (commented):
I can understand the rational for this. But from the end user perspective, this type of error is still *Internal* ones for a database. 

Imagine in our cloud service, the authentication backend is built and maintained by ourselves, this is not essential to expose this type of errors to end user.

Of course we need to log this type of errors for database admin to track

MichaelScofield:
@sunng87 I think the demand for the status code for "external" errors do exist, especially when greptimedb will have connected to external systems. Two other databases both have this kind of error codes, [postgresql](https://www.postgresql.org/docs/current/errcodes-appendix.html) and [clickhouse](https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/ErrorCodes.cpp).

The status codes are involved in errors, not the other way around. When we implementing features, we can freely choose to return what kind of error (and the status code) to the user. If, like in our cloud service, we need to make the external system's error internal to greptimedb, we can simply change the status code of it.

src/auth/src/error.rs
```diff
@@ -89,7 +89,7 @@ impl ErrorExt for Error {
             Error::FileWatch { .. } => StatusCode::InvalidArguments,
             Error::InternalState { .. } => StatusCode::Unexpected,
             Error::Io { .. } => StatusCode::StorageUnavailable,
-            Error::AuthBackend { .. } => StatusCode::Internal,
+            Error::AuthBackend { .. } => StatusCode::External,
```
shuiyisong, line 92:
```suggestion
            Error::AuthBackend { source, .. } => source.status_code(),
```

Since `AuthBackend` is a `BoxedError`, we can use the actual error's source code instead.

shuiyisong, line 92:
```suggestion
            Error::AuthBackend { source, .. } => source.status_code(),
```

Since `AuthBackend` is a `BoxedError`, we can use the actual error's source code instead.

If so, we'll be able to set the error status code for each implementation use case. For example, remaining `Internal` for our cloud service. cc @sunng87 @MichaelScofield

src/auth/src/lib.rs
```diff
@@ -15,7 +15,7 @@
 mod common;
 pub mod error;
 mod permission;
-mod user_info;
+pub mod user_info;
```
shuiyisong, line 18: I think we should be very careful to open this crate up. We should only be using the public `UserInfo` trait and not explicitly use any implementation struct in codebase.

MichaelScofield:
@shuiyisong PTAL

sunng87 (approved):

sunng87:
@MichaelScofield Makes sense. Thank you for explanation.

src/servers/src/mysql/handler.rs
```diff
@@ -253,7 +277,16 @@ impl<W: AsyncWrite + Send + Sync + Unpin> AsyncMysqlShim<W> for MysqlInstanceShi
             let user_id = Identity::UserId(&username, addr.as_deref());
 
             let password = match auth_plugin {
-                "mysql_native_password" => Password::MysqlNativePassword(auth_data, salt),
+                MYSQL_NATIVE_PASSWORD => Password::MysqlNativePassword(auth_data, salt),
+                MYSQL_CLEAR_PASSWORD => {
+                    if auth_data.is_empty() {
+                        return false;
+                    }
+                    // The raw bytes received are represented in C-like string, ended in '\0'.
```
killme2008, line 285: I think we should add an assertion to ensure that it's ended with `\0`. To make sure the behavior has not changed.

killme2008 (approved):
Almost LGTM

shuiyisong (approved):

Issue #3685, opened 2024-04-10
Limit CPU and memory comsumption in resource-limited environment


v0y4g3r:
### What problem does the new feature solve?

GreptimeDB is designed to scale from even embedded devices to mega scale cloud services. But when it runs on resource-limited devices, like industrial controller based on.Android and Windows, it does not have a framework to limit the resource consumption, namely CPU and memory usage.

### What does the feature do?

This issue calls for a resource-limit framework, just like cgroup in Linux kernel, to limit the CPU and memory usage for those dedicated spawned tasks, like flush, compaction, etc.

### Implementation challenges

Tokio does not provide instrumentation tools to probe the CPU and memory usage of submitted tasks, we can only wrap the tasks with our own metrics and using rate limiting strategies to limit inflight tasks.

sunng87:
This reminds me a discussion with @zyy17 about an general abstraction of task spawning:

- an implementation independent spawn function as entrypoint
- option to spawn to current thread, to thread pool, to local subprocess or remote process (managed by various resource manager, like k8s)
- optional option to sepecify compute resources for the task

v0y4g3r:
@ActivePeter we can start with introducing a new wrapper runtime that can set different priorities for different tasks even if it's not yet referenced in the code base. Feel free to draft a PR once you're ready.

PR #4782, opened 2024-09-28
feat: Limit CPU in runtime (#3685)

ActivePeter:

https://github.com/GreptimeTeam/greptimedb/issues/3685

- Add parking_lot in root Cargo.toml, use `parking_lot.workspace = true` in sub project.
- Expand the runtime with optional resource limitation level.

## Checklist


v0y4g3r:
@fengys1996 Please also take a look at this PR, aiming to provide a new relaxed runtime to constrain CPU usage of spawned tasks.

src/common/runtime/src/ratelimit.rs
```diff
@@ -0,0 +1,608 @@
...
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+struct Parameters {
+    capacity: u64,
+    refill_amount: u64,
+    refill_interval: Duration,
+}
+
+pub struct Ratelimiter {
```
v0y4g3r, line 96: We may just use the ratelimit crate instead of writing our own:https://docs.rs/ratelimit/0.7.1/ratelimit/

fengys1996:
It looks like there are some conflicts and some compilation issues.
Please resolve them first.

src/common/runtime/src/future_throttle_count_mode.rs
```diff
@@ -0,0 +1,112 @@
...
+    fn unwrap_backoff(&mut self) -> &mut Pin<Box<Sleep>> {
+        match self {
+            State::Backoff(sleep) => sleep,
+            _ => panic!("unwrap_backoff failed"),
+        }
+    }
+}
+
+#[pin_project::pin_project]
+pub struct ThrottleFuture<F: Future + Send + 'static> {
```
sunng87, line 27: consider renaming to ThrottleableFuture

src/common/runtime/src/future_throttle_count_mode.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use futures::FutureExt;
+use tokio::time::Sleep;
+
+use crate::runtime_throttle_count_mode::RuntimeThrottleShareWithFuture;
+
+enum State {
+    Common,
+    Backoff(Pin<Box<Sleep>>),
+}
```
sunng87, line 15: Consider better naming:

- Common -> Pollable,
- Backoff -> Throttled

src/common/runtime/src/future_throttle_count_mode.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        match this.state {
+            State::Common => {}
+            State::Backoff(ref mut sleep) => match sleep.poll_unpin(cx) {
+                Poll::Ready(_) => {
+                    *this.state = State::Common;
+                }
+                Poll::Pending => return Poll::Pending,
+            },
+        };
```
sunng87, line 75: if let State::Backoff(...) = this.state {
...
}

src/common/runtime/src/future_throttle_count_mode.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        let poll_res = this.future.poll(cx);
+
+        match poll_res {
+            Poll::Ready(r) => Poll::Ready(r),
+            Poll::Pending => {
+                #[cfg(test)]
+                {
+                    *this.pend_cnt += 1;
+                }
```
sunng87, line 106: we can add twice for a single task in a single poll, is it expected?

src/common/runtime/src/runtime.rs
```diff
@@ -50,45 +46,42 @@ impl Drop for Dropper {
     }
 }
 
-impl Runtime {
-    pub fn builder() -> Builder {
+pub trait RuntimeTrait {
+    /// Get a runtime builder
+    fn builder() -> Builder {
```
sunng87, line 51: returning a fixed Builder may not make sense. In most case it should be

pub trait Runtime {
  type Builder;

  fn builder() -> Self::Builder;

}

Also consider removing Trait from trait name

src/common/runtime/src/runtime_throttle_count_mode.rs
```diff
@@ -0,0 +1,106 @@
...
+            .finish()
+    }
+}
+
+/// A runtime to run future tasks
+#[derive(Clone, Debug)]
+pub struct RuntimeThrottleCount {
+    name: String,
+    handle: Handle,
+    share_with_future: Arc<RuntimeThrottleShareWithFuture>,
```
sunng87, line 52: This can be more accrue as `shared_with_futures`.

Or it can be as simple as `throttler`

src/common/runtime/src/runtime_throttle_count_mode.rs
```diff
@@ -0,0 +1,106 @@
...
+                "ratelimiter_refill_amount",
+                &self.ratelimiter.as_ref().map(|v| v.refill_amount()),
+            )
+            .finish()
+    }
+}
+
+/// A runtime to run future tasks
+#[derive(Clone, Debug)]
+pub struct RuntimeThrottleCount {
```
sunng87, line 49: Consider renaming to `ThrottleableRuntime`

src/common/runtime/src/runtime.rs
ActivePeter: The builder is reused, only the fn build() need to be override

ActivePeter: Runtime is used by the type Runtime

src/common/runtime/src/runtime_throttle_count_mode.rs
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    fn name(&self) -> &str {
+        &self.name
+    }
+}
+
+enum State {
+    Pollable,
+    Throttled(Pin<Box<Sleep>>),
```
evenyag, line 113: Can we avoid this Box?

src/common/runtime/src/runtime_throttle_count_mode.rs
```diff
@@ -0,0 +1,405 @@
...
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub enum Priority {
+    VeryLow = 0,
+    Low = 1,
+    Middle = 2,
+    High = 3,
+    VeryHigh = 4,
+}
+impl Priority {
+    pub fn ratelimiter_count(&self) -> Option<Ratelimiter> {
```
evenyag, line 221: Do we need to make it pub?

src/common/runtime/src/runtime_throttle_count_mode.rs
```diff
@@ -0,0 +1,405 @@
...
+            Priority::Low => Some(4000),
+            Priority::Middle => Some(6000),
+            Priority::High => Some(8000),
+            Priority::VeryHigh => None,
+        };
+        if let Some(gen_per_10ms) = gen_per_10ms {
+            Ratelimiter::builder(gen_per_10ms, Duration::from_millis(10)) // generate poll count per 10ms
+                .max_tokens(max) // reserved token for batch request
+                .build()
+                .unwrap()
```
evenyag, line 234: When will `build()` return an error? Can we return an error instead?

src/common/runtime/src/runtime_throttle_count_mode.rs
```diff
@@ -0,0 +1,405 @@
...
+        let gen_per_10ms = match self {
+            Priority::VeryLow => Some(2000),
+            Priority::Low => Some(4000),
+            Priority::Middle => Some(6000),
+            Priority::High => Some(8000),
+            Priority::VeryHigh => None,
+        };
+        if let Some(gen_per_10ms) = gen_per_10ms {
+            Ratelimiter::builder(gen_per_10ms, Duration::from_millis(10)) // generate poll count per 10ms
+                .max_tokens(max) // reserved token for batch request
```
evenyag, line 232: How do we get these magic numbers? Is it possible to make them configurable?

src/common/runtime/src/runtime.rs
```diff
@@ -98,11 +91,17 @@ impl Default for Builder {
             runtime_name: format!("runtime-{}", RUNTIME_ID.fetch_add(1, Ordering::Relaxed)),
             thread_name: "default-worker".to_string(),
             builder: RuntimeBuilder::new_multi_thread(),
+            priority: Priority::VeryHigh,
         }
     }
 }
 
 impl Builder {
+    pub fn priority(&mut self, priority: Priority) -> &mut Self {
```
evenyag, line 100: Looks like the priority is runtime level, not task level. What's the effect of setting different priorities? We may need some documents.

src/common/runtime/src/runtime_throttle_count_mode.rs
```diff
@@ -0,0 +1,405 @@
...
+                .unwrap()
+                .into()
+        } else {
+            None
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
evenyag, line 243: The unit test can focus on the correctness of the ThrottleFuture (covers the branches returning pending or ready).

evenyag (commented):
Do we have a document to describe how this runtime works and how to find a proper priority for the runtime? I'd suggest providing more docs in the runtime crate.

src/common/runtime/src/runtime_throttle_count_mode.rs
v0y4g3r: > Can we avoid this Box?

Look like `tokio::time::Sleep` does not implement `Unpin`.

src/common/runtime/src/lib.rs
```diff
@@ -17,6 +17,8 @@ pub mod global;
 mod metrics;
 mod repeated_task;
 pub mod runtime;
+mod runtime_default;
+mod runtime_throttle_count_mode;
```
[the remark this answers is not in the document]
ActivePeter: For this throttleable capability, the default spawn runtime has the highest priority, meaning there will be no scheduling restrictions on tasks; thus, it will not affect the original capabilities. Additionally, implementing this as a feature will lead to version divergence. For a large project like Greptime, switching features may cause the logic of basic library calls to diverge, which means conditional judgments for features would need to be introduced in the core code. Personally, I think this is unnecessary.
```suggestion
mod runtime_default;
mod runtime_throttleable;
```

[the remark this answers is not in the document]
ActivePeter: For this throttleable capability, the default spawn runtime has the highest priority, meaning there will be no scheduling restrictions on tasks; thus, it will not affect the original capabilities. Additionally, implementing this as a feature will lead to version divergence. For a large project like Greptime, switching features may cause the logic of basic library calls to diverge, which means conditional judgments for features would need to be introduced in the core code. Personally, I think this is unnecessary.

src/common/runtime/src/bin.rs
```diff
@@ -0,0 +1,214 @@
...
+                    .await
+                    .unwrap(),
+                );
+            }
+        }
+    }
+
+    pub fn test_diff_workload_priority() {
+        common_telemetry::init_default_ut_logging();
+        // let _ = std::fs::create_dir_all(TEST_GENERATE_DIR);
```
evenyag, line 143: Remove unused codes.

src/common/runtime/README
```diff
@@ -0,0 +1,5 @@
+# Run performance test for different priority & workload type
+
+```
```
evenyag, line 3: The readme is outdated.

src/common/runtime/src/bin.rs
```diff
@@ -0,0 +1,214 @@
...
+        debug!(
+            "testing cpu usage for priority {:?} workload_id {}",
+            priority, workload_id,
+        );
+        // start monitor thread
+        let mut tasks = vec![];
+        let start = std::time::Instant::now();
+        for i in 0..500 {
+            // persist cpu usage in json: {priority}.{workload_id}
+            match workload_id {
```
evenyag, line 183: A str workload id is more readable. e.g. compute_heavily/blocking_write

src/common/runtime/src/bin.rs
```diff
@@ -0,0 +1,214 @@
...
+pub struct Command {
+    #[arg(long, action = clap::ArgAction::Count)]
+    test_diff_priority_cpu: u8,
+}
+
+fn main() {
+    let cmd = Command::parse();
+
+    // make sure one and only one of them is set
+    let cmds_state = [(cmd.test_diff_priority_cpu > 0, "test-diff-priority-cpu")];
```
evenyag, line 27: Why do we need this cmd argument? An argument to control the times of the loop to run the test in `test_spec_priority_and_workload()` is more helpful than this argument. For example, a `num_loops` argument.

src/common/runtime/src/runtime_throttleable.rs
```diff
@@ -0,0 +1,305 @@
...
+pub struct ThrottleableRuntime {
+    name: String,
+    handle: Handle,
+    shared_with_future: Arc<RuntimeThrottleSharedWithFuture>,
+    // Used to receive a drop signal when dropper is dropped, inspired by databend
+    _dropper: Arc<Dropper>,
+}
+
+impl ThrottleableRuntime {
+    #[allow(dead_code)]
```
evenyag, line 63: We can pub this mod in `lib.rs` than we don't need this allow atrribute.

src/common/runtime/src/bin.rs
ActivePeter: Maybe we will add multi-command selection in the future, is there better solution?

src/common/runtime/README
```diff
@@ -0,0 +1,60 @@
...
+    }
+}
+```
+
+This is the preliminary experimental effect so far:
+
+![](resources/rdme-exp.png)
+
+## TODO
+- Introduce PID to achieve more accurate limitation.
```
evenyag, line 60:
```suggestion
- Introduce PID to achieve more accurate limitation.

```

src/common/runtime/src/bin.rs
```diff
@@ -0,0 +1,213 @@
...
+            tokio::task::yield_now().await;
+            std::thread::yield_now();
+        }
+    }
+    async fn workload_compute_heavily2() {
+        let prefix = 30;
+        for _ in 0..2000 {
+            let _ = compute_pi_str(prefix);
+            tokio::task::yield_now().await;
+            std::thread::yield_now();
```
evenyag, line 92: Is `std::thread::yield_now()` necessary? What if we only use `tokio::task::yield_now()`.

src/common/runtime/src/runtime_throttleable.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    state: State,
+}
+
+impl<F> ThrottleFuture<F>
+where
+    F: Future + Send + 'static,
+    F::Output: Send + 'static,
+{
+    pub(crate) fn new(handle: Arc<RuntimeThrottleSharedWithFuture>, future: F) -> Self {
```
evenyag, line 151:
```suggestion
    fn new(handle: Arc<RuntimeThrottleSharedWithFuture>, future: F) -> Self {
```

src/common/runtime/src/runtime_throttleable.rs
```diff
@@ -0,0 +1,304 @@
...
+pub use tokio::task::JoinHandle;
+use tokio::time::Sleep;
+
+use crate::error::{BuildRuntimeRateLimiterSnafu, Result};
+use crate::runtime::{Dropper, Priority, RuntimeTrait};
+use crate::Builder;
+
+pub struct RuntimeThrottleSharedWithFuture {
+    pub ratelimiter: Option<Ratelimiter>,
+}
```
evenyag, line 35: Maybe `RuntimeRateLimiter` is clearer than this name.
```suggestion
struct RuntimeRateLimiter {
    ratelimiter: Option<Ratelimiter>,
}
```

src/common/runtime/src/runtime_throttleable.rs
```diff
@@ -0,0 +1,304 @@
...
+                #[cfg(test)]
+                {
+                    *this.pend_cnt += 1;
+                }
+                *this.state = State::Throttled(Box::pin(tokio::time::sleep(wait)));
+                match this.state.unwrap_backoff().poll_unpin(cx) {
+                    Poll::Ready(_) => {
+                        *this.state = State::Pollable;
+                        cx.waker().wake_by_ref();
+                        return Poll::Pending;
```
evenyag, line 193: What if we don't return pending here but poll `this.future`? Since we already sleep enough time, it should be fine to poll `this.future`.
```suggestion
                        *this.state = State::Pollable;
```

src/common/runtime/resources/rdme-exp.png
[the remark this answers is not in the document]
ActivePeter: > Did you test this under release mode? BTW, better to display the workload type by name instead of the id.

I’ll try it

src/common/runtime/src/runtime_throttleable.rs
ActivePeter: Actully this shouldn't happen because we just started sleep, but I think change to `*this.state = State::Pollable;` is ok too.

src/common/runtime/README
```diff
@@ -0,0 +1,60 @@
...
```
evenyag, line 60:
```suggestion
- Introduce PID to achieve more accurate limitation.

```

src/common/runtime/README
evenyag, line 1: README.md

src/common/runtime/src/runtime_throttleable.rs
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+}
+
+#[pin_project::pin_project]
+pub struct ThrottleFuture<F: Future + Send + 'static> {
+    #[pin]
+    future: F,
+
+    /// priority of this future
```
evenyag, line 136:
```suggestion
    /// Rate limiter of this future
```

src/common/runtime/src/bin.rs
```diff
@@ -0,0 +1,206 @@
...
+            let _ = compute_pi_str(prefix);
+            tokio::task::yield_now().await;
+        }
+    }
+    async fn workload_compute_heavily2() {
+        let prefix = 30;
+        for _ in 0..2000 {
+            let _ = compute_pi_str(prefix);
+            tokio::task::yield_now().await;
+            std::thread::yield_now();
```
evenyag, line 85:
```suggestion
            tokio::task::yield_now().await;
```

evenyag (approved):
LGTM. There are some remaining enhancements we can acheive in other PRs. e.g. Making tokens of different priorities configurable.

v0y4g3r (approved):
LGTM

v0y4g3r:
@fengys1996 I think we can test this feature on the new version of GreptimeDB edge now to see if it can manage the total CPU usage.

killme2008:
Great work! Thanks a lot.

PR #4787, opened 2024-09-30
feat: set max log files to 720 by default, info log only

### LINKED ISSUE #4777: Allow configuring the max log files to keep

User experience

Currently, we don't limit the total number of log files. A long-running database may generate too many log files. We need to limit the max log files to keep by default.

### Implementation challenges

The tracing-appender already supports setting the max log files. We can expose this in our `LoggingOptions`.
https://docs.rs/tracing-appender/latest/tracing_appender/rolling/struct.Builder.html#method.max_log_files

Kev1n8:

Closes #4777.

Exposed `max_log_files` in `LoggingOptions`, set it to 720 by default, and build `info_log` with this restriction.

## Checklist

Not sure how to test this. It passed CI tho.

### DIFF
--- a/src/common/telemetry/src/logging.rs
+++ b/src/common/telemetry/src/logging.rs
@@ -68,6 +68,9 @@ pub struct LoggingOptions {
 
     /// The logging options of slow query.
     pub slow_query: SlowQueryOptions,
+
+    /// The maximum number of log files set by default.
+    pub max_log_files: usize,
 }
 
 /// The options of slow query.
@@ -116,6 +119,8 @@ impl Default for LoggingOptions {
             tracing_sample_ratio: None,
             append_stdout: true,
             slow_query: SlowQueryOptions::default(),
+            // Rotation hourly, 24 files per day, keeps info log files of 30 days
+            max_log_files: 720,
         }
     }
 }
@@ -206,8 +211,12 @@ pub fn init_global_logging(
 
         // Configure the file logging layer with rolling policy.
         let file_logging_layer = if !opts.dir.is_empty() {
-            let rolling_appender =
-                RollingFileAppender::new(Rotation::HOURLY, &opts.dir, "greptimedb");
+            let rolling_appender = RollingFileAppender::builder()
+                .rotation(Rotation::HOURLY)
+                .filename_prefix("greptimedb")
+                .max_log_files(opts.max_log_files)
+                .build(&opts.dir)
+                .expect("initializing rolling file appender failed");
             let (writer, guard) = tracing_appender::non_blocking(rolling_appender);
             guards.push(guard);
 
--- a/src/common/telemetry/src/logging.rs
+++ b/src/common/telemetry/src/logging.rs
@@ -216,7 +216,10 @@ pub fn init_global_logging(
                 .filename_prefix("greptimedb")
                 .max_log_files(opts.max_log_files)
                 .build(&opts.dir)
-                .expect("initializing rolling file appender failed");
+                .expect(&format!(
+                    "initializing rolling file appender at {} failed",
+                    &opts.dir
+                ));
             let (writer, guard) = tracing_appender::non_blocking(rolling_appender);
             guards.push(guard);
 
@@ -237,8 +240,16 @@ pub fn init_global_logging(
 
         // Configure the error file logging layer with rolling policy.
         let err_file_logging_layer = if !opts.dir.is_empty() {
-            let rolling_appender =
-                RollingFileAppender::new(Rotation::HOURLY, &opts.dir, "greptimedb-err");
+            let rolling_appender = RollingFileAppender::builder()
+                .rotation(Rotation::HOURLY)
+                .filename_prefix("greptimedb-err")
+                // Simply set maximum to constant 30 here
+                .max_log_files(30)
+                .build(&opts.dir)
+                .expect(&format!(
+                    "initializing rolling file appender at {} failed",
+                    &opts.dir
+                ));
             let (writer, guard) = tracing_appender::non_blocking(rolling_appender);
             guards.push(guard);
 
@@ -265,8 +276,15 @@ pub fn init_global_logging(
         };
 
         let slow_query_logging_layer = if !opts.dir.is_empty() && opts.slow_query.enable {
-            let rolling_appender =
-                RollingFileAppender::new(Rotation::HOURLY, &opts.dir, "greptimedb-slow-queries");
+            let rolling_appender = RollingFileAppender::builder()
+                .rotation(Rotation::HOURLY)
+                .filename_prefix("greptimedb-slow-queries")
+                .max_log_files(opts.max_log_files)
+                .build(&opts.dir)
+                .expect(&format!(
+                    "initializing rolling file appender at {} failed",
+                    &opts.dir
+                ));
             let (writer, guard) = tracing_appender::non_blocking(rolling_appender);
             guards.push(guard);
 
--- a/src/common/telemetry/src/logging.rs
+++ b/src/common/telemetry/src/logging.rs
@@ -54,6 +54,9 @@ pub struct LoggingOptions {
     /// The log format that can be one of "json" or "text". Default is "text".
     pub log_format: LogFormat,
 
+    /// The maximum number of log files set by default.
+    pub max_log_files: usize,
+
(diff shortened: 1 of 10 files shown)


src/common/telemetry/src/logging.rs
```diff
@@ -206,8 +211,12 @@ pub fn init_global_logging(
...
         // Configure the file logging layer with rolling policy.
         let file_logging_layer = if !opts.dir.is_empty() {
-            let rolling_appender =
-                RollingFileAppender::new(Rotation::HOURLY, &opts.dir, "greptimedb");
+            let rolling_appender = RollingFileAppender::builder()
+                .rotation(Rotation::HOURLY)
+                .filename_prefix("greptimedb")
+                .max_log_files(opts.max_log_files)
+                .build(&opts.dir)
+                .expect("initializing rolling file appender failed");
```
v0y4g3r, line 219: Maybe also carry log dir in panic info?

src/common/telemetry/src/logging.rs
```diff
@@ -68,6 +68,9 @@ pub struct LoggingOptions {
 
     /// The logging options of slow query.
     pub slow_query: SlowQueryOptions,
+
+    /// The maximum number of log files set by default.
+    pub max_log_files: usize,
```
v0y4g3r, line 73: Can we expose this option in [configurtion file's log section](https://github.com/GreptimeTeam/greptimedb/blob/aa03d3b11ccff692ad348cd113da46a7e758b527/config/standalone.example.toml#L604)?

src/common/telemetry/src/logging.rs
evenyag, line 1: We may set the max file number for the error log. I think a constant value 30 is fine.
https://github.com/GreptimeTeam/greptimedb/blob/e39a9e6feb30dcd7f58742e1bc296856a9b27b04/src/common/telemetry/src/logging.rs#L231-L232

We can add max log files for the slow log:
https://github.com/GreptimeTeam/greptimedb/blob/e39a9e6feb30dcd7f58742e1bc296856a9b27b04/src/common/telemetry/src/logging.rs#L258-L261

src/common/telemetry/src/logging.rs
Kev1n8: > Better to return an error.
I simply add the dir info to it for now. Maybe we could do it in another pr .

v0y4g3r: > > Better to return an error.
> 
> I simply add the dir info to it for now. Maybe we could do it in another pr .

@evenyag I think maybe it's reasonable to panic when log fails to initialize with given config instead of silently dropping all output.

v0y4g3r:
@Kev1n8 Looks like some CI checks failed, please run those checks locally to ensure CI can pass:
```bash
// Check code style
cargo clippy --workspace --all-targets -- -D warnings
// Check if doc is up-to-date
make config-docs
// Run all unit tests
cargo test --workspace --all-targets 
```

The failure in unit test relates to your changes in configuration file.

Kev1n8:
Seems like there's still something wrong. I'll look into it. Any suggestions will be hugely appreciated too!

v0y4g3r:
Looks like there're some typos detected (though those files are not touched in this PR). Cloud you please also fix them?

killme2008:
@Kev1n8 Looks like there are some typos in checking https://github.com/GreptimeTeam/greptimedb/actions/runs/11120980442/job/30935597697?pr=4787

It's not related to this PR, but would you like to fix them in this PR too? Thanks a lot.

src/common/telemetry/src/logging.rs
```diff
@@ -228,8 +239,15 @@ pub fn init_global_logging(
 
         // Configure the error file logging layer with rolling policy.
         let err_file_logging_layer = if !opts.dir.is_empty() {
-            let rolling_appender =
-                RollingFileAppender::new(Rotation::HOURLY, &opts.dir, "greptimedb-err");
+            let rolling_appender = RollingFileAppender::builder()
+                .rotation(Rotation::HOURLY)
+                .filename_prefix("greptimedb-err")
+                // Simply set maximum to constant 30 here
+                .max_log_files(30)
```
killme2008, line 246: Why not keep it the same with other logs? I think we should keep them consistent.

killme2008 (approved):
Almost LGTM

src/common/telemetry/src/logging.rs
evenyag: There won't be too much error logs. But it's ok to use the same option.

src/common/telemetry/src/logging.rs
```diff
@@ -206,8 +211,14 @@ pub fn init_global_logging(
...
         let file_logging_layer = if !opts.dir.is_empty() {
-            let rolling_appender =
-                RollingFileAppender::new(Rotation::HOURLY, &opts.dir, "greptimedb");
+            let rolling_appender = RollingFileAppender::builder()
+                .rotation(Rotation::HOURLY)
+                .filename_prefix("greptimedb")
+                .max_log_files(opts.max_log_files)
+                .build(&opts.dir)
+                .unwrap_or_else(|_| {
+                    panic!("initializing rolling file appender at {} failed", &opts.dir)
```
evenyag, line 220: Please also print the error in the panic message.

src/common/telemetry/src/logging.rs
```diff
@@ -228,8 +239,15 @@ pub fn init_global_logging(
...
-            let rolling_appender =
-                RollingFileAppender::new(Rotation::HOURLY, &opts.dir, "greptimedb-err");
+            let rolling_appender = RollingFileAppender::builder()
+                .rotation(Rotation::HOURLY)
+                .filename_prefix("greptimedb-err")
+                // Simply set maximum to constant 30 here
+                .max_log_files(30)
+                .build(&opts.dir)
+                .unwrap_or_else(|_| {
+                    panic!("initializing rolling file appender at {} failed", &opts.dir)
```
evenyag, line 249: Same as above

src/common/telemetry/src/logging.rs
```diff
@@ -256,8 +274,14 @@ pub fn init_global_logging(
...
         let slow_query_logging_layer = if !opts.dir.is_empty() && opts.slow_query.enable {
-            let rolling_appender =
-                RollingFileAppender::new(Rotation::HOURLY, &opts.dir, "greptimedb-slow-queries");
+            let rolling_appender = RollingFileAppender::builder()
+                .rotation(Rotation::HOURLY)
+                .filename_prefix("greptimedb-slow-queries")
+                .max_log_files(opts.max_log_files)
+                .build(&opts.dir)
+                .unwrap_or_else(|_| {
+                    panic!("initializing rolling file appender at {} failed", &opts.dir)
```
evenyag, line 283: Same as above.

src/common/telemetry/src/logging.rs
evenyag, line 73: You may need to define max_log_files before slow_query to avoid toml deserialization error. I remember that we have a unit test for it so we can check the CI later.

Kev1n8:
> @Kev1n8 Looks like there are some typos in checking  https://github.com/GreptimeTeam/greptimedb/actions/runs/11120980442/job/30935597697?pr=4787
> 
> It's not related to this PR, but would you like to fix them in this PR too? Thanks a lot.

Sure.

src/common/telemetry/src/logging.rs
Kev1n8: Will do.

src/common/telemetry/src/logging.rs
Kev1n8: Got it. I am now putting it right after `log_format`.

v0y4g3r (approved):
@Kev1n8 Looks like config api test still fails.

tests-integration/tests/http.rs
```diff
@@ -835,6 +835,7 @@ retry_delay = "500ms"
 [logging]
 append_stdout = true
 enable_otlp_tracing = false
+max_log_files = 720
```
v0y4g3r, line 838:
```suggestion
max_log_files = 720
append_stdout = true
enable_otlp_tracing = false
```

Kev1n8:
> @Kev1n8 Looks like config api test still fails.

My bad. Thanks for fixing that!

PR #4790, opened 2024-09-30
fix: ts conversion during transform phase

shuiyisong:

Previously during the transform phase of the pipeline, the timestamp value coerce process ignores the targeting timestamp resolution of the definition.

## Checklist

### DIFF
--- a/src/pipeline/src/etl/transform/transformer/greptime/coerce.rs
+++ b/src/pipeline/src/etl/transform/transformer/greptime/coerce.rs
@@ -176,16 +176,23 @@ pub(crate) fn coerce_value(val: &Value, transform: &Transform) -> Result<Option<
         Value::Boolean(b) => coerce_bool_value(*b, transform),
         Value::String(s) => coerce_string_value(s, transform),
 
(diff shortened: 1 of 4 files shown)


src/pipeline/src/etl/transform/transformer/greptime/coerce.rs
```diff
@@ -176,16 +176,23 @@ pub(crate) fn coerce_value(val: &Value, transform: &Transform) -> Result<Option<
...
+                    input_timestamp.timestamp_micros(),
+                ))),
+                Timestamp::Millisecond(_) => Ok(Some(ValueData::TimestampMillisecondValue(
+                    input_timestamp.timestamp_millis(),
+                ))),
+                Timestamp::Second(_) => Ok(Some(ValueData::TimestampSecondValue(
+                    input_timestamp.timestamp(),
+                ))),
+            },
+            _ => unimplemented!("Timestamp can only be coerced to another timestamp"),
```
killme2008, line 194: I think we must return an error instead of panic, including the following `unimplemented!` invocations.

killme2008 (approved):
Almost LGTM

sunng87 (approved):

Issue #4760, opened 2024-09-23
Timeout support for MySQL and Postgres channel


sunng87:
### What problem does the new feature solve?

Add variables to MySQL and Postgres for controlling statement execution timeout.

This is also a compatibility requirement from Sigma.

### What does the feature do?

For MySQL protocol, we need to support `SET MAX_EXECUTION_TIME = 5000`: https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_max_execution_time

For Postgres protocol, we need to support `SET statement_timeout TO 5000`: https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-STATEMENT-TIMEOUT

lyang24:
This looks interesting. I wonder are timeout settings per session? I am trying to measure the scope of the work, do we have similar session variable implementation examples on the codebase?

sunng87:
@lyang24 good to hear. Yes this variable is session-scoped like our time zone setting. You can check `src/session/src/lib.rs` as the start for timezone impl, including how we set and how we use it.

lyang24:
Thanks for the code pointers. may i take this issue?

sunng87:
Sure

Issue #1712, opened 2023-06-03
Find a better way to deal with mysql's special federated queries


MichaelScofield:
### What type of enhancement is this?

Refactor

### What does the enhancement do?

Some mysql tools or protocols will emit special queries upon connection established. For example, sqlalchemy's `SHOW VARIABLES LIKE 'sql_mode'`, DBeaver's `SHOW WARNINGS` or simply `select 1`. We are currently using regex to pattern matching these special federated queries, then return static, fake values, just like what Databend does. Instead of really executing them. It seems like a quick and dirty way to resolve the problems, but we are not trying to be compatible with the whole of mysql, so I think we can live with that. So I'm hesitating to label this issue as "wont fix" ...

tisonkun:
An issue I found for the regex matching way is that it's too tight to tolerate whitespace. So perhaps a special handle after tokenizer is needed instead of matching plain text.

killme2008:
Looks like we can't find a better way. The current implementation is acceptable and working well.

PR #4792, opened 2024-10-02
feat: implement statement/execution timeout session variable

lyang24:

https://github.com/GreptimeTeam/greptimedb/issues/4760

- Summarize your change (**mandatory**):
This pr implements session variables for query timeout - the show and set dialects are supported.

The query timeouts are per session and its implemented with applying a tokio timeout to the query processing function
* need to write test for show/set new session variables.

## Checklist


shuiyisong:
Nice work 👍 

The input with time unit of PostgreSQL seems not implemented, would you like to implement it in this PR?

Also, it would be better to add time unit while outputting the variable value to the client. Here is an example in PostgreSQL:

```
postgres=# SHOW statement_timeout;
 statement_timeout
-------------------
 100ms
(1 row)
```

lyang24:
> Nice work 👍
> 
> The input with time unit of PostgreSQL seems not implemented, would you like to implement it in this PR?
> 
> Also, it would be better to add time unit while outputting the variable value to the client. Here is an example in PostgreSQL:
> 
> ```
> postgres=# SHOW statement_timeout;
>  statement_timeout
> -------------------
>  100ms
> (1 row)
> ```

Thank you for the feedbacks, I added an implementation of postgres time unit support in the newer iteration of the pr.

src/operator/src/statement.rs
```diff
@@ -120,7 +121,15 @@ impl StatementExecutor {
     ) -> Result<Output> {
         let _slow_query_timer = self.stats.start_slow_query_timer(stmt.clone());
         match stmt {
-            QueryStatement::Sql(stmt) => self.execute_sql(stmt, query_ctx).await,
+            QueryStatement::Sql(stmt) => {
+                if let Some(timeout) = query_ctx.query_timeout() {
+                    return tokio::time::timeout(timeout, self.execute_sql(stmt, query_ctx))
```
evenyag, line 126: `execute_sql` returns a stream (the output is a stream) and we may spend most time waiting on the stream.

We may need to add the timeout to the stream (But this is per-item timeout)
https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.timeout

lyang24: This is a great catch I need to think more about where to apply the timeouts. After another review of docs, the implementation is tricker than i thought - [mysql](https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_max_execution_time ) query timeout only apply to the read statements. 

I will turn this pr to draft for now and add a second commit for applying timeout soon

lyang24: I have been thinking about how to apply per statement timeout in GreptimeDB.

Hopefully my understanding is correct with the life of a statement execution.
**execute**
(https://github.com/GreptimeTeam/greptimedb/blob/2f2b4b306cb480c1396f452f2c27dc2206705d38/src/frontend/src/instance.rs#L252C1-L295C14) create streams for each statement and the results are collected into a vec.
**result yielding**
postgres - the stream [yielding](https://github.com/sunng87/pgwire/blob/368e49ed5932de78deae24f20c3e806463b0506a/src/api/query.rs#L375C1-L380C6) at upstream pgwire package.
mysql - the stream [yielding](https://github.com/GreptimeTeam/greptimedb/blob/a889ea88ca595c184c14a3c65fe64027d9ed23a6/src/servers/src/mysql/writer.rs#L164C1-L182C22) at the writer.

lyang24: I thought about 3 additional approaches for the querytimeout.
1. attach timeouts at very beginning of query execution. i.e.  (add timeout to `on_query` method from `SimpleQueryHandler` for postgres).
pros: easy to implement, its also capture the query's entire life time.
cons: its wouldn't work on per statement timeout with bulk statements.
2. Only apply timeouts while waiting for the stream's results, attach timeout duration to the stream and use tokio select while yielding results from stream.
pros: easy to implement, can achieve per statement timeout
cons: may need to patch upstream pgwire
3. Alarm approach
- start alarm struct that waits for timeouts duration and send a query cancellation request after it wake up.
- when stream is created we could bind alarm and stream into the same lifetime.
cons: implement complexity



should we also update the slow query timer to account for the waiting time for stream records?
`let _slow_query_timer = self.stats.start_slow_query_timer(stmt.clone());`

lyang24: cc @evenyag @sunng87

src/operator/src/statement.rs
```diff
@@ -343,6 +352,23 @@ impl StatementExecutor {
...
+                        feat: format!("Unsupported set variable {}", var_name),
+                    }
+                    .fail();
+                }
+            }
+            "STATEMENT_TIMEOUT" => {
+                if query_ctx.channel() == Channel::Postgres {
+                    set_query_timeout(set_var.value, query_ctx)?
+                } else {
+                    query_ctx.set_warning(format!("Unsupported set variable {}", var_name));
```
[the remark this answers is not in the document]
lyang24: thanks for the catch, I updated the logic here

src/operator/src/statement.rs
evenyag: @lyang24 Sorry for the late reply. I've been quite busy in the past two weeks 😢 

> cons: its wouldn't work on per statement timeout with bulk statements.

Per statement timeout is enough.

I'd vote for approach 2 but also apply the timeout to the method call itself.

We can implement the timeout logic in the [plan_exec()](https://github.com/GreptimeTeam/greptimedb/blob/2f2b4b306cb480c1396f452f2c27dc2206705d38/src/operator/src/statement.rs#L383-L390).

> [mysql](https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_max_execution_time) query timeout only apply to the read statements.

We may check whether the query is a read query in `execute_sql()`.
https://github.com/GreptimeTeam/greptimedb/blob/2f2b4b306cb480c1396f452f2c27dc2206705d38/src/operator/src/statement.rs#L128-L132

We can wrap the implementation without a timeout into a method like `plan_exec_inner()` and attach the timeout to it if provided in `plan_exec()` first.

Then we can use `tokio select` and the `stream!` macro to implement a timeout stream that tracks the timeout in the statement level.

Once we get the `Vec<Output>` from `plan_exec_inner()`, we can attach the timeout stream to the output. The `attach_timer()` is an example.
https://github.com/GreptimeTeam/greptimedb/blob/2f2b4b306cb480c1396f452f2c27dc2206705d38/src/frontend/src/instance.rs#L365-L376

lyang24: thanks for the guidance I learned a lot and I have implemented a draft of second commit. I wonder if i could get another around of feedback/ suggestions on testing when you have time (no rush at all).

src/operator/src/statement.rs
```diff
@@ -429,6 +432,115 @@ impl StatementExecutor {
...
+
+impl Stream for TimeoutStream
+{
+    type Item = common_recordbatch::error::Result<RecordBatch>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+
+        if let Some(last_poll) = this.last_poll {
+            if last_poll.elapsed() >= this.timeout_duration {
```
lyang24, line 513: a small issue: if a waker is send back after first pending poll and result is not polled again until waker signals ready, but its probably ok because client wont see this.

src/operator/src/statement.rs
```diff
@@ -408,6 +442,111 @@ impl StatementExecutor {
...
+/// For MySQL, it applies only to read-only statements.
+fn derive_timeout(stmt: &QueryStatement, query_ctx: &QueryContextRef) -> Option<Duration> {
+    if let Some(query_timeout) = query_ctx.query_timeout() {
+        return match (query_ctx.channel(), stmt) {
+            (Channel::Mysql, QueryStatement::Sql(Statement::Query(_)))
+            | (Channel::Postgres, QueryStatement::Sql(_)) => Some(query_timeout),
+            (_, _) => None,
+        };
+    }
+    None
```
evenyag, line 549:
```suggestion
    let query_timeout = query_ctx.query_timeout()?;
    match (query_ctx.channel(), stmt) {
        (Channel::Mysql, QueryStatement::Sql(Statement::Query(_)))
        | (Channel::Postgres, QueryStatement::Sql(_)) => Some(query_timeout),
        (_, _) => None,
    }
```

src/operator/src/statement.rs
```diff
@@ -390,8 +424,8 @@ impl StatementExecutor {
 
     #[tracing::instrument(skip_all)]
     async fn plan_exec(&self, stmt: QueryStatement, query_ctx: QueryContextRef) -> Result<Output> {
-        let plan = self.plan(&stmt, query_ctx.clone()).await?;
-        self.exec_plan(plan, query_ctx).await
+        let timeout = derive_timeout(&stmt, &query_ctx);
+        self.plan_exec_inner(stmt, query_ctx, timeout).await
```
evenyag, line 428: We can put the timeout logic in this method. The `plan_exec_inner()` only plans and executes the statement.
```suggestion
        let timeout = derive_timeout(&stmt, &query_ctx);
        match timeout {
            Some(timeout) => {
                let start = Instant::now();
                let output = tokio::time::timeout(timeout, self.plan_exec_inner()).await?;
                // compute remaining timeout......
                attach_timeout_stream(result?, remaining_timeout)
            },
            None => self.plan_exec_inner().await
        }
```

src/operator/src/statement.rs
```diff
@@ -343,6 +352,31 @@ impl StatementExecutor {
...
+                    query_ctx.set_warning(format!("Unsupported set variable {}", var_name))
+                }
+                _ => {
+                    return NotSupportedSnafu {
+                        feat: format!("Unsupported set variable {}", var_name),
+                    }
+                    .fail()
+                }
+            },
+            "STATEMENT_TIMEOUT" => {
```
evenyag, line 370: Maybe we can always support these two variables without considering the channel. @sunng87 @MichaelScofield

src/operator/src/statement/set.rs
```diff
@@ -177,3 +182,93 @@ pub fn set_datestyle(exprs: Vec<Expr>, ctx: QueryContextRef) -> Result<()> {
...
+
+// support time units in ms, s, min, h, d for postgres protocol.
+// https://www.postgresql.org/docs/8.4/config-setting.html#:~:text=Valid%20memory%20units%20are%20kB,%2C%20and%20d%20(days).
+fn parse_pg_query_timeout_input(input: &str) -> Result<u64> {
+    // Regex rules:
+    // The string must start with a number (one or more digits).
+    // The number must be followed by one of the valid time units (ms, s, min, h, d).
+    // The string must end immediately after the unit, meaning there can be no extra
+    // characters or spaces after the valid time specification.
+    let re = regex::Regex::new(r"^(\d+)(ms|s|min|h|d)$").context(BuildRegexSnafu)?;
```
evenyag, line 234: Use oncecell or lazystatic to initialize the regex

src/operator/src/statement/set.rs
```diff
@@ -177,3 +182,93 @@ pub fn set_datestyle(exprs: Vec<Expr>, ctx: QueryContextRef) -> Result<()> {
...
+                expr
+            ),
+        }
+        .fail(),
+    }
+}
+
+// support time units in ms, s, min, h, d for postgres protocol.
+// https://www.postgresql.org/docs/8.4/config-setting.html#:~:text=Valid%20memory%20units%20are%20kB,%2C%20and%20d%20(days).
+fn parse_pg_query_timeout_input(input: &str) -> Result<u64> {
```
evenyag, line 228: Can we use the humantime crate to parse it?

src/operator/src/statement/set.rs
```diff
@@ -177,3 +182,93 @@ pub fn set_datestyle(exprs: Vec<Expr>, ctx: QueryContextRef) -> Result<()> {
...
+        }
+        // postgres support time units i.e. SET STATEMENT_TIMEOUT = '50ms';
+        Expr::Value(Value::SingleQuotedString(timeout))
+        | Expr::Value(Value::DoubleQuotedString(timeout)) => {
+            if ctx.channel() != Postgres {
+                return NotSupportedSnafu {
+                    feat: format!("Invalid timeout expr {} in set variable statement", timeout),
+                }
+                .fail();
+            }
```
evenyag, line 211: It should be fine to support this in non-pg channels.

src/query/src/sql.rs
```diff
@@ -650,6 +650,23 @@ pub fn show_variable(stmt: ShowVariables, query_ctx: QueryContextRef) -> Result<
...
+                query_ctx.query_timeout_as_millis().to_string()
+            } else {
+                return UnsupportedVariableSnafu { name: variable }.fail();
+            }
+        }
+        "STATEMENT_TIMEOUT" => {
+            // Add time units to postgres query timeout display.
+            if query_ctx.channel() == Channel::Postgres {
+                let mut timeout = query_ctx.query_timeout_as_millis().to_string();
+                timeout.push_str("ms");
```
evenyag, line 664: Also consider using humantime.

src/session/src/context.rs
```diff
@@ -282,6 +283,22 @@ impl QueryContext {
...
+        self.mutable_session_data.read().unwrap().query_timeout
+    }
+
+    pub fn query_timeout_as_millis(&self) -> u128 {
+        let timeout = self.mutable_session_data.read().unwrap().query_timeout;
+        if let Some(t) = timeout {
+            return t.as_millis();
+        }
+        0
+    }
```
evenyag, line 297: This is unused.
```suggestion
```

evenyag (commented):
Please resolve the conflicts. You might need to merge the latest main branch.

src/query/src/sql.rs
lyang24: my concern is "2min 2s" is a valid parse to duraiton in humantime crate but is invalid in pg syntax
https://docs.rs/humantime/latest/humantime/

src/operator/src/statement/set.rs
lyang24: my concern is "2min 2s" is a valid parse to duraiton in humantime crate but is invalid in pg syntax
https://docs.rs/humantime/latest/humantime/

src/operator/src/statement.rs
lyang24: This will make testing easier, will make the change if this gets another vote.

src/operator/src/statement/set.rs
lyang24: good call fixed this with lazystatic

src/operator/src/statement/set.rs
```diff
@@ -177,3 +191,87 @@ pub fn set_datestyle(exprs: Vec<Expr>, ctx: QueryContextRef) -> Result<()> {
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::statement::set::parse_pg_query_timeout_input;
+
+    #[test]
+    fn test_parse_pg_query_timeout_input() {
+        assert!(parse_pg_query_timeout_input("").is_err());
```
evenyag, line 267: Add cases like:
```
"50"
"ms"
"a"
```

evenyag (approved):
Others LGTM

src/operator/src/statement/set.rs
lyang24: good call I just learned that postgres syntax allows 
`postgres=# set statement_timeout to '60000';`
i modified the parse_pg_query_timeout_input method slightly

src/operator/src/statement.rs
```diff
@@ -405,6 +450,89 @@ impl StatementExecutor {
...
+            return Poll::Ready(Some(StreamTimeoutSnafu.fail()));
+        }
+
+        Pin::new(&mut this.stream).poll_next(cx)
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.stream.size_hint()
+    }
+}
```
evenyag, line 525: I just realized that this stream implementation has a small issue: If the inner stream is pending and waiting for the result, the executor won't wake up the stream.

We should implement this using `stream!` and `select!`. For example:
```rust
stream! {
    let start = Instant::now();
    while let Some(item) = timeout(remaining_timeout, stream.try_next()).await? {
        yield item;
        // TODO: Use checked sub
        remaining_timeout -= Instant::now() - start;
    }
}
```

lyang24: yes i was [concerned](https://github.com/GreptimeTeam/greptimedb/pull/4792#discussion_r1825451682) about this but not sure how to implement. Thanks for the example.

src/operator/src/statement.rs
MichaelScofield: I'm thinking maybe it's better to move the protocol specific variables to their server implementations. For example, the `STATEMENT_TIMEOUT` is only valid in pg, then we can deal with it in the pg related mods in crate server. In doing so we can get rid of these `if`s to the `channel`. WDYT? @sunng87 @evenyag

killme2008:
Good job! Is it ready for review? @lyang24

src/operator/src/statement.rs
lyang24: Hi @evenyag, I tried to implement the suggestion but failed. I struggled with conversion of interfaces - In the end output accepts a `Pin<Box<dyn RecordBatchStream + Send>>;` . RecordBatchStream inherit from Stream trait and `stream!` return an impl of Stream trait - I am not sure how to convert the input and output generic type between  RecordBatchStream and Stream. Can you give me some guidence?


some of my code is here
```
fn attach_timeout(output: Output, timeout: Duration) -> Result<Output> {
    match output.data {
        OutputData::AffectedRows(_) | OutputData::RecordBatches(_) => Ok(output),
        OutputData::Stream(stream) => {
            let stream = attach_timeout_stream(stream, timeout)?;
            Ok(Output::new(OutputData::Stream(Box::pin(stream)), output.meta))
        }
    }
}

fn attach_timeout_stream<S>(
    mut stream: S,
    mut timeout: Duration,
) -> std::result::Result<impl Stream, common_recordbatch::error::Error>
where
    S: Stream<Item = Result<RecordBatch>> + Unpin
{
    Ok(stream! {
        let start = tokio::time::Instant::now();
        while let Some(item) = tokio::time::timeout(timeout, stream.next()).await.unwrap() {
            yield 
            timeout = timeout.checked_sub(tokio::time::Instant::now() - start).unwrap_or(Duration::ZERO);
        }
    })
}
```

evenyag: You can try this adapter.
https://github.com/GreptimeTeam/greptimedb/blob/d616bd92ef0ca8809ee0387d792714ccbcefb899/src/common/recordbatch/src/lib.rs#L229-L237

src/operator/src/statement.rs
evenyag: > the STATEMENT_TIMEOUT is only valid in pg, then we can deal with it in the pg related mods in crate server. In doing so we can get rid of these ifs to the channel.

Do we need to refactor the way we handle the variables?

src/operator/src/statement.rs
lyang24: Hi I am able to get the rust compiler to work with the new impl with stream! macro - i added Send + Unpin train bound to RecordBatchStreamWrapper just fyi.

src/common/recordbatch/src/lib.rs
```diff
@@ -227,14 +227,14 @@ impl Stream for SimpleRecordBatchStream {
 }
 
 /// Adapt a [Stream] of [RecordBatch] to a [RecordBatchStream].
-pub struct RecordBatchStreamWrapper<S> {
+pub struct RecordBatchStreamWrapper<S: Stream<Item = Result<RecordBatch>> + Send + Unpin> {
```
evenyag, line 230:
```suggestion
pub struct RecordBatchStreamWrapper<S> {
```

src/common/recordbatch/src/lib.rs
```diff
@@ -227,14 +227,14 @@ impl Stream for SimpleRecordBatchStream {
...
     pub schema: SchemaRef,
     pub stream: S,
     pub output_ordering: Option<Vec<OrderOption>>,
     pub metrics: Arc<ArcSwapOption<RecordBatchMetrics>>,
 }
 
-impl<S> RecordBatchStreamWrapper<S> {
+impl<S: Stream<Item = Result<RecordBatch>> + Send + Unpin> RecordBatchStreamWrapper<S> {
```
evenyag, line 237:
```suggestion
impl<S> RecordBatchStreamWrapper<S> {
```

src/common/recordbatch/src/lib.rs
```diff
@@ -246,7 +246,7 @@ impl<S> RecordBatchStreamWrapper<S> {
     }
 }
 
-impl<S: Stream<Item = Result<RecordBatch>> + Unpin> RecordBatchStream
+impl<S: Stream<Item = Result<RecordBatch>> + Send + Unpin> RecordBatchStream
```
evenyag, line 249: Why do we need additional `Send` bound?

src/operator/src/statement.rs
evenyag: Why do we need `Send` bound in `RecordBatchStreamWrapper`? Can we remove it?

src/common/recordbatch/src/lib.rs
lyang24: I thought we need the send to satisfy the Send trait bound for SendableRecordBatchStream but it turns out we don't need it. I will revert the changes src/common/recordbatch/src/lib.rs
`pub type SendableRecordBatchStream = Pin<Box<dyn RecordBatchStream + Send>>;`

src/operator/src/statement.rs
lyang24: removed

src/operator/src/statement.rs
MichaelScofield: @evenyag yes, actually I think it's an opportunity to get rid of the ugly "federated.rs" in mysql mod.

src/operator/src/statement.rs
```diff
@@ -338,6 +345,31 @@ impl StatementExecutor {
             "DATESTYLE" => set_datestyle(set_var.value, query_ctx)?,
 
             "CLIENT_ENCODING" => validate_client_encoding(set_var)?,
+            // TODO: write sqlness test for query timeout variables
```
MichaelScofield, line 348: An integration test that is mysql only is better. Let sqlness focus more on sql correctness.

src/operator/src/statement.rs
evenyag: @MichaelScofield Could you create an issue for this? We may not address it in this PR.

MichaelScofield: @evenyag https://github.com/GreptimeTeam/greptimedb/issues/1712

src/operator/src/statement.rs
lyang24: just learned `-- SQLNESS PROTOCOL MYSQL` can trigger mysql channel very useful.
In commit 3: removed the old comment and added basic set and show mysql only test, wonder if we have additional tools to trigger a actual timeout in test (I was think about `SELECT SLEEP(3);`  but its not supported yet).

tests/cases/standalone/common/basic.result
```diff
@@ -179,3 +179,22 @@ DROP TABLE foo;
...
+
+affected_rows: 0
+
+-- SQLNESS PROTOCOL MYSQL
+SHOW VARIABLES MAX_EXECUTION_TIME;
+
++---------------+-------+
+| Variable_name | Value |
++---------------+-------+
+|               |       |
```
MichaelScofield, line 193: Empty result?

lyang24: This is expected see https://github.com/GreptimeTeam/greptimedb/issues/4799

MichaelScofield (approved):

PR #4797, opened 2024-10-05
feat: add json format output for http interface

### LINKED ISSUE #4764: JSON format output for http query interface

API improvement

Some BI or visualization tools like PowerBI accepts json data from custom web apis. So if we offer json based http API, it's possible to integration these type of tool without any modification from both side. 

Before, we need to modify the database driver of these type of tools, like [this](https://github.com/GreptimeTeam/metabase-greptimedb-driver) and [this](https://github.com/sunng87/greptimedb-sqlalchemy), in order to make it work with GreptimeDB. But for close-source tools or hosted ones, there is no way to do that.

Our default json output is in a table based data mode, which is not very friendly for client to consume directly. 

```json
{
  "output": [
    {
      "records": {
        "schema": {
          "column_schemas": [
            {
              "name": "host",
              "data_type": "String"
            },
            {
              "name": "ts",
              "data_type": "TimestampMillisecond"
            },
            {
              "name": "cpu",
              "data_type": "Float64"
            },
            {
              "name": "memory",
              "data_type": "Float64"
            }
          ]
        },
        "rows": [
          [
            "127.0.0.1",
            1667446797450,
            0.1,
            0.4
          ],
          [
            "127.0.0.1",
            1667446798450,
            0.5,
            0.2
          ],
          [
            "127.0.0.2",
            1667446798450,
            0.2,
            0.3
          ]
        ],
        "total_rows": 3
      }
    }
  ],
  "execution_time_ms": 23
}
```

The idea it to create an alternative format, say `format=json`, and the data format should look like:

```json
{
  "data": [
    {
      "host": "127.0.0.1",
      "ts": 1667446797450,
      "cpu": 0.2,
      "memory": 0.3
    },
    ...
  ],
  "execution_time_ms": 23
}
```

### Implementation challenges

_No response_

Kev1n8:

Closes #4764.

> Some BI or visualization tools like PowerBI accepts json data from custom web apis.

As explained in #4764, it's better if the http service return a json format that is already supported by other products.

- Added JsonResponse.
- Placed relevant tests.

## Checklist

### DIFF
--- a/src/servers/src/http/json_result.rs
+++ b/src/servers/src/http/json_result.rs
@@ -1,0 +1,135 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use axum::http::{header, HeaderValue};
+use axum::response::{IntoResponse, Response};
+use common_error::status_code::StatusCode;
+use common_query::Output;
+use mime_guess::mime;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Map};
+
+use super::process_with_limit;
+use crate::http::error_result::ErrorResponse;
+use crate::http::header::{GREPTIME_DB_HEADER_EXECUTION_TIME, GREPTIME_DB_HEADER_FORMAT};
+use crate::http::{handler, GreptimeQueryOutput, HttpResponse, ResponseFormat};
+
+/// The json format here is different from the default json output of greptime_result.
+/// `JsonResponse` is intended to make it easier for user to consume data.
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+pub struct JsonResponse {
+    output: Vec<GreptimeQueryOutput>,
+    execution_time_ms: u64,
+}
+
+impl JsonResponse {
+    pub async fn from_output(outputs: Vec<crate::error::Result<Output>>) -> HttpResponse {
+        match handler::from_output(outputs).await {
+            Err(err) => HttpResponse::Error(err),
+            Ok((output, _)) => {
+                if output.len() > 1 {
+                    HttpResponse::Error(ErrorResponse::from_error_message(
+                        StatusCode::InvalidArguments,
+                        "cannot output multi-statements result in json format".to_string(),
+                    ))
+                } else {
+                    HttpResponse::Json(JsonResponse {
+                        output,
+                        execution_time_ms: 0,
+                    })
+                }
+            }
+        }
+    }
+
+    pub fn output(&self) -> &[GreptimeQueryOutput] {
+        &self.output
+    }
+
+    pub fn with_execution_time(mut self, execution_time: u64) -> Self {
+        self.execution_time_ms = execution_time;
+        self
+    }
+
+    pub fn execution_time_ms(&self) -> u64 {
+        self.execution_time_ms
+    }
+
+    pub fn with_limit(mut self, limit: usize) -> Self {
+        self.output = process_with_limit(self.output, limit);
+        self
+    }
+}
+
+impl IntoResponse for JsonResponse {
+    fn into_response(mut self) -> Response {
+        debug_assert!(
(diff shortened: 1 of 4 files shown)


Kev1n8:
The branch is including my last PR. Not sure what the cause is.

sunng87:
I think it's because you created this branch from your last feature branch. If your previous branch has been merged, you can rebase to main to fix that

Kev1n8:
> I think it's because you created this branch from your last feature branch. If your previous branch has been merged, you can rebase to main to fix that

I thought so too. thx

sunng87:
Almost LGTM. It would be nice if we can add a test case in `tests-integration/tests/http.rs`

Kev1n8:
Added a test case. Let me know if anything else is needed.

src/servers/src/http/json_result.rs
```diff
@@ -0,0 +1,135 @@
...
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Map};
+
+use super::process_with_limit;
+use crate::http::error_result::ErrorResponse;
+use crate::http::header::{GREPTIME_DB_HEADER_EXECUTION_TIME, GREPTIME_DB_HEADER_FORMAT};
+use crate::http::{handler, GreptimeQueryOutput, HttpResponse, ResponseFormat};
+
+/// The json format here is different from the default json output of greptime_result.
```
killme2008, line 29:
```suggestion
/// The json format here is different from the default json output of `GreptimedbV1` result.
```

src/servers/src/http/json_result.rs
```diff
@@ -0,0 +1,135 @@
...
+            self.output.len()
+        );
+
+        let execution_time = self.execution_time_ms;
+        let payload = match self.output.pop() {
+            None => String::default(),
+            Some(GreptimeQueryOutput::AffectedRows(n)) => json!({
+                "data": [
+                    {
+                        "affectedrows": n
```
killme2008, line 90:
```suggestion
                        "affected_rows": n
```

src/servers/src/http/json_result.rs
```diff
@@ -0,0 +1,135 @@
...
+                let schema = records.schema();
+
+                for row in records.rows.iter() {
+                    let mut row_map = Map::new();
+                    for (i, col) in schema.column_schemas.iter().enumerate() {
+                        row_map.insert(col.name.clone(), row[i].clone());
+                    }
+                    data.push(row_map);
+                }
+
```
killme2008, line 108: Prefer to use the iterators to generate the results.

src/servers/src/http/json_result.rs
```diff
@@ -0,0 +1,135 @@
...
+        let mut resp = (
+            [(
+                header::CONTENT_TYPE,
+                HeaderValue::from_static(mime::APPLICATION_JSON.as_ref()),
+            )],
+            payload,
+        )
+            .into_response();
+        resp.headers_mut().insert(
```
killme2008, line 125: Why not move these headers above into the construction of `resp`? Then we don't need to `insert` them.

src/servers/src/http/json_result.rs
shuiyisong, line 90: How about we put `affectedrows` into the outer layer with `execution_time_ms`, treating it like meta info. 

cc @sunng87 

For example:

```JSON
{
    "data": [],
    "affected_rows": 10,
    "execution_time_ms": 10
}
```

Kev1n8: I did this because  in the other format it is `affectedrows`. maybe we should change them all together.

killme2008: I am not quite sure, i think we are always using `affected_rows ` or `AffectedRows` right now.

src/servers/src/http/json_result.rs
Kev1n8: will do

src/servers/src/http/json_result.rs
sunng87: sounds good

src/servers/src/http/json_result.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        let execution_time = self.execution_time_ms;
+        let payload = match self.output.pop() {
+            None => String::default(),
+            Some(GreptimeQueryOutput::AffectedRows(n)) => json!({
+                "data": [
+                    {
+                        "affected_rows": n
+                    },
+                ],
```
sunng87, line 92: @Kev1n8 as per @shuiyisong , can we move `affected_rows` to top level? this will make client easier to check the resultset.

killme2008 (approved):
LGTM. Thanks a lot.

shuiyisong (approved):
LGTM

BTW: it would be nice to add more tests, such as `affected_rows` cases and `select count(1)` cases where the column name is not the schema column name.
Possible future works: add `total_rows` in metadata for queries.

PR #4798, opened 2024-10-05
feat: add function to aggregate path into a geojson path

sunng87:

In greptimedb, we store waypoints of a path row by row. However, when using with data process or visualization software, it treats a trajectory as a single value, either in geojson or other formats. 

This patch introduces a new uadf `geojson_encode_path` that takes rows of latitude, longitude and timestamp to output a path that compatible with deckgl's path input.

It also refactors our macro for aggregation function to remove `use` from macro to avoid multiple use error when using the macro.

sqlness tests included.

## Checklist


killme2008 (approved):
LGTM

fengjiachun (approved):
LGTM

PR #4804, opened 2024-10-08
feat(mito): Support background task priority

v0y4g3r:

This PR adds priorities to background tasks like flush/compaction/file purge.

In current local scheduler implementation, the scheduler spawns some corountines to poll background tasks from a FIFO and wait until the job finishes then poll the next one. In intense insert workloads, the flush jobs would be blocked by compaction jobs if we don't assign a priority to background jobs.

This PR introduces an asynchronous priority channel with two levels: High and Low. It includes an optional deadline for low-priority jobs. The channel first attempts to poll jobs from the high-priority queue, switching to the low-priority queue only if the high-priority queue is empty. If a job at the head of the low-priority queue reaches its deadline, it will be polled immediately. Although jobs in low-priority channel have deadline, all jobs enqueued still follow the FIFO convention.

## Checklist

### DIFF
--- a/src/mito2/src/compaction.rs
+++ b/src/mito2/src/compaction.rs
@@ -63,7 +63,7 @@ use crate::request::{OptionOutputTx, OutputTx, WorkerRequest};
 use crate::schedule::remote_job_scheduler::{
     CompactionJob, DefaultNotifier, RemoteJob, RemoteJobSchedulerRef,
 };
-use crate::schedule::scheduler::{Job, SchedulerRef};
+use crate::schedule::scheduler::{Job, Priority, SchedulerRef};
 use crate::sst::file::{FileHandle, FileId, FileMeta, Level};
 use crate::sst::version::LevelMeta;
 use crate::worker::WorkerListener;
@@ -362,12 +362,13 @@ impl CompactionScheduler {
 
         // Submit the compaction task.
         self.scheduler
-            .schedule(Job {
-                r#type: "compaction",
-                task: Box::pin(async move {
+            .schedule(Job::new(
+                "compaction",
+                Priority::Low(Some(Duration::from_secs(60))),
+                Box::pin(async move {
                     local_compaction_task.run().await;
                 }),
-            })
+            ))
             .map_err(|e| {
                 error!(e; "Failed to submit compaction request for region {}", region_id);
                 // If failed to submit the job, we need to remove the region from the scheduler.
--- a/src/mito2/src/compaction.rs
+++ b/src/mito2/src/compaction.rs
@@ -63,7 +63,7 @@ use crate::request::{OptionOutputTx, OutputTx, WorkerRequest};
 use crate::schedule::remote_job_scheduler::{
     CompactionJob, DefaultNotifier, RemoteJob, RemoteJobSchedulerRef,
 };
-use crate::schedule::scheduler::SchedulerRef;
+use crate::schedule::scheduler::{Job, SchedulerRef};
 use crate::sst::file::{FileHandle, FileId, FileMeta, Level};
 use crate::sst::version::LevelMeta;
 use crate::worker::WorkerListener;
@@ -362,9 +362,12 @@ impl CompactionScheduler {
 
         // Submit the compaction task.
         self.scheduler
-            .schedule(Box::pin(async move {
-                local_compaction_task.run().await;
-            }))
+            .schedule(Job {
+                r#type: "compaction",
+                task: Box::pin(async move {
+                    local_compaction_task.run().await;
+                }),
+            })
             .map_err(|e| {
                 error!(e; "Failed to submit compaction request for region {}", region_id);
                 // If failed to submit the job, we need to remove the region from the scheduler.
--- a/src/mito2/src/schedule.rs
+++ b/src/mito2/src/schedule.rs
@@ -12,5 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[allow(dead_code)]
+mod priority;
 pub mod remote_job_scheduler;
 pub mod scheduler;
--- a/src/mito2/src/schedule/priority.rs
+++ b/src/mito2/src/schedule/priority.rs
@@ -1,0 +1,323 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::ops::Add;
+use std::pin::Pin;
+use std::time::{Duration, Instant};
+
+use async_stream::stream;
+use futures::Stream;
+use tokio::select;
+use tokio::time::Sleep;
+
+use crate::error;
+
+pub fn bounded<T>(size: usize) -> (Sender<T>, Receiver<T>) {
+    let (tx_low, rx_low) = async_channel::bounded(size);
+    let (tx_high, rx_high) = async_channel::bounded(size);
+    (Sender { tx_low, tx_high }, Receiver { rx_low, rx_high })
+}
+
+pub fn unbounded<T>() -> (Sender<T>, Receiver<T>) {
+    let (tx_low, rx_low) = async_channel::unbounded();
+    let (tx_high, rx_high) = async_channel::unbounded();
+    (Sender { tx_low, tx_high }, Receiver { rx_low, rx_high })
+}
+
+#[derive(Clone)]
+pub struct Sender<T> {
+    tx_low: async_channel::Sender<(Instant, T)>,
+    tx_high: async_channel::Sender<T>,
+}
+
+impl<T> Sender<T> {
+    pub async fn send_high(&self, item: T) -> error::Result<()> {
+        self.tx_high.send(item).await.map_err(|e| {
+            error::SendToChannelSnafu {
+                err_msg: format!("{:?}", e),
+            }
+            .build()
+        })
+    }
+
+    pub async fn send_low(&self, item: T, deadline: Option<Duration>) -> error::Result<()> {
+        self.tx_low
+            .send((
+                Instant::now().add(deadline.unwrap_or(Duration::from_secs(0))),
+                item,
+            ))
+            .await
+            .map_err(|e| {
+                error::SendToChannelSnafu {
+                    err_msg: format!("{:?}", e),
+                }
+                .build()
+            })
+    }
+}
+
+#[derive(Clone)]
+pub struct Receiver<T> {
+    rx_low: async_channel::Receiver<(Instant, T)>,
+    rx_high: async_channel::Receiver<T>,
+}
+
+impl<T> Receiver<T>
+where
+    T: Send + 'static,
+{
+    pub fn into_stream(self) -> Pin<Box<dyn Stream<Item = T> + Send>> {
+        let s = stream!({
+            let mut timer: Option<Pin<Box<Sleep>>> = None;
+            let mut pending: Option<T> = None;
+            loop {
+                select! {
+                    biased;
+                    Some(_) = maybe_timeout(&mut timer) => {
+                        if let Some(timed) = std::mem::take(&mut pending){
+                            yield timed
+                        }
+                    }
+
+                    Ok(high) = self.rx_high.recv() => {
+                        // also check if low channel has pending.
+                        if pending.is_none() && let Ok((deadline, low_item)) = self.rx_low.try_recv() {
+                            let now = Instant::now();
+                            if deadline > now {
+                                let tta = deadline - now;
+                                timer = Some(Box::pin(tokio::time::sleep(tta)));
+                            }
+                            pending = Some(low_item);
+                        }
+                        yield high;
+                    }
+
+                    Ok((_, low)) = self.rx_low.recv() => {
+                        if let Some(timed) = std::mem::take(&mut pending){
+                            timer = None;
+                            yield timed;
+                        }
+                        yield low
+                    }
+
+                    else => {
+                        if self.rx_high.is_closed() && self.rx_low.is_closed() {
+                            return;
+                        }
+                    }
+                }
+            }
+        });
+        Box::pin(s)
+    }
+}
+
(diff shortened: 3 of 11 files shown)


v0y4g3r:
@coderabbitai review

src/mito2/src/compaction.rs
```diff
@@ -362,9 +362,13 @@ impl CompactionScheduler {
...
-            .schedule(Box::pin(async move {
-                local_compaction_task.run().await;
-            }))
+            .schedule(Job::new(
+                "compaction",
+                Priority::Low(Some(Duration::from_secs(60))),
+                Box::pin(async move {
+                    local_compaction_task.run().await;
+                }),
+            ))
```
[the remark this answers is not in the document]
v0y4g3r: false positive.

src/mito2/src/schedule/priority.rs
```diff
@@ -0,0 +1,416 @@
...
+                    biased;
+                    Some(_) = maybe_timeout(&mut timer) => {
+                        if let Some(timed) = std::mem::take(&mut pending){
+                            yield timed
+                        }
+                    }
+
+                    Ok(high) = self.rx_high.recv() => {
+                        // also check if low channel has pending.
+                        if pending.is_none() && let Ok((deadline, low_item)) = self.rx_low.try_recv() {
```
evenyag, line 117: If we have many low-priority tasks exceed their deadline, it is still possible that we have no chance to execute the task with high prority.

src/mito2/src/schedule/priority.rs
```diff
@@ -0,0 +1,416 @@
...
+impl<T> Receiver<T> {
+    /// Converts the receiver into a async stream.
+    pub fn into_stream(self) -> impl Stream<Item = T> {
+        stream!({
+            let mut timer: Option<Pin<Box<Sleep>>> = None;
+            let mut pending: Option<T> = None;
+            loop {
+                select! {
+                    biased;
+                    Some(_) = maybe_timeout(&mut timer) => {
```
evenyag, line 109: Maybe we can simplify the channel to always execute a pending low priority task by a fixed interval or after executing N high priority task.

We can remove the deadline argument from the prority.

src/mito2/src/schedule/scheduler.rs
```diff
@@ -16,16 +16,58 @@ use std::future::Future;
...
+pub enum Priority {
+    /// High priority job.
+    High,
+    /// Low priority job with an optional deadline.
+    Low(Option<Duration>),
+}
+
+pub struct Job {
+    /// Job type.
+    r#type: &'static str,
```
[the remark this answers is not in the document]
evenyag: I think `job_type` has better readability.

src/mito2/src/schedule/scheduler.rs
```diff
@@ -16,16 +16,58 @@ use std::future::Future;
...
+    /// Job priority.
+    priority: Priority,
```
evenyag, line 44: I think we can provide a method that send by priority in the channel side.

src/mito2/src/schedule/scheduler.rs
```diff
@@ -61,34 +103,36 @@ impl LocalScheduler {
     ///
     /// concurrency: the number of bounded receiver
     pub fn new(concurrency: usize) -> Self {
-        let (tx, rx) = async_channel::unbounded();
+        let (tx, rx) = priority::unbounded::<Job>();
```
evenyag, line 106: Maybe we need a gauge for the number of pending jobs.

src/mito2/src/schedule/priority.rs
```diff
@@ -0,0 +1,416 @@
...
+                        if pending.is_none() && let Ok((deadline, low_item)) = self.rx_low.try_recv() {
+                            let now = Instant::now();
+                            if let Some(deadline) = deadline {
+                                if deadline > now {
+                                    let tta = deadline - now;
+                                    timer = Some(Box::pin(tokio::time::sleep(tta)));
+                                    pending = Some(low_item);
+                                } else {
+                                    // already
+                                    yield low_item;
```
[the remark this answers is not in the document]
v0y4g3r: > We will likely lose the high task in this branch.

No, the high priority item will still be emitted here: https://github.com/GreptimeTeam/greptimedb/blob/e7cb80ee6d84a309774ac5b72d50e2bf7f88418b/src/mito2/src/schedule/priority.rs#L133

src/mito2/src/schedule.rs
```diff
@@ -12,5 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
v0y4g3r: fixed in 26c48af902

src/mito2/src/compaction.rs
```diff
@@ -362,9 +362,13 @@ impl CompactionScheduler {
 
         // Submit the compaction task.
         self.scheduler
-            .schedule(Box::pin(async move {
-                local_compaction_task.run().await;
-            }))
+            .schedule(Job::new(
+                "compaction",
+                Priority::Low(Some(Duration::from_secs(60))),
```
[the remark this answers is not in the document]
v0y4g3r: fixed in 26c48af902

src/mito2/src/schedule/scheduler.rs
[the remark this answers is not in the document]
v0y4g3r: fixed in fixed in 26c48af902

src/mito2/src/schedule/priority.rs
v0y4g3r: > If we have many low-priority tasks exceed their deadline, it is still possible that we have no chance to execute the task with high prority.

Maybe we should figure out a proper deadline.

src/mito2/src/schedule/priority.rs
v0y4g3r: > Maybe we can simplify the channel to always execute a pending low priority task by a fixed interval or after executing N high priority task.
> 
> We can remove the deadline argument from the prority.

The initial implementation was like this, but the task waiting time will be indefinite.

src/mito2/src/schedule/scheduler.rs
v0y4g3r: > I think we can provide a method that send by priority in the channel side.

Maybe add a priority argument to `Scheduler::schedule`:https://github.com/GreptimeTeam/greptimedb/blob/f3fc31708a3a213dbc12d59a42b4860d5f658532/src/mito2/src/schedule/scheduler.rs#L81

src/mito2/src/schedule/scheduler.rs
v0y4g3r: > Maybe we need a gauge for the number of pending jobs.

fixed in f3fc31708a

src/mito2/src/schedule/priority.rs
```diff
@@ -0,0 +1,447 @@
...
+                        }
+                    }
+
+                    Ok(high) = self.rx_high.recv() => {
+                        // also check if low channel has pending.
+                        if pending.is_none() && let Ok((deadline, low_item)) = self.rx_low.try_recv() {
+                            let now = Instant::now();
+                            if let Some(deadline) = deadline {
+                                if deadline > now {
+                                    let tta = deadline - now;
```
evenyag, line 114: What does `tta` mean?

src/mito2/src/compaction.rs
```diff
@@ -362,9 +362,13 @@ impl CompactionScheduler {
 
         // Submit the compaction task.
         self.scheduler
-            .schedule(Box::pin(async move {
-                local_compaction_task.run().await;
-            }))
+            .schedule(Job::new(
+                "compaction",
+                Priority::Low(Some(Duration::from_secs(5 * 60))),
```
evenyag, line 367: Do we need to define a constant for this?

src/mito2/src/schedule/priority.rs
evenyag: > but the task waiting time will be indefinite.

If we send low-pri jobs in this order: `None`, `deadline 5m`, then the job with `None` can still block the job with `deadline 5m` longer than 5 minutes.

src/mito2/src/schedule/priority.rs
```diff
@@ -0,0 +1,447 @@
...
+impl<T> Clone for Receiver<T> {
+    fn clone(&self) -> Self {
+        Self {
+            rx_high: self.rx_high.clone(),
+            rx_low: self.rx_low.clone(),
+        }
+    }
+}
+
+impl<T> Receiver<T> {
```
evenyag, line 93: I wonder if separating workers for low and high priorities like rocksdb is more robust
https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide#parallelism-options

Each channel services a worker pool. Then if the number of low priority workers is less than the number of global runtime workers, we can ensure compaction never blocks flush. (We may need to limit the concurrency of a single compaction task, now it spawns many sub tasks)

src/mito2/src/compaction.rs
evenyag, line 367: out-of-topic: We may assign high-pri or allow configuring the priority if the compaction is a manually triggered by users.

src/mito2/src/schedule/priority.rs
v0y4g3r: > > but the task waiting time will be indefinite.
> 
> If we send low-pri jobs in this order: `None`, `deadline 5m`, then the job with `None` can still block the job with `deadline 5m` longer than 5 minutes.

Yes, the enqueued jobs in either high or low priority channels are still guaranteed FIFO

src/mito2/src/schedule/priority.rs
v0y4g3r: > What does `tta` mean?

time-to-arrival

src/mito2/src/schedule/priority.rs
v0y4g3r: > I wonder if separating workers for low and high priorities like rocksdb is more robust https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide#parallelism-options
> 
> Each channel services a worker pool. Then if the number of low priority workers is less than the number of global runtime workers, we can ensure compaction never blocks flush. (We may need to limit the concurrency of a single compaction task, now it spawns many sub tasks)

I tried not to touch the scheduler itself in this PR, but tbh it's a bit confusing, especially in that flush jobs are executed in place while compaction tasks are spawned to separate runtime but the scheduler still has to wait for it to complete. I did some benchmark but not yet decided whether to refactor the scheduler so that we can  control the parallelism a in better manner.

src/mito2/src/schedule/scheduler.rs
```diff
@@ -115,13 +163,19 @@ impl Scheduler for LocalScheduler {
...
-            .context(InvalidSchedulerStateSnafu)?
-            .try_send(job)
-            .map_err(|_| InvalidSenderSnafu {}.build())
+        let binding = self.sender.read().unwrap();
+        let sender = binding.as_ref().context(InvalidSchedulerStateSnafu)?;
+        let job_type = job.job_type;
+        let res = match job.priority {
+            Priority::High => sender.try_send_high(job),
+            Priority::Low(deadline) => sender.try_send_low(job, deadline),
+        };
```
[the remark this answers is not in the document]
v0y4g3r: fixed in [0ce7860a2d](https://github.com/GreptimeTeam/greptimedb/commit/0ce7860a2d5b5fd8cc33a8b40b343a1182ad461b#diff-259752964dd67102a325e6a30328b3a4015bddebd0f8f9c8553806142d7bde32R42)

src/mito2/src/compaction.rs
v0y4g3r: > Do we need to define a constant for this?

fixed in f1354bf049

src/mito2/src/schedule/scheduler.rs
```diff
@@ -115,13 +163,19 @@ impl Scheduler for LocalScheduler {
...
+        if res.is_ok() {
+            SCHEDULER_PENDING_JOBS.with_label_values(&[job_type]).add(1);
+            Ok(())
+        } else {
+            InvalidSenderSnafu {}.fail()
```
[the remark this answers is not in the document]
v0y4g3r: > We already mapped the error into `SendToChannel` error in the channel so we don't need to return this error here. Maybe we only need to keep one of them.

fixed in https://github.com/GreptimeTeam/greptimedb/commit/0ce7860a2d5b5fd8cc33a8b40b343a1182ad461b#diff-259752964dd67102a325e6a30328b3a4015bddebd0f8f9c8553806142d7bde32R42

src/mito2/src/compaction.rs
v0y4g3r: > out-of-topic: We may assign high-pri or allow configuring the priority if the compaction is a manually triggered by users.

Good catch. I fixed in b68a73a5d0 so that manually triggered compaction tasks will obtain High priorities.

src/mito2/src/schedule/priority.rs
```diff
@@ -0,0 +1,447 @@
...
+use std::pin::Pin;
+use std::time::Instant;
+
+use async_stream::stream;
+use futures::Stream;
+use tokio::select;
+use tokio::time::Sleep;
+
+use crate::error;
+use crate::schedule::scheduler::Priority;
```
evenyag, line 25: non-blocking: We can define Priority in this mod.

src/mito2/src/schedule/priority.rs
evenyag: We should avoid spawning tasks inside the compaction job. We might create a job for each output instead of processing jobs in parallel inside the scheduler.

> compaction tasks are spawned to separate runtime but the scheduler still has to wait for it to complete.

Yes, it also blocks the scheduler. Maybe the scheduler for compaction should run in the compact runtime.

> not yet decided whether to refactor the scheduler so that we can control the parallelism a in better manner.

We may refactor the local compactor first. What's more, we may need a different concurrency control mechanism if we use a remote compactor. A trivial approach is providing a larger max compaction jobs limit if we use the remote compactor.

Maybe we can implement the scheduler in a simple style: It just has a semaphore and always spawns the job and acquires the semaphore.

evenyag (approved):
This is a workaround before we refactor the scheduler.

waynexia:
This implementation itself looks good to me. But I'm thinking this mechanism of assigning priority outside of runtime has some natural limitations...

>This is a workaround before we refactor the scheduler.

Are you working on this? @evenyag

v0y4g3r:
> This implementation itself looks good to me. But I'm thinking this mechanism of assigning priority outside of runtime has some natural limitations...
> 
> > This is a workaround before we refactor the scheduler.
> 
> Are you working on this? @evenyag

I think we can close this PR now.

PR #4757, opened 2024-09-23
feat: unordered scanner scans data by time ranges

evenyag:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

This PR adds the initial implementation for the unordered scanner to support reading rows from time ranges individually.

The scanner groups sources (files and memtables) into ranges and maintains a `RangeMeta` to store the metadata of that range. Each `PartitionRange` refers to a `RangeMeta` in the `StreamContext` so we can use the `PartitionRange::identifier` to locate the `RangeMeta`.

`SeqScan` and `UnorderedScan` have different grouping strategies:
- `SeqScan` merges all overlapping files and memtables to ensure output rows are sorted. So it puts overlapping files and memtables in the same range.
  - If the range only consists of one source, then it can further split row groups of the source into multiple ranges
- `UnorderedScan` can assign each row group to a dedicated range as it doesn't sort and merge duplication rows

The implementations of grouping strategies are similar to `SeqDistributor` and `UnorderedDistributor`.

It also refactors the `scan_partition()` method by wrapping some methods to scan memtables and files.

## Checklist

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.

## Release Notes

- **New Features**
  - Introduced a new module for managing partition ranges, enhancing data scanning capabilities.
  - Added methods to handle pruning of memtable and file ranges, improving efficiency during scans.

- **Improvements**
  - Updated return types for several methods to utilize `BTreeMap`, enhancing data organization and access.
  - Enhanced internal state management with the addition of `num_ranges` to track range counts.

- **Bug Fixes**
  - Removed unnecessary code attributes to streamline the codebase.

- **Documentation**
  - Added comments for future development tasks to improve code maintainability.

PR #4809, opened 2024-10-10
feat: Seq scanner scans data by time range

evenyag:

https://github.com/GreptimeTeam/greptimedb/pull/4757

Seq scanner scans data according to the order of partition ranges.

Moves `scan_file_ranges` and `scan_mem_ranges` to `scan_util`. Refactors them so that `SeqScan` and `UnorderedScan` can reuse them.

To reuse codes, this PR adds a new metric struct `PartitionMetrics`, and shares it between streams in the same partition. This struct also prints the debug log in `drop()` so we can still get the log when the stream is dropped before exhausted.

It removes all unused codes.

There are still some remaining works:
- Remove the scan parallelism. But we need to support file-level parallelism.
- Support splitting multiple row groups in `SeqScan`.
- Support field pruning in last_non_null mode if the time range only has one file.
- More tests for `RangeMeta` and `StreamContext`

## Checklist


src/mito2/src/read/seq_scan.rs
```diff
@@ -525,292 +290,66 @@ impl RegionScanner for SeqScan {
...
-        for ranges in part.file_ranges[0].chunks(ranges_per_part) {
-            let new_part = ScanPart {
-                memtable_ranges: Vec::new(),
-                file_ranges: smallvec![ranges.to_vec()],
-                time_range: part.time_range,
+            let read_type = if compaction {
+                "compaction"
+            } else {
+                "seq_scan_files"
             };
```
waynexia, line 331: why not use `get_scanner_type` here

waynexia (approved):

src/mito2/src/read/seq_scan.rs
evenyag: It doesn't return the expected label value for this metric.

evenyag:
https://github.com/GreptimeTeam/greptimedb/pull/4809/commits/74b4088b325cd2ff1dffacd693e58512b9d47a8a changed the `PartitionRange::end` to exclusive. @waynexia @discord9

src/mito2/src/read/range.rs
```diff
@@ -264,3 +287,77 @@ fn maybe_split_ranges_for_seq_scan(ranges: Vec<RangeMeta>) -> Vec<RangeMeta> {
...
+                (1, 1000, 2000),
+                (2, 6000, 7000),
+                (3, 0, 1500),
+                (4, 1500, 3000),
+            ],
+            &[(vec![3, 1, 4], 0, 3000), (vec![2], 6000, 7000)],
+        );
+
+        // 1, 2, 3 => [3], [1], [2],
+        run_group_ranges_test(
```
discord9, line 354: Maybe add a `(1, 3000, 4000), (2, 4000, 5000)` case which will concat to `[1,2]`?

discord9 (approved):
others LGTM

PR #4814, opened 2024-10-11
feat: window sort physical plan

discord9:

window sort physical plan, which take a list of `PartitionRange` and a input stream coming in as the concated, sorted only in given `PartitionRange` stream, and merge sort them using `streaming_merge` and emit merge results as soon as possible.

- `WindowedSortExec`: The physical plan, which create a `WindowedSortStream`
- `WindowedSortStream`: the core logic of merge sort, which is basically find sorted run, and merge sort them when possible
- `compute_all_working_ranges` and other util functions: find the smallest possible merge sort duration, so once data in input stream excess this range future values can't appeal in this range again, hence safe to merge sort this range, the core logic of merge sort happens in `WindowedSortStream::poll_next_inner` which determine when to merge sort
- test: most util function and `WindowedSortExec`&`WindowedSortStream` are tested, and most cases are covered, welcome to add more

Alternative options:
sort algo that sort partial-ordered list quickly, but since we have more prior information（i.e. the range of partial sorted runs) we have this window sort, which can be think of like a variant of tim sort I think?
## Checklist


src/query/src/window_sort.rs
```diff
@@ -0,0 +1,3101 @@
...
+    batch: DfRecordBatch,
+    expression: &PhysicalSortExpr,
+) -> datafusion_common::Result<SortedRunSet<Timestamp>> {
+    // split input rb to sorted runs
+    let sort_column = expression.evaluate_to_sort_column(&batch)?;
+    let sorted_runs_offset = get_sorted_runs(sort_column.clone())?;
+    if let Some(run) = sorted_runs_offset.first()
+        && run.offset == 0
+        && run.len == batch.num_rows()
+        && sorted_runs_offset.len() == 1
```
evenyag, line 592: if `sorted_runs_offset.len() == 1`, it must satisfy `run.offset == 0` and `run.len == batch.num_rows()`. Maybe we should check this.

src/query/src/window_sort.rs
```diff
@@ -0,0 +1,3101 @@
...
+    {
+        // input rb is already sorted, we can emit it directly
+        Ok(SortedRunSet {
+            runs_with_batch: vec![(batch, run.clone())],
+            sort_column,
+        })
+    } else {
+        // those slice should be zero copy, so supposedly no new reservation needed
+        let mut ret = Vec::new();
+        for run in sorted_runs_offset {
```
evenyag, line 602: nitpick: reserve capacity for `ret`?

src/query/src/window_sort.rs
```diff
@@ -0,0 +1,3101 @@
...
+    // get slice that in which all data that `min_val<=data<max_val`
+    let (start, end) = if descending {
+        // note that `data < max_val`
+        // i,e, for max_val = 4, array = [5,3,2] should be start=1
+        // max_val = 4, array = [5, 4, 3, 2] should be start= 2
+        let start = bisect::<false>(&[array.clone()], &[max_val.clone()], &[*opt])?;
+        // min_val = 1, array = [3, 2, 1, 0], end = 2
+        // min_val = 1, array = [3, 2, 0], end = 2
+        let end = bisect::<false>(&[array.clone()], &[min_val.clone()], &[*opt])?;
+        (start, end)
```
evenyag, line 696: if `min_val = 1, array = [3, 2, 1, 0], end = 2`, we won't take value `1` as end is exlusive.

evenyag:
I still need some time to review test cases.

src/query/src/window_sort.rs
```diff
@@ -0,0 +1,3101 @@
...
+            .map(|i| Timestamp::new(i.value() + 1, i.unit()));
+        start.zip(end).map(|(s, e)| TimeRange::new(s, e))
+    }
+}
+
+/// find all successive runs in the input iterator
+fn find_successive_runs<T: Iterator<Item = (usize, Option<N>)>, N: Ord + Copy>(
+    iter: T,
+    sort_opts: &Option<SortOptions>,
+) -> Vec<SucRun<N>> {
```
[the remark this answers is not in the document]
discord9: > We can assume the time index column is always non-null. This should simplify this function. Then we can use the `values()` to get a slice of i64. https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveArray.html#method.values

This function is not complex anyway, and I want to make it more useful in that it can also accept null column in case future changes are needed here and we forget this invariant assumption. Also it's a lot of unnecessary chore to modify all unit tests for it

src/query/src/window_sort.rs
discord9: > if `min_val = 1, array = [3, 2, 1, 0], end = 2`, we won't take value `1` as end is exlusive.

min_val is inclusive as it's get the mininal value from PartitionRange

src/query/src/window_sort.rs
```diff
@@ -0,0 +1,3101 @@
...
+            } else {
+                "Input `PartitionRange`s's lower bound is not monotonic non-decrease"
+            };
+            let plain_error = PlainError::new(msg.to_string(), StatusCode::Unexpected);
+            return Err(BoxedError::new(plain_error)).context(QueryExecutionSnafu {});
+        }
+
+        let properties = PlanProperties::new(
+            input.equivalence_properties().clone(),
+            Partitioning::UnknownPartitioning(1),
```
waynexia, line 106: This plan won't collapse partitions.

```suggestion
            input.output_partitioning().clone(),
```

src/query/src/window_sort.rs
```diff
@@ -0,0 +1,3101 @@
...
+    pub fn to_stream(
+        &self,
+        context: Arc<TaskContext>,
+        partition: usize,
+    ) -> datafusion_common::Result<DfSendableRecordBatchStream> {
+        if 0 != partition {
+            return Err(DataFusionError::Internal(format!(
+                "WindowedSortExec invalid partition {partition}"
+            )));
+        }
```
waynexia, line 135: According to the previous change.

```suggestion
```

src/query/src/window_sort.rs
```diff
@@ -0,0 +1,3105 @@
...
+            // test for off by one case
+            (
+                Arc::new(TimestampMillisecondArray::from_iter_values([1, 2, 3, 4, 5])) as ArrayRef,
+                false,
+                TimeRange {
+                    start: Timestamp::new_millisecond(2),
+                    end: Timestamp::new_millisecond(4),
+                },
+                Ok((1, 2)),
+            ),
```
evenyag, line 2063: Do we need to test start <= 1? e.g. `[0, 4)` and `[1, 4)`.

src/query/src/window_sort.rs
```diff
@@ -0,0 +1,3105 @@
...
+                ranges,
+                Arc::new(mock_input),
+            )
+            .unwrap();
+
+            let exec_stream = exec.execute(0, Arc::new(TaskContext::default())).unwrap();
+
+            let real_output = exec_stream.collect::<Vec<_>>().await;
+            let real_output: Vec<_> = real_output.into_iter().try_collect().unwrap();
+            // let concat_batch = concat_batches(&self.output.schema(), &real_output).unwrap();
```
evenyag, line 2305:
```suggestion
```

evenyag (approved):

waynexia (approved):
Let's merge this first

PR #4820, opened 2024-10-11
fix: error in admin function is not formatted properly

waynexia:

Prior to this change, the error is something like
```
MySQL [(none)]> admin flush_table(phy);
ERROR 1210 (HY000): (InvalidArguments): Failed to build admin function args: unsupported function arg {arg}
```

which doesn't format `arg` in error message.

## Checklist


WenyXu (approved):
Interesting

killme2008 (approved):

PR #4827, opened 2024-10-12
feat: cache logical region's metadata

waynexia:

Cache logical region's metadata to accelerate query.

Given some metrics are likely never to be used, this cache is loaded lazily.

## Checklist

### DIFF
--- a/src/metric-engine/src/engine/region_metadata.rs
+++ b/src/metric-engine/src/engine/region_metadata.rs
@@ -23,13 +23,25 @@ use crate::error::Result;
 impl MetricEngineInner {
     /// Load column metadata of a logical region.
     ///
-    /// The return value is ordered on [ColumnId].
+    /// The return value is ordered on column name.
     pub async fn load_logical_columns(
         &self,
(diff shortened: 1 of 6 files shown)


fengjiachun (approved):
LGTM

src/metric-engine/src/engine/region_metadata.rs
```diff
@@ -23,25 +23,74 @@ use crate::error::Result;
...
-
-        // sort columns on column id to ensure the order
+        // Sort columns on column name to ensure the order
         logical_column_metadata
             .sort_unstable_by(|c1, c2| c1.column_schema.name.cmp(&c2.column_schema.name));
+        // Update cache
+        self.state
+            .write()
+            .unwrap()
+            .add_logical_columns(logical_region_id, logical_column_metadata.clone());
```
evenyag, line 59: Is it possible that another thread altered the logical regions and invalidated the cache before this step? Then we will cache outdated columns.

We may need another lock to avoid altering the region while reading columns.

evenyag (approved):

PR #4835, opened 2024-10-15
feat: introduce the `PluginOptions`

WenyXu:

introduce the `PluginOptions`

## Checklist

### DIFF
--- a/src/cmd/src/datanode.rs
+++ b/src/cmd/src/datanode.rs
@@ -272,9 +272,10 @@ impl StartCommand {
         info!("Datanode start command: {:#?}", self);
         info!("Datanode options: {:#?}", opts);
 
+        let plugin_opts = opts.plugins;
         let opts = opts.component;
         let mut plugins = Plugins::new();
-        plugins::setup_datanode_plugins(&mut plugins, &opts)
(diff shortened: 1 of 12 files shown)


fengjiachun (approved):
LGTM

waynexia:
Can we add something like `PluginConfig` in the config struct? It may look like
```toml
[[plugin.a]]
a1 = "xxx"
a2 = "xxx"

[[plugin.b]]
b1 = "xxx"
b2 = "xxx"
```

So every plugin can register and parse their options programly. And we won't duplicate the parsing step of config structs twice.

src/cmd/src/datanode.rs
```diff
@@ -272,9 +272,10 @@ impl StartCommand {
         info!("Datanode start command: {:#?}", self);
         info!("Datanode options: {:#?}", opts);
 
+        let plugin_opts = opts.plugins;
         let opts = opts.component;
         let mut plugins = Plugins::new();
-        plugins::setup_datanode_plugins(&mut plugins, &opts)
+        plugins::setup_datanode_plugins(&mut plugins, &opts, &plugin_opts)
```
fengjiachun, line 278: The args order suggestion:
`(&mut plugins, &plugins_opts, &common_opts)`

waynexia (approved):

Issue #1084, opened 2023-02-26
Alter table options


killme2008:
### What type of enhancement is this?

API improvement, Configuration

### What does the enhancement do?

For example, modify the table's `TTL` to 30 days:
```sql
ALTER TABLE monitor MODIFY TTL '30d';
```
Not all of the table options can be modified. The `regions` option can't be changed right now. Maybe we can manage a white list that contains alterable options.

### Implementation challenges

How to make the change takes effect dynamically? I am not sure. I think we can implement a observer pattern for it.

killme2008:
After discussing and investigating, decide to use the following syntax:

```sql
alter table set options(ttl='30d', write_buffer_size='32mb');
```

The syntax for options is the same as that for creating tables. The other keywords in the `options` place can be used to set other table features.

NiwakaDev:
I guess that we don't allow changing `storage` option.

killme2008:
> I guess that we don't allow changing `storage` option.

Yes, of course.  I've done some preliminary work in #210 .

chienguo:
@killme2008 Hi! why did you stop this work? I am also interested in this issue, and I also noticed the discussion under  https://github.com/GreptimeTeam/greptimedb/issues/4394.  Seems that alter table options and alter database options can be done parallelly. Can I have a try on this issue?

evenyag:
Feel free to try it. Could you please share your design and steps for implementing this feature?

chienguo:
here is my rough design:
- [ ] add `AlterKind::ChangeOptions`, record the option names that are about to change and their corresponding value
- [ ] for all the places that deal with AlterKind, add support for `ChangeOptions`, like `alter_expr_to_request`, `create_proto_alter_kind` to create region proto, `builder_with_alter_kind` to construct a `TableMetaBuilder` actually to set table options for the upper `AlterTableProcedure` to use
- [ ] for the table options white list, for now, I can only think of doing it in a hardcode way
- [ ] modify parser to support input like this `alter table set options(ttl='30d', write_buffer_size='32mb');`

And where the proto file to generate `greptime.v1.meta.rs` locates? I cannot find it

evenyag:
> And where the proto file to generate `greptime.v1.meta.rs` locates? I cannot find it

Are you looking for this?
https://github.com/GreptimeTeam/greptime-proto/tree/main/proto/greptime/v1

What's more, we also need to support altering options in the region engine and update the option of the region.

[...]

PR #4848, opened 2024-10-18
feat: Support altering table TTL

v0y4g3r:

Supporting altering table options.

The syntax would be:
```sql
ALTER TABLE <table_name> SET ['<option_name>'='option_value'];
```

This also changes the syntax of altering column data types:
```
-- this no longer works
ALTER TABLE <table> <column_name> <new_type>;

-- COLUMN keyword is required
ALTER TABLE <table> COLUMN <column_name> <new_type>;
```
But since the [doc](https://docs.greptime.com/reference/sql/alter#modify-column-type) does require a `COLUMN` keyword, so it will not be a breaking change.

## Checklist


tests/cases/standalone/common/alter/alter_table_options.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE ato(i INTEGER, j TIMESTAMP TIME INDEX);
+
+ALTER TABLE ato MODIFY 'ttl'='1d';
+
+SHOW CREATE TABLE ato;
+
+ALTER TABLE ato MODIFY 'ttl'='2d';
```
killme2008, line 7: Looks like it's not the syntax we discussed before.https://github.com/GreptimeTeam/greptimedb/issues/1084

v0y4g3r: > Looks like it's not the syntax we discussed before.#1084

In Ansi-SQL, SET keyword is often used as setting session-wise variables. Databases like Clickhouse use `MODIFY` keyword to change TTL parameters.

killme2008: I got it, but I prefer `MODIFY OPTIONS ttl='2d'` like that, it represents that you want to modify the option just like `MODIFY COLUMN`, which means you want to modify the column options.

v0y4g3r: > I got it, but I prefer `MODIFY OPTIONS ttl='2d'` like that, it represents that you want to modify the option just like `MODIFY COLUMN`, which means you want to modify the column options.

OK. That will also make parser kind of easier.

v0y4g3r: fixed in f89b532dd1

tests/cases/standalone/common/alter/alter_table_options.result
MichaelScofield, line 1: Suggest adding a case for `ALTER TABLE ato MODIFY OPTIONS 'ttl'='0d';`

MichaelScofield (approved):

waynexia:
I'd prefer the `SET` way for two reasons:
- It keeps the same grammar as PG
- `SET` has a corresponding `UNSET` which is convenient for removing a custom value and resetting it to the default option

src/mito2/src/region/opener.rs
```diff
@@ -313,6 +313,10 @@ impl RegionOpener {
         let manifest = manifest_manager.manifest();
         let metadata = manifest.metadata.clone();
 
+        if let Some(ttl) = manifest.ttl {
+            region_options.ttl = Some(ttl);
+        }
```
WenyXu, line 318: It's better to store the ttl in the table metadata instead of the manifest. cc @v0y4g3r

src/mito2/src/manifest/action.rs
```diff
@@ -45,6 +45,8 @@ pub enum RegionMetaAction {
 pub struct RegionChange {
     /// The metadata after changed.
     pub metadata: RegionMetadataRef,
+    #[serde(default, with = "humantime_serde")]
+    pub ttl: Option<Duration>,
```
evenyag, line 49: We don't need to persist this as we already store table options in meta server and we only need to update the in memory state of the region.

src/sql/src/parsers/alter_parser.rs
```diff
@@ -74,13 +74,44 @@ impl<'a> ParserContext<'a> {
                 )));
             }
         } else if self.consume_token("MODIFY") {
-            let _ = self.parser.parse_keyword(Keyword::COLUMN);
-            let column_name = Self::canonicalize_identifier(self.parser.parse_identifier(false)?);
-            let target_type = self.parser.parse_data_type()?;
+            // Modify column type
```
evenyag, line 77: Can we use https://docs.rs/sqlparser/latest/sqlparser/parser/struct.Parser.html#method.parse_options_with_keywords

This is how sql parser parses this statement.
https://docs.rs/sqlparser/latest/src/sqlparser/parser/mod.rs.html#6961-6967

We may need to slightly change the syntax. If we only need to parse options, maybe we can refer to the implementation of this method
https://docs.rs/sqlparser/latest/src/sqlparser/parser/mod.rs.html#6412-6419

Use `parse_comma_separated()`
https://docs.rs/sqlparser/latest/sqlparser/parser/struct.Parser.html#method.parse_comma_separated

src/store-api/src/region_request.rs
```diff
@@ -389,6 +392,8 @@ pub enum AlterKind {
         /// Columns to change.
         columns: Vec<ChangeColumnType>,
     },
+    /// Change table options.
+    ChangeTableOptions { options: Vec<ChangeTableOption> },
```
evenyag, line 396: This is region level.
```suggestion
    /// Change region options.
    ChangeRegionOptions { options: Vec<ChangeRegionOption> },
```

src/table/src/metadata.rs
```diff
@@ -208,9 +210,34 @@ impl TableMeta {
...
             }
+            AlterKind::ChangeTableOptions { options } => self.change_table_attrs(options),
         }
     }
 
+    /// Creates a [TableMetaBuilder] with modified table options.
+    fn change_table_attrs(
+        &self,
+        requests: &[ChangeTableOptionRequest],
+    ) -> Result<TableMetaBuilder> {
```
evenyag, line 221: Can we rename to `change_table_options`? We didn't mention attrs.
```suggestion
    fn change_table_options(
        &self,
        requests: &[ChangeTableOptionRequest],
    ) -> Result<TableMetaBuilder> {
```

src/table/src/requests.rs
```diff
@@ -212,6 +214,35 @@ pub enum AlterKind {
...
+}
+
+impl TryFrom<&ChangeTableOption> for ChangeTableOptionRequest {
+    type Error = Error;
+
+    fn try_from(value: &ChangeTableOption) -> std::result::Result<Self, Self::Error> {
+        let ChangeTableOption { key, value } = value;
+        if key == TTL_KEY {
+            let ttl = if value.is_empty() {
+                None
```
evenyag, line 234: Do we need to return `duration(0)` to indicate this?

src/mito2/src/manifest/action.rs
v0y4g3r: @evenyag @WenyXu This approach is not viable.
When datanode starts, it only scans all regions it owns, which is a set of `DatanodeTableValue`.
https://github.com/GreptimeTeam/greptimedb/blob/6e776d5f988c54ce8481064c5179c387985ec6cd/src/datanode/src/datanode.rs#L213

The name `DatanodeTableValue` is misleading, in fact, it only contains the region-related info:

https://github.com/GreptimeTeam/greptimedb/blob/e328c7067ce95f042b273030dde6875b643934e7/src/common/meta/src/key/datanode_table.rs#L117-L123

So if you need table metadata here, you can only do another lookup according to the table id, for every region.

So simply store the option in table metadata does not work, we have two other approaches:
- A: Write the TTL to region metadata in the alter procedure
- B: Write the TTL to some place that datanode can access during startup, which is manifest

For A, we need to introduce another step to alter region metadata, and I can't see clear benefit.

src/mito2/src/region/opener.rs
v0y4g3r: Discuss about this [here](https://github.com/GreptimeTeam/greptimedb/pull/4848#discussion_r1812311692).

src/table/src/metadata.rs
v0y4g3r: > Can we rename to `change_table_options`? We didn't mention attrs.

fixed in https://github.com/GreptimeTeam/greptimedb/commit/dce487e66e70eb38617be606b3fcbbb74b809111#diff-839e23ed322c90eacc42fdddeb1f7108335e6434eba0a7a1c6dbc42052a8b9e9R396

src/store-api/src/region_request.rs
v0y4g3r: > ChangeTableOptions { options: Vec<Chan

fixed in https://github.com/GreptimeTeam/greptimedb/commit/dce487e66e70eb38617be606b3fcbbb74b809111#diff-839e23ed322c90eacc42fdddeb1f7108335e6434eba0a7a1c6dbc42052a8b9e9R396

tests/cases/standalone/common/alter/alter_table_options.result
v0y4g3r: fixed in https://github.com/GreptimeTeam/greptimedb/commit/dce487e66e70eb38617be606b3fcbbb74b809111#diff-d5bf576e62305e940175a21bc0f1cdfd4f5ddf8ca10161f161f4f7c286324cfbR66

v0y4g3r:
> // Modify column type



> I'd prefer the `SET` way for two reasons:
> 
> * It keeps the same grammar as PG
> * `SET` has a corresponding `UNSET` which is convenient for removing a custom value and resetting it to the default option

With PG syntax we will need to surround parameters with parentheses, which is verbose and errorprone and I can't see the necessity. We are neither PG or PG-based database after all.

src/sql/src/parsers/alter_parser.rs
v0y4g3r: fixed in 08a3b30da7

src/mito2/src/manifest/action.rs
evenyag: RegionInfo contains a field `region_options`. @WenyXu Does it contain all the table options we need?

v0y4g3r: The proposed changes are included in https://github.com/GreptimeTeam/greptimedb/pull/4848/commits/9df3a4a6218ad03e9e13c65df8674de282484773 and https://github.com/GreptimeTeam/greptimedb/pull/4848/commits/1f8297f6ec9706bbfb60bdba41b1e79cd9516827

src/common/meta/src/key.rs
```diff
@@ -926,22 +924,35 @@ impl TableMetadataManager {
...
-
         let new_table_info_value = current_table_info_value.update(new_table_info);
 
         // Updates table info.
         let (update_table_info_txn, on_update_table_info_failure) = self
             .table_info_manager()
             .build_update_txn(table_id, current_table_info_value, &new_table_info_value)?;
 
-        let mut r = self.kv_backend.txn(update_table_info_txn).await?;
+        let txn = if let Some(region_distribution) = region_distribution {
```
WenyXu, line 943: make sure `region_distribution.len() != 0` is true

WenyXu (commented):
Rest LGTM

WenyXu (approved):

waynexia:
> > // Modify column type
> 
> > I'd prefer the `SET` way for two reasons:
> > 
> > * It keeps the same grammar as PG
> > * `SET` has a corresponding `UNSET` which is convenient for removing a custom value and resetting it to the default option
> 
> With PG syntax we will need to surround parameters with parentheses, which is verbose and errorprone and I can't see the necessity. But this does make parser simpler, so I changed it to the non-parentheses version [08a3b30](https://github.com/GreptimeTeam/greptimedb/commit/08a3b30da70f78a82d9d79d1c25e5856cefd67ca)

How about supporting both? For the sake of reducing unnecessary incompatibility.

src/table/src/requests.rs
```diff
@@ -212,6 +214,33 @@ pub enum AlterKind {
...
         new_table_name: String,
     },
+    ChangeTableOptions {
+        options: Vec<ChangeTableOptionRequest>,
+    },
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum ChangeTableOptionRequest {
+    TTL(Duration),
```
evenyag, line 224:
```suggestion
    Ttl(Duration),
```

src/store-api/src/region_request.rs
```diff
@@ -639,6 +685,11 @@ impl From<v1::ChangeColumnType> for ChangeColumnType {
     }
 }
 
+#[derive(Debug, Eq, PartialEq, Clone)]
+pub enum ChangeRegionOption {
+    TTL(Option<Duration>),
```
evenyag, line 690:
```suggestion
    Ttl(Option<Duration>),
```

src/mito2/src/worker/handle_alter.rs
```diff
@@ -120,11 +140,31 @@ impl<S> RegionWorkerLoop<S> {
...
+        region: MitoRegionRef,
+        version: VersionRef,
+        options: Vec<ChangeRegionOption>,
+        sender: OptionOutputTx,
+    ) {
+        let mut current_options = version.options.clone();
+        for option in options {
+            match option {
+                ChangeRegionOption::TTL(new_ttl) => {
+                    info!("update ttl to {:?}", new_ttl);
```
evenyag, line 160: We should also log the region id.

src/store-api/src/region_request.rs
evenyag, line 690: Maybe we should document whether the duration should be zero.

src/store-api/src/region_request.rs
```diff
@@ -473,12 +484,47 @@ impl TryFrom<alter_request::Kind> for AlterKind {
...
 
+impl TryFrom<&v1::ChangeTableOption> for ChangeRegionOption {
+    type Error = MetadataError;
+
+    fn try_from(option: &v1::ChangeTableOption) -> std::result::Result<Self, Self::Error> {
+        if option.key == TTL_KEY {
+            let ttl_value = if option.value.is_empty() {
+                None
+            } else {
+                let d = humantime::parse_duration(&option.value).map_err(|_| {
```
evenyag, line 508: Should we check whether `d` is zero and return None?

src/table/src/requests.rs
```diff
@@ -212,6 +214,33 @@ pub enum AlterKind {
     RenameTable {
         new_table_name: String,
     },
+    ChangeTableOptions {
+        options: Vec<ChangeTableOptionRequest>,
+    },
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum ChangeTableOptionRequest {
```
evenyag, line 223: We have a similar struct `ChangeRegionOption`. They have the same fields. Can we reuse the `ChangeRegionOption` struct?

evenyag (commented):
The two option enum variants are a little confusing.
```
TTL(Duration),

TTL(Option<Duration>),
```

Others LGTM

src/common/meta/src/key.rs
v0y4g3r: Checked here https://github.com/GreptimeTeam/greptimedb/blob/3a492fecf82d31e00d340f3baf5bd6683e5ac731/src/common/meta/src/key/datanode_table.rs#L223

src/table/src/requests.rs
v0y4g3r: > 

I prefer `TTL`, it's not a word, just an abbreviation.

src/mito2/src/worker/handle_alter.rs
v0y4g3r: fixed in https://github.com/GreptimeTeam/greptimedb/pull/4848/commits/56f4be4c66137553ad3261fd1163e44d9939bdf7

src/store-api/src/region_request.rs
v0y4g3r: We now only use zero duration to indicate unset fixed in https://github.com/GreptimeTeam/greptimedb/pull/4848/commits/56f4be4c66137553ad3261fd1163e44d9939bdf7

src/table/src/requests.rs
v0y4g3r: fixed in fixed in https://github.com/GreptimeTeam/greptimedb/pull/4848/commits/56f4be4c66137553ad3261fd1163e44d9939bdf7

src/table/src/requests.rs
evenyag: Well, in Rust we usually treat it as a word in naming, like `Arc`. But this is not a blocking issue.

src/table/src/requests.rs
```diff
@@ -212,8 +213,35 @@ pub enum AlterKind {
...
+//                 humantime::parse_duration(value)
+//                     .map_err(|_| error::InvalidTableOptionValueSnafu { key, value }.build())?
+//             };
+//             Ok(Self::TTL(ttl))
+//         } else {
+//             UnsupportedTableOptionChangeSnafu { key }.fail()
+//         }
+//     }
+// }
+
```
evenyag, line 244: We should remove them.

src/sql/src/statements/alter.rs
```diff
@@ -97,6 +100,27 @@ impl Display for AlterTableOperation {
             } => {
                 write!(f, r#"MODIFY COLUMN {column_name} {target_type}"#)
             }
+            AlterTableOperation::ChangeTableOptions { options } => {
+                for ChangeTableOption { key, value } in options {
+                    write!(f, r#"MODIFY '{key}'='{value}', "#)?;
```
killme2008, line 105: Should be 

```rust
write!(f, r#"SET '{key}'='{value}',
```

PR #4855, opened 2024-10-18
chore: better column schema check for flow

discord9:

adjust flow's column check so that `update_at` column is only appended when possible, and give a better msg when column is mismatched

## Checklist

### DIFF
--- a/tests/cases/standalone/common/flow/flow_basic.result
+++ b/tests/cases/standalone/common/flow/flow_basic.result
@@ -740,8 +740,8 @@ CREATE TABLE ngx_distribution (
     bucket_size INT,
     total_logs BIGINT,
     time_window TIMESTAMP TIME INDEX,
-    update_at TIMESTAMP,
     -- auto generated column by flow engine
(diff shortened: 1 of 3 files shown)


evenyag (approved):

tests/cases/standalone/common/flow/flow_basic.result
```diff
@@ -739,7 +740,8 @@ CREATE TABLE ngx_distribution (
     bucket_size INT,
     total_logs BIGINT,
     time_window TIMESTAMP TIME INDEX,
-    update_at TIMESTAMP, -- auto generated column by flow engine
+    update_at TIMESTAMP,
+    -- auto generated column by flow engine
```
waynexia, line 744:
```suggestion
    -- auto generated column by flow engine
    update_at TIMESTAMP,
```

waynexia:
> adjust flow's column check so that update_at column is only appended when possible

Do you think it's necessary to describe the "when possible" in document? To my understanding this PR is actually deprecating the `update_at` information.

PR #4859, opened 2024-10-21
feat: Add functionality to the Opentelemetry write interface to extract fields from attr to top-level data. 

paomian:

* It is convenient to extract part of the data in attr to the top level for indexing data.
* If there are fields with the same name in different attr's, the log scope resource is selected in descending order of priority.
* The selected field defaults to tag
* If the selected field is not a simple field type (int string bool), an error is reported.

## Checklist


src/servers/src/otlp/logs.rs
```diff
@@ -355,35 +358,343 @@ fn build_identity_row(
...
+                    let column_schema = &schema_info.schema[index];
+                    if column_schema.datatype != schema.datatype {
+                        return IncompatibleSchemaSnafu {
+                            column_name: k.clone(),
+                            datatype: column_schema.datatype().as_str_name(),
+                            expected: column_schema.datatype,
+                            actual: schema.datatype,
+                        }
+                        .fail();
+                    }
```
shuiyisong, line 387:
```suggestion
                    ensure!(
                        column_schema.datatype == schema.datatype,
                        IncompatibleSchemaSnafu {
                            column_name: k.clone(),
                            datatype: column_schema.datatype().as_str_name(),
                            expected: column_schema.datatype,
                            actual: schema.datatype,
                        }
                    );
```

src/servers/src/otlp/logs.rs
```diff
@@ -355,35 +358,343 @@ fn build_identity_row(
...
+        let mut final_extra_values = vec![GreptimeValue { value_data: None }; extra_schema_len];
+        for (idx, (_, (attr_type, index, _))) in final_extra_schema_info.iter().enumerate() {
+            let value = match attr_type {
+                OpenTelemetryLogRecordAttrType::Resource => resource_values.get_mut(*index),
+                OpenTelemetryLogRecordAttrType::Scope => scope_values.get_mut(*index),
+                OpenTelemetryLogRecordAttrType::Log => log_values.get_mut(*index),
+            };
+            if let Some(value) = value {
+                mem::swap(&mut final_extra_values[idx], value);
+                // final_extra_values[*index] = value.clone();
```
shuiyisong, line 568:
```suggestion
```

src/servers/src/otlp/logs.rs
```diff
@@ -355,35 +358,343 @@ fn build_identity_row(
...
+            extract_field_from_attr_and_combine_schema(extra_log_schema, select_info, &log_attr)?;
+
+        let parse_info = ParseInfo {
+            values: row,
+            resource_values: resource_values.clone(),
+            scope_values: scope_values.clone(),
+            log_values,
+        };
+        result.push(parse_info);
+        // result.push(row);
```
shuiyisong, line 688:
```suggestion
```

src/servers/src/otlp/logs.rs
```diff
@@ -355,35 +358,343 @@ fn build_identity_row(
...
+                (
+                    OpenTelemetryLogRecordAttrType::Log,
+                    *log_schema_index,
+                    column_schema,
+                ),
+            );
+        }
+    }
+    schemas
+}
```
shuiyisong, line 520: By changing the invocation to the following
```Rust
    let final_extra_schema_info = merge_schema(vec![
        (
            &extra_resource_schema,
            OpenTelemetryLogRecordAttrType::Resource,
        ),
        (&extra_scope_schema, OpenTelemetryLogRecordAttrType::Scope),
        (&extra_log_schema, OpenTelemetryLogRecordAttrType::Log),
    ]);
```
We can reduce the duplicated code.

```suggestion
fn merge_schema(
    input_schemas: Vec<(&SchemaInfo, OpenTelemetryLogRecordAttrType)>,
) -> BTreeMap<&String, (OpenTelemetryLogRecordAttrType, usize, &ColumnSchema)> {
    let mut schemas = BTreeMap::new();
    input_schemas
        .into_iter()
        .for_each(|(schema_info, attr_type)| {
            for (key, index) in schema_info.index.iter() {
                if let Some(col_schema) = schema_info.schema.get(*index) {
                    schemas.insert(key, (attr_type, *index, col_schema));
                }
            }
        });

    schemas
}
```

src/servers/src/otlp/logs.rs
```diff
@@ -355,35 +358,343 @@ fn build_identity_row(
...
+
+fn parse_export_logs_service_request_to_rows(
+    request: ExportLogsServiceRequest,
+    select_info: Box<SelectInfo>,
+) -> Result<Rows> {
+    let mut schemas = build_otlp_logs_identity_schema();
+    let mut extra_resource_schema = SchemaInfo::default();
+    let mut extra_scope_schema = SchemaInfo::default();
+    let mut extra_log_schema = SchemaInfo::default();
```
shuiyisong, line 529: On second thought, can we use one `SchemaInfo` to collect resource, scope and log attributes in the same place?

shuiyisong (approved):
Rest LGTM

src/servers/src/otlp/logs.rs
paomian: It might work, but the logic might be a little more complicated.

shuiyisong: We can open a separate PR to refactor this logic later.

paomian: I can think of a situation that is not easy to deal with, that is, resource attr and log attr have the same name field but different types, and sharing a schema info will result in a type mismatch error when the first log attr is processed.

shuiyisong: All three schema lists will eventually be merged in `merge_schema` before the value is appended to the row, so even if they have the same name it's not going to matter. Also we don't differentiate selector keys with prefixes like `resource.key_name` or `scope.key_name`.

paomian: The field with the same name, can not determine whether it is in other types of attr or its own type of attr, other types of attr in the field with the same name type can be overridden, but if it is in its own type of attr need to report an error, this may not be good to deal with.

sunng87:
We will need to update document for this feature.

src/servers/src/otlp/logs.rs
sunng87: Yes, I suggest to use a single SchemaInfo to reduce allocate. The later ones should always override previous. Also I'm not sure if we can always use `String` for tag types.

paomian: I think we can merge this pr first and follow up with a detailed discussion about merging schema info.

src/servers/src/otlp/logs.rs
```diff
@@ -355,35 +358,321 @@ fn build_identity_row(
...
+                None,
+            ))),
+            JsonbNumber::Float64(f) => Ok(Some((
+                GreptimeValue {
+                    value_data: Some(ValueData::F64Value(f)),
+                },
+                ColumnDataType::Float64,
+                SemanticType::Tag,
+                None,
+            ))),
```
evenyag, line 433: Maybe we should set it as field if it is float as a float tag is usually meaningless.

src/servers/src/otlp/logs.rs
```diff
@@ -355,35 +358,321 @@ fn build_identity_row(
...
+        JsonbValue::Bool(b) => Ok(Some((
+            GreptimeValue {
+                value_data: Some(ValueData::BoolValue(b)),
+            },
+            ColumnDataType::Boolean,
+            SemanticType::Field,
+            None,
+        ))),
+        JsonbValue::Array(_) | JsonbValue::Object(_) => {
+            OpenTelemetryNestValueIsComplicatedSnafu { key: column_name }.fail()
```
sunng87, line 452: Consider an alternative name for this error like: `UnsupportedJsonDataTypeForTag(column_name, data type)`

src/servers/src/otlp/logs.rs
sunng87: We only extract columns for tag, so here we can move float type to unsupported data type

src/servers/src/otlp/logs.rs
```diff
@@ -355,35 +358,321 @@ fn build_identity_row(
...
+                SemanticType::Tag,
+                None,
+            ))),
+        },
+        JsonbValue::Bool(b) => Ok(Some((
+            GreptimeValue {
+                value_data: Some(ValueData::BoolValue(b)),
+            },
+            ColumnDataType::Boolean,
+            SemanticType::Field,
```
sunng87, line 448: As we are extracting columns for tag, I guess this is a typo?

sunng87 (approved):

sunng87 (approved):

PR #4862, opened 2024-10-22
test: add fuzz test for metric region migration

WenyXu:

Add fuzz test for metric region migration

## Checklist

### DIFF
--- a/tests-fuzz/targets/migration/fuzz_migrate_metric_regions.rs
+++ b/tests-fuzz/targets/migration/fuzz_migrate_metric_regions.rs
@@ -1,0 +1,447 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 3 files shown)


fengjiachun (approved):
LGTM

tests-fuzz/targets/migration/fuzz_migrate_metric_regions.rs
```diff
@@ -0,0 +1,495 @@
...
+        .collect::<Vec<_>>();
+    info!("List datanodes: {:?}", datanodes);
+
+    // Generates region migration task.
+    let mut migrations = Vec::with_capacity(num_partitions);
+    let mut new_distribution: BTreeMap<u64, HashSet<_>> = BTreeMap::new();
+    for (datanode_id, regions) in region_distribution {
+        let step = rng.gen_range(1..datanodes.len());
+        for region in regions {
+            let to_peer = (datanode_id + step as u64) % datanodes.len() as u64;
```
evenyag, line 253: Is it possible that the `to_peer` is the same as the current datanode id?

tests-fuzz/targets/migration/fuzz_migrate_metric_regions.rs
```diff
@@ -0,0 +1,495 @@
...
+            |partition| {
+                info!("Region: {region_id},  datanode: {}", partition.datanode_id);
+                partition.datanode_id == migration.to_peer
+            },
+            Duration::from_secs(1),
+        )
+        .await;
+    }
+
+    // Creates more logical tables and inserts values
```
evenyag, line 415: Why do we need to repeat this so many times?

tests-fuzz/targets/migration/fuzz_migrate_metric_regions.rs
WenyXu: Nope

evenyag (approved):

PR #4807, opened 2024-10-09
feat: json_path_exists udf

CookiePieWw:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

#4805 

## What's changed and what's your intention?

As title.

## Checklist

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [x] This PR requires documentation updates.

Issue #4805, opened 2024-10-09
[Tracking issue] Add more JSON UDFs


WenyXu:
### What problem does the new feature solve?

The previous tracking issue #4230 introduced some basic JSON UDFs. This tracking issue aims to expand on that by adding more advanced UDFs, enabling more powerful querying capabilities.

### What does the feature do?

- [ ] [JSON_OBJECT_KEYS](https://docs.databend.com/sql/sql-functions/semi-structured-functions/json-object-keys) Returns an Array containing the list of keys in the input JSON Object.
- [ ] [JSON_PATH_EXISTS](https://docs.databend.com/sql/sql-functions/semi-structured-functions/json-path-exists) Checks whether a specified path exists in JSON data.
- [ ] [JSON_PATH_MATCH](https://docs.databend.com/sql/sql-functions/semi-structured-functions/json-path-match) Checks whether a specified JSON path expression matches certain conditions within a JSON data.
...

sunng87:
reference https://www.depesz.com/2024/10/11/sql-json-is-here-kinda-waiting-for-pg-17/

qazxcdswe123:
Hi @WenyXu , I came across this issue and I woner if `JSON_OBJECT_KEYS` is still relavant? (do we still want it in greptimedb?)

The previous attempt https://github.com/GreptimeTeam/greptimedb/pull/5382 was closed because MySQL `Unsupported data type:
    List<String>`, but in #6935 #7096 the datatype get introduced.

I think the issue is still relavant as https://github.com/GreptimeTeam/greptimedb/issues/7685 said we need more json support 😆 

If it's still relavant. I can make a PR on this

MichaelScofield:
@qazxcdswe123 it's still relavant, feel free to make a PR!

qazxcdswe123:
Hi @MichaelScofield, when working on this, I found a potential performance improvement。

For now we are forced to read and process the entire jsonb input (i.e: `parse_jsonb` that materialize both key and value), but databends' approach use `RawJsonb::object_keys()`, which only reads in keys, and this could be 9x perf diff from my benchmark, but it's added in a later version of `jsonb`

| Scenario | Time (full) | Time (key only) | Slowdown |
|---|---|---|---|
| 10 keys | ~2134 ns | ~294 ns | ~7.3× |
| 100 keys | ~23313 ns | ~2640 ns | ~8.8× |
[...]

PR #4864, opened 2024-10-22
feat: add json_path_match udf

Kev1n8:

part of #4805.

- Referring the implementation of #4807
- Added relevant sql tests

## Checklist


src/common/function/src/scalars/json/json_path_match.rs
```diff
@@ -0,0 +1,184 @@
...
+    fn return_type(&self, _input_types: &[ConcreteDataType]) -> Result<ConcreteDataType> {
+        Ok(ConcreteDataType::boolean_datatype())
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::exact(
+            vec![
+                ConcreteDataType::json_datatype(),
+                ConcreteDataType::string_datatype(),
+            ],
```
Kev1n8, line 48: I wonder what should the UDF behaves when given a vector of json with both null and non-null values. For example, I could do 
```
Signature::one_of(
            vec![
                TypeSignature::Exact(vec![
                    ConcreteDataType::json_datatype(),
                    ConcreteDataType::string_datatype(),
                ]),
                TypeSignature::Exact(vec![
                    ConcreteDataType::null_datatype(),
                    ConcreteDataType::string_datatype(),
                ]),
                TypeSignature::Exact(vec![
                    ConcreteDataType::json_datatype(),
                    ConcreteDataType::null_datatype(),
                ]),
                TypeSignature::Exact(vec![
                    ConcreteDataType::null_datatype(),
                    ConcreteDataType::null_datatype(),
                ]),
            ],
```
in the signature. But in `eval()`, the `dataype` retrieved by `jsons.data_type()` indicates the whole column should be the same type, which causing me fail to recognise which rows of json in the column is null(because it's all `Binay`), I'm not sure how to deal with it.

tests/cases/standalone/common/function/json/json.sql
```diff
@@ -6,3 +6,19 @@ SELECT json_path_exists(parse_json('{"a": 1, "b": 2}'), '$.c');
 SELECT json_path_exists(parse_json('[1, 2]'), '[0]');
 
 SELECT json_path_exists(parse_json('[1, 2]'), '[2]');
+
+SELECT json_path_exists(parse_json('[1, 2]'), NULL);
+
+-- SELECT json_path_exists(NULL, '$.a');
```
Kev1n8, line 12: I would like to fix [this](#4865) in a follow-on PR.

src/common/function/src/scalars/json/json_path_match.rs
```diff
@@ -0,0 +1,222 @@
...
+        Ok(ConcreteDataType::boolean_datatype())
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::one_of(
+            vec![
+                TypeSignature::Exact(vec![
+                    ConcreteDataType::json_datatype(),
+                    ConcreteDataType::string_datatype(),
+                ]),
```
killme2008, line 49: This sub signature should be enough for datafusion, it will take care of null type by default.

killme2008 (approved):
Good job!

src/common/function/src/scalars/json/json_path_match.rs
Kev1n8: I'm not entirely certain. It might be helpful to check issue #4865, where I noticed that the function seems unable to handle null JSON with this sub signature.

tests/cases/standalone/common/function/json/json.sql
WenyXu: The JSON null is not the same as the null in a datatype system. 
```sql
SELECT json_path_exists(parse_json('null'), '$.a');
```

Kev1n8: > The JSON null is not the same as the null in a datatype system.
> `SELECT json_path_exists(parse_json('null'), '$.a');` 

Thx for pointing that out. I wonder how it should behave in two different cases?
1. Given case like `json_path_exists(NULL, '$.a')`, leave it unsupported or return NULL (as https://docs.databend.com/sql/sql-functions/semi-structured-functions/json-path-exists)?
2. When given the json string is "null", return NULL?

Please let me know the preferred behaviour so that I can fix potential bias.

WenyXu: let's throw a unsupported error

src/common/function/src/scalars/json/json_path_match.rs
```diff
@@ -0,0 +1,192 @@
...
+            let path = paths.get_ref(i);
+
+            match json.data_type() {
+                // JSON data type uses binary vector
+                ConcreteDataType::Binary(_) => {
+                    let json = json.as_binary();
+                    let path = path.as_string();
+                    let result = match (json, path) {
+                        (Ok(Some(json)), Ok(Some(path))) => {
+                            if jsonb::is_object(json) {
```
WenyXu, line 80: It should work for JSON arrays too.

Kev1n8: oops, fixed.

tests/cases/standalone/common/function/json/json.result
```diff
@@ -31,3 +31,60 @@ SELECT json_path_exists(parse_json('[1, 2]'), '[2]');
...
+| false                                                     |
++-----------------------------------------------------------+
+
+SELECT json_path_exists(parse_json('null'), '$.a');
+
++--------------------------------------------------------+
+| json_path_exists(parse_json(Utf8("null")),Utf8("$.a")) |
++--------------------------------------------------------+
+| false                                                  |
++--------------------------------------------------------+
```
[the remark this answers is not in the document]
WenyXu: What's the behavior in another system? seems it should be expected.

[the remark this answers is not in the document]
Kev1n8: I was referring to [this](https://docs.databend.com/sql/sql-functions/semi-structured-functions/json-path-exists). It seems if one of the parameters is null, the result should be null.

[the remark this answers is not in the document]
WenyXu: Good catch!

WenyXu (approved):
LGTM

PR #4875, opened 2024-10-24
refactor: make use of the "pre_execute" in sql execution interceptor

MichaelScofield:

as title

## Checklist


src/cmd/src/options.rs
```diff
@@ -44,6 +47,64 @@ pub struct GreptimeOptions<T> {
...
+    /// ```
+    /// (Note that the `type` must equals to the name of the plugin option's definition struct.)
+    ///
+    /// Finally, after loading the GreptimeDB config file, we can `downcast` the
+    /// [`Box<dyn PluginOptions>`] to its real type:
+    ///
+    /// ```no_run
+    /// let options = greptime_options.plugins.remove("my-plugin1").unwrap();
+    /// let plugin1_options: MyPlugin1Option = options.as_any().downcast_ref().unwrap();
+    /// assert_eq(plugin1_options, "foo");
```
sunng87, line 105:
```suggestion
    /// assert_eq(plugin1_options.a, "foo");
```

src/cmd/src/standalone.rs
```diff
@@ -445,11 +447,13 @@ impl StartCommand {
...
 
         let mut plugins = Plugins::new();
+
+        let mut plugin_options = opts.plugins;
         let opts = opts.component;
         let fe_opts = opts.frontend_options();
         let dn_opts = opts.datanode_options();
 
-        plugins::setup_frontend_plugins(&mut plugins, &fe_opts)
+        plugins::setup_frontend_plugins(&mut plugins, &fe_opts, &mut plugin_options)
```
sunng87, line 456: since `plugin_options` is for plugin setup and the data is from `fe_opts`, can we make it a internal variable of `setup_frontend_plugins`?  Will we still be using it after this `setup_frontend_plugins`?

src/cmd/src/options.rs
sunng87: Can we include a test for this? I'm curious how this works. Judging from the API, it doesn't have a central registry for all types. Maybe typetag maintains a global state internally? But rust doesn't have reflection so how doesn't access those type by, just macros. The macro will no access to the type's fully qualified name...

MichaelScofield: It does have a central registry for all the trait implementations, check this out: https://github.com/dtolnay/typetag?tab=readme-ov-file#so-many-questions

Since there's alreay a plugin option implementation, I decide to discard this one.

src/cmd/src/standalone.rs
MichaelScofield: these codes are no longer used

MichaelScofield:
@sunng87 PTAL

sunng87 (approved):

fengjiachun (approved):
LGTM

PR #4877, opened 2024-10-24
refactor: change trace/span id to field by default

sunng87:

Just treat traceid/spanid as fields.

We should do the same thing for trace but we can defer it for our refactor on trace impl

## Checklist


shuiyisong:
Since it's a very small change, I'm merging it with https://github.com/GreptimeTeam/greptimedb/pull/4876

PR #4881, opened 2024-10-25
fix: json_path_exists null results

### LINKED ISSUE #4865: json_path_exists can't deal with null json
### What type of bug is this?

Incorrect result

Standalone mode, Frontend

### Minimal reproduce step

The recently added `json_path_exists` behaves wrongly when json param is null.
For example:
```
SELECT json_path_exists(parse_json('null'), '$.a')
SELECT json_paht_exists(parse_json('{}'),'null')

OUTPUT:
false
false
```

According to [JSON_PATH_MATCH](https://docs.databend.com/sql/sql-functions/semi-structured-functions/json-path-match), the result should be null.

### What did you see instead?

Error.

macOS 15.0.1

0.9.5

_No response_

Kev1n8:

Closes #4865.

As explained in referred issue, the result should be `null` if any of the parameters is `null` .

## Checklist


WenyXu (approved):

WenyXu:
Thanks @Kev1n8 ❤️

src/common/function/src/scalars/json/json_path_exists.rs
```diff
@@ -64,38 +64,40 @@ impl Function for JsonPathExistsFunction {
...
-            ConcreteDataType::Binary(_) => {
-                for i in 0..size {
-                    let json = jsons.get_ref(i);
-                    let path = paths.get_ref(i);
+        for i in 0..size {
+            let json = jsons.get_ref(i);
+            let path = paths.get_ref(i);
 
+            match json.data_type() {
+                ConcreteDataType::Binary(_) => {
```
evenyag, line 74: We should use `jsons.data_type()` because the data type of `ValueRef::Null` isn't `Binary`. But we can match `json` and handle `ValueRef::Binary` and `ValueRef::Null`.

tests/cases/standalone/common/function/json/json.result
```diff
@@ -36,15 +36,15 @@ SELECT json_path_exists(parse_json('[1, 2]'), 'null');
 +-----------------------------------------------------------+
 | json_path_exists(parse_json(Utf8("[1, 2]")),Utf8("null")) |
```
evenyag, line 37: Consider adding a test for NULL value. (Not json "null")

tests/cases/standalone/common/function/json/json.result
```diff
@@ -36,15 +36,15 @@ SELECT json_path_exists(parse_json('[1, 2]'), 'null');
...
 +-----------------------------------------------------------+
 
 SELECT json_path_exists(parse_json('null'), '$.a');
 
 +--------------------------------------------------------+
 | json_path_exists(parse_json(Utf8("null")),Utf8("$.a")) |
 +--------------------------------------------------------+
-| false                                                  |
+|                                                        |
 +--------------------------------------------------------+
```
CookiePieWw, line 48: I'm not quite sure about this. "null" is a valid json string and "$.a" is a valid path. For me, it should return false since the path does not exist in the json here.

Kev1n8: I thought so too. @WenyXu and I had a discusstion in another [PR](https://github.com/GreptimeTeam/greptimedb/pull/4864#discussion_r1812018609) on this, maybe we need to delve more into this.

WenyXu: > I thought so too. @WenyXu and I had a discusstion in another [PR](https://github.com/GreptimeTeam/greptimedb/pull/4864#discussion_r1812018609) on this, maybe we need to delve more into this.

I checked the [docs of the databend](https://docs.databend.com/sql/sql-functions/semi-structured-functions/json-path-exists).

```
SELECT JSON_PATH_EXISTS(NULL, '$.a');

----
NULL
```

The test should use the `NULL`  instead of `parse_json("null")`

src/common/function/src/scalars/json/json_path_exists.rs
Kev1n8: I see your point, but I’m curious about how we would handle cases where some items in the array are `ValueRef::Null` while others are not. Or, all items' datatype are supposed to be the same?

Kev1n8:
To be clear, I'd like to specify the expected results of all SQLs below:
```
SELECT json_path_exists(NULL, '$.a');
NULL

SELECT json_path_exists(parse_json('{}'), NULL);
NULL

SELECT json_path_exists(parse_json('null'), '$.a');
false

SELECT json_path_exists(parse_json('{}'), 'null');
false
```

Please let me know if anything is incorrect.

WenyXu:
> To be clear, I'd like to specify the expected results of all SQLs below:
> 
> ```
> SELECT json_path_exists(NULL, '$.a');
> NULL
> 
> SELECT json_path_exists(parse_json('{}'), NULL);
> NULL
> 
> SELECT json_path_exists(parse_json('null'), '$.a');
> false
> 
> SELECT json_path_exists(parse_json('{}'), 'null');
> false
> ```
> 
> Please let me know if anything is incorrect.

looks good

src/common/function/src/scalars/json/json_path_exists.rs
```diff
@@ -72,25 +86,39 @@ impl Function for JsonPathExistsFunction {
...
+                        // unwrap it without checking if it's `Null`.
+                        let path = if let Ok(Some(str)) = path.as_string() {
+                            str
+                        } else {
+                            return InvalidFuncArgsSnafu {
+                                err_msg: format!("Illegal json path: {:?}", path),
+                            }
+                            .fail();
+                        };
+                        let path = jsonb::jsonpath::parse_json_path(path.as_bytes()).unwrap();
```
WenyXu, line 111: we need to handle the error here

Kev1n8: Fixed.

tests/cases/standalone/common/function/json/json.result
Kev1n8: Added.

src/common/function/src/scalars/json/json_path_exists.rs
Kev1n8: Okay, I'll update the code.

src/common/function/src/scalars/json/json_path_exists.rs
```diff
@@ -64,32 +78,61 @@ impl Function for JsonPathExistsFunction {
...
                     };
 
                     results.push(result);
                 }
             }
+
+            // Any null args existence causes the result to be NULL.
+            (ConcreteDataType::Null(_), ConcreteDataType::String(_)) => results.push_nulls(size),
+            (ConcreteDataType::Binary(_), ConcreteDataType::Null(_)) => results.push_nulls(size),
+            (ConcreteDataType::Null(_), ConcreteDataType::Null(_)) => results.push_nulls(size),
```
WenyXu, line 134: According to the sqlness tests, this should be the only change that needs to be done?

Kev1n8: There are 2 major changes: function signatures and datatype cases.

1. I believe function signature changes matter for accepting null type. I've tested if, with only `Exact(vec![Json, String])`, system will throw error when the signature does not match.
2. Apart from "this" change, I think the error handlers also make sense, while the original simply push a `None` into results when there's a parsing or casting error.

src/common/function/src/scalars/json/json_path_exists.rs
```diff
@@ -64,32 +78,61 @@ impl Function for JsonPathExistsFunction {
...
+                                err_msg: format!("Illegal json path: {:?}", path),
+                            }
+                            .fail();
+                        };
+
+                        jsonb::path_exists(json, path).ok()
+                    } else {
+                        // Should never reach here since we `match` in advance.
+                        None
                     };
```
WenyXu, line 125: nit: can we simplify this logic

Kev1n8: Sure. I tried simplifying, please take a look. Let me know if there's anything more.

src/common/function/src/scalars/json/json_path_exists.rs
```diff
@@ -64,32 +78,63 @@ impl Function for JsonPathExistsFunction {
...
+                    let result = if !path.is_null() {
+                        // Get path as str.
+                        let path_str = match path.as_string() {
+                            Ok(Some(str)) => str,
+                            _ => {
+                                return InvalidFuncArgsSnafu {
+                                    err_msg: format!("Unable to parse path as string: {:?}", path),
+                                }
+                                .fail();
+                            }
```
WenyXu, line 109: I don't think we need to handle this solution. The path must be a string.

src/common/function/src/scalars/json/json_path_exists.rs
```diff
@@ -64,32 +78,63 @@ impl Function for JsonPathExistsFunction {
...
-                                Ok(json_path) => jsonb::path_exists(json, json_path).ok(),
-                                Err(_) => None,
+                    // Get json as bytes.
+                    let json = match json.as_binary() {
+                        Ok(Some(bytes)) => bytes,
+                        _ => {
+                            return InvalidFuncArgsSnafu {
+                                err_msg: format!("Illegal json binary: {:?}", json),
                             }
+                            .fail();
```
WenyXu, line 96: ditto

src/common/function/src/scalars/json/json_path_exists.rs
WenyXu: cc @CookiePieWw

CookiePieWw: We can simply put a `None` here as other udf does.
https://github.com/GreptimeTeam/greptimedb/blob/191755fc42f45ecc3628c55427e7e5abbecd2ede/src/common/function/src/scalars/date/date_format.rs#L94-L103

WenyXu: Yes, I prefer the original logic.

Kev1n8:
Reduce some unnecessary checks. Please check out.

src/common/function/src/scalars/json/json_path_exists.rs
```diff
@@ -64,32 +78,39 @@ impl Function for JsonPathExistsFunction {
...
-                                Ok(json_path) => jsonb::path_exists(json, json_path).ok(),
-                                Err(_) => None,
-                            }
+                            // Get `JsonPath`.
+                            let json_path = match jsonb::jsonpath::parse_json_path(path.as_bytes())
+                            {
+                                Ok(json_path) => json_path,
+                                Err(_) => {
+                                    return InvalidFuncArgsSnafu {
+                                        err_msg: format!("Illegal json path: {:?}", path),
```
CookiePieWw, line 95: We can add a test for invalid json path.

CookiePieWw (approved):
Rest LGTM

WenyXu:
Thanks! ❤️

Kev1n8:
Thanks for reviewing! ❤️

PR #4887, opened 2024-10-27
feat(index): support building inverted index for the field column on Mito

zhongzc:

Respect field columns when building the index. See the `do_update` function in `src/mito2/src/sst/index/inverted_index/creator.rs`

## Checklist

### DIFF
--- a/src/mito2/src/read/scan_region.rs
+++ b/src/mito2/src/read/scan_region.rs
@@ -393,20 +393,29 @@ impl ScanRegion {
             .and_then(|c| c.index_cache())
             .cloned();
 
+        // TODO(zhongzc): currently we only index tag columns, need to support field columns.
+        let ignore_column_ids = &self
+            .version
+            .options
+            .index_options
+            .inverted_index
+            .ignore_column_ids;
+        let indexed_column_ids = self
+            .version
+            .metadata
+            .primary_key
+            .iter()
+            .filter(|id| !ignore_column_ids.contains(id))
+            .copied()
+            .collect::<HashSet<_>>();
+
         InvertedIndexApplierBuilder::new(
             self.access_layer.region_dir().to_string(),
             self.access_layer.object_store().clone(),
             file_cache,
             index_cache,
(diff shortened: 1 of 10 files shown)


src/mito2/src/read/scan_region.rs
```diff
@@ -393,20 +393,29 @@ impl ScanRegion {
             .and_then(|c| c.index_cache())
             .cloned();
 
+        // TODO(zhongzc): currently we only index tag columns, need to support field columns.
```
waynexia, line 396: This comment doesn't seem to match the PR title 🤔

src/mito2/src/sst/index/inverted_index/creator.rs
```diff
@@ -210,6 +203,32 @@ impl InvertedIndexer {
...
         }
 
+        for field in batch.fields() {
+            if !self.indexed_column_ids.contains(&field.column_id) {
+                continue;
+            }
+
+            let sort_field = SortField::new(field.data.data_type());
+            let col_id_str = field.column_id.to_string();
+            for i in 0..n {
```
waynexia, line 213: I'm a little worried about the overhead for this loop 😢

waynexia (approved):
LGTM in general

src/mito2/src/read/scan_region.rs
zhongzc: It's not completely assembled, but this is the key entrance and everything inside is ready.

src/mito2/src/sst/index/inverted_index/creator.rs
zhongzc: which part? `to_string` or encoding?

v0y4g3r: > I'm a little worried about the overhead for this loop 😢

Looks like this loop lieterally iterates and builds index for each row?

v0y4g3r (approved):
LGTM

PR #4890, opened 2024-10-28
feat: show flow's mem usage in INFORMATION_SCHEMA.FLOWS

discord9:

Add a column to show flow's mem usage in system table, so one can tune flow's ttl option as needed.

in distributed mode Flownode's heartbeat task will reguarly ask for a `StateReport` which include all flows' mem usage data, and send them to metasrv, then metasrv can use this info to build system table, however there is a small latency in data update since heartbeat duration is in seconds but this usually is ok since flow's mem usage grow very slowly usually. In standalone mode can just query flow's mem size

## Checklist


killme2008:
What's the status of this PR?

discord9:
I need to refactor this pr to make it more similiar to `CollectStatHandler` to make it more clean, convert it to draft for now

src/catalog/src/system_schema/information_schema/flows.rs
```diff
@@ -195,6 +212,12 @@ impl InformationSchemaFlowsBuilder {
             .flow_names(&catalog_name)
             .await;
 
+        let flow_stat = {
+            let information_extension =
+                utils::information_extension(&self.catalog_manager).unwrap();
+            information_extension.flow_stats().await?.clone()
```
fengjiachun, line 218: Why is cloning needed? Can't we just move?

src/cmd/src/standalone.rs
```diff
@@ -532,6 +533,14 @@ impl StartCommand {
                 .context(OtherSnafu)?,
         );
 
+        // set the ref to query for the local flow state
+        {
+            let flow_worker_manager = flownode.flow_worker_manager();
+            information_extension
+                .set_flow_worker_manager(flow_worker_manager.clone())
+                .await;
```
fengjiachun, line 541: Same as above

src/catalog/src/system_schema/information_schema/flows.rs
```diff
@@ -195,6 +212,12 @@ impl InformationSchemaFlowsBuilder {
             .flow_names(&catalog_name)
             .await;
 
+        let flow_stat = {
+            let information_extension =
+                utils::information_extension(&self.catalog_manager).unwrap();
```
killme2008, line 217: Is it safe to `unwrap`?

src/cmd/src/standalone.rs
killme2008: Why not move it up?  Before the creation of `catalog_manager`. And you don't need to clone the `information_extension` at all.

src/cmd/src/standalone.rs
```diff
@@ -750,6 +767,18 @@ impl InformationExtension for StandaloneInformationExtension {
...
+    async fn flow_stats(&self) -> std::result::Result<Option<FlowStat>, Self::Error> {
+        let state = self
+            .flow_worker_manager
+            .read()
+            .await
+            .as_ref()
+            .unwrap()
+            .gen_state_report()
+            .await;
+        Ok(Some(state))
```
killme2008, line 780:
```suggestion
        self
            .flow_worker_manager
            .read()
            .await
            .as_ref()
            .unwrap()
            .gen_state_report()
            .await
```

src/flow/src/utils.rs
```diff
@@ -18,16 +18,73 @@ use std::collections::{BTreeMap, BTreeSet};
...
+    /// Query the size report, will timeout after one second if no response
+    pub async fn query(&self) -> crate::Result<FlowStat> {
+        let (tx, rx) = oneshot::channel();
+        self.inner.send(tx).await.map_err(|_| {
+            InternalSnafu {
+                reason: "failed to send size report request due to receiver dropped",
+            }
+            .build()
+        })?;
+        let timeout = tokio::time::timeout(std::time::Duration::from_secs(1), rx);
```
killme2008, line 67: Why choose 1 second?

src/cmd/src/standalone.rs
discord9: > Why not move it up? Before the creation of `catalog_manager`. And you don't need to clone the `information_extension` at all.

because `flownode` need `catalog_manager` to translate sql and resolve table source, so there is a ugly cycle here(`a->b` means b need a)...:
```
information_extension -> catalog_manager -> flow_builder -> flownode -> flow_worker_manager -set-> information_extension
```
will later make flownode can set `catalog_manager` later in a separate PR so other struct's init can be easier

src/flow/src/utils.rs
discord9: > Why choose 1 second?

changed to half of the heartbeat duration now, just in case

fengjiachun (approved):
Rest LGTM

PR #4909, opened 2024-10-30
feat: adds the number of rows and index files size to region_statistics table

### LINKED ISSUE #4896: Adds index data size to region_statistics table

User experience

After this PR https://github.com/GreptimeTeam/greptimedb/pull/4771, we already have the table SST files and memtables data size etc, but missing the index data size.

I think we should add it too. @WenyXu @zhongzc 

Not sure if we have this info in the index service.

### Implementation challenges

_No response_

killme2008:

Close #4896 

Adds new columns to `information_schema.region_statistics`:

* `disk_size`: the approximate disk size in the region.
* `region_rows`:  the number of rows in the region.
* `index_size`: The SST index files size in the region.

And split out the index file size from `sst_size` of `RegionStat`.

## Checklist


waynexia (approved):

tests/cases/standalone/common/information_schema/region_statistics.sql
```diff
@@ -0,0 +1,25 @@
...
+);
+
+
+INSERT INTO test VALUES
+       (1, 'a', 1),
+       (11, 'b', 11),
+       (21, 'c', 21);
+
+-- SQLNESS SLEEP 11s
+-- FIXME(dennis): we need to wait the datanode reporting stats info to metasrv.
```
killme2008, line 20: @WenyXu It's advisable to adjust the heartbeat frequency for region stats information; it's currently set to 10 by default, but we can reduce it for faster test execution.

zhongzc (approved):

Issue #4828, opened 2024-10-14
Scan data by time range


evenyag:
### What type of enhancement is this?

Performance, Refactor

### What does the enhancement do?

A region partitions data by time ranges. The region can scan data by time range so it only reads a small number of files each time. This can reduce the latency since downloading files from all ranges may take longer.

### Implementation challenges

We need to postpone the pruning phase and reimplement scanners. `UnorderedScan` and `SeqScan` have different requirements for ranges.

evenyag:
The only remaining issue is `Support field pruning in last_non_null mode if the time range only has one file.`

We can fix this if this becomes a bottleneck. Now I think we can close this issue.

PR #4913, opened 2024-10-30
fix: prune batches from memtable by time range

evenyag:

https://github.com/GreptimeTeam/greptimedb/issues/4828

This PR fixes that memtable may return rows out of the expected time range. The memtable updates the min/max stats so it can yield rows out of the range when there are in-progress rows during query.

It adds a `PruneTimeIterator` to filter batches by time range.

## Checklist


waynexia:
This case seems hard to test in UT, could you add some description near the related code?

discord9 (approved):
LGTM

v0y4g3r (approved):
LGTM

PR #4915, opened 2024-10-31
fix: database option not reflected in region options

### LINKED ISSUE #4914: Database options do not take effect
### What type of bug is this?

Incorrect result

Other

### Minimal reproduce step

1. Create a database with ttl
```bash
mysql -h 127.0.0.1 -P 4002 -e "create database benchmark with (ttl='1d');"
```

2. Create a table in that database
```bash
mysql -h 127.0.0.1 -P 4002 benchmark -e "CREATE TABLE IF NOT EXISTS cpu ( hostname STRING NULL, region STRING NULL, datacenter STRING NULL, rack STRING NULL, os STRING NULL, arch STRING NULL, team STRING NULL, service STRING NULL, service_version STRING NULL, service_environment STRING NULL, usage_user BIGINT NULL, usage_system BIGINT NULL, usage_idle BIGINT NULL, usage_nice BIGINT NULL, usage_iowait BIGINT NULL, usage_irq BIGINT NULL, usage_softirq BIGINT NULL, usage_steal BIGINT NULL, usage_guest BIGINT NULL, usage_guest_nice BIGINT NULL, ts TIMESTAMP(9) NOT NULL, TIME INDEX (ts), PRIMARY KEY (hostname, region, datacenter, rack, os, arch, team, service, service_version, service_environment));"
```

The table (and regions in that table) created should inherit the database `ttl` option

### What did you see instead?

The ttl option does not take effect in region level.

NA

0.9.5

In #2341 we introduced procedure framework for fault-tolerant DDL tasks like creating tables. create table procedure are split into several steps, namely creating table metadata and invoking datanodes to create regions. 

At the [first step](https://github.com/GreptimeTeam/greptimedb/blob/fd8eba36a8d03d7ce7b9566ebc17220d728f9fd5/src/operator/src/statement/ddl.rs#L1313), table metadata inherits the database level options. However, the merge result is not reflected in `CreateTableExpr` itself. So at the second step, which is invoking datanodes to create regions, the `CreateTableExpr` does not carry any database-level options:

https://github.com/GreptimeTeam/greptimedb/blob/6a634f8e5d870e7dd9000eb68ce30a936d09fb9c/src/common/meta/src/ddl/create_table.rs#L170

So the created regions do not contain schema-level options.

This bug is elusive in that you cannot find that schema options is not applied to regions since all table metadata are correct, so `SHOW CREATE TABLE` statement can give the correct result.

v0y4g3r:

- fixes #4914 

## Checklist

### DIFF
--- a/src/operator/src/statement/ddl.rs
+++ b/src/operator/src/statement/ddl.rs
@@ -59,6 +59,7 @@ use sql::statements::sql_value_to_value;
 use sql::statements::statement::Statement;
 use sqlparser::ast::{Expr, Ident, UnaryOperator, Value as ParserValue};
 use store_api::metric_engine_consts::{LOGICAL_TABLE_METADATA_KEY, METRIC_ENGINE_NAME};
+use store_api::mito_engine_options::TTL_KEY;
 use substrait::{DFLogicalSubstraitConvertor, SubstraitPlan};
 use table::dist_table::DistTable;
 use table::metadata::{self, RawTableInfo, RawTableMeta, TableId, TableInfo, TableType};
@@ -238,6 +239,9 @@ impl StatementExecutor {
             &create_table.schema_name,
             &create_table.table_name,
         );
+
+        // fill schema options in create table expr;
+        fill_schema_opts(create_table, &schema_opts);
 
         let (partitions, partition_cols) = parse_partitions(create_table, partitions, &query_ctx)?;
         let mut table_info = create_table_info(create_table, partition_cols, schema_opts)?;
@@ -1221,6 +1225,13 @@ impl StatementExecutor {
             .submit_ddl_task(&ExecutorContext::default(), request)
             .await
             .context(error::ExecuteDdlSnafu)
+    }
+}
+
+/// Fills schema-level options to [CreateTableExpr].
+fn fill_schema_opts(expr: &mut CreateTableExpr, schema: &SchemaNameValue) {
+    if let Some(ttl) = schema.human_ttl() {
+        expr.table_options.insert(TTL_KEY.to_string(), ttl);
     }
 }
 
--- a/src/operator/src/statement/ddl.rs
+++ b/src/operator/src/statement/ddl.rs
@@ -1230,7 +1230,9 @@ impl StatementExecutor {
 
 /// Fills schema-level options to [CreateTableExpr].
 fn fill_schema_opts(expr: &mut CreateTableExpr, schema: &SchemaNameValue) {
-    if let Some(ttl) = schema.human_ttl() {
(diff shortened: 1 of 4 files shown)


WenyXu (approved):

evenyag (approved):

src/operator/src/statement/ddl.rs
```diff
@@ -1224,6 +1228,13 @@ impl StatementExecutor {
     }
 }
 
+/// Fills schema-level options to [CreateTableExpr].
+fn fill_schema_opts(expr: &mut CreateTableExpr, schema: &SchemaNameValue) {
+    if let Some(ttl) = schema.human_ttl() {
+        expr.table_options.insert(TTL_KEY.to_string(), ttl);
```
killme2008, line 1234: Will a ttl key in `expr.table_options` be overridden if it already exists?

src/operator/src/statement/ddl.rs
```diff
@@ -239,6 +240,9 @@ impl StatementExecutor {
             &create_table.table_name,
         );
 
+        // fill schema options in create table expr;
+        fill_schema_opts(create_table, &schema_opts);
```
fengjiachun, line 244: Question: How should we respond if the `alter database ttll` operation is performed and the tables already exist?

src/operator/src/statement/ddl.rs
v0y4g3r: > Will a ttl key in `expr.table_options` be overridden if it already exists?

This has been addressed in d033ae8bbb

src/operator/src/statement/ddl.rs
v0y4g3r: > Question: How should we respond if the `alter database ttll` operation is performed and the tables already exist?

This PR does not break current behavior, it just fix the problem that database ttl is not applied to regions (neither in memory and region options in metadata).

Altering database options is another story and won't be covered in this PR.

src/operator/src/statement/ddl.rs
```diff
@@ -1224,6 +1228,15 @@ impl StatementExecutor {
     }
 }
 
+/// Fills schema-level options to [CreateTableExpr].
+fn fill_schema_opts(expr: &mut CreateTableExpr, schema: &SchemaNameValue) {
+    if !expr.table_options.contains_key(TTL_KEY)
+        && let Some(ttl) = schema.human_ttl()
+    {
+        expr.table_options.insert(TTL_KEY.to_string(), ttl);
+    }
```
evenyag, line 1237: I forgot that this will also persist the database ttl in table options.

Is it possible that we add another field `db_options` so we know we won't persist it in table value?

PR #4917, opened 2024-10-31
chore: minor refactor for weighted choose

fengjiachun:

As the title.

## Checklist

### DIFF
--- a/src/meta-srv/src/selector/weighted_choose.rs
+++ b/src/meta-srv/src/selector/weighted_choose.rs
@@ -92,11 +92,6 @@ where
     }
 
     fn choose_multiple(&mut self, amount: usize) -> Result<Vec<Item>> {
-        if amount >= self.items.len() {
-            // fast path
-            return Ok(self.items.iter().map(|item| item.item.clone()).collect());
-        }
-
         Ok(self
             .items
             .choose_multiple_weighted(&mut thread_rng(), amount, |item| item.weight as f64)
--- a/src/meta-srv/src/selector/weighted_choose.rs
+++ b/src/meta-srv/src/selector/weighted_choose.rs
@@ -92,6 +92,11 @@ where
     }
 
     fn choose_multiple(&mut self, amount: usize) -> Result<Vec<Item>> {
+        if amount >= self.items.len() {
+            // fast path
+            return Ok(self.items.iter().map(|item| item.item.clone()).collect());
+        }
+
         Ok(self
             .items
             .choose_multiple_weighted(&mut thread_rng(), amount, |item| item.weight as f64)
--- a/src/meta-srv/src/selector/common.rs
+++ b/src/meta-srv/src/selector/common.rs
@@ -22,7 +22,7 @@ use crate::metasrv::SelectTarget;
 use crate::selector::SelectorOptions;
 
 /// According to the `opts`, choose peers from the `weight_array` through `weighted_choose`.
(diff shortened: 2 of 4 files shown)


src/meta-srv/src/selector/weighted_choose.rs
```diff
@@ -92,6 +92,10 @@ where
     }
 
     fn choose_multiple(&mut self, amount: usize) -> Result<Vec<Item>> {
+        if amount >= self.items.len() {
+            return Ok(self.items.iter().map(|item| item.item.clone()).collect());
+        }
+
```
fengjiachun, line 98: Respect the interface comment: https://github.com/GreptimeTeam/greptimedb/blob/f3a1e699ec358c980cdac2904f6fa99991eae7c2/src/meta-srv/src/selector/weighted_choose.rs#L29

src/meta-srv/src/selector/weighted_choose.rs
```diff
@@ -92,6 +92,10 @@ where
     }
 
     fn choose_multiple(&mut self, amount: usize) -> Result<Vec<Item>> {
+        if amount >= self.items.len() {
+            return Ok(self.items.iter().map(|item| item.item.clone()).collect());
```
fengys1996, line 96: It seems that the items are not actually choosed according to their weights.

fengjiachun: If the `amount` equals the count of items, then the only possibility is to get all the items.

fengjiachun: The method `sample_efraimidis_spirakis` said: 
```
/// Randomly sample exactly `amount` distinct indices from `0..length`, and
/// return them in an arbitrary order (there is no guarantee of shuffling or
/// ordering). The weights are to be provided by the input function `weights`,
/// which will be called once for each index.
```

I think it mean:
The `sample_efraimidis_spirakis` function ensures that each index is selected only once through the algorithm when amount <= length. Therefore, when amount == length, the final result will be all unique indices within the range 0..length without duplicates.

fengjiachun: The method also said:

```
/// This implementation uses `O(length + amount)` space and `O(length)` time
/// if the "nightly" feature is enabled, or `O(length)` space and
/// `O(length + amount * log length)` time otherwise.
```

So I think we can give a fast path here(when the `amount` > items.len).

fengys1996:
The [choose_multiple](https://github.com/GreptimeTeam/greptimedb/blob/main/src/meta-srv/src/selector/weighted_choose.rs#L27-L30) method in the WeightedChoose trait does not clearly state whether duplicate items are allowed to be returned, which may cause some confusion.

src/meta-srv/src/selector/weighted_choose.rs
fengjiachun: The fast path code has already been removed. Perhaps we shouldn't optimize prematurely before any issues arise. cc @fengys1996

WenyXu:
> The [choose_multiple](https://github.com/GreptimeTeam/greptimedb/blob/main/src/meta-srv/src/selector/weighted_choose.rs#L27-L30) method in the WeightedChoose trait does not clearly state whether duplicate items are allowed to be returned, which may cause some confusion. cc @WenyXu

It wouldn't by default, let's add some comments.

WenyXu (approved):

fengys1996 (approved):

PR #4924, opened 2024-11-01
feat: heartbeat_flush_threshold option

### LINKED ISSUE #4911: Allow adjusting the heartbeat frequency for region stats information
@WenyXu It's better to allow adjusting the heartbeat frequency for region stats information; it's currently set to 10 by default, but we can reduce it for faster test execution.

fengjiachun:

Close #4911 

## Checklist

### DIFF
--- a/src/meta-srv/src/handler.rs
+++ b/src/meta-srv/src/handler.rs
@@ -478,8 +478,10 @@ pub struct HeartbeatHandlerGroupBuilder {
     /// The handler to handle region lease.
     region_lease_handler: Option<RegionLeaseHandler>,
 
-    /// The heartbeat flush threshold.
-    heartbeat_flush_threshold: Option<usize>,
+    /// The factor that determines how often statistics should be flushed,
+    /// based on the number of received heartbeats. When the number of heartbeats
+    /// reaches this factor, a flush operation is triggered.
+    flush_stats_factor: Option<usize>,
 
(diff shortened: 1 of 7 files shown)


WenyXu (approved):

src/meta-srv/src/handler.rs
```diff
@@ -478,6 +478,9 @@ pub struct HeartbeatHandlerGroupBuilder {
     /// The handler to handle region lease.
     region_lease_handler: Option<RegionLeaseHandler>,
 
+    /// The heartbeat flush threshold.
+    heartbeat_flush_threshold: Option<usize>,
```
killme2008, line 482: I don't like this name. In factor, it's related to the heartbeat interval. So  I think we can call it `flush_stats_factor`, it looks better. It's not a threshold, but a factor.

killme2008 (approved):
LGTM. But a little suggestion about the variable name.

src/meta-srv/src/handler/collect_stats_handler.rs
```diff
@@ -262,7 +277,7 @@ mod tests {
         let kv = res.unwrap();
         let val: DatanodeStatValue = kv.value.try_into().unwrap();
         // refresh every 10 stats
```
fengjiachun, line 279:
```suggestion
```

Issue #4905, opened 2024-10-30
Support altering table's compaction options


MichaelScofield:
### What type of enhancement is this?

User experience

### What does the enhancement do?

Like altering table's TTL option, compaction options are best able to be modified after table is created.

lyang24:
I think alter compaction option can be implemented in the front end starting with adding a new variant to [ChangeOption](https://github.com/GreptimeTeam/greptimedb/blob/be72d3bedba9202cc3c938fdfde41f591f5c7355/src/store-api/src/region_request.rs#L661C1-L664C2). However, [TWCS](https://github.com/GreptimeTeam/greptimedb/blob/be72d3bedba9202cc3c938fdfde41f591f5c7355/src/mito2/src/region/options.rs#L151C1-L159C2) is the only compaction strategy available. Do we plan to offer some compaction option in the near future?

MichaelScofield:
@lyang24 Right now changing the configs of TWCS is sufficient.

lyang24:
sounds good can i work on this issue?

MichaelScofield:
@lyang24 Sure, assigned to you.

killme2008:
Closed via #4965 
Thank you @lyang24

PR #4939, opened 2024-11-05
feat: support alter twcs compression options

lyang24:
https://github.com/GreptimeTeam/greptimedb/issues/4905

support alter twcs compaction options on sql level

expand alter table set options to include  twcs compaction options.

The approach here is to expand ChangeOption enum with twcs compaction options. 
The key and value type is validated at store-api end. 
If value is missing (empty string) it will be set with default value.
Changes will be reflected on show create table statements.

## Checklist

### DIFF
--- a/src/mito2/src/worker/handle_alter.rs
+++ b/src/mito2/src/worker/handle_alter.rs
@@ -27,6 +27,8 @@ use crate::error::{
 };
 use crate::flush::FlushReason;
 use crate::manifest::action::RegionChange;
+use crate::region::options::CompactionOptions::Twcs;
+use crate::region::options::TwcsOptions;
 use crate::region::version::VersionRef;
 use crate::region::MitoRegionRef;
 use crate::request::{DdlRequest, OptionOutputTx, SenderDdlRequest};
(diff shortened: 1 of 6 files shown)


src/store-api/src/region_request.rs
```diff
@@ -661,23 +663,138 @@ impl From<v1::ChangeColumnType> for ChangeColumnType {
...
+
+impl ChangeOption {
+    pub fn to_change_table_option(&self) -> ChangeTableOption {
+        match self {
+            ChangeOption::TTL(duration) => ChangeTableOption {
+                key: "ttl".to_string(),
+                value: humantime::format_duration(*duration).to_string(),
+            },
+            ChangeOption::TwscMaxActiveWindowRuns(runs) => ChangeTableOption {
+                key: "compaction.twcs.max_active_window_runs".to_string(),
```
v0y4g3r, line 684: Can we extract these string literals to consts? They appear multiple times in codebase.

src/store-api/src/region_request.rs
```diff
@@ -661,23 +663,138 @@ impl From<v1::ChangeColumnType> for ChangeColumnType {
...
+                value: size.map_or(String::new(), |value| value.to_string()),
+            },
+            ChangeOption::TwscTimeWindow(window) => ChangeTableOption {
+                key: "compaction.twcs.time_window".to_string(),
+                value: window.map_or(String::new(), |value| {
+                    humantime::format_duration(value).to_string()
+                }),
+            },
+            ChangeOption::TwscRemoteCompaction(flag) => ChangeTableOption {
+                key: "compaction.twcs.remote_compaction".to_string(),
```
v0y4g3r, line 710: Not sure if this option can be modified dynamically? @zyy

src/store-api/src/region_request.rs
```diff
@@ -661,23 +663,138 @@ impl From<v1::ChangeColumnType> for ChangeColumnType {
...
+                value: flag.map_or(String::new(), |value| value.to_string()),
+            },
+            ChangeOption::TwscFallbackToLocal(flag) => ChangeTableOption {
+                key: "compaction.twcs.fallback_to_local".to_string(),
+                value: flag.map_or(String::new(), |value| value.to_string()),
```
v0y4g3r, line 715: nitpick:
```suggestion
                value: flag.map(|s| s.to_string()).unwrap_or_default(),
```

src/mito2/src/worker/handle_alter.rs
```diff
@@ -167,6 +169,92 @@ impl<S> RegionWorkerLoop<S> {
...
+                ChangeOption::TwscMaxActiveWindowRuns(runs) => {
+                    let Twcs(options) = current_options.compaction;
+                    let runs = runs.unwrap_or(TwcsOptions::default().max_active_window_runs);
+                    info!(
+                        "Update region compaction.twcs.max_active_window_runs: {}, previous: {} new: {}",
+                        region.region_id, options.max_active_window_runs, runs
+                    );
+                    let mut new_option = options.clone();
+                    new_option.max_active_window_runs = runs;
+                    current_options.compaction = Twcs(new_option)
```
v0y4g3r, line 181: Can we avoid the clone here?

```suggestion
                    let Twcs(options) = &mut current_options.compaction;
                    let runs = runs.unwrap_or(TwcsOptions::default().max_active_window_runs);
                    info!(
                        "Update region compaction.twcs.max_active_window_runs: {}, previous: {} new: {}",
                        region.region_id, options.max_active_window_runs, runs
                    );
                    options.max_active_window_runs = runs;
```

src/store-api/src/region_request.rs
lyang24: I see i left this option out in this new update

src/store-api/src/region_request.rs
lyang24: yep its a good idea

src/mito2/src/worker/handle_alter.rs
lyang24: good call fixed it

src/store-api/src/region_request.rs
```diff
@@ -661,23 +667,126 @@ impl From<v1::ChangeColumnType> for ChangeColumnType {
 #[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize)]
 pub enum ChangeOption {
     TTL(Duration),
+    TwscMaxActiveWindowRuns(Option<usize>),
```
killme2008, line 670: What about changing these variants into one 

```Rust
Twcs(String)
```

The `String` value is the option name such as `compaction.twcs.max_active_window_runs` etc.

I think it looks much better and we can add new options in the future without adding new variant of `ChangeOption`.

src/mito2/src/worker/handle_alter.rs
```diff
@@ -167,6 +169,67 @@ impl<S> RegionWorkerLoop<S> {
                         current_options.ttl = Some(new_ttl);
                     }
                 }
+                ChangeOption::TwscMaxActiveWindowRuns(runs) => {
```
killme2008, line 172: If we change all these variants into `Twcs(String)`, then we can refactor these match clauses into:

```rust
ChangeOption::Twcs(options) =>
{
    match options.as_str() => {
       "compaction.twcs.max_active_window_runs" => { .. }
      ......
    }
}
```

src/store-api/src/region_request.rs
lyang24: I think the draw back would be that we cannot validate the type of value at storage api layer - i.e user can pass max_active_window_runs='abcde' and this will error out at mito engine code path

killme2008: > I think the draw back would be that we cannot validate the type of value at storage api layer - i.e user can pass max_active_window_runs='abcde' and this will error out at mito engine code path (where we handle the alter options).

We can match the option to verify its validity, as the next comment explains.

lyang24: ah i see great call

lyang24: I took a try at implement this approach i found it quite difficult below is my approach:
1. we also need the value of the change, the enum would look like Twcs(String, String)
2. the second comments suggest that we match and validate values at Mito side when handling the alter request, if there is an error we have to return a Mito error which will cause a lot of drift of logic on alter tables. (i.e. ttl change values are parsed and validated when building the request and it will return a metadata error - InvalidRegionOptionChangeRequest).

If our goal is to reduce the variant of ChangeOption. 
I propose a nested enum approach - to parse and validate the values when we are building the request to stay consistent with ttl changes.
```
pub enum ChangeOption {
    TTL(Duration),
    Twcs(TwcsChangeOptions),
}

pub enum TwcsChangeOptions {
       MaxActiveWindowFiles(Option<Duration>),
       ...
}
```

lyang24:
the new approach look cleaner going to close this one in favor of 
https://github.com/GreptimeTeam/greptimedb/pull/4965

PR #4955, opened 2024-11-06
fix: do not pick compacting/expired files

evenyag:

This PR fixes that twcs compaction picker may pick expired files to compact.

We already set the `compacting` here:
https://github.com/GreptimeTeam/greptimedb/blob/22a662f6bcb811e2c6e36a2a29d035175f4131df/src/mito2/src/compaction/twcs.rs#L218-L222

## Checklist


waynexia (approved):
LGTM, but where do you check if a file is expired?

v0y4g3r (approved):
LGTM

evenyag:
> LGTM, but where do you check if a file is expired?

Here:
https://github.com/GreptimeTeam/greptimedb/blob/22a662f6bcb811e2c6e36a2a29d035175f4131df/src/mito2/src/compaction/twcs.rs#L216-L217

PR #4962, opened 2024-11-08
feat: memory parquet reader

v0y4g3r:

Enables parquet reader to read from memory files, as pre work for parquet-based memtable.

## Checklist


v0y4g3r:
Due to the lacking of keyword generics and we cannot find a workaround currently, this proposal should be abandoned and we can only reused the reader part code be composition.

PR #4969, opened 2024-11-11
fix: dbg assert may fail

CookiePieWw:

The dbg_assert may fail for:

1. It asserts that the number of params in the sql text is the same as the number of params in the datafusion plan
2. but datafusion may optimize sql like `SELECT ? + 1 FROM 1 = 0` to an empty plan, so they may mismatch.

## Checklist


CookiePieWw:
Seems not only about this `assert`, there's other bugs. I'll come to it later.

image

CookiePieWw:
track in #4970

PR #4972, opened 2024-11-11
fix: alter table add column id alloc mismatch

discord9:

Fix a bug when altering a metric table and add a already existing columns in physical table caused `column_id` mismatch,
which would cause datafusion to panic in certain condition, related sqlness tests are also added

### Bug Explain

The bug is:

1. Two metrics tables have been created that share a single physical table.
2. Table A has added a column named "at", with a column ID of 1.
3. After some time, Table B attempts to add the "at" column:
   - It uses the `column_semantic_type` function to check and finds that it does not have this column, so it adds it to the `columns_to_add` list.
   - It sends an alter request to Mito, specifying the column definition as ("at" String null Tag 2), knowing that the maximum column ID is 1.
   - Mito receives the request and discovers that the "at" column already exists (as seen in `store_api::metadata::RegionMetadataBuilder::add_columns`). It ignores the request, does not add the column to the physical table, and does not return an error.
   - The metrics engine, not receiving an error, assumes the column addition was successful and adds ("at" String null Tag 2) to the logical table.
4. At this point, the column ID for "at" is seen as 1 in both Table A and the physical table, while Table B sees it as 2.

Solution is to check physical region for actual `column_id`(and other metadata) to update to logical region

## Checklist

### DIFF
--- a/src/metric-engine/src/data_region.rs
+++ b/src/metric-engine/src/data_region.rs
@@ -150,6 +150,7 @@ impl DataRegion {
             })
             .collect::<Result<_>>()?;
 
+        info!("Adding (Column id assigned) columns {new_columns:?} to region {region_id:?}");
         // assemble alter request
         let alter_request = RegionRequest::Alter(RegionAlterRequest {
             schema_version: version,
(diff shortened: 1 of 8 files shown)


waynexia (approved):
💯

fengjiachun (approved):
LGTM

PR #4984, opened 2024-11-13
feat: parallel build_file_ranges

discord9:

parallel build_file_ranges, so that reader can read in parallel in happy path

previously here is a `Mutex`, which would make even just read sequentially execute, change to `RwLock` to allow for parallel build ranges

## Checklist

### DIFF
--- a/src/mito2/src/read/scan_region.rs
+++ b/src/mito2/src/read/scan_region.rs
@@ -29,7 +29,7 @@ use smallvec::SmallVec;
 use store_api::region_engine::{PartitionRange, RegionScannerRef};
 use store_api::storage::{ScanRequest, TimeSeriesRowSelector};
 use table::predicate::{build_time_range_predicate, Predicate};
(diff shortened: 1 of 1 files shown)


src/mito2/src/read/scan_region.rs
```diff
@@ -891,13 +891,27 @@ impl RangeBuilderList {
         reader_metrics: &mut ReaderMetrics,
     ) -> Result<()> {
         let file_index = index.index - self.mem_builders.len();
-        let mut builder_opt = self.file_builders[file_index].lock().await;
-        match &mut *builder_opt {
+        let builder_opt = self.file_builders[file_index].read().await;
```
evenyag, line 894: We can remove the global builder list and use per-partition builder list. So a partition won't wait for other partitions. (This happens on UnorderedScan).

I implemented a prototype in https://github.com/GreptimeTeam/greptimedb/commit/3f9bf48161bcc26ace6787894f68c296a87760d7.

PR #4956, opened 2024-11-07
fix: round euclidean result in sqlness

v0y4g3r:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)
fix #4946

## Checklist

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.

PR #4987, opened 2024-11-13
feat: add distance functions

zhongzc:

Added

* `cos_distance`
* `l2sq_distance`
* `dot_prodcut`

to calculate the distance between vectors.

## Checklist


src/common/function/src/scalars/vector/distance.rs
```diff
@@ -0,0 +1,469 @@
...
+    unsafe {
+        let num_floats = bytes.len() / 4;
+        let floats: &[f32] = std::slice::from_raw_parts(bytes.as_ptr() as *const f32, num_floats);
+        Ok(floats)
+    }
+}
+
+/// Parse a string to a vector value.
+/// Valid inputs are strings like "[1.0, 2.0, 3.0]".
+fn parse_vector_from_string(s: &str) -> Result<Vec<f32>> {
```
WenyXu, line 184: I'm unsure whether we should directly support parsing vectors from strings, or if we should instead provide a UDF to handle the string-to-vector parsing.

zhongzc: I'm comfortable with this concise usage unless some strong obstacle changes my mind.

v0y4g3r: > I'm comfortable with this concise usage unless some strong obstacle changes my mind.

maybe rename to `parse_f32_vector_from_string` since it accepts valid f32 values only.

tests/cases/standalone/common/types/vector/vector.result
```diff
@@ -31,6 +31,180 @@ SELECT * FROM t;
...
+
++-------------------------+--------------------------+----------------------+
+| ts                      | v                        | d                    |
++-------------------------+--------------------------+----------------------+
+| 1970-01-01T00:00:00.001 | 0000803f0000004000004040 | 5.960464477539063e-8 |
+| 1970-01-01T00:00:00.002 | 000080400000a0400000c040 | 0.025368154048919678 |
+| 1970-01-01T00:00:00.003 | 0000e0400000004100001041 | 0.04058808088302612  |
++-------------------------+--------------------------+----------------------+
+
+SELECT cos_distance(v, v) FROM t;
```
v0y4g3r, line 74: We may need to round the results to avoid potential sqlness inconsistence on different platforms. see https://github.com/GreptimeTeam/greptimedb/pull/4956

waynexia (approved):
LGTM

WenyXu (approved):
LGTM

PR #4988, opened 2024-11-13
test: more sqlness tests for flow

discord9:

more sqlness tests for flow, include a wired regex to redact flush_flow's results since it's not deterministic

## Checklist


tests/cases/standalone/common/flow/flow_basic.result
```diff
@@ -19,12 +19,11 @@ Affected Rows: 0
 
 -- TODO(discord9): confirm if it's necessary to flush flow here?
 -- because flush_flow result is at most 1
-admin flush_flow('test_numbers_basic');
+-- SQLNESS REPLACE ADMIN\sFLUSH_FLOW\('\w+'\)\s+\|\n\+-+\+\n\|\s+[0-9]+\s+ FLOW_FLUSHED
```
evenyag, line 22: Why do we replace the name of this column?

tests/cases/standalone/common/flow/flow_basic.result
```diff
@@ -916,3 +894,397 @@ DROP TABLE requests;
 
 Affected Rows: 0
 
+-- blog usecase
```
evenyag, line 897: non-blocking: Maybe add two new files for blog and user guide?

zhongzc (approved):

evenyag (approved):

tests/cases/standalone/common/flow/flow_basic.result
discord9: > Why do we replace the name of this column?

the return of flush flow is the output row count of flow which is not determinstic so remove

PR #4994, opened 2024-11-14
refactor: introduce `MaintenanceModeManager`

WenyXu:

introduce `MaintenanceModeManager`

## Checklist

### DIFF
--- a/src/common/meta/src/key/maintenance.rs
+++ b/src/common/meta/src/key/maintenance.rs
@@ -1,0 +1,63 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+use crate::error::Result;
+use crate::key::MAINTENANCE_KEY;
+use crate::kv_backend::KvBackendRef;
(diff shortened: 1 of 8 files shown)


src/meta-srv/src/metasrv/builder.rs
```diff
@@ -195,6 +196,9 @@ impl MetasrvBuilder {
         let flow_metadata_manager = Arc::new(FlowMetadataManager::new(
             leader_cached_kv_backend.clone() as _,
         ));
+        let maintenance_mode_manager = Arc::new(MaintenanceModeManager::new(
+            leader_cached_kv_backend.clone() as _,
```
fengjiachun, line 200: Seems we cannot use `leader_cached_kv_backend` since a follower can also set `maintenance_mode`

src/common/meta/src/key/maintenance.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /// Enables maintenance mode.
+    pub async fn set_maintenance_mode(&self) -> Result<()> {
+        let req = PutRequest {
+            key: Vec::from(MAINTENANCE_KEY),
+            value: vec![],
+            prev_kv: false,
+        };
+        self.kv_backend.put(req.clone()).await?;
```
MichaelScofield, line 44: unnecessary `clone`

src/common/meta/src/key/maintenance.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /// Unsets maintenance mode.
+    pub async fn unset_maintenance_mode(&self) -> Result<()> {
+        let req = PutRequest {
+            key: Vec::from(MAINTENANCE_KEY),
+            value: vec![],
+            prev_kv: false,
+        };
+        self.kv_backend.put(req.clone()).await?;
```
MichaelScofield, line 55: unnecessary `clone`

src/common/meta/src/key/maintenance.rs
WenyXu: Good catch!

MichaelScofield (approved):

PR #4998, opened 2024-11-14
fix: prune memtable/files range independently in each partition

evenyag:

https://github.com/GreptimeTeam/greptimedb/issues/4828

This PR removes the scanner level global lock for range builders. Each partition will create a range builder for a file to prune.
In UnorderedScan, each partition may prune the same file multiple times. But this usually doesn't increase the total query time.

The global lock may have potential issues while the query engine is doing something like streaming merge. Since partitions may wait for each other. If the top partition currently polling is waiting for a lock owned by another partition in the merge heap, then the top partition may never have a chance to get the lock.

## Checklist


evenyag:
@coderabbitai summary

src/mito2/src/compaction/twcs.rs
```diff
@@ -114,7 +114,7 @@ impl TwcsPicker {
                 // Files in window exceeds file num limit
                 vec![enforce_file_num(&files.files, max_files)]
             } else {
-                debug!("Skip building compaction output, active window: {:?}, current window: {}, max runs: {}, found runs: {}, ", active_window, *window, max_runs, found_runs);
+                trace!("Skip building compaction output, active window: {:?}, current window: {}, max runs: {}, found runs: {}, ", active_window, *window, max_runs, found_runs);
```
v0y4g3r, line 117: Do we need to change this log level?

src/mito2/src/read/range.rs
```diff
@@ -334,6 +341,160 @@ fn maybe_split_ranges_for_seq_scan(ranges: Vec<RangeMeta>) -> Vec<RangeMeta> {
...
+        index: RowGroupIndex,
+        reader_metrics: &mut ReaderMetrics,
+    ) -> Result<SmallVec<[FileRange; 2]>> {
+        let mut ranges = SmallVec::new();
+        let file_index = index.index - self.num_memtables;
+        let builder_opt = self.get_file_builder(file_index);
+        match builder_opt {
+            Some(builder) => builder.build_ranges(index.row_group_index, &mut ranges),
+            None => {
+                let builder = input.prune_file(file_index, reader_metrics).await?;
```
v0y4g3r, line 459: Is the key point of this PR is to remove the std::mutex across the await point here so that top node in merge reader will not be affected by some other node silently holding the lock without being polled?

src/mito2/src/compaction/twcs.rs
evenyag: This log is quite verbose while using debug level.

src/mito2/src/read/range.rs
evenyag: Yes. The removed mutex is an async tokio mutex. It's better to avoid a partition waiting for the execution of another partition. It's fine to use mutex inside a partition if it doesn't need to wait for other partitions.

v0y4g3r (approved):
LGTM

PR #5001, opened 2024-11-15
feat: CREATE OR REPLACE FLOW

discord9:

impl replace flow

- impl replace flow on flownode by drop and create on flownode
- impl replace on metasrv by add a new `update_txn` for flowname&flow info value
- change `show create flow`'s output to contain `IF NOT EXISTS` only to prevent accidently clear internal state if need to run

## Checklist

### DIFF
--- a/src/common/meta/src/ddl/create_flow.rs
+++ b/src/common/meta/src/ddl/create_flow.rs
@@ -76,7 +76,7 @@ impl CreateFlowProcedure {
                 source_table_ids: vec![],
                 query_context,
                 state: CreateFlowState::Prepare,
-                flow_already_exists: false,
+                prev_flow_info_value: None,
             },
         }
     }
@@ -135,7 +135,22 @@ impl CreateFlowProcedure {
             self.data.flow_id = Some(flow_id);
             self.data.peers = peers;
             info!("Replacing flow, flow_id: {}", flow_id);
-            self.data.flow_already_exists = true;
+
+            let flow_info_value = self
+                .context
+                .flow_metadata_manager
+                .flow_info_manager()
+                .get(flow_id)
+                .await?;
+
+            ensure!(
+                flow_info_value.is_some(),
+                error::UnexpectedSnafu {
+                    err_msg: format!("Flow info value not found for flow_id: {}", flow_id)
+                }
+            );
+
+            self.data.prev_flow_info_value = flow_info_value;
         }
 
         //  Ensures sink table doesn't exist.
@@ -183,7 +198,10 @@ impl CreateFlowProcedure {
                     .map_err(add_peer_context_if_needed(peer.clone()))
             });
         }
-        info!("Creating flow({}) on flownodes", self.data.flow_id.unwrap());
+        info!(
+            "Creating flow({:?}) on flownodes with peers={:?}",
+            self.data.flow_id, self.data.peers
+        );
         join_all(create_flow)
             .await
             .into_iter()
@@ -201,10 +219,10 @@ impl CreateFlowProcedure {
         // Safety: The flow id must be allocated.
         let flow_id = self.data.flow_id.unwrap();
         let (flow_info, flow_routes) = (&self.data).into();
-        if self.data.flow_already_exists {
+        if let Some(prev_flow_value) = self.data.prev_flow_info_value.as_ref() {
             self.context
                 .flow_metadata_manager
-                .update_flow_metadata(flow_id, flow_info, flow_routes)
+                .update_flow_metadata(flow_id, flow_info, prev_flow_value.clone(), flow_routes)
                 .await?;
             info!("Replaced flow metadata for flow {flow_id}");
         } else {
@@ -312,7 +330,8 @@ pub struct CreateFlowData {
     pub(crate) peers: Vec<Peer>,
     pub(crate) source_table_ids: Vec<TableId>,
     pub(crate) query_context: QueryContext,
-    pub(crate) flow_already_exists: bool,
+    /// For verify if prev value is consistent when need to update flow metadata.
+    pub(crate) prev_flow_info_value: Option<FlowInfoValue>,
 }
 
 impl From<&CreateFlowData> for CreateRequest {
--- a/src/common/meta/src/ddl/create_flow.rs
+++ b/src/common/meta/src/ddl/create_flow.rs
@@ -224,7 +224,7 @@ impl CreateFlowProcedure {
         {
             self.context
                 .flow_metadata_manager
-                .update_flow_metadata(flow_id, &flow_info, prev_flow_value, flow_routes)
+                .update_flow_metadata(flow_id, prev_flow_value, &flow_info, flow_routes)
                 .await?;
             info!("Replaced flow metadata for flow {flow_id}");
         } else {
--- a/src/common/meta/src/ddl/create_flow.rs
+++ b/src/common/meta/src/ddl/create_flow.rs
@@ -44,7 +44,7 @@ use crate::instruction::{CacheIdent, CreateFlow};
 use crate::key::flow::flow_info::FlowInfoValue;
 use crate::key::flow::flow_route::FlowRouteValue;
 use crate::key::table_name::TableNameKey;
-use crate::key::{FlowId, FlowPartitionId};
+use crate::key::{DeserializedValueWithBytes, FlowId, FlowPartitionId};
 use crate::lock_key::{CatalogLock, FlowNameLock, TableNameLock};
 use crate::peer::Peer;
 use crate::rpc::ddl::{CreateFlowTask, QueryContext};
@@ -140,13 +140,13 @@ impl CreateFlowProcedure {
                 .context
                 .flow_metadata_manager
                 .flow_info_manager()
-                .get(flow_id)
+                .get_raw(flow_id)
                 .await?;
 
             ensure!(
                 flow_info_value.is_some(),
-                error::UnexpectedSnafu {
-                    err_msg: format!("Flow info value not found for flow_id: {}", flow_id)
+                error::FlowNotFoundSnafu {
+                    flow_name: format_full_flow_name(catalog_name, flow_name),
                 }
             );
 
@@ -219,10 +219,12 @@ impl CreateFlowProcedure {
         // Safety: The flow id must be allocated.
         let flow_id = self.data.flow_id.unwrap();
         let (flow_info, flow_routes) = (&self.data).into();
-        if let Some(prev_flow_value) = self.data.prev_flow_info_value.as_ref() {
+        if let Some(prev_flow_value) = self.data.prev_flow_info_value.as_ref()
+            && self.data.task.or_replace
+        {
             self.context
                 .flow_metadata_manager
-                .update_flow_metadata(flow_id, flow_info, prev_flow_value.clone(), flow_routes)
+                .update_flow_metadata(flow_id, &flow_info, prev_flow_value, flow_routes)
                 .await?;
             info!("Replaced flow metadata for flow {flow_id}");
         } else {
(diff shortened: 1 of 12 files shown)


src/common/meta/src/ddl/create_flow.rs
```diff
@@ -153,7 +183,7 @@ impl CreateFlowProcedure {
                     .map_err(add_peer_context_if_needed(peer.clone()))
             });
         }
-
+        info!("Creating flow({}) on flownodes", self.data.flow_id.unwrap());
```
WenyXu, line 186: let's better to add peer id

src/common/meta/src/key/flow/flow_name.rs
```diff
@@ -237,6 +237,39 @@ impl FlowNameManager {
...
+    )> {
+        let key = FlowNameKey::new(catalog_name, flow_name);
+        let raw_key = key.to_bytes();
+        let flow_flow_name_value = FlowNameValue::new(flow_id);
+        let txn = Txn::new()
+            .when(vec![Compare::new(
+                raw_key.clone(),
+                CompareOp::NotEqual,
+                None,
+            )])
```
WenyXu, line 261: Compare with the previous value?

src/common/meta/src/key/flow/flow_info.rs
```diff
@@ -215,6 +215,32 @@ impl FlowInfoManager {
...
+        &self,
+        flow_id: FlowId,
+        flow_value: &FlowInfoValue,
+    ) -> Result<(
+        Txn,
+        impl FnOnce(&mut TxnOpGetResponseSet) -> FlowInfoDecodeResult,
+    )> {
+        let key = FlowInfoKey::new(flow_id).to_bytes();
+        let txn = Txn::new()
+            .when(vec![Compare::new(key.clone(), CompareOp::NotEqual, None)])
```
WenyXu, line 232: ditto

src/common/meta/src/key/flow.rs
```diff
@@ -230,6 +230,93 @@ impl FlowMetadataManager {
...
+        );
+
+        let mut resp = self.kv_backend.txn(txn).await?;
+        if !resp.succeeded {
+            let mut set = TxnOpGetResponseSet::from(&mut resp.responses);
+            let remote_flow_flow_name =
+                on_create_flow_flow_name_failure(&mut set)?.with_context(|| {
+                    error::UnexpectedSnafu {
+                        err_msg: format!(
+                        "Reads the empty flow name during the creating flow, flow_id: {flow_id}"
```
WenyXu, line 287:
```suggestion
                        "Reads the empty flow name during the updating flow, flow_id: {flow_id}"
```

src/common/meta/src/key/flow.rs
```diff
@@ -230,6 +230,93 @@ impl FlowMetadataManager {
...
+                return error::FlowAlreadyExistsSnafu {
+                    flow_name: format!("{}.{}", flow_info.catalog_name, flow_info.flow_name),
+                }
+                .fail();
+            }
+
+            let remote_flow =
+                on_create_flow_failure(&mut set)?.with_context(|| error::UnexpectedSnafu {
+                    err_msg: format!(
+                        "Reads the empty flow during the creating flow, flow_id: {flow_id}"
```
WenyXu, line 310:
```suggestion
                        "Reads the empty flow during the updating flow, flow_id: {flow_id}"
```

src/common/meta/src/key/flow.rs
```diff
@@ -230,6 +230,93 @@ impl FlowMetadataManager {
...
+                    ),
+                })?;
+            let op_name = "creating flow";
```
WenyXu, line 313:
```suggestion
            let op_name = "updating flow";
```

src/common/meta/src/key/flow.rs
```diff
@@ -230,6 +230,93 @@ impl FlowMetadataManager {
         Ok(())
     }
 
+    /// Update metadata for flow and returns an error if old metadata IS NOT exists.
+    pub async fn update_flow_metadata(
```
WenyXu, line 234: Add unit tests

src/common/meta/src/key/flow/flow_name.rs
discord9: > Compare with the previous value?

I want to make sure the value exists before update txn, if set to compare with previous value then wouldn't it allow the value to be non-existing?

src/common/meta/src/key/flow/flow_info.rs
discord9: > ditto

I think here we can leave out cmp with previous value to allow `CREATE OR REPLACE` to replace previous flow even when flow's info value is exactly the same

src/common/meta/src/ddl/create_flow.rs
```diff
@@ -170,18 +218,27 @@ impl CreateFlowProcedure {
...
         // Safety: The flow id must be allocated.
         let flow_id = self.data.flow_id.unwrap();
-        // TODO(weny): Support `or_replace`.
         let (flow_info, flow_routes) = (&self.data).into();
-        self.context
-            .flow_metadata_manager
-            .create_flow_metadata(flow_id, flow_info, flow_routes)
-            .await?;
-        info!("Created flow metadata for flow {flow_id}");
+        if let Some(prev_flow_value) = self.data.prev_flow_info_value.as_ref() {
```
WenyXu, line 222:
```suggestion
        if let Some(prev_flow_value) = self.data.prev_flow_info_value.as_ref() && or_replace {
```

src/common/meta/src/ddl/create_flow.rs
```diff
@@ -98,16 +101,56 @@ impl CreateFlowProcedure {
...
+                .flow_info_manager()
+                .get(flow_id)
+                .await?;
+
+            ensure!(
+                flow_info_value.is_some(),
+                error::UnexpectedSnafu {
+                    err_msg: format!("Flow info value not found for flow_id: {}", flow_id)
+                }
+            );
```
WenyXu, line 151:
```suggestion
                .await
                .context(error::FlowNotFoundSnafu)?;
```

src/common/meta/src/key/flow.rs
```diff
@@ -230,6 +230,99 @@ impl FlowMetadataManager {
...
+        &self,
+        flow_id: FlowId,
+        flow_info: FlowInfoValue,
+        prev_flow_info: FlowInfoValue,
+        flow_routes: Vec<(FlowPartitionId, FlowRouteValue)>,
+    ) -> Result<()> {
```
WenyXu, line 240: We need to use the `DeserializedValueWithBytes`

Here is an example:
https://github.com/discord9/greptimedb/blob/e3c9e2e3bffbdf5432a7a0210acd6608a98fada5/src/common/meta/src/key/table_info.rs#L171-L176

src/common/meta/src/key/flow/flow_info.rs
```diff
@@ -215,6 +215,36 @@ impl FlowInfoManager {
...
+
+    /// Builds a update flow transaction.
+    /// It is expected that the `__flow/info/{flow_id}` IS ALREADY occupied and equal to `prev_flow_value`,
+    /// but the new value can be the same, so to allow replace operation to happen even when the value is the same.
+    /// Otherwise, the transaction will retrieve existing value and fail.
+    pub(crate) fn build_update_txn(
+        &self,
+        flow_id: FlowId,
+        flow_value: &FlowInfoValue,
+        prev_flow_value: &FlowInfoValue,
```
WenyXu, line 227: ditto

src/common/meta/src/key/flow/flow_info.rs
```diff
@@ -215,6 +215,36 @@ impl FlowInfoManager {
...
         ))
     }
+
+    /// Builds a update flow transaction.
+    /// It is expected that the `__flow/info/{flow_id}` IS ALREADY occupied and equal to `prev_flow_value`,
+    /// but the new value can be the same, so to allow replace operation to happen even when the value is the same.
+    /// Otherwise, the transaction will retrieve existing value and fail.
+    pub(crate) fn build_update_txn(
+        &self,
+        flow_id: FlowId,
```
WenyXu, line 225: I think we can use the FlowId in the `FlowInfoValue`?

src/common/meta/src/key/flow/flow_name.rs
```diff
@@ -237,6 +237,42 @@ impl FlowNameManager {
...
+        let key = FlowNameKey::new(catalog_name, flow_name);
+        let raw_key = key.to_bytes();
+        let flow_flow_name_value = FlowNameValue::new(flow_id);
+        let raw_value = flow_flow_name_value.try_as_raw_value()?;
+        let txn = Txn::new()
+            .when(vec![Compare::new(
+                raw_key.clone(),
+                CompareOp::Equal,
+                Some(raw_value),
+            )])
```
WenyXu, line 263: We can just remove the `when` cmp
```suggestion
```

src/common/meta/src/key/flow/flow_info.rs
discord9: > I think we can use the FlowId in the `FlowInfoValue`?

There is no FlowId in `FlowInfoValue`?

WenyXu (approved):
Rest LGTM

src/common/meta/src/key/flow/flow_info.rs
```diff
@@ -215,6 +228,36 @@ impl FlowInfoManager {
...
+    )> {
+        let key = FlowInfoKey::new(flow_id).to_bytes();
+        let raw_value = new_flow_value.try_as_raw_value()?;
+        let prev_value = current_flow_value.get_raw_bytes();
+        let txn = Txn::new()
+            .when(vec![
+                Compare::new(key.clone(), CompareOp::NotEqual, None),
+                Compare::new(key.clone(), CompareOp::Equal, Some(prev_value)),
+            ])
+            .and_then(vec![TxnOp::Put(key.clone(), raw_value.clone())])
```
killme2008, line 253:
```suggestion
            .and_then(vec![TxnOp::Put(key.clone(), raw_value)])
```

src/flow/src/adapter.rs
```diff
@@ -673,6 +676,20 @@ impl FlowWorkerManager {
     }
 }
 
+#[derive(Debug, Clone)]
```
killme2008, line 679:
```suggestion
/// The arguments to create a flow in [`FlowWorkerManager`].
#[derive(Debug, Clone)]
```

src/flow/src/adapter.rs
```diff
@@ -694,18 +711,47 @@ impl FlowWorkerManager {
...
+            query_ctx,
+        } = args;
+
+        let already_exist = {
+            let mut flag = false;
+
+            // check if the task already exists
+            for handle in self.worker_handles.iter() {
+                if handle.lock().await.contains_flow(flow_id).await? {
+                    flag = true;
```
killme2008, line 734:
```suggestion
                    flag = true;
                    break;
```

src/flow/src/server.rs
```diff
@@ -355,23 +355,23 @@ impl FlownodeBuilder {
...
-                            .current_catalog(info.catalog_name().clone())
-                            .build(),
-                    ),
-                )
-                .await?;
+            let args = CreateFlowArgs {
+                flow_id: flow_id as _,
+                sink_table_name,
+                source_table_ids: info.source_table_ids().to_vec(),
+                create_if_not_exists: true,
```
killme2008, line 362: Why hard coding these parameters? `create_if_not_exists: true` etc.

discord9: > Why hard coding these parameters? `create_if_not_exists: true` etc.

because recover should only happen on restart the `create_if_not_exists` and `or_replace` can be arbitrary value(since flow doesn't exist)
 but for the sake of consistency and to make sure recover of flow actually happen, set both to true
 (which is also fine since checks for not allow both to be true is on metasrv and we already pass that)
 those comments and a new sqlness restart tests are also added for recover flow

killme2008 (approved):
LGTM

Issue #4959, opened 2024-11-07
Window sort support limit offset


evenyag:
### What type of enhancement is this?

Performance

### What does the enhancement do?

Now window sort only supports limit without offset.
```
mysql> explain analyze select ts from cpu order by ts desc limit 1, 100;
+-------+------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| stage | node | plan                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
+-------+------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|     0 |    0 |  MergeScanExec: peers=[4398046511104(1024, 0), ] metrics=[output_rows: 100, greptime_exec_read_cost: 0, ready_time: 1367518, first_consume_time: 2405064866, finish_time: 2405182055, ]
                                                                                                                                                                                                                                                                                   |
|     1 |    0 |  GlobalLimitExec: skip=1, fetch=100 metrics=[output_rows: 100, elapsed_compute: 1112, ]
  SortExec: TopK(fetch=101), expr=[ts@0 DESC], preserve_partitioning=[false] metrics=[output_rows: 101, elapsed_compute: 1542010568, row_replacements: 26402, ]
    SeqScan: region=4398046511104(1024, 0), partition_count=1 (0 memtable ranges, 6 file 1016 ranges) metrics=[output_rows: 103680000, mem_used: 1058324480, elapsed_poll: 626485335, elapsed_await: 809918628, ]
 |
|  NULL | NULL | Total rows: 100                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
+-------+------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
3 rows in set (2.43 sec)
```

Support offset can reduce fetching all rows.

### Implementation challenges

_No response_

LYZJU2019:
Can I give it a try?

evenyag:
@waynexia Are you working on this?

waynexia:
[...]

PR #5018, opened 2024-11-19
feat: reimplement limit in PartSort to reduce memory footprint

waynexia:

Closes https://github.com/GreptimeTeam/greptimedb/issues/4959

Use `TopK` buffer to leverage a heap-based limit implementation which should greatly reduce the memory footprint

## Checklist


evenyag (approved):

src/query/src/part_sort.rs
```diff
@@ -338,16 +375,42 @@ impl PartSortStream {
         Ok(None)
     }
 
+    fn push_buffer(&mut self, batch: DfRecordBatch) -> datafusion_common::Result<()> {
+        match &mut self.buffer {
+            PartSortBuffer::All(v) => v.push(batch),
+            PartSortBuffer::Top(top, cnt) => {
+                *cnt += batch.num_rows();
```
killme2008, line 382: Looks like the row count number records how many rows were pushed into this buffer, but maybe it is not precise because the `TopK` only keeps top `k` rows. Am I right?

src/query/src/part_sort.rs
```diff
@@ -418,15 +481,64 @@ impl PartSortStream {
             )
         })?;
 
-        // only clear after sorted for better debugging
-        self.buffer.clear();
         self.produced += sorted.num_rows();
```
killme2008, line 484: Looks like removing the `buffer.clear()`, any considerations about it?

src/query/src/part_sort.rs
```diff
@@ -418,15 +481,64 @@ impl PartSortStream {
...
+            self.partition,
+            self.schema().clone(),
+            vec![self.expression.clone()],
+            self.limit.unwrap(),
+            self.context.session_config().batch_size(),
+            self.context.runtime_env(),
+            &self.root_metrics,
+            self.partition,
+        )?;
+        let PartSortBuffer::Top(top_k, _) =
```
killme2008, line 503: Hmmm. It seems the row count number is useless.

src/query/src/part_sort.rs
```diff
@@ -418,15 +481,64 @@ impl PartSortStream {
...
+            std::mem::replace(&mut self.buffer, PartSortBuffer::Top(new_top_buffer, 0))
+        else {
+            unreachable!("buffer type is checked before and should be Top variant")
+        };
+
+        let mut result_stream = top_k.emit()?;
+        let mut placeholder_ctx = std::task::Context::from_waker(futures::task::noop_waker_ref());
+        let mut results = vec![];
```
killme2008, line 511: Not sure if we can give the vector an initial capacity.

killme2008 (commented):
Great work!

src/query/src/part_sort.rs
```diff
@@ -418,15 +481,64 @@ impl PartSortStream {
...
+        let mut row_count = 0;
+        // according to the current implementation of `TopK`, the result stream will always be ready
+        while let Poll::Ready(Some(batch)) = result_stream.poll_next_unpin(&mut placeholder_ctx) {
```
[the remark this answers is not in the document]
waynexia: Done

src/query/src/part_sort.rs
waynexia: We can't know how long an async stream is

src/query/src/part_sort.rs
waynexia: It's used to check if the buffer is empty. I've updated the doc comment

src/query/src/part_sort.rs
waynexia: `clear` is moved to the front (it's `std::mem::replace` now)

src/query/src/part_sort.rs
waynexia: Yes, this `cnt` might be greater than `k`. It's usage is to check if the buffer is empty. I've updated the doc comment.

discord9 (approved):

PR #5021, opened 2024-11-19
chore(cli): set default timeout for cli commands

WenyXu:

Update the timeout option from `Option<Duration>` to `Duration` . And users can explicitly set the timeout to `0s` to disable server-side timeout configuration.

## Checklist


fengjiachun (approved):
LGTM

shuiyisong (approved):

waynexia:
Can you provide more context about this change? Like why we need it. We can't set it to none once there is a default value.

WenyXu:
> Can you provide more context about this change? Like why we need it. We can't set it to none once there is a default value.

It is used to override the server-side timeout setting

Issue #4970, opened 2024-11-11
Parameter number mismatch because of datafusion optimization


CookiePieWw:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Frontend

### Minimal reproduce step

MySQL CLI:
```
PREPARE stmt FROM 'SELECT ? + 1 WHERE 1 = 0';
EXECUTE stmt USING 1;
```

### What did you expect to see?

Execute the query successfully

### What did you see instead?

ERROR 1314 (0A000): prepare statement params number mismatch

### What operating system did you use?

doesnt matter

### What version of GreptimeDB did you use?

0.9.5

### Relevant log output and stack trace

Currently `EXECUTE` checks its param num through a cached datafusion plan generated by PREPARE stmt, but datafusion may optimize the plan so that the required param may be optimized out. e.g. `PREPARE stmt FROM 'SELECT ? + 1 WHERE 1 = 0';` is optimized to an empty plan but still require 1 params for execution.The param inference is affected as well.

CookiePieWw:
Closed by #5025

PR #5025, opened 2024-11-20
fix: prepare param mismatch

CookiePieWw:

#4970 

Datafusion may convert the prepared sql to an optimized plan without initial params, so that we can't confirm if the param is needed when executing the prepared plan.

This patch introduces a workaroud that just drop the optimized plan if the param number mismatches and use the initial sql. It may affect the performance of the sql, but previously these sqls even cannot execute.

e.g.
```SQL
PREPARE stmt FROM 'SELECT ? + 1 WHERE 1 = 0';
EXECUTE stmt USING 1;
```

## Checklist

### DIFF
--- a/src/servers/src/mysql/handler.rs
+++ b/src/servers/src/mysql/handler.rs
@@ -189,8 +189,6 @@ impl MysqlInstanceShim {
             dummy_params(param_num)?
         };
 
-        debug_assert_eq!(params.len(), param_num - 1);
-
         let columns = schema
             .as_ref()
             .map(|schema| {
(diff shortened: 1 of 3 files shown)


waynexia:
I'm curious about if this case will only generate a static plan without any input, or it will have other cases of mismatch. If not we can just drop the input and evaluate the static plan instead.

CookiePieWw:
> I'm curious about if this case will only generate a static plan without any input, or it will have other cases of mismatch. If not we can just drop the input and evaluate the static plan instead.

Initially I thought there're cases that part of the params are optimized and others are kept,  thus we cannot match the params given by users to the optimized plan. But I'm not sure though, it's up to datafusion's optimization rules.

killme2008 (approved):
LGTM

zhongzc (approved):

PR #5032, opened 2024-11-20
fix: remove metric engine's internal column from promql's query 

waynexia:

Filter out internal columns in promql's query. Those columns are valid for SQL query but PromQL requires them to be string.

## Checklist


shuiyisong:
https://github.com/GreptimeTeam/greptimedb/blob/c199604ece2060c0616bbf2ec182ca17f68f1cf6/src/servers/src/http/prometheus.rs#L695-L708

When listing metric names from the database, currently we don't have any filter, which means tables without `greptime_timestamp` and `greptime_value` will also be displayed. These tables do not fit Prometheus rules and will not generate any value in Grafana when queried using PromQL. 

It would be better to filter out all tables without `greptime_timestamp` and `greptime_value` in this API in advance.

waynexia:
>When listing metric names from the database, currently we don't have any filter, which means tables without greptime_timestamp and greptime_value will also be displayed. These tables do not fit Prometheus rules and will not generate any value in Grafana when queried using PromQL.

We only require a time index column to run a basic promql query, which is satisfied by all tables. But I do add a logic to filter out physical table which should not be expected here.

killme2008 (approved):
LGTM

killme2008:
Can you add some sqlness test for this fix? @waynexia

shuiyisong (approved):

PR #5044, opened 2024-11-24
fix: prevent metadata region from inheriting database ttl

waynexia:

A quick bad fix for 0.10.1.

Metadata region has a constant 0 as time index column's value. When inheriting TTL from database it will remove all metadata. 

## Checklist


evenyag (approved):
😢

killme2008 (approved):
Tricky but works

Issue #5042, opened 2024-11-23
Alter request to physical region is forbidden


killme2008:
### What type of bug is this?

Unexpected error, User Experience

### What subsystems are affected?

Storage Engine

### Minimal reproduce step

Try to change a physical table `ttl`:

```sql
mysql> ALTER TABLE greptime_physical_table set ttl='2years';
ERROR 1235 (42000): (Unsupported): Alter request to physical region is forbidden
```

But it fails.

### What did you expect to see?

Success to alter

### What did you see instead?

Throws error.

### What operating system did you use?

All

### What version of GreptimeDB did you use?

0.10.0

killme2008:
Actually, we don't need to forbid this operation for physical tables at all.

evenyag:
Can we close this issue now?

killme2008:
Yes, fixed via #5046 , thansk to @lyang24

PR #5046, opened 2024-11-25
fix: allow physical region alter region options

lyang24:

https://github.com/GreptimeTeam/greptimedb/issues/5042

allow set and unset of region options in alter physical region request.
allow Alter/ unset region options by passing the request to DataNode's mito engine. 

## Checklist

### DIFF
--- a/src/metric-engine/src/engine.rs
+++ b/src/metric-engine/src/engine.rs
@@ -96,9 +96,10 @@ use crate::utils;
 /// |    Read    |       ✅        |        ✅        |
 /// |   Close    |       ✅        |        ✅        |
 /// |    Open    |       ✅        |        ✅        |
-/// |   Alter    |       ✅        |        ❌        |
+/// |   Alter    |       ✅        |        ❓*       |
 ///
 /// *: Physical region can be dropped only when all related logical regions are dropped.
+/// *: Alter: Physical regions only support altering region options.
 ///
 /// ## Internal Columns
 ///
--- a/src/metric-engine/src/engine/alter.rs
+++ b/src/metric-engine/src/engine/alter.rs
@@ -14,7 +14,7 @@
 
 use std::collections::HashMap;
 
-use common_telemetry::{error, info};
+use common_telemetry::error;
 use snafu::{OptionExt, ResultExt};
 use store_api::metadata::ColumnMetadata;
 use store_api::metric_engine_consts::ALTER_PHYSICAL_EXTENSION_KEY;
@@ -22,10 +22,7 @@ use store_api::region_request::{AffectedRows, AlterKind, RegionAlterRequest};
 use store_api::storage::RegionId;
 
 use crate::engine::MetricEngineInner;
-use crate::error::{
-    ForbiddenPhysicalAlterSnafu, LogicalRegionNotFoundSnafu, Result, SerializeColumnMetadataSnafu,
-};
-use crate::metrics::FORBIDDEN_OPERATION_COUNT;
+use crate::error::{LogicalRegionNotFoundSnafu, Result, SerializeColumnMetadataSnafu};
 use crate::utils::{to_data_region_id, to_metadata_region_id};
 
 impl MetricEngineInner {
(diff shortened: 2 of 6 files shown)


src/metric-engine/src/engine.rs
```diff
@@ -96,9 +96,10 @@ use crate::utils;
 /// |    Read    |       ✅        |        ✅        |
 /// |   Close    |       ✅        |        ✅        |
 /// |    Open    |       ✅        |        ✅        |
-/// |   Alter    |       ✅        |        ❌        |
+/// |   Alter    |       ✅        |        ❓*       |
 ///
 /// *: Physical region can be dropped only when all related logical regions are dropped.
+/// *: Alter: Physical regions only support altering region options.
```
waynexia, line 102: 👍 Nice Document

waynexia (approved):
LGTM, thanks @lyang24

waynexia (changes requested):
Sorry, I forget one thing. The alter request must be processed by the underlying physical region (mito region) as well. Otherwise it's only presented in metasrv and causing inconsistency.

src/metric-engine/src/engine/alter.rs
```diff
@@ -150,10 +150,16 @@ impl MetricEngineInner {
...
+        match request.kind {
+            AlterKind::SetRegionOptions { options: _ }
+            | AlterKind::UnsetRegionOptions { keys: _ } => Ok(()),
+            _ => {
+                info!("Metric region received alter request {request:?} on physical region {region_id:?}");
+                FORBIDDEN_OPERATION_COUNT.inc();
+
+                ForbiddenPhysicalAlterSnafu.fail()
+            }
+        }
```
WenyXu, line 162: You need to alter the underlying mito region as well. i.e., `self.mito.handle_request(region_id, request)`

lyang24: Thank you for catching this. I have two follow up questions:
- is there a place that i can add a test for this senario?
- I see `alter_logical_region` only support add column request, does that mean we can only set ttls to physical table on metric engine?

WenyXu: > is there a place that i can add a test for this scenario?

You can look at a test named `test_alter_column_fulltext_options` in the `mito` crate.

> does that mean we can only set ttls to physical table on metric engine?

Yes, we can set TTLs and other table options of the physical table.

lyang24:
> LGTM, It would be perfect if you could add a `test_alter_column_ttl_options` test in the mito crate.

Hi I added a test but I noticed the ttl did not persist after region restart. Can you help me understand with if its my test structure (region metadata not flushed?) or its a bug.

src/mito2/src/engine/alter_test.rs
```diff
@@ -573,6 +574,80 @@ async fn test_alter_column_fulltext_options() {
...
+                engine: String::new(),
+                region_dir,
+                options: HashMap::default(),
+                skip_wal_replay: false,
+            }),
+        )
+        .await
+        .unwrap();
+    // Verify the ttl again after engine restart.
+    // check_ttl(&engine, &Duration::from_secs(500));
```
lyang24, line 648: Hi @WenyXu, the ttl verify will fail after region restart. The restarted region had `None` as ttl.

WenyXu: https://github.com/lyang24/greptimedb/blob/5faf2f483a2399960cd10df7380d517591977f66/src/mito2/src/region/options.rs#L102-L149

It's expected that you need to open the region with the specified options, including the `ttl`.

i.e.,
```rust
   let options = HashMap::from([("ttl".to_string(), "500s".to_string())]);
    engine
        .handle_request(
            region_id,
            RegionRequest::Open(RegionOpenRequest {
                engine: String::new(),
                region_dir,
                options,
                skip_wal_replay: false,
            }),
        )
        .await
        .unwrap();
```

lyang24: I see thanks, i will just remove the reopen engine part of the test as hardcode an option do not provide much value for this test.

killme2008:
@lyang24 Could you please rebase this branch with the main branch? There is a typo config issue.

killme2008:
@waynexia Please take a look.

lyang24:
> @waynexia Please take a look.

thank you

waynexia (approved):
LGTM, this change won't affect metadata region.

PR #5051, opened 2024-11-26
fix(metric-engine): set ttl also on opening metadata regions

v0y4g3r:

#5044

The region options for metadata region in metric engines should stay independent from database options, no matter during opening or creating regions.

## Checklist

### DIFF
--- a/src/metric-engine/src/engine/create.rs
+++ b/src/metric-engine/src/engine/create.rs
@@ -32,7 +32,7 @@ use store_api::metric_engine_consts::{
     METADATA_SCHEMA_TIMESTAMP_COLUMN_INDEX, METADATA_SCHEMA_TIMESTAMP_COLUMN_NAME,
     METADATA_SCHEMA_VALUE_COLUMN_INDEX, METADATA_SCHEMA_VALUE_COLUMN_NAME,
 };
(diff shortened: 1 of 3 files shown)


src/metric-engine/src/engine/create.rs
```diff
@@ -539,6 +535,13 @@ impl MetricEngineInner {
     }
 }
 
+/// Creates the region options for metadata region in metric engine.
+pub(crate) fn region_options_for_metadata_region() -> HashMap<String, String> {
+    [(TTL_KEY.to_string(), "10000 years".to_string())]
```
killme2008, line 540: I think we must clone the original option and do the stuff.

v0y4g3r: > I think we must clone the original option and do the stuff.

Metadata region for metric engine should stay unaffected from any table level options.

evenyag (approved):

src/metric-engine/src/engine/create.rs
killme2008: I disagree. For example, Maye the user prefers to explicitly set the compaction options. That way, if we add new options in the future, we won't run into unexpected behavior in this region.

killme2008 (approved):
LGTM

PR #5061, opened 2024-11-27
fix(flow): minor fix about count(*)&sink keyword

discord9:

- fix can't use `count(*)` in flow creation
- fix `SINK` should be case insensitive but not

## Checklist

### DIFF
--- a/src/sql/src/parsers/create_parser.rs
+++ b/src/sql/src/parsers/create_parser.rs
@@ -259,9 +259,16 @@ impl<'a> ParserContext<'a> {
 
         let flow_name = self.intern_parse_table_name()?;
 
-        self.parser
-            .expect_token(&Token::make_keyword(SINK))
-            .context(SyntaxSnafu)?;
+        // make `SINK` case in-sensitve
+        if let Token::Word(word) = self.parser.peek_token().token
+            && word.value.to_uppercase() == SINK
+        {
+        } else {
+            Err(ParserError::ParserError(
+                "Expect `SINK` keyword".to_string(),
+            ))
+            .context(SyntaxSnafu)?
+        }
         self.parser
             .expect_keyword(Keyword::TO)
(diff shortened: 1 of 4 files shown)


tests/cases/standalone/common/flow/flow_basic.sql
```diff
@@ -61,6 +61,35 @@ DROP TABLE numbers_input_basic;
...
+SELECT
+    COUNT(*) as wildcard
+FROM
+    input_basic;
+
+INSERT INTO
+    input_basic
+VALUES
+    (23, "2021-07-01 00:00:01.000"),
+    (24, "2021-07-01 00:00:01.500");
```
shuiyisong, line 82: nitpick: we can add a test to use lowercase as keywords to see if it works, for example:
```
create flow test_wildcard_basic sink to out_basic as
select
    count(*) as wildcard
from
    input_basic;

insert into
    input_basic
values
    (23, "2021-07-01 00:00:01.000"),
    (24, "2021-07-01 00:00:01.500");
```

shuiyisong (approved):
Rest LGTM

src/sql/src/parsers/create_parser.rs
```diff
@@ -259,9 +259,17 @@ impl<'a> ParserContext<'a> {
 
         let flow_name = self.intern_parse_table_name()?;
 
-        self.parser
-            .expect_token(&Token::make_keyword(SINK))
-            .context(SyntaxSnafu)?;
+        // make `SINK` case in-sensitive
+        if let Token::Word(word) = self.parser.peek_token().token
+            && word.value.to_uppercase() == SINK
```
zhongzc, line 264:
```suggestion
            && word.value.eq_ignore_ascii_case(SINK)
```

waynexia (approved):

PR #5074, opened 2024-11-29
feat(fuzz): add alter table options for alter fuzzer

CookiePieWw:

As title. Also move some files for clearer structure.

Currently there's no validation for the alter table option stmt, just add them to the fuzzer.

## Checklist


WenyXu:
I think we need to verify table options after alterations.

CookiePieWw:
> I think we need to verify table options after alterations.

Agree. We can add a table option validator like row/column validator. The column types/names can be validate in validator as well.

CookiePieWw:
Updated with a table option validator, ptal :) @WenyXu

WenyXu (approved):
LGTM

tests-fuzz/src/translator/postgres/alter_expr.rs
```diff
@@ -112,15 +118,40 @@ impl AlterTableExprTranslator {
...
+
+    fn format_unset_table_options(name: impl Display, keys: &[String]) -> String {
+        format!(
+            "ALTER TABLE {name} UNSET {};",
+            keys.iter()
+                .map(|key| format!("'{}'", key))
+                .collect::<Vec<_>>()
+                .join(", ")
+        )
+    }
```
evenyag, line 141: Is it possible to share some codes between `tests-fuzz/src/translator/mysql/alter_expr.rs` and `tests-fuzz/src/translator/postgres/alter_expr.rs`?

tests-fuzz/src/generator/alter_expr.rs
```diff
@@ -191,6 +194,105 @@ impl<R: Rng> Generator<AlterTableExpr, R> for AlterExprModifyDataTypeGenerator<R
...
+                AlterTableOption::TwcsTimeWindow(_) => {
+                    let time_window: u32 = rng.gen();
+                    AlterTableOption::TwcsTimeWindow((time_window as i64).into())
+                }
+                AlterTableOption::TwcsMaxOutputFileSize(_) => {
+                    let max_output_file_size: u64 = rng.gen();
+                    // Use B as the unit by default
+                    let max_output_file_size_str = max_output_file_size.to_string() + "B";
+                    AlterTableOption::TwcsMaxOutputFileSize(
+                        ReadableSize::from_str(&max_output_file_size_str).unwrap(),
```
evenyag, line 235: What about constructing the readable size directly?
```suggestion
                    AlterTableOption::TwcsMaxOutputFileSize(
                        ReadableSize(max_output_file_size),
```

tests-fuzz/targets/ddl/fuzz_alter_table.rs
```diff
@@ -179,6 +198,28 @@ async fn execute_alter_table(ctx: FuzzContext, input: FuzzInput) -> Result<()> {
...
+            .iter()
+            .zip(expected_table_options.iter())
+            .for_each(|(a, b)| {
+                if let (
+                    AlterTableOption::TwcsMaxOutputFileSize(a),
+                    AlterTableOption::TwcsMaxOutputFileSize(b),
+                ) = (a, b)
+                {
+                    // File size is not exact, so we compare string representation
+                    assert_eq!(a.to_string(), b.to_string());
```
evenyag, line 218: We should compare the `ReadableSize` struct if possible.

tests-fuzz/src/validator/table.rs
```diff
@@ -0,0 +1,100 @@
...
+/// | Table | Create Table                                                                                                                                                                                           |
+/// +-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+/// | json  | CREATE TABLE IF NOT EXISTS `json` (`ts` TIMESTAMP(3) NOT NULL, `j` JSON NULL, TIME INDEX (`ts`)) ENGINE=mito WITH(compaction.twcs.max_output_file_size = '1M', compaction.type = 'twcs', ttl = '1day') |
+/// +-------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+fn parse_show_create(show_create: &str) -> Result<Vec<AlterTableOption>> {
+    let option_start = show_create.find("WITH(");
+    if option_start.is_none() {
+        Ok(vec![])
+    } else {
+        let option_start = option_start.unwrap() + 5;
```
evenyag, line 34: We can use `if-let-else` to get option_start without `unwrap`.

tests-fuzz/src/validator/table.rs
```diff
@@ -0,0 +1,100 @@
...
+        let option_end = {
+            let mut i = 1;
+            while show_create.chars().nth(option_start + i).unwrap() != ')' {
+                i += 1;
+            }
+            i + option_start
```
evenyag, line 40: We can get a str from option_start and use find to find the position of `)`.

tests-fuzz/targets/ddl/fuzz_alter_table.rs
CookiePieWw: Currently the file size is extracted and parsed from `SHOW CREATE` results, which is not exact. I wonder if there's some method to get the exact `TwcsMaxOutputFileSize` through sql interface.

tests-fuzz/src/generator/alter_expr.rs
```diff
@@ -191,6 +194,105 @@ impl<R: Rng> Generator<AlterTableExpr, R> for AlterExprModifyDataTypeGenerator<R
...
+            let option = rng.gen_range(0..all_options.len());
+            if !option_templates_idx.contains(&option) {
+                option_templates_idx.push(option);
+            }
+        }
+        let options = option_templates_idx
+            .iter()
+            .map(|idx| match all_options[*idx] {
+                AlterTableOption::Ttl(_) => {
+                    let ttl: u32 = rng.gen();
```
discord9, line 223: might consider add more possible values for ttl once #5089 merged

CookiePieWw:
Updated with new ttl options, PTAL :) @WenyXu @evenyag

tests-fuzz/targets/ddl/fuzz_alter_table.rs
evenyag: If the string representations are equal, why does the file size extracted is not exact?

CookiePieWw: Seems a float point issue, example:
The fuzzer set `ALTER TABLE `AUt` SET 'compaction.twcs.max_output_file_size' = '11547.5PiB'`, and read 11547.5PiB as well, but the `u64` inside the readable size is:
image
So the assertion failed in a weird way:
image

PR #5082, opened 2024-12-02
fix: correct `is_exceeded_size_limit` behavior for in-memory store

WenyXu:

This PR fixes the behavior of `PaginationStream`, ensuring it automatically adjusts the next request's `page_size` when the range response size from the in-memory store exceeds the size limit.

## Checklist

### DIFF
--- a/src/common/meta/src/error.rs
+++ b/src/common/meta/src/error.rs
@@ -425,8 +425,8 @@ pub enum Error {
         source: BoxedError,
     },
 
-    #[snafu(display("Meta client exceeded size limit"))]
-    MetaClientExceededSizeLimit {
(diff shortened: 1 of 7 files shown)


src/common/meta/src/error.rs
```diff
@@ -425,6 +425,13 @@ pub enum Error {
         source: BoxedError,
     },
 
+    #[snafu(display("Meta client exceeded size limit"))]
+    MetaClientExceededSizeLimit {
```
WenyXu, line 429:
```suggestion
    ResponseExceededSizeLimit {
```

MichaelScofield (approved):

PR #5086, opened 2024-12-03
feat: define basic structures and implement TimeFilter

waynexia:

Define some basic log query structures

## Checklist

### DIFF
--- a/src/log-query/src/log_query.rs
+++ b/src/log-query/src/log_query.rs
@@ -200,12 +200,20 @@ pub struct ColumnFilters {
 }
 
 pub enum ContentFilter {
-    Term(String),
+    /// Only match the exact content.
+    ///
+    /// For example, if the content is "pale blue dot", the filter "pale" or "pale blue" will match.
+    Exact(String),
+    /// Match the content with a prefix.
+    ///
+    /// For example, if the content is "error message", the filter "err" or "error mess" will match.
     Prefix(String),
+    /// Match the content with a postfix. Similar to `Prefix`.
     Postfix(String),
+    /// Match the content with a substring.
     Contains(String),
+    /// Match the content with a regex pattern. The pattern should be a valid Rust regex.
     Regex(String),
-    Phase(String),
     Compound(Vec<ContentFilter>, BinaryOperator),
 }
 
--- a/src/log-query/src/log_query.rs
+++ b/src/log-query/src/log_query.rs
@@ -27,6 +27,8 @@ pub struct LogQuery {
     pub columns: Vec<ColumnFilters>,
     /// Maximum number of logs to return. If not provided, it will return all matched logs.
     pub limit: Option<usize>,
+    /// Adjacent lines to return.
+    pub context: Context,
 }
 
 /// Represents a time range for log query.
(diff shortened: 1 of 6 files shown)


src/log-query/src/log_query.rs
```diff
@@ -0,0 +1,323 @@
...
+    Regex(String),
+    Compound(Vec<ContentFilter>, BinaryOperator),
+}
+
+pub enum BinaryOperator {
+    And,
+    Or,
+}
+
+pub enum Context {
```
zyy17, line 242: question: In the context structure, does `usize` refer to the number of lines? If so, can we add the `range` semantics to express the data from a time before and after a specific match? For example, query the logs from 5 minutes before and after a particular keyword appears.

src/log-query/src/log_query.rs
```diff
@@ -0,0 +1,322 @@
...
+///     is provided, `end` is calculated as `start + span` and vice versa.
+/// - Only `span` is provided: the `span` string can be any valid "interval" as mentioned
+///     above. In this case, the current time (on the server side) is considered as the `end`.
+/// - All fields are provided: in this case, the `start` and `end` fields are considered
+///     with higher priority, and the `span` field is ignored.
+///
+/// This struct doesn't require a timezone to be presented. When the timezone is not
+/// provided, it will fill the default timezone with the same rules akin to other queries.
+#[derive(Debug, Clone)]
+pub struct TimeFilter {
```
paomian, line 62: Can support timestamp milliseconds number？

paomian (approved):
REST LGTM

zyy17 (approved):
Mostly LGTM

src/log-query/src/log_query.rs
```diff
@@ -0,0 +1,322 @@
...
+/// GreptimeDB's log query request.
+pub struct LogQuery {
+    /// A fully qualified table name to query logs from.
+    pub table_name: TableName,
+    /// Specifies the time range for the log query. See [`TimeFilter`] for more details.
+    pub time_filter: TimeFilter,
+    /// Columns with filters to query.
+    pub columns: Vec<ColumnFilters>,
+    /// Maximum number of logs to return. If not provided, it will return all matched logs.
+    pub limit: Option<usize>,
```
shuiyisong, line 32: Maybe we should set the default to 1000 lines or so. It's easy to write a careless query and result in huge returning lines during debugging. Also, it seems a common way to do it (with default limits).
Explicitly set to `0` to return all lines.

src/log-query/src/log_query.rs
```diff
@@ -0,0 +1,322 @@
...
+
+        if let Some(end) = end_dt {
+            self.end = Some(end.to_rfc3339());
+        }
+
+        Ok(())
+    }
+
+    /// Util function returns a start and optional end DateTime
+    fn parse_datetime(s: &str) -> Result<(DateTime<Utc>, Option<DateTime<Utc>>)> {
```
shuiyisong, line 155: It looks like a very generic parse function, should it be moved to `common-time`?
Same as `parse_span` below.

shuiyisong (approved):
Rest LGTM

src/log-query/src/log_query.rs
waynexia: Timestamp format can be millisecond or higher resolutions

src/log-query/src/log_query.rs
waynexia: Good suggestion 👍 The default behavior is not going to be placed in the query definition. I'll set one on executing

src/log-query/src/log_query.rs
waynexia: `parse_datetime` is not that generic for now as here is the first and only place that accepts both date and timestamp. I'd do a refactor afterward for `parse_span` to combine the same piece of logic in the plan rewriter.

PR #5089, opened 2024-12-03
feat: ttl=0/instant/forever/humantime&ttl refactor

discord9:

support `ttl='instant'/'forever'`

- support `ttl='instant'/'forever'`, however to keep compatibilty, `ttl='0s'/'null'/''` still equals to `ttl='forever'`
- refactor ttl to enum
- refactor table/database ttl, so that once table ttl is set, it can't be override by database, only after unset table ttl can database ttl be applied on a table
- add more sqlness tests for alter ttl

## Checklist


src/common/base/src/ttl.rs
```diff
@@ -0,0 +1,182 @@
...
+pub enum TimeToLive {
+    /// immediately throw away on insert
+    Immediate,
+    /// Duration to keep the data, this duration should be non-zero
+    Duration(Duration),
+    /// Keep the data forever
+    #[default]
+    Forever,
+    // TODO(discord9): add a new variant
+    // that can't be overridden by database level ttl? call it ForceForever?
```
v0y4g3r, line 33: If table-level ttl is set to forever, it won't be overriden by database ttl.

Table level ttl should be an `Option<TimeToLive>`.

If it's None, then it follows database ttl. If database ttl is not set, then it's forever
If it's Some(*), it will always ignore database ttl.

src/common/base/src/ttl.rs
```diff
@@ -0,0 +1,182 @@
...
+use std::time::Duration;
+
+use serde::de::Visitor;
+use serde::{Deserialize, Serialize};
+
+/// Time To Live
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Default)]
+pub enum TimeToLive {
+    /// immediately throw away on insert
```
v0y4g3r, line 25:
```suggestion
    /// Immediately throw away on insert
```

src/common/meta/src/ddl/alter_database.rs
```diff
@@ -46,19 +47,15 @@ fn build_new_schema_value(
...
+                        value.ttl = *ttl;
                     }
                 }
             }
         }
         AlterDatabaseKind::UnsetDatabaseOptions(keys) => {
             for key in keys.0.iter() {
                 match key {
-                    UnsetDatabaseOption::Ttl => value.ttl = None,
+                    UnsetDatabaseOption::Ttl => value.ttl = TimeToLive::default(),
```
v0y4g3r, line 58: Like I mentioned, it will be a None.

src/store-api/src/region_request.rs
```diff
@@ -746,7 +746,7 @@ impl From<v1::ModifyColumnType> for ModifyColumnType {
 
 #[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize)]
 pub enum SetRegionOption {
-    TTL(Duration),
+    TTL(Option<TimeToLive>),
```
evenyag, line 749: Maybe we can rename the variant with `Ttl`.

src/session/src/context.rs
```diff
@@ -59,6 +60,8 @@ pub struct QueryContext {
 #[derive(Debug, Builder, Clone, Default)]
 pub struct QueryContextMutableFields {
     warning: Option<String>,
+    /// regions with ttl=0
+    ttl_zero_regions: HashSet<u64>,
```
evenyag, line 64: Use `immediate` instead of `ttl_zero` which is confusing.

src/query/src/sql/show_create_table.rs
```diff
@@ -46,13 +45,13 @@ fn create_sql_options(table_meta: &TableMeta, schema_options: Option<SchemaOptio
...
-    } else if let Some(database_ttl) = schema_options.and_then(|o| o.ttl) {
-        options.insert(
-            TTL_KEY.to_string(),
-            format_duration(database_ttl).to_string(),
-        );
+    if let Some(ttl) = table_opts.ttl.and_then(|t| t.as_repr_opt()) {
+        options.insert(TTL_KEY.to_string(), ttl);
+    } else if let Some(database_ttl) = schema_options
+        .and_then(|o| o.ttl)
+        .and_then(|ttl| ttl.as_repr_opt())
```
evenyag, line 52: We can call `as_repr_opt` in the first `and_then`
```suggestion
        .and_then(|o| o.ttl.as_repr_opt())
```

src/operator/src/insert.rs
```diff
@@ -265,6 +267,15 @@ impl Inserter {
...
 }
 
+/// Check and add regions with ttl=0 to context
+pub(crate) fn check_ttl_zero_table(ctx: &QueryContextRef, table_info: &TableInfo) {
+    let ttl = table_info.meta.options.ttl;
+
+    if ttl.map(|t| t.is_immediate()).unwrap_or(false) {
+        ctx.add_ttl_zero_regions(table_info.region_ids().into_iter().map(|i| i.as_u64()));
+    }
+}
```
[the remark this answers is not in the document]
evenyag: It not only checks the table but also modifies the context so we may consider a new name for this function.

src/operator/src/insert.rs
```diff
@@ -558,6 +575,17 @@ impl Inserter {
...
 
+        // add ttl zero regions to context
+        for table_name in table_name_to_ids.keys() {
+            let table = if let Some(table) = self.get_table(catalog, &schema, table_name).await? {
+                table
+            } else {
+                continue;
+            };
+            let table_info = table.table_info();
+            check_ttl_zero_table(ctx, &table_info);
```
evenyag, line 586: I see we add this check in a lot of places. It's easy to make mistakes. I think we should refactor the return value of `RowToRegion`, `TableToRegion`, and `StatementToRegion` to tell us which request is for immediate regions.

killme2008:
Use `instant` instead of `immediate` for brevity and clarity. What do you think? @discord9

src/query/src/sql/show_create_table.rs
discord9: > We can call `as_repr_opt` in the first `and_then`

the first ttl is a Option so two and_then for flatten

src/operator/src/insert.rs
discord9: > I see we add this check in a lot of places. It's easy to make mistakes. I think we should refactor the return value of `RowToRegion`, `TableToRegion`, and `StatementToRegion` to tell us which request is for immediate regions.

How about just filter those insert requests out in this step? Like rename `TableToRegion.convert` to `convert_no_imme_region` and etc.

evenyag:
> > Use `instant` instead of `immediate` for brevity and clarity. What do you think? @discord9
> 
> How about `imme` or `now`? `instant` also seems a bit long

`now` sounds good. `imme` isn't well-known.

discord9:
Seems to be encounter https://github.com/rust-lang/rust/issues/64552
Trying to bypass it

src/operator/src/insert.rs
```diff
@@ -558,7 +595,21 @@ impl Inserter {
             }
         }
 
-        Ok(table_name_to_ids)
+        let mut imme_table_ids = HashSet::new();
+        // add ttl zero regions to context
+        for table_name in table_name_to_ids.keys() {
```
killme2008, line 600: Looks like this iteration can be inlined to above loop at https://github.com/GreptimeTeam/greptimedb/blob/66c04459749703c9926b0aabfd52fccf1371bd93/src/operator/src/insert.rs#L504

Avoid to `get_table` so many times.

src/operator/src/req_convert/insert/table_to_region.rs
```diff
@@ -44,7 +45,19 @@ impl<'a> TableToRegion<'a> {
         let requests = Partitioner::new(self.partition_manager)
             .partition_insert_requests(self.table_info.table_id(), rows)
             .await?;
-        Ok(RegionInsertRequests { requests })
+
+        let requests = RegionInsertRequests { requests };
+        if is_ttl_imme_table(self.table_info) {
```
killme2008, line 50: I think `is_ttl_imme_table` should be renamed too, and can be moved into `TableInfo` struct.

tests/cases/standalone/common/ttl/ttl_imme.result
```diff
@@ -0,0 +1,57 @@
+CREATE TABLE test_ttl(ts TIMESTAMP TIME INDEX, val INT) WITH (ttl = 'instant');
```
killme2008, line 1: Don't forget to rename the file.

tests/cases/standalone/common/ttl/ttl_imme.result
```diff
@@ -0,0 +1,57 @@
...
+|          |   TIME INDEX ("ts")                     |
+|          | )                                       |
+|          |                                         |
+|          | ENGINE=mito                             |
+|          | WITH(                                   |
+|          |   ttl = 'instant'                       |
+|          | )                                       |
++----------+-----------------------------------------+
+
+INSERT INTO test_ttl VALUES
```
killme2008, line 22: It's better to insert more values. And test the instant table with the flow task.

discord9:
added a fix for alter ttl to instant should remove expired ssts too

src/operator/src/insert.rs
```diff
@@ -91,6 +94,20 @@ impl AutoCreateTableType {
...
 }
 
+/// Split insert requests into normal and instant requests.
+///
+/// Where instant requests are requests with ttl=instant,
+/// and normal requests are requests with ttl set to other values.
+///
+/// This is used to split requests for different processing.
+#[derive(Clone)]
+pub struct InstantInsertRequests {
```
evenyag, line 104: The struct also has normal requests, maybe we should and normal to the struct name.

src/operator/src/insert.rs
```diff
@@ -558,7 +604,7 @@ impl Inserter {
             }
         }
 
-        Ok(table_name_to_ids)
+        Ok((table_name_to_ids, instant_table_ids))
```
evenyag, line 607: Will we auto create an instant table in the future? Do we need to check this after L596?

src/operator/src/req_convert/insert/stmt_to_region.rs
```diff
@@ -134,7 +135,18 @@ impl<'a> StatementToRegion<'a> {
...
+            InstantInsertRequests {
+                normal_requests: Default::default(),
+                instant_requests: requests,
+            }
+        } else {
+            InstantInsertRequests {
+                normal_requests: requests,
+                instant_requests: Default::default(),
+            }
+        })
```
evenyag, line 149: non-blocking: Use the same style.
```suggestion
        if table_info.is_ttl_instant_table() {
            Ok(InstantInsertRequests {
                normal_requests: Default::default(),
                instant_requests: requests,
            })
        } else {
            Ok(InstantInsertRequests {
                normal_requests: requests,
                instant_requests: Default::default(),
            })
        }
```

src/session/Cargo.toml
evenyag, line 1: Can we revert this change?

tests/cases/standalone/common/flow/flow_basic.result
```diff
@@ -321,16 +338,6 @@ where
 
 Affected Rows: 0
 
-SHOW CREATE FLOW filter_numbers_basic;
```
evenyag, line 324: Why do we remove this show create flow test?

tests/cases/standalone/common/ttl/ttl_instant.sql
```diff
@@ -0,0 +1,145 @@
...
+       test_ttl
+VALUES
+       (now(), 1),
+       (now(), 2),
+       (now(), 3);
+
+SELECT
+       val
+from
+       test_ttl;
```
evenyag, line 19: non-blocking: Consider adding order val or sqlness sort result to ensure result is stable.

tests/cases/standalone/common/ttl/ttl_instant.sql
```diff
@@ -0,0 +1,145 @@
...
+
+ALTER TABLE
+       test_ttl
+SET
+       ttl = '1s';
```
evenyag, line 121: We already test ttl 1s above, do we need to test this again?

evenyag (approved):
Others LGTM. Most comments are non-blocking.

evenyag:
Don't forget to update the PR title and description.

killme2008 (approved):
LGTM

src/operator/src/insert.rs
discord9: > Will we auto create an instant table in the future? Do we need to check this after L596?

I don't think we should auto create an instant table for it's a very unexpected behavior and should be set by user explictly anyway

tests/cases/standalone/common/ttl/ttl_instant.sql
discord9: > We already test ttl 1s above, do we need to test this again?

to makesure alter back and forth from duration to/from instant wouldn't break anything

src/common/time/src/ttl.rs
```diff
@@ -75,15 +77,6 @@ impl TimeToLive {
         })
     }
 
-    /// Print TimeToLive as string
-    pub fn to_string_opt(&self) -> Option<String> {
-        match self {
-            TimeToLive::Instant => Some(INSTANT.to_string()),
-            TimeToLive::Duration(d) => Some(humantime::format_duration(*d).to_string()),
```
killme2008: Could `duration.to_string()` replace `humantime::format_duration(*d).to_string()`?

src/operator/src/insert.rs
```diff
@@ -101,7 +101,7 @@ impl AutoCreateTableType {
 ///
 /// This is used to split requests for different processing.
 #[derive(Clone)]
-pub struct InstantInsertRequests {
+pub struct InstantOrNormalInsertRequests {
```
killme2008, line 104: Should be `and`?

src/operator/src/insert.rs
```diff
@@ -91,6 +94,20 @@ impl AutoCreateTableType {
...
 }
 
+/// Split insert requests into normal and instant requests.
+///
+/// Where instant requests are requests with ttl=instant,
+/// and normal requests are requests with ttl set to other values.
+///
+/// This is used to split requests for different processing.
+#[derive(Clone)]
+pub struct InstantOrNormalInsertRequests {
```
evenyag, line 104: Does this struct only have normal or instant requests? If not, may be we should use `InstantAndNormal` instead of `Or`.

tests/cases/standalone/common/ttl/ttl_instant.sql
```diff
@@ -0,0 +1,166 @@
...
+ADMIN compact_table('test_ttl');
+
+SELECT
+       val
+from
+       test_ttl
+ORDER BY
+       val;
+
+-- to makesure alter back and forth from duration to/from instant wouldn't break anything
```
evenyag, line 134:
```suggestion
-- to make sure alter back and forth from duration to/from instant wouldn't break anything
```

src/common/time/src/ttl.rs
discord9: > Could `duration.to_string()` replace `humantime::format_duration(*d).to_string()`?

` humantime::Duration` wrapper std::Duration and impl `Display`(which `std::Duration` didn't not because too many date format), but `ttl` store std::Duration for simplicity and `humantime::Duration` missing some derived traits

src/common/time/src/error.rs
```diff
@@ -93,12 +93,20 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
+
+    #[snafu(display("Failed to parse duration: {raw:?}"))]
+    ParseDuration {
+        raw: humantime::DurationError,
+        #[snafu(implicit)]
+        location: Location,
+    },
```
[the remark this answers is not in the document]
discord9: > 

humantime::DurationError didn't impl StackedError(and wouldn't make sense to, because it's the last stack frame on the error frame)

src/common/time/src/ttl.rs
killme2008: Yes. So that's my actual question. If you remove `to_string_opt`, are you sure all the cases can be replaced by `ttl.map(|ttl| ttl.to_string())` ?

discord9:
should `SET DATABASE 'ttl'='instant'` be forbidden? Or should on insert also check database ttl?

killme2008:
> should `SET DATABASE 'ttl'='instant'` be forbidden? Or should on insert also check database ttl?

Let's prohibit the database from setting TTL to `instant`, as it is illogical.

src/common/time/src/ttl.rs
discord9: > Yes. So that's my actual question. If you remove `to_string_opt`, are you sure all the cases can be replaced by `ttl.map(|ttl| ttl.to_string())` ?

impl ttl for enum like this so it should be ok?, and made sure that enum' variant is never access directly outside `ttl.rs` file:
```
impl Display for DatabaseTimeToLive {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            DatabaseTimeToLive::Forever => write!(f, "{}", FOREVER),
            DatabaseTimeToLive::Duration(d) => write!(f, "{}", humantime::Duration::from(*d)),
        }
    }
}
```

Issue #5052, opened 2024-11-26
Performance Regression in TSBS Query Time from v0.9 to v0.10


discord9:
### What type of enhancement is this?

Performance

### What does the enhancement do?

__edited__: could be false alarm since the old 0.9 tsbs tests is on a different machine(although with similar spec), still working to confirm if test on the same machine can reproduce this
__edited 2__: add v0.9.5 and v0.9.4 tsbs(on same machine) results, 
__edited 3__: after compare more version, it seems ok, but still `groupby-orderby-limit` need to look into since it cost more time somehow

The TSBS conducts a series of queries on the database, and our comparison of test results from versions 0.9 to 0.10 indicates some performance regression. Details are as follows:

![gtdb0 9vs0 10_p1](https://github.com/user-attachments/assets/d93f2666-8d59-4e33-9c2f-d1546bf1aa48)
![gtdb0 9vs0 10_p2](https://github.com/user-attachments/assets/8be7ab2f-663f-4a72-86bb-c82973abc90d)

As shown above, many of the GROUP BY queries are slower compared to version 0.9. The most significant regression is observed in the lastpoint query, which is five times slower in version 0.10 than in 0.9.

### Implementation challenges

We suspect that changes in the way sequential scans are handled between versions 0.9 and 0.10 may have caused these performance issues. Below are the metrics collected after repeatedly testing TSBS on version 0.10 of the database, obtained from `curl localhost:4000/metrics`:
[metrics.log](https://github.com/user-attachments/files/17914621/metrics.log)

discord9:
we could set up a simple script running TSBS on a daily or weekly basis to detect such perf regression just like sqlite's speedtest?

MichaelScofield:
[...]

PR #5090, opened 2024-12-03
fix!: fix regression caused by unbalanced partitions and splitting ranges

evenyag:

- https://github.com/GreptimeTeam/greptimedb/issues/5052
- https://github.com/GreptimeTeam/greptimedb/issues/4828

Fix several issues in SeqScan related to #5052
- Avoids pushing row group indices before split parts. It may cause the merge reader to merge all row groups of the same file, which is unexpected
    - **Before this PR, compaction will load all row groups into memory.** This is a regression introduced in https://github.com/GreptimeTeam/greptimedb/pull/4809
- Adds the number of row groups to the `SourceIndex` so we can know how to split the source.
- Makes the number of rows in partitions close to each other if possible and doesn't always create `target_partitions` partitions so the query engine can repartition inputs.

## Checklist


config/datanode.example.toml
```diff
@@ -492,12 +492,6 @@ experimental_write_cache_ttl = "8h"
 ## Buffer size for SST writing.
 sst_write_buffer_size = "8MB"
 
-## Parallelism to scan a region (default: 1/4 of cpu cores).
```
killme2008, line 495: So it's another breaking change.

src/query/src/optimizer/parallelize_scan.rs
```diff
@@ -109,21 +110,67 @@ impl ParallelizeScan {
...
+        mut ranges: Vec<PartitionRange>,
         expected_partition_num: usize,
     ) -> Vec<Vec<PartitionRange>> {
-        let actual_partition_num = expected_partition_num.min(ranges.len()).max(1);
+        if ranges.is_empty() {
+            // Returns a single partition with no range.
+            return vec![vec![]];
+        }
+
+        // Sort ranges by number of rows in descending order.
```
killme2008, line 127: Consider implementing a fast path for `ranges.len() == 1` to skip the costly calculation below.

src/mito2/src/read/seq_scan.rs
```diff
@@ -112,23 +105,16 @@ impl SeqScan {
...
-        )
-        .await?;
+        let reader =
+            Self::build_all_merge_reader(&self.stream_ctx, partition_ranges, &part_metrics).await?;
         Ok(Box::new(reader))
     }
 
     /// Builds a merge reader that reads all data.
+    /// Callers MUST not split ranges before calling this method.
     async fn build_all_merge_reader(
```
killme2008, line 115: Is it only used for compaction reader? If so , i think it must be renamed too.

src/mito2/src/read/scan_region.rs
```diff
@@ -104,6 +96,17 @@ impl Scanner {
             Scanner::Unordered(unordered_scan) => unordered_scan.input().file_ids(),
         }
     }
+
+    /// Sets the target partitions for the scanner. It can controls the parallelism of the scanner.
+    pub(crate) fn set_target_partitions(&mut self, target_partitions: usize) {
+        use store_api::region_engine::{PrepareRequest, RegionScanner};
```
killme2008, line 102: Move the `use` out of the function.

src/query/src/optimizer/parallelize_scan.rs
fengjiachun: The `sort_by` function also takes this internally (fast path)

src/mito2/src/read/scan_region.rs
evenyag: This method is under `#[cfg(test)]` so moving the `use` to outside will cause `unused` warning.

config/datanode.example.toml
evenyag: Yes. It's still compatible with the old toml file.

killme2008 (approved):
LGTM

PR #5093, opened 2024-12-04
feat!: enable read cache and write cache when using remote object stores

### LINKED ISSUE #4813: Enable read/write cache by default if using remote object stores

Performance

The write cache and object store read cache significantly affect the performance. We should enable these caches if the storage is remote object stores.

### Implementation challenges

There are several challenges:
- The default path of the cache.
- The default capacity of the cache. We may infer this from memory or disk. We may not always get the correct or proper disk size. Using memory size may be easier.
- The mito engine may need to deal with multiple storages.

killme2008:

Close #4813 

* Enable read and write cache by default for remote object storages.  
* Set the default read cache path to `{data_home}/object_cache/read` and create subdirectories for multiplex storages by their names.  
* Set the default write cache path to `{data_home}/object_cache/write`, as the write cache key, `IndexKey`, includes `region_id`, eliminating the need for subdirectories for multiplex storages.
* Change the default cache capacity from 1Gib to 5Gib.

@evenyag Please take a look.

## Checklist

## Release Notes

- **New Features**
	- Enhanced configuration documentation for object storage and caching.
	- Added new test functions for OpenTelemetry Protocol (OTLP) metrics, traces, and logs.
	- Introduced a new error variant for directory creation failures.

- **Improvements**
	- Default cache size increased from 1GiB to 5GiB for object storage.
	- Clarified and updated configuration paths for experimental write cache.
	- Improved error handling and logging for cache directory creation.

- **Bug Fixes**
	- Adjusted cache path handling for S3 storage type to prevent unintended caching.


killme2008:
@coderabbitai review

src/datanode/src/config.rs
```diff
@@ -60,6 +60,14 @@ impl ObjectStoreConfig {
         }
     }
 
+    /// Returns true when it's a remote object storage such as AWS s3 etc.
+    pub fn is_object_storage(&self) -> bool {
+        matches!(
+            self,
+            Self::S3(_) | Self::Oss(_) | Self::Azblob(_) | Self::Gcs(_)
+        )
+    }
```
waynexia, line 69: Nit: how about only excluding `Self::File`? This is not likely to change compared to the OSS services.

waynexia (approved):
Others LGTM.

Is it necessary to change the default path? I'm ok with the new naming but since it's a breaking change

killme2008:
> Others LGTM.
> 
> Is it necessary to change the default path? I'm ok with the new naming but since it's a breaking change

I was hesitant as well, but I believe that failing to standardize the paths for read and write caches will lead to growing chaos in `data_home`. It's wiser to implement these breaking changes sooner rather than later.

WenyXu (approved):

src/datanode/src/config.rs
WenyXu: +1

killme2008:
@evenyag Please help to take a look

src/datanode/src/config.rs
```diff
@@ -60,6 +60,14 @@ impl ObjectStoreConfig {
         }
     }
 
+    /// Returns true when it's a remote object storage such as AWS s3 etc.
+    pub fn is_object_storage(&self) -> bool {
+        matches!(
+            self,
+            Self::S3(_) | Self::Oss(_) | Self::Azblob(_) | Self::Gcs(_)
+        )
```
evenyag, line 68:
```suggestion
    pub fn is_object_storage(&self) -> bool {
        !matches!(self, Self::File(_))
```

src/datanode/src/datanode.rs
```diff
@@ -399,10 +399,17 @@ impl DatanodeBuilder {
         for engine in &opts.region_engine {
             match engine {
                 RegionEngineConfig::Mito(config) => {
+                    let mut config = config.clone();
+                    if opts.storage.is_object_storage() {
+                        // Enable the write cache when setting object storage
+                        config.enable_experimental_write_cache = true;
+                        info!("Configured 'enable_experimental_write_cache=true' for mito engine.");
+                    }
```
evenyag, line 407: Can we do this inside `Self::build_mito_engine()`?

src/mito2/src/worker.rs
```diff
@@ -373,6 +373,12 @@ async fn write_cache_from_config(
     // TODO(yingwen): Remove this and document the config once the write cache is ready.
     warn!("Write cache is an experimental feature");
 
+    tokio::fs::create_dir_all(Path::new(&config.experimental_write_cache_path))
+        .await
+        .context(CreateDirSnafu {
+            dir: &config.experimental_write_cache_path,
+        })?;
```
evenyag, line 380: nitpick: Only do this when `config.enable_experimental_write_cache` is true

killme2008: Yes, it's checked above.

src/mito2/src/config.rs
```diff
@@ -233,7 +234,8 @@ impl MitoConfig {
 
         // Sets write cache path if it is empty.
         if self.experimental_write_cache_path.is_empty() {
```
evenyag, line 236:
```suggestion
        if self.experimental_write_cache_path.trim().is_empty() {
```

PR #5081, opened 2024-12-02
feat: add more transaction related statement for postgres interface

sunng87:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

This patch adds transaction related statement for postgres interface:

- start transaction [isolation level ...]
- abort transaction
- commit transaction

## Checklist

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.

Issue #5077, opened 2024-12-01
Support for `postgres_fdw` in GreptimeDB to enable integration with PostgreSQL


frin1:
### What problem does the new feature solve?

The current PostgreSQL protocol implementation in GreptimeDB does not fully support Foreign Data Wrappers (FDWs), particularly the `postgres_fdw` extension. This limitation makes it challenging to integrate GreptimeDB with existing PostgreSQL databases. By adding this support, it would be easier to use PostgreSQL as a metadata store while querying and interacting with GreptimeDB for time-series data. This would allow seamless integration of GreptimeDB in PostgreSQL-based systems and enhance its adoption in existing PostgreSQL workflows.

### What does the feature do?

This feature would enable `postgres_fdw` to interact with GreptimeDB by:
- Supporting required PostgreSQL protocol features to make `postgres_fdw` functional.
- Allowing users to define foreign tables in PostgreSQL that link directly to tables in GreptimeDB.

The functionality would allow users to query GreptimeDB tables (e.g. `numbers`) directly from PostgreSQL.

### Implementation challenges

1. **Protocol Compliance**: Ensuring that GreptimeDB's PostgreSQL protocol implementation meets the requirements of `postgres_fdw`.
2. **Query Translation**: Mapping queries issued via `postgres_fdw` into GreptimeDB's query execution engine. Differences in SQL dialects or functions might need careful handling.
3. **Performance Optimization**: Avoiding inefficiencies caused by translating queries or transferring large volumes of time-series data between systems.
4. **Testing and Compatibility**: Ensuring that the feature works with common PostgreSQL versions and setups where `postgres_fdw` is used.

killme2008:
@frin1 It's interesting! I don't known `postgres_fdw` before.

@sunng87 Is it posibble to make `pgwire` supporting `postgres_fdw`?

sunng87:
Thank you for the information. I just tried foreign data wrapper by myself. At wire protocol level we are fully compatible. It's mostly unsupported SQL statements:

- some postgres specific transaction statements and we can bypass them with #5081 
- cursor: the foreign data wrapper uses a cursor mechanism to fetch data from remote server https://www.postgresql.org/docs/current/plpgsql-cursors.html

The statements ran by fdw are like:

- `DECLARE c1 CURSOR FOR\nSELECT id FROM public.tree`
- `FETCH 100 FROM c1`
- `CLOSE c1`

[...]

PR #5094, opened 2024-12-04
feat: add cursor statements

sunng87:

This patch is to address #5077 

In this patch, I added initial support for postgres cursor related statements:

- `DECLARE $cursor CURSOR FOR $query`
- `FETCH $size FROM $cursor`
- `CLOSE $cursor`

This patch, together with #5081 , will hopefully enable us as a postgresql foreign data wrapper compatible server.

Some utility functions are added to recordbatch module, allows to slice and merge recordbatch(es). 

~~Tests to be added.~~

## Checklist


src/sql/src/parsers/cursor_parser.rs
```diff
@@ -0,0 +1,146 @@
...
+            .parser
+            .parse_object_name(false)
+            .context(error::SyntaxSnafu)?;
+        let _ = self
+            .parser
+            .expect_keywords(&[Keyword::CURSOR, Keyword::FOR]);
+
+        let mut is_select = false;
+        if let Token::Word(w) = self.parser.peek_token().token {
+            if w.keyword == Keyword::SELECT {
```
MichaelScofield, line 37: What about [`WITH`](https://docs.greptime.cn/en/reference/sql/with), it's select too.

src/operator/src/statement/cursor.rs
```diff
@@ -0,0 +1,85 @@
...
+        let cursor_name = declare_cursor.cursor_name.to_string();
+        let query_stmt = Statement::Query(declare_cursor.query);
+
+        let output = self
+            .plan_exec(QueryStatement::Sql(query_stmt), query_ctx.clone())
+            .await?;
+        match output.data {
+            OutputData::RecordBatches(rb) => {
+                let rbs = rb.as_stream();
+                query_ctx.insert_cursor(cursor_name, RecordBatchStreamCursor::new(rbs));
```
MichaelScofield, line 42: I think for the best of user experience, we should return "cursor already exists" error if the same cursor name was declared before, instead of replacing the old one here.

src/common/recordbatch/src/error.rs
```diff
@@ -168,6 +168,11 @@ pub enum Error {
         #[snafu(source)]
         error: tokio::time::error::Elapsed,
     },
+    #[snafu(display("RecordBatch slice index overflow"))]
+    RecordBatchSliceIndexOverflow {
```
killme2008, line 172: Please print the offset and limit to the error message.

src/operator/src/statement/cursor.rs
```diff
@@ -0,0 +1,93 @@
...
+use session::context::QueryContextRef;
+use snafu::ResultExt;
+use sql::statements::cursor::{CloseCursor, DeclareCursor, FetchCursor};
+use sql::statements::statement::Statement;
+
+use crate::error::{self, Result};
+use crate::statement::StatementExecutor;
+
+impl StatementExecutor {
+    pub(super) async fn declare_cursor(
```
killme2008, line 28: Adds tracing for these functions.
```suggestion
    #[tracing::instrument(skip_all)]
    pub(super) async fn declare_cursor(
```

src/operator/src/statement/cursor.rs
```diff
@@ -0,0 +1,93 @@
...
+                query_ctx.insert_cursor(cursor_name, RecordBatchStreamCursor::new(rbs));
+            }
+            OutputData::Stream(rbs) => {
+                query_ctx.insert_cursor(cursor_name, RecordBatchStreamCursor::new(rbs));
+            }
+            OutputData::AffectedRows(_) => error::NotSupportedSnafu {
+                feat: "Non-query statement on cursor",
+            }
+            .fail()?,
+        }
```
killme2008, line 59: Better to add some debug/info logs for cusor management.

src/sql/src/parsers/cursor_parser.rs
```diff
@@ -0,0 +1,149 @@
...
+            is_select,
+            error::InvalidSqlSnafu {
+                msg: "Expect select query in cursor statement".to_string(),
+            }
+        );
+
+        let query_stmt = self.parse_query()?;
+        match query_stmt {
+            Statement::Query(query) => Ok(Statement::DeclareCursor(DeclareCursor {
+                cursor_name,
```
killme2008, line 54: Called `Self::canonicalize_object_name` for cusor name before returning.

src/sql/src/parsers/cursor_parser.rs
```diff
@@ -0,0 +1,149 @@
...
+            .context(error::SyntaxSnafu)?;
+        let _ = self.parser.parse_keyword(Keyword::FROM);
+
+        let cursor_name = self
+            .parser
+            .parse_object_name(false)
+            .context(error::SyntaxSnafu)?;
+
+        Ok(Statement::FetchCursor(FetchCursor {
+            cursor_name,
```
killme2008, line 79: The same suggestion as above

src/sql/src/statements/cursor.rs
```diff
@@ -0,0 +1,55 @@
...
+// limitations under the License.
+
+use std::fmt::Display;
+
+use sqlparser::ast::ObjectName;
+use sqlparser_derive::{Visit, VisitMut};
+
+use super::query::Query;
+
+#[derive(Debug, Clone, PartialEq, Eq, Visit, VisitMut)]
```
killme2008, line 22: Always add documents for pub strucutures.

src/sql/src/statements/cursor.rs
```diff
@@ -0,0 +1,55 @@
...
+pub struct DeclareCursor {
+    pub cursor_name: ObjectName,
+    pub query: Box<Query>,
+}
+
+impl Display for DeclareCursor {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "DECLARE {} CURSOR FOR {}", self.cursor_name, self.query)
```
killme2008, line 30: Need to test the display result. Assert them in parser test. Just like

```rust
if let Statement::CloseCursor(cc) = &result[0] {           
            assert_eq!(sql , cc.to_string());
        } else {
            panic!("Unexpected statement")
        }
````

File: src/common/recordbatch/src/cursor.rs [hunk 10]
```diff
@@ -0,0 +1,173 @@
...
+
+            // If we still have no batch after attempting to fetch
+            let current_batch = match &inner.current_batch {
+                Some(batch) => batch,
+                None => break,
+            };
+
+            // Calculate how many rows we can take from this batch
+            let rows_to_take_from_batch = remaining_rows_to_take
+                .min(inner.total_rows_in_current_batch - inner.current_row_index);
```
killme2008, line 81: May be impossible, but for safety.

```suggestion
                .min(inner.total_rows_in_current_batch - inner.current_row_index).max(0);
```

sunng87: No worry, these indexes are carefully managed so `remaining_row_to_take` and `inner.total_rows_in_current_batch - inner.current_row_index` will not be less than 0 unless `recordbatch.num_rows` returns negative number which is impossible.

Even if we assigned `0` here it causes other issues like busy loop or wrong data.

src/operator/src/statement/cursor.rs
sunng87: Added in insert_cursor to check connection opened cursor count and warns when there are more than 10 opened

killme2008 (approved):
LGTM

src/operator/src/statement/cursor.rs
```diff
@@ -0,0 +1,98 @@
...
+        Ok(Output::new_with_affected_rows(0))
+    }
+
+    #[tracing::instrument(skip_all)]
+    pub(super) async fn fetch_cursor(
+        &self,
+        fetch_cursor: FetchCursor,
+        query_ctx: QueryContextRef,
+    ) -> Result<Output> {
+        let cursor_name = fetch_cursor.cursor_name.to_string();
```
MichaelScofield, line 73: Seems `cursor_name` doesn't have to `to_string` here.

MichaelScofield (approved):

src/operator/src/statement/cursor.rs
sunng87: cursor_name is an `ObjectName`. 

but perhaps we can convert it to string in parser. I'm not sure if that's good practice because I still see a lot of usage of `ObjectName` in parser types.

PR #5112, opened 2024-12-06
feat(flow): check sink table mismatch on flow creation

discord9:

check sink table mismatch on flow creation

- on flow creation, if sink table exist, check if schema's datatype is mismatched
- if sink table not exist when creating flow, create sink table with inferred schema

## Checklist


tests/cases/standalone/common/flow/flow_basic.sql
```diff
@@ -678,3 +711,128 @@ DROP FLOW calc_android_log_abnormal;
...
+  total_connect,
+  conection_rate,
+  avg_first_frame_time,
+  max_first_frame_time,
+  ok_conection_rate,
+  record_time_window FROM live_connection_statistics_detail;
+
+DROP FLOW live_connection_aggregation_detail;
+DROP TABLE live_connection_log;
+DROP TABLE live_connection_statistics_detail;
```
waynexia, line 838: nit: new line

src/common/error/src/lib.rs
```diff
@@ -18,9 +18,22 @@ pub mod ext;
...
 // please define in `src/servers/src/http/header.rs`.
 pub const GREPTIME_DB_HEADER_ERROR_CODE: &str = "x-greptime-err-code";
 pub const GREPTIME_DB_HEADER_ERROR_MSG: &str = "x-greptime-err-msg";
+
+/// Create a http header map from error code and message.
+/// using `GREPTIME_DB_HEADER_ERROR_CODE` and `GREPTIME_DB_HEADER_ERROR_MSG` as keys.
+pub fn from_err_code_msg_to_header(
+    code: impl Into<HeaderValue>,
+    msg: impl Into<HeaderValue>,
+) -> HeaderMap {
```
waynexia, line 34: Does `TypedHeader` suit this case better? https://docs.rs/axum/0.6.19/axum/struct.TypedHeader.html

tests/cases/standalone/common/flow/flow_auto_sink_table.sql
```diff
@@ -0,0 +1,58 @@
...
+-- SQLNESS ARG restart=true
+SHOW CREATE FLOW test_numbers_basic;
+
+SHOW CREATE TABLE out_num_cnt_basic;
+
+DROP FLOW test_numbers_basic;
+
+DROP TABLE numbers_input_basic;
+
+DROP TABLE out_num_cnt_basic;
```
fengjiachun, line 58: Add an empty line

src/common/error/src/lib.rs
discord9: > Does `TypedHeader` suit this case better? https://docs.rs/axum/0.6.19/axum/struct.TypedHeader.html

I can't find it in newer version of axum, so no?:
https://docs.rs/axum/0.7.9/axum/?search=TypedHeader

killme2008:
Changing the status and code header will result in many SQL tests failing due to the modified error message.

Focus on your needs rather than altering the underlying code.

@discord9

killme2008:
As you can see the results in https://github.com/GreptimeTeam/greptimedb/actions/runs/12377853661/job/34548598017?pr=5112

killme2008:
We don't need to send the stack trace to the frontend; users only require the root cause message and status code. For further investigation, the stack trace and detailed error information can be found in the data nodes or flow nodes.

discord9:
revert back those commit and not use typed header for now and separate to another PR once this is merged

fengjiachun (approved):
Rest LGTM

waynexia (approved):

PR #5126, opened 2024-12-09
test: sqlness upgrade compatibility tests

discord9:

Add a way to test inter-version compatibility in sqlness

- add a `-- SQLNESS ARG version=v0.x.x` which will restart greptime from given binary in a directory
- add a simple poc sqlness test, to run those tests
- add ability to pull different version binary from github release when encounter `version=xxx`, default to pull, but can set to not pull and manually download it. 
- more sqlness tests can be added later in separate PR to prevent bloat
- add ability to setup a standalone etcd for metasrv metadata storage so restarting metasrv wouldn't clean up metadata

to playaround with this test, run:
 ```bash
cargo sqlness -c $(pwd)/tests/upgrade-compat --store-addrs "0.0.0.0:2379" --setup-etcd 
``` 
(make sure to use absolute path for case dir, the `-c` option, can also add `--preserve-state` to look into `/tmp/sqlness*****/greptime-sqlness-standalone.log` to make sure it actually switched version, note this require docker and create a `etcd` docker container for metadata storage during testing)

## Summary
Basically just restart greptime using another binary
Well this can't currently test http api, but I intended to add ability to invoke bash in sqlness tests,
so one can just call `curl` in sqlness tests
## Checklist

### CODE REVIEW

File: tests/runner/src/env.rs [hunk 1]
```diff
@@ -693,8 +780,28 @@ impl GreptimeDB {
 impl Database for GreptimeDB {
     async fn query(&self, ctx: QueryContext, query: String) -> Box<dyn Display> {
         if ctx.context.contains_key("restart") && self.env.server_addrs.server_addr.is_none() {
-            self.env.restart_server(self).await;
+            self.env.restart_server(self, false).await;
+        } else if let Some(version) = ctx.context.get("version") {
+            if self.env.old_bins_dir.lock().unwrap().is_none() {
+                // save old bins dir
+                *self.env.old_bins_dir.lock().unwrap() = self.env.bins_dir.lock().unwrap().clone();
```
MichaelScofield, line 787: Seems `old_bins_dir` is always latest bins dir. I suggest using a hashmap to manage bins dirs across versions, to make codes more understandable.

fengjiachun (approved):

MichaelScofield (approved):

Issue #5080, opened 2024-12-02
Create inverted index for JSON fields


killme2008:
### What type of enhancement is this?

Performance, User experience

### What does the enhancement do?

We should support creating an inverted index for `context.trace_id` and `context.span_id` in the `context` column from OpenTelemetry Traces(https://opentelemetry.io/docs/concepts/signals/traces/):

```json
"context": {
    "trace_id": "5b8aa5a2d2c872e8321cf37308d69df2",
    "span_id": "051581bf3cb55c13"
  }
```

Additionally, enabling the query `where context.trace_id = '5b8aa5a2d2c872e8321cf37308d69df2'` would allow the index to optimize query performance.

CookiePieWw:
Seems duplicated to #4849? We can close one of them.

killme2008:
It will be implemented in new JSON type.

PR #5131, opened 2024-12-10
feat: Alter inverted index

lyang24:

This pr enables `alter table x modify column set/ unset inverted index syntax`.
1. modify index proto has been refactored.
2. parse support the new syntax with refactoring.
3. metadata update with alter inverted index.

## Checklist

### DIFF
--- a/src/datatypes/src/schema/column_schema.rs
+++ b/src/datatypes/src/schema/column_schema.rs
@@ -158,11 +158,22 @@ impl ColumnSchema {
         self
     }
 
-    pub fn set_inverted_index(mut self, value: bool) -> Self {
-        let _ = self
-            .metadata
-            .insert(INVERTED_INDEX_KEY.to_string(), value.to_string());
-        self
+    pub fn with_inverted_index(&mut self, value: bool) {
+        match value {
+            true => {
+                self.metadata
+                    .insert(INVERTED_INDEX_KEY.to_string(), value.to_string());
+            }
+            false => {
+                self.metadata.remove(INVERTED_INDEX_KEY);
+            }
+        }
+    }
+
+    // Put a placeholder to invalidate schemas.all(!has_inverted_index_key).
+    pub fn insert_inverted_index_placeholder(&mut self) {
+        self.metadata
+            .insert(INVERTED_INDEX_KEY.to_string(), "".to_string());
     }
 
     pub fn is_inverted_indexed(&self) -> bool {
--- a/src/store-api/src/metadata.rs
+++ b/src/store-api/src/metadata.rs
@@ -34,7 +34,10 @@ use serde::{Deserialize, Deserializer, Serialize};
 use snafu::{ensure, Location, OptionExt, ResultExt, Snafu};
 
 use crate::codec::PrimaryKeyEncoding;
-use crate::region_request::{AddColumn, AddColumnLocation, AlterKind, ModifyColumnType};
+use crate::region_request::{
+    AddColumn, AddColumnLocation, AlterKind, ApiSetIndexOptions, ApiUnsetIndexOptions,
+    ModifyColumnType,
+};
 use crate::storage::consts::is_internal_column;
 use crate::storage::{ColumnId, RegionId};
 
@@ -570,13 +573,23 @@ impl RegionMetadataBuilder {
             AlterKind::AddColumns { columns } => self.add_columns(columns)?,
             AlterKind::DropColumns { names } => self.drop_columns(&names),
             AlterKind::ModifyColumnTypes { columns } => self.modify_column_types(columns),
-            AlterKind::SetColumnFulltext {
-                column_name,
-                options,
-            } => self.change_column_fulltext_options(column_name, true, Some(options))?,
-            AlterKind::UnsetColumnFulltext { column_name } => {
-                self.change_column_fulltext_options(column_name, false, None)?
-            }
+            AlterKind::SetIndex { options } => match options {
+                ApiSetIndexOptions::Fulltext {
+                    column_name,
+                    options,
+                } => self.change_column_fulltext_options(column_name, true, Some(options))?,
+                ApiSetIndexOptions::Inverted { column_name } => {
+                    self.change_column_inverted_index_options(column_name, true)?
+                }
+            },
+            AlterKind::UnsetIndex { options } => match options {
+                ApiUnsetIndexOptions::Fulltext { column_name } => {
+                    self.change_column_fulltext_options(column_name, false, None)?
+                }
+                ApiUnsetIndexOptions::Inverted { column_name } => {
+                    self.change_column_inverted_index_options(column_name, false)?
+                }
+            },
             AlterKind::SetRegionOptions { options: _ } => {
                 // nothing to be done with RegionMetadata
             }
@@ -682,6 +695,19 @@ impl RegionMetadataBuilder {
                 column_meta.column_schema.data_type = target_type;
             }
         }
+    }
+
+    fn change_column_inverted_index_options(
+        &mut self,
+        column_name: String,
+        value: bool,
+    ) -> Result<()> {
+        for column_meta in self.column_metadatas.iter_mut() {
+            if column_meta.column_schema.name == column_name {
+                column_meta.column_schema.with_inverted_index(value)
+            }
+        }
+        Ok(())
     }
 
     fn change_column_fulltext_options(
@@ -1381,12 +1407,14 @@ mod test {
 
         let mut builder = RegionMetadataBuilder::from_existing(metadata);
         builder
-            .alter(AlterKind::SetColumnFulltext {
-                column_name: "b".to_string(),
-                options: FulltextOptions {
-                    enable: true,
-                    analyzer: datatypes::schema::FulltextAnalyzer::Chinese,
-                    case_sensitive: true,
+            .alter(AlterKind::SetIndex {
+                options: ApiSetIndexOptions::Fulltext {
+                    column_name: "b".to_string(),
+                    options: FulltextOptions {
+                        enable: true,
+                        analyzer: datatypes::schema::FulltextAnalyzer::Chinese,
+                        case_sensitive: true,
+                    },
                 },
             })
             .unwrap();
@@ -1407,8 +1435,10 @@ mod test {
 
         let mut builder = RegionMetadataBuilder::from_existing(metadata);
         builder
-            .alter(AlterKind::UnsetColumnFulltext {
-                column_name: "b".to_string(),
+            .alter(AlterKind::UnsetIndex {
+                options: ApiUnsetIndexOptions::Fulltext {
+                    column_name: "b".to_string(),
+                },
             })
             .unwrap();
         let metadata = builder.build().unwrap();
--- a/src/table/src/metadata.rs
+++ b/src/table/src/metadata.rs
@@ -32,7 +32,10 @@ use store_api::region_request::{SetRegionOption, UnsetRegionOption};
 use store_api::storage::{ColumnDescriptor, ColumnDescriptorBuilder, ColumnId, RegionId};
 
 use crate::error::{self, Result};
-use crate::requests::{AddColumnRequest, AlterKind, ModifyColumnTypeRequest, TableOptions};
+use crate::requests::{
+    AddColumnRequest, AlterKind, ModifyColumnTypeRequest, SetIndexOptions, TableOptions,
+    UnsetIndexOptions,
+};
 
 pub type TableId = u32;
 pub type TableVersion = u64;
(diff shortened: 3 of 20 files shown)


src/store-api/src/metadata.rs
```diff
@@ -666,6 +679,26 @@ impl RegionMetadataBuilder {
...
     }
 
+    fn change_column_inverted_index_options(
+        &mut self,
+        column_name: String,
+        value: bool,
+    ) -> Result<()> {
+        for column_meta in self.column_metadatas.iter_mut() {
+            if column_meta.column_schema.name == column_name {
+                ensure!(
```
lyang24, line 689: discussion: i wonder if `String` is the only data type that support inverted index? what about json or other datatypes.

CookiePieWw: Currently we only support string type. Other datatypes like json may be supported in the future, and we can adjust this correspondingly. See #5080

CookiePieWw (commented):
Nice work! Left some nit comments.

I wonder if the behaviour of changing inverted index is the same as fulltext index: disabling it would only stop constructing new indexes, and the previously contructed indexes will still take effect? @zhongzc

waynexia:
FYI: https://github.com/GreptimeTeam/greptimedb/issues/5178

This issue is related to the PR, in case anyone might be interested in it.

tests/cases/standalone/common/alter/change_col_inverted_index.sql
```diff
@@ -0,0 +1,32 @@
+CREATE TABLE fox (
```
zhongzc, line 1: Waiting for #5297 to be merged, you can check the effect here by checking the results of `SHOW INDEX FROM <table>`.

src/table/src/metadata.rs
```diff
@@ -252,6 +270,66 @@ impl TableMeta {
...
+                table_name,
+            })?;
+
+        ensure!(
+            column.data_type.is_string(),
+            error::InvalidColumnOptionSnafu {
+                column_name,
+                msg: "INVERTED index only supports string type",
+            }
+        );
```
zhongzc, line 296: inverted index has no type restrictions

zhongzc:
> Thanks for your nice work! Left some nit comments.
> 
> I wonder if the behaviour of changing inverted index is the same as fulltext index: disabling it would only stop constructing new indexes, and the previously contructed indexes will still take effect? @zhongzc

Yes, we can think that alter index only modifies the column schema, leaving other things to the engine.

tests/cases/standalone/common/alter/change_col_inverted_index.sql
lyang24: added show index to the sqlness test

src/table/src/metadata.rs
```diff
@@ -252,6 +270,66 @@ impl TableMeta {
...
+
+        let mut columns = Vec::with_capacity(table_schema.column_schemas().len());
```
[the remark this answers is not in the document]
lyang24: thank you, I am trying to see what would go wrong here.

Are we worried about the case when remove invert index and accidentally build inverted index on primary keys? when we alter table remove inverted index it will set the inverted index value to false in column metadata - i think that could prevent the case as `is_inverted_indexed` check if the value is true.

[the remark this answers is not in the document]
zhongzc: Let me try to explain this.

When the condition `schemas.all(!has_inverted_index_key)` is met, the columns in the primary key implicitly have a shadow attribute: `is_inverted_index`. However, this attribute is not observable from the column schema.

Once `SET | UNSET INVERTED INDEX` disrupts the condition `schemas.all(!has_inverted_index_key)`, we need to materialize the shadow attribute.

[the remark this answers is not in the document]
zhongzc: My explanation might have become more abstract.

In practice, when we set an inverted index for a schema that previously had no columns with an inverted index set, we need to reset `inverted_indexed=true` for the columns in the primary key.

tests/cases/standalone/common/alter/change_col_inverted_index.sql
```diff
@@ -0,0 +1,36 @@
...
+    ts TIMESTAMP TIME INDEX,
+    fox STRING,
+);
```
zhongzc, line 4: You can add some case:

```
CREATE TABLE test_pk (..., foo STRING, bar INT, PRIMARY KEY (foo, bar));
SHOW INDEX;
UNSET foo;
SHOW INDEX;
UNSET bar;
SHOW INDEX;
SET foo;
SHOW INDEX;
```

src/datatypes/src/schema/column_schema.rs
```diff
@@ -149,6 +149,19 @@ impl ColumnSchema {
...
+            true => {
+                self.metadata
+                    .insert(INVERTED_INDEX_KEY.to_string(), value.to_string());
+            }
+
+            false => {
+                self.metadata.remove(INVERTED_INDEX_KEY);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
lyang24: there is another method has_inverted_index that is used from show index i will removing the value on unset so the show index will not produce confusing result.

src/table/src/metadata.rs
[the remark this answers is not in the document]
lyang24: I see the primary key will be inverted index if there are no inverted index on the table, so when we add a new inverted index we need to set the inverted index = true for primary keys as well.

tests/cases/standalone/common/alter/change_col_inverted_index.result
```diff
@@ -0,0 +1,183 @@
...
+| test_pk | 1          | PRIMARY, INVERTED INDEX | 3            | bar         | A         |             |          |        | YES  | greptime-primary-key-v1, greptime-inverted-index-v1 |         |               | YES     |            |
+| test_pk | 1          | PRIMARY, INVERTED INDEX | 2            | foo         | A         |             |          |        | YES  | greptime-primary-key-v1, greptime-inverted-index-v1 |         |               | YES     |            |
+| test_pk | 1          | TIME INDEX              | 1            | ts          | A         |             |          |        | NO   |                                                     |         |               | YES     |            |
++---------+------------+-------------------------+--------------+-------------+-----------+-------------+----------+--------+------+-----------------------------------------------------+---------+---------------+---------+------------+
+
+ALTER TABLE test_pk MODIFY COLUMN bar UNSET INVERTED INDEX;
+
+Affected Rows: 0
+
+SHOW INDEX FROM test_pk;
```
lyang24, line 156: not sure if this is the result we are expecting, nothing changes when we set/ unset inverted index on primary keys that comes with inverted index.

tests/cases/standalone/common/alter/change_col_inverted_index.sql
lyang24: added the case the results looks a bit strange to me. If its not what we expected i can debug more.

tests/cases/standalone/common/alter/change_col_inverted_index.result
CookiePieWw: The table has two primary keys with no explicit inverted index, so `is_inverted_indexed` and `has_inverted_index_key` will return false. So that `SHOW INDEX` finds no inverted index in the table and add primary keys into the inverted index sets, see:
https://github.com/GreptimeTeam/greptimedb/blob/812a775b3d02da1430364dfa7ef14d31bfab800a/src/catalog/src/system_schema/information_schema/key_column_usage.rs#L231-L234

Seems we need to set `inverted_indexed=true` for column `bar` when we unset inverted index for column `foo` to get the expected result. When it comes to unset inverted index in all columns, seems it's impossible? cc @zhongzc

lyang24: > Seems we need to set inverted_indexed=true for column bar when we unset inverted index for column foo to get the expected result. 
thanks for pointing out the logic, i had a follow up when we remove foo in the next iteration there will be no inverted index on the table - do we expecting primary key added as inverted index again?

CookiePieWw: In this case, we expect all primary key shows no inverted index in the query, ref https://github.com/GreptimeTeam/greptimedb/pull/5131#discussion_r1906577069:
```
CREATE TABLE test_pk (..., foo STRING, bar INT, PRIMARY KEY (foo, bar));
SHOW INDEX; // inverted indexes: [foo, bar]

UNSET foo;
SHOW INDEX; // inverted indexes: [bar]

UNSET bar;
SHOW INDEX; // inverted indexes: []

SET foo;
SHOW INDEX; // inverted indexes: [foo]
```

There're two cases:
1. All primary key are marked non-inverted index but we implicitly build inverted index for primary keys. We should show them as inverted index columns.
2. All primary key are marked non-inverted index, and it's explicitly unset by users. We don't need to show them as inverted index columns.

lyang24: thanks i missed the expected result from previous comment. I added logic to make the results converge but the logic is not straight forward :(

tests/cases/standalone/common/alter/change_col_inverted_index.sql
lyang24: modified the logic to produce expected results

src/table/src/metadata.rs
```diff
@@ -252,6 +270,77 @@ impl TableMeta {
...
+                // If user explicitly unset an inverted index in primary keys.
+                // We should invalidate the primary key as inverted index
+                // on the condition of schemas.all(!has_inverted_index_key).
+                if !value && self.primary_key_indices.contains(&i) {
+                    let mut new_column_schema = column_schema.clone();
+                    new_column_schema.insert_inverted_index_placeholder();
+                    columns.push(new_column_schema);
+                } else {
+                    let mut new_column_schema = column_schema.clone();
+                    new_column_schema.update_inverted_index(value);
```
zhongzc, line 307: nit: replace `set_inverted_index` and `update_inverted_index` with `with_inverted_index(mut self, value) -> Self`, and remove the key when `value` is `false`.

Then, use `insert_inverted_index_placeholder` instead here:
https://github.com/GreptimeTeam/greptimedb/blob/b3c8d8bd6da4e06758b24828362f8f6d3b3c7b15/src/sql/src/statements.rs#L490

zhongzc (approved):

zhongzc:
great job!

src/table/src/metadata.rs
lyang24: yes I combined the two functions. However, i have made it to take a mutable reference and not return a value to avoid some cloning.

for an example avoid cloning here
```
    fn change_column_inverted_index_options(
        &mut self,
        column_name: String,
        value: bool,
    ) -> Result<()> {
        for column_meta in self.column_metadatas.iter_mut() {
            if column_meta.column_schema.name == column_name {
                column_meta.column_schema.with_inverted_index(value)
            }
        }
        Ok(())
    }
```

lyang24: actually I was looking at this test case. There are primary keys and there is empty inverted index defined, should primary key has inverted index properties?

```
CREATE TABLE IF NOT EXISTS test_no_inverted_index (
    a STRING,
    b STRING SKIPPING INDEX,
    c DOUBLE,
    ts TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
    PRIMARY KEY(a, b),
    INVERTED INDEX(),
    TIME INDEX(ts)
);
```

CookiePieWw: I think we can follow behaviors of the recent rev, it appears the primary keys don't have inverted index:
```SQL
mysql> CREATE TABLE IF NOT EXISTS test_no_inverted_index ( a STRING, b STRING SKIPPING INDEX, c DOUBLE, ts TIMESTAMP DEFAULT CURRENT_TIMESTAMP, PRIMARY KEY(a, b), INVERTED INDEX(), TIME INDEX(ts) );
Query OK, 0 rows affected (0.02 sec)

mysql> show index from test_no_inverted_index;
+------------------------+------------+-------------------------+--------------+-------------+-----------+-------------+----------+--------+------+---------------------------------------------------+---------+---------------+---------+------------+
| Table                  | Non_unique | Key_name                | Seq_in_index | Column_name | Collation | Cardinality | Sub_part | Packed | Null | Index_type                                        | Comment | Index_comment | Visible | Expression |
+------------------------+------------+-------------------------+--------------+-------------+-----------+-------------+----------+--------+------+---------------------------------------------------+---------+---------------+---------+------------+
| test_no_inverted_index |          1 | PRIMARY                 |            1 | a           | A         |        NULL |     NULL |   NULL | YES  | greptime-primary-key-v1                           |         |               | YES     |       NULL |
| test_no_inverted_index |          1 | PRIMARY, SKIPPING INDEX |            2 | b           | A         |        NULL |     NULL |   NULL | YES  | greptime-primary-key-v1, greptime-bloom-filter-v1 |         |               | YES     |       NULL |
| test_no_inverted_index |          1 | TIME INDEX              |            1 | ts          | A         |        NULL |     NULL |   NULL | NO   |                                                   |         |               | YES     |       NULL |
+------------------------+------------+-------------------------+--------------+-------------+-----------+-------------+----------+--------+------+---------------------------------------------------+---------+---------------+---------+------------+
3 rows in set (0.09 sec)
```

killme2008:
@lyang24 There is a sqlness failure.

@CookiePieWw Have you reviewed this PR? Thanks.

CookiePieWw (approved):
LGTM :)

lyang24:
> LGTM :)
> 
> I'm a little confused by the ci failure in [a3f44fb](https://github.com/GreptimeTeam/greptimedb/commit/a3f44fb43e1890174b3c0ed6cb099ecad5e10bea) since the behavior of `CREATE TABLE` should not be changed.

I think it makes sense for the test to fail the behavior has changed because we merged `set_inverted_index` and `update_inverted_index` where we remove the value if its false.

> nit: replace set_inverted_index and update_inverted_index with with_inverted_index(mut self, value) -> Self, and remove the key when value is false.


i think `insert_inverted_index_placeholder` has not been placed when the empty inverted indexes are defined, thus test will fail.

killme2008:
@lyang24 Thanks a lot.

PR #5135, opened 2024-12-10
perf: avoid cache during compaction

v0y4g3r:

This pr fixes the memory consumption issue during log ingestion. From prior benchmark, we observed unexpected memory usage which won't be reclaimed even after stopping benchmark and flushing all pending data to disk. 

In memory profiling we can see that it's related to merge reader that caches row group data. This behavior is necessary in normal query paths, but not desired in compaction, in that the files read during compaction will be deleted shortly after compaction.
This can also be confirmed through metrics:
```
$ curl -s -XGET localhost:4000/metrics | grep page
greptime_mito_cache_bytes{type="page"} 3776304887
```

## Memory usage after this fix
image

## Checklist

### DIFF
--- a/src/mito2/src/engine.rs
+++ b/src/mito2/src/engine.rs
@@ -424,12 +424,16 @@ impl EngineInner {
         // Get cache.
         let cache_manager = self.workers.cache_manager();
 
-        let scan_region =
-            ScanRegion::new(version, region.access_layer.clone(), request, Some(cache_manager))
-                .with_parallel_scan_channel_size(self.config.parallel_scan_channel_size)
-                .with_ignore_inverted_index(self.config.inverted_index.apply_on_query.disabled())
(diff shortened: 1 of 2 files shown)


fengjiachun (approved):
LGTM

evenyag (approved):

src/mito2/src/compaction.rs
```diff
@@ -584,7 +583,7 @@ impl<'a> CompactionSstReaderBuilder<'a> {
         let mut scan_input = ScanInput::new(self.sst_layer, ProjectionMapper::all(&self.metadata)?)
             .with_files(self.inputs.to_vec())
             .with_append_mode(self.append_mode)
-            .with_cache(self.cache)
```
evenyag, line 587: If we don't pass the cache manager to the compaction reader, we can't use the  write cache to speed up the compaction and reduce the API calls to the object store. @v0y4g3r

PR #5138, opened 2024-12-11
chore: decide tag column in log api follow table schema if table exists

paomian:

* The write log api uses the `greptime_identity` pipeline to follow the tag column of a table that already exists when writing.

## Checklist


src/servers/src/http/event.rs
```diff
@@ -612,10 +612,30 @@ async fn ingest_logs_inner(
...
+        let schema = query_ctx.current_schema();
+        let table = state
+            .get_table(catalog, &schema, &table_name)
+            // FIXME: handle error
+            .await
+            .context(CatalogSnafu)?;
+
+        if let Some(table) = table {
+            let schema = &table.table_info().meta;
+            schema.row_key_column_names().for_each(|column_name| {
```
paomian, line 628: I think the name of the function `row_key_column_names` is not clear enough.
Shall we change it to other like `tag_column_names`.

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -374,13 +374,13 @@ fn json_value_to_row(
 /// 3. The pipeline assumes that the json format is fixed
 /// 4. The pipeline will return an error if the same column datatype is mismatched
 /// 5. The pipeline will analyze the schema of each json record and merge them to get the final schema.
-pub fn identity_pipeline(array: Vec<serde_json::Value>) -> Result<Rows> {
+pub fn identity_pipeline(array: Vec<serde_json::Value>) -> Result<(SchemaInfo, Vec<Row>)> {
```
sunng87, line 377: seems like we don't need to modify this. Or we can add a new argument `table: Option<Arc<Table>>` and build the `Rows` inside this method.

paomian: The reason I didn't pass table is because I didn't want to introduce the table struct in the pipeline crate.

sunng87: Sounds reasonable. Is `Schema` or `SchemaInfo` capable to carry that information?

paomian: Maybe it would be better to pass a tag field name vec(`schema.row_key_column_names()`) to identity_pipeline.

sunng87:
almost lgtm. we just need to 1. update comments of identity_pipeline to describe why the `table` is here. 2. provide some unit tests when table is not none.

src/frontend/src/instance/log_handler.rs
```diff
@@ -84,6 +85,17 @@ impl PipelineHandler for Instance {
             .await
             .context(PipelineSnafu)
     }
+
+    async fn get_table(
+        &self,
+        catalog: &str,
+        schema: &str,
+        table: &str,
+    ) -> std::result::Result<Option<Arc<Table>>, catalog::error::Error> {
```
shuiyisong, line 94: nitpick: I would prefer passing `&QueryContext` instead of `catalog` and `schema` alone

shuiyisong (approved):
Mostly LGTM. Please add some tests.

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -395,23 +391,57 @@ pub fn identity_pipeline(array: Vec<serde_json::Value>) -> Result<Rows> {
...
+/// 5. The pipeline will analyze the schema of each json record and merge them to get the final schema.
+pub fn identity_pipeline(
+    array: Vec<serde_json::Value>,
+    table: Option<Arc<table::Table>>,
+) -> Result<Rows> {
+    match table {
+        Some(table) => {
+            let table_info = table.table_info();
+            let tag_column_names = table_info.meta.row_key_column_names().collect();
+            identity_pipeline_inner(array, Some(tag_column_names))
```
shuiyisong, line 434: nitpick: can we pass `impl Iterator` directly instead of `collect` and `iter` later.

paomian: If impl Iterator is used, line 436 needs an additional type annotation `Option::<std::iter::Empty<&String>>::None`, and `identity_pipeline_inner` needs a lifetime parameter for `Option<impl Iterator<Item = &'a String>>`. Usually the number of columns is not particularly large, and I don't think collect should be a problem here.

Issue #5151, opened 2024-12-12
PromQL queries with binary operators after an `or` operator result in error


dekelpilli:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Query Engine

### Minimal reproduce step

On a metric with data in the given time range, execute 
```cpp
 {__name__="my_metric"} or {__name__="my_metric"} + 5
 ```
 
 The actual query I want to run is 
 
 ```cpp
(delta({__name__="my_metric"}[1m] offset 15s) * 60) > 100
or 
({__name__="my_metric"} * 0)
 ```

As a way to default missing/filtered values to 0. Usually I would use `... or label_replace(vector(0), ...)`, but `label_replace` isn't supported yet.

### What did you expect to see?

The left query values being ORed with 0

### What did you see instead?

PlanQuery: Internal error during building DataFusion plan: No field named my_metric."greptime_value * Float64(0)". Valid fields are my_metric.my_tag, my_metric.greptime_timestamp, "greptime_value * Float64(0)".

### What operating system did you use?

Mac OS 15.1

### What version of GreptimeDB did you use?

0.11.0

waynexia:
Thanks for reporting this. It seems caused by column name mismatching, let me file a fix.

BTW, does this work for you as a workaround? `(delta({__name__="my_metric"}[1m] offset 15s) * 60) > 100 or vector(0)`

dekelpilli:
Thanks for getting on that so quickly.

> BTW, does this work for you as a workaround? (delta({__name__="my_metric"}[1m] offset 15s) * 60) > 100 or vector(0)

No, because we expect the delta portion to return a series (or multiple series) with labels, which then won't match the `vector(0)` portion. This is a workaround because `label_replace` isn't ready.

killme2008:
I implemented the `label_join` and `label_replace` functions in #5153. Thank you for your report.

killme2008:
@waynexia Is this issue fixed?

killme2008:
Not sure if it is fixed @waynexia

waynexia:
I've verified on the latest `v0.15.3` and it's fixed. Sorry for the late response

PR #5153, opened 2024-12-12
feat: impl label_join and label_replace for promql

killme2008:

#5151 

Impl the following functions for PromQL:

* [label_join](https://prometheus.io/docs/prometheus/latest/querying/functions/#label_join)
* [label_replace](https://prometheus.io/docs/prometheus/latest/querying/functions/#label_replace)

## Checklist


src/query/src/promql/planner.rs
```diff
@@ -1425,6 +1457,127 @@ impl PromPlanner {
         Ok(exprs)
     }
 
+    /// Build expr for `label_replace` function
+    fn build_regexp_replace_label_expr(
+        other_input_exprs: &mut VecDeque<DfExpr>,
+        session_state: &SessionState,
+    ) -> Result<DfExpr> {
+        // label_replace(vector, dst_label, replacement, src_label, regex)
```
evenyag, line 1465: What if `dst_label` has the same name as `src_label`?

tests/cases/standalone/common/promql/label.result
```diff
@@ -0,0 +1,103 @@
...
+| 1970-01-01T00:00:00 | 1   | idc1-host1       | host1 | idc1       |
+| 1970-01-01T00:00:05 | 1   | idc1-host1       | host1 | idc1       |
+| 1970-01-01T00:00:05 | 3   | idc2:zone1-host1 | host1 | idc2:zone1 |
+| 1970-01-01T00:00:10 | 1   | idc1-host1       | host1 | idc1       |
+| 1970-01-01T00:00:10 | 3   | idc2:zone1-host1 | host1 | idc2:zone1 |
+| 1970-01-01T00:00:10 | 5   | idc3:zone2-host1 | host1 | idc3:zone2 |
+| 1970-01-01T00:00:15 | 1   | idc1-host1       | host1 | idc1       |
+| 1970-01-01T00:00:15 | 3   | idc2:zone1-host1 | host1 | idc2:zone1 |
+| 1970-01-01T00:00:15 | 5   | idc3:zone2-host1 | host1 | idc3:zone2 |
+| 1970-01-01T00:00:15 | 7   | idc4:zone3-host1 | host1 | idc4:zone3 |
```
waynexia, line 43: I'm not sure if this is valid from the output order. values within the same time-series (after label_join) are not continuous. Could you add a test like rate with label_join to validate it?

src/query/src/promql/planner.rs
killme2008: Good catch! Fixed in https://github.com/GreptimeTeam/greptimedb/pull/5153/commits/9750e5fc83f6eeb4b502db00581540c03cf720f0

It will replace the source label with the generated new label value.

tests/cases/standalone/common/promql/label.result
killme2008: Confirm that Prometheus also does not support the range selector for these function results, so it is irrelevant.

dekelpilli:
It's worth noting that in PromQL, passing an empty string for `label_replace`'s `replacement` argument will result in the label being removed. That is, if
```
{__name__="test",host="abc"} => test{host="abc",other_label="my_value"}
```
then
```
label_repalce({__name__="test",host="abc"}, "host", "", "", "", "") => test{other_label="my_value"}
```

I'm not sure if this would be handled downstream here, but I don't believe `regexp_replace` handles that.

killme2008:
> It's worth noting that in PromQL, passing an empty string for `label_replace`'s `replacement` argument will result in the label being removed. That is, if
> 
> ```
> {__name__="test",host="abc"} => test{host="abc",other_label="my_value"}
> ```
> 
> then
> 
> ```
> label_repalce({__name__="test",host="abc"}, "host", "", "", "", "") => test{other_label="my_value"}
> ```
> 
> I'm not sure if this would be handled downstream here, but I don't believe `regexp_replace` handles that.

Thank you for your note; I'll test it now.

killme2008:
@dekelpilli Looks like we can't remove the label in the current implementation right now. cc @waynexia 

We don't know the replaced results until executing the plan, it's not easy to implement in the current planner.

I leave a todo and try to find a solution in the future.

killme2008:
@waynexia @evenyag Ready for review, please take a look. Thanks @dekelpilli

evenyag (approved):

waynexia (approved):

PR #5161, opened 2024-12-13
fix: loki write row len error

paomian:

* fix loki write row len error
* The previous test passed as is because there was only one record in a request. Either the schema doesn't trigger recapacity, or it doesn't reuse the schema.

## Checklist

### DIFF
--- a/src/servers/src/http/event.rs
+++ b/src/servers/src/http/event.rs
@@ -514,8 +514,8 @@ pub async fn loki_ingest(
             let line = entry.line;
 
             // create and init row
-            let mut row = Vec::with_capacity(schemas.capacity());
-            for _ in 0..row.capacity() {
+            let mut row = Vec::with_capacity(schemas.len());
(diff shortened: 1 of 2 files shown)


shuiyisong (approved):

shuiyisong (approved):

src/servers/src/http/event.rs
```diff
@@ -514,7 +514,7 @@ pub async fn loki_ingest(
             let line = entry.line;
 
             // create and init row
-            let mut row = Vec::with_capacity(schemas.capacity());
+            let mut row = Vec::with_capacity(schemas.len());
             for _ in 0..row.capacity() {
```
sunng87, line 518: the key is to change this `row.capacity()` to schemas_len

paomian: I will change `row.capacity()` to `schemas.len()`

PR #5162, opened 2024-12-13
test: flow rebuild

discord9:

add some tests related to flow's rebuild. Which will become more useful once flow refill(recover of internal state) is implemented.

## Checklist


shuiyisong (approved):

evenyag (approved):
The test cases should be sufficient. Maybe we can write a fuzz test for this instead of using sqlness to test it.
cc @WenyXu @waynexia

PR #5164, opened 2024-12-13
feat: introduce Buffer for non-continuous bytes

CookiePieWw:

As title. The implementation is copied from `opendal` actually.

Also refactor the inverted index cache to avoid a copy when reading bytes from remote blob. But we still need to copy the bytes to a continuous `Vec<u8>` for reading it, maybe we can avoid it by using `Buffer` more heavily. But `Buffer` itself is cache-unfriendly, so I'm not sure if it's good.

Seems `Bytes` it self is a thread-safe and copy-efficient, so we don't need wrap it up with an `Arc`.

## Checklist


src/mito2/src/cache/index.rs
```diff
@@ -190,31 +188,14 @@ impl IndexDataPageKey {
         (end_page + 1 - start_page) as u32
     }
 
-    /// Computes the byte range in the first page based on the offset and size.
-    /// For example, if offset is 1000 and size is 5000 with PAGE_SIZE of 4096, the first page range is 1000..4096.
-    fn calculate_first_page_range(offset: u64, size: u32, page_size: u64) -> Range<usize> {
+    /// Computes the byte range based on the offset and size.
+    /// For example, if offset is 5000 and size is 5000 with PAGE_SIZE of 4096, the range is 904..5904.
```
WenyXu, line 192:
```suggestion
/// Calculates the byte range for data retrieval based on the specified offset and size.
    ///
    /// This function determines the starting and ending byte positions required for reading data.
    /// For example, with an offset of 5000 and a size of 5000, using a PAGE_SIZE of 4096,
    /// the resulting byte range will be 904..5904. This indicates that:
    /// - The reader will first access fixed-size pages [4096, 8192) and [8192, 12288).
    /// - To read the range [5000..10000), it only needs to fetch bytes within the range [904, 5904) across two pages.
```

WenyXu (approved):

src/mito2/src/cache/index.rs
```diff
@@ -97,24 +99,20 @@ where
         if !cache_miss_range.is_empty() {
             let pages = self.inner.read_vec(&cache_miss_range).await?;
             for (i, page) in cache_miss_idx.into_iter().zip(pages.into_iter()) {
-                let page = Arc::new(page);
+                let page = page;
                 let key = keys[i].clone();
                 data[i] = page.clone();
                 self.cache.put_index(key, page.clone());
             }
```
WenyXu, line 106:
```suggestion
        if !cache_miss_range.is_empty() {
            let pages = self.inner.read_vec(&cache_miss_range).await?;
            for (i, page) in cache_miss_idx.into_iter().zip(pages.into_iter()) {
                let key = keys[i].clone();
                data[i] = page.clone();
                self.cache.put_index(key, page);
            }
```

WenyXu (commented):
Rest LGTM, remember to remove unnecessary TODO comments

v0y4g3r:
We don't have to copy the buffer code from opendal in that we've already re-exported that struct:
https://github.com/GreptimeTeam/greptimedb/blob/03a28320d6f322000790eeb2b81ce7cb2cb9a972/src/object-store/src/lib.rs#L17

zhongzc (approved):

Issue #5073, opened 2024-11-29
DELETE not work in last_non_null mode


evenyag:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Standalone mode, Distributed Cluster

### Minimal reproduce step

```
mysql> CREATE TABLE IF NOT EXISTS `test` (
    ->   `time` TIMESTAMP(0) NOT NULL,
    ->   `code` STRING NULL,
    ->   `name` STRING NULL,
    ->   `status` TINYINT NULL,
    ->   TIME INDEX (`time`),
    ->   PRIMARY KEY (`code`)
    -> ) ENGINE=mito WITH(
    ->   append_mode = 'false',
    ->   merge_mode = 'last_non_null'
    -> );
Query OK, 0 rows affected (0.19 sec)

mysql>
mysql> INSERT INTO `test` (`time`, `code`, `name`, `status`) VALUES ('2024-11-26 10:00:00', 'achn', '1.png', 0);
Query OK, 1 row affected (0.03 sec)

mysql> INSERT INTO `test` (`time`, `code`, `name`, `status`) VALUES ('2024-11-26 10:01:00', 'achn', '2.png', 0);
Query OK, 1 row affected (0.00 sec)

mysql> INSERT INTO `test` (`time`, `code`, `name`, `status`) VALUES ('2024-11-26 10:02:00', 'achn', '3.png', 1);
Query OK, 1 row affected (0.00 sec)

mysql>
mysql> SELECT * FROM `test`;
+---------------------+------+-------+--------+
| time                | code | name  | status |
+---------------------+------+-------+--------+
| 2024-11-26 10:00:00 | achn | 1.png |      0 |
| 2024-11-26 10:01:00 | achn | 2.png |      0 |
| 2024-11-26 10:02:00 | achn | 3.png |      1 |
+---------------------+------+-------+--------+
3 rows in set (0.15 sec)

mysql> DELETE FROM `test`;

mysql>
mysql> INSERT INTO `test` (`time`, `code`, `name`) VALUES ('2024-11-26 10:00:00', 'achn', '1.png');
Query OK, 1 row affected (0.00 sec)

mysql> INSERT INTO `test` (`time`, `code`, `name`) VALUES ('2024-11-26 10:01:00', 'achn', '2.png');
Query OK, 1 row affected (0.00 sec)

[...]

PR #5168, opened 2024-12-14
fix: deletion between two put may not work in `last_non_null` mode

evenyag:

fixes https://github.com/GreptimeTeam/greptimedb/issues/5073

Before this PR, the `LastFieldsBuilder` only updated the deletion flag while initializing the builder. So, a delete op may not prevent previous rows from updating the null fields in the last row.

This PR updates the `contains_deletion` flag in `push_first_row()` to fix this issue.

## Checklist

### DIFF
--- a/src/mito2/src/read/dedup.rs
+++ b/src/mito2/src/read/dedup.rs
@@ -311,6 +311,12 @@ impl LastFieldsBuilder {
             return;
         }
 
+        self.contains_deletion = batch.op_types().get_data(0).unwrap() == OpType::Delete as u8;
+        if self.contains_deletion {
+            // Deletes this row.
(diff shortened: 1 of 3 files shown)


src/mito2/src/read/dedup.rs
```diff
@@ -311,6 +311,12 @@ impl LastFieldsBuilder {
             return;
         }
 
+        self.contains_deletion = batch.op_types().get_data(0).unwrap() == OpType::Delete as u8;
```
v0y4g3r, line 314: Can you add some doc for this early return condition?

v0y4g3r (approved):
LGTM

PR #5180, opened 2024-12-16
feat: do not remove time filters in ScanRegion

evenyag:

Don't remove time filters when building time ranges in `ScanRegion`. Then the parquet reader can use the time filters to prune row groups. This PR also removes the time range from the parquet reader.

Now the scanner only uses the time range to find files to scan.

## Checklist


killme2008 (approved):
LGTM

waynexia (approved):

PR #5128, opened 2024-12-09
feat: extract hints from http header

fengjiachun:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

As the title, and close #5043 

## Checklist

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.

PR #5194, opened 2024-12-18
feat: hints all in one

fengjiachun:

Another hint format: `x-greptime-hints: auto_create_table=true, ttl=7d`.

Currently, the following two formats are supported:
1. x-greptime-hint-xxx: value
2. x-greptime-hints: xxx=yyy, aaa=bbb

This PR implements the second format and we may remove support for the first one in the future as it may lead too many keys.

This PR is a continuation of #5128.

## Checklist

### DIFF
--- a/src/servers/src/hint_headers.rs
+++ b/src/servers/src/hint_headers.rs
@@ -15,6 +15,9 @@
 use http::HeaderMap;
 use tonic::metadata::MetadataMap;
 
+// For the given format: `x-greptime-hints: auto_create_table=true, ttl=7d`
+pub const HINTS_KEY: &str = "x-greptime-hints";
+
 pub const HINT_KEYS: [&str; 5] = [
     "x-greptime-hint-auto_create_table",
     "x-greptime-hint-ttl",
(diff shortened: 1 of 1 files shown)


src/servers/src/hint_headers.rs
```diff
@@ -25,6 +28,14 @@ pub const HINT_KEYS: [&str; 5] = [
...
 pub(crate) fn extract_hints<T: ToHeaderMap>(headers: &T) -> Vec<(String, String)> {
     let mut hints = Vec::new();
+    if let Some(value_str) = headers.get(HINTS_KEY) {
+        value_str.split(',').for_each(|hint| {
+            let mut parts = hint.splitn(2, '=');
+            if let (Some(key), Some(value)) = (parts.next(), parts.next()) {
+                hints.push((key.trim().to_string(), value.trim().to_string()));
+            }
+        });
+    }
```
killme2008, line 38: If the header includes `HINTS_KEY`, it should be prioritized, allowing for an early return upon its discovery.

fengjiachun: You mean if we find `x-greptime-hints`, we should ignore keys starting with `x-greptime-hint-`?

killme2008: > You mean if we find `x-greptime-hints`, we should ignore keys starting with `x-greptime-hint-`?

Yes, always prefers `x-greptime-hints`.

fengjiachun: Ok, done in [e44f279](https://github.com/GreptimeTeam/greptimedb/pull/5194/commits/e44f27925582696c56d426606fcf897abb009205)

killme2008 (approved):
LGTM

WenyXu (approved):

PR #5195, opened 2024-12-18
fix(flow): batch builder with type

discord9:

fix a bug in which null type vector builder is used incorrectly

- previously, some vector builder is used wrong and could be null vector builder when encounter null value, this is fixed now as passing actual type is supported to build vector&batch in flow

## Checklist


discord9:
Fixing failing tests now

src/flow/src/expr.rs
```diff
@@ -221,10 +223,28 @@ impl Batch {
...
+                } else {
+                    // both are null, so we will push null type
+                    dts.push(datatypes::prelude::ConcreteDataType::null_datatype())
+                }
+            }
+            if self.batch.is_empty() {
+                other.batch.iter().map(|v| v.data_type()).collect_vec()
+            } else {
+                self.batch.iter().map(|v| v.data_type()).collect_vec()
+            }
```
evenyag, line 247: Looks like we need to remove them.

src/flow/src/expr.rs
```diff
@@ -221,10 +223,28 @@ impl Batch {
...
+                    dts.push(v.data_type())
+                } else if let Some(v) = other.batch().get(i)
+                    && !v.data_type().is_null()
+                {
+                    dts.push(v.data_type())
+                } else {
+                    // both are null, so we will push null type
+                    dts.push(datatypes::prelude::ConcreteDataType::null_datatype())
+                }
+            }
```
evenyag, line 242: We never read this `dts` outside of the block.

src/flow/src/adapter.rs
```diff
@@ -555,6 +555,8 @@ impl FlowWorkerManager {
         let default_interval = Duration::from_secs(1);
         let mut avg_spd = 0; // rows/sec
         let mut since_last_run = tokio::time::Instant::now();
+        let run_per_trace = 5;
```
evenyag, line 558: I wonder if `5` is too small.

src/flow/src/compute/render/reduce.rs
```diff
@@ -47,6 +47,7 @@ impl Context<'_, '_> {
         reduce_plan: &ReducePlan,
         output_type: &RelationType,
     ) -> Result<CollectionBundle<Batch>, Error> {
+        common_telemetry::debug!("render reduce batch");
```
evenyag, line 50: Do we still need this log?

fengjiachun (approved):
LGTM

fengjiachun:
@discord9 Some conflits

PR #5201, opened 2024-12-19
feat: init PgElection

CookiePieWw:

As title. Details:

1. For election, candidates try acquire advisory lock regularly.
    - If successful, the candidate is the leader. Leader should keep alive by update leader value's expire time and acquire lock multiple times. If failed once, step down.
    - Edge case:
      1. Leader hold the lock but leader value expire:
          - Happens when: leader stuck between acquire lock and renew lease;
          - Result: either a follower or the leader will delete the value and re-init a campaign.
          - Follower state: a follower finds it expire, re-init a campaign. Does not delete key to avoid conflict.
          - Leader state 1: If stuck over `IDLE_SESSION_TIMEOUT`, the lock is released and the campaign restart. 
            - When leader come back, it find he cannot acquire the advisory lock, just become a follower.
          - Leader state 2: If stuck less then `IDLE_SESSION_TIMEOUT`, it will:
            - Find the leader value deleted by followers, just step down with lock released and join a new campaign,
            - Find the leader value expired, just step down with value deleted, lock released and re-init a new campaign.
       2. Leader value is valid but cannot acquire the lock.
           - Happens when: leader stuck over `IDLE_SESSION_TIMEOUT` and comes back.
           - Actually has been explained above in **Leader state 1**. The `IDLE_SESSION_TIMEOUT` is longer than the `META_LEASE_TIME` is the key that this case happens only if case 1 happens.

    - If not successful, the candidate is a follower. Follower should check the leader value's expire time and acquire lock multiple times. If the leader value expired, delete the value and re-init a campaign. If get the lock, just become the leader.

4. For registering candidates, just regularly register themselves with expire time.

## Checklist


src/meta-srv/src/bootstrap.rs
```diff
@@ -225,8 +227,14 @@ pub async fn metasrv_builder(
         (None, BackendImpl::PostgresStore) => {
             let pg_client = create_postgres_client(opts).await?;
             let kv_backend = PgStore::with_pg_client(pg_client).await.unwrap();
-            // TODO(jeremy, weny): implement election for postgres
-            (kv_backend, None)
+            let election_client = create_postgres_client(opts).await?;
```
CookiePieWw, line 230: We have to create another client since `tokio_postgres::Client` is stateful and cannot be cloned. Would it be possible that one of the connections crashed and cause unpredictable fault?

src/meta-srv/src/election/postgres.rs
```diff
@@ -0,0 +1,547 @@
...
+            {
+                Ok(value_with_lease.value.into())
+            } else {
+                return NoLeaderSnafu {}.fail();
+            }
+        }
+    }
+
+    async fn resign(&self) -> Result<()> {
+        todo!()
```
CookiePieWw, line 362: Corresponding `todo!` to `EtcdElection`.

src/meta-srv/src/bootstrap.rs
sunng87: I think it's ok. If this election client is broken, the lock is auto released thanks to this advisory lock mechanism. And if the kv client is broken, it will try to reconnect on next request.

src/meta-srv/src/election/postgres.rs
```diff
@@ -0,0 +1,569 @@
...
+    Election, LeaderChangeMessage, LeaderKey, CANDIDATES_ROOT, CANDIDATE_LEASE_SECS, ELECTION_KEY,
+    KEEP_ALIVE_INTERVAL_SECS,
+};
+use crate::error::{
+    DeserializeFromJsonSnafu, NoLeaderSnafu, PostgresExecutionSnafu, Result, SerializeToJsonSnafu,
+    UnexpectedSnafu,
+};
+use crate::metasrv::{ElectionRef, LeaderValue, MetasrvNodeInfo};
+
+const CAMPAIGN: &str = "SELECT pg_try_advisory_lock(1)";
```
sunng87, line 41: We might need to make this keyid configurable in case that it's in use by other applications. But for now we can just leave a comment here.

sunng87:
The cool part of this advisory lock is it's got unlocked when the connection is down. So I think we don't need to manually unlock it. 

And in order to acquire the lock, each metasrv node should call `pg_try_advisory_lock` periodically. The connection that holds the lock will always get `t` and it also acts as heartbeat to keep connection live.

CookiePieWw:
> The cool part of this advisory lock is it's got unlocked when the connection is down. So I think we don't need to manually unlock it.
> 
> And in order to acquire the lock, each metasrv node should call `pg_try_advisory_lock` periodically. The connection that holds the lock will always get `t` and it also acts as heartbeat to keep connection live.

Initially I think there might be situations that a metasrv holds the lock but fails to renew the lease. Since we can think holding the lock itself indicates the metasrv is still alive, we can remove the lease/expire-time part of campaign?

src/meta-srv/src/election/postgres.rs
CookiePieWw: Sure. I'd like to make key-id and the table name of pg backend (currently it's hard-coded as greptime_metakv) configurable in the next pr.

sunng87:
> Initially I think there might be situations that a metasrv holds the lock but fails to renew the lease. Since we can think holding the lock itself indicates the metasrv is still alive, we can remove the lease/expire-time part of campaign/leader value and check the lock as heartbeat instead?

Conclusion from offline discussion:

- `pg_try_advisory_lock` can be used as both lock and heartbeat
- we need to set session variable `idle_session_timeout` to prevent lock leak
- we need to ensure leader key updated correctly by lock owner

CookiePieWw:
Split to multiple patches. See https://github.com/GreptimeTeam/greptimedb/pull/5209 for the first part

PR #5202, opened 2024-12-19
feat: logs query endpoint

waynexia:

A basic implementation of `v1/logs` query interface. Example:

```shell
curl -X "POST" "http://localhost:4000/v1/logs" \                                                                                                                                                                            ─╯
    -H "Content-Type: application/json" \
    -d $'
    {
        "table": {
            "catalog_name": "greptime",
            "schema_name": "public",
            "table_name": "test_mem"
        },
        "time_filter": {
            "start": "2024-10-17"
        },
        "columns": [
            {
                "column_name": "message",
                "filters": [
                    {
                       "Contains": "production"
                    }
                ]
            },
            {
                "column_name": "ip",
                "filters": []
            },
            {
                "column_name": "timestamp",
                "filters": []
            }
        ],
        "context": "None"
    }
'
```

## Checklist


src/log-query/src/log_query.rs
```diff
@@ -34,6 +36,18 @@ pub struct LogQuery {
...
+        Self {
+            table: TableName::new("", "", ""),
+            time_filter: Default::default(),
+            columns: vec![],
+            limit: None,
+            context: Default::default(),
+        }
+    }
+}
+
```
shuiyisong, line 50: How about removing the `impl Default`? It's strange to have a table name with empty strings. We can create a `new` function if needed.

src/query/src/log_query/planner.rs
```diff
@@ -0,0 +1,353 @@
...
+
+fn escape_pattern(pattern: &str) -> String {
+    pattern
+        .chars()
+        .flat_map(|c| match c {
+            '\\' | '%' | '_' => vec!['\\', c],
+            _ => vec![c],
+        })
+        .collect::<String>()
+}
```
shuiyisong, line 179: Looks like the function is the exact same as the one in `common-function`'s `PatternAst::escape_pattern`. Can we use import instead?

src/query/src/log_query/planner.rs
```diff
@@ -0,0 +1,353 @@
...
+use log_query::{ColumnFilters, LogQuery, TimeFilter};
+use snafu::{OptionExt, ResultExt};
+use table::table::adapter::DfTableProviderAdapter;
+
+use crate::log_query::error::{
+    CatalogSnafu, DataFusionPlanningSnafu, Result, TimeIndexNotFoundSnafu, UnimplementedSnafu,
+    UnknownTableSnafu,
+};
+
+const DEFAULT_LIMIT: usize = 1000;
```
fengjiachun, line 31: Isn't 1000 too much? 100 should be enough, shouldn't it?

src/query/src/log_query/planner.rs
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    fn build_time_filter(&self, time_filter: &TimeFilter, schema: &Schema) -> Result<Expr> {
+        let timestamp_col = schema
+            .timestamp_column()
+            .with_context(|| TimeIndexNotFoundSnafu {})?
+            .name
+            .clone();
+
+        let expr = col(timestamp_col.clone())
```
fengjiachun, line 111: It's difficult to understand; can we refactor it as:

```
let start_time = ScalarValue::Utf8(time_filter.start.clone());
let end_time = ScalarValue::Utf8(time_filter.end.clone().unwrap_or_else(|| "9999-12-31T23:59:59Z".to_string()));

let expr = col(timestamp_col.clone())
        .gt_eq(lit(start_time))
        .and(col(timestamp_col).lt_eq(lit(end_time)));

```

fengjiachun (approved):
Rest LGTM

src/query/src/log_query/planner.rs
waynexia: Looks like 1000 is preferred https://github.com/grafana/loki/issues/2226

src/log-query/src/log_query.rs
```diff
@@ -34,6 +36,18 @@ pub struct LogQuery {
     pub context: Context,
 }
 
+impl Default for LogQuery {
```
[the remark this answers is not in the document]
waynexia: For deserialization purpose, in case some fields are missing

src/log-query/src/log_query.rs
waynexia: This is for deserializing, when some fields are missing in the incoming query payload

shuiyisong (approved):
🚀

PR #5203, opened 2024-12-19
fix: auto created table ttl check

discord9:

auto created table also needed to check for ttl==instant

- previous auto created table is not check for ttl==instant which will cause the first batch to create the table to be in the table even if ttl==instant
- fix is to also check for them so ttl=instant table wouldn't have any data anyway 

## PR Checklist

### DIFF
--- a/src/operator/src/insert.rs
+++ b/src/operator/src/insert.rs
@@ -576,6 +576,9 @@ impl Inserter {
 
                     for table in tables {
                         let table_info = table.table_info();
+                        if table_info.is_ttl_instant_table() {
+                            instant_table_ids.insert(table_info.table_id());
+                        }
                         table_name_to_ids.insert(table_info.name.clone(), table_info.table_id());
                     }
                 }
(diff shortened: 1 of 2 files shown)


src/operator/src/insert.rs
```diff
@@ -576,6 +576,9 @@ impl Inserter {
 
                     for table in tables {
                         let table_info = table.table_info();
+                        if table_info.is_ttl_instant_table() {
```
killme2008, line 579: It's better to add a test.

killme2008 (approved):

src/operator/src/insert.rs
discord9: > It's better to add a test.

Might be a little tricky to test, is considering extending sqlness to test it but that might be the scope of another PR, syntax might be like this:
```sql
-- SQLNESS ARG use_http=true
-- SQLNESS ARG http_header="xxxx"
INSERT INTO xxx VALUES (...);
```

killme2008: Maybe you can try http test

https://github.com/GreptimeTeam/greptimedb/blob/main/tests-integration/tests/http.rs

evenyag (approved):

Issue #5208, opened 2024-12-20
Implement leader election over postgresql


CookiePieWw:
### What type of enhancement is this?

Other

### What does the enhancement do?

- [x] Candidate registration 
   - [x] Init #5209 
   - [x] Separate registration and keep alive logics #5320 
- [x] Leader election #5249 
- [x] Txn
   - [x] Connection Pool #5260 
   - [x] Txn Impl #5266
- [x] Config
   - [x] Configurable pg kv backend table name #5244
   - [x] Configurable advisory lock id #5244
   - [x] ~~Configurable isolation level for txn~~ Only serializable can be set.
- [ ] Tests
  - [x] Unit test for Txn
     - [x] Init #5266
     - [x] Fix concurrent problem when multiple unit test are executed
            ~~Hopefully we can~~ get it done in #5266 too.
  - [x] Sqlness #5309 
  - [ ] Integration

killme2008:
Let's close this issue. Thanks a lot.

PR #5209, opened 2024-12-20
feat: init PgElection with candidate registration

CookiePieWw:

As title. Part of #5208 

Previous pr #5201 

A candidate first write a (candidate_key, node info + expire time) pair into the pg kv backend, and regularly check the expire time.
- If the value expired, raise an error and re-registrate
- If not, just renew the expire time

For `all_candidates`, just get all candidate using candidate_key_prefix, and filter the expired value.

## PR Checklist

### DIFF
--- a/src/meta-srv/src/bootstrap.rs
+++ b/src/meta-srv/src/bootstrap.rs
@@ -288,8 +288,15 @@ async fn create_postgres_client(opts: &MetasrvOptions) -> Result<tokio_postgres:
     let postgres_url = opts.store_addrs.first().context(InvalidArgumentsSnafu {
         err_msg: "empty store addrs",
     })?;
-    let (client, _) = tokio_postgres::connect(postgres_url, NoTls)
+    let (client, connection) = tokio_postgres::connect(postgres_url, NoTls)
         .await
         .context(error::ConnectPostgresSnafu)?;
+
+    tokio::spawn(async move {
+        connection
+            .await
+            .context(error::PostgresExecutionSnafu)
+            .unwrap()
+    });
     Ok(client)
 }
--- a/src/meta-srv/src/bootstrap.rs
+++ b/src/meta-srv/src/bootstrap.rs
@@ -45,13 +45,7 @@ use tonic::codec::CompressionEncoding;
 use tonic::transport::server::{Router, TcpIncoming};
 
 use crate::election::etcd::EtcdElection;
-#[cfg(feature = "pg_kvbackend")]
-use crate::election::postgres::PgElection;
-#[cfg(feature = "pg_kvbackend")]
-use crate::election::CANDIDATE_LEASE_SECS;
-#[cfg(feature = "pg_kvbackend")]
-use crate::error::InvalidArgumentsSnafu;
-use crate::error::{InitExportMetricsTaskSnafu, TomlFormatSnafu};
+use crate::error::{InitExportMetricsTaskSnafu, InvalidArgumentsSnafu, TomlFormatSnafu};
 use crate::metasrv::builder::MetasrvBuilder;
 use crate::metasrv::{BackendImpl, Metasrv, MetasrvOptions, SelectorRef};
 use crate::selector::lease_based::LeaseBasedSelector;
@@ -231,15 +225,7 @@ pub async fn metasrv_builder(
             let kv_backend = PgStore::with_pg_client(pg_client)
                 .await
                 .context(error::KvBackendSnafu)?;
-            let election_client = create_postgres_client(opts).await?;
-            let election = PgElection::with_pg_client(
-                opts.server_addr.clone(),
-                election_client,
-                opts.store_key_prefix.clone(),
-                CANDIDATE_LEASE_SECS,
-            )
-            .await?;
-            (kv_backend, Some(election))
+            (kv_backend, None)
         }
     };
 
--- a/src/meta-srv/src/bootstrap.rs
+++ b/src/meta-srv/src/bootstrap.rs
@@ -46,6 +46,8 @@ use tonic::transport::server::{Router, TcpIncoming};
 
 use crate::election::etcd::EtcdElection;
 #[cfg(feature = "pg_kvbackend")]
+use crate::election::postgres::PgElection;
+#[cfg(feature = "pg_kvbackend")]
 use crate::error::InvalidArgumentsSnafu;
 use crate::error::{InitExportMetricsTaskSnafu, TomlFormatSnafu};
 use crate::metasrv::builder::MetasrvBuilder;
@@ -224,9 +226,17 @@ pub async fn metasrv_builder(
         #[cfg(feature = "pg_kvbackend")]
         (None, BackendImpl::PostgresStore) => {
             let pg_client = create_postgres_client(opts).await?;
-            let kv_backend = PgStore::with_pg_client(pg_client).await.unwrap();
-            // TODO(jeremy, weny): implement election for postgres
-            (kv_backend, None)
+            let kv_backend = PgStore::with_pg_client(pg_client)
+                .await
+                .context(error::KvBackendSnafu)?;
+            let election_client = create_postgres_client(opts).await?;
+            let election = PgElection::with_pg_client(
+                opts.server_addr.clone(),
+                election_client,
+                opts.store_key_prefix.clone(),
+            )
+            .await?;
+            (kv_backend, Some(election))
         }
     };
 
--- a/src/meta-srv/src/bootstrap.rs
+++ b/src/meta-srv/src/bootstrap.rs
@@ -45,7 +45,9 @@ use tonic::codec::CompressionEncoding;
 use tonic::transport::server::{Router, TcpIncoming};
 
 use crate::election::etcd::EtcdElection;
-use crate::error::{InitExportMetricsTaskSnafu, InvalidArgumentsSnafu, TomlFormatSnafu};
+#[cfg(feature = "pg_kvbackend")]
+use crate::error::InvalidArgumentsSnafu;
+use crate::error::{InitExportMetricsTaskSnafu, TomlFormatSnafu};
 use crate::metasrv::builder::MetasrvBuilder;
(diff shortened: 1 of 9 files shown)


src/meta-srv/src/election.rs
```diff
@@ -26,10 +27,20 @@ use crate::metasrv::MetasrvNodeInfo;
...
+
 #[derive(Debug, Clone)]
 pub enum LeaderChangeMessage {
-    Elected(Arc<LeaderKey>),
-    StepDown(Arc<LeaderKey>),
+    Elected(Arc<dyn LeaderKey>),
+    StepDown(Arc<dyn LeaderKey>),
+}
+
+pub trait LeaderKey: Send + Sync + Debug {
```
fengjiachun, line 39: Always add documentation for public traits and their methods.

src/meta-srv/src/election/postgres.rs
```diff
@@ -0,0 +1,488 @@
...
+    }
+
+    async fn get_value_with_lease_by_prefix(
+        &self,
+        key_prefix: &String,
+    ) -> Result<Vec<ValueWithLease>> {
+        let key_prefix = format!("{}%", key_prefix);
+        let prev = self
+            .client
+            .query(PREFIX_SCAN, &[&key_prefix])
```
CookiePieWw, line 215: Maybe we can wrap the query up with a function for type safety. It's quite dangerous here to directly use strings.

src/meta-srv/src/election.rs
```diff
@@ -26,10 +27,31 @@ use crate::metasrv::MetasrvNodeInfo;
...
+    /// The name in byte. name is the election identifier that corresponds to the leadership key.
+    fn name(&self) -> &[u8];
+
+    /// The key in byte. key is an opaque key representing the ownership of the election. If the key
+    /// is deleted, then leadership is lost.
+    fn key(&self) -> &[u8];
+
+    /// The creation revision of the key.
+    fn rev(&self) -> i64;
```
fengjiachun, line 51: It's better to use `revision` as the method name for clarity.

src/meta-srv/src/election.rs
```diff
@@ -26,10 +27,31 @@ use crate::metasrv::MetasrvNodeInfo;
...
+
+    /// The lease ID of the election leader.
+    fn lease(&self) -> i64;
```
fengjiachun, line 54: It's better to use `lease_id` as the method name for clarity.

src/meta-srv/src/bootstrap.rs
```diff
@@ -275,8 +276,15 @@ async fn create_postgres_client(opts: &MetasrvOptions) -> Result<tokio_postgres:
...
+    let (client, connection) = tokio_postgres::connect(postgres_url, NoTls)
         .await
         .context(error::ConnectPostgresSnafu)?;
+
+    tokio::spawn(async move {
+        connection
+            .await
+            .context(error::PostgresExecutionSnafu)
+            .unwrap()
+    });
```
WenyXu, line 288: It doesn't seem necessary?

src/meta-srv/src/election/postgres.rs
```diff
@@ -0,0 +1,488 @@
...
+                    ),
+                }
+            );
+
+            let updated = ValueWithLease {
+                value: prev.value.clone(),
+                expire_time: now + CANDIDATE_LEASE_SECS as f64,
+            };
+            self.update_value_with_lease(&key, &prev, &updated).await?;
+        }
```
WenyXu, line 152: it seems to take several seconds?

src/meta-srv/src/bootstrap.rs
CookiePieWw: I actually follows the document of `tokio_postgres` [here](https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.Connection.html):
> This is one half of what is returned when a new connection is established. It performs the actual IO with the server, and should generally be spawned off onto an executor to run in the background.

Not quite sure what will happen if not spawn a test with it.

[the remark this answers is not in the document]
CookiePieWw: I actually follows the document of `tokio_postgres` [here](https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.Connection.html):
> This is one half of what is returned when a new connection is established. It performs the actual IO with the server, and should generally be spawned off onto an executor to run in the background.

Not quite sure what will happen if not spawn a test with it.

src/meta-srv/src/election/postgres.rs
```diff
@@ -0,0 +1,488 @@
...
+    infancy: AtomicBool,
+    leader_watcher: broadcast::Sender<LeaderChangeMessage>,
+    store_key_prefix: String,
+    candidate_lease_ttl_secs: u64,
+}
+
+impl PgElection {
+    pub async fn with_pg_client(
+        leader_value: String,
+        client: Client,
```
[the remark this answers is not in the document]
CookiePieWw: Got you. It's more safe and convenient to directly use `KvBackend` and then we don't really need to do some crud tests.

src/meta-srv/src/election/postgres.rs
```diff
@@ -0,0 +1,488 @@
...
+        store_key_prefix: String,
+        candidate_lease_ttl_secs: u64,
+    ) -> Result<ElectionRef> {
+        let (tx, _) = broadcast::channel(100);
+        Ok(Arc::new(Self {
+            leader_value,
+            client,
+            is_leader: AtomicBool::new(false),
+            infancy: AtomicBool::new(true),
```
fengjiachun, line 64: The initial value should probably be false.
```suggestion
            infancy: AtomicBool::new(false),
```

src/meta-srv/src/election/postgres.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    async fn register_candidate(&self, node_info: &MetasrvNodeInfo) -> Result<()> {
+        let key = self.candidate_key();
+        let node_info =
+            serde_json::to_string(node_info).with_context(|_| SerializeToJsonSnafu {
+                input: format!("{node_info:?}"),
+            })?;
+        let value_with_lease = ValueWithLease {
+            value: node_info,
+            expire_time: time::SystemTime::now()
```
fengjiachun, line 110: Can we use the database server's time? I'm concerned that using client-side time might be inaccurate.

src/meta-srv/src/election/postgres.rs
```diff
@@ -0,0 +1,488 @@
...
+            client,
+            is_leader: AtomicBool::new(false),
+            infancy: AtomicBool::new(true),
+            leader_watcher: tx,
+            store_key_prefix,
+            candidate_lease_ttl_secs,
+        }))
+    }
+
+    fn _election_key(&self) -> String {
```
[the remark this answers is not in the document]
CookiePieWw: Not used for now so add a underscore. Next pr implementing the election logic will use it.

src/meta-srv/src/election/postgres.rs
CookiePieWw: Yes `EtcdElection` uses false. But seems we should use true here? Since a newly created election is in infancy from my perspective.

src/meta-srv/src/election/postgres.rs
```diff
@@ -0,0 +1,519 @@
...
+        Ok((value.to_string(), expire_time))
+    } else {
+        UnexpectedSnafu {
+            violated: format!(
+                "Invalid value {}, expect node info || {} || expire time",
+                value, LEASE_SEP
+            ),
+        }
+        .fail()
+    }
```
fengjiachun, line 82:
```suggestion
    let (value, expire_time) = value
        .split(LEASE_SEP)
        .collect_tuple()
        .context(UnexpectedSnafu {
            violated: format!(
                "Invalid value {}, expect node info || {} || expire time",
                value, LEASE_SEP
            ),
        })?;
    // Given expire_time is in the format 'YYYY-MM-DD HH24:MI:SS.MS'
    let expire_time = match Timestamp::from_str(expire_time, None) {
        Ok(ts) => ts,
        Err(_) => UnexpectedSnafu {
            violated: format!("Invalid timestamp: {}", expire_time),
        }
        .fail()?,
    };
    Ok((value.to_string(), expire_time))
```

src/meta-srv/src/election/postgres.rs
```diff
@@ -0,0 +1,519 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+
+use common_time::Timestamp;
+use itertools::Itertools;
+use snafu::{ensure, ResultExt};
```
fengjiachun, line 21:
```suggestion
use snafu::{ensure, OptionExt, ResultExt};
```

src/meta-srv/src/election/postgres.rs
```diff
@@ -0,0 +1,519 @@
...
+        format!("{}{}", self.store_key_prefix, CANDIDATES_ROOT)
+    }
+
+    fn candidate_key(&self) -> String {
+        format!("{}{}", self.candidate_root(), self.leader_value)
+    }
+}
+
+#[async_trait::async_trait]
+impl Election for PgElection {
```
fengjiachun, line 130: nit: the order of `impl` members differs from the trait

fengjiachun (approved):
Rest LGTM

src/meta-srv/src/bootstrap.rs
```diff
@@ -275,8 +276,15 @@ async fn create_postgres_client(opts: &MetasrvOptions) -> Result<tokio_postgres:
...
-    let (client, _) = tokio_postgres::connect(postgres_url, NoTls)
+    let (client, connection) = tokio_postgres::connect(postgres_url, NoTls)
         .await
         .context(error::ConnectPostgresSnafu)?;
+
+    tokio::spawn(async move {
+        connection
+            .await
+            .context(error::PostgresExecutionSnafu)
+            .unwrap()
```
WenyXu, line 287: doc required

WenyXu (approved):
Rest LGTM

PR #5227, opened 2024-12-24
fix: correct write cache's metric labels

waynexia:

Several changes are included in this patch.
- Change the behavior of read cache and write cache's configuration. Now user only config for the root dir for cache files. Leave the detailed subdir allocation to DB.
- Change the default path to write cache and read cache. In some cases they are considered breaking.
- Fixes the issue that write cache file's UUID are reported in prometheus layers.

Detailed explanation of path config changes:

From now on, user only config the root file for caches, with default path to `data_home`. And DB will allocate proper paths for different caches, instead of letting user to care about this. An example local fs dir structure is:
```
- <data_home>
  - data/<catalog>/<schema>
  - greptimedb/object_cache
    - read
    - write
```

## PR Checklist

### DIFF
--- a/src/mito2/src/cache/file_cache.rs
+++ b/src/mito2/src/cache/file_cache.rs
@@ -37,8 +37,10 @@ use crate::sst::file::FileId;
 use crate::sst::parquet::helper::fetch_byte_ranges;
 use crate::sst::parquet::metadata::MetadataLoader;
 
-/// Subdirectory of cached files.
-const FILE_DIR: &str = "files/";
+/// Subdirectory of cached files for write.
+///
+/// This must contain three layers, corresponding to [`build_prometheus_metrics_layer`](object_store::layers::build_prometheus_metrics_layer).
+const FILE_DIR: &str = "greptimedb/object_cache/write/";
 
 /// A file cache manages files on local store and evict files based
 /// on size.
--- a/src/mito2/src/cache/file_cache.rs
+++ b/src/mito2/src/cache/file_cache.rs
@@ -40,7 +40,7 @@ use crate::sst::parquet::metadata::MetadataLoader;
 /// Subdirectory of cached files for write.
 ///
 /// This must contain three layers, corresponding to [`build_prometheus_metrics_layer`](object_store::layers::build_prometheus_metrics_layer).
-const FILE_DIR: &str = "greptimedb/object_cache/write/";
+const FILE_DIR: &str = "cache/object/write/";
 
 /// A file cache manages files on local store and evict files based
 /// on size.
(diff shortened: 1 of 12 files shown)


src/mito2/src/cache/file_cache.rs
```diff
@@ -37,8 +37,10 @@ use crate::sst::file::FileId;
 use crate::sst::parquet::helper::fetch_byte_ranges;
 use crate::sst::parquet::metadata::MetadataLoader;
 
-/// Subdirectory of cached files.
-const FILE_DIR: &str = "files/";
+/// Subdirectory of cached files for write.
+///
+/// This must contain three layers, corresponding to [`build_prometheus_metrics_layer`](object_store::layers::build_prometheus_metrics_layer).
```
killme2008, line 42: Just want to adapt to the new layer, and then we break changing the default cache directories. It's not a good idea totally.

waynexia: Despite the layer, this change also simplifies how users config the system

evenyag (approved):

src/mito2/src/cache/file_cache.rs
killme2008: Sure. You can retain the default paths as they were; they are separate matters. Simply the system config but keeps the default values.

waynexia: The previous one is not ideal. They are `greptimedb/object_cache/write/files/` and `object_cache/read/`

killme2008: Got it. My question is, why do we add the `greptimedb/` prefix to the `object_cache/{ read | write}`, it's ugly and unnecessary. If the metric layer needs the three layers of the path, I think we must change it.

src/object-store/src/layers/lru_cache/read_cache.rs
```diff
@@ -20,14 +20,18 @@ use moka::future::Cache;
...
 use crate::metrics::{
     OBJECT_STORE_LRU_CACHE_BYTES, OBJECT_STORE_LRU_CACHE_ENTRIES, OBJECT_STORE_LRU_CACHE_HIT,
     OBJECT_STORE_LRU_CACHE_MISS, OBJECT_STORE_READ_ERROR,
 };
 
 const RECOVER_CACHE_LIST_CONCURRENT: usize = 8;
+/// Subdirectory of cached files for read.
+///
+/// This must contain three layers, corresponding to [`build_prometheus_metrics_layer`](object_store::layers::build_prometheus_metrics_layer).
+const READ_CACHE_DIR: &str = "cache/object/read";
```
killme2008, line 34: The default path before is `{data_home}/object_cache/read` and `{data_home}/object_cache/write`, let's keep them the same as before.

killme2008 (approved):
LGTM

src/mito2/src/cache/file_cache.rs
waynexia: It's now changed to `cache/object/<read|write>`

PR #5231, opened 2024-12-24
feat: introduce the Limiter in frontend to limit the requests by in-flight write bytes size.

zyy17:

Introduce the `Limiter` in frontend to limit the requests by in-flight write bytes size.

## PR Checklist

### DIFF
--- a/src/frontend/src/limiter.rs
+++ b/src/frontend/src/limiter.rs
@@ -1,0 +1,291 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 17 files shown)


src/frontend/src/limiter.rs
```diff
@@ -0,0 +1,291 @@
...
+        }
+    }
+
+    /// Returns the current in-flight write bytes.
+    #[allow(dead_code)]
+    pub fn in_flight_write_bytes(&self) -> u64 {
+        self.in_flight_write_bytes.load(Ordering::Relaxed)
+    }
+
+    fn insert_requests_data_size(&self, request: &InsertRequests) -> usize {
```
waynexia, line 128: I'm wondering if it's possible to read the payload size from protocol layer instead of calculating them ourselves. An extra pass to walk the entire data might be costly.

waynexia (commented):
Can we hold the frontend from receiving new requests instead of rejecting them?

sunng87 (approved):

sunng87:
@waynexia client may still get tcp level error like send buffer is full if it sends faster than server can read.

waynexia (approved):

PR #5234, opened 2024-12-25
fix: flow compare null values

discord9:

fix a bug in flow which fail to handle null array comparison:
- forget to handle null scalar case causing compare with null array and some non-null array

future works:
- need to check all logic in flow which could handle null array incorrectly

## PR Checklist


src/flow/src/compute/render/reduce.rs
```diff
@@ -398,20 +399,53 @@ fn reduce_batch_subgraph(
...
+                            .get(key_idx)
+                            .context(InternalSnafu {
+                                reason: format!(
+                                    "Key index out of bound, expected at most {} but got {}",
+                                    output_type.column_types.len(),
+                                    key_idx
+                                ),
+                            })?
+                            .clone()
+                            .scalar_type;
```
waynexia, line 424: Can we place this above 2 for-loops?

tests/cases/standalone/common/flow/flow_null.sql
waynexia, line 1: suggs:
- empty line in the end
- don't select source table as their result is obvious and is not what we are testing for. This can make the test more concise

zhongzc (approved):
most LGTM

WenyXu (approved):

waynexia (approved):

PR #5242, opened 2024-12-26
refactor: support to convert time string to timestamp in `convert_value()`

zyy17:

This PR supports converting the time string(for example, `2024-12-26 10:00:25.899`) to the `Timestamp` type in `convert_value()`.  It fixes the insert data error when users insert timestamp in string format by using a prepared statement. The logic is compatible with native MySQL.

## PR Checklist

### DIFF
--- a/src/servers/src/mysql/helper.rs
+++ b/src/servers/src/mysql/helper.rs
@@ -17,7 +17,9 @@ use std::time::Duration;
 
 use chrono::NaiveDate;
 use common_query::prelude::ScalarValue;
+use common_time::Timestamp;
 use datatypes::prelude::ConcreteDataType;
+use datatypes::types::TimestampType;
 use datatypes::value::{self, Value};
 use itertools::Itertools;
 use opensrv_mysql::{to_naive_datetime, ParamValue, ValueInner};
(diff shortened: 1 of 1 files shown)


src/servers/src/mysql/helper.rs
```diff
@@ -161,7 +163,28 @@ pub fn convert_value(param: &ParamValue, t: &ConcreteDataType) -> Result<ScalarV
...
+                        Ok(ScalarValue::TimestampNanosecond(Some(ts.value()), None))
+                    }
+                    TimestampType::Microsecond(_) => {
+                        Ok(ScalarValue::TimestampMicrosecond(Some(ts.value()), None))
+                    }
+                    TimestampType::Millisecond(_) => {
+                        Ok(ScalarValue::TimestampMillisecond(Some(ts.value()), None))
+                    }
+                    TimestampType::Second(_) => {
+                        Ok(ScalarValue::TimestampSecond(Some(ts.value()), None))
```
sunng87, line 184: I'm afraid we cannot simply use `ts.value` here for different time unit. The input timestamp string will have no unit information so the timestamp we parsed in line 167 will have a fixed unit.

zyy17: Maybe we can use `covert_to()`, for example:

```
let ts = Timestamp::from_str_utc(&String::from_utf8_lossy(b))
                    .map_err(|e| {
                        error::MysqlValueConversionSnafu {
                            err_msg: e.to_string(),
                        }
                        .build()
                    })?
                    .convert_to(ts_type.unit())
                    .ok_or_else(|| {
                        error::MysqlValueConversionSnafu {
                            err_msg: "Overflow when converting timestamp to target unit"
                                .to_string(),
                        }
                        .build()
                    })?;
```

sunng87: sounds good. we can add some unit tests to verify this.

sunng87 (approved):

fengjiachun (approved):
LGTM

PR #5247, opened 2024-12-27
fix: disable path label in opendal for now

shuiyisong:

Since switching to openDAL's Prometheus layer, we have not been able to set the layer multiple times, causing the path label to include the index`s uuid symbol, which causes the size of the label to explode.

Disable the path label for now because we're not using it for debugging anyway. We may reopen it later if openDAL supports dynamic path trim or some hook to decide path label dynamically.

## PR Checklist

### DIFF
--- a/src/object-store/src/layers.rs
+++ b/src/object-store/src/layers.rs
@@ -30,13 +30,18 @@ mod prometheus {
     /// pattern `<data|index>/catalog/schema/table_id/...` OR `greptimedb/object_cache/<read|write>/...`
     ///
     /// We'll get the data/catalog/schema from path.
-    pub fn build_prometheus_metrics_layer(with_path_label: bool) -> PrometheusLayer {
+    pub fn build_prometheus_metrics_layer(_with_path_label: bool) -> PrometheusLayer {
         PROMETHEUS_LAYER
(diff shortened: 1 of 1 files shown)


waynexia (approved):

src/object-store/src/layers.rs
```diff
@@ -30,13 +30,18 @@ mod prometheus {
...
     /// We'll get the data/catalog/schema from path.
-    pub fn build_prometheus_metrics_layer(with_path_label: bool) -> PrometheusLayer {
+    pub fn build_prometheus_metrics_layer(_with_path_label: bool) -> PrometheusLayer {
         PROMETHEUS_LAYER
             .get_or_init(|| {
-                let path_level = if with_path_label { 3 } else { 0 };
+                // let path_level = if with_path_label { 3 } else { 0 };
+
+                // opendal doesn't support dynamic path label trim
+                // we have uuid in index path, which casues the label size to explode
```
waynexia, line 39:
```suggestion
                // we have uuid in index path, which causes the label size to explode
```

WenyXu (approved):

sunng87 (approved):

PR #5249, opened 2024-12-27
feat: add election logic for PgElection

CookiePieWw:

#5208

As title. No integration tests for now.

## PR Checklist


src/meta-srv/src/election/postgres.rs
```diff
@@ -16,18 +16,30 @@ use std::sync::atomic::{AtomicBool, Ordering};
...
-use crate::election::{Election, LeaderChangeMessage, CANDIDATES_ROOT, ELECTION_KEY};
+use crate::election::{Election, LeaderChangeMessage, LeaderKey, CANDIDATES_ROOT, ELECTION_KEY};
 use crate::error::{
-    DeserializeFromJsonSnafu, PostgresExecutionSnafu, Result, SerializeToJsonSnafu, UnexpectedSnafu,
+    DeserializeFromJsonSnafu, NoLeaderSnafu, PostgresExecutionSnafu, Result, SerializeToJsonSnafu,
+    UnexpectedSnafu,
 };
 use crate::metasrv::{ElectionRef, LeaderValue, MetasrvNodeInfo};
 
+// TODO: The lock id should be configurable.
```
CookiePieWw, line 35:
```suggestion
// TODO(CookiePie): The lock id should be configurable.
```

src/meta-srv/src/election/postgres.rs
```diff
@@ -343,6 +457,148 @@ impl PgElection {
...
+        // Case 2: Current leader exists and the lease expired. Raise an error to re-init campaign.
+        //         If the leader failed to renew the lease, its session will expire and the lock will be released.
+        ensure!(expire_time > current, NoLeaderSnafu);
+        // Case 3: All checks passed, just return.
+        Ok(())
+    }
+
+    /// Step down the leader. The leader should delete the key and notify the leader watcher.
+    /// Do not check if the deletion is successful, since the key may be deleted by other followers.
+    /// Caution: Should only step down while holding the advisory lock.
```
WenyXu, line 522: nit
```suggestion
    /// Step down the leader. The leader should delete the key and notify the leader watcher.
    /// 
    /// __DO NOT__ check if the deletion is successful, since the key may be deleted by other followers.
    ///
    /// ## Caution:
    /// Should only step down while holding the advisory lock.
```

src/meta-srv/src/election/postgres.rs
```diff
@@ -343,6 +457,148 @@ impl PgElection {
 
         Ok(res.len() == 1)
     }
+
+    async fn leader_action(&self) -> Result<()> {
```
WenyXu, line 461:
```suggestion
    /// Handles the actions of a leader in the election process.
    /// 
    /// This function performs the following checks and actions:
    /// 
    /// - **Case 1**: If the current instance believes it is the leader from the previous term,
    ///   it attempts to renew the lease. It checks if the lease is still valid and either renews it
    ///   or steps down if it has expired.
    /// 
    ///   - **Case 1.1**: If the instance is still the leader and the lease is valid, it renews the lease
    ///     by updating the value associated with the election key.
    ///   - **Case 1.2**: If the instance is still the leader but the lease has expired, it logs a warning
    ///     and steps down, initiating a new campaign for leadership.
    ///   - **Case 1.3**: If the instance is not the leader (which is a rare scenario), it logs a warning
    ///     indicating that it still holds the lock and steps down to re-initiate the campaign.
    ///   - **Case 1.4**: If no lease information is found, it also steps down and re-initiates the campaign.
    /// 
    /// - **Case 2**: If the current instance does not believe it is the leader, it calls the
    ///   `elected` method to participate in the election process, indicating that it is ready to
    ///   contend for leadership.
    async fn leader_action(&self) -> Result<()> {
```

src/meta-srv/src/election/postgres.rs
```diff
@@ -198,11 +272,51 @@ impl Election for PgElection {
     }
 
     async fn campaign(&self) -> Result<()> {
```
WenyXu, line 274:
```suggestion
    /// Attempts to acquire leadership by executing a campaign. This function continuously checks
    /// if the current instance can become the leader by acquiring an advisory lock in the PostgreSQL database.
    ///
    /// The function operates in a loop, where it:
    ///
    /// 1. Waits for a predefined interval before attempting to acquire the lock again.
    /// 2. Executes the `CAMPAIGN` SQL query to try to acquire the advisory lock.
    /// 3. Checks the result of the query:
    ///    - If the lock is successfully acquired (result is true), it calls the `leader_action` method
    ///      to perform actions as the leader.
    ///    - If the lock is not acquired (result is false), it calls the `follower_action` method
    ///      to perform actions as a follower.
    async fn campaign(&self) -> Result<()> {
```

src/meta-srv/src/election/postgres.rs
```diff
@@ -198,11 +272,51 @@ impl Election for PgElection {
...
-        todo!()
+        let mut keep_alive_interval =
+            tokio::time::interval(Duration::from_secs(META_KEEP_ALIVE_INTERVAL_SECS));
```
WenyXu, line 276: Should we use `MissedTickBehavior::Skip`?
```suggestion
            tokio::time::interval(Duration::from_secs(META_KEEP_ALIVE_INTERVAL_SECS))
                .set_missed_tick_behavior(MissedTickBehavior::Skip);
```

src/meta-srv/src/election/postgres.rs
```diff
@@ -343,6 +457,148 @@ impl PgElection {
...
+            name: self.leader_value.clone().into_bytes(),
+            key: key.clone().into_bytes(),
+            ..Default::default()
+        };
+        if self
+            .is_leader
+            .compare_exchange(true, false, Ordering::Relaxed, Ordering::Relaxed)
+            .is_ok()
+        {
+            self.delete_value(&key).await?;
```
fengjiachun, line 535: Is it possible that this operation mistakenly deletes the key written by the new leader?

CookiePieWw: The deletion will only happens when the metasrv holds an advisory lock. If a new leader is elected, it should have held the lock to prevent others from the deletion.

src/meta-srv/src/bootstrap.rs
```diff
@@ -229,7 +233,15 @@ pub async fn metasrv_builder(
             let kv_backend = PgStore::with_pg_client(pg_client)
                 .await
                 .context(error::KvBackendSnafu)?;
-            (kv_backend, None)
+            let election_client = create_postgres_client(opts).await?;
+            let election = PgElection::with_pg_client(
+                opts.server_addr.clone(),
+                election_client,
+                opts.store_key_prefix.clone(),
+                CANDIDATE_LEASE_SECS,
```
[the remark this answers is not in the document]
CookiePieWw: > I think the two options (CANDIDATE_LEASE_SECS and KEEP_ALIVE_INTERVAL_SECS) should control the entire election backend.

Yes it is. The newly added param is just for unit test since it's too long for tests to wait for expiration.

src/meta-srv/src/election/postgres.rs
CookiePieWw: etcd uses `.set_missed_tick_behavior(MissedTickBehavior::Delay);`, so I changed it to `Delay`.

src/meta-srv/src/bootstrap.rs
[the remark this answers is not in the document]
fengjiachun: Got it.

src/meta-srv/src/election/postgres.rs
fengjiachun: Got it

fengjiachun (approved):
Rest LGTM

WenyXu (approved):

PR #5251, opened 2024-12-27
fix(flow): flow's table schema cache

discord9:

fix a bug in which flow's table schema cache will become incorrect after table altered

- check from kvbackended(cached so query is cheap) table info manager for real table schema instead

in the future also need to check table schema version to prevent using a altered table as flow's src/sink, but this is not in the scope of this fix

## PR Checklist


src/flow/src/adapter/flownode_impl.rs
```diff
@@ -34,11 +34,19 @@ use crate::error::InternalSnafu;
 use crate::metrics::METRIC_FLOW_TASK_COUNT;
 use crate::repr::{self, DiffRow};
 
-fn to_meta_err(err: crate::error::Error) -> common_meta::error::Error {
+/// TODO(discord9): add location
```
killme2008, line 37: What does the todo mean? Looks like the function already accepts the location parameter.

src/flow/src/adapter/flownode_impl.rs
```diff
@@ -156,15 +173,14 @@ impl Flownode for FlowWorkerManager {
...
-                    .table_repr
-                    .get_by_table_id(&table_id)
-                    .map(|r| r.1)
-                    .and_then(|id| ctx.schema.get(&id))
-                    .map(|desc| &desc.names)
-                    .context(UnexpectedSnafu {
-                        err_msg: format!("Table not found: {}", table_id),
-                    })?;
+
+                // TODO(discord9): also check schema version so that altered table can be reported
```
killme2008, line 177: Why not fix it right now?

src/flow/src/adapter/table_source.rs
```diff
@@ -25,18 +26,57 @@ use crate::adapter::TableName;
 use crate::error::{
     Error, ExternalSnafu, TableNotFoundMetaSnafu, TableNotFoundSnafu, UnexpectedSnafu,
 };
-use crate::repr::RelationDesc;
+use crate::repr::{ColumnType, RelationDesc, RelationType};
+
+#[async_trait::async_trait]
+pub trait FlowTableSource: Send + Sync + std::fmt::Debug {
```
killme2008, line 32: Please add the doc.

src/flow/src/adapter/table_source.rs
```diff
@@ -126,3 +169,89 @@ impl TableSource {
...
+        }
+        TableNotFoundSnafu {
+            name: format!("Table id = {:?}, couldn't found table desc", table_id),
+        }
+        .fail()?
+    }
+
+    async fn table_name_from_id(&self, table_id: &TableId) -> Result<TableName, Error> {
+        for (id, name, _desc) in &self.id_names_to_desc {
+            if id == table_id {
```
killme2008, line 230: This is so inefficient, can we improve it?

src/flow/src/adapter/flownode_impl.rs
discord9: > Why not fix it right now?

Tried once, and realize it's a somewhere large change(need change metadata all the way, change proto and add compat tests etc) so want to put it in another PR

killme2008 (approved):
LGTM

killme2008:
@zhongzc PTAL

src/flow/src/adapter/table_source.rs
```diff
@@ -25,18 +28,58 @@ use crate::adapter::TableName;
...
+        let id = self.table_id_from_name(name).await?;
+        self.table_from_id(&id).await
+    }
+    async fn table_from_id(&self, table_id: &TableId) -> Result<RelationDesc, Error>;
+}
 
 /// mapping of table name <-> table id should be query from tableinfo manager
-pub struct TableSource {
+#[derive(Clone)]
+pub struct KvBackendTableSource {
```
fengjiachun, line 49: I think it's fine to name it `TableSource` since the KvBackend is part of XXXManager.

src/flow/src/adapter/table_source.rs
```diff
@@ -126,3 +172,107 @@ impl TableSource {
...
     }
 }
+
+impl std::fmt::Debug for ManagedTableSource {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("KvBackendTableSource").finish()
+    }
+}
+
+pub struct FlowDummyTableSource {
```
zhongzc, line 182: move to the test mod.

tests/cases/standalone/common/flow/flow_aft_alter.sql
```diff
@@ -0,0 +1,64 @@
+-- test if flow can get table schema correctly after table have benn altered
```
zhongzc, line 1:
```suggestion
-- test if flow can get table schema correctly after table have been altered
```

zhongzc (approved):

PR #5252, opened 2024-12-27
feat: override `__sequence` on creating SST to save space and CPU

waynexia:

The sequence is `u64` and is allocated with write request. They are different from row to row and thus has bad compression efficiency in both CPU and space. We can assign the greatest sequence number to the file to reduce this overhead while keeping this mechanism works.

In TSBS test scenario with 100M rows, it can save 48MB of disk space. And in the metric monitor scenario, it can save up to 50%~60% in file size. We can also observe an (unstable) flush performance improvement.

## PR Checklist

### DIFF
--- a/src/mito2/src/engine/basic_test.rs
+++ b/src/mito2/src/engine/basic_test.rs
@@ -580,7 +580,7 @@ async fn test_region_usage() {
     flush_region(&engine, region_id, None).await;
 
     let region_stat = region.region_statistic();
-    assert_eq!(region_stat.sst_size, 2773);
+    assert!(region_stat.sst_size > 0); // Chief says this assert can ensure the size is counted.
     assert_eq!(region_stat.num_rows, 10);
 
     // region total usage
--- a/src/mito2/src/engine/basic_test.rs
+++ b/src/mito2/src/engine/basic_test.rs
@@ -580,7 +580,7 @@ async fn test_region_usage() {
     flush_region(&engine, region_id, None).await;
 
     let region_stat = region.region_statistic();
-    assert_eq!(region_stat.sst_size, 2790);
+    assert_eq!(region_stat.sst_size, 2773);
     assert_eq!(region_stat.num_rows, 10);
 
     // region total usage
--- a/src/mito2/src/flush.rs
+++ b/src/mito2/src/flush.rs
@@ -15,6 +15,7 @@
 //! Flush related utilities and structs.
 
 use std::collections::HashMap;
+use std::num::NonZeroU64;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 
@@ -337,18 +338,6 @@ impl RegionFlushTask {
         }
 
         let memtables = version.memtables.immutables();
-        // get max_sequence from all non-empty memetables
-        let max_sequence = memtables
-            .iter()
-            .filter_map(|m| {
-                if !m.is_empty() {
-                    Some(m.stats().max_sequence())
-                } else {
-                    None
-                }
-            })
-            .max();
-
         let mut file_metas = Vec::with_capacity(memtables.len());
         let mut flushed_bytes = 0;
         for mem in memtables {
@@ -357,6 +346,7 @@ impl RegionFlushTask {
                 continue;
             }
 
+            let max_sequence = mem.stats().max_sequence();
             let file_id = FileId::random();
             let iter = mem.iter(None, None)?;
             let source = Source::Iter(iter);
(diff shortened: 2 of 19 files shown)


MichaelScofield:
good for merging ssts when ingesting them, better if the `__sequence` can be removed

evenyag:
> good for merging ssts when ingesting them, better if the `__sequence` can be removed

We can share the same sequence because here we assume we always dedup all rows in the SST in non append-only mode.

src/mito2/src/sst/parquet/writer.rs
```diff
@@ -112,9 +113,11 @@ where
     pub async fn write_all(
         &mut self,
         mut source: Source,
+        override_sequence: Option<SequenceNumber>, // override the `sequence` field from `Source`
         opts: &WriteOptions,
```
evenyag, line 117: Can we move the `override_sequence` to `WriteOptions`. It should be an option to control the writer.

src/mito2/src/memtable/time_series.rs
```diff
@@ -199,6 +201,10 @@ impl Memtable for TimeSeriesMemtable {
         // so that we can ensure writing to memtable will succeed.
         self.update_stats(local_stats);
 
+        // update max_sequence
+        let sequence = kvs.mutation.sequence;
```
evenyag, line 205: The sequence in the `kvs` is the start sequence in the mutation. We can encapsulate a method to get the maximum sequence.
https://github.com/GreptimeTeam/greptimedb/blob/11bab0c47cce4fe09cd37763f7e1877d11178c7b/src/mito2/src/memtable/key_values.rs#L95-L104

We can also add a max sequence field to the metrics and update the sequence in `update_stats()`.

src/mito2/src/memtable/partition_tree.rs
```diff
@@ -140,6 +141,12 @@ impl Memtable for PartitionTreeMemtable {
 
         self.update_stats(&metrics);
 
+        // update max_sequence
+        if res.is_ok() {
+            let sequence = kvs.mutation.sequence;
```
evenyag, line 146: This is the start sequence of the mutation.

src/mito2/src/engine/basic_test.rs
```diff
@@ -580,7 +580,7 @@ async fn test_region_usage() {
     flush_region(&engine, region_id, None).await;
 
     let region_stat = region.region_statistic();
-    assert_eq!(region_stat.sst_size, 2790);
+    assert_eq!(region_stat.sst_size, 2773);
```
evenyag, line 583: Maybe we can assert the size > 0.

src/mito2/src/flush.rs
```diff
@@ -337,6 +337,18 @@ impl RegionFlushTask {
...
+        let max_sequence = memtables
+            .iter()
+            .filter_map(|m| {
+                if !m.is_empty() {
+                    Some(m.stats().max_sequence())
+                } else {
+                    None
+                }
+            })
+            .max();
```
evenyag, line 350: I prefer not to share the max sequence across memtables.

src/mito2/src/flush.rs
```diff
@@ -382,6 +395,7 @@ impl RegionFlushTask {
                 index_file_size: sst_info.index_metadata.file_size,
                 num_rows: sst_info.num_rows as u64,
                 num_row_groups: sst_info.num_row_groups,
+                max_sequence: max_sequence.unwrap_or(0),
```
evenyag, line 398: We can get the max sequence from this memtable.
```suggestion
                max_sequence: max_sequence.unwrap_or(0),
```

src/mito2/src/sst/file.rs
```diff
@@ -134,6 +134,8 @@ pub struct FileMeta {
     /// the default value `0` doesn't means the file doesn't contains any rows,
     /// but instead means the number of rows is unknown.
     pub num_row_groups: u64,
+    /// Max sequence in this file
+    pub max_sequence: u64,
```
evenyag, line 138: Maybe we should rename it to `sequence`? It is a global sequence for all rows in the file.

TODO: While ingesting SSTs, we can also set this field to assign a sequence to the whole SST. We can do it in the mito engine so the mito engine can correctly bump the sequence number and assign it to the file meta. @MichaelScofield

evenyag, line 138: idea: Can we use `Option<NonZeroU64>`? So we won't forget that this field can be zero.

src/mito2/src/engine/basic_test.rs
```diff
@@ -580,7 +580,7 @@ async fn test_region_usage() {
     flush_region(&engine, region_id, None).await;
 
     let region_stat = region.region_statistic();
-    assert_eq!(region_stat.sst_size, 2790);
+    assert!(region_stat.sst_size > 0); // Chief says this assert can ensure the size is counted.
```
evenyag, line 583: 😕

evenyag (approved):

fengjiachun (approved):
LGTM

PR #5254, opened 2024-12-28
fix: implement a CacheStrategy to ensure compaction use cache correctly

evenyag:

Some changes revert https://github.com/GreptimeTeam/greptimedb/pull/5135/

This PR implements a `CompactionStrategy` to replace `Option<CacheManagerRef>` so we can control how compaction uses caches.
- For queries, the strategy forwards all operations to the cache manager.
- The strategy only uses the write cache and the metadata cache for compactions.
- The `Disabled` strategy means disable all caches.

## PR Checklist

### DIFF
--- a/src/mito2/src/cache.rs
+++ b/src/mito2/src/cache.rs
@@ -66,6 +66,8 @@ pub enum CacheStrategy {
     /// Enables the write cache so that the compaction can read files cached
     /// in the write cache and write the compacted files back to the write cache.
     Compaction(CacheManagerRef),
+    /// No cache.
+    Disabled,
 }
 
 impl CacheStrategy {
@@ -86,6 +88,7 @@ impl CacheStrategy {
                     .get_parquet_meta_data(region_id, file_id)
                     .await
             }
+            CacheStrategy::Disabled => None,
         }
     }
 
@@ -102,6 +105,7 @@ impl CacheStrategy {
             CacheStrategy::Compaction(cache_manager) => {
                 cache_manager.get_parquet_meta_data_from_mem_cache(region_id, file_id)
             }
+            CacheStrategy::Disabled => None,
         }
     }
 
@@ -119,6 +123,7 @@ impl CacheStrategy {
             CacheStrategy::Compaction(cache_manager) => {
                 cache_manager.put_parquet_meta_data(region_id, file_id, metadata);
             }
+            CacheStrategy::Disabled => {}
         }
     }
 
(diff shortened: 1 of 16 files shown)


src/mito2/src/cache.rs
```diff
@@ -55,6 +55,193 @@ const FILE_TYPE: &str = "file";
...
+    /// Doesn't disable any cache.
+    Normal(CacheManagerRef),
+    /// Strategy for compaction.
+    /// Disables some caches that are not needed during compaction.
+    /// Enables the write cache so that the compaction can read files cached
+    /// in the write cache and write the compacted files back to the write cache.
+    Compaction(CacheManagerRef),
+    /// No cache.
+    Disabled,
+}
```
waynexia, line 71: naming strategy with target use scenarios seems misleading. Can we put how it actually works in the name? Like `EnableAll`, `MetaOnly`, `Disabled` (feel free to adjust these names)

evenyag: `EnableAll` is great. But the compaction needs a special cache strategy so I prefer to keep using `Compaction` as the strategy name.

evenyag:
https://github.com/GreptimeTeam/greptimedb/actions/runs/12532626343/job/34951585899?pr=5254

I don't know why all runners already installed fuzz and gc-bin......
```
warning: tool `rust-analyzer` is already installed, remove it from `/home/runner/.cargo/bin`, then run `rustup update` to have rustup manage this tool.
warning: tool `rustfmt` is already installed, remove it from `/home/runner/.cargo/bin`, then run `rustup update` to have rustup manage this tool.
warning: tool `cargo-fmt` is already installed, remove it from `/home/runner/.cargo/bin`, then run `rustup update` to have rustup manage this tool.
    Updating crates.io index
 Downloading crates ...
  Downloaded cargo-fuzz v0.12.0
error: binary `cargo-fuzz` already exists in destination
Add --force to overwrite
    Updating crates.io index
 Downloading crates ...
  Downloaded cargo-gc-bin v0.1.4
error: binary `cargo-gc` already exists in destination
Add --force to overwrite
     Summary Failed to install cargo-fuzz, cargo-gc-bin (see error(s) above).
error: some crates failed to install
```

src/mito2/src/read/scan_region.rs
```diff
@@ -582,7 +564,7 @@ impl ScanInput {
             predicate: None,
             memtables: Vec::new(),
             files: Vec::new(),
-            cache_manager: None,
+            cache_strategy: CacheStrategy::EnableAll(CacheManagerRef::default()),
```
v0y4g3r, line 567: Not sure if `CacheStrategy::EnableAll(CacheManagerRef::default())` makes any sense when `CacheManagerRef::default()` does not contain any cache?

v0y4g3r (commented):
Rest LGTM

src/mito2/src/read/scan_region.rs
evenyag: fixed in 58b83032f111191c5ec17abcff4c2aef2dc1ec29

waynexia (approved):

PR #5261, opened 2024-12-31
feat: support elasticsearch `_bulk` API to ingest logs

zyy17:

1. Support elasticsearch `_bulk` API to ingest logs;
2. Mock some elasticsearch APIs that are used for Logstash;

## PR Checklist

### DIFF
--- a/src/servers/src/elasticsearch.rs
+++ b/src/servers/src/elasticsearch.rs
@@ -34,6 +34,9 @@ use crate::error::{
 use crate::http::event::{
     ingest_logs_inner, LogIngesterQueryParams, LogState, GREPTIME_INTERNAL_IDENTITY_PIPELINE_NAME,
 };
+use crate::metrics::{
+    METRIC_ELASTICSEARCH_LOGS_DOCS_COUNT, METRIC_ELASTICSEARCH_LOGS_INGESTION_ELAPSED,
+};
 
 // The headers for every response of Elasticsearch API.
 static ELASTICSEARCH_HEADERS: Lazy<HeaderMap> = Lazy::new(|| {
@@ -97,9 +100,11 @@ pub async fn handle_bulk_api(
     // The `schema` is already set in the query_ctx in auth process.
     query_ctx.set_channel(Channel::Elasticsearch);
 
+    let db = params.db.unwrap_or_else(|| "public".to_string());
+
     // Record the ingestion time histogram.
-    let _timer = crate::metrics::METRIC_ELASTICSEARCH_LOGS_INGESTION_ELAPSED
-        .with_label_values(&[params.db.unwrap_or_else(|| "public".to_string()).as_str()])
+    let _timer = METRIC_ELASTICSEARCH_LOGS_INGESTION_ELAPSED
+        .with_label_values(&[&db])
         .start_timer();
 
     let table = if let Some(table) = params.table {
@@ -164,6 +169,11 @@ pub async fn handle_bulk_api(
             )),
         );
     }
+
+    // Record the number of documents ingested.
+    METRIC_ELASTICSEARCH_LOGS_DOCS_COUNT
+        .with_label_values(&[&db])
+        .inc_by(log_num as u64);
 
     (
         StatusCode::OK,
--- a/src/servers/src/elasticsearch.rs
+++ b/src/servers/src/elasticsearch.rs
@@ -21,7 +21,8 @@ use axum::http::{HeaderMap, HeaderName, HeaderValue, StatusCode};
 use axum::response::IntoResponse;
 use axum::{Extension, TypedHeader};
 use common_error::ext::ErrorExt;
-use common_telemetry::{debug, warn};
+use common_telemetry::{debug, error};
+use once_cell::sync::Lazy;
 use serde_json::{json, Deserializer, Value};
 use session::context::{Channel, QueryContext};
 use snafu::{ensure, ResultExt};
@@ -33,6 +34,20 @@ use crate::error::{
 use crate::http::event::{
     ingest_logs_inner, LogIngesterQueryParams, LogState, GREPTIME_INTERNAL_IDENTITY_PIPELINE_NAME,
 };
+
+// The headers for every response of Elasticsearch API.
+static ELASTICSEARCH_HEADERS: Lazy<HeaderMap> = Lazy::new(|| {
+    HeaderMap::from_iter([
+        (
+            axum::http::header::CONTENT_TYPE,
+            HeaderValue::from_static("application/json"),
+        ),
+        (
+            HeaderName::from_static("x-elastic-product"),
+            HeaderValue::from_static("Elasticsearch"),
+        ),
+    ])
+});
 
 // The fake version of Elasticsearch and used for `_version` API.
 const ELASTICSEARCH_VERSION: &str = "8.16.0";
@@ -84,7 +99,7 @@ pub async fn handle_bulk_api(
 
     // Record the ingestion time histogram.
     let _timer = crate::metrics::METRIC_ELASTICSEARCH_LOGS_INGESTION_ELAPSED
-        .with_label_values(&[params.db.unwrap_or("public".to_string()).as_str()])
+        .with_label_values(&[params.db.unwrap_or_else(|| "public".to_string()).as_str()])
         .start_timer();
 
     let table = if let Some(table) = params.table {
@@ -137,7 +152,7 @@ pub async fn handle_bulk_api(
     )
     .await
     {
-        warn!("Failed to ingest logs: {:?}", e);
+        error!(e; "Failed to ingest logs");
         return (
             status_code_to_http_status(&e.status_code()),
             elasticsearch_headers(),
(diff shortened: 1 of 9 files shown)


src/servers/src/elasticsearch.rs
```diff
@@ -0,0 +1,362 @@
...
+                    StatusCode::BAD_REQUEST,
+                    e.to_string().as_str(),
+                )),
+            );
+        }
+    };
+    let log_num = log_values.len();
+
+    // The `schema` is already set in the query_ctx in auth process.
+    query_ctx.set_channel(Channel::Http);
```
sunng87, line 119: We can add a new channel type `elastic`. This http is mostly for our own protocols via http.

src/servers/src/elasticsearch.rs
```diff
@@ -0,0 +1,362 @@
...
+use session::context::{Channel, QueryContext};
+use snafu::{ensure, ResultExt};
+
+use crate::error::{InvalidElasticsearchInputSnafu, ParseJsonSnafu, Result as ServersResult};
+use crate::http::event::{
+    ingest_logs_inner, LogIngesterQueryParams, LogState, GREPTIME_INTERNAL_IDENTITY_PIPELINE_NAME,
+};
+
+// The fake version of Elasticsearch and used for `_version` API.
+const ELASTICSEARCH_VERSION: &str = "8.16.0";
```
sunng87, line 34: nit: if es client allows, we can add a surfix `-greptimedb-{version}`

src/servers/src/elasticsearch.rs
```diff
@@ -0,0 +1,362 @@
...
+    .await
+    {
+        warn!("Failed to ingest logs: {:?}", e);
+        return (
+            StatusCode::INTERNAL_SERVER_ERROR,
+            elasticsearch_headers(),
+            axum::Json(write_bulk_response(
+                start.elapsed().as_millis() as i64,
+                0,
+                StatusCode::INTERNAL_SERVER_ERROR,
```
sunng87, line 138: We should use the status_code from this `e`. It may not be our fault.

sunng87: and use the message of `e`

sunng87 (commented):
Looks mostly good to me. Is there any chance to specify and invoke pipeline during the ingestion process? We can address it in next patch.

src/servers/src/elasticsearch.rs
```diff
@@ -0,0 +1,363 @@
...
+        None,
+        table,
+        log_values,
+        Arc::new(query_ctx),
+    )
+    .await
+    {
+        warn!("Failed to ingest logs: {:?}", e);
+        return (
+            StatusCode::INTERNAL_SERVER_ERROR,
```
sunng87, line 139:
```suggestion
            e.status_code(),
```

zyy17: We should return a legal http status code here. The custom status code has already been returned in the HTTP body.

sunng87: I see. I remember there is a method to convert our status code to corresponding http status.

sunng87: see `status_code_to_http_status`

zyy17:
> Looks mostly good to me. Is there any chance to specify and invoke pipeline during the ingestion process? We can address it in next patch.

It's already in the request parameter. You can use &pipeline_name=xxx to specify the pipeline. Default it's greptime_identity.

sunng87 (approved):

src/servers/src/elasticsearch.rs
```diff
@@ -0,0 +1,366 @@
...
+        "Received bulk request, params: {:?}, payload: {:?}",
+        params, payload
+    );
+
+    // The `schema` is already set in the query_ctx in auth process.
+    query_ctx.set_channel(Channel::Elasticsearch);
+
+    // Record the ingestion time histogram.
+    let _timer = crate::metrics::METRIC_ELASTICSEARCH_LOGS_INGESTION_ELAPSED
+        .with_label_values(&[params.db.unwrap_or("public".to_string()).as_str()])
```
killme2008, line 87: Prefer `unwarp_or_else(|| "public".to_string())` to avoid `to_string` when db is present.

src/servers/src/elasticsearch.rs
```diff
@@ -0,0 +1,366 @@
...
+        log_state.log_handler,
+        pipeline,
+        None,
+        table,
+        log_values,
+        Arc::new(query_ctx),
+    )
+    .await
+    {
+        warn!("Failed to ingest logs: {:?}", e);
```
killme2008, line 140: Looks like `error!` is more proper:
```suggestion
        error!("Failed to ingest logs: {:?}"; e);
```

src/servers/src/elasticsearch.rs
```diff
@@ -0,0 +1,366 @@
...
+            "items": [
+                { "create": { "status": status_code, "error": { "type": "illegal_argument_exception", "reason": error_reason } } }
+            ]
+        })
+    }
+}
+
+/// Returns the headers for every response of Elasticsearch API.
+pub fn elasticsearch_headers() -> HeaderMap {
+    HeaderMap::from_iter([
```
killme2008, line 212: Can we use lazy static to create the header map and clone it every time?

Then we can remove this function, but just clone the global variable instead.

src/servers/src/elasticsearch.rs
```diff
@@ -0,0 +1,366 @@
...
+            );
+            is_document = true;
+            continue;
+        }
+
+        // It means the second object is the document data.
+        if is_document {
+            // If the msg_field is provided, fetch the value of the field from the document data.
+            if let Some(msg_field) = msg_field {
+                if let Some(Value::String(message)) = v.get(msg_field) {
```
killme2008, line 275: Can we use `v.remove(msg_field)` to consume the message? Then we can avoid the `to_string()` in the if body.

src/servers/src/elasticsearch.rs
```diff
@@ -0,0 +1,366 @@
...
+                        // If the message is not a valid JSON, just use the original message as the log value.
+                        Err(_) => Value::String(message.to_string()),
+                    };
+                    log_values.push(value);
+                }
+            } else {
+                log_values.push(v);
+            }
+            is_document = false;
+        }
```
killme2008, line 287: If `is_document=false`, do we need to print a warning message? Otherwise, it would be hard to debug such requests.

src/servers/src/elasticsearch.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    // Check if the input is empty.
+    ensure!(
+        !values.is_empty(),
+        InvalidElasticsearchInputSnafu {
+            reason: "empty bulk request".to_string(),
+        }
+    );
+
+    let mut log_values: Vec<Value> = Vec::new();
```
killme2008, line 249: Is the initial capacity of `log_values` `values.len()/2`?

src/servers/src/http.rs
```diff
@@ -752,6 +766,82 @@ impl HttpServer {
...
+            // Return fake response for Elasticsearch ilm request.
+            .route(
+                "/_ilm/policy/*path",
+                routing::any((
+                    HttpStatusCode::OK,
+                    elasticsearch::elasticsearch_headers(),
+                    axum::Json(serde_json::json!({})),
+                )),
+            )
+            // Return fake response for Elasticsearch index template request.
```
killme2008, line 790: Is it possible to return the actual index template by rendering the table schema?

src/servers/src/http/event.rs
```diff
@@ -64,15 +64,30 @@ lazy_static! {
         ContentType::from_str(CONTENT_TYPE_PROTOBUF_STR).unwrap();
 }
 
+/// LogIngesterQueryParams is used for query params of log ingester API.
 #[derive(Debug, Default, Serialize, Deserialize)]
 pub struct LogIngesterQueryParams {
-    pub table: Option<String>,
+    /// The database where log data will be written to.
     pub db: Option<String>,
+
```
killme2008, line 72: The blank line between fields is unnecessary.

killme2008 (commented):
Good job! Can we add a http integration test for this API? Please refer to https://github.com/GreptimeTeam/greptimedb/blob/main/tests-integration/tests/http.rs

src/servers/src/elasticsearch.rs
zyy17: TIL

src/servers/src/http.rs
zyy17: It's possible but seems necessary. All the fake api is prepared for ingestion. We can take it as the future PR when necessary.

src/servers/src/elasticsearch.rs
zyy17: `is_document` is a flag variable that indicates that the current object is a document. I add some comments for its usage.

src/servers/src/elasticsearch.rs
killme2008: @zyy17 What about this refactoring? I think it can avoid some copies.

zyy17: Already done in the new change.

killme2008 (approved):
LGTM. Good job!

Issue #5248, opened 2024-12-27
mito and metasrv have inconsistent behaviors while adding columns


evenyag:
### What type of bug is this?

Data corruption

### What subsystems are affected?

Standalone mode, Distributed Cluster

### Minimal reproduce step

1. Create a table test with column a, ts
2. add columns a, b in a gRPC request, ignore errors
3. add column c, d in a gRPC request
4. insert some rows
5. select rows from the table

### What did you expect to see?

Insert and select success

### What did you see instead?

Insert failed
```
thread 'grpc::test::test_distributed_handle_multi_ddl_request' panicked at tests-integration/src/grpc.rs:94:14:
called `Result::unwrap()` on an `Err` value: 0: Table operation error, at src/frontend/src/instance.rs:276:16
1: Cannot find column by name: Column b not found in table table_created_through_grpc, at /Users/evenyag/codes/greptime/procedure-traits/src/operator/src/req_convert/insert/stmt_to_region.rs:106:18
```

### What operating system did you use?

Unrelated

### What version of GreptimeDB did you use?

0.11.1

### Relevant log output and stack trace

```bash
thread 'grpc::test::test_distributed_handle_multi_ddl_request' panicked at tests-integration/src/grpc.rs:94:14:
called `Result::unwrap()` on an `Err` value: 0: Table operation error, at src/frontend/src/instance.rs:276:16
1: Cannot find column by name: Column b not found in table table_created_through_grpc, at /Users/evenyag/codes/greptime/procedure-traits/src/operator/src/req_convert/insert/stmt_to_region.rs:106:18
```

killme2008:
Do we have a plan to fix it? It's a fatal issue.

evenyag:
> Do we have a plan to fix it? It's a fatal issue.

I'm working at this.

PR #5273, opened 2025-01-02
feat: support add if not exists in the gRPC alter kind

evenyag:

- fixes https://github.com/GreptimeTeam/greptimedb/issues/5248
- the proto branch: https://github.com/GreptimeTeam/greptime-proto/pull/207

This PR adds the `add_if_not_exists` flag to the AddColumn alter kind and validates the alteration to the table in the prepare step of the AlterTableProcedure. Before this PR, the procedure may not be able to build table info in the update metadata step, which causes regions to have an inconsistent schema with the meta server.

It also fixes an issue that the procedure always alters tables without `add_if_not_exists` which is different from the behavior of auto alteration.

## PR Checklist

### DIFF
--- a/src/operator/src/expr_factory.rs
+++ b/src/operator/src/expr_factory.rs
@@ -505,7 +505,7 @@ pub(crate) fn to_alter_table_expr(
                         .context(ExternalSnafu)?,
                 ),
                 location: location.as_ref().map(From::from),
-                // We don't support `IF NOT EXISTS` for `ADD COLUMN` yet.
+                // TODO(yingwen): We don't support `IF NOT EXISTS` for `ADD COLUMN` yet.
                 add_if_not_exists: false,
             }],
         }),
--- a/src/operator/src/expr_factory.rs
+++ b/src/operator/src/expr_factory.rs
@@ -477,6 +477,7 @@ pub fn column_schemas_to_defs(
         .collect()
 }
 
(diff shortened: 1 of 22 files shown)


src/operator/src/expr_factory.rs
```diff
@@ -504,6 +505,8 @@ pub(crate) fn to_alter_table_expr(
                         .context(ExternalSnafu)?,
                 ),
                 location: location.as_ref().map(From::from),
+                // We don't support `IF NOT EXISTS` for `ADD COLUMN` yet.
```
killme2008, line 508: Leave a TODO and convert it into a GFI.

src/operator/src/insert.rs
```diff
@@ -741,6 +741,8 @@ impl Inserter {
         Ok(create_table_expr)
     }
 
+    /// Returns an alter table expression if it finds new columns in the request.
+    /// It always add columns if not exist.
```
killme2008, line 745:
```suggestion
    /// It always adds columns if not exist.
```

killme2008 (approved):
LGTM

v0y4g3r (approved):
LGTM. Do we need to change the documentation of `ALTER TABLE` statements?

evenyag:
> LGTM. Do we need to change the documentation of `ALTER TABLE` statements?

This PR doesn't change the behavior so we don't have to update it.

Issue #5217, opened 2024-12-23
/v1/prometheus/api/v1/label/<label_name>/values returns InvalidArguments


audricsun:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Standalone mode

### Minimal reproduce step




Use influxdb protocol to write some data into greptime
```shell
❯ curl -X POST -H "Content-Type:text/plain" "localhost:4000/v1/influxdb/api/v2/write?db=public" --data-binary "test-schema,a=1,b=2.0,c=3 value=4" -vvv
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying ::1:4000...
* Connected to localhost (::1) port 4000 (#0)
> POST /v1/influxdb/api/v2/write?db=public HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/7.76.1
> Accept: */*
> Content-Type:text/plain
> Content-Length: 33
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 204 No Content
< date: Mon, 23 Dec 2024 10:05:46 GMT
<
* Connection #0 to host localhost left intact
```

fetch label and label values with prometheus API

```shell
❯ curl -X GET localhost:4000/v1/prometheus/api/v1/labels
{"status":"success","data":["__name__","a","b","c","client","interval_type","name","number","time"]}%                                                                                                                                                                                                                                                                                                                                    
❯ curl -X GET localhost:4000/v1/prometheus/api/v1/label/a/values
{"status":"error","data":{"resultType":"","result":[]},"error":"match[] parameter is required","errorType":"InvalidArguments"}%                                                                                                                                                                                                                                                                                                          
```

### What did you expect to see?

return label values like prometheus [Prometheus' HTTP API](https://docs.greptime.com/user-guide/query-data/promql#prometheus-http-api)

```shell
# curl localhost:9090/api/v1/labels
[...]

PR #5274, opened 2025-01-03
fix: do not support v1/label/<label_name>/values

yihong0618:

This patch fix v1/label/<label_name>/values returns InvalidArguments close issue #5217 by using the same logic as prometheus api.

## PR Checklist

### DIFF
--- a/src/servers/src/http/prometheus.rs
+++ b/src/servers/src/http/prometheus.rs
@@ -719,22 +719,63 @@ pub async fn label_values_query(
     }
 
     let queries = params.matches.0;
-    if queries.is_empty() {
-        return PrometheusJsonResponse::error(
-            StatusCode::InvalidArguments,
-            "match[] parameter is required",
-        );
-    }
-
     let start = params.start.unwrap_or_else(yesterday_rfc3339);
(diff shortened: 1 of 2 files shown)


yihong0618:
will figure why the test failed

src/servers/src/http/prometheus.rs
```diff
@@ -719,22 +719,63 @@ pub async fn label_values_query(
...
 
+        for table_name in table_names {
+            let prom_query = PromQuery {
+                query: table_name,
+                start: start.clone(),
+                end: end.clone(),
+                step: DEFAULT_LOOKBACK_STRING.to_string(),
+                lookback: lookback.clone(),
+            };
+            let result = handler.do_query(&prom_query, query_ctx.clone()).await;
```
sunng87, line 749: Executing queries in a loop may introduce huge latency. We can run these tasks in parallel.

But I still think we need to add some restrictions because there can be a large number of tables to query.

yihong0618: agreed seems it need to discuss, and I will make this Pull request to draft

yihong0618:
closing for now since can be a large number of tables to query.

PR #5293, opened 2025-01-06
ci: disable docker/rust cache temporarily and merge docker compose files

sunng87:

Checking action usage data we found the docker cache steps takes about 3 mins for every run even when no docker images are changed. 

In this patch, we disabled this cache step to see if it's have positive effects on usage.

I also merged docker compose files from each fixtures folders in order to use the checksum as docker cache key in future.

## PR Checklist


sunng87:
4 mins (about 20%) save is expected with this patch.

WenyXu (approved):

zyy17 (approved):

PR #5298, opened 2025-01-06
refactor: optimize out partition split insert requests

MichaelScofield:

If there's only one region in the table, skip spliting insert request rows by partition rules.

Saves about 0.4% cpu on our testing device.

## PR Checklist

### DIFF
--- a/src/operator/src/req_convert/insert/row_to_region.rs
+++ b/src/operator/src/req_convert/insert/row_to_region.rs
@@ -55,7 +55,6 @@ impl<'a> RowToRegion<'a> {
             let table_id = self.get_table_id(&request.table_name)?;
             let region_numbers = self.region_numbers(&request.table_name)?;
             let requests = if let Some(region_id) = match region_numbers[..] {
-                [] => Some(RegionId::new(table_id, 0)),
                 [singular] => Some(RegionId::new(table_id, singular)),
                 _ => None,
             } {
--- a/src/operator/src/req_convert/insert/row_to_region.rs
+++ b/src/operator/src/req_convert/insert/row_to_region.rs
@@ -13,30 +13,31 @@
 // limitations under the License.
 
 use ahash::{HashMap, HashSet};
-use api::v1::region::InsertRequests as RegionInsertRequests;
+use api::v1::region::{InsertRequest, InsertRequests as RegionInsertRequests};
 use api::v1::RowInsertRequests;
 use partition::manager::PartitionRuleManager;
 use snafu::OptionExt;
-use table::metadata::TableId;
+use store_api::storage::{RegionId, RegionNumber};
+use table::metadata::{TableId, TableInfoRef};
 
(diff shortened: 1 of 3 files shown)


src/operator/src/req_convert/insert/row_to_region.rs
```diff
@@ -49,10 +50,36 @@ impl<'a> RowToRegion<'a> {
...
+                    if rest.is_empty() {
+                        Some(first.region.id)
+                    } else {
+                        None
+                    }
+                }
+                None => Some(RegionId::new(table_id, 0)),
+            } {
+                vec![InsertRequest {
+                    region_id: region_id.as_u64(),
```
fengjiachun, line 74: Does it work well in the metric engine? It seems to be the physical region ID.

src/operator/src/req_convert/insert/row_to_region.rs
```diff
@@ -49,10 +50,36 @@ impl<'a> RowToRegion<'a> {
...
         for request in requests.inserts {
+            let Some(rows) = request.rows else { continue };
+
             let table_id = self.get_table_id(&request.table_name)?;
-            let requests = Partitioner::new(self.partition_manager)
-                .partition_insert_requests(table_id, request.rows.unwrap_or_default())
-                .await?;
+            let table_route = self
+                .partition_manager
+                .find_physical_table_route(table_id)
```
fengjiachun, line 58: I am concerned this will cause performance loss with multi-partition, as `find_physical_table_route` will be called again internally.

fengjiachun (approved):
LGTM

src/operator/src/req_convert/insert/row_to_region.rs
MichaelScofield: Now the codes use the table info to find the singular region id, must work with logical table.

src/operator/src/req_convert/insert/row_to_region.rs
MichaelScofield: use already retrived table info now, no extra perf penalty

src/operator/src/req_convert/insert/row_to_region.rs
```diff
@@ -49,10 +50,25 @@ impl<'a> RowToRegion<'a> {
...
         for request in requests.inserts {
+            let Some(rows) = request.rows else { continue };
+
             let table_id = self.get_table_id(&request.table_name)?;
-            let requests = Partitioner::new(self.partition_manager)
-                .partition_insert_requests(table_id, request.rows.unwrap_or_default())
-                .await?;
+            let region_numbers = self.region_numbers(&request.table_name)?;
+            let requests = if let Some(region_id) = match region_numbers[..] {
+                [] => Some(RegionId::new(table_id, 0)),
```
killme2008, line 58: Should it return an error?

killme2008 (approved):
Almost LGTM

PR #5181, opened 2024-12-16
chore: add aquamarine to dep lists

evenyag:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)
https://github.com/GreptimeTeam/greptimedb/pull/5163

## What's changed and what's your intention?
Adds aquamarine back.

Otherwise, docs ci will fail
https://github.com/GreptimeTeam/greptimedb/actions/runs/12352863566/job/34470766661

## Checklist

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.

PR #5301, opened 2025-01-06
fix: drop unused dep using udeps to minial the size

yihong0618:

This patch drop unused deps using https://github.com/est31/cargo-udeps 

command `cargo  udeps --all-targets`

update: 

better use: 

`cargo  udeps --all-targets --all-features`

future maybe we can move all tests deps together when `cargo build` do not contain them

## PR Checklist


src/metric-engine/Cargo.toml
```diff
@@ -9,7 +9,6 @@ workspace = true
 
 [dependencies]
 api.workspace = true
-aquamarine.workspace = true
```
evenyag, line 12: We use this dep to generate docs.
https://github.com/GreptimeTeam/greptimedb/pull/5181

src/common/meta/Cargo.toml
```diff
@@ -35,8 +35,6 @@ common-wal.workspace = true
 datafusion-common.workspace = true
 datafusion-expr.workspace = true
 datatypes.workspace = true
-deadpool.workspace = true
-deadpool-postgres.workspace = true
```
evenyag, line 39: @CookiePieWw @WenyXu Does the postgres feature use them?

src/metric-engine/Cargo.toml
yihong0618: will bring it back maybe to dev?

evenyag:
Does cargo udeps support `--workspace --all-features`? There may be some false positives.

yihong0618:
> aquamarine.workspace = true

yes will do this

waynexia:
cross-reference an earlier process: https://github.com/GreptimeTeam/greptimedb/pull/5163

yihong0618:
> Does cargo udeps support `--workspace --all-features`? There may be some false positives.

address the comments thanks

yihong0618:
> cross-reference an earlier process: #5163

copy that thanks

src/metric-engine/Cargo.toml
evenyag: `dev-dependencies` may not work.
https://github.com/mersinvald/aquamarine/blob/master/demo/Cargo.toml

src/session/Cargo.toml
```diff
@@ -22,6 +22,5 @@ common-telemetry.workspace = true
 common-time.workspace = true
 derive_builder.workspace = true
 derive_more = { version = "1", default-features = false, features = ["debug"] }
-meter-core.workspace = true
```
evenyag, line 25: @shuiyisong Does 3rd party crates rely on this?

evenyag (approved):

src/session/Cargo.toml
shuiyisong: I don't think so. We can remove this now.

shuiyisong (approved):

Issue #5296, opened 2025-01-06
Storing necessary fields instead of the entire `QueryContext` struct


WenyXu:
It's better to store only the necessary fields instead of the entire `QueryContext` struct. I'm going to convert this PR into an issue and plan to refactor it in the future.

_Originally posted by @WenyXu in https://github.com/GreptimeTeam/greptimedb/issues/5200#issuecomment-2572399879_

evenyag:
@WenyXu Can we turn this into a good first issue if you provide more context for it?

PR #5307, opened 2025-01-06
fix: handle insert default value

discord9:

- fix handling insert default values into flow
- add related sqlness tests
- ALSO, add fill default value for impure functions(like `now()`) in frontend(previously it's on datanode), so for default value like `now()` the result will be consistent

## Future works
maybe put `now()`'s value into query context, and assign it in frontend, and fill default value using that info in datanode(so that wouldn't need to fill rows in frontend, improving perfmance) , but that would be too large a change to fit into current PR, so future works maybe can be track by #5296 ?

## PR Checklist


tests/cases/standalone/common/flow/flow_ins_default.sql
waynexia, line 1: nit: empty line

waynexia (commented):
I'm thinking if the default value should be filled in the flownode. For those inpure function (`now()` or `random()`) as default value, this can lead to wired results. A more proper way in my perspective is filling default value in the frontend before mirroring.

discord9:
> I'm thinking if the default value should be filled in the flownode. For those inpure function (`now()` or `random()`) as default value, this can lead to wired results. A more proper way in my perspective is filling default value in the frontend before mirroring.

If that's the case, then for i.e. mito engine, wouldn't it mean for dist's datanode, if each datanode have slight derivation in time(Or just different write request arrives in different times), then the default values being filled would be different for different regions in different datanode would have subtle different? Is this consider a bug or feature? see [column_default_value](https://github.com/GreptimeTeam/greptimedb/blob/05f115e047cd246b940b4f2f627260d2c95df177/src/mito2/src/request.rs#L310) function in mito worker?

src/operator/src/req_convert/insert/fill_impure_default.rs
```diff
@@ -0,0 +1,236 @@
...
+                .create_impure_default()
+                .context(ConvertColumnDefaultConstraintSnafu {
+                    column_name: column.name.clone(),
+                })?
+                .expect("impure default value");
+            let grpc_default_value = api::helper::to_proto_value(default_value.clone());
+            let def = column_schemas_to_defs(vec![column], &pk_names)?
+                .into_iter()
+                .next()
+                .expect("column def have one element");
```
v0y4g3r, line 64:
```suggestion
            let def = column_schemas_to_defs(vec![column], &pk_names)?.swap_remove(0);
```

src/operator/src/req_convert/insert/fill_impure_default.rs
```diff
@@ -0,0 +1,236 @@
...
+            .collect::<Vec<_>>();
+        let mut impure_columns = HashMap::new();
+        for column in impure_column_list {
+            let default_value = column
+                .create_impure_default()
+                .context(ConvertColumnDefaultConstraintSnafu {
+                    column_name: column.name.clone(),
+                })?
+                .expect("impure default value");
+            let grpc_default_value = api::helper::to_proto_value(default_value.clone());
```
v0y4g3r, line 60: Curious why clippy did not find that
```suggestion
            let grpc_default_value = api::helper::to_proto_value(default_value);
```

src/mito2/src/request.rs
```diff
@@ -333,6 +333,14 @@ impl WriteRequest {
             }
             OpType::Put => {
                 // For put requests, we use the default value from column schema.
+                if column.column_schema.is_default_impure() {
```
evenyag, line 336: Add a test for this?

src/operator/src/insert.rs
```diff
@@ -252,6 +266,8 @@ impl Inserter {
         .convert(requests)
         .await?;
 
+        // Fill impure default values in the request
+        let inserts = fill_reqs_with_impure_default(&table_infos, inserts)?;
         self.do_request(inserts, &ctx).await
```
evenyag, line 271: Can we move `fill_reqs_with_impure_default()` into `do_request()`? We can refactor `do_request()` to accept the `table_infos`.

src/operator/src/req_convert/insert/fill_impure_default.rs
```diff
@@ -0,0 +1,238 @@
...
+                    Some((schema.clone(), val.clone().unwrap_or_default()))
+                } else {
+                    None
+                }
+            })
+            .unzip();
+
+        rows.schema.extend(schema_append);
+        for row in rows.rows.iter_mut() {
+            row.values.extend(row_append.clone());
```
evenyag, line 110: Can we use `extend_from_slice()`?

src/operator/src/req_convert/insert/fill_impure_default.rs
```diff
@@ -0,0 +1,238 @@
...
+            .schema
+            .iter()
+            .filter_map(|schema| {
+                if self.impure_columns.contains_key(&schema.column_name) {
+                    Some(&schema.column_name)
+                } else {
+                    None
+                }
+            })
+            .collect();
```
evenyag, line 94: Do we need to add a fast path to skip filling if `self.impure_columns.len() == impure_columns_in_reqs.len()`?

I'm not sure if this can improve performance without a benchmark.

src/operator/src/req_convert/insert/fill_impure_default.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+impl ImpureDefaultFiller {
+    pub fn new(table_info: TableInfoRef) -> Result<Self> {
+        let impure_column_list = find_all_impure_columns(&table_info);
+        let pks = table_info.meta.primary_key_indices.clone();
+        let pk_names = pks
+            .iter()
+            .map(|&i| table_info.meta.schema.column_name_by_index(i).to_string())
+            .collect::<Vec<_>>();
```
evenyag, line 51: This clone is unnecessary
```rust
let pks = table_info.meta.primary_key_indices.clone();
```

v0y4g3r: > This clone is unnecessary
> 
> ```rust
> let pks = table_info.meta.primary_key_indices.clone();
> ```

Why clippy failed to figure out that?

src/operator/src/insert.rs
```diff
@@ -202,18 +204,26 @@ impl Inserter {
...
 
-        let (tables_info, instant_table_ids) = self
+        let CreateAlterTableResult {
+            instant_table_ids,
+            table_infos,
+        } = self
             .create_or_alter_tables_on_demand(&requests, &ctx, create_type, statement_executor)
             .await?;
+
+        let name2info = table_infos
```
evenyag, line 214: non-blocking: I remember that we usually use name_to_info in the codebase.

evenyag (approved):

waynexia (approved):

Issue #5271, opened 2025-01-02
Unable to view table structure in db clients (e.g. dbeaver or table plus)


weyert:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Table Engine

### Minimal reproduce step

1. Start the `nginx` demo from https://github.com/GreptimeTeam/demo-scene/tree/main/nginx-log-metrics
2. Connect via postgreSQL to the database in DBeaver or Table Plus

### What did you expect to see?

I would expect that I can browse the table structure

### What did you see instead?

Depending on the user db client I am seeing different behaviour but the end result of no structure table is the same:
1. DBeaver Community: I am getting an error when trying to connect: `ERROR: Failed to describe statement`
2. Table Plus (macOS): Successfully connects but when navigating to the `structure` view of a table it is empty

I can execute the query `DESCRIBE "public"."ngx_access_log"` in the latter db client (Table Plus) and returns the following:

| Column         | Type                 | Key | Null | Default | Semantic Type |
|----------------|----------------------|-----|------|---------|---------------|
| client         | String               |     | YES  |         | FIELD         |
| application_id | String               |     | YES  |         | FIELD         |
| upstream_id    | String               |     | YES  |         | FIELD         |
| referer        | String               |     | YES  |         | FIELD         |
| method         | String               |     | YES  |         | FIELD         |
| endpoint       | String               |     | YES  |         | FIELD         |
| trace_id       | String               |     | YES  |         | FIELD         |
| protocol       | String               |     | YES  |         | FIELD         |
| status         | UInt16               |     | YES  |         | FIELD         |
| response_time  | Int16                |     | YES  |         | FIELD         |
| response_size  | Float64              |     | YES  |         | FIELD         |
| agent          | String               |     | YES  |         | FIELD         |
| access_time    | TimestampMillisecond | PRI | NO   |         | TIMESTAMP     |
| log_time       | TimestampMillisecond |     | NO   |         | FIELD         |
| ip_address     | String               |     | YES  |         | FIELD         |

### What operating system did you use?

Docker (linux/amd64) on Mac OS X 10.5 ARM

### What version of GreptimeDB did you use?

0.11.1

### Relevant log output and stack trace

```bash
TablePlus seems to log the following queries that might be related to fetching the table structure? I can't see any relevant logs inside `greptimedb` itself.


-- 2025-01-02 12:30:55.4400
[...]

PR #5313, opened 2025-01-08
feat: support `select session_user;`

yihong0618:
This commit is part of support DBeaver that support function select session_user like postgres did.

first part of https://github.com/GreptimeTeam/greptimedb/issues/5271#issuecomment-2575070285
will support `select pg_catalog.pg_database;` in next patch

screen shot;
image

## PR Checklist

### DIFF
--- a/src/common/function/src/system.rs
+++ b/src/common/function/src/system.rs
@@ -22,7 +22,7 @@ mod version;
 use std::sync::Arc;
 
 use build::BuildFunction;
-use database::{CurrentSchemaFunction, DatabaseFunction};
+use database::{CurrentSchemaFunction, DatabaseFunction, SessionUserFunction};
 use pg_catalog::PGCatalogFunction;
 use procedure_state::ProcedureStateFunction;
 use timezone::TimezoneFunction;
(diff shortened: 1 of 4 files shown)


sunng87 (commented):
Awesome! Can we add a test case for this? It can be a sqlness test.

By the way, why it fails when we call `select session_user()`

yihong0618:
> Awesome! Can we add a test case for this? It can be a sqlness test.
> 
> By the way, why it fails when we call `select session_user()`

the same as pg
FYI you can use pglite for quick pg command test its cool
image

https://pglite.dev/repl/

sunng87:
@yihong0618 That's very helpful. Bookmarked.

killme2008 (approved):
Thank you. LGTM. Please add a sqlness test just like https://github.com/GreptimeTeam/greptimedb/blob/main/tests/cases/standalone/common/system/pg_catalog.sql

You can run the sqlness tests via `cargo sqlness -t <file_name>`

yihong0618:
> Thank you. LGTM. Please add a sqlness test just like https://github.com/GreptimeTeam/greptimedb/blob/main/tests/cases/standalone/common/system/pg_catalog.sql
> 
> You can run the sqlness tests via `cargo sqlness -t <file_name>`

Done～

and found <file_name> is not file_name is regrex....

sunng87:
@yihong0618 Thank you!

PR #5315, opened 2025-01-08
feat: more than one flow workers

discord9:

- more than one flow workers, configurable, using round robin to decide which flow goes to which worker

## PR Checklist

## Summary by CodeRabbit

- **New Features**
	- Added support for configurable number of workers in flow node initialization
	- Implemented round-robin worker selection mechanism for flow creation
	- Enhanced concurrent worker management with multi-threaded worker support

- **Improvements**
	- Simplified worker thread creation and management process
	- Improved error handling during worker initialization

### DIFF
--- a/config/flownode.example.toml
+++ b/config/flownode.example.toml
@@ -5,6 +5,8 @@ mode = "distributed"
 ## @toml2docs:none-default
 node_id = 14
 
+## flow engine options.
+[flow]
 ## The number of flow worker in flownode.
 ## Not setting this value will use the number of CPU cores divided by 2.
 #+num_workers=8
--- a/config/flownode.example.toml
+++ b/config/flownode.example.toml
@@ -7,7 +7,7 @@ node_id = 14
 
 ## The number of flow worker in flownode.
 ## Not setting this value will use the number of CPU cores divided by 2.
-num_workers = 8
+#+num_workers=8
 
 ## The gRPC server options.
 [grpc]
--- a/config/flownode.example.toml
+++ b/config/flownode.example.toml
@@ -6,7 +6,7 @@ mode = "distributed"
 node_id = 14
 
 ## The number of flow worker in flownode
-num_workers = 2
+num_workers = 8
 
 ## The gRPC server options.
 [grpc]
--- a/config/flownode.example.toml
+++ b/config/flownode.example.toml
@@ -5,6 +5,9 @@ mode = "distributed"
 ## @toml2docs:none-default
 node_id = 14
 
+## The number of flow worker in flownode
+num_workers = 2
+
 ## The gRPC server options.
 [grpc]
 ## The address to bind the gRPC server.
--- a/config/flownode.example.toml
+++ b/config/flownode.example.toml
@@ -9,7 +9,7 @@ node_id = 14
 [flow]
 ## The number of flow worker in flownode.
 ## Not setting(or set to 0) this value will use the number of CPU cores divided by 2.
-#+num_workers=8
+#+num_workers=0
 
 ## The gRPC server options.
 [grpc]
--- a/config/flownode.example.toml
+++ b/config/flownode.example.toml
@@ -8,7 +8,7 @@ node_id = 14
 ## flow engine options.
 [flow]
 ## The number of flow worker in flownode.
-## Not setting this value will use the number of CPU cores divided by 2.
+## Not setting(or set to 0) this value will use the number of CPU cores divided by 2.
 #+num_workers=8
 
 ## The gRPC server options.
--- a/config/flownode.example.toml
+++ b/config/flownode.example.toml
@@ -5,7 +5,8 @@ mode = "distributed"
 ## @toml2docs:none-default
 node_id = 14
 
-## The number of flow worker in flownode
+## The number of flow worker in flownode.
+## Not setting this value will use the number of CPU cores divided by 2.
 num_workers = 8
 
 ## The gRPC server options.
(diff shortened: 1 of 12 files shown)


discord9:
@coderabbitai review

src/flow/src/adapter.rs
```diff
@@ -109,7 +112,28 @@ impl Default for FlownodeOptions {
...
+                MAX_WORKERS
+            )))
+            .context(LoadLayeredConfigSnafu)?
+        }
+        Ok(())
+    }
+}
+
+/// Define a reasonable maximum number of workers
+const MAX_WORKERS: usize = 16;
```
killme2008, line 136: Looks like it's not necessary. For example, tokio runtime doesn't limit the worker threads number.

src/flow/src/adapter.rs
```diff
@@ -181,15 +208,20 @@ impl FlowWorkerManager {
...
         query_engine: Arc<dyn QueryEngine>,
         table_meta: TableMetadataManagerRef,
-    ) -> (Self, Worker<'s>) {
+        num_workers: usize,
+    ) -> (Self, Vec<Worker<'s>>) {
         let mut zelf = Self::new(node_id, query_engine, table_meta);
-        let (handle, worker) = create_worker();
-        zelf.add_worker_handle(handle);
-        (zelf, worker)
+        let mut workers = Vec::new();
```
killme2008, line 218: Love to use iterator to generate the workers.

src/flow/src/adapter.rs
```diff
@@ -100,6 +102,7 @@ impl Default for FlownodeOptions {
             mode: servers::Mode::Standalone,
             cluster_id: None,
             node_id: None,
+            num_workers: 2,
```
killme2008, line 105: What about setting it to the CPU numbers by default?

```
common_config::utils::get_cpus()
```

src/flow/src/adapter.rs
```diff
@@ -828,7 +860,8 @@ impl FlowWorkerManager {
             .write()
             .await
             .insert(flow_id, err_collector.clone());
-        let handle = &self.worker_handles[0].lock().await;
+        // TODO(discord9): load balance?
```
killme2008, line 863: It's a good idea to introduce a load balancer trait, but we can leave it later.

killme2008 (commented):
Don't forget  to update the example config files.

config/flownode.example.toml
```diff
@@ -5,6 +5,9 @@ mode = "distributed"
 ## @toml2docs:none-default
 node_id = 14
 
+## The number of flow worker in flownode
+num_workers = 2
```
killme2008, line 9: What about supports `auto` value, just like many options in mito engine?

discord9: > What about supporting `auto` value, just like many options in mito engine?

supporting both str("auto") and number will be unnecessarily complex, how about just say not set it will use auto default value in docs?

killme2008: You can modify the example to:

```toml
#+num_workers=2
```

This will be rendered correctly but treated as a comment.

killme2008 (approved):
LGTM

src/flow/src/adapter.rs
```diff
@@ -100,6 +102,7 @@ impl Default for FlownodeOptions {
             mode: servers::Mode::Standalone,
             cluster_id: None,
             node_id: None,
+            num_workers: common_config::utils::get_cpus() / 2,
```
killme2008, line 105: Looks like we need to ensure that the `cpus/2` is not less than one.

```rust
(common_config::utils::get_cpus() / 2).max(1)
```

config/flownode.example.toml
```diff
@@ -5,6 +5,10 @@ mode = "distributed"
 ## @toml2docs:none-default
 node_id = 14
 
+## The number of flow worker in flownode.
+## Not setting this value will use the number of CPU cores divided by 2.
+#+num_workers=8
```
waynexia, line 10: Sugg: also provides `0` as default value (which is num cores divided by 2)

waynexia (commented):
Other points:
- Do we need to support CLI options for this as well? but this is not a blocker IMO
- Do we support graceful shutdown after this? which needs to stop workers individually before exit

src/flow/src/server.rs
```diff
@@ -414,24 +414,30 @@ impl FlownodeBuilder {
 
         register_function_to_query_engine(&query_engine);
 
-        let (tx, rx) = oneshot::channel();
+        let num_workers = self.opts.opts.num_workers;
```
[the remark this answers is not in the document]
discord9: > ensure this is greater than 0

Should already be checked in `Configureable::validate`

config/flownode.example.toml
discord9: > Sugg: also provides `0` as default value (which is num cores divided by 2)

a little bit wired that config has to be `mut` all the way down and everywhere that use it need to check for non zero, maybe just validate them to be greater than zero is good enough?or make num_workers a `NonZero<usize>` instead?

config/flownode.example.toml
```diff
@@ -5,6 +5,12 @@ mode = "distributed"
 ## @toml2docs:none-default
 node_id = 14
 
+## flow engine options.
+[flow]
+## The number of flow worker in flownode.
+## Not setting(or set to 0) this value will use the number of CPU cores divided by 2.
+#+num_workers=8
```
killme2008, line 12:
```suggestion
#+num_workers=0
```

src/common/config/src/config.rs
```diff
@@ -73,14 +73,21 @@ pub trait Configurable: Serialize + DeserializeOwned + Default + Sized {
...
+        opts.validate_sanitize()?;
+
         Ok(opts)
     }
 
+    /// Validate(and possibly sanitize) the configuration.
+    fn validate_sanitize(&mut self) -> Result<()> {
+        Ok(())
+    }
+
```
zhongzc, line 90: No need to change here?

zhongzc (approved):

PR #5320, opened 2025-01-08
refactor: split up candidate registration and keep alive

CookiePieWw:

#5208

A refactor that has been discussed before. For etcd it's relatively complicated since the keep alive is operated through a keep alive stream returned by lease client. So the keep alive is still operated inside the `candidate registration`.

Not sure if it's the expected structure, so I just init a draft pr.

## PR Checklist


CookiePieWw:
Candidate register and keep alive shoud be progressive, so it's simpler to put them into a function instead of split them up.

PR #5325, opened 2025-01-09
fix: drop all python embedding code for docker and doc

yihong0618:

Since all python feature had been dropped from #5321 
Docker side and requirements can be drop too, also drop  `rustpython` mention in doc. 

## PR Checklist

### DIFF
--- a/docker/dev-builder/android/Dockerfile
+++ b/docker/dev-builder/android/Dockerfile
@@ -13,8 +13,7 @@ RUN apt-get update && apt-get install -y \
     curl \
     git \
     build-essential \
-    pkg-config \
-    python3
+    pkg-config
 
 # Trust workdir
 RUN git config --global --add safe.directory /greptimedb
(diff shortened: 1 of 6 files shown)


docker/dev-builder/android/Dockerfile
```diff
@@ -14,11 +14,7 @@ RUN apt-get update && apt-get install -y \
     git \
     build-essential \
     pkg-config \
-    python3 \
-    python3-dev \
-    python3-pip \
-    && pip3 install --upgrade pip \
-    && pip3 install pyarrow
+    python3
```
yihong0618, line 17: not sure this python3 better drop or not, need disscuss

sunng87: I believe we can remove this

sunng87 (approved):

discord9 (approved):

PR #5328, opened 2025-01-09
feat: add show search_path for pg

yihong0618:

part of https://github.com/GreptimeTeam/greptimedb/issues/5271
for now `pg_database` is better to hold waiting for the upstream https://github.com/apache/datafusion/pull/13489

this patch support pg `show search_path` for now seems no search path for greptime so use database instead

## PR Checklist

### DIFF
--- a/src/frontend/src/instance.rs
+++ b/src/frontend/src/instance.rs
@@ -565,6 +565,7 @@ pub fn check_permission(
             validate_db_permission!(stmt, query_ctx);
         }
         Statement::ShowStatus(_stmt) => {}
+        Statement::ShowSearchPath(_stmt) => {}
         Statement::DescribeTable(stmt) => {
             validate_param(stmt.name(), query_ctx)?;
         }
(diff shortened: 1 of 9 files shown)


sunng87 (approved):

killme2008 (commented):
The implementation seems too complex for our needs. We should consider a simpler approach like the one used in GrepTimeDB's MySQL implementation, which uses regex matching for queries and returns placeholder responses. This would be more straightforward and easier to maintain. https://github.com/GreptimeTeam/greptimedb/blob/main/src/servers/src/mysql/federated.rs

yihong0618:
> The implementation seems too complex for our needs. We should consider a simpler approach like the one used in MySQL implementation, which uses regex matching for queries and returns placeholder responses. This would be more straightforward and easier to maintain. https://github.com/GreptimeTeam/greptimedb/blob/main/src/servers/src/mysql/federated.rs

will take a look but I followed the command like `show status` should that need to change too? if so I'd like maybe change that to this way too.

killme2008:
@yihong0618 

Yes, `show status` doesn't need to be changed.

If we need to keep the `set search_path to xxxx` in session, the current implementation is good.

yihong0618:
> @yihong0618 
> 
> 
> 
> Yes, `show status` doesn't need to be changed.
> 
> 
> 
> If we need to keep the `set search_path to xxxx` in session, the current implementation is good.
> 
> 
> 
> 

thanks will try to implement set search_path next

killme2008 (approved):
LGTM

PR #5335, opened 2025-01-10
refactor(elasticsearch): use `_index` as greptimedb table in log ingestion and add `/${index}/_bulk` API

zyy17:

1. Take `_index` from bulk requests as greptimedb table;
2. Add `${index}/_bulk` API;
3. Add `LogIngestRequest` and refactor `ingest_logs_inner` to support ingest logs into multiple tables by one request;

## PR Checklist

### DIFF
--- a/src/servers/src/elasticsearch.rs
+++ b/src/servers/src/elasticsearch.rs
@@ -92,7 +92,7 @@ pub async fn handle_bulk_api(
     TypedHeader(_content_type): TypedHeader<ContentType>,
     payload: String,
 ) -> impl IntoResponse {
-    do_handle_bulk_api(None, log_state, params, query_ctx, payload).await
+    do_handle_bulk_api(log_state, None, params, query_ctx, payload).await
 }
 
 /// Process `/${index}/_bulk` API requests. Only support to create logs.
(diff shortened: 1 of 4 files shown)


src/servers/src/elasticsearch.rs
```diff
@@ -261,49 +272,92 @@ fn convert_es_input_to_log_values(
...
+    //   1. The first object is the command, it should be `create` or `index`.
+    //   2. The second object is the document data.
+    loop {
+        // Read the first object to get the command, it should be `create` or `index`.
+        // Stop the loop if there is no command.
+        let next_cmd = values.next();
+        if next_cmd.is_none() {
+            break;
         }
+        let mut cmd = next_cmd.unwrap();
```
sunng87, line 289: We can use 

```rust
while let Some(cmd) = values.next() { ... }
```

to avoid use `unwrap` even it's safe

src/servers/src/elasticsearch.rs
```diff
@@ -261,49 +272,92 @@ fn convert_es_input_to_log_values(
...
-            log_values.push(log_value);
-
-            // Reset the flag for the next chunk.
-            is_document = false;
+        // Read the second object to get the document data. Stop the loop if there is no document.
+        let document = values.next();
+        if document.is_none() {
+            break;
         }
+        let document = document.unwrap();
```
sunng87, line 311: similar here

```rust

if let Some(document) = values.next() {
    // ...
} else {
    break;
}

```

src/servers/src/elasticsearch.rs
zyy17: TIL

sunng87 (approved):

shuiyisong (approved):
LGTM

By the way, we may also want to accept table names specified in the HTTP header. Since we have all kinds of protocols using each of their own table naming methods, we can build a common extraction function to provide Greptime header parameters at the bottom and layered protocol-specified parameters at the top, much like our layered configuration structure.

Issue #5346, opened 2025-01-12
Support to flatten JSON logs in internal pipeline


zyy17:
### What type of enhancement is this?

API improvement

### What does the enhancement do?

The current internal greptimedb pipeline will write JSON objects as the following schema:

```
<key>: <json blob>
```

It's easy to implement with good performance. However, it's very hard to use unless you write another pipeline to parse it.

Another user-friendly process is to flatten JSON objects. For example, if we want to write the following JSON:

```json
{
  "kubernetes": {
    "container_id": "containerd://21086f05ddd5a29d194babf3879dafb85de278e31b3ac564ea08c174e8f039bc",
    "container_image": "localhost:5001/printer:latest",
    "container_image_id": "localhost:5001/printer@sha256:21b8881e5e2bcea49cca791ef20540f89881957895fd049293ee82ad8bf13e99",
    "container_name": "printer",
    "namespace_labels": {
      "control-plane": "controller-manager",
      "kubernetes.io/metadata.name": "default"
    }
  }
}
```

It can be flatten as:

```json
{
  "kubernetes.container_id": "containerd://21086f05ddd5a29d194babf3879dafb85de278e31b3ac564ea08c174e8f039bc",
  "kubernetes.container_image": "localhost:5001/printer:latest",
  "kubernetes.container_image_id": "localhost:5001/printer@sha256:21b8881e5e2bcea49cca791ef20540f89881957895fd049293ee82ad8bf13e99",
  "kubernetes.container_name": "printer",
  "kubernetes.namespace_labels.control-plane": "controller-manager",
  "kubernetes.namespace_labels.kubernetes.io/metadata.name": "default"
}
```

It's similar to the vector VRL function [`flatten()`](https://vector.dev/docs/reference/vrl/functions/#flatten).

For an array-type object, it's hard to flatten it as an object with a unique key. We can marshal it as JSON string, for example:

```
{
  "a": {
    "b": {
      "c": 1,
      "d": 2,
      "e": {
        "f": "g",
        "h": true
      },
      "u": [
        "a",
        "b",
        "c"
      ]
    }
  }
}
=>
{
  "a.b.c": 1,
  "a.b.d": 2,
  "a.b.e.f": "g",
  "a.b.e.h": true,
  "a.b.u": "[\"a\",\"b\",\"c\"]"
}
```

### Implementation challenges

1. Limit the depth when flattening the big JSON object;
2. Performance;

PR #5358, opened 2025-01-14
refactor: support to flatten json object in `greptime_identity` pipeline

zyy17:

Resolve https://github.com/GreptimeTeam/greptimedb/issues/5346.

Support to flatten JSON object in `greptime_identity` pipeline.

## PR Checklist


zyy17:
> This is going to be a breaking change if we modify this identity pipeline directly. An alternative solution is to create a new built-in pipeline, say `greptime_json_flattern`. WDYT?

Make sense. But I am still considering whether we should flatten JSON object by default, as it would make it easier to use.

shuiyisong:
How about a processor to flatten a field from input JSON, or all fields if the field parameter is absent?

Now that we support JSON datatype in the database, it's possible that the user wants to store the nested raw JSON data in one column, for example, some extra optional commenting data with JSON format(like structured metadata in Loki protocol, which doesn't need to be indexed, and only used for display upon query).

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -431,11 +437,72 @@ pub fn identity_pipeline(
...
+    current_level: usize,
+    max_nested_levels: usize,
+) -> Result<()> {
+    // For safety, we do not allow the depth to be greater than the max_object_depth.
+    if current_level > max_nested_levels {
+        return ReachedMaxNestedLevelsSnafu { max_nested_levels }.fail();
+    }
+
+    for (key, value) in object {
+        let new_key = base.map_or_else(|| key.clone(), |base_key| format!("{base_key}.{key}"));
```
killme2008, line 471: Does the `key.clone()` look unnecessary? We can consume it.

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -370,7 +373,10 @@ fn identity_pipeline_inner<'a>(
     let mut schema_info = SchemaInfo::default();
     for value in array {
         if let serde_json::Value::Object(map) = value {
-            let row = json_value_to_row(&mut schema_info, map)?;
+            let row = json_value_to_row(
```
killme2008, line 376: I think the `flatten` should be an option for the identity pipeline.

zyy17:
@sunng87 @shuiyisong How about adding the new query parameters, for example, `flatten_json=true`? Adding the new internal pipeline seems unnecessary.

src/pipeline/src/etl/transform/transformer/greptime.rs
zyy17: `key.clone()` occurs when the `base` is `None`, which is during the initial call, and in this case, the `clone()` is necessary.

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -47,6 +51,25 @@ pub struct GreptimeTransformer {
...
+pub struct GreptimeIdentityPipelineParams {
+    /// Whether to flatten the JSON object.
+    pub flatten_json_object: bool,
+}
+
+impl GreptimeIdentityPipelineParams {
+    /// Create a `GreptimeIdentityPipelineParams` from the HTTP headers.
+    pub fn from_headers(headers: &HeaderMap) -> Self {
+        Self {
+            flatten_json_object: headers
```
killme2008, line 65: Don't use one header for one parameter, just use `x-greptime-pipeline-params` as the header key and make the values in the form of  query string such as ''a=b&b=c`  like that.

sunng87 (approved):

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -420,22 +461,84 @@ fn identity_pipeline_inner<'a>(
...
+                    dest,
+                    Some(&new_key),
+                    object,
+                    current_level + 1,
+                    max_nested_levels,
+                )?;
+            }
+            // To simplify the process of logs collection scenario, we will convert the array into a string that can be easily handled by full-text search.
+            JsonValue::Array(array) => {
+                let array_str = serde_json::to_string(&array).context(SerializeToJsonSnafu {
```
shuiyisong, line 521: Since we have the JSON datatype, can we store the array as JSON array directly instead of `serde_json::to_string`?

src/servers/src/http/event.rs
```diff
@@ -569,9 +572,13 @@ pub(crate) async fn ingest_logs_inner(
                 .get_table(&request.table, &query_ctx)
                 .await
                 .context(CatalogSnafu)?;
-            pipeline::identity_pipeline(request.values, table)
-                .context(PipelineTransformSnafu)
-                .context(PipelineSnafu)?
+            pipeline::identity_pipeline(
+                request.values,
+                table,
+                GreptimeIdentityPipelineParams::from_headers(&headers),
```
shuiyisong, line 578: How about parsing the headers to params here in `event.rs` so that the `pipeline` crate doesn't have to depend on the `axum` crate.

shuiyisong (approved):
Rest LGTM

src/servers/src/http/event.rs
sunng87: I can refactor this in #5409

zyy17: make sense.

src/pipeline/src/etl/transform/transformer/greptime.rs
paomian: +1

paomian (approved):
rest LGTM

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -38,6 +41,8 @@ use crate::etl::transform::{Transform, Transformer, Transforms};
 use crate::etl::value::{Timestamp, Value};
 
 const DEFAULT_GREPTIME_TIMESTAMP_COLUMN: &str = "greptime_timestamp";
+const DEFAULT_MAX_NESTED_LEVELS_FOR_JSON_FLATTENING: usize = 10;
+const GREPTIME_IDENTITY_PIPELINE_PARAMS_HEADER: &str = "x-greptime-identity-pipeline-params";
```
killme2008, line 45: Why limit it to just the identity pipeline? Other pipelines can use these parameters too. The query params can simply be a string map.
We won't be able to add more headers in the future. It makes more sense to have all pipelines share a single header for custom parameters.

evenyag:
Waits for https://github.com/GreptimeTeam/greptimedb/pull/5417

There are conflicts in the merge queue

Issue #5326, opened 2025-01-09
Pipeline dispatcher


sunng87:
### What problem does the new feature solve?

Currently we can store all the logs/events from one API call to a single table. However, it's possible that several different types of logs share the same source. 

### What does the feature do?

So in our pipeline, we need a mechanism to dispatch data into different pipelines, and eventually different tables, based on certain fields like `type`, or any other data that user can extract from the pipeline.

```yaml
processors:

dispatcher:
  field: type
  routes:
    - value: http
      pipeline: http_pipeline
      table_name: ${prefix}_http
    - value: database
      pipeline: database_pipeline
      table_name: ${prefix}_database
    - default: yes
      pipeline: default_pipeline
      table_name: ${prefix}_events
```

```mermaid
flowchart TD
    A[entry_pipeline] --> D(dispatcher)
    D --> |type=http| B[http_pipeline]
    D --> |type=database| C[database_pipeline]
    D --> |type=unknown| E[default_pipeline]
```

killme2008:
Should we support regex matching for values? Users may not be able to specify all possible values and instead provide a regex expression.

sunng87:
I think it's possible with processors. They can process all possible values and generate fixed `type`

sunng87:
closed in #5409

PR #5359, opened 2025-01-14
feat: pipeline dispatch part 1, add definition

sunng87:

This is patch adds dispatcher definition to pipeline yaml, as the part 1 of pipeline dispatcher #5326 

The implementation is slight different from what I described in the issue. The default rule is removed and current pipeline will be treated as the default one.

## PR Checklist

### DIFF
--- a/src/pipeline/src/dispatcher.rs
+++ b/src/pipeline/src/dispatcher.rs
@@ -42,10 +42,10 @@ use crate::{
 ///
 /// If none of the rules match the value, this pipeline will continue to process
 /// current log entry
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub(crate) struct Dispatcher {
-    field: String,
-    rules: Vec<Rule>,
+    pub field: String,
+    pub rules: Vec<Rule>,
 }
 
 /// The rule definition for dispatcher
@@ -55,11 +55,11 @@ pub(crate) struct Dispatcher {
 ///   `greptime_identity`
 /// - `table_part`: the table name segment that we use to construct full table
 ///   name
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub(crate) struct Rule {
-    value: Value,
-    table_part: String,
-    pipeline: Option<String>,
+    pub value: Value,
+    pub table_part: String,
+    pub pipeline: Option<String>,
 }
 
 impl TryFrom<&Yaml> for Dispatcher {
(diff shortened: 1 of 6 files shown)


src/pipeline/src/dispatcher.rs
```diff
@@ -0,0 +1,109 @@
...
+///   field: type
+///   rules:
+///     - value: http
+///       pipeline: http_pipeline
+///       table_part: http_log
+///     - value: db
+///       pipeline: db_pipeline
+///       table_part: db_log
+///
+///
```
waynexia, line 47:
```suggestion
```

src/pipeline/src/dispatcher.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub field: String,
+    pub rules: Vec<Rule>,
+}
+
+/// The rule definition for dispatcher
+///
+/// - `value`: for pattern matching
+/// - `pipeline`: the pipeline to call, if it's unspecified, we use default
+///   `greptime_identity`
+/// - `table_part`: the table name segment that we use to construct full table
```
waynexia, line 63: what's "part" (partial) means

src/pipeline/src/etl/error.rs
```diff
@@ -582,6 +582,12 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
+    #[snafu(display("Field is required for dispatcher"))]
+    FieldRequiredForDispatcher,
+    #[snafu(display("table_part is required for dispatcher rule"))]
+    TablePartRequiredForDispatcherRule,
+    #[snafu(display("value is required for dispatcher rule"))]
+    ValueRequiredForDispatcherRule,
```
waynexia, line 590: sugg: add `Location` to them

waynexia (approved):

src/pipeline/src/etl/error.rs
sunng87: I removed it because this validation error seems obvious and only used once

src/pipeline/src/dispatcher.rs
sunng87: Let me add more description for this in next patch.
This partial is to construct a full table name when the write request is dispatched to another pipeline. In that situation, we cannot use original table name from input, but to construct a new table using this pattern and input table name. 

Perhaps table_suffix is a name easier to understand. Will update this in next patch.

shuiyisong (approved):
LGTM

PR #5366, opened 2025-01-15
fix: panic when received invalid query string

waynexia:

Encounter this panic with MySQL client:
```
2025-01-15T09:24:45.917687Z ERROR on_query{protocol="mysql"}:mysql::do_query: common_telemetry::panic_hook: panicked at src/servers/src/mysql/federated.rs:276:32:
byte index 6 is not a char boundary; it is inside '；' (bytes 4..7) of `exit；`
```
Here the query contains a full-width character and causes panic. This is because `str` type in rust doesn't perform utf-8 aware indexing but based on plain bytes.

## PR Checklist


fengjiachun (approved):
LGTM

evenyag (approved):

Issue #5282, opened 2025-01-03
Tracking Issue: Sparse primary key codec


WenyXu:
### What type of enhancement is this?

Performance

### What does the enhancement do?

In the metrics engine scenario, when the number of primary key columns on a physical table becomes excessively large, the CPU overhead for encoding all primary keys increases dramatically. This results in significant performance bottlenecks for both write and query operations.

**Proposal**
Introduce a sparse primary key codec that encodes only non-null primary key columns. By skipping the encoding of null or empty columns, this change aims to:
- Reduce CPU overhead associated with primary key encoding.
- Improve overall write and query performance.

**TODO**
- [x] sparse primary key codec implementation #5365
  - [x] read, backward compatibility  #5408 
  - [x] write #5424
- [x]  pluggable codec implementation #5312
- [ ]  codec benchmark
- [x] #5373
- [x] #5374
- [x] #5380

evenyag:
We also need to stabilize the format of the sparse primary key. The current format is still temporary and performs poorly on strings.

PR #5374, opened 2025-01-15
feat(metric-engine): introduce index options from metric engine

WenyXu:

#5282 

This PR introduces index options to the metric engine, addressing scenarios with high cardinality where inverted index files can grow excessively large. By allowing users to specify index types and parameters, such as skipping indexes with customizable granularity, users can better control index behavior and optimize storage.

For example:
In this case, all auto-added columns will apply the index options specified by the user.
```sql
CREATE TABLE phy (
    ts TIMESTAMP TIME INDEX, 
    val DOUBLE
) ENGINE=metric WITH (
    "physical_metric_table" = "", 
    "index.type" = "skipping", 
    "index.granularity" = "8192"
);
```

If a logical table is created on the physical table:

```sql
CREATE TABLE t1 (
    ts TIMESTAMP TIME INDEX, 
    val DOUBLE, 
    host STRING PRIMARY KEY
) ENGINE=metric WITH (
    "on_physical_table" = "phy"
);
```
The resulting physical table will have the following schema:

```sql
+-------+---------------------------------------------------------------------------------+
| Table | Create Table                                                                    |
+-------+---------------------------------------------------------------------------------+
| phy   | CREATE TABLE IF NOT EXISTS "phy" (                                              |
|       |   "ts" TIMESTAMP(3) NOT NULL,                                                   |
|       |   "val" DOUBLE NULL,                                                            |
|       |   "host" STRING NULL SKIPPING INDEX WITH(granularity = '8192', type = 'BLOOM'), |
|       |   TIME INDEX ("ts"),                                                            |
|       |   PRIMARY KEY ("host"),                                                         |
|       | )                                                                               |
|       |                                                                                 |
|       | ENGINE=metric                                                                   |
|       | WITH(                                                                           |
|       |   'index.granularity' = '8192',                                                 |
|       |   'index.type' = 'skipping',                                                    |
|       |   physical_metric_table = ''                                                    |
|       | )                                                                               |
+-------+---------------------------------------------------------------------------------+
```

## PR Checklist

### DIFF
--- a/src/metric-engine/src/engine/options.rs
+++ b/src/metric-engine/src/engine/options.rs
@@ -16,7 +16,6 @@
 
 use std::collections::HashMap;
 
-use serde::{Deserialize, Serialize};
 use store_api::metric_engine_consts::{
     METRIC_ENGINE_INDEX_SKIPPING_INDEX_GRANULARITY_OPTION,
     METRIC_ENGINE_INDEX_SKIPPING_INDEX_GRANULARITY_OPTION_DEFAULT, METRIC_ENGINE_INDEX_TYPE_OPTION,
@@ -30,20 +29,14 @@ use crate::error::{Error, ParseRegionOptionsSnafu, Result};
 /// value and appropriately increasing the size of the index, it results in an improved indexing effect.
 const SEG_ROW_COUNT_FOR_DATA_REGION: u32 = 256;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
-pub enum IndexType {
-    Inverted,
-    Skipping,
-}
-
 /// Physical region options.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct PhysicalRegionOptions {
     pub index: IndexOptions,
 }
 
 /// Index options for auto created columns
-#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
 pub enum IndexOptions {
     #[default]
     Inverted,


tests/cases/standalone/common/create/create_metric_table.result
```diff
@@ -71,6 +71,27 @@ DESC TABLE phy;
 | job        | String               | PRI | YES  |         | TAG           |
 +------------+----------------------+-----+------+---------+---------------+
 
+SHOW CREATE TABLE phy;
```
zhongzc, line 74: Why can’t see the `__table_id` and `__tsid` columns?

src/metric-engine/src/engine/create.rs
```diff
@@ -520,7 +532,8 @@ impl MetricEngineInner {
                 DATA_SCHEMA_TABLE_ID_COLUMN_NAME,
                 ConcreteDataType::uint32_datatype(),
                 false,
-            ),
+            )
+            .with_inverted_index(true),
```
zhongzc, line 536: This is an incompatible change. Once the schema of a certain column is explicitly set to have an inverted key, the condition for indexing the tag by default will be destroyed.

Is it expected?

zhongzc: oh is no problem for newly created tables.

src/metric-engine/src/engine/create.rs
```diff
@@ -529,7 +542,8 @@ impl MetricEngineInner {
                 DATA_SCHEMA_TSID_COLUMN_NAME,
                 ConcreteDataType::uint64_datatype(),
                 false,
-            ),
+            )
+            .with_inverted_index(false),
```
zhongzc, line 546: Let's clean up https://github.com/GreptimeTeam/greptimedb/blob/7eaabb3ca232be4135c3d31df567751e743c8d3f/src/metric-engine/src/engine/options.rs#L36-L39

src/metric-engine/src/data_region.rs
```diff
@@ -142,6 +148,19 @@ impl DataRegion {
 
                 c.column_id = new_column_id_start + delta as u32;
                 c.column_schema.set_nullable();
+                match index_options {
+                    IndexOptions::Inverted => {
+                        c.column_schema.set_inverted_index(true);
```
zhongzc, line 153: This is an incompatible change. Once the schema of a certain column is explicitly set to have an inverted key, the condition for indexing the tag by default will be destroyed.

tests/cases/standalone/common/create/create_metric_table.result
WenyXu: The `__table_id` and `__tsid` are internal columns, and these columns and column options will be generated via the metric engine.

src/metric-engine/src/data_region.rs
WenyXu: Yes... it's ok for the newly created table. But it's problematic to alter these options.

src/metric-engine/src/engine/create.rs
WenyXu: Resolved in [fbed01e](https://github.com/GreptimeTeam/greptimedb/pull/5374/commits/fbed01e22f13943a6684b0f7b6f45428e4f77312)

src/metric-engine/src/data_region.rs
WenyXu: Resolved in [2b8b907](https://github.com/GreptimeTeam/greptimedb/pull/5374/commits/2b8b9072b00950ff4616f9206d3dfa2df9228556)

zhongzc (approved):

src/metric-engine/src/engine/options.rs
```diff
@@ -16,31 +16,144 @@
...
 /// Compared to the mito engine, the pattern of the metric engine constructs smaller indices.
 /// Therefore, compared to the default seg row count of 1024, by adjusting it to a smaller
 /// value and appropriately increasing the size of the index, it results in an improved indexing effect.
 const SEG_ROW_COUNT_FOR_DATA_REGION: u32 = 256;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum IndexType {
+    Inverted,
+    Skipping,
+}
```
fengjiachun, line 37: Why do we need a new IndexType? Can't we just use Mito's?

fengjiachun (approved):

PR #5382, opened 2025-01-16
feat(json): Add `json_object_keys` function.

linyihai:

https://github.com/GreptimeTeam/greptimedb/issues/4805

Add `JSON_OBJECT_KEYS` functions to query all json keys.

## PR Checklist

### DIFF
--- a/src/common/function/src/scalars/json/json_object_keys.rs
+++ b/src/common/function/src/scalars/json/json_object_keys.rs
@@ -1,0 +1,202 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fmt::{self, Display};
+
+use common_query::error::{InvalidFuncArgsSnafu, Result, UnsupportedInputDataTypeSnafu};
+use common_query::prelude::Signature;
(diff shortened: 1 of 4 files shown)


src/common/function/src/scalars/json/json_object_keys.rs
```diff
@@ -0,0 +1,173 @@
...
+const NAME: &str = "json_object_keys";
+
+impl Function for JsonObjectKeysFunction {
+    fn name(&self) -> &str {
+        NAME
+    }
+
+    fn return_type(&self, _input_types: &[ConcreteDataType]) -> Result<ConcreteDataType> {
+        Ok(ConcreteDataType::string_datatype())
+    }
```
WenyXu, line 41: I'm unsure whether it's a good idea to return string type here.

CookiePieWw: String type seems to be the only choice for now, or we need to activate the list type first.

CookiePieWw:
Thanks for your contribution! After some discussion, we think it's better to use a list type instead of a string type as the return type. Since we don't have a list type yet, we need to hold this for a while 😢

linyihai:
OK, waiting on  `list type` landing .

linyihai:
> OK, waiting on `list type` landing .

I found `ListVector` could be an alternative for returing List type.  Let me update the PR and wait the next reveiw.

CookiePieWw:
Not works on my machine😢 . The list type only works on grpc protocol now.
```SQL
mysql> SELECT json_object_keys(parse_json('{"a": 1, "b": [1, 2, 3]}')) AS keys;
ERROR 1235 (42000): (Unsupported): Unsupported data type: List<String>, reason: not implemented
```

linyihai:
https://github.com/GreptimeTeam/greptimedb/blob/0a169980b71de9178d26bb3c07c70cfe0967b5a2/src/servers/src/mysql/writer.rs#L238-L245

Indeed.`List` hadn't supported by MySQL implementation.

WenyXu:
I'm going to close this PR for now, as we currently don't have a proper way to handle the list type in UDF results.
Please feel free to reopen it once support for list types is available, or if there's a better solution for handling them.

PR #5391, opened 2025-01-17
fix(flow): deal with flow drop leftover

discord9:

previously, if a flow& it's sink table is dropped, but in flownode's output buf there is still leftover for this flow, calling flush_flow here will cause a bug which when send insert request back, can't found the sink table(as expected because sink table is dropped)
this is fixed now by check if table no longer exist and not make send back request failed
related failed test: https://github.com/GreptimeTeam/greptimedb/actions/runs/12825330060/job/35763543123

## PR Checklist

### DIFF
--- a/src/flow/src/adapter.rs
+++ b/src/flow/src/adapter.rs
@@ -300,7 +300,7 @@ impl FlowWorkerManager {
                     {
                         // deal with both flow&sink table no longer exists
                         // but some output is still in output buf
-                        common_telemetry::warn!(e; "Table `{}` no longer exists, skip writeback", table_name);
+                        common_telemetry::warn!(e; "Table `{}` no longer exists, skip writeback", table_name.join("."));
                         continue;
                     } else {
                         return Err(e);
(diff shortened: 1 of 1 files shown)


src/flow/src/adapter.rs
```diff
@@ -284,12 +284,29 @@ impl FlowWorkerManager {
...
+                Err(e) => {
+                    if self
+                        .table_info_source
+                        .get_opt_table_id_from_name(&table_name)
+                        .await?
+                        .is_none()
+                    {
+                        // deal with both flow&sink table no longer exists
+                        // but some output is still in output buf
+                        common_telemetry::warn!(e; "Table no longer exists, skip writeback");
```
waynexia, line 303:
```suggestion
                        common_telemetry::warn!(e; "Table `{}` no longer exists, skip writeback", table_name);
```

waynexia (approved):

killme2008 (approved):

PR #5396, opened 2025-01-18
feat: overwrites inferred compaction window by region options

evenyag:

This PR always overwrites the compaction time window by the window size in the TWCS options if it is available.

After this PR, a SWCS compaction will not update the time window if we set the TWCS window size option. We always respect the window size specified in the region option.

## PR Checklist

### DIFF
--- a/src/mito2/src/region/version.rs
+++ b/src/mito2/src/region/version.rs
@@ -404,7 +404,7 @@ impl VersionBuilder {
             && compaction_time_window != self.compaction_time_window
         {
             info!(
-                "VersionBuilder overwrite region compaction time window from {:?} to {:?}, region: {}",
+                "VersionBuilder overwrites region compaction time window from {:?} to {:?}, region: {}",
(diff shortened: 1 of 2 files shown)


src/mito2/src/region/version.rs
```diff
@@ -389,15 +393,32 @@ impl VersionBuilder {
...
+        let compaction_time_window = self
+            .options
+            .compaction
+            .time_window()
+            .or(self.compaction_time_window);
+        if self.compaction_time_window.is_some()
+            && compaction_time_window != self.compaction_time_window
+        {
+            info!(
+                "VersionBuilder overwrite region compaction time window from {:?} to {:?}, region: {}",
```
waynexia, line 407:
```suggestion
                "VersionBuilder overwrites region compaction time window from {:?} to {:?}, region: {}",
```

waynexia (approved):

WenyXu (approved):

PR #5398, opened 2025-01-19
fix: vector function for PromQL need to ignore the time index also

### LINKED ISSUE #5392: PromQL queries with binary operators and `vector(0)` results error
### What type of bug is this?

Incorrect result

Query Engine

### Minimal reproduce step

execute a PromQL like `vector(1) * metric`. 

```
TQL EVAL (1737086400, 1737107500, '10s') vector(1) * host_cpu_seconds_total
```

I guess it's because of the time index for the metric is `greptime_timestamp` while the default time index for `vector()` is `time`

.

### What did you see instead?

```
Internal error during building DataFusion plan: No field named greptime_timestamp. Valid fields are .time, .greptime_value, .collector.
```

macos

0.11.2

```json
{
    "code": 3000,
    "error": "Internal error during building DataFusion plan: No field named greptime_timestamp. Valid fields are .time, .greptime_value, .collector.",
    "execution_time_ms": 0
}
```

yihong0618:
close #5392

The error is from index for vector function,
when `join_on_non_field_columns` and `only_join_time_index` we also need to care about the index,
so the easy way is 
```rs
        // push time index column if it exists
        if let Some(time_index_column) = &self.ctx.time_index_column {
            // issue #5392 if is special vector function
            if !is_vector_matching {
                tag_columns.push(Column::from_name(time_index_column));
            }
        }
```

also add the tests in tests/cases/standalone/common/promql/set_operation.sql

## PR Checklist


yihong0618:
seems sqlness test down is not according this, will figure out why

waynexia (approved):
👍 Thanks @yihong0618

killme2008 (approved):
LGTM

PR #5399, opened 2025-01-19
fix: avoid suppress manual compaction

v0y4g3r:

This PR fixes the issue that manually triggered compaction will be ignored when compaction is in progress.

## PR Checklist

### DIFF
--- a/src/mito2/src/compaction.rs
+++ b/src/mito2/src/compaction.rs
@@ -27,6 +27,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use api::v1::region::compact_request;
+use api::v1::region::compact_request::Options;
 use common_base::Plugins;
 use common_meta::key::SchemaMetadataManagerRef;
 use common_telemetry::{debug, error, info, warn};
@@ -143,8 +144,20 @@ impl CompactionScheduler {
         max_parallelism: usize,
     ) -> Result<()> {
         if let Some(status) = self.region_status.get_mut(&region_id) {
-            // Region is compacting. Add the waiter to pending list.
-            status.merge_waiter(waiter);
+            match compact_options {
+                Options::Regular(_) => {
+                    // Region is compacting. Add the waiter to pending list.
+                    status.merge_waiter(waiter);
+                }
+                options @ Options::StrictWindow(_) => {
+                    // Incoming compaction request is manually triggered.
+                    status.set_pending_request(PendingCompaction { options, waiter });
+                    info!(
+                        "Region {} is compacting, manually compaction will be re-scheduled.",
+                        region_id
+                    );
+                }
+            }
             return Ok(());
         }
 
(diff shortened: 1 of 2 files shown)


src/mito2/src/compaction.rs
```diff
@@ -150,8 +144,24 @@ impl CompactionScheduler {
...
+                    // Region is compacting. Add the waiter to pending list.
+                    status.merge_waiter(waiter);
+                }
+                options @ Options::StrictWindow(_) => {
+                    // Incoming compaction request is manually triggered.
+                    status.set_pending_request(PendingCompaction {
+                        options,
+                        waiter,
+                        max_parallelism,
+                    });
```
evenyag, line 158: I wonder if there will be multiple pending requests.

v0y4g3r: > I wonder if there will be multiple pending requests.

The last manual compaction (SWCS) will override previous manual compactions.

evenyag: We may notify the waiter that we will override its operation. Now we just drop the waiter.

v0y4g3r: > We may notify the waiter that we will override its operation. Now we just drop the waiter.

Sounds good.

v0y4g3r: > We may notify the waiter that we will override its operation. Now we just drop the waiter.

f807687041 PTAL @evenyag

src/mito2/src/compaction.rs
```diff
@@ -495,23 +513,36 @@ impl CompactionStatus {
...
     }
 
-    fn on_failure(self, err: Arc<Error>) {
-        if let Some(mut pending) = self.pending_compaction {
-            pending.on_failure(self.region_id, err.clone());
+    fn on_failure(mut self, err: Arc<Error>) {
+        for waiter in self.waiters.drain(..) {
+            waiter.send(Err(err.clone()).context(CompactRegionSnafu {
+                region_id: self.region_id,
+            }));
```
evenyag, line 545: Do we need to also cancel the pending compaction?

evenyag (approved):

src/mito2/src/compaction.rs
v0y4g3r: > Do we need to also cancel the pending compaction?

fixed in 6b71d0b911

fengjiachun (approved):
LGTM

src/mito2/src/compaction.rs
```diff
@@ -495,23 +513,44 @@ impl CompactionStatus {
...
+        }
+
+        if let Some(mut pending_compaction) = self.pending_request
+            && let Some(waiter) = pending_compaction.waiter.take_inner()
+        {
+            waiter.send(Err(err.clone()).context(CompactRegionSnafu {
+                region_id: self.region_id,
+            }));
```
evenyag, line 553:
```suggestion
        if let Some(mut pending_compaction) = self.pending_request
        {
            pending_compaction.waiter.send(Err(err.clone()).context(CompactRegionSnafu {
                region_id: self.region_id,
            }));
```

v0y4g3r: > Some(mut pending_compaction) = sel

fixed in 5d56727968

PR #5409, opened 2025-01-20
feat: pipeline dispatcher part 2: execution

sunng87:

#5326

Add dispatcher execution for pipeline. This patch includes changes:

1. refactor pipeline execution in `pipeline.exec_mut` and return a new type response called `Dispatched` to indicate that this pipeline want to dispatch its intermediate results to another one.
1. refactor pipeline execution in `event.rs` so pipeline can be executed recursively when it dispatched to other pipeline
1. refactored `event.rs` to move some pipeline execution logic to `servers/src/pipeline.rs`. we will eventually move these type of logic into the pipeline module by defining `PipelineHandler` alternatives in pipeline module
1. allow an intermediate results to be fed into the special `identity` pipeline
1. changed otlp pipeline execution to use `serde_json::Value` from otlp data instead of pipeline value
1. changed dryrun output to a json array because pipeline now returns multiple insert requests (for multiple tables)
2. Renamed `table_part` of dispatch rule to `table_suffix` for more accurate meaning.

This is the second part of dispatcher implementation. ~~Note that we will still have part 3 for full and correct dispatcher.~~ We already included full implementation of pipeline dispatch mechanism in this patch. Because the foundation of previous code design has changed, we refactored the whole pipeline execution system to use `BTreeMap` as intermediate state. 

## PR Checklist


src/servers/src/pipeline.rs
```diff
@@ -0,0 +1,224 @@
...
+        let pipeline = get_pipeline(pipeline_definition, state, query_ctx).await?;
+
+        let transform_timer = std::time::Instant::now();
+        let mut intermediate_state = pipeline.init_intermediate_state();
+
+        let mut transformed = Vec::with_capacity(values.len());
+        let mut dispatched: BTreeMap<DispatchedTo, Vec<Vec<pipeline::Value>>> = BTreeMap::new();
+
+        match values {
+            PipelineExecInput::Original(array) => {
```
shuiyisong, line 127: `let mut intermediate_state = pipeline.init_intermediate_state();`
Can we move this line under 
`PipelineExecInput::Original(array) => {`

shuiyisong (approved):
LGTM
note: We might need to update the dashboard since the return for `dryrun` is changed.

sunng87:
@shuiyisong Yes I confirmed with Changlong the dashboard only dumps response in the textarea so it's safe to change this.

sunng87:
Note that this patch is still imcomplete for dispatcher implementation. The accuracy of keys from `intermediate_keys` needs further confirmation from @paomian

src/servers/src/pipeline.rs
sunng87: good catch!

paomian:
Because of the need to implement dispatch functionality, the original intermediate state could not be adapted to this new requirement. So we almost rewrote the pipeline's intermediate state. @sunng87

paomian (approved):

PR #5412, opened 2025-01-20
chore: extract `service_name` in OTLP traces by default

shuiyisong:

Although OTeL doesn't treat `service.name` differently, the popular GUI tool for displaying traces, [Jaeger](https://www.jaegertracing.io/), seems to have a dedicated API for querying all distinct service names. In that case, we need to extract the `service_name` in advance for indexing.

## PR Checklist


zyy17 (approved):
lgtm

sunng87 (approved):

sunng87:
@zyy17 I'm afraid we cannot. In trace we need to ensure all entries with same trace id are in the same table and same partition. Otherwise I think it will be difficult to do the query

PR #5430, opened 2025-01-23
chore: update toolchain to `2024-12-25`

shuiyisong:

This PR updates the nightly toolchain to `2024-12-25`. There seems to be a [false-positive check](https://github.com/rust-lang/rust-clippy/issues/13885) in clippy which has not been fixed, so we're not updating to a higher version.

## PR Checklist


shuiyisong:
https://github.com/GreptimeTeam/greptimedb/blob/f0d30a0f2610be3df7ebfddb423661fa03b16f86/Makefile#L11
@zyy17 how do we update the dev_builder?

src/common/version/build.rs
```diff
@@ -19,10 +19,13 @@ use build_data::{format_timestamp, get_source_time};
...
-        "cargo:rerun-if-changed={}/.git/refs/heads",
-        env!("CARGO_RUSTC_CURRENT_DIR")
-    );
+    // The "CARGO_WORKSPACE_DIR" is set manually (not by Rust itself) in Cargo config file, to
+    // solve the problem where the "CARGO_MANIFEST_DIR" is not what we want when this repo is
+    // made as a submodule in another repo.
+    let src_path = env::var("CARGO_WORKSPACE_DIR").or_else(|_| env::var("CARGO_MANIFEST_DIR"))?;
+
+    // `CARGO_RUSTC_CURRENT_DIR` is removed, use `src_path` instead.
+    println!("cargo:rerun-if-changed={}/.git/refs/heads", src_path);
```
shuiyisong, line 28: The `CARGO_RUSTC_CURRENT_DIR` is removed due to [this](https://github.com/rust-lang/cargo/issues/14798). However, I haven't found a way to set this ENV correctly. `CARGO_WORKSPACE_DIR` is not set and `CARGO_MANIFEST_DIR` is wrong.

src/puffin/src/puffin_manager/cache.rs
```diff
@@ -27,8 +27,9 @@ pub struct PuffinMetadataCache {
     cache: moka::sync::Cache<String, Arc<FileMetadata>>,
 }
 
+#[allow(clippy::ptr_arg)]
 fn puffin_metadata_weight(k: &String, v: &Arc<FileMetadata>) -> u32 {
-    (k.as_bytes().len() + v.memory_usage()) as u32
+    (k.len() + v.memory_usage()) as u32
 }
```
shuiyisong, line 33: Clippy complains about changing `&String` to `&str`. However, if we do so it then complains about `k` in `.eviction_listener(|k, v, _cause| {` doesn't have a size known at compile-time. So I just add `#[allow(clippy::ptr_arg)]` to pass the check 😢 @zhongzc

sunng87 (approved):

sunng87:
I need to rebase main and update a checksum in `flake.nix` for this. Just a moment.

sunng87:
done

src/common/version/build.rs
shuiyisong: According to https://github.com/baoyachi/shadow-rs/issues/175#issuecomment-2543936776 upgrade `shadow-rs` and use `BuildPattern::Lazy` should fix the problem. Let's try and see.

src/puffin/src/puffin_manager/cache.rs
sunng87: We can use `&str` for key while changing line 41 to 

```rust
weigher(|k: &String, v| puffin_metadata_weight(k, v))
```

evenyag (approved):

Issue #5072, opened 2024-11-29
Support kill query thread_id


adminpass:
### What problem does the new feature solve?

Query can be interrupted in advance

### What does the feature do?

kill thread_id

evenyag:
Kill will send the connection id. We may need to track the query and link it to the connection id.
https://github.com/databendlabs/databend/issues/4871

evenyag:
Let's keep it open to track the kill query feature.

PR #5448, opened 2025-01-24
feat: support sleep function

yihong0618:

This patch is first part of https://github.com/GreptimeTeam/greptimedb/issues/5072 
to support sleep function to let the query sleep a time, then the kill query `ctrl +c` can easily to use.

Some notes or need to disscuss
- for now all the function seems not use `async` this patch follows it
- the sleep function using thread sleep when use `async` it can use `tokio`
- the pg_sleep is tricky way(just using regrex and replace as the other code tricky
- for now seems most of them do not support  table function, this sleep also follows it like `select now(); --ok` `select * from now(); --not ok`

## PR Checklist


src/common/function/src/scalars/sleep.rs
```diff
@@ -0,0 +1,170 @@
...
+                Value::Int64(x) => x as f64,
+                Value::Int32(x) => x as f64,
+                Value::Float64(x) => x.into_inner(),
+                _ => {
+                    result.push(None);
+                    continue;
+                }
+            };
+            // Sleep for the specified seconds TODO: use tokio::time::sleep when the scalars are async
+            thread::sleep(Duration::from_secs_f64(secs));
```
evenyag, line 88: I'd suggest not to sleep in a scalar function. This pauses the runtime thread which is risky.

src/common/function/src/function_registry.rs
```diff
@@ -122,6 +123,9 @@ pub static FUNCTION_REGISTRY: Lazy<Arc<FunctionRegistry>> = Lazy::new(|| {
     // Vector related functions
     VectorFunction::register(&function_registry);
 
+    // Sleep function
+    SleepFunction::register(&function_registry);
```
evenyag, line 127: I'm hesitant about whether to provide a sleep function. We may only provide a sleep function under the admin statement if we really need this.
```sql
admin sleep(60);
```

If our final goal is to support kill query, the sleep function may not be the first thing we need to implement. We can implement a way to cancel a query first.

evenyag (commented):
Sorry for the late review. I'm not sure whether we should provide the sleep function, since this is mostly for test purpose.

yihong0618:
> Sorry for the late review. I'm not sure whether we should provide the sleep function, since this is mostly for test purpose.

make sense will draft it for now until we discussed it

yihong0618:
closing this for now, since maybe a better solution

PR #5450, opened 2025-01-24
feat: provide options to disable or customize http corss-origin settings

sunng87:

This patch adds two new options to customize http cross-origin behaviour to customize it or configure a whitelist of origins.

## PR Checklist


config/frontend.example.toml
```diff
@@ -31,6 +31,12 @@ timeout = "30s"
 ## The following units are supported: `B`, `KB`, `KiB`, `MB`, `MiB`, `GB`, `GiB`, `TB`, `TiB`, `PB`, `PiB`.
 ## Set to 0 to disable limit.
 body_limit = "64MB"
+## HTTP CORS support, it's turned on by default
+## This allows browser to access http APIs without CORS restrictions
+disable_cors = false
```
killme2008, line 36: Always use a positive name like `enable_xxx`.

src/cmd/tests/load_config_test.rs
```diff
@@ -234,6 +239,10 @@ fn test_load_standalone_example_config() {
                 remote_write: Some(Default::default()),
                 ..Default::default()
             },
+            http: HttpOptions {
+                cors_allowed_origins: vec!["https://example.com".to_owned()],
```
killme2008, line 243: Keep consistent with the above.
```suggestion
                cors_allowed_origins: vec!["https://example.com".to_string()],
```

src/servers/src/http.rs
```diff
@@ -1177,17 +1203,123 @@ mod test {
...
+            origin
+        );
+
+        let res = client
+            .get("/health")
+            .header("Origin", "https://notallowed.com")
+            .send()
+            .await;
+
+        assert_eq!(res.status(), StatusCode::OK);
```
killme2008, line 1296: A 403 Forbidden response would be more appropriate.

sunng87: This behaviour is controlled by tower's cors middleware. Also in cors, server doesn't reject requests, it's totally decided by client /browser.

killme2008 (approved):
LGTM. Don't forget to update the PR title and content.

shuiyisong (approved):

sunng87:
@killme2008  It's still option to disable because it's enabled by default.

PR #5451, opened 2025-01-24
perf: optimize time series memtable ingestion

waynexia:

Some tiny optimizations for time series memtable (~5% improvement).

TSBS result:

Before:

image

After:

image

## PR Checklist


yihong0618:
learned how to perf from this, thank you very much

src/mito2/src/memtable/time_series.rs
```diff
@@ -54,7 +54,7 @@ use crate::region::options::MergeMode;
 use crate::row_converter::{DensePrimaryKeyCodec, PrimaryKeyCodecExt};
 
 /// Initial vector builder capacity.
-const INITIAL_BUILDER_CAPACITY: usize = 0;
+const BUILDER_CAPACITY: usize = 512;
```
evenyag, line 57: The init capacity 0 is to avoid each Series pre-allocating too much memory when the points in each series are far less than the capacity.

For example, if we have 8k series and each series has 100 points, then the memtable may use 5x memory than before. What's more, the write buffer size doesn't consider the builder capacity now, so we may need to bump the write buffer size once we create a builder.

Maybe we can freeze the builder earlier once the number of items reaches a threshold.

evenyag (commented):
@v0y4g3r PTAL

src/datatypes/src/vectors.rs
```diff
@@ -202,13 +202,16 @@ pub trait MutableVector: Send + Sync {
...
     /// Panics if error if data types mismatch.
     fn push_value_ref(&mut self, value: ValueRef) {
+        #[cfg(debug_assertions)]
         self.try_push_value_ref(value).unwrap_or_else(|_| {
             panic!(
                 "expecting pushing value of datatype {:?}, actual {:?}",
                 self.data_type(),
                 value
             );
         });
```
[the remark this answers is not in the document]
waynexia: change to ignoring errors on `push` in memtable.

src/mito2/src/memtable/time_series.rs
waynexia: Add a two-step capacity that starts from 16

v0y4g3r (approved):
LGTM

src/mito2/src/memtable/time_series.rs
```diff
@@ -625,12 +628,24 @@ impl Series {
...
+        &mut self,
+        ts: ValueRef<'a>,
+        sequence: u64,
+        op_type: OpType,
+        values: impl Iterator<Item = ValueRef<'a>>,
+    ) -> usize {
+        // + 10 to avoid potential reallocation.
+        if self.active.len() + 10 > BUILDER_CAPACITY {
+            let region_metadata = self.region_metadata.clone();
+            self.freeze(&region_metadata);
```
evenyag, line 646: Looks like we should use `BUILDER_CAPACITY` as the capacity of freeze. We may update the capacity in `freeze()`.
https://github.com/GreptimeTeam/greptimedb/blob/5b6279f191f44f3460583f03a39e9265fa1c7047/src/mito2/src/memtable/time_series.rs#L640-L647

evenyag (approved):
Others LGTM

Issue #5401, opened 2025-01-20
Support Jaeger query APIs


zyy17:
### What problem does the new feature solve?

The current GreptimeDB supports ingesting trace data by OTel; however, we don't provide a better way to render the data. To use Grafana or Jaeger UI, we should support Jaeger query [APIs](https://www.jaegertracing.io/docs/2.2/apis/). 

### What does the feature do?

Implement Jaeger query APIs.

### Implementation challenges

Although the Jaeger has v2 and v3 APIs, the v2 APIs are more compatible. We can implement v2 APIs at first.

The following APIs are used in [Grafana Jaeger datasource](https://github.com/grafana/grafana/tree/main/public/app/plugins/datasource/jaeger):

- `/api/services/`;
- `/api/traces`;
- `/api/dependencies`: Actually, get dependencies can be a flow task in GreptimeDB;

PR #5452, opened 2025-01-24
feat: implement Jaeger query APIs

zyy17:

Resolve https://github.com/GreptimeTeam/greptimedb/issues/5401.

This PR implements the following Jaeger query HTTP APIs with the **`/v1/jaeger`** endpoint prefix:

- `/api/services`;
- `/api/operations`;
- `/api/traces`;

Btw, the `/api/dependencies` API will be supported by the next PR.

The current OTel table design may not be the most efficient. When the table size is large, the efficiency of Jaeger queries may be relatively slow. However, performance tuning could be the next topic for tracing observability.

The PR implements Jaeger APIs by following **key steps**:

1. Define the `JaegerQueryHandler` in `src/servers/src/query_handler.rs`;
2. Accept the query parameters from http requests and pass them to `JaegerQueryHandler` to execute the query;
3. Use DataFusion DataFrame APIs to query trace data;

## PR Checklist


src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+
+        // Set trace id.
+        if let Some(JsonValue::String(trace_id)) = row_iter.next() {
+            span.trace_id = trace_id.clone();
+            trace_id_to_processes.entry(trace_id).or_default();
+        }
+
+        // Convert timestamp from nanoseconds to microseconds.
+        if let Some(JsonValue::Number(timestamp)) = row_iter.next() {
+            span.start_time = timestamp.as_u64().unwrap_or(0) / 1000;
```
sunng87, line 661: Do you think if we need to raise an error here instead of using default 0?

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+fn traces_from_records(records: HttpRecordsOutput) -> Vec<Trace> {
+    // maintain the mapping: trace_id -> (process_id -> service_name).
+    let mut trace_id_to_processes: HashMap<String, HashMap<String, String>> = HashMap::new();
+    // maintain the mapping: trace_id -> spans.
+    let mut trace_id_to_spans: HashMap<String, Vec<Span>> = HashMap::new();
+
+    // FIXME(zyy17): Should return the error if the schema is not correct.
+    for row in records.rows.into_iter() {
+        let mut span = Span::default();
+        let mut row_iter = row.into_iter();
```
sunng87, line 651: please double check if the column order is guaranteed

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+        );
+    }
+    query
+}
+
+fn query_trace_sql(trace_table_name: &str, query_params: QueryTraceParams) -> String {
+    let mut query = format!(
+        "SELECT trace_id, timestamp, duration_nano, service_name, span_name, span_id, span_attributes FROM {}.{} WHERE service_name = '{}'",
+        query_params.db, trace_table_name, query_params.service_name
+    );
```
sunng87, line 822: sql injection warning. we should use parser to validate the query generated from this method contains only 1 statement.

tests-integration/tests/http.rs
```diff
@@ -946,6 +947,9 @@ enable = true
 [influxdb]
 enable = true
 
+[jaeger]
+enable = true
```
sunng87, line 951: remember to update sample configurations in `config/`

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+
+/// JaegerAPIResponse is the response of Jaeger HTTP API.
+/// The original version is `structuredResponse` which is defined in https://github.com/jaegertracing/jaeger/blob/main/cmd/query/app/http_handler.go.
+#[derive(Default, Debug, Serialize, Deserialize, PartialEq)]
+pub struct JaegerAPIResponse<T> {
+    pub data: Option<T>,
+    pub total: i32,
+    pub limit: i32,
+    pub offset: i32,
+    pub errors: Option<Vec<JaegerAPIError>>,
```
killme2008, line 42: What about changing the errors to `Vec<JaegerAPIError>`?

And adds `#[serde(skip_serializing_if = "Vec::is_empty")]` for it.

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+}
+
+/// Trace is a collection of spans.
+#[derive(Debug, Default, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct Trace {
+    #[serde(rename = "traceID")]
+    pub trace_id: String,
+    pub spans: Vec<Span>,
+    pub processes: Option<HashMap<String, Process>>,
```
killme2008, line 71: The same suggestions as above for `processes` and `warnings`. Usually, we don't use `Option` to wrap vector or map; the empty means `None`.

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+    pub logs: Vec<Log>,
+
+    #[serde(rename = "processID")]
+    #[serde(skip_serializing_if = "String::is_empty")]
+    pub process_id: String,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub process: Option<Process>,
+
+    pub warnings: Option<Vec<String>>,
```
killme2008, line 106: The same suggestion as above.

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+    pub tags: Option<String>,
+
+    #[serde(rename = "spanKind")]
+    pub span_kind: Option<String>,
+}
+
+impl QueryTraceParams {
+    fn from_jaeger_query_params(query_params: JaegerQueryParams) -> Result<Self> {
+        let mut internal_query_params: QueryTraceParams = QueryTraceParams {
+            db: query_params.db.unwrap_or("public".to_string()),
```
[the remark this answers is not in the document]
killme2008: Use `unwrap_or_else(|| public".to_string())` instead to avoid unnecessary copy.

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+                    *v = number;
+                }
+                if let Some(boolean) = convert_string_to_boolean(v) {
+                    *v = boolean;
+                }
+            }
+            internal_query_params.tags = Some(tags_map);
+        }
+
+        if let Some(limit) = query_params.limit {
```
killme2008, line 278: Use `Option::map` to simply the code.

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+struct QueryTraceParams {
+    db: String,
+    service_name: String,
+    operation_name: Option<String>,
+
+    // The limit of the number of traces to return.
+    limit: Option<i32>,
+
+    // FIXME(zyy17): Support tags query in the future.
+    tags: Option<HashMap<String, JsonValue>>,
```
killme2008, line 299: Don't use `Option` to wrap collections.

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+                        services.push(service_name);
+                    }
+                }
+            }
+            let services_num = services.len();
+            (
+                StatusCode::OK,
+                axum::Json(JaegerAPIResponse {
+                    data: Some(services),
+                    total: services_num as i32,
```
killme2008, line 347: What about chang the type of total into `usize`?

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+    Extension(mut query_ctx): Extension<QueryContext>,
+) -> impl IntoResponse {
+    debug!(
+        "Received Jaeger '/api/traces/{}' request, query_params: {:?}",
+        trace_id, query_params
+    );
+    query_ctx.set_channel(Channel::Jaeger);
+    let query_ctx = Arc::new(query_ctx);
+
+    let db = query_params.db.unwrap_or("public".to_string());
```
killme2008, line 451: Always use `unwrap_or_else`

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+            }
+            operations.push(operation);
+        }
+    }
+    operations
+}
+
+// FIXME(zyy17): It may encounter performance issue when the trace table is too large.
+fn get_services_sql(db: &str, trace_table_name: &str) -> String {
+    format!(
```
killme2008, line 783: This will introduce SQL injection. I think we should use `DataFrame` instead.

Just like https://github.com/GreptimeTeam/greptimedb/blob/88c3d331a1039832e8cbe4e1ac71147a687e5c33/src/query/src/sql.rs#L222

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+    }
+
+    if let Some(limit) = query_params.limit {
+        query = format!("{} LIMIT {}", query, limit);
+    }
+
+    query
+}
+
+fn normalize_span_kind(span_kind: &str) -> String {
```
killme2008, line 883: What about moving thins function into `SpanKind`?

src/servers/src/http/jaeger.rs
```diff
@@ -0,0 +1,1244 @@
...
+                InvalidJaegerQuerySnafu {
+                    reason: format!("parse maxDuration '{}' failed: {}", max_duration, e),
+                }
+                .build()
+            })?;
+            internal_query_params.max_duration = Some(duration.as_nanos() as u64);
+        }
+
+        if let Some(min_duration) = query_params.min_duration {
+            let duration = humantime::parse_duration(&min_duration).map_err(|e| {
```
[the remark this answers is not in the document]
zyy17: I don't define the new error for duration errors. Instead, I use the error output as part of the reason in `InvalidJaegerQuerySnafu`. It can't use `context` in Snafu.

src/servers/src/http/jaeger.rs
zyy17:
```
internal_query_params.limit = Some(query_params.limit.unwrap_or(100));
```

src/servers/src/http/jaeger.rs
zyy17: use `let db = query_ctx.get_db_string();` instead.

src/servers/src/http/jaeger.rs
zyy17: I create `check_schema()` to check it.

src/servers/src/http/jaeger.rs
zyy17: I use the DataFrame APIs to refactor all the query process. It may not have SQL injection problem in theory.

src/servers/src/http/jaeger.rs
zyy17: I use the DataFrame APIs to refactor all the query process. It may not have SQL injection problem in theory.

src/servers/src/http/jaeger.rs
zyy17: `span_kind` is a `String` type and I think it may be unnecessary to define the the new type and implement the above method.

src/servers/src/http/jaeger.rs
zyy17: Sounds reasonable.

zyy17:
@WenyXu I use [some](https://github.com/zyy17/greptimedb/blob/feat/jaeger-api/src/frontend/src/instance/jaeger.rs#L265-L330) JSON functions for tags filter, PTAL.

src/frontend/src/instance/jaeger.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    let output = Output::new_with_stream(Box::pin(
+        RecordBatchStreamAdapter::try_new(stream).context(CollectRecordbatchSnafu)?,
+    ));
+
+    Ok(output)
+}
+
+// To decouple from the global session context, it will create a new session context and register the necessary UDFs.
+fn create_df_context(
```
sunng87, line 242: Any background for this? Is it really needed to use a separated context?

sunng87 (approved):
Much improved now

src/frontend/src/instance/jaeger.rs
zyy17: Guide by @evenyag . Also, I add the following comment for background:

```
// The current implementation registers UDFs during the planning stage, which makes it difficult
// to utilize them through DataFrame APIs. To address this limitation, we create a new session
// context and register the required UDFs, allowing them to be decoupled from the global context.
```

src/frontend/src/instance/jaeger.rs
```diff
@@ -0,0 +1,336 @@
...
+) -> ServerResult<Output> {
+    let db = ctx.get_db_string();
+    let table = catalog_manager
+        .table(ctx.current_catalog(), &db, TRACE_TABLE_NAME, Some(&ctx))
+        .await
+        .context(CatalogSnafu)?
+        .with_context(|| TableNotFoundSnafu {
+            table: TRACE_TABLE_NAME,
+            catalog: ctx.current_catalog().to_string(),
+            schema: db.to_string(),
```
killme2008, line 201: Snafu will auto convert the string references into strings.
```suggestion
            catalog: ctx.current_catalog(),
            schema: db,
```

src/frontend/src/instance/jaeger.rs
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// The current implementation registers UDFs during the planning stage, which makes it difficult
+// to utilize them through DataFrame APIs. To address this limitation, we create a new session
+// context and register the required UDFs, allowing them to be decoupled from the global context.
+fn create_df_context(
+    query_engine: &QueryEngineRef,
+    ctx: QueryContextRef,
+) -> ServerResult<SessionContext> {
+    let df_context = SessionContext::new_with_state(
```
killme2008, line 251: I think we can leave a TODO here. It's possible to resue this `SessionContext`.

src/frontend/src/service_config/jaeger.rs
```diff
@@ -0,0 +1,37 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
```
killme2008, line 17: Add a description for this public structure.

killme2008 (approved):
Almost LGTM

PR #5456, opened 2025-01-25
feat: write memtable in parallel

waynexia:

Write memtable in parallel. No more 1 core sucks 127 cores rest.

Previous ~550k, After:

image

And the CPU usages are all below 50%. The next step is pipelining the write process to alleviate WAL's influence.

## PR Checklist


src/mito2/src/region_write_ctx.rs
```diff
@@ -197,23 +197,38 @@ impl RegionWriteCtx {
...
+        if mutations.len() == 1 {
+            if let Err(err) = mutable.write(&mutations[0].1) {
+                self.notifiers[mutations[0].0].err = Some(Arc::new(err));
+            }
+        } else {
+            let write_tasks = mutations.into_iter().map(|(i, kvs)| {
+                let mutable = mutable.clone();
+                // use tokio runtime to schedule tasks.
+                common_runtime::spawn_blocking_global(move || (i, mutable.write(&kvs)))
+            });
```
v0y4g3r, line 226: Can we use something like rayon instead of spawn futures that inherently not async?

src/mito2/src/region_write_ctx.rs
```diff
@@ -197,23 +197,38 @@ impl RegionWriteCtx {
...
+            for result in futures::future::join_all(write_tasks).await {
+                let (i, result) = result.unwrap();
```
v0y4g3r, line 228: Is the unwrap safe here?

src/mito2/src/region_write_ctx.rs
waynexia: rayon can't reuse existing tokio managed threads, and I'm hesitant to create another group of runtimes... all those tasks go into spawn_blocking which isn't async, I think that's fine.

BTW I do find a combination of rayon + tokio https://github.com/andybarron/tokio-rayon, but the last update time is 4 years ago 🤣

src/mito2/src/region_write_ctx.rs
waynexia: That Result is from `spawn`, I'll make it more explicit

src/mito2/src/region_write_ctx.rs
```diff
@@ -197,23 +197,39 @@ impl RegionWriteCtx {
...
+            if let Err(err) = mutable.write(&mutations[0].1) {
+                self.notifiers[mutations[0].0].err = Some(Arc::new(err));
+            }
+        } else {
+            let write_tasks = mutations.into_iter().map(|(i, kvs)| {
+                let mutable = mutable.clone();
+                // use tokio runtime to schedule tasks.
+                common_runtime::spawn_blocking_global(move || (i, mutable.write(&kvs)))
+            });
+            for result in futures::future::join_all(write_tasks).await {
```
discord9, line 227: join_all is ordered, maybe using [`FuturesUnordered`](https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html) makes more sense since already record the order?

waynexia: from its document the ordered is more performant, let me have a try  https://docs.rs/futures/latest/futures/future/fn.join_all.html#see-also

discord9: > from its document the ordered is more performant, let me have a try https://docs.rs/futures/latest/futures/future/fn.join_all.html#see-also

by performant it mean compare to itself, `JoinAll` is like this:
```rs
pub struct JoinAll<F>
where
    F: Future,
{
    kind: JoinAllKind<F>,
}

#[cfg_attr(target_os = "none", cfg(target_has_atomic = "ptr"))]
pub(crate) const SMALL: usize = 30;

enum JoinAllKind<F>
where
    F: Future,
{
    Small {
        elems: Pin<Box<[MaybeDone<F>]>>,
    },
    #[cfg_attr(target_os = "none", cfg(target_has_atomic = "ptr"))]
    Big {
        fut: Collect<FuturesOrdered<F>, Vec<F::Output>>,
    },
}
```
So it's ordered even <30 futures, but do we need ordered here or is unordered good enough

discord9 (approved):
otherwise LGTM

v0y4g3r (approved):
LGTM

Issue #5414, opened 2025-01-21
GreptimeDB Crash Report


zengfanlin:
### What type of bug is this?

Crash

### What subsystems are affected?

Standalone mode

### Minimal reproduce step

I send a export command line .\greptime cli export --addr localhost:4000  --output-dir /tmp/backup/timerange --start-time "2024-01-01 00:00:00"  --end-time "2024-01-31 23:59:59"

### What did you expect to see?

export data

### What did you see instead?

crash

[report-e1ce3ef0-c21f-4ba9-ae01-ef93cb3455dd.txt](https://github.com/user-attachments/files/18485201/report-e1ce3ef0-c21f-4ba9-ae01-ef93cb3455dd.txt)

### What operating system did you use?

Windows 10 x64

### What version of GreptimeDB did you use?

0.11.2

### Relevant log output and stack trace

```bash

```

yihong0618:
when you use windows -output-dir /tmp/backup/timerange  the dir  name maybe need to follow windows rules

evenyag:
> when you use windows -output-dir /tmp/backup/timerange the dir name maybe need to follow windows rules

Yes, you may need to check if the database can write to this path.

Could you please check the logs of the server? Does it report any errors?

discord9:
I test it on windows and it seems if the provided http addr can't be connected then will report this error, maybe check again if the db processs is actually hosting a http server at localhost:4000, meanwhile we can add a more friendly message when failed to connect

PR #5459, opened 2025-01-26
feat(cli): add proxy options

discord9:

as title, if not set `proxy` or `no_proxy` still auto use sys proxy like before

## PR Checklist

### DIFF
--- a/src/cli/src/database.rs
+++ b/src/cli/src/database.rs
@@ -45,14 +45,13 @@ pub fn parse_proxy_opts(
     if no_proxy {
         return Ok(None);
     }
-    if let Some(proxy) = proxy {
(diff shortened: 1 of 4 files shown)


killme2008:
What is the scenario that needs this proxy server setting?

discord9:
> What is the scenario that needs this proxy server setting?

the no-proxy option is necessary since system proxy on windows will check not only env var but also registry for system wide proxy, so it helps to have a no-proxy option to opt out of this(#5414 encounter this) the proxy setting is just only have a no-proxy opt seems wired

evenyag (approved):

killme2008 (approved):
LGTM

PR #5464, opened 2025-01-27
ci: move components to flakes so it won't affect builders

sunng87:

This change removes rust components from the toolchain file and add them to flake so the dev builders won't need to install components that required for dev and ci.  

## PR Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,7 +18,11 @@
           libgit2
           libz
         ];
-
+        lib = nixpkgs.lib;
+        rustToolchain = fenix.packages.${system}.fromToolchainName {
+          name = (lib.importTOML ./rust-toolchain.toml).toolchain.channel;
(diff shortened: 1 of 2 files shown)


waynexia (approved):

flake.nix
```diff
@@ -18,7 +18,11 @@
           libgit2
           libz
         ];
-
+        lib = nixpkgs.lib;
+        rustToolchain = fenix.packages.${system}.fromToolchainName {
+          name = (lib.importTOML ./rust-toolchain.toml).toolchain.channel;
+          sha256 = "sha256-f/CVA1EC61EWbh0SjaRNhLL0Ypx2ObupbzigZp8NmL4=";
```
evenyag, line 24: Does it mean that we need to update the sha256 if we want to bump the toolchain?

evenyag (approved):

flake.nix
sunng87: Yes, it's designed to ensure all dependencies are tracked by nix. If toolchain file changed, there will be a checksum mismatch.

Issue #5466, opened 2025-02-04
a table with logical region that can not delete but can delete at the second time, and break others


yihong0618:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Other, Standalone mode, Distributed Cluster

### Minimal reproduce step

- cargo build
- ./target/debug/greptime standalone start

run this sql in psql(do not restart the standalone
```sql
CREATE TABLE phy (ts timestamp time index, val double) engine=metric with ("physical_metric_table" = "");

CREATE TABLE t1 (ts timestamp time index, val double, host string primary key) engine = metric with ("on_physical_table" = "phy");

CREATE TABLE t2 (ts timestamp time index, job string primary key, val double) engine = metric with ("on_physical_table" = "phy");

```

then

- try to delete the phy

```sql
drop table phy;
-- show
-- Physical region 4398046511104(1024, 0) is busy, there are still some logical regions using it
```

then try to delete it again(no need the sam session but need to keep the same `standalone`(without restart)
```sql
drop table phy;
-- show
-- OK 0
```

### What did you expect to see?

the second also can not be deleted 

### What did you see instead?

as the desc

![Image](https://github.com/user-attachments/assets/54263eee-8c5f-44bf-9b01-718a35bb22b5)

and break all t1 and t2

![Image](https://github.com/user-attachments/assets/6af865e2-eebd-4f77-92a1-2ecc0572e688)

### What operating system did you use?

Mac but no matters

### What version of GreptimeDB did you use?

main branch but seems all version

### Relevant log output and stack trace

```bash
2025-02-04T08:08:50.140790Z ERROR LocalManager::submit_root_procedure: common_procedure::local::runner: Failed to execute procedure metasrv-procedure::DropTable-301d097a-f93d-4fb8-92cd-92704c016ef5, retry: false err=0: Failed to execute procedure due to external error
1: Failed to operate on datanode: peer-0(), at src/common/meta/src/ddl/utils.rs:37:18
2: External error, at src/frontend/src/instance/standalone.rs:87:14
3: Operation to region server failed, at src/frontend/src/instance/standalone.rs:69:14
4: Execute gRPC request error, at src/datanode/src/region_server.rs:396:22
5: Failed to handle request for region 4398046511104(1024, 0), at src/datanode/src/region_server.rs:770:14
6: Physical region 4398046511104(1024, 0) is busy, there are still some logical regions using it, at src/metric-engine/src/engine/drop.rs:60:18
[...]

PR #5467, opened 2025-02-04
fix: close issue #5466 by do not shortcut the drop command

yihong0618:

Happy new year.

after some debugging for #5466 just found that
the second `drop table` is from short cut in handle_request:

```rs
        let region_change = match &request {
            RegionRequest::Create(create) => {
                let attribute = parse_region_attribute(&create.engine, &create.options)?;
                RegionChange::Register(attribute)
            }
            RegionRequest::Open(open) => {
                let attribute = parse_region_attribute(&open.engine, &open.options)?;
                RegionChange::Register(attribute)
            }
            RegionRequest::Close(_) => RegionChange::Deregisters,
            RegionRequest::Put(_)
            | RegionRequest::Drop(_)
            | RegionRequest::Delete(_)
            | RegionRequest::Alter(_)
            | RegionRequest::Flush(_)
            | RegionRequest::Compact(_)
            | RegionRequest::Truncate(_) => RegionChange::None,
            RegionRequest::Catchup(_) => RegionChange::Catchup,
        };

        let engine = match self.get_engine(region_id, &region_change)? {
            CurrentEngine::Engine(engine) => engine,
            CurrentEngine::EarlyReturn(rows) => return Ok(RegionResponse::new(rows)),
        };
```

the reason is that for the `RegionRequest::Drop(_)` the old logic is `Deregisters` for that it can be early return,
but for this issue case it can not  be do this way, the table had logical region when early return it will not pass the 
` logical region` check in function `drop_region` 

```rs 
    fn get_engine(
        &self,
        region_id: RegionId,
        region_change: &RegionChange,
    ) -> Result<CurrentEngine> {
        let current_region_status = self.region_map.get(&region_id);

        let engine = match region_change {
            RegionChange::Register(attribute) => match current_region_status {
                Some(status) => match status.clone() {
                    RegionEngineWithStatus::Registering(engine) => engine,
                    RegionEngineWithStatus::Deregistering(_) => {
                        return error::RegionBusySnafu { region_id }.fail()
                    }
                    RegionEngineWithStatus::Ready(_) => status.clone().into_engine(),
                },
                _ => self
                    .engines
                    .read()
                    .unwrap()
                    .get(attribute.engine())
                    .with_context(|| RegionEngineNotFoundSnafu {
                        name: attribute.engine(),
                    })?
                    .clone(),
            },
            RegionChange::Deregisters => match current_region_status {
                Some(status) => match status.clone() {
                    RegionEngineWithStatus::Registering(_) => {
                        return error::RegionBusySnafu { region_id }.fail()
                    }
                    RegionEngineWithStatus::Deregistering(_) => {
                        return Ok(CurrentEngine::EarlyReturn(0))
                    }
                    RegionEngineWithStatus::Ready(_) => status.clone().into_engine(),
                },
                None => return Ok(CurrentEngine::EarlyReturn(0)),
            },
            RegionChange::None | RegionChange::Catchup => match current_region_status {
                Some(status) => match status.clone() {
                    RegionEngineWithStatus::Registering(_) => {
                        return error::RegionNotReadySnafu { region_id }.fail()
                    }
                    RegionEngineWithStatus::Deregistering(_) => {
                        return error::RegionNotFoundSnafu { region_id }.fail()
                    }
                    RegionEngineWithStatus::Ready(engine) => engine,
                },
                None => return error::RegionNotFoundSnafu { region_id }.fail(),
            },
        };

        Ok(CurrentEngine::Engine(engine))
    }
``` 

also change the tests

seems sqlness limitation the case is not add to sqlness tests

---

update address the comments by set it back to `READY` when fail

## PR Checklist

### DIFF
--- a/src/datanode/src/region_server.rs
+++ b/src/datanode/src/region_server.rs
@@ -661,7 +661,7 @@ impl RegionServerInner {
                             }
                         }
                         Err(e) => {
-                            self.unset_region_status(region_id, *region_change);
+                            self.unset_region_status(region_id, &engine, *region_change);
                             error!(e; "Failed to open region: {}", region_id);
                             errors.push(e);
                         }
@@ -670,7 +670,7 @@ impl RegionServerInner {
             }
             Err(e) => {
                 for (&region_id, region_change) in &region_changes {
-                    self.unset_region_status(region_id, *region_change);
+                    self.unset_region_status(region_id, &engine, *region_change);
                 }
                 error!(e; "Failed to open batch regions");
                 errors.push(BoxedError::new(e));
@@ -780,7 +780,7 @@ impl RegionServerInner {
             }
             Err(err) => {
                 // Removes the region status if the operation fails.
-                self.unset_region_status(region_id, region_change);
+                self.unset_region_status(region_id, &engine, region_change);
                 Err(err)
             }
         }
@@ -809,11 +809,20 @@ impl RegionServerInner {
         }
     }
 
-    fn unset_region_status(&self, region_id: RegionId, region_change: RegionChange) {
+    fn unset_region_status(
+        &self,
+        region_id: RegionId,
+        engine: &RegionEngineRef,
+        region_change: RegionChange,
+    ) {
         match region_change {
             RegionChange::None => {}
-            RegionChange::Register(_) | RegionChange::Deregisters => {
+            RegionChange::Register(_) => {
                 self.region_map.remove(&region_id);
+            }
+            RegionChange::Deregisters => {
+                self.region_map
+                    .insert(region_id, RegionEngineWithStatus::Ready(engine.clone()));
             }
             RegionChange::Catchup => {}
         }
@@ -1195,7 +1204,7 @@ mod tests {
             .unwrap_err();
 
         let status = mock_region_server.inner.region_map.get(&region_id);
-        assert!(status.is_none());
+        assert!(status.is_some());
     }
 
     struct CurrentEngineTest {


WenyXu:
@yihong0618 Thank you for reporting the issue! However, we can't do that, we need to restore the `region_map` state instead of changing the logic.

WenyXu:
> @yihong0618 Thank you for reporting the issue! However, we can't do that, we need to restore the `region_map` state instead of changing the logic.

If the dropping table exists in `region_map` with the `Ready` state, we need to restore it to `Ready` if the operation fails; otherwise, we need to remove it from the `region_map` to avoid memory leaks.

yihong0618:
> @yihong0618 Thank you for reporting the issue! However, we can't do that, we need to restore the `region_map` state instead of changing the logic.

yes, also consider that, but the logic seems  a bit complicated than what we have now.

will try the READY way, maybe need some advice, if so will ping you tomorrow

yihong0618:
> > @yihong0618 Thank you for reporting the issue! However, we can't do that, we need to restore the `region_map` state instead of changing the logic.
> 
> If the dropping table exists in `region_map` with the `Ready` state, we need to restore it to `Ready` if the operation fails; otherwise, we need to remove it from the `region_map` to avoid memory leaks.

address the comments

killme2008:
@WenyXu @yihong0618 Do we need to ensure the previous state is `Ready` in the region map before putting it back?

yihong0618:
> @WenyXu @yihong0618 Do we need to ensure the previous state is `Ready` in the region map before putting it back?

seems we not not need it

evenyag (approved):

PR #5470, opened 2025-02-05
docs: fix memory perf command wrong

yihong0618:

for now the memory perf command using `tikv-jemallocator` had been changed
using the doc or blog way get error ` {"error":"Memory profiling is not enabled"}` and very hard to debug.
this patch fix it. about the `greptime blog` side I do not fix

more info:

- blog: https://magiroux.com/posts/rust-jemalloc-profiling
- issue: https://github.com/tikv/jemallocator/issues/65

## PR Checklist

### DIFF
--- a/docs/how-to/how-to-profile-memory.md
+++ b/docs/how-to/how-to-profile-memory.md
@@ -23,7 +23,7 @@ curl https://raw.githubusercontent.com/brendangregg/FlameGraph/master/flamegraph
 Start GreptimeDB instance with environment variables:
 
 ```bash
-MALLOC_CONF=prof:true ./target/debug/greptime standalone start
+_RJEM_MALLOC_CONF=prof:true ./target/debug/greptime standalone start
 ```
 
 Dump memory profiling data through HTTP API:
--- a/docs/how-to/how-to-profile-memory.md
+++ b/docs/how-to/how-to-profile-memory.md
@@ -4,6 +4,16 @@ This crate provides an easy approach to dump memory profiling info.
 
 ## Prerequisites
 ### jemalloc
+jeprof is already compiled in the target directory of GreptimeDB. You can find the binary and use it.
+```
+# find jeprof binary
+find . -name 'jeprof'
+# add executable permission
+chmod +x <path_to_jeprof>
+```
+The path is usually under `./target/${PROFILE}/build/tikv-jemalloc-sys-${HASH}/out/build/bin/jeprof`.
+The default version of jemalloc installed from the package manager may not have the `--collapsed` option.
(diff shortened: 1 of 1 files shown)


shuiyisong:
Interesting. I guess you're using this command on macOS. However, on Linux (arm64) it does the opposite, where `MALLOC_CONF` yields the correct output and `_RJEM_MALLOC_CONF` gives the error.

Stating commands per OS in the comments seems to be a better solution.

yihong0618:
> Interesting. I guess you're using this command on macOS. However, on Linux (amd64) it does the opposite, where `MALLOC_CONF` yields the correct output and `_RJEM_MALLOC_CONF` gives the error.
> 
> Stating commands per OS in the comments seems to be a better solution.

OK will add both command

docs/how-to/how-to-profile-memory.md
```diff
@@ -5,7 +5,7 @@ This crate provides an easy approach to dump memory profiling info.
 ## Prerequisites
 ### jemalloc
 ```bash
-# for macOS
+# for MacOS
```
shuiyisong, line 8:
```suggestion
# for macOS
```

evenyag, line 8: Apple uses `macOS` since Sierra. 
```suggestion
# for macOS
```

shuiyisong (approved):
Rest LGTM

docs/how-to/how-to-profile-memory.md
```diff
@@ -23,7 +23,11 @@ curl https://raw.githubusercontent.com/brendangregg/FlameGraph/master/flamegraph
 Start GreptimeDB instance with environment variables:
 
 ```bash
+# for Linux
 MALLOC_CONF=prof:true ./target/debug/greptime standalone start
+
+# for MacOS
```
evenyag, line 29:
```suggestion
# for macOS
```

docs/how-to/how-to-profile-memory.md
```diff
@@ -5,7 +5,7 @@ This crate provides an easy approach to dump memory profiling info.
 ## Prerequisites
 ### jemalloc
```
evenyag, line 6: If you don't mind, you can add a section for installing the correct version of jeprof. I wrote a doc for it but haven't submitted a PR.
https://github.com/GreptimeTeam/greptimedb/blob/build/add-jeprof/docs/how-to/how-to-profile-memory.md

yihong0618: of course will add it now, and add you to co-author

yihong0618: Done

evenyag (approved):

PR #5473, opened 2025-02-05
fix: Delete statement not supported in metric engine close #4649

### LINKED ISSUE #4649: Delete statement not supported in metric engine
### What type of bug is this?

Unexpected error

Table Engine

### Minimal reproduce step

Can't delete data from logical tables or physical table:

```sql
delete from up;
delete from greptime_physical_table;
```

Both report errors as:
```
Unsupported region request: Delete
```

```
Unsupported region request: Delete
```

### What did you see instead?

Execute successfully.

all

0.9.2

_No response_

yihong0618:

close #4649 also add sqlness tests

## PR Checklist

### DIFF
--- a/src/metric-engine/src/engine.rs
+++ b/src/metric-engine/src/engine.rs
@@ -164,7 +164,16 @@ impl RegionEngine for MetricEngine {
             }
             RegionRequest::Flush(req) => self.inner.flush_region(region_id, req).await,
             RegionRequest::Delete(_) | RegionRequest::Truncate(_) => {
-                UnsupportedRegionRequestSnafu { request }.fail()
+                if self.inner.is_physical_region(region_id) {
+                    self.inner
+                        .mito
+                        .handle_request(region_id, request)
+                        .await
+                        .context(error::MitoFlushOperationSnafu)
+                        .map(|response| response.affected_rows)
+                } else {
+                    UnsupportedRegionRequestSnafu { request }.fail()
+                }
             }
             RegionRequest::Catchup(req) => self.inner.catchup_region(region_id, req).await,
         };
--- a/tests/cases/standalone/common/basic.sql
+++ b/tests/cases/standalone/common/basic.sql
@@ -47,6 +47,15 @@ INSERT INTO t1 VALUES ('host1',0, 0), ('host2', 1, 1,);
 
 SELECT * from t1;
 
+-- issue #4649 should fail (do not support delete from logical table for now)
+delete from t1;
+
+-- issue #4649 should succeed
+delete from phy;
+
+-- insert the value again
+INSERT INTO t1 VALUES ('host1',0, 0), ('host2', 1, 1,);
+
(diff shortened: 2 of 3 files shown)


src/metric-engine/src/engine.rs
```diff
@@ -164,7 +164,16 @@ impl RegionEngine for MetricEngine {
             }
             RegionRequest::Flush(req) => self.inner.flush_region(region_id, req).await,
             RegionRequest::Delete(_) | RegionRequest::Truncate(_) => {
```
evenyag, line 166: Truncate should be more complicated. We may need to truncate the meta region and clear all memory states related to the meta region.

yihong0618: OK will do this

src/metric-engine/src/engine.rs
```diff
@@ -163,8 +163,18 @@ impl RegionEngine for MetricEngine {
...
-            RegionRequest::Delete(_) | RegionRequest::Truncate(_) => {
-                UnsupportedRegionRequestSnafu { request }.fail()
+            RegionRequest::Truncate(_) => UnsupportedRegionRequestSnafu { request }.fail(),
+            RegionRequest::Delete(_) => {
+                if self.inner.is_physical_region(region_id) {
+                    self.inner
+                        .mito
+                        .handle_request(region_id, request)
+                        .await
+                        .context(error::MitoFlushOperationSnafu)
```
waynexia, line 173: This should be something like `MitoDeleteOperationSnafu`. as well as line 159

waynexia (approved):
Rest LGTM, thanks @yihong0618

src/metric-engine/src/engine.rs
yihong0618: Done

tests/cases/standalone/common/basic.sql
```diff
@@ -47,6 +47,15 @@ INSERT INTO t1 VALUES ('host1',0, 0), ('host2', 1, 1,);
 
 SELECT * from t1;
 
+-- issue #4649 should fail (do not support delete from logical table for now)
+delete from t1;
+
+-- issue #4649 should succeed
+delete from phy;
```
evenyag, line 54: Do we need to select the table again?
```suggestion
delete from phy;

SELECT * from t1;
```

evenyag (approved):

tests/cases/standalone/common/basic.sql
yihong0618: Yes t1 affects by the phy

evenyag: We can select t1 again to verify it is empty.

yihong0618: FYI: there's another select t1 below in the sql file
and I can just drop the insert

killme2008:
Good job, thank you!

PR #5477, opened 2025-02-06
feat: add snapshot seqs field to query context

discord9:

~~proto definition:
https://github.com/GreptimeTeam/greptime-proto/pull/214~~

add snapshot seqs field to query context, and make query context get translate to `ScanRequest` 's sequence fields for snapshot read

### Details
- add snapshot_seqs field to QueryContext, which should be filled by something similar to  `SqlQueryInterceptor`
- make region server get region last seq num when get table provider to rewrite logical plan from `QueryContext` and use it for `ScanRequest`
- add a variant for `RegionRequest` which is `Sequences` for query regions' last seq num
- add `get_last_seq_num` for `RegionEngine` trait to get last seq num
- plugin to actually fill query context with sequences snapshot is in next PR

## PR Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced snapshot sequence management to enhance query consistency and reliability.
  - Added functionality to retrieve the latest sequence numbers, improving tracking of data regions.
  - Enhanced data retrieval by integrating optional session context into scan operations.

- **Chores**
  - Updated backend dependencies to boost overall stability and performance.

### DIFF
--- a/src/mito2/src/engine.rs
+++ b/src/mito2/src/engine.rs
@@ -543,6 +543,11 @@ impl RegionEngine for MitoEngine {
             .with_label_values(&[request.request_type()])
             .start_timer();
 
+        // Get sequences can happen outside region worker.
+        if let RegionRequest::Sequences(seqs) = request {
+            return self.get_region_sequences(seqs).await;
+        }
+
         self.inner
             .handle_request(region_id, request)
             .await
--- a/src/mito2/src/engine.rs
+++ b/src/mito2/src/engine.rs
@@ -539,11 +539,6 @@ impl RegionEngine for MitoEngine {
             .with_label_values(&[request.request_type()])
             .start_timer();
 
-        // Get sequences can happen outside region worker.
-        if let RegionRequest::Sequences(seqs) = request {
-            return self.get_region_sequences(seqs).await;
-        }
-
         self.inner
             .handle_request(region_id, request)
             .await
--- a/src/mito2/src/engine.rs
+++ b/src/mito2/src/engine.rs
@@ -430,12 +430,8 @@ impl EngineInner {
             .workers
             .get_region(region_id)
             .context(RegionNotFoundSnafu { region_id })?;
-        let version = region.version();
-        let seq = version
-            .memtables
-            .list_memtables()
-            .last()
-            .map(|m| m.stats().max_sequence());
+        let version_ctrl = &region.version_control;
+        let seq = Some(version_ctrl.committed_sequence());
         Ok(seq)
     }
 
--- a/src/mito2/src/engine.rs
+++ b/src/mito2/src/engine.rs
@@ -82,7 +82,7 @@ use store_api::region_engine::{
     SetRegionRoleStateResponse, SettableRegionRoleState,
 };
 use store_api::region_request::{AffectedRows, RegionOpenRequest, RegionRequest};
-use store_api::storage::{RegionId, ScanRequest};
+use store_api::storage::{RegionId, ScanRequest, SequenceNumber};
 use tokio::sync::{oneshot, Semaphore};
 
 use crate::cache::CacheStrategy;
@@ -424,6 +424,21 @@ impl EngineInner {
         receiver.await.context(RecvSnafu)?
     }
 
+    fn get_last_seq_num(&self, region_id: RegionId) -> Result<Option<SequenceNumber>> {
+        // Reading a region doesn't need to go through the region worker thread.
+        let region = self
+            .workers
+            .get_region(region_id)
+            .context(RegionNotFoundSnafu { region_id })?;
+        let version = region.version();
+        let seq = version
+            .memtables
+            .list_memtables()
+            .last()
+            .map(|m| m.stats().max_sequence());
+        Ok(seq)
+    }
+
     /// Handles the scan `request` and returns a [ScanRegion].
     fn scan_region(&self, region_id: RegionId, request: ScanRequest) -> Result<ScanRegion> {
         let query_start = Instant::now();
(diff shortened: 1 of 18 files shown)


src/mito2/src/engine.rs
```diff
@@ -424,6 +424,21 @@ impl EngineInner {
...
+    fn get_last_seq_num(&self, region_id: RegionId) -> Result<Option<SequenceNumber>> {
+        // Reading a region doesn't need to go through the region worker thread.
+        let region = self
+            .workers
+            .get_region(region_id)
+            .context(RegionNotFoundSnafu { region_id })?;
+        let version = region.version();
+        let seq = version
+            .memtables
+            .list_memtables()
```
fengjiachun, line 436: Can `memtables` provide a `last_memtable` method? cc @evenyag

src/servers/src/hint_headers.rs
```diff
@@ -18,12 +18,13 @@ use tonic::metadata::MetadataMap;
...
 pub const HINTS_KEY: &str = "x-greptime-hints";
 
-pub const HINT_KEYS: [&str; 5] = [
+pub const HINT_KEYS: [&str; 6] = [
     "x-greptime-hint-auto_create_table",
     "x-greptime-hint-ttl",
     "x-greptime-hint-append_mode",
     "x-greptime-hint-merge_mode",
     "x-greptime-hint-physical_table",
+    "x-greptime-hint-snapshot_read",
```
fengjiachun, line 27: It seems unnecessary to add it to the hint keys.

src/mito2/src/engine.rs
evenyag: We should use the committed sequence in the version as the last sequence.

src/store-api/src/region_request.rs
```diff
@@ -60,6 +60,7 @@ pub enum RegionRequest {
     Compact(RegionCompactRequest),
     Truncate(RegionTruncateRequest),
     Catchup(RegionCatchupRequest),
+    Sequences(RegionSequencesRequest),
```
evenyag, line 63: I don't think we need to add the `Sequences` request to the `RegionRequest`. We can add a new API to the `RegionEngine` to acquire sequences. Also we don't have to define a proto for this.

src/store-api/src/region_engine.rs
```diff
@@ -414,6 +418,34 @@ pub trait RegionEngine: Send + Sync {
         request: RegionRequest,
     ) -> Result<RegionResponse, BoxedError>;
 
+    /// Returns the last sequence number of the region.
+    #[allow(unused)]
+    async fn get_last_seq_num(
+        &self,
+        region_id: RegionId,
+    ) -> Result<Option<SequenceNumber>, BoxedError> {
+        Ok(None)
```
zhongzc, line 427: leave a todo?

src/store-api/src/region_engine.rs
```diff
@@ -414,6 +418,34 @@ pub trait RegionEngine: Send + Sync {
...
+    }
+
+    async fn get_region_sequences(
+        &self,
+        seqs: RegionSequencesRequest,
+    ) -> Result<RegionResponse, BoxedError> {
+        let mut results = HashMap::new();
```
zhongzc, line 434: with_capacity

src/session/src/context.rs
```diff
@@ -43,6 +43,10 @@ const CURSOR_COUNT_WARNING_LIMIT: usize = 10;
 #[builder(build_fn(skip))]
 pub struct QueryContext {
     current_catalog: String,
+    /// mapping of RegionId to SequenceNumber, for snapshot read, meaning that the read should only
+    /// container data that was committed before(and include) the given sequence number
+    /// this field will only be filled if extensions contains a pair of "snapshot_read" and "true"
+    snapshot_seqs: Arc<RwLock<Option<HashMap<u64, u64>>>>,
```
zhongzc, line 49: Do we have a suitable type alias to replace these two u64s?

discord9: > Do we have a suitable type alias to replace these two u64s?

both `RegionId` and `SequenceNumber` is defined in downstream crate(`store-api`), so no alias for them in `session` crate

src/mito2/src/region/version.rs
```diff
@@ -78,6 +78,14 @@ impl VersionControl {
         data.last_entry_id = entry_id;
     }
 
+    /// Sequence number of last committed data.
+    pub(crate) fn committed_sequence(&self) -> SequenceNumber {
+        self.data
+            .read()
+            .expect("lock is already held by the current thread")
```
evenyag, line 85: Maybe `unwrap()` is fine. Other places use `unwrap()`.

fengjiachun (approved):
LGTM

src/session/src/context.rs
evenyag, line 49: Do we need to use `Option`? We can use an empty map if there are no seqs.

fengjiachun:
@zhongzc PTAL

src/store-api/src/region_engine.rs
```diff
@@ -436,6 +438,27 @@ pub trait RegionEngine: Send + Sync {
         request: RegionRequest,
     ) -> Result<RegionResponse, BoxedError>;
 
+    /// Returns the last sequence number of the region.
+    #[allow(unused)]
```
evenyag, line 442: Why do we need `allow(unused)` here?

src/session/src/context.rs
```diff
@@ -324,6 +342,19 @@ impl QueryContext {
...
     }
+
+    pub fn snapshots(&self) -> Option<HashMap<u64, u64>> {
+        self.snapshot_seqs.read().expect("lock poisoned").clone()
+    }
+
+    pub fn get_snapshot(&self, region_id: u64) -> Option<u64> {
+        self.snapshot_seqs
+            .read()
+            .expect("lock poisoned")
```
evenyag, line 353: Maybe unwrap.
```suggestion
        self.snapshot_seqs.read().unwrap().clone()
    }

    pub fn get_snapshot(&self, region_id: u64) -> Option<u64> {
        self.snapshot_seqs
            .read()
            .unwrap()
```

src/session/src/context.rs
discord9: > Do we need to use `Option`? We can use an empty map if there are no seqs.

previous using Option is to differ this three case apart:
- not need snapshot: None
- need snapshot but haven't get all snapshot ids: Some(empty)
- need snapshot and snapshots' id are available: Some(regions to ids)

I'm not sure now if first two case be distinguishable is useful anymore ,perhaps not?

PR #5479, opened 2025-02-06
refactor: otlp logs insertion

shuiyisong:

This PR refactors the write path of OTLP Logs insertion. The main change is to eliminate the following three schemas and use one single `SchemaInfo`.
```Rust
let mut extra_resource_schema = SchemaInfo::default();
let mut extra_log_schema = SchemaInfo::default();
let mut extra_scope_schema = SchemaInfo::default();
```

The OTLP Logs request itself has 3 ordered layers: resource, scope, and log, and each has its own attributes. The attributes are stored using JSON datatype. However, we enable users to set `select keys` to find values in the attributes and uplift the values to the top level, which is an independent column, to speed up queries.

The original implementation uses three different `SchemaInfo` to store each layer's schema change and merge them in the end. If the select key appears in the multiple layers, say both resource and scope, it actually overrides the former one. Therefore we can use one single `SchemaInfo` to record the global extra schema information and perform an override in place if needed.

The second change is to build and return a `Row` structure in the `parse_log` function instead of returning an intermediate  `ParseInfo`. In the `parse_log` we actually have all we need to build a `Row`.

## PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6608,17 +6608,18 @@ dependencies = [
 [[package]]
 name = "meter-core"
 version = "0.1.0"
-source = "git+https://github.com/GreptimeTeam/greptime-meter.git?rev=a10facb353b41460eeb98578868ebf19c2084fac#a10facb353b41460eeb98578868ebf19c2084fac"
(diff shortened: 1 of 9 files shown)


shuiyisong:
> Looks like it's a breaking change.

This PR only simplifies the transformation logic from the  OTLP request to the RowInsertRequest. It doesn't change the final output content, thus it's not a breaking change.

Issue #5333, opened 2025-01-10
Move `api/v1/event/pipeline` to `api/v1/pipeline`


waynexia:
### What type of enhancement is this?

API improvement

### What does the enhancement do?

Move the pipeline to the first level in the API. Reflecting:
- It's not only for logs
- Event is not an implementation level thing

PR #5480, opened 2025-02-06
feat: move pipelines to the first-class endpoint

waynexia:

Closes https://github.com/GreptimeTeam/greptimedb/issues/5333

Endpoint changes:
- `v1/events/logs` -> `v1/ingest`. This endpoint will become our universal HTTP ingest endpoint. More formats and scenarios support are on the way
- `v1/events/pipelines/dryrun` -> `v1/pipelines/_dryrun` for escape user pipeline named "dryrun"
- `v1/events/pipelines/<pipeline_name>` -> `v1/pipelines/<pipeline_name>`

All old APIs will remain for a while, so this PR is not breaking.

## PR Checklist

### DIFF
--- a/src/servers/src/http.rs
+++ b/src/servers/src/http.rs
@@ -605,9 +605,16 @@ impl HttpServerBuilder {
             log_validator: validator,
             ingest_interceptor,
         };
+
         let router = self.router.nest(
+            &format!("/{HTTP_API_VERSION}"),
+            HttpServer::route_pipelines(log_state.clone()),
+        );
+        // deprecated since v0.11.0. Use `/logs` and `/pipelines` instead.
+        let router = router.nest(
             &format!("/{HTTP_API_VERSION}/events"),
-            HttpServer::route_log(log_state.clone()),
+            #[allow(deprecated)]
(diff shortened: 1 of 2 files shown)


src/servers/src/http.rs
```diff
@@ -912,6 +920,25 @@ impl HttpServer {
             .with_state(log_state)
     }
 
+    fn route_pipelines<S>(log_state: LogState) -> Router<S> {
+        Router::new()
+            .route("/pipelines/ingest", routing::post(event::log_ingester))
```
killme2008, line 925: We need to ensure the pipeline name can't be `ingest`.

But I think changing the `ingest` to `_ingest` would be better.

killme2008 (approved):
LGTM

src/servers/src/http.rs
shuiyisong: Same as `dryrun` below.

waynexia: Changed `dryrun` to `_dryrun`. And `pipelines/ingest` to `ingest` according to the discussion with @sunng87

shuiyisong (approved):

PR #5481, opened 2025-02-06
fix: lose decimal precision when using decimal type as tag

evenyag:

When using decimal types as tags, querying the table will get an error because `replicate()` the tag vector will lose the precision.

## PR Checklist


fengjiachun (approved):
LGTM

WenyXu (approved):

Issue #5493, opened 2025-02-08
Support setting tcp keepalive duration


evenyag:
### What problem does the new feature solve?

Sometimes we may want to configure the TCP keepalive idle for a connection in the Postgres or MySQL handler of GreptimeDB

https://www.postgresql.org/docs/current/runtime-config-connection.html#RUNTIME-CONFIG-TCP-SETTINGS

It may be useful if we run the fronted after an LB so we can use the keepalive probe to keep the connection.

What's more, we may also find a way to configure the keepalive idle for the HTTP server since we can't set that in axum after https://github.com/GreptimeTeam/greptimedb/pull/5417

### What does the feature do?

Adds server side configurations for keepalive related options

### Implementation challenges

We have the following different handlers to support
- MySQL https://github.com/GreptimeTeam/greptimedb/issues/5493
- PostgreSQL https://github.com/GreptimeTeam/greptimedb/issues/5493
- HTTP

It isn't easy to test in an integration test.

PR #5496, opened 2025-02-08
feat: support server-side keep-alive for mysql and pg protocols

waynexia:

https://github.com/GreptimeTeam/greptimedb/issues/5493

Add a new option `keep_alive_secs` to MySQL and PostgreSQL protocol for server-side keep-alive. Similar to `tcp_keepalives_idle` and `tcp_keepalives_interval` in PostgreSQL server.

Verified by `tcpdump`:

image

## PR Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -186,7 +186,6 @@ shadow-rs = "0.38"
 similar-asserts = "1.6.0"
 smallvec = { version = "1", features = ["serde"] }
 snafu = "0.8"
-socket2 = "0.5"
 sysinfo = "0.30"
 # on branch v0.52.x
 sqlparser = { git = "https://github.com/GreptimeTeam/sqlparser-rs.git", rev = "71dd86058d2af97b9925093d40c4e03360403170", features = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -186,6 +186,7 @@ shadow-rs = "0.38"
 similar-asserts = "1.6.0"
 smallvec = { version = "1", features = ["serde"] }
 snafu = "0.8"
+socket2 = "0.5"
 sysinfo = "0.30"
 # on branch v0.52.x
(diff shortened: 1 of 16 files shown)


config/config.md
```diff
@@ -234,6 +234,7 @@
 | `mysql.enable` | Bool | `true` | Whether to enable. |
 | `mysql.addr` | String | `127.0.0.1:4002` | The addr to bind the MySQL server. |
 | `mysql.runtime_size` | Integer | `2` | The number of server worker threads. |
+| `mysql.keep_alive_secs` | Integer | `0` | Server-side keep-alive time in seconds.<br/>Set to 0 (default) to disable. |
```
evenyag, line 237: Should we use human time to parse the duration?

evenyag (approved):

Cargo.toml
```diff
@@ -186,6 +186,7 @@ shadow-rs = "0.38"
 similar-asserts = "1.6.0"
 smallvec = { version = "1", features = ["serde"] }
 snafu = "0.8"
+socket2 = "0.5"
```
sunng87, line 189: I suggest to put this to server/Cargo.toml if it's only used in server module. Maintaining workspace dependencies has been a pain for downstream projects.

src/servers/src/server.rs
```diff
@@ -157,6 +158,15 @@ impl AcceptTask {
                 let addr = listener.local_addr()?;
                 info!("{name} server started at {addr}");
 
+                // set keep-alive
+                if keep_alive_secs > 0 {
+                    let socket_ref = socket2::SockRef::from(&listener);
```
sunng87, line 163: is it to set on listener or on socket returned from accept?

waynexia: It looks like this can be inherited by the client socket https://stackoverflow.com/questions/43441913/so-keepalive-set-on-the-server-socket-or-on-a-per-client-basis

sunng87 (approved):

PR #5498, opened 2025-02-08
refactor: change traversal order during index construction

zhongzc:

Previously, the default was to build an index for the tag column, hence the traversal method involved iterating through all tag columns.

Now, this assumption has been invalidated as the process has shifted to building indexes for specified columns. Therefore, it is no longer appropriate to continue traversing all tag columns, especially in scenarios where the number of columns is vast but only a few columns require index construction.

## PR Checklist

### DIFF
--- a/src/mito2/src/read.rs
+++ b/src/mito2/src/read.rs
@@ -626,8 +626,11 @@ impl Batch {
             );
         }
 
-        let idx = self.fields_idx.as_ref().unwrap().get(&column_id)?;
-        Some(&self.fields[*idx])
+        self.fields_idx
+            .as_ref()
+            .unwrap()
+            .get(&column_id)
+            .map(|&idx| &self.fields[idx])
     }
 }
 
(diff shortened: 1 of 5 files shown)


WenyXu (approved):
LGTM

src/mito2/src/read.rs
```diff
@@ -588,6 +591,44 @@ impl Batch {
...
+                self.fields
+                    .iter()
+                    .enumerate()
+                    .map(|(i, c)| (c.column_id, i))
+                    .collect(),
+            );
+        }
+
+        let idx = self.fields_idx.as_ref().unwrap().get(&column_id)?;
+        Some(&self.fields[*idx])
```
fengjiachun, line 630: Use map()?
```suggestion
        self.fields_idx
            .as_ref()
            .unwrap()
            .get(&column_id)
            .map(|&idx| &self.fields[idx])
```

fengjiachun (approved):

PR #5422, opened 2025-01-22
fix: flush table panic when table has interval column close #3235

yihong0618:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

early raise error if create table with interval field.

close: #3235 as discuss.

---

**Note**: this change will break old some create table as discuss

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [ ] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #5506, opened 2025-02-10
feat remove datetime type

fengjiachun:

## PR Checklist


tests/cases/standalone/common/create/create_type_alias.result
```diff
@@ -36,7 +36,7 @@ SHOW CREATE TABLE data_types;
 |            |   "b" BOOLEAN NULL,                                        |
 |            |   "vb" VARBINARY NULL,                                     |
 |            |   "dt" DATE NULL,                                          |
-|            |   "dtt" DATETIME NULL,                                     |
+|            |   "dtt" TIMESTAMP(6) NULL,                                 |
```
evenyag, line 39: What if we add a datetime column? Maybe we need a test.

src/common/grpc-expr/src/util.rs
```diff
@@ -151,6 +151,15 @@ pub fn build_create_table_expr(
 
         let column_type = infer_column_datatype(datatype, datatype_extension)?;
 
+        ensure!(
```
evenyag, line 154: https://github.com/GreptimeTeam/greptimedb/pull/5422 also forbids the interval type in another place.

Where is the best place to do this validation? We may need to ensure all create table operations use the same validation rule.

evenyag: Looks like we have many different helper method to build a create table expr......

tests/cases/standalone/common/create/create_type_alias.result
fengjiachun: I'll add a test case later.

fengjiachun: [ff3ba74](https://github.com/GreptimeTeam/greptimedb/pull/5506/commits/ff3ba7407446d385cf51d6be128a9b76db06518f)

fengjiachun:
@evenyag PTAL

src/operator/src/expr_helper.rs
```diff
@@ -295,9 +298,52 @@ pub fn validate_create_expr(create: &CreateTableExpr) -> Result<()> {
         }
     }
 
+    if auto_schema {
+        // `DateTime` type is not supported in auto schema mode.
```
killme2008, line 302: Can't we automatically convert it to `Timestamp(6)`?

fengjiachun: No, because it requires transferring data from `data_time_values` to `timestamp_values`, which is inefficient and inelegant.

src/operator/src/expr_helper.rs
```diff
@@ -287,9 +290,52 @@ pub fn validate_create_expr(create: &CreateTableExpr) -> Result<()> {
...
```
evenyag, line 294: If auto schema is false, we already converted the datetime type to timestamp(3) before this function. Maybe we can also return an error?

fengjiachun: If the auto schema is false, it means the table was manually created. In this case, DateTime is an alias for Timestamp(6), so the actual type will be Timestamp(6). If you specify this column as `DateTime` through the gRPC protocol during writing, a type check error will occur (this check has always existed).

v0y4g3r (approved):
LGTM. Maybe we also need to remove [`Value::DateTime`](https://github.com/GreptimeTeam/greptimedb/blob/35b635f6399dc0ee33bc61d7b22d6dd76b3b5626/src/datatypes/src/value.rs#L78) but may not be included in this PR.

evenyag (approved):

PR #5523, opened 2025-02-13
chore: use the same version of chrono-tz

yihong0618:

for now there are three place of `chrono-tz` see the diff

its better for change all of them the same version to min the size of cargo lock.

## PR Checklist


shuiyisong:
If the crate is used across multiple packages, we can set it in the workspace's `Cargo.toml` and use `chrono-tz.workspace = true` for unified version control.

yihong0618:
> If the crate is used across multiple packages, we can set it in the workspace's `Cargo.toml` and use `chrono-tz.workspace = true` for unified version control.

will change to that way, thanks

yihong0618:
> If the crate is used across multiple packages, we can set it in the workspace's `Cargo.toml` and use `chrono-tz.workspace = true` for unified version control.

there are some other package use different version and not workplace.true way, I do not quick sure they are by design and code, so do not change them in this patch

shuiyisong (approved):

waynexia (approved):

PR #5526, opened 2025-02-13
perf: optimize table alteration speed in metric engine

### LINKED ISSUE #5492: Optimize table creation speed in Metric Engine

User experience

Previously, we merged multiple table creation requests at a higher level, but in the Metric Engine implementation, these requests were still executed sequentially. This change aims to enable batch execution of table creation requests within the Metric Engine to improve efficiency.

### Implementation challenges

_No response_

WenyXu:

close #5492

This PR optimizes the handling of ALTER TABLE requests by processing them in batches, reducing the number of calls to Mito engine. Additionally, it caches necessary states in memory to improve efficiency.

## PR Checklist


src/metric-engine/src/data_region.rs
```diff
@@ -65,33 +61,30 @@ impl DataRegion {
...
-                }
-                Err(e) => {
-                    return Err(e).context(MitoWriteOperationSnafu)?;
-                }
-            }
-        }
+        let _ = self
+            .mito
+            .handle_request(region_id, request)
+            .await
```
fengjiachun, line 84: Question: why are we removing the retry logic?

src/metric-engine/src/engine/alter/extract_new_columns.rs
```diff
@@ -0,0 +1,51 @@
...
+use store_api::storage::{ColumnId, RegionId};
+
+use crate::error::Result;
+
+/// Extract new columns from the create requests.
+///
+/// # Panics
+///
+/// This function will panic if the alter kind is not `AddColumns`.
+pub fn extract_new_columns<'a>(
```
fengjiachun, line 28: What about using a hashemap instead of hashset and vec, then just let the hashmap as the return value and remove the new_column_names new_columns args?

fengjiachun (approved):

src/metric-engine/src/data_region.rs
WenyXu: There are two main reasons. First, retries might be incorrect if the schema version changes, as we need to filter out potential duplicate columns. Second, procedure locks ensure that only one exclusive operation is allowed at a time.

src/metric-engine/src/engine/alter.rs
```diff
@@ -12,148 +12,157 @@
...
+    /// Alter multiple logical regions on the same physical region.
+    pub async fn alter_logical_regions(
         &self,
-        logical_region_id: RegionId,
-        request: RegionAlterRequest,
-    ) -> Result<RegionId> {
+        requests: Vec<(RegionId, RegionAlterRequest)>,
+        extension_return_value: &mut HashMap<String, Vec<u8>>,
+    ) -> Result<AffectedRows> {
+        validate_alter_region_requests(&requests)?;
```
discord9, line 69: if alter region in batches only support add columns should we rename some methods to emphasize this?

discord9:
Rest LGTM

Issue #5463, opened 2025-01-27
Some `order by` SQL didn't use window sort


evenyag:
### What type of bug is this?

Performance issue

### What subsystems are affected?

Distributed Cluster, Standalone mode

### Minimal reproduce step

Execute the following SQL:

```sql
CREATE TABLE IF NOT EXISTS `orderby_test` (
  `file` STRING NULL,
  `host` STRING NULL,
  `message` STRING NULL FULLTEXT WITH(analyzer = 'English', case_sensitive = 'false'),
  `source_type` STRING NULL,
  `timestamp` STRING NULL,
  `greptime_timestamp` TIMESTAMP(9) NOT NULL,
  TIME INDEX (`greptime_timestamp`)
)
ENGINE=mito
WITH(
  append_mode = 'true'
);

explain analyze select * from orderby_test where greptime_timestamp >= now() - '10m'::interval and greptime_timestamp < now() order by greptime_timestamp desc limit 10;

explain analyze select greptime_timestamp as timestamp, message from orderby_test order by greptime_timestamp desc limit 10;

```

### What did you expect to see?

We expect that we should use window sort like this SQL
```sql
explain analyze select * from orderby_test order by greptime_timestamp desc limit 10;
+-------+------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| stage | node | plan                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
+-------+------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|     0 |    0 |  MergeScanExec: peers=[4440996184064(1034, 0), ] metrics=[output_rows: 0, greptime_exec_read_cost: 0, first_consume_time: 2662581, finish_time: 2665042, ready_time: 2489376, ]
                                                                                                                                                                                                                                                                                                                                                                |
|     1 |    0 |  GlobalLimitExec: skip=0, fetch=10 metrics=[output_rows: 0, elapsed_compute: 1, ]
  WindowedSortExec: expr=greptime_timestamp@5 DESC num_ranges=1 fetch=10 metrics=[output_rows: 0, elapsed_compute: 1, ]
[...]

PR #5527, opened 2025-02-13
feat: window sort supports where on fields and time index

evenyag:

https://github.com/GreptimeTeam/greptimedb/issues/5463

This PR fixes some cases that can't use window sort in #5463. The repartition plan between `FilterExec` and `RegionScanExec` prevents the window sort. We match this pattern and remove the repartition plan before checking region partitions so we can apply window sort optimization.

Using window sort usually should have more benefits than adding repartition before the filter.

We need to check there is no regression in performance.

## PR Checklist


waynexia (approved):

discord9 (approved):
LGTM, any test on possible regression?

PR #5529, opened 2025-02-13
chore: format all toml files

yihong0618:

this patch follow this comment https://github.com/GreptimeTeam/greptimedb/pull/5523#issuecomment-2655659231
from @shuiyisong all use workplace.true to `DRY`

about this patch:

- I search all the toml using regex but there are too many of them
- so I try to search if some static check and luck to find https://github.com/mainmatter/cargo-autoinherit
- fix most of them but the sort is a mess
- and found https://github.com/DevinR528/cargo-sort 
- install the sort need the main `cargo install --git=https://github.com/DevinR528/cargo-sort --branch=main cargo-sort`

This patch not fix them all, can check the output below, I am not sure these dependence can be merged like https://github.com/GreptimeTeam/greptimedb/pull/5523

```cli
`async-compression` won't be auto-inherited because there are multiple sources for it:
  - version: ^0.3
  - version: ^0.4
`chrono-tz` won't be auto-inherited because there are multiple sources for it:
  - version: ^0.9.0
  - version: ^0.8
`criterion` won't be auto-inherited because there are multiple sources for it:
  - version: ^0.5
  - version: ^0.4
`hyper` won't be auto-inherited because there are multiple sources for it:
  - version: ^1.1
  - version: ^0.14
```

## PR Checklist


shuiyisong:
> install the sort need the main cargo install --git=https://github.com/DevinR528/cargo-sort --branch=main cargo-sort

You can use `make fmt-toml` to sort the cargo file first and then double-check the change. We use `taplo` to do the sorting with some pre-defined configs.

yihong0618:
> > install the sort need the main cargo install --git=https://github.com/DevinR528/cargo-sort --branch=main cargo-sort
> 
> You can use `make fmt-toml` to sort the cargo file first and then double-check the change. We use `taplo` to do the sorting with some pre-defined configs.

yes used it too

sunng87:
I would suggest only lifting common dependencies of multiple modules to workspace level. For our close-sourced downstream projects, we use a trick to manage both open-source code and close-source code with the workspace level cargo.toml. It requires syncing the dependencies section between cargo.tomls, which is a pain.

sunng87:
I just add https://github.com/GreptimeTeam/greptimedb/pull/5536 to possibly improve the sync experience.

killme2008 (approved):
LGTM

PR #5543, opened 2025-02-14
fix: window sort support alias time index

evenyag:

closes https://github.com/GreptimeTeam/greptimedb/issues/5463

This PR collects aliases of the time index column in the `ProjectionExec`, so both distributed query and ordering by alias can use window sort.

It also uses the commutativity of `alias.expr` as the commutativity of `Alias` so we can push the plan to the datanode.

## PR Checklist


waynexia (approved):

discord9 (approved):
LGTM

Issue #5525, opened 2025-02-13
[Sprint] Tracking Issue: VictoriaMetrics to GreptimeDB Migration


WenyXu:
### What type of enhancement is this?

User experience

### What does the enhancement do?

This tracking issue documents the challenges encountered during migrating from VictoriaMetrics to GreptimeDB. 

Unsupported 
- [ ] topk (`topk(5, irate(redis_commands_total{instance=~"$instance"} [1m]))`)
- [ ] sort_desc (`sort_desc(sum(certmanager_certificate_expiration_timestamp_seconds{exported_namespace=~"$Namespace", name=~"$Certificate"} - time()) BY (name,exported_namespace) )`)
- [ ] incorrect alias (`sum(rate(nginx_ingress_controller_requests{status!~"[4-5].*", controller_class=~"$controller_class", ingress=~"$ingress", namespace=~"$namespace", controller_pod=~"$pod"}[$__interval])) / 
(sum(rate(nginx_ingress_controller_requests{ controller_class=~"$controller_class", ingress=~"$ingress", namespace=~"$namespace", controller_pod=~"$pod"}[$__interval])) - 
(sum(rate(nginx_ingress_controller_requests{status=~"404|499", controller_class=~"$controller_class", ingress=~"$ingress", namespace=~"$namespace", controller_pod=~"$pod"}[$__interval])) or vector(0)))`)
- [ ] incorrect function signature (`round(sum(irate(nginx_ingress_controller_requests{controller_pod=~"$controller",controller_class=~"$controller_class",namespace=~"$namespace"}[2m])), 0.001)`)
- [ ] `quantile(0.5,
    rate(process_cpu_seconds_total{job=~\"kubernetes-endpoints\", instance=~\".*\"}[1m0s])
    /
    process_cpu_cores_available{job=~\"kubernetes-endpoints\", instance=~\".*\"}
)`
- [ ] le column not found (`sum(histogram_quantile(0.99, sum(rate(vm_series_read_per_query_bucket{job=~"$job_select", instance=~"$instance"}[$__rate_interval])) by (instance, vmrange)))`)
- [ ] sub query (`sort_desc(sum(irate(container_network_receive_bytes_total{cluster="$cluster",namespace=~".+"}[$interval:$resolution])) by (namespace))`)
- [ ] #5521 
- [ ] #5520 
- [ ] #5519

UX
- [ ] match[] parameter is required (series query must include a metric matcher)
[...]

PR #5556, opened 2025-02-17
fix(promql-planner): update ctx field columns of OR operator

WenyXu:

#5525 

This PR ensures that the field column in the context correctly corresponds to the field column of the left input.

The `self.ctx` is updated after invoking `self.prom_expr_to_plan`. 
```rust 
        ...
        let right_input = self.prom_expr_to_plan(rhs, session_state).await?;
        let right_field_columns = self.ctx.field_columns.clone();
        let right_time_index_column = self.ctx.time_index_column.clone();
        let mut right_table_ref = self
            .table_ref()
            .unwrap_or_else(|_| TableReference::bare(""));
        let right_context = self.ctx.clone();
        ...
        if Self::is_token_a_set_op(*op) {
            return self.set_op_on_non_field_columns(
        ...
```

Therefore, we need to explicitly update the field column in `or_operator`.
```rust
        fn or_operator(
        ...
        self.ctx.field_columns = vec![left_field_col.to_string()];
```

## PR Checklist


evenyag (approved):

fengjiachun (approved):
LGTM

PR #5558, opened 2025-02-17
feat: allow purging a given puffin file in staging area

zhongzc:

There were many dangling contents in the staging area before even though the sst file had been cleared. Now, when actively purging the sst file, clean up the staging area as well.

## PR Checklist


src/puffin/src/puffin_manager/stager/bounded_stager.rs
```diff
@@ -284,6 +306,15 @@ impl Stager for BoundedStager {
...
     }
+
+    async fn purge(&self, puffin_file_name: &str) -> Result<()> {
+        let keys = self.file_mapping.take_keys(puffin_file_name).await;
+        for key in keys {
+            let cache_key = Self::encode_cache_key(puffin_file_name, &key);
+            self.cache.invalidate(&cache_key).await;
+        }
+        Ok(())
+    }
```
evenyag, line 317: Maybe we can use `invalidate_entries_if` so we don't have to maintain the mapping.
https://docs.rs/moka/latest/moka/future/struct.Cache.html#method.invalidate_entries_if

src/puffin/src/puffin_manager/stager/bounded_stager.rs
```diff
@@ -337,6 +368,9 @@ impl BoundedStager {
     }
 
     /// Recovers the staging area by iterating through the staging directory.
+    ///
+    /// Note: It can't recover the mapping between puffin files and keys, so TTL
+    ///       is configured to purge the dangling files and directories.
```
evenyag, line 373: non-blocking: Now we hash the file name and key to get the cache key so we can't recover the relationship, we may consider a better way to compute the cache key.

zhongzc: It seems that there is no solution to reversibly encode the concatenation of two strings without introducing any assumptions?

evenyag: Yes, we have to use `/` as delimiter and use the key as path directly. The drawback of encoding the key is that we don't know which puffin file the cache entry belongs to.

src/puffin/src/puffin_manager/stager/bounded_stager.rs
```diff
@@ -284,6 +294,14 @@ impl Stager for BoundedStager {
 
         Ok(())
     }
+
+    async fn purge(&self, puffin_file_name: &str) -> Result<()> {
+        let file_name = puffin_file_name.to_string();
+        self.cache
+            .invalidate_entries_if(move |_k, v| v.puffin_file_name() == file_name)
+            .unwrap(); // SAFETY: `support_invalidation_closures` is enabled
```
evenyag, line 302: Can we call `self.cache.run_pending_tasks()` here? `invalidate_entries_if()` only sets the callback. We can execute the maintenance task here.

evenyag (approved):

fengjiachun (approved):
LGTM

PR #5560, opened 2025-02-18
fix(promql-planner): remove le tag in ctx

WenyXu:

#5525 

This PR removes the `le` tag in ctx after the planner returns a `HistogramFold` node to ensure consistency with the `HistogramFold` output schema, which also omits the le tag.

## PR Checklist

### DIFF
--- a/src/query/src/promql/planner.rs
+++ b/src/query/src/promql/planner.rs
@@ -634,7 +634,6 @@ impl PromPlanner {
         let mut func_exprs = self.create_function_expr(func, args.literals, session_state)?;
         func_exprs.insert(0, self.create_time_index_column_expr()?);
         func_exprs.extend_from_slice(&self.create_tag_column_exprs()?);
-
         LogicalPlanBuilder::from(input)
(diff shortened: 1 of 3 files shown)


killme2008 (approved):
LGTM

src/query/src/promql/planner.rs
```diff
@@ -1803,6 +1803,12 @@ impl PromPlanner {
                 fn_name: SPECIAL_HISTOGRAM_QUANTILE.to_string(),
             })?
             .clone();
+        // remove le column from tag columns
+        let tag_columns = std::mem::take(&mut self.ctx.tag_columns);
+        self.ctx.tag_columns = tag_columns
+            .into_iter()
+            .filter(|col| col != LE_COLUMN_NAME)
+            .collect();
```
evenyag, line 1811: Can we use retain?
https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.retain

evenyag (approved):

PR #5561, opened 2025-02-18
perf: close issue 4974 by do not delete columns when drop logical region about 100 times faster

### LINKED ISSUE #4974: Slow `DROP DATABASE` with large number of metric tables

User experience

When a database contains a large number of metric tables (thousands), the `DROP DATABASE` operation becomes excessively slow, taking several minutes to complete. This performance issue is due to the metric engine’s behavior of scanning the metadata region each time a table is dropped. The p50 latency for this metadata lookup is around 300ms, leading to significant delays when dropping numerous tables.

### Implementation challenges

_No response_

yihong0618:

After some search and learn the region server found maybe we can skip the drop columns in drop logical region
since all data can be drop in `phy` table I wonder.

because the function `get_all` or `get_all_with_prefix` is very expensive, cost almost all the time for the drop database

before this patch

image

after this patch
image

the generate data scripts

```py
import psycopg2
import time
from datetime import datetime, timedelta

# --- Configuration ---
DB_HOST = "localhost"
DB_PORT = "4003"
DB_USER = "greptime"
DB_NAME = "public"
SENSORS_PER_LOCATION = 4000
SIMULATION_DURATION_HOURS = 5
INSERT_BATCH_SIZE = 4000




def insert_data_now(simulation_start_time):
    conn = create_connection()
    try:
        # First create physical table
        with conn.cursor() as cur:
            cur.execute("CREATE DATABASE test;")
            cur.execute("use test;")
            physical_table_sql = "CREATE TABLE phy (ts timestamp time index, val double) engine=metric with (\"physical_metric_table\" = \"\");"
            physical2_table_sql = "CREATE TABLE phy2 (ts timestamp time index, val double) engine=metric with (\"physical_metric_table\" = \"\");"
            try:
                cur.execute(physical_table_sql)
                cur.execute(physical2_table_sql)
                conn.commit()
            except Exception as e:
                print(f"Physical table creation error: {e}")

        current_time = simulation_start_time

        with conn.cursor() as cur:
            index = 1
            while current_time <= simulation_start_time + timedelta(seconds=200):
                insert_sql = f"CREATE TABLE t{index} (ts timestamp time index, val double, host string primary key) engine = metric with (\"on_physical_table\" = \"phy\");"
                insert2_sql = f"CREATE TABLE b{index} (ts timestamp time index, val double, host string primary key) engine = metric with (\"on_physical_table\" = \"phy2\");"
                index += 1
                try:
                    cur.execute(insert_sql)
                    cur.execute(insert2_sql)
                    conn.commit()
                except Exception as e:
                    print(f"Table creation error: {e}")
                    continue
                print(f"Created table t{index-1} at {current_time}")
                print(f"Created table b{index-1} at {current_time}")
                current_time += timedelta(seconds=1)
    finally:
        if conn:
            conn.close() # close connect in the thread.

def create_connection():
    """Creates a database connection."""
    conn = psycopg2.connect(
        host=DB_HOST,
        port=DB_PORT,
        user=DB_USER,
        dbname=DB_NAME
    )
    return conn

def main():
    """Main function."""
    simulation_start_time = datetime.utcnow() - timedelta(seconds=200)

    insert_data_now(simulation_start_time=simulation_start_time)

if __name__ == "__main__":
    main()
```

close #4974

## PR Checklist

### DIFF
--- a/src/datanode/src/region_server.rs
+++ b/src/datanode/src/region_server.rs
@@ -1218,7 +1218,12 @@ mod tests {
         );
 
         let response = mock_region_server
-            .handle_request(region_id, RegionRequest::Drop(RegionDropRequest {}))
+            .handle_request(
+                region_id,
+                RegionRequest::Drop(RegionDropRequest {
+                    fast_drop_database_path: false,
+                }),
+            )
             .await
             .unwrap();
         assert_eq!(response.affected_rows, 0);
@@ -1310,7 +1315,12 @@ mod tests {
             .insert(region_id, RegionEngineWithStatus::Ready(engine.clone()));
 
         mock_region_server
-            .handle_request(region_id, RegionRequest::Drop(RegionDropRequest {}))
+            .handle_request(
+                region_id,
+                RegionRequest::Drop(RegionDropRequest {
+                    fast_drop_database_path: false,
+                }),
+            )
             .await
             .unwrap_err();
 
--- a/src/datanode/src/region_server.rs
+++ b/src/datanode/src/region_server.rs
@@ -1220,9 +1220,7 @@ mod tests {
         let response = mock_region_server
             .handle_request(
                 region_id,
-                RegionRequest::Drop(RegionDropRequest {
-                    force_drop_all_logical_tables: false,
-                }),
+                RegionRequest::Drop(RegionDropRequest { fast_path: false }),
             )
             .await
             .unwrap();
@@ -1317,9 +1315,7 @@ mod tests {
         mock_region_server
             .handle_request(
                 region_id,
-                RegionRequest::Drop(RegionDropRequest {
-                    force_drop_all_logical_tables: false,
-                }),
+                RegionRequest::Drop(RegionDropRequest { fast_path: false }),
             )
             .await
             .unwrap_err();
--- a/src/datanode/src/region_server.rs
+++ b/src/datanode/src/region_server.rs
@@ -1221,7 +1221,7 @@ mod tests {
             .handle_request(
                 region_id,
                 RegionRequest::Drop(RegionDropRequest {
-                    fast_drop_path: false,
+                    force_drop_all_logical_tables: false,
                 }),
             )
             .await
@@ -1318,7 +1318,7 @@ mod tests {
             .handle_request(
                 region_id,
                 RegionRequest::Drop(RegionDropRequest {
-                    fast_drop_path: false,
+                    force_drop_all_logical_tables: false,
                 }),
             )
             .await
--- a/src/datanode/src/region_server.rs
+++ b/src/datanode/src/region_server.rs
@@ -1221,7 +1221,7 @@ mod tests {
             .handle_request(
                 region_id,
                 RegionRequest::Drop(RegionDropRequest {
-                    fast_drop_database_path: false,
+                    fast_drop_path: false,
                 }),
             )
             .await
@@ -1318,7 +1318,7 @@ mod tests {
             .handle_request(
                 region_id,
                 RegionRequest::Drop(RegionDropRequest {
-                    fast_drop_database_path: false,
+                    fast_drop_path: false,
                 }),
             )
             .await
--- a/src/metric-engine/src/engine/drop.rs
+++ b/src/metric-engine/src/engine/drop.rs
@@ -33,6 +33,7 @@ impl MetricEngineInner {
         req: RegionDropRequest,
     ) -> Result<AffectedRows> {
         let data_region_id = utils::to_data_region_id(region_id);
+        let force_drop_all_logical_tables = req.force_drop_all_logical_tables;
 
         // enclose the guard in a block to prevent the guard from polluting the async context
         let (is_physical_region, is_physical_region_busy) = {
@@ -52,7 +53,7 @@ impl MetricEngineInner {
 
         if is_physical_region {
             // check if there is no logical region relates to this physical region
-            if is_physical_region_busy {
+            if is_physical_region_busy && !force_drop_all_logical_tables {
                 // reject if there is any present logical region
                 return Err(PhysicalRegionBusySnafu {
                     region_id: data_region_id,
@@ -60,9 +61,9 @@ impl MetricEngineInner {
                 .build());
             }
 
-            self.drop_physical_region(data_region_id).await
-        } else {
-            // cannot merge these two `if` otherwise the stupid type checker will complain
+            self.drop_physical_region(data_region_id, force_drop_all_logical_tables)
+                .await
+        } else if !force_drop_all_logical_tables {
             let metadata_region_id = self
                 .state
                 .read()
@@ -71,15 +72,21 @@ impl MetricEngineInner {
                 .get(&region_id)
                 .copied();
             if let Some(metadata_region_id) = metadata_region_id {
-                self.drop_logical_region(region_id, metadata_region_id, req.fast_drop_path)
+                self.drop_logical_region(region_id, metadata_region_id)
                     .await
             } else {
                 Err(LogicalRegionNotFoundSnafu { region_id }.build())
             }
+        } else {
+            Ok(0)
         }
     }
 
-    async fn drop_physical_region(&self, region_id: RegionId) -> Result<AffectedRows> {
+    async fn drop_physical_region(
+        &self,
+        region_id: RegionId,
+        force_drop_all_logical_tables: bool,
+    ) -> Result<AffectedRows> {
         let data_region_id = utils::to_data_region_id(region_id);
         let metadata_region_id = utils::to_metadata_region_id(region_id);
 
@@ -90,7 +97,7 @@ impl MetricEngineInner {
             .handle_request(
                 data_region_id,
                 RegionRequest::Drop(RegionDropRequest {
-                    fast_drop_path: false,
+                    force_drop_all_logical_tables,
                 }),
             )
             .await
@@ -99,7 +106,7 @@ impl MetricEngineInner {
             .handle_request(
                 metadata_region_id,
                 RegionRequest::Drop(RegionDropRequest {
-                    fast_drop_path: false,
+                    force_drop_all_logical_tables,
                 }),
             )
             .await
@@ -120,11 +127,10 @@ impl MetricEngineInner {
         &self,
         logical_region_id: RegionId,
         physical_region_id: RegionId,
-        fast_drop_path: bool,
     ) -> Result<AffectedRows> {
         // Update metadata
         self.metadata_region
-            .remove_logical_region(physical_region_id, logical_region_id, fast_drop_path)
+            .remove_logical_region(physical_region_id, logical_region_id)
             .await?;
 
         // Update engine state
--- a/src/metric-engine/src/engine/drop.rs
+++ b/src/metric-engine/src/engine/drop.rs
@@ -71,7 +71,7 @@ impl MetricEngineInner {
                 .get(&region_id)
                 .copied();
             if let Some(metadata_region_id) = metadata_region_id {
-                self.drop_logical_region(region_id, metadata_region_id, req.fast_drop_database_path)
+                self.drop_logical_region(region_id, metadata_region_id, req.fast_drop_path)
                     .await
             } else {
                 Err(LogicalRegionNotFoundSnafu { region_id }.build())
@@ -90,7 +90,7 @@ impl MetricEngineInner {
             .handle_request(
                 data_region_id,
                 RegionRequest::Drop(RegionDropRequest {
-                    fast_drop_database_path: false,
+                    fast_drop_path: false,
                 }),
             )
             .await
@@ -99,7 +99,7 @@ impl MetricEngineInner {
             .handle_request(
                 metadata_region_id,
                 RegionRequest::Drop(RegionDropRequest {
-                    fast_drop_database_path: false,
+                    fast_drop_path: false,
                 }),
             )
             .await
@@ -120,15 +120,11 @@ impl MetricEngineInner {
         &self,
         logical_region_id: RegionId,
         physical_region_id: RegionId,
-        fast_drop_database_path: bool,
+        fast_drop_path: bool,
     ) -> Result<AffectedRows> {
         // Update metadata
         self.metadata_region
-            .remove_logical_region(
-                physical_region_id,
-                logical_region_id,
-                fast_drop_database_path,
-            )
+            .remove_logical_region(physical_region_id, logical_region_id, fast_drop_path)
             .await?;
 
         // Update engine state
--- a/src/metric-engine/src/engine/drop.rs
+++ b/src/metric-engine/src/engine/drop.rs
@@ -30,7 +30,7 @@ impl MetricEngineInner {
     pub async fn drop_region(
         &self,
         region_id: RegionId,
-        _req: RegionDropRequest,
+        req: RegionDropRequest,
     ) -> Result<AffectedRows> {
         let data_region_id = utils::to_data_region_id(region_id);
 
@@ -71,7 +71,7 @@ impl MetricEngineInner {
                 .get(&region_id)
                 .copied();
             if let Some(metadata_region_id) = metadata_region_id {
-                self.drop_logical_region(region_id, metadata_region_id)
+                self.drop_logical_region(region_id, metadata_region_id, req.fast_drop_database_path)
                     .await
             } else {
                 Err(LogicalRegionNotFoundSnafu { region_id }.build())
@@ -87,13 +87,20 @@ impl MetricEngineInner {
         // Since the physical regions are going to be dropped, we don't need to
         // update the contents in metadata region.
         self.mito
-            .handle_request(data_region_id, RegionRequest::Drop(RegionDropRequest {}))
+            .handle_request(
+                data_region_id,
+                RegionRequest::Drop(RegionDropRequest {
+                    fast_drop_database_path: false,
+                }),
+            )
             .await
             .with_context(|_| CloseMitoRegionSnafu { region_id })?;
         self.mito
             .handle_request(
                 metadata_region_id,
-                RegionRequest::Drop(RegionDropRequest {}),
+                RegionRequest::Drop(RegionDropRequest {
+                    fast_drop_database_path: false,
+                }),
             )
             .await
             .with_context(|_| CloseMitoRegionSnafu { region_id })?;
@@ -113,10 +120,15 @@ impl MetricEngineInner {
         &self,
         logical_region_id: RegionId,
         physical_region_id: RegionId,
+        fast_drop_database_path: bool,
     ) -> Result<AffectedRows> {
         // Update metadata
         self.metadata_region
-            .remove_logical_region(physical_region_id, logical_region_id)
+            .remove_logical_region(
+                physical_region_id,
+                logical_region_id,
+                fast_drop_database_path,
+            )
             .await?;
 
         // Update engine state
--- a/src/metric-engine/src/engine/drop.rs
+++ b/src/metric-engine/src/engine/drop.rs
@@ -33,7 +33,7 @@ impl MetricEngineInner {
         req: RegionDropRequest,
     ) -> Result<AffectedRows> {
         let data_region_id = utils::to_data_region_id(region_id);
-        let force_drop_all_logical_tables = req.force_drop_all_logical_tables;
+        let fast_path = req.fast_path;
 
         // enclose the guard in a block to prevent the guard from polluting the async context
         let (is_physical_region, is_physical_region_busy) = {
@@ -53,7 +53,7 @@ impl MetricEngineInner {
 
         if is_physical_region {
             // check if there is no logical region relates to this physical region
-            if is_physical_region_busy && !force_drop_all_logical_tables {
+            if is_physical_region_busy && !fast_path {
                 // reject if there is any present logical region
                 return Err(PhysicalRegionBusySnafu {
                     region_id: data_region_id,
@@ -61,9 +61,21 @@ impl MetricEngineInner {
                 .build());
             }
 
-            self.drop_physical_region(data_region_id, force_drop_all_logical_tables)
-                .await
-        } else if !force_drop_all_logical_tables {
+            return self.drop_physical_region(data_region_id).await;
+        }
+
+        if fast_path {
+            // for fast path, we don't delete the metadata in the metadata region.
+            // it only remove the logical region from the engine state.
+            //
+            // The drop database procedure will ensure the metadata region and data region are dropped eventually.
+            self.state
+                .write()
+                .unwrap()
+                .remove_logical_region(region_id)?;
+
+            Ok(0)
+        } else {
             let metadata_region_id = self
                 .state
                 .read()
@@ -77,16 +89,10 @@ impl MetricEngineInner {
             } else {
                 Err(LogicalRegionNotFoundSnafu { region_id }.build())
             }
-        } else {
-            Ok(0)
         }
     }
 
-    async fn drop_physical_region(
-        &self,
-        region_id: RegionId,
-        force_drop_all_logical_tables: bool,
-    ) -> Result<AffectedRows> {
+    async fn drop_physical_region(&self, region_id: RegionId) -> Result<AffectedRows> {
         let data_region_id = utils::to_data_region_id(region_id);
         let metadata_region_id = utils::to_metadata_region_id(region_id);
 
@@ -96,18 +102,14 @@ impl MetricEngineInner {
         self.mito
             .handle_request(
                 data_region_id,
-                RegionRequest::Drop(RegionDropRequest {
-                    force_drop_all_logical_tables,
-                }),
+                RegionRequest::Drop(RegionDropRequest { fast_path: false }),
             )
             .await
             .with_context(|_| CloseMitoRegionSnafu { region_id })?;
         self.mito
             .handle_request(
                 metadata_region_id,
-                RegionRequest::Drop(RegionDropRequest {
-                    force_drop_all_logical_tables,
-                }),
+                RegionRequest::Drop(RegionDropRequest { fast_path: false }),
             )
             .await
             .with_context(|_| CloseMitoRegionSnafu { region_id })?;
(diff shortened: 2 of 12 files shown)


src/metric-engine/src/metadata_region.rs
```diff
@@ -122,26 +122,9 @@ impl MetadataRegion {
...
-            .await?;
-        let mut column_keys = logical_columns
-            .into_iter()
-            .map(|(col, _)| Self::concat_column_key(logical_region_id, &col))
-            .collect::<Vec<_>>();
-
-        // remove region key and column keys
-        column_keys.push(region_key);
-        self.delete(region_id, &column_keys).await?;
-
```
fengjiachun, line 144: What will happen if we just want to drop a logical region?

yihong0618: seems we can not here i searched the code base this function not affect that

fengjiachun: This part may need to be retained when dropping a logic table, but there should also be a fast path to execute `drop phy_table`. cc @WenyXu @waynexia

yihong0618: if add a fast_path arg in Drop database seems need to through some functions 
do we need that? maybe I can fix it in this way after some discuss.

WenyXu: > if add a fast_path arg in Drop database seems need to through some functions do we need that? maybe I can fix it in this way after some discuss.

We can add a fast_path in the drop database procedure. If the datanode receives a drop physical table request with a force flag, it will drop all logical tables and the physical table in a single request.

yihong0618: > > if add a fast_path arg in Drop database seems need to through some functions do we need that? maybe I can fix it in this way after some discuss.
> 
> We can add a fast_path in the drop database procedure. If the datanode receives a drop physical table request with a force flag, it will drop all logical tables and the physical table in a single request.

yes that's my point, it will go through some functions if that is OK, I will do it in This way

yihong0618: find it is kind of hard add it in the drop database procedure

yihong0618:
@fengjiachun @WenyXu done using `fast_drop_database_path` way
for the ci using my fork repo
and upstream patch link https://github.com/GreptimeTeam/greptime-proto/pull/218

fengjiachun (approved):
LGTM, thanks you! @yihong0618

fengjiachun:
@WenyXu PTAL

src/store-api/src/region_request.rs
```diff
@@ -222,7 +222,13 @@ fn make_region_creates(creates: CreateRequests) -> Result<Vec<(RegionId, RegionR
 
 fn parse_region_drop(drop: DropRequest) -> Result<(RegionId, RegionDropRequest)> {
     let region_id = drop.region_id.into();
-    Ok((region_id, RegionDropRequest {}))
+    let fast_drop_database_path = drop.fast_drop_database_path;
```
fengjiachun, line 225: How about using `fast_drop_path`? I believe we might also offer the ability to fast drop physical tables in the future, so it's not just for dropping databases.

yihong0618: Done

src/datanode/src/region_server.rs
```diff
@@ -1218,7 +1218,12 @@ mod tests {
         );
 
         let response = mock_region_server
-            .handle_request(region_id, RegionRequest::Drop(RegionDropRequest {}))
+            .handle_request(
+                region_id,
+                RegionRequest::Drop(RegionDropRequest {
+                    fast_drop_path: false,
```
WenyXu, line 1224: I suggest introducing a force_drop_all_logical_tables flag. When set to true, it will drop the physical table along with all logical tables associated with it.
```suggestion
                    force_drop_all_logical_tables: false,
```

src/metric-engine/src/engine/drop.rs
```diff
@@ -71,7 +71,7 @@ impl MetricEngineInner {
                 .get(&region_id)
                 .copied();
             if let Some(metadata_region_id) = metadata_region_id {
-                self.drop_logical_region(region_id, metadata_region_id)
+                self.drop_logical_region(region_id, metadata_region_id, req.fast_drop_path)
```
WenyXu, line 74: The `force_drop_all_logical_tables` flag should only apply to the physical region

src/store-api/src/region_request.rs
```diff
@@ -397,8 +398,12 @@ impl RegionCreateRequest {
...
 }
 
-#[derive(Debug, Clone, Default)]
-pub struct RegionDropRequest {}
+#[derive(Debug, Clone)]
+pub struct RegionDropRequest {
+    /// fast drop database path for the shortcut that do not need to delete logical
+    /// columns first more to check issue #4974 and #5561
+    pub fast_drop_path: bool,
+}
```
WenyXu, line 406: Any idea about defining the field? @evenyag @fengjiachun

src/mito2/src/request.rs
```diff
@@ -690,6 +690,7 @@ impl WorkerRequest {
 #[derive(Debug)]
 pub(crate) enum DdlRequest {
     Create(RegionCreateRequest),
+    #[allow(dead_code)]
```
WenyXu, line 693:
```suggestion
```

src/metric-engine/src/metadata_region.rs
```diff
@@ -124,23 +124,26 @@ impl MetadataRegion {
...
+                .map(|(col, _)| Self::concat_column_key(logical_region_id, &col))
+                .collect::<Vec<_>>();
 
-        // remove region key and column keys
-        column_keys.push(region_key);
-        self.delete(region_id, &column_keys).await?;
+            // remove region key and column keys
+            column_keys.push(region_key);
+            self.delete(region_id, &column_keys).await?;
+        }
```
WenyXu, line 146: The `force_drop_all_logical_tables` flag should only apply to the physical region. So we only need to handle the flag in `drop_physical_region` method. It will drop the metadata region and the datanode region directly without any checks

yihong0618: I wonder using this way will also go through `remove_logical_region` when logical region and will cost the most time, will check it now

src/store-api/src/region_request.rs
yihong0618: will change the name after discuss 
IMO force_drop_all_logical_tables is kind of better.
but as mention by @fengjiachun 
> How about using fast_drop_path? I believe we might also offer the ability to fast drop physical tables in the future, so it's not just for dropping databases.

fengjiachun: `force_drop_all_logical_tables` +1

src/metric-engine/src/metadata_region.rs
yihong0618: will try not go to the `else` part using this flag

WenyXu: If the user submits a drop database request, we don't need to send a drop logical table request to datanode anymore. So we can only handle the flag in `drop_physical_region` method.  That is we don't need to handle the  `(DropTableTarget::Logical, TableRouteValue::Logical(route))` branch anymore.

https://github.com/GreptimeTeam/greptimedb/blob/f6f617d6672d0d854b0a5dd052c8e40f4add48d5/src/common/meta/src/ddl/drop_database/cursor.rs#L64-L107

yihong0618: Thanks for the info learned that

yihong0618: I don't think this way is simple, its hard to pass the force_drop_all_logical_tables to the cursor

yihong0618: did not follow this way

yihong0618:
address by use else path way.
and frankly speaking, I prefer the previous implementation in the drop logical function seems no particular issue for that?

WenyXu:
> address by use else path way. and frankly speaking, I prefer the previous implementation in the drop logical function seems no particular issue for that?

If all logical tables are force dropped via drop physical region request, we don't need to send drop logical table requests to datanode. By the way, we don't need to handle the `(DropTableTarget::Logical, TableRouteValue::Logical(route))` branch anymore; just ignore it.
https://github.com/GreptimeTeam/greptimedb/blob/f6f617d6672d0d854b0a5dd052c8e40f4add48d5/src/common/meta/src/ddl/drop_database/cursor.rs#L64-L91

yihong0618:
> > address by use else path way. and frankly speaking, I prefer the previous implementation in the drop logical function seems no particular issue for that?
> 
> If all logical tables are force dropped via drop physical region request, we don't need to send drop logical table requests to datanode. By the way, we don't need to handle the `(DropTableTarget::Logical, TableRouteValue::Logical(route))` branch anymore; just ignore it.
> 
> https://github.com/GreptimeTeam/greptimedb/blob/f6f617d6672d0d854b0a5dd052c8e40f4add48d5/src/common/meta/src/ddl/drop_database/cursor.rs#L64-L91

dropped

yihong0618:
still think old way is better and maybe for the recover, so holding this

WenyXu:
> > > address by use else path way. and frankly speaking, I prefer the previous implementation in the drop logical function seems no particular issue for that?
> > 
> > 
> > If all logical tables are force dropped via drop physical region request, we don't need to send drop logical table requests to datanode. By the way, we don't need to handle the `(DropTableTarget::Logical, TableRouteValue::Logical(route))` branch anymore; just ignore it.
> > https://github.com/GreptimeTeam/greptimedb/blob/f6f617d6672d0d854b0a5dd052c8e40f4add48d5/src/common/meta/src/ddl/drop_database/cursor.rs#L64-L91
> 
> using this way do not check is logical region can be recover? the `test_drop_database_recover` also failed @WenyXu
> 
> error local
> 
> ```
> running 1 test
> 2025-02-21T13:04:34.395303Z  INFO common_telemetry::logging: logs dir = /tmp/__unittest_logs
> 2025-02-21T13:04:34.395877Z DEBUG common_meta::ddl::table_meta: Allocated region wal options {0: "{\"wal.provider\":\"raft_engine\"}"} for table 1024
> 2025-02-21T13:04:34.396312Z  WARN common_meta::ddl::create_logical_tables::update_metadata: No physical columns found, leaving the physical table's schema unchanged when creating logical tables
> 2025-02-21T13:04:34.396379Z  INFO common_meta::ddl::create_logical_tables::update_metadata: Created 1 tables [1025] metadata for physical table 1024
> 2025-02-21T13:04:34.397300Z DEBUG common_meta::ddl::drop_table::executor: Dropping region 4398046511104(1024, 0) on Datanode Peer { id: 0, addr: "" }
> 2025-02-21T13:04:34.397335Z DEBUG common_meta::ddl::test_util::datanode_handler: Returning Ok(0) for request: RegionRequest { header: Some(RegionRequestHeader { tracing_context: {}, dbname: "", query_context: None }), body: Some(Drop(DropRequest { region_id: 4398046511104, force_drop_all_logical_tables: true })) }, peer: Peer { id: 0, addr: "" }
> 2025-02-21T13:04:34.397358Z  INFO common_meta::ddl::drop_database::executor: Table: greptime.public.phy(1024) is dropped
> test ddl::tests::drop_database::test_drop_database_recover ... FAILED
> 
> failures:
> 
> failures:
>     ddl::tests::drop_database::test_drop_database_recover
> 
> test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 248 filtered out; finished in 0.01s
> 
> ──── TRY 3 STDERR:       common-meta ddl::tests::drop_database::test_drop_database_recover
> thread 'ddl::tests::drop_database::test_drop_database_recover' panicked at /Users/hyi/prs/greptimedb/src/common/procedure-test/src/lib.rs:137:5:
> procedure 'metasrv-procedure::DropDatabase' did not reach the expected state
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> 
>    RETRY 4/4 [         ] common-meta ddl::tests::drop_database::test_drop_database_recover
>         PASS [   0.006s] common-meta key::flow::flow_info::tests::test_key_serialization
>         PASS [   0.007s] common-meta key::flow::flow_name::tests::test_key_serialization
>         PASS [   0.011s] common-meta key::flow::flow_info::tests::test_key_deserialization
>         PASS [   0.008s] common-meta key::flow::flow_name::tests::test_key_start_range
>         PASS [   0.007s] common-meta key::flow::flow_route::tests::test_key_deserialization
>         PASS [   0.014s] common-meta key::flow::flow_name::tests::test_key_deserialization
>         PASS [   0.008s] common-meta key::flow::flow_route::tests::test_key_serialization
>         PASS [   0.009s] common-meta key::flow::flow_route::tests::test_key_start_range
>         PASS [   0.008s] common-meta key::flow::flownode_flow::tests::test_key_serialization
>         PASS [   0.009s] common-meta key::flow::flownode_flow::tests::test_key_deserialization
>   TRY 4 FAIL [   0.019s] common-meta ddl::tests::drop_database::test_drop_database_recover
> ──── TRY 4 STDOUT:       common-meta ddl::tests::drop_database::test_drop_database_recover
> 
> running 1 test
> 2025-02-21T13:04:34.415082Z  INFO common_telemetry::logging: logs dir = /tmp/__unittest_logs
> 2025-02-21T13:04:34.415863Z DEBUG common_meta::ddl::table_meta: Allocated region wal options {0: "{\"wal.provider\":\"raft_engine\"}"} for table 1024
> 2025-02-21T13:04:34.416323Z  WARN common_meta::ddl::create_logical_tables::update_metadata: No physical columns found, leaving the physical table's schema unchanged when creating logical tables
> 2025-02-21T13:04:34.416394Z  INFO common_meta::ddl::create_logical_tables::update_metadata: Created 1 tables [1025] metadata for physical table 1024
> 2025-02-21T13:04:34.417372Z DEBUG common_meta::ddl::drop_table::executor: Dropping region 4398046511104(1024, 0) on Datanode Peer { id: 0, addr: "" }
> test ddl::tests::drop_database::test_drop_database_recover ... FAILED
> 
> failures:
> 
> failures:
>     ddl::tests::drop_database::test_drop_database_recover
> 
> test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 248 filtered out; finished in 0.01s
> 
> ──── TRY 4 STDERR:       common-meta ddl::tests::drop_database::test_drop_database_recover
> thread 'ddl::tests::drop_database::test_drop_database_recover' panicked at /Users/hyi/prs/greptimedb/src/common/procedure-test/src/lib.rs:137:5:
> procedure 'metasrv-procedure::DropDatabase' did not reach the expected state
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> 
>   Cancelling due to test failure: 7 tests still running
>         PASS [   0.008s] common-meta key::flow::table_flow::tests::test_key_deserialization
>         PASS [   0.008s] common-meta key::flow::table_flow::tests::test_key_serialization
>         PASS [   0.008s] common-meta key::flow::tests::test_create_flow_metadata_flow_exists_err
>         PASS [   0.010s] common-meta key::flow::tests::test_create_flow_metadata
>         PASS [   0.009s] common-meta key::flow::tests::test_create_flow_metadata_unexpected_err
>         PASS [   0.531s] common-greptimedb-telemetry tests::test_gretimedb_telemetry
>         PASS [   1.962s] cmd cli::tests::test_export_create_table_with_quoted_names
> ────────────
>      Summary [   2.140s] 411/3137 tests run: 410 passed, 1 failed, 8 skipped
>   TRY 4 FAIL [   0.019s] common-meta ddl::tests::drop_database::test_drop_database_recover
> warning: 2726/3137 tests were not run due to test failure (run with --no-fail-fast to run all tests, or run with --max-fail)
> ```

Yes, you can update the unit test.

WenyXu:
> * check in the drop logical region side do not break any test

We can just update the test.

> * to ignore the logical table seems left some empty dir in the data(which I will confirm)

It wouldn't.

> * and easy to understand, and also for the later change for the code

I think left some comments are aceeptable.

yihong0618:
> * and easy to unders

copy that and thank you very much for the confirm

WenyXu:
> * check in the drop logical region side do not break any test

BTW, I’ll help check the test when I have some free time. It shouldn’t be too tricky.

yihong0618:
> > * check in the drop logical region side do not break any test
> 
> BTW, I’ll help check the test when I have some free time. It shouldn’t be too tricky.

Thanks again learned that!

src/common/meta/src/ddl/drop_database/cursor.rs
```diff
@@ -248,38 +229,6 @@ mod tests {
         assert_eq!(executor.target, DropTableTarget::Physical);
     }
 
-    #[tokio::test]
-    async fn test_next_with_logical_tables() {
```
WenyXu, line 252: Let's cover the case with logical tables. you can update the test instead of removing it.

src/common/meta/src/ddl/tests/drop_database.rs
```diff
@@ -74,7 +74,9 @@ async fn test_drop_database_with_logical_tables() {
         .try_collect::<Vec<_>>()
         .await
         .unwrap();
-    assert!(tables.is_empty());
+    // the logical region will shortcut drop when DropDatabaseProcedure so the table is not empty now
+    // more to check issue #5561
+    assert!(!tables.is_empty());
```
WenyXu, line 79: Oh, I see. we don't need to send drop logical table requests to datanode. However, we should drop the table metadata.

src/common/meta/src/ddl/tests/drop_database.rs
```diff
@@ -196,5 +187,5 @@ async fn test_drop_database_recover() {
         .try_collect::<Vec<_>>()
         .await
         .unwrap();
-    assert!(tables.is_empty());
+    assert!(!tables.is_empty());
```
WenyXu, line 190: ditto

src/mito2/src/request.rs
```diff
@@ -690,7 +690,7 @@ impl WorkerRequest {
 #[derive(Debug)]
 pub(crate) enum DdlRequest {
     Create(RegionCreateRequest),
-    Drop(RegionDropRequest),
+    Drop(()),
```
WenyXu, line 693:
```suggestion
    Drop,
```

WenyXu:
I realize the changes in this PR might be a bit tricky. However, this is a critical feature that I’d like to include in the 0.12 release. Would you mind if I make the adjustments directly and push them to this PR? Feel free to share your thoughts.

src/common/meta/src/ddl/tests/drop_database.rs
yihong0618: in the test or in the function?

yihong0618:
> I realize the changes in this PR might be a bit tricky. However, this is a critical feature that I’d like to include in the 0.12 release. Would you mind if I make the adjustments directly and push them to this PR? Feel free to share your thoughts.

just continue commit on this Pull request is nice I wonder, And I'd like to learn from you, thanks again

src/common/meta/src/ddl/tests/drop_database.rs
WenyXu: I realized we can't remove the code directly. Instead, we should pass an argument to indicate that it's a task to drop the logical table—this should only remove the table metadata and invalidate the table caches, while skipping requests to the datanode.

```rust
            (DropTableTarget::Logical, TableRouteValue::Logical(route)) => {
                 let physical_table_id = route.physical_table_id();

                 let (_, table_route) = ddl_ctx
                     .table_metadata_manager
                     .table_route_manager()
                     .get_physical_table_route(physical_table_id)
                     .await?;
                 Ok((
                     Box::new(DropDatabaseExecutor::new(
                         table_id,
                         table_id,
                         TableName::new(&ctx.catalog, &ctx.schema, &table_name),
                         table_route.region_routes,
                         self.target,

                     )),
                     Status::executing(true),
                 ))
             }
```

WenyXu: e.g.,
```rust
                     Box::new(DropDatabaseExecutor::new(
                         table_id,
                         table_id,
                         TableName::new(&ctx.catalog, &ctx.schema, &table_name),
                         table_route.region_routes,
                         self.target,
                         // drop_logical_table
                         true,
                     )),
```

WenyXu:
> > I realize the changes in this PR might be a bit tricky. However, this is a critical feature that I’d like to include in the 0.12 release. Would you mind if I make the adjustments directly and push them to this PR? Feel free to share your thoughts.
> 
> just continue commit if you have some thought I missed, on this Pull request is nice I wonder, And I'd like to learn from you, thanks again
> 
> and I will fix the review comments

Great, I’ve added some detailed comments. Feel free to reach out if you have any questions.

src/common/meta/src/ddl/tests/drop_database.rs
yihong0618: yes....I also found that, that's why I left the comments 
> still think old way is better and maybe for the recover, so holding this.

check in the drop logical region side do not break any test
to ignore the logical table seems left some empty dir in the data(which I will confirm)

the old way I use can pass the problem, because we only ignore the `function get_all or get_all_with_prefix` 
and the columns delete in the logical table, the meta table also delete.

But your comment is fit for me now, will sleep tonight, feel free to continue or I can try to do it tomorrow.

good night.

yihong0618: Thinking about it, since we need to pass the arg and it is tricky
why do not consider https://github.com/GreptimeTeam/greptimedb/pull/5561/commits/051a1acbc91965e18c2ca50d626198af3ba0a842

its simple then this. I think

src/mito2/src/request.rs
yihong0618: because of 
```rs
            RegionRequest::Drop(_) => WorkerRequest::Ddl(SenderDdlRequest {
                region_id,
                sender: sender.into(),
                request: DdlRequest::Drop(()),
            }),
```
if change that can not compile

yihong0618:
```
error[E0532]: expected tuple struct or tuple variant, found unit variant `RegionRequest::Drop`
    --> src/store-api/src/region_request.rs:1138:13
     |
119  |     Drop,
     |     ---- `RegionRequest::Drop` defined here
...
1138 |             RegionRequest::Drop(_) => write!(f, "Drop"),
     |             ^^^^^^^^^^^^^^^^^^^^^^
     |
help: use this syntax instead
     |
1138 |             RegionRequest::Drop => write!(f, "Drop"),
     |             ~~~~~~~~~~~~~~~~~~~
help: consider importing one of these tuple variants instead
     |
15   + use crate::region_request::BatchRegionDdlRequest::Drop;
     |
15   + use crate::region_request::region_request::Body::Drop;
     |
15   + use crate::region_request::v1::flow::flow_request::Body::Drop;
     |
15   + use api::v1::flow::flow_request::Body::Drop;
     |
       and 1 other candidate
help: if you import `Drop`, refer to it directly
     |
1138 -             RegionRequest::Drop(_) => write!(f, "Drop"),
1138 +             Drop(_) => write!(f, "Drop"),
     |

error[E0618]: expected function, found `region_request::RegionRequest`
   --> src/store-api/src/region_request.rs:104:57
    |
104 |                 .map(|(region_id, request)| (region_id, RegionRequest::Drop(request...
    |                                                         ^^^^^^^^^^^^^^^^^^^---------
    |                                                         |
    |                                                         call expression requires function
...
119 |     Drop,
    |     ---- `region_request::RegionRequest::Drop` defined here

error[E0618]: expected function, found `region_request::RegionRequest`
   --> src/store-api/src/region_request.rs:236:25
    |
119 |     Drop,
    |     ---- `region_request::RegionRequest::Drop` defined here
...
236 |     Ok(vec![(region_id, RegionRequest::Drop(request))])
    |                         ^^^^^^^^^^^^^^^^^^^---------
    |                         |
    |                         call expression requires function
```

we can not just use `Drop` here

two ways can pass this

1. `#[allow(dead_code)]`
2. in the code now
3. use it println!("Dropping region: {:?}", req.region_id);

src/common/meta/src/ddl/tests/drop_database.rs
WenyXu: > I am just thinking about it, since we need to pass the arg and it is tricky why do not consider [051a1ac](https://github.com/GreptimeTeam/greptimedb/commit/051a1acbc91965e18c2ca50d626198af3ba0a842)
> 
> its simple and easy to change then this and no need to add more comments to explain. I think cc @WenyXu @fengjiachun

just want to avoid sending RPC calls to the datanode

WenyXu: > I am just thinking about it, since we need to pass the arg and it is tricky

why it's tricky?

yihong0618: sorry I mean the old way.

yihong0618:
rebased for easy change

yihong0618:
> > > I realize the changes in this PR might be a bit tricky. However, this is a critical feature that I’d like to include in the 0.12 release. Would you mind if I make the adjustments directly and push them to this PR? Feel free to share your thoughts.
> > 
> > 
> > just continue commit if you have some thought I missed, on this Pull request is nice I wonder, And I'd like to learn from you, thanks again
> > and I will fix the review comments
> 
> Great, I’ve added some detailed comments. Feel free to reach out if you have any questions.

address and add co-author

src/common/meta/src/ddl/tests/drop_database.rs
WenyXu: > just want to avoid sending RPC calls to the datanode

My apologies. After attempting to refactor your PR, I realized that skipping the sending of drop logical table requests to the datanode could lead to resource leaks.

WenyXu:
Hi, I realized this PR could be a little tricky. Would you be okay if I made changes directly on your code base? If we can't avoid sending RPCs to the datanode, `fast_path` would be more appropriate than `force_drop_all_logical_tables`.

WenyXu:
> using your way is sure had empty dir

It's expected.

WenyXu (approved):

PR #5567, opened 2025-02-19
feat: support to generate json output for explain analyze in http api

waynexia:

Now the `/v1/sql` API can generate something like the following for `explain analyze` clause 
image

It can be turned on with a new HTTP header `x-greptime-analyze-format`

## PR Checklist

### DIFF
--- a/src/query/src/analyze.rs
+++ b/src/query/src/analyze.rs
@@ -215,14 +215,14 @@ fn create_output_batch(
     let stage_0_metrics = collector.record_batch_metrics;
 
     // Append the metrics of the current stage
-    builder.append_metric(0, 0, metrics_to_string(stage_0_metrics, format));
+    builder.append_metric(0, 0, metrics_to_string(stage_0_metrics, format)?);
 
     // Find merge scan and append its sub_stage_metrics
     input.apply(|plan| {
         if let Some(merge_scan) = plan.as_any().downcast_ref::<MergeScanExec>() {
             let sub_stage_metrics = merge_scan.sub_stage_metrics();
             for (node, metric) in sub_stage_metrics.into_iter().enumerate() {
-                builder.append_metric(1, node as _, metrics_to_string(metric, format));
+                builder.append_metric(1, node as _, metrics_to_string(metric, format)?);
             }
             return Ok(TreeNodeRecursion::Stop);
         }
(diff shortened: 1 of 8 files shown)


sunng87 (approved):
LGTM

src/query/src/analyze.rs
```diff
@@ -226,3 +234,61 @@ fn create_output_batch(
 
     builder.finish()
 }
+
+fn metrics_to_string(metrics: RecordBatchMetrics, format: AnalyzeFormat) -> String {
+    match format {
+        AnalyzeFormat::JSON => JsonMetrics::from_record_batch_metrics(metrics).to_string(),
+        _ => metrics.to_string(),
```
killme2008, line 241: List the variant explicitly:
```suggestion
        AnalyzeFormat::TEXT => metrics.to_string(),
```

src/session/src/context.rs
```diff
@@ -38,6 +38,8 @@ pub type ConnInfoRef = Arc<ConnInfo>;
 
 const CURSOR_COUNT_WARNING_LIMIT: usize = 10;
 
+pub const ANALYZE_FORMAT_HEADER_NAME: &str = "x-greptime-analyze-format";
```
killme2008, line 41: What about adding `query` to the header name? It's more precise.
```suggestion
pub const ANALYZE_FORMAT_HEADER_NAME: &str = "x-greptime-query-analyze-format";
```

killme2008 (approved):
Almost LGTM

src/session/src/context.rs
evenyag: Isn't `explain-analyze-format`?

src/query/src/analyze.rs
waynexia: I mean to ignore it. There is another format like `graphviz`, which we are not likely to support shortly nor want to throw an error.

src/session/src/context.rs
waynexia: I'll pick `x-greptime-query-analyze-format`, `explain analyze` is a verbose grammar thing.

src/query/src/analyze.rs
```diff
@@ -226,3 +234,61 @@ fn create_output_batch(
...
+    }
+}
+
+#[derive(Debug, Default, Serialize)]
+struct JsonMetrics {
+    name: String,
+    param: String,
+    metrics: HashMap<String, usize>,
```
sunng87, line 249: we can lift common attributes like time, data size and memory to top level

sunng87:
Refactored to use format option from `EXPLAIN FORMAT json ...`

src/query/src/analyze.rs
killme2008: Our code style prefers listing all variants in the match enumeration type if I don't get wrong.

killme2008: If the format doesn't support, let's return an error?

PR #5571, opened 2025-02-19
feat: submit node's cpu cores number to metasrv in heartbeat

MichaelScofield:

as title

metasrv then is aware of cluster's total cpus, and able to do something about it further

## PR Checklist

### DIFF
--- a/src/common/meta/src/cluster.rs
+++ b/src/common/meta/src/cluster.rs
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::hash::{DefaultHasher, Hash, Hasher};
 use std::str::FromStr;
 
+use api::v1::meta::HeartbeatRequest;
 use common_error::ext::ErrorExt;
 use lazy_static::lazy_static;
 use regex::Regex;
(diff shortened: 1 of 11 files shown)


src/common/meta/src/cluster.rs
```diff
@@ -69,6 +71,28 @@ pub struct NodeInfoKey {
...
+    pub fn new(request: &HeartbeatRequest) -> Option<Self> {
+        let HeartbeatRequest { header, peer, .. } = request;
+        let header = header.as_ref()?;
+        let peer = peer.as_ref()?;
+
+        let role = header.role.try_into().ok()?;
+        let node_id = match role {
+            // Because the Frontend is stateless, it's too easy to neglect choosing a unique id
+            // for it when setting up a cluster. So we calculate its id from its address.
+            Role::Frontend => calculate_node_id(&peer.addr),
```
sunng87, line 85: This relies on the `peer.addr` is correctly detected. 
@zyy17 wdyt

sunng87 (approved):

src/common/meta/src/cluster.rs
MichaelScofield: I think we should set the frontend's id correctly at the first place, can be simply calculated out from its pod address at the operator side

fengjiachun (approved):
LGTM

Issue #5139, opened 2024-12-11
support for exporting data to client or S3


WenyXu:
### What type of enhancement is this?

User experience

### What does the enhancement do?

Our data export tool currently sends a `COPY DATABASE ... TO` statement to the frontend node, resulting in the exported data being saved only on the server's (frontend node's) local disk, regardless of where the command is executed.

We should enhance the tool to support exporting data directly to the client or an S3 bucket.

See also #5136

PR #5585, opened 2025-02-22
feat: support export command export data to s3

### LINKED ISSUE #5533: Data Not Exported in `greptime cli export` When Running in Docker/Kubernetes
### What type of bug is this?

Incorrect result

Distributed Cluster

### Minimal reproduce step

1. Deploy GreptimeDB in a Docker containers using docker compose or using the provided Helm chart on Kubernetes (cluster mode).
2. Run the following command using the GreptimeDB CLI binary (downloaded from the GitHub Releases section):
   ```sh
   .\greptime.exe cli export --addr <REDACTED_IP>:4000 --output-dir .\backup\
   ```
3. Observe that only SQL schema files are generated, but no Parquet files containing data.
4. Running the same command on a standalone Windows 11 instance of GreptimeDB successfully exports both schemas and data.

The `greptime cli export` command should generate both SQL schema definitions and Parquet files containing actual database data when running in a containerized environment, just as it does when running a standalone instance.

### What did you see instead?

- The CLI exports SQL schema files but does not generate any Parquet data files (even though I have a lot of data).
- Logs indicate that the `COPY DATABASE` command executes, but no data is written to the specified backup directory.

Windows 11 x64, Ubuntu 24.04

0.11.3

```bash
**Environment:**

- **GreptimeDB Version:** 0.11.3
- **GreptimeDB CLI Version:** 0.11.3 (Downloaded from GitHub Releases)
- **Operating System (CLI):** Windows 11
- **Server OS (When Not Working):** Ubuntu Server 24.04
- **Deployment Method (When Not Working):** Docker Compose / Helm Chart (I have tried both)

This is just a small piece of the log, but I have many tables.

2025-02-13T13:59:28.428719Z  INFO cli::export: Executing sql: COPY DATABASE "greptime"."private_schema" TO '.\backup\greptime\private_schema/' WITH (FORMAT='parquet');
2025-02-13T13:59:28.511580Z  INFO cli::export: Finished exporting greptime.private_schema data into path: .\backup\greptime\private_schema/
...

**Additional Notes:**

- The issue occurs only in containerized environments (Docker/Kubernetes).
- The same command works correctly on a standalone Windows installation of GreptimeDB.
- The `COPY DATABASE` SQL statement executes but does not produce Parquet files.
```

yihong0618:

This patch add S3 Export Support to export cli 

## desc 

This patch aim to support for exporting database data and schema using opendal

- use opendal crate not the object-store is because maybe cli can be a standalone tool to make it simple
- introduce three new errors
- about the database data this patch use a trick way( using connection to make the code easy)
- because we can export to s3 so this patch make the `output-dir` optional
- add some new args to command can check below

args now
```
pub struct ExportCommand {
    /// Server address to connect
    #[clap(long)]
    addr: String,

    /// Directory to put the exported data. E.g.: /tmp/greptimedb-export
    /// for local export.
    #[clap(long)]
    output_dir: Option<String>,

    /// The name of the catalog to export.
    #[clap(long, default_value = "greptime-*")]
    database: String,

    /// Parallelism of the export.
    #[clap(long, short = 'j', default_value = "1")]
    export_jobs: usize,

    /// Max retry times for each job.
    #[clap(long, default_value = "3")]
    max_retry: usize,

    /// Things to export
    #[clap(long, short = 't', value_enum, default_value = "all")]
    target: ExportTarget,

    /// A half-open time range: [start_time, end_time).
    /// The start of the time range (time-index column) for data export.
    #[clap(long)]
    start_time: Option<String>,

    /// A half-open time range: [start_time, end_time).
    /// The end of the time range (time-index column) for data export.
    #[clap(long)]
    end_time: Option<String>,

    /// The basic authentication for connecting to the server
    #[clap(long)]
    auth_basic: Option<String>,

    /// The timeout of invoking the database.
    ///
    /// It is used to override the server-side timeout setting.
    /// The default behavior will disable server-side default timeout(i.e. `0s`).
    #[clap(long, value_parser = humantime::parse_duration)]
    timeout: Option<Duration>,

    /// The proxy server address to connect, if set, will override the system proxy.
    ///
    /// The default behavior will use the system proxy if neither `proxy` nor `no_proxy` is set.
    #[clap(long)]
    proxy: Option<String>,

    /// Disable proxy server, if set, will not use any proxy.
    #[clap(long)]
    no_proxy: bool,

    /// if export data to s3
    #[clap(long)]
    s3: bool,

    /// The s3 bucket name
    /// if s3 is set, this is required
    #[clap(long)]
    s3_bucket: Option<String>,

    /// The s3 endpoint
    /// if s3 is set, this is required
    #[clap(long)]
    s3_endpoint: Option<String>,

    /// The s3 access key
    /// if s3 is set, this is required
    #[clap(long)]
    s3_access_key: Option<String>,

    /// The s3 secret key
    /// if s3 is set, this is required
    #[clap(long)]
    s3_secret_key: Option<String>,

    /// The s3 region
    /// if s3 is set, this is required
    #[clap(long)]
    s3_region: Option<String>,
}
```

## example command

```cli
./target/debug/greptime cli export --addr 127.0.0.1:4000 --s3 --s3-bucket ttt --s3-access-key minioadmin --s3-secret-key minioadmin --s3-region us-east-1 --s3-endpoint http://localhost:9000
```

future to do maybe the fs way also can use opendal to make code simple

close #5533 #5139 #5138

cc @fengjiachun @zyy17 

## PR Checklist

### DIFF
--- a/src/cli/src/export.rs
+++ b/src/cli/src/export.rs
@@ -21,6 +21,8 @@ use async_trait::async_trait;
 use clap::{Parser, ValueEnum};
 use common_error::ext::BoxedError;
 use common_telemetry::{debug, error, info};
+use opendal::layers::LoggingLayer;
+use opendal::{services, Operator};
 use serde_json::Value;
 use snafu::{OptionExt, ResultExt};
 use tokio::fs::File;
@@ -29,12 +31,11 @@ use tokio::sync::Semaphore;
 use tokio::time::Instant;
 
 use crate::database::{parse_proxy_opts, DatabaseClient};
-use crate::error::{EmptyResultSnafu, Error, FileIoSnafu, OpenDalSnafu, Result, S3ConfigNotSetSnafu, SchemaNotFoundSnafu};
+use crate::error::{
+    EmptyResultSnafu, Error, FileIoSnafu, OpenDalSnafu, OutputDirNotSetSnafu, Result,
+    S3ConfigNotSetSnafu, SchemaNotFoundSnafu,
+};
 use crate::{database, Tool};
-
-use opendal::services;
-use opendal::Operator;
-use opendal::layers::LoggingLayer;
 
 type TableReference = (String, String, String);
 
@@ -57,8 +58,9 @@ pub struct ExportCommand {
     addr: String,
 
     /// Directory to put the exported data. E.g.: /tmp/greptimedb-export
+    /// for local export.
     #[clap(long)]
-    output_dir: String,
+    output_dir: Option<String>,
 
     /// The name of the catalog to export.
     #[clap(long, default_value = "greptime-*")]
@@ -141,13 +143,16 @@ impl ExportCommand {
     pub async fn build(&self) -> std::result::Result<Box<dyn Tool>, BoxedError> {
         if self.s3
             && (self.s3_bucket.is_none()
-            || self.s3_endpoint.is_none()
-            || self.s3_access_key.is_none()
-            || self.s3_secret_key.is_none()
-            || self.s3_region.is_none())
-            {
-                return Err(BoxedError::new(S3ConfigNotSetSnafu {}.build()));
-            }
+                || self.s3_endpoint.is_none()
+                || self.s3_access_key.is_none()
+                || self.s3_secret_key.is_none()
+                || self.s3_region.is_none())
+        {
+            return Err(BoxedError::new(S3ConfigNotSetSnafu {}.build()));
+        }
+        if !self.s3 && self.output_dir.is_none() {
+            return Err(BoxedError::new(OutputDirNotSetSnafu {}.build()));
+        }
         let (catalog, schema) =
             database::split_database(&self.database).map_err(BoxedError::new)?;
         let proxy = parse_proxy_opts(self.proxy.clone(), self.no_proxy)?;
@@ -159,7 +164,6 @@ impl ExportCommand {
             self.timeout.unwrap_or_default(),
             proxy,
         );
-
 
         Ok(Box::new(Export {
             catalog,
@@ -185,7 +189,7 @@ pub struct Export {
     catalog: String,
     schema: Option<String>,
     database_client: DatabaseClient,
-    output_dir: String,
+    output_dir: Option<String>,
     parallelism: usize,
     target: ExportTarget,
     start_time: Option<String>,
@@ -200,7 +204,13 @@ pub struct Export {
 
 impl Export {
     fn catalog_path(&self) -> PathBuf {
-        PathBuf::from(&self.output_dir).join(&self.catalog)
+        if self.s3 {
+            PathBuf::from(&self.catalog)
+        } else if let Some(dir) = &self.output_dir {
+            PathBuf::from(dir).join(&self.catalog)
+        } else {
+            unreachable!("catalog_path: output_dir must be set when not using s3")
+        }
     }
 
     async fn get_db_names(&self) -> Result<Vec<String>> {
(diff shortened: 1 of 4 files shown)


fengjiachun:
Sorry for the delay; I'll review this PR after we release version 0.12.

yihong0618:
> Sorry for the delay; I'll review this PR after we release version 0.12.

Thanks and cool

src/cli/src/export.rs
```diff
@@ -101,10 +107,51 @@ pub struct ExportCommand {
...
+
+    /// The s3 secret key
+    /// if s3 is set, this is required
+    #[clap(long)]
+    s3_secret_key: Option<String>,
+
+    /// The s3 region
+    /// if s3 is set, this is required
+    #[clap(long)]
+    s3_region: Option<String>,
```
fengjiachun, line 138: Can we offer a configuration that supports multiple types of object storage, like GreptimeDB does currently?

https://github.com/GreptimeTeam/greptimedb/blob/dee76f0a73c2252675b7837bedd695b47bc9ac58/src/datanode/src/config.rs#L44

yihong0618: maybe since `export` is cli I wonder we offer cli command is needed

fengjiachun: I think it's necessary. For instance, if GreptimeDB stores data on `oss`, backups may also need to be stored on `oss`.

yihong0618: got your point, follow the greptimedb config is fine? if so will try to use that way.
if that, seems we need to change the config logic

fengjiachun: Yes, I think following the GreptimeDB config is fine.

src/cli/src/export.rs
```diff
@@ -375,84 +485,127 @@ impl Export {
...
                 };
 
                 let sql = format!(
-                    r#"COPY DATABASE "{}"."{}" TO '{}' {};"#,
-                    self.catalog,
-                    schema,
-                    db_dir.to_str().unwrap(),
-                    with_options
+                    r#"COPY DATABASE "{}"."{}" TO '{}' WITH ({}){};"#,
+                    export_self.catalog, schema, path, with_options_clone, connection_part
```
yihong0618, line 578: @fengjiachun do we support copy database with connection to all `OSS`?
I checked the doc only find s3
will check the code later

fengjiachun: Yes, the `copy database` supports connecting to both `oss` and `s3` using the same arguments: https://docs.greptime.com/reference/sql/copy/#options

yihong0618: just check the code about copy parquet part for now its only support s3 and f3
https://github.com/GreptimeTeam/greptimedb/blob/9860bca986c3639468d261468802930e8dec79d2/src/common/datasource/src/object_store.rs#L56-L75

So I think just support s3 is OK and maybe support others in the future

fengjiachun: Oh my fault, let's just consider s3 for now.

yihong0618: forget to support two things the latest commit support them, I think this is ready for review

src/cli/src/export.rs
```diff
@@ -300,19 +366,47 @@ impl Export {
         let timer = Instant::now();
         let db_names = self.get_db_names().await?;
         let db_count = db_names.len();
+        let s3_operator = if self.s3 {
+            Some(self.build_s3_operator().await?)
+        } else {
+            None
+        };
+
```
fengjiachun, line 374: Can we use the opendal operator for both S3 and FS?

fengjiachun (commented):
Others LGTM, Thank you @yihong0618

src/cli/src/export.rs
yihong0618: of course will try and fix them

yihong0618: Done

fengjiachun (approved):
LGTM

WenyXu (approved):

fengjiachun:
@yihong0618 Thank you!

PR #5588, opened 2025-02-24
fix: drop useless docker file ubuntu18

yihong0618:

seems ubuntu 18 docker file never use for now, this patch drop it.

and FYI centos7 seems can not build in arm64 I am not sure it still useful

```
 => CANCELED [base 2/4] RUN yum install -y epel-release     openssl     openssl-devel      centos-release-scl      2.3s
 => ERROR [builder 3/8] RUN ulimit -n 1024000 && yum groupinstall -y 'Development Tools'                           0.2s
------                                                                                                                  
 > [builder 3/8] RUN ulimit -n 1024000 && yum groupinstall -y 'Development Tools':                                      
0.100 Loaded plugins: fastestmirror, ovl                                                                                
0.124 There is no installed groups file.
0.124 Maybe run: yum groups mark convert (see man yum)
0.126 Determining fastest mirrors
0.131 Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=aarch64&repo=os&infra=container error was
0.131 14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"
0.132 
0.132 
0.132  One of the configured repositories failed (Unknown),
0.132  and yum doesn't have enough cached data to continue. At this point the only
0.132  safe thing yum can do is fail. There are a few ways to work "fix" this:
0.132 
0.132      1. Contact the upstream for the repository and get them to fix the problem.
0.132 
0.132      2. Reconfigure the baseurl/etc. for the repository, to point to a working
0.132         upstream. This is most often useful if you are using a newer
0.132         distribution release than is supported by the repository (and the
0.132         packages for the previous distribution release still work).
0.132 
0.132      3. Run the command with the repository temporarily disabled
0.132             yum --disablerepo=<repoid> ...
0.132 
0.132      4. Disable the repository permanently, so yum won't use it by default. Yum
0.132         will then just ignore the repository until you permanently enable it
0.132         again or use --enablerepo for temporary usage:
0.132 
0.132             yum-config-manager --disable <repoid>
0.132         or
0.132             subscription-manager repos --disable=<repoid>
0.132 
0.132      5. Configure the failing repository to be skipped, if it is unavailable.
0.132         Note that yum will try to contact the repo. when it runs most commands,
0.132         so will have to try and fail each time (and thus. yum will be be much
0.132         slower). If it is a very temporary problem though, this is often a nice
0.132         compromise:
0.132 
0.132             yum-config-manager --save --setopt=<repoid>.skip_if_unavailable=true
0.132 
0.132 Cannot find a valid baseurl for repo: base/7/aarch64
------
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load

 5 warnings found (use docker --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 36)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 7)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 25)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 48)
Dockerfile:11
--------------------
   9 |     
  10 |     # Install dependencies
  11 | >>> RUN ulimit -n 1024000 && yum groupinstall -y 'Development Tools'
  12 |     RUN yum install -y epel-release  \
  13 |         openssl \
--------------------
ERROR: failed to solve: process "/bin/sh -c ulimit -n 1024000 && yum groupinstall -y 'Development Tools'" did not complete successfully: exit code: 1
```

## PR Checklist


yihong0618:
> Already delete in: #5554

cool closing

yihong0618:
@daviderli614 about the centos part?

PR #5603, opened 2025-02-26
feat: introduce roaring bitmap to optimize sparse value scenarios

zhongzc:

Wait for https://github.com/GreptimeTeam/greptime-proto/pull/219

By adding `BitmapType` to `InvertedIndexMeta`, we can indicate the bitmap encoding in the index. Currently, there are two choices: bitvec and roaring. For compatibility, bitvec is the default.

* During writing, mito passes the bitmap type to the `finish` method of `InvertedIndexCreator` to control which encoding to use for writing the index.
* During reading, the bitmap is decoded based on the bitmap type in `InvertedIndexMeta`.

Introduced a `Bitmap` structure that provides various methods such as `union` and `intersect` to enable interoperation between the two different encodings, ensuring compatibility between the new and old versions.

In the sort create step, BitVec is changed to Roaring to optimize for the worst-case scenarios.

## PR Checklist


src/index/src/inverted_index/create/sort.rs
```diff
@@ -17,22 +17,24 @@ mod intermediate_rw;
...
+use roaring::RoaringBitmap;
 
+use crate::bitmap::Bitmap;
 use crate::inverted_index::error::Result;
+use crate::inverted_index::format::writer::ValueStream;
 use crate::{Bytes, BytesRef};
 
 /// A stream of sorted values along with their associated bitmap
-pub type SortedStream = Box<dyn Stream<Item = Result<(Bytes, BitVec)>> + Send + Unpin>;
+pub type SortedStream = Box<dyn Stream<Item = Result<(Bytes, RoaringBitmap)>> + Send + Unpin>;
```
waynexia, line 29: why it's not `Bitmap` but a concrete `RoaringBitmap`?

src/index/src/inverted_index/create/sort/intermediate_rw.rs
```diff
@@ -19,26 +19,20 @@
 //! The serialization format is as follows:
 //!
 //! ```text
-//! [magic][bitmap leading zeros][item][item]...[item]
-//!    [4]          [4]              [?]
+//! [magic][item][item]...[item]
+//!    [4]       [?]
```
waynexia, line 23: Is this change backward compatible?

src/index/src/inverted_index/create/sort.rs
zhongzc: Well, using `Bitmap` is also good, I'll change it.

Cargo.toml
```diff
@@ -129,7 +129,9 @@ etcd-client = "0.14"
 fst = "0.4.7"
 futures = "0.3"
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "072ce580502e015df1a6b03a185b60309a7c2a7a" }
+# TODO(zhongzc): wait for greptime-proto#pull/219 to be merged
```
evenyag, line 132: The proto is merged now.

src/index/src/inverted_index/format/reader.rs
```diff
@@ -67,17 +67,25 @@ pub trait InvertedIndexReader: Send + Sync {
...
     }
 
     /// Retrieves the multiple bitmaps from the given ranges.
-    async fn bitmap_deque(&mut self, ranges: &[Range<u64>]) -> Result<VecDeque<BitVec>> {
-        Ok(self
-            .read_vec(ranges)
-            .await?
+    async fn bitmap_deque(
+        &mut self,
+        ranges: &[(Range<u64>, BitmapType)],
```
evenyag, line 79: Is it possible that we use different bitmap types in the same index?

zhongzc: You are worried that this implementation is not efficient because two vecs were allocated, right?

evenyag: The method looks like that we can use different bitmap type for different ranges. But this isn't a public interface, so it should be fine.

evenyag (approved):

waynexia (approved):

Issue #5524, opened 2025-02-13
Remove `Value::DateTime` and `ValueRef::DateTime`


fengjiachun:
LGTM. Maybe we also need to remove [`Value::DateTime`](https://github.com/GreptimeTeam/greptimedb/blob/35b635f6399dc0ee33bc61d7b22d6dd76b3b5626/src/datatypes/src/value.rs#L78) but may not be included in this PR.

_Originally posted by @v0y4g3r in https://github.com/GreptimeTeam/greptimedb/pull/5506#pullrequestreview-2613903764_

evenyag:
We may also remove the `DateTimeType` and `DateTimeVector`.

linyihai:
> So actually we should remove the whole DateTime data type from the database.

If you don't mind, I can try to remove the relevant code. It might take a while. More changes than I thought

Issue #5489, opened 2025-02-07
Deprecate `DateTime` data types


v0y4g3r:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Query Engine

### Minimal reproduce step

```sql
CREATE TABLE data_types (
  s STRING,
  dtt DATETIME,
  ts9 TIMESTAMPNANOSECOND DEFAULT CURRENT_TIMESTAMP() TIME INDEX,
  PRIMARY KEY(s));

INSERT INTO data_types (s, dtt) VALUES ('test3', '2024-02-06 15:30:01');

select * from data_types;

admin flush_table('data_types');

select * from data_types;
```

### What did you expect to see?

```
mysql> CREATE TABLE data_types (
    ->   s STRING,
    ->   dtt DATETIME,
    ->   ts9 TIMESTAMPNANOSECOND DEFAULT CURRENT_TIMESTAMP() TIME INDEX,
    ->   PRIMARY KEY(s));
Query OK, 0 rows affected (0.03 sec)

mysql> INSERT INTO data_types (s, dtt) VALUES ('test3', '2024-02-06 15:30:01');
Query OK, 1 row affected (0.01 sec)

mysql> select * from data_types;
+-------+---------------------+----------------------------+
| s     | dtt                 | ts9                        |
+-------+---------------------+----------------------------+
| test3 | 2024-02-06 15:30:01 | 2025-02-07 12:27:57.806796 |
+-------+---------------------+----------------------------+
1 row in set (0.02 sec)


mysql> admin flush_table('data_types');
+---------------------------------+
| ADMIN flush_table('data_types') |
+---------------------------------+
|                               0 |
+---------------------------------+
1 row in set (0.03 sec)

mysql> select * from data_types;
+-------+---------------------+----------------------------+
| s     | dtt                 | ts9                        |
+-------+---------------------+----------------------------+
| test3 | 2024-02-06 15:30:01 | 2025-02-07 12:27:57.806796 |
+-------+---------------------+----------------------------+
1 row in set (0.02 sec)
```

### What did you see instead?

```
mysql> CREATE TABLE data_types (
    ->   s STRING,
    ->   dtt DATETIME,
    ->   ts9 TIMESTAMPNANOSECOND DEFAULT CURRENT_TIMESTAMP() TIME INDEX,
    ->   PRIMARY KEY(s));
Query OK, 0 rows affected (0.03 sec)

[...]

PR #5616, opened 2025-02-28
refactor!: Remove `Value::DateTime` and `ValueRef::DateTime`.

linyihai:

https://github.com/GreptimeTeam/greptimedb/issues/5524

Tries to removed the `Value::DateTime` and `ValueRef::DateTime`.

## PR Checklist


killme2008 (approved):
Good job! We need to update the document too. Could you please update it too? 

https://github.com/GreptimeTeam/docs/blob/main/docs/reference/sql/data-types.md#date-and-time-types-1

linyihai:
> Good job! We need to update the document too. Could you please update it too?
> 
> https://github.com/GreptimeTeam/docs/blob/main/docs/reference/sql/data-types.md#date-and-time-types-1

I'll take a look.

src/datatypes/src/vectors/helper.rs
```diff
@@ -286,7 +287,14 @@ impl Helper {
                 Arc::new(StringVector::try_from_arrow_array(array)?)
             }
             ArrowDataType::Date32 => Arc::new(DateVector::try_from_arrow_array(array)?),
-            ArrowDataType::Date64 => Arc::new(DateTimeVector::try_from_arrow_array(array)?),
+            ArrowDataType::Date64 => {
+                let array = cast(
+                    array.as_ref(),
+                    &ArrowDataType::Timestamp(TimeUnit::Millisecond, None),
+                )
+                .expect("Date64 should be castable to TimestampMillisecond");
```
evenyag, line 295: Is it safe to panic here?

src/flow/src/expr/relation/func.rs
```diff
@@ -267,7 +267,7 @@ impl AggregateFunc {
             MaxBool => (boolean_datatype, Max),
             MaxString => (string_datatype, Max),
             MaxDate => (date_datatype, Max),
-            MaxDateTime => (datetime_datatype, Max),
+            MaxDateTime => (timestamp_millisecond_datatype, Max),
```
evenyag, line 270: @discord9 We'll remove the datetime type from flow.

src/datatypes/src/data_type.rs
```diff
@@ -68,7 +68,6 @@ pub enum ConcreteDataType {
 
     // Date and time types:
     Date(DateType),
-    DateTime(DateTimeType),
```
evenyag, line 71: This will be a breaking change.

@v0y4g3r @fengjiachun Do we need to consider existing tables with DateTime type? People may not be able to open these tables anymore.

fengjiachun: Yes, this is a breaking change, but I still believe we can remove the `DateTime` type because even if users are using this type, the stored data is incorrect. We can mark this PR as `breaking`.

src/api/src/helper.rs
```diff
@@ -589,7 +584,9 @@ pub fn pb_value_to_value_ref<'a>(
             ValueRef::Timestamp(Timestamp::new_nanosecond(*t))
         }
         ValueData::TimeSecondValue(t) => ValueRef::Time(Time::new_second(*t)),
-        ValueData::TimeMillisecondValue(t) => ValueRef::Time(Time::new_millisecond(*t)),
+        ValueData::DatetimeValue(t) | ValueData::TimeMillisecondValue(t) => {
+            ValueRef::Time(Time::new_millisecond(*t))
+        }
```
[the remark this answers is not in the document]
linyihai: `ValueData::DatetimeValue(t) ` needs to be coverd as match syntax needs.

Does `ValueData::DatetimeValue` need to be discard in the `greptime-proto` firstly?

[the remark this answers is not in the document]
evenyag: We should convert `ValueData::DatetimeValue` to `ValueData::TimestampMillisecondValue` instead of `ValueData::TimeMillisecondValue`.

src/api/src/helper.rs
```diff
@@ -1234,6 +1232,10 @@ mod tests {
             ConcreteDataType::date_datatype(),
             ColumnDataTypeWrapper::date_datatype().into()
         );
+        assert_eq!(
+            ConcreteDataType::timestamp_microsecond_datatype(),
+            ColumnDataTypeWrapper::datetime_datatype().into()
+        );
```
linyihai, line 1238: > We may need to add some tests for these cases.

Added this for this case

src/flow/src/expr/relation/func.rs
evenyag, line 270:
```suggestion
            MaxDateTime => (timestamp_microsecond_datatype, Max),
```

src/flow/src/expr/relation/func.rs
```diff
@@ -283,7 +283,7 @@ impl AggregateFunc {
             MinBool => (boolean_datatype, Min),
             MinString => (string_datatype, Min),
             MinDate => (date_datatype, Min),
-            MinDateTime => (datetime_datatype, Min),
+            MinDateTime => (timestamp_millisecond_datatype, Min),
```
evenyag, line 286:
```suggestion
            MinDateTime => (timestamp_microsecond_datatype, Min),
```

src/flow/src/compute/render/map.rs
```diff
@@ -290,7 +290,9 @@ mod test {
         let mfp = MapFilterProject::new(1)
             .filter(vec![
                 ScalarExpr::Column(0)
-                    .call_unary(expr::UnaryFunc::Cast(ConcreteDataType::datetime_datatype()))
+                    .call_unary(expr::UnaryFunc::Cast(
+                        ConcreteDataType::timestamp_millisecond_datatype(),
```
evenyag, line 294:
```suggestion
                        ConcreteDataType::timestamp_microsecond_datatype(),
```

src/flow/src/compute/render/map.rs
```diff
@@ -300,7 +302,9 @@ mod test {
                         ScalarExpr::literal(4i64.into(), ConcreteDataType::int64_datatype()),
                         BinaryFunc::SubInt64,
                     )
-                    .call_unary(expr::UnaryFunc::Cast(ConcreteDataType::datetime_datatype()))
+                    .call_unary(expr::UnaryFunc::Cast(
+                        ConcreteDataType::timestamp_millisecond_datatype(),
```
evenyag, line 306:
```suggestion
                        ConcreteDataType::timestamp_microsecond_datatype(),
```

src/datatypes/src/vectors/helper.rs
```diff
@@ -180,7 +181,7 @@ impl Helper {
                 ConstantVector::new(Arc::new(DateVector::from(vec![v])), length)
             }
             ScalarValue::Date64(v) => {
-                ConstantVector::new(Arc::new(DateTimeVector::from(vec![v])), length)
+                ConstantVector::new(Arc::new(TimestampMillisecondVector::from(vec![v])), length)
```
evenyag, line 184: In datafusion, this is in millisecond. Maybe we need to convert it into microsecond or return an error.
https://docs.rs/datafusion/latest/datafusion/common/enum.ScalarValue.html#variant.Date64

src/datatypes/src/data_type.rs
```diff
@@ -441,7 +437,7 @@ impl TryFrom<&ArrowDataType> for ConcreteDataType {
             ArrowDataType::Float32 => Self::float32_datatype(),
             ArrowDataType::Float64 => Self::float64_datatype(),
             ArrowDataType::Date32 => Self::date_datatype(),
-            ArrowDataType::Date64 => Self::datetime_datatype(),
+            ArrowDataType::Date64 => Self::timestamp_millisecond_datatype(),
```
evenyag, line 440: I wonder if we should map this to millisecond or microsecond.

src/api/src/helper.rs
```diff
@@ -577,9 +573,8 @@ pub fn pb_value_to_value_ref<'a>(
         ValueData::BinaryValue(bytes) => ValueRef::Binary(bytes.as_slice()),
         ValueData::StringValue(string) => ValueRef::String(string.as_str()),
         ValueData::DateValue(d) => ValueRef::Date(Date::from(*d)),
-        ValueData::DatetimeValue(d) => ValueRef::DateTime(DateTime::new(*d)),
         ValueData::TimestampSecondValue(t) => ValueRef::Timestamp(Timestamp::new_second(*t)),
-        ValueData::TimestampMillisecondValue(t) => {
+        ValueData::DatetimeValue(t) | ValueData::TimestampMillisecondValue(t) => {
```
evenyag, line 577: Oh, we should map to microsecond according to https://github.com/GreptimeTeam/greptimedb/issues/5489

evenyag (commented):
I'm so sorry that we should map datetime type to microsecond timestamp type, not millisecond type according to https://github.com/GreptimeTeam/greptimedb/issues/5489. However, we can still replace datetime with timestamp millisecond in some places.

There is a remaining problem that arrow treats `Date64` as `timestamp millis`. I'm not sure how to handle this when we find `Date64`. Should we convert it to microsecond? This may be very confusing. should we return an error instead?

@fengjiachun @v0y4g3r

fengjiachun:
> * If we are dealing with Date64 type, we convert it between TimestampMillisecond



> > I'm so sorry that we should map datetime type to microsecond timestamp type, not millisecond type according to #5489. However, we can still replace datetime with timestamp millisecond in some places.
> > There is a remaining problem that arrow treats `Date64` as `timestamp millis`. I'm not sure how to handle this when we find `Date64`. Should we convert it to microsecond? This may be very confusing. should we return an error instead?
> > @fengjiachun @v0y4g3r
> 
> Maybe the rule is:
> 
> * If we want a Datetime SQL type or handling this type, we map it to TimestampMicrosecond
> * If we are dealing with Date64 type, we convert it between TimestampMillisecond
> * Dateime and Date64 are totally different things

1. After removing the `DateTime` type, our type no longer maps to `Date64`.
2. Tables with `DateTime` data before version 0.11 will encounter an error when parsing `Date64` during data reading. The upgrade document should clarify how to address this issue (export and reimport data?).

@evenyag cc @linyihai

evenyag:
> > * If we are dealing with Date64 type, we convert it between TimestampMillisecond
> 
> > > I'm so sorry that we should map datetime type to microsecond timestamp type, not millisecond type according to #5489. However, we can still replace datetime with timestamp millisecond in some places.
> > > There is a remaining problem that arrow treats `Date64` as `timestamp millis`. I'm not sure how to handle this when we find `Date64`. Should we convert it to microsecond? This may be very confusing. should we return an error instead?
> > > @fengjiachun @v0y4g3r
> > 
> > 
> > Maybe the rule is:
> > 
> > * If we want a Datetime SQL type or handling this type, we map it to TimestampMicrosecond
> > * If we are dealing with Date64 type, we convert it between TimestampMillisecond
> > * Dateime and Date64 are totally different things
> 
> 1. After removing the `DateTime` type, our type no longer maps to `Date64`.
> 2. Tables with `DateTime` data before version 0.11 will encounter an error when parsing `Date64` during data reading. The upgrade document should clarify how to address this issue (export and reimport data?).
> 
> @evenyag cc @linyihai

Maybe we can return an error if we need to convert Date64 to our types or arrays. What do you think? @linyihai

linyihai:
> Maybe we can return an error if we need to convert Date64 to our types or arrays. What do you think? @linyihai

It's seems reasonable. I'll have a try.

linyihai:
I'd like to replace with all `datetime` with `timestamp_microsecond`, what do you think?

src/catalog/src/system_schema/utils/tables.rs
```diff
@@ -54,7 +54,7 @@ pub fn bigint_column(name: &str) -> ColumnSchema {
 pub fn datetime_column(name: &str) -> ColumnSchema {
```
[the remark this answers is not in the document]
linyihai: >  I'd like to replace with all `datetime` with `timestamp_microsecond`, what do you think?  

Would it be timestamp_micro_column?,  I want to replace `datetime` with `timestamp_micro`

evenyag:
> I'd like to replace with all `datetime` with `timestamp_microsecond`, what do you think?

It should be fine.

src/meta-srv/src/service/admin/node_lease.rs
```diff
@@ -41,7 +41,8 @@ impl HttpHandler for NodeLeaseHandler {
             .into_iter()
             .map(|(k, v)| HumanLease {
                 name: k,
-                human_time: common_time::DateTime::new(v.timestamp_millis).to_string(),
+                human_time: common_time::Timestamp::new_microsecond(v.timestamp_millis)
```
evenyag, line 44: We have to use millisecond timestamp here as `v` is millisecond.

evenyag (commented):
We are almost done

src/meta-srv/src/service/admin/node_lease.rs
linyihai: Done. Thank you!

fengjiachun (approved):
LGTM

src/meta-srv/src/service/admin/node_lease.rs
fengjiachun: @evenyag PTAL

PR #5626, opened 2025-03-03
fix: refactor region leader state validation

WenyXu:

The region state may change when a downgrading instruction is received from the metasrv. For example:
1. A region flush is triggered (when the buffer is full).
2. The region is set to downgrading (upon receiving the downgrading instruction).
3. The flush completes, but the manifest update fails (due to an unexpected state).
While a failure to update the manifest is unexpected, a downgrading leader rejects user writes but still allows flushing the memtable and updating the manifest. This enhancement aims to reduce the number of flush failures caused by unexpected states.

This PR introduces a conditional check for when the expected state is writable, allowing the current state to be either writable or downgrading.

## PR Checklist

### DIFF
--- a/src/mito2/src/region.rs
+++ b/src/mito2/src/region.rs
@@ -36,7 +36,7 @@ use store_api::storage::RegionId;
 use crate::access_layer::AccessLayerRef;
 use crate::error::{
     FlushableRegionStateSnafu, RegionLeaderStateSnafu, RegionNotFoundSnafu, RegionTruncatedSnafu,
-    Result,
+    Result, UpdateManifestSnafu,
 };
 use crate::manifest::action::{RegionMetaAction, RegionMetaActionList};
 use crate::manifest::manager::RegionManifestManager;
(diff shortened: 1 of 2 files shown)


src/mito2/src/region.rs
```diff
@@ -371,14 +371,30 @@ impl ManifestContext {
...
-                state: current_state,
-                expect: expect_state,
-            }
-        );
+
+        // If expect_state is writable, the current state must be either writable or downgrading.
+        //
+        // A downgrading leader rejects user writes but still allows
+        // flushing the memtable and updating the manifest.
+        if expect_state == RegionLeaderState::Writable {
```
evenyag, line 379: Will that be other states?

WenyXu: resolved in [775d93c](https://github.com/GreptimeTeam/greptimedb/pull/5626/commits/775d93cbe20fba19299e69dae4321b0c6ea12f3d)

src/mito2/src/region.rs
```diff
@@ -371,14 +371,30 @@ impl ManifestContext {
...
-        ensure!(
-            current_state == RegionRoleState::Leader(expect_state),
-            RegionLeaderStateSnafu {
-                region_id: manifest.metadata.region_id,
-                state: current_state,
-                expect: expect_state,
-            }
-        );
+
+        // If expect_state is not downgrading, the current state must be either `expect_state` or downgrading.
```
fengjiachun, line 375:
```suggestion
        // If `expect_state` is not downgrading, the current state must be either `expect_state` or downgrading.
```

fengjiachun (approved):
LGTM

src/mito2/src/region.rs
```diff
@@ -371,14 +371,30 @@ impl ManifestContext {
...
-        );
+
+        // If `expect_state` is not downgrading, the current state must be either `expect_state` or downgrading.
+        //
+        // A downgrading leader rejects user writes but still allows
+        // flushing the memtable and updating the manifest.
+        if expect_state != RegionLeaderState::Downgrading {
+            ensure!(
+                current_state == RegionRoleState::Leader(expect_state)
+                    || current_state == RegionRoleState::Leader(RegionLeaderState::Downgrading),
```
evenyag, line 382: We can log this case.

evenyag (approved):

PR #5632, opened 2025-03-03
fix: prevent failover of regions to the same peer

WenyXu:

Prevent failover to the same peer for region

## PR Checklist


sunng87 (approved):

waynexia (approved):
I'm not sure if this will fix the root cause. E.g., skip checking for failover if there is only one node available.

Issue #5630, opened 2025-03-03
window_sort: Invalid range: Current batch have data on the left side of working range


evenyag:
### What type of bug is this?

Crash

### What subsystems are affected?

Standalone mode

### Minimal reproduce step

This occurs in a user deployment:
- Create a table
- Ingest some data
- Order by data by time index

The user has two timestamp column:
- The timezone for the time index `collect_time` is UTC+8
- The timezone for a timestamp column `collect_time_utc` is UTC

Alias `collect_time_utc` to `collect_time` and order by `collect_time`.

### What did you expect to see?

Not panic

### What did you see instead?

Internal error:
```
(EngineExecuteQuery): Internal error: Current batch have data on the left side of working range, something is very wrong.
This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker
; uncategorized SQLException; SQL state [HY000]; error code [1815]; (EngineExecuteQuery): Internal error: Current batch have data on the left side of working range, something is very wrong.
This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker] with root cause
```

### What operating system did you use?

Windows

### What version of GreptimeDB did you use?

0.12

### Relevant log output and stack trace

```bash
2025-03-03T07:33:42.049906Z ERROR query::window_sort: Invalid range: TimeRange { start: 1740985923991935000::Nanosecond, end: 1740986320648185001::Nanosecond } < TimeRange { start: 1741014000185750000::Nanosecond, end: 1741015120648185001::Nanosecond }
2025-03-03T07:33:42.050029Z ERROR servers::mysql::writer: Failed to handle mysql query, code: EngineExecuteQuery, error: Internal error: Current batch have data on the left side of working range, something is very wrong.
This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker, db: lig_data err=0: , at src\common\recordbatch\src\adapter.rs:291:55
1: External(0: , at src\common\recordbatch\src\adapter.rs:291:55
1: Internal("Current batch have data on the left side of working range, something is very wrong"))
2025-03-03T07:33:42.956401Z ERROR query::window_sort: Invalid range: TimeRange { start: 1740986009346439000::Nanosecond, end: 1740986320648185001::Nanosecond } < TimeRange { start: 1741014000185750000::Nanosecond, end: 1741015120648185001::Nanosecond }
[...]

PR #5634, opened 2025-03-03
fix: window sort not apply when other column alias to time index name

discord9:

#5630 

fix a bug when other columns alias to name same as time index column, wrongly applied window sort optimize when it shouldn't.
i.e.:
```sql
CREATE TABLE IF NOT EXISTS lightning (
  collect_time TIMESTAMP(9) NOT NULL,
  collect_time_utc TIMESTAMP(9) NULL,
  peak_current FLOAT NULL,
  TIME INDEX (collect_time)
)
ENGINE=mito
WITH(
  'compaction.twcs.time_window' = '7d',
  'compaction.type' = 'twcs'
);

-- this shouldn't use window sort since is not order by time index
SELECT
  collect_time_utc AS collect_time,
  peak_current,
FROM
  lightning
ORDER BY
  collect_time ASC;
```
will wrongly use `WindowSort` optimized physical plan before this fix.
Related sqlness test have another issue due to substrait encode/decode cause slightly different logical plan in standalone/distributed mode(results are correct) so sqlness test are duplicated in standalone/distributed

a better way to impl how to determine if a column is time index column is maybe optimize in `LogicalPlan` in which field's metadata contain related keys if it is time index, and add a new planner to turn `WindowSort` Logical Plan to `WindowSortExec` physical plan, but that is not the scope of this fix.

## PR Checklist


tests/cases/standalone/optimizer/order_by.result
```diff
@@ -120,9 +120,7 @@ EXPLAIN ANALYZE SELECT i, t AS alias_ts FROM test_pk ORDER BY alias_ts DESC LIMI
 +-+-+-+
 | 0_| 0_|_MergeScanExec: REDACTED
 |_|_|_|
-| 1_| 0_|_SortPreservingMergeExec: [alias_ts@1 DESC] REDACTED
-|_|_|_WindowedSortExec: expr=alias_ts@1 DESC num_ranges=1 fetch=5 REDACTED
-|_|_|_PartSortExec: expr=alias_ts@1 DESC num_ranges=1 limit=5 REDACTED
+| 1_| 0_|_SortExec: TopK(fetch=5), expr=[alias_ts@1 DESC], preserve_partitioning=[false] REDACTED
```
evenyag, line 123: Strange, this should still use window sort.

tests/cases/standalone/common/order/windowed_sort.sql
```diff
@@ -111,3 +111,52 @@ SELECT * FROM test_pk where pk > 7 ORDER BY t LIMIT 5;
...
+-- SQLNESS REPLACE region=\d+\(\d+,\s+\d+\) region=REDACTED
+EXPLAIN ANALYZE SELECT
+  collect_time_utc AS collect_time,
+  peak_current,
+FROM
+  lightning
+ORDER BY
+  collect_time ASC;
+
+DROP TABLE lightning;
```
waynexia, line 162:
```suggestion
DROP TABLE lightning;

```

evenyag (approved):

src/query/src/optimizer/windowed_sort.rs
```diff
@@ -231,3 +237,80 @@ fn remove_repartition(
         Ok(Transformed::no(plan))
     })
 }
+
+/// Resolves alias of the time index column.
+///
+/// i.e if a is time index, alias= {a:b, b:c}, then result should be {b}(not {c})
+/// if alias={b:a} and a is time index, then return empty
+fn resolver_alias(alias_map: &[(String, String)], time_index: &HashSet<String>) -> HashSet<String> {
```
evenyag, line 245:
```suggestion
fn resolve_alias(alias_map: &[(String, String)], time_index: &HashSet<String>) -> HashSet<String> {
```

src/query/src/optimizer/windowed_sort.rs
```diff
@@ -231,3 +237,80 @@ fn remove_repartition(
         Ok(Transformed::no(plan))
     })
 }
+
+/// Resolves alias of the time index column.
+///
+/// i.e if a is time index, alias= {a:b, b:c}, then result should be {b}(not {c})
```
evenyag, line 243: Looks like we should return {a, b}?

src/query/src/optimizer/windowed_sort.rs
```diff
@@ -231,3 +237,80 @@ fn remove_repartition(
...
+
+    use super::*;
+
+    #[test]
+    fn test_alias() {
+        let testcases = [
+            (
+                vec![("a", "b"), ("b", "c")],
+                HashSet::from(["a"]),
+                HashSet::from(["b"]),
```
evenyag, line 275:
```suggestion
                HashSet::from(["a", "b"]),
```

evenyag (approved):

Issue #5595, opened 2025-02-25
On opentelemetry trace data modeling


sunng87:
### What type of enhancement is this?

API improvement

### What does the enhancement do?

Trace is a key data type in OpenTelemetry, known for its efficiency and wide
adoption. Our roadmap focuses on supporting direct ingestion of OTEL tracing
data but may also consider broader plans like observability 2.0, where each span
record holds events and attributes.

## How tracing data in Otel looks like

By using otel's built-in console exporter, we can dump the tracing data
structure.

```json
    {
        "name": "SELECT",
        "context": {
            "trace_id": "0xa83a8a797828edd9fb904f0050626b8a",
            "span_id": "0xcbf9e87107730473",
            "trace_state": "[]"
        },
        "kind": "SpanKind.CLIENT",
        "parent_id": "0x64d38d765bee316b",
        "start_time": "2025-02-25T04:28:29.337988Z",
        "end_time": "2025-02-25T04:28:29.338085Z",
        "status": {
            "status_code": "UNSET"
        },
        "attributes": {
            "db.system": "sqlite",
            "db.name": "",
            "db.statement": "SELECT QUOTE(?), QUOTE(?), QUOTE(?)"
        },
        "events": [],
        "links": [],
        "resource": {
            "attributes": {
                "telemetry.sdk.language": "python",
                "telemetry.sdk.name": "opentelemetry",
                "telemetry.sdk.version": "1.30.0",
                "service.name": "myproject"
            },
            "schema_url": ""
        }
    }
    {
        "name": "POST todos/",
        "context": {
            "trace_id": "0xa83a8a797828edd9fb904f0050626b8a",
            "span_id": "0x64d38d765bee316b",
            "trace_state": "[]"
        },
        "kind": "SpanKind.SERVER",
        "parent_id": null,
        "start_time": "2025-02-25T04:28:29.335006Z",
        "end_time": "2025-02-25T04:28:29.340459Z",
        "status": {
            "status_code": "UNSET"
        },
        "attributes": {
            "http.request.method": "POST",
            "url.scheme": "http",
            "server.address": "django:8000",
            "server.port": 8000,
            "url.path": "/todos/",
            "client.address": "10.89.0.6",
            "client.port": 33788,
            "user_agent.original": "python-requests/2.32.3",
            "network.protocol.version": "1.1",
            "http.route": "todos/",
            "http.response.status_code": 201
        },
        "events": [],
        "links": [],
        "resource": {
[...]

PR #5655, opened 2025-03-06
refactor: update jaeger api implementation for new trace modeling

sunng87:

#5595 

This patch updates how we assemble jaeger API response in order to fit both legacy trace modeling and the upcoming newer one `greptime_trace_v1`.

changes included in this patch are:

1. use table option meta key `TABLE_DATA_MODEL` to mark special table models we built into database by ourselves
2. allow customization of table name for jaeger API via new http parameter `table`
3. support both `v0` and `v1` data model for jaeger API. This is done by
   1. using wildcard `*` to retrieve all columns in trace/span query
   1. parse them accordingly: if there is a `span_attributes` column, it's `v0` and we will assemble the span tags as json. Otherwise it's `v1` model and we will push all columns start with `span_attributes.` into span tags.
4. in order to get stable results, sort is added to:
   1. spans: added `order by timestamp` by default
   1. span.tags: sorted by key when assembling the results
5. using header to pass table name (or database name which is already done in middleware) for jaeger api, because jaeger's grafana data source doesn't allow us to configure more query strings.

## PR Checklist


src/servers/src/http/jaeger.rs
```diff
@@ -690,50 +663,203 @@ fn error_response(err: Error) -> (HttpStatusCode, axum::Json<JaegerAPIResponse>)
...
+                    // for v0 data model, span_attributes are nested as a json
+                    // data structure
 
-        // Set trace id.
-        if let Some(JsonValue::String(trace_id)) = row_iter.next() {
-            span.trace_id = trace_id.clone();
-            trace_id_to_processes.entry(trace_id).or_default();
-        }
+                    if let JsonValue::Object(span_attrs) = cell {
+                        span.tags = object_to_tags(span_attrs);
```
shuiyisong, line 727: Let's use `span.tags.extend()` in case if the order is not ensured.

src/servers/src/http/jaeger.rs
```diff
@@ -690,50 +663,203 @@ fn error_response(err: Error) -> (HttpStatusCode, axum::Json<JaegerAPIResponse>)
...
+                                span.logs.push(Log {
+                                    timestamp: t,
+                                    fields,
+                                });
+                            }
+                        }
+                    }
+                }
+                SCOPE_NAME_COLUMN => {
+                    if let JsonValue::String(scope_name) = cell {
```
shuiyisong, line 788: Looks like the empty check is removed, is it on purpose?
Same as below.

src/servers/src/http/jaeger.rs
sunng87: There is only one `span_attributes` in the legacy data model

sunng87:
The idea of this patch is to support both v0 and v1 data model in single query without adding additional check. It works for most cases except when the query contains tag filter, it's a little bit tricky to build the filter.

sunng87:
thinking about this for a while and there is no easy approach to use single statement for tag filters on two table models. I will need to add table attribute to mark trace table version model

src/servers/src/http/jaeger.rs
shuiyisong: https://github.com/GreptimeTeam/greptimedb/blob/c8db4b286d0bdc733e6a605416bcbcf7bfb6c919/src/servers/src/http/jaeger.rs#L826-L835

`scope_name`, `scope_version`, and some other tags are selected and pushed into the `span.tags` below, so it might have order issues if we use assignment here.

shuiyisong:
Is there a way to add table comments stating whether it's using v0 or v1?

sunng87:
@shuiyisong yes I'm working on similar mechanism to add this information in table options

src/servers/src/http/jaeger.rs
sunng87: I see. Let me fix this.

sunng87:
The refactor is done, please take a look @zyy17 @shuiyisong

zyy17 (approved):
mostly lgtm

shuiyisong (approved):

PR #5675, opened 2025-03-09
feat: time series distribution in scanner

waynexia:

Implement `TimeSeriesDistribution` in `RegionScanExec` which would output data grouped by series.

## PR Checklist


src/query/src/optimizer/scan_hint.rs
```diff
@@ -121,6 +121,30 @@ impl ScanHintRule {
...
+        let region_metadata = adapter.region_metadata();
+        let mut pk_column_iter = region_metadata.primary_key_columns();
+        let mut curr_sort_expr = sort_expr_cursor.next();
+        let mut curr_pk_col = pk_column_iter.next();
+
+        while let (Some(sort_expr), Some(pk_col)) = (curr_sort_expr, curr_pk_col) {
+            if sort_expr.name == pk_col.column_schema.name {
+                curr_sort_expr = sort_expr_cursor.next();
+            }
+            curr_pk_col = pk_column_iter.next();
```
evenyag, line 135: This hint may not be suitable for non-metric situations.
If a table has pk `a, b, c, ...` and the following rows:
```
a, b, c
1, 1, 1
1, 1, 2
1, 2, 1
1, 2, 2
```
When the sort expr is `a, c, ts`, maybe time series distribution doesn't have benefits.

src/table/src/table/scan.rs
```diff
@@ -64,14 +69,60 @@ impl RegionScanExec {
...
+                    &[LexOrdering::new(pk_columns)],
+                )
+            }
+            Some(TimeSeriesDistribution::TimeWindowed) => {
+                pk_columns.insert(0, ts_col);
+                EquivalenceProperties::new_with_orderings(
+                    arrow_schema.clone(),
+                    &[LexOrdering::new(pk_columns)],
+                )
+            }
```
evenyag, line 114: Is this ordering different from the time window distribution? Since in the same time window, we sort data in `time series, timestamp` order now. But we didn't implement this distribution so this may not be a problem.

src/query/src/optimizer/scan_hint.rs
```diff
@@ -121,6 +121,30 @@ impl ScanHintRule {
...
+        }
+
+        let next_remaining = sort_expr_cursor.next();
+        match (curr_sort_expr, next_remaining) {
+            (Some(expr), None)
+                if expr.name == region_metadata.time_index_column().column_schema.name =>
+            {
+                adapter.with_distribution(TimeSeriesDistribution::PerSeries);
```
evenyag, line 143: Do we need to modify the `parallelize_scan`? Now, it assigns partition ranges without considering the time series distribution. We may output the same time series in different partitions.

evenyag (approved):
Others LGTM

src/query/src/optimizer/scan_hint.rs
waynexia: I change the check to require order expr to be exact the same with pk

src/query/src/optimizer/scan_hint.rs
waynexia: Now it'll skip parallelizing for time series distribution

src/table/src/table/scan.rs
waynexia: I left a todo. I expect to handle this via time window distribution in the future

WenyXu (approved):

PR #5685, opened 2025-03-11
chore: add some method for log query handler

paomian:

* add some method to LogQueryHandler for some enterprise functionals.

## PR Checklist

### DIFF
--- a/src/servers/src/query_handler.rs
+++ b/src/servers/src/query_handler.rs
@@ -173,6 +173,17 @@ pub trait PipelineHandler {
 #[async_trait]
 pub trait LogQueryHandler {
     async fn query(&self, query: LogQuery, ctx: QueryContextRef) -> Result<Output>;
+    async fn get_table(
+        &self,
+        catalog: &str,
+        schema: &str,
+        table: &str,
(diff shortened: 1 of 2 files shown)


src/servers/src/query_handler.rs
```diff
@@ -173,6 +173,17 @@ pub trait PipelineHandler {
...
+        &self,
+        catalog: &str,
+        schema: &str,
+        table: &str,
+    ) -> std::result::Result<Option<Arc<table::Table>>, catalog::error::Error>;
+    async fn table_names(
+        &self,
+        catalog: &str,
+        schema: &str,
+    ) -> std::result::Result<Vec<String>, catalog::error::Error>;
```
fengjiachun, line 186: Add an empty line between methods?

killme2008:
These functions don't belong in the log query handler. They are catalog manager functions and should be called via a catalog manager reference.

waynexia:
+1. `CatalogManager` does have these methods, as well as other methods that may also be useful, like getting schema names or check if a table exists

waynexia (approved):
i thought there is a better way to get catalog manager. maybe left a todo to inspect the usage after we see the actual use case

PR #5688, opened 2025-03-11
feat: add simple extract processor

shuiyisong:

This PR introduces `simple_extract` processor, which allows the user to use a key like `a.b` to extract a field from a nested JSON object quickly. The purpose of this processor is to provide a simple yet performant way to extract JSON fields other than `json_path` processor, which is relatively complex and slow.

See the test for usage.

## PR Checklist

### DIFF
--- a/src/pipeline/src/etl/processor/simple_extract.rs
+++ b/src/pipeline/src/etl/processor/simple_extract.rs
@@ -28,7 +28,8 @@ pub(crate) const PROCESSOR_SIMPLE_EXTRACT: &str = "simple_extract";
 #[derive(Debug, Default)]
 pub struct SimpleExtractProcessor {
     fields: Fields,
-    // a.b -> [a, b]
+    /// simple keys to extract nested JSON field
+    /// key `a.b` is saved as  ['a', 'b'], each key represents a level of the JSON tree
     key: Vec<String>,
     ignore_missing: bool,
 }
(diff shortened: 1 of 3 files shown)


src/pipeline/src/etl/processor/simple_extract.rs
```diff
@@ -0,0 +1,147 @@
...
+                        .fail();
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+}
+
+mod test {
```
sunng87, line 126:
```suggestion
#[cfg(test)
mod test {
```

src/pipeline/tests/simple_extract.rs
```diff
@@ -0,0 +1,69 @@
...
+      }
+    ]
+"#;
+
+    let pipeline_yaml = r#"
+---
+processors:
+  - simple_extract:
+      field: commit, commit_author
+      key: "commitAuthor"
```
sunng87, line 54: Can we check again this UI? if I understand correctly it means we use the first field as input, apply the processor "commitAuthor" and use the second field as output. 

If this follows how we have been using "field" it's ok.

shuiyisong: Yes, it's tricky at first sight. In the `field`, the first key is the name of the input object, currently there is no way to point to the 'root context'. The second key represents the output's name; if not provided, it uses the input's name, which leads to a replacement of the original object. cc @paomian

paomian (approved):

PR #5697, opened 2025-03-12
chore: support `application/x-ndjson` for log ingest

shuiyisong:

The current `application/json`(which is also NDJSON) uses serde's stream deserializer, which stops immediately when an error is encountered, leaving all the following possible valid JSON discarded.
In the real world, the data is often in the format of lines of JSON objects. We can simply skip one line of invalid JSON string and continue parsing.
See test for example difference.

## PR Checklist

### DIFF
--- a/src/servers/src/http/event.rs
+++ b/src/servers/src/http/event.rs
@@ -43,7 +43,7 @@ use crate::error::{
     Result, UnsupportedContentTypeSnafu,
 };
 use crate::http::header::constants::GREPTIME_PIPELINE_PARAMS_HEADER;
-use crate::http::header::CONTENT_TYPE_PROTOBUF_STR;
+use crate::http::header::{CONTENT_TYPE_NDJSON_STR, CONTENT_TYPE_PROTOBUF_STR};
 use crate::http::result::greptime_manage_resp::GreptimedbManageResponse;
 use crate::http::result::greptime_result_v1::GreptimedbV1Response;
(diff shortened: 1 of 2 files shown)


sunng87 (approved):

src/servers/src/http/event.rs
```diff
@@ -571,6 +573,29 @@ fn extract_pipeline_value_by_content_type(
             Deserializer::from_str(&payload).into_iter(),
             ignore_errors,
         )?,
+        ct if ct == *NDJSON_CONTENT_TYPE => {
+            let mut result = Vec::with_capacity(1000);
+            for (index, line) in payload.lines().enumerate() {
```
sunng87, line 578: can we add a test case when there is a line breaker in a string in json

shuiyisong:
``` rust
    fn a() {
        let s1 = r#"
{"a": "a
"}
"#;
        let s2 = r#"
{"a": "a\n"}
"#;
        println!("s1: {:?}", s1.lines().collect::<Vec<&str>>());
        println!("s2: {:?}", s2.lines().collect::<Vec<&str>>());
    }
// output
s1: ["", "{\"a\": \"a", "\"}"]
s2: ["", "{\"a\": \"a\\n\"}"]
```

Unfortunately, if the line contains a __true__ line breaker, it will be treated as two lines. However if it's been escaped, it's fine.

waynexia (approved):

PR #5719, opened 2025-03-14
feat: move default data path from /tmp to current directory

### LINKED ISSUE #4947: Change the default working directory from `/tmp` to ...

User experience

Currently the default working directory for GreptimeDB is `/tmp/greptimedb`, it's hardcoded in our codes, like [here](https://github.com/GreptimeTeam/greptimedb/blob/main/src/datanode/src/config.rs#L92) and [here](https://github.com/GreptimeTeam/greptimedb/blob/main/src/common/telemetry/src/logging.rs#L114). That could cause problems for some impatient users who don't read our configuration document carefully enough:

- the `/tmp` does not exist in Windows OS;
- the contents in `/tmp` usually do not survive across system reboot;
- `/tmp` could be treated(mounted) differently than normal file system on disk, not good for getting consistent benchmarking results between users.

I think from perspective of user experience, we'd better change it from `/tmp` to some places on disk. The options I could think of are:

- `$HOME/.greptimedb`
- `./greptimedb`

I'd prefer the first one. 

### Implementation challenges

Find a place to put the default working directory const in codes.

sunng87:

Fixes #4947 

Change default data location to a relative path: `./greptimedb_data`. A relative path from current directory is almost the best location we can choose which has permission to write.

We will also update our demos and docs to wipe out content using `/tmp`.

Let's say goodbye to our /tmp culture.

## PR Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ tests-fuzz/corpus/
 # Nix
 .direnv
 .envrc
+
+## default data home
+greptimedb_data
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ Start a GreptimeDB container with:
 
 ```shell
 docker run -p 127.0.0.1:4000-4003:4000-4003 \
-  -v "$(pwd)/greptimedb:/tmp/greptimedb" \
+  -v "$(pwd)/greptimedb:./greptimedb_data" \
   --name greptime --rm \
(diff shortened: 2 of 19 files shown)


zyy17 (approved):
Great! Feel no more temporary for a database.

fengjiachun (approved):
LGTM

PR #5733, opened 2025-03-18
refactor: remove trace id in pk and use original trace / span id

zyy17:

1. Remove trace id, spane id in the primary key.

   The current memtable implementation is still not friendly for high cardinality keys. If we use trace id in the primary key, it will easy to cause OOM;

2. Use the original format trace id / span id instead of using `bytes_to_hex_string()`

   It will cause two problems:

   - When using `bytes_to_hex_string()` in trace id, the first string is always the same, and it will break the default partitions rules. Because of that, the write traffic will always write to some fixed regions.

   - User cannot run the query based on the original trace id; 

## PR Checklist


paomian (approved):

sunng87 (approved):

shuiyisong (approved):
Please fix the test and we are good to go 🚀

PR #5734, opened 2025-03-18
fix: mysql prepare limit&offset param

discord9:

fix a bug where prepare statement in mysql can't get params when param is in `LIMIT` or `OFFSET`
before:
```sql
SELECT `cake`.`domain`, `cake`.`is_expire`, `cake`.`ts` FROM `cake` WHERE `cake`.`domain` = ? LIMIT ? OFFSET ?
```
will fail to replace placeholder in limit/offset.
after this fix it can deal with placeholder in those position.

Curiously, this bug wouldn't trigger in sqlness or mysql-client, but only in `sea-orm`'s mysql backend will trigger it(__edited__: it's because both mysql-client and sqlness use `CliParams` which is different from `ProtocolParams` 's path which sea-orm use and have bugs), could related to how it set sql mode:
```sql
SET sql_mode=(SELECT CONCAT(@@sql_mode, ',PIPES_AS_CONCAT,NO_ENGINE_SUBSTITUTION')),time_zone='+00:00',NAMES …" db.statement="\n\nSET sql_mode=(SELECT CONCAT(@@sql_mode, ',PIPES_AS_CONCAT,NO_ENGINE_SUBSTITUTION')),time_zone='+00:00',NAMES utf8mb4 COLLATE utf8mb4_unicode_ci;
```

## PR Checklist


killme2008 (approved):
LGTM

src/servers/src/mysql/helper.rs
```diff
@@ -77,6 +79,41 @@ pub fn transform_placeholders(stmt: Statement) -> Statement {
...
+        }
+    };
+    let plan = plan
+        .transform(|p| {
+            let LogicalPlan::Limit(mut limit) = p else {
+                return Ok(Transformed::no(p));
+            };
+
+            if let Some(fetch) = &mut limit.fetch {
+                *fetch = Box::new(fetch.clone().transform(give_placeholder_types)?.data);
```
killme2008, line 103: Can we take the `fetch` and `skip` to avoid the clone?

src/servers/src/mysql/handler.rs
```diff
@@ -194,6 +194,7 @@ impl MysqlInstanceShim {
         };
 
         let params = if let Some(plan) = &plan {
+            let plan = fix_placeholder_types(plan.clone())?;
```
MichaelScofield, line 197: Seems the `plan` can be borrowed mutably to avoid the `clone` here.

tests/cases/standalone/common/prepare/mysql_prepare.result
```diff
@@ -70,3 +70,19 @@ DEALLOCATE stmt;
 
 affected_rows: 0
 
+-- test if placeholder at limit and offset are parsed correctly
+-- SQLNESS PROTOCOL MYSQL
+PREPARE stmt FROM 'SELECT 1 LIMIT ? OFFSET ?;';
```
MichaelScofield, line 75: It would be better if the sql actually yields some data.

tests/cases/standalone/common/prepare/mysql_prepare.sql
```diff
@@ -48,3 +48,29 @@ EXECUTE stmt USING 1, 2;
...
+PREPARE stmt FROM 'SELECT `cake`.`domain`, `cake`.`is_expire`, `cake`.`ts` FROM `cake` WHERE `cake`.`domain` = ? LIMIT ? OFFSET ?';
+
+-- SQLNESS PROTOCOL MYSQL
+EXECUTE stmt USING 'happy', 42, 0;
+
+-- SQLNESS PROTOCOL MYSQL
+DEALLOCATE stmt;
+
+-- SQLNESS PROTOCOL MYSQL
+DROP TABLE cake;
```
killme2008, line 76: Always add a new line at the end of file

killme2008 (approved):
LGTM

MichaelScofield (approved):

PR #5740, opened 2025-03-19
feat(mito): allow skipping wal while creating tables

### LINKED ISSUE #1267: Table-level WAL switch

Not all tables require durability at the cost of insert throughput. With periodical table flush, losing a short period of data in memtable is acceptable in some scenarios.

### What does the feature do?

Allow user to specify whether to skip writing WAL on data ingestion.

### Implementation challenges

Add a table option on creation, maybe `skip_wal`.
In `write_to_wal`, check if this table should skip WAL.
https://github.com/GreptimeTeam/greptimedb/blob/b31a6cb506c5fd4cee417ddb3c0780d25272b505/src/storage/src/wal.rs#L103-L113

If `skip_wal` is true, simply return the sequence number as entry id without actually writing to underlying log store.

v0y4g3r:

close #1267

Allow users to create tables with `skip_wal` set to true to disable WAL per table.

## PR Checklist

### DIFF
--- a/src/mito2/src/wal.rs
+++ b/src/mito2/src/wal.rs
@@ -24,6 +24,7 @@ use std::sync::Arc;
 
 use api::v1::WalEntry;
 use common_error::ext::BoxedError;
+use common_telemetry::debug;
 use entry_reader::NoopEntryReader;
 use futures::future::BoxFuture;
 use futures::stream::BoxStream;
@@ -89,6 +90,7 @@ impl<S: LogStore> Wal<S> {
         let store = self.store.clone();
         move |region_id, last_entry_id, provider| -> BoxFuture<'_, Result<()>> {
             if let Provider::Noop = provider {
+                debug!("Skip obsolete for region: {}", region_id);
                 return Box::pin(async move { Ok(()) });
             }
             Box::pin(async move {
--- a/src/mito2/src/wal.rs
+++ b/src/mito2/src/wal.rs
@@ -24,6 +24,7 @@ use std::sync::Arc;
 
 use api::v1::WalEntry;
 use common_error::ext::BoxedError;
+use entry_reader::NoopEntryReader;
 use futures::future::BoxFuture;
 use futures::stream::BoxStream;
 use prost::Message;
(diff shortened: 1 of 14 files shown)


src/table/src/requests.rs
```diff
@@ -129,6 +134,16 @@ impl TableOptions {
             options.ttl = Some(ttl_value);
         }
 
+        if let Some(skip_wal) = kvs.get(SKIP_WAL_KEY) {
+            options.skip_wal = skip_wal.parse().unwrap_or_else(|_| {
+                warn!(
```
MichaelScofield, line 139: Better to return some error to user, otherwise it might cause surprise.

MichaelScofield (approved):

src/mito2/src/wal.rs
```diff
@@ -87,6 +88,9 @@ impl<S: LogStore> Wal<S> {
     ) -> impl FnOnce(RegionId, EntryId, &Provider) -> BoxFuture<Result<()>> {
         let store = self.store.clone();
         move |region_id, last_entry_id, provider| -> BoxFuture<'_, Result<()>> {
+            if let Provider::Noop = provider {
+                return Box::pin(async move { Ok(()) });
```
killme2008, line 92: It's better to log when skipping WAL.

killme2008 (approved):
LGTM.

I think we can add a sqlness CRUD test for skippping WAL table.
Because we'll skip the wal, should call `admin flush_table` like that.

src/mito2/src/worker/handle_write.rs
```diff
@@ -75,13 +76,21 @@ impl<S: LogStore> RegionWorkerLoop<S> {
                 .start_timer();
             let mut wal_writer = self.wal.writer();
             for region_ctx in region_ctxs.values_mut() {
+                if let WalOptions::Noop = &region_ctx.version().options.wal_options {
```
evenyag, line 79: Do we need to add a unit test that the wal is disabled?

src/table/src/requests.rs
v0y4g3r: fixed in https://github.com/GreptimeTeam/greptimedb/pull/5740/commits/2c067d85a27a9076a7224e15c1e747e2aa957d0f#diff-0d62a3ecad039e20e5695ee9c5e911179c7da14ec7e963f75394e23b7b34f080R137

src/mito2/src/worker/handle_write.rs
v0y4g3r: Added a sqlness case for that https://github.com/GreptimeTeam/greptimedb/pull/5740/commits/2c067d85a27a9076a7224e15c1e747e2aa957d0f#diff-0d620b4ddaeef2e408d31f800ca10cf11d8c64aa9884772cdaba5cf46d20c911

PR #5741, opened 2025-03-19
refactor: construct `show index` results through table schema

zhongzc:

Construct the `show index` results in a more direct way.

## PR Checklist

### DIFF
--- a/src/query/src/sql.rs
+++ b/src/query/src/sql.rs
@@ -432,7 +432,7 @@ pub async fn show_index(
     let mut index_types = StringVectorBuilder::with_capacity(capacity);
     let mut comments = StringVectorBuilder::with_capacity(capacity);
     let mut index_comments = StringVectorBuilder::with_capacity(capacity);
-    let mut visibles = StringVectorBuilder::with_capacity(capacity);
+    let mut visible = StringVectorBuilder::with_capacity(capacity);
     let mut expressions = StringVectorBuilder::with_capacity(capacity);
 
     let mut append_index_column =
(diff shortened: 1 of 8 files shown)


src/query/src/sql.rs
```diff
@@ -403,102 +401,140 @@ pub async fn show_index(
...
+            tables.push(Some(table_name));
+            non_uniques.push(Some(1));
+            key_names.push(Some(key_name));
+            seq_in_indices.push(Some(1));
+            column_names.push(Some(column_name));
+            collations.push(if sorted { Some("A") } else { None });
+            cardinalities.push(None);
+            sub_parts.push(None);
+            packeds.push(None);
+            nullables.push(Some(if nullable { YES_STR } else { "" }));
```
zhongzc, line 449: image

killme2008:
Why do we need to refactor in this way? In fact, `show` statement is just an extension for querying `information_schema`.

zhongzc:
> Why do we need to refactor in this way? In fact, `show` statement is just an extension for querying `information_schema`.

Some column schema information, I did not find a suitable placement in the `key_column_usage`, so it is difficult to retrieve from the `key_column_usage.

killme2008:
> > Why do we need to refactor in this way? In fact, `show` statement is just an extension for querying `information_schema`.
> 
> Some column schema information, I did not find a suitable placement in the `key_column_usage`, so it is difficult to retrieve from the `key_column_usage.

We can extend the table columns just like `partitions. greptime_partition_id `

https://docs.greptime.com/reference/sql/information-schema/partitions

PR #5743, opened 2025-03-19
refactor: add `partition_rules_for_uuid()`

zyy17:

Add a function `partition_rules_for_uuid()` to create partition rules based on the expected partition number when using otlp traces.

It will make the partition rules less confusing, and we may configure the partition number by top-level hint in the future.

## PR Checklist


sunng87 (approved):

src/sql/src/partition.rs
```diff
@@ -38,117 +40,105 @@ macro_rules! between_string {
...
-            between_string!(ident_expr, "e", "f")
-        ),
-        or!(
-            between_string!(ident_expr, "F", "a"),
-            Expr::BinaryOp {
+// partition_rules_for_uuid can creates partition rules up to 256 partitions.
+fn partition_rules_for_uuid(partition_num: u32, ident: &str) -> Vec<Expr> {
+    let ident_expr = Expr::Identifier(Ident::new(ident).clone());
+
+    let total_partitions = 256;
```
sunng87, line 54: I think we can go one step further by removing limit cap of 256. We can find the smallest total partition number of 16*n and generate the boundaries.

src/sql/src/partition.rs
```diff
@@ -38,117 +40,105 @@ macro_rules! between_string {
...
+            "trace_id >= '60' AND trace_id < '70'",
+            "trace_id >= '70' AND trace_id < '80'",
+            "trace_id >= '80' AND trace_id < '90'",
+            "trace_id >= '90' AND trace_id < 'a0'",
+            "trace_id >= 'a0' AND trace_id < 'b0'",
+            "trace_id >= 'b0' AND trace_id < 'c0'",
+            "trace_id >= 'c0' AND trace_id < 'd0'",
+            "trace_id >= 'd0' AND trace_id < 'e0'",
+            "trace_id >= 'e0' AND trace_id < 'f0'",
+            "trace_id >= 'f0'",
```
waynexia, line 141: we need to handle big letter and small letter together like previous rule

src/sql/src/partition.rs
zyy17: Maybe up to 65536? For example:

```
[2,16]: use the first one hex char to partition;
[16, 256]: use the first two hex chars to partition;
...
[4097, 65536]
```

Maybe it's enough for a very, very long time.

src/sql/src/partition.rs
zyy17: Is it necessary to handle partition rules? It makes things too complicated. Most UUIDs are lowercase. Can we change all the UUIDs to lowercase when applying the rules? Actually, in the OLTP trace scenario, we will use https://github.com/GreptimeTeam/greptimedb/blob/main/src/servers/src/otlp/utils.rs#L21-L23 to convert the random bytes to lower-hex strings.

zyy17: @sunng87 @waynexia I found out the partition rules will be complicated(typically, it's hard to program) and hard to read with an increase in partition number:

In opentelemetry trace spec:

1. TraceId will be a 16 random bytes;
2. When retrieving the trace id, it will return **lowercase** hex-encoded format;

Since we always handle lowercase hex, why wouldn't we simplify it?

src/sql/src/partition.rs
zyy17: Also, I add a constraint: the partition number should be the order of `2`(https://github.com/GreptimeTeam/greptimedb/pull/5743/files#diff-f01f6e55255c1b45965db4f395a14ff3d1b7a50a0c1d8b8163b88e066ee1f1caR59).

Then we can create 2,4,8,16,32,...,65536 partitions by this API.

killme2008:
@waynexia PTAL

src/sql/src/partition.rs
sunng87: https://opentelemetry.io/docs/specs/otel/trace/api/#retrieving-the-traceid-and-spanid

It makes to me. We can support lowercase only.

zyy17:
After some tests, I still thought it was very ugly and complicated to create partition rules that consider upper and lower case at the same time. Also, it was not the case in the current OTEL implementation(always lower case).

To make things easier, I would strongly suggest considering the UUID in lower case.

waynexia (approved):

PR #5748, opened 2025-03-20
fix: wrap table name with ``

CookiePieWw:

As title.

## PR Checklist

### DIFF
--- a/src/common/meta/src/kv_backend/rds/mysql.rs
+++ b/src/common/meta/src/kv_backend/rds/mysql.rs
@@ -158,18 +158,18 @@ impl<'a> MySqlTemplateFactory<'a> {
                 "CREATE TABLE IF NOT EXISTS `{table_name}`(k VARBINARY(3072) PRIMARY KEY, v BLOB);",
             ),
             range_template: RangeTemplate {
-                point: format!("SELECT k, v FROM {table_name} WHERE k = ?"),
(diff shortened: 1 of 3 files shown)


yihong0618:
cool, Had this problem of not being able to delete the table, thank you

yihong0618:
> cool, Had this problem of not being able to delete the table, thank you

@CookiePieWw FYI we can not use \`tablename\` like mysql

the error I met is I created a table using mysql client with  \`tablename\`  and try to delete it in pg clinet

CookiePieWw:
@yihong0618 Thanks for your reminder! PostgreSQL uses "" instead.

But I'm not sure if the problem you encountered has something to do with the patch, since the table name of kvbackend only effects distributed version of GreptimeDB with PostgreSQL/MySQL as a kvbackend (instead of etcd).

Your description seems to refers to SQL dialect and client-server communications in `server` crate of GreptimeDB.

yihong0618:
> @yihong0618 Thanks for your reminder! PostgreSQL uses "" instead.
> 
> But I'm not sure if the problem you encountered has something to do with the patch, since the table name of kvbackend only effects distributed version of GreptimeDB with PostgreSQL/MySQL as a kvbackend (instead of etcd).
> 
> Your description seems to refers to SQL dialect and client-server communications in `server` crate of GreptimeDB.

next time when I have this kind of problem will open an issue~

WenyXu (approved):

PR #5754, opened 2025-03-21
refactor: make frontend instance clear

fengys1996:

### Motivation

Currently, it looks like there might be a circular reference like this: [Instance](https://github.com/GreptimeTeam/greptimedb/blob/a0739a96e418147129528e5eff8a689b309a6f9d/src/frontend/src/instance.rs#L116) -> [Servers](https://github.com/GreptimeTeam/greptimedb/blob/a0739a96e418147129528e5eff8a689b309a6f9d/src/frontend/src/instance.rs#L123) -> [Instance](https://github.com/GreptimeTeam/greptimedb/blob/a0739a96e418147129528e5eff8a689b309a6f9d/src/frontend/src/instance.rs#L116). However, digging deeper into the code, find that we maintain two instances, one instance contains all servers, and the other instance does not have any servers. The instance without servers is passed to the servers, so the circular reference is eliminated, but it is confusing. The related code see [here](https://github.com/GreptimeTeam/greptimedb/blob/main/src/cmd/src/frontend.rs#L375-L381).

### What has changed 

1. Added a new structure, `Frontend`, which is responsible for starting and stopping services. It contains `Instance`, `Servers`, and some background tasks.
2. Remove `servers` and some background tasks from `Instance` struct. The `Instance` has only become the implementation of various `xxx handler `traits.
3. Finally `Frontend -> Servers -> Instance`, it make sense.

## PR Checklist

### DIFF
--- a/src/cmd/src/standalone.rs
+++ b/src/cmd/src/standalone.rs
@@ -291,7 +291,7 @@ impl App for StandaloneApp {
 
     async fn stop(&self) -> Result<()> {
         self.frontend
-            .stop()
+            .shutdown()
             .await
             .context(error::StartFrontendSnafu)?;
 
@@ -621,16 +621,15 @@ impl StartCommand {
 
         let (tx, _rx) = broadcast::channel(1);
 
-        let servers = Services::new(opts.clone(), fe_instance.clone(), plugins.clone())
+        let export_metrics_task = ExportMetricsTask::try_new(&opts.export_metrics, Some(&plugins))
+            .context(error::ServersSnafu)?;
+
+        let servers = Services::new(opts, fe_instance.clone(), plugins)
             .build()
             .await
             .context(error::StartFrontendSnafu)?;
 
-        let export_metrics_task = ExportMetricsTask::try_new(&opts.export_metrics, Some(&plugins))
-            .context(error::ServersSnafu)?;
-
         let frontend = Frontend {
-            opts: fe_opts,
             instance: fe_instance,
             servers,
             heartbeat_task: None,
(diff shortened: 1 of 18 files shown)


fengys1996:
This pr also remove the `FrontendIntance` trait, please take a look @shuiyisong @MichaelScofield

MichaelScofield (approved):

src/cmd/src/standalone.rs
```diff
@@ -281,39 +266,45 @@ impl App for Instance {
...
+
         Ok(())
     }
 
     async fn stop(&self) -> Result<()> {
         self.frontend
             .shutdown()
             .await
-            .context(ShutdownFrontendSnafu)?;
+            .context(error::StartFrontendSnafu)?;
```
shuiyisong, line 296: It should be shutdown snafu.

src/cmd/src/standalone.rs
```diff
@@ -281,39 +266,45 @@ impl App for Instance {
...
-        self.frontend.start().await.context(StartFrontendSnafu)?;
         self.flow_worker_manager
             .clone()
             .run_background(Some(self.flow_shutdown.subscribe()));
+
+        self.frontend
+            .start()
+            .await
+            .context(error::StartFrontendSnafu)?;
+
```
shuiyisong, line 288: Should we start the frontend and then start the flow worker?

src/cmd/src/standalone.rs
fengys1996: Good catch!

shuiyisong (approved):

PR #5757, opened 2025-03-21
feat: add/correct some kafka-related metrics

waynexia:

- Use actual used data structure's size instead of allocated memory size (`.len()` vs. `.capacity()`)
- Add per-partition traffic and workload metrics

## PR Checklist

### DIFF
--- a/src/log-store/src/kafka/producer.rs
+++ b/src/log-store/src/kafka/producer.rs
@@ -24,7 +24,10 @@ use tokio::sync::mpsc::{self, Receiver, Sender};
 use crate::error::{self, Result};
 use crate::kafka::index::IndexCollector;
 use crate::kafka::worker::{BackgroundProducerWorker, ProduceResultHandle, WorkerRequest};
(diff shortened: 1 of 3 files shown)


sunng87 (approved):

WenyXu (approved):

WenyXu:
Should we need to add metrics to measure record-producing latency?

sunng87:
Yes, we'd better add per-partition latency stat

src/log-store/src/kafka/producer.rs
```diff
@@ -106,6 +107,15 @@ impl ProducerClient for PartitionClient {
         records: Vec<Record>,
         compression: Compression,
     ) -> rskafka::client::error::Result<Vec<i64>> {
+        let total_size = records.iter().map(|r| r.approximate_size()).sum::<usize>();
+        let partition = self.partition().to_string();
```
killme2008, line 111: Adding a `&str` return to `partition` might be efficient due to string creation for metrics.

waynexia: `.partition()` returns an i32, it has to be formatted to a string

sunng87 (approved):

PR #5775, opened 2025-03-26
feat(pipeline): support table name suffix templating in pipeline

### LINKED ISSUE #5755: Ingest logs to different tables according to log input field

There's a common use cases in log ingestion that logs collected from different application/cluster should be ingested to different tables, so a common practice is to use app name/cluster name as a part of table name.

Image we have two logs:
```
app_name=hbase, timestamp=1739929347000, message="...."
app_name=redis, timestamp=1739929347000, message="...."
```

Then I would need the table name in pipeline configuration that specifies `table = app_logs_${app_name}` so the first log goes to `app_logs_hbase` and the second `app_logs_redis`

### What does the feature do?

Ingest logs from different applications to different tables.

### Implementation challenges

_No response_

shuiyisong:

close #5755

This PR adds support to the config table name suffix template in the pipeline config. For example, you can now use the following config to add a suffix to the table name, so that the input data from log ingest API can be written to multiple tables. The variable inside `${}` should exist in the pipeline context; if any error occurs during runtime execution, no suffix will be appended to the table name.
```YAML
transform:
  - fields:
      - type
      - log
      - logger
    type: string
  - field: time
    type: time
    index: timestamp
tablesuffix: _${type}
```

## PR Checklist

### DIFF
--- a/src/pipeline/src/etl.rs
+++ b/src/pipeline/src/etl.rs
@@ -29,6 +29,7 @@ use crate::dispatcher::{Dispatcher, Rule};
 use crate::error::{
     IntermediateKeyIndexSnafu, PrepareValueMustBeObjectSnafu, Result, YamlLoadSnafu, YamlParseSnafu,
 };
+use crate::tablename::TableNameTemplate;
 use crate::GreptimeTransformer;
 
 const DESCRIPTION: &str = "description";
@@ -36,6 +37,7 @@ const PROCESSORS: &str = "processors";
 const TRANSFORM: &str = "transform";
 const TRANSFORMS: &str = "transforms";
 const DISPATCHER: &str = "dispatcher";
+const TABLENAME: &str = "tablename";
 
 pub type PipelineMap = std::collections::BTreeMap<String, Value>;
 
@@ -76,11 +78,18 @@ pub fn parse(input: &Content) -> Result<Pipeline> {
                 None
             };
 
+            let tablename = if !doc[TABLENAME].is_badvalue() {
+                Some(TableNameTemplate::try_from(&doc[TABLENAME])?)
+            } else {
+                None
+            };
+
             Ok(Pipeline {
                 description,
                 processors,
                 transformer,
                 dispatcher,
+                tablename,
             })
         }
         Content::Json(_) => unimplemented!(),
@@ -93,6 +102,7 @@ pub struct Pipeline {
     processors: processor::Processors,
     dispatcher: Option<Dispatcher>,
     transformer: GreptimeTransformer,
+    tablename: Option<TableNameTemplate>,
 }
 
 /// Where the pipeline executed is dispatched to, with context information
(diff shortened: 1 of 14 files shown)


src/pipeline/src/tablename.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+impl TableNameTemplate {
+    pub fn apply(&self, val: &PipelineMap) -> Option<String> {
+        let values = self
+            .keys
+            .iter()
+            .filter_map(|key| {
+                let v = val.get(key)?;
+                match v {
```
paomian, line 54: Should we add boolean conversion?

src/servers/src/pipeline.rs
```diff
@@ -134,8 +136,12 @@ async fn run_custom_pipeline(
             .context(PipelineSnafu)?;
 
         match r {
-            PipelineExecOutput::Transformed(row) => {
-                transformed.push(row);
+            PipelineExecOutput::Transformed((row, table_name)) => {
+                let table_name = table_name.unwrap_or(default_table.clone());
+                req_map
+                    .entry(table_name)
+                    .or_insert_with(|| Vec::with_capacity(arr_len))
```
paomian, line 143: There may be a waste of memory here, but there is currently no way to determine the distribution of the data, so there's not much to do here at the moment, maybe in the future RowInsertRequest could accept an arbitrary type that implements IntoIterator. It would be better to use a chained table type

paomian: @sunng87

paomian:
rest lgtm

src/pipeline/src/tablename.rs
shuiyisong: I think it's strange for a table name to be something like `normal_name_true`. We can add it later if there is a request.

shuiyisong:
Changed the modification to append table suffix. PTAL @paomian @sunng87

src/pipeline/src/etl.rs
```diff
@@ -29,13 +29,15 @@ use crate::dispatcher::{Dispatcher, Rule};
...
 };
+use crate::tablesuffix::TableSuffixTemplate;
 use crate::GreptimeTransformer;
 
 const DESCRIPTION: &str = "description";
 const PROCESSORS: &str = "processors";
 const TRANSFORM: &str = "transform";
 const TRANSFORMS: &str = "transforms";
 const DISPATCHER: &str = "dispatcher";
+const TABLESUFFIX: &str = "tablesuffix";
```
sunng87, line 40: we can use `table_suffix` for consistency with dispatcher

src/servers/src/pipeline.rs
sunng87: Here we trade space for time, I think it's fine. Doing one-time allocation is better.

paomian (approved):

sunng87 (approved):

PR #5777, opened 2025-03-26
chore: upgrade some dependencies

fengys1996:

This pr mainly upgrade some dependencies.

It is worth noting that rand 0.9 has a lot of API changes compared to the 0.8. The detail see [here](https://github.com/rust-random/rand/releases/tag/0.9.0).

## PR Checklist


Cargo.toml
```diff
@@ -176,15 +176,15 @@ reqwest = { version = "0.12", default-features = false, features = [
...
+rstest = "0.25"
 rstest_reuse = "0.7"
 rust_decimal = "1.33"
 rustc-hash = "2.0"
 rustls = { version = "0.23.20", default-features = false } # override by patch, see [patch.crates-io]
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", features = ["float_roundtrip"] }
 serde_with = "3"
-shadow-rs = "0.38"
+shadow-rs = "1.1"
```
sunng87, line 187: by the way, is there any new feature in shadown-rs that can improve our compilation speed for common-version?

MichaelScofield (approved):

Cargo.toml
fengys1996: The shadown-rs has three BuildPattern. The BuildPattern enum defines strategies for triggering package rebuilding. Default mode is `Lazy`.
- Lazy: The lazy mode. In this mode, if the current Rust environment is set to debug, the rebuild package will not run every time the build script is triggered. If the environment is set to release, it behaves the same as the RealTime mode.
- RealTime: The real-time mode. It will always trigger rebuilding a package upon any change, regardless of whether the Rust environment is set to debug or release.
- Custom: The custom build mode, an enhanced version of RealTime mode, allowing for user-defined conditions to trigger rebuilding a package.

We use Lazy BuildPattern in GreptimeDB, the related code see [here](https://github.com/GreptimeTeam/greptimedb/blob/e107bd5529ad420c7c70c61be9d4b99f0daa37f2/src/common/version/build.rs#L39).

So theoretically this problem should be solved.

shuiyisong (approved):
🚀

PR #5791, opened 2025-03-28
refactor: add time range for jager get operations API

zyy17:

The native Jaeger query API doesn't have a parameter to limit the time range for getting operations(span) for a specific service. When the trace table gets larger, its performance will be terrible.

In this PR, I add the http header `x-greptime-jaeger-time-range-for-operations` to add a time range explicitly. For example:

```
x-greptime-jaeger-time-range-for-operations: 3 days
```

We don't add the time range by default and let the user configure it based on their scenarios.

## PR Checklist

### DIFF
--- a/src/frontend/src/instance/jaeger.rs
+++ b/src/frontend/src/instance/jaeger.rs
@@ -60,9 +60,10 @@ impl JaegerQueryHandler for Instance {
             self.query_engine(),
             vec![col(SERVICE_NAME_COLUMN)],
             vec![],
+            vec![],
             Some(DEFAULT_LIMIT),
             None,
-            true,
+            vec![col(SERVICE_NAME_COLUMN)],
         )
         .await?)
     }
@@ -72,6 +73,8 @@ impl JaegerQueryHandler for Instance {
         ctx: QueryContextRef,
         service_name: &str,
         span_kind: Option<&str>,
+        start_time: Option<i64>,
+        end_time: Option<i64>,
     ) -> ServerResult<Output> {
         let mut filters = vec![col(SERVICE_NAME_COLUMN).eq(lit(service_name))];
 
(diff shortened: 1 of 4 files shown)


src/frontend/src/instance/jaeger.rs
```diff
@@ -244,13 +261,19 @@ async fn query_trace_table(
...
-            .sort_by(vec![col(TIMESTAMP_COLUMN)])
+            .distinct_on(distincts.clone(), distincts, None)
             .context(DataFusionSnafu)?
+    } else {
+        dataframe
+    };
+
+    // Apply the sorts if needed.
+    let dataframe = if !sorts.is_empty() {
+        dataframe.sort_by(sorts).context(DataFusionSnafu)?
```
shuiyisong, line 274: `sort_by` uses the `asc` order on default. Is it intentional?

shuiyisong (approved):
Rest LGTM

src/frontend/src/instance/jaeger.rs
killme2008: +1

src/servers/src/http/jaeger.rs
```diff
@@ -49,6 +50,7 @@ pub const JAEGER_QUERY_TABLE_NAME_KEY: &str = "jaeger_query_table_name";
 
 const REF_TYPE_CHILD_OF: &str = "CHILD_OF";
 const SPAN_KIND_TIME_FMTS: [&str; 2] = ["%Y-%m-%d %H:%M:%S%.6f%z", "%Y-%m-%d %H:%M:%S%.9f%z"];
+const JAEGER_TIME_RANGE_FOR_OPERATIONS_HEADER: &str = "x-greptime-jaeger-time-range-for-operations";
```
evenyag, line 53: The header may be too long. Is it possible to use a shorter header? e.g. `x-greptime-jaeger-query-time-range`

sunng87 (approved):
LGTM. And I think we can use a generic header to limit time range for all metadata APIs.

Issue #5789, opened 2025-03-27
Overwhelming pending DDL tasks crashes metasrv and etcd


waynexia:
### What type of bug is this?

Crash

### What subsystems are affected?

Metasrv

### Minimal reproduce step

When ingest with protocol supports auto create table, but that table can't be created for some reason, metasrv will store a huge number of pending DDL tasks and crash 

### What did you expect to see?

Metasrv can protect itself as well as backend etcd from crashing.

### What did you see instead?

as above

### What operating system did you use?

unrelated

### What version of GreptimeDB did you use?

v0.13

### Relevant log output and stack trace

```bash

```

WenyXu:
implemented in #5793

PR #5793, opened 2025-03-28
feat: add limit for the number of running procedures

WenyXu:

#5789

If we do not impose a limit on the number of procedures, certain scenarios may lead to an uncontrolled increase in their count. **e.g.,** When the metric engine is enabled with remote WAL, if remote WAL becomes unavailable, the upper layer continues to return a "storage unavailable" error. In procedures, such errors trigger retries. Since table creation keeps failing, write operations continuously trigger the auto-create table procedure. Eventually, this leads to metasrv running out of memory.

This PR  
- Adds a limit on the number of running procedures.
- Implements ErrorExt::status_code for log-store to allow fine-grained control over error status codes.

## PR Checklist

### DIFF
--- a/src/common/procedure/src/local.rs
+++ b/src/common/procedure/src/local.rs
@@ -15,7 +15,7 @@
 mod runner;
 mod rwlock;
 
-use std::collections::{HashMap, VecDeque};
+use std::collections::{HashMap, HashSet, VecDeque};
 use std::sync::atomic::{AtomicBool, AtomicI64, Ordering};
 use std::sync::{Arc, Mutex, RwLock};
 use std::time::{Duration, Instant};
@@ -33,6 +33,7 @@ use self::rwlock::KeyRwLock;
 use crate::error::{
     self, DuplicateProcedureSnafu, Error, LoaderConflictSnafu, ManagerNotStartSnafu, Result,
     StartRemoveOutdatedMetaTaskSnafu, StopRemoveOutdatedMetaTaskSnafu,
+    TooManyRunningProceduresSnafu,
 };
 use crate::local::runner::Runner;
 use crate::procedure::{BoxedProcedureLoader, InitProcedureState, ProcedureInfo};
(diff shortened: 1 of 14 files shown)


src/common/procedure/src/local.rs
```diff
@@ -210,19 +209,28 @@ impl ManagerContext {
...
+
     /// Try to insert the `procedure` to the context if there is no procedure
     /// with same [ProcedureId].
     ///
     /// Returns `false` if there is already a procedure using the same [ProcedureId].
     fn try_insert_procedure(&self, meta: ProcedureMetaRef) -> bool {
+        let procedure_id = meta.id;
         let mut procedures = self.procedures.write().unwrap();
-        if procedures.contains_key(&meta.id) {
+        if procedures.contains_key(&procedure_id) {
```
fengjiachun, line 224: The `insert` method returns None to indicate that the key is not present, so we don't need to call `contains_key`.

fengjiachun (approved):
LGTM

src/common/procedure/src/local/runner.rs
```diff
@@ -348,24 +348,14 @@ impl Runner {
...
-        if let Some(loaded_procedure) = self.manager_ctx.load_one_procedure(procedure_id) {
-            // Try to load procedure state from the message to avoid re-run the subprocedure
-            // from initial state.
-            assert_eq!(self.meta.id, loaded_procedure.parent_id.unwrap());
-
-            // Use the dumped procedure from the procedure store.
-            procedure = loaded_procedure.procedure;
-            // Update step number.
-            step = loaded_procedure.step;
-        }
```
evenyag, line 368: Why do we remove this?

src/common/procedure/src/local.rs
```diff
@@ -492,6 +487,13 @@ impl LocalManager {
 
         let watcher = meta.state_receiver.clone();
 
+        ensure!(
+            self.manager_ctx.num_running_procedures() < self.config.max_running_procedures,
+            TooManyRunningProceduresSnafu {
+                max_running_procedures: self.config.max_running_procedures,
+            }
+        );
```
evenyag, line 495: When will we have so many running procedures?

src/common/procedure/src/local/runner.rs
WenyXu: It seems like it's not used anywhere.

src/common/procedure/src/local.rs
WenyXu: When the metric engine is enabled with remote WAL, if remote WAL becomes unavailable, the upper layer continues to return a "storage unavailable" error. In procedures, such errors trigger retries. Since automatic table creation keeps failing, write operations continuously trigger the auto-create table procedure. Eventually, this leads to metasrv running out of memory

evenyag (approved):

PR #5794, opened 2025-03-28
perf: introduce `simd_json` for parsing ndjson

shuiyisong:

As the title suggests. For reference, to load 10 x 100M data under the JSONBench scene
- before: 49,502 ms
- after: 37,620 ms

Note ⚠️: For positive integer numbers, `serde_json` will try to convert them into `int64` while `simd_json` will try to convert them into `uint64`. In most cases, we think it's ok.

## PR Checklist

### DIFF
--- a/src/pipeline/src/etl.rs
+++ b/src/pipeline/src/etl.rs
@@ -160,6 +160,23 @@ pub fn json_array_to_map(val: Vec<serde_json::Value>) -> Result<Vec<PipelineMap>
     val.into_iter().map(json_to_map).collect()
 }
 
+pub fn simd_json_to_map(val: simd_json::OwnedValue) -> Result<PipelineMap> {
+    match val {
+        simd_json::OwnedValue::Object(map) => {
+            let mut intermediate_state = PipelineMap::new();
+            for (k, v) in map.into_iter() {
+                intermediate_state.insert(k, Value::try_from(v)?);
+            }
+            Ok(intermediate_state)
+        }
+        _ => PrepareValueMustBeObjectSnafu.fail(),
+    }
+}
+
+pub fn simd_json_array_to_map(val: Vec<simd_json::OwnedValue>) -> Result<Vec<PipelineMap>> {
+    val.into_iter().map(simd_json_to_map).collect()
+}
+
 impl Pipeline {
     pub fn exec_mut(&self, val: &mut PipelineMap) -> Result<PipelineExecOutput> {
         for processor in self.processors.iter() {
--- a/src/pipeline/src/etl/value.rs
+++ b/src/pipeline/src/etl/value.rs
@@ -367,6 +367,37 @@ impl std::fmt::Display for Value {
     }
 }
 
+impl TryFrom<simd_json::value::OwnedValue> for Value {
+    type Error = Error;
+
+    fn try_from(v: simd_json::value::OwnedValue) -> Result<Self> {
+        match v {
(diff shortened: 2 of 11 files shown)


src/pipeline/src/etl.rs
```diff
@@ -160,6 +160,23 @@ pub fn json_array_to_map(val: Vec<serde_json::Value>) -> Result<Vec<PipelineMap>
     val.into_iter().map(json_to_map).collect()
 }
 
+pub fn simd_json_to_map(val: simd_json::OwnedValue) -> Result<PipelineMap> {
+    match val {
+        simd_json::OwnedValue::Object(map) => {
+            let mut intermediate_state = PipelineMap::new();
```
MichaelScofield, line 166:
```suggestion
            let mut intermediate_state = PipelineMap::with_capacity(map.len());
```

src/pipeline/src/etl/value.rs
```diff
@@ -367,6 +367,37 @@ impl std::fmt::Display for Value {
...
+            simd_json::OwnedValue::String(s) => Ok(Value::String(s.to_string())),
+            simd_json::OwnedValue::Array(values) => {
+                let mut re = Vec::with_capacity(values.len());
+                for v in values.into_iter() {
+                    re.push(Value::try_from(v)?);
+                }
+                Ok(Value::Array(Array { values: re }))
+            }
+            simd_json::OwnedValue::Object(map) => {
+                let mut values = PipelineMap::new();
```
MichaelScofield, line 391:
```suggestion
                let mut values = PipelineMap::with_capacity(map.len());
```

src/servers/src/http/event.rs
```diff
@@ -602,7 +606,8 @@ fn extract_pipeline_value_by_content_type(
                     }
                 };
 
-                if let Ok(v) = serde_json::from_str(&line) {
+                if let Ok(v) = unsafe { simd_json::to_owned_value(line.as_mut_vec()) } {
```
MichaelScofield, line 609: Can the safety be hold? [`as_mut_vec`](https://doc.rust-lang.org/std/string/struct.String.html#method.as_mut_vec) safety requires that bytes are still valid UTF-8 after changed. Does the `to_owned_value` guarantee that?

src/servers/src/http/event.rs
```diff
@@ -617,8 +622,13 @@ fn extract_pipeline_value_by_content_type(
         ct if ct == *TEXT_CONTENT_TYPE || ct == *TEXT_UTF8_CONTENT_TYPE => payload
             .lines()
             .filter_map(|line| line.ok().filter(|line| !line.is_empty()))
-            .map(|line| json!({"message": line}))
-            .collect(),
+            .map(|line| {
+                let mut map = PipelineMap::new();
```
MichaelScofield, line 626:
```suggestion
                let mut map = PipelineMap::with_capacity(1);
```

src/servers/src/http/event.rs
shuiyisong: Change `as_mut_vec` to `as_bytes_mut`, for it doesn't require the unstable feature.
From the description, `to_owned_value` only de-escapes the string. It should work on a character level like any other JSON deserializer.

MichaelScofield (approved):

src/servers/src/http/event.rs
MichaelScofield: safety should be commented

fengys1996 (approved):
LGTM

Issue #5803, opened 2025-03-30
End-to-end column-wise data ingestion path


v0y4g3r:
### What type of enhancement is this?

Performance

### What does the enhancement do?

This issue proposes an end-to-end column format data ingestion path to boost ingestion performance.

Row-wise data ingestion, as implemented currently in GreptimeDB, is easy to understand and use, but the performance is becoming a major concern. 

For example, as a benchmark result shows, Protobuf encoding result is 2x slower than Arrow IPC but the encoded size is quite similar

![Image](https://github.com/user-attachments/assets/f8d39fc1-c82d-49a4-9408-7aa37666662e)

> (The benchmark code can be found [here](https://github.com/v0y4g3r/greptimedb/commit/b1e75072d4d7f0b48d59914b531cb8155c40cf66))

IIRC GreptimeDB used to have a column format ingestion API but was removed sometime because it was becoming challenging to implement the partition rules and after all all column-wise data would be converted to row-wise structs (namely `KeyValues`)

https://github.com/GreptimeTeam/greptimedb/blob/62cd1f1f870ee6811ef760f34caf6ce0e7e66359/src/mito2/src/memtable/key_values.rs#L29-L39

So in this PR I proposes an "end-to-end" ingestion path so we can spare such conversions and benefit the most for column-wise format.

### Implementation challenges

- Client should be able to encode data to columnar format 
- Implement efficient column-wise partition rules so we don't have to convert columns to rows to find the belonging regions
- Implement bulk memtable to accommodate columnar data and avoid conversion.s

killme2008:
Already done.

PR #5804, opened 2025-03-30
feat: Column-wise partition rule implementation

v0y4g3r:

- #5803

This PR addes `split_record_batch` method to `PartitionRule` trait and implements it in `MultiDimPartitionRule` using datafusion's physical expr evaluation methods.

## Performance

This PR adds a benchmark suite to test the performance of `PartitionRule::split_record_batch`. As can be seen in the following chart, column-wise partition implementation can be up to 2x to 10x faster than row-wise implementation.

image

## PR Checklist


src/partition/src/multi_dim.rs
```diff
@@ -134,6 +151,128 @@ impl MultiDimPartitionRule {
...
+                        .clone(),
+                ))
+            })
+            .collect::<error::Result<_>>()?;
+
+        let mut selected = BooleanArray::new(BooleanBuffer::new_unset(num_rows), None);
+        for region_selection in result.values() {
+            selected = arrow::compute::kernels::boolean::or(&selected, region_selection)
+                .context(error::ComputeArrowKernelSnafu)?;
+        }
```
MichaelScofield, line 263:
```suggestion
        }
        let all_rows_selected = selected.false_count() == 0;
        if all_rows_selected {
            return Ok(result);
        }
```

but how can a row not selected? I think it's expected that partition rule works on all rows.

fengjiachun (approved):
LGTM

src/partition/src/multi_dim.rs
v0y4g3r: Note that in the row-wise `find_region` method, if a row is not selected by any parittion rule, it will be assigned to the default region:

https://github.com/GreptimeTeam/greptimedb/blob/62cd1f1f870ee6811ef760f34caf6ce0e7e66359/src/partition/src/multi_dim.rs#L91-L108

This behavior lacks a documentation [here](https://docs.greptime.com/user-guide/administration/manage-data/table-sharding/#create-a-distributed-table), in this case, we defined 4 partition rules:

```
PARTITION ON COLUMNS (device_id, area) (
  device_id < 100 AND area = 'North',
  device_id < 100 AND area = 'South',
  device_id >= 100 AND area = 'East',
  device_id >= 100 AND area = 'West'
);
```

And it will be natural for readers to ask: what if I insert a row with, for example, `device_id = 99` and `area = 'East'`?

MichaelScofield: then the "amendment" here can be removed since the default region is already choosed inside

src/partition/src/expr.rs
```diff
@@ -37,6 +48,70 @@ pub enum Operand {
...
+                    Value::IntervalDayTime(v) => ScalarValue::IntervalDayTime(Some((*v).into())),
+                    Value::IntervalMonthDayNano(v) => {
+                        ScalarValue::IntervalMonthDayNano(Some((*v).into()))
+                    }
+                    Value::Duration(d) => duration_to_scalar_value(d.unit(), Some(d.value())),
+                    Value::Decimal128(d) => {
+                        let (v, p, s) = d.to_scalar_value();
+                        ScalarValue::Decimal128(v, p, s)
+                    }
+                    _ => unreachable!(),
```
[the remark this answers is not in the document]
v0y4g3r: resolved in 1ff8a7ce28

MichaelScofield (approved):

src/partition/src/multi_dim.rs
```diff
@@ -134,6 +151,133 @@ impl MultiDimPartitionRule {
...
+
+                let mut cache_write_guard = self.physical_expr_cache.write().unwrap();
+                cache_write_guard.replace((new_cache.clone(), schema));
+                new_cache
+            }
+        };
+
+        let mut result: HashMap<u32, BooleanArray> = physical_exprs
+            .iter()
+            .zip(self.regions.iter())
```
evenyag, line 238: Can we ensure that the order of physical_exprs is exactly the same as the order of regions? We may also check `physical_exprs.len() == self.regions.len()`.

PR #5806, opened 2025-03-31
feat: add backend field to fulltext options

zhongzc:

Wait for https://github.com/GreptimeTeam/greptime-proto/pull/226

Add a backend field to FulltextOptions to prepare for future multi-backend support in fulltext index. Currently supports:
- Bloom
- Tantivy

## PR Checklist


fengjiachun:
@zhongzc CI failed

src/datatypes/src/schema/column_schema.rs
```diff
@@ -522,11 +526,29 @@ impl fmt::Display for FulltextOptions {
...
             write!(f, ", case_sensitive={}", self.case_sensitive)?;
+            write!(f, ", backend={}", self.backend)?;
         }
         Ok(())
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default, Visit, VisitMut)]
+#[serde(rename_all = "kebab-case")]
+pub enum FulltextBackend {
```
killme2008, line 537: Please add doc for this enum

killme2008 (approved):
Good job! I wonder what happens in the following scenario:

* A column uses the tantivy backend initially, some rows are inserted, and indexes are built.
* The backend is altered to bloom.

Questions:
* Do the old tantivy indexes still work?
* Can both the new and old indexes work together?

zhongzc:
The application of the index is determined by the generated index file, not the table structure, so the original index file can still take effect.

The problem translates to whether both `bloom` and `tantivy` can work for the same query.

In principle, whether it's `matches` or `matches_term`, both have the potential to leverage `bloom` and `tantivy` for filtering and query acceleration. However, in the current implementation, `matches_term` is bound to `bloom` while `matches` is bound to `tantivy`. Thus, when using `matches`, the `tantivy` backend index takes effect, and when using `matches_term`, the `bloom` backend index takes effect. In the future, there may be opportunities to further refine and interconnect their applications.

PR #5807, opened 2025-03-31
feat(flow): batching mode engine

discord9:

introduce recording rule engine to drive recording rule:
- `RecordingRuleEngine`: provide same api as streaming engine, like `create/drop/flush_flow`
- `RecordingRuleTask`: the task which actually run the recording rule
- `RecordingRuleState`: freqly changing states of recording rule, including states like update time and dirty time window
- `DirtyTimeWindows`: to instruction which time window is dirty and use the info to update them

## PR Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a new batching mode system for flow processing that replaces legacy configurations.
  - Enabled the GRPC interface to execute logical plan queries, providing enhanced query handling.

- **Refactor**
  - Improved API accessibility and naming conventions for greater clarity.
  - Enhanced error handling with more granular error types for robust reporting.

- **Metrics**
  - Added new instrumentation to track batching task execution times and detect slow queries.
  - Extended GRPC performance monitoring to capture logical plan execution durations.


src/common/meta/src/ddl/create_flow.rs
```diff
@@ -328,27 +328,28 @@ pub enum CreateFlowState {
...
-    /// The flow is a recording rule task.
-    RecordingRule,
+    /// The flow is a batching task.
+    Batching,
     /// The flow is a streaming task.
     Streaming,
 }
 
 impl FlowType {
     pub const RECORDING_RULE: &str = "recording_rule";
```
fengjiachun, line 338: batching

src/flow/src/metrics.rs
```diff
@@ -28,6 +28,32 @@ lazy_static! {
         &["table_id"]
     )
     .unwrap();
+    pub static ref METRIC_FLOW_RULE_ENGINE_QUERY_TIME: HistogramVec = register_histogram_vec!(
```
fengjiachun, line 31: rule_engine -> batching_engine

src/flow/src/metrics.rs
```diff
@@ -28,6 +28,32 @@ lazy_static! {
...
+        "greptime_flow_rule_engine_query_time",
+        "flow rule engine query time",
```
fengjiachun, line 33:
```suggestion
        "greptime_flow_batching_engine_query_time",
        "flow batching engine query time",
```

src/flow/src/metrics.rs
```diff
@@ -28,6 +28,32 @@ lazy_static! {
...
+            60.,
+            2. * 60.,
+            5. * 60.,
+            10. * 60.
+        ]
+    )
+    .unwrap();
+    pub static ref METRIC_FLOW_RULE_ENGINE_SLOW_QUERY: HistogramVec = register_histogram_vec!(
+        "greptime_flow_rule_engine_slow_query",
+        "flow rule engine slow query",
```
fengjiachun, line 52:
```suggestion
        "greptime_flow_batching_engine_slow_query",
        "flow batching engine slow query",
```

src/flow/src/metrics.rs
```diff
@@ -28,6 +28,32 @@ lazy_static! {
...
+            20.,
+            30.,
+            60.,
+            2. * 60.,
+            5. * 60.,
+            10. * 60.
+        ]
+    )
+    .unwrap();
+    pub static ref METRIC_FLOW_RULE_ENGINE_SLOW_QUERY: HistogramVec = register_histogram_vec!(
```
fengjiachun, line 50: rule -> batching

src/flow/src/batching_mode/frontend_client.rs
```diff
@@ -25,8 +25,8 @@ use common_meta::rpc::store::RangeRequest;
 use meta_client::client::MetaClient;
 use snafu::ResultExt;
 
+use crate::batching_mode::DEFAULT_RULE_ENGINE_QUERY_TIMEOUT;
```
fengjiachun, line 28: batching_engine

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,1151 @@
...
+        let engine = self.query_engine.clone();
+        let frontend = self.frontend_client.clone();
+
+        task.check_execute(&engine, &frontend).await?;
+
+        // TODO(discord9): also save handle & use time wheel or what for better
+        let _handle = common_runtime::spawn_global(async move {
+            match task_inner.start_executing_loop(engine, frontend).await {
+                Ok(()) => info!("Flow {} shutdown", task_inner.flow_id),
+                Err(err) => common_telemetry::error!(
```
killme2008, line 264: Should we store the start error and report it immediately before returning from this function?

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,1151 @@
...
+                    "Flow {} encounter unrecoverable error: {err:?}",
+                    task_inner.flow_id
+                ),
+            }
+        });
+
+        // only replace here not earlier because we want the old one intact if something went wrong before this line
+        let replaced_old_task_opt = self.tasks.write().await.insert(flow_id, task);
+        drop(replaced_old_task_opt);
```
killme2008, line 273: For example, if we can keep the starting error somewhere, we can check it here and return the error ASAP.

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,1151 @@
...
+        let task = task.with_context(|| UnexpectedSnafu {
+            reason: format!("Can't found task for flow {flow_id}"),
+        })?;
+
+        task.gen_exec_once(&self.query_engine, &self.frontend_client)
+            .await?;
+        Ok(())
+    }
+
+    pub async fn flow_exist(&self, flow_id: FlowId) -> bool {
```
killme2008, line 307: Add documents to all these pub functions. And I wonder if we can add the flow engine and task trait?

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,1151 @@
...
+                name: full_table_name.clone(),
+            })?
+            .into_inner();
+
+        let new_query = self
+            .gen_query_with_time_window(engine.clone(), &table.table_info.meta)
+            .await?;
+
+        let insert_into = if let Some((new_query, _column_cnt)) = new_query {
+            // TODO(discord9): also assign column name to compat update_at column
```
killme2008, line 452: SQL injection warning!

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,1151 @@
...
+
+        let req = api::v1::greptime_request::Request::Query(api::v1::QueryRequest {
+            query: Some(api::v1::query_request::Query::Sql(sql.to_string())),
+        });
+
+        let res = db_client.database.handle(req).await;
+        drop(timer);
+
+        let elapsed = instant.elapsed();
+        if let Ok(res1) = &res {
```
killme2008, line 487: Why `res1` here but `res` at the following branch?

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,1151 @@
...
+        }
+    }
+
+    /// Generate the create table SQL
+    ///
+    /// the auto created table will automatically added a `update_at` Milliseconds DEFAULT now() column in the end
+    /// (for compatibility with flow streaming mode)
+    ///
+    /// and it will use first timestamp column as time index, all other columns will be added as normal columns and nullable
+    async fn gen_create_table_sql(&self, engine: QueryEngineRef) -> Result<String, Error> {
```
killme2008, line 563: SQL injection warning! Why not create the plan or statement instead?

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,1151 @@
...
+    let all_pk_cols: Vec<_> = all_pk_cols
+        .into_iter()
+        .filter(|col| first_time_stamp != Some(col.to_string()))
+        .collect();
+
+    Ok((first_time_stamp, all_pk_cols))
+}
+
+#[derive(Debug)]
+pub struct TaskState {
```
killme2008, line 772: This file/mod is too large and should be split into `task`, `engine`, and `state`.

src/flow/src/error.rs
```diff
@@ -248,9 +263,11 @@ impl ErrorExt for Error {
             | Self::FlowNotFound { .. }
             | Self::ListFlows { .. } => StatusCode::TableNotFound,
             Self::Plan { .. } | Self::Datatypes { .. } => StatusCode::PlanQuery,
-            Self::InvalidQuery { .. } | Self::CreateFlow { .. } | Self::Arrow { .. } => {
-                StatusCode::EngineExecuteQuery
-            }
+            Self::InvalidQuery { .. }
+            | Self::InvalidRequest { .. }
```
killme2008, line 267: InvalidXXX must return `IllegalArgument` status.

killme2008 (commented):
I think this PR needs to be refactored and requires more unit tests.

src/flow/src/batching_mode/engine.rs
discord9: > SQL injection warning! Why not create the plan or statement instead?

will restore frontend's grpc logical plan handle path for this in this PR, previously it's deemed unnecessary since user of grpc protocol shouldn't send logical plan

src/flow/src/batching_mode/engine.rs
```diff
@@ -185,9 +187,9 @@ impl BatchingEngine {
         let flow_type = flow_options.get(FlowType::FLOW_TYPE_KEY);
 
         ensure!(
-            flow_type == Some(&FlowType::Batching.to_string()) || flow_type.is_none(),
+            matches!(flow_type, Some(ty) if ty == "batching"),
```
killme2008, line 160: Don't forget the `None`.

src/flow/src/batching_mode/engine.rs
discord9: > Should we store the start error and report it immediately before returning from this function?

Found out actually errors in executing loop should just get logged instead of throw out as start error

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,1151 @@
...
+                    &column_name,
+                    &df_schema,
+                    &self.query_engine.engine_state().session_state(),
+                )
+            })
+            .transpose()?;
+
+        info!("Flow id={}, found time window expr={:?}", flow_id, phy_expr);
+
+        let task = BatchingTask::new(
```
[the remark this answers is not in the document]
discord9: > Using a builder to construct the task would be better.

wanted to make this task building more explicit and not to accidentally forget to set a field(which did happen before)

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,1151 @@
...
+            rx,
+        );
+
+        let task_inner = task.clone();
+        let engine = self.query_engine.clone();
+        let frontend = self.frontend_client.clone();
+
+        task.check_execute(&engine, &frontend).await?;
+
+        // TODO(discord9): also save handle & use time wheel or what for better
```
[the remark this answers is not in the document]
discord9: > Please create an issue for this todo, it's important.

#5821

src/flow/src/batching_mode/engine.rs
killme2008: Does it need to check if it is `None`? It was checked before refactoring.

discord9: > Does it need to check if it is `None`? It was checked before refactoring.

Yes, as default behavior is to use batching mode if no type was given?

src/flow/src/batching_mode/state.rs
```diff
@@ -0,0 +1,375 @@
...
+
+        self.last_update_time + next_duration
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct DirtyTimeWindows {
+    /// windows's `start -> end` and non-overlapping
+    /// `end` is exclusive(and optional)
+    windows: BTreeMap<Timestamp, Option<Timestamp>>,
```
zhongzc, line 85: non-blocking: using a sorted Vec can be more effective.

src/flow/src/batching_mode/state.rs
```diff
@@ -0,0 +1,375 @@
...
+            };
+
+            let std_window_size = window_size.to_std().map_err(|e| {
+                InternalSnafu {
+                    reason: e.to_string(),
+                }
+                .build()
+            })?;
+
+            // if cur.lower - prev.upper <= window_size * 2, merge
```
zhongzc, line 223: `Self::MERGE_DIST` is 3..

src/flow/src/batching_mode/state.rs
discord9: > non-blocking: using a sorted Vec can be more effective.

need to insert at random position so need a ordered map

src/frontend/src/instance/grpc.rs
```diff
@@ -82,11 +84,15 @@ impl GrpcQueryHandler for Instance {
                         let output = result.remove(0)?;
                         attach_timer(output, timer)
                     }
-                    Query::LogicalPlan(_) => {
-                        return NotSupportedSnafu {
-                            feat: "Execute LogicalPlan in Frontend",
-                        }
-                        .fail();
+                    Query::LogicalPlan(plan) => {
+                        let timer = GRPC_HANDLE_SQL_ELAPSED.start_timer();
```
fengjiachun, line 88: We might need to add a `GRPC_HANDLE_LOGICAL_PLAN_ELAPSED` timer.

src/flow/src/batching_mode/task.rs
```diff
@@ -0,0 +1,836 @@
...
+            info!(
+                "Try creating sink table(if not exists) with expr: {:?}",
+                create_table
+            );
+            self.create_table(frontend_client, create_table).await?;
+            info!(
+                "Sink table {}(if not exists) created",
+                self.sink_table_name.join(".")
+            );
+        }
```
killme2008, line 134: Let's enhance `information_schema.flows` to include recent execution results (success/error messages).

src/flow/src/batching_mode/task.rs
```diff
@@ -0,0 +1,836 @@
...
+        let ty = ConcreteDataType::from_arrow_type(field.data_type());
+        let col_schema = if first_time_stamp == Some(name.clone()) {
+            ColumnSchema::new(name, ty, false).with_time_index(true)
+        } else {
+            ColumnSchema::new(name, ty, true)
+        };
+        column_schemas.push(col_schema);
+    }
+
+    let update_at_schema = ColumnSchema::new(
```
killme2008, line 608: The name is too common to confuse. Maybe we can add a prefix such as `greptime_flow_update_at` or like `AUTO_CREATED_PLACEHOLDER_TS_COL`.

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    pub async fn handle_inserts(
+        &self,
+        request: api::v1::region::InsertRequests,
+    ) -> Result<FlowResponse, Error> {
+        let table_info_mgr = self.table_meta.table_info_manager();
+        let mut group_by_table_name: HashMap<TableName, Vec<api::v1::Rows>> = HashMap::new();
+        for r in request.requests {
+            let tid = RegionId::from(r.region_id).table_id();
+            let name = get_table_name(table_info_mgr, &tid).await?;
```
fengjiachun, line 77: We should collect all table IDs and retrieve their names in batches, as fetching each table name individually in a loop may be very slow.

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,292 @@
...
+            table_info_mgr
+                .batch_get(&tids)
+                .await
+                .with_context(|_| TableNotFoundMetaSnafu {
+                    msg: format!("Failed to get table info for table ids: {:?}", tids),
+                })?;
+        let group_by_table_name = group_by_table_id
+            .into_iter()
+            .map(|(id, rows)| {
+                let table_name = table_infos[&id].table_name();
```
fengjiachun, line 95: We need to verify if `table_infos` includes the `table_id`.

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,292 @@
...
+        for r in request.requests {
+            let tid = RegionId::from(r.region_id).table_id();
+            let entry = group_by_table_id.entry(tid).or_default();
+            if let Some(rows) = r.rows {
+                entry.push(rows);
+            }
+        }
+
+        let tids = group_by_table_id.keys().cloned().collect::<Vec<TableId>>();
+        let table_infos =
```
fengjiachun, line 85: If a table doesn't exist, it won't be included in the results, so we need to check if any tables are missing.

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,292 @@
...
+        );
+
+        let Some(query_ctx) = query_ctx else {
+            UnexpectedSnafu {
+                reason: "Query context is None".to_string(),
+            }
+            .fail()?
+        };
+        let query_ctx = Arc::new(query_ctx);
+        let mut source_table_names = Vec::new();
```
fengjiachun, line 198: with_capacity

src/flow/src/batching_mode/state.rs
```diff
@@ -0,0 +1,397 @@
...
+        let expr = expr_lst.into_iter().reduce(|a, b| a.or(b));
+        Ok(expr)
+    }
+
+    /// Merge time windows that overlaps or get too close
+    pub fn merge_dirty_time_windows(
+        &mut self,
+        window_size: chrono::Duration,
+        expire_lower_bound: Option<Timestamp>,
+    ) -> Result<(), Error> {
```
fengjiachun, line 201: // fast path
if self.windows.is_empty() {
    return Ok(());
}

src/flow/src/batching_mode/state.rs
```diff
@@ -0,0 +1,397 @@
...
+                prev_tw = Some((lower_bound, upper_bound));
+                continue;
+            };
+
+            let std_window_size = window_size.to_std().map_err(|e| {
+                InternalSnafu {
+                    reason: e.to_string(),
+                }
+                .build()
+            })?;
```
fengjiachun, line 224: Move this outof the for loop?

src/flow/src/batching_mode/task.rs
```diff
@@ -0,0 +1,769 @@
...
+        } else {
+            Ok(None)
+        }
+    }
+
+    pub async fn gen_insert_plan(
+        &self,
+        engine: &QueryEngineRef,
+    ) -> Result<Option<LogicalPlan>, Error> {
+        let table_name_mgr = self.table_meta.table_name_manager();
```
fengjiachun, line 168:
```suggestion
```

src/flow/src/batching_mode/task.rs
```diff
@@ -0,0 +1,769 @@
...
+
+        let full_table_name = self.sink_table_name.clone().join(".");
+
+        let table_id = table_name_mgr
```
fengjiachun, line 172:
```suggestion
        let table_id = self
            .table_meta
            .table_name_manager()
```

src/flow/src/batching_mode/task.rs
```diff
@@ -0,0 +1,768 @@
...
+// auto created table have a auto added column `update_at`, and optional have a `AUTO_CREATED_PLACEHOLDER_TS_COL` column for time index placeholder if no timestamp column is specified
+// TODO(discord9): unit test
+fn create_table_with_expr(
+    plan: &LogicalPlan,
+    sink_table_name: &[String; 3],
+) -> Result<CreateTableExpr, Error> {
+    let catalog_name = &sink_table_name[0];
+    let schema_name = &sink_table_name[1];
+    let table_name = &sink_table_name[2];
+
```
fengjiachun, line 524:
```suggestion
```

src/flow/src/batching_mode/task.rs
```diff
@@ -0,0 +1,768 @@
...
+                AUTO_CREATED_PLACEHOLDER_TS_COL,
+                ConcreteDataType::timestamp_millisecond_datatype(),
+                false,
+            )
+            .with_time_index(true)
+            .with_default_constraint(Some(ColumnDefaultConstraint::Value(Value::Timestamp(
+                Timestamp::new_millisecond(0),
+            ))))
+            .context(DatatypesSnafu {
+                extra: "Failed to build column `{} TimestampMillisecond default 0`",
```
fengjiachun, line 565: Is this `{}` in the message right?

fengjiachun (approved):
Rest LGTM

src/flow/src/metrics.rs
```diff
@@ -28,6 +28,32 @@ lazy_static! {
...
+            10.,
+            20.,
+            30.,
+            60.,
+            2. * 60.,
+            5. * 60.,
+            10. * 60.
+        ]
+    )
+    .unwrap();
```
waynexia, line 49: we usually use a fixed step, like 2, 3 or 10. E.g.:
```
vec![1, 2, 4, 8]
```

And btw what's the unit of this metric?

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,306 @@
...
+                    msg: format!("Failed to get table info for table ids: {:?}", tids),
+                })?;
+
+        if !tids.iter().all(|id| table_infos.contains_key(id)) {
+            warn!(
+                "Failed to get all the table info for table ids, expected table ids: {:?}, actual table infos' table id: {:?}, those table doesn't exist: {:?}",
+                tids,
+                table_infos.keys(),
+                tids.iter().filter(|id| !table_infos.contains_key(id)).collect::<Vec<_>>()
+            );
```
waynexia, line 99: nit: make it more concise
```suggestion
            warn!(
                "Failed to get all the table info for table ids, expected table ids: {:?}, those table doesn't exist: {:?}",
                tids,
                tids.iter().filter(|id| !table_infos.contains_key(id)).collect::<Vec<_>>()
            );
```

waynexia: And the missing table ids is calculated twice in this case

tests/cases/standalone/common/flow/flow_advance_ttl.result
```diff
@@ -60,7 +60,7 @@ SHOW CREATE TABLE out_distinct_basic;
 +--------------------+---------------------------------------------------+
 | out_distinct_basic | CREATE TABLE IF NOT EXISTS "out_distinct_basic" ( |
 |                    |   "dis" INT NULL,                                 |
-|                    |   "update_at" TIMESTAMP(3) NULL,                  |
+|                    |   "__update_at" TIMESTAMP(3) NULL,                |
```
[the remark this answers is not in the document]
discord9: > Is this change necessary?

unnecessary, and might break existing flows, should I revert it for now?

src/flow/src/batching_mode/task.rs
```diff
@@ -0,0 +1,767 @@
...
+    pub flow_id: FlowId,
+    pub query: String,
+    plan: LogicalPlan,
+    pub time_window_expr: Option<TimeWindowExpr>,
+    /// in seconds
+    pub expire_after: Option<i64>,
+    sink_table_name: [String; 3],
+    pub source_table_names: HashSet<[String; 3]>,
+    table_meta: TableMetadataManagerRef,
+    pub state: Arc<RwLock<TaskState>>,
```
waynexia, line 73: It looks like this RwLock need not to be an async lock.

src/flow/src/batching_mode/engine.rs
```diff
@@ -0,0 +1,306 @@
...
+                    table_name.catalog_name,
+                    table_name.schema_name,
+                    table_name.table_name,
+                ];
+                Some((table_name, rows))
+            })
+            .collect::<HashMap<_, _>>();
+
+        // TODO(discord9): not use one lock for all tasks
+        for (_flow_id, task) in self.tasks.read().await.iter() {
```
waynexia, line 120: can we somehow make the handling logic parallel? I'm afraid this will become a bottleneck

src/flow/src/batching_mode/state.rs
```diff
@@ -0,0 +1,401 @@
...
+                )
+            }
+        }
+
+        // get the first `MAX_FILTER_NUM` time windows
+        let nth = self
+            .windows
+            .iter()
+            .nth(Self::MAX_FILTER_NUM)
+            .map(|(key, _)| *key);
```
waynexia, line 160: IMO we need either (1) print warnings when some time windows are ignored, or (2) handle all dirty windows in multiple runs, with each run has `MAX_FILTER_NUM`

tests/cases/standalone/common/flow/flow_advance_ttl.result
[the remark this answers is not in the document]
waynexia: If it's not necessary then I'd prefer to keep it as is. Or do you have any consideration for prefixing it?

src/flow/src/batching_mode/state.rs
discord9: > > IMO we need either (1) print warnings when some time windows are ignored, or (2) handle all dirty windows in multiple runs, with each run has `MAX_FILTER_NUM`
> 
> warning is print in above, will refactor to handle all dirty windows in multiple runs too later

also done in multiple runs by setting wait time to zero if still have dirty windows

src/frontend/src/instance/grpc.rs
```diff
@@ -82,11 +86,16 @@ impl GrpcQueryHandler for Instance {
                         let output = result.remove(0)?;
                         attach_timer(output, timer)
                     }
-                    Query::LogicalPlan(_) => {
-                        return NotSupportedSnafu {
-                            feat: "Execute LogicalPlan in Frontend",
-                        }
-                        .fail();
+                    Query::LogicalPlan(plan) => {
+                        // this path is useful internally when flownode need to execute a logical plan through GRPC interface
```
waynexia, line 90:
```suggestion
                        // this path is useful internally when flownode needs to execute a logical plan through gRPC interface
```

waynexia (approved):
LGTM

Issue #5667, opened 2025-03-07
Upgrade datafusion


evenyag:
### What type of enhancement is this?

Refactor, Tech debt reduction

### What does the enhancement do?

We have some workarounds for the current version of datafusion
- https://github.com/GreptimeTeam/greptimedb/issues/5659
- https://github.com/GreptimeTeam/greptimedb/issues/5469

### Implementation challenges

There will be some breaking changes, we may need to resolve them.

MichaelScofield:
now that datafusion has updated to newer version, can some walkaround codes be removed now? or does this issue can be closed? @evenyag

evenyag:
Looks like we still can't revert https://github.com/GreptimeTeam/greptimedb/pull/5662

```
[
    ... skip(159) ...,
~   +---------------------+----------------------------+ -> Error: 1003(Internal), Internal error: 1003,
-   | ts                  | avg(latencies.avg_latency) |,
-   +---------------------+----------------------------+,
-   | 2023-10-01T10:00:00 | 135.0                      |,
-   | 2023-10-01T10:00:02 | 120.0                      |,
-   | 2023-10-01T10:00:04 | 130.0                      |,
-   +---------------------+----------------------------+,
    ... skip(5) ...
]
```

evenyag:
We may need to upgrade to a more latest one.

PR #5814, opened 2025-04-02
chore: update datafusion family

MichaelScofield:

https://github.com/GreptimeTeam/greptimedb/issues/5667

update datafusion family:

arrow/parquet: 53.0 -> 54.2
sqlparser: 0.52 -> 0.54
datafusion: 2024.12 -> 2025.2

to use more recent arrow and parquet

## PR Checklist


fengjiachun (approved):
LGTM

src/query/src/error.rs
```diff
@@ -126,7 +126,7 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("DataFusion error"))]
+    #[snafu(display(""))]
```
evenyag, line 129: It'd be better not to provide an empty display message. If some places use `to_string()` or `Display` to show the error, we will get an empty message.

src/sql/src/statements/insert.rs
```diff
@@ -41,7 +41,10 @@ macro_rules! parse_fail {
 impl Insert {
     pub fn table_name(&self) -> &ObjectName {
         match &self.inner {
-            Statement::Insert(insert) => &insert.table_name,
+            Statement::Insert(insert) => match &insert.table {
+                TableObject::TableName(name) => name,
+                TableObject::TableFunction(_) => unimplemented!(),
```
evenyag, line 46: Is it unreachable?

tests/cases/standalone/common/range/special_aggr.result
```diff
@@ -64,12 +64,12 @@ SELECT ts, host, first_value(addon ORDER BY val DESC) RANGE '5s', last_value(add
 | ts                  | host  | first_value(host.addon) ORDER BY [host.val DESC NULLS FIRST] RANGE 5s | last_value(host.addon) ORDER BY [host.val DESC NULLS FIRST] RANGE 5s |
 +---------------------+-------+-----------------------------------------------------------------------+----------------------------------------------------------------------+
 | 1970-01-01T00:00:00 | host1 | 3                                                                     | 1                                                                    |
-| 1970-01-01T00:00:05 | host1 | 4                                                                     | 4                                                                    |
+| 1970-01-01T00:00:05 | host1 | 4                                                                     | 6                                                                    |
```
evenyag, line 67: Is it expected?

src/common/datasource/Cargo.toml
```diff
@@ -31,7 +31,8 @@ derive_builder.workspace = true
 futures.workspace = true
 lazy_static.workspace = true
 object-store.workspace = true
-orc-rust = { version = "0.5", default-features = false, features = [
+object_store_opendal.workspace = true
+orc-rust = { git = "https://github.com/GreptimeTeam/orc-rust", rev = "dbb9d7906751deaf2c9cc0c490c6c78406fc70da", default-features = false, features = [
```
evenyag, line 35: Do we have a plan to update the official orc-rust?

src/query/src/error.rs
MichaelScofield: It seems that new DataFusion can pushdown functions on "limit" clause, then the error messages could be different between standalone and distributed modes, causing sqlness to fall. See this action run before I changed this error: https://github.com/GreptimeTeam/greptimedb/actions/runs/14212982871/job/39823883300

To pass sqlness, I have to either add "DataFusion error" in `PollStreamSnafu`'s display, and produces a lot of verbose "DataFusion error:" in the error output; or do it like this here, and some verbose “DataFusion error:" are reduced as well. I choose the latter for better message conciseness.

tests/cases/standalone/common/range/special_aggr.result
MichaelScofield: I think so. It's related to how "null"s are ordered in DataFusion, and it doesn't really matter here.

MichaelScofield:
@fengys1996 @evenyag PTAL

fengys1996 (approved):

src/query/src/error.rs
evenyag: > It seems that new DataFusion can pushdown functions on "limit" clause, then the error messages could be different between standalone and distributed modes, causing sqlness to fall.

You can move inconsistent test results to standalone and distributed tests separately.

MichaelScofield: that considered and less desired because i thought it does not worth the cost of duplicating the codes, and what's more important, now the error messages are less verbose

MichaelScofield: i may find a way to make `to_string` on these errors not accidently produce empty result while still keeping the sqlness happy, will do it in next pr

PR #5817, opened 2025-04-02
feat: add `matches_term` function

zhongzc:

Adds a new `matches_term` function that performs exact term/phrase matching with non-alphanumeric boundaries. The function is designed for:
- Whole-word matching (e.g. "cat" in "cat!" but not in "category")
- Phrase matching (e.g. "hello world" in "note:hello world!")

## PR Checklist


fengjiachun (approved):
LGTM

src/common/function/src/scalars/matches_term.rs
```diff
@@ -0,0 +1,349 @@
...
+                            result.push_null();
+                            continue;
+                        }
+                        Some(term) => term,
+                    };
+                    MatchesTermFinder::new(term).find(text)
+                }
+            };
+            result.push(Some(contains));
+        }
```
evenyag, line 162: non-blocking: We can downcast the vector or array to a string array. This can avoid using `get_ref().as_string().unwrap()` for each row.

src/common/function/src/scalars/matches_term.rs
```diff
@@ -0,0 +1,349 @@
...
+
+        let mut pos = 0;
+        while let Some(found_pos) = self.finder.find(text[pos..].as_bytes()) {
+            let actual_pos = pos + found_pos;
+
+            let prev_ok = self.starts_with_non_alnum
+                || text[..actual_pos]
+                    .chars()
+                    .last()
+                    .map(|c| !c.is_alphanumeric())
```
evenyag, line 227: Do we need to test non-ascii words?

zhongzc: ref: `is_alphanumeric`'s doc

```
assert!('٣'.is_alphanumeric());
assert!('7'.is_alphanumeric());
assert!('৬'.is_alphanumeric());
assert!('¾'.is_alphanumeric());
assert!('①'.is_alphanumeric());
assert!('K'.is_alphanumeric());
assert!('و'.is_alphanumeric());
assert!('藏'.is_alphanumeric());
```

src/common/function/src/scalars/matches_term.rs
zhongzc: applied

evenyag (approved):

src/common/function/src/scalars/matches_term.rs
```diff
@@ -0,0 +1,354 @@
...
+                        .next()
+                        .map(|c| !c.is_alphanumeric())
+                        .unwrap_or(true);
+
+                if next_ok {
+                    return true;
+                }
+            }
+
+            pos = actual_pos + 1;
```
discord9, line 246: I'm a bit confused about if this `pos` should always be on next unicode character boundary by calling [`ceil_char_boundary`](https://doc.rust-lang.org/std/primitive.str.html#method.ceil_char_boundary), Or does it need to be? Maybe some comment here would be nice?

discord9: > I'm a bit confused about if this `pos` should always be on next unicode character boundary by calling [`ceil_char_boundary`](https://doc.rust-lang.org/std/primitive.str.html#method.ceil_char_boundary), Or does it need to be? Maybe some comment here would be nice?

edited: probably not worthy as it's just 3 bytes less

src/common/function/src/scalars/matches_term.rs
```diff
@@ -0,0 +1,375 @@
...
+
+        assert!(MatchesTermFinder::new("error").find("errorlog_error_case"));
+        assert!(MatchesTermFinder::new("test").find("atestbtestc_test_end"));
+        assert!(MatchesTermFinder::new("data").find("database_data_store"));
+        assert!(!MatchesTermFinder::new("data").find("database_datastore"));
+        assert!(MatchesTermFinder::new("log.txt").find("catalog.txt_log.txt!"));
+        assert!(!MatchesTermFinder::new("log.txt").find("catalog.txtlog.txt!"));
+        assert!(MatchesTermFinder::new("data-set").find("bigdata-set_data-set!"));
+
+        assert!(MatchesTermFinder::new("中文").find("这是中文测试，中文！"));
```
discord9, line 372: might as well:
```rust
assert!(!MatchesTermFinder::new("中文").find("这是中文测试"));
```

PR #5818, opened 2025-04-02
chore: update rust toolchain

shuiyisong:

As the title suggests, this PR updates the rust toolchain to `2025-04-15`.
On my local machine, a clean `make check` reduces ~10s compared to the former version.
~~Let's keep the PR draft until v0.14 is released.~~

## PR Checklist


sunng87:
@shuiyisong we can move on

Cargo.toml
```diff
@@ -73,10 +73,11 @@ edition = "2021"
 license = "Apache-2.0"
 
 [workspace.lints]
-clippy.print_stdout = "warn"
-clippy.print_stderr = "warn"
 clippy.dbg_macro = "warn"
 clippy.implicit_clone = "warn"
+clippy.result_large_err = "allow"
+clippy.large_enum_variant = "allow"
```
MichaelScofield, line 79: these two are generally considered bad for performance, why allow them?

shuiyisong: These are reported right after changing the toolchain version. 
After a brief look, I found that there are too many places to change to fix these two. We can fix them in a separate PR later.

sunng87:
@shuiyisong We need to change the regex to 

```
-- SQLNESS REPLACE (\d+\s*) OID
```

Because the oid may contain multiple digits and it will affect the column width

tests/cases/standalone/common/system/pg_catalog.sql
```diff
@@ -69,7 +69,7 @@ ORDER BY 1,2;
 
 -- make sure oid of namespace keep stable
 -- SQLNESS PROTOCOL POSTGRES
--- SQLNESS REPLACE (\d+) OID
+-- SQLNESS REPLACE (\d+\s*) OID
 SELECT * FROM pg_namespace ORDER BY oid;
```
sunng87, line 73: We can change this from order by `oid` to `nspname`, just for a stable result.

sunng87: Let me push a fix for this.

Cargo.toml
MichaelScofield: That's fine. FYI if the large err is caused by snafu, you can simply do like [this](https://github.com/GreptimeTeam/greptimedb-edge-ingester-java/blob/47f823d29567013e42d3c94515757e2b9d575fa4/core/src/main/rust/ingester/src/error.rs#L26)

MichaelScofield (approved):

sunng87 (approved):

PR #5842, opened 2025-04-08
fix: alway rejects write while downgrading region

WenyXu:

This PR makes the datanode immediately reject write operations while handling downgrade region instructions.

**Previous behavior:**
A region would only transition to a "downgraded leader" state after discovering its downgrade status through heartbeats, creating a potential timing gap.

**New behavior:**
Write operations are now immediately rejected when handling downgrade region instructions, and then flushes the memtable data.

With write operations rejected and memtable data flushed, the upgrade candidate process becomes faster since there's no data that would need to be replayed during the upgrade candidate region operation.

This PR also 
- Changes the default region migration timeout from 30s to 300s.
- Changes the `CLOSE_DOWNGRADED_REGION_TIMEOUT` to 10s.
- Introduces metadata region last entry id to improve replay speed of metadata regions.

## PR Checklist

### DIFF
--- a/src/common/function/src/admin/migrate_region.rs
+++ b/src/common/function/src/admin/migrate_region.rs
@@ -25,12 +25,12 @@ use session::context::QueryContextRef;
 use crate::handlers::ProcedureServiceHandlerRef;
 use crate::helper::cast_u64;
 
-const DEFAULT_TIMEOUT_SECS: u64 = 30;
+const DEFAULT_TIMEOUT_SECS: u64 = 120;
 
 /// A function to migrate a region from source peer to target peer.
 /// Returns the submitted procedure id if success. Only available in cluster mode.
 ///
-/// - `migrate_region(region_id, from_peer, to_peer)`, with timeout(30 seconds).
+/// - `migrate_region(region_id, from_peer, to_peer)`, with timeout(120 seconds).
 /// - `migrate_region(region_id, from_peer, to_peer, timeout(secs))`.
 ///
 /// The parameters:
--- a/src/common/function/src/admin/migrate_region.rs
+++ b/src/common/function/src/admin/migrate_region.rs
@@ -25,12 +25,13 @@ use session::context::QueryContextRef;
 use crate::handlers::ProcedureServiceHandlerRef;
 use crate::helper::cast_u64;
 
-const DEFAULT_TIMEOUT_SECS: u64 = 120;
+/// The default timeout for migrate region procedure.
+const DEFAULT_TIMEOUT_SECS: u64 = 300;
 
 /// A function to migrate a region from source peer to target peer.
 /// Returns the submitted procedure id if success. Only available in cluster mode.
 ///
-/// - `migrate_region(region_id, from_peer, to_peer)`, with timeout(120 seconds).
+/// - `migrate_region(region_id, from_peer, to_peer)`, with timeout(300 seconds).
 /// - `migrate_region(region_id, from_peer, to_peer, timeout(secs))`.
 ///
 /// The parameters:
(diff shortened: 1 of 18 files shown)


MichaelScofield (approved):

src/common/function/src/admin/migrate_region.rs
```diff
@@ -25,12 +25,12 @@ use session::context::QueryContextRef;
 use crate::handlers::ProcedureServiceHandlerRef;
 use crate::helper::cast_u64;
 
-const DEFAULT_TIMEOUT_SECS: u64 = 30;
+const DEFAULT_TIMEOUT_SECS: u64 = 120;
```
killme2008, line 28: Better to document the timeout, what is its usage?

src/common/meta/src/instruction.rs
```diff
@@ -157,6 +157,8 @@ pub struct UpgradeRegion {
     pub region_id: RegionId,
     /// The `last_entry_id` of old leader region.
     pub last_entry_id: Option<u64>,
+    /// The `last_entry_id` of old leader metadata region (Only used for metrics).
```
killme2008, line 160:
```suggestion
    /// The `last_entry_id` of old leader metadata region (Only used for metric engine).
```

src/datanode/src/heartbeat/handler/downgrade_region.rs
```diff
@@ -89,33 +96,35 @@ impl HandlerContext {
...
+                    warn!("Region: {region_id} is not found");
+                    return Some(InstructionReply::DowngradeRegion(DowngradeRegionReply {
+                        last_entry_id: None,
+                        metadata_last_entry_id: None,
+                        exists: false,
+                        error: None,
+                    }));
+                }
+                Err(err) => {
+                    warn!(err; "Failed to convert region to downgrading leader");
```
killme2008, line 121: The error must be logged as an error.
```suggestion
                    error!(err; "Failed to convert region to downgrading leader");
```

src/meta-srv/src/procedure/region_migration.rs
```diff
@@ -127,6 +127,8 @@ pub struct VolatileContext {
     leader_region_lease_deadline: Option<Instant>,
     /// The last_entry_id of leader region.
     leader_region_last_entry_id: Option<u64>,
+    /// The last_entry_id of leader metadata region (Only used for metrics).
```
killme2008, line 130:
```suggestion
    /// The last_entry_id of leader metadata region (Only used for metric engine).
```

src/meta-srv/src/procedure/region_migration/downgrade_leader_region.rs
```diff
@@ -331,41 +316,6 @@ mod tests {
         assert!(!err.is_retryable());
     }
 
-    #[tokio::test]
-    async fn test_should_reject_writes() {
```
killme2008, line 335: Do we still need to write a unit test for rejecting writes?

killme2008 (commented):
Question: When region downgrading fails due to a flushing timeout or other unexpected behaviors, what happens to this migration? Will it retry or abort?

src/common/function/src/admin/migrate_region.rs
evenyag, line 28: The timeout may not be enough.
```suggestion
const DEFAULT_TIMEOUT_SECS: u64 = 300;
```

src/meta-srv/src/procedure/region_migration.rs
WenyXu: resolved in [cb27076](https://github.com/GreptimeTeam/greptimedb/pull/5842/commits/cb27076f79c8421627c23b041cb1dd2e79a9905e)

src/meta-srv/src/procedure/region_migration/downgrade_leader_region.rs
WenyXu: Now, rejecting writes has become the default behavior, we don't need this unit test anymore

fengjiachun (approved):
LGTM

PR #5858, opened 2025-04-09
refactor: avoid empty display in errors

MichaelScofield:

#5814 

Some errors are annotated with `#[snafu(display("")]`, which could produce empty result if we accidently call `to_string` on it, like [here](https://github.com/GreptimeTeam/greptimedb/pull/5841/files). So it's not a good practice. This PR is trying to use Snafu's "transparent" attribute to avoid that.

## PR Checklist

### DIFF
--- a/src/common/error/tests/ext.rs
+++ b/src/common/error/tests/ext.rs
@@ -14,7 +14,7 @@
 
 use std::any::Any;
 
-use common_error::ext::{ErrorExt, PlainError};
+use common_error::ext::{ErrorExt, PlainError, StackError};
 use common_error::status_code::StatusCode;
 use common_macro::stack_trace_debug;
 use snafu::{Location, ResultExt, Snafu};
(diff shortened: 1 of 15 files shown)


src/common/error/tests/ext.rs
```diff
@@ -0,0 +1,102 @@
...
+fn test_output_msg() {
+    let result = normal_error();
+    assert_eq!(
+        result.unwrap_err().output_msg(),
+        r#"A normal error with "display" attribute, message "blabla": <root cause>"#
+    );
+
+    let result = transparent_error();
+    assert_eq!(result.unwrap_err().output_msg(), "<root cause>");
+}
```
evenyag, line 73: Can we also add a test to assert the `transparent()` method?

evenyag (approved):
Cool 👍

src/common/error/tests/ext.rs
MichaelScofield: sure, added

fengjiachun (approved):
LGTM

PR #5865, opened 2025-04-09
feat: implement process manager and  information_schema.process_list 

v0y4g3r:

- #5866 

This PR adds implementation of `ProcessManager` and `information_schema.process_list` table used to track running queries.

This is the first step towards process management. To reduce PR size, currently no query will be registered to `ProcessManager`.

## PR Checklist


src/common/meta/src/key/process_list.rs
```diff
@@ -0,0 +1,222 @@
...
+            id,
+        })
+    }
+}
+
+/// Detail value of process.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ProcessValue {
+    /// Database name.
+    pub database: String,
```
sunng87, line 80: Note that a query can cross schema, so it's hard to judge which "schema" the query is belongs to. We can use catalog here. When using greptimedb in multi-tenant mode, we use catalog to isolate tenants. Each connection should belongs to only one catalog.

src/common/meta/src/key/process_list.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// The running query sql.
+    pub query: String,
+    /// Query start timestamp in milliseconds.
+    pub start_timestamp_ms: i64,
```
sunng87, line 84: We will also need to include some information of the client, there is a chance multiple clients issuing same query at same time, it's hard to judge process owner just with `start_timestamp_ms`.

src/catalog/src/system_schema/information_schema/process_list.rs
```diff
@@ -0,0 +1,163 @@
...
+    let predicates = Predicates::from_scan_request(&Some(request));
+    let current_time = current_time_millis();
+    let queries = process_manager.list_all_processes()?;
+
+    let mut id_builder = UInt64VectorBuilder::with_capacity(queries.len());
+    let mut database_builder = StringVectorBuilder::with_capacity(queries.len());
+    let mut query_builder = StringVectorBuilder::with_capacity(queries.len());
+    let mut start_time_builder = TimestampMillisecondVectorBuilder::with_capacity(queries.len());
+    let mut elapsed_time_builder = DurationMillisecondVectorBuilder::with_capacity(queries.len());
+
```
sunng87, line 124: We need to ensure this table only contains queries from current catalog. When using in multi-tenant mode, we should not allow one tenant to see global process list.

There is an exception that when connected to greptime catalog, we can show global state.

sunng87: client information like user, client ip/port are better to be included in this table.
https://dev.mysql.com/doc/refman/8.4/en/information-schema-processlist-table.html

sunng87:
One more idea, instead of this push model that each frontend reports this information to meta, can we make it a pull model that when we doing query against information.process_list, the frontend issues a request to fetch instant state from all siblings.

Pros:

1. More instant data
2. No overhead when no query is on process_list

Cons:

1. Need cache when large quality of queries on process_list

sunng87:
If we use pull mode, we can first query meta to get a list of live frontend nodes, then broadcast to all these frontends. I think we can afford the cost because it won't be a frequent query.

The question is I remember we cannot do async operation on information schema. Not sure if it's still valid.

MichaelScofield:
There's another issue regarding push mode: the pressure on the metasrv. The process list might be huge in a busy cluster, and the reporting tasks may submit too frequent for metasrv in a large cluster. Both produced extra network and storage issues for metasrv.

v0y4g3r:
> There's another issue regarding push mode: the pressure on the metasrv. The process list might be huge in a busy cluster, and the reporting tasks may submit too frequent for metasrv in a large cluster. Both produced extra network and storage issues for metasrv.

For most queries that live less than the report interval, they won't be stored in metasrv or backend storage. It's not expected to have many long-running queries in any cluster.

killme2008:
What's the status of this PR?

v0y4g3r:
@sunng87 @MichaelScofield I refactored the original implementation to a pull based mode. Now the implementation will ask every frontend instance for all running queries when executing `SELECT * FROM information_schema.process_list`.

There're still some todos:
- client info now is not provided. We should carry client info in `QueryContext`
- catalog filtering is not implemented. We should add a new optimizaition rule to filter only the queries running in current user's catalog.

I think these does not block current PR and I will implement them in a following PR because current PR is already very big.

src/catalog/src/process_manager.rs
```diff
@@ -0,0 +1,181 @@
...
+impl ProcessManager {
+    /// Registers a submitted query.
+    pub fn register_query(
+        self: &Arc<Self>,
+        catalog: String,
+        schemas: Vec<String>,
+        query: String,
+        client: String,
+    ) -> Ticket {
+        let id = self.next_id.fetch_add(1, Ordering::Relaxed);
```
sunng87: Due to historical reason, the process_id in postgres protocol is assigned to connection because there is no multiplex support in postgres protocol, so killing a query is killing a connection, which is backed by a process. That's why it's called Process. I think mysql works like this too.

I will suggest to add an optional `pid` argument for this function so when query is registered from postgres or mysql, we will use the pre-assigned pid, while if it's from http or other stateless interface, pid is generated per-query.

This will also require a new API from `ProcessManager` called like `next_pid()` and it has to be called upon connection of postgres and mysql.

src/catalog/src/process_manager.rs
```diff
@@ -0,0 +1,181 @@
...
+        let mut processes = vec![];
+        if let Some(remote_frontend_selector) = self.frontend_selector.as_ref() {
+            let frontends = remote_frontend_selector
+                .select(|node| node.peer.addr != self.server_addr)
+                .await
+                .context(error::ListProcessSnafu)?;
+            for mut f in frontends {
+                processes.extend(
+                    f.list_process(ListProcessRequest {
+                        catalog: catalog.unwrap_or_default().to_string(),
```
sunng87, line 136: For catalog filtering:

1. if current catalog is greptime, and catalog parameter is not provided, we can list all processes from current cluster
2. if current catalog is not greptime, we can only list processes from current catalog.

sunng87 (approved):
LGTM as the first part. I will be working on pid management and termination in parallel, start from postgres first.

src/catalog/src/process_manager.rs
v0y4g3r: added in e516ae49e

sunng87: Looks good. I have opened https://github.com/GreptimeTeam/greptimedb/pull/6301/ for `Session` and `QueryContext`.

src/catalog/src/system_schema/information_schema/process_list.rs
```diff
@@ -0,0 +1,189 @@
...
+/// Build running process list.
+async fn make_process_list(
+    process_manager: ProcessManagerRef,
+    request: ScanRequest,
+) -> error::Result<RecordBatch> {
+    let predicates = Predicates::from_scan_request(&Some(request));
+    let current_time = current_time_millis();
+    // todo(hl): find a way to extract user catalog to filter queries from other users.
+    let queries = process_manager.local_processes(None)?;
```
MichaelScofield: Is it `list_all_processes`?

src/common/frontend/src/error.rs
```diff
@@ -36,6 +65,9 @@ impl ErrorExt for Error {
         use Error::*;
         match self {
             External { source, .. } => source.status_code(),
+            Meta { source, .. } => source.status_code(),
+            ParseProcessId { .. } => StatusCode::InvalidArguments,
+            ListProcess { .. } | CreateChannel { .. } => StatusCode::External,
```
MichaelScofield: Why not use its source's status code for `CreateChannel`?

PR #5874, opened 2025-04-10
fix: conn pool leak

discord9:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

fix a bug when `ChannelManager` starting recycle connection pool, the pool can't be released even all ChannelManager using it dropped, hence cause pool being leaked

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #5876, opened 2025-04-10
fix: gRPC connection pool leak

fengys1996:

#5874

This pr mainly do two things:
1. fix gRPC connection pool leak.
2. refactor the ChannelManager to make clone more lightweight.

## PR Checklist

### DIFF
--- a/src/common/grpc/src/channel_manager.rs
+++ b/src/common/grpc/src/channel_manager.rs
@@ -240,8 +240,8 @@ impl ChannelManager {
         }
 
         endpoint = endpoint
-            .tcp_keepalive(self.inner.config.tcp_keepalive)
-            .tcp_nodelay(self.inner.config.tcp_nodelay);
+            .tcp_keepalive(self.config().tcp_keepalive)
+            .tcp_nodelay(self.config().tcp_nodelay);
 
         Ok(endpoint)
     }
--- a/src/common/grpc/src/channel_manager.rs
+++ b/src/common/grpc/src/channel_manager.rs
@@ -63,10 +63,8 @@ impl Default for Inner {
 
 impl Drop for Inner {
     fn drop(&mut self) {
-        // Cancel the channel recycle task if it is running.
-        if self.channel_recycle_started.load(Ordering::Relaxed) {
-            self.cancel.cancel();
-        }
+        // Cancel the channel recycle task.
+        self.cancel.cancel();
     }
 }
 
--- a/src/common/grpc/src/channel_manager.rs
+++ b/src/common/grpc/src/channel_manager.rs
@@ -63,7 +63,10 @@ impl Default for Inner {
 
 impl Drop for Inner {
     fn drop(&mut self) {
-        self.cancel.cancel();
+        // Cancel the channel recycle task if it is running.
+        if self.channel_recycle_started.load(Ordering::Relaxed) {
+            self.cancel.cancel();
+        }
     }
 }
 
--- a/src/common/grpc/src/channel_manager.rs
+++ b/src/common/grpc/src/channel_manager.rs
@@ -671,4 +671,49 @@ mod tests {
             true
         });
     }
+
+    #[tokio::test]
+    async fn test_pool_release_with_channel_recycle() {
+        let mgr = ChannelManager::new();
+
+        let pool_holder = mgr.pool().clone();
+
+        // start channel recycle task
+        let addr = "test_addr";
+        let _ = mgr.get(addr);
+
+        let mgr_clone_1 = mgr.clone();
+        let mgr_clone_2 = mgr.clone();
+        assert_eq!(3, Arc::strong_count(mgr.pool()));
+
+        drop(mgr_clone_1);
+        drop(mgr_clone_2);
+        assert_eq!(3, Arc::strong_count(mgr.pool()));
+
+        drop(mgr);
+
+        // wait for the channel recycle task to finish
+        tokio::time::sleep(Duration::from_millis(10)).await;
+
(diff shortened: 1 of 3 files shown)


src/common/grpc/src/channel_manager.rs
```diff
@@ -39,40 +40,67 @@ lazy_static! {
...
 
-impl ChannelManager {
-    pub fn new() -> Self {
-        Default::default()
+impl Drop for Inner {
+    fn drop(&mut self) {
+        // Cancel the channel recycle task if it is running.
+        if self.channel_recycle_started.load(Ordering::Relaxed) {
+            self.cancel.cancel();
+        }
```
fengys1996, line 69: Do we need to determine whether the channel recycle task is running?

src/common/grpc/src/channel_manager.rs
```diff
@@ -172,65 +206,67 @@ impl ChannelManager {
...
         {
             return;
         }
 
-        let pool = self.pool.clone();
+        let pool = self.pool().clone();
+        let cancel = self.inner.cancel.clone();
         let _handle = common_runtime::spawn_global(async {
-            recycle_channel_in_loop(pool, RECYCLE_CHANNEL_INTERVAL_SECS).await;
+            recycle_channel_in_loop(pool, cancel, RECYCLE_CHANNEL_INTERVAL_SECS).await;
```
discord9, line 265: how about just pass `Inner` into it and remove arc on both pool&channel_recycle_started?

src/common/grpc/src/channel_manager.rs
```diff
@@ -443,11 +479,15 @@ impl Pool {
...
 }
 
-async fn recycle_channel_in_loop(pool: Arc<Pool>, interval_secs: u64) {
+async fn recycle_channel_in_loop(pool: Arc<Pool>, cancel: CancellationToken, interval_secs: u64) {
     let mut interval = tokio::time::interval(Duration::from_secs(interval_secs));
 
     loop {
-        let _ = interval.tick().await;
+        tokio::select! {
+            _ = cancel.cancelled() => break,
```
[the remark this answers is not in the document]
fengys1996: Good suggestion!

src/common/grpc/src/channel_manager.rs
fengys1996: The Arc of channel_recycle_started can be removed. Pool not, since Pool will share between `Inner` and `channel recyle task`.

fengys1996:
@discord9 PTAL

src/common/grpc/src/channel_manager.rs
discord9: > The Arc of channel_recycle_started can be removed. Pool not, since Pool will share between `Inner` and `channel recyle task`.

How about directly pass Arc<Inner> to recycle task? Since now most field of `Inner` is passed to `recycle_channel_in_loop` anyway

fengys1996: This will cause the loop always hold a reference of inner, and then the drop logic of inner cannot be executed.

discord9: > This will cause the loop always hold a Arc, and then the drop logic of inner cannot be executed.

how about a weak ref

discord9:
a unit test to make sure pool get freed after clone channel manager many times and drop them would be nice

src/common/grpc/src/channel_manager.rs
fengys1996: Rust's ownership mechanism ensures that this does not happen.

src/common/grpc/src/channel_manager.rs
fengys1996: In general, I tend not to use weak refs.

fengys1996:
> a unit test to make sure pool get freed after clone channel manager many times and drop them would be nice

Added in https://github.com/GreptimeTeam/greptimedb/pull/5876/commits/3b2a88b04292057c14e346a088cf50c20e8058b5

src/common/grpc/src/channel_manager.rs
discord9: > Rust's ownership mechanism ensures that this does not happen.

Both the `load` and `cas` atomic op have `Relaxed` order so out of order execution could bring surprise.
i.e. if `cas` happen sightly early, then `drop` is called immediately `load` may not read `true`, of course in reality this case might never happen but the potential data race is still here

fengys1996: >  of course in reality this case might never happen but the potential data race is still here

Makes sense. While this will never happen, the potential for a data race exists. 

> Do we need to determine whether the channel recycle task is running before cancel?

Then we directly execute `self.cancel.cancel();` in the drop method.

fengys1996: Fix in https://github.com/GreptimeTeam/greptimedb/pull/5876/commits/b18831cedffd57a0c4c3c2eac64f1735286f3798

src/common/grpc/src/channel_manager.rs
discord9: > In general, I tend not to use weak ref.

not using weak still sound semantically misfitting, a task that refresh stuffs for the pool shouldn't hold a strong count in isolation, this kind of break the tree model of data that Arc should be using(creating a root for tree that floating in the void), this is exactly where weak should be use I think, see more at https://effective-rust.com/references.html?highlight=Weak#smart-pointer-types

discord9 (approved):
LGTM

shuiyisong (approved):
🚀

PR #5877, opened 2025-04-10
feat: introduce `high_watermark` for remote wal logstore

CookiePieWw:

For regions without write for a long time, it will have a legacy `flush_entry_id`. This patch update a `high_watermark` for a region when a flush request found memtables are empty and then send it to metasrv for further active pruning.

Since the `high_watermark` is updated when memtables are empty, it's safe for metasrv to delete all records before `high_watermark` for this region.

Correspondingly, `log_store` will maintain the `topic-high_watermark` map for all topics. Regions retrieve them when they need to update the `high_watermark`s of themselves.

## PR Checklist

### DIFF
--- a/src/log-store/src/kafka/log_store.rs
+++ b/src/log-store/src/kafka/log_store.rs
@@ -18,6 +18,7 @@ use std::time::Duration;
 
 use common_telemetry::{debug, warn};
 use common_wal::config::kafka::DatanodeKafkaConfig;
+use dashmap::DashMap;
 use futures::future::try_join_all;
 use futures_util::StreamExt;
 use rskafka::client::partition::OffsetAt;
@@ -52,6 +53,8 @@ pub struct KafkaLogStore {
     consumer_wait_timeout: Duration,
     /// Ignore missing entries during read WAL.
     overwrite_entry_start_id: bool,
+    /// High watermark offset of the last record in the topic.
+    high_watermark: DashMap<String, u64>,
 }
 
 impl KafkaLogStore {
@@ -68,6 +71,7 @@ impl KafkaLogStore {
             max_batch_bytes: config.max_batch_bytes.as_bytes() as usize,
             consumer_wait_timeout: config.consumer_wait_timeout,
             overwrite_entry_start_id: config.overwrite_entry_start_id,
+            high_watermark: DashMap::new(),
         })
     }
 }
@@ -158,6 +162,7 @@ impl LogStore for KafkaLogStore {
             .collect::<HashSet<_>>();
         let mut region_grouped_records: HashMap<RegionId, (OrderedBatchProducerRef, Vec<_>)> =
             HashMap::with_capacity(region_ids.len());
+        let mut region_to_provider = HashMap::with_capacity(region_ids.len());
         for entry in entries {
             let provider = entry.provider().as_kafka_provider().with_context(|| {
                 error::InvalidProviderSnafu {
@@ -165,6 +170,7 @@ impl LogStore for KafkaLogStore {
                     actual: entry.provider().type_name(),
                 }
             })?;
+            region_to_provider.insert(entry.region_id(), provider.clone());
             let region_id = entry.region_id();
             match region_grouped_records.entry(region_id) {
                 std::collections::hash_map::Entry::Occupied(mut slot) => {
@@ -198,6 +204,13 @@ impl LogStore for KafkaLogStore {
                 },
             ))
             .await?;
+
+        // Updates the high watermark offset of the last record in the topic.
+        for (region_id, offset) in &region_grouped_max_offset {
+            // Safety: `region_id` is always valid.
+            let provider = region_to_provider.get(region_id).unwrap();
+            self.high_watermark.insert(provider.topic.clone(), *offset);
+        }
 
         Ok(AppendBatchResponse {
             last_entry_ids: region_grouped_max_offset.into_iter().collect(),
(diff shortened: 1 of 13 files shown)


src/log-store/src/kafka/log_store.rs
```diff
@@ -199,6 +205,13 @@ impl LogStore for KafkaLogStore {
             ))
             .await?;
 
+        // Updates the high watermark offset of the last record in the topic.
+        for (region_id, offset) in &region_grouped_max_offset {
+            // Safety: `region_id` is always valid.
+            let provider = region_to_provider.get(region_id).unwrap();
+            self.high_watermark.insert(provider.topic.clone(), *offset);
+        }
```
WenyXu, line 213: The watermark is only updated on writes. We might need to consider adding a periodic task to fetch the latest offset to keep the watermark up to date even when there are no writes.

src/mito2/src/worker.rs
```diff
@@ -915,6 +915,9 @@ impl<S: LogStore> RegionWorkerLoop<S> {
                 self.handle_manifest_region_change_result(req).await
             }
             BackgroundNotify::RegionEdit(req) => self.handle_region_edit_result(req).await,
+            BackgroundNotify::UpdateHighWatermark => {
+                self.handle_update_high_watermark_request(region_id).await
+            }
```
WenyXu, line 920: Can we handle the high-watermark updating in `handle_flush_request`?

src/mito2/src/flush.rs
```diff
@@ -487,8 +501,12 @@ impl FlushScheduler {
...
         if version.memtables.is_empty() {
             debug_assert!(!self.region_status.contains_key(&region_id));
-            // The region has nothing to flush.
-            task.on_success();
+            // The region has nothing to flush. Submit a job to update high watermark.
+            let job = task.into_update_high_watermark_job();
+            if let Err(e) = self.scheduler.schedule(job) {
+                error!(e; "Update high watermark for region {}", region_id);
+                return Err(e);
+            }
```
WenyXu, line 509: Do we need to update it again?

CookiePieWw: Next time when we recieve a flush request it will be updated if the memtable is still empty.

fengjiachun (approved):
LGTM

fengjiachun:
PTAL at the `mito` engine section.  @evenyag

src/mito2/src/region.rs
```diff
@@ -119,6 +119,9 @@ pub(crate) struct MitoRegion {
     last_compaction_millis: AtomicI64,
     /// Provider to get current time.
     time_provider: TimeProviderRef,
+    /// The highest watermark of remote WAL since last flushing.
+    /// Sends this to meta-srv for active remote WAL pruning.
+    pub(crate) high_watermark_since_flush: AtomicU64,
```
WenyXu, line 124:
```suggestion
    /// The topic's latest entry id since the region's last flushing.
    /// **Only used for remote WAL pruning.**
    /// 
    /// The value will be updated to the high watermark of the topic
    /// if region receives a flush request or schedules a periodic flush task 
    /// and the region's memtable is empty.    
    ///
    /// There are no WAL entries in range [flushed_entry_id, latest_entry_id_since_flush] for current region,
    /// which means these WAL entries maybe able to be pruned up to `latest_entry_id_since_flush`.
```

src/mito2/src/worker/handle_alter.rs
```diff
@@ -40,7 +41,7 @@ use crate::region::MitoRegionRef;
 use crate::request::{DdlRequest, OptionOutputTx, SenderDdlRequest};
 use crate::worker::RegionWorkerLoop;
 
-impl<S> RegionWorkerLoop<S> {
+impl<S: LogStore> RegionWorkerLoop<S> {
```
WenyXu, line 44:
```suggestion
impl<S> RegionWorkerLoop<S> {
```

src/mito2/src/region.rs
evenyag: Please explain what the watermark is or what the highest watermark is.
Is it identical to the `latest_entry_id_since_flush`?

If so, can we rename this field to `topic_latest_entry_id`?

src/mito2/src/region.rs
```diff
@@ -292,7 +295,9 @@ impl MitoRegion {
         let manifest_usage = self.stats.total_manifest_size();
         let num_rows = version.ssts.num_rows() + version.memtables.num_rows();
         let manifest_version = self.stats.manifest_version();
-        let flushed_entry_id = version.flushed_entry_id;
+        let flushed_entry_id = version
+            .flushed_entry_id
+            .max(self.high_watermark_since_flush.load(Ordering::Relaxed));
```
evenyag, line 300: I think they are two different things. Please add a new field.

src/mito2/src/worker/handle_flush.rs
```diff
@@ -39,6 +40,12 @@ impl<S> RegionWorkerLoop<S> {
         let Some(region) = self.regions.flushable_region_or(region_id, &mut sender) else {
             return;
         };
+        self.update_high_watermark(&region);
+        info!(
+            "Region {} flush request, high watermark: {}",
+            region_id,
+            region.high_watermark_since_flush.load(Ordering::Relaxed)
+        );
```
evenyag, line 48: We should update when the flush is finished (e.g. `handle_flush_finished()`). Here we only trigger the flush job.

WenyXu: No, it's expected to update the high watermark before going the `schedule_flush` method

WenyXu (approved):
Rest LGTM

src/mito2/src/region.rs
CookiePieWw: Renamed to `topic_latest_entry_id` and added a new field in [RegionStatistics] for it.

Issue #5739, opened 2025-03-19
[Tracking Issues] Follower regions


WenyXu:
### What problem does the new feature solve?

Region Followers allow read and write operations to be separated across different datanodes, improving query performance and overall system throughput. The core idea is to introduce Follower regions, where read requests can be served by follower regions, while writes remain handled by the leader.

### What does the feature do?

**Goals**

- Support Region Follower: Allow regions to exist as followers across multiple datanodes
- Leader-Follower data synchronization: Ensure that followers receive updates from the leader
- Follower read support: Enable query execution on follower regions to reduce load on the leader
- Failover mechanism: Design a process where a follower can take over leadership if needed

**TODO**
- [x] Install manifest #5742 
- [x] Sync region manifest #5765 
- [x] Region lease renewer #5762 
- [x] Leader region info collector #5811
- [x] Support metric engine
   - [x] Sync region
   - [x] Handling logical table registration 
- [x] Notify followers to sync the manifest in DDL procedures #5901
- [ ] Handling table deregistration
- [ ] Alter options
- [x] Auto flush metadata region
- [ ] Support flow

PR #5878, opened 2025-04-10
fix: ensure logical regions are synced during region sync

WenyXu:

#5739 

This PR makes the metric engine region sync to properly handle logical regions synchronization. Previously, logical regions were not being synchronized even when the metadata manifest version was updated. Now, when `metadata_synced` is true, the system correctly recovers states from the metadata region and syncs logical regions. 

## PR Checklist


src/datanode/src/region_server.rs
```diff
@@ -308,20 +308,36 @@ impl RegionServer {
...
+        else {
+            return Ok(());
+        };
+
+        for region in new_opened_regions {
+            self.inner.region_map.insert(
+                region,
+                RegionEngineWithStatus::Ready(engine_with_status.engine().clone()),
+            );
+            debug!("Logical region {} is opened", region);
```
fengjiachun, line 337: Should it be at the `info` level?

src/metric-engine/src/metadata_region.rs
```diff
@@ -76,11 +78,34 @@ impl MetadataRegion {
         }
     }
 
-    pub async fn open_logical_region(&self, logical_region_id: RegionId) {
+    /// Open a logical region.
+    ///
+    /// Returns true if the logical region is opened for the first time.
+    pub async fn open_logical_region(&self, logical_region_id: RegionId) -> bool {
+        info!("open_logical_region: {:?}", logical_region_id);
```
fengjiachun, line 85: It's better to move it to the end and log the return value (true or false).

WenyXu: Ooops, I just forgot to remove the log here.

src/store-api/src/region_engine.rs
```diff
@@ -583,6 +583,60 @@ impl RegionStatistic {
...
+    /// Returns true if new manifest is installed.
+    pub fn synced(&self) -> bool {
+        match self {
+            SyncManifestResponse::NotSupported => false,
+            SyncManifestResponse::Mito { synced } => *synced,
+            SyncManifestResponse::Metric {
+                metadata_synced,
+                data_synced,
+                ..
+            } => *metadata_synced || *data_synced,
```
fengjiachun, line 610: A bit confusing, why it is `or` but not `and`

fengjiachun (approved):
Others LGTM

src/store-api/src/region_engine.rs
```diff
@@ -583,6 +583,60 @@ impl RegionStatistic {
...
+#[derive(Debug)]
+pub enum SyncManifestResponse {
+    NotSupported,
+    Mito {
+        synced: bool,
+    },
+    Metric {
+        metadata_synced: bool,
+        data_synced: bool,
+        new_opened_logical_region_ids: Vec<RegionId>,
```
evenyag, line 596: nitpick: Better to document the relationship between `new_opened_logical_region_ids` and other fields.

src/store-api/src/region_engine.rs
evenyag: We need some documents for the behavior. Maybe we should also comment when the region is considered sync above the enum variant definitions.

src/metric-engine/src/metadata_region.rs
```diff
@@ -509,6 +534,10 @@ impl MetadataRegion {
             .await
             .context(MitoWriteOperationSnafu)?;
 
+        for (logical_region_id, _) in logical_regions {
+            self.add_logical_region(logical_region_id).await?;
+        }
```
evenyag, line 539: Is it safe to add all logical regions without considering existence? When do we call `add_logical_regions()`? Maybe more documents are needed.

src/store-api/src/region_engine.rs
WenyXu: resolved in [cf4fc5a](https://github.com/GreptimeTeam/greptimedb/pull/5878/commits/cf4fc5a5cf4118988545c04892de6777b42272a6)

src/store-api/src/region_engine.rs
WenyXu: added in [cf4fc5a](https://github.com/GreptimeTeam/greptimedb/pull/5878/commits/cf4fc5a5cf4118988545c04892de6777b42272a6)

evenyag (approved):

PR #5893, opened 2025-04-14
feat: report per-region metrics on region server

waynexia:

- Add `region_id` label to `greptime_datanode_handle_region_request_elapsed` metric
- Add new metric `greptime_datanode_region_changed_row_count` with per-region insert/delete row count

## PR Checklist

### DIFF
--- a/src/datanode/src/metrics.rs
+++ b/src/datanode/src/metrics.rs
@@ -20,13 +20,21 @@ pub const REGION_REQUEST_TYPE: &str = "datanode_region_request_type";
 
 pub const REGION_ROLE: &str = "region_role";
 pub const REGION_ID: &str = "region_id";
+pub const RESULT_TYPE: &str = "result";
 
 lazy_static! {
     /// The elapsed time of handling a request in the region_server.
     pub static ref HANDLE_REGION_REQUEST_ELAPSED: HistogramVec = register_histogram_vec!(
         "greptime_datanode_handle_region_request_elapsed",
         "datanode handle region request elapsed",
-        &[REGION_REQUEST_TYPE]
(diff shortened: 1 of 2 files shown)


src/datanode/src/region_server.rs
```diff
@@ -1104,6 +1112,7 @@ enum RegionChange {
     Register(RegionAttribute),
     Deregisters,
     Catchup,
+    Update,
```
sunng87, line 1115: The `Update` here can be confusing in future. How about changing to `Data`, which means the change is a *Data Change*?

src/datanode/src/metrics.rs
```diff
@@ -20,13 +20,21 @@ pub const REGION_REQUEST_TYPE: &str = "datanode_region_request_type";
...
         "datanode handle region request elapsed",
-        &[REGION_REQUEST_TYPE]
+        &[REGION_ID, REGION_REQUEST_TYPE]
+    )
+    .unwrap();
+    /// The number of rows in region request received by region server, labeled with request type.
+    pub static ref REGION_CHANGED_ROW_COUNT: IntCounterVec = register_int_counter_vec!(
+        "greptime_datanode_region_changed_row_count",
+        "datanode region changed row count",
+        &[REGION_ID, REGION_REQUEST_TYPE]
```
evenyag, line 37: If users use the metric engine, I wonder the metrics may have too many label values.

evenyag: Do we need changed row count for each logical region? @sunng87

Is it possible to report this in physical region level? Maybe we can add region id to mito's ingest rows metrics.

evenyag (approved):

PR #5895, opened 2025-04-14
feat: add query engine options

waynexia:

Add `[query]` section to frontend/datanode/flownode config

## PR Checklist


sunng87:
Just to confirm is this a limitation on global query concurrency or a parallelism of a single query. For example, with this configuration, if there are two concurrent queries coming in, will it runs on parallelism*2 cores on worst case?

The purpose is to have a limitation on max CPU cores that queries can use, and leave some resources for writes. It's a little a bit like read/write runtime we had.

sunng87 (approved):

evenyag (approved):

Issue #5892, opened 2025-04-14
Filter on new column with default value doesn't work


waynexia:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Datanode, Storage Engine

### Minimal reproduce step

- create a table
- write some data
- alter table add a new column with default value
- query the table with a filter on that new column

### What did you expect to see?

- correct query result

### What did you see instead?

- incorrect result, the filter doesn't work on new column

### What operating system did you use?

unrelated

### What version of GreptimeDB did you use?

0.13.2

### Relevant log output and stack trace

```bash

```

PR #5912, opened 2025-04-16
fix: filter doesn't consider default values after schema change

evenyag:

fixes https://github.com/GreptimeTeam/greptimedb/issues/5892

This PR uses the default value of the newly added column for pruning.
- The row group stats will use the default value as min/max
- If we have inverted indices, then we may not prune row groups by min-max stats. It evaluates the filter while creating the `SimpleFilterContext` if the SSt doesn't have the column.

### Unresolved Problem
If users add a column with `CURRENT_TIMESTAMP` as the default value, we will always create a new default value each time.
Fixing this may require us to record the added time of a column in the manifest and use that as the default value. We can create another issue for it, as it is not as urgent as this issue.

## PR Checklist


WenyXu (approved):

waynexia (approved):

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -1009,10 +1016,20 @@ impl ReaderState {
...
-/// Context to evaluate the column filter.
+/// The filter to evaluate or the prune result of the default value.
+pub(crate) enum MaybeFilter {
+    /// The filter to evaluate.
+    Filter(SimpleFilterEvaluator),
+    /// The filter matches the default value.
+    Matched,
+    /// The filter is pruned.
+    Pruned,
+}
```
waynexia, line 1027: This does not taste good...

PR #5919, opened 2025-04-17
perf: keep compiled regex in SimpleFilterEvaluator to avoid re-compiling

evenyag:

This PR compiles the regex while building the `SimpleFilterEvaluator` and uses the pre-compiled `Regex` to evaluate datums.

It also slightly modifies the `regexp_is_match_scalar()` function to support reusing the `Regex`.

## PR Checklist

### DIFF
--- a/src/common/recordbatch/src/filter.rs
+++ b/src/common/recordbatch/src/filter.rs
@@ -24,12 +24,17 @@ use datafusion_common::arrow::buffer::BooleanBuffer;
 use datafusion_common::arrow::compute::kernels::cmp;
 use datafusion_common::cast::{as_boolean_array, as_null_array, as_string_array};
 use datafusion_common::{internal_err, DataFusionError, ScalarValue};
-use datatypes::arrow::array::{Array, BooleanArray, RecordBatch};
+use datatypes::arrow::array::{
+    Array, ArrayAccessor, ArrayData, BooleanArray, BooleanBufferBuilder, RecordBatch,
+    StringArrayType,
+};
 use datatypes::arrow::compute::filter_record_batch;
+use datatypes::arrow::datatypes::DataType;
 use datatypes::arrow::error::ArrowError;
 use datatypes::compute::kernels::regexp;
 use datatypes::compute::or_kleene;
 use datatypes::vectors::VectorRef;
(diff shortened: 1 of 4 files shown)


src/common/recordbatch/src/filter.rs
```diff
@@ -254,6 +286,44 @@ pub fn batch_filter(
...
+    let null_bit_buffer = array.nulls().map(|x| x.inner().sliced());
+    let mut result = BooleanBufferBuilder::new(array.len());
+
+    if let Some(re) = regex {
+        for i in 0..array.len() {
+            let value = array.value(i);
+            result.append(re.is_match(value));
+        }
+    } else {
+        result.append_n(array.len(), true);
```
killme2008, line 308: So if the regexp is invalid or not a string, we assume that the results are matched, is that as expected?

killme2008 (approved):
Almost LGTM

src/common/recordbatch/src/filter.rs
```diff
@@ -179,10 +191,10 @@ impl SimpleFilterEvaluator {
...
             Operator::Gt => cmp::gt(input, &self.literal),
             Operator::GtEq => cmp::gt_eq(input, &self.literal),
-            Operator::RegexMatch => self.regex_match(input, false, false),
-            Operator::RegexIMatch => self.regex_match(input, true, false),
-            Operator::RegexNotMatch => self.regex_match(input, false, true),
-            Operator::RegexNotIMatch => self.regex_match(input, true, true),
+            Operator::RegexMatch => self.regex_match(input, false),
+            Operator::RegexIMatch => self.regex_match(input, false),
+            Operator::RegexNotMatch => self.regex_match(input, true),
+            Operator::RegexNotIMatch => self.regex_match(input, true),
```
waynexia, line 197: why not wrap the remaining boolean flag into the state

waynexia (approved):
rest LGTM

src/common/recordbatch/src/filter.rs
evenyag: In the prune phase, this is fine. But I'll improve this by not building a `FilterEvaluator` if the regex is invalid.

PR #5943, opened 2025-04-21
chore: use `once_cell` to avoid parse everytime in pipeline exec

shuiyisong:

As the title suggests.

## PR Checklist


waynexia (commented):
one potential hazard is both `options` and `flatten_json_object` are `pub`, which means they can be modified by outer code and lead to inconsistent states. how about wrapping getter/setter for them?

sunng87 (approved):

PR #5951, opened 2025-04-22
feat: introduce flush metadata region task for metric engine

WenyXu:

Added a new `FlushMetadataRegionTask` to periodically flush metadata regions in the metric engine

## PR Checklist

### DIFF
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -602,6 +602,9 @@ fork_dictionary_bytes = "1GiB"
 ## Whether to enable the experimental sparse primary key encoding.
 experimental_sparse_primary_key_encoding = false
 
+## The interval to flush the metadata region.
+flush_metadata_region_interval = "30s"
+
 ## The logging options.
 [logging]
 ## The directory to store the log files. If set to empty, logs will not be written to files.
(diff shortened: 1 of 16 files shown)


src/metric-engine/src/engine.rs
```diff
@@ -359,19 +362,39 @@ impl RegionEngine for MetricEngine {
...
+        let state = Arc::new(RwLock::default());
+        let flush_interval = if config.flush_metadata_region_interval.is_zero() {
+            warn!(
+                "Flush metadata region interval is zero, override with default value: {:?}. Disable metadata region flush is forbidden.",
+                DEFAULT_FLUSH_METADATA_REGION_INTERVAL
+            );
+            DEFAULT_FLUSH_METADATA_REGION_INTERVAL
+        } else {
+            config.flush_metadata_region_interval
+        };
```
evenyag, line 377: non-blocking: We can implement this as the `sanitize()` method of the `EngineConfig`.

config/datanode.example.toml
```diff
@@ -602,6 +602,9 @@ fork_dictionary_bytes = "1GiB"
 ## Whether to enable the experimental sparse primary key encoding.
 experimental_sparse_primary_key_encoding = false
 
+## The interval to flush the metadata region.
+flush_metadata_region_interval = "30s"
```
evenyag, line 606: Is it necessary to expose this config now?

evenyag (approved):

config/datanode.example.toml
fengjiachun: I think we should allow metasrv to handle this, but it may be in the next RP.

fengjiachun (approved):
Others LGTM

Issue #5959, opened 2025-04-23
Failed to recover flows at service restart


adminpass:
### What type of bug is this?

Crash

### What subsystems are affected?

Standalone mode

### Minimal reproduce step

```sql
CREATE DATABASE jsdp_log;
USE jsdp_log;

CREATE TABLE IF NOT EXISTS `api_log` (
  `time` TIMESTAMP(9) NOT NULL,
  `key` STRING NULL SKIPPING INDEX WITH(granularity = '1024', type = 'BLOOM'),
  `status_code` TINYINT NULL,
  `method` STRING NULL,
  `path` STRING NULL,
  `raw_query` STRING NULL,
  `user_agent` STRING NULL,
  `client_ip` STRING NULL,
  `duration` INT NULL,
  `count` INT NULL,
  TIME INDEX (`time`)
) ENGINE=mito WITH(
  append_mode = 'true'
);

/* 创建 sink 表 */
CREATE TABLE IF NOT EXISTS `api_stats` (
  `time` TIMESTAMP(0) NOT NULL,
  `key` STRING NULL,
  `qpm` BIGINT NULL,
  `rpm` BIGINT NULL,
  `update_at` TIMESTAMP(3) NULL,
  TIME INDEX (`time`),
  PRIMARY KEY (`key`)
) ENGINE=mito WITH(
  append_mode = 'false',
  merge_mode = 'last_row'
);

/* 创建 flow  */
CREATE FLOW IF NOT EXISTS api_stats_flow
SINK TO api_stats EXPIRE AFTER '10 minute'::INTERVAL AS
SELECT date_trunc('minute', `time`::TimestampSecond) AS `time1`, `key`, count(*), sum(`count`)
FROM api_log
GROUP BY `time1`, `key`;

INSERT INTO `api_log` (`time`, `key`, `status_code`, `method`, `path`, `raw_query`, `user_agent`, `client_ip`, `duration`, `count`) VALUES ('2025-01-15 09:02:57.976857', '1', 0, 'GET', '/lightning/v1/query', 'key=1&since=600', 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36', '1', 21, 1);
```

### GreptimeDB server restart...

```SQL
[...]

PR #5963, opened 2025-04-23
fix: store flow query ctx on creation

discord9:

#5959

flow need to parse `raw_sql`, ~hence if `schema!=public`, on recover flows it will failed to find correct table if `raw_sql` ref to tables not in `public` schema, fix is to store schema for flows, and when recover flows using it~
hence when recovering flow using `raw_sql`, need the same `QueryContext` so that the sql can be parsed as the same logical plan

- local test confirm upgrade metadata is compatible(the new query context field is `#[serde(default)]`)

## PR Checklist


fengjiachun (approved):
LGTM

WenyXu (approved):

PR #5974, opened 2025-04-24
fix: conn timeout&refactor: better err msg

discord9:

some fixes for flow, including:
- add a 3s conn timeout for grpc
- refactor for better error message when rewrite failed
- fix `udd_calc` panic by checking for empty first, before this fix a empty input(or a null array input) will cause panic when calling `udd_calc`
- add sqlness for `udd_calc` and `hll`
- add retry for frontend client so network fluctuate won't affect flow execution (also when recovering flow, this retry make it more resistant in case frontend is still starting) 
- fix datafusion to substrait encode by replace `cast(null, <data type>)` to Typed Null in substrait, see https://github.com/waynexia/arrow-datafusion/pull/116 and new unit test
- use the same analyzer as normal query use (`CountWildcardToTimeIndexRule` instead of `CountWildcardRule`)

## PR Checklist

### DIFF
--- a/src/flow/src/batching_mode.rs
+++ b/src/flow/src/batching_mode.rs
@@ -34,7 +34,7 @@ pub const SLOW_QUERY_THRESHOLD: Duration = Duration::from_secs(60);
 const MIN_REFRESH_DURATION: Duration = Duration::new(5, 0);
 
 /// Grpc connection timeout
-const GRPC_CONN_TIMEOUT: Duration = Duration::from_secs(3);
+const GRPC_CONN_TIMEOUT: Duration = Duration::from_secs(5);
 
 /// Grpc max retry number
 const GRPC_MAX_RETRIES: u32 = 3;
--- a/src/flow/src/batching_mode.rs
+++ b/src/flow/src/batching_mode.rs
@@ -32,3 +32,7 @@ pub const SLOW_QUERY_THRESHOLD: Duration = Duration::from_secs(60);
 
 /// The minimum duration between two queries execution by batching mode task
 const MIN_REFRESH_DURATION: Duration = Duration::new(5, 0);
+
+const GRPC_CONN_TIMEOUT: Duration = Duration::from_secs(3);
+
+const GRPC_MAX_RETRIES: u32 = 3;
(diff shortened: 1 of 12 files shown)


MichaelScofield (approved):

src/flow/src/batching_mode.rs
```diff
@@ -32,3 +32,9 @@ pub const SLOW_QUERY_THRESHOLD: Duration = Duration::from_secs(60);
 
 /// The minimum duration between two queries execution by batching mode task
 const MIN_REFRESH_DURATION: Duration = Duration::new(5, 0);
+
+/// Grpc connection timeout
+const GRPC_CONN_TIMEOUT: Duration = Duration::from_secs(3);
```
killme2008, line 37: I think the default timeout value may be too small. A value of at least 5 seconds is much more reasonable.

src/flow/src/batching_mode/frontend_client.rs
```diff
@@ -223,12 +228,32 @@ impl FrontendClient {
...
-                    })
+                let mut retry = 0;
+
+                loop {
+                    let ret = db.database.handle(req.clone()).await.with_context(|_| {
+                        InvalidRequestSnafu {
+                            context: format!("Failed to handle request: {:?}", req),
+                        }
+                    });
+                    if let Err(err) = ret {
```
killme2008, line 239: I think it's better to retry only when it's retryable.

evenyag, line 239: nitpick: Do we need to check the status code of the error `StatusCode::is_retryable()` to ensure we can retry this? e.g. It's not invalid argument error.

killme2008 (approved):
Almost LGTM

evenyag (approved):
Others LGTM

src/flow/src/batching_mode.rs
killme2008, line 37:
```suggestion
const GRPC_CONN_TIMEOUT: Duration = Duration::from_secs(5);
```

src/flow/src/batching_mode/frontend_client.rs
discord9: #5987

PR #5983, opened 2025-04-25
ci: make install script to start greptimedb standalone by default

killme2008:

As the title said

## PR Checklist


sunng87:
This is not recommended. Start a process in install script is unexpected. Also someone could use this script in there docker build file and this will break it.

killme2008:
> This is not recommended. Start a process in install script is unexpected. Also someone could use this script in there docker build file and this will break it.

Good point!

PR #5995, opened 2025-04-27
feat: flush leader region before downgrading

WenyXu:

This change helps the system handle region migration more smoothly. By flushing the leader region before downgrading it, we can:
- Minimize the time window where the region is not writable
- Reduce the impact on write operations during region migration

### Changes
- Added a new `FlushLeaderRegion` state in the region migration procedure
- Modified the region migration flow to flush the leader region before downgrading it
- Added new metrics for tracking flush leader region operation duration
- Added new test cases for the flush leader region functionality

## PR Checklist

### DIFF
--- a/src/common/meta/src/instruction.rs
+++ b/src/common/meta/src/instruction.rs
@@ -217,7 +217,9 @@ pub enum Instruction {
     /// Invalidates batch cache.
     InvalidateCaches(Vec<CacheIdent>),
     /// Flushes regions.
-    FlushRegion(FlushRegions),
+    FlushRegions(FlushRegions),
+    /// Flushes a single region.
+    FlushRegion(RegionId),
 }
 
 /// The reply of [UpgradeRegion].
@@ -248,6 +250,7 @@ pub enum InstructionReply {
     CloseRegion(SimpleReply),
     UpgradeRegion(UpgradeRegionReply),
     DowngradeRegion(DowngradeRegionReply),
+    FlushRegion(SimpleReply),
 }
 
 impl Display for InstructionReply {
@@ -259,6 +262,7 @@ impl Display for InstructionReply {
             Self::DowngradeRegion(reply) => {
                 write!(f, "InstructionReply::DowngradeRegion({})", reply)
             }
+            Self::FlushRegion(reply) => write!(f, "InstructionReply::FlushRegion({})", reply),
         }
     }
 }
--- a/src/meta-srv/src/procedure/region_migration.rs
+++ b/src/meta-srv/src/procedure/region_migration.rs
@@ -124,10 +124,14 @@ pub struct Metrics {
 
 impl Display for Metrics {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let total = self.flush_leader_region_elapsed
+            + self.downgrade_leader_region_elapsed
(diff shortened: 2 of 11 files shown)


src/meta-srv/src/procedure/region_migration.rs
```diff
@@ -111,6 +112,8 @@ impl PersistentContext {
 pub struct Metrics {
     /// Elapsed time of downgrading region and upgrading region.
     operations_elapsed: Duration,
+    /// Elapsed time of flushing leader region.
+    flush_leader_region_elapsed: Duration,
```
fengjiachun, line 116: I don't think we need to include this in the migration operation timeout.

src/meta-srv/src/procedure/region_migration/flush_leader_region.rs
```diff
@@ -0,0 +1,281 @@
...
+
+use crate::error::{self, Error, Result};
+use crate::handler::HeartbeatMailbox;
+use crate::procedure::region_migration::update_metadata::UpdateMetadata;
+use crate::procedure::region_migration::{Context, State};
+use crate::service::mailbox::Channel;
+
+/// Flushes the leader region.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct FlushLeaderRegion;
```
fengjiachun, line 33: `PreFlushRegion` ?

WenyXu: Resolved in [f55902e](https://github.com/GreptimeTeam/greptimedb/pull/5995/commits/f55902e25aa608f25e2c35009dd694e204134036)

src/meta-srv/src/procedure/region_migration.rs
WenyXu: Resolved in [f55902e](https://github.com/GreptimeTeam/greptimedb/pull/5995/commits/f55902e25aa608f25e2c35009dd694e204134036)

fengjiachun (approved):
LGTM

src/common/meta/src/instruction.rs
```diff
@@ -217,7 +217,9 @@ pub enum Instruction {
     /// Invalidates batch cache.
     InvalidateCaches(Vec<CacheIdent>),
     /// Flushes regions.
-    FlushRegion(FlushRegions),
+    FlushRegions(FlushRegions),
+    /// Flushes a single region.
+    FlushRegion(RegionId),
```
MichaelScofield, line 222: Why do we need another instruction for flushing a single region provided that there's already an instruction for flushing multiple (including one of course) regions?

WenyXu: FlushRegions works asynchronously and serves as a non-mandatory hint, while FlushRegion enforces a synchronous flush operation. We will later rename FlushRegions to FlushRegionsHint or FlushRegionsSuggestion to make its purpose clearer.

MichaelScofield (commented):
Why not make the leader **flushing** inside downgrading it? That way you can avoid addition writes between flush and downgrade.

WenyXu:
> Why not make the leader **flushing** inside downgrading it? That way you can avoid addition writes between flush and downgrade.

By flushing the leader region before downgrading it, we can minimize the time window where the region is not writable.

MichaelScofield (approved):

PR #5996, opened 2025-04-27
fix: prune primary key with multiple columns may use default value as statistics

evenyag:

- https://github.com/GreptimeTeam/greptimedb/pull/5912

This PR fixes a critical issue introduced in #5912 

We will use the default value for non-first columns in the primary key to fill the min/max statistics. This is incorrect.

This PR adds a new enum `StatValues` to distinguish no column and no stats. It only uses the default value for stats when there is no column.

## PR Checklist


v0y4g3r (approved):
LGTM

Issue #5861, opened 2025-04-09
Casting types automatically for insertions from MySQL protocol


evenyag:
### What type of enhancement is this?

User experience, API improvement

### What does the enhancement do?

MySQL can automatically cast strings to integers in an INSERT statement.

For example, if you have a table with an integer column and you execute an INSERT statement like this:
```sql
INSERT INTO your_table (integer_column) VALUES ('1234');
```

MySQL will convert the string '1234' to the integer 1234 without throwing an error.

### Implementation challenges

1. We may only enable this feature in MySQL channel
2. Error handling: What if we can't cast a string to another type? MySQL inserts 0 in this case.
3. Do we only support strings to other types?

PR #6015, opened 2025-04-28
feat: cast strings to numerics automatically in mysql connections

evenyag:

closes https://github.com/GreptimeTeam/greptimedb/issues/5861

This PR casts strings to numerics automatically when the channel is MySQL. For a MySQL connection, it now only does the automatic conversion in insert statements.

It adds a new flag `auto_string_to_numeric` to `sql_value_to_value()` and only does this conversion when the flag is true. Now we enable the flag according to the `channel` in the `QueryContext`.

When using a prepared statement, it will cast all variables from strings to numerics, not only in insert statements. MySQL's prepared statement behaves similarly.

The inconsistency between the prepared statement and non-prepared statement should be fine in our case.

## PR Checklist


src/sql/src/statements.rs
```diff
@@ -133,6 +140,32 @@ fn parse_string_to_value(
     }
 }
 
+/// Casts string to value of specified numeric data type.
+/// If the string cannot be parsed, returns the default value for the numeric type.
+///
+/// Returns None if the data type doesn't support auto casting.
+fn auto_cast_to_numeric(s: &str, data_type: &ConcreteDataType) -> Option<Value> {
+    let value = match data_type {
+        ConcreteDataType::Boolean(_) => Value::Boolean(s.parse::<bool>().unwrap_or_default()),
```
MichaelScofield, line 149: Why not return error (or do nothing, pass through) if the casting is failed? Using default value could cause surprise to users.

v0y4g3r: > Why not return error (or do nothing, pass through) if the casting is failed? Using default value could cause surprise to users.

IMHO here we should return  `Value::Null` instead of the default value of target data type.

tests/cases/standalone/common/insert/mysql_insert.sql
```diff
@@ -0,0 +1,18 @@
...
+    val INT,
+    ts TIMESTAMP,
+    TIME INDEX(ts)
+);
+
+-- SQLNESS PROTOCOL MYSQL
+INSERT INTO integers VALUES (11, 1), (12, 2);
+
+-- SQLNESS PROTOCOL MYSQL
+INSERT INTO integers VALUES ('13', 3), ('14', 4), ('15', 5);
```
v0y4g3r, line 12: Maybe add test for incorrect numeric string:

```sql
-- SQLNESS PROTOCOL MYSQL
INSERT INTO integers VALUES ('16', 6), ('1a', 7);
```

src/sql/src/statements.rs
evenyag: Maybe MySQL changed this behavior. I can follow the latest behavior.

v0y4g3r (approved):
LGTM

MichaelScofield (approved):

PR #6026, opened 2025-04-30
fix: improve region migration error handling and optimize leader downgrade with lease check

WenyXu:

This PR improves the region migration error handling and lease management in the meta server. 

### Changes

1. **Enhanced Lease Management**
   - Added `find_datanode_lease_value` helper to check when a datanode last contacted the meta server
   - Improved leader downgrade process:
     - If datanode has been disconnected longer than region lease period, skip the sleep wait
     - If datanode disconnected recently, adjust sleep time based on remaining lease period
     - This ensures we don't wait unnecessarily when the datanode is already considered dead
   - Added proper error handling for unreachable datanodes (PusherNotFound)

2. **Better Error Handling**
   - Added new `DowngradeLeader` error type for better error tracking
   - Added deregistration of failure detectors for candidate regions
   - Enhanced logging with more context and RegionId
   - Improved error propagation for non-retryable errors

3. **Additional Logging**
   - Added logging for memtable state during region operations
   - Enhanced logging for region migration status and errors

## PR Checklist

### DIFF
--- a/src/meta-srv/src/procedure/region_migration.rs
+++ b/src/meta-srv/src/procedure/region_migration.rs
@@ -705,6 +705,8 @@ impl Procedure for RegionMigrationProcedure {
                         .inc();
                     Err(ProcedureError::retry_later(e))
                 } else {
+                    // Consumes the opening region guard before deregistering the failure detectors.
+                    self.context.volatile_ctx.opening_region_guard.take();
                     self.context
                         .deregister_failure_detectors_for_candidate_region()
                         .await;
--- a/src/meta-srv/src/procedure/region_migration.rs
+++ b/src/meta-srv/src/procedure/region_migration.rs
@@ -37,6 +37,7 @@ use common_meta::key::datanode_table::{DatanodeTableKey, DatanodeTableValue};
 use common_meta::key::table_info::TableInfoValue;
 use common_meta::key::table_route::TableRouteValue;
 use common_meta::key::{DeserializedValueWithBytes, TableMetadataManagerRef};
+use common_meta::kv_backend::ResettableKvBackendRef;
 use common_meta::lock_key::{CatalogLock, RegionLock, SchemaLock};
 use common_meta::peer::Peer;
 use common_meta::region_keeper::{MemoryRegionKeeperRef, OperatingRegionGuard};
@@ -266,6 +267,7 @@ pub trait ContextFactory {
 #[derive(Clone)]
 pub struct DefaultContextFactory {
     volatile_ctx: VolatileContext,
+    in_memory_key: ResettableKvBackendRef,
     table_metadata_manager: TableMetadataManagerRef,
     opening_region_keeper: MemoryRegionKeeperRef,
     region_failure_detector_controller: RegionFailureDetectorControllerRef,
@@ -277,6 +279,7 @@ pub struct DefaultContextFactory {
 impl DefaultContextFactory {
     /// Returns an [`DefaultContextFactory`].
     pub fn new(
+        in_memory_key: ResettableKvBackendRef,
         table_metadata_manager: TableMetadataManagerRef,
         opening_region_keeper: MemoryRegionKeeperRef,
         region_failure_detector_controller: RegionFailureDetectorControllerRef,
@@ -286,6 +289,7 @@ impl DefaultContextFactory {
     ) -> Self {
         Self {
             volatile_ctx: VolatileContext::default(),
+            in_memory_key,
             table_metadata_manager,
             opening_region_keeper,
             region_failure_detector_controller,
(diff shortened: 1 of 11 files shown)


src/meta-srv/src/region/supervisor.rs
```diff
@@ -360,10 +360,9 @@ impl RegionSupervisor {
             Ok(false) => {}
             Ok(true) => {
                 warn!(
-                    "Skipping failover since maintenance mode is enabled. Removing failure detectors for regions: {:?}",
+                    "Skipping failover since maintenance mode is enabled. Detected region failures: {:?}",
                     regions
                 );
-                self.deregister_failure_detectors(regions).await;
```
WenyXu, line 366: When the operator correctly sets maintenance mode, the current code that actively removes the failover detector causes the datanode to not trigger failover during scaling down. However, without the active removal of the failover detector logic, the failover detector would naturally stop triggering failure detection after receiving datanode heartbeats.

fengjiachun:
We should use `LeaseValue.timestamp` instead of the pusher's disconnect_time.

src/meta-srv/src/procedure/region_migration/downgrade_leader_region.rs
```diff
@@ -220,6 +236,59 @@ impl DowngradeLeaderRegion {
         }
     }
 
+    async fn update_region_lease_deadline_with_pusher_disconnect_time(&self, ctx: &mut Context) {
```
fengjiachun, line 239:
```suggestion
    async fn update_region_lease_deadline(&self, ctx: &mut Context) {
```

fengjiachun: It is no longer related to pusher disconnection.

src/meta-srv/src/procedure/region_migration/downgrade_leader_region.rs
```diff
@@ -220,6 +236,59 @@ impl DowngradeLeaderRegion {
...
+                ctx.volatile_ctx.reset_leader_region_lease_deadline();
+                info!(
+                    "Datanode {}({}) has been disconnected for longer than the region lease period ({:?}), reset leader region lease deadline to None, region: {}", 
+                    leader,
+                    last_connection_at,
+                    region_lease,
+                    ctx.persistent_ctx.region_id
+                );
+            } else if elapsed > 0 {
+                let deadline =
```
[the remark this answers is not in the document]
WenyXu: Resolved in [fa8322e](https://github.com/GreptimeTeam/greptimedb/pull/6026/commits/fa8322ebc8dc3a6adc34584f25f33c5ac835e95b)

src/meta-srv/src/procedure/region_migration/downgrade_leader_region.rs
```diff
@@ -220,6 +236,61 @@ impl DowngradeLeaderRegion {
...
     }
 
+    async fn update_leader_region_lease_deadline(&self, ctx: &mut Context) {
+        let leader = &ctx.persistent_ctx.from_peer;
+
+        let last_connection_at = match find_datanode_lease_value(leader.id, &ctx.in_memory).await {
+            Ok(lease_value) => lease_value.map(|lease_value| lease_value.timestamp_millis),
+            Err(err) => {
+                error!(err; "Failed to find datanode lease value for datanode: {}, during region migration, region: {}", leader, ctx.persistent_ctx.region_id);
+                return;
```
MichaelScofield, line 246: If lease is not found, does it mean datanode is unreachable or metasrv is in some illegal state? Either way is simply log the err enough? Maybe we should deal with the error instead, for example, abort the procedure.

src/meta-srv/src/procedure/region_migration.rs
```diff
@@ -428,6 +434,19 @@ impl Context {
             .await;
     }
 
+    /// Notifies the RegionSupervisor to deregister failure detectors for the candidate region on the destination peer.
+    ///
+    /// The candidate region may be created on the destination peer,
+    /// so we need to deregister the failure detectors for the candidate region if the procedure is aborted.
```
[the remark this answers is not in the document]
WenyXu: Once the candidate region is opened on the datanode, the region info will be included in the datanode's heartbeat.
This automatically registers it with the failover detector, so we must explicitly deregister it here.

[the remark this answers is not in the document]
WenyXu: I think the "activation" action will also introduce additional state management overhead. I believe the current approach is also reasonable — deregistering the candidate failover detector when the procedure fails feels intuitive. Additionally, when a false positive failure detection is triggered, the system should be capable of recognizing and handling it appropriately.

MichaelScofield (approved):

Issue #5907, opened 2025-04-16
Grafana label_values doesn’t work


nathanwang-comp:
### What type of bug is this?

Other

### What subsystems are affected?

Query Engine

### Minimal reproduce step

Add greptimedb as otlp backend for metrics,
When using gremtimedb as Prometheus datasource in Grafana, try to add template variables with laebel_values function, when one variable depends one the other variable, it throw errors
DataFusion error: No field named "?table?".label_2. Valid fields are "?table?".label_1, public.table.greptime_timestamp.

### What did you expect to see?

The values should be retrieve

### What did you see instead?

Errors
DataFusion error: No field named "?table?".label2. Valid fields are "?table?".label1, public.table1.greptime_timestamp.

### What operating system did you use?

Ubuntu 21.04

### What version of GreptimeDB did you use?

0.5.1

### Relevant log output and stack trace

```bash

```

nathanwang-comp:
Grafana dashboard link https://github.com/adinhodovic/argo-cd-mixin/blob/main/dashboards_out/argo-cd-application-overview.json

waynexia:
Could you please provide the result of `SHOW CREATE TABLE`?

nathanwang-comp:
CREATE TABLE IF NOT EXISTS "argocd_app_info" ( "autosync_enabled" STRING NULL, "container" STRING NULL, "dest_namespace" STRING NULL, "dest_server" STRING NULL, "endpoint" STRING NULL, "exported_namespace" STRING NULL, "greptime_timestamp" TIMESTAMP(3) NOT NULL, "greptime_value" DOUBLE NULL, "health_status" STRING NULL, "instance" STRING NULL, "job" STRING NULL, "name" STRING NULL, "namespace" STRING NULL, "operation" STRING NULL, "pod" STRING NULL, "project" STRING NULL, "repo" STRING NULL, "service" STRING NULL, "sync_status" STRING NULL, TIME INDEX ("greptime_timestamp"), PRIMARY KEY ("autosync_enabled", "container", "dest_namespace", "dest_server", "endpoint", "exported_namespace", "health_status", "instance", "job", "name", "namespace", "operation", "pod", "project", "repo", "service", "sync_status") ) ENGINE=metric WITH( on_physical_table = 'greptime_physical_table' )

@waynexia

evenyag:
The metric argocd_app_info doesn't have the `cluster` label. Is it expected?
[...]

PR #5653, opened 2025-03-05
perf(prom): optimize label values query

WenyXu:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)
#5525 

## What's changed and what's your intention?

This PR optimizes the label values query by using a specified logical plan, eliminating unnecessary overhead from the PromQL plan. This change improves performance by avoiding extra operations like sorting labels and dividing timeseries.

1h time range
PromQL 1.21s vs SQL 0.29s

24h time range
PromQL 4.32s vs SQL 0.75s
```
tql eval(1741183200,1741186800,300) up{job="node_exporter"};

|     0 |    0 |  PromInstantManipulateExec: range=[1741183200000..1741186800000], lookback=[300000], interval=[300000], time index=[greptime_timestamp] metrics=[output_rows: 13057, elapsed_compute: 208885377, num_series: 1057, ]
  PromSeriesNormalizeExec: offset=[0], time index=[greptime_timestamp], filter NaN: [false] metrics=[output_rows: 1400003, elapsed_compute: 203537424, num_series: 1057, ]
    PromSeriesDivideExec: tags=["instance", "job", "replica", "revision"] metrics=[output_rows: 1400003, elapsed_compute: 150884149, num_series: 1057, ]
      SortPreservingMergeExec: [instance@2 ASC NULLS LAST, job@3 ASC NULLS LAST, replica@4 ASC NULLS LAST, revision@5 ASC NULLS LAST] metrics=[output_rows: 1400003, elapsed_compute: 106749221, ]
        SortExec: expr=[instance@2 ASC NULLS LAST, job@3 ASC NULLS LAST, replica@4 ASC NULLS LAST, revision@5 ASC NULLS LAST], preserve_partitioning=[true] metrics=[output_rows: 1400003, elapsed_compute: 244091535, spill_count: 0, spilled_bytes: 0, spilled_rows: 0, ]
          MergeScanExec: peers=[5205500362752(1212, 0), ] metrics=[output_rows: 1400003, greptime_exec_read_cost: 0, finish_time: 873571638, first_consume_time: 564254392, ready_time: 9807817, ]
 |
[...]

PR #6040, opened 2025-05-03
fix: do not add projection to cast timestamp in label_values

evenyag:
Use cast to build time filter directly instead of adding a projection, which will cause column not found

- https://github.com/GreptimeTeam/greptimedb/pull/5653
- https://github.com/GreptimeTeam/greptimedb/issues/5907

This PR fixes an issue that `label_values()` returns `No field named` error if the following conditions are met:
- The timestamp of the metrics isn't in millisecond resolution
- There are filters in the matcher.

#5653 introduced the cast plan, but it should have other issues in earlier versions, as the timestamp resolution is invalid.

The previous implementation adds a projection node. This PR casts the timestamp previously and removes the cast and additional projection node.

## PR Checklist

### DIFF
--- a/src/query/src/promql/error.rs
+++ b/src/query/src/promql/error.rs
@@ -17,6 +17,7 @@ use std::any::Any;
 use common_error::ext::ErrorExt;
 use common_error::status_code::StatusCode;
 use common_macro::stack_trace_debug;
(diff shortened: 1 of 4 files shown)


WenyXu (approved):

killme2008 (approved):
LGTM

PR #6041, opened 2025-05-03
fix: ensures logical and physical region have the same timestamp unit

evenyag:

This PR ensures the logical and physical regions have the same timestamp unit. Otherwise, users can create a logical region without a different time unit but finally get an error when they try to insert a record.

To check the time unit, it stores the unit in the state of the physical region.

## PR Checklist

### DIFF
--- a/tests/cases/standalone/common/alter/alter_table.result
+++ b/tests/cases/standalone/common/alter/alter_table.result
@@ -155,9 +155,9 @@ ADD
 
 Affected Rows: 0
 
-ALTER TABLE 
-    t2 
-ADD 
+ALTER TABLE
+    t2
+ADD
     COLUMN at4 UINT16;
 
 Affected Rows: 0
@@ -215,7 +215,7 @@ SELECT * FROM grpc_latencies;
 | 2024-07-11T20:00:06 | host1 | GetUser     | 103.0   |
 +---------------------+-------+-------------+---------+
 
-ALTER TABLE grpc_latencies SET ttl = '10d';
+ALTER TABLE grpc_latencies SET ttl = '10000d';
 
 Affected Rows: 0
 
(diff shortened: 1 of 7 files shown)


tests/cases/standalone/common/alter/alter_table.result
```diff
@@ -215,7 +215,7 @@ SELECT * FROM grpc_latencies;
 | 2024-07-11T20:00:06 | host1 | GetUser     | 103.0   |
 +---------------------+-------+-------------+---------+
 
-ALTER TABLE grpc_latencies SET ttl = '10d';
+ALTER TABLE grpc_latencies SET ttl = '10000d';
```
MichaelScofield, line 218: Why change this test, how does it relate to this PR?

tests/cases/standalone/common/create/create_metric_table.result
```diff
@@ -48,6 +48,11 @@ CREATE TABLE t5 (ts timestamp time index, valval double, host string primary key
 
 Error: 1004(InvalidArguments), Adding field column valval to physical table
 
+-- create logical table with different time unit on time index column
+CREATE TABLE t6 (ts timestamp(6) time index, job string primary key, val double) engine = metric with ("on_physical_table" = "phy");
```
MichaelScofield, line 52: Not particularly related to this PR but why don't we create the time index column automatically for logical table? Then we don't even bother checking the time unit.

waynexia (approved):

tests/cases/standalone/common/alter/alter_table.result
evenyag: The TTL is too short, so the data expires and may be purged during sqlness test, and give an empty query result.

tests/cases/standalone/common/create/create_metric_table.result
evenyag: We usually create this table automatically. I just found this issue occasionally. It may not be a big problem.

Currently, logical tables mostly behave like a normal table, we don't expect users to create logical tables manually.

MichaelScofield (approved):

PR #6048, opened 2025-05-06
feat: flow add static user/pwd auth

discord9:

add static user provider to support username/password auth for flownode so flow can connect to frontend if frontend had set a username/password auth, notice this is not auth for flownode itself since it doesn't have port exposed to user, but only for connecting to frontend

tested locally by modify sqlness runner

need to add tls support and need an integration test later 

## PR Checklist


src/flow/src/batching_mode/frontend_client.rs
```diff
@@ -101,7 +102,8 @@ impl FrontendClient {
         )
     }
 
-    pub fn from_meta_client(meta_client: Arc<MetaClient>) -> Self {
+    pub fn from_meta_client(meta_client: Arc<MetaClient>, auth: Option<FlowAuthHeader>) -> Self {
+        common_telemetry::info!("Frontend client build with auth={:?}", auth);
```
WenyXu, line 106: The password might be logged

src/auth/src/error.rs
```diff
@@ -38,6 +38,14 @@ pub enum Error {
         location: Location,
     },
 
+    #[snafu(display("Failed to convert to utf8, error={:?}", error))]
```
shuiyisong, line 41:
```suggestion
    #[snafu(display("Failed to convert to UTF-8"))]
```

We don't need to print the error manually, the stack should print itself to the root source.

src/plugins/src/flownode.rs
```diff
@@ -0,0 +1,57 @@
...
+                    provider.type_id()
+                ),
+            }
+            .fail()?
+        };
+        let (usr, pwd) = static_provider
+            .get_one_user_pwd()
+            .context(IllegalAuthConfigSnafu)?;
+        let auth_header = FlowAuthHeader::from_user_pwd(&usr, &pwd);
+        plugins.insert(auth_header);
```
MichaelScofield, line 50: Seems you are using the plugins to pass parameters, which contradicts to its purpose. Is there another way?

src/auth/src/user_provider/static_user_provider.rs
```diff
@@ -60,6 +60,18 @@ impl StaticUserProvider {
                 .fail(),
         }
     }
+
+    /// Make sure there is only one username/password pair
+    /// And return it
+    pub fn get_one_user_pwd(&self) -> Result<(String, String)> {
+        let kv = self.users.iter().next().context(InvalidConfigSnafu {
+            value: "",
+            msg: "Expect one and only one pair of username and password",
```
shuiyisong, line 69: If there are multiple auth pairs, we can take one of them. 
I assume that the config would be the same for both frontend and flownode for convenience. In that case the frontend can have multiple auth pairs, where the flownode can take one valid pair for later usage.

src/flow/src/batching_mode/frontend_client.rs
```diff
@@ -216,7 +220,13 @@ impl FrontendClient {
             {
                 let addr = &node_info.peer.addr;
                 let client = Client::with_manager_and_urls(chnl_mgr.clone(), vec![addr.clone()]);
-                let database = Database::new(catalog, schema, client);
+                let database = {
+                    let mut db = Database::new(catalog, schema, client);
+                    if let Some(auth) = auth {
+                        db.set_auth(auth.auth().clone());
```
shuiyisong, line 226: Is this `clone` invoked on every flow request in the hot path?

discord9: > Is this `clone` invoked on every flow request in the hot path?

not really a hot path since already wait on many thing else, so clone is acceptable

src/auth/src/user_provider/static_user_provider.rs
discord9: > If there are multiple auth pairs, we can take one of them. I assume that the config would be the same for both frontend and flownode for convenience. In that case the frontend can have multiple auth pairs, where the flownode can take one valid pair for later usage.

is take one randomly acceptable? since the underline one is hash map

shuiyisong: I think it is acceptable. The goal here is to take a pair of valid ones, it doesn't seem matter to me if it's the first pair or the last. 
Also, we can add some comments to state the usage of this function (to enable invoking other components within the cluster), otherwise the function itself seems strange.

MichaelScofield (approved):

shuiyisong (approved):
LGTM

PR #6049, opened 2025-05-06
fix: alter table modify type should also modify default value

discord9:

- fix a bug when alter table modify datatype should also modify default value
- added sqlness for aforementioned bug
- actually insert a row in fuzz_alter_table test
- add a `SELECT *` after alter table in fuzz_alter_table to see if data is ok after alter, to debugging a bug in distributed mode that only occur in CI and is also very rare(can't reproduce yet), see https://github.com/GreptimeTeam/greptimedb/actions/runs/14753095448/job/41415159460 this one still haven't been reproduced

## PR Checklist

### DIFF
--- a/src/datatypes/src/schema/column_schema.rs
+++ b/src/datatypes/src/schema/column_schema.rs
@@ -61,7 +61,7 @@ pub struct ColumnSchema {
     pub data_type: ConcreteDataType,
     is_nullable: bool,
     is_time_index: bool,
-    pub default_constraint: Option<ColumnDefaultConstraint>,
+    default_constraint: Option<ColumnDefaultConstraint>,
     metadata: Metadata,
 }
 
--- a/src/datatypes/src/schema/column_schema.rs
+++ b/src/datatypes/src/schema/column_schema.rs
@@ -61,7 +61,7 @@ pub struct ColumnSchema {
     pub data_type: ConcreteDataType,
     is_nullable: bool,
     is_time_index: bool,
-    default_constraint: Option<ColumnDefaultConstraint>,
+    pub default_constraint: Option<ColumnDefaultConstraint>,
     metadata: Metadata,
 }
 
--- a/src/store-api/src/metadata.rs
+++ b/src/store-api/src/metadata.rs
@@ -701,16 +701,25 @@ impl RegionMetadataBuilder {
                 // also cast default value to target_type if default value exist
                 let new_default =
                     if let Some(default_value) = column_meta.column_schema.default_constraint() {
-                        Some(default_value.cast_to_datatype(&target_type).context(
-                            CastDefaultValueSnafu {
-                                default_value: default_value.clone(),
-                                target_type,
-                            },
-                        )?)
+                        Some(
+                            default_value
+                                .cast_to_datatype(&target_type)
+                                .with_context(|_| CastDefaultValueSnafu {
+                                    default_value: Some(default_value.clone()),
+                                    target_type: target_type.clone(),
+                                })?,
+                        )
                     } else {
                         None
                     };
-                column_meta.column_schema.default_constraint = new_default;
+                column_meta.column_schema = column_meta
+                    .column_schema
+                    .clone()
+                    .with_default_constraint(new_default)
+                    .with_context(|_| CastDefaultValueSnafu {
+                        default_value: column_meta.column_schema.default_constraint().cloned(),
+                        target_type,
+                    })?;
             }
         }
 
@@ -987,7 +996,7 @@ pub enum MetadataError {
 
     #[snafu(display("Failed to cast default value {:?} to {}", default_value, target_type))]
     CastDefaultValue {
-        default_value: ColumnDefaultConstraint,
+        default_value: Option<ColumnDefaultConstraint>,
         target_type: ConcreteDataType,
         source: datatypes::Error,
         #[snafu(implicit)]
--- a/src/store-api/src/metadata.rs
+++ b/src/store-api/src/metadata.rs
@@ -705,9 +705,10 @@ impl RegionMetadataBuilder {
                             default_value
                                 .cast_to_datatype(&target_type)
                                 .with_context(|_| CastDefaultValueSnafu {
-                                    default_value: Some(default_value.clone()),
-                                    target_type: target_type.clone(),
-                                    reason: "Failed to cast default value",
+                                    reason: format!(
+                                        "Failed to cast default value from {:?} to type {:?}",
+                                        default_value, target_type
+                                    ),
                                 })?,
                         )
                     } else {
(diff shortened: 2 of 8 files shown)


src/datatypes/src/schema/column_schema.rs
```diff
@@ -61,7 +61,7 @@ pub struct ColumnSchema {
     pub data_type: ConcreteDataType,
     is_nullable: bool,
     is_time_index: bool,
-    default_constraint: Option<ColumnDefaultConstraint>,
+    pub default_constraint: Option<ColumnDefaultConstraint>,
```
evenyag, line 64: Don't publish this field and use `with_default_constraint()`. It can do some validation.

src/store-api/src/metadata.rs
```diff
@@ -694,9 +697,24 @@ impl RegionMetadataBuilder {
...
+                        Some(default_value.cast_to_datatype(&target_type).context(
+                            CastDefaultValueSnafu {
+                                default_value: default_value.clone(),
+                                target_type,
+                            },
+                        )?)
+                    } else {
+                        None
+                    };
+                column_meta.column_schema.default_constraint = new_default;
```
evenyag, line 713: Should we cast the default value or use NULL as the new default value? How do other databases behave?

src/store-api/src/metadata.rs
```diff
@@ -694,9 +697,24 @@ impl RegionMetadataBuilder {
 
         for column_meta in self.column_metadatas.iter_mut() {
             if let Some(target_type) = change_type_map.remove(&column_meta.column_schema.name) {
-                column_meta.column_schema.data_type = target_type;
+                column_meta.column_schema.data_type = target_type.clone();
+                // also cast default value to target_type if default value exist
+                let new_default =
+                    if let Some(default_value) = column_meta.column_schema.default_constraint() {
+                        Some(default_value.cast_to_datatype(&target_type).context(
```
MichaelScofield, line 704:
```suggestion
                        Some(default_value.cast_to_datatype(&target_type).with_context(|_|
```

src/store-api/src/metadata.rs
discord9: > Should we cast the default value or use NULL as the new default value? How do other databases behave?

I don't know which one is better:
1. postgre: https://www.postgresql.org/docs/current/ddl-alter.html#DDL-ALTER-COLUMN-TYPE
- "PostgreSQL will attempt to convert the column's default value (if any) to the new type, as well as any constraints that involve the column. But these conversions might fail, or might produce surprising results. It's often best to drop any constraints on the column before altering its type, and then add back suitably modified constraints afterwards."
2. MySQL: https://dev.mysql.com/doc/refman/8.4/en/alter-table.html you either use `ALTER TABLE <table-name> MODIFY <column-name> <New Full column definition>` for change the whole column definition or `SET DEFAULT`/`DROP DEFAULT` to change only the default value

src/store-api/src/metadata.rs
```diff
@@ -694,9 +697,33 @@ impl RegionMetadataBuilder {
...
+                        None
+                    };
+                column_meta.column_schema = column_meta
+                    .column_schema
+                    .clone()
+                    .with_default_constraint(new_default)
+                    .with_context(|_| CastDefaultValueSnafu {
+                        default_value: column_meta.column_schema.default_constraint().cloned(),
+                        target_type,
+                    })?;
```
evenyag, line 722: nitpick: Maybe we can change the CastDefaultValue to keep a message. So you can provide a different message to distinguish errors from `cast_to_datatype` and `with_default_constraint`.
```
CastDefaultValueSnafu {
    reason: format!("error message"),
}
```

tests/cases/standalone/common/alter/alter_table.result
```diff
@@ -78,14 +124,40 @@ DESC TABLE test_alt_table;
...
+| m      | Boolean              |     | YES  |         | FIELD         |
 | dt     | TimestampMicrosecond |     | YES  |         | FIELD         |
 +--------+----------------------+-----+------+---------+---------------+
 
 DROP TABLE test_alt_table;
 
 Affected Rows: 0
 
+-- test if column with default value can change type properly
+CREATE TABLE test_alt_table_default(h INTEGER, i INTEGER DEFAULT 0, j TIMESTAMP TIME INDEX, PRIMARY KEY (h));
```
evenyag, line 136: non-blocking: Can we add a case that the default value is an integer value, and we alter the column to string type?

evenyag (approved):
Others LGTM

src/store-api/src/metadata.rs
waynexia: I prefer the pg way. But we may also need to support `SET DEFAULT`/`DROP DEFALUT` clauses (or any other equivalence)

src/store-api/src/metadata.rs
```diff
@@ -967,6 +996,21 @@ pub enum MetadataError {
...
+        reason
+    ))]
+    CastDefaultValue {
+        default_value: Option<ColumnDefaultConstraint>,
+        target_type: ConcreteDataType,
+        reason: String,
+        source: datatypes::Error,
+        #[snafu(implicit)]
+        location: Location,
+    },
```
evenyag, line 1012: We already have a reason, we can format the context in the reason string.
```suggestion
    #[snafu(display("{}", reason))]
    CastDefaultValue {
        reason: String,
        source: datatypes::Error,
        #[snafu(implicit)]
        location: Location,
    },
```

src/store-api/src/metadata.rs
```diff
@@ -694,9 +697,35 @@ impl RegionMetadataBuilder {
...
+                let new_default =
+                    if let Some(default_value) = column_meta.column_schema.default_constraint() {
+                        Some(
+                            default_value
+                                .cast_to_datatype(&target_type)
+                                .with_context(|_| CastDefaultValueSnafu {
+                                    default_value: Some(default_value.clone()),
+                                    target_type: target_type.clone(),
+                                    reason: "Failed to cast default value",
+                                })?,
```
evenyag, line 711:
```suggestion
                                .with_context(|_| CastDefaultValueSnafu {
                                    reason: format!("Failed to cast default value {:?} to {:?}", default_value, target_type),
                                })?,
```

src/store-api/src/metadata.rs
```diff
@@ -694,9 +697,35 @@ impl RegionMetadataBuilder {
...
+                        None
+                    };
+                column_meta.column_schema = column_meta
+                    .column_schema
+                    .clone()
+                    .with_default_constraint(new_default.clone())
+                    .with_context(|_| CastDefaultValueSnafu {
+                        default_value: column_meta.column_schema.default_constraint().cloned(),
+                        target_type,
+                        reason: format!("Failed to set new default: {:?}", new_default),
```
evenyag, line 723:
```suggestion
                    .with_context(|_| CastDefaultValueSnafu {
                        reason: format!("Failed to set new default: {:?}", new_default),
```

waynexia (approved):
LGTM

PR #6050, opened 2025-05-06
refactor: respect `data_home` as root data home directory

zyy17:

respect `data_home` as root data home directory:

1. When `logs_dir` is not specified, use `${data_home}/${logs_dir}` as default logs dir;
2. Rename `METASRV_HOME` to `METASRV_DATA_DIR` and change its default value to `metasrv/`. The actual `metasrv_home` will be `${data_home}/${METASRV_DATA_DIR}`;
3. Normalize some constants about `*DIR`;

## PR Checklist

### DIFF
--- a/config/config.md
+++ b/config/config.md
@@ -100,7 +100,7 @@
 | `query` | -- | -- | The query engine options. |
 | `query.parallelism` | Integer | `0` | Parallelism of the query engine.<br/>Default to 0, which means the number of CPU cores. |
 | `storage` | -- | -- | The data storage options. |
-| `storage.data_home` | String | `./greptimedb_data/` | The working home directory. |
+| `storage.data_home` | String | `./greptimedb_data` | The working home directory. |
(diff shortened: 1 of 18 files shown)


config/config.md
```diff
@@ -314,7 +314,7 @@
 
 | Key | Type | Default | Descriptions |
 | --- | -----| ------- | ----------- |
-| `data_home` | String | `./greptimedb_data/metasrv/` | The working home directory. |
+| `data_home` | String | `./greptimedb_data` | The working home directory. |
```
MichaelScofield, line 317: Is this a breaking change if users are using the default configuration?

zyy17: It's not a breaking change.

In the original implementation, it uses the following code to create a data home for metasrv:

```rust
let metasrv_home = options.data_home.to_string();
```

This PR change to(`METASRV_DATA_DIR` is `metasrv`):

```rust
        let metasrv_home = Path::new(&options.data_home)
            .join(METASRV_DATA_DIR)
            .to_string_lossy()
            .to_string();

```

The final config is the same.

zyy17: After some thought, it's not breaking change for default configs. However, it could be a breaking change for the scenarios that specify `data_home`.

evenyag: What does this directory store?

killme2008: Let's break it @zyy17

killme2008: > What does this directory store?

Appears to contain only log files and temporary data like telemetry UUIDs.

killme2008:
@MichaelScofield @zyy17 Let's move forward!

PR #6054, opened 2025-05-06
feat: impl bulk memtable and bridge bulk inserts

v0y4g3r:

- Do not merge until https://github.com/GreptimeTeam/greptime-proto/pull/237 is merged

This PR bridges bulk insert write path with region worker by adding a separate `bulk_req_buffer` to `RegionWorkerLoop`:
https://github.com/GreptimeTeam/greptimedb/blob/be0c482a373cc9f7ee803dc4348065fa73a44ce9/src/mito2/src/worker.rs#L728

This change also involves handling stalled requests because we now have stalled bulk requests:

https://github.com/GreptimeTeam/greptimedb/blob/be0c482a373cc9f7ee803dc4348065fa73a44ce9/src/mito2/src/worker.rs#L598

To further accelerates bulk write performance, this PR introduces `SimpleBulkMemtable` when tables do not have primary keys:
https://github.com/GreptimeTeam/greptimedb/blob/bdc3d3a11bc53fd6c86f5f2c0690d27a2897a06a/src/mito2/src/memtable/time_series.rs#L92-L93

`SimpleBulkMemtable` supports append an arrow `RecordBatch` by extending arrays in `Series`:

https://github.com/GreptimeTeam/greptimedb/blob/bdc3d3a11bc53fd6c86f5f2c0690d27a2897a06a/src/mito2/src/memtable/time_series.rs#L843-L921

## PR Checklist


evenyag:
What a huge PR 🥵 

Could you provide some description?

v0y4g3r:
> What a huge PR 🥵
> 
> Could you provide some description?

I added some details of this PR

src/operator/src/bulk_insert.rs
```diff
@@ -50,16 +53,20 @@ impl Inserter {
         let FlightMessage::Recordbatch(rb) = message else {
             return Ok(0);
         };
+        let record_batch = rb.df_record_batch();
         metrics::BULK_REQUEST_MESSAGE_SIZE.observe(body_size as f64);
         decode_timer.observe_duration();
+        metrics::BULK_REQUEST_MESSAGE_SIZE.observe(body_size as f64);
```
fengjiachun, line 59:
```suggestion
        decode_timer.observe_duration();
        metrics::BULK_REQUEST_MESSAGE_SIZE.observe(body_size as f64);
```

src/operator/src/bulk_insert.rs
```diff
@@ -83,84 +90,88 @@ impl Inserter {
         let mut mask_per_datanode = HashMap::with_capacity(region_masks.len());
         for (region_number, mask) in region_masks {
             let region_id = RegionId::new(table_id, region_number);
+            metrics::BULK_REQUEST_ROWS
+                .with_label_values(&["rows_per_region"])
+                .observe(mask.true_count() as f64);
```
fengjiachun, line 95: Is it(compute true_count) an expensive operation?

src/mito2/src/region_write_ctx.rs
```diff
@@ -92,6 +94,10 @@ pub(crate) struct RegionWriteCtx {
     ///
     /// The i-th notify is for i-th mutation.
     notifiers: Vec<WriteNotify>,
+    /// Notifiers for bulk requests.
+    bulk_notifiers: Vec<WriteNotify>,
+    /// Pending bulk write requests
+    pub(crate) bulk_parts: Vec<BulkPart>,
```
evenyag, line 100: Maybe we should replace WriteRequest with an enum so we can support bulk/normal in the same struct in the future.

Now we have to copy-paste similar codes: WriteNotify, pending requests, stalled requests......

src/mito2/src/region_write_ctx.rs
```diff
@@ -243,4 +251,53 @@ impl RegionWriteCtx {
...
             .set_sequence_and_entry_id(self.next_sequence - 1, self.next_entry_id - 1);
     }
+
+    pub(crate) fn push_bulk(&mut self, sender: OptionOutputTx, mut bulk: BulkPart) {
+        self.bulk_notifiers
+            .push(WriteNotify::new(sender, bulk.num_rows));
+        bulk.sequence = self.next_sequence;
+        self.next_sequence += bulk.num_rows as u64;
+        self.bulk_parts.push(bulk);
+    }
```
evenyag, line 261: So we won't write bulk parts to WAL now, right?

src/mito2/src/worker.rs
```diff
@@ -593,35 +594,55 @@ pub(crate) struct StalledRequests {
     ///
     /// Key: RegionId
     /// Value: (estimated size, stalled requests)
-    pub(crate) requests: HashMap<RegionId, (usize, Vec<SenderWriteRequest>)>,
+    pub(crate) requests:
+        HashMap<RegionId, (usize, Vec<SenderWriteRequest>, Vec<SenderBulkRequest>)>,
```
evenyag, line 598: We should also update the `stalled_count()` method. Well, maybe we can reduce some potential issues if we can merge the two requests into one.

src/mito2/src/worker/handle_write.rs
```diff
@@ -294,8 +311,89 @@ impl<S> RegionWorkerLoop<S> {
...
         }
+    }
 
-        region_ctxs
+    /// Processes bulk insert requests.
+    fn process_bulk_requests(
+        &mut self,
+        region_ctxs: &mut HashMap<RegionId, RegionWriteCtx>,
+        requests: &mut Vec<SenderBulkRequest>,
+    ) {
```
evenyag, line 321: Mostly the same as `process_write_requests()`? I think we need to refactor this to avoid duplicate code.

src/operator/src/metrics.rs
```diff
@@ -93,4 +93,11 @@ lazy_static! {
         ]
     )
     .unwrap();
+    pub static ref BULK_REQUEST_ROWS: HistogramVec = register_histogram_vec!(
+        "greptime_table_operator_bulk_insert_message_rows",
+        "table operator bulk inserts message rows",
+        &["type"],
+        linear_buckets(0.0, 10000.0, 10).unwrap()
+    )
+    .unwrap();
```
evenyag, line 102: The buckets num is too large. better to use exponential buckets.

src/store-api/src/region_request.rs
```diff
@@ -334,165 +329,30 @@ fn make_region_truncate(truncate: TruncateRequest) -> Result<Vec<(RegionId, Regi
...
-
-    let decode_timer = metrics::CONVERT_REGION_BULK_REQUEST
+    let decoder_timer = metrics::CONVERT_REGION_BULK_REQUEST
         .with_label_values(&["decode"])
         .start_timer();
     let schema_data = FlightData::decode(request.schema.clone()).context(ProstSnafu)?;
     let payload_data = FlightData::decode(request.payload.clone()).context(ProstSnafu)?;
     let mut decoder = FlightDecoder::default();
-    let _schema_message = decoder.try_decode(schema_data).context(FlightCodecSnafu)?;
+    let _ = decoder.try_decode(schema_data).context(FlightCodecSnafu)?;
```
evenyag, line 343: So why do we need the schema_data?

src/operator/src/bulk_insert.rs
evenyag: Maybe we can observe this after we filter the batch. Then we can use the num_rows of the batch.

fengjiachun: Can we also include `region_id` in the labels?

src/store-api/src/region_request.rs
v0y4g3r: > So why do we need the schema_data?

`FlightDecode` needs this function call to fill the `FlightDecode::schema` field:
https://github.com/GreptimeTeam/greptimedb/blob/aab7451aa23d3e141fea0ee3a8b9478e82f15cdb/src/common/grpc/src/flight.rs#L164

Otherwise it will raise an error here: https://github.com/GreptimeTeam/greptimedb/blob/aab7451aa23d3e141fea0ee3a8b9478e82f15cdb/src/common/grpc/src/flight.rs#L169

src/operator/src/bulk_insert.rs
v0y4g3r: > Can we also include `region_id` in the labels?

That will bring too many buckets

src/mito2/src/memtable/bulk/part.rs
```diff
@@ -32,34 +34,105 @@ use datatypes::arrow::datatypes::SchemaRef;
...
+                        if let Some(v) = &vectors[col_idx] {
+                            value_to_grpc_value(v.get(row_idx))
+                        } else {
+                            api::v1::Value { value_data: None }
+                        }
+                    })
+                    .collect::<Vec<_>>();
+                api::v1::Row { values }
+            })
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
v0y4g3r: nevermind, this is just the fallback path for `TimeSeriesMemtable`

src/mito2/src/memtable/simple_bulk_memtable.rs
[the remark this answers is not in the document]
v0y4g3r: > Looks like this memtable is almost the same as the time series memtable. I didn't review all the details.

Mostlt the same, with only special consideration for `write_bulk`.

src/mito2/src/memtable/time_series.rs
```diff
@@ -727,21 +758,21 @@ impl Series {
 struct ValueBuilder {
     timestamp: Vec<i64>,
     timestamp_type: ConcreteDataType,
-    sequence: Vec<u64>,
-    op_type: Vec<u8>,
-    fields: Vec<Option<Box<dyn MutableVector>>>,
+    sequence: UInt64VectorBuilder,
+    op_type: UInt8VectorBuilder,
```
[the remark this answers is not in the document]
v0y4g3r: > We changed this from vector builder to `Vec` as `Vec` is slightly faster. I think it's fine to keep using `Vec`?

I didn't remember the details of this change, maybe change back to `Vec` can simplify the code.

src/mito2/src/memtable/time_series.rs
```diff
@@ -780,20 +811,27 @@ impl ValueBuilder {
...
-                        .create_mutable_vector(num_rows.max(INITIAL_BUILDER_CAPACITY));
+                    let mut mutable_vector =
+                        if let ConcreteDataType::String(_) = &self.field_types[idx] {
+                            FieldBuilder::String(StringBuilder::with_capacity(256, 4096))
+                        } else {
+                            FieldBuilder::Other(
+                                self.field_types[idx]
+                                    .create_mutable_vector(num_rows.max(INITIAL_BUILDER_CAPACITY)),
+                            )
+                        };
```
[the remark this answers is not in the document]
v0y4g3r: > non-blocking: Can we implement this as a constructor of the `FieldBuilder`?

There's a `FieldBuilder::create` constructor. But here we need to specify different initial capacity for string builder and other builders, because `StringBuilder` have `item_capacity` and `data_capacity`.

src/mito2/src/region_write_ctx.rs
v0y4g3r: > So we won't write bulk parts to WAL now, right?

Yes

src/operator/src/metrics.rs
v0y4g3r: fixed in 065cf78f1cea9ae9f502084c4c498ddadb3d3a94

src/mito2/src/worker.rs
v0y4g3r: fixed in 065cf78f1cea9ae9f502084c4c498ddadb3d3a94

src/mito2/src/memtable/time_series.rs
```diff
@@ -802,6 +840,94 @@ impl ValueBuilder {
...
+                FieldBuilder::String(builder) => {
+                    let string_vector = field_src.as_any().downcast_ref::<StringVector>().ok_or(
+                        error::InvalidBatchSnafu {
+                            reason: format!(
+                                "Field type mismatch, expecting String, given: {}",
+                                field_src.data_type()
+                            ),
+                        }
+                        .build(),
+                    )?;
```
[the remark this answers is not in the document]
v0y4g3r: fixed in 7f5aab6a7b3ac5608009b0f62c798e24ac66df29

src/mito2/src/metrics.rs
```diff
@@ -392,6 +392,15 @@ lazy_static! {
...
         exponential_buckets(0.01, 10.0, 6).unwrap(),
     ).unwrap();
+
+
+    pub static ref REGION_WORKER_HANDLE_WRITE_ELAPSED: HistogramVec = register_histogram_vec!(
+        "greptime_region_worker_handle_write",
+        "elapsed time for handling writes in region worker loop",
+        &["stage"],
+        vec![0.001, 0.002, 0.005, 0.01, 0.02, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.5, 1.0, 1.5, 2.0, 2.5, 5.0, 10.0, 15.0]
+    ).unwrap();
```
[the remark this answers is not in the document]
v0y4g3r: > I wonder if we defined too many buckets. Maybe we can use exponential buckets. e.g. 0.001 ~ 10.

fixed in 7e3555e186760c36c686d0c5f0849a6715616394

src/mito2/src/memtable/time_partition.rs
```diff
@@ -55,6 +57,11 @@ impl TimePartition {
     fn write(&self, kvs: &KeyValues) -> Result<()> {
         self.memtable.write(kvs)
     }
+
+    /// Writes a record batch to memtable.
+    fn write_record_batch(&self, rb: BulkPart) -> error::Result<()> {
+        self.memtable.write_bulk(rb)
+    }
```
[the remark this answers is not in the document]
v0y4g3r: > So we don't support splitting the record batch by time window right now? We should fix this later.

In fe09a5da485b0c9296ccbe3651c706789c9960cd I implemented basic partition handling that works as expected when the bulk part falls into only one time partition. 

If the part intersect with multiple time partitions, it only write the bulk part to the first time paritition. So this violates the contraction of time partition semantic and need a fixme in future pull requests.

fengjiachun (approved):
LGTM

src/operator/src/metrics.rs
```diff
@@ -93,4 +93,12 @@ lazy_static! {
         ]
     )
     .unwrap();
+    pub static ref BULK_REQUEST_ROWS: HistogramVec = register_histogram_vec!(
+        "greptime_table_operator_bulk_insert_message_rows",
+        "table operator bulk inserts message rows",
+        &["type"],
+        // 10 ~ 1_000_000
+        exponential_buckets(10.0, 10.0, 5).unwrap()
```
evenyag, line 101: The comment is incorrect. It only creates the buckets in `[10 ~ 10000]`.

evenyag (approved):

PR #6062, opened 2025-05-08
fix(query): ensure proper handling of escaped regex in PromQL

WenyXu:

In VictoriaMetrics, escaped characters in regex matchers (such as instance=~"172\\.16\\.222\\.77:4000") are correctly processed and the query works as expected. Special characters like dots in IP addresses need to be escaped with backslashes to be interpreted as literal characters in regex patterns.

This change fixes the regression introduced in PR #5920 and ensures that PromQL queries with escaped special characters (particularly in IP addresses) work consistently with VictoriaMetrics behavior, improving compatibility for users migrating from these systems to GreptimeDB.

## PR Checklist


waynexia (changes requested):
#5920 makes us compatible with Prometheus. This is not a regression.

WenyXu:
> #5920 makes us compatible with Prometheus. This is not a regression.

Screenshot 2025-05-08 at 14 11 48
Screenshot 2025-05-08 at 14 11 26

Prometheus also supports query conditions with escaped characters.

killme2008:
If VM and Prometheus support both behaviors, we should as well.

killme2008:
The #6094 should fix this.

PR #6063, opened 2025-05-08
feat(flow): support prom ql(in tql) in flow

discord9:

support running prom ql in flow, for now it's very basic, and lacking some checks, i.e.:
```sql
CREATE FLOW calc_reqs SINK TO cnt_reqs AS
TQL EVAL (0, 15, '5s') count_values("status_code", http_requests);
```
it might be worth to only allow `now()` related expr in `TQL EVAL(start, end, step)` 's `start` and `end` expr

~TODO: more sqlness tests~ done
~TODO: fix a bug when `now()` is used in `TQL EVAL` the `now()` is instantised to a fixed time(it should be remain as a function)~ fixed, this is fix by passing the raw query string without change it

## PR Checklist

### DIFF
--- a/src/flow/src/batching_mode/task.rs
+++ b/src/flow/src/batching_mode/task.rs
@@ -592,8 +592,6 @@ impl BatchingTask {
             return Ok(None);
         };
 
-        // TODO(discord9): add auto column or not? This might break compatibility for auto created sink table before this, but that's ok right?
-
         let mut add_filter = AddFilterRewriter::new(expr);
         let mut add_auto_column = AddAutoColumnRewriter::new(sink_table_schema.clone());
 
--- a/src/flow/src/batching_mode/task.rs
+++ b/src/flow/src/batching_mode/task.rs
@@ -37,6 +37,8 @@ use query::query_engine::DefaultSerializer;
 use query::QueryEngineRef;
 use session::context::QueryContextRef;
 use snafu::{ensure, OptionExt, ResultExt};
+use sql::parser::{ParseOptions, ParserContext};
+use sql::statements::statement::Statement;
 use substrait::{DFLogicalSubstraitConvertor, SubstraitPlan};
 use tokio::sync::oneshot;
 use tokio::sync::oneshot::error::TryRecvError;
@@ -76,6 +78,34 @@ pub struct TaskConfig {
     sink_table_name: [String; 3],
     pub source_table_names: HashSet<[String; 3]>,
     catalog_manager: CatalogManagerRef,
+    query_type: QueryType,
+}
+
+fn determine_query_type(query: &str, query_ctx: &QueryContextRef) -> Result<QueryType, Error> {
+    let stmts =
+        ParserContext::create_with_dialect(query, query_ctx.sql_dialect(), ParseOptions::default())
+            .map_err(BoxedError::new)
+            .context(ExternalSnafu)?;
+
+    ensure!(
+        stmts.len() == 1,
+        InvalidQuerySnafu {
+            reason: format!("Expect only one statement, found {}", stmts.len())
+        }
+    );
+    let stmt = &stmts[0];
+    match stmt {
+        Statement::Tql(_) => Ok(QueryType::Tql),
+        _ => Ok(QueryType::Sql),
+    }
+}
+
+#[derive(Debug, Clone)]
+enum QueryType {
+    /// query is a tql query
+    Tql,
+    /// query is a sql query
+    Sql,
 }
 
 #[derive(Clone)]
@@ -86,7 +116,7 @@ pub struct BatchingTask {
 
 impl BatchingTask {
     #[allow(clippy::too_many_arguments)]
-    pub fn new(
+    pub fn try_new(
         flow_id: FlowId,
         query: &str,
         plan: LogicalPlan,
@@ -97,8 +127,8 @@ impl BatchingTask {
         query_ctx: QueryContextRef,
         catalog_manager: CatalogManagerRef,
         shutdown_rx: oneshot::Receiver<()>,
-    ) -> Self {
-        Self {
+    ) -> Result<Self, Error> {
+        Ok(Self {
             config: Arc::new(TaskConfig {
                 flow_id,
                 query: query.to_string(),
@@ -108,9 +138,10 @@ impl BatchingTask {
                 source_table_names: source_table_names.into_iter().collect(),
                 catalog_manager,
                 output_schema: plan.schema().clone(),
+                query_type: determine_query_type(query, &query_ctx)?,
             }),
             state: Arc::new(RwLock::new(TaskState::new(query_ctx, shutdown_rx))),
-        }
+        })
     }
 
     /// mark time window range (now - expire_after, now) as dirty (or (0, now) if expire_after not set)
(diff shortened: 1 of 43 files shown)


waynexia (approved):
I give a quick peek and rest LGTM

tests/cases/standalone/common/flow/show_create_flow.result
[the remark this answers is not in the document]
discord9: > why are they changed?

Now create flow will store the original raw query as it is(the part after "AS"), before the sql string will get convert to Statement then `to_string()` back to string, this is to get around that `TQL EVAL(now(),now(),'15s')` will become fixed numbers problem

src/sql/src/util.rs
```diff
@@ -58,14 +59,36 @@ pub fn parse_option_string(option: SqlOption) -> Result<(String, String)> {
...
+            // since tql have sliding time window, so we don't need to extract tables from it
+            // (because we are going to eval it fully anyway)
+        }
+    }
 
     names.into_iter()
 }
 
+/// translate the start location to the index in the sql string
+pub fn location_to_index(sql: &str, location: &sqlparser::tokenizer::Location) -> usize {
```
fengjiachun, line 77: Please add a unit test for this.

fengjiachun (approved):
Others LGTM

src/sql/src/parsers/create_parser.rs
```diff
@@ -324,7 +325,22 @@ impl<'a> ParserContext<'a> {
             .expect_keyword(Keyword::AS)
             .context(SyntaxSnafu)?;
 
-        let query = self.parser.parse_query().context(error::SyntaxSnafu)?;
+        let start_loc = self.parser.peek_token().span.start;
```
killme2008, line 328: I don't get  the refactor of the parser. Could you explain in details?

tests/cases/standalone/common/flow/flow_advance_ttl.result
```diff
@@ -82,6 +82,7 @@ ADMIN FLUSH_FLOW('test_distinct_basic');
 |  FLOW_FLUSHED  |
 +-----------------------------------------+
 
+-- SQLNESS SLEEP 1s
```
killme2008, line 85: Why add so many `sleep`? The test case will be slower.

tests/cases/standalone/common/flow/flow_tql.result
```diff
@@ -0,0 +1,83 @@
...
+|          |   TIME INDEX ("ts"),                      |
+|          |   PRIMARY KEY ("status_code")             |
+|          | )                                         |
+|          |                                           |
+|          | ENGINE=mito                               |
+|          |                                           |
++----------+-------------------------------------------+
+
+INSERT INTO TABLE http_requests VALUES
+    (now() - '17s'::interval, 'host1', 'idc1', 200),
```
killme2008, line 35: I think we can use fixed time to assert the results.

tests/cases/standalone/common/flow/flow_basic.result
```diff
@@ -45,6 +45,7 @@ ADMIN FLUSH_FLOW('test_numbers_basic');
 |  FLOW_FLUSHED  |
 +----------------------------------------+
 
+-- SQLNESS SLEEP 1s
```
evenyag, line 48: So many sleep 😱

killme2008 (commented):
The PR is missing a design and implementation overview. I did my best to review the code, but honestly, it's still quite challenging without some context. It would be helpful if you could add a brief introduction explaining the design and main changes in this PR. Thank you!

tests/cases/standalone/common/flow/flow_basic.result
discord9: > So many sleep 😱

tmp fix, will create a separate PR to remove this

src/flow/src/batching_mode/task.rs
```diff
@@ -494,104 +526,103 @@ impl BatchingTask {
...
-            let Some(expr) = expr else {
-                // no new data, hence no need to update
-                debug!("Flow id={:?}, no new data, not update", self.config.flow_id);
-                return Ok(None);
-            };
-
-            // TODO(discord9): add auto column or not? This might break compatibility for auto created sink table before this, but that's ok right?
-
-            let mut add_filter = AddFilterRewriter::new(expr);
+            // TODO(discord9): not add auto column for tql query?
```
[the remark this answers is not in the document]
discord9: > I don't understand why not.

it's just that most prom ql query comes with a useful time index already? but I guess a `update_at` column is still useful, will remove this TODO

src/flow/src/batching_mode/task.rs
```diff
@@ -494,104 +526,103 @@ impl BatchingTask {
...
+                .map(|s| s.to_string())
+        );
+
+        let Some(expr) = expr else {
+            // no new data, hence no need to update
+            debug!("Flow id={:?}, no new data, not update", self.config.flow_id);
+            return Ok(None);
+        };
+
+        // TODO(discord9): add auto column or not? This might break compatibility for auto created sink table before this, but that's ok right?
```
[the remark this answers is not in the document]
discord9: > I don't understand what's meaning of this TODO, could you explain it in details?

This one is fixed now, forgot to remove it, now batching mode also create same auto created column like streaming mode

src/sql/src/statements/create.rs
```diff
@@ -741,7 +778,7 @@ WITH(
                     r#"
 CREATE FLOW filter_numbers
 SINK TO out_num_cnt
-AS SELECT number FROM numbers_input WHERE number > 10"#,
+AS SELECT number FROM numbers_input where number > 10"#,
```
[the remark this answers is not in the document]
discord9: > Why the `where` becomes lower case in this PR?

previous behavior is wrong, previously the raw query string get parsed into ast, then unparse(`to_string`) back to string, hence many subtle difference are created in the process, this PR make sure the raw query string get stored intact(mainly due to prom ql will parse `now()` function into a fixed timestamp, but also this is more consistent)

src/sql/src/parsers/create_parser.rs
discord9: > I don't get the refactor of the parser. Could you explain in details?

this is for fixing the problem that prom ql parser will parse `now()` to a fixed timestamp, so instead just slice the entire raw query string away and save it

tests/cases/standalone/common/flow/flow_tql.result
discord9: > I think we can use fixed time to assert the results.

need to test if `now()` is working

tests/cases/standalone/common/flow/flow_advance_ttl.result
discord9: > Why add so many `sleep`? The test case will be slower.

was investigating why the sqlness test with posgresql sometimes failed as if wasn't wait for long enough, now it seems it is due to not waiting for long enough after flush flow, but not have a good enough solution now, maybe wait a while after called `flush_flow` in batching engine?

tests/cases/standalone/common/flow/flow_tql.result
killme2008: They are both needed.

killme2008:
@discord9 Let's move forward, thanks

PR #6066, opened 2025-05-08
fix!: disable append mode in trace services table

zyy17:

Duplicated `service_name` in default services table `opentelemetry_traces_services` since it turn on append only mode.

## PR Checklist

### DIFF
--- a/src/operator/src/insert.rs
+++ b/src/operator/src/insert.rs
@@ -580,6 +580,10 @@ impl Inserter {
                 // for it's a very unexpected behavior and should be set by user explicitly
                 for mut create_table in create_tables {
                     if create_table.table_name == trace_services_table_name(trace_table_name) {
+                        // Disable append mode for trace services table since it requires upsert behavior.
+                        create_table
+                            .table_options
+                            .insert(APPEND_MODE_KEY.to_string(), "false".to_string());
                         let table = self
                             .create_physical_table(create_table, None, ctx, statement_executor)
                             .await?;
--- a/tests-integration/tests/http.rs
+++ b/tests-integration/tests/http.rs
@@ -2838,7 +2838,7 @@ pub async fn test_otlp_traces_v1(store_type: StorageType) {
     )
     .await;
 
-    let expected_ddl = r#"[["mytable_services","CREATE TABLE IF NOT EXISTS \"mytable_services\" (\n  \"timestamp\" TIMESTAMP(9) NOT NULL,\n  \"service_name\" STRING NULL,\n  TIME INDEX (\"timestamp\")\n)\n\nENGINE=mito\nWITH(\n  append_mode = 'true'\n)"]]"#;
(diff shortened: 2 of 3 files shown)


zyy17:
> Any background for this change? This can affect ingestion performance.

When I tested opentelmetry, I found out it has duplicated `services_name` in the `opentelemetry_traces_services` table, which should be a unique value. If we use an append-only table, we still need to use `SELECT DISTINCT(service_name)` to get the unique service names. 

We have aggregated the `service_name` in each hotel request, resulting in only one insert operation per service_name. However, I'm uncertain about the magnitude of the performance impact.

tests-integration/tests/http.rs
```diff
@@ -2838,7 +2838,7 @@ pub async fn test_otlp_traces_v1(store_type: StorageType) {
     )
     .await;
 
-    let expected_ddl = r#"[["mytable_services","CREATE TABLE IF NOT EXISTS \"mytable_services\" (\n  \"timestamp\" TIMESTAMP(9) NOT NULL,\n  \"service_name\" STRING NULL,\n  TIME INDEX (\"timestamp\")\n)\n\nENGINE=mito\nWITH(\n  append_mode = 'true'\n)"]]"#;
+    let expected_ddl = r#"[["mytable_services","CREATE TABLE IF NOT EXISTS \"mytable_services\" (\n  \"timestamp\" TIMESTAMP(9) NOT NULL,\n  \"service_name\" STRING NULL,\n  TIME INDEX (\"timestamp\")\n)\n\nENGINE=mito\nWITH(\n  append_mode = 'false'\n)"]]"#;
```
evenyag, line 2841: I think non-append-only table is necessary. But if we need to duplicate service_name, we must add it to the primary key.

src/operator/src/insert.rs
```diff
@@ -580,6 +580,10 @@ impl Inserter {
                 // for it's a very unexpected behavior and should be set by user explicitly
                 for mut create_table in create_tables {
                     if create_table.table_name == trace_services_table_name(trace_table_name) {
+                        // Disable append mode for trace services table since it requires upsert behavior.
+                        create_table
+                            .table_options
+                            .insert(APPEND_MODE_KEY.to_string(), "false".to_string());
```
evenyag, line 586: Will it be a breaking change? Does it matter if users already have the service name table.

evenyag:
> > Any background for this change? This can affect ingestion performance.
> 
> When I tested opentelmetry, I found out it has duplicated `services_name` in the `opentelemetry_traces_services` table, which should be a unique value. If we use an append-only table, we still need to use `SELECT DISTINCT(service_name)` to get the unique service names.
> 
> We have aggregated the `service_name` in each hotel request, resulting in only one insert operation per service_name. However, I'm uncertain about the magnitude of the performance impact. For most of the scenarios, the services table will be very small(maybe the cost of deduplication isn't that significant).
> 
> @evenyag @waynexia What do you think?

The performance should be acceptable since the flow table does a similar thing in batching mode. Maybe we can further reduce the request number by caching service name in the frontend if this becomes the bottleneck.

sunng87:
I see. It's only for service table. It should be fine.

sunng87 (approved):

sunng87:
@evenyag I think eventually we should make this service table a type of internal index. It's like prometheus' label values. We can store it by time window.

src/operator/src/insert.rs
zyy17: Yes. It's a breaking change. Let me add the `!`.

Issue #6070, opened 2025-05-09
Getting errors when running a CLI command locally to export databases to S3 within a cluster.


0neSe7en:
### What type of enhancement is this?

User experience

### What does the enhancement do?

We already support export data to s3. https://github.com/GreptimeTeam/greptimedb/pull/5585

But if I run the greptime command locally to backup to S3 which is running within a cluster, it will get error

```
export job failed err=0: OpenDAL operator failed, at src/cli/src/export.rs:602:44
1: Unexpected (temporary) at Writer::close => send http request

Context:
   url: http://minio.svc.cluster.local/greptime/backup/greptime/public/copy_from.sql
```

the command I ran. (I port foward the greptime from cluster to local)

```
greptime cli export --addr 127.0.0.1:4000 --s3 --s3-bucket greptime/backup --s3-endpoint http://minio.svc.cluster.local/ --s3-access-key <> --s3-secret-key <> --s3-region <>
```

I think we can use the `--output-dir` for SQL files output, instead of export all the SQLs to S3 as well, so that we can support `greptime cli export --target all --output-dir ./... --s3`
the schemas( `copy-tables.sql`, `copy-database.sql`, `copy_from.sql` ) are generated in `--output-dir`, and the parquet files are exported to S3.

additionally, we can add `--s3-root` as we already have in storage options. From the command I ran, I use `greptime/backup` for `--s3-bucket`, it would be better to use `--s3-bucket greptime --s3-root backup` or maybe just `--s3-url s3://greptime/backup` which is the same as `copy database to <URL>`

fengjiachun:
> I think we can use the --output-dir for SQL files output, instead of export all the SQLs to S3 as well, so that we can support greptime cli export --target all --output-dir ./... --s3
the schemas( copy-tables.sql, copy-database.sql, copy_from.sql ) are generated in --output-dir, and the parquet files are exported to S3.

I believe running the CLI command twice will resolve the issue:

```
cli export --target schema --output-dir ..
cli export --target data --s3..
```

> additionally, we can add --s3-root as we already have in storage options. From the command I ran, I use greptime/backup for --s3-bucket, it would be better to use --s3-bucket greptime --s3-root backup or maybe just --s3-url s3://greptime/backup which is the same as copy database to <URL>

Sounds good

0neSe7en:
>I believe running the CLI command twice will resolve the issue:

[...]

PR #6091, opened 2025-05-13
feat: export to s3 add more options

discord9:

#6070

-- add `--s3-ddl-local-dir` to export sql to local even when `s3` is set and is exporting parquet data file to s3, this is useful for when export client can't access s3 directly(because the *.sql files is created by export client directly)
-- add `--s3-root` to allow state root path separately 
tested locally with and without `--s3-root` option

## PR Checklist

### DIFF
--- a/src/cli/src/export.rs
+++ b/src/cli/src/export.rs
@@ -376,7 +376,7 @@ impl Export {
         let timer = Instant::now();
         let db_names = self.get_db_names().await?;
         let db_count = db_names.len();
-        let operator = self.build_operator().await?;
+        let operator = self.build_prefer_fs_operator().await?;
 
         for schema in db_names {
             let create_database = self
@@ -406,7 +406,7 @@ impl Export {
         let semaphore = Arc::new(Semaphore::new(self.parallelism));
         let db_names = self.get_db_names().await?;
         let db_count = db_names.len();
-        let operator = Arc::new(self.build_operator().await?);
+        let operator = Arc::new(self.build_prefer_fs_operator().await?);
         let mut tasks = Vec::with_capacity(db_names.len());
 
         for schema in db_names {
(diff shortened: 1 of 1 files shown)


src/cli/src/export.rs
```diff
@@ -115,6 +120,11 @@ pub struct ExportCommand {
     #[clap(long)]
     s3_bucket: Option<String>,
 
+    // The s3 root path
+    /// if s3 is set, this is required
+    #[clap(long)]
+    s3_root: Option<String>,
```
killme2008, line 126: I am not sure why we only support s3. Is Are there any reasons that user can't export data into azure blob storage or gcs?

src/cli/src/export.rs
```diff
@@ -56,6 +56,11 @@ pub struct ExportCommand {
 
     /// Directory to put the exported data. E.g.: /tmp/greptimedb-export
     /// for local export.
+    ///
+    /// if both `output_dir` and `s3` are set, `output_dir` will be only used for
+    /// exported SQL files, and the data will be exported to s3.
+    ///
+    /// if `s3` is set but `output_dir` is not set, both SQL&data will be exported to s3.
```
killme2008, line 63: If i understand correctly, the proper name must be `ddl_dir`?

src/cli/src/export.rs
```diff
@@ -56,6 +56,11 @@ pub struct ExportCommand {
 
     /// Directory to put the exported data. E.g.: /tmp/greptimedb-export
     /// for local export.
+    ///
+    /// if both `output_dir` and `s3` are set, `output_dir` will be only used for
```
MichaelScofield, line 60: Non-orthogonal arguments are bad. How about separate the sql export destination and the data export destination to two different arguments?

src/cli/src/export.rs
discord9: > If i understand correctly, the proper name must be `ddl_dir`?

added a `s3_ddl_local_dir` option, this verbose name is to prevent misunderstanding(i.e. misunderstand as s3 path or else)

src/cli/src/export.rs
yihong0618: because of this
https://github.com/GreptimeTeam/greptimedb/pull/5585#discussion_r1986041379

we need to support it first

yihong0618: cc @fengjiachun

fengjiachun: Once issue #6105 is resolved, we can upgrade the CLI tool to support GCS and AzBlob.

fengjiachun (approved):
LGTM

killme2008 (approved):
LGTM

We need to refactor these options after we support other storages.

PR #6092, opened 2025-05-13
fix: append noop entry when auto topic creation is disabled

### LINKED ISSUE #3045: Ensure the Kafka record offset is must greater than 0
> What if the topic creation is success but the append is failed here? Or What if the append returns failure because of some network errors but it is actually succeed in kafka site? Will the append be executed the next time and thus two noop records are appended? There might be other situations that could make inconsistency, I think what we ought to do here is to make the "create and append" a transaction. But can we?

Yes, we need to leave an issue for this. We must ensure the topic offset is greater than 0.

_Originally posted by @WenyXu in https://github.com/GreptimeTeam/greptimedb/pull/3040#discussion_r1437986394_

### LINKED ISSUE #5649: Truncating stale records after create metadata for kafka topics

User experience

A topic may exist before it is created via metasrv. In such cases, stale records from previous operations may persist, potentially causing inconsistencies or unexpected behaviors. To ensure data integrity and prevent conflicts, we propose truncating stale records immediately after the topic metadata is created.

### Implementation challenges

_No response_

WenyXu:

close #3045 

This PR fixes a bug in the topic management system where noop entries were not being appended to topics when auto topic creation was disabled.

1. **Bug Fix**
   - Fixed the issue where topics were not properly prepared (no noop entry appended) when `auto_create_topics` was disabled
   - Removed early return when `auto_create_topics` is false, ensuring topics are always prepared
   - Separated topic creation and preparation phases to ensure proper topic initialization

2. **Topic Management**
   - Changed topic persistence strategy to only store newly created topics
   - Added better logging for topic creation and preparation status

3. **Configuration Updates**
   - Disabled auto topic creation in Kafka by default (`KAFKA_CFG_AUTO_CREATE_TOPICS_ENABLE: false`)
   - Maintained `auto_create_topics` configuration for controlling topic creation behavior

## PR Checklist

### DIFF
--- a/src/common/meta/src/wal_options_allocator/topic_pool.rs
+++ b/src/common/meta/src/wal_options_allocator/topic_pool.rs
@@ -78,31 +78,31 @@ impl KafkaTopicPool {
     }
 
     /// Tries to activate the topic manager when metasrv becomes the leader.
+    ///
     /// First tries to restore persisted topics from the kv backend.
-    /// If not enough topics retrieved, it will try to contact the Kafka cluster and request creating more topics.
+    /// If there are unprepared topics (topics that exist in the configuration but not in the kv backend),
+    /// it will create these topics in Kafka if `auto_create_topics` is enabled.
+    ///
+    /// Then it prepares all unprepared topics by appending a noop record if the topic is empty,
+    /// and persists them in the kv backend for future use.
     pub async fn activate(&self) -> Result<()> {
         let num_topics = self.topics.len();
         ensure!(num_topics > 0, InvalidNumTopicsSnafu { num_topics });
 
-        let topics_to_be_created = self
-            .topic_manager
-            .get_topics_to_create(&self.topics)
-            .await?;
+        let unprepared_topics = self.topic_manager.unprepare_topics(&self.topics).await?;
 
-        if !topics_to_be_created.is_empty() {
+        if !unprepared_topics.is_empty() {
             if self.auto_create_topics {
-                info!("Creating {} topics", topics_to_be_created.len());
-                self.topic_creator
-                    .create_topics(&topics_to_be_created)
-                    .await?;
+                info!("Creating {} topics", unprepared_topics.len());
+                self.topic_creator.create_topics(&unprepared_topics).await?;
             } else {
                 info!("Auto create topics is disabled, skipping topic creation.");
             }
             self.topic_creator
-                .prepare_topics(&topics_to_be_created)
+                .prepare_topics(&unprepared_topics)
                 .await?;
             self.topic_manager
-                .persist_topics(&topics_to_be_created)
+                .persist_prepared_topics(&unprepared_topics)
                 .await?;
         }
         info!("Activated topic pool with {} topics", self.topics.len());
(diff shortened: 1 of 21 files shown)


src/common/wal/src/config/kafka/common.rs
```diff
@@ -25,9 +25,9 @@ use snafu::{OptionExt, ResultExt};
 /// The default backoff config for kafka client.
 pub const DEFAULT_BACKOFF_CONFIG: BackoffConfig = BackoffConfig {
     init_backoff: Duration::from_millis(100),
-    max_backoff: Duration::from_secs(10),
-    base: 2.0,
-    deadline: Some(Duration::from_secs(120)),
+    max_backoff: Duration::from_secs(1),
+    base: 3.0,
+    deadline: Some(Duration::from_secs(3)),
```
fengjiachun, line 30: Please add comments for these configurations.

WenyXu: resolved in [4f58a40](https://github.com/GreptimeTeam/greptimedb/pull/6092/commits/4f58a40c78c4e8194176e271b02e422f67e26d8d)

src/common/meta/src/wal_options_allocator/topic_pool.rs
```diff
@@ -77,27 +78,35 @@ impl KafkaTopicPool {
...
+        if !unprepared_topics.is_empty() {
+            if self.auto_create_topics {
+                info!("Creating {} topics", unprepared_topics.len());
+                self.topic_creator.create_topics(&unprepared_topics).await?;
+            } else {
+                info!("Auto create topics is disabled, skipping topic creation.");
+            }
             self.topic_creator
-                .prepare_topics(&topics_to_be_created)
+                .prepare_topics(&unprepared_topics)
```
MichaelScofield, line 102: `prepare_topics` should receive the actually created topics, otherwise the already created topics that are skipped above could be appended more than one noop records.

src/common/meta/src/wal_options_allocator/topic_pool.rs
```diff
@@ -77,27 +78,35 @@ impl KafkaTopicPool {
...
+                .await?;
+            self.topic_manager
+                .persist_prepared_topics(&unprepared_topics)
```
MichaelScofield, line 105: Seems the `persist_prepared_topics` simply stores all the topics. The `unprepared_topics` can be different, will multiple prepare operations do persist here replace each other?

src/common/wal/src/config/kafka/common.rs
MichaelScofield: Are max_backoff and deadline meaning the same?

src/common/meta/src/wal_options_allocator/topic_pool.rs
WenyXu: > prepare_topics should receive the actually created topics

Yes, it will 

> could be appended more than one noop records.

It only appends a noop record if the topic is empty.

src/common/meta/src/wal_options_allocator/topic_pool.rs
WenyXu: > will multiple prepare operations

No, Only metasrv leader will do this operation.

src/common/wal/src/config/kafka/common.rs
WenyXu: The `max_backoff` represents the max wait time between retry attempts, while deadline represents the total time limit for all retry operations.

MichaelScofield (approved):

fengjiachun (approved):

Issue #6078, opened 2025-05-09
Support using SST as plain parquet without encoding primary keys


evenyag:
### What type of enhancement is this?

Tech debt reduction, Refactor, Performance

### What does the enhancement do?

Currently, the internal struct and format of the mito engine are highly coupled with the encoded primary key (a time series).

This improves performance when the cardinality of the primary key is low. However, the performance has substantial degradation when the unique value of the primary key increases.

This design also has the following disadvantage under the append mode:
- We always read all primary key columns
- Unable to use parquet's min-max statistics

### Implementation challenges

We plan to support a new SST format that stores the primary key without encoding. We called this the plain format.

We can support this for append mode first. We have to implement the following parts:
- [ ] PlainFormat to help reading/writing the new format
- [ ] A new batch struct
- [ ] new Projection mapper
- [ ] Writing the new format
- [ ] Reading the new format
- [ ] Merge and compaction
- [ ] Use the new format for apppend only table

For tables with merge mode, we may need further support features like deduplication and deletion. This may require an internal column to distinguish time series. We plan to support it later and may create a new issue for it.

evenyag:
> My first thought about this is to decouple append-only and plain format. Is there any connection between them, except they are mostly in the same scenario?

They are unrelated, but I want to gradually adopt the new format. Using plain format for append-only table is due to the hardness of implementation and performance reasons. Finally, we may switch to plain parquet and only keep the old format for backward compatibility.

Supporting merge mode for the plain format requires more work. We may also have to address some performance issues. Using the plain format under append-only mode is helpful to test the performance and is easier to implement.

However, if we found that the old format really could not be replaced, we may need to choose the proper format according to the table workload.

evenyag:
We can also apply the dedup/merge algorithm for plain parquets to parquets with primary keys. It would be useful to dedup parquets that contain duplications.

We may extend https://github.com/GreptimeTeam/greptimedb/issues/6505 to support parts with different formats. So let's wait for that.

evenyag:
follow up in https://github.com/GreptimeTeam/greptimedb/issues/7732

PR #6097, opened 2025-05-14
feat: Adds format helper for the plain format

evenyag:

This PR adds helpers for the plain format:
- PlainWriteFormat
  - Processes record batches to write
- PlainReadFormat
  - Gets statistics
  - Adds internal columns to read
  - SST level projection handling

The `PlainWriteFormat` is quite simple as it only overrides sequences on demand.

The `PlainReadFormat` is similar to `PrimaryKeyReadFormat`. Since there is no primary key encoding, the read format can get the min/max value for each column.

It renames the old helpers to `PrimaryKeyWriteFormat` and `PrimaryKeyReadFormat`. The `ReadFormat` now is an enum that wraps the two different formats.

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet/format.rs
+++ b/src/mito2/src/sst/parquet/format.rs
@@ -61,19 +61,19 @@ pub(crate) type PrimaryKeyArray = DictionaryArray<UInt32Type>;
 /// Contains: time index and internal columns.
 const FIXED_POS_COLUMN_NUM: usize = 4;
 
-/// Helper for writing the SST format.
-pub(crate) struct WriteFormat {
+/// Helper for writing the SST format with primary key.
+pub(crate) struct PrimaryKeyWriteFormat {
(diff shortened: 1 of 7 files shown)


src/mito2/src/sst/parquet/format.rs
```diff
@@ -132,8 +133,190 @@ impl WriteFormat {
...
+    }
+
+    /// Returns min/max values of specific columns.
+    /// Returns None if the column does not have statistics.
+    /// The column should not be encoded as a part of a primary key.
+    pub(crate) fn column_values(
+        row_groups: &[impl Borrow<RowGroupMetaData>],
+        column: &ColumnMetadata,
+        column_index: usize,
+        is_min: bool,
```
evenyag, line 244: Just move the implementation of `column_values()` and `column_null_counts()` to this place.

PR #6100, opened 2025-05-14
fix(flow): flow task run interval

discord9:

- fix flow task run interval by setting it to half of the time window duration，so that flow task will refresh not that frequent
- fix a bug when flow query went wrong can't shutdown flow exec loop
- correct some log msg

## PR Checklist

### DIFF
--- a/src/flow/src/batching_mode/time_window.rs
+++ b/src/flow/src/batching_mode/time_window.rs
@@ -55,6 +55,9 @@ use crate::error::{
 use crate::expr::error::DataTypeSnafu;
 use crate::Error;
 
+/// Represents a test timestamp in seconds since the Unix epoch.
+const DEFAULT_TEST_TIMESTAMP: Timestamp = Timestamp::new_second(17_0000_0000);
+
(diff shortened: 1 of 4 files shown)


src/flow/src/batching_mode/time_window.rs
```diff
@@ -84,19 +88,40 @@ impl std::fmt::Display for TimeWindowExpr {
...
+                UnexpectedSnafu {
+                    reason: format!(
+                        "Expect upper bound older than lower bound, found upper={u:?} and lower={l:?}"
+                    ),
+                }
+                .build()
+            })?,
+            _ => None,
+        };
+        zelf.eval_time_window_size = time_window_size;
```
evenyag, line 123: If we already know the time window, does it mean that we can use the time window to `eval` each timestamp later?

discord9: > If we already know the time window, does it mean that we can use the time window to `eval` each timestamp later?

Could be, might optimize later in separate PR

v0y4g3r (approved):
LGTM

zhongzc (approved):

PR #6109, opened 2025-05-15
feat: don't hide atomic write dir

waynexia:

Change the atomic write dir from `.tmp` to `tmp`. Sometimes, a hidden dir may cause confusion. This is not a breaking change since this directory is "temporary"

## PR Checklist

### DIFF
--- a/src/mito2/src/access_layer.rs
+++ b/src/mito2/src/access_layer.rs
@@ -42,6 +42,8 @@ pub type AccessLayerRef = Arc<AccessLayer>;
 /// SST write results.
 pub type SstInfoArray = SmallVec<[SstInfo; 2]>;
 
+pub const ATOMIC_WRITE_DIR: &'static str = "tmp/";
+
 /// A layer to access SST files under the same directory.
 pub struct AccessLayer {
     region_dir: String,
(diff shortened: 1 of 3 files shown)


src/mito2/src/access_layer.rs
```diff
@@ -42,6 +42,10 @@ pub type AccessLayerRef = Arc<AccessLayer>;
 /// SST write results.
 pub type SstInfoArray = SmallVec<[SstInfo; 2]>;
 
+pub const ATOMIC_WRITE_DIR: &str = "tmp/";
+/// For compatibility. Remove this after a major version release.
+pub const OLD_ATOMIC_WRITE_DIR: &str = "tmp/";
```
evenyag, line 47: I guess the old dir should be `.tmp/`? Now the two paths are the same path.
```suggestion
pub const ATOMIC_WRITE_DIR: &str = "tmp/";
/// For compatibility. Remove this after a major version release.
pub const OLD_ATOMIC_WRITE_DIR: &str = ".tmp/";
```

evenyag (approved):

discord9 (approved):

PR #6111, opened 2025-05-15
chore: increment sqlparser version

happysalada:
This is done in order to facilitate nix builds

This is a proof of concept PR to illustrate a potential solution to facilitate nix builds.
I don't think it's mergeable as is, but it illustrates what the full solution would entail.

sqlparser custom dependency has been patch incremented to solve having the same dependency in the same version with 2 different urls.

In order for this to work fully, the same thing needs to be done for sqlparser_derive too.

## PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2439,7 +2439,7 @@ dependencies = [
  "futures-util",
  "serde",
  "snafu 0.8.5",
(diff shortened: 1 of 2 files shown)


MichaelScofield:
`sqlparser` is "[overrided](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html)" by our fork now (see the `[patch.crates-io]` section), this way we keep all the crates (including those in dependencies) use the same `sqlparser`. I think this PR can be closed.

PR #6115, opened 2025-05-16
chore: Add more data format support to the pipeline dryrun api.

paomian:

* Add more data format support to the pipeline dryrun api.
* Add the `data_type` field to specify the format of the data content.
* The `data` field must be of type string

**This pr is a breaking change to the dry_run API.**

## PR Checklist

### DIFF
--- a/src/pipeline/src/error.rs
+++ b/src/pipeline/src/error.rs
@@ -561,8 +561,8 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
-    #[snafu(display("Prepare value must be an object"))]
-    PrepareValueMustBeObject {
+    #[snafu(display("Input value must be an object"))]
+    InputValueMustBeObject {
         #[snafu(implicit)]
         location: Location,
     },
@@ -833,7 +833,7 @@ impl ErrorExt for Error {
             | ValueYamlKeyMustBeString { .. }
             | YamlLoad { .. }
             | YamlParse { .. }
-            | PrepareValueMustBeObject { .. }
+            | InputValueMustBeObject { .. }
             | ColumnOptions { .. }
             | UnsupportedIndexType { .. }
             | UnsupportedNumberType { .. }
--- a/src/servers/src/http/event.rs
+++ b/src/servers/src/http/event.rs
@@ -680,15 +680,15 @@ impl TryFrom<&ContentType> for EventPayloadResolverInner {
 }
 
 #[derive(Debug)]
-struct EventPayloadResolver {
+struct EventPayloadResolver<'a> {
     inner: EventPayloadResolverInner,
     /// The content type of the payload.
     /// keep it for logging original content type
     #[allow(dead_code)]
-    content_type: ContentType,
+    content_type: &'a ContentType,
 }
 
-impl EventPayloadResolver {
+impl EventPayloadResolver<'_> {
     pub(super) fn support_content_type_list() -> Vec<String> {
         EventPayloadResolverInner::iter()
             .map(|x| x.to_string())
@@ -696,19 +696,19 @@ impl EventPayloadResolver {
     }
 }
 
-impl TryFrom<&ContentType> for EventPayloadResolver {
+impl<'a> TryFrom<&'a ContentType> for EventPayloadResolver<'a> {
     type Error = Error;
 
-    fn try_from(content_type: &ContentType) -> Result<Self> {
+    fn try_from(content_type: &'a ContentType) -> Result<Self> {
         let inner = EventPayloadResolverInner::try_from(content_type)?;
         Ok(EventPayloadResolver {
             inner,
-            content_type: content_type.clone(),
+            content_type,
         })
     }
 }
 
-impl EventPayloadResolver {
+impl EventPayloadResolver<'_> {
     fn parse_payload(&self, payload: Bytes, ignore_errors: bool) -> Result<Vec<PipelineMap>> {
         match self.inner {
             EventPayloadResolverInner::Json => {
--- a/src/servers/src/http/event.rs
+++ b/src/servers/src/http/event.rs
@@ -475,6 +475,9 @@ fn add_step_info_for_pipeline_dryrun_error(step_msg: &str, e: Error) -> Response
     (status_code_to_http_status(&e.status_code()), body).into_response()
 }
 
+/// Parse the data with given content type
+/// If the content type is invalid, return error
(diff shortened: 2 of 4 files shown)


src/pipeline/src/error.rs
```diff
@@ -561,8 +561,8 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
-    #[snafu(display("Prepare value must be an object"))]
-    PrepareValueMustBeObject {
+    #[snafu(display("Input value must be an object"))]
```
sunng87, line 564: Is there a chance that the original input is not an object but to be converted to object internally? The user may not be aware of the concept of "object".

src/servers/src/http/event.rs
```diff
@@ -474,6 +475,21 @@ fn add_step_info_for_pipeline_dryrun_error(step_msg: &str, e: Error) -> Response
...
 }
 
+/// Parse the data with given content type
+/// If the content type is invalid, return error
+/// content type is one of application/json, text/plain, application/x-ndjson
+fn parse_dryrun_data(data_type: String, data: String) -> Result<Vec<PipelineMap>> {
+    let content_type = ContentType::from_str(&data_type);
+    if content_type.is_err() {
+        return InvalidParameterSnafu {
+            reason: format!("invalid content type: {}, expected: one of: application/json, text/plain, application/x-ndjson", data_type),
```
sunng87, line 485: This error message is liked to be out of date when we add new content-type support. 

And avoid two `:`s

src/servers/src/http/event.rs
```diff
@@ -474,6 +475,21 @@ fn add_step_info_for_pipeline_dryrun_error(step_msg: &str, e: Error) -> Response
...
+        }
+        .fail();
+    }
+    // Safety : the content type is valid
+    extract_pipeline_value_by_content_type(content_type.unwrap(), Bytes::from(data), false)
```
sunng87, line 490: refactor this to 

```
if let Ok(content_type) = content_type {
   ...
} else {
   ...
}
```

src/pipeline/src/error.rs
paomian: > Is there a chance that the original input is not an object but to be converted to object internally?

This could be quite complex, and there may be some tricky edge cases to handle. For example, `1 2 3 4` would be parsed by JSON parse as a sequence of `Value::Number()`. There are also issues like deeply nested arrays such as `[[[[{"message":"i am object"}]]]]`.

sunng87: The user may not aware of its input will be converted to obejct if the content type is plain/text or even ndjson. Can we guarantee that we won't hit this error when input is text based?

sunng87 (approved):

src/pipeline/src/error.rs
paomian: If the user selects the correct content type and uses text/plain, they will not encounter this error. My concern lies in a design mistake made at the beginning. We did not differentiate between JSON and NDJSON, and used https://docs.rs/serde_json/latest/serde_json/de/struct.StreamDeserializer.html. This tool can simultaneously support JSON and NDJSON. However, serde_json will parse content separated by spaces or line breaks. For example, if the payload is 1 2 3 4, it will be parsed into a value sequence containing 1, 2, 3, and 4.

If a user provides text payload data but selects JSON as the type, parsing errors may occur. The text will be recognized as multiple values rather than a single string, making it difficult to infer the user's intent. Such errors are not easily recognizable by users and can lead to confusion. Perhaps we should strengthen the restrictions. If the request header is JSON, we should only allow the payload to be an object array type. This way, we can thoroughly distinguish between NDJSON and JSON types, enabling us to verify if the user's data and type match.

For the sake of efficiency, I prefer not to directly parse the input format to determine the user's input format.

src/servers/src/http/event.rs
```diff
@@ -616,63 +639,139 @@ pub async fn log_ingester(
...
+            InvalidParameterSnafu {
+                reason: format!(
+                    "invalid content type: {}, expected: one of {}",
+                    content_type,
+                    EventPayloadResolver::support_content_type_list().join(", ")
+                ),
+            }
+            .fail()
+        }
+    }
```
shuiyisong, line 679:
```suggestion
    fn try_from(content_type: &ContentType) -> Result<Self> {
        match content_type {
            x if *x == *JSON_CONTENT_TYPE => Ok(EventPayloadResolverInner::Json),
            x if *x == *NDJSON_CONTENT_TYPE => Ok(EventPayloadResolverInner::Ndjson),
            x if *x == *TEXT_CONTENT_TYPE || *x == *TEXT_UTF8_CONTENT_TYPE => {
                Ok(EventPayloadResolverInner::Text)
            }
            _ => InvalidParameterSnafu {
                reason: format!(
                    "invalid content type: {}, expected: one of {}",
                    content_type,
                    EventPayloadResolver::support_content_type_list().join(", ")
                ),
            }
            .fail(),
        }
    }
```

src/servers/src/http/event.rs
```diff
@@ -616,63 +639,139 @@ pub async fn log_ingester(
...
+impl TryFrom<&ContentType> for EventPayloadResolver {
+    type Error = Error;
+
+    fn try_from(content_type: &ContentType) -> Result<Self> {
+        let inner = EventPayloadResolverInner::try_from(content_type)?;
+        Ok(EventPayloadResolver {
+            inner,
+            content_type: content_type.clone(),
+        })
+    }
```
shuiyisong, line 708: We can pass in the `content_type` here to avoid `clone`

shuiyisong (approved):
Rest LGTM

src/pipeline/src/error.rs
sunng87: > If the request header is JSON, we should only allow the payload to be an object array type. 

Agreed. We should enforce this restriction for json input.

src/servers/src/http/event.rs
```diff
@@ -616,62 +639,153 @@ pub async fn log_ingester(
...
-                let mut map = PipelineMap::new();
-                map.insert("message".to_string(), pipeline::Value::String(line));
-                map
-            })
-            .collect::<Vec<_>>(),
+    }
+}
 
-        _ => UnsupportedContentTypeSnafu { content_type }.fail()?,
+fn extract_pipeline_value_by_content_type(
```
paomian, line 768: @shuiyisong

Issue #6114, opened 2025-05-16
Fails to execute SQL statements with comments on the first line


killme2008:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Query Engine, Other

### Minimal reproduce step

GreptimeDB fails to correctly execute SQL statements with comments on the first line, causing the MySQL client to hang.

### What did you expect to see?

Should ignore the comment.

### What did you see instead?

Hang

### What operating system did you use?

Not relevant

### What version of GreptimeDB did you use?

Not relevant

### Relevant log output and stack trace

```bash

```

QuakeWang:
@killme2008 Hi, I want to try this issue : )

killme2008:
Looks like it's not an issue; it doesn't hang, just awaits valid input. The real problem is that kill queries are currently unsupported (https://github.com/GreptimeTeam/greptimedb/issues/5899).

PR #6118, opened 2025-05-16
fix: comments on the first line hang

QuakeWang:

fix: #6114 

GreptimeDB fails to correctly execute SQL statements with comments on the first line, causing the MySQL client to hang.

image

After fix this error, test passed.

image

## PR Checklist


killme2008:
Looks like it's not an issue; it doesn't hang, just awaits valid input. The real problem is that `kill` queries are currently unsupported (#5072).

killme2008:
@QuakeWang Thanks for your contribution. It's not a bug, so i closed this PR right now.
Looking foward to your future contribution!

PR #6126, opened 2025-05-19
feat: accommodate default column name with pre-created table schema

waynexia:

For protocols with default column names like Prometheus Remote Write, the default name must match the destination table's schema, otherwise it will trigger an auto schema change and leading to unexpected behavior, like having two time index columns.

This patch implements an "accommodate" mode that can overwrite the schema in write request with corresponding column name from table schema that already exists.

## PR Checklist

### DIFF
--- a/src/operator/src/insert.rs
+++ b/src/operator/src/insert.rs
@@ -164,6 +164,7 @@ impl Inserter {
             ctx,
             statement_executor,
             AutoCreateTableType::Physical,
+            false,
         )
         .await
     }
@@ -180,6 +181,7 @@ impl Inserter {
             ctx,
             statement_executor,
             AutoCreateTableType::Log,
+            false,
         )
         .await
     }
@@ -195,6 +197,7 @@ impl Inserter {
             ctx,
             statement_executor,
             AutoCreateTableType::Trace,
+            false,
         )
         .await
     }
@@ -205,12 +208,14 @@ impl Inserter {
         requests: RowInsertRequests,
         ctx: QueryContextRef,
         statement_executor: &StatementExecutor,
+        accommodate_existing_schema: bool,
     ) -> Result<Output> {
         self.handle_row_inserts_with_create_type(
             requests,
             ctx,
             statement_executor,
             AutoCreateTableType::LastNonNull,
+            accommodate_existing_schema,
         )
         .await
     }
@@ -222,6 +227,7 @@ impl Inserter {
         ctx: QueryContextRef,
         statement_executor: &StatementExecutor,
         create_type: AutoCreateTableType,
+        accommodate_existing_schema: bool,
     ) -> Result<Output> {
         // remove empty requests
         requests.inserts.retain(|req| {
@@ -241,7 +247,7 @@ impl Inserter {
                 &ctx,
                 create_type,
                 statement_executor,
-                false,
+                accommodate_existing_schema,
             )
             .await?;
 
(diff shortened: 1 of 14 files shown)


src/operator/src/insert.rs
```diff
@@ -448,12 +455,18 @@ impl Inserter {
     ///
     /// Returns a mapping from table name to table id, where table name is the table name involved in the requests.
     /// This mapping is used in the conversion of RowToRegion.
+    ///
+    /// `accommodate_existing_schema` is used to determine if the existing schema should override the new schema.
+    /// It only works for TIME_INDEX and VALUE columns. This is for the case where the user creates a table with
+    /// custom schema, and then inserts data with endpoints that have default schema setting, like prometheus
```
killme2008, line 461: We don't need to process the field columns since only the time index is unique; the fields are not.

If users want to add new fields, just let them—they’ll notice any issues themselves. This approach also avoids our special validation logic, which could impact performance.

src/operator/src/insert.rs
```diff
@@ -791,9 +807,10 @@ impl Inserter {
     /// It always adds columns if not exist.
```
killme2008, line 807: The function comment needs to be updated.

src/operator/src/insert.rs
```diff
@@ -281,10 +287,11 @@ impl Inserter {
             table_infos,
         } = self
             .create_or_alter_tables_on_demand(
-                &requests,
+                &mut requests,
                 &ctx,
                 AutoCreateTableType::Logical(physical_table.to_string()),
                 statement_executor,
+                true,
```
killme2008, line 294: Not quite sure, but the influxdb line protocol will go into this function too?

src/operator/src/insert.rs
waynexia: It could be a surprise if users create a table with its own field name, but all data goes into `greptime_value`. In such case they'll need to modify all their queries to specify which field to query

src/operator/src/insert.rs
waynexia: No, only PromStoreProtocolHandler will use this

killme2008: But the influxdb needs this fix too.

src/operator/src/insert.rs
killme2008: Got it. But the single field detection feels inefficient and might hurt performance.

The user will likely understand the behavior through testing or evaluation, which may be acceptable. The time index can't be alter and is unique.

waynexia: >The time index can't be alter and is unique.

This also holds for the field column in some scenarios like above.

>But the single field detection feels inefficient and might hurt performance.

I tried to reduce the overhead. This won't bring extra logic if the request doesn't trigger a schema change, or the table has default column names

src/operator/src/insert.rs
waynexia: added

src/operator/src/insert.rs
```diff
@@ -802,10 +828,60 @@ impl Inserter {
...
+            }
+
+            // Remove from add_columns any column that is timestamp or field (if there is only one field column)
+            add_columns.add_columns.retain(|col| {
+                let def = col.column_def.as_ref().unwrap();
+                def.semantic_type != SemanticType::Timestamp as i32
+                    && (def.semantic_type != SemanticType::Field as i32 && field_col_name.is_some())
+            });
+        }
+
```
killme2008, line 884: Return `Ok(None)` if `add_columns` is empty after removing to avoid unnecessary alterations.

src/operator/src/insert.rs
killme2008: I dislike it, but it's acceptable. The opentsdb must support it too.

killme2008 (approved):
Almost LGTM

evenyag (approved):

PR #6141, opened 2025-05-20
chore: support `[<schema>.]<pipeline_name>` for cross reference pipeline

shuiyisong:

Introduce `PipelineName` to hold and reference the pipeline using name, schema, and version.
Now we can use `schema.pipeline_name` to refer to pipelines from other databases while ingesting logs.

## PR Checklist

### DIFF
--- a/src/frontend/src/instance/log_handler.rs
+++ b/src/frontend/src/instance/log_handler.rs
@@ -21,7 +21,7 @@ use client::Output;
 use common_error::ext::BoxedError;
 use datatypes::timestamp::TimestampNanosecond;
 use pipeline::pipeline_operator::PipelineOperator;
-use pipeline::{Pipeline, PipelineInfo, PipelineVersion};
+use pipeline::{Pipeline, PipelineName, PipelineVersion};
 use servers::error::{
     AuthSnafu, Error as ServerError, ExecuteGrpcRequestSnafu, InFlightWriteBytesExceededSnafu,
     PipelineSnafu, Result as ServerResult,
@@ -70,7 +70,7 @@ impl PipelineHandler for Instance {
         content_type: &str,
         pipeline: &str,
         query_ctx: QueryContextRef,
-    ) -> ServerResult<PipelineInfo> {
+    ) -> ServerResult<PipelineName> {
         self.pipeline_operator
             .insert_pipeline(name, content_type, pipeline, query_ctx)
             .await
--- a/src/frontend/src/instance/log_handler.rs
+++ b/src/frontend/src/instance/log_handler.rs
@@ -21,7 +21,7 @@ use client::Output;
 use common_error::ext::BoxedError;
 use datatypes::timestamp::TimestampNanosecond;
 use pipeline::pipeline_operator::PipelineOperator;
-use pipeline::{Pipeline, PipelineName, PipelineVersion};
+use pipeline::{Pipeline, PipelineName};
 use servers::error::{
     AuthSnafu, Error as ServerError, ExecuteGrpcRequestSnafu, InFlightWriteBytesExceededSnafu,
     PipelineSnafu, Result as ServerResult,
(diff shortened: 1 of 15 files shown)


sunng87:
Do we really want to make pipeline isolated by schema? Do you think it's a good idea to make pipeline namespace agnostic? Any pipeline in the catalog can be accessed by any schema/namespace

src/frontend/src/instance/log_handler.rs
```diff
@@ -54,12 +54,11 @@ impl PipelineHandler for Instance {
 
     async fn get_pipeline(
         &self,
-        name: &str,
-        version: PipelineVersion,
+        pipeline_name: &PipelineName,
```
killme2008, line 57: I prefer separating the name and version in code. If a combined representation is necessary, consider using the format `schema.pipeline:{version}`.

src/pipeline/src/manager.rs
```diff
@@ -40,8 +43,88 @@ pub mod util;
...
 pub type PipelineVersion = Option<TimestampNanosecond>;
 
-/// Pipeline info. A tuple of timestamp and pipeline reference.
-pub type PipelineInfo = (Timestamp, PipelineRef);
+/// Unique reference to pipeline name.
+/// A valid pipeline name should be in the format of [<schema>.]<name>
+/// The optional schema is for cross-schema reference.
+/// Note the version can be None while query, which means the latest version of the pipeline.
+#[derive(Debug, Clone)]
+pub struct PipelineName {
```
killme2008, line 51: `PipelineIdentity` is the more appropriate term.

src/pipeline/src/manager.rs
```diff
@@ -40,8 +43,88 @@ pub mod util;
...
+        let version = to_pipeline_version(version_str)?;
+
+        ensure!(
+            !name.is_empty(),
+            InvalidPipelineNameSnafu {
+                reason: "pipeline name cannot be empty",
+            }
+        );
+
+        let parts = name.split('.').collect::<Vec<&str>>();
```
killme2008, line 81: Do we need to consider quoted names such as `"test.db"."test pipeline"`?

killme2008: Looks like we need to process it like 

https://github.com/GreptimeTeam/greptimedb/blob/56e696bd55cf7d708d21b19729801703e5a44a3e/src/session/src/table_name.rs#L43

Use sql parser to parse the name into `ObjectName` at first.

src/pipeline/src/manager.rs
```diff
@@ -40,8 +43,88 @@ pub mod util;
...
+                .name
+                .starts_with(GREPTIME_INTERNAL_PIPELINE_NAME_PREFIX),
+            InvalidPipelineNameSnafu {
+                reason: "custom pipeline name cannot start with 'greptime_'",
+            }
+        );
+        Ok(())
+    }
+
+    pub fn check_schema(&self, ctx_schema: &str) -> Result<()> {
```
killme2008, line 113: Adds a document for public functions.
Notice that it is only used when adding pipelines.

src/pipeline/src/manager/table.rs
```diff
@@ -321,53 +315,61 @@ impl PipelineTable {
...
-                generate_pipeline_cache_key(schema, name, None),
+                generate_pipeline_cache_key(&pipeline_name),
                 compiled_pipeline.clone(),
             );
-            self.pipelines.insert(
-                generate_pipeline_cache_key(schema, name, Some(TimestampNanosecond(version))),
-                compiled_pipeline.clone(),
+            self.original_pipelines.insert(
+                generate_pipeline_cache_key(&pipeline_name),
+                (pipeline.to_owned(), version),
```
killme2008, line 333: Prefer `clone` over `to_owned`. Apply this suggestion throughout the file.

tests-integration/tests/http.rs
```diff
@@ -1542,7 +1542,52 @@ transform:
...
+    let res = client
+        .post("/v1/ingest?db=test_db&table=logs1&pipeline_name=test")
+        .header("Content-Type", "application/json")
+        .body(data_body)
+        .send()
+        .await;
+    assert_eq!(res.status(), StatusCode::BAD_REQUEST);
+
+    // cross ref using public's pipeline
+    let res = client
```
killme2008, line 1573: Should pipeline permissions for cross-database access be checked, or does the database access plugin already handle this? If not, should we intercept the request?

shuiyisong:
see https://github.com/GreptimeTeam/greptimedb/pull/6143

PR #6143, opened 2025-05-21
chore: shared pipeline under same catalog with compatibility

shuiyisong:

~~As mentioned [here](https://github.com/GreptimeTeam/greptimedb/pull/6141#issuecomment-2894561441), remove the schema in the pipeline information and cache key, so that all databases under the same catalog can access the shared pipelines.~~

~~**Note**: This is a breaking change. If the pipeline is already in use, check if pipelines have the same names but different schemas before upgrading.~~

Allow shared pipeline under the same catalog.
1. The newly created pipeline would be labeled under an empty string schema. **Note**, this would include the 'update' process, which means only the empty string schema is getting 'updated' or created.
2. The query of the pipeline would try to be as compatible as possible

**cache**
1. Try to find the pipeline labeled empty string schema, return if found.
2. Try to find the pipeline with the current schema, return if found.
3. Search all pipelines with the name and version suffix: if exactly one is found, return; else, throw an error.

**database**
1. Find all pipelines with the given name and version
2. If only one result is found, use the result
3. Check if there's an empty string schema result, use it if found
4. Check if there's a result labeled with the current schema, use it if found
5. Throw an error for multiple schema pipelines are found, but none of them is under the empty string schema and the current schema

## PR Checklist

### DIFF
--- a/src/pipeline/src/manager/table.rs
+++ b/src/pipeline/src/manager/table.rs
@@ -41,7 +41,7 @@ use table::TableRef;
 use crate::error::{
     BuildDfLogicalPlanSnafu, CastTypeSnafu, CollectRecordsSnafu, DataFrameSnafu,
     ExecuteInternalStatementSnafu, InsertPipelineSnafu, InvalidPipelineVersionSnafu,
-    MultiPipelineWithDiffSchemaSnafu, PipelineNotFoundSnafu, Result,
+    MultiPipelineWithDiffSchemaSnafu, PipelineNotFoundSnafu, RecordBatchLenNotMatchSnafu, Result,
 };
 use crate::etl::{parse, Content, Pipeline};
 use crate::manager::pipeline_cache::PipelineCache;
@@ -539,17 +539,19 @@ impl PipelineTable {
             );
 
             ensure!(
-                pipeline_content.len() == 1
-                    && pipeline_schema.len() == 1
-                    && pipeline_created_at.len() == 1,
-                PipelineNotFoundSnafu { name, version }
+                pipeline_content.len() == pipeline_schema.len()
+                    && pipeline_schema.len() == pipeline_created_at.len(),
+                RecordBatchLenNotMatchSnafu
             );
 
-            re.push((
-                pipeline_content.get_data(0).unwrap().to_string(),
-                pipeline_schema.get_data(0).unwrap().to_string(),
-                pipeline_created_at.get_data(0).unwrap(),
-            ));
+            let len = pipeline_content.len();
+            for i in 0..len {
+                re.push((
+                    pipeline_content.get_data(i).unwrap().to_string(),
+                    pipeline_schema.get_data(i).unwrap().to_string(),
+                    pipeline_created_at.get_data(i).unwrap(),
+                ));
+            }
         }
 
         Ok(re)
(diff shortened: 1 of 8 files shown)


killme2008:
> Note: This is a breaking change. If the pipeline is already in use, check if pipelines have the same names but different schemas before upgrading.

I am afraid some users already use it in this way. The breaking may not be acceptable.

sunng87:
Is it possible to do it in a semi-compatible way:

- Always find pipeline by name
  - if there are pipelines with same name from different schema, use the pipeline of current schema
  - otherwise use the only pipeline returned (this is the breaking behaviour)
- Do not store schema for newly created pipeline so there won't be duplicated pipelines
- When updated existed pipeline, keep the schema name

shuiyisong:
> Is it possible to do it in a semi-compatible way:
> 
>     * Always find pipeline by name
>       
>       * if there are pipelines with same name from different schema, use the pipeline of current schema
>       * otherwise use the only pipeline returned (this is the breaking behaviour)
> 
>     * Do not store schema for newly created pipeline so there won't be duplicated pipelines
> 
>     * When updated existed pipeline, keep the schema name

There is one obscure case. If there are multiple pipelines with the same name but none of them is under the current schema.

sunng87:
> This should be rare, we can throw an error if it occurs.

Sounds good.

shuiyisong:
@sunng87 @killme2008 PTAL

killme2008:
I wonder that the cloud catalog doesn't contain a public schema; how do we process it? @shuiyisong

shuiyisong:
> I wonder that the cloud catalog doesn't contain a public schema; how do we process it? @shuiyisong

This doesn't check the actual existence of the schema, which in this case is public. It's merely a string field in the `greptime_private.pipelines`.

src/pipeline/src/manager/table.rs
```diff
@@ -230,7 +216,7 @@ impl PipelineTable {
                 rows: vec![Row {
                     values: vec![
                         ValueData::StringValue(name.to_string()).into(),
-                        ValueData::StringValue(schema.to_string()).into(),
+                        ValueData::StringValue(DEFAULT_SCHEMA_NAME.to_string()).into(),
```
sunng87, line 219: Do you think if we can just use empty field here? This will prevent confusion in long term.

shuiyisong: The format of the cache key is `{schema}/{pipeline_name}/{version}`, using an empty string seems strange in here.

src/pipeline/src/manager/table.rs
```diff
@@ -297,55 +278,84 @@ impl PipelineTable {
...
+            .find(|v| v.1 == EMPTY_SCHEMA_NAME)
+            .or_else(|| pipeline_vec.iter().find(|v| v.1 == schema));
+
+        // multiple pipeline with no empty or current schema
+        // throw an error
+        let (pipeline_content, found_schema, version) =
+            pipeline.context(MultiPipelineWithDiffSchemaSnafu {
+                schemas: pipeline_vec
+                    .iter()
+                    .map(|v| v.1.split_once('/').unwrap().0)
```
sunng87, line 318: make sure the unwrap is safe here

tests-integration/tests/http.rs
```diff
@@ -1560,7 +1560,43 @@ transform:
...
+    assert_eq!(res.status(), StatusCode::OK);
+
+    // check write success
+    validate_data(
+        "pipeline_db_schema",
+        &client,
+        "select * from test_db.logs1",
+        "[[2436,2528,\"INTERACT.MANAGER\",\"I\",\"ClusterAdapter:enter sendTextDataToCluster\\\\n\",1716668197217000000]]",
+    )
+    .await;
```
paomian, line 1597: Should we make a test that updates from an old pipeline to a new pipeline?

paomian:
Rest LGTM

Issue #6139, opened 2025-05-20
Flow sqlness tests occasionally fail


v0y4g3r:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Other

### Minimal reproduce step

https://github.com/GreptimeTeam/greptimedb/actions/runs/15133172550/job/42539789720?pr=6138

### What did you expect to see?

All sqlness passes

### What did you see instead?

Flow related cases sometimes fail, including `flow_blog.sql` and `flow_basic.sql`

### What operating system did you use?

NA

### What version of GreptimeDB did you use?

Latest main branch

### Relevant log output and stack trace

```bash

```

discord9:
apprently there is no strict happen-before relation for two query in two seprarate connection, i.e.:
at time 0 conn 1: insert into table
at time 1 conn 2: select from that table, doesn't promise the insert into result must show up
~and there is no way to sync anything, so currently can only sleep and hope conn 2 catch up~

edited: still doesn't explain why first insert then select can get empty results since they do happen  in order

discord9:
as it turned out it's because if insert first then flush flow,
some times mirror insert task that spawn by frontend may not finish, but sql client already issue a `flush_flow` to flownode, which cause flush flow to flush nothing hence cause results not updated, a temporary fix is put in flush_flow to wait for a extra 100 ms see #6295  , a more fitting fix to sync between mirror insert task on frontend and flush_flow in flownode is being worked on now

PR #6147, opened 2025-05-21
fix: temp fix for flow flush flow sleep 1 sec

discord9:

https://github.com/GreptimeTeam/greptimedb/issues/6139

a temp fix that let flow flush flow sleep for one second before return, this temp fix help with passing posgre kvbackend sqlness,
but this is only a temporary fix to allow CI to pass, a proper fix is needed later

## PR Checklist


src/flow/src/batching_mode/engine.rs
```diff
@@ -375,6 +376,8 @@ impl BatchingEngine {
         let res = task
             .gen_exec_once(&self.query_engine, &self.frontend_client)
             .await?;
+        // TODO(discord9): found a better way to make sure results get into result table instead of waiting
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
evenyag, line 380: So the `gen_exec_once()` can't ensure this? Is it possible to use oneshot channels or something to get notifications?

evenyag:
> a temp fix that let flow flush flow sleep for one second before return, this temp fix help with passing posgre kvbackend sqlness, but this is only a temporary fix to allow CI to pass, a proper fix is needed later

Maybe we can create an issue for this.

discord9:
> > a temp fix that let flow flush flow sleep for one second before return, this temp fix help with passing posgre kvbackend sqlness, but this is only a temporary fix to allow CI to pass, a proper fix is needed later
> 
> Maybe we can create an issue for this.

#6139

src/flow/src/batching_mode/engine.rs
discord9: > So the `gen_exec_once()` can't ensure this? Is it possible to use oneshot channels or something to get notifications?

apprently there is no strict happen-before relation for two query in two seprarate connection, i.e.:
at time 0 conn 1: insert into table
at time 1 conn 2: select from that table, doesn't promise the `insert into` result must show up
and there is no way to sync anything, so can only sleep

killme2008:
@evenyag @discord9 Should we merge this PR? I am not sure if we want to. It hangs so long.

evenyag:
The failure rate is quite low, and the root cause is still unclear. I tend not to merge it.

PR #6155, opened 2025-05-22
fix: alter table update table column default

discord9:

as title

## PR Checklist

### DIFF
--- a/src/table/src/metadata.rs
+++ b/src/table/src/metadata.rs
@@ -817,17 +817,33 @@ impl TableMeta {
             );
         }
         // Collect columns after changed.
-        let columns: Vec<_> = table_schema
-            .column_schemas()
-            .iter()
-            .cloned()
-            .map(|mut column| {
-                if let Some(change_column) = modify_column_types.get(&column.name) {
-                    column.data_type = change_column.target_type.clone();
-                }
-                column
-            })
-            .collect();
+        let mut columns: Vec<_> = vec![];
+        for mut column in table_schema.column_schemas().iter().cloned() {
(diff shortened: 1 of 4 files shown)


fengjiachun (approved):
LGTM

src/table/src/metadata.rs
```diff
@@ -817,17 +817,33 @@ impl TableMeta {
...
-            .iter()
-            .cloned()
-            .map(|mut column| {
-                if let Some(change_column) = modify_column_types.get(&column.name) {
-                    column.data_type = change_column.target_type.clone();
-                }
-                column
-            })
-            .collect();
+        let mut columns: Vec<_> = vec![];
```
MichaelScofield, line 820:
```suggestion
        let mut columns: Vec<_> = Vec::with_capacity(table_schema.column_schemas().len());
```

src/table/src/metadata.rs
```diff
@@ -817,17 +817,33 @@ impl TableMeta {
...
+                                    default_value, &change_column.target_type
+                                ),
+                            })?,
+                    )
+                } else {
+                    None
+                };
+                column = column
+                    .clone()
+                    .with_default_constraint(new_default.clone())
```
MichaelScofield, line 840: Seems these two clones are not needed:

```suggestion
                    .with_default_constraint(new_default)
```

discord9: > Seems these two clones are not needed:

was used in the error msg below, also not too expensive to clone anyway

MichaelScofield (approved):

PR #6165, opened 2025-05-23
fix: bulk insert case sensitive

v0y4g3r:

Fix the problem when bulk ingesting record batch with uppercase columns, GreptimeDB will silently return an "internal error" to client.

## PR Checklist


fengjiachun:
Our gRPC insert is case-insensitive, so I think the bulk insert should be too.

fengjiachun (approved):
LGTM

PR #6191, opened 2025-05-27
chore: flush pipelines table immediately after add pipeline

paomian:

## PR Checklist


src/pipeline/src/manager/table.rs
```diff
@@ -261,6 +271,21 @@ impl PipelineTable {
             output
         );
 
+        // Flush the pipeline table to make sure the inserted pipeline is visible.
```
v0y4g3r, line 274: Maybe leave a todo here until we find a better solution

v0y4g3r (approved):
Rest LGTM

Issue #6156, opened 2025-05-22
Support Dynamic Routing of Prometheus Remote Write to Physical Tables via Metric Labels


walcool:
### What type of enhancement is this?

API improvement

### What does the enhancement do?

Currently, GreptimeDB’s Prometheus Remote Write ingestion supports specifying the dbname and physical_table in the remote write configuration. This requires setting up a separate remote write target in Prometheus for each physical table and database, which introduces operational overhead, especially when dealing with dynamic workloads and multi-tenant environments.

### Implementation challenges

We propose enhancing the frontend nodes in GreptimeDB to dynamically route incoming remote write requests to the appropriate physical_table and dbname based on specific label values present in the incoming time-series data.

**Motivation:**
Simplified Prometheus Configuration:
Avoid the need to configure separate remote write blocks per database/table combination.

**Dynamic Workload Support:**
Enable Prometheus to push all metrics to a single remote write endpoint. Let the GreptimeDB frontend route metrics based on labels like 

`__physical_table__` and `__db__`.

**Multi-Tenancy and Automation Friendly:**
Ideal for setups where workloads are dynamically provisioned (e.g., by tenants, namespaces, or services) and require isolation at the physical table or database level.

**Proposed Behavior:**
When a Prometheus remote write request is received, the frontend node will inspect each time series’ labels for:

`__physical_table__`

`__db__` 
Based on the label values, the data will be routed to the correct backend physical table and database dynamically.

If either label is missing, fallback behavior can be:

Reject the series with an error, or

Use a default database and table if configured.

After routing is determined, the `__db__` and `__physical_table__` labels should be stripped from the final stored series.

Example:
http_requests_total{`__db__`="tenant1", `__physical_table__`="metrics_table", service="api", code="200"} 1
physical table -> metrics_table
db -> tenant1

**Impact:**

1. Greatly reduces Prometheus config complexity.
2.  Makes GreptimeDB more attractive for dynamic, multi-tenant, or SaaS-style deployments.
3. Aligns with Prometheus best practices by leveraging label-based routing.

killme2008:
I think the new pipeline will solve this problem. https://github.com/GreptimeTeam/greptimedb/pull/5981

Dispatch the requests based on column values.


@shuiyisong

waynexia:
related to https://github.com/GreptimeTeam/greptimedb/issues/5930

sunng87:
[...]

Issue #5980, opened 2025-04-25
[Tracking Issue] Pipeline 2.0


shuiyisong:
### What problem does the new feature solve?

The Pipeline engine was originally designed for text pre-processing. It's a lightweight ETL component that extracts and transforms input text into the desired data type for storage before log ingestion.

As GreptimeDB moves towards a unified database for metrics, logs and traces, we hope the pipeline engine can work with all the input data, not only text logs. For example, we should be able to alter and modify label values from Prometheus remote write requests.

### What does the feature do?

The main goal of pipeline 2.0 is to 
1. Simplify the configuration, make the pipeline transformation more automatically inferred
2. Enable pipeline in all data entries like Prometheus remote write and OTLP, etc.
3. Integrate VRL for complex calculation

### Implementation challenges

Note: the TODO list is not decisive and can be modified

- [x] Automatic inferred transform https://github.com/GreptimeTeam/greptimedb/pull/6013
- [ ] Pipeline in data entries
  - [x] Prometheus remote write https://github.com/GreptimeTeam/greptimedb/pull/5981
  - [ ] OTLP
- [x] Integration with VRL
- [x] Select and dissoc processor https://github.com/GreptimeTeam/greptimedb/pull/6019
- [x] Default insert into database with transform specifying extra options
- [ ] Dispatch to specific physical table based on given label value
  - This requires setting table options in the pipeline config
- [ ] Drop requests on conditions at the beginning
- [x] Allow shared pipelines under the same catalog https://github.com/GreptimeTeam/greptimedb/pull/6143

killme2008:
Will the pipeline for prometheus remote write fix the #5361 ? Can it add external label for prometheus metrics.

shuiyisong:
> Will the pipeline for prometheus remote write fix the [#5361](https://github.com/GreptimeTeam/greptimedb/issues/5361) ? Can it add external label for prometheus metrics.

Yes, it should be able to add/modify labels during the pipeline process. To be more precise, we should be able to process any fields in requests (including labels in the prometheus remote write requests) with CRUDs just like the pipeline now.

killme2008:
@shuiyisong I think we can close this issue right now.

Issue #5930, opened 2025-04-18
pipeline supports simple ingestion routing logic based on column value


waynexia:
### What problem does the new feature solve?

Routing on write can make Greptimedb simpler to configure and use. For example, client can ingest into different tables or schemas with one endpoint, letting pipeline to generate `db` parameter or some headers based on the input data's value

### What does the feature do?

Be able to configure routing rule in pipeline

killme2008:
Already supported via dispatcher

https://docs.greptime.com/user-guide/logs/pipeline-config/#dispatcher

PR #6192, opened 2025-05-27
feat: pipeline with insert options

shuiyisong:

#5930 #6156 #5980 

Now the pipeline execution checks for hints from the context and uses them for later insertion.

Originally, the `run_pipeline` returns one `RowInsertRequests`. Since hints are set into the query context, we have to split the insertion into multiple calls, each of them starts with a cloned query context from the original query context, then sets the options from pipeline execution.

TODO: 
- [ ] add tests for metrics physical table option

## PR Checklist

### DIFF
--- a/src/pipeline/src/etl/ctx_req.rs
+++ b/src/pipeline/src/etl/ctx_req.rs
@@ -20,24 +20,34 @@ use ahash::{HashMap, HashMapExt};
 use api::v1::{RowInsertRequest, RowInsertRequests, Rows};
 use itertools::Itertools;
 use session::context::{QueryContext, QueryContextRef};
-use session::hints::{HINTS_KEY_PREFIX, HINT_KEYS};
 
 use crate::PipelineMap;
 
 const DEFAULT_OPT: &str = "";
 
+pub const PIPELINE_HINT_KEYS: [&str; 6] = [
+    "greptime_auto_create_table",
+    "greptime_ttl",
+    "greptime_append_mode",
+    "greptime_merge_mode",
+    "greptime_physical_table",
+    "greptime_skip_wal",
+];
+
+const PIPELINE_HINT_PREFIX: &str = "greptime_";
+
 // Remove hints from the pipeline context and form a option string
 // e.g: skip_wal=true,ttl=1d
 pub fn from_pipeline_map_to_opt(pipeline_map: &mut PipelineMap) -> String {
     let mut btreemap = BTreeMap::new();
-    for k in HINT_KEYS {
+    for k in PIPELINE_HINT_KEYS {
         if let Some(v) = pipeline_map.remove(k) {
             btreemap.insert(k, v.to_str_value());
         }
     }
     btreemap
         .into_iter()
-        .map(|(k, v)| format!("{}={}", k.replace(HINTS_KEY_PREFIX, ""), v))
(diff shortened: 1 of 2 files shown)


src/pipeline/src/etl.rs
```diff
@@ -153,21 +154,23 @@ impl DispatchedTo {
 /// The result of pipeline execution
 #[derive(Debug)]
 pub enum PipelineExecOutput {
-    Transformed((Row, Option<String>)),
+    Transformed((String, Row, Option<String>)),
```
paomian, line 157: Maybe it would be better to use a struct here.

tests-integration/tests/http.rs
```diff
@@ -1972,6 +1973,125 @@ table_suffix: _${type}
...
+    type: int32
+  - fields:
+      - type
+      - log
+      - logger
+    type: string
+  - field: time
+    type: time
+    index: timestamp
+table_suffix: _${type}
```
sunng87, line 2005: Can we make `table_suffix` a hint field instead of top-level configuration? I think this will allow us to use processor and VRL to generate any type of suffix, which is much richer than declarative way.

tests-integration/tests/http.rs
```diff
@@ -1972,6 +1973,125 @@ table_suffix: _${type}
...
+    let data_body = r#"
+[
+  {
+    "id1": "2436",
+    "id2": "2528",
+    "logger": "INTERACT.MANAGER",
+    "type": "http",
+    "time": "2024-05-25 20:16:37.217",
+    "log": "ClusterAdapter:enter sendTextDataToCluster\\n",
+    "x-greptime-hint-ttl": "1d"
```
sunng87, line 2028: I would prefer python style private syntax `__greptime_hint_ttl`. But `x-greptime-` should be ok too. Let me see if any of other reviewer has an idea for this.

src/session/src/context.rs
```diff
@@ -306,6 +306,10 @@ impl QueryContext {
         self.extensions.clone()
     }
 
+    pub fn clone_inner(&self) -> QueryContext {
+        self.clone()
+    }
```
sunng87, line 311: Why we need this?

src/pipeline/src/etl/ctx_req.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    pub fn merge(&mut self, other: Self) {
+        for (opt, req) in other.req {
+            self.req.entry(opt).or_default().extend(req);
+        }
+    }
+
+    pub fn as_req_iter(self, ctx: QueryContextRef) -> ContextReqIter {
+        let ctx = ctx.clone_inner();
```
sunng87, line 102: we should be able to use:

```suggestion
        let ctx = (*ctx).clone();
```

sunng87 (approved):

shuiyisong:
Note: A valid vrl variable can not start with an underscore, nor can it contain a hyphen. We might need to change the hint naming syntax.

ref: https://vector.dev/docs/reference/vrl/expressions/#variable

paomian (approved):

tests-integration/tests/http.rs
shuiyisong: Using `greptime_` as the prefix, since vrl doesn't allow underscore as a starting character.

Issue #6016, opened 2025-04-29
FlushRegions works asynchronously and serves as a non-mandatory hint, while FlushRegion enforces a synchronous flush operation. We will later rename FlushRegions to FlushRegionsHint or FlushRegionsSuggestion to make its purpose clearer.


MichaelScofield:
FlushRegions works asynchronously and serves as a non-mandatory hint, while FlushRegion enforces a synchronous flush operation. We will later rename FlushRegions to FlushRegionsHint or FlushRegionsSuggestion to make its purpose clearer.

_Originally posted by @WenyXu in https://github.com/GreptimeTeam/greptimedb/pull/5995#discussion_r2063261934_

yinheli:
I'd like to propose merging the current `FlushRegions` and `FlushRegion` instructions into a single, unified `FlushRegion` instruction. 

**Proposal:**

- Introduce a single `FlushRegion` struct:
  - `region_ids: Vec<RegionId>` — supports both single and batch flush.
  - `is_hint: bool` — if true, the flush is best-effort/async (like the current `FlushRegions`); if false, it’s enforced/sync (like the current `FlushRegion`).
- Update the instruction handler to branch on `is_hint` and handle accordingly.

May I have a try?

evenyag:
> May I have a try?

Sure!

@WenyXu @MichaelScofield What do you think about the proposal?

evenyag:
In batch mode, maybe we could use the try-all approach. We can control whether to fail-fast by a flag if necessary.

PR #6193, opened 2025-05-27
refactor: Unified FlushRegion instruction (#6016)

yinheli:

close https://github.com/GreptimeTeam/greptimedb/issues/6016

Unified FlushRegion instruction 

## PR Checklist

### DIFF
--- a/src/datanode/src/heartbeat/handler/flush_region.rs
+++ b/src/datanode/src/heartbeat/handler/flush_region.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use common_meta::instruction::{FlushRegions, InstructionReply, SimpleReply};
+use common_meta::instruction::{FlushRegion, InstructionReply, SimpleReply};
 use common_telemetry::warn;
 use futures_util::future::BoxFuture;
 use store_api::region_request::{RegionFlushRequest, RegionRequest};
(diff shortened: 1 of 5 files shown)


WenyXu (commented):
Merging the two types of flush region into a single instruction seems to introduce unnecessary complexity, especially since some code paths are not actually used. I believe combining the two instructions into one may not be the best fit. Would it be simpler to just rename `FlushRegions` to something like `FlushRegionsHint` or `FlushRegionsSuggestion` to better convey its non-mandatory nature?

src/datanode/src/heartbeat/handler/flush_region.rs
```diff
@@ -22,102 +22,186 @@ use crate::error;
...
+                Some(InstructionReply::FlushRegion(SimpleReply {
+                    result: success_count > 0, // true if at least one region succeeded
+                    error: Some(error_msg),
+                }))
             }
-        })
+        }
+    }
+
+    async fn flush_single_region(&self, region_id: RegionId) -> Option<InstructionReply> {
```
MichaelScofield, line 136: Can you use the `result = true/false` to indicate the flush result, instead of `Some/None`? I think it makes codes more understandable.

```suggestion
    async fn flush_single_region(&self, region_id: RegionId) -> InstructionReply {
```

fengjiachun:
Close it due to lack of progress; it can be reopened if there are any updates.

PR #6202, opened 2025-05-28
fix(mito): use 1day as default time partition duration

### LINKED ISSUE #6164: Optimize time window inferring logic

User experience

Make `1d` the default time window so that we won't end up with a huge time window when initial data ingested spans a very large time range.

### Implementation challenges

NA

v0y4g3r:

- fixes #6164

This PR changes the default partition duration from infinite to 1day.

`TimeParitions` aims to split inserted data according to a "part_duration". Originally the part_duration is not present, so all rows are written to a default time partition without time range. When the first compaction is triggered, GreptiemDB will infer a proper time window according to flushed files.

However this would cause problem when the initial data covers a very large time span, for example, 10 years. Then we would end up with an SST file covers a large time span, which would further make it overlap with possibly every new SST files.

In this PR, we make 1day the default `part_duration` when there's no compaction time window specified while creating tables. This default value should suit most cases. 

## PR Checklist


src/mito2/src/memtable/time_partition.rs
```diff
@@ -229,23 +220,7 @@ impl TimePartitions {
...
-                memtable.id(),
-            );
-            let part = TimePartition {
-                memtable,
-                time_range: None,
-            };
-            inner.parts.push(part);
-        }
-
+        let inner = PartitionsInner::new(next_memtable_id);
```
evenyag, line 223: The following comment is outdated. The time_range is not None.
https://github.com/GreptimeTeam/greptimedb/blob/b4d00fb49957030c0577a84cffbf8a7d0b1366ac/src/mito2/src/memtable/time_partition.rs#L211-L215

Maybe we can always initialize the duration to 1 day so we don't need to wrap `TimePartitions::part_duration` into `Option`.

tests/cases/standalone/common/flow/flow_basic.result
```diff
@@ -1382,31 +1388,32 @@ ADMIN FLUSH_FLOW('requests_long_term');
 SELECT
     *
 FROM
-    requests_without_ip;
+    requests_without_ip
+ORDER BY ts ASC;;
```
evenyag, line 1392: Maybe we only need to add `order by` to unstable results? Most results usually won't change.

v0y4g3r: > Maybe we only need to add `order by` to unstable results? Most results usually won't change. This PR modifies too many sqlness tests.

The result of `select` is always not guaranteed to be sorted, I think it should be mandatory to add `ORDER BY` in sqlness cases.

src/mito2/src/memtable/time_partition.rs
v0y4g3r: fixed in 47abc221669a550fc48b03ba01ff0bb995319ccd

src/mito2/src/memtable/time_partition.rs
```diff
@@ -418,26 +393,33 @@ impl TimePartitions {
...
-        let new_part = TimePartition {
-            memtable,
-            time_range: new_time_range,
-        };
+        let partitions_inner = old_stats
+            .time_range()
+            .and_then(|(_, old_stats_end_timestamp)| {
+                partition_start_timestamp(
+                    old_stats_end_timestamp,
+                    part_duration.unwrap_or(Duration::from_days(1)),
```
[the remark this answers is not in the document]
v0y4g3r: fixed in 2f534f8d0a6a24b3f9c08a16977d8550c1c47ff4

src/mito2/src/memtable/time_partition.rs
```diff
@@ -329,19 +304,18 @@ impl TimePartitions {
...
         {
             Some(pos) => pos,
             None => {
-                let range = PartTimeRange::from_start_duration(part_start, part_duration)
+                let range = PartTimeRange::from_start_duration(part_start, self.part_duration)
                     .with_context(|| InvalidRequestSnafu {
                         region_id: self.metadata.region_id,
                         reason: format!(
-                            "Partition time range for {part_start:?} is out of bound, bucket size: {part_duration:?}",
+                            "Partition time range for {part_start:?} is out of bound, bucket size: {:?}",self.part_duration
```
evenyag:
```suggestion
                            "Partition time range for {part_start:?} is out of bound, bucket size: {:?}", self.part_duration
```

fengjiachun (approved):
LGTM

PR #6205, opened 2025-05-28
feat(pipeline): vrl processor

shuiyisong:

## PR Checklist

### DIFF
--- a/src/pipeline/src/etl/processor/vrl.rs
+++ b/src/pipeline/src/etl/processor/vrl.rs
@@ -1,0 +1,206 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::BTreeMap;
+
+use chrono_tz::Tz;
+use snafu::{OptionExt, ResultExt};
+use vrl::compiler::runtime::Runtime;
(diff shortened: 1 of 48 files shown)


src/pipeline/src/etl/processor/vrl.rs
```diff
@@ -0,0 +1,319 @@
...
+                    msg: e.get_expression_error().to_string(),
+                }
+                .build()
+            })?;
+
+        vrl_value_to_pipeline_value(re)
+    }
+}
+
+impl TryFrom<&yaml_rust::yaml::Hash> for VrlProcessor {
```
sunng87, line 92: We can make this `source` key optional for better ergonomic. If the value of `vrl` is a string, just use it as source.

src/pipeline/src/etl/processor/vrl.rs
```diff
@@ -0,0 +1,319 @@
...
+        let pipeline_vrl = m
+            .into_iter()
+            .map(|(k, v)| pipeline_value_to_vrl_value(v).map(|v| (KeyString::from(k), v)))
+            .collect::<Result<BTreeMap<_, _>>>()?;
+
+        let mut target = TargetValue {
+            value: VrlValue::Object(pipeline_vrl),
+            metadata: VrlValue::Object(BTreeMap::new()),
+            secrets: Secrets::default(),
+        };
```
sunng87, line 75: In near future, we will need to implement `Target` on our own `PipelineMap`.

src/pipeline/src/etl/processor/vrl.rs
shuiyisong: Due to implementation details, we have to keep the source key(for parsing the YAML of each processor, the string literal is converted into a hash map before it goes into the processor's independent parsing logic). Also, we might add other keys like config options later for compiling the vrl source code.

sunng87 (approved):

src/pipeline/src/etl/processor/vrl.rs
paomian: I also think that the current btreemap doesn't meet our needs anymore, and we should wrap a type to provide more functionality and optimization.

paomian (approved):

Issue #6105, opened 2025-05-15
Enable common datasource support for Azblob, Oss, and Gcs storage


fengjiachun:
@fengjiachun 
just check the code and do the tests,
about copy parquet part for now its only support s3 and fs
https://github.com/GreptimeTeam/greptimedb/blob/9860bca986c3639468d261468802930e8dec79d2/src/common/datasource/src/object_store.rs#L56-L75

So I think just support s3 is OK and maybe support others in the future

_Originally posted by @yihong0618 in https://github.com/GreptimeTeam/greptimedb/pull/5585#discussion_r1986041379_

Currently, both `copy from/to table/database` and `external table` utilize `build_backend` to create object storage, but only S3 is supported at this time. We need to extend support to other types of object storage.

Issue #6220, opened 2025-05-30
Refactor CLI tools to support more object storage options


fengjiachun:
### What type of enhancement is this?

User experience

### What does the enhancement do?

Currently, several CLI tools like `data export` and `metadata export` support exporting data to S3 object storage. Ideally, exporting to other object storages should also be possible. However, this is limited by the `common datasource` only supporting S3 and FS. Once #6105  is completed, refactoring work can begin.

### Implementation challenges

Waiting for #6105

PR #6225, opened 2025-05-31
feat: Support export cli export to OSS

zqr10159:

#6105 
Support export cli export to OSS

Currently, both copying from/to tables/databases and external tables utilize build_backend to create object storage, but only S3 is supported at this time. This PR extends support to OSS.
image
image
image
image

cli example:
`greptime cli export --addr 127.0.0.1:4000 --oss --oss-bucket <your-bucket> --oss-endpoint <your-endopint> --oss-access-key-id <id> --oss-access-key-secret <secret>`

`greptime cli export --addr 127.0.0.1:4000 --target data --oss --oss-bucket <your-bucket> --oss-endpoint <your-endopint> --oss-access-key-id <id> --oss-access-key-secret <secret>`

`greptime cli export --addr 127.0.0.1:4000 --target schema --oss --oss-bucket <your-bucket> --oss-endpoint <your-endopint> --oss-access-key-id <id> --oss-access-key-secret <secret>`
## PR Checklist

### DIFF
--- a/src/cli/src/export.rs
+++ b/src/cli/src/export.rs
@@ -22,6 +22,7 @@ use clap::{Parser, ValueEnum};
 use common_error::ext::BoxedError;
 use common_telemetry::{debug, error, info};
 use object_store::layers::LoggingLayer;
+use object_store::services::Oss;
 use object_store::{services, ObjectStore};
 use serde_json::Value;
 use snafu::{OptionExt, ResultExt};
@@ -149,20 +150,44 @@ pub struct ExportCommand {
     /// if s3 is set, this is required
     #[clap(long)]
     s3_region: Option<String>,
+
+    /// if export data to oss
+    #[clap(long)]
+    oss: bool,
+
+    /// The oss bucket name
+    /// if oss is set, this is required
+    #[clap(long)]
+    oss_bucket: Option<String>,
+
+    /// The oss endpoint
+    /// if oss is set, this is required
+    #[clap(long)]
+    oss_endpoint: Option<String>,
+
+    /// The oss access key id
+    /// if oss is set, this is required
+    #[clap(long)]
+    oss_access_key_id: Option<String>,
+
+    /// The oss access key secret
+    /// if oss is set, this is required
+    #[clap(long)]
+    oss_access_key_secret: Option<String>,
 }
 
 impl ExportCommand {
     pub async fn build(&self) -> std::result::Result<Box<dyn Tool>, BoxedError> {
         if self.s3
             && (self.s3_bucket.is_none()
-                || self.s3_endpoint.is_none()
-                || self.s3_access_key.is_none()
-                || self.s3_secret_key.is_none()
-                || self.s3_region.is_none())
+            || self.s3_endpoint.is_none()
+            || self.s3_access_key.is_none()
+            || self.s3_secret_key.is_none()
+            || self.s3_region.is_none())
         {
             return Err(BoxedError::new(S3ConfigNotSetSnafu {}.build()));
         }
-        if !self.s3 && self.output_dir.is_none() {
+        if !self.s3 && !self.oss && self.output_dir.is_none() {
             return Err(BoxedError::new(OutputDirNotSetSnafu {}.build()));
         }
         let (catalog, schema) =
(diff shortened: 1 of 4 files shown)


yihong0618:
is OpenDAL support this?

zqr10159:
> is OpenDAL support this?

Sure! You can find it here. `https://nightlies.apache.org/opendal/opendal-docs-stable/core/#services`

killme2008:
Good job!

Please take a look @fengjiachun @WenyXu

WenyXu:
Hi @zqr10159 Considering that we might need to support additional object storage backends like GCS or Azure in the future, do you think it would be worth exploring a more general solution, such as introducing a macro?

zqr10159:
Yes, I agree with that! Suggested refactoring after completion, and the previous issue illustrates this as well. #6220

zqr10159:
> Hi @zqr10159 Considering that we might need to support additional object storage backends like GCS or Azure in the future, do you think it would be worth exploring a more general solution, such as introducing a macro?

Yes, I agree with that! Suggested refactoring after completion, and the previous issue illustrates this as well. #6220

src/common/datasource/src/object_store/oss.rs
```diff
@@ -0,0 +1,116 @@
...
+        ALLOW_ANONYMOUS,
+        BUCKET,
+        ENDPOINT,
+        ACCESS_KEY_ID,
+        ACCESS_KEY_SECRET,
+    ]
+    .contains(&key)
+}
+
+pub fn build_oss_backend(
```
WenyXu: Add comments for public functions

src/common/datasource/src/object_store/oss.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::error::{self, Result};
+
+const BUCKET: &str = "bucket";
+const ENDPOINT: &str = "endpoint";
+const ACCESS_KEY_ID: &str = "access_key_id";
+const ACCESS_KEY_SECRET: &str = "access_key_secret";
+const ROOT: &str = "root";
+const ALLOW_ANONYMOUS: &str = "allow_anonymous";
+
+pub fn is_supported_in_oss(key: &str) -> bool {
```
WenyXu, line 31: Isn't used anywhere?

src/cli/src/export.rs
```diff
@@ -216,6 +246,11 @@ pub struct Export {
     s3_access_key: Option<String>,
     s3_secret_key: Option<String>,
     s3_region: Option<String>,
+    oss: bool,
```
WenyXu, line 265: Should we use a field like scheme or object_storage_type? @fengjiachun

zqr10159:
@WenyXu Hi, I have solved the problems.

WenyXu:
@zqr10159 Thanks! Just a few minor nitpicks: we need to ensure that users accessing secrets are not logging into the shell. To improve security and avoid accidental leakage, perhaps we should introduce something like a `SecretString` type, which prevents the secret from being printed.

zqr10159:
> @zqr10159 Thanks! Just a few minor nitpicks: we need to ensure that users accessing secrets are not logging into the shell. To improve security and avoid accidental leakage, perhaps we should introduce something like a `SecretString` type, which prevents the secret from being printed.

Ok, I see.

zqr10159:
image
@WenyXu

WenyXu (approved):
Thanks @zqr10159, Rest LGTM

src/cli/src/export.rs
```diff
@@ -490,6 +543,8 @@ impl Export {
             Ok(op)
         } else if self.s3 {
             self.build_s3_operator().await
+        } else if self.oss {
```
fengjiachun, line 546: Possibly missing an if branch: `if self.oss && self.oss_ddl_local.is_some()` cc @discord9

src/cli/src/export.rs
fengjiachun: It's fine for me now; maybe we can do this after adding support for more object storage types.

src/cli/src/export.rs
discord9: > Possibly missing an if branch: `if self.oss && self.oss_ddl_local.is_some()` cc @discord9

Yes the reasoning(and counter-argument) is in here: https://github.com/GreptimeTeam/greptimedb/issues/6070#issuecomment-2909614790
TLDR: export client might not have direct access to underlying object store, so store sql files in local dir
However the behavior could need further discussion

zqr10159: Is there anything I can do?

fengjiachun: Just follow lines L537 to L543, but add a field like `oss_ddl_local_dir`, or rename `s3_ddl_local_dir` to `ddl_local_dir` for both `oss` and `s3`. What do you think?

zqr10159: I'm leaning toward a unifying name!

fengjiachun:
@zqr10159 Many thanks.

killme2008:
@zqr10159 Contribution appreciated!

PR #6227, opened 2025-06-03
feat: bloom filter index applier support or eq chain

zhongzc:

Apply bloom filter for exprs like `column = lit OR column = lit OR ...`

## PR Checklist


src/mito2/src/sst/index/bloom_filter/applier/builder.rs
```diff
@@ -356,6 +446,45 @@ mod tests {
         assert_eq!(column_predicates[0].list.len(), 3);
     }
 
+    #[test]
+    fn test_build_with_or_chain() {
```
evenyag, line 450: Also test case that is not the `or-eq` chain? e.g. eq null, with and operator

PR #6229, opened 2025-06-03
feat(flow): flow streaming mode in list expr support

discord9:

as title, useful in using a flow without aggr and just filter things out
and since optimizer sometimes will optimize long `OR` chain into in list, this is also useful

## PR Checklist

### DIFF
--- a/src/flow/src/transform/expr.rs
+++ b/src/flow/src/transform/expr.rs
@@ -490,7 +490,10 @@ impl TypedExpr {
                         expr: Box::new(typed_expr.expr),
                         list,
                     };
-                    Ok(TypedExpr::new(in_list_expr, typed_expr.typ))
+                    Ok(TypedExpr::new(
+                        in_list_expr,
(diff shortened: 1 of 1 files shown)


src/flow/src/expr/scalar.rs
```diff
@@ -222,9 +232,55 @@ impl ScalarExpr {
...
+            }
+        );
+
+        let found = eval_list
+            .iter()
+            .map(|v| v.to_arrow_array())
+            .try_fold(
+                BooleanArray::new(BooleanBuffer::new_unset(batch.row_count()), None),
+                |result, in_list| -> Result<BooleanArray, DataFusionError> {
+                    let rhs = compare_with_eq(&eval_expr.to_arrow_array(), &in_list, false)?;
```
zhongzc: `eval_expr.to_arrow_array()` can be moved outside the loop, and `in_list` should be changed to `in_list_elem` for better accuracy.

PR #6236, opened 2025-06-03
refactor: support to get trace id with time range

zyy17:

Support to get trace ID with a time range.

NOTE: The grafana jaeger plugin disables time parameters by default; you should enable it explicitly

image

## PR Checklist

### DIFF
--- a/src/frontend/src/instance/jaeger.rs
+++ b/src/frontend/src/instance/jaeger.rs
@@ -130,7 +130,13 @@ impl JaegerQueryHandler for Instance {
         .await?)
     }
 
-    async fn get_trace(&self, ctx: QueryContextRef, trace_id: &str) -> ServerResult<Output> {
+    async fn get_trace(
+        &self,
+        ctx: QueryContextRef,
+        trace_id: &str,
+        start_time: Option<i64>,
+        end_time: Option<i64>,
+    ) -> ServerResult<Output> {
         // It's equivalent to
         //
         // ```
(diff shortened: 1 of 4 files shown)


sunng87 (approved):
We can add a test case for this.

src/frontend/src/instance/jaeger.rs
```diff
@@ -139,13 +145,25 @@ impl JaegerQueryHandler for Instance {
...
+
+        if let Some(start_time) = start_time {
+            // Microseconds to nanoseconds.
+            filters.push(col(TIMESTAMP_COLUMN).gt_eq(lit_timestamp_nano(start_time * 1_000)));
+        }
+
+        if let Some(end_time) = end_time {
+            // Microseconds to nanoseconds.
+            filters.push(col(TIMESTAMP_COLUMN).lt_eq(lit_timestamp_nano(end_time * 1_000)));
+        }
```
evenyag, line 166: Grafana always uses microsecond for the start/end in the jaeger plugin?

zyy17: Yes, it always use microseconds.

src/servers/src/query_handler.rs
```diff
@@ -203,7 +203,13 @@ pub trait JaegerQueryHandler {
...
 
     /// Get trace by trace id. It's used for `/api/traces/{trace_id}` API.
-    async fn get_trace(&self, ctx: QueryContextRef, trace_id: &str) -> Result<Output>;
+    async fn get_trace(
+        &self,
+        ctx: QueryContextRef,
+        trace_id: &str,
+        start_time: Option<i64>,
+        end_time: Option<i64>,
+    ) -> Result<Output>;
```
shuiyisong, line 212: Nit: We can add comments to state that these i64 are microseconds

shuiyisong (approved):
LGTM 🚀

PR #6255, opened 2025-06-04
fix: convert JSON type to JSON string in COPY TABLE TO statment

### LINKED ISSUE #6240: COPY TABLE does not correctly export JSON format
### What type of bug is this?

Unexpected error

Other

### Minimal reproduce step

1.  Create table
```sql
CREATE TABLE jsons (j JSON, t timestamp time index);
```

2. Insert data
```sql
INSERT INTO jsons VALUES('[null]', 0),
('[true]', 1),
('[false]', 2);
```
3. Copy table to fs

```
COPY jsons TO '/tmp/export/demo.csv' WITH (format='csv');
```

All JSON values should be exported as JSON strings, like this:
```bash
cat /tmp/export/demo.csv
j,t
"[null]",1970-01-01T00:00:00
"[true]",1970-01-01T00:00:00.001
"[false]",1970-01-01T00:00:00.002
```

### What did you see instead?

```bash
cat /tmp/export/demo.csv
j,t
8000000100000000,1970-01-01T00:00:00
8000000140000000,1970-01-01T00:00:00.001
8000000130000000,1970-01-01T00:00:00.002
```

doesn't matter

doesn't matter

```bash

```

WenyXu:

fixes #6240

This PR fixes the JSON type export in COPY TABLE TO statement. Previously, it directly output the internal representation of JSON type, now it correctly converts JSON type to JSON string format.

#### Changes
- Add `SendableRecordBatchMapper` to handle type conversion in `RecordBatch` stream
- Implement JSON type to JSON string conversion logic
- Updated test cases to verify JSON type export functionality

## PR Checklist

### DIFF
--- a/src/common/recordbatch/src/lib.rs
+++ b/src/common/recordbatch/src/lib.rs
@@ -92,14 +92,20 @@ pub fn map_json_type_to_string(
     for (vector, schema) in batch.columns.iter().zip(original_schema.column_schemas()) {
         if let ConcreteDataType::Json(j) = schema.data_type {
             let mut string_vector_builder = StringVectorBuilder::with_capacity(vector.len());
+            let binary_vector = vector
+                .cast(&ConcreteDataType::binary_datatype())
+                .with_context(|_| error::CastVectorSnafu {
+                    from_type: schema.data_type.clone(),
+                    to_type: ConcreteDataType::string_datatype(),
+                })?;
+
             for i in 0..vector.len() {
-                let value = vector.get(i);
-                if value.is_null() {
+                if vector.is_null(i) {
                     string_vector_builder.push(None);
                 } else {
                     let string_value = json_type_value_to_string(
-                        // Safety: value is must be a binary value.
-                        value.as_value_ref().as_binary().unwrap().unwrap(),
+                        // Safety: value must be a binary value.
+                        binary_vector.get_ref(i).as_binary().unwrap().unwrap(),
                         &j.format,
                     )
                     .with_context(|_| error::CastVectorSnafu {
--- a/src/common/recordbatch/src/lib.rs
+++ b/src/common/recordbatch/src/lib.rs
@@ -30,8 +30,11 @@ pub use datafusion::physical_plan::SendableRecordBatchStream as DfSendableRecord
 use datatypes::arrow::compute::SortOptions;
 pub use datatypes::arrow::record_batch::RecordBatch as DfRecordBatch;
 use datatypes::arrow::util::pretty;
-use datatypes::prelude::VectorRef;
-use datatypes::schema::{Schema, SchemaRef};
+use datatypes::prelude::{ConcreteDataType, VectorRef};
+use datatypes::scalars::ScalarVectorBuilder;
+use datatypes::schema::{ColumnSchema, Schema, SchemaRef};
+use datatypes::types::json_type_value_to_string;
+use datatypes::vectors::StringVectorBuilder;
 use error::Result;
 use futures::task::{Context, Poll};
 use futures::{Stream, TryStreamExt};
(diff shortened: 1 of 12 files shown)


tests/cases/standalone/common/copy/copy_from_fs_csv.sql
```diff
@@ -15,6 +19,18 @@ Copy with_filename FROM '${SQLNESS_HOME}/demo/export/csv/demo.csv' with (format=
 
 select * from with_filename order by ts;
 
+CREATE TABLE with_json(host string, cpu double, memory double, jsons JSON, ts timestamp time index);
+
+Copy with_json FROM '${SQLNESS_HOME}/demo/export/json/demo.json' with (format='json');
+
+select host, cpu, memory, json_to_string(jsons), ts from with_json order by ts;
```
fengjiachun, line 26:
```suggestion
SELECT host, cpu, memory, json_to_string(jsons), ts FROM with_json order by ts;
```

src/common/recordbatch/src/lib.rs
```diff
@@ -58,6 +61,142 @@ pub struct OrderOption {
...
+    for (vector, schema) in batch.columns.iter().zip(original_schema.column_schemas()) {
+        if let ConcreteDataType::Json(j) = schema.data_type {
+            let mut string_vector_builder = StringVectorBuilder::with_capacity(vector.len());
+            for i in 0..vector.len() {
+                let value = vector.get(i);
+                if value.is_null() {
+                    string_vector_builder.push(None);
+                } else {
+                    let string_value = json_type_value_to_string(
+                        // Safety: value is must be a binary value.
```
killme2008:
```suggestion
                        // Safety: value must be a binary value.
```

src/common/recordbatch/src/lib.rs
```diff
@@ -58,6 +61,142 @@ pub struct OrderOption {
...
+pub fn map_json_type_to_string(
+    batch: RecordBatch,
+    original_schema: &SchemaRef,
+    mapped_schema: &SchemaRef,
+) -> Result<RecordBatch> {
+    let mut vectors = Vec::with_capacity(original_schema.column_schemas().len());
+    for (vector, schema) in batch.columns.iter().zip(original_schema.column_schemas()) {
+        if let ConcreteDataType::Json(j) = schema.data_type {
+            let mut string_vector_builder = StringVectorBuilder::with_capacity(vector.len());
+            for i in 0..vector.len() {
```
killme2008: I think it's better to use `iter_data` to iterate the vector:

https://greptimedb.rs/datatypes/vectors/struct.BinaryVector.html#method.get_data

We can downcast the vector into `BinaryVector`.

src/common/recordbatch/src/lib.rs
WenyXu: resolved in [4b29189](https://github.com/GreptimeTeam/greptimedb/pull/6255/commits/4b29189786d3ad9f9f446bc89366d0648f2c0c53)

src/common/recordbatch/src/lib.rs
WenyXu: resolved in [4b29189](https://github.com/GreptimeTeam/greptimedb/pull/6255/commits/4b29189786d3ad9f9f446bc89366d0648f2c0c53)

killme2008: Why not use `iter_data` instead of for loop?

WenyXu: Oh, I see. resolved in [d4c1e79](https://github.com/GreptimeTeam/greptimedb/pull/6255/commits/d4c1e79da311a97ba0bd52864d3d308d8c332754)

killme2008 (approved):
LGTM

PR #6268, opened 2025-06-08
feat: dist auto step aggr pushdown

discord9:

before:
```
frontend:
  aggr_fn(input)
datanode:
  send input
```
after:
```
frontend:
Projection(back to original column name for parent to recognize)
  aggr_calc(aggr_merge(state))
datanode:
  aggr_state(input) as state
  send state
```
greatly reduce codec cost and network cost, except for case where each time series have very small amount of rows, but massive amount of time series count, but in that case move data to frontend directly wouldn't help much either

currently it's hard-coded, but later can impl a `AggrStateWrapper` and `AggrMergeWrapper` to automatically create state/merge udaf impl for all supporting aggregate functions

p.s. this is a small pr, most changes are just sqlness test to test if it works

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/analyzer.rs
+++ b/src/query/src/dist_plan/analyzer.rs
@@ -156,6 +156,7 @@ struct PlanRewriter {
     partition_cols: Option<Vec<String>>,
     column_requirements: HashSet<Column>,
     expand_on_next_call: bool,
+    new_child_plan: Option<LogicalPlan>,
 }
 
 impl PlanRewriter {
@@ -211,6 +212,7 @@ impl PlanRewriter {
                     self.stage
                         .extend(transformer_actions.extra_parent_plans.into_iter().rev());
                     self.expand_on_next_call = true;
+                    self.new_child_plan = transformer_actions.new_child_plan;
                 }
             }
             Commutativity::NonCommutative
@@ -286,6 +288,10 @@ impl PlanRewriter {
     }
 
     fn expand(&mut self, mut on_node: LogicalPlan) -> DfResult<LogicalPlan> {
+        if let Some(new_child_plan) = self.new_child_plan.take() {
+            // if there is a new child plan, use it as the new root
+            on_node = new_child_plan;
+        }
         // store schema before expand
         let schema = on_node.schema().clone();
         let mut rewriter = EnforceDistRequirementRewriter {
--- a/src/query/src/dist_plan/commutativity.rs
+++ b/src/query/src/dist_plan/commutativity.rs
@@ -250,6 +250,7 @@ impl Categorizer {
                             debug!("After Step Optimize: {ret:?}");
                             ret.ok().map(|s| TransformerAction {
                                 extra_parent_plans: s.to_vec(),
+                                new_child_plan: None,
                             })
                         })),
                     };
@@ -430,6 +431,8 @@ pub struct TransformerAction {
     ///             MergeScan: ...
     /// ```
     pub extra_parent_plans: Vec<LogicalPlan>,
+    /// new child plan, if None, use the original plan.
+    pub new_child_plan: Option<LogicalPlan>,
 }
 
 pub fn partial_commutative_transformer(plan: &LogicalPlan) -> Option<LogicalPlan> {


discord9:
This is not the final design, but a poc that might be also useful to certain case i.e. flow batching mode

discord9:
apprently this optimize success on a large and complex query but failed on a simple one, I wonder why:
[try_analyze.zip](https://github.com/user-attachments/files/20691111/try_analyze.zip)

discord9:
This PoC Optimize is done, need review

src/query/src/dist_plan/commutativity.rs
```diff
@@ -15,20 +15,201 @@
...
+    let projection = LogicalPlan::Projection(new_projection);
+    // return the new logical plan
+    Ok(vec![projection, upper_aggr_plan])
+}
+
+/// Check if the given aggregate expression is steppable.
+/// As in if it can be split into multiple steps:
+/// i.e. on datanode first call `state(input)` then
+/// on frontend call `calc(merge(state))` to get the final result.
+///
```
v0y4g3r:
```suggestion
```

src/query/src/dist_plan/commutativity.rs
```diff
@@ -55,7 +236,21 @@ impl Categorizer {
             LogicalPlan::Filter(filter) => Self::check_expr(&filter.predicate),
             LogicalPlan::Window(_) => Commutativity::Unimplemented,
             LogicalPlan::Aggregate(aggr) => {
-                if !Self::check_partition(&aggr.group_expr, &partition_cols) {
+                let is_all_steppable = is_all_aggr_exprs_steppable(&aggr.aggr_expr);
+                let is_partition = Self::check_partition(&aggr.group_expr, &partition_cols);
```
evenyag: `is_partition` is confusing. Should be `matches_partition`.

src/query/src/dist_plan/commutativity.rs
```diff
@@ -15,20 +15,201 @@
...
+                )))
+            }
+        };
+
+        // deal with nested alias case
+        let mut new_aggr_expr = aggr_expr.clone();
+        {
+            let new_aggr_func = get_aggr_func_mut(&mut new_aggr_expr).unwrap();
+            *new_aggr_func = upper_func;
+        }
```
evenyag, line 105: Can you give some examples of how it solves the nested case?

src/query/src/dist_plan/analyzer.rs
```diff
@@ -190,12 +196,20 @@ impl PlanRewriter {
...
-                    && let Some(plan) = transformer(plan)
+                    && let Some(changed_plan) = transformer(plan)
                 {
-                    self.update_column_requirements(&plan);
-                    self.stage.push(plan)
+                    debug!("PlanRewriter: transformed plan: {changed_plan:#?} from {plan}");
+                    if let Some(last_stage) = changed_plan.last() {
+                        // update the column requirements from the last stage
+                        self.update_column_requirements(last_stage);
+                    }
```
evenyag: Should we update column requirements for all changed plans instead of only the last one?

src/query/src/dist_plan/commutativity.rs
discord9: > Can you give some examples of how it solves the nested case?

it only handle cases like simple Alias(AggregateExpr) lucky this seems to be the only kind of aggr expr datafusion will provide other than a simple and plain AggregateExpr

src/query/src/dist_plan/analyzer.rs
discord9: > Should we update column requirements for all changed plans instead of only the last one?

seems not needed? since only the last stage need to have all those column requirements, so merge scan can bring up those required columns

v0y4g3r (approved):
LGTM

killme2008:
@waynexia Please take a look.This PR changed the dist planner.

src/query/src/dist_plan/commutativity.rs
```diff
@@ -15,20 +15,206 @@
...
 pub enum Commutativity {
     Commutative,
     PartialCommutative,
     ConditionalCommutative(Option<Transformer>),
-    TransformedCommutative(Option<Transformer>),
+    TransformedCommutative {
+        /// Return plans from parent to child order
+        transformer: Option<StageTransformer>,
+        /// whether the transformer changes the child to parent
+        expand_on_parent: bool,
```
waynexia: when will this be false?

src/query/src/dist_plan/analyzer.rs
```diff
@@ -190,12 +196,20 @@ impl PlanRewriter {
...
                 }
             }
-            Commutativity::TransformedCommutative(transformer) => {
+            Commutativity::TransformedCommutative {
+                transformer,
+                expand_on_parent,
+            } => {
                 if let Some(transformer) = transformer
-                    && let Some(plan) = transformer(plan)
+                    && let Some(changed_plan) = transformer(plan)
```
waynexia:
```suggestion
                    && let Some(changed_plans) = transformer(plan)
```

src/query/src/dist_plan/commutativity.rs
```diff
@@ -15,20 +15,206 @@
...
+    let mut new_group_expr = new_aggr.group_expr.clone();
+    for expr in &mut new_group_expr {
+        if let Expr::Column(_) = expr {
+            // already a column, no need to change
+            continue;
+        }
+        let col_name = expr.qualified_name();
+        let input_column = Expr::Column(datafusion_common::Column::new(col_name.0, col_name.1));
+        *expr = input_column;
+    }
```
waynexia, line 125: what if I group by `a+1, b/2`?

src/query/src/dist_plan/analyzer.rs
```diff
@@ -190,12 +196,20 @@ impl PlanRewriter {
...
                 {
-                    self.update_column_requirements(&plan);
-                    self.stage.push(plan)
+                    debug!("PlanRewriter: transformed plan: {changed_plan:#?} from {plan}");
+                    if let Some(last_stage) = changed_plan.last() {
+                        // update the column requirements from the last stage
+                        self.update_column_requirements(last_stage);
+                    }
+                    self.stage.extend(changed_plan.into_iter().rev());
```
waynexia: why `.rev()`?

src/query/src/dist_plan/commutativity.rs
```diff
@@ -15,20 +15,206 @@
...
+    let upper_aggr_plan = LogicalPlan::Aggregate(new_aggr);
+    debug!("Before recompute schema: {upper_aggr_plan:?}");
+    let upper_aggr_plan = upper_aggr_plan.recompute_schema()?;
+    debug!("After recompute schema: {upper_aggr_plan:?}");
+    // create a projection on top of the new aggregate plan
+    let new_projection =
+        Projection::try_new(new_projection_exprs, Arc::new(upper_aggr_plan.clone()))?;
+    let projection = LogicalPlan::Projection(new_projection);
+    // return the new logical plan
+    Ok(vec![projection, upper_aggr_plan])
```
waynexia: If the length is fixed here, prefer using tuple or struct instead of vec

src/query/src/dist_plan/commutativity.rs
```diff
@@ -217,6 +417,8 @@ impl Categorizer {
 
 pub type Transformer = Arc<dyn Fn(&LogicalPlan) -> Option<LogicalPlan>>;
 
+pub type StageTransformer = Arc<dyn Fn(&LogicalPlan) -> Option<Vec<LogicalPlan>>>;
```
waynexia: 1. document about the signature of `Fn(&LogicalPlan) -> Option<Vec<LogicalPlan>>>`
2. this is not uniform.  `TransformedCommutative` should be able to modify the input plan as well. I.e., both new parent and child plan might be changed.

src/query/src/dist_plan/commutativity.rs
discord9: > what if I group by `a+1, b/2`?

The same as any complex projection expr, it get aliased as a new column with name being the schema name

discord9: details see this log file:
[sqlness-frontend.log](https://github.com/user-attachments/files/20794703/sqlness-frontend.log)
search for "transformed plan" and "integers.ts + Int64(1)"

src/query/src/dist_plan/analyzer.rs
discord9: > why `.rev()`?

because `step_aggr_to_upper_aggr` pushs plan from parent to child order, but `self.stage` is child to parent order, maybe `step_aggr_to_upper_aggr` can push from child to parent order instead?

src/query/src/dist_plan/commutativity.rs
discord9: > 1. document about the signature of `Fn(&LogicalPlan) -> Option<Vec<LogicalPlan>>>`
> 
>     2. this is not uniform.  `TransformedCommutative` should be able to modify the input plan as well. I.e., both new parent and child plan might be changed.

if it doesn't expand_on_next_call then it can modify child plan too? Also expand_on_next_call is always true for now

discord9: > 1. document about the signature of `Fn(&LogicalPlan) -> Option<Vec<LogicalPlan>>>`
> 
>     2. this is not uniform.  `TransformedCommutative` should be able to modify the input plan as well. I.e., both new parent and child plan might be changed.

seems no clean way to mutate child node here, and for now mutate child plan is not necessary so is ok for now?

src/query/src/dist_plan/commutativity.rs
discord9: > when will this be false?

removed

src/query/src/dist_plan/commutativity.rs
waynexia: it's necessary for cases like `avg -> sum/count` or `hll -> hll_calc(hll_state)`.

You can return both parent and child in this lambda, or a named structure for better clarity.

src/query/src/dist_plan/analyzer.rs
waynexia: not a huge point. let leave it as is

Issue #6242, opened 2025-06-04
Implement `Target` on `PipelineMap` for vrl


shuiyisong:
In near future, we will need to implement `Target` on our own `PipelineMap`.

_Originally posted by @sunng87 in https://github.com/GreptimeTeam/greptimedb/pull/6205#discussion_r2125055646_

PR #6278, opened 2025-06-09
refactor: remove `PipelineMap` and use `Value` instead

shuiyisong:

This is a pre PR to #6242, to replace `PipelineMap` with the upper level `Value`, so that we can later implement `Target` on the `Value` directly.

## PR Checklist

### DIFF
--- a/src/pipeline/src/error.rs
+++ b/src/pipeline/src/error.rs
@@ -809,14 +809,13 @@ impl ErrorExt for Error {
         use Error::*;
         match self {
             CastType { .. } => StatusCode::Unexpected,
-            ValueMustBeMap { .. } => StatusCode::Unexpected,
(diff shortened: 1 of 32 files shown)


src/pipeline/src/error.rs
```diff
@@ -803,6 +809,7 @@ impl ErrorExt for Error {
         use Error::*;
         match self {
             CastType { .. } => StatusCode::Unexpected,
+            ValueMustBeMap { .. } => StatusCode::Unexpected,
```
paomian: Maybe `StatusCode::InvalidArguments` better.

shuiyisong: If it's from the HTTP interfaces this error shouldn't happen. So it's not related to the user's input. 
Also, I'm thinking about putting this under `StatusCode::IllegalState`.

PR #6285, opened 2025-06-10
fix: ignore missing columns and tables in PromQL

evenyag:

This PR ignores missing columns and tables in PromQL, especially when using the `OR` set operator. This improves the compatibility of PromQL.
- The planner returns an `EmptyMetric` during planning when it can't find the table. The `EmptyMetric` will return 0 rows.
- The planner checks the label existence and ignores missing labels while planning `AggregateExpr`.

Users may only report some labels or metrics under specific conditions, so their PromQLs may refer to some labels/metrics that are not currently present.

It also modifies the `series` query API to return an empty result when the table to query doesn't exist.

## PR Checklist

### DIFF
--- a/src/query/src/promql/planner.rs
+++ b/src/query/src/promql/planner.rs
@@ -19,6 +19,8 @@ use std::time::UNIX_EPOCH;
 use arrow::datatypes::IntervalDayTime;
 use async_recursion::async_recursion;
 use catalog::table_source::DfTableSourceProvider;
+use common_error::ext::ErrorExt;
+use common_error::status_code::StatusCode;
 use common_query::prelude::GREPTIME_VALUE;
 use datafusion::common::DFSchemaRef;
 use datafusion::datasource::DefaultTableSource;
@@ -215,6 +217,7 @@ impl PromPlanner {
             PromExpr::Call(expr) => self.prom_call_expr_to_plan(session_state, expr).await?,
             PromExpr::Extension(expr) => self.prom_ext_expr_to_plan(session_state, expr).await?,
         };
+
         Ok(res)
     }
 
@@ -273,6 +276,7 @@ impl PromPlanner {
         } = aggr_expr;
 
         let input = self.prom_expr_to_plan(expr, session_state).await?;
+
         match (*op).id() {
             token::T_TOPK | token::T_BOTTOMK => {
                 self.prom_topk_bottomk_to_plan(aggr_expr, input).await
@@ -676,7 +680,9 @@ impl PromPlanner {
             at: _,
         } = vector_selector;
         let matchers = self.preprocess_label_matchers(matchers, name)?;
-        self.setup_context().await?;
+        if let Some(empty_plan) = self.setup_context().await? {
+            return Ok(empty_plan);
+        }
         let normalize = self
             .selector_to_series_normalize_plan(offset, matchers, false)
             .await?;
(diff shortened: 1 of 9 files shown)


src/query/src/promql/planner.rs
```diff
@@ -1336,13 +1346,18 @@ impl PromPlanner {
     }
 
     /// Setup [PromPlannerContext]'s state fields.
-    async fn setup_context(&mut self) -> Result<()> {
+    ///
+    /// Returns a logical plan for an empty metric.
```
killme2008, line 1350: Adds `must_use` hint?

tests-integration/tests/http.rs
```diff
@@ -667,12 +667,11 @@ pub async fn test_prom_http_api(store_type: StorageType) {
...
     assert_eq!(prom_resp.status, "error");
-    assert!(prom_resp
-        .error_type
-        .is_some_and(|err| err.eq_ignore_ascii_case("TableNotFound")));
-    assert!(prom_resp
-        .error
-        .is_some_and(|err| err.eq_ignore_ascii_case("Table not found: greptime.public.up")));
+    assert_eq!(prom_resp.error_type.unwrap(), "TableNotFound");
+    assert_eq!(
+        prom_resp.error.unwrap(),
```
killme2008: Is it compatible with the Prometheus API?

evenyag: No, but I'm not sure whether this PR should modify the logic related to label_values.

src/query/src/promql/planner.rs
evenyag: The hint doesn't check that we use the inner content of the Result.

killme2008: Yes, but it will warn if we don't take care the result.

tests-integration/tests/http.rs
killme2008: Ensure the API is compatible with Prometheus. If not addressed in this PR, please open an issue.

src/query/src/promql/planner.rs
evenyag: No, `must_use` only warns when we don't use the `Result`. There is no difference.

killme2008 (approved):
LGTM

killme2008:
@waynexia Please take a look. I think it can be merged.

src/query/src/promql/planner.rs
```diff
@@ -1385,7 +1400,32 @@ impl PromPlanner {
...
+
+        // The table doesn't have any data, so we set start to 0 and end to -1.
+        let plan = LogicalPlan::Extension(Extension {
+            node: Arc::new(
+                EmptyMetric::new(
+                    0,
+                    -1,
+                    self.ctx.interval,
+                    SPECIAL_TIME_FUNCTION.to_string(),
+                    DEFAULT_FIELD_COLUMN.to_string(),
```
waynexia, line 1422: not sure if we should use `GREPTIME_TIMESTAMP` and `GREPTIME_VALUE` instead. Both are Ok to me

waynexia (approved):
LGTM

src/query/src/promql/planner.rs
evenyag: Most places in the promql planner use these names, so I follow their patterns.

PR #6289, opened 2025-06-10
fix: null value handling on PromQL's join

waynexia:

We should consider NULL equals NULL when joining in PromQL. Theoretically, they are simply not considered when comparing two rows, hence "vacant equals vacant".

This patch updates the forked datafusion to include a fix patch where the variable `null_equals_null` are dropped in optimizer.

## PR Checklist


tests/cases/standalone/common/promql/set_operation.result
```diff
@@ -678,8 +678,14 @@ Affected Rows: 4
 -- null!=null, so it will returns the empty set.
```
evenyag: We can update the comment now.

killme2008 (approved):
LGTM

PR #6293, opened 2025-06-11
fix: multiple jaeger api compatible issues

zyy17:

## PR Checklist


src/frontend/src/instance/jaeger.rs
```diff
@@ -188,6 +188,8 @@ impl JaegerQueryHandler for Instance {
 
         let mut filters = vec![];
 
+        filters.push(col(SERVICE_NAME_COLUMN).eq(lit(query_params.service_name)));
```
killme2008, line 190: Should we validate that service_name is not empty before adding it to the filters?

zyy17: Already checked in https://github.com/GreptimeTeam/greptimedb/blob/main/src/servers/src/http/jaeger.rs#L251-L254. Also, I add some comments to describe it.

src/frontend/src/instance/jaeger.rs
```diff
@@ -208,15 +209,73 @@ impl JaegerQueryHandler for Instance {
...
+        //   timestamp <= {end_time}
+        // ```
+        let mut filters = vec![col(TRACE_ID_COLUMN).in_list(
+            trace_ids_from_output(output)
+                .await?
+                .iter()
+                .map(lit)
+                .collect::<Vec<Expr>>(),
+            false,
+        )];
```
sunng87, line 260: Just curious about why we need to run two calls for this? Is it designed to include some spans that might be out of the time range?

If we can use subquery we will be able to retrieve data in one call. But this is ok for now. I guess we may not be able to use subquery in dataframe api.

src/servers/src/http/jaeger.rs
```diff
@@ -215,7 +215,7 @@ pub struct JaegerQueryParams {
     pub operation_name: Option<String>,
 
     /// Limit the return data.
-    pub limit: Option<usize>,
+    pub limit: Option<String>,
```
sunng87: if we use `Option<usize>`, what we will get for empty limit? Using `String` here is a little weird.

sunng87: I wonder if there is any option in axum extractor for this

zyy17: Only parse if the string is not empty:

```
        if let Some(limit) = query_params.limit.filter(|d| !d.is_empty()) {
            internal_query_params.limit = Some(limit.parse::<usize>().map_err(|e| {
                InvalidJaegerQuerySnafu {
                    reason: format!("parse limit '{}' failed: {}", limit, e),
                }
                .build()
            })?);
        }
```

When I use jaeger-ui, we will get the following request if limit is empty:

```
http://localhost:5173/search?end=1749785009644000&limit=&lookback=1h&maxDuration&minDuration&operation=access-mysql&service=test-jaeger-query-api&start=1749781409644000
```

If we use `usize`, it will raise a error in middleware.

src/frontend/src/instance/jaeger.rs
zyy17: I refer the logic in [jaeger-clickhouse](https://github.com/jaegertracing/jaeger-clickhouse/blob/main/storage/clickhousespanstore/reader.go#L257-L267).

src/servers/src/http/jaeger.rs
sunng87: See this https://github.com/tokio-rs/axum/blob/main/examples/query-params-with-empty-strings/src/main.rs#L38

src/frontend/src/instance/jaeger.rs
```diff
@@ -458,3 +517,34 @@ fn tags_filters(
...
+        let recordbatches = util::collect(stream)
+            .await
+            .context(CollectRecordbatchSnafu)?;
+
+        // Only contains `trace_id` column in string type.
+        if !recordbatches.is_empty()
+            && schema.num_columns() == 1
+            && schema.contains_column(TRACE_ID_COLUMN)
+        {
+            let mut trace_ids = vec![];
```
shuiyisong, line 534: Nit: The capacity of the vec can be inferred from the limit.

src/servers/src/http/jaeger.rs
```diff
@@ -238,6 +243,20 @@ pub struct JaegerQueryParams {
     pub span_kind: Option<String>,
 }
 
+/// Serde deserialization decorator to map empty Strings to None,
+fn empty_string_as_none<'de, D, T>(de: D) -> Result<Option<T>, D::Error>
```
sunng87, line 247: nit: we can move this to a share module like util.

src/servers/src/http/jaeger.rs
```diff
@@ -403,8 +422,12 @@ pub async fn handle_get_trace(
         .with_label_values(&[&db, "/api/traces"])
         .start_timer();
 
+    // Convert start time and end time from microseconds to nanoseconds.
+    let start_time = query_params.start.map(|start| start * 1000);
+    let end_time = query_params.end.map(|end| end * 1000);
```
sunng87:
```suggestion
    let start_time_ns = query_params.start.map(|start_us| start_us * 1000);
    let end_time_ns = query_params.end.map(|end_us| end_us * 1000);
```

src/servers/src/http/jaeger.rs
```diff
@@ -403,8 +422,12 @@ pub async fn handle_get_trace(
...
+
     let output = match handler
-        .get_trace(query_ctx, &trace_id, query_params.start, query_params.end)
+        .get_trace(query_ctx, &trace_id, start_time, end_time)
```
sunng87:
```suggestion
        .get_trace(query_ctx, &trace_id, start_time_ns, end_time_ns)
```

sunng87 (approved):

PR #6295, opened 2025-06-11
feat: truly limit time range by split window

discord9:

as title, split large dirty time window into smaller ones for query, so that each flow can run a moderate size query without waiting for it to finish for too long, and avoid repeat computation due to query too eargely

## PR Checklist


src/flow/src/batching_mode/state.rs
```diff
@@ -206,47 +206,59 @@ impl DirtyTimeWindows {
...
+                        .context(TimeSnafu)?;
+                    to_be_query.insert(*start, Some(split_at));
+
+                    // remove the original window
+                    new_windows.remove(start);
+                    new_windows.insert(split_at, Some(end));
+                    cur_time_range += split_offset;
+                    break;
+                }
             }
```
v0y4g3r:
```suggestion
            let Some(x) = end.sub(start) else {
                continue;
            };
            if cur_time_range + x <= max_time_range {
                to_be_query.insert(*start, Some(end));
                new_windows.remove(start);
                cur_time_range += x;
            } else {
                // too large a window, split it
                // split at window_size * times
                let surplus = max_time_range - cur_time_range;
                let times = surplus.num_seconds() / window_size.num_seconds();

                let split_offset = window_size * times as i32;
                let split_at = start
                    .add_duration(split_offset.to_std().unwrap())
                    .context(TimeSnafu)?;
                to_be_query.insert(*start, Some(split_at));

                // remove the original window
                new_windows.remove(start);
                new_windows.insert(split_at, Some(end));
                cur_time_range += split_offset;
                break;
            }
```

src/flow/src/batching_mode/engine.rs
```diff
@@ -370,7 +370,7 @@ impl BatchingEngine {
         let task = self.tasks.read().await.get(&flow_id).cloned();
         let task = task.with_context(|| FlowNotFoundSnafu { id: flow_id })?;
 
-        task.mark_all_windows_as_dirty()?;
+        // task.mark_all_windows_as_dirty()?;
```
killme2008: Sorry, why is this line of code commented out?

v0y4g3r:
bugbot run

src/flow/src/batching_mode/engine.rs
discord9: > Sorry, why is this line of code commented out?

Added explain in comment, as it is usually prohibitive to update the entire time range(unless no time window is found, but that's also handled now), change to only update dirty time window instead, this behavior shouldn't be breaking though, as it still update dirty time window

src/flow/src/batching_mode/engine.rs
```diff
@@ -476,15 +476,36 @@ impl BatchingEngine {
         Ok(())
     }
 
+    /// Only flush the dirty windows of the flow task with given flow id, by running the query on it.
+    /// As flush the whole time range is usually prohibitively expensive.
     pub async fn flush_flow_inner(&self, flow_id: FlowId) -> Result<usize, Error> {
         debug!("Try flush flow {flow_id}");
+        // need to wait a bit to ensure previous mirror insert is handled
+        // this is only useful for the case when we are flushing the flow right after inserting data into it
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
discord9, line 486: as it turned out, it's the mirror insert(which is a async task) that happen before flush_flow in sql file, which might not complete insert into when admin flush_flow is called, this temp fix is to wait a bit so previous sql might be able to complete, proper fix need some kind of sync but it's complex

tests/cases/standalone/common/flow/flow_flush.result
```diff
@@ -42,11 +42,8 @@ SELECT
 FROM
     out_num_cnt_basic;
 
-+---------------------------------+-------------------------+
-| sum(numbers_input_basic.number) | time_window             |
-+---------------------------------+-------------------------+
-| 42                              | 2021-07-01T00:00:00.100 |
-+---------------------------------+-------------------------+
+++
+++
```
waynexia, line 46: is this expected?

waynexia (approved):

tests/cases/standalone/common/flow/flow_flush.result
discord9: > is this expected?

Currently this is expected, as data is inserted before flow's create, hence not possible to notify flow to flush correct time range, this can be refill later once we added `refill_flow` admin functions

v0y4g3r (approved):
LGTM

Issue #6182, opened 2025-05-26
`copy from` shouldn't create not exist directories


MichaelScofield:
### What type of bug is this?

User Experience

### What subsystems are affected?

Standalone mode

### Minimal reproduce step

Start a standalone locally and execute:

```sql
create table foo (ts timestamp time index);
copy foo from '/tmp/should-not-have-been-existed-dir/foo.parquet' with (format = 'parquet');
```

You'll see the directory `/tmp/should-not-have-been-existed-dir`been created. I think it's a bad side effect which must be resolved, since "copy from" is expected to be a read only operation for the file system, from the perspective of users.

### What did you expect to see?

If the directory is not existed, don't create it.

### What did you see instead?

Directory been created.

### What operating system did you use?

macos

### What version of GreptimeDB did you use?

latest

### Relevant log output and stack trace

```bash

```

Arshdeep54:
Hey @MichaelScofield , can you assign me this issue?

MichaelScofield:
@Arshdeep54 assigned!

PR #6300, opened 2025-06-11
fix: add path exist check in copy_table_from (#6182)

Arshdeep54:

Solves [Issue](https://github.com/GreptimeTeam/greptimedb/issues/6182) 

This PR adds a validation step in the `copy_table_from`  to ensure that the source path exists before proceeding. If the specified path does not exist, the operation now fails early with a clear `PathNotFound` error.

The intention is to prevent unexpected side effects—such as automatically creating non-existent directories—when running copy from in standalone mode. 

## PR Checklist

### DIFF
--- a/src/operator/src/error.rs
+++ b/src/operator/src/error.rs
@@ -831,6 +831,13 @@ pub enum Error {
         location: Location,
     },
 
+    #[snafu(display("Path not found: {path}"))]
+    PathNotFound {
+        path: String,
+        #[snafu(implicit)]
+        location: Location,
+    },
+
     #[cfg(feature = "enterprise")]
     #[snafu(display("Trigger related operations are not currently supported"))]
     UnsupportedTrigger {
@@ -975,6 +982,7 @@ impl ErrorExt for Error {
             Error::ComputeArrow { .. } => StatusCode::Internal,
             Error::InvalidProcessId { .. } => StatusCode::InvalidArguments,
             Error::ProcessManagerMissing { .. } => StatusCode::Unexpected,
+            Error::PathNotFound { .. } => StatusCode::InvalidArguments,
         }
     }
 
--- a/src/operator/src/statement/copy_table_from.rs
+++ b/src/operator/src/statement/copy_table_from.rs
@@ -14,6 +14,7 @@
 
 use std::collections::HashMap;
 use std::future::Future;
+use std::path::Path;
 use std::sync::Arc;
 
 use client::{Output, OutputData, OutputMeta};
@@ -23,7 +24,7 @@ use common_datasource::file_format::json::JsonFormat;
 use common_datasource::file_format::orc::{infer_orc_schema, new_orc_stream_reader, ReaderAdapter};
 use common_datasource::file_format::{FileFormat, Format};
 use common_datasource::lister::{Lister, Source};
-use common_datasource::object_store::{build_backend, parse_url};
+use common_datasource::object_store::{build_backend, parse_url, FS_SCHEMA};
 use common_datasource::util::find_dir_and_filename;
 use common_query::{OutputCost, OutputRows};
 use common_recordbatch::adapter::RecordBatchStreamTypeAdapter;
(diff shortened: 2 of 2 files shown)


src/operator/src/statement/copy_table_from.rs
```diff
@@ -98,6 +99,10 @@ impl StatementExecutor {
     ) -> Result<(ObjectStore, Vec<Entry>)> {
         let (_schema, _host, path) = parse_url(&req.location).context(error::ParseUrlSnafu)?;
 
+        if !Path::new(&path).exists() {
+            return Err(error::Error::PathNotFound { path });
+        }
```
MichaelScofield: To pass the ci:
```suggestion
        ensure!(
            Path::new(&path).exists(),
            PathNotFoundSnafu { path },
        );
```

MichaelScofield (commented):
Can you add some tests?

killme2008:
@MichaelScofield PTAL

src/operator/src/error.rs
```diff
@@ -831,6 +831,9 @@ pub enum Error {
         location: Location,
     },
 
+    #[snafu(display("Path not found: {path}"))]
+    PathNotFound { path: String },
```
evenyag: Please add location to the error. You can refer to other variants.

src/operator/src/statement/copy_table_from.rs
```diff
@@ -98,6 +99,11 @@ impl StatementExecutor {
     ) -> Result<(ObjectStore, Vec<Entry>)> {
         let (_schema, _host, path) = parse_url(&req.location).context(error::ParseUrlSnafu)?;
 
+        ensure!(
+            Path::new(&path).exists(),
+            PathNotFoundSnafu { path },
+        );
```
evenyag: The path can be an object store path, so `Path::exists()` may not work. We may use object_store to check the path's existence.

MichaelScofield:
@Arshdeep54 can you fix the ci? seems a simple rustfmt run is sufficient

src/operator/src/statement/copy_table_from.rs
Arshdeep54: I tried this 
```
 let object_store = build_backend(&req.location, &req.connection).context(error::BuildBackendSnafu)?;
object_store
                .stat(&path)
                .await
                .context(error::ReadObjectSnafu { path: &path })?;
```
But the directory was created before in `build_backend` funciton which calls  `FS_SCHEMA => Ok(build_fs_backend(&root)?),` if schema is FS . 

So should I do `Path::exists()` only for `schema==FS_schema` before calling `build_backend`  ?

```
if schema.to_uppercase() == FS_SCHEMA {
            ensure!(Path::new(&path).exists(), PathNotFoundSnafu { path });
        }
```

evenyag: Make sense. I guess the fs backend will create the directory. @WenyXu What do you think?

WenyXu: Hi @Arshdeep54 
> But the directory was created before in build_backend funciton which calls  FS_SCHEMA => Ok(build_fs_backend(&root)?), if schema is FS .

Can you find out why OpenDAL tries to create the directory?

v0y4g3r: > Hi @Arshdeep54
> 
> > But the directory was created before in build_backend funciton which calls  FS_SCHEMA => Ok(build_fs_backend(&root)?), if schema is FS .
> 
> Can you find out why OpenDAL tries to create the directory?

https://github.com/apache/opendal/blob/f5a9fb310a0c99474074f3d1767f506f8f10bbe2/core/src/services/fs/backend.rs#L92-L102

Looks like we cannot control that behavior. Manually checking url schema would be the best solution.

fengjiachun (approved):
LGTM

src/operator/src/statement/copy_table_from.rs
Arshdeep54: @fengjiachun Made these [changes](https://github.com/GreptimeTeam/greptimedb/pull/6300#discussion_r2145110670).

evenyag (approved):

fengjiachun (approved):

killme2008:
@Arshdeep54 Thanks for your contribution!

PR #6302, opened 2025-06-11
feat: support special labels parsing in prom remote write

shuiyisong:

#6156 

This PR adds support for parsing and setting the schema and physical table options from the Prom remote write's label value.
- `__database__` for setting the schema
- `__physical_table__` for setting the physical table info

Note: `__physical_table__` only takes effect upon first insertion(which is create-table-upon-insert process).

## PR Checklist

### DIFF
--- a/src/servers/src/prom_row_builder.rs
+++ b/src/servers/src/prom_row_builder.rs
@@ -20,16 +20,25 @@ use api::prom_store::remote::Sample;
 use api::v1::value::ValueData;
 use api::v1::{ColumnDataType, ColumnSchema, Row, RowInsertRequest, Rows, SemanticType, Value};
 use common_query::prelude::{GREPTIME_TIMESTAMP, GREPTIME_VALUE};
+use pipeline::{ContextOpt, ContextReq};
 use prost::DecodeError;
 
 use crate::http::PromValidationMode;
 use crate::proto::{decode_string, PromLabel};
 use crate::repeated_field::Clear;
 
+#[derive(Debug, Default, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct PromCtx {
+    pub schema: Option<String>,
+    pub physical_table: Option<String>,
+}
+
 /// [TablesBuilder] serves as an intermediate container to build [RowInsertRequests].
 #[derive(Default, Debug)]
 pub(crate) struct TablesBuilder {
-    tables: HashMap<String, TableBuilder>,
+    // tables: HashMap<String, TableBuilder>,
+    // schema -> table -> table_builder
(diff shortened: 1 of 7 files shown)


sunng87:
@waynexia @v0y4g3r please confirm this, we have user waiting for this feature. thanks!

src/servers/src/prom_row_builder.rs
```diff
@@ -20,16 +20,25 @@ use api::prom_store::remote::Sample;
...
 use common_query::prelude::{GREPTIME_TIMESTAMP, GREPTIME_VALUE};
+use pipeline::{ContextOpt, ContextReq};
 use prost::DecodeError;
 
 use crate::http::PromValidationMode;
 use crate::proto::{decode_string, PromLabel};
 use crate::repeated_field::Clear;
 
+#[derive(Debug, Default, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct PromCtx {
```
killme2008, line 32:
```suggestion
// Prometheus remote write context
pub struct PromCtx {
```

src/servers/src/prom_store.rs
```diff
@@ -39,9 +39,14 @@ use crate::error::{self, Result};
 use crate::row_writer::{self, MultiTableData};
 
 pub const METRIC_NAME_LABEL: &str = "__name__";
-
 pub const METRIC_NAME_LABEL_BYTES: &[u8] = b"__name__";
 
+pub const SCHEMA_LABEL: &str = "__schema__";
```
killme2008: The PR content said the key is `__database__`, could you confirm that? Do we need to support both of `schema` and `database`?

shuiyisong: Good catch, only `__database__` should be supported. We should not expose the idea of the catalog and the schema to end users.

src/servers/src/prom_row_builder.rs
```diff
@@ -20,16 +20,26 @@ use api::prom_store::remote::Sample;
...
+    pub schema: Option<String>,
+    pub physical_table: Option<String>,
+}
+
 /// [TablesBuilder] serves as an intermediate container to build [RowInsertRequests].
 #[derive(Default, Debug)]
 pub(crate) struct TablesBuilder {
-    tables: HashMap<String, TableBuilder>,
+    // tables: HashMap<String, TableBuilder>,
+    // schema -> table -> table_builder
```
v0y4g3r:
```suggestion
    // schema -> table -> table_builder
```

v0y4g3r (approved):
Rest LGTM

killme2008 (approved):
LGTM

PR #6306, opened 2025-06-12
feat: parallelism hint in grpc

discord9:

as title, query parallelism for each query, set in hint

## PR Checklist

### DIFF
--- a/src/session/src/hints.rs
+++ b/src/session/src/hints.rs
@@ -19,6 +19,7 @@ pub const HINTS_KEY_PREFIX: &str = "x-greptime-hint-";
 
 pub const READ_PREFERENCE_HINT: &str = "read_preference";
 
+/// Deprecated, use `HINTS_KEY` instead.
 pub const HINT_KEYS: [&str; 7] = [
     "x-greptime-hint-auto_create_table",
     "x-greptime-hint-ttl",


src/session/src/hints.rs
```diff
@@ -18,12 +18,14 @@ pub const HINTS_KEY_PREFIX: &str = "x-greptime-hint-";
...
+pub const HINT_KEYS: [&str; 8] = [
     "x-greptime-hint-auto_create_table",
     "x-greptime-hint-ttl",
     "x-greptime-hint-append_mode",
     "x-greptime-hint-merge_mode",
     "x-greptime-hint-physical_table",
     "x-greptime-hint-skip_wal",
     "x-greptime-hint-read_preference",
+    // same as `query.parallelism` in query options, but only for this query
+    "x-greptime-hint-query_parallelism",
```
evenyag: I can't find a place to set this hint.

src/query/src/datafusion.rs
```diff
@@ -478,6 +478,17 @@ impl QueryEngine for DatafusionQueryEngine {
...
+            if let Ok(n) = parallelism.parse::<u64>() {
+                let new_cfg = state.config().clone().with_target_partitions(n as usize);
+                *state.config_mut() = new_cfg;
+            } else {
+                common_telemetry::warn!(
+                    "Failed to parse query_parallelism: {}, using default value",
+                    parallelism
+                );
+            }
+        }
```
evenyag, line 493: Should we implement this by extension? Maybe we should define it as a field in the query context.

discord9: > Should we implement this by extension? Maybe we should define it as a field in the query context.

Might do it later in separate pr

src/session/src/hints.rs
discord9: > I can't find a place to set this hint.

key value pairs in hint is auto copy to query_ctx when running a query

fengjiachun (approved):
LGTM

Issue #5866, opened 2025-04-09
Process management


v0y4g3r:
### What problem does the new feature solve?

GreptimeDB lacks process management and users cannot know what's currently happening inside GreptimeDB.

### What does the feature do?

This issue proposes a process manager mechanism in GreptimeDB that:
- tracks running queries on GreptimeDB, by implementing an `information_schema.process_list` system table that records all running processes.
- allows user to manually kill long-running queries using `kill <process_id>` statement.

### Implementation challenges

- Implement process tracking container that can be shared between standalone and distributed mode
- Allow user to interrupt running queries

evenyag:
related issue: https://github.com/GreptimeTeam/greptimedb/issues/5072

v0y4g3r:
We can implement the process tracking container using `KvBackend`. The design of keys is crucial. I'd propose using frontend ip/id + a sequence number. In that case, each frontend can register queries and get an id, and once the query finishes, frontend can deregister the query using that id returned. Once frontend instances shutdown, we can safely remove all query keys start with this frontend ip address.

```rs
pub struct ProcessKey {
    pub frontend_ip: String,
    pub id: u64,
}

pub struct ProcessValue {
    pub database: String,
    pub query: String,
    pub start_timestamp_ms: i64,
}
```

And when implementing `kill` statement, we can find the frontend which is running that query using the `frontend_ip` part in the key.

killme2008:
@v0y4g3r In MySQL, the tables for processes is `processlist`: https://dev.mysql.com/doc/refman/8.4/en/information-schema-processlist-table.html

v0y4g3r:
> [@v0y4g3r](https://github.com/v0y4g3r) In MySQL, the tables for processes is `processlist`: https://dev.mysql.com/doc/refman/8.4/en/information-schema-processlist-table.html

I deliberately chose a different name for that because we have a very different schema for that table.

PR #6320, opened 2025-06-13
feat: handle `Ctrl-C` command in MySQL client

v0y4g3r:

- [x] Please don't merge until https://github.com/GreptimeTeam/greptime-proto/pull/245 is merged first.
- #5866 

Properly handle `Ctrl+C` command in MySQL client. Essentially, when user press `Ctrl+C` on the client-side, MySQL client will issue a `KILL QUERY <CONNCTION_ID>` statement to serverside. The conneciton id is allocated every time a client is connected to MySQL server. Thus, when we handle `KILL QUERY` statement, we should iterate all local processes to find out the process with the corresponding connection id.

Now we can elegantly terminate long running queries with `Ctrl+C`:

```sql
mysql> SELECT * FROM my_bench_table;
^C^C -- query aborted
ERROR 1317 (70100): (Cancelled): Stream has been cancelled
```

### Key Changes:
 - **Connection ID Management:**
   - Added `connection_id` to `Session` and `QueryContext` in `src/session/src/lib.rs` and `src/session/src/context.rs`.
   - Updated `MysqlInstanceShim` and `MysqlServer` to handle `connection_id` in `src/servers/src/mysql/handler.rs` and `src/servers/src/mysql/server.rs`.

 - **KILL Statement Enhancements:**
   - Introduced `Kill` enum to handle both `ProcessId` and `ConnectionId` in `src/sql/src/statements/kill.rs`.
   - Updated `ParserContext` to parse `KILL QUERY <connection_id>` in `src/sql/src/parser.rs`.
   - Modified `StatementExecutor` to support killing queries by `connection_id` in `src/operator/src/statement/kill.rs`.

 - **Process Management:**
   - Refactored `ProcessManager` to include `connection_id` in `src/catalog/src/process_manager.rs`.
   - Added `kill_local_process` method for local query termination.

 - **Testing:**
   - Added tests for `KILL` statement parsing and execution in `src/sql/src/parser.rs`.

## PR Checklist

### DIFF
--- a/src/operator/src/statement/kill.rs
+++ b/src/operator/src/statement/kill.rs
@@ -74,21 +74,19 @@ impl StatementExecutor {
         connection_id: u32,
     ) -> error::Result<usize> {
         let current_user_catalog = query_ctx.current_catalog().to_string();
-        let catalog_processes = pm
-            .list_all_processes(Some(current_user_catalog.as_ref()))
+        let matches = pm
+            .find_processes_by_connection_id(current_user_catalog.as_ref(), connection_id)
             .await
             .context(error::CatalogSnafu)?;
         let mut killed = 0;
-        for process in catalog_processes {
-            if process.connection_id == connection_id {
-                let succ = pm
-                    .kill_local_process(current_user_catalog.clone(), process.id)
-                    .await
-                    .context(error::CatalogSnafu)?;
-                if succ {
-                    killed += 1;
-                }
-            };
+        for process in matches {
+            let succ = pm
+                .kill_local_process(current_user_catalog.clone(), process.id)
+                .await
+                .context(error::CatalogSnafu)?;
+            if succ {
+                killed += 1;
+            }
         }
         Ok(killed)
     }
--- a/src/operator/src/statement/kill.rs
+++ b/src/operator/src/statement/kill.rs
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use catalog::process_manager::ProcessManagerRef;
 use common_frontend::DisplayProcessId;
 use common_query::Output;
 use common_telemetry::error;
 use session::context::QueryContextRef;
 use snafu::ResultExt;
+use sql::statements::kill::Kill;
 
 use crate::error;
 use crate::statement::StatementExecutor;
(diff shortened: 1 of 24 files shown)


src/operator/src/statement/kill.rs
```diff
@@ -25,22 +27,69 @@ impl StatementExecutor {
...
+
+    /// Handles MySQL `KILL QUERY <CONNECTION_ID>` statements.
+    pub async fn kill_connection_id(
+        &self,
+        pm: &ProcessManagerRef,
+        query_ctx: QueryContextRef,
+        connection_id: u32,
+    ) -> error::Result<usize> {
+        let current_user_catalog = query_ctx.current_catalog().to_string();
+        let catalog_processes = pm
```
killme2008: What about adding a new function `find_process_by_conn_id` to process manager?

src/sql/src/statements/statement.rs
```diff
@@ -139,7 +140,7 @@ pub enum Statement {
     // CLOSE
     CloseCursor(CloseCursor),
     // KILL <process>
-    Kill(String),
+    Kill(Kill),
```
killme2008, line 143: When to add sqlness tests?

killme2008 (commented):
Almost LGTM

src/servers/src/mysql/handler.rs
```diff
@@ -112,10 +114,12 @@ impl MysqlInstanceShim {
                 Default::default(),
                 // TODO(sunng87): generate process id properly
                 0,
+                Some(connection_id),
```
sunng87: Can we just use this connection_id for process_id, the `0` above?

sunng87: The mysql connection id is exact the process id: https://dev.mysql.com/doc/refman/8.4/en/information-functions.html#function_connection-id

v0y4g3r: > Can we just use this connection_id for process_id, the `0` above?

I don't think we can use process id as connection id.  The protocol requires an `u32`  connect_id:

https://github.com/GreptimeTeam/greptimedb/blob/5c8013832f3c22c55f8b2c1b27a75ad7c55504a5/src/servers/src/mysql/handler.rs#L348-L350

If we cast the `u64` process id as `connect_id`, that would result inconsistence.

Or should we just use `u32` for process id?

src/sql/src/statements/statement.rs
v0y4g3r: > When to add sqlness tests?

I'm wondering if sqlness support ctrl+c or creating a new session to issue kill command. I will try to add support for that.

src/operator/src/statement/kill.rs
v0y4g3r: fixed in https://github.com/GreptimeTeam/greptimedb/pull/6320/commits/77e65d707a260b415fb6f395ca87f69fce8c02b3

src/servers/src/mysql/handler.rs
sunng87: Yes. I suggest to use u32 for process id. In postgres it's 32 bit integer too.

src/sql/src/statements/statement.rs
yihong0618: we can support sleep for easy test
like:
https://github.com/GreptimeTeam/greptimedb/pull/5448

src/servers/src/mysql/handler.rs
v0y4g3r: > Yes. I suggest to use u32 for process id. In postgres it's 32 bit integer too.

fixed in https://github.com/GreptimeTeam/greptimedb/pull/6320/commits/8f300eeb3f343e0b4a0035ae0e0ce6d507a9d362

src/sql/src/statements/statement.rs
v0y4g3r: > we can support sleep for easy test like: #5448

Synchronous sleep would not work for current [process cancellation implementation](https://github.com/GreptimeTeam/greptimedb/blob/5449d791343d5525a743d1fa276af5c6f42da968/src/frontend/src/instance.rs#L242). Perhaps we need to support async sleep function.

sunng87 (approved):

killme2008:
@v0y4g3r The proto PR was merged, please update this PR.

PR #6349, opened 2025-06-18
refactor: make scanner creation async

MichaelScofield:

also with other small refactors

## PR Checklist

### DIFF
--- a/src/mito2/src/engine.rs
+++ b/src/mito2/src/engine.rs
@@ -98,9 +98,12 @@ use crate::error::{
     SerdeJsonSnafu,
 };
 use crate::manifest::action::RegionEdit;
+use crate::memtable::MemtableStats;
 use crate::metrics::HANDLE_REQUEST_ELAPSED;
 use crate::read::scan_region::{ScanRegion, Scanner};
+use crate::region::MitoRegionRef;
 use crate::request::{RegionEditRequest, WorkerRequest};
+use crate::sst::file::FileMeta;
 use crate::wal::entry_distributor::{
     build_wal_entry_distributor_and_receivers, DEFAULT_ENTRY_RECEIVER_BUFFER_SIZE,
 };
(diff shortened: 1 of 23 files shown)


src/mito2/src/engine.rs
```diff
@@ -245,6 +249,37 @@ impl MitoEngine {
...
+
+        Ok(region
+            .version()
+            .ssts
+            .levels()
+            .iter()
+            .flat_map(|x| x.files().map(|f| f.meta_ref()))
+            .cloned()
+            .collect())
+    }
```
evenyag: It'd be better to get memtables and SSTs from the same `Version` to ensure consistency. So you might return them in the same method.

src/mito2/src/read/projection.rs
```diff
@@ -199,6 +199,10 @@ impl ProjectionMapper {
         batch: &Batch,
         cache_strategy: &CacheStrategy,
     ) -> common_recordbatch::error::Result<RecordBatch> {
+        if batch.is_empty() {
+            return Ok(self.empty_record_batch());
+        }
```
evenyag: Is it possible that the `batch` is empty?

Currently, an empty record batch may have special meaning.
https://github.com/GreptimeTeam/greptimedb/blob/2ab08a8f935cf2e9157ca8905224903ba91995ba/src/mito2/src/read/seq_scan.rs#L293-L297

But we can change this later if we ensure the PartSort never relies on the `distinguish_range` flag, and this feature is no longer needed. @discord9 @waynexia

src/mito2/src/error.rs
```diff
@@ -1020,6 +1020,13 @@ pub enum Error {
         location: Location,
         source: mito_codec::error::Error,
     },
+
+    #[snafu(display("Unexpected: {reason}"))]
+    Unexpected {
+        reason: String,
+        #[snafu(implicit)]
+        location: Location,
+    },
```
evenyag: We already have some Unexpected errors: UnexpectedReplay and UnexpectedImpureDefault. Maybe we should merge them into one unexpected error.

src/mito2/src/read/projection.rs
discord9: Yes at least for now PartSortExec rely on empty batch to indicate `PartitionRange` see this comment:
https://github.com/GreptimeTeam/greptimedb/blob/2ab08a8f935cf2e9157ca8905224903ba91995ba/src/query/src/part_sort.rs#L46-L52

MichaelScofield:
@evenyag PTAL

v0y4g3r:
bugbot run

v0y4g3r (approved):

src/mito2/src/read/projection.rs
waynexia: We may file an issue to track this if needed.

waynexia:
bugbot run

PR #5852, opened 2025-04-09
fix: remove metadata region options

v0y4g3r:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

### Add `SKIP_WAL_KEY` Option to Metric Engine

 - **Enhancements**:
   - Introduced `SKIP_WAL_KEY` to the metric engine options in `create.rs` and `mito_engine_options.rs`.
   - Updated test cases in `create.rs` to include `skip_wal` option and ensure it is removed for metadata regions.

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #6350, opened 2025-06-18
refactor: make metadata region option opt-in

sunng87:

We have encountered a few errors with current opt-out approach #5852, #6340 . This patch is to change it to an opt-in way hopefully it gets easier to maintain in long term.

Please add other options to opt-in. @waynexia @WenyXu 

## PR Checklist

### DIFF
--- a/src/metric-engine/src/engine/create.rs
+++ b/src/metric-engine/src/engine/create.rs
@@ -33,7 +33,7 @@ use store_api::metric_engine_consts::{
     METADATA_SCHEMA_TIMESTAMP_COLUMN_INDEX, METADATA_SCHEMA_TIMESTAMP_COLUMN_NAME,
     METADATA_SCHEMA_VALUE_COLUMN_INDEX, METADATA_SCHEMA_VALUE_COLUMN_NAME,
 };
-use store_api::mito_engine_options::TTL_KEY;
+use store_api::mito_engine_options::{TTL_KEY, WAL_OPTIONS_KEY};
 use store_api::region_engine::RegionEngine;
 use store_api::region_request::{AffectedRows, RegionCreateRequest, RegionRequest};
 use store_api::storage::consts::ReservedColumnId;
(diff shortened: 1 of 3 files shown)


src/metric-engine/src/engine/create.rs
```diff
@@ -600,17 +597,17 @@ fn parse_physical_region_id(request: &RegionCreateRequest) -> Result<RegionId> {
...
-    original.remove(MEMTABLE_TYPE);
-    original.insert(TTL_KEY.to_string(), FOREVER.to_string());
-    original.remove(SKIP_WAL_KEY);
-    original
+    let mut metadata_region_options = HashMap::new();
+    metadata_region_options.insert(TTL_KEY.to_string(), FOREVER.to_string());
+
+    if let Some(wal_options) = original.get(WAL_OPTIONS_KEY) {
+        metadata_region_options.insert(WAL_OPTIONS_KEY.to_string(), wal_options.to_string());
+    }
```
evenyag, line 607: Should we reuse the compaction options for the data region? Maybe the metadata region should have different compaction options because it mainly stores KV pairs.

src/metric-engine/src/engine/create.rs
```diff
@@ -600,17 +597,17 @@ fn parse_physical_region_id(request: &RegionCreateRequest) -> Result<RegionId> {
 
 /// Creates the region options for metadata region in metric engine.
 pub(crate) fn region_options_for_metadata_region(
-    mut original: HashMap<String, String>,
+    original: &HashMap<String, String>,
 ) -> HashMap<String, String> {
     // TODO(ruihang, weny): add whitelist for metric engine options.
```
waynexia:
```suggestion
```

sunng87:
Please double-check if we need include other options from `original`, or add some metadata region specific config

src/metric-engine/src/engine/create.rs
v0y4g3r: meatadata region should not inherit the compaction parameters. Maybe we can manually specify a metadata region specific compaction parameter set in the future.

PR #6356, opened 2025-06-19
refactor: always write `parent_span_id` for otlp traces ingestion

zyy17:

1. **Ensure `parent_span_id` column always exist**

    When `parent_span_id` is not provided, the ingestion will not create the `parent_span_id` field in the trace table, which will lead to a `No field` error when querying `parent_span_id`.
  
    Since `parent_span_id` is a special field in otlp, we should always keep `parent_span_id` as a valid field. In this PR, it will always write the `parent_span_id`. If the `parent_span_id` is not provided, use the empty string.

2. **Support to write `None` type value**

   To reach 1, this PR also modifies the API of `write_fields()` and `write_by_schema()`: **`ValueData` -> `Option<ValueData>`**; 

## PR Checklist


src/servers/src/otlp/trace/v1.rs
```diff
@@ -105,22 +106,10 @@ pub fn write_span_to_row(writer: &mut TableData, span: TraceSpan) -> Result<()>
...
-                parent_span_id,
-            )),
-            &mut row,
-        )?;
-    }
-
-    let fields = vec![
+        make_string_column_data(
+            PARENT_SPAN_ID_COLUMN,
+            span.parent_span_id.unwrap_or_default(), // If the parent span id is not set, use an empty string.
```
sunng87: Don't use empty string here, we will query root span with `parent_span_id = NULL` in sql, using empty string will break this. Can we simply set it to `None`?

I understand this patch is to make sure the `parent_span_id` column will always exists.

zyy17: make sense.

zyy17: it seems we need to support write `None` value type in row writer. cc @fengjiachun

zyy17:
@fengjiachun, Please help me review the modification of `row_writer.rs`.

sunng87 (approved):

PR #6360, opened 2025-06-20
feat(pipeline): introduce pipeline doc version 2 for combine-transform

### LINKED ISSUE #6337: Opt-in transforms in pipeline

API improvement

In pipeline 1.0, all fields must by configured explicitly in `transform`. If I configured only 1 field then there will be only 1 field in result table, all other fields are dropped.

In pipeline 2.0, we introduced auto transform so user don't have to configure transform explicitly. We should allow user to configure it with on demand. All those fields that are not configured will apply our auto inference rules and ingested as well.

For those who want to stick with 1.0 approach, we can provide a toggle for that.

This is a must-have before 0.15 release.

### Implementation challenges

_No response_

shuiyisong:

close #6337 

Currently, we have a fully-set transform mode and an auto-transform mode
- fully-set transform: only fields explicitly noted in the transform section will be persisted into the database
- auto-transform: no transform section is written, and the pipeline engine will try to set all the fields from the pipeline context. But in this case, there is no way to set other indexes other than the time index.

What might come in handy is a new transform combining the advantages of the two existing ones. We only set some fields in the transform section, specifying particular datatype and index for them, then the rest of the fields are set by the pipeline engine. With the `select` processor, we can limit what we want and what we don't want in the final table.

However, this new mode is incompatible with the current version. If the user has already used pipeline with `dissect` or `regex` processors, after upgrading the database, the original message string gets immediately inserted into the database and there's no way to stop this behaviour.

So we have to introduce the concept of the pipeline doc version, just like the version in a Docker Compose file. Here is an example:
```YAML
version: 2
processors:
  - date:
    field: input_str
    formats:
      - "%Y-%m-%dT%H:%M:%S%.3fZ"

transform:
  - field: input_str, ts
    type: time
```

Simply add a `version` field at the top of the config file, and the pipeline engine will treat the transform process as the combine mode:
1. Process any written transform rules. Note that if the transform section is explicitly written, it must contain a time index field.
2. Add the remaining fields of the pipeline context to the final table.

There are some minor changes that come with this patch:
1. The `timestamp` processor is removed for its uncertainty regarding the resolution of the output. It's never on our public documentation, so it should be fine.
2. Although `table_suffix` is the top-level field in the pipeline file, it is evaluated after the processor section and before the transform section. So, referring to a field in the transform section won't work.

TODO: 
- [x] update description
- [x] add test

## PR Checklist


src/pipeline/src/etl.rs
```diff
@@ -153,31 +204,21 @@ impl DispatchedTo {
...
 pub enum PipelineExecOutput {
-    Transformed(TransformedOutput),
-    AutoTransform(AutoTransformOutput),
+    NewTransformed(NewTransformedInner),
     DispatchedTo(DispatchedTo, Value),
 }
 
 #[derive(Debug)]
-pub struct TransformedOutput {
+pub struct NewTransformedInner {
```
sunng87: what is the meaning of `New` here?

src/pipeline/src/etl.rs
```diff
@@ -243,51 +293,80 @@ impl Pipeline {
...
+                .context(AutoTransformOneTimestampSnafu)?;
+
+            // Create pipeline context with the found timestamp
+            let def = crate::PipelineDefinition::GreptimeIdentityPipeline(Some(
+                IdentityTimeIndex::Epoch(name.to_string(), ts.get_unit(), false),
+            ));
+            let n_ctx =
+                PipelineContext::new(&def, pipeline_ctx.pipeline_param, pipeline_ctx.channel);
+            values_to_row(schema_info, val, &n_ctx, Some(transformed_values))?
+        };
```
sunng87, line 352: Can we simplify this logic as:

match self.transformer():
  case GreptimeTransformer: check v1 or v2, apply transformer or values_to_row
  case AutoTransformer: ...

paomian:
This pr is a breaking change. After that, the user won't be able to transform a column more than once in the transform.

shuiyisong:
> This pr is a breaking change. After that, the user won't be able to transform a column more than once in the transform.

Fixed this by still using `get`, but removing the field in the end if it's `v2` mode.

sunng87 (approved):

paomian (approved):

PR #6358, opened 2025-06-19
feat(storage): Add skip_ssl_validation option for object storage HTTP client

rgidda:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

### Summary
Added a `skip_ssl_validation` configuration option to the HTTP client settings for object storage connections. This allows users to skip SSL certificate validation when connecting to object storage endpoints, which is useful for:
1. Development environments with self-signed certificates
2. Private networks with internal certificate authorities
3. Testing scenarios where certificate validation isn't feasible

### Implementation Details
1. Added `skip_ssl_validation` field to `HttpClientConfig` struct
2. Set default value to `false` to maintain secure defaults
3. Added comprehensive unit test to verify configuration parsing
4. Ensured the setting propagates through all storage backends (S3, OSS, Azure, GCS)
5. Added security warning in documentation comments

### Why this is needed
- Resolves connection issues with object storage providers using self-signed certificates
- Provides a workaround for environments where certificate authorities aren't available
- Maintains security by default while offering flexibility for specific use cases

### Security Note
⚠️ Skipping SSL validation exposes connections to man-in-the-middle attacks. This option should only be used in controlled environments with trusted networks. The default value (`false`) maintains secure validation.

## PR Checklist

- [x] I have written the necessary rustdoc comments
  - Added documentation for `skip_ssl_validation` with security warning
- [x] I have added the necessary unit tests and integration tests
  - Added `test_skip_ssl_validation_config` test case
- [x] This PR requires documentation updates
  - Will update configuration reference docs in separate PR
- [x] API changes are backward compatible
  - New field with default value maintains backward compatibility
- [x] Schema or data changes are backward compatible
  - No schema or data format changes

PR #6367, opened 2025-06-20
chore: remove log that prints skip_ssl_validation

evenyag:

https://github.com/GreptimeTeam/greptimedb/pull/6358

We have already printed options during startup phase:
```
2025-06-20T09:05:16.185336Z  INFO cmd::standalone: Standalone options: GreptimeOptions { ...
```

We can remove the log for skip_ssl_validation.

## PR Checklist


src/datanode/src/store.rs
```diff
@@ -212,10 +212,7 @@ pub(crate) fn build_http_client(config: &HttpClientConfig) -> Result<HttpClient>
         .connect_timeout(config.connect_timeout)
         .pool_idle_timeout(config.pool_idle_timeout)
         .timeout(config.timeout)
-        .danger_accept_invalid_certs({
-            info!("skip_ssl_validation: {}", config.skip_ssl_validation);
-            config.skip_ssl_validation
-        })
+        .danger_accept_invalid_certs(config.skip_ssl_validation)
```
killme2008, line 219: I think it's better to keep or improve the log.

evenyag: We already log the storage config, I don't think we need to print this flag again.

evenyag: I changed this to a warning when the flag is true.

killme2008: Yes. I think we should warn it when skipping the ssl validation.

v0y4g3r (approved):
LGTM

PR #6369, opened 2025-06-20
refactor: scan `Batch`es directly

MichaelScofield:

Because sometimes we want to scan out `Batch` directly, instead of getting a `RecordBatch` stream. So this PR split out the `Batch` to `RecordBatch` convention from scan stream.

## PR Checklist

### DIFF
--- a/src/mito2/src/read/scan_util.rs
+++ b/src/mito2/src/read/scan_util.rs
@@ -360,8 +360,8 @@ impl Drop for PartitionMetricsInner {
         self.in_progress_scan.dec();
 
         debug!(
-            "{} finished, region_id: {}, partition: {}, metrics: {:?}",
-            self.scanner_type, self.region_id, self.partition, metrics
+            "{} finished, region_id: {}, partition: {}, scan_metrics: {:?}, convert_batch_costs: {}",
+            self.scanner_type, self.region_id, self.partition, metrics, self.convert_cost,
         );
     }
 }
--- a/src/mito2/src/read/scan_util.rs
+++ b/src/mito2/src/read/scan_util.rs
@@ -23,7 +23,6 @@ use common_telemetry::debug;
 use datafusion::physical_plan::metrics::{ExecutionPlanMetricsSet, MetricBuilder, Time};
 use futures::Stream;
 use prometheus::IntGauge;
-use smallvec::SmallVec;
 use store_api::storage::RegionId;
 
 use crate::error::Result;
@@ -35,7 +34,6 @@ use crate::read::range::{RangeBuilderList, RowGroupIndex};
 use crate::read::scan_region::StreamContext;
 use crate::read::{Batch, ScannerMetrics, Source};
 use crate::sst::file::FileTimeRange;
-use crate::sst::parquet::file_range::FileRange;
 use crate::sst::parquet::reader::{ReaderFilterMetrics, ReaderMetrics};
 
 /// Verbose scan metrics for a partition.
@@ -47,8 +45,6 @@ struct ScanMetricsSet {
     build_reader_cost: Duration,
     /// Duration to scan data.
     scan_cost: Duration,
-    /// Duration to convert batches.
-    convert_cost: Duration,
     /// Duration while waiting for `yield`.
     yield_cost: Duration,
     /// Duration of the scan.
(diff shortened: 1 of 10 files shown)


MichaelScofield:
bugbot run

src/servers/src/grpc/flight/stream.rs
```diff
@@ -77,7 +78,11 @@ impl FlightRecordBatchStream {
         }
 
         while let Some(batch_or_err) = recordbatches.next().in_current_span().await {
-            match batch_or_err {
+            match batch_or_err.inspect_err(|e| {
+                if e.status_code().should_log_error() {
+                    error!("{e:?}");
+                }
+            }) {
```
evenyag: Why not log this in the `Err(e)` match arm?

src/store-api/src/storage/requests.rs
```diff
@@ -69,7 +69,7 @@ impl Display for ScanRequest {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         write!(f, "ScanRequest {{")?;
         if let Some(projection) = &self.projection {
-            write!(f, "projection: {:?},", projection)?;
+            write!(f, "projection: {:?}", projection)?;
```
evenyag: Perhaps we should always print the projection, as all subsequent `write!` calls use `, ` as a prefix.

src/mito2/src/read/stream.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let batch = futures::ready!(self.inner.poll_next_unpin(cx));
+        let Some(batch) = batch else {
+            self.partition_metrics
+                .inc_convert_batch_cost(self.convert_cost);
+            self.convert_cost = Duration::from_secs(0);
+
+            return Poll::Ready(None);
+        };
```
evenyag: It will acquire the metrics mutex for each batch in order to observe the convert cost. Before this, it only observed the metrics after completing one PartRange.

It'd be better to only update the metrics every N rows.

src/mito2/src/read/scan_util.rs
```diff
@@ -548,6 +553,28 @@ pub(crate) fn scan_file_ranges(
...
+        let stream = build_file_range_scan_stream(
+            stream_ctx,
+            part_metrics,
+            &mut reader_metrics,
+            read_type,
+            ranges,
+        );
+        for await batch in stream {
+            yield batch?;
+        }
```
evenyag: Can't we return the stream directly?

src/mito2/src/read/seq_scan.rs
```diff
@@ -418,7 +433,7 @@ pub(crate) fn build_sources(
     part_metrics: &PartitionMetrics,
     range_builder_list: Arc<RangeBuilderList>,
     sources: &mut Vec<Source>,
-) {
+) -> Result<()> {
```
evenyag: This method doesn't need to return a Result.

src/mito2/src/read/unordered_scan.rs
```diff
@@ -93,52 +94,67 @@ impl UnorderedScan {
...
+            } else {
+                scan_file_ranges(
+                    context,
+                    metrics,
+                    index,
+                    "unordered_scan_files",
+                    range_builder_list,
+                )
+                .boxed()
+            };
```
evenyag: Why do we wrap it into a function? It also boxes the stream, which could be avoided.

src/mito2/src/read/stream.rs
```diff
@@ -0,0 +1,121 @@
...
+        match batch {
+            ScanBatch::Normal(batch) => {
+                if batch.is_empty() {
+                    Ok(self.projection_mapper.empty_record_batch())
+                } else {
+                    self.projection_mapper.convert(&batch, &self.cache_strategy)
+                }
+            }
+            ScanBatch::Series(series) => {
+                let mut record_batches = Vec::with_capacity(series.batches.len());
```
evenyag: We should reuse the `record_batches`.

MichaelScofield:
@evenyag PTAL

src/mito2/src/read/scan_util.rs
MichaelScofield: Can't because `try_stream!` has to yield the item.

PR #6377, opened 2025-06-23
feat: state/merge wrapper for aggr func

discord9:

a helper to break a aggregate udf to it's `state` and `merge` function, so can send logical plan with `state` function to datanode to aggr, then `merge` at frontend, previously the relation from a aggregate udf to it's `state`/`merge` function is hard-coded in optimizer, this helper will make optimizer accept all aggregate udf(with a correctly defined `state_fields`) , there is some limitation in datafusion that's make it hard to deduce input types from state types, hence there is some mess in there.

## PR Checklist


waynexia:
@v0y4g3r please say "bugbot run" 🐶

src/query/src/dist_plan/aggr_wrapper.rs
```diff
@@ -0,0 +1,655 @@
...
+        }
+    }
+}
+
+impl StateMergeWrapper {
+    pub fn new<'a>(
+        original: AggregateUDF,
+        args: WrapperArgs<'a>,
+    ) -> datafusion_common::Result<Self> {
+        let state_functions = (0..original.state_fields(args.into())?.len())
```
zhongzc: How about saving the retrieved `state_field` as a field of `StateWrapper`? This way, the return type of `StateWrapper` doesn't need to be recalculated.

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -0,0 +1,934 @@
...
+            ordering_fields: args.ordering_fields,
+            is_distinct: args.is_distinct,
+        };
+        let state_fields = self.inner.state_fields(state_fields_args)?;
+        Ok(state_fields)
+    }
+}
+
+#[derive(Debug)]
+pub struct MergeAccum {
```
waynexia, line 477: brief document for public APIs

killme2008:
@zhongzc @evenyag @waynexia PTAL

discord9:
~working on better tests, meanwhile rest part of logic is ready for review~
ready for review

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -0,0 +1,486 @@
...
+/// The merge function is used to merge the states of the state functions.
+/// The merge function's name is in the format `__<aggr_name>_merge
+pub fn aggr_merge_func_name(aggr_name: &str) -> String {
+    format!("__{}_merge", aggr_name)
+}
+
+/// A wrapper to make an aggregate function out of the state and merge functions of the original aggregate function.
+/// It contains the original aggregate function, the state functions, and the merge function.
+///
+/// Notice state functions may have multiple output columns, so it's return type is always a struct array, and the merge function is used to merge the states of the state functions.
```
zhongzc, line 59: outdated

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -0,0 +1,534 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Wrapper for making aggregate functions out of state/merge functions of original aggregate functions.
+//!
+//! i.e. for a aggregate function `foo`, we will have a state function `foo_state` and a merge function `foo_merge`.
+//!
+//! `foo_state` i's input args is the same as `foo`'s, and its output is a state object.
```
waynexia:
```suggestion
//! `foo_state`'s input args is the same as `foo`'s, and its output is a state object.
```

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -0,0 +1,534 @@
...
+//! Note that `foo_state` might have multiple output columns, so it's a struct array.
+//! `foo_merge`'s input args is the same as `foo_state`'s, and its output is the same as `foo`'s.
```
waynexia:
```suggestion
//! `foo_merge`'s input arg is the same as `foo_state`'s output, and its output is the same as `foo`'s input.
```

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -0,0 +1,534 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Wrapper for making aggregate functions out of state/merge functions of original aggregate functions.
+//!
+//! i.e. for a aggregate function `foo`, we will have a state function `foo_state` and a merge function `foo_merge`.
+//!
+//! `foo_state` i's input args is the same as `foo`'s, and its output is a state object.
+//! Note that `foo_state` might have multiple output columns, so it's a struct array.
```
waynexia:
```suggestion
//! Note that `foo_state` might have multiple output columns, so it's a struct array
//! that each output column is a struct field.
```

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -0,0 +1,534 @@
...
+    let mut expr_ref = expr;
+    while let Expr::Alias(alias) = expr_ref {
+        expr_ref = &alias.expr;
+    }
+    if let Expr::AggregateFunction(aggr_func) = expr_ref {
+        Some(aggr_func)
+    } else {
+        None
+    }
+}
```
waynexia, line 83: Implement this as a `TreeNode` visitor would be more robust. Considering cases like
```sql
select max(c)+1 from t;
```

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -0,0 +1,534 @@
...
+        let aggr_plan = {
+            // certain aggr func need type coercion to work correctly, so we need to analyze the plan first.
+            if let LogicalPlan::Aggregate(aggr) = TypeCoercion::new().analyze(
+                LogicalPlan::Aggregate(aggr_plan).clone(),
+                &Default::default(),
+            )? {
+                aggr
+            } else {
+                return Err(datafusion_common::DataFusionError::Internal(
+                    "Expected an Aggregate plan".to_string(),
```
waynexia: This error should be "failed to coerce expressions in aggregate plan"

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -0,0 +1,534 @@
...
+        let mut upper = aggr_plan.clone();
+        upper.aggr_expr = upper_aggr_exprs;
+        upper.input = lower_plan.clone();
+        // upper schema's output schema should be the same as the original aggregate plan's output schema
+        let upper_check = upper.clone();
+        let upper_plan = Arc::new(LogicalPlan::Aggregate(upper_check).recompute_schema()?);
+        if *upper_plan.schema() != aggr_plan.schema {
+            return Err(datafusion_common::DataFusionError::Internal(format!(
+                "Upper aggregate plan's schema is not the same as the original aggregate plan's schema: {:?} != {:?}",
+                upper_plan.schema(), aggr_plan.schema
```
waynexia:
```suggestion
                upper_plan.schema(), aggr_plan.schema()
```

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -0,0 +1,534 @@
...
+
+        let mut upper = aggr_plan.clone();
+        upper.aggr_expr = upper_aggr_exprs;
+        upper.input = lower_plan.clone();
+        // upper schema's output schema should be the same as the original aggregate plan's output schema
+        let upper_check = upper.clone();
+        let upper_plan = Arc::new(LogicalPlan::Aggregate(upper_check).recompute_schema()?);
+        if *upper_plan.schema() != aggr_plan.schema {
+            return Err(datafusion_common::DataFusionError::Internal(format!(
+                "Upper aggregate plan's schema is not the same as the original aggregate plan's schema: {:?} != {:?}",
```
waynexia: schema might be very verbose
```suggestion
                "Upper aggregate plan's schema is not the same as the original aggregate plan's schema: \n[transformed]:{:?}\n[   original]{:?}",
```

waynexia (commented):
mostly LGTM

src/common/function/src/aggrs/aggr_wrapper.rs
discord9: > Implement this as a `TreeNode` visitor would be more robust. Considering cases like
> 
> ```sql
> select max(c)+1 from t;
> ```

case like this is parsed into plan like
```
Projection ["max(c)" + 1]
  Aggregate aggr_exprs: [max(c)]
...
```
so it's necessary to only handle aggregate with alias as other expr variant in here is unexpected and should error out anyway(and later code rely on this assumption to alias the output column of merge plan)
see the impl of `create_aggregate_expr_and_maybe_filter` which have the same assumption:

https://github.com/apache/datafusion/blob/d1e6eb4b3fcf1e16391360c48bfe4bf9748ce557/datafusion/core/src/physical_planner.rs#L1767

PR #6384, opened 2025-06-24
feat: supports CsvWithNames and CsvWithNamesAndTypes formats

### LINKED ISSUE #6067: Supports CSVWithNames and CSVWithNamesAndTypes formats

API improvement

We currently support rendering SQL and PromQL results in CSV format, as detailed in [CSV result format](https://github.com/GreptimeTeam/greptimedb/blob/main/src/servers/src/http/result/csv_result.rs) and its implementation in the `/sql` API ([link](https://github.com/GreptimeTeam/greptimedb/blob/51f2cb10531ba488c5f849748c6631888fd32a23/src/servers/src/http/handler.rs#L141)). However, the current output only includes rows, lacking column names and types.

ClickHouse's `CSVWithNames` and `CSVWithNamesAndTypes` formats, which include column names as headers and both names and types respectively, would greatly improve usability. These additions would facilitate easier data import and analysis in tools like Excel. Therefore, supporting these formats would be beneficial.

### Implementation challenges

_No response_

### LINKED ISSUE #6333: problem with v1/sql?db=public&format=csv api called exporting data as csv
### What type of bug is this?

Unexpected error

Query Engine

### Minimal reproduce step

invoking v1/sql?db=public&format=csv api for a certain table.
it may has relation with my table data, it worked well before, after insert some new rows to the table, it cause export error.but i can not specify which data may has problem

export export success

### What did you see instead?

`{
    "code": 1002,
    "error": "CSV error: found record with 34 fields, but the previous record has 36 fields",
    "execution_time_ms": 0
}`

mac

0.15.0

```bash

```

killme2008:

Close #6067 
Close #6333 

* Supports new response formats: `CsvWithNames` and `CsvWithNamesAndTypes`
* Fixed #6333, which is caused by JSON array type in trace data.

Once this PR is merged, we should update the dashboard to use `format=csvWithNames` instead of `format=csv` when downloading result CSVs. @sunchanglong @ZonaHex

## PR Checklist

### DIFF
--- a/src/servers/src/http.rs
+++ b/src/servers/src/http.rs
@@ -306,7 +306,8 @@ pub enum GreptimeQueryOutput {
 #[derive(Default, Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ResponseFormat {
     Arrow,
-    Csv,
+    // (with_names, with_types)
+    Csv(bool, bool),
     Table,
     #[default]
     GreptimedbV1,
@@ -318,7 +319,9 @@ impl ResponseFormat {
     pub fn parse(s: &str) -> Option<Self> {
         match s {
             "arrow" => Some(ResponseFormat::Arrow),
-            "csv" => Some(ResponseFormat::Csv),
+            "csv" => Some(ResponseFormat::Csv(false, false)),
+            "csvwithnames" => Some(ResponseFormat::Csv(true, false)),
+            "csvwithnamesandtypes" => Some(ResponseFormat::Csv(true, true)),
             "table" => Some(ResponseFormat::Table),
             "greptimedb_v1" => Some(ResponseFormat::GreptimedbV1),
             "influxdb_v1" => Some(ResponseFormat::InfluxdbV1),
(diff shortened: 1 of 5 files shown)


sunng87 (approved):

ZonaHex:
Does this affect v1/promql api?

killme2008:
> Does this affect v1/promql api?

Yes, all of the `format` parameters.

src/servers/src/http/result/csv_result.rs
```diff
@@ -18,6 +18,7 @@ use common_error::status_code::StatusCode;
 use common_query::Output;
 use mime_guess::mime;
 use serde::{Deserialize, Serialize};
+use serde_json::Value as JsonValue;
 
 use crate::http::header::{GREPTIME_DB_HEADER_EXECUTION_TIME, GREPTIME_DB_HEADER_FORMAT};
 // use super::process_with_limit;
```
evenyag:
```suggestion
```

src/servers/src/http.rs
```diff
@@ -306,7 +306,8 @@ pub enum GreptimeQueryOutput {
 #[derive(Default, Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ResponseFormat {
     Arrow,
-    Csv,
+    // (with_names, with_types)
+    Csv(bool, bool),
```
evenyag, line 310: nitpick: Can we name these fields?
```suggestion
    Csv { with_names: bool, with_types: bool },
```

evenyag (approved):

src/servers/src/http.rs
killme2008: There are only two fields, and the current implementation is appropriate.

PR #6390, opened 2025-06-25
feat(pipeline): support Loki API

### LINKED ISSUE #4800: Loki ingestion with greptime pipeline

API improvement

The loki ingestion API seems just staightforward. 

There is a concept called stream, just like labels of promethueus. It's key-value based pairs.
https://grafana.com/docs/loki/latest/reference/loki-http-api/#ingest-logs

The log body is just timestamp to content mapping.

```json
{
  "streams": [
    {
      "stream": {
        "label": "value"
      },
      "values": [
          [ "<unix epoch in nanoseconds>", "<log line>" ],
          [ "<unix epoch in nanoseconds>", "<log line>" ]
      ]
    }
  ]
}
```

And there are also a few thirdparty clients that can write to loki.
https://grafana.com/docs/loki/latest/send-data/

So it's possible by defining a pipeline that transform loki data into our table based format. We can even use the pipeline to parse content of log body for more key-value information.

### Implementation challenges

_No response_

shuiyisong:

close #4800

This PR adds support for the Loki API to run pipelines.
The process becomes a little complicated; see the diagram below for a quick glimpse.

Untitled Diagram

For entering the pipeline process
1. The original ts (in nanosecond timestamp datatype), Loki text line, and all the labels and structure metadata are merged into one big map
2. This map is used as the input data of the pipeline

## PR Checklist

### DIFF
--- a/src/servers/src/http/loki.rs
+++ b/src/servers/src/http/loki.rs
@@ -19,7 +19,7 @@ use std::time::Instant;
 use api::v1::value::ValueData;
 use api::v1::{
     ColumnDataType, ColumnDataTypeExtension, ColumnSchema, JsonTypeExtension, Row,
-    RowInsertRequest, RowInsertRequests, Rows, SemanticType, Value as GreptimeValue,
+    RowInsertRequest, Rows, SemanticType, Value as GreptimeValue,
 };
 use axum::extract::State;
 use axum::Extension;
@@ -32,8 +32,9 @@ use headers::ContentType;
 use jsonb::Value;
 use lazy_static::lazy_static;
 use loki_proto::logproto::LabelPairAdapter;
-use loki_proto::prost_types::Timestamp;
-use pipeline::SchemaInfo;
+use loki_proto::prost_types::Timestamp as LokiTimestamp;
+use pipeline::util::to_pipeline_version;
+use pipeline::{ContextReq, PipelineContext, PipelineDefinition, SchemaInfo};
 use prost::Message;
 use quoted_string::test_utils::TestSpec;
 use session::context::{Channel, QueryContext};
@@ -41,16 +42,17 @@ use snafu::{ensure, OptionExt, ResultExt};
 
 use crate::error::{
     DecodeOtlpRequestSnafu, InvalidLokiLabelsSnafu, InvalidLokiPayloadSnafu, ParseJsonSnafu,
-    Result, UnsupportedContentTypeSnafu,
+    PipelineSnafu, Result, UnsupportedContentTypeSnafu,
 };
-use crate::http::event::{LogState, JSON_CONTENT_TYPE, PB_CONTENT_TYPE};
-use crate::http::extractor::LogTableName;
+use crate::http::event::{LogState, PipelineIngestRequest, JSON_CONTENT_TYPE, PB_CONTENT_TYPE};
+use crate::http::extractor::{LogTableName, PipelineInfo};
 use crate::http::result::greptime_result_v1::GreptimedbV1Response;
 use crate::http::HttpResponse;
 use crate::metrics::{
     METRIC_FAILURE_VALUE, METRIC_LOKI_LOGS_INGESTION_COUNTER, METRIC_LOKI_LOGS_INGESTION_ELAPSED,
     METRIC_SUCCESS_VALUE,
 };
+use crate::pipeline::run_pipeline;
 use crate::prom_store;
 
 const LOKI_TABLE_NAME: &str = "loki_logs";
@@ -97,6 +99,7 @@ pub async fn loki_ingest(
     Extension(mut ctx): Extension<QueryContext>,
     TypedHeader(content_type): TypedHeader<ContentType>,
     LogTableName(table_name): LogTableName,
+    pipeline_info: PipelineInfo,
     bytes: Bytes,
 ) -> Result<HttpResponse> {
     ctx.set_channel(Channel::Loki);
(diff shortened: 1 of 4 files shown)


src/servers/src/http/loki.rs
```diff
@@ -106,213 +112,453 @@ pub async fn loki_ingest(
...
+                        warn!("not a valid value, {:?}", e);
+                    }
+                }
+            }
+        }
+        if let Some(v) = labels {
+            v.into_iter().for_each(|(k, v)| {
+                map.insert(k, pipeline::Value::String(v));
+            });
+        }
```
shuiyisong: Discussion: Do we need to add a prefix to the structured_metadata and labels?

sunng87: it may conflict with content generated by pipeline from `line` right?

sunng87: I think we can prefix them with `loki` just like the timestamp and line column

shuiyisong: Prefixing them with `loki_metadata_` and `loki_label_` to distinguish them.

sunng87 (approved):

src/servers/src/http/loki.rs
```diff
@@ -32,31 +31,40 @@ use common_telemetry::{error, warn};
...
     METRIC_SUCCESS_VALUE,
 };
+use crate::pipeline::run_pipeline;
 use crate::prom_store;
 
 const LOKI_TABLE_NAME: &str = "loki_logs";
 const LOKI_LINE_COLUMN: &str = "line";
 const LOKI_STRUCTURED_METADATA_COLUMN: &str = "structured_metadata";
 
+const LOKI_TIMESTAMP_COLUMN_NAME: &str = "loki_timestamp";
```
killme2008: Do we need to reuse `greptime_timestamp` just like other protocols?

src/servers/src/http/loki.rs
```diff
@@ -106,213 +115,464 @@ pub async fn loki_ingest(
...
-    // init schemas
-    let mut schemas = LOKI_INIT_SCHEMAS.clone();
+    let handler = log_state.log_handler;
 
-    let mut rows = match content_type {
-        x if x == *JSON_CONTENT_TYPE => handle_json_req(bytes, &mut schemas).await,
-        x if x == *PB_CONTENT_TYPE => handle_pb_req(bytes, &mut schemas).await,
-        _ => UnsupportedContentTypeSnafu { content_type }.fail(),
-    }?;
+    let c = if let Some(pipeline_name) = pipeline_info.pipeline_name {
```
killme2008: `c` is not a good name, what about `request`?

src/servers/src/http/loki.rs
```diff
@@ -106,213 +115,464 @@ pub async fn loki_ingest(
...
-            .with_label_values(&[db_str])
-            .inc_by(*rows as u64);
-        METRIC_LOKI_LOGS_INGESTION_ELAPSED
-            .with_label_values(&[db_str, METRIC_SUCCESS_VALUE])
-            .observe(exec_timer.elapsed().as_secs_f64());
-    } else {
-        METRIC_LOKI_LOGS_INGESTION_ELAPSED
-            .with_label_values(&[db_str, METRIC_FAILURE_VALUE])
-            .observe(exec_timer.elapsed().as_secs_f64());
+    let mut outputs = Vec::new();
```
killme2008: Since we know the exact output size, can we create it with a capacity based on `c` requests?

src/servers/src/http/loki.rs
```diff
@@ -106,213 +115,464 @@ pub async fn loki_ingest(
...
-    column_indexer.insert(GREPTIME_TIMESTAMP.to_string(), 0);
-    column_indexer.insert(LOKI_LINE_COLUMN.to_string(), 1);
-
-    let payload: serde_json::Value =
-        serde_json::from_slice(bytes.as_ref()).context(ParseJsonSnafu)?;
-
-    let streams = payload
-        .get(STREAMS_KEY)
-        .context(InvalidLokiPayloadSnafu {
+pub struct LokiMiddleItem<T> {
```
killme2008, line 202: Love to document this and the following pub strucutures.

src/servers/src/http/loki.rs
```diff
@@ -106,213 +115,464 @@ pub async fn loki_ingest(
...
+            return self.next();
+        };
+        let Some(lines) = map.remove(LINES_KEY) else {
+            warn!("missing lines on stream, {:?}", map);
+            return self.next();
+        };
+        let serde_json::Value::Array(lines) = lines else {
+            warn!("lines is not an array, {:?}", lines);
+            return self.next();
+        };
```
killme2008: We could use a while loop instead of recursion in the above code.

src/servers/src/http/loki.rs
```diff
@@ -106,213 +115,464 @@ pub async fn loki_ingest(
...
+        );
+
+        if let Some(serde_json::Value::Object(m)) = structured_metadata {
+            for (k, v) in m {
+                match pipeline::Value::try_from(v) {
+                    Ok(v) => {
+                        map.insert(format!("{}{}", LOKI_PIPELINE_METADATA_PREFIX, k), v);
+                    }
+                    Err(e) => {
+                        warn!("not a valid value, {:?}", e);
```
killme2008, line 465: Print the value into the log.

src/servers/src/http/loki.rs
```diff
@@ -106,213 +115,464 @@ pub async fn loki_ingest(
...
+    type Item = LokiMiddleItem<Vec<LabelPairAdapter>>;
 
-            process_labels(&mut column_indexer, schemas, &mut row, labels.as_ref());
+    fn next(&mut self) -> Option<Self::Item> {
+        let entry = self.entries.pop_front()?;
 
-            rows.push(row);
+        let ts = if let Some(ts) = entry.timestamp {
+            ts
+        } else {
```
killme2008: Should we warn if the timestamp is missing?

src/servers/src/http/loki.rs
shuiyisong: The value is printed into the error already. Also the value is moved so we couldn't print it without an additional clone.

killme2008 (approved):
LGTM

Issue #6402, opened 2025-06-25
Index result cache may cache bloom query result after applying minmax stats


evenyag:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Storage Engine

### Minimal reproduce step

Creates a table with fulltext index

```sql
CREATE TABLE IF NOT EXISTS `test_cache` (
  `timestamp` TIMESTAMP(9) NOT NULL DEFAULT current_timestamp(),
  `message` STRING NULL NULL FULLTEXT INDEX WITH(analyzer = 'English', backend = 'bloom', case_sensitive = 'false'),
  TIME INDEX (`timestamp`),
)
ENGINE=mito
WITH(
  append_mode = 'true',
);
```

Inserts rows with unique timestamps and messages for more than 1 row group.

Queries with `matches_term` in the following step:
- Runs a query that only selects a row group (by timestamp) but doesn't match any message
- Runs a query with another time range and matches some messages
- Runs a query with the same time range as the second query but uses like.

### What did you expect to see?

The first query returns 0 rows, the second and third queries return the same number of rows.

### What did you see instead?

The second query also returns 0 rows.

### What operating system did you use?

Unrelated

### What version of GreptimeDB did you use?

0.14.4

### Relevant log output and stack trace

```bash
SELECT count(*) FROM public."test_cache" where timestamp >= '2025-06-23 22:00:00' and timestamp < '2025-06-23 22:00:09' and  matches_term(message, '<some-trace-id>');
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.05 sec)

SELECT count(*) FROM public."test_cache" where timestamp >= '2025-06-23 21:57:29' and timestamp < '2025-06-23 22:12:29' and  matches_term(message, '<some-trace-id>');
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.06 sec)

SELECT count(*) FROM public."test_cache" where timestamp >= '2025-06-23 21:57:29' and timestamp < '2025-06-23 22:12:29' and message like '%<some-trace-id>%';
+----------+
| count(*) |
+----------+
|       39 |
+----------+
1 row in set (0.77 sec)
```

PR #6403, opened 2025-06-25
fix: complete partial index search results in cache

zhongzc:

close https://github.com/GreptimeTeam/greptimedb/issues/6402

Previously, when using index caches, we only cached results for row groups that were actively searched, but marked unsearched row groups as missing from the cache. 
This caused subsequent queries to incorrectly assume these row groups had been searched and filtered out, potentially leading to incorrect query results.

This fix ensures all row groups are properly tracked in the cache by:

1. Only searching row groups that are required by the query and not already in cache
2. Merging new search results with existing cached results
3. Updating the cache with the complete set of searched row groups

This optimization prevents incorrect query results while still avoiding redundant index searches when accessing the same data with similar predicates.

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet.rs
+++ b/src/mito2/src/sst/parquet.rs
@@ -88,11 +88,12 @@ pub struct SstInfo {
 
 #[cfg(test)]
 mod tests {
+    use std::collections::HashSet;
     use std::sync::Arc;
 
     use common_time::Timestamp;
     use datafusion_common::{Column, ScalarValue};
-    use datafusion_expr::{BinaryExpr, Expr, Operator};
+    use datafusion_expr::{col, lit, BinaryExpr, Expr, Operator};
     use datatypes::arrow;
     use datatypes::arrow::array::RecordBatch;
     use datatypes::arrow::datatypes::{DataType, Field, Schema};
@@ -104,12 +105,17 @@ mod tests {
     use tokio_util::compat::FuturesAsyncWriteCompatExt;
 
     use super::*;
-    use crate::access_layer::{FilePathProvider, RegionFilePathFactory};
+    use crate::access_layer::{FilePathProvider, OperationType, RegionFilePathFactory};
     use crate::cache::{CacheManager, CacheStrategy, PageKey};
     use crate::read::BatchReader;
-    use crate::sst::index::{Indexer, IndexerBuilder};
+    use crate::region::options::{IndexOptions, InvertedIndexOptions};
+    use crate::sst::file::{FileHandle, FileMeta};
+    use crate::sst::file_purger::NoopFilePurger;
+    use crate::sst::index::bloom_filter::applier::BloomFilterIndexApplierBuilder;
+    use crate::sst::index::inverted_index::applier::builder::InvertedIndexApplierBuilder;
+    use crate::sst::index::{Indexer, IndexerBuilder, IndexerBuilderImpl};
     use crate::sst::parquet::format::WriteFormat;
-    use crate::sst::parquet::reader::ParquetReaderBuilder;
+    use crate::sst::parquet::reader::{ParquetReader, ParquetReaderBuilder, ReaderMetrics};
     use crate::sst::parquet::writer::ParquetWriter;
     use crate::sst::{location, DEFAULT_WRITE_CONCURRENCY};
     use crate::test_util::sst_util::{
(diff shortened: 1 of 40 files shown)


src/mito2/src/sst/parquet/reader.rs
```diff
@@ -425,14 +425,15 @@ impl ParquetReaderBuilder {
...
-            return true;
+        let cached = self
+            .cache_strategy
+            .index_result_cache()
+            .and_then(|cache| cache.get(predicate_key, self.file_handle.file_id()));
+        if let Some(result) = cached.as_ref() {
+            if all_required_row_groups_searched(output, result) {
+                apply_selection_and_update_metrics(output, result, metrics, INDEX_TYPE_FULLTEXT);
+                return true;
+            }
```
waynexia, line 446: If not all row groups are cached, can we reuse the cached part and only apply from file for those aren't cached?

zhongzc: For bloom filters, this can be done—just take note, that's how I modified it too.
But for inverted-like indexes, it's fundamentally impossible to achieve.

src/mito2/src/sst/parquet/row_selection.rs
```diff
@@ -191,7 +204,19 @@ impl RowGroupSelection {
...
+        // This is to indicate that the row groups are searched even if no rows are found.
+        for rg_id in 0..num_row_groups {
+            selection_in_rg
+                .entry(rg_id)
+                .or_insert(RowSelectionWithCount {
+                    selection: RowSelection::from(vec![]),
+                    row_count: 0,
+                    selector_len: 0,
+                });
+        }
```
evenyag: nitpick: We can wrap this into a helper function
```suggestion
        // Fill the missing row groups with empty selections.
        // This is to indicate that the row groups are searched even if no rows are found.
        fill_missing_row_groups(&mut selection_in_rg, num_row_groups);
```

src/mito2/src/sst/parquet/reader.rs
evenyag, line 1: nitpick: We may implement some tests similar to this filter test for different indices. It's hard to cover some of these cases in sqlness tests without inserting a large number of rows. We can use a smaller row group size in unit tests.

https://github.com/GreptimeTeam/greptimedb/blob/4737285275109506fdeb9948e574c124740e4d82/src/mito2/src/sst/parquet.rs#L319-L369

But we don't have to write all these tests in this PR.

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -498,9 +501,11 @@ impl ParquetReaderBuilder {
             .apply(self.file_handle.file_id(), Some(file_size_hint))
             .await;
         let selection = match apply_res {
-            Ok(output) => {
-                RowGroupSelection::from_inverted_index_apply_output(row_group_size, output)
-            }
+            Ok(output) => RowGroupSelection::from_inverted_index_apply_output(
+                row_group_size,
+                parquet_meta.num_row_groups(),
```
evenyag: non-blocking: We only use `num_row_groups`. Do we need to pass the whole `parquet_meta`?

src/mito2/src/sst/parquet/reader.rs
zhongzc: I'll try.

src/mito2/src/test_util.rs
```diff
@@ -218,35 +220,28 @@ impl Default for TestEnv {
...
 
     /// Returns a new env with specific `data_home` for test.
     pub fn with_data_home(data_home: TempDir) -> TestEnv {
         let (schema_metadata_manager, kv_backend) = mock_schema_metadata_manager();
+
+        let index_aux_path = data_home.path().join("index_aux");
+        let f = PuffinManagerFactory::new(&index_aux_path, 4096, None, None);
+        let puffin_manager = futures::executor::block_on(f).unwrap();
+        let f = IntermediateManager::init_fs(index_aux_path.to_str().unwrap());
+        let intermediate_manager = futures::executor::block_on(f).unwrap();
```
evenyag: Can we change this method (with_datahome) to async?

src/mito2/src/sst/parquet.rs
```diff
@@ -593,4 +599,300 @@ mod tests {
         }
         assert_eq!(total_rows, rows_read);
     }
+
+    #[tokio::test]
+    async fn test_write_read_with_index() {
```
evenyag, line 604: Great, can we also add tests for fulltext indices?

evenyag (approved):

src/mito2/src/test_util.rs
zhongzc: done

src/mito2/src/sst/parquet.rs
zhongzc: Will put it in the next PR

PR #6405, opened 2025-06-26
refactor: pass pipeline name through http header and get db from query context

zyy17:

1. The original implementation passes the pipeline name from the URL parameter, but it's not general if the collector (for example, Telegraf) doesn't support configuring it. This PR modifies the logic of extracting the pipeline name from the HTTP header.

2. Get DB from query context;

## PR Checklist

### DIFF
--- a/src/servers/src/elasticsearch.rs
+++ b/src/servers/src/elasticsearch.rs
@@ -39,6 +39,7 @@ use crate::http::event::{
     extract_pipeline_params_map_from_headers, ingest_logs_inner, LogIngesterQueryParams, LogState,
     PipelineIngestRequest,
 };
+use crate::http::header::constants::GREPTIME_PIPELINE_NAME_HEADER_NAME;
 use crate::metrics::{
     METRIC_ELASTICSEARCH_LOGS_DOCS_COUNT, METRIC_ELASTICSEARCH_LOGS_INGESTION_ELAPSED,
 };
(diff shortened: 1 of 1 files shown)


shuiyisong (approved):
Nit: do we need to support both? first check the header, if not exist,  then param?

sunng87 (approved):

src/servers/src/elasticsearch.rs
```diff
@@ -132,19 +133,18 @@ async fn do_handle_bulk_api(
     // The `schema` is already set in the query_ctx in auth process.
     query_ctx.set_channel(Channel::Elasticsearch);
 
-    let db = params.db.unwrap_or_else(|| "public".to_string());
+    let db = query_ctx.current_schema();
```
killme2008, line 136: We can support both query parameters and headers, giving preference to headers when both are present.

killme2008 (commented):
I don't like this breaking.

killme2008:
> Nit: do we need to support both? first check the header, if not exist, then param?

Agree

zyy17:
@killme2008 @shuiyisong I don't agree to support both configurations for the following reasons:

1. Using HTTP headers instead of URL parameters will be more general and easier to configure. The additional URL parameters are not natural for users who are already using ES ingestion APIs. It's also consistent with other GreptimeDB Logs APIs.
2. Although we can support both configuration styles, maintaining one style will be easier.

killme2008:
> @killme2008 @shuiyisong I don't agree to support both configurations for the following reasons:
> 
> 1. Using HTTP headers instead of URL parameters will be more general and easier to configure. The additional URL parameters are not natural for users who are already using ES ingestion APIs. It's also consistent with other GreptimeDB Logs APIs.
> 2. Although we can support both configuration styles, maintaining one style will be easier.

We already support both in Prometheus and other endpoints, so maintaining consistency is important.

sunng87:
The middleware which generates the db in `query_context` should already supported both. You can try it.

PR #6420, opened 2025-06-28
fix: add cfg for DecodeSqlValue error

fengys1996:

Running `cargo check -p meta-srv` will fail. This PR fixes it.

## PR Checklist

### DIFF
--- a/src/meta-srv/src/error.rs
+++ b/src/meta-srv/src/error.rs
@@ -374,6 +374,7 @@ pub enum Error {
         location: Location,
     },
 
+    #[cfg(feature = "mysql_kvbackend")]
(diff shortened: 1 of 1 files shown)


src/meta-srv/src/error.rs
```diff
@@ -374,6 +374,7 @@ pub enum Error {
         location: Location,
     },
 
+    #[cfg(feature = "mysql_kvbackend")]
```
sunng87, line 377: do we need this for pg backend?

fengys1996: It seems that only use for mysql rds now.

sunng87 (approved):

Issue #5971, opened 2025-04-24
System table for recording important events for region migration


zyy17:
### What type of enhancement is this?

Refactor

### What does the enhancement do?

The following reasons may trigger the region migration:

- failover
- manual region migration
- auto rebalance(for Enterprise feature)

Those events should be recorded as key system events for DB observability.

zyy17:
cc @WenyXu @evenyag

killme2008:
Already implemented in https://github.com/GreptimeTeam/greptimedb/pull/6501

PR #6413, opened 2025-06-27
refactor: add `RegionMigrationTriggerReason` in `RegionMigrationProcedureTask`

zyy17:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

Add `RegionMigrationTriggerReason` in `RegionMigrationProcedureTask`

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [x] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #6423, opened 2025-06-29
feat: introduce RegionMigrationEventRecorder

zyy17:

Resolve https://github.com/GreptimeTeam/greptimedb/issues/5971 and wait for https://github.com/GreptimeTeam/greptimedb/pull/6413 to be merged.

The PR introduces `RegionMigrationEventRecorder` to record every region migration and send it to the frontend to store as a mito system table. For example:

```
> SELECT * FROM greptime_private.region_migration_events;
+--------------------------------------+-------------------------+------------------+--------------------+----------------+------------------+----------------+------------+----------+----------------------------+
| procedure_id                         | region_id               | from_datanode_id | from_datanode_addr | to_datanode_id | to_datanode_addr | trigger_reason | timeout_ms | status   | timestamp                  |
+--------------------------------------+-------------------------+------------------+--------------------+----------------+------------------+----------------+------------+----------+----------------------------+
| 590a85f0-6723-4f56-b547-9cbff09d0292 | 4402341478413(1025, 13) |                0 | 0.0.0.0:13003      |              1 | 0.0.0.0:13004    | Manual         |     300000 | Starting | 2025-06-29 04:32:53.383000 |
| 590a85f0-6723-4f56-b547-9cbff09d0292 | 4402341478413(1025, 13) |                0 | 0.0.0.0:13003      |              1 | 0.0.0.0:13004    | Manual         |     300000 | Running  | 2025-06-29 04:32:53.387000 |
| 590a85f0-6723-4f56-b547-9cbff09d0292 | 4402341478413(1025, 13) |                0 | 0.0.0.0:13003      |              1 | 0.0.0.0:13004    | Manual         |     300000 | Finished | 2025-06-29 04:32:53.763000 |
+--------------------------------------+-------------------------+------------------+--------------------+----------------+------------------+----------------+------------+----------+----------------------------+
```

The following tasks need to be finished before being ready to review:

- [ ] Some integration tests;
- [ ] Batch the requests for performance;
- [ ] Error handling for database client;

## PR Checklist


src/meta-srv/src/procedure/region_migration/manager.rs
```diff
@@ -399,15 +403,38 @@ impl RegionMigrationManager {
...
         common_runtime::spawn_global(async move {
             let watcher = &mut match procedure_manager.submit(procedure_with_id).await {
                 Ok(watcher) => watcher,
                 Err(e) => {
                     error!(e; "Failed to submit region migration procedure {procedure_id} for {task}");
+                    record_region_migration_event(
+                        &event_recorder,
+                        task.clone(),
+                        procedure_id,
+                        RegionMigrationStatus::Failed(e),
```
WenyXu: How about introduce a status for failed to submit?

src/meta-srv/src/procedure/region_migration.rs
```diff
@@ -788,8 +788,8 @@ mod tests {
         }
     }
 
-    #[test]
-    fn test_data_serialization() {
+    #[tokio::test]
+    async fn test_data_serialization() {
```
WenyXu: unrelated change

src/meta-srv/src/procedure/region_migration.rs
```diff
@@ -770,8 +770,8 @@ mod tests {
         test_util::new_persistent_context(1, 2, RegionId::new(1024, 1))
     }
 
-    #[test]
-    fn test_lock_key() {
+    #[tokio::test]
+    async fn test_lock_key() {
```
WenyXu: unrelated change

src/meta-srv/src/event_recorder/region_migration_event.rs
```diff
@@ -0,0 +1,191 @@
...
+                        ValueData::TimestampNanosecondValue(self.timestamp).into(),
+                    ],
+                }],
+            }),
+        };
+
+        RowInsertRequests {
+            inserts: vec![insert],
+        }
+    }
```
WenyXu: What do you think about splitting it into separate methods `schema()`, `to_rows()`, and `table()`?

src/meta-srv/src/event_recorder.rs
```diff
@@ -0,0 +1,312 @@
...
+}
+
+// DefaultEventHandlerImpl is the default event handler implementation. It sends the received events to the frontend instances.
+struct DefaultEventHandlerImpl {
+    in_memory_key: ResettableKvBackendRef,
+}
+
+#[async_trait]
+impl EventHandler for DefaultEventHandlerImpl {
+    async fn handle(&self, event: Box<dyn Event>) -> Result<()> {
```
WenyXu: How about accepting events instead of a single event?

src/meta-srv/src/event_recorder.rs
```diff
@@ -0,0 +1,312 @@
...
+            .in_memory_key
+            .range(range_request)
+            .await
+            .context(KvBackendSnafu)?;
+
+        let mut urls = Vec::with_capacity(response.kvs.len());
+        for kv in response.kvs {
+            let node_info = NodeInfo::try_from(kv.value).context(KvBackendSnafu)?;
+            urls.push(node_info.peer.addr);
+        }
```
WenyXu: How about implementing this logic in `PeerLookupServiceRef`? The frontend address is only available in the memory store of the meta leader, so you'll need to use `MetaPeerClientRef`. Also, remember to filter out inactive nodes.

src/meta-srv/src/event_recorder.rs
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    async fn process(mut self) {
+        info!("Start the background processor in event recorder to handle the received events.");
+        while let Some(event) = self.rx.recv().await {
+            debug!("Received event: {:?}", event);
+            if let Err(e) = self.event_handler.handle(event).await {
+                error!(e; "Failed to handle event");
+            }
+        }
```
WenyXu: We can process events in batches, for example:
```suggestion
let mut interval = time::interval(Duration::from_secs(1));
interval.set_missed_tick_behavior(time::MissedTickBehavior::Delay);

let mut buffer = VecDeque::with_capacity(channel_size);

loop {
    tokio::select! {
        maybe_event = self.rx.recv() => {
            if let Some(event) = maybe_event {
                debug!("Buffered event: {:?}", event);
                buffer.push_back(event);
            } else {
                // handle close
                if !buffer.is_empty() {
                    if let Err(e) = self.event_handler.handle_batch(buffer.drain(..).collect()).await {
                        error!(e; "Failed to handle final batch of events");
                    }
                }
                break;
            }
        }
        _ = interval.tick() => {
            if !buffer.is_empty() {
                let batch: Vec<_> = buffer.drain(..).collect();
                debug!("Handling batch of {} events", batch.len());
                if let Err(e) = self.event_handler.handle_batch(batch).await {
                    error!(e; "Failed to handle batch of events");
                }
            }
        }
    }
}
```

src/meta-srv/src/event_recorder/region_migration_event.rs
```diff
@@ -0,0 +1,191 @@
...
+    Failed(Error),
+}
+
+impl Display for RegionMigrationStatus {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Starting => write!(f, "Starting"),
+            Self::Running => write!(f, "Running"),
+            Self::Finished => write!(f, "Finished"),
+            Self::Failed(e) => write!(f, "Failed: {}", e),
```
WenyXu: use debug fmt
```suggestion
            Self::Failed(e) => write!(f, "Failed: {:?}", e),
```

src/meta-srv/src/error.rs
```diff
@@ -953,6 +969,8 @@ impl ErrorExt for Error {
             | Error::BuildPartitionClient { .. }
             | Error::BuildKafkaClient { .. }
             | Error::DeleteRecords { .. }
+            | Error::InsertEvents { .. }
+            | Error::SendEvent { .. }
```
MichaelScofield: Better use their "source"s status.

src/meta-srv/src/event_recorder.rs
```diff
@@ -0,0 +1,312 @@
...
+
+// DefaultEventHandlerImpl is the default event handler implementation. It sends the received events to the frontend instances.
+struct DefaultEventHandlerImpl {
+    in_memory_key: ResettableKvBackendRef,
+}
+
+#[async_trait]
+impl EventHandler for DefaultEventHandlerImpl {
+    async fn handle(&self, event: Box<dyn Event>) -> Result<()> {
+        let database_client = self.build_database_client().await?;
```
MichaelScofield: The database client is better to be reused, creating one is not trivial.

src/meta-srv/src/event_recorder/region_migration_event.rs
```diff
@@ -0,0 +1,191 @@
...
+pub const REGION_MIGRATION_EVENTS_TABLE_STATUS_COLUMN_NAME: &str = "status";
+pub const REGION_MIGRATION_EVENTS_TABLE_TIMESTAMP_COLUMN_NAME: &str = "timestamp";
+
+/// RegionMigrationEvent is the event of region migration.
+#[derive(Debug)]
+pub struct RegionMigrationEvent {
+    pub(crate) task: RegionMigrationProcedureTask,
+    pub(crate) procedure_id: ProcedureId,
+    pub(crate) status: RegionMigrationStatus,
+    pub(crate) timestamp: i64,
```
MichaelScofield: Better use our `Timestamp` type in common-time. It has time unit, more readable than a plain `i64` here.

src/meta-srv/src/procedure/region_migration.rs
MichaelScofield: Why these two tests need to be async? There are no async functions calls inside.

zyy17: This is because `TestingEnv::new()` creates the `event_recorder`, which in turn spawns a tokio process within `event_recorder::new()`.

I will add the new function `new_default_context_factory()` to avoid it:

```rust
/// Generates a [DefaultContextFactory].
pub fn new_default_context_factory() -> DefaultContextFactory {
    let kv_backend = Arc::new(MemoryKvBackend::new());
    let table_metadata_manager = Arc::new(TableMetadataManager::new(kv_backend.clone()));
    let mailbox_sequence =
        SequenceBuilder::new("test_heartbeat_mailbox", kv_backend.clone()).build();
    let mailbox_ctx = MailboxContext::new(mailbox_sequence);
    let opening_region_keeper = Arc::new(MemoryRegionKeeper::default());
    let server_addr = "localhost".to_string();

    DefaultContextFactory {
        table_metadata_manager,
        opening_region_keeper,
        volatile_ctx: Default::default(),
        in_memory_key: Arc::new(MemoryKvBackend::default()),
        mailbox: mailbox_ctx.mailbox().clone(),
        server_addr: server_addr.clone(),
        region_failure_detector_controller: Arc::new(NoopRegionFailureDetectorControl),
        cache_invalidator: Arc::new(MetasrvCacheInvalidator::new(
            mailbox_ctx.mailbox().clone(),
            MetasrvInfo { server_addr },
        )),
    }
}
```

The two test cases only need `context` instead of the whole `TestingEnv`.

cc @WenyXu

src/meta-srv/src/event_recorder/region_migration_event.rs
zyy17: make sense.

src/meta-srv/src/event_recorder.rs
zyy17: Make sense. I add the new `frontends()` in `PeerLookupService`.

src/meta-srv/src/event_recorder.rs
zyy17: The advice will be the enhancement. Since the frontends may change due to operational events, it's simple and safe to create a client every time; however, it's indeed less efficient.

In my opinion, it is still acceptable because sending the events to the frontend is not a high-frequency behavior. By default, the event recorder sends events to the frontend every second in batches.

zyy17: @WenyXu @fengjiachun What do you think?

src/meta-srv/src/event_recorder/region_migration_event.rs
```diff
@@ -0,0 +1,201 @@
...
+            },
+        ]
+    }
+
+    fn to_row(&self) -> Row {
+        Row {
+            values: vec![
+                ValueData::StringValue(self.procedure_id.to_string()).into(),
+                ValueData::U64Value(self.task.region_id.as_u64()).into(),
+                ValueData::U32Value(self.task.region_id.table_id()).into(),
```
WenyXu: Let's also add region number.

src/meta-srv/src/event_recorder.rs
WenyXu: We can add a TODO here

src/meta-srv/src/event_recorder.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+}
+
+impl DefaultEventHandlerImpl {
+    async fn build_database_client(&self) -> Result<Database> {
+        let frontends = self
+            .peer_lookup_service
+            .frontends()
+            .await
+            .context(KvBackendSnafu)?;
```
WenyXu: The `frontends` method will returns outdated frontend address.

src/common/meta/src/peer.rs
```diff
@@ -24,6 +24,9 @@ use crate::{DatanodeId, FlownodeId};
 pub trait PeerLookupService {
     async fn datanode(&self, id: DatanodeId) -> Result<Option<Peer>, Error>;
     async fn flownode(&self, id: FlownodeId) -> Result<Option<Peer>, Error>;
+
+    /// Get all available frontends from the memory backend.
+    async fn frontends(&self) -> Result<Vec<Peer>, Error>;
 }
```
WenyXu: Let's add some comments as well.
```suggestion
/// A service that can query peer given a node id.
#[async_trait::async_trait]
pub trait PeerLookupService {
    /// Returns the peer of the datanode if the metasrv receives a heartbeat from the datanode.
    ///
    /// Note: It may return an inactive peer.
    async fn datanode(&self, id: DatanodeId) -> Result<Option<Peer>, Error>;

    /// Returns the peer of the flownode if the metasrv receives a heartbeat from the flownode.
    ///
    /// Note: It may return an inactive peer.
    async fn flownode(&self, id: FlownodeId) -> Result<Option<Peer>, Error>;

    /// Get all available frontends from the memory backend.
    async fn frontends(&self) -> Result<Vec<Peer>, Error>;
}
```

WenyXu: How about adding a flag to return the address of the active peer?

src/meta-srv/src/lease.rs
```diff
@@ -270,6 +295,12 @@ impl PeerLookupService for MetaPeerLookupService {
             .map_err(BoxedError::new)
             .context(common_meta::error::ExternalSnafu)
     }
+    async fn frontends(&self) -> common_meta::error::Result<Vec<Peer>> {
+        lookup_frontends(&self.meta_peer_client, u64::MAX)
+            .await
+            .map_err(BoxedError::new)
+            .context(common_meta::error::ExternalSnafu)
+    }
```
WenyXu: How about adding a flag to optionally return the active peer addresses, using `lookup_frontends(&self.meta_peer_client, u64::MAX, FRONTEND_HEARTBEAT_INTERVAL_MILLIS / 2000)`?

src/meta-srv/src/event_recorder.rs
```diff
@@ -0,0 +1,354 @@
...
+                        }
+                        break;
+                    }
+                }
+                // When the interval is triggered, flush the buffer and send the events to the event handler.
+                _ = interval.tick() => {
+                    if !buffer.is_empty() {
+                        debug!("Flushing buffer by trigger with {} events", buffer.len());
+                        if let Err(e) = self.event_handler.handle(std::mem::take(&mut buffer)).await {
+                            error!(e; "Failed to handle events");
```
WenyXu: How about buffering the events for retry if processing fails?

zyy17: Make sense. Let me add it later.

src/common/meta/src/peer.rs
zyy17: How about using the function name `active_frontends()`?

src/meta-srv/src/event_recorder.rs
MichaelScofield: How about caching the client first and recreate one if the requesting is failed on "bad route"?

src/meta-srv/src/event_recorder.rs
```diff
@@ -0,0 +1,355 @@
...
+            .into_iter()
+            .map(|peer| peer.addr)
+            .collect::<Vec<_>>();
+
+        debug!("Available frontend addresses: {:?}", urls);
+
+        Ok(Database::new(
+            DEFAULT_CATALOG_NAME,
+            DEFAULT_PRIVATE_SCHEMA_NAME,
+            Client::with_urls(urls),
```
fengjiachun: It's better to use `with_manager_and_urls` so we can share the `ChannelManager`.

src/meta-srv/src/event_recorder.rs
fengjiachun: If we share the `ChannelManager` instance, the `Database` instance should be lightweight.

src/meta-srv/src/event_recorder.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    // The default event handler implementation sends the received events to the frontend instances.
+    fn build_event_handler(&self) -> Box<dyn EventHandler> {
+        Box::new(DefaultEventHandlerImpl {
+            peer_lookup_service: self.peer_lookup_service.clone(),
+        })
+    }
+}
+
+impl EventRecorderImpl {
```
WenyXu: Better to implement drop trait for `EventRecorderImpl`.
```rust
impl Drop for EventRecorderImpl {
    fn drop(&mut self) {
        if let Some(handle) = self._handle.take() {
            handle.abort();
            info!("Event recorder background task stopped");
        }
    }
}
```

fengjiachun:
Others LGTM

src/common/meta/src/peer.rs
WenyXu: > How about using the function name active_fronends()?

It looks good to me

src/meta-srv/src/event_recorder.rs
zyy17: I add the fixed interval retry mechanism: https://github.com/GreptimeTeam/greptimedb/pull/6423/files#diff-df13ee08ce124870c4933329e6056fbce7976c9153888554d830422200cc76feR249-R278.

Also, I change to `DEFAULT_EVENTS_PROCESS_INTERVAL_SECONDS` from `1` to `5` to lower the frequency.

src/meta-srv/src/event_recorder.rs
zyy17: @fengjiachun, already add the `channel_manager` field in `DefaultEventHandlerImpl` and share it in `build_database_client()`(https://github.com/GreptimeTeam/greptimedb/pull/6423/files#diff-df13ee08ce124870c4933329e6056fbce7976c9153888554d830422200cc76feR188-R192).

src/meta-srv/src/event_recorder.rs
```diff
@@ -0,0 +1,394 @@
...
+}
+
+#[async_trait]
+impl EventHandler for DefaultEventHandlerImpl {
+    async fn handle(&self, events: &[Box<dyn Event>]) -> Result<()> {
+        debug!("Received {} events: {:?}", events.len(), events);
+        let database_client = self.build_database_client().await?;
+        let row_inserts = RowInsertRequests {
+            inserts: events.iter().map(|e| e.to_row_insert()).collect(),
+        };
```
WenyXu: nit: If the events have the same schema, it's more efficient to send them in one insert request.

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,383 @@
...
+
+    /// Add the extra schema to the event with the default schema.
+    fn extra_schema(&self) -> Vec<ColumnSchema> {
+        vec![]
+    }
+
+    /// Add the extra row to the event with the default row.
+    fn extra_row(&self) -> Result<Row> {
+        Ok(Row { values: vec![] })
+    }
```
WenyXu, line 85: Can we ensure that events of the same type always have the same `extra_schema`?
If so, we could use a single `InsertRequest` with multiple rows to reduce schema validation overhead.

src/meta-srv/src/events.rs
```diff
@@ -0,0 +1,99 @@
...
+impl EventHandler for EventHandlerImpl {
+    async fn handle(&self, events: &[Box<dyn Event>]) -> Result<()> {
+        debug!("Received {} events: {:?}", events.len(), events);
+        let database_client = self.build_database_client().await?;
+        let row_inserts = RowInsertRequests {
+            inserts: events
+                .iter()
+                .map(|e| build_row_insert_request(e.as_ref()))
+                .collect::<Result<Vec<_>>>()?,
+        };
```
WenyXu: It’s better to group events by `event_type` and merge all events with the same `event_type` into a single RowInsertRequest.

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,383 @@
...
+            schema,
+            rows: vec![row],
+        }),
+    })
+}
+
+/// EventRecorder trait defines the interface for recording events.
+pub trait EventRecorder: Send + Sync + 'static {
+    /// Records an event for persistence and processing by [EventHandler].
+    fn record(&self, event: Box<dyn Event>) -> Result<()>;
```
WenyXu: We can add a `stop` method to EventRecorder and call it before shutting down the server.

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,383 @@
...
+        let mut buffer = Vec::with_capacity(batch_size);
+        let mut interval = tokio::time::interval(self.process_interval);
+
+        loop {
+            tokio::select! {
+                maybe_event = self.rx.recv() => {
+                    if let Some(maybe_event) = maybe_event {
+                        debug!("Received event: {:?}", maybe_event);
+                        // Push the event to the buffer, the buffer will be flushed when the interval is triggered or received a closed signal.
+                        buffer.push(maybe_event);
```
WenyXu, line 324: Should we flush the buffer when it’s full?

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,383 @@
...
+                if let Err(e) = self.event_handler.handle(buffer).await {
+                    error!(
+                        e; "Failed to handle events, retrying... ({}/{})",
+                        retry, self.max_retry_times
+                    );
+                    retry += 1;
+                    sleep(Duration::from_millis(
+                        self.process_interval.as_millis() as u64 / self.max_retry_times,
+                    ))
+                    .await;
```
WenyXu, line 356: How about using `backon` crate to handle retry?

Cargo.toml
```diff
@@ -192,6 +193,7 @@ rust_decimal = "1.33"
 rustc-hash = "2.0"
 # It is worth noting that we should try to avoid using aws-lc-rs until it can be compiled on various platforms.
 rustls = { version = "0.23.25", default-features = false }
+sea-query = "0.32.6"
```
killme2008:
```suggestion
sea-query = "0.32"
```

Cargo.toml
```diff
@@ -192,6 +193,7 @@ rust_decimal = "1.33"
 rustc-hash = "2.0"
 # It is worth noting that we should try to avoid using aws-lc-rs until it can be compiled on various platforms.
 rustls = { version = "0.23.25", default-features = false }
```
killme2008:
```suggestion
rustls = { version = "0.23", default-features = false }
```

config/config.md
```diff
@@ -369,6 +369,12 @@
...
 | `wal.replication_factor` | Integer | `1` | Expected number of replicas of each partition. |
 | `wal.create_topic_timeout` | String | `30s` | Above which a topic creation operation will be cancelled. |
+| `event_recorder` | -- | -- | Configuration options for the event recorder. |
+| `event_recorder.channel_size` | Integer | `2048` | The capacity of the tokio channel for transmitting events to background processor. |
+| `event_recorder.buffer_size` | Integer | `100` | The size of the buffer for batching events before flushing to event handler. |
+| `event_recorder.flush_interval` | String | `5s` | The time interval for flushing batched events to the event handler. |
+| `event_recorder.max_retry_times` | Integer | `3` | The maximum number of retry attempts when event handler processing fails. |
+| `event_recorder.ttl` | String | `30d` | TTL for the events table that will be used to store the events. |
```
killme2008: The user only needs to be aware of the `ttl` parameter; most other parameters are irrelevant to them. We should follow the CoC principle.

src/client/src/database.rs
```diff
@@ -172,6 +172,26 @@ impl Database {
         from_grpc_response(response)
     }
 
+    pub async fn row_inserts(&self, requests: RowInsertRequests) -> Result<u32> {
```
killme2008, line 182: Document all public functions please.

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,395 @@
...
+    fn extra_row(&self) -> Result<Row> {
+        Ok(Row { values: vec![] })
+    }
+
+    /// Returns the event as any type.
+    fn as_any(&self) -> &dyn Any;
+}
+
+/// Returns the hints for the insert operation.
+pub fn insert_hints() -> Vec<(&'static str, &'static str)> {
```
killme2008, line 92: Doesn't need to pub?
```suggestion
fn insert_hints() -> Vec<(&'static str, &'static str)> {
```

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,395 @@
...
+        // Spawn a background task to process the events.
+        let handle = tokio::spawn(async move {
+            processor.process(opts.buffer_size).await;
+        });
+
+        recorder.handle = Some(handle);
+
+        // It only sets the ttl once, so it's safe to skip the error.
+        if EVENTS_TABLE_TTL.set(opts.ttl).is_err() {
+            info!("Events table ttl already set, skip setting it");
```
killme2008: Better to log the current value.

src/common/event-recorder/src/recorder.rs
killme2008: There are many potential callees for the event recorder, making it unclear which one should call `stop`. Therefore, I think it's unnecessary—the recorder should manage this itself.

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,395 @@
...
+                        e; "Failed to handle events, retrying... ({}/{})",
+                        retry, self.max_retry_times
+                    );
+                    retry += 1;
+                    sleep(Duration::from_millis(
+                        self.process_interval.as_millis() as u64 / self.max_retry_times,
+                    ))
+                    .await;
+                } else {
+                    info!("Successfully handled {} events", buffer.len());
```
killme2008: The log is annoying. Please change it to `debug`.

killme2008:
```suggestion
                    debug!("Successfully handled {} events", buffer.len());
```

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,395 @@
...
+                    break;
+                }
+
+                if retry > self.max_retry_times {
+                    error!(
```
killme2008: I think warning is enough.

killme2008:
```suggestion
                    warn!(
```

src/meta-srv/src/events.rs
```diff
@@ -0,0 +1,99 @@
...
+        let frontends = self
+            .peer_lookup_service
+            .active_frontends()
+            .await
+            .context(KvBackendSnafu)?;
+
+        if frontends.is_empty() {
+            return NoAvailableFrontendSnafu.fail();
+        }
+
```
killme2008: Use snafu::ensure
```suggestion
        enesure!(!frontends.is_empty(), NoAvailableFrontendSnafu);

```

src/meta-srv/src/lease.rs
```diff
@@ -203,6 +206,29 @@ pub async fn lookup_flownode_peer(
...
+) -> Result<Vec<Peer>> {
+    let range_request =
+        RangeRequest::new().with_prefix(NodeInfoKey::key_prefix_with_role(ClusterRole::Frontend));
+
+    let response = meta_peer_client.range(range_request).await?;
+
+    let mut peers = Vec::with_capacity(response.kvs.len());
+    for kv in response.kvs {
+        let node_info = NodeInfo::try_from(kv.value).context(KvBackendSnafu)?;
+        let is_alive = ((time_util::current_time_millis() - node_info.last_activity_ts) as u64)
```
killme2008: Refactor a function `is_alive` to `NodeInfo` struct and call it directly.

src/meta-srv/src/lease.rs
```diff
@@ -270,6 +296,17 @@ impl PeerLookupService for MetaPeerLookupService {
             .map_err(BoxedError::new)
             .context(common_meta::error::ExternalSnafu)
     }
+    async fn active_frontends(&self) -> common_meta::error::Result<Vec<Peer>> {
+        // Get the active frontends within the last heartbeat interval.
+        lookup_frontends(
+            &self.meta_peer_client,
+            // TODO(zyy17): How to get the heartbeat interval of the frontend if it uses a custom heartbeat interval?
```
killme2008, line 312: @fengjiachun @WenyXu Does metasrv know the configured heartbeat  interval value?

src/meta-srv/src/metasrv/builder.rs
```diff
@@ -322,6 +331,7 @@ impl MetasrvBuilder {
                 options.grpc.server_addr.clone(),
                 cache_invalidator.clone(),
             ),
+            event_recorder.clone(),
```
killme2008:
```suggestion
            event_recorder,
```

src/meta-srv/src/procedure/region_migration/manager.rs
```diff
@@ -418,16 +450,48 @@ impl RegionMigrationManager {
...
 }
 
+fn record_region_migration_event(
+    event_recorder: &EventRecorderRef,
+    task: RegionMigrationProcedureTask,
+    procedure_id: ProcedureId,
+    status: RegionMigrationStatus,
+    error: Option<Error>,
+) {
+    if let Err(e) = event_recorder.record(Box::new(RegionMigrationEvent::new(
```
killme2008: We can have `record` return nothing and handle logging within the recorder itself. The logic remains the same as when `record` returns an error—we simply log the error.

tests-integration/src/cluster.rs
```diff
@@ -394,33 +400,45 @@ impl GreptimeDbClusterBuilder {
             Arc::new(InvalidateCacheHandler::new(cache_registry.clone())),
         ]);
 
-        let options = FrontendOptions::default();
+        let mut fe_opts = FrontendOptions::default();
+        // Setup the server addr for local test.
+        fe_opts.grpc.server_addr = format!("127.0.0.1:{}", DEFAULT_GRPC_ADDR_PORT);
```
killme2008, line 405: Why configure this? It's the default value. https://github.com/GreptimeTeam/greptimedb/blob/29739b556e6daec6776999b0ef8e93a7b026d3d8/src/servers/src/grpc.rs#L124

tests-integration/tests/region_migration.rs
```diff
@@ -1226,3 +1273,76 @@ async fn run_sql(
...
+    }
+}
+
+async fn check_region_migration_events_system_table(
+    fe_instance: &Arc<Instance>,
+    procedure_id: &str,
+    region_id: u64,
+) {
+    // Sleep for while to ensure the event is recorded.
+    tokio::time::sleep(Duration::from_secs(5)).await;
```
killme2008: It's better to sleep more than 5 seconds -- the default flushing interval.

src/common/event-recorder/src/recorder.rs
zyy17: There may be multiple events in one handler in the future, so it may not be easy to aggregate the same events. However, it can be a performance enhancement.

src/common/event-recorder/src/recorder.rs
zyy17: Yes. It should be public because of https://github.com/GreptimeTeam/greptimedb/pull/6423/files#diff-49754ac5c614313905ff4fe65c4a6b6d1e2df8ae4234803b29c28e944ea3489fR55.

tests-integration/src/cluster.rs
zyy17: If we don't configure the `server_addr`, it will use the local IP of the default network device, which does not work in the integration test. We have to specify localhost.

The `server_addr` and `bind_addr` are not the same by default.

src/meta-srv/src/procedure/region_migration/manager.rs
zyy17: make sense.

src/common/event-recorder/src/recorder.rs
zyy17: Prefer to add it for the next PR.

src/common/event-recorder/src/recorder.rs
zyy17: Refactor `build_row_inserts_request()` to support it: https://github.com/GreptimeTeam/greptimedb/pull/6423/files#diff-c297317fbd8fa14f06599204d7d1f7a414bde144d115ed4338dd487b7642ba36R106-R173.

src/common/event-recorder/src/recorder.rs
WenyXu: I see. Maybe we shouldn't add a stop method here. However, I still think `EventRecorderImpl` should hold a `CancellationToken` to ensure it can still attempt to record events when metasrv is shutting down. e.g., during an upgrade.

src/meta-srv/src/lease.rs
WenyXu: No

zyy17: @WenyXu Can we use the report hearbeat interval in [request](https://github.com/GreptimeTeam/greptime-proto/blob/main/proto/greptime/v1/meta/heartbeat.proto#L40)?

WenyXu: Maybe it's better to store all these heartbeat intervals in the metasrv. During startup, components can fetch these heartbeat configuration from metasrv instead of relying on a static configuration file. (We can track this in a separate issue.) cc @fengjiachun

zyy17:
After discussing with @WenyXu, we decided to split the huge PR into multiple small PRs, which are easier to review and test:

1. PR1: Introduce the common event recorder(Only contains `src/common/event-recorder`): https://github.com/GreptimeTeam/greptimedb/pull/6501.
2. PR3: Add row_inserts() and row_inserts_with_hints(): https://github.com/GreptimeTeam/greptimedb/pull/6503.
3. PR3: Add the `active_frontends()` in `PeerLookupService`:https://github.com/GreptimeTeam/greptimedb/pull/6504
4. PR4: Record region migration events by using the event recorder: https://github.com/GreptimeTeam/greptimedb/pull/6579.

cc @fengjiachun @killme2008 @MichaelScofield

PR #6430, opened 2025-07-01
refactor: replace `pipeline::value` with `vrl::value`

shuiyisong:

In this PR, we replace the original handmade `pipeline::Value` with `vrl::value::Value`. This allows us to 
1. Remove the `serde_json::Value` while parsing the JSON input
2. Remove the transform between `pipeline::Value` and `vrl::value::Value` while using the vrl script
3. Cut down the boilerplate code by a large amount

The downside might be the frequent `Bytes` to `String` conversion. We'll run a load test later to see if there's a performance benefit.

## PR Checklist


paomian:
tl

src/pipeline/src/etl/ctx_req.rs
```diff
@@ -86,32 +86,34 @@ impl ContextOpt {
...
-    pub fn from_pipeline_map_to_opt(pipeline_map: &mut Value) -> Result<Self> {
-        let pipeline_map = pipeline_map.as_map_mut().context(ValueMustBeMapSnafu)?;
+    pub fn from_pipeline_map_to_opt(value: &mut VrlValue) -> Result<Self> {
+        let map = value.as_object_mut().context(ValueMustBeMapSnafu)?;
+
         let mut opt = Self::default();
         for k in PIPELINE_HINT_KEYS {
-            if let Some(v) = pipeline_map.remove(k) {
+            if let Some(v) = map.remove(k) {
+                let v = v.to_string_lossy().to_string();
```
paomian, line 95: should we check v is not array or map？

shuiyisong: I don't think it's necessary: the hint value is ultimately used by the engine, `v` is an array or a map doesn't affect the validity of the value.

src/pipeline/src/etl/processor/date.rs
```diff
@@ -225,20 +228,16 @@ impl Processor for DateProcessor {
 }
 
 /// try to parse val with timezone first, if failed, parse without timezone
```
killme2008: Looks like the comment is not correct, it always try to parse value into utc datetime first.

src/servers/src/proto.rs
```diff
@@ -379,25 +382,33 @@ impl PromSeriesProcessor {
...
             let value = prom_validation_mode.decode_string(&l.value)?;
-            pipeline_map.insert(name, Value::String(value));
+            pipeline_map.insert(KeyString::from(name), VrlValue::Bytes(value.into()));
         }
 
         let one_sample = series.samples.len() == 1;
 
         for s in series.samples.iter() {
+            let Ok(value) = NotNan::new(s.value) else {
+                warn!("invalid float value: {}", s.value);
```
killme2008:
```suggestion
                warn!("Invalid float value: {}", s.value);
```
It's bettero to print the pipeline name.

paomian:
Surprisingly, the value type in VRL does not include a string type.

src/pipeline/src/etl/processor/gsub.rs
```diff
@@ -117,11 +119,12 @@ impl crate::etl::processor::Processor for GsubProcessor {
         self.ignore_missing
     }
 
-    fn exec_mut(&self, mut val: Value) -> Result<Value> {
+    fn exec_mut(&self, mut val: VrlValue) -> Result<VrlValue> {
         for field in self.fields.iter() {
             let index = field.input_field();
+            let val = val.as_object_mut().context(ValueMustBeMapSnafu)?;
```
sunng87, line 125: We need to check for this type of error, do we provide enough context information like which *value*. We need to avoid giving user a plain error message without context, the context of their yaml input

sunng87: This can be a bigger topic and if it's complex enough, we can address it in next patch

src/servers/src/otlp/logs.rs
```diff
@@ -652,43 +658,41 @@ fn parse_export_logs_service_request(request: ExportLogsServiceRequest) -> Vec<V
...
-                Some(value) => any_value_to_pipeline_value(value),
-                None => Value::Null,
+                Some(value) => any_value_to_vrl_value(value),
+                None => VrlValue::Null,
             },
-            None => Value::Null,
+            None => VrlValue::Null,
         };
-        map.insert(kv.key.clone(), value);
+        if value != VrlValue::Null {
```
paomian: Do we need this check?

src/servers/src/http/loki.rs
```diff
@@ -325,13 +326,15 @@ impl Iterator for LokiJsonParser {
...
-                    serde_json::Value::Object(labels) => Some(labels),
+                    VrlValue::Object(labels) => Some(labels),
                     _ => None,
                 })
                 .map(|m| {
                     m.into_iter()
                         .filter_map(|(k, v)| match v {
-                            serde_json::Value::String(v) => Some((k, v)),
+                            VrlValue::Bytes(v) => {
+                                Some((k.into(), String::from_utf8_lossy(&v).to_string()))
```
paomian, line 336: can we use [from_utf8_lossy_owned](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy_owned)
`String::from_utf8_lossy_owned(v.to_vec())`

src/servers/src/http/loki.rs
```diff
@@ -413,13 +416,16 @@ impl From<LokiMiddleItem<serde_json::Value>> for LokiRawItem {
...
+                VrlValue::Object(m) => Some(m),
                 _ => None,
             })
             .map(|m| {
                 m.into_iter()
                     .filter_map(|(k, v)| match v {
-                        serde_json::Value::String(v) => Some((k, Value::String(v.into()))),
+                        VrlValue::Bytes(bytes) => Some((
+                            k.into(),
+                            Value::String(String::from_utf8_lossy(&bytes).to_string().into()),
```
paomian, line 427: ibid

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -484,76 +448,59 @@ fn resolve_value(
...
-                ValueData::TimestampMicrosecondValue(us),
+                ValueData::StringValue(String::from_utf8_lossy(&v).to_string()),
                 column_name,
-                ColumnDataType::TimestampMicrosecond,
+                ColumnDataType::String,
             )?;
         }
 
-        Value::Timestamp(Timestamp::Millisecond(ms)) => {
+        VrlValue::Regex(v) => {
```
paomian, line 484: It would be good to add some logs here, I don't think this should be a reasonable branch.

src/servers/src/otlp/logs.rs
shuiyisong: This happens if the input key-value pair has the key, but the value itself is `None`. It's ok to skip these keys.

src/pipeline/src/etl/processor/gsub.rs
shuiyisong: In this specific case, the error should be unreachable, since we check and wrap the input data as an object.
It's because we use the VrlValue struct directly, so we have to convert it to an object before using it as the context.

src/servers/src/otlp/logs.rs
paomian: I think we may have to keep these keys, which may provide some help when debugging.

src/pipeline/src/etl/processor/gsub.rs
paomian: Maybe in the future we can use ObjectMap as a data container for the Pipeline.

paomian:
other LGTM

src/pipeline/src/etl/processor/gsub.rs
sunng87: This is just an instance, we need to confirm any error that throws from processor function should carry a context for better debug-ability

sunng87: This is non-blocking. We can address this later.

PR #6431, opened 2025-07-01
feat!: point matrix based partition rule checker

waynexia:

follow up of #6399 

Implement a new point matrix based rule checker to replace the previous AxisRuleChecker. The new checker is more robust. It can recognize bad cases can't be handled before, and won't have false negatives for correct cases.

This new checker will check for both completeness and uniqueness. I.e., we don't need a mechanism like "default region" before as the rule will guarantee all values has a declared region. And we also don't rely on the rule apply order as all region won't non-overlapping with others.

Moreover, the error message is also improved. It will give the uncovered or overlapped value range for invalid rule. E.g.:
```
Error: 1004(InvalidArguments), Checkpoint `a=10` is not covered

Error: 1004(InvalidArguments), Checkpoint `b=42` is overlapped
```

## PR Checklist

### DIFF
--- a/src/partition/src/checker.rs
+++ b/src/partition/src/checker.rs
@@ -12,8 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::{BTreeMap, BTreeSet, HashMap};
-use std::ops::Deref;
+use std::collections::{BTreeMap, HashMap};
 use std::sync::Arc;
 
 use datatypes::arrow::array::{BooleanArray, Float64Builder, RecordBatch};
@@ -74,7 +73,7 @@ impl<'a> PartitionChecker<'a> {
         if matrix_fundation.is_empty() {
             todo!("error: matrix fundation is empty");
         }
-        let matrix_generator = MatrixGenerator { matrix_fundation };
+        let matrix_generator = MatrixGenerator::new(matrix_fundation);
         let batch = matrix_generator.all_points();
         let mut results = Vec::with_capacity(self.collider.atomic_exprs.len());
         for expr in self.collider.atomic_exprs.iter() {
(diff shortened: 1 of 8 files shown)


src/partition/src/checker.rs
```diff
@@ -0,0 +1,571 @@
...
+}
+
+// Logic of checking rules
+impl<'a> PartitionChecker<'a> {
+    fn run(&self) -> Result<()> {
+        // Sort atomic exprs and check uniqueness
+        let mut atomic_exprs = BTreeMap::new();
+        for expr in self.collider.atomic_exprs.iter() {
+            let key = &expr.nucleons;
+            atomic_exprs.insert(key, expr);
```
discord9, line 53: maybe just dedup right here instead?

waynexia: on the correct case we don't need to find which is duplicated. So I put the logic in error path to avoid affecting normal case

discord9:
Other than this two corner case(which should be documented to not recommend to use anyway), this checker LGTM:
```sql
CREATE TABLE tricky_float_partition (
  f FLOAT PRIMARY KEY,
  ts TIMESTAMP TIME INDEX,
)
PARTITION ON COLUMNS (f) (
  f < 10.0,
  f >= 10.000000000000002
);
ok 0 # which it shouldn't pass check
```
and
```sql
CREATE TABLE invalid_rule (
  a INT PRIMARY KEY,
  b STRING,
  ts TIMESTAMP TIME INDEX,
)
PARTITION ON COLUMNS (a) (
  a <= 10,
  a >= 11,
);
ERROR:  Checkpoint `10<a<11` is not covered #which should pass check as a is int
```

v0y4g3r:
bugbot run

v0y4g3r:
bugbot run

src/partition/src/collider.rs
```diff
@@ -114,6 +164,10 @@ impl<'a> Collider<'a> {
         for (index, expr) in source_exprs.iter().enumerate() {
             Self::collide_expr(expr, index, &normalized_values, &mut atomic_exprs)?;
         }
+        // sort nucleon exprs
+        for expr in &mut atomic_exprs {
+            expr.nucleons.sort_unstable();
+        }
```
evenyag, line 170: Why do we sort the exprs here?

src/partition/src/multi_dim.rs
```diff
@@ -84,12 +82,20 @@ impl MultiDimPartitionRule {
             physical_expr_cache: RwLock::new(None),
         };
 
-        let mut checker = RuleChecker::new(&rule);
+        // let mut checker = RuleChecker::new(&rule);
+        // #[allow(deprecated)]
+        // checker.check()?;
+
```
evenyag:
```suggestion
```

src/partition/src/checker.rs
```diff
@@ -0,0 +1,594 @@
...
+                // Fallback if column not found in normalized values
+                check_point.push_str(&format!("{}:unknown", col_name));
+            }
+        }
+
+        check_point
+    }
+}
+
+/// Generates a point matrix that contains premutations of `matrix_foundation`'s values
```
evenyag: Did you mean permutation?
```suggestion
/// Generates a point matrix that contains permutations of `matrix_foundation`'s values
```

src/partition/src/checker.rs
```diff
@@ -0,0 +1,594 @@
...
+        }
+
+        let matrix_generator = MatrixGenerator::new(matrix_foundation);
+
+        // Process data in batches using iterator
+        let mut results = Vec::with_capacity(self.collider.atomic_exprs.len());
+        for batch in matrix_generator {
+            results.clear();
+            for expr in self.collider.atomic_exprs.iter() {
+                let physical_expr = expr.to_physical_expr(&batch.schema());
```
evenyag: non-blocking: We can create physical exprs outside the loop first.

src/partition/src/checker.rs
```diff
@@ -0,0 +1,594 @@
...
+impl<'a> PartitionChecker<'a> {
+    pub fn try_new(rule: &'a MultiDimPartitionRule) -> Result<Self> {
+        let collider = Collider::new(rule.exprs())?;
+        Ok(Self { rule, collider })
+    }
+
+    pub fn check(&self) -> Result<()> {
+        self.run()?;
+        Ok(())
+    }
```
evenyag, line 43: non-blocking: We may benchmark the check performance under a large number of partition expression combinations.

evenyag (approved):
Most comments are non-blocking.

src/partition/src/checker.rs
waynexia: will do in the following PR(s)

src/partition/src/collider.rs
waynexia: from its comment, this field is ordered. this is for follow-up steps (in todo) of merging exprs

evenyag:
This may break existing partition rules.

PR #6421, opened 2025-06-28
feat: add `SET DEFAULT` syntax

linyihai:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)
https://github.com/GreptimeTeam/greptimedb/issues/6068
## What's changed and what's your intention?

This adds `SET DEFAULT` syntax. For example, "ALTER TABLE test ALTER a SET DEFAULT 1, ALTER b SET DEFAULT 2". 
`SET DEFAULT` doesn't know the concreate datatype of the column,  so it can't infer the correct `default-contraint` until it pass to the `table` crate, where we can get the column datatype from the `meta`.

Split `parse_column_default_constraint` into `sql-common` crate, as `table` and `sql` crate all need it. Previous `sql` crate holds the `parse_column_default_constraint`, `table` crate need depends on `sql`  crate, but `sql` also depends on `table` crate, which causes the cyclic dependence. 
 

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [x] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #6452, opened 2025-07-03
refactor: split some convert function into `common-sql` crate

linyihai:

This is refactor PR needed for https://github.com/GreptimeTeam/greptimedb/pull/6421

moving `parse_column_default_constraint ` into a sperate crate then other crate can also use it.  This is my intetion: 

```
This adds SET DEFAULT syntax. For example, "ALTER TABLE test ALTER a SET DEFAULT 1, ALTER b SET DEFAULT 2".
SET DEFAULT doesn't know the concreate datatype of the column, so it can't infer the correct default-contraint until it pass to the table crate, where we can get the column datatype from the meta.

Split parse_column_default_constraint into sql-common crate, as table and sql crate all need it. Previous sql crate holds the parse_column_default_constraint, table crate need depends on sql crate, but sql also depends on table crate, which causes the cyclic dependence.
```

## PR Checklist


src/operator/src/error.rs
```diff
@@ -851,6 +851,13 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
+
+    #[snafu(display("SQLCommon error: {}", source))]
+    SQLCommon {
```
evenyag: We don't print the source error in the message. We have a macro to generate error messages later.
```suggestion
    #[snafu(display("Sql common error"))]
    SqlCommon {
```

Cargo.toml
```diff
@@ -294,6 +295,7 @@ query = { path = "src/query" }
 servers = { path = "src/servers" }
 session = { path = "src/session" }
 sql = { path = "src/sql" }
+sql-common = { path = "src/sql-common" }
```
evenyag: We have a common directory. I think we can move the sql-common crate to `common/sql` and rename it to `common-sql`.

You can refer to the `common-grpc` crate
https://github.com/GreptimeTeam/greptimedb/blob/main/src/common/grpc/Cargo.toml

linyihai: Had moved to common/sql

src/operator/src/error.rs
linyihai: Done.

src/servers/Cargo.toml
```diff
@@ -89,6 +89,7 @@ opentelemetry-proto.workspace = true
 otel-arrow-rust.workspace = true
 parking_lot.workspace = true
 #pgwire = { version = "0.30", default-features = false, features = ["server-api-ring"] }
+common-sql.workspace = true
```
evenyag: We should move `common-sql` to L50 to keep dependencies sorted.

src/operator/src/error.rs
```diff
@@ -851,6 +851,13 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
+
+    #[snafu(display("Sql common error"))]
+    SqlCommon {
+        source: common_sql::error::Error,
+        #[snafu(implicit)]
+        location: Location,
+    },
```
evenyag, line 860: Should we use `#[snafu(transparent)]` here? @MichaelScofield

MichaelScofield: We may, it makes codes clearer. But it's not mandated.

killme2008 (approved):
LGTM

Issue #1915, opened 2023-07-10
Support TQL in CTE position


waynexia:
### What problem does the new feature solve?

Support SQL like 
```sql
WITH prom_result AS (
    TQL EVAL (0, 100, '10s') sum(rate(http_requests_total[5m])) BY (job)
)
SELECT * FROM prom_result
ORDER BY sum_rate_http_requests_total_5m
LIMIT 10
```

### What does the feature do?

Currently TQL and SQL are still very dissociated. Interoperability between independent components such as CTE is a great starting point.

etolbakov:
@waynexia I would like to give it a go if you don't mind

killme2008:
Already supported in v0.17 and 1.0 beta.1

PR #6456, opened 2025-07-03
feat(parser): parse TQL in CTE position

waynexia:

part of https://github.com/GreptimeTeam/greptimedb/issues/1915

Supports queries like 
```sql
WITH 
    sql_cte(ts, value, label) AS (SELECT timestamp, val, name FROM metrics),
    tql_cte(time, metric_value) AS (TQL EVAL (0, 100, '5s') cpu_usage)
SELECT s.ts, s.value, t.metric_value 
FROM sql_cte s JOIN tql_cte t ON s.ts = t.time
```

Before this patch, such a query would get a parse error:
```
ERROR 1149 (42000): (InvalidSyntax): Invalid SQL syntax: sql parser error: Expected: SELECT, VALUES, or a subquery in the query body, found: TQL at Line: 1, Column: 83
```

## PR Checklist

### DIFF
--- a/src/sql/src/parsers/with_tql_parser.rs
+++ b/src/sql/src/parsers/with_tql_parser.rs
@@ -16,7 +16,10 @@ use std::fmt;
 
 use serde::Serialize;
 use snafu::ResultExt;
-use sqlparser::ast::{Ident, ObjectName, Query as SpQuery};
+use sqlparser::ast::helpers::attached_token::AttachedToken;
+use sqlparser::ast::{
+    Cte, Ident, ObjectName, Query as SpQuery, TableAlias, TableAliasColumnDef, With,
+};
 use sqlparser::keywords::Keyword;
 use sqlparser::parser::IsOptional;
 use sqlparser::tokenizer::Token;
@@ -102,11 +105,31 @@ impl ParserContext<'_> {
         let recursive = self.parser.parse_keyword(Keyword::RECURSIVE);
 
         // Parse the CTE list
-        let mut cte_tables = Vec::new();
+        let mut tql_cte_tables = Vec::new();
+        let mut sql_cte_tables = Vec::new();
 
         loop {
             let cte = self.parse_hybrid_cte()?;
-            cte_tables.push(cte);
+            match cte.content {
+                CteContent::Sql(body) => sql_cte_tables.push(Cte {
+                    alias: TableAlias {
+                        name: cte.name,
+                        columns: cte
+                            .columns
+                            .into_iter()
+                            .map(|col| TableAliasColumnDef {
+                                name: col.0[0].clone(),
+                                data_type: None,
+                            })
+                            .collect(),
+                    },
+                    query: body,
+                    from: None,
+                    materialized: None,
+                    closing_paren_token: AttachedToken::empty(),
+                }),
+                CteContent::Tql(_) => tql_cte_tables.push(cte),
+            }
 
             if !self.parser.consume_token(&Token::Comma) {
                 break;
@@ -119,12 +142,17 @@ impl ParserContext<'_> {
         // Convert the hybrid CTEs to a standard query with hybrid metadata
         let hybrid_cte = HybridCteWith {
             recursive,
-            cte_tables,
+            cte_tables: tql_cte_tables,
         };
 
         // Create a Query statement with hybrid CTE metadata
         let mut query = Query::try_from(*main_query)?;
         query.hybrid_cte = Some(hybrid_cte);
+        query.inner.with = Some(With {
+            recursive,
+            cte_tables: sql_cte_tables,
+            with_token: AttachedToken::empty(),
+        });
 
         Ok(Statement::Query(Box::new(query)))
     }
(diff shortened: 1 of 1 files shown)


src/sql/src/parsers/with_tql_parser.rs
```diff
@@ -0,0 +1,551 @@
...
+    pub(crate) fn contains_tql_in_with(&mut self) -> Result<bool> {
+        // Save the current parser position
+        let checkpoint = self.parser.index();
+
+        // Perform lookahead parsing to detect TQL
+        let has_tql = self.scan_for_tql_in_ctes().unwrap_or(false);
+
+        // Restore parser position by creating a new parser with the original SQL
+        // and advancing to the checkpoint position
+        let mut new_parser = sqlparser::parser::Parser::new(&sqlparser::dialect::GenericDialect {})
```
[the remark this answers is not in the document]
waynexia: CTEs are usually located at the beginning, so the overhead is minimal.

src/sql/src/parsers/with_tql_parser.rs
```diff
@@ -0,0 +1,550 @@
...
+    fn parse_tql_query_in_cte(&mut self) -> Result<String> {
+        // Skip any whitespace or commas
+        while matches!(
+            self.parser.peek_token().token,
+            Token::Comma | Token::Whitespace(_)
+        ) {
+            self.parser.next_token();
+        }
+
+        // Get the starting position for extracting the query text
```
killme2008: Could we reuse the `tql_parser` code here? https://github.com/GreptimeTeam/greptimedb/blob/c2f1447345901754ab003bed7964f3f4d22ddc90/src/sql/src/parsers/tql_parser.rs#L100C8-L100C24

src/sql/src/parsers/with_tql_parser.rs
```diff
@@ -0,0 +1,550 @@
...
+    pub(crate) fn contains_tql_in_with(&mut self) -> Result<bool> {
+        // Save the current parser position
+        let checkpoint = self.parser.index();
+
+        // Perform lookahead parsing to detect TQL
+        let has_tql = self.scan_for_tql_in_ctes().unwrap_or(false);
+
+        // Restore parser position by creating a new parser with the original SQL
+        // and advancing to the checkpoint position
+        let mut new_parser = sqlparser::parser::Parser::new(&sqlparser::dialect::GenericDialect {})
```
killme2008: Using the exists parser dialect looks better.

src/sql/src/statements/query.rs
```diff
@@ -45,8 +50,50 @@ impl TryFrom<Query> for SpQuery {
...
+
+        if hybrid_cte.recursive {
+            write!(f, "RECURSIVE ")?;
+        }
+
+        for (i, cte) in hybrid_cte.cte_tables.iter().enumerate() {
+            if i > 0 {
+                write!(f, ", ")?;
+            }
+            write!(f, "{}", cte.name)?;
```
evenyag: Can we implement Display for HybridCteWith?

src/sql/src/parsers/with_tql_parser.rs
evenyag: Can we check the tql in the new parser instead? The old parser still remains unchanged and keeps the old dialect.

src/sql/src/parsers/with_tql_parser.rs
```diff
@@ -0,0 +1,550 @@
...
+
+                // Skip to matching closing parenthesis
+                let mut paren_depth = 1;
+                while paren_depth > 0 && self.parser.peek_token() != Token::EOF {
+                    match self.parser.next_token().token {
+                        Token::LParen => paren_depth += 1,
+                        Token::RParen => paren_depth -= 1,
+                        _ => {}
+                    }
+                }
```
evenyag: Can we wrap a method to do this?

src/sql/src/parsers/with_tql_parser.rs
```diff
@@ -0,0 +1,550 @@
...
+        let columns = if self.parser.consume_token(&Token::LParen) {
+            let mut columns = Vec::new();
+            loop {
+                let column = self.parser.parse_identifier().context(error::SyntaxSnafu)?;
+                columns.push(Self::canonicalize_identifier(column));
+
+                if !self.parser.consume_token(&Token::Comma) {
+                    break;
+                }
+            }
```
evenyag: https://docs.rs/sqlparser/latest/sqlparser/parser/struct.Parser.html#method.parse_comma_separated
https://docs.rs/sqlparser/latest/sqlparser/parser/struct.Parser.html#method.parse_parenthesized_qualified_column_list

Can we use `parse_comma_separated` or functions like `parse_parenthesized_qualified_column_list`?

src/sql/src/parsers/with_tql_parser.rs
waynexia: we can't retrieve the current dialect from `Parser`. But it should be Ok as this is for parsing promql, which doesn't have quote issues

src/sql/src/parsers/with_tql_parser.rs
waynexia: it now only occurs once

waynexia:
I've refined the vibe garbage and cut the diff size by half, PTAL again @killme2008 @evenyag

evenyag (approved):

Issue #3728, opened 2024-04-17
Compatibility test suit


waynexia:
### What type of enhancement is this?

Tech debt reduction, User experience

### What does the enhancement do?

With DB going stable, we should care about backward compatibility and forward compatibility between essential versions.

There is an existing compatibility [test suit](https://github.com/GreptimeTeam/greptimedb/blob/main/tests/compat/README.md) which leverages sqlness as harness. But it has some problems:
- Depending on a "bug" behavior of sqlness https://github.com/GreptimeTeam/greptimedb/pull/3724
- Doesn't aware of feature version. I.e., some features are not available in the old version, and we may deprecate some features in the new version. Mechanisms like `#[since(v0.7.0)]` and `#[till(v0.8.0)]` looks necessary to compatibility test.
- It's based on bash script 😣 Prefer Rust/Python/Go etc that are easy to maintain

MichaelScofield:
Seems this feature is not done?

waynexia:
yes, it's still pending

PR #6460, opened 2025-07-04
docs(rfc): compatibility test framework

waynexia:

Related to https://github.com/GreptimeTeam/greptimedb/issues/3728

RFC for a backward/forward compatibility test framework

[rendered](https://github.com/waynexia/greptimedb/blob/rfc-compat-test/docs/rfcs/2025-07-04-compatibility-test-framework.md)

## PR Checklist


docs/rfcs/2025-07-04-compatibility-test-framework.md
```diff
@@ -0,0 +1,151 @@
...
+
+`TEMPLATE` is another new interceptor, it can generate queries from a template based on a runtime data.
+
+In above example, we need to run the `SHOW CREATE TABLE` query for all existing tables, so we can use the `TEMPLATE` interceptor to generate the query with a dynamic table list.
+
+### RUNNER
+
+There are also some extra requirement for the runner itself:
+
+- It should run the test cases in sequence, first `1.feature/`, then `2.verify/`, and finally `3.cleanup/`.
```
discord9, line 127: How do one extend verify step if certain feature need to do some thing that no other features require?
Also why keep features/verify/cleanup in separate folder, wouldn't it make more sense to put everything related to one features to one folder(Or might even in one .sql file?)

waynexia: > How do one extend verify step if certain feature need to do some thing that no other features require?

One can still write normal sqlness case in `2.verify`

> Also why keep features/verify/cleanup in separate folder

In the execution order. If this is not clear, we also merge `1.feature` and `3.cleanup` to file level, like
```
feature/
    prepare.sql
    cleanup.sql
verify/
   ....
```

docs/rfcs/2025-07-04-compatibility-test-framework.md
```diff
@@ -0,0 +1,151 @@
...
+
+Follows the `ARG` interceptor in sqlness, we can mark a feature is available between two given versions. Only the `since` is required:
+
+```sql
+-- SQLNESS ARG since=VERSION_STRING [till=VERSION_STRING]
+```
+
+### IGNORE_RESULT
+
+`IGNORE_RESULT` is a new interceptor, it tells the runner to ignore the result of the query, only check whether the query is executed successfully.
```
evenyag, line 113: `IGNORE_RESULT` looks like it ignores whether the query succeeded or failed.

docs/rfcs/2025-07-04-compatibility-test-framework.md
```diff
@@ -0,0 +1,151 @@
...
+
+# check forward compatibility when downgrading from v0.15.0 to v0.13.0
+./sqlness run --from=0.15.0 --to=0.13.0
+```
+
+We can also use a script to run the compatibility test for all the versions in a given range to give a quick report with all versions we need.
+
+And we always bump the version in `Cargo.toml` to the next major release version, so the next major release version can be used as "latest" unpublished version for scenarios like local testing.
+
+# Alternatives
```
evenyag, line 149: A limitation may be that we can't cover HTTP/gRPC APIs.

discord9: > A limitation may be that we can't cover HTTP/gRPC APIs.

maybe can expand sqlness to send http/grpc request?

discord9 (approved):

docs/rfcs/2025-07-04-compatibility-test-framework.md
discord9: > > How do one extend verify step if certain feature need to do some thing that no other features require?
> 
> One can still write normal sqlness case in `2.verify`
> 
> > Also why keep features/verify/cleanup in separate folder
> 
> In the execution order. If this is not clear, we also merge `1.feature` and `3.cleanup` to file level, like
> 
> ```
> feature/
>     prepare.sql
>     cleanup.sql
> verify/
>    ....
> ```

So it can be something like this?:
```
feature/
     1.prepare.sql
     2.verify.sql
     (optionally, if feature have special needs, most feature should not need this, but a escape hatch is always good to have)
     3.cleanup.sql
2.verify/
    ....(general verification steps)
```
This should be both fixable and easy to maintain for most features, using `<num>.name` and `<num>` to denoting executing order?(If file not denoted, inherit from parent folder?)
And sqlness will run files in that order, i.e.:
```
feature/1.prepare.sql
2.verify/*
feature/2.verify.sql
feature/3.cleanup.sql
```

docs/rfcs/2025-07-04-compatibility-test-framework.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Maintenance
+
+Each time implement a new feature that should be covered by the compatibility test, we should create a new test case in `1.feature/` and `3.cleanup/` for them. And check if existing cases in `2.verify/` can be reused to verify the database behavior.
+
+This simulates an enthusiastic user who uses all the new features at the first time. All the new Maintenance burden is on the feature implementer to write one more test case for the new feature, to "fixation" the behavior. And once there is a breaking change in the future, it can be detected by the compatibility test framework automatically.
+
+Another topic is about deprecation. If a feature is deprecated, we should also mark it in the test case. Still use above example, assume we deprecate the `index.granularity` and `index.false_positive_rate` index options in `v0.99.0`, we can mark them as:
+```sql
+-- SQLNESS ARG since=0.15.0 till=0.99.0
```
discord9, line 93: Should enterprise version also maintain a mapping from open source version number to enterprise version number, so that it can run sqlness with something like
```
./sqlness run --from=0.15.0 --to=0.16.0 --from-url=<enterprise-binary-url>
```
so it can download old enterprise version as the old version binary to use in step one?
And also enterprise version number like `v20.0.0ee` or something like that?For enterprise feature testing?

sunng87: I wonder how we will make use of the version number here. Is it designed to fetch docker tag?

Our enterprise version will soon reuse open source version number in its cargo's `package.number` (but maintain its own version number to match enterprise release cycle). If necessary, we can add a command line option to output it's `package.number` for the framework to check, like `greptime --base-version`

discord9: > I wonder how we will make use of the version number here. Is it designed to fetch docker tag?
> 
> Our enterprise version will soon reuse open source version number in its cargo's `package.number` (but maintain its own version number to match enterprise release cycle). If necessary, we can add a command line option to output it's `package.number` for the framework to check, like `greptime --base-version`

sqlness can already fetch&run binary for standalone/distributed now, it's not rely on docker yet but might add a option to run db on docker instead

sunng87: Got it. Just to confirm the usage of version. I will add a `--base-version` option for enterprise binary that outputs CARGO_PKG_VERSION, which is designed to sync with open source.

docs/rfcs/2025-07-04-compatibility-test-framework.md
waynexia: that's a large topic, it may exceed the scope of current PR. we can include it once it's supported

docs/rfcs/2025-07-04-compatibility-test-framework.md
waynexia: well, let's make a new one then

Issue #5405, opened 2025-01-20
PromQL: result from query_range not ordered by timestamp


0neSe7en:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Query Engine

### Minimal reproduce step

When using the `/v1/query_range` endpoint with a PromQL query like: `host_disk_reads_completed_total * host_disk_reads_completed_total` returns unordered values. 
seems like using binary operator between two metrics will cause this issue.

### What did you expect to see?

the results should be ordered by timestamp

### What did you see instead?

not ordered by timestamp

```json
{
    "status": "success",
    "data": {
        "resultType": "matrix",
        "result": [
            {
                "metric": {
                    "__name__": "host_disk_reads_completed_total",
                    "collector": "disk",
                    "device": "disk0",
                    "host": "<mask>"
                },
                "values": [
                    [
                        1737326594.0,
                        "2672939735631361"
                    ],
                    [
                        1737326600.0,
                        "2673913964849604"
                    ],
                    [
                        1737326588.0,
                        "2672939735631361"
                    ],
                    [
                        1737326589.0,
                        "2672939735631361"
                    ],
                    [
                        1737326595.0,
                        "2672939735631361"
                    ],
                    [
                        1737326596.0,
                        "2673913964849604"
                    ],
                    [
                        1737326606.0,
                        "2673913964849604"
                    ],
                    [
                        1737326592.0, <---------------------------
                        "2672939735631361"
                    ],
                    [
                        1737326598.0,
                        "2673913964849604"
                    ],
                    [
                        1737326599.0,
                        "2673913964849604"
                    ],
                    [
                        1737326601.0,
                        "2673913964849604"
                    ],
                    [
                        1737326604.0,
                        "2673913964849604"
                    ],
                    [
                        1737326607.0,
                        "2673913964849604"
                    ],
                    [
                        1737326610.0,
                        "2673913964849604"
                    ],
                    [
                        1737326597.0,
                        "2673913964849604"
                    ],
                    [
                        1737326603.0,
                        "2673913964849604"
                    ],
                    [
                        1737326608.0,
                        "2673913964849604"
                    ],
                    [
                        1737326605.0,
                        "2673913964849604"
                    ],
                    [
                        1737326590.0,
                        "2672939735631361"
                    ],
                    [
                        1737326591.0,
                        "2672939735631361"
                    ],
                    [
                        1737326602.0,
                        "2673913964849604"
                    ],
                    [
[...]

PR #6474, opened 2025-07-08
chore: sort range query return values

shuiyisong:

#5405 

This PR does
1. Sort timestamp values within a metric
2. Sort metrics by labels

Although the first rule is not explicitly stated in the official Prometheus documents, the existing dashboards happen to rely on this 'feature'. Also we find that compatible third-party backends also implement sorting upon returning timestamp-value pairs.

## PR Checklist

### DIFF
--- a/tests-integration/tests/http.rs
+++ b/tests-integration/tests/http.rs
@@ -2455,7 +2455,7 @@ transform:
         "test_pipeline_with_vrl",
         &client,
         "select * from d_table",
-        "[[2436,1716668197217000000]]",
+        "[[\"channel_1\",1716668197217000000]]",
     )
     .await;
 


src/servers/src/http/result/prometheus_resp.rs
```diff
@@ -320,6 +321,8 @@ impl PrometheusJsonResponse {
                     });
                 }
                 PromQueryResult::Matrix(ref mut v) => {
+                    // sort values by timestamp
+                    values.sort_by(|a, b| a.0.partial_cmp(&b.0).unwrap_or(Ordering::Equal));
```
evenyag: Do we need to check with https://doc.rust-lang.org/stable/std/primitive.slice.html#method.is_sorted first? We don't need to sort them if the are already sorted.

evenyag (approved):

Issue #6482, opened 2025-07-08
Promql sum query stopped working in 0.15.1


Stephan3555:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Frontend

### Minimal reproduce step

Unable to run following query within greptime dashboard:

<img width="2534" height="754" alt="Image" src="https://github.com/user-attachments/assets/2e660a34-3f63-449f-b76d-b33ae09b80b3" />

### What did you expect to see?

Results, like it was the case in 0.14.3

### What did you see instead?

Error message, attached photo

### What operating system did you use?

your docker image

### What version of GreptimeDB did you use?

0.15.1

### Relevant log output and stack trace

```bash
No log output in frontend nodes when i run the query in the greptime dashboard
```

Stephan3555:
I rolled back to 0.14.3 again. Same query is now giving back results

killme2008:
This is only an update to the error message, will be fixed in https://github.com/GreptimeTeam/greptimedb/pull/6481

Ensure the table exists.

Stephan3555:
You are right, the table was not found. But my system is broken in the newest version. I just made the wrong assumption.

<img width="2559" height="1126" alt="Image" src="https://github.com/user-attachments/assets/83214188-33ec-4fb0-88fb-6db6d884aa50" />

Here is an example. On my prod system 0.14.3 is running. on my dev system 0.15.1 is running. Same query is not working anymore on 0.15.1

Stephan3555:
On 0.15.1 the return status code is 500. Nothing else changed except an upgrade of greptime-operator and greptime-cluster to the lates helm chart versions:

- greptimedb-operator: 0.3.0 --> 0.4.0
- greptimedb-cluster: 0.4.2 --> 0.6.4

<img width="2550" height="1149" alt="Image" src="https://github.com/user-attachments/assets/139a7d6d-3a86-4d9f-9cc0-85d0770db2d0" />

field organization_id is still present. Here are the metadata of the table (version 0.15.1)

```
CREATE TABLE IF NOT EXISTS "jmeter_threads" (
  "greptime_timestamp" TIMESTAMP(3) NOT NULL,
  "greptime_value" DOUBLE NULL,
  "organization_id" STRING NULL INVERTED INDEX,
  "heliosrun_id" STRING NULL INVERTED INDEX,
  "app" STRING NULL INVERTED INDEX,
  "kubernetes_namespace" STRING NULL INVERTED INDEX,
[...]

PR #6484, opened 2025-07-09
fix: expand on conditional commutative as well

waynexia:

Closes https://github.com/GreptimeTeam/greptimedb/issues/6482

`ConditionalCommutative` and `PartialCommutative` may change column requirements. We should stop checking the following plans if we don't maintain a proper per-stage column list. I.e., this is not the best fix or most appropriate implementation, but it does fix the problem. Will submit a follow-up PR to implement that mechanism. 

## PR Checklist


tests/cases/distributed/tql-explain-analyze/analyze.result
```diff
@@ -140,26 +138,18 @@ TQL ANALYZE (0, 10, '5s') rate(test[10s]);
...
+| 0_| 0_|_CoalesceBatchesExec: target_batch_size=8192 REDACTED
+|_|_|_FilterExec: prom_rate(j_range,i,j,Int64(10000))@1 IS NOT NULL REDACTED
+|_|_|_ProjectionExec: expr=[j@1 as j, prom_rate(j_range@4, i@0, j@1, 10000) as prom_rate(j_range,i,j,Int64(10000)), k@2 as k, l@3 as l] REDACTED
 |_|_|_PromRangeManipulateExec: req range=[0..10000], interval=[5000], eval range=[10000], time index=[j] REDACTED
 |_|_|_PromSeriesNormalizeExec: offset=[0], time index=[j], filter NaN: [true] REDACTED
 |_|_|_PromSeriesDivideExec: tags=["k", "l"] REDACTED
+|_|_|_MergeScanExec: REDACTED
+|_|_|_|
+| 1_| 0_|_SortPreservingMergeExec: [k@2 ASC, l@3 ASC, j@1 ASC] REDACTED
 |_|_|_SeriesScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0}, "distribution":"PerSeries" REDACTED
```
evenyag: Should we put SortPreservingMergeExec right after SeriesScan? It may have performance issue.

killme2008:
There is a test failure @waynexia

killme2008:
FYI. No need to rush a fix; we can wait for a better solution.

discord9:
I think this node with topk is necessary for sqlness to always scan both partition(not necessary for correctness)
```
|_|_|_SortExec: TopK(fetch=2), expr=[ts@1 DESC], preserve_partitioning=[true] REDACTED
```
If this line is removed, then it seems the query might only scan one region and get enough for the parent node `SortPreservingMergeExec: [ts@1 DESC], fetch=2 REDACTED` and return early, completely ignoring another region
however this is unstable as another region might comlete scan early with zero rows returned
However I don't know why seems `SortPreservingMergeExec` can end early even with fetch=2

discord9:
[rep_db.log](https://github.com/user-attachments/files/21140903/rep_db.log)
[rep_cli.log](https://github.com/user-attachments/files/21140905/rep_cli.log)
local reproduce, notice how mysql explain analyze sometimes end up with one region, somtimes end up with two region(see near end of file in `rep_cli.log`, looks like somehow one of the merge scan get cancelled early(notice how in db log only one merge scan is spawnned some times) and make metrics to (1024,1) get missed and never collected
could be like this:
1. (1024,0) get polled, and produced zero rows,
2. (1024,1) get polled, produced 4 rows,

the merge sort in `SortPreservingMergeExec fetch=2` sees it can now yield final result(as both merge scan is sorted)
so it cancelled the stream of (1024,1), which cause (1024, 1) to miss collect metrics(because merge scan collect metrics after all data is drained.)
but if the upper merge sort execute in a different order:
1. (1024, 1) yield 4 rows, then (1024, 1) yield again, finish and collect metrics
2. (1024, 0) didn't yield and just ended and collect metrics

then all metrics of merge scan is collected and `EXPLAIN ANALYZE` get correct results
Previously frontend have a `SortExec` with `TopK(n)` so it maybe didn't cancel any merge scan stream early?

see #6491 for fix(notice that this fix still seems failed to collect cancelled metrics in time somehow)

discord9:
> FYI. No need to rush a fix; we can wait for a better solution.

This is actually a logic bug, because sort get expanded, then it's parent node aggr also get expanded(while it shouldn't), this fix make sure that for case like this:
```
Aggregate:
  Sort:
```
only the `Sort` part get expanded to like:
```
frontend:
Aggregate: ...
  MergeSort: ...
datanode:
  Sort: ...
```
instead of wrongly expand to (due to the step aggr push down)
```
frontend:
UpperAggregate: ...
  MergeSort: ...
datanode:
LowerAggregate: ...
  Sort: ...
```

discord9 (approved):
LGTM

discord9:
@waynexia PTAL

PR #6488, opened 2025-07-09
fix(grpc): check grpc client unavailable

v0y4g3r:

`tokio::sync::mpsc::bounded::Sender::try_send` will return an error when the bounded channel is full, but in that case we should break the stream and return error to client. 

This PR changes `Sender::try_send` to `Sender::send`, which correctly return an error only when client disconnects.

## PR Checklist

### DIFF
--- a/src/servers/src/grpc/greptime_handler.rs
+++ b/src/servers/src/grpc/greptime_handler.rs
@@ -42,6 +42,7 @@ use session::hints::READ_PREFERENCE_HINT;
 use snafu::{OptionExt, ResultExt};
 use table::TableRef;
 use tokio::sync::mpsc;
+use tokio::sync::mpsc::error::TrySendError;
 
 use crate::error::Error::UnsupportedAuthScheme;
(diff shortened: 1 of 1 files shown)


killme2008 (approved):
LGTM

src/servers/src/grpc/greptime_handler.rs
```diff
@@ -176,7 +176,7 @@ impl GreptimeRequestHandler {
                 let result = result
                     .map(|x| DoPutResponse::new(request_id, x))
                     .map_err(Into::into);
-                if result_sender.try_send(result).is_err() {
+                if result_sender.send(result).await.is_err() {
```
MichaelScofield: `send` will wait until there's capacity. What if the client is not responding, will the results be cached in our db forever? How about checking the `try_send`'s error instead?

v0y4g3r: fixed in fe70ee94a

PR #6489, opened 2025-07-09
fix: skip nan in prom remote write pipeline

shuiyisong:

Skip the NaN value in Prometheus remote write for pipeline execution.

## PR Checklist

### DIFF
--- a/src/servers/src/proto.rs
+++ b/src/servers/src/proto.rs
@@ -390,6 +390,10 @@ impl PromSeriesProcessor {
         let one_sample = series.samples.len() == 1;
 
         for s in series.samples.iter() {
+            // skip NaN value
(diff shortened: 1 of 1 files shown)


src/servers/src/proto.rs
```diff
@@ -390,6 +390,10 @@ impl PromSeriesProcessor {
         let one_sample = series.samples.len() == 1;
 
         for s in series.samples.iter() {
+            // skip NaN value
+            if s.value.is_nan() {
+                continue;
+            }
```
evenyag, line 396: Will we output NaN to the result?

shuiyisong: No, the pipeline cannot handle NaN, so we have to filter it beforehand.

PR #6494, opened 2025-07-09
fix: correctly update partition key indices during alter table operations

WenyXu:

When performing `ALTER TABLE` operations (such as adding with the location or dropping columns), the partition key indices might not be correctly updated in the new table metadata.

- Add partition key indices mapping in both AddColumns and DropColumns alter operations
- Prevent removal of partition columns

## PR Checklist


src/common/meta/src/ddl/alter_table/executor.rs
```diff
@@ -304,5 +304,10 @@ fn build_new_table_info(
         | AlterKind::DropDefaults { .. } => {}
     }
 
+    info!(
+        "Built new table info: {:?} for table {}, table_id: {}",
+        new_info.meta, table_name, table_id
+    );
```
evenyag, line 310: Do we implement Debug for TableMeta? I wonder the default implementation maybe too verbose

killme2008 (approved):

killme2008:
LGTM
But I think we can add a sqlness test for it.

tests/cases/standalone/common/alter/alter_table_first_after.sql
```diff
@@ -47,3 +47,36 @@ SELECT * FROM t;
...
+    (2000, 'e', 5),
+    (2100, 'f', 6),
+    (2200, 'g', 7),
+    (2400, 'h', 8);
+
+SELECT * FROM my_table WHERE a > 100 order by a;
+
+SELECT count(*) FROM my_table WHERE a > 100;
+
+ALTER TABLE my_table ADD COLUMN c STRING FIRST;
```
killme2008, line 76: Add a test to remove partition key column, which should raise an error.

WenyXu: It's added in `drop_col.sql`

PR #6495, opened 2025-07-09
refactor(flow): faster time window expr

discord9:

faster time window expr with flow batch mode

## PR Checklist

### DIFF
--- a/src/flow/src/batching_mode/time_window.rs
+++ b/src/flow/src/batching_mode/time_window.rs
@@ -50,7 +50,8 @@ use snafu::{ensure, OptionExt, ResultExt};
 
 use crate::adapter::util::from_proto_to_data_type;
 use crate::error::{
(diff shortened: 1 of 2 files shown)


src/flow/src/batching_mode/time_window.rs
```diff
@@ -121,13 +124,53 @@ impl TimeWindowExpr {
...
+            // a - (a % n) impl ceil to nearest n * stride
+            let time_delta = time_diff_ns - (time_diff_ns % stride_ns);
+
+            if time_diff_ns < 0 && time_delta != time_diff_ns {
+                // The origin is later than the source timestamp, round down to the previous bin
+
+                time_delta - stride_ns
+            } else {
+                time_delta
+            }
```
waynexia, line 150: found a std api for this https://doc.rust-lang.org/std/primitive.i64.html#method.next_multiple_of

>If rhs is negative, calculates the largest value less than or equal to self that is a multiple of rhs.

(God knows why a numeric arithmetic API also needs feature gating...

waynexia (approved):
need some comments about the new fast path, rest lgtm

PR #6499, opened 2025-07-10
feat: update partial execution metrics

waynexia:

Stream may be terminated earlier (no follow-up poll) by upper plan. In this case we will lose its execution metrics, as they are collected when a stream drains dry.

This patch changes the behavior to update metrics every time. To make sure the recorded metrics are always up-to-date.

RCU calculation logic remains unchanged, as it's some kind of "counter" which doesn't support partial update.

## PR Checklist

### DIFF
--- a/src/client/src/region.rs
+++ b/src/client/src/region.rs
@@ -159,8 +159,9 @@ impl RegionRequester {
             ));
 
             let mut buffered_message: Option<FlightMessage> = None;
+            let mut stream_ended = false;
 
-            loop {
+            while !stream_ended {
                 // get the next message from the buffered message or read from the flight message stream
                 let flight_message_item = if let Some(msg) = buffered_message.take() {
                     Some(Ok(msg))
@@ -212,7 +213,8 @@ impl RegionRequester {
                                 }
                             }
                         } else {
-                            // the loop will break in the next iteration
+                            // the stream has ended
+                            stream_ended = true;
                         }
 
                         yield result_to_yield;


waynexia:
one thing left to do is handle the distributed case, where the metrics are reported through a special message to the frontend at the end of a stream.

src/client/src/region.rs
```diff
@@ -157,19 +157,68 @@ impl RegionRequester {
...
                             schema_cloned.clone(),
                             record_batch,
-                        )
+                        );
+
+                        // get the next message from the stream. normally it should be a metrics message.
+                        if let Some(next_flight_message_result) = flight_message_stream.next().await
+                        {
+                            match next_flight_message_result {
+                                Ok(FlightMessage::Metrics(s)) => {
```
discord9, line 192: What if datanode's query scan a very large amount of files(i.e. 2000+) and displayed in metrics, that caused this metrics string to be extremely large, could it hurt performance?(It seems typical record batch size is 8192, so getting metrics would be somewhere often?

src/servers/src/grpc/flight/stream.rs
```diff
@@ -89,6 +89,15 @@ impl FlightRecordBatchStream {
                         warn!(e; "stop sending Flight data");
                         return;
                     }
+                    if let Some(metrics) = recordbatches
+                        .metrics()
+                        .and_then(|m| serde_json::to_string(&m).ok())
```
discord9: maybe send metrics every X rows or every X recordbatch?

evenyag: Updating metrics too frequently may affect performance.

waynexia: The poll is based on record batch, only send right after every record batch can ensure the metrics must exists. The overhead is a consideration. I'm also hesitate for this.

waynexia:
I changed the behavior to only send partial metrics from region server under EXPLAIN/ANALYZE VERBOSE mode. PTAL @evenyag @discord9

src/client/src/region.rs
```diff
@@ -157,19 +157,68 @@ impl RegionRequester {
...
+                                    break;
+                                }
+                                Err(e) => {
+                                    yield Err(BoxedError::new(e)).context(ExternalSnafu);
+                                    break;
+                                }
+                            }
+                        } else {
+                            // the loop will break in the next iteration
+                        }
```
evenyag: Will we poll the flight_message_stream again in the next iteration? If so, the stream may panic.

waynexia: this branch is when stream returns `None`, which means the entire stream is empty and should not be pulled again

discord9: > this branch is when stream returns `None`, which means the entire stream is empty and should not be pulled again

The next loop will try to pull again even if this time it yield a `None`, consider break after yield if stream is empty?

evenyag: If `buffered_message` is `None`, we will poll `flight_message_stream` again at L168 in the `loop` block. Did I understand the control flow correctly?

Do we need a flag to mark the stream is done?

waynexia: Got it, add a flag

PR #6501, opened 2025-07-10
feat: introduce common event recorder

zyy17:

**Adds the following traits to define the event recorder(`src/common/event-recorder`)**

   - `Event`: Event trait defines the interface for events that can be recorded and persisted to system tables:
  
     ```rust
     pub trait Event: Send + Sync + Debug {
         /// Returns the type of the event.
         fn event_type(&self) -> &str;

         /// Returns the timestamp of the event. Default to the current time.
         fn timestamp(&self) -> Timestamp {
             Timestamp::current_time(TimeUnit::Nanosecond)
         }

         /// Returns the JSON bytes of the event as the payload.
         fn json_payload(&self) -> Result<String>;

         /// Add the extra schema to the event with the default schema.
         fn extra_schema(&self) -> Vec<ColumnSchema> {
             vec![]
         }

         /// Add the extra row to the event with the default row.
         fn extra_row(&self) -> Result<Row> {
             Ok(Row { values: vec![] })
         }

         /// Returns the event as any type.
         fn as_any(&self) -> &dyn Any;
     }
     ```

     By default, the event recorder will record the event with the following schema: `type`, `payload`, `timestamp`. If the developer wants to add more columns to the event, they can override the `extra_schema()` and `extra_row()` methods.

   - `EventRecorder`: EventRecorder trait defines the interface for recording events:
   
     ```rust
     pub trait EventRecorder: Send + Sync + 'static {
         /// Records an event for persistence and processing by [EventHandler].
         fn record(&self, event: Box<dyn Event>) -> Result<()>;
     }
     ```

     The `EventRecorder` is the major API that can be used in the application code to record events.
  
   - `EventHandler`: EventHandler trait defines the interface for how to handle the event:
  
     ```rust
     pub trait EventHandler: Send + Sync + 'static {
         /// Processes and handles incoming events. The [DefaultEventHandlerImpl] implementation forwards events to frontend instances for persistence.
         /// We use `&[Box<dyn Event>]` to avoid consuming the events, so the caller can buffer the events and retry if the handler fails.
         async fn handle(&self, events: &[Box<dyn Event>]) -> Result<()>;
     }
     ```

     Different roles(e.g. frontend, meta-srv, datanode) can implement the `EventHandler` trait to handle the events. **All the events will be pesisted as a mito system table `greptime_private.events`**.

## PR Checklist

### DIFF
--- a/src/common/event-recorder/src/error.rs
+++ b/src/common/event-recorder/src/error.rs
@@ -28,13 +28,6 @@ pub enum Error {
         location: Location,
     },
 
-    #[snafu(display("Keyvalue backend error"))]
-    KvBackend {
-        source: common_meta::error::Error,
-        #[snafu(implicit)]
-        location: Location,
-    },
-
     #[snafu(display("Mismatched schema, expected: {:?}, actual: {:?}", expected, actual))]
     MismatchedSchema {
         #[snafu(implicit)]
@@ -50,7 +43,7 @@ impl ErrorExt for Error {
     fn status_code(&self) -> StatusCode {
         match self {
             Error::MismatchedSchema { .. } => StatusCode::InvalidArguments,
-            Error::NoAvailableFrontend { .. } | Error::KvBackend { .. } => StatusCode::Internal,
+            Error::NoAvailableFrontend { .. } => StatusCode::Internal,
         }
     }
 
--- a/src/common/event-recorder/src/error.rs
+++ b/src/common/event-recorder/src/error.rs
@@ -1,0 +1,60 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use api::v1::ColumnSchema;
+use common_error::ext::ErrorExt;
+use common_error::status_code::StatusCode;
+use common_macro::stack_trace_debug;
+use snafu::{Location, Snafu};
+
+#[derive(Snafu)]
+#[snafu(visibility(pub))]
+#[stack_trace_debug]
+pub enum Error {
+    #[snafu(display("No available frontend"))]
+    NoAvailableFrontend {
+        #[snafu(implicit)]
+        location: Location,
+    },
+
+    #[snafu(display("Keyvalue backend error"))]
+    KvBackend {
+        source: common_meta::error::Error,
+        #[snafu(implicit)]
+        location: Location,
+    },
+
+    #[snafu(display("Mismatched schema, expected: {:?}, actual: {:?}", expected, actual))]
+    MismatchedSchema {
+        #[snafu(implicit)]
+        location: Location,
+        expected: Vec<ColumnSchema>,
+        actual: Vec<ColumnSchema>,
+    },
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+impl ErrorExt for Error {
+    fn status_code(&self) -> StatusCode {
+        match self {
+            Error::MismatchedSchema { .. } => StatusCode::InvalidArguments,
+            Error::NoAvailableFrontend { .. } | Error::KvBackend { .. } => StatusCode::Internal,
+        }
+    }
+
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
+    }
+}
--- a/src/common/event-recorder/src/recorder.rs
+++ b/src/common/event-recorder/src/recorder.rs
@@ -204,7 +204,7 @@ pub trait EventRecorder: Send + Sync + 'static {
     fn record(&self, event: Box<dyn Event>);
 
     /// Cancels the event recorder.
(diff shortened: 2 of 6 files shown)


zyy17:
This PR is split from https://github.com/GreptimeTeam/greptimedb/pull/6423.

src/common/event-recorder/src/error.rs
```diff
@@ -0,0 +1,60 @@
...
+        #[snafu(implicit)]
+        location: Location,
+    },
+
+    #[snafu(display("Keyvalue backend error"))]
+    KvBackend {
+        source: common_meta::error::Error,
+        #[snafu(implicit)]
+        location: Location,
+    },
```
WenyXu: unused?

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    // NOTE: While we implement a retry mechanism for failed event handling, there is no guarantee that all events will be processed successfully.
+    async fn flush_events_to_handler(&self, buffer: &mut Vec<Box<dyn Event>>) {
+        if !buffer.is_empty() {
+            debug!("Flushing {} events to the event handler", buffer.len());
+
+            // Use the fixed interval to retry the failed events.
+            let mut retry = 1;
+            while retry <= self.max_retry_times {
```
WenyXu: nit: we can use `backon` crate

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,443 @@
...
+
+    async fn process(mut self, buffer_size: usize) {
+        info!("Start the background processor in event recorder to handle the received events.");
+
+        let mut buffer = Vec::with_capacity(buffer_size);
+        let mut interval = tokio::time::interval(self.process_interval);
+
+        loop {
+            tokio::select! {
+                maybe_event = self.rx.recv() => {
```
WenyXu, line 343: nit: Consider handling the upper-layer cancel/stop signal and flushing pending events. Introducing a `CancellationToken` could help.

WenyXu (approved):
Rest LGTM

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,456 @@
...
+        let (tx, rx) = channel(DEFAULT_CHANNEL_SIZE);
+
+        let mut recorder = Self { tx, handle: None };
+
+        let processor = EventProcessor::new(
+            rx,
+            event_handler,
+            DEFAULT_FLUSH_INTERVAL_SECONDS,
+            DEFAULT_MAX_RETRY_TIMES,
+        );
```
WenyXu: nit: Someone should hold the cancel token

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,456 @@
...
+                        buffer.push(maybe_event);
+                    } else {
+                        // When received a closed signal, flush the buffer and exit the loop.
+                        self.flush_events_to_handler(&mut buffer).await;
+                        break;
+                    }
+                }
+                // Cancel the processor through the cancel token.
+                _ = self.cancel_token.cancelled() => {
+                    break;
```
WenyXu: nit: better flush the events, and log the cancel event.

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    // NOTE: While we implement a retry mechanism for failed event handling, there is no guarantee that all events will be processed successfully.
+    async fn flush_events_to_handler(&self, buffer: &mut Vec<Box<dyn Event>>) {
+        if !buffer.is_empty() {
+            debug!("Flushing {} events to the event handler", buffer.len());
+
+            let mut backoff = ExponentialBuilder::default()
+                .with_min_delay(Duration::from_millis(
+                    DEFAULT_FLUSH_INTERVAL_SECONDS.as_millis() as u64 / self.max_retry_times,
```
WenyXu: nit: ensure `self.max_retry_times` is not zero

zyy17: use `self.max_retry_times.max(1)` to prevent it(although it is always `DEFAULT_MAX_RETRY_TIMES`).

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,518 @@
...
+        let mut interval = tokio::time::interval(self.process_interval);
+
+        loop {
+            tokio::select! {
+                maybe_event = self.rx.recv() => {
+                    if let Some(maybe_event) = maybe_event {
+                        debug!("Received event: {:?}", maybe_event);
+
+                        if buffer.len() >= buffer_size {
+                            warn!(
```
killme2008: The log is unnecessary since this is expected behavior and doesn't warrant a warning. Using `debug!` is appropriate.

src/common/event-recorder/src/recorder.rs
```diff
@@ -0,0 +1,518 @@
...
+    Ok(())
+}
+
+/// EventRecorder trait defines the interface for recording events.
+pub trait EventRecorder: Send + Sync + 'static {
+    /// Records an event for persistence and processing by [EventHandler].
+    fn record(&self, event: Box<dyn Event>);
+
+    /// Cancels the event recorder.
+    fn cancel(&self);
```
killme2008: `stop` or `close` is more appropriate, as it stops the background processor.

killme2008 (approved):
Almost LGTM

src/common/event-recorder/src/recorder.rs
zyy17: Add the `close()` API to let the caller to decide when to close the event recorder.

PR #6529, opened 2025-07-15
feat: supports more db options

### LINKED ISSUE #5549: Table options not available in CREATE DATABASE
### What type of bug is this?

Configuration

Standalone mode, Metasrv

### Minimal reproduce step

It's mentioned in [doc](https://docs.greptime.com/reference/sql/create#create-database) table options are applicable to CREATE DATABSE WITH as well, but the [code](https://github.com/GreptimeTeam/greptimedb/blob/8026b1d/src/sql/src/parsers/create_parser.rs#L61) suggests it's hard coded to allow only TTL.

Execute the following SQL.

```
create database log with (ttl = '7d', storage = 's3');
```

```toml
# config file
[storage]
type = "File"
data_home = "/var/lib/greptimedb"
cache_path = "/var/lib/greptimedb/cache"
cache_capacity = "2GiB"

[wal]
dir = "/var/lib/greptimedb/wal"

[logging]
dir = "/var/lib/greptimedb/log"

[[storage.providers]]
name = "s3"
type = "S3"
# more s3 configs redacted
```

database created with default storage options.

### What did you see instead?

```
11:31:35 PM [error] create database log with (ttl = '7d', storage = 's3'); Unrecognized database option key: storage
```

docker.io/greptime/greptimedb:v0.11.3 on arm64 orangepi 5 pro

v0.11.3

```bash

```

killme2008:

Close #5549

## PR Checklist

### DIFF
--- a/src/common/meta/src/key/schema_name.rs
+++ b/src/common/meta/src/key/schema_name.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::fmt::Display;
 
 use common_catalog::consts::{DEFAULT_CATALOG_NAME, DEFAULT_SCHEMA_NAME};
@@ -59,7 +59,7 @@ pub struct SchemaNameValue {
     #[serde(default)]
     pub ttl: Option<DatabaseTimeToLive>,
     #[serde(default)]
-    pub extra_options: HashMap<String, String>,
+    pub extra_options: BTreeMap<String, String>,
 }
 
 impl Display for SchemaNameValue {
@@ -451,7 +451,7 @@ mod tests {
             parsed,
             Some(SchemaNameValue {
                 ttl: Some(Duration::from_secs(10).into()),
-                extra_options: HashMap::new(),
+                extra_options: BTreeMap::new(),
             })
         );
 
@@ -466,7 +466,7 @@ mod tests {
         // Simulate new format: ttl + extra_options
         let json = r#"{"ttl":"15s","extra_options":{"foo":"bar","baz":"qux"}}"#;
         let parsed = SchemaNameValue::try_from_raw_value(json.as_bytes()).unwrap();
-        let mut expected_options = HashMap::new();
+        let mut expected_options = BTreeMap::new();
         expected_options.insert("foo".to_string(), "bar".to_string());
         expected_options.insert("baz".to_string(), "qux".to_string());
         assert_eq!(
--- a/src/common/meta/src/key/schema_name.rs
+++ b/src/common/meta/src/key/schema_name.rs
@@ -58,12 +58,17 @@ impl Default for SchemaNameKey<'_> {
 pub struct SchemaNameValue {
     #[serde(default)]
     pub ttl: Option<DatabaseTimeToLive>,
+    #[serde(default)]
+    pub extra_options: HashMap<String, String>,
 }
 
 impl Display for SchemaNameValue {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         if let Some(ttl) = self.ttl.map(|i| i.to_string()) {
-            write!(f, "ttl='{}'", ttl)?;
+            write!(f, "'ttl'='{}'", ttl)?;
+        }
+        for (k, v) in self.extra_options.iter() {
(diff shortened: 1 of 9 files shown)


src/common/meta/src/key/schema_name.rs
```diff
@@ -58,12 +58,17 @@ impl Default for SchemaNameKey<'_> {
 pub struct SchemaNameValue {
     #[serde(default)]
     pub ttl: Option<DatabaseTimeToLive>,
+    #[serde(default)]
+    pub extra_options: BTreeMap<String, String>,
```
MichaelScofield, line 62: `ttl` and `extra_options` are two conversely ways to organize the options: `ttl` is to make options explicit, while `extra_options` is implicit. I think it's better to stick to one, either "flatten" the `extra_options`, make every options appeared like the "ttl", or absorb the "ttl" in the `extra_options`.

killme2008: The `ttl` setting is unique: tables without a specified `ttl` will inherit this value, and changing it will impact all such tables.

The other options act as template variables and should be separated.

src/table/src/requests.rs
```diff
@@ -65,6 +71,23 @@ pub const VALID_TABLE_OPTION_KEYS: [&str; 11] = [
     TABLE_DATA_MODEL,
 ];
 
+// Valid option keys when creating a db.
+static VALID_DB_OPT_KEYS: Lazy<HashSet<&str>> = Lazy::new(|| {
+    let mut set = HashSet::new();
```
MichaelScofield, line 77:
```suggestion
    let mut set = HashSet::with_capacity(6);
```

MichaelScofield (approved):

src/sql/src/parsers/create_parser.rs
```diff
@@ -215,6 +209,14 @@ impl<'a> ParserContext<'a> {
...
             );
         }
+        if let Some(append_mode) = options.get("append_mode") {
+            if append_mode == "true" && options.contains_key("merge_mode") {
+                return InvalidDatabaseOptionSnafu {
+                    key: "merge_mode".to_string(),
+                }
+                .fail();
+            }
+        }
```
evenyag, line 219: I think we should not put this check on the parser side. Is it possible to move the check to the ddl phase?

src/operator/src/statement/ddl.rs
```diff
@@ -107,7 +107,33 @@ impl StatementExecutor {
...
         let create_expr = &mut expr_helper::create_to_expr(&stmt, &ctx)?;
+        // We don't put ttl into the table options
+        // Because it will be used directly while compaction.
+        if let Some(schema_options) = schema_options {
+            for (key, value) in schema_options.extra_options.iter() {
+                create_expr
+                    .table_options
+                    .entry(key.clone())
+                    .or_insert(value.clone());
+            }
```
evenyag, line 134: If we set these options while creating the table, what if we want to change the database level setting?

For TTL, it affects all tables.

src/sql/src/parsers/create_parser.rs
killme2008: Why not check in advance? Checking later has already wasted a lot of resources

src/operator/src/statement/ddl.rs
killme2008: As I said and the issue's description, there are just as the template variables to create tables.

src/operator/src/statement/ddl.rs
```diff
@@ -107,7 +107,33 @@ impl StatementExecutor {
...
+        let schema_options = self
+            .table_metadata_manager
+            .schema_manager()
+            .get(SchemaNameKey {
+                catalog: &catalog,
+                schema: &schema,
+            })
+            .await
+            .context(TableMetadataManagerSnafu)?
+            .map(|v| v.into_inner());
```
evenyag, line 123: Realized that we should also skip this for system tables like the pipeline table...
Does TTL have this problem?

src/sql/src/parsers/create_parser.rs
evenyag: I don't think the parser should handle this business logic... But this isn't a blocker.

src/operator/src/statement/ddl.rs
evenyag: The pipeline is under greptime_private schema so this may not be a problem.

evenyag (approved):

PR #6534, opened 2025-07-16
fix: aggr group by all partition cols use partial commutative

discord9:

**this is a small fix(400+ lines), the rest of new code is unit testing**
- fix a bug where aggr group by all partition cols should use partial commutative, not commutative, so that if a push down is already in progress(either being `Sort` or `Limit`), aggr group by all partition cols will not be push-down
- fix a problem when join query have multiple merge scan, all merge scan metrics should be shown when `explain analyze`(in `analyze.rs`)
- add stack traced column requirement for expanding `Sort` properly, see `expand_proj_sort_proj` testcase
- fix a problem where partition columns is not checked for aliased column, see `proj_multi_alias_tracker` testcase
- add more unit test cases, almost exhausting all non-trivial combine of `Sort`, `Limit`, `Aggregate(Group by Partition Columns)` and `Aggregate(Step-able)`
- found bug with `Sort` and alias, choose to fix in next pr as this alias case is too complex to fix now and this is already a complex fix pr, see `expand_sort_alias_conflict_limit` unit test for failed optimize
- TODO: probably will add in next pr to choice to fallback to only push down table scan when `DistPlanner` produce error when trying to push down query
```
Aggr group by partition cols
  [ multiple comm node in the middle]
      Sort expr_list
```
which is currently wrongly optimized to :
```
frontend:
MergeSort expr_list
  MergeScan

datanode:
Aggr group by partition cols
  [ multiple comm node in the middle]
      Sort expr_list
```

which it shoudn't, as correct behavior is to optimize to like:
```
frontend:
Aggr group by partition cols
  MergeSort expr_list
    MergeScan

datanode:
  [ multiple comm node in the middle]
      Sort expr_list
```

TODO: more testcases around tql&aggr push down is needed, currently this path haven't been test thoroughly enough

## PR Checklist


src/query/src/dist_plan/analyzer.rs
```diff
@@ -223,7 +239,7 @@ impl PlanRewriter {
             Commutativity::Commutative => {}
             Commutativity::PartialCommutative => {
                 if let Some(plan) = partial_commutative_transformer(plan) {
-                    self.update_column_requirements(&plan);
+                    self.update_column_requirements(&plan, self.level - 1);
```
evenyag, line 272: Why do we use `level - 1` instead of `level`?

src/query/src/dist_plan/analyzer.rs
```diff
@@ -381,38 +417,85 @@ impl PlanRewriter {
 /// Requirements enforced by this rewriter:
 /// - Enforce column requirements for `LogicalPlan::Projection` nodes. Makes sure the
 ///   required columns are available in the sub plan.
+///
+#[derive(Debug)]
 struct EnforceDistRequirementRewriter {
-    column_requirements: HashSet<Column>,
+    // only enforce column requirements after the expanding node in question
```
evenyag: What does "the expanding node in question" mean?

tests/cases/distributed/explain/step_aggr_advance.sql
```diff
@@ -0,0 +1,307 @@
...
+    ORDER BY
+      a,
+      b
+  )
+GROUP BY
+  a;
+
+drop table aggr_optimize_not_count;
+
+drop table aggr_optimize_not;
```
evenyag: non-blocking
```suggestion
drop table aggr_optimize_not;

```

src/query/src/dist_plan/analyzer.rs
discord9: > Why do we use `level - 1` instead of `level`?

Because plan is parent of node that might expand

src/query/src/dist_plan/analyzer.rs
discord9: > What does "the expanding node in question" mean?

the node in `self.expand(node)`

discord9:
~found another bug with partition columns and alias, working on a fix too~
fixed

waynexia (approved):

src/query/src/dist_plan/analyzer/utils.rs
```diff
@@ -0,0 +1,318 @@
...
+                self.mapping
+            );
+        }
+    }
+
+    pub fn get_all_alias_for_col(&self, col_name: &str) -> Option<&HashSet<Column>> {
+        self.mapping.get(col_name)
+    }
+
+    #[allow(unused)]
```
evenyag, line 141: Did we use this?

src/query/src/dist_plan/commutativity.rs
```diff
@@ -405,8 +421,14 @@ impl Categorizer {
...
-        for col in partition_cols {
-            if !ref_cols.contains(col) {
+        for all_alias in partition_cols.values() {
+            let all_alias = all_alias
+                .iter()
+                .map(|c| c.name.clone())
+                .collect::<HashSet<_>>();
+            // check if ref columns intersect with all alias of partition columns
+            // is empty, if it's empty, not all partition columns show up in `exprs`
+            if ref_cols.intersection(&all_alias).count() == 0 {
```
evenyag, line 420: Maybe we can iterate ref_cols to check them in all_alias manually to avoid constructing the hash set again.

src/query/src/dist_plan/commutativity.rs
```diff
@@ -322,17 +335,20 @@ impl Categorizer {
...
         match plan.name() {
             name if name == SeriesDivide::name() => {
                 let series_divide = plan.as_any().downcast_ref::<SeriesDivide>().unwrap();
                 let tags = series_divide.tags().iter().collect::<HashSet<_>>();
-                for partition_col in partition_cols {
-                    if !tags.contains(partition_col) {
+
+                for all_alias in partition_cols.values() {
+                    let all_alias = all_alias.iter().map(|c| &c.name).collect::<HashSet<_>>();
+                    if tags.intersection(&all_alias).count() == 0 {
```
evenyag, line 336: nitpick: Can we iterate series_divide.tags() and check existence in the all_alias set?

src/query/src/dist_plan/analyzer.rs
```diff
@@ -381,42 +486,96 @@ impl PlanRewriter {
...
-            if column_requirements.is_empty() {
-                return Ok(Transformed::no(node));
-            }
+        // check that node doesn't have multiple children, i.e. join/subquery
+        if node.inputs().len() > 1 {
+            return Err(datafusion_common::DataFusionError::Internal(
+                "EnforceDistRequirementRewriter: node with multiple inputs is not supported"
+                    .to_string(),
+            ));
+        }
```
evenyag, line 526: If we return an error here, will we get an error when we execute a query with JOIN/SUBQUERY?

discord9: > If we return an error here, will we get an error when we execute a query with JOIN/SUBQUERY?

no, `should_expand` make sure that join/subquery never get expanded(so if any does have >1 input, it's unexpected)

src/query/src/dist_plan/commutativity.rs
discord9: > Maybe we can iterate ref_cols to check them in all_alias manually to avoid constructing the hash set again.

not avoidable still, since `all_alias` is a HashSet<Column> so can't get by string

src/query/src/dist_plan/analyzer/utils.rs
discord9: > Did we use this?

not yet but could be useful later

PR #6543, opened 2025-07-17
refactor(otlp_metric): make otlp metric compatible with promql

shuiyisong:

In this PR, we are trying to make OTLP metric input data compatible with Prometheus and PromQL, that it can be used directly as Prometheus metrics.
In order to be backward compatible, we check the table existence and its table option. If the table is created by the old path, we continue to write data in the old path, which is the original OTLP metric format.
If no table exists, we use the new Prometheus compatible path:
1. we now use metric engine by default
2. the resolution of timestamp is changed from nanosecond to millisecond to comply with metric engine's behavior.
3. only certain resource attributes are by default remain persistent in the table, others will be discarded.
4. introduce `x-greptime-otlp-metric-promote-all-resource-attrs` and `x-greptime-otlp-metric-promote-scope-attrs` to control the behavior.

## PR Checklist


src/frontend/src/instance/builder.rs
```diff
@@ -223,6 +224,7 @@ impl FrontendBuilder {
             slow_query_recorder,
             limiter,
             process_manager,
+            otlp_metrics_table_legacy_cache: DashMap::new(),
```
sunng87, line 227: is it possible to make this a member of inserter?

src/frontend/src/instance/otlp.rs
```diff
@@ -63,10 +86,21 @@ impl OpenTelemetryProtocolHandler for Instance {
             None
         };
 
-        self.handle_row_inserts(requests, ctx, false, false)
-            .await
-            .map_err(BoxedError::new)
-            .context(error::ExecuteGrpcQuerySnafu)
+        if is_legacy || !with_metric_engine {
```
sunng87: provide comments for why we use different logic, and what is considered legacy?

src/frontend/src/instance.rs
```diff
@@ -331,6 +348,112 @@ impl Instance {
...
+
+            let flag = hit_legacy;
+            // set cache for all names
+            names.iter().for_each(|name| {
+                if !cache.contains_key(*name) {
+                    cache.insert(name.to_string(), flag);
+                }
+            });
+            return Ok(flag);
+        }
```
sunng87, line 385: Is it possible to cache whether the database has legacy tables only?
If there is a cache for the database, use the cached result, otherwise we check the table info for result and fill the cache.

tests-integration/src/otlp.rs
```diff
@@ -81,12 +81,11 @@ mod test {
...
-| resource   | scope | telemetry_sdk_name | host       | greptime_timestamp            | greptime_value |
-+------------+-------+--------------------+------------+-------------------------------+----------------+
-| greptimedb | otel  | java               | testsevrer | 1970-01-01T00:00:00.000000100 | 100.0          |
-| greptimedb | otel  | java               | testserver | 1970-01-01T00:00:00.000000105 | 105.0          |
-+------------+-------+--------------------+------------+-------------------------------+----------------+",
++---------------------+----------------+
+| greptime_timestamp  | greptime_value |
++---------------------+----------------+
+| 1970-01-01T00:00:00 | 105.0          |
++---------------------+----------------+",
```
sunng87: There is no more labels for this table?

shuiyisong: Yes, by default only selected labels are allowed into persistence. Let me add some of these labels to the test.

https://github.com/GreptimeTeam/greptimedb/blob/ad3d35954d9c76d524ed914ed2c7c846f83b70a6/src/servers/src/otlp/metrics.rs#L40-L64

src/frontend/src/instance/builder.rs
shuiyisong: Yes. All we need is a place to put the cache, the checking logic itself doesn't have any state.
Putting it inside the inserter seems wired, it's more like a frontend check.

src/frontend/src/instance.rs
shuiyisong: This can still cause inconsistency. For example: 
1. the agent collects metric A, B, C from source 1 using the old OTLP metric path.
2. the cluster gets upgraded and restarted, all caches are gone
3. the agent collects metric E, F from source 2, since they're new, use the new compatible path

Now we have a database with both format of data. The next time A, B and C come in, it will consider the database is new mode regarding the cache, and lead to error because the table schema is wrong.

sunng87:
How do we add custom labels? For example, the Claude Code otlp exporter has its own attributes

shuiyisong:
> How do we add custom labels? For example, the Claude Code otlp exporter has its own attributes

I'm thinking about it too. Should we make it a config that gets read from the configuration file upon startup?

sunng87:
No. In managed service there is no chance to touch configuration files.

sunng87:
I think for the label whitelist behaviour we will have to be a little different with prometheus.

paomian:
How about replacing legacy to before_0.16 ？If the format is outdated again, it will be difficult to name it。

src/servers/src/http/extractor.rs
```diff
@@ -129,6 +131,43 @@ where
     }
 }
 
+pub struct OtlpMetricOptions {
```
paomian, line 135: Add some doc for this pub struct.

src/servers/src/http/extractor.rs
```diff
@@ -129,6 +131,43 @@ where
...
+    // persist all resource attributes to the table
```
paomian: We can change it from comment to doc.

src/frontend/src/instance.rs
```diff
@@ -331,6 +348,112 @@ impl Instance {
             .await
             .context(TableOperationSnafu)
     }
+
+    async fn check_otlp_legacy(
+        &self,
+        names: &Vec<&String>,
```
paomian:
```suggestion
        names: &[&String],
```

paomian:
I think our upgrade documentation should explain this difference in detail, otherwise users may experience unexpected behavior when migrating. 
For example, if the user was using the old mode before and had already written to it after the upgrade, the table version would be cached. Even if these tables are deleted, the old path would be used?

paomian (approved):
Rest LGTM

PR #6545, opened 2025-07-17
feat: adds uptime telemetry

killme2008:

Adds `uptime` telemetry, but only records approximate time ranges (e.g., hours, weeks) rather than exact values.

## PR Checklist


sunng87:
We need to make sure this is sent at right time.

killme2008:
> We need to make sure this is sent at right time.

It doesn't matter. It will be sent hourly, and we don't need exact results.

sunng87 (approved):

paomian (approved):

PR #6550, opened 2025-07-17
fix: estimate mem size for bulk ingester

fengys1996:

I encountered a problem: under the same write load, bluk ingester has much higher CPU than row-based gRPC ingester.

After some investigation, the reasons are as follows:

Bulk ingester estimates memory usage that is much larger than the actual memory usage, resulting in excessive flush and compact operations.

> The statistics of `get_array_memory_size` will include the underlying bytes, which causes duplicate statistics.

image

To the left of the red arrow is row-based ingester, and to the right is bulk ingester. It is clear that under the same write load, the bulk ingester flushes more.

## PR Checklist

### DIFF
--- a/src/mito2/src/memtable/bulk/part.rs
+++ b/src/mito2/src/memtable/bulk/part.rs
@@ -136,7 +136,12 @@ impl TryFrom<&BulkPart> for BulkWalEntry {
 
 impl BulkPart {
     pub(crate) fn estimated_size(&self) -> usize {
-        self.batch.get_array_memory_size()
+        self.batch
+            .columns()
+            .iter()
+            // If can not get slice memory size, assume 0 here.
+            .map(|c| c.to_data().get_slice_memory_size().unwrap_or(0))
+            .sum()
     }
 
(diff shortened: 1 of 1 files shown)


src/mito2/src/memtable/bulk/part.rs
```diff
@@ -136,7 +136,12 @@ impl TryFrom<&BulkPart> for BulkWalEntry {
 
 impl BulkPart {
     pub(crate) fn estimated_size(&self) -> usize {
-        self.batch.get_array_memory_size()
+        self.batch
+            .columns()
+            .iter()
+            // If can not get slice memory size, assume 0 here.
+            .map(|c| c.to_data().get_slice_memory_size().unwrap_or(0))
```
waynexia, line 143: How about this
```suggestion
            .map(|c| c.get_buffer_memory_size())
```

fengys1996: `get_buffer_memory_size` will have the same problem. 😀

v0y4g3r (approved):
LGTM

PR #6551, opened 2025-07-18
fix: flow mirror cache

discord9:

fix flow mirror insert cache not updated when flownode's address change

## PR Checklist

### DIFF
--- a/src/common/meta/src/cache_invalidator.rs
+++ b/src/common/meta/src/cache_invalidator.rs
@@ -22,6 +22,7 @@ use crate::key::flow::flow_name::FlowNameKey;
 use crate::key::flow::flow_route::FlowRouteKey;
 use crate::key::flow::flownode_flow::FlownodeFlowKey;
 use crate::key::flow::table_flow::TableFlowKey;
+use crate::key::node_address::NodeAddressKey;
 use crate::key::schema_name::SchemaNameKey;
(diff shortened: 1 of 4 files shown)


src/common/meta/src/instruction.rs
```diff
@@ -174,6 +174,8 @@ pub struct UpgradeRegion {
 /// The identifier of cache.
 pub enum CacheIdent {
     FlowId(FlowId),
+    /// Indicate change of address of flownode.
+    FlowNode(u64),
```
killme2008: What about using an explict name such as `FlowNodeAddressChange` like that?

src/common/meta/src/cache_invalidator.rs
```diff
@@ -137,6 +138,12 @@ where
                     let key = FlowInfoKey::new(*flow_id);
                     self.invalidate_key(&key.to_bytes()).await;
                 }
+                CacheIdent::FlowNode(node_id) => {
+                    // other caches doesn't need to be invalidated
+                    // since this is only for flownode address change not id change
+                    let key = NodeAddressKey::with_flownode(*node_id);
```
discord9, line 149: please confirm is this the only key in kv backend cache that need to update for flownode's address?(Since others have a remap mechanism?) @WenyXu

discord9: confirmed

PR #6552, opened 2025-07-18
chore: Add explicit channels to Grpc and Prometheus query contexts

paomian:

I found some query channel is unknow. I'd like to set explicit channels where there are none.

## PR Checklist


src/servers/src/grpc/greptime_handler.rs
```diff
@@ -286,10 +289,12 @@ pub(crate) async fn auth(
     })
 }
 
+/// Creates a new `QueryContext` from the provided request header and extensions.
+/// You need to set channel by yourself if you want to use it in a query context.
 pub(crate) fn create_query_context(
     header: Option<&RequestHeader>,
```
killme2008, line 293: Why not just add a parameter `channel` to this function? Then you don't need to change its signature.

PR #6553, opened 2025-07-18
feat: trigger alter parse

fengys1996:

## PR Checklist

### DIFF
--- a/src/sql/src/parsers/alter_parser/trigger.rs
+++ b/src/sql/src/parsers/alter_parser/trigger.rs
@@ -7,8 +7,8 @@ use crate::error::{self, DuplicateClauseSnafu, InvalidSqlSnafu, Result};
 use crate::parser::ParserContext;
 use crate::parsers::create_parser::trigger::{ANNOTATIONS, LABELS, NOTIFY, ON};
 use crate::statements::alter::trigger::{
-    AnnotationChange, AnnotationOperations, LabelChange, LabelOperations, NotifyChannelChange,
-    NotifyChannelOperations,
+    AlterTrigger, AlterTriggerOperation, AnnotationChange, AnnotationOperations, LabelChange,
+    LabelOperations, NotifyChannelChange, NotifyChannelOperations,
 };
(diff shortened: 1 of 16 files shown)


src/sql/src/parsers/alter_parser/trigger.rs
```diff
@@ -0,0 +1,791 @@
...
+    /// ```sql
+    /// RENAME TO <new_trigger_name>
+    /// ```
+    ///
+    /// ## Parameters
+    ///
+    /// - `is_first_keyword_matched`: indicates whether the first keyword `ON`
+    ///     has been matched.
+    fn parse_rename_to(&mut self, is_first_keyword_matched: bool) -> Result<Ident> {
+        if !is_first_keyword_matched {
```
MichaelScofield, line 269: Why renaming a trigger requires "ON"?

src/sql/src/parsers/alter_parser/trigger.rs
```diff
@@ -0,0 +1,791 @@
...
+    /// ```sql
+    /// LABELS (key1, key2)
+    /// ```
+    ///
+    /// ## Parameters
+    ///
+    /// - `is_first_keyword_matched`: indicates whether the first keyword `LABELS`
+    ///     has been matched.
+    fn parse_trigger_label_names(&mut self, is_first_keyword_matched: bool) -> Result<Vec<String>> {
+        if !is_first_keyword_matched {
```
MichaelScofield, line 301: ditto

src/sql/src/parsers/alter_parser/trigger.rs
fengys1996: Any better suggestions?

fengys1996: Good catch. Wrong document, it should be "RENAME", instead of "ON". Fixed in https://github.com/GreptimeTeam/greptimedb/pull/6553/commits/ea6382d4bdf00034af4d5cbd273f2d488baa43e1.

MichaelScofield (approved):

PR #6560, opened 2025-07-20
fix: ignore target files in make fmt-check

yihong0618:

This patch ignore target dir when `make fmt-check`

to ignore this kind of error

```cli
➜  greptimedb git:(main) make fmt-check
cargo fmt --all -- --check
python3 scripts/check-snafu.py
python3 scripts/check-super-imports.py
Found 'use super::' without leading tab in the following files:
    1 ./target/debug/build/vrl-fe0e1005044935b6/out/src/parser/parser.rs:4 - use super::ast::*;
    2 ./target/debug/build/vrl-fe0e1005044935b6/out/src/parser/parser.rs:5 - use super::template_string::TemplateString;
    3 ./target/debug/build/vrl-fe0e1005044935b6/out/src/parser/parser.rs:6 - use super::lex::*;
    4 ./target/debug/build/vrl-fe0e1005044935b6/out/src/parser/parser.rs:26 - use super::super::ast::*;
    5 ./target/debug/build/vrl-fe0e1005044935b6/out/src/parser/parser.rs:27 - use super::super::template_string::TemplateString;
    6 ./target/debug/build/vrl-fe0e1005044935b6/out/src/parser/parser.rs:28 - use super::super::lex::*;
    7 ./target/debug/build/vrl-fe0e1005044935b6/out/src/parser/parser.rs:42 - use super::__ToTriple;
    8 ./target/debug/build/vrl-fe0e1005044935b6/out/src/parser/parser.rs:15877 - use super::super::ast::*;
    9 ./target/debug/build/vrl-fe0e1005044935b6/out/src/parser/parser.rs:15878 - use 
```

## PR Checklist


scripts/check-super-imports.py
```diff
@@ -24,6 +23,10 @@ def find_rust_files(directory):
         if "test" in root.lower():
             continue
 
+        # Skip the target directory
+        if "target" in root:
```
v0y4g3r: Should we use exact match instead of "contains"?

yihong0618: thought about but just follow the test dir logic

The reason is that the windows name maybe different

./target or .\\target

so just using this logic

yihong0618: thought about but just follow the test dir logic

The reason is that the windows name maybe different

./target or .\\target

so just using this logic

shuiyisong: Maybe we can use [Path](https://docs.python.org/3/library/pathlib.html) to be platform-neutral.

evenyag (approved):

PR #6569, opened 2025-07-23
fix: closee issue #6555 return empty result

### LINKED ISSUE #6555: Cannot find column le when metric doesn't exist

User experience

The promql planner reports an error when querying histogram if the metric doesn't exist

```
tql eval(0, 10, '10s') histogram_quantile(0.99, sum by(pod,instance,le) (rate(greptime_servers_postgres_query_elapsed_bucket{instance=~"xxx"}[1m])));

ERROR 1210 (HY000): (InvalidArguments): Cannot find column le
```

We should tolerate this error in the planner.

### Implementation challenges

We may return an empty result in this case.

Image

yihong0618:

This patch close #6555

fix promql planner reports an error when querying histogram if the metric doesn't exist

return empty result instead

## PR Checklist


tests/cases/standalone/common/promql/simple_histogram.sql
```diff
@@ -187,3 +187,17 @@ insert into histogram4_bucket values
...
+    le STRING,
+    t TIMESTAMP TIME INDEX,
+    v DOUBLE,
+    PRIMARY KEY (pod, instance, le)
+);
+
+-- should return empty result instead of error
+tql eval(0, 10, '10s') histogram_quantile(0.99, sum by(pod,instance, fff) (rate(greptime_servers_postgres_query_elapsed_bucket{instance=~"xxx"}[1m])));
+
+drop table greptime_servers_postgres_query_elapsed_bucket;
```
waynexia: nit: add an empty line

waynexia (approved):
thank you @yihong0618

tests/cases/standalone/common/promql/simple_histogram.result
```diff
@@ -332,3 +332,24 @@ drop table histogram4_bucket;
...
+    le STRING,
+    t TIMESTAMP TIME INDEX,
+    v DOUBLE,
+    PRIMARY KEY (pod, instance, le)
+);
+
+Affected Rows: 0
+
+-- should return empty result instead of error
+tql eval(0, 10, '10s') histogram_quantile(0.99, sum by(pod,instance, fff) (rate(greptime_servers_postgres_query_elapsed_bucket{instance=~"xxx"}[1m])));
```
evenyag: We should not create the table greptime_servers_postgres_query_elapsed_bucket before the query, so we test the case that the table doesn't exist.

We can also add a case that the greptime_servers_postgres_query_elapsed_bucket doesn't have the column `le`.

evenyag (approved):
LGTM

PR #6574, opened 2025-07-23
feat: fallback when failed to push down using `DistPlanner`

discord9:

add a fallback to only push down table scan when `PlanRewriter` failed to push down as many node as possible, this is useful when we encounter a previously unknown bug in `PlanRewriter` or encounter known but unfixed bugs like alias bug to provide at least some level of usability  

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/analyzer.rs
+++ b/src/query/src/dist_plan/analyzer.rs
@@ -16,6 +16,7 @@ use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
 use common_telemetry::debug;
+use datafusion::config::{ConfigExtension, ExtensionOptions};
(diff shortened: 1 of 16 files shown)


src/query/src/dist_plan/analyzer.rs
```diff
@@ -64,15 +65,35 @@ impl AnalyzerRule for DistPlannerAnalyzer {
             .rewrite(plan, &optimizer_context)?
             .data;
 
+        let result = match self.try_push_down(plan.clone()) {
+            Ok(plan) => plan,
+            Err(err) => {
+                common_telemetry::error!(err; "Failed to push down plan, using fallback plan rewriter for plan: {plan}");
```
evenyag: Maybe in warn level?

PR #6587, opened 2025-07-24
fix: close issue #6586 make pg also show error as mysql

### LINKED ISSUE #6586: different behavior for pg and mysql

User experience

when I try issue #6177 which is very hard to fix

I found the error message are different for pg and mysql
pg does not show all error only ERROR:  External error: External error

```cli
mysql -h 127.0.0.1 -P 4002 -e "select * from monitor where val in (select val from monitor) and ts >= now() - interval 30 minutes"
ERROR 1815 (HY000) at line 1: (EngineExecuteQuery): Error during planning: No installed planner was able to convert the custom node to an execution plan: MergeScanLogicalPlan { input: Projection(Projection { expr: [Column(Column { relation: Some(Bare { table: "monitor" }), name: "val" })], input: TableScan(TableScan { table_name: Bare { table: "monitor" }, source: "...", projection: None, projected_schema: DFSchema { inner: Schema { fields: [Field { name: "ts", data_type: Timestamp(Millisecond, None), nullable: false, dict_id: 0, dict_
➜  ~ psql -h 127.0.0.1 -p 4003 -U public  -d public -c "select * from monitor where val in (select val from monitor) and ts >= now() - interval 30 minutes"
ERROR:  External error: External error
```

### Implementation challenges

_No response_

yihong0618:

This patch try to fix #6586 but there are quite some same logic left
do not know need to fix or not.

after this patch
```cli
➜  ~ psql -h 127.0.0.1 -p 4003 -U public  -d public -c "select * from monitor where val in (select val from monitor) and ts >= now() - interval 30 minutes"
ERROR:  External error: External error3 0: Region query error, at src/frontend/src/instance/region_query.rs:57:14
1: Failed to query, at src/frontend/src/instance/region_query.rs:83:14
2: External error2, at src/frontend/src/instance/standalone.rs:106:14
3: Failed to execute logical plan, at src/datanode/src/region_server.rs:1196:14
4: <transparent>, at src/query/src/datafusion.rs:313:29
5: Plan("No installed planner was able to convert the custom node to an execution plan: MergeScanLogicalPlan { input: Projection(Projection { expr: [Column(Column { relation: Some(Bare { table: \"monitor\" }), name: \"val\" })], input: TableScan(TableScan { table_name: Bare { table: \"monitor\" }, source: \"...\", projection: None, projected_schema: DFSchema { inner: Schema { fields: [Field { name: \"ts\", data_type: Timestamp(Millisecond, None), nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {\"greptime:time_index\": \"true\"} }, Field { name: \"val\", data_type: Int32, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }], metadata: {\"greptime:version\": \"0\"} }, field_qualifiers: [Some(Bare { table: \"monitor\" }), Some(Bare { table: \"monitor\" })], functional_dependencies: FunctionalDependencies { deps: [] } }, filters: [], fetch: None, .. }), schema: DFSchema { inner: Schema { fields: [Field { name: \"val\", data_type: Int32, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }], metadata: {\"greptime:version\": \"0\"} }, field_qualifiers: [Some(Bare { table: \"monitor\" })], functional_dependencies: FunctionalDependencies { deps: [] } } }), is_placeholder: false, partition_cols: [] }")
➜  ~ mysql -h 127.0.0.1 -P 4002 -e "select * from monitor where val in (select val from monitor) and ts >= now() - interval 30 minutes"
ERROR 1815 (HY000) at line 1: (EngineExecuteQuery): Error during planning: No installed planner was able to convert the custom node to an execution plan: MergeScanLogicalPlan { input: Projection(Projection { expr: [Column(Column { relation: Some(Bare { table: "monitor" }), name: "val" })], input: TableScan(TableScan { table_name: Bare { table: "monitor" }, source: "...", projection: None, projected_schema: DFSchema { inner: Schema { fields: [Field { name: "ts", data_type: Timestamp(Millisecond, None), nullable: false, dict_id: 0, dict_
```

## PR Checklist

### DIFF
--- a/src/common/recordbatch/src/error.rs
+++ b/src/common/recordbatch/src/error.rs
@@ -44,7 +44,7 @@ pub enum Error {
         source: datatypes::error::Error,
     },
 
-    #[snafu(display("External error"))]
+    #[snafu(display("External error3 {:?}", source))]
     External {
         #[snafu(implicit)]
         location: Location,


sunng87:
I think the correct fix is to use `e.output_msg()` for all postgres error just like mysql's `handle_err` in `writer.rs`. The pgwire API will take standard `e.message()` when error returned as `PgError::ApiError`.

We can create a new function say `convert_err` that takes a `impl ErrorExt` and return `Response::Error` in `postgres/handler.rs`. The error we need to deal with is from  `recordbatches_to_query_response`L182

yihong0618:
> convert_err

copy that, thanks

sunng87:
We can reuse code from L139-154

yihong0618:
> I think the correct fix is to use `e.output_msg()` for all postgres error just like mysql's `handle_err` in `writer.rs`. The pgwire API will take standard `e.message()` when error returned as `PgError::ApiError`.
> 
> We can create a new function say `convert_err` that takes a `impl ErrorExt` and return `Response::Error` in `postgres/handler.rs`. The error we need to deal with is from `recordbatches_to_query_response`L182

the error is from error future~

sunng87:
I see. We may need to return early with executing recordbatches. Let me pickup your modification and continue it.

sunng87:
I'm afraid we cannot do early return because we will have to hold all record batches in memory in that way. That best way is to reassemble `ApiError` just like in this patch.

yihong0618:
> I'm afraid we cannot do early return because we will have to hold all record batches in memory in that way. That best way is to reassemble `ApiError` just like in this patch.

cool

src/servers/src/postgres/handler.rs
```diff
@@ -111,6 +112,38 @@ where
...
+    }
+    Response::Error(Box::new(
+        PgErrorCode::from(status_code).to_err_info(e.output_msg()),
+    ))
+}
+
+fn convert_error_future<E: ErrorExt>(
+    e: E,
+) -> BoxStream<'static, Result<Vec<datatypes::prelude::Value>, PgWireError>> {
+    let response = convert_err(e);
```
sunng87: here we don't need to convert to response then convert to error back. Just convert e to

`PgWireError::UserError(Box::new(PgErrorCode::from(e.status_code()).to_err_info(e.output_msg()));`

sunng87:
Just a moment, I'm working on my changes

sunng87:
I'm good. Sorry for some misleading comment. Sometimes I need to dive up for best solution.

sunng87 (approved):

yihong0618:
> I'm good. Sorry for some misleading comment. Sometimes I need to dive up for best solution.

learn something how to handle error in Rust.  thanks ~

yihong0618:
and there are still some of this kind error, are they need to change the new error type?

sunng87:
@yihong0618  Yes, it's better to update those in extended query as well. Good catch!
Do you want to update all of them?

yihong0618:
> @yihong0618 Yes, it's better to update those in extended query as well. Good catch! Do you want to update all of them?

of course

yihong0618:
> @yihong0618 Yes, it's better to update those in extended query as well. Good catch! Do you want to update all of them?

image

Done and thress similar left but they are datafusion error not sure if need to change

src/servers/src/postgres/types.rs
```diff
@@ -351,6 +352,23 @@ fn encode_array(
     }
 }
 
+pub fn convert_err(e: impl ErrorExt) -> PgWireError {
```
sunng87: It's just a little weird to have this in types.rs. We can create a new module like utils or common in postgres for this.

yihong0618: copy

fengjiachun (approved):
LGTM

PR #6589, opened 2025-07-24
feat: try all underlying streams when polling WindowedSortStream

waynexia:

`WindowedSortStream` used to poll underlying streams in sequence, only start polling next one when current is finished. This may bring extra wait time if data between each input stream is not evenly distributed. This will not only let a slow stream hang the `WindowedSortStream`, but also hang other input streams because their ready content is not consumed. 

This patch changed the behavior to try all streams until it finds one available, which should reduce the TTFB and overall execution time by reducing wait time.

If this turns out to be effective, we can do a similar optimization to the merge scan stream

## PR Checklist

### DIFF
--- a/src/query/src/window_sort.rs
+++ b/src/query/src/window_sort.rs
@@ -387,6 +387,7 @@ impl WindowedSortStream {
         &mut self,
         cx: &mut Context<'_>,
     ) -> Poll<Option<datafusion_common::Result<DfRecordBatch>>> {
+        let mut rotate_count = 0;
         while let Some(merge_stream) = &mut self.merge_stream.front_mut() {
             match merge_stream.as_mut().poll_next(cx) {
                 Poll::Ready(Some(Ok(batch))) => {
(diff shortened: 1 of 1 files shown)


src/query/src/window_sort.rs
```diff
@@ -417,7 +418,14 @@ impl WindowedSortStream {
...
                 Poll::Pending => {
-                    return Poll::Pending;
+                    if rotate_count < self.merge_stream.len() {
+                        // try consume next stream first
+                        rotate_count += 1;
+                        self.merge_stream.rotate_left(1);
+                        continue;
+                    } else {
+                        return Poll::Pending;
+                    }
```
evenyag, line 428: I remember that each stream is a working set result. The windowed sort stream can't return batches from the next working set until the current working set is exhausted, ensuring the output order is sorted.

So we can't return the batch directly. What do you think? @discord9

discord9: > I remember that each stream is a working set result. The windowed sort stream can't return batches from the next working set until the current working set is exhausted, ensuring the output order is sorted.
> 
> So we can't return the batch directly. What do you think? @discord9

Yes, first all the input stream for the `merge_stream` is actually `MemoryStream` so the data is always ready(or input stream is finished), second `merge_stream`'s order matter as they are a result of a working set(a given input range that's fully scaned)'s merge sort result so using different order(with stream that can pending, `MemoryStream` doesn't pend so test passed) will cause incorrect result,
although a optimize possible here is maybe a dynamic merge sort which can add new input stream with newer lower bound somehow and emit first row earlier

waynexia: got it

PR #6593, opened 2025-07-28
feat: HTTP API to activate/deactive heap prof (activate by default)

evenyag:

The Docker image enables heap profile by default, but doesn't activate it
```
MALLOC_CONF="prof:true,prof_active:false"
```

The DB side activates heap profile by default and provides the following HTTP APIs
- `/debug/prof/mem/activate` activate the profiling
- `/debug/prof/mem/deactivate` deactivate the profiling
- `/debug/prof/mem/status` get whether the profiling is active

It also adds a config to disable the heap profiling:
```
[memory]
enable_heap_profiling = false
```

This is similar to tikv, which enables profiling by default.
https://github.com/tikv/tikv/blob/ed504baa35c36fa2f05ee1e5bb8f5e3f60da46d6/src/config/mod.rs#L3483-L3486

According to its document, the performance overhead is acceptable.
> Whether enables the heap profiling which may have a bit performance overhead about 2% for
> the default sample rate.

## PR Checklist


waynexia (approved):

src/servers/src/http/mem_prof.rs
```diff
@@ -56,6 +56,30 @@ pub async fn mem_prof_handler(
     Ok((StatusCode::OK, dump))
 }
 
+#[cfg(feature = "mem-prof")]
```
MichaelScofield, line 59: not related to this PR but I think it's time to remove this feature since it's default to true now and is controlled by configuration

PR #6599, opened 2025-07-28
feat: poll result stream more often

discord9:

poll result stream more often when window sort and polling for input stream, this should reduce time to first bytes time for many cases, before this window sort would wait for first pending of input stream then to emit result

## PR Checklist

### DIFF
--- a/src/query/src/window_sort.rs
+++ b/src/query/src/window_sort.rs
@@ -445,31 +445,24 @@ impl WindowedSortStream {
         // consume input stream
         while !self.is_terminated {
             // then we get a new RecordBatch from input stream
-            let new_input_rbs = match self.input.as_mut().poll_next(cx) {
-                Poll::Ready(Some(Ok(batch))) => {
-                    Some(split_batch_to_sorted_run(batch, &self.expression)?)
-                }
+            let SortedRunSet {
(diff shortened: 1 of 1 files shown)


src/query/src/window_sort.rs
```diff
@@ -564,8 +557,18 @@ impl WindowedSortStream {
...
+
+            // poll result stream again to see if we can emit more results
+            match self.poll_result_stream(cx) {
+                Poll::Ready(None) => {
+                    if self.is_terminated {
+                        return Poll::Ready(None);
+                    }
+                }
+                x => return x,
+            };
```
evenyag, line 569: Is it safe to poll the result stream here?

evenyag, line 569: Is it safe to poll the result stream here?

discord9: > Is it safe to poll the result stream here?

yes, and necessary too as above code might just create a new merge sort stream, and it's always safe to poll result stream as it's a fifo stream of result so order is always correct, polling early here is just output early and no side effect whatsoever

evenyag (approved):

waynexia (approved):

PR #6602, opened 2025-07-29
feat: add partial truncate

discord9:

https://github.com/GreptimeTeam/greptime-proto/pull/262 proto def

add partial truncate which remove files in a list, using
```sql
TRUNCATE TABLE foo FILE RANGE (start, end), (start2, end2) ... 
```
to truncate table with time ranges, notice this doesn't promise to truncate all data in that ranges, just trying it's best to remove files that are fully contained in that range and could left some data un-truncated if they happen to be in a file that isn't a subset of truncate ranges

## PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5310,7 +5310,7 @@ dependencies = [
 [[package]]
 name = "greptime-proto"
 version = "0.1.0"
-source = "git+https://github.com/GreptimeTeam/greptime-proto.git?rev=91c3d7b97a2850c014aa5ce4ffa4caeb6b918446#91c3d7b97a2850c014aa5ce4ffa4caeb6b918446"
+source = "git+https://github.com/GreptimeTeam/greptime-proto.git?rev=ccfd4da48bc0254ed865e479cd981a3581b02d84#ccfd4da48bc0254ed865e479cd981a3581b02d84"
 dependencies = [
  "prost 0.13.5",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -140,7 +140,7 @@ etcd-client = "0.14"
 fst = "0.4.7"
 futures = "0.3"
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "91c3d7b97a2850c014aa5ce4ffa4caeb6b918446" }
+greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "ccfd4da48bc0254ed865e479cd981a3581b02d84" }
 hex = "0.4"
 http = "1"
 humantime = "2.1"


src/mito2/src/region/version.rs
```diff
@@ -210,13 +209,24 @@ impl VersionControl {
...
+            } => Arc::new(
+                VersionBuilder::new(version.metadata.clone(), new_mutable)
+                    .flushed_entry_id(truncated_entry_id)
+                    .flushed_sequence(truncated_sequence)
+                    .truncated_entry_id(Some(truncated_entry_id))
+                    .build(),
+            ),
+            TruncateKind::Partial { files_to_remove } => Arc::new(
+                VersionBuilder::from_version(version)
+                    .memtables(MemtableVersion::new(new_mutable))
```
evenyag: This may remove existing memtables.

src/api/src/helper.rs
```diff
@@ -1076,6 +1076,62 @@ pub fn value_to_grpc_value(value: Value) -> GrpcValue {
...
+        .map(|r| {
+            (
+                Timestamp::new(r.start, time_unit),
+                Timestamp::new(r.end, time_unit),
+            )
+        })
+        .collect())
+}
+
+pub fn to_pb_time_ranges(time_ranges: &[(Timestamp, Timestamp)]) -> v1::TimeRanges {
```
evenyag: Document that input time ranges must have the same unit.

evenyag: Do we need a debug assertion on units?

src/common/meta/src/rpc/ddl.rs
```diff
@@ -878,6 +890,29 @@ impl TryFrom<TruncateTableTask> for PbTruncateTableTask {
...
+                                start: start.value(),
+                                end: end.value(),
+                            })
+                            .collect();
+                        Some(api::v1::TimeRanges {
+                            time_unit: to_pb_time_unit(time_unit) as i32,
+                            time_ranges,
+                        })
+                    }
+                },
```
evenyag: Can we use to_pb_time_ranges?

tests/cases/standalone/common/truncate/truncate.result
```diff
@@ -44,6 +44,53 @@ SELECT ts, host, cpu, memory FROM monitor ORDER BY ts;
...
+| 2023-09-20T13:47:32 | host3 | 66.6 | 1024.0 |
+| 2023-09-20T13:47:34 | host1 | 77.7 | 2048.0 |
+| 2023-09-20T13:47:34 | host2 | 77.7 | 2048.0 |
+| 2023-09-20T13:47:34 | host3 | 77.7 | 2048.0 |
+| 2023-09-20T13:47:36 | host1 | 88.8 | 4096.0 |
+| 2023-09-20T13:47:36 | host2 | 88.8 | 4096.0 |
+| 2023-09-20T13:47:36 | host3 | 88.8 | 4096.0 |
++---------------------+-------+------+--------+
+
+TRUNCATE monitor RANGE (0, 1700000000000);
```
evenyag: Missing test cases for multiple ranges.

src/sql/src/parsers/truncate_parser.rs
```diff
@@ -153,5 +380,29 @@ mod tests {
...
+            "result is: {result:?}"
+        );
+
+        let sql = "TRUNCATE TABLE foo RANGE ()";
+        let result =
+            ParserContext::create_with_dialect(sql, &GreptimeDbDialect {}, ParseOptions::default());
+        assert!(
+            result.is_err() && format!("{result:?}").contains("expected: 'a timestamp value'"),
+            "result is: {result:?}"
+        );
```
evenyag: Add more invalid cases (missing `,`, misssing `)`).

src/sql/src/parsers/truncate_parser.rs
```diff
@@ -41,7 +43,93 @@ impl ParserContext<'_> {
             }
         );
 
-        Ok(Statement::TruncateTable(TruncateTable::new(table_ident)))
+        let have_range = self.parser.parse_keyword(Keyword::RANGE);
+        // if no range is specified, we just truncate the table
+        if !have_range {
+            return Ok(Statement::TruncateTable(TruncateTable::new(table_ident)));
+        }
```
evenyag: Should we mention that we only truncate files in the sql syntax?
```
truncate table xxx file range (), ();
```

src/sql/src/statements/truncate.rs
```diff
@@ -71,5 +126,25 @@ TRUNCATE TABLE t1"#,
                 unreachable!();
             }
         }
+
+        let sql = r"truncate table t1 range (1,2);";
```
evenyag: Add a case for multiple ranges?

src/api/src/helper.rs
```diff
@@ -1076,6 +1076,81 @@ pub fn value_to_grpc_value(value: Value) -> GrpcValue {
...
+    time_ranges.iter().fold(None, |acc, (start, _end)| {
+        {
+            if let Some(acc) = acc {
+                if acc != start.unit() {
+                    panic!("All time ranges must have the same time unit");
+                }
+            }
+        }
+        Some(start.unit())
+    });
```
evenyag: Maybe we can do this only when `cfg!(debug_assertions)` is true.

src/api/src/helper.rs
```diff
@@ -1079,6 +1079,82 @@ pub fn value_to_grpc_value(value: Value) -> GrpcValue {
...
```
MichaelScofield:
```suggestion
    debug_assert!(time_ranges.windows(2).all(|x| x[0].unit() == x[1].unit()));
```

src/api/src/helper.rs
```diff
@@ -1079,6 +1079,82 @@ pub fn value_to_grpc_value(value: Value) -> GrpcValue {
...
+    let mut pb_time_ranges = v1::TimeRanges {
+        // default time unit is Millisecond
+        time_unit: v1::TimeUnit::Millisecond as i32,
+        time_ranges: vec![],
```
MichaelScofield:
```suggestion
        time_ranges: Vec::with_capacity(time_ranges.len()),
```

src/mito2/src/worker/handle_truncate.rs
```diff
@@ -28,24 +29,65 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+                RegionTruncate {
+                    region_id,
+                    kind: TruncateKind::All {
+                        truncated_entry_id,
+                        truncated_sequence,
+                    },
+                }
+            }
+            RegionTruncateRequest::ByTimeRanges { time_ranges } => {
+                debug!(
```
MichaelScofield:
```suggestion
                info!(
```

src/mito2/src/worker/handle_truncate.rs
```diff
@@ -28,24 +29,65 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+                // find all files that are fully contained in the time ranges
+                let mut files_to_truncate = Vec::new();
+                for level in version_data.version.ssts.levels() {
+                    for file in level.files() {
+                        let file_time_range = file.time_range();
+                        // TODO(discord9): This is a naive way to check if is contained, we should
+                        // optimize it later.
+                        let is_subset = time_ranges.iter().any(|(start, end)| {
+                            file_time_range.0 >= *start && file_time_range.1 <= *end
+                        });
```
MichaelScofield, line 70: I think you may want to merge contiguous time ranges first, otherwise some files are not selected. For example, if the input time ranges are [(1, 3), (3, 5)], then the file with time range (2, 4) will not be considered as "subset" here, while it actually should.

src/operator/src/statement.rs
```diff
@@ -530,6 +535,84 @@ impl StatementExecutor {
...
+            .as_timestamp()
+            .with_context(|| UnexpectedSnafu {
+                violated: format!(
+                    "Table {}'s time index column must be a timestamp type, found: {:?}",
+                    table_name, time_index_dt
+                ),
+            })?
+            .unit();
+
+        let mut time_ranges = vec![];
```
MichaelScofield:
```suggestion
        let mut time_ranges = Vec::with_capacity(sql_values_time_range.len());
```

src/sql/src/parsers/truncate_parser.rs
```diff
@@ -41,7 +43,95 @@ impl ParserContext<'_> {
...
+                    continue;
+                }
+                _ => Err(ParserError::ParserError(format!(
+                    "Expected a comma or end of statement at {}",
+                    self.parser.peek_token().span.start
+                )))
+                .context(UnexpectedSnafu {
+                    expected: "a comma or end of statement",
+                    actual: self.peek_token_as_string(),
+                })?,
```
MichaelScofield:
```suggestion
                _ => return UnexpectedToken {
                    expected: "a comma or end of statement",
                    actual: self.peek_token_as_string(),
                }.fail();
```

src/sql/src/statements/truncate.rs
```diff
@@ -15,30 +15,85 @@
...
+        let mut is_first = true;
+        for (start, end) in &self.time_ranges {
+            if is_first {
+                is_first = false;
+            } else {
+                write!(f, ", ")?;
+            }
+            write!(f, "({}, {})", start, end)?;
+        }
+        Ok(())
```
MichaelScofield:
```suggestion
        write!(f, "{}", self
            .time_ranges
            .iter()
            .map(|x| format!("({}, {})", x.start, x.end))
            .join(", ")
        )
```

src/api/src/helper.rs
MichaelScofield: I don't think this check should be level debug, it's by input.

src/mito2/src/worker/handle_truncate.rs
discord9: > I think you may want to merge contiguous time ranges first, otherwise some files are not selected. For example, if the input time ranges are [(1, 3), (3, 5)], then the file with time range (2, 4) will not be considered as "subset" here, while it actually should.

don't want to over-complex stuff here, this `truncate foo file range` thing will be only basically used by flow, so complex interval merging stuff should happen there not here for better debug&simplicity

MichaelScofield: ok, then this should be added to the comments

MichaelScofield (approved):

PR #6610, opened 2025-07-29
feat: support `__schema__` and `__database__` in Prom Remote Read

waynexia:

Support `__schema__` and `__database__` special labels in Prometheus Remote Read service.

## PR Checklist

### DIFF
--- a/src/servers/src/prom_store.rs
+++ b/src/servers/src/prom_store.rs
@@ -81,7 +81,7 @@ pub fn table_name(q: &Query) -> Result<String> {
 pub fn extract_schema_from_read_request(request: &ReadRequest) -> Option<String> {
     for query in &request.queries {
         for matcher in &query.matchers {
-            if matcher.name == SCHEMA_LABEL {
+            if matcher.name == SCHEMA_LABEL && matcher.r#type == MatcherType::Eq as i32 {
                 return Some(matcher.value.clone());
             }
         }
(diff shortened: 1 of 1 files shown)


src/servers/src/http/prom_store.rs
```diff
@@ -53,6 +53,16 @@ pub const DEFAULT_ENCODING: &str = "snappy";
...
 
+/// Extract schema from TablesBuilder. Returns the first schema found from any PromCtx.
+/// Prioritizes non-None schemas.
+fn extract_schema_from_tables_builder(tables_builder: &TablesBuilder) -> Option<String> {
+    tables_builder
+        .tables
+        .keys()
+        .find_map(|prom_ctx| prom_ctx.schema.as_ref())
+        .cloned()
+}
```
sunng87: Please double check this logic, in a batch of metrics sent from prometheus remote write, if only one series contains `__database__`, what will be the behavior? Does it mean the whole batch will be written to `__database__`?

shuiyisong: This setting probably won't take effect anyway, the schema of query context is reset in https://github.com/GreptimeTeam/greptimedb/blob/2f9a10ab745e34b573d517e05233fe572e1b6d6d/src/servers/src/http/prom_store.rs#L145

waynexia: Ahh, ingestion with this special label is already supported. I only need to implement it for remote read

src/servers/src/prom_store.rs
```diff
@@ -73,6 +76,29 @@ pub fn table_name(q: &Query) -> Result<String> {
         })
 }
 
+/// Extract schema from remote read request. Returns the first schema found from any query's matchers.
+/// Prioritizes __schema__ over __database__ labels.
+pub fn extract_schema_from_read_request(request: &ReadRequest) -> Option<String> {
+    for query in &request.queries {
+        for matcher in &query.matchers {
+            if matcher.name == SCHEMA_LABEL {
```
killme2008: Need to verify the matcher type: only if it's `Eq`, use the value as the schema name.

sunng87 (approved):

killme2008 (approved):
LGTM

Issue #6616, opened 2025-07-30
OTLP metric follow-up tasks


shuiyisong:
### What type of enhancement is this?

API improvement

### What does the enhancement do?

This is a follow-up to-do list for #6543 
- [ ] support allow-list and ignore-list options in HTTP header 
- [ ] add unit to name normalize function
- [ ] save `AggregationTemporality` but filter them out in PromQL results

PR #6617, opened 2025-07-30
chore(otlp_metric): support attr list in header opts

shuiyisong:

#6616 

This is a follow-up PR of #6543. This PR mainly does:
1. add `x-greptime-otlp-metric-resource-attrs-list` header to include or exclude resource attrs.
2. change default behavior to persist none resource attributes, to align with [Prometheus's behavior](https://prometheus.io/docs/guides/opentelemetry/#promoting-resource-attributes).

## PR Checklist

### DIFF
--- a/src/servers/src/http/header.rs
+++ b/src/servers/src/http/header.rs
@@ -62,8 +62,10 @@ pub mod constants {
     // OTLP headers
     pub const GREPTIME_OTLP_METRIC_PROMOTE_ALL_RESOURCE_ATTRS_HEADER_NAME: &str =
         "x-greptime-otlp-metric-promote-all-resource-attrs";
-    pub const GREPTIME_OTLP_METRIC_RESOURCE_ATTRS_LIST_HEADER_NAME: &str =
-        "x-greptime-otlp-metric-resource-attrs-list";
+    pub const GREPTIME_OTLP_METRIC_PROMOTE_RESOURCE_ATTRS_HEADER_NAME: &str =
+        "x-greptime-otlp-metric-promote-resource-attrs";
+    pub const GREPTIME_OTLP_METRIC_IGNORE_RESOURCE_ATTRS_HEADER_NAME: &str =
+        "x-greptime-otlp-metric-ignore-resource-attrs";
     pub const GREPTIME_OTLP_METRIC_PROMOTE_SCOPE_ATTRS_HEADER_NAME: &str =
         "x-greptime-otlp-metric-promote-scope-attrs";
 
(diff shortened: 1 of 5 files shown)


src/session/src/protocol_ctx.rs
```diff
@@ -33,7 +35,12 @@ impl ProtocolCtx {
 #[derive(Debug, Clone, Default)]
 pub struct OtlpMetricCtx {
     pub promote_all_resource_attrs: bool,
+    pub resource_attrs: HashSet<String>,
     pub promote_scope_attrs: bool,
     pub with_metric_engine: bool,
+    // If the user uses OTLP metrics ingestion before v0.16, it uses the old path.
```
paomian: we can change them to doc.

src/servers/src/http/header.rs
```diff
@@ -62,6 +62,8 @@ pub mod constants {
     // OTLP headers
     pub const GREPTIME_OTLP_METRIC_PROMOTE_ALL_RESOURCE_ATTRS_HEADER_NAME: &str =
         "x-greptime-otlp-metric-promote-all-resource-attrs";
+    pub const GREPTIME_OTLP_METRIC_RESOURCE_ATTRS_LIST_HEADER_NAME: &str =
+        "x-greptime-otlp-metric-resource-attrs-list";
```
sunng87: follow the naming pattern, use `promote` as the verb

src/servers/src/otlp/metrics.rs
```diff
@@ -38,33 +37,7 @@ const INSTANCE_KEY: &str = "instance";
...
-    "k8s.container.name",
-    "k8s.cronjob.name",
-    "k8s.daemonset.name",
-    "k8s.deployment.name",
-    "k8s.job.name",
-    "k8s.namespace.name",
-    "k8s.pod.name",
-    "k8s.replicaset.name",
-    "k8s.statefulset.name",
-];
```
sunng87, line 63: Wait, we still need this list as default promote list.

sunng87: We will always promote these items, until there is a request. When user specify the promote resource attirbute lists, we promote those in additional to these.

src/servers/src/http/header.rs
shuiyisong: Do we use this `x-greptime-otlp-metric-resource-attrs-list` as an exclude list when `promote-all` is checked like Prometheus?

sunng87: Is prometheus using the same config item for include and exclude? I think we should use two different item to avoid confusion.

sunng87 (approved):

paomian (approved):

Issue #6505, opened 2025-07-10
Bulk Memtable with full primary key support


evenyag:
### What problem does the new feature solve?

In https://github.com/GreptimeTeam/greptimedb/issues/5803, we need a memtable that supports bulk ingestion and primary keys.

In https://github.com/GreptimeTeam/greptimedb/issues/6078, we also need a memtable that can support encoding Parquet in plain format.

None of the current memtable implementations support these features. They also don't perform well under high cardinality primary keys. We want to implement a memtable that works well under bulk ingestion and high cardinality primary keys.

### What does the feature do?

Implement a new memtable that is similar to the `SimpleBulkMemtable`, but supports bulk ingestion.

### Implementation challenges

1. How to support primary keys
2. How to provide realtime visibility

killme2008:
It is done? @evenyag  
If so, we can close #5803 too

evenyag:
We still need some optimization and I'd like to track them in this issue.

evenyag:
We can track further optimizations in other issues.

PR #6620, opened 2025-07-31
feat: Implement a converter to converts KeyValues into BulkPart

evenyag:

This PR implements a BulkPartConverter and necessary utilities to convert KeyValues into BulkPart.
- The BulkPartConverter can store primary keys additionally, it uses DictionaryArray for string columns in memory.
- It adds a use_bulk_insert method to the memtable builder to control whether to use bulk insert with BulkPartConverter.
- It defines a new function `to_flat_sst_arrow_schema()` to generate the new schema for the format with the primary key.

It introduced a new format that contains both the primary key and the raw primary key columns (It flattens the primary key, thus the flat format). This new format stores all key columns according to their position in the primary key before fields.

## PR Checklist

### DIFF
--- a/src/mito2/src/memtable/bulk/part.rs
+++ b/src/mito2/src/memtable/bulk/part.rs
@@ -24,36 +24,39 @@ use bytes::Bytes;
 use common_grpc::flight::{FlightDecoder, FlightEncoder, FlightMessage};
 use common_recordbatch::DfRecordBatch as RecordBatch;
 use common_time::timestamp::TimeUnit;
-use datafusion::arrow::array::{TimestampNanosecondArray, UInt64Builder};
 use datatypes::arrow;
 use datatypes::arrow::array::{
-    Array, ArrayRef, BinaryBuilder, DictionaryArray, TimestampMicrosecondArray,
-    TimestampMillisecondArray, TimestampSecondArray, UInt32Array, UInt64Array, UInt8Array,
-    UInt8Builder,
+    Array, ArrayRef, BinaryBuilder, BinaryDictionaryBuilder, DictionaryArray,
+    TimestampMicrosecondArray, TimestampMillisecondArray, TimestampNanosecondArray,
+    TimestampSecondArray, UInt32Array, UInt64Array, UInt64Builder, UInt8Array, UInt8Builder,
 };
-use datatypes::arrow::compute::TakeOptions;
-use datatypes::arrow::datatypes::SchemaRef;
+use datatypes::arrow::compute::{SortColumn, SortOptions, TakeOptions};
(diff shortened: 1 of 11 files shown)


src/mito2/src/memtable/bulk/part.rs
```diff
@@ -209,6 +217,284 @@ impl BulkPart {
...
+impl PrimaryKeyColumnBuilder {
+    /// Appends a value to the builder.
+    fn push_value_ref(&mut self, value: ValueRef) -> Result<()> {
+        match self {
+            PrimaryKeyColumnBuilder::StringDict(builder) => {
+                if let Ok(Some(s)) = value.as_string() {
+                    // We know the value is a string.
+                    builder.append_value(s);
+                } else {
+                    builder.append_null();
```
discord9, line 240: is fail instead of push null here a better option to discover bugs early?

src/mito2/src/memtable/bulk/part.rs
```diff
@@ -209,6 +217,284 @@ impl BulkPart {
...
+            region_metadata.primary_key_encoding,
+            primary_key_codec.encoding()
+        );
+
+        if region_metadata.primary_key_encoding == PrimaryKeyEncoding::Sparse {
+            // We can't get primary key columns now.
+            store_primary_key_columns = false;
+        }
+
+        let primary_key_column_builders = if store_primary_key_columns {
```
discord9: how about just `store_primary_key_columns && region_metadata.primary_key_encoding != PrimaryKeyEncoding::Sparse`?

src/mito2/src/memtable/bulk/part.rs
```diff
@@ -1084,4 +1368,226 @@ mod tests {
             1,
         );
     }
+
+    #[test]
+    fn test_bulk_part_converter_append_and_convert() {
+        let metadata = metadata_for_test();
```
discord9, line 1379: maybe test using sparse mode in metadata as well?

discord9 (approved):
others LGTM

waynexia (approved):

src/mito2/src/memtable/bulk/part.rs
```diff
@@ -209,6 +217,281 @@ impl BulkPart {
...
+                })?
+                .as_binary()
+                .context(DataTypeMismatchSnafu)?
+            {
+                self.key_array_builder
+                    .append(encoded)
+                    .context(ComputeArrowSnafu)?;
+            } else {
+                self.key_array_builder
+                    .append("")
```
v0y4g3r, line 355: Can we use append_null instead?

evenyag: This column is expected to be not null. And the value won't be null actually.

PR #6622, opened 2025-07-31
feat:LRU cache(with TTL) for prepared statements

### LINKED ISSUE #1790: LRU cache(with TTL) for prepared statements

Performance, Tech debt reduction

GreptimeDB caches the query or logical plan for prepared statements, It's a `HashMap` with `RwLock` in 

https://github.com/GreptimeTeam/greptimedb/blob/128c6ec98c57996de08a12e958feb9695afa2204/src/servers/src/mysql/handler.rs#L54

We can use something like LRU cache which supports TTL to replace the naive implement.

### Implementation challenges

_No response_

sunheyi6:

close #1790 
## PR Checklist

### DIFF
--- a/src/servers/Cargo.toml
+++ b/src/servers/Cargo.toml
@@ -130,6 +130,7 @@ urlencoding = "2.1"
 uuid.workspace = true
 vrl.workspace = true
 zstd.workspace = true
+moka = { workspace = true, features = ["sync"] }
 
 [target.'cfg(not(windows))'.dependencies]
 tikv-jemalloc-ctl = { version = "0.6", features = ["use_std", "stats"] }
@@ -153,6 +154,7 @@ table.workspace = true
 tempfile = "3.0.0"
 tokio-postgres = "0.7"
 tokio-postgres-rustls = "0.12"
+moka = { workspace = true, features = ["sync"] }
 
 [target.'cfg(unix)'.dev-dependencies]
 pprof = { version = "0.14", features = ["criterion", "flamegraph"] }
--- a/src/servers/src/mysql/handler.rs
+++ b/src/servers/src/mysql/handler.rs
@@ -29,6 +29,8 @@ use datafusion_common::ParamValues;
 use datafusion_expr::LogicalPlan;
 use datatypes::prelude::ConcreteDataType;
 use itertools::Itertools;
+use loki_proto::stats::Cache;
+use moka::sync::Cache;
 use opensrv_mysql::{
     AsyncMysqlShim, Column, ErrorKind, InitWriter, ParamParser, ParamValue, QueryResultWriter,
     StatementMetaWriter, ValueInner,
@@ -80,7 +82,7 @@ pub struct MysqlInstanceShim {
     salt: [u8; 20],
     session: SessionRef,
     user_provider: Option<UserProviderRef>,
-    prepared_stmts: Arc<RwLock<HashMap<String, SqlPlan>>>,
+    prepared_stmts: Arc<Cache<String, SqlPlan>>,
     prepared_stmts_counter: AtomicU32,
     process_id: u32,
 }
@@ -105,6 +107,11 @@ impl MysqlInstanceShim {
             }
         }
 
+        let cache = Cache::builder()
+            .time_to_live(Duration::from_secs(30 * 60))
+            .time_to_idle(Duration::from_secs(5 * 60))
+            .build();
+
         MysqlInstanceShim {
             query_handler,
             salt: scramble,
@@ -115,7 +122,7 @@ impl MysqlInstanceShim {
                 process_id,
             )),
             user_provider,
-            prepared_stmts: Default::default(),
+            prepared_stmts: Arc::new(cache),
             prepared_stmts_counter: AtomicU32::new(1),
             process_id,
         }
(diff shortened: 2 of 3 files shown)


killme2008:
Thank you for your contribution.

After some investigation, I believe we should not set a TTL for the statement cache. If a statement expires, applications relying on the cache would fail immediately.

Instead, we could set a cache size limit, similar to MySQL’s `max_prepared_stmt_count`:

https://blog.searce.com/how-max-prepared-stmt-count-bring-down-the-production-mysql-system-6ca28e577663

If the cache reaches this limit, we can return an error.

sunheyi6:
> Thank you for your contribution.
> 
> After some investigation, I believe we should not set a TTL for the statement cache. If a statement expires, applications relying on the cache would fail immediately.
> 
> Instead, we could set a cache size limit, similar to MySQL’s `max_prepared_stmt_count`:
> 
> https://blog.searce.com/how-max-prepared-stmt-count-bring-down-the-production-mysql-system-6ca28e577663
> 
> If the cache reaches this limit, we can return an error.

that meens only add a param to control cache size,don't remove ttl?

killme2008:
> > Thank you for your contribution.
> > After some investigation, I believe we should not set a TTL for the statement cache. If a statement expires, applications relying on the cache would fail immediately.
> > Instead, we could set a cache size limit, similar to MySQL’s `max_prepared_stmt_count`:
> > https://blog.searce.com/how-max-prepared-stmt-count-bring-down-the-production-mysql-system-6ca28e577663
> > If the cache reaches this limit, we can return an error.
> 
> that meens only add a param to control cache size,don't remove ttl?

Yes, controlling the cache size should suffice.

sunheyi6:
@killme2008 done

src/servers/Cargo.toml
```diff
@@ -153,6 +154,7 @@ table.workspace = true
 tempfile = "3.0.0"
 tokio-postgres = "0.7"
 tokio-postgres-rustls = "0.12"
+moka = { workspace = true, features = ["sync"] }
```
killme2008: I think we don't need this dep at all. Just use the original HashMap, and check the capacity before inserting.

src/servers/src/mysql/handler.rs
```diff
@@ -159,14 +172,12 @@ impl MysqlInstanceShim {
 
     /// Save query and logical plan with a given statement key
     fn save_plan(&self, plan: SqlPlan, stmt_key: String) {
-        let mut prepared_stmts = self.prepared_stmts.write();
-        let _ = prepared_stmts.insert(stmt_key, plan);
+        self.prepared_stmts.insert(stmt_key, plan);
```
killme2008: Checking capacity during insertion is sufficient.

src/servers/src/mysql/handler.rs
```diff
@@ -105,6 +107,17 @@ impl MysqlInstanceShim {
             }
         }
 
+        let max_capacity = std::env::var("PREPARED_STMT_MAX_CAPACITY")
```
killme2008: Would be better to make it configurable.Prefer to add the capacity into `MySQLOptions`:

https://github.com/GreptimeTeam/greptimedb/blob/869f8bf68ad39f32fb9b40a3c7f25b98765bafc3/src/frontend/src/service_config/mysql.rs#L19

Don't forget to update the example configurations:

https://github.com/GreptimeTeam/greptimedb/tree/main/config

src/servers/Cargo.toml
sunheyi6: ok,i will try to use hashmap

src/servers/src/mysql/handler.rs
sunheyi6: These two files are not under the same src, so I cannot directly import them in handle.rs by using use crate::frontend::FrontendOptions;

killme2008: Since the frontend creates the servers, you can simply add a field to the MySQL server and pass it when creating servers in the frontend crate.

See https://github.com/GreptimeTeam/greptimedb/blob/869f8bf68ad39f32fb9b40a3c7f25b98765bafc3/src/frontend/src/server.rs#L228

sunheyi6: ok

sunheyi6: @killme2008 done

src/frontend/src/service_config/mysql.rs
```diff
@@ -29,6 +29,7 @@ pub struct MysqlOptions {
     #[serde(default = "Default::default")]
     #[serde(with = "humantime_serde")]
     pub keep_alive: std::time::Duration,
+    pub prepared_max_capacity: usize,
```
killme2008: The name is not good. I think `prepared_stmt_cache_capacity` is better.

src/frontend/src/service_config/mysql.rs
```diff
@@ -40,6 +41,7 @@ impl Default for MysqlOptions {
             tls: TlsOption::default(),
             reject_no_database: None,
             keep_alive: std::time::Duration::from_secs(0),
+            prepared_max_capacity: 1000,
```
killme2008: The default value could be more large, e.g. 10000

src/servers/src/mysql/handler.rs
```diff
@@ -160,6 +163,19 @@ impl MysqlInstanceShim {
     /// Save query and logical plan with a given statement key
     fn save_plan(&self, plan: SqlPlan, stmt_key: String) {
         let mut prepared_stmts = self.prepared_stmts.write();
+        let max_capacity = self.prepared_max_capacity;
+
+        let is_update = prepared_stmts.contains_key(&stmt_key);
+
+        if !is_update && prepared_stmts.len() >= max_capacity {
+            // Log a warning but don't insert
```
killme2008: We must return an error instead of just printing an warning log.

killme2008 (commented):
FYI, we need to update the sample configs as I mentioned before:

https://github.com/GreptimeTeam/greptimedb/tree/main/config

No rush on the coding - I'd recommend reading through the contribution guide first:

https://github.com/GreptimeTeam/greptimedb/blob/main/CONTRIBUTING.md

yihong0618:
about how to pass the DCO

```
git commit -m "xxxxxxxxxxxx" -s
```

add the `-s` for every commit is fine

sunheyi6:
sorry, I messed up the submission record of this PR. I will submit a new one.

yihong0618:
> sorry, I messed up the submission record of this PR. I will submit a new one.

you can use rebase.

sunheyi6:
> > sorry, I messed up the submission record of this PR. I will submit a new one.
> 
> you can use rebase.

I can't seem to reopen it after closing it.Fine, I'll just submit a new PR.

Issue #6628, opened 2025-08-01
Crash when there are too many WHERE conditions in the SQL query


0neSe7en:
### What type of bug is this?

Crash

### What subsystems are affected?

Query Engine

### Minimal reproduce step

There is a bug in my application that generates a SQL query with about 500 WHERE conditions.

```sql
SELECT * FROM "logs" where timestamp between 0 and 1 ... 500 more timestamp between 0 and 1
```

### What did you expect to see?

results or throw an error.

### What did you see instead?

the greptime crashed

### What operating system did you use?

macOS

### What version of GreptimeDB did you use?

0.15.3

### Relevant log output and stack trace

```bash
greptime-1  |
greptime-1  | thread '<unknown>' has overflowed its stack
greptime-1  | fatal runtime error: stack overflow, aborting
```

sunheyi6:
Is this a real business scenario? What business scenario requires such query conditions?

0neSe7en:
no, the query was incorrect, it was generated by the code that had bug. 
But I think we'd better throw an error rather than crash the greptime.

sunheyi6:
> no, the query was incorrect, it was generated by the code that had bug. But I think we'd better throw an error rather than crash the greptime.

yes, throwing exceptions is better.

yihong0618:
This is interesting.


how to reproduce

```py
def generate_sql_with_many_conditions():
    base_query = 'SELECT * FROM "logs" WHERE\n'
    condition = 'timestamp BETWEEN 0 AND 1'
    conditions = [condition for _ in range(400)]
    full_query = base_query + ' OR\n'.join(conditions) + ';'
    return full_query

if __name__ == "__main__":
    sql = generate_sql_with_many_conditions()
    print(sql)
```

```cli
python3 t.py > t.sql
```

start sever

psql -h 127.0.0.1 -p 4003 -U public  -d public -f t.sql

try run it lldb

error: 
```
lldb 
2025-08-01T07:34:21.142934Z  INFO flow::adapter::flownode_impl: FlowDualEngine done recovering
Process 11421 stopped
* thread #19, name = 'global-worker', stop reason = EXC_BAD_ACCESS (code=2, address=0x1780cf8a0)
[...]

PR #6634, opened 2025-08-01
fix: bump greptime-sqlparser to avoid convert statement overflow

kemingy:

- fix https://github.com/GreptimeTeam/greptimedb/issues/6628

The SIGSEGV is from the `s.into()`, which convert the `Statement`:
- from `sqlparser-rs-6d6510b5e3af714f`
- to `sqlparser-0.54.0`

## PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2646,7 +2646,7 @@ dependencies = [
  "futures-util",
  "serde",
  "snafu 0.8.5",
- "sqlparser 0.54.0 (git+https://github.com/GreptimeTeam/sqlparser-rs.git?rev=0cf6c04490d59435ee965edd2078e8855bd8471e)",
+ "sqlparser 0.54.0-greptime",
  "sqlparser_derive 0.1.1",
  "statrs",
  "store-api",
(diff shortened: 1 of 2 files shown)


src/sql/src/statements/statement.rs
```diff
@@ -305,6 +305,9 @@ impl TryFrom<&Statement> for DfStatement {
                 .fail();
             }
         };
-        Ok(DfStatement::Statement(Box::new(s.into())))
+        // SAFETY: both are from sqlparser (shallow and v0.54), they are identical.
+        Ok(DfStatement::Statement(Box::new(unsafe {
+            std::mem::transmute(s)
+        })))
```
waynexia: Thank you for this fix!

To my understanding this is not guaranteed to be a sound behavior in Rust https://doc.rust-lang.org/nomicon/repr-rust.html

Image

One possible way to make this sound is by annotating both structs with `#[repr(C)]`, but it is not very viable in this scenario.

Are you interested in optimizing the conversion macro itself to avoid overflow? It's implemented here https://github.com/GreptimeTeam/sqlparser-rs/blob/f1bf51f225da3021709f28fc80b196482c4690a1/derive/src/lib.rs#L608-L616

kemingy: You're right. This is not guaranteed. I'll check the proc macro derive.

kemingy:
- bug fixed in https://github.com/GreptimeTeam/sqlparser-rs/pull/18

cc @sunng87

waynexia (approved):
LGTM :+1: Thanks for taking this

sunng87 (approved):

PR #6641, opened 2025-08-04
fix: show create flow's expire after

discord9:

as title, add time unit in show create flow, also add more testcases

## PR Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -195,6 +195,7 @@ rstest_reuse = "0.7"
 rust_decimal = "1.33"
 rustc-hash = "2.0"
 # It is worth noting that we should try to avoid using aws-lc-rs until it can be compiled on various platforms.
+pretty_assertions = "1.4.0"
 rustls = { version = "0.23.25", default-features = false }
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -195,6 +195,7 @@ rstest_reuse = "0.7"
 rust_decimal = "1.33"
 rustc-hash = "2.0"
 # It is worth noting that we should try to avoid using aws-lc-rs until it can be compiled on various platforms.
+pretty_assertions = "1.4.0"
```
killme2008: Used only in the query crate; does not need to be added to the workspace or other crates.

FYI. The comment above is for `rustls`.

killme2008 (approved):
Almost LGTM

Cargo.toml
discord9: > Used only in the query crate; does not need to be added to the workspace or other crates.
> 
> FYI. The comment above is for `rustls`.

pretty_assertions now get used in four crates(all dev-dependencies though) so added to workspace

PR #6646, opened 2025-08-04
feat: use column expr with filters in LogQuery

waynexia:

`filters` is a Vector of `ColumnFilters`, and inside `ColumnFilters` it requires a column. In `LogExpr::Filter`, it has an `expr` field along with `filter: ContentFilter`. There is some duplication about column reference in these cases. Make it hard to use. This patch made a refactor to how a column is provided to filters

## PR Checklist

### DIFF
--- a/src/query/src/log_query/planner.rs
+++ b/src/query/src/log_query/planner.rs
@@ -213,32 +213,21 @@ impl LogQueryPlanner {
                 Ok(Some(left.and(right)))
             }
             log_query::ContentFilter::GreatThan { value, inclusive } => {
-                if *inclusive {
-                    Ok(Some(
-                        col_expr
-                            .clone()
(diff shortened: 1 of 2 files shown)


src/query/src/log_query/planner.rs
```diff
@@ -138,118 +140,133 @@ impl LogQueryPlanner {
...
+                            .clone()
+                            .gt_eq(lit(ScalarValue::Utf8(Some(value.clone())))),
+                    ))
+                } else {
+                    Ok(Some(
+                        col_expr
+                            .clone()
+                            .gt(lit(ScalarValue::Utf8(Some(value.clone())))),
+                    ))
+                }
```
evenyag: Is it possible to remove the clone inside if-else? Can we avoid some clone in this function?

evenyag (approved):

PR #6650, opened 2025-08-05
feat: truncate files when flow

discord9:

as title, truncate table before sending flow's query

~TODO: add rate limit for truncate operation~
rate limit impl-ed
should help with improve query performance of massively repeatedly updating table

## PR Checklist


discord9:
tests show that setting limit to 10 min is still too frequent and create too many files(3000+ files after 3 days), setting to 1h instead and will try test again

discord9:
realized `partial truncate` still need a sequence number (and only truncate files before seq number) and only partial truncate after flow's insert into query is completed so to make sure sink table always have data(even just slightly outdated)

PR #6651, opened 2025-08-05
perf: Reduce fulltext bloom load time

evenyag:

This PR improves the first load time on bloom fulltext index.

Before this PR, fulltext index always enabled compression. When using bloom backend, compression doesn't reduce the size of the puffin blob a lot, but introduces extra time to decompress the blob to the staging directory.
```
Init blob to stager, type: greptime-fulltext-index-bloom-0, blob size: 14763795, after size: 14927141, cost: 187.463416ms
```

It removes `try_join_all` from cached blob readers. This can reduce duplicated IO and load time for bloom blobs.
When reading multiple bloom filter ranges, `get_or_load()` usually fetches the same underlying page to merge IO.
If the blob reader uses `try_join_all` to fetch pages concurrently, it will fetch the same page multiple times from the puffin file.

This can lead to worse performance when the `granularity` is small:
```
INFO mito2::cache::index::bloom_filter_index: bloom filter read vec from cache, ranges: 400
INFO mito2::cache::index: Cache miss range: [0..65536]
INFO mito2::cache::index: Cache miss range: [0..65536]
INFO mito2::cache::index: Cache miss range: [0..65536]
INFO mito2::cache::index: Cache miss range: [0..65536]
INFO mito2::cache::index: Cache miss range: [0..65536]
INFO mito2::cache::index: Cache miss range: [0..65536]
INFO mito2::cache::index: Cache miss range: [0..65536]
......
```

Removing this can reduce about 50% of the first load time with the default granularity.

## PR Checklist

### DIFF
--- a/src/mito2/src/cache/index/bloom_filter_index.rs
+++ b/src/mito2/src/cache/index/bloom_filter_index.rs
@@ -18,7 +18,6 @@ use std::sync::Arc;
 use api::v1::index::BloomFilterMeta;
 use async_trait::async_trait;
 use bytes::Bytes;
-use futures::future::try_join_all;
 use index::bloom_filter::error::Result;
 use index::bloom_filter::reader::BloomFilterReader;
 use store_api::storage::ColumnId;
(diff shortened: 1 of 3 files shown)


zhongzc (approved):
Very reasonable and well worth it. 👍

WenyXu (approved):

src/mito2/src/cache/index/bloom_filter_index.rs
```diff
@@ -120,21 +119,24 @@ impl<R: BloomFilterReader + Send> BloomFilterReader for CachedBloomFilterIndexBl
...
-                range.start,
-                (range.end - range.start) as u32,
-                move |ranges| async move { inner.read_vec(&ranges).await },
-            )
-        });
-        Ok(try_join_all(fetch)
-            .await?
-            .into_iter()
-            .map(Bytes::from)
-            .collect::<Vec<_>>())
```
waynexia, line 137: would this slow down when ranges are not very overlapping?

evenyag: Ranges here belong to the same row group and are usually smaller than the default cache page size to fetch.

PR #6658, opened 2025-08-05
perf: improve bloom filter reader's byte reading logic

waynexia:

think of this when reviewing https://github.com/GreptimeTeam/greptimedb/pull/6651

Converting bytes to u64 one by one is not optimal. We can do it more efficiently

## PR Checklist

### DIFF
--- a/src/index/src/bloom_filter/reader.rs
+++ b/src/index/src/bloom_filter/reader.rs
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::ops::Range;
+use std::ops::{Range, Rem};
 
 use async_trait::async_trait;
+use bytemuck::try_cast_slice;
 use bytes::Bytes;
 use common_base::range_read::RangeReader;
 use fastbloom::BloomFilter;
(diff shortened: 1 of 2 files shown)


src/index/src/bloom_filter/reader.rs
```diff
@@ -33,6 +34,45 @@ const BLOOM_META_LEN_SIZE: u64 = 4;
...
+
+        // SAFETY: We're creating a properly sized slice from uninitialized but allocated memory
+        // to copy bytes into. The slice has exactly the right size for the byte data.
+        let dest_slice = unsafe {
+            std::slice::from_raw_parts_mut(u64_vec.as_mut_ptr() as *mut u8, byte_slice.len())
+        };
+        dest_slice.copy_from_slice(byte_slice);
+
+        // SAFETY: We've just initialized exactly u64_count elements worth of bytes
+        unsafe { u64_vec.set_len(u64_count) };
```
evenyag, line 62: May need some tests for this function.

src/index/src/bloom_filter/reader.rs
```diff
@@ -33,6 +34,45 @@ const BLOOM_META_LEN_SIZE: u64 = 4;
 /// Default prefetch size of bloom filter meta.
 pub const DEFAULT_PREFETCH_SIZE: u64 = 8192; // 8KiB
 
+/// Safely converts bytes to Vec<u64> using bytemuck for optimal performance.
+/// Faster than chunking and converting each piece individually.
+fn bytes_to_u64_vec(bytes: &Bytes) -> Vec<u64> {
```
evenyag: Can we write some benchmarks (aligned and unaligned) for this against the old implementation?

src/index/src/bloom_filter/reader.rs
waynexia: there is one with real file IO, and it will fail without this if-else

src/index/src/bloom_filter/reader.rs
waynexia: 846a7a0

src/index/src/bloom_filter/reader.rs
```diff
@@ -33,6 +34,45 @@ const BLOOM_META_LEN_SIZE: u64 = 4;
 /// Default prefetch size of bloom filter meta.
 pub const DEFAULT_PREFETCH_SIZE: u64 = 8192; // 8KiB
 
+/// Safely converts bytes to Vec<u64> using bytemuck for optimal performance.
```
zhongzc, line 37: It should be noted that these bytes are a sequence of little-endian u64s.

evenyag (approved):

fengjiachun (approved):
LGTM

PR #6645, opened 2025-08-04
feat: support TQL CTE in planner

waynexia:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

part of https://github.com/GreptimeTeam/greptimedb/issues/1915

## What's changed and what's your intention?

Now TQL CTE works

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #6661, opened 2025-08-06
fix: box Explain node in Statement to reduce stack size

waynexia:

Maybe related to https://github.com/GreptimeTeam/greptimedb/pull/6634

I find https://github.com/GreptimeTeam/greptimedb/pull/6645 will stack overflowed on a totally unrelated place (range query test). And from the debugger, `stacker` chooses not to grow the stack size every time.

I suspect this is caused by some large enum or struct. And it turns out this is at least related. Before `Statement` is 1856 size, because of `Explain` which is the same size as well. After boxing it the `Statement`'s size shrinks to 384, and the overflow in #6645 is gone

## PR Checklist

### DIFF
--- a/src/sql/src/parsers/explain_parser.rs
+++ b/src/sql/src/parsers/explain_parser.rs
@@ -31,7 +31,9 @@ impl ParserContext<'_> {
                 actual: self.peek_token_as_string(),
             })?;
 
-        Ok(Statement::Explain(Explain::try_from(explain_statement)?))
+        Ok(Statement::Explain(Box::new(Explain::try_from(
+            explain_statement,
+        )?)))
     }
 }
 
@@ -118,6 +120,6 @@ mod tests {
         })
         .unwrap();
 
-        assert_eq!(stmts[0], Statement::Explain(explain))
+        assert_eq!(stmts[0], Statement::Explain(Box::new(explain)))
     }
 }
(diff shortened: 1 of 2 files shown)


waynexia:
cc @kemingy @yihong0618 @sunng87, you might be interested in this

kemingy:
I saw the comments from the upstream sqlparser: https://github.com/apache/datafusion-sqlparser-rs/blob/35835146020f0f4782fd4e5caf05aa8b7e17e7b6/src/lib.rs#L152-L155

There are lots of `large_enum_variant` warnings. Shall we box more enum elements?

waynexia:
> I saw the comments from the upstream sqlparser: [apache/datafusion-sqlparser-rs@`3583514`/src/lib.rs#L152-L155](https://github.com/apache/datafusion-sqlparser-rs/blob/35835146020f0f4782fd4e5caf05aa8b7e17e7b6/src/lib.rs#L152-L155)

I suppose 1.8KB is not just an "Extra memory"... For other cases, like ~100B overhead looks acceptable. But I'm also curious about why `recursive` doesn't help in this case.

> There are lots of `large_enum_variant` warnings. Shall we box more enum elements?

You are right, maybe we need to have another look to see if there are any other unexpected large things

Issue #6571, opened 2025-07-23
Tracking issue: global gc worker


discord9:
### What problem does the new feature solve?

Remove obsolete parquet files after table repartitioning and such

### What does the feature do?

remove stale files generating from repartition or compaction

### Implementation challenges

- Preventing Incorrect Deletion of In-Use Files
- Differentiating Leaked vs. In-Progress Files
- Designing a Robust Read Replica Protection Mechanism

PR #6669, opened 2025-08-06
feat: gc worker phase 1

discord9:

https://github.com/GreptimeTeam/greptimedb/issues/6571

simple impl of single region garbage collector for sst/index files that are not in use for a while
added a `removed_files` field to region manifest, which will store removed files from certain manifest version(determined by gc worker) to now

## PR Checklist


src/mito2/src/gc.rs
```diff
@@ -0,0 +1,252 @@
...
+                current_files.keys().cloned().collect(),
+                concurrency,
+            )
+            .await?;
+
+        let old_unused_files = unused_files
+            .into_iter()
+            .filter(|e| {
+                // TODO(discord9): is use last_modified time correct?
+                if let Some(last_modified) = e.metadata().last_modified() {
```
discord9: should I instead get a tag of `delete_time` from metadata in `user_metadata` instead(and setting a tag when delete files)? @evenyag

evenyag: If we delete the file, how can we set a tag?

discord9: > If we delete the file, how can we set a tag?

exactly, ~modify the file handler to not delete files, but instead set a tag with `expel_time` being now(), meaning it stops being ref from manifest at current time, and gc worker can then delete lingering files based on that~
can't modify metadata in-place using object store so maybe just upload a new tmp empty file to indicate it's expel time?

edited: decided to use delta manifest files in region manifest instead, although it have a limit of 10 by default, but it should be ok for most case

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,252 @@
...
+    pub(crate) manifest_ctx: ManifestContextRef,
+    /// Lingering time before deleting files.
+    pub(crate) lingering_time: Duration,
+}
+
+impl LocalGcWorker {
+    /// concurrency of listing files per region.
+    /// This is used to limit the number of concurrent listing operations and speed up listing
+    pub const CONCURRENCY_LIST_PER_FILES: usize = 512;
+    pub async fn do_region_gc(&self) -> Result<()> {
```
evenyag: Also document do_region_gc briefly.
```suggestion
    /// concurrency of listing files per region.
    /// This is used to limit the number of concurrent listing operations and speed up listing
    pub const CONCURRENCY_LIST_PER_FILES: usize = 512;

    pub async fn do_region_gc(&self) -> Result<()> {
```

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,252 @@
...
+use store_api::storage::RegionId;
+use tokio_stream::StreamExt;
+
+use crate::access_layer::AccessLayerRef;
+use crate::error::{JoinSnafu, OpenDalSnafu, Result};
+use crate::region::ManifestContextRef;
+use crate::sst::file::{FileId, RegionFileId};
+use crate::sst::location;
+
+pub struct LocalGcWorker {
```
evenyag, line 87: So it only supports one region, right? Looks like it doesn't support cross region gc, too. Consider providing more information in the PR description and comment.

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,420 @@
...
+        region_id: RegionId,
+        in_used: HashSet<FileId>,
+        recently_removed_files: BTreeMap<Option<Timestamp>, HashSet<FileMeta>>,
+        concurrency: usize,
+    ) -> Result<Vec<Entry>> {
+        let may_linger_until =
+            chrono::Utc::now() - chrono::Duration::from_std(self.opt.lingering_time).unwrap();
+
+        let unknown_file_may_linger_until = chrono::Utc::now()
+            - chrono::Duration::from_std(self.opt.unknown_file_lingering_time).unwrap();
```
evenyag: Is the unwrap safe?

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,420 @@
...
+        let mut entries = vec![];
+        while let Some(stream) = rx.recv().await {
+            entries.extend(
+                stream
+                    .into_iter()
+                    .filter_map(Result::ok)
+                    .filter(|e| !in_use_filenames.contains(e.name()))
+                    .filter(|e| !may_linger_filenames.contains(e.name()))
+                    .filter(|e| {
+                        if !all_files_appear_in_delta_manifests.contains(e.name()) {
```
evenyag: This may only contains the file name, not the path.

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,420 @@
...
+                        self.access_layer.path_type(),
+                    ),
+                    location::index_file_path(
+                        self.access_layer.table_dir(),
+                        meta.file_id(),
+                        self.access_layer.path_type(),
+                    ),
+                ]
+            })
+            .collect::<HashSet<_>>();
```
evenyag: Will it be wasteful to keep all path strings in the set? Can we only maintain a `HashSet<FileId>`? We list and parse the files to check their file id.

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,420 @@
...
+/// Generate partition prefixes based on concurrency and
+/// assume file names are evenly-distributed uuid string,
+/// to evenly distribute files across partitions.
+/// For example, if concurrency is 2, partition prefixes will be:
+/// ["8"] so it divide uuids into two partitions based on the first character.
+/// If concurrency is 32, partition prefixes will be:
+/// ["08", "10", "18", "20", "28", "30", "38" ..., "f0", "f8"]
+/// if concurrency is 1, it returns an empty vector.
+///
+fn gen_partition_from_concurrency(concurrency: usize) -> Vec<String> {
```
evenyag, line 355: Since we may have multiple regions, maybe we should limit the max concurrency per gc job. We can also use a fixed concurrency for simplicity.

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,420 @@
...
+            .delete_iter(unused_files.iter().map(|e| e.name()))
+            .await
+            .context(OpenDalSnafu)
+    }
+
+    /// Get all the removed files in delta manifest files and their expel times.
+    /// The expel time is the time when the file is considered as removed.
+    /// Which is the last modified time of delta manifest which contains the remove action.
+    ///
+    pub async fn get_removed_files_expel_times(
```
evenyag, line 141: We can get removed files from the FilePurger or just store a list of recently deleted files in the manifest.

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,420 @@
...
+use crate::region::ManifestContextRef;
+use crate::sst::file::{FileId, FileMeta, RegionFileId};
+use crate::sst::location;
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct FileGcOption {
+    /// Lingering time before deleting files.
+    #[serde(with = "humantime_serde")]
+    pub lingering_time: Duration,
+    /// Lingering time before deleting unknown files(files with undetermine expel time).
```
evenyag, line 53: Maybe recycle/remove/expire time is better than expel time.

discord9: > Maybe recycle/remove/expire time is better than expel time.

sound can be confusing with actual delete time

evenyag: Expel is rarely used in this context. LLM suggests using "expiration time" or "retention time" instead.

discord9: > retention

expel time is the time file is removed from manifest, not the duration for which file continue to linger after being removed from manifest, maybe there is this confusing?

src/mito2/src/gc.rs
discord9: > So it only supports one region, right? Looks like it doesn't support cross region gc, too. Consider providing more information in the PR description and comment.

added support for multiple region, will call those methods from metasrv in later pr

src/mito2/src/gc.rs
discord9: > We can get removed files from the FilePurger or just store a list of recently deleted files in the manifest.

store a list of recently deleted files in the manifest instead.

src/mito2/src/manifest/action.rs
```diff
@@ -84,12 +93,24 @@ pub enum TruncateKind {
...
+    /// This is a list of (set of files, timestamp) pairs, where the timestamp is the time when
+    /// the files are removed from manifest. The timestamp is in milliseconds since unix epoch.
+    ///
+    /// Using same checkpoint files and action files, the recovered manifest may differ in this
+    /// `removed_files` field, because the checkpointer may evict some removed files using
+    /// current machine time. This is acceptable because the removed files are not used in normal
+    /// read/write path.
+    ///
+    #[serde(default)]
+    pub removed_files: Vec<(HashSet<FileMeta>, i64)>,
```
evenyag, line 114: Better to define a struct. Can we only keep RegionId and FileId instead of the whole FileMeta to save some space?

src/mito2/src/manifest/checkpointer.rs
```diff
@@ -118,6 +120,40 @@ impl Checkpointer {
         self.inner.last_checkpoint_version.load(Ordering::Relaxed)
     }
 
+    /// Update the `removed_files` field in the manifest by the options in `manifest_options`.
+    /// This should be called before maybe do checkpoint to update the manifest.
+    pub(crate) fn update_manifest_removed_files(
```
evenyag, line 125: We haven't called this method?

src/mito2/src/manifest/checkpointer.rs
```diff
@@ -118,6 +120,40 @@ impl Checkpointer {
...
+            .filter_map(|(files, ts)| {
+                if ts < can_evict_until.timestamp_millis() {
+                    // can evict all files
+                    // TODO(discord9): maybe only evict to below keep_count? Maybe not, or the update might be too frequent.
+                    None
+                } else {
+                    Some((files, ts))
+                }
+            });
+        manifest.removed_files = updated.collect();
```
evenyag, line 152: We may need to sort them by time and only keep `keep_count` records.

src/mito2/src/manifest/checkpointer.rs
discord9: > We haven't called this method?

it's called just before `maybe_do_checkpoint`

discord9:
closed for https://github.com/GreptimeTeam/greptimedb/pull/6751 need manifest change to merge first

PR #6671, opened 2025-08-06
fix: TQL CTE parser take raw query string

waynexia:

Part of #1915 

Fix one parser problem, which used to reconstruct the query string from tokens instead of taking the string directly. 

## PR Checklist


src/sql/src/parsers/with_tql_parser.rs
```diff
@@ -234,34 +247,30 @@ impl ParserContext<'_> {
...
             }
 
-            // Consume the token and push it into our buffer
+            // Consume the token and track parentheses depth
             let consumed = self.parser.next_token();
             match consumed.token {
                 Token::LParen => paren_depth += 1,
                 Token::RParen => {
-                    // This RParen must belong to a nested expression since
-                    // `paren_depth > 0` here. Decrease depth accordingly.
```
evenyag: Why do we remove this comment?

evenyag (commented):
Sqlness failed

src/sql/src/parsers/with_tql_parser.rs
waynexia: looks unnecessary here. added back

evenyag (approved):

PR #6675, opened 2025-08-07
chore: update datafusion family

MichaelScofield:
pass clippy and fmt

most notably updates are:

arrow 54 -> 56
datafusion months ago to today
tonic 0.12 -> 0.13
sqlparser 0.54 -> 0.55

## PR Checklist


.github/workflows/dependency-check.yml
```diff
@@ -18,6 +18,10 @@ jobs:
     - name: Set up Rust
       uses: actions-rust-lang/setup-rust-toolchain@v1
 
+    - uses: webfactory/ssh-agent@v0.9.0
+      with:
+        ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+
```
evenyag: Why do we need this in the CI action?

MichaelScofield: because otel-arrow has a submodule starts with "git@github": https://github.com/GreptimeTeam/otel-arrow/blob/main/.gitmodules

I tried to provide a dummy ssh key, or configure the git to replace ssh with https, neither succeeded. So I have to create a new ssh key of mine and set it here.

src/common/datasource/src/file_format/orc.rs
```diff
@@ -119,7 +120,7 @@ impl OrcOpener {
 }
 
 impl FileOpener for OrcOpener {
-    fn open(&self, meta: FileMeta) -> DfResult<FileOpenFuture> {
+    fn open(&self, meta: FileMeta, _: PartitionedFile) -> DfResult<FileOpenFuture> {
```
evenyag: nitpick: Maybe we can use the size in the ObjectMeta of the PartitionedFile

src/query/src/dist_plan/analyzer.rs
```diff
@@ -155,7 +155,7 @@ impl DistPlannerAnalyzer {
             .collect::<DfResult<Vec<_>>>()?;
 
         // Some plans that are special treated (should not call `with_new_exprs` on them)
-        if !matches!(plan, LogicalPlan::Unnest(_)) {
+        if !matches!(plan, LogicalPlan::Unnest(_) | LogicalPlan::Join(_)) {
```
discord9: is the join suppose to be here?

src/query/src/dist_plan/analyzer/test.rs
```diff
@@ -646,7 +646,8 @@ fn expand_part_col_aggr_part_col_aggr() {
         .unwrap();
 
     let expected_original = [
-        "Aggregate: groupBy=[[t.pk1, t.pk2, max(t.number)]], aggr=[[min(max(t.number))]]", // notice here `max(t.number)` is added to groupBy due to aggr exprs depend on this column
+        // See DataFusion #14860 for change details.
```
discord9, line 649: by the way 14860 have a small bug(`LogicalPlanBuilder`'s option can't persist be set) so later might need to bump version by a bit(but that will be much easier)

src/query/src/dist_plan/analyzer.rs
MichaelScofield: calling join with `with_new_exprs` here will cause a panic, see datafusion #15055

src/query/src/dist_plan/analyzer/test.rs
MichaelScofield: got it, thx

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -247,15 +243,8 @@ impl StateWrapper {
...
-        let input_exprs = acc_args.exprs;
-        let input_schema = acc_args.schema;
-        let input_types = input_exprs
-            .iter()
-            .map(|e| e.data_type(input_schema))
-            .collect::<Result<Vec<_>, _>>()?;
-        let return_type = self.inner.return_type(&input_types)?;
-        Ok(return_type)
+    ) -> datafusion_common::Result<FieldRef> {
+        self.inner.return_field(acc_args.schema.fields())
```
evenyag, line 247: Is it the new recommended way to get the return field?

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -297,11 +285,15 @@ impl AggregateUDFImpl for StateWrapper {
     /// Return state_fields as the output struct type.
     ///
     fn return_type(&self, arg_types: &[DataType]) -> datafusion_common::Result<DataType> {
-        let old_return_type = self.inner.return_type(arg_types)?;
+        let input_fields = &arg_types
+            .iter()
+            .map(|x| Arc::new(Field::new("x", x.clone(), false)))
```
evenyag, line 290: Is it ok to use `x` as field names?

src/flow/src/batching_mode/task.rs
```diff
@@ -340,16 +348,6 @@ impl BatchingTask {
...
-        let expanded_plan = CountWildcardRule::new()
-            .analyze(fixed_plan.clone(), &Default::default())
-            .with_context(|_| DatafusionSnafu {
-                context: format!(
-                    "Failed to expand wildcard in logical plan, plan={:?}",
-                    fixed_plan
-                ),
-            })?;
-
-        let plan = expanded_plan;
```
evenyag, line 352: Can we remove this? @discord9

src/mito2/src/sst/parquet/metadata.rs
```diff
@@ -95,13 +95,14 @@ impl<'a> MetadataLoader<'a> {
         let mut footer = [0; 8];
         footer.copy_from_slice(&buffer[buffer_len - FOOTER_SIZE..]);
 
-        let metadata_len = ParquetMetaDataReader::decode_footer(&footer).map_err(|e| {
+        let footer_tail = ParquetMetaDataReader::decode_footer_tail(&footer).map_err(|e| {
```
evenyag, line 98: out-of-topic: We can refactor this to use `ArrowReaderMetadata` in the future.
https://docs.rs/parquet/latest/parquet/arrow/arrow_reader/struct.ArrowReaderMetadata.html#method.load_async

src/object-store/src/layers.rs
```diff
@@ -40,11 +40,7 @@ mod prometheus {
                 // remove path label first, waiting for later fix
                 // TODO(shuiyisong): add dynamic path label trim for opendal
 
-                let layer = PrometheusLayer::builder()
-                    .path_label(0)
```
evenyag, line 44: Is it safe to remove the `path_label()` call?

src/operator/src/statement/copy_table_from.rs
```diff
@@ -280,32 +264,24 @@ impl StatementExecutor {
                         .context(error::PhysicalExprSnafu)?,
                 ))
             }
-            FileMetadata::Json {
-                format,
-                path,
-                schema,
-            } => {
-                let projected_file_schema = Arc::new(
```
evenyag, line 288: Is it ok to remove the `format`? cc @WenyXu

src/promql/src/extension_plan/absent.rs
```diff
@@ -278,16 +280,18 @@ impl ExecutionPlan for AbsentExec {
...
             expr: Arc::new(
                 ColumnExpr::new_with_schema(&self.time_index_column, &self.input.schema()).unwrap(),
             ),
             options: Some(SortOptions {
                 descending: false,
                 nulls_first: false,
             }),
-        }]))]
+        }]
+        .into();
```
evenyag: non-blocking: I'd prefer from over into.

src/promql/src/extension_plan/empty_metric.rs
```diff
@@ -60,6 +62,12 @@ pub struct EmptyMetric {
     time_index_schema: DFSchemaRef,
     /// Schema of the output record batch
     result_schema: DFSchemaRef,
+    // This dummy input's sole purpose is to provide a schema for use in DataFusion's
+    // `SimplifyExpressions`. Otherwise it may report a "no field name ..." error.
+    // The error is caused by an optimization that tries to rewrite "A = A", during
+    // which will find the field in plan's schema. However, the schema is empty if the
+    // plan does not have an input.
+    dummy_input: LogicalPlan,
```
evenyag, line 70: Why is the schema empty when it doesn't have input? Isn't the schema built by fields?

src/promql/src/functions/round.rs
```diff
@@ -118,10 +120,17 @@ mod tests {
...
+            Arc::new(Field::new("b", input[1].data_type(), false)),
+        ];
+        let return_field = Arc::new(Field::new("c", DataType::Float64, false));
         let args = ScalarFunctionArgs {
             args: input,
+            arg_fields,
             number_rows: 1,
-            return_type: &DataType::Float64,
+            return_field,
+            config_options: Arc::new(ConfigOptions::default()),
```
evenyag, line 133: out-of-topic: Now the ScalarFunctionArgs contains ConfigOptions, we may use it to pass time zone and rewrite our time-related functions.

src/query/src/dist_plan/analyzer.rs
evenyag: So expressions_consider_join doesn't work now? Is it possible to fix it? If datafusion has fixed this, we can change the expressions_consider_join implementation now.

src/query/src/lib.rs
```diff
@@ -17,6 +17,7 @@
 #![feature(try_blocks)]
 #![feature(stmt_expr_attributes)]
 #![feature(iterator_try_collect)]
+#![feature(box_patterns)]
```
evenyag, line 20: Where did we use it? Looks like it will be replaced.
https://doc.rust-lang.org/beta/unstable-book/language-features/box-patterns.html

src/query/src/optimizer/count_wildcard.rs
```diff
@@ -105,24 +105,30 @@ impl CountWildcardToTimeIndexRule {
...
-    fn is_wildcard(expr: &Expr) -> bool {
-        matches!(expr, Expr::Wildcard { .. })
+    #[expect(deprecated)]
+    fn args_at_most_wildcard_or_literal_one(args: &[Expr]) -> bool {
+        match args {
+            [] => true,
+            [Expr::Literal(ScalarValue::Int64(Some(v)), _)] => *v == 1,
+            [Expr::Wildcard { .. }] => true,
+            _ => false,
+        }
```
evenyag, line 115: What'll be deprecated?

src/query/src/dist_plan/analyzer/test.rs
discord9: related datafusion issue here: https://github.com/apache/datafusion/issues/17166

src/query/src/query_engine/state.rs
```diff
@@ -152,9 +147,6 @@ impl QueryEngineState {
         physical_optimizer
             .rules
             .insert(1, Arc::new(PassDistribution));
-        physical_optimizer
-            .rules
-            .insert(2, Arc::new(EnforceSorting {}));
```
evenyag, line 157: Is it fine to remove these rules?

src/query/src/range_select/plan.rs
```diff
@@ -545,6 +544,7 @@ impl RangeSelect {
                 // At this time, aggregate plan has been replaced by a custom range plan,
                 // so `CountWildcardRule` has not been applied.
                 // We manually modify it when creating the physical plan.
+                #[allow(deprecated)]
```
evenyag: Do we need to track this in the issue?

src/query/src/window_sort.rs
```diff
@@ -238,7 +238,7 @@ impl ExecutionPlan for WindowedSortExec {
     /// and is expected to run directly on storage engine's output
     /// distribution / partition.
     fn benefits_from_input_partitioning(&self) -> Vec<bool> {
-        vec![false; self.ranges.len()]
+        vec![false]
```
evenyag, line 241: Is it correct to return a vec with only one element? @discord9

src/table/src/table/scan.rs
```diff
@@ -332,7 +330,11 @@ impl ExecutionPlan for RegionScanExec {
         Some(self.metric.clone_inner())
     }
 
-    fn statistics(&self) -> DfResult<Statistics> {
+    fn partition_statistics(&self, partition: Option<usize>) -> DfResult<Statistics> {
+        if partition.is_some() {
+            return Ok(Statistics::new_unknown(self.schema().as_ref()));
+        }
```
evenyag, line 336: Will this affect the optimization for count in append mode?

tests/cases/distributed/explain/step_aggr_advance.result
```diff
@@ -80,6 +84,11 @@ tql analyze (1752591864, 1752592164, '30s') max by (a, b, c) (max_over_time(aggr
 |_|_|_PromRangeManipulateExec: req range=[1752591864000..1752592164000], interval=[30000], eval range=[120000], time index=[greptime_timestamp] REDACTED
 |_|_|_PromSeriesNormalizeExec: offset=[0], time index=[greptime_timestamp], filter NaN: [true] REDACTED
 |_|_|_PromSeriesDivideExec: tags=["a", "b", "c", "d"] REDACTED
+|_|_|_SortExec: expr=[a@0 ASC, b@1 ASC, c@2 ASC, d@3 ASC, greptime_timestamp@4 ASC], preserve_partitioning=[true] REDACTED
+|_|_|_CoalesceBatchesExec: target_batch_size=8192 REDACTED
+|_|_|_RepartitionExec: partitioning=REDACTED
+|_|_|_CoalescePartitionsExec REDACTED
+|_|_|_CooperativeExec REDACTED
```
evenyag, line 91: This has some performance impact on promql queries. How can we fix it? @waynexia @discord9

tests/cases/standalone/common/types/string/unicode.result
```diff
@@ -19,7 +19,10 @@ SELECT * FROM emojis ORDER BY id;
 +----+--------+-------------------------+
 
 -- substr on unicode
-SELECT substr(s, 1, 1), substr(s, 2, 1) FROM emojis ORDER BY id;
+-- Because "substr" always use "Utf8View" as its return type, this query will return error
+-- "column types must match schema types, expected Utf8View but found Utf8" if the "CAST" is not used.
+-- TODO(LFC): support "Utf8View" Arrow array so that this "CAST" walkaround can be removed.
+SELECT CAST(substr(s, 1, 1) AS STRING), CAST(substr(s, 2, 1) AS STRING) FROM emojis ORDER BY id;
```
evenyag, line 25: out-of-topic: Maybe we can switch to using Utf8View by default.

tests/cases/standalone/tql-explain-analyze/analyze.result
```diff
@@ -272,14 +339,26 @@ TQL ANALYZE sum(test2);
...
+|_|_|_CooperativeExec REDACTED
 |_|_|_MergeScanExec: REDACTED
 |_|_|_|
 | 1_| 0_|_PromInstantManipulateExec: range=[0..0], lookback=[300000], interval=[300000], time index=[greptime_timestamp] REDACTED
 |_|_|_PromSeriesDivideExec: tags=["shard"] REDACTED
+|_|_|_SortExec: expr=[shard@2 ASC, greptime_timestamp@0 ASC], preserve_partitioning=[true] REDACTED
+|_|_|_CoalesceBatchesExec: target_batch_size=8192 REDACTED
+|_|_|_RepartitionExec: partitioning=REDACTED
+|_|_|_CoalescePartitionsExec REDACTED
+|_|_|_CooperativeExec REDACTED
```
discord9, line 352: This `CooperativeExec` could be blocking later optimize `SortExec`? Might need a issue to investigate this after pr is merged

src/query/src/query_engine/state.rs
discord9: > Is it fine to remove these rules?

Oh I think this `EnforceSorting` remove unnecessary sorts, see: https://docs.rs/datafusion/latest/datafusion/physical_optimizer/enforce_sorting/struct.EnforceSorting.html

src/query/src/window_sort.rs
discord9: > Is it correct to return a vec with only one element? @discord9

I think this `Vec<bool>`'s length need to be equal to input plan number, so it's ok?

tests/cases/distributed/explain/step_aggr_advance.result
```diff
@@ -534,8 +610,23 @@ tql explain (1752591864, 1752592164, '30s') sum by (a, b, c) (rate(aggr_optimize
...
+|               |                             CooperativeExec                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
+|               |                               MergeScanExec: REDACTED
```
discord9, line 629: I'm confused why this large chunk of new execute plan exist now

src/common/function/src/aggrs/aggr_wrapper.rs
discord9: > Is it ok to use `x` as field names?

probably fine

src/flow/src/batching_mode/task.rs
discord9: > Can we remove this? @discord9

Yes, seems optimize `count(*)` away before here

src/query/src/dist_plan/analyzer.rs
MichaelScofield: @evenyag not sure how to fix it, can you give some hints?

datafusion requires the size of `exprs` must be at least two times of `on`'s size:

```rust
// inside datafusion's with_new_exprs method:

let equi_expr_count = on.len() * 2;
assert!(expr.len() >= equi_expr_count);
```

while the `exprs` seems very unlikely given it is produced by folding the `on`'s exprs then chaining some filters.

tests/cases/standalone/tql-explain-analyze/analyze.result
MichaelScofield: you mean our window sort rule?

src/promql/src/extension_plan/empty_metric.rs
MichaelScofield: see how the schema is set in datafusion's `SimplifyExpressions`:

```rust
let schema = if !plan.inputs().is_empty() {
    DFSchemaRef::new(merge_schema(&plan.inputs()))
} else if let LogicalPlan::TableScan(scan) = &plan {
    Arc::new(DFSchema::try_from_qualified_schema(
        scan.table_name.clone(),
        &scan.source.schema(),
    )?)
} else {
    Arc::new(DFSchema::empty())
};
```

src/query/src/optimizer/count_wildcard.rs
MichaelScofield: don't know, the proposal to deprecate wildcard expr in datafusion is a long awaited issue

src/table/src/table/scan.rs
MichaelScofield: no, the `partition` input is none for that optimization.

the optimization is guarded in sqlness. should it changed in datafusion, we will know

tests/cases/standalone/common/types/string/unicode.result
MichaelScofield: will do it in the follow-ups

tests/cases/standalone/tql-explain-analyze/analyze.result
discord9: > you mean our window sort rule?

Oh I get it wrong might be something else need look further

src/promql/src/extension_plan/empty_metric.rs
evenyag: You can provide this link in the comment.

src/common/datasource/src/file_format/orc.rs
MichaelScofield: just found out the `OrcOpener` is no longer used, I'll remove it

.github/workflows/develop.yml
```diff
@@ -58,6 +58,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
+
```
evenyag: Can we remove these empty newlines?

src/common/function/src/aggrs/aggr_wrapper.rs
MichaelScofield: I believe so. The original codes' intention was to find return values' types according to the function's input values' types. Which is already included in the the new `return_field` method. Please confirm @discord9

src/common/function/src/aggrs/aggr_wrapper.rs
MichaelScofield: to quote: "The name you specify on the field is not important." 

see datafusion's upgrading guide: https://github.com/apache/datafusion/blob/main/docs/source/library-user-guide/upgrading.md#processing-fieldref-instead-of-datatype-for-user-defined-functions

src/operator/src/statement/copy_table_from.rs
WenyXu: yes

src/mito2/src/sst/parquet/metadata.rs
MichaelScofield: you can record it here: https://github.com/GreptimeTeam/greptimedb/issues/6749

src/promql/src/functions/round.rs
MichaelScofield: record it in https://github.com/GreptimeTeam/greptimedb/issues/6749

src/query/src/lib.rs
MichaelScofield: in `planner.rs` line 103~108:

```rust
if let datafusion::sql::parser::Statement::Statement(
    box datafusion::sql::sqlparser::ast::Statement::Explain { format, .. },
) = &mut df_stmt
{
    format.take();
}
```

if it's superseded, compiler will warn us

src/object-store/src/layers.rs
MichaelScofield: the `path_label` is removed in opendal since 0.53, with a major refactor regarding its prometheus layer. didn't got the time to investigate the details. however, `path_label(0)` was basically doing nothing (see https://docs.rs/opendal/0.52.0/opendal/layers/struct.PrometheusLayerBuilder.html#method.path_label), and we don't change anything about the metrics paths(do we?), so I think it's probably fine to simply remove the call. 

I've also started the greptimedb to see the opendal's metrics, and they stay the same before and after:

```text
# HELP opendal_operation_bytes Current operation size in bytes, represents the size of data being processed in the current operation
# TYPE opendal_operation_bytes histogram
opendal_operation_bytes_bucket{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs",le="4096"} 1
opendal_operation_bytes_bucket{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs",le="65536"} 1
opendal_operation_bytes_bucket{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs",le="262144"} 1
opendal_operation_bytes_bucket{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs",le="1048576"} 1
opendal_operation_bytes_bucket{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs",le="4194304"} 1
opendal_operation_bytes_bucket{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs",le="16777216"} 1
opendal_operation_bytes_bucket{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs",le="67108864"} 1
opendal_operation_bytes_bucket{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs",le="268435456"} 1
opendal_operation_bytes_bucket{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs",le="1073741824"} 1
opendal_operation_bytes_bucket{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs",le="5368709120"} 1
opendal_operation_bytes_bucket{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs",le="+Inf"} 1
opendal_operation_bytes_sum{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs"} 403
opendal_operation_bytes_count{namespace="",operation="write",root="/private/tmp/greptimedb_data",scheme="fs"} 1
```

src/query/src/range_select/plan.rs
MichaelScofield: I change it to `#[expect(deprecated)]`. once the deprecation state is changed, compiler will let us know

src/query/src/dist_plan/analyzer.rs
evenyag: Can you try to remove the special handling for Join node here?
https://github.com/GreptimeTeam/greptimedb/blob/351826cd32f7e11d8a1fad83ab0b96191fb7ae01/src/query/src/plan.rs#L134-L152


```
fn expressions_consider_join(&self) -> Vec<Expr> {
    self.expressions()
}
```

MichaelScofield: it works, but are you sure to remove those codes?

src/query/src/plan.rs
```diff
@@ -132,24 +132,7 @@ pub trait ExtractExpr {
 
 impl ExtractExpr for LogicalPlan {
     fn expressions_consider_join(&self) -> Vec<Expr> {
-        match self {
```
discord9, line 135: why remove this though, does it cause panic or incorrect results? do we need to add it to todo list in #6749 ?

src/query/src/dist_plan/analyzer.rs
evenyag: Yes, it was a workaround before https://github.com/apache/datafusion/pull/15055/

src/query/src/plan.rs
evenyag: It was a workaround before https://github.com/apache/datafusion/pull/15055.

discord9 (approved):
LGTM, but I think we need a way to keep track of those sqlness result change and resolve them later

sunng87:
Can we confirm if there is any user space breaking change?

MichaelScofield:
@sunng87 I can think of two minor ones:
1. `substr` function needs to be "cast"ed: `SELECT substr(s, 1, 1) FROM t` => `SELECT cast(substr(s, 1, 1) as string) FROM t`
2. "`service`" is now a reserved word in sqlparser, must be quoted in SQL

Issue #6177, opened 2025-05-25
No installed planner for sub query


killme2008:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Query Engine

### Minimal reproduce step

A simple reproduce case:

```sql
mysql> create table monitor (ts timestamp time index, val int);
Query OK, 0 rows affected (0.03 sec)

mysql> insert into monitor values(now(), 1), (now(), 2);
Query OK, 2 rows affected (0.02 sec)

mysql>  select * from monitor where val in (select val from monitor) and ts >=now();
ERROR 1815 (HY000): (EngineExecuteQuery): Error during planning: No installed planner was able to convert the custom node to an execution plan: MergeScanLogicalPlan { input: Projection(Projection { expr: [Column(Column { relation: Some(Bare { table: "monitor" }), name: "val" })], input: TableScan(TableScan { table_name: Bare { table: "monitor" }, source: "...", projection: None, projected_schema: DFSchema { inner: Schema { fields: [Field { name: "ts", data_type: Timestamp(Millisecond, None), nullable: false, dict_id: 0, dict_
```

The query ` select * from monitor where val in (select val from monitor) and ts >=now();` failed.

### What did you expect to see?

The query success

### What did you see instead?

Failed

### What operating system did you use?

All

### What version of GreptimeDB did you use?

v0.14.3

### Relevant log output and stack trace

```bash

```

killme2008:
@waynexia Please take a look. The issue still exists 0.15, but the error message changed:

```text
ERROR 1815 (HY000): (EngineExecuteQuery): Error during planning: No installed planner was able to convert the custom node to an execution plan: MergeScanLogicalPlan { input: Projection(Projection { expr: [Column(Column { relation: Some(Bare { table: "monitor" }), name: "val" })], input: TableScan(TableScan { table_name: Bare { table: "monitor" }, source: "...", projection: None, projected_schema: DFSchema { inner: Schema { fields: [Field { name: "ts", data_type: Timestamp(Millisecond, None), nullable: false, dict_id: 0, dict_
```

sunheyi6:
> [@waynexia](https://github.com/waynexia) Please take a look. The issue still exists 0.15, but the error message changed:
> 
> ```
[...]

PR #6687, opened 2025-08-07
feat: add rule to convert scalar subquery to join

sunheyi6:

close https://github.com/GreptimeTeam/greptimedb/issues/6177
## PR Checklist


evenyag:
> No installed planner was able to convert the custom node to an execution plan

The error message in the issue is about no installed planner. Could you please explain why we need a new rule? @sunheyi6 

cc @waynexia @killme2008

waynexia:
Thank you. Could you please add some sqlness cases https://github.com/GreptimeTeam/greptimedb/tree/main/tests#sqlness-manual for this scenario?

sunheyi6:
> > No installed planner was able to convert the custom node to an execution plan
> 
> The error message in the issue is about no installed planner. Could you please explain why we need a new rule? @sunheyi6
> 
> cc @waynexia @killme2008

I'm sorry. Maybe I misunderstood. Then how can I add a planner?

evenyag:
> > > No installed planner was able to convert the custom node to an execution plan
> > 
> > 
> > The error message in the issue is about no installed planner. Could you please explain why we need a new rule? @sunheyi6
> > cc @waynexia @killme2008
> 
> I'm sorry. Maybe I misunderstood. Then how can I add a planner?

I'm just curious about how this fix works.

I'm not sure why the SQL in https://github.com/GreptimeTeam/greptimedb/issues/6177 failed in the current version. I only found that the error message mentioned in the issue is:
```
ERROR 1815 (HY000): (EngineExecuteQuery): Error during planning: No installed planner was able to convert the custom node to an execution plan: MergeScanLogicalPlan { input: Projection(Projection { expr: [Column(Column { relation: Some(Bare { table: "monitor" }), name: "val" })], input: TableScan(TableScan { table_name: Bare { table: "monitor" }, source: "...", projection: None, projected_schema: DFSchema { inner: Schema { fields: [Field { name: "ts", data_type: Timestamp(Millisecond, None), nullable: false, dict_id: 0, dict_
```

I don't know why it throws this error. This PR doesn't explain this.

So I have a few questions:
1. What's the root cause of that issue?
2. How does this rule fix that issue?

Could you please provide more context about this fix in the description?

killme2008:
Agreed. We need to investigate the root cause, then discuss potential fixes.

FYI: Please read the contribution guide and always include tests with your PRs.

yihong0618:
This can not compile

sunheyi6:
> Agreed. We need to investigate the root cause, then discuss potential fixes.
> 
> FYI: Please read the contribution guide and always include tests with your PRs.

​​The problem is the lack of subquery handling, which causes the subquery to fail directly.​

evenyag:
Datafusion returns this error when all extension planners return None
https://github.com/apache/datafusion/blob/43718d5fff20c2dc048db820491633aace2498f6/datafusion/core/src/physical_planner.rs#L1249-L1272

However, the dist planner should handle the MergeScanLogicalPlan.
https://github.com/GreptimeTeam/greptimedb/blob/dea87b7e57efde412a6458a9e8dcaeaff2e9c90e/src/query/src/dist_plan/planner.rs#L119-L131

sunheyi6:
> Datafusion returns this error when all extension planners return None https://github.com/apache/datafusion/blob/43718d5fff20c2dc048db820491633aace2498f6/datafusion/core/src/physical_planner.rs#L1249-L1272
> 
> However, the dist planner should handle the MergeScanLogicalPlan.
> 
> https://github.com/GreptimeTeam/greptimedb/blob/dea87b7e57efde412a6458a9e8dcaeaff2e9c90e/src/query/src/dist_plan/planner.rs#L119-L131

I'm sorry, but I still don't quite understand what role this code plays in subqueries. Could you please explain it in more detail? I'm not very familiar with the logic here.

evenyag:
> I'm sorry, but I still don't quite understand what role this code plays in subqueries. Could you please explain it in more detail? I'm not very familiar with the logic here.

This does nothing special for subqueries. I'm worried that it's not just the subquery that can cause this error.

I'd suggest reading the documents of datafusion and GreptimeDB, then dive into the related code. Datafusion has some examples and tests to demonstrate the usage.
- https://github.com/apache/datafusion/blob/49.0.0/datafusion-examples/examples/plan_to_sql.rs
- https://docs.rs/datafusion/latest/datafusion/physical_planner/trait.ExtensionPlanner.html
- https://docs.greptime.com/contributor-guide/frontend/distributed-querying/

You can add some logs to debug and learn how a query executes. Without the knowledge of datafusion and our query engine, it's hard to understand and debug the issue.

I'm going to close this PR later since the root cause is still unclear. We can update the issue if we have further progress.

PR #6692, opened 2025-08-08
fix: correctly set extension range source index

MichaelScofield:
refactor: extract the common codes of creating proto ColumnSchema and Row to helper functions

fix: explicitly set the follower max sequence when finding extension ranges to avoid potential concurrency hazard

the extension range source index in the StreamContext was not correctly set to the extension ranges array index in the ScanInput, which caused out of bound error during scanning

also refactor some codes

also eliminate a potential concurrency hazard in finding follower max sequence

## PR Checklist

### DIFF
--- a/src/api/src/v1.rs
+++ b/src/api/src/v1.rs
@@ -14,6 +14,8 @@
 
 pub mod column_def;
 
+pub mod helper;
+
 pub mod meta {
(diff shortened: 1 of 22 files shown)


fengjiachun (approved):
LGTM

Issue #6680, opened 2025-08-07
`count()` returns multiple rows


WenyXu:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Storage Engine

### Minimal reproduce step

1. Create physical table
```
CREATE TABLE IF NOT EXISTS `greptime_physical_table` (
    greptime_timestamp TIMESTAMP(3) NOT NULL,
    greptime_value DOUBLE NULL,
    `namespace` STRING NULL INVERTED INDEX,
    PRIMARY KEY (`namespace`),
    TIME INDEX (greptime_timestamp),
)
PARTITION ON COLUMNS (`namespace`) (
    `namespace` < 'app-0',                                                                       
    `namespace` >= 'app-0' AND `namespace` < 'app-1',
    `namespace` >= 'app-1' AND `namespace` < 'app-2',
    `namespace` >= 'app-2' AND `namespace` < 'app-3',
    `namespace` >= 'app-3' AND `namespace` < 'app-4',
    `namespace` >= 'app-4' AND `namespace` < 'app-5',
    `namespace` >= 'app-5' AND `namespace` < 'app-6',
    `namespace` >= 'app-6' AND `namespace` < 'app-7',
    `namespace` >= 'app-7' AND `namespace` < 'app-8',
    `namespace` >= 'app-8'
)
engine = metric with (
    "physical_metric_table" = "",
    "memtable.type" = "partition_tree",
    "index.type" = "inverted",
);

```

2. Create logical table
CREATE TABLE IF NOT EXISTS `metrics_table_0` (
  `column_1035` STRING NULL,
  `column_422` STRING NULL,
  `column_747` STRING NULL,
  `greptime_timestamp` TIMESTAMP(3) NOT NULL,
  `greptime_value` DOUBLE NULL,
  TIME INDEX (`greptime_timestamp`),
  PRIMARY KEY (`column_1035`, `column_422`, `column_747`)
)
ENGINE=metric
WITH(
  on_physical_table = 'greptime_physical_table'
);

3. Run ruery
```
select count() from metrics_table_0;
```

### What did you expect to see?

```
+---------+
| count() |
+---------+
|       0 |
+---------+
```

### What did you see instead?

```
+---------+
| count() |
+---------+
|       0 |
|       0 |
|       0 |
|       0 |
|       0 |
|       0 |
|       0 |
|       0 |
|       0 |
|       0 |
+---------+
```

### What operating system did you use?

Doesn't matter

[...]

PR #6694, opened 2025-08-08
fix: metrics without physical partition columns query push down

discord9:

https://github.com/GreptimeTeam/greptimedb/issues/6680

fix  query part of metrics table that doesn't have physical table's partition columns should still prevent certain push down, impl by add a placeholder column that never appear hence prevent those push down

metrics table inherit physical table's partition column should be rethink however, by auto add those columns while creating logical table maybe

**this is a temporary workaround for backward compatibility**

## PR Checklist

### DIFF
--- a/src/catalog/src/kvbackend/manager.rs
+++ b/src/catalog/src/kvbackend/manager.rs
@@ -132,7 +132,7 @@ impl KvBackendCatalogManager {
         {
             let mut new_table_info = (*table.table_info()).clone();
 
-            let mut phy_cols_not_in_logical_table = vec![];
+            let mut phy_part_cols_not_in_logical_table = vec![];
 
             // Remap partition key indices from physical table to logical table
             new_table_info.meta.partition_key_indices = physical_table_info_value
@@ -156,7 +156,8 @@ impl KvBackendCatalogManager {
                                 .column_index_by_name(physical_column.name.as_str());
                             if idx.is_none() {
                                 // not all part columns in physical table that are also in logical table
-                                phy_cols_not_in_logical_table.push(physical_column.name.clone());
+                                phy_part_cols_not_in_logical_table
+                                    .push(physical_column.name.clone());
                             }
 
                             idx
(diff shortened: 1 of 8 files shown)


src/catalog/src/kvbackend/manager.rs
```diff
@@ -148,14 +150,32 @@ impl KvBackendCatalogManager {
...
+                                phy_part_cols_not_in_logical_table
+                                    .push(physical_column.name.clone());
+                            }
+
+                            idx
                         })
                 })
                 .collect();
 
+            if !phy_part_cols_not_in_logical_table.is_empty() {
```
WenyXu: How about adding this flag into `Table` struct instead of options? like:

```rust
struct Table {
    table_info: TableInfoRef,
    filter_pushdown: FilterPushDownType,
    data_source: DataSourceRef,
    /// Columns default [`Expr`]
    column_defaults: HashMap<String, Expr>,
    extra_attributes: Vec<ExtraAttribute>
}

```

WenyXu (commented):
Rest LGTM

src/query/src/dist_plan/analyzer.rs
```diff
@@ -482,14 +482,32 @@ impl PlanRewriter {
...
                             .into_iter()
                             .map(|index| schema.column_name_by_index(index).to_string())
                             .collect::<Vec<String>>();
+
+                        let partition_rules = table.partition_rules();
+                        let exist_phy_part_cols_not_in_logical_table = partition_rules
+                            .map(|r| !r.extra_phy_cols_not_in_logical_table.is_empty())
+                            .unwrap_or(false);
+
+                        if exist_phy_part_cols_not_in_logical_table && partition_cols.is_empty() {
```
evenyag, line 500: Do we have tests for other cases here? e.g. `exist_phy_part_cols_not_in_logical_table && !partition_cols.is_empty()`.

discord9: > Do we have tests for other cases here? e.g. `exist_phy_part_cols_not_in_logical_table && !partition_cols.is_empty()`.

logical_table_1 and 2 are both this case, because phy table have three partition columns

evenyag (approved):

waynexia (approved):

PR #6697, opened 2025-08-09
chore: impl default for SlowQueriesRecordType and LogFormat

killme2008:

This PR implements default traits and custom deserialization for configuration enums to handle empty string values gracefully during deserialization. The changes allow configuration fields to use default values when empty strings are provided, improving the robustness of configuration parsing.

Key changes:

* Added Default trait implementations with `#[default]` attributes for `SlowQueriesRecordType`, `LogFormat`, `OtlpExportProtocol`, and `CompressionType` enums
* Creates a new procedural macro `DeserializeWithEmptyDefault` that generates custom `Deserialize` implementations treating empty strings as default values
* Added comprehensive tests to verify the deserialization behavior for all affected enums

## PR Checklist

### DIFF
--- a/src/common/datasource/src/compression.rs
+++ b/src/common/datasource/src/compression.rs
@@ -19,6 +19,7 @@ use std::str::FromStr;
 use async_compression::tokio::bufread::{BzDecoder, GzipDecoder, XzDecoder, ZstdDecoder};
 use async_compression::tokio::write;
 use bytes::Bytes;
+use common_macro::DeserializeWithEmptyDefault;
 use datafusion::datasource::file_format::file_compression_type::FileCompressionType;
 use futures::Stream;
 use serde::Serialize;
@@ -28,7 +29,18 @@ use tokio_util::io::{ReaderStream, StreamReader};
 
 use crate::error::{self, Error, Result};
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EnumIter, Serialize, Default)]
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    Hash,
+    EnumIter,
+    Serialize,
+    Default,
+    DeserializeWithEmptyDefault,
+)]
 #[serde(rename_all = "lowercase")]
 pub enum CompressionType {
     /// Gzip-ed file
@@ -43,15 +55,6 @@ pub enum CompressionType {
     #[default]
     Uncompressed,
 }
-
-common_macro::impl_deserialize_with_empty_default!(
-    CompressionType,
-    Gzip => "gzip",
-    Bzip2 => "bzip2",
-    Xz => "xz",
-    Zstd => "zstd",
-    Uncompressed => "uncompressed",
-);
 
 impl FromStr for CompressionType {
     type Err = Error;
--- a/src/common/datasource/src/compression.rs
+++ b/src/common/datasource/src/compression.rs
@@ -215,3 +215,48 @@ impl From<CompressionType> for FileCompressionType {
         }
     }
 }
+#[cfg(test)]
+mod deserialization_tests {
+    use serde::Deserialize;
+
+    use super::*;
+
+    #[derive(Deserialize)]
+    struct Wrapper {
+        #[serde(default)]
+        compression: CompressionType,
+    }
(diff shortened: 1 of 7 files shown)


src/common/telemetry/src/logging.rs
```diff
@@ -125,17 +125,19 @@ impl Default for SlowQueryOptions {
     }
 }
 
-#[derive(Clone, Debug, Serialize, Deserialize, Copy, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize, Copy, PartialEq, Default)]
```
evenyag: You may also need `#[serde(default)]`. Maybe adding some test cases?

zyy17:
Although the `SlowQueryOptions` has the default options, it's more complete to add a default value for sub-fields.

src/common/telemetry/src/logging.rs
killme2008: Addressed https://github.com/GreptimeTeam/greptimedb/pull/6697/commits/53226374a0e9c6be93dad1429e9285ed6dcc889d

sunng87:
I'm not big fan of implementing this as procedural macro, the macro is difficult to read and maintain, and even difficult for rustfmt to format. If we want to do it as macro, a declarative macro like #[derive(MyDeserialize)] is better: we don't need to type the string representations of the enum repeatly.

Also, serde has a mechanism called [custom deserializer](https://serde.rs/field-attrs.html#deserialize_with) for this.

```rust
fn deserialize_enum_or_default<'de, D, T>(deserializer: D) -> Result<T, D::Error>
where
    D: Deserializer<'de>,
    T: Default + Deserialize<'de>,
{
    let s = String::deserialize(deserializer)?;
    
    if s.is_empty() {
        Ok(T::default())
    } else {
        // Try to deserialize the string as the enum
        T::deserialize(de::value::StringDeserializer::new(s))
            .map_err(de::Error::custom)
    }
}
```

src/common/datasource/src/compression.rs
```diff
@@ -205,3 +215,48 @@ impl From<CompressionType> for FileCompressionType {
...
+
+        for (json, expected) in cases {
+            let wrapper: Wrapper = serde_json::from_str(json).unwrap();
+            assert_eq!(wrapper.compression, expected, "Failed for json: {}", json);
+        }
+    }
+
+    #[test]
+    fn test_deserialize_empty_string_defaults_to_uncompressed() {
+        let json = r#"{"compression": ""}"#;
```
evenyag, line 254: Do we need to support an empty string as compression?

killme2008: Yes, and which is `uncompression`.

Looks like it's used in manifest.

killme2008:
> I'm not big fan of implementing this as procedural macro, the macro is difficult to read and maintain, and even difficult for rustfmt to format. If we want to do it as macro, a declarative macro like #[derive(MyDeserialize)] is better: we don't need to type the string representations of the enum repeatly.
> 
> Also, serde has a mechanism called [custom deserializer](https://serde.rs/field-attrs.html#deserialize_with) for this.
> 
> ```rust
> fn deserialize_enum_or_default<'de, D, T>(deserializer: D) -> Result<T, D::Error>
> where
>     D: Deserializer<'de>,
>     T: Default + Deserialize<'de>,
> {
>     let s = String::deserialize(deserializer)?;
>     
>     if s.is_empty() {
>         Ok(T::default())
>     } else {
>         // Try to deserialize the string as the enum
>         T::deserialize(de::value::StringDeserializer::new(s))
>             .map_err(de::Error::custom)
>     }
> }
> ```

Agreed. Addressed in https://github.com/GreptimeTeam/greptimedb/pull/6697/commits/968947373cbb766f230ee215dd2a42c8661eef37

src/common/datasource/src/compression.rs
killme2008: The behaviour is the same as `from_str`:
https://github.com/GreptimeTeam/greptimedb/blob/5eb491df126fe34fa55c11e9dabb39e9ed57c955/src/common/datasource/src/compression.rs#L56

killme2008:
Over design. 

I'll create a new PR for it.

Issue #6703, opened 2025-08-11
feat: make promql 100%


yihong0618:
### What type of enhancement is this?

User experience

### What does the enhancement do?

how to:

- git clone https://github.com/promlabs/promql-compliance-tester.git
- go build -o promql-compliance-tester ./cmd/promql-compliance-tester/
- start greptime standalone
- ./promql-compliance-tester

make promql 100%

ref: #1042

for now:

```cli

INFO[0002] Test execution summary:                       source="main.go:127"
INFO[0002]   Total test cases: 529                       source="main.go:128"
INFO[0002]   Successful: 494 (93.38%)                    source="main.go:129"
INFO[0002]   Failed: 35 (6.62%)                          source="main.go:130"

=== FAILED PROMQL QUERIES (Copy-paste ready) ===
{type="free", instance!="demo.promlabs.com:10000"}
demo_memory_usage_bytes offset -1m
demo_memory_usage_bytes offset -5m
demo_memory_usage_bytes offset -10m
timestamp(demo_memory_usage_bytes * 1)
timestamp(-demo_memory_usage_bytes)
time() % 1
time() * time()
time() % time()
abs(-demo_memory_usage_bytes)
ceil(-demo_memory_usage_bytes)
floor(-demo_memory_usage_bytes)
exp(-demo_memory_usage_bytes)
sqrt(-demo_memory_usage_bytes)
ln(-demo_memory_usage_bytes)
log2(-demo_memory_usage_bytes)
log10(-demo_memory_usage_bytes)
round(-demo_memory_usage_bytes)
label_replace(demo_num_cpus, "job", "value-$1", "src", "(.*")
label_replace(demo_num_cpus, "~invalid", "", "src", "(.*)")
label_replace(demo_num_cpus, "instance", "", "", "")
label_join(demo_num_cpus, "new_label", "-", "instance", "job")
label_join(demo_num_cpus, "job", "-", "instance", "job")
label_join(demo_num_cpus, "job", "-", "instance")
histogram_quantile(-0.5, rate(demo_api_request_duration_seconds_bucket[1m]))
histogram_quantile(0.1, rate(demo_api_request_duration_seconds_bucket[1m]))
histogram_quantile(0.5, rate(demo_api_request_duration_seconds_bucket[1m]))
histogram_quantile(0.75, rate(demo_api_request_duration_seconds_bucket[1m]))
[...]

PR #6704, opened 2025-08-11
fix: time() function should the same as behavior prometheus

### LINKED ISSUE #6701: promsql for time() is different from Prometheus
### What type of bug is this?

Incorrect result

Query Engine

### Minimal reproduce step

for Prometheus
Image

but for greptime is 0

same behavior as Prometheus

### What did you see instead?

```cli
mysql> tql eval time();
+---------------------+----------------------+
| time                | time / Float64(1000) |
+---------------------+----------------------+
| 1970-01-01 00:00:00 |                    0 |
+---------------------+----------------------+
1 row in set (0.041 sec)
```

mysql> 

0

no matter

all

```bash

```

yihong0618:

ref #6703

close #6701

This patch make promql `time()` the same as Prometheus 
as doc https://prometheus.io/docs/prometheus/latest/querying/functions/#time

also this patch fix threes promql tests that it do not support

```cli
tql eval time() % 1
tql eval time() * time()
tql eval time() % time()
```

This three also can work after this patch

## PR Checklist


yihong0618:
also change some tests, maybe need to discuss

tests/cases/standalone/common/promql/time_fn.result
```diff
@@ -2,39 +2,24 @@
...
+-- Test that time() is consistent within the same evaluation
+tql eval (3000, 3000, '1s') time() > bool 1577836800; -- 2020-01-01
+tql eval (0, 0, '1s') time() == bool time();
 
-+---------------------+----------------------+
-| time                | time / Float64(1000) |
-+---------------------+----------------------+
-| 1970-01-01T00:50:00 | 3000.0               |
-+---------------------+----------------------+
+Error: 2000(InvalidSyntax), unexpected character: ';'
```
sunng87: error

src/query/src/promql/planner.rs
```diff
@@ -2602,7 +2599,14 @@ impl PromPlanner {
...
             PromExpr::Call(Call { func, .. }) => {
                 if func.name == SPECIAL_TIME_FUNCTION {
-                    Some(build_special_time_expr(SPECIAL_TIME_FUNCTION))
+                    // For the time() function, return current evaluation time as a literal.
+                    // According to Prometheus documentation: https://prometheus.io/docs/prometheus/latest/querying/functions/#time
+                    // time() returns the number of seconds since January 1, 1970 UTC.
+                    let now = std::time::SystemTime::now()
+                        .duration_since(std::time::UNIX_EPOCH)
+                        .unwrap()
+                        .as_secs_f64();
```
sunng87: We can extract a function for this

tests/cases/standalone/common/promql/time_fn.result
yihong0618: ...my fault

src/query/src/promql/planner.rs
yihong0618: copy

sunng87:
I just checked protmetheus docs about `time()`: https://prometheus.io/docs/prometheus/latest/querying/functions/#time

It turns the time that it's evaluated.

And the graph shows that as well
image

yihong0618:
> I just checked protmetheus docs about `time()`: https://prometheus.io/docs/prometheus/latest/querying/functions/#time
> 
> It turns the time that it's evaluated.
> 
> And the graph shows that as well image

fully understand it, you are right

for time() function its two behaviors

- when barely return instant current timestamp
- when range query its need math to evaluate

will change the code and all tests

yihong0618:
old logic is right.

the problem is that `tql eval time();` it is a range query start default 0 end default 0 and interval 300000

I do not know if it is right, so left the latest commit here waiting for discuss

cc @waynexia

waynexia:
>the problem is that tql eval time(); it is a range query start default 0 end default 0 and interval 300000

So the problem now is about default parameters of TQL EVAL. To be honest, I just knew we could omit them yesterday... I'm OK to remove this logic. It should not be a frequent use case. 

We do allow omitting them for TQL EXPLAIN, I'm thinking about if this is added to TQL EVAL (and maybe TQL ANALYZE) unexpectedly.

yihong0618:
> > the problem is that tql eval time(); it is a range query start default 0 end default 0 and interval 300000
> 
> So the problem now is about default parameters of TQL EVAL. To be honest, I just knew we could omit them yesterday... I'm OK to remove this logic. It should not be a frequent use case.
> 
> We do allow omitting them for TQL EXPLAIN, I'm thinking about if this is added to TQL EVAL (and maybe TQL ANALYZE) unexpectedly.

yes this is the first problem 

the sencond one is that we can not run this `tql eval time() * time()` this pull request solved

and I think you can continue commit on this pull request about 
> To be honest, I just knew we could omit them yesterday... I'm OK to remove this logic. It should not be a frequent use case.


Thank you very much

waynexia (approved):

waynexia:
`time() * time()` is fixed now, thanks @yihong0618. Would you be willing to make a follow-up PR to prohibit using `TQL EVAL/ANALYZE` without parameters?

yihong0618:
> `time() * time()` is fixed now, thanks @yihong0618. Would you be willing to make a follow-up PR to prohibit using `TQL EVAL/ANALYZE` without parameters?

of course

sunng87 (approved):

Issue #6707, opened 2025-08-11
timestamp function for promql should support unknown


yihong0618:
### What type of enhancement is this?

User experience

### What does the enhancement do?

<img width="490" height="376" alt="Image" src="https://github.com/user-attachments/assets/2f33ef58-d944-4356-b5e3-c50cc695f332" />

timestamp(demo_memory_usage_bytes * 1)
timestamp(-demo_memory_usage_bytes)

PR #6708, opened 2025-08-11
fix: support unknown for timestamp function

yihong0618:

#6707 

also fix all these:

```
abs(-demo_memory_usage_bytes)
ceil(-demo_memory_usage_bytes)
floor(-demo_memory_usage_bytes)
exp(-demo_memory_usage_bytes)
sqrt(-demo_memory_usage_bytes)
ln(-demo_memory_usage_bytes)
log2(-demo_memory_usage_bytes)
log10(-demo_memory_usage_bytes)
round(-demo_memory_usage_bytes)
```

cc @waynexia 

as diff

-->

## PR Checklist


yihong0618:
some sql now in math has no error!

src/query/src/promql/planner.rs
```diff
@@ -1552,23 +1552,36 @@ impl PromPlanner {
...
-                    if result.input.replace(*arg.clone()).is_some() {
-                        MultipleVectorSnafu { expr: *arg.clone() }.fail()?;
+            // First try to parse as literal expression (including binary expressions like 100.0 + 3.0)
+            if Self::try_build_literal_expr(arg).is_some() {
+                let expr =
+                    Self::get_param_as_literal_expr(&Some(Box::new(*arg.clone())), None, None)?;
+                result.literals.push(expr);
+            } else {
+                // If not a literal, treat as vector input
+                match *arg.clone() {
```
sunng87: check if clone is still necessary here because we don't use any value in match

waynexia (approved):
LGTM, thank you @yihong0618 💯

src/query/src/promql/planner.rs
yihong0618: addressed

src/query/src/promql/planner.rs
```diff
@@ -1560,23 +1560,31 @@ impl PromPlanner {
...
-                | PromExpr::MatrixSelector(_)
-                | PromExpr::Extension(_)
-                | PromExpr::Aggregate(_)
-                | PromExpr::Paren(_)
-                | PromExpr::Call(_) => {
-                    if result.input.replace(*arg.clone()).is_some() {
-                        MultipleVectorSnafu { expr: *arg.clone() }.fail()?;
+            // First try to parse as literal expression (including binary expressions like 100.0 + 3.0)
+            if Self::try_build_literal_expr(arg).is_some() {
+                let expr = Self::get_param_as_literal_expr(&Some(arg.clone()), None, None)?;
```
killme2008: Can we directly use the `try_build_literal_expr` result? Avoid to call `get_param_as_literal_expr`. It calls `try_build_literal_expr` inside and check the data type, but we don't check data type here.

yihong0618: Done

PR #6695, opened 2025-08-08
feat: Implements last row dedup strategy for flat format

evenyag:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)
https://github.com/GreptimeTeam/greptimedb/issues/6505

## What's changed and what's your intention?

This PR defines the `RecordBatchDedupStrategy` trait and `DedupIterator` to dedup `RecordBatches` in flat format.

It adds the `FlatLastRow` strategy to dedup record batches by keeping the last row for the same key.

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [x] API changes are backward compatible.
- [x] Schema or data changes are backward compatible.

PR #6709, opened 2025-08-11
feat: Implements last-non-null dedup strategy for flat format

evenyag:

Follow-up of https://github.com/GreptimeTeam/greptimedb/pull/6695

Implements the `FlatLastNonNull` strategy for flat format.

The difference between this strategy and `FlatLastRow` is that `FlatLastNonNull` takes the latest non null value for each field.

However, it has to handle deletion specially to avoid a deleted row overriding the latest row's field. It tracks whether the duplicated rows of the last key contain a deletion. If they have, then we should not include non-null fields from the next batch.

It ports all tests from the old dedup strategy.

## PR Checklist

### DIFF
--- a/src/mito2/src/read/flat_dedup.rs
+++ b/src/mito2/src/read/flat_dedup.rs
@@ -386,10 +386,7 @@ impl FlatLastNonNull {
             return Ok((batch, contains_delete));
         }
 
-        let mut is_field = vec![false; batch.num_columns()];
-        // Iterates fields, skips internal columns.
-        is_field[field_column_start..batch.num_columns() - FIXED_POS_COLUMN_NUM].fill(true);
-
+        let field_column_end = batch.num_columns() - FIXED_POS_COLUMN_NUM;
         let take_options = Some(TakeOptions {
             check_bounds: false,
         });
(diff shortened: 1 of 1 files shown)


src/mito2/src/read/flat_dedup.rs
```diff
@@ -205,6 +210,264 @@ impl RecordBatchDedupStrategy for FlatLastRow {
...
+        let contains_delete = Self::last_range_has_delete(&ranges, op_types);
+
+        // Each range at least has 1 row.
+        let num_duplications: usize = ranges.iter().map(|r| r.end - r.start - 1).sum();
+        if num_duplications == 0 {
+            // Fast path, no duplication.
+            return Ok((batch, contains_delete));
+        }
+
+        let mut is_field = vec![false; batch.num_columns()];
```
fengjiachun: Maybe we can avoid allocating the vector?

```
let field_end = batch.num_columns() - FIXED_POS_COLUMN_NUM;

if col_idx >= field_column_start && col_idx < field_end {
    // is field
}
```

fengjiachun (approved):
Rest LGTM

v0y4g3r (approved):
LGTM

PR #6720, opened 2025-08-12
fix: two label_replace different from promql

yihong0618:

ref #6703

two label replace same as promlabs

image

## PR Checklist

### DIFF
--- a/src/query/src/promql/planner.rs
+++ b/src/query/src/promql/planner.rs
@@ -1921,10 +1921,7 @@ impl PromPlanner {
     fn validate_label_name(label_name: &str) -> Result<()> {
         // Check if label name starts with double underscores (reserved)
         if label_name.starts_with("__") {
-            return InvalidDestinationLabelNameSnafu {
-                label_name: label_name.to_string(),
-            }
-            .fail();
+            return InvalidDestinationLabelNameSnafu { label_name }.fail();
         }
         // Check if label name matches the required pattern
         if !LABEL_NAME_REGEX.is_match(label_name) {


src/query/src/promql/error.rs
```diff
@@ -207,6 +207,20 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
+
+    #[snafu(display("invalid regular expression in label_replace(): {}", regex))]
```
fengjiachun:
```suggestion
    #[snafu(display("Invalid regular expression in label_replace(): {}", regex))]
```

src/query/src/promql/error.rs
```diff
@@ -207,6 +207,20 @@ pub enum Error {
...
+    InvalidRegularExpression {
+        regex: String,
+        #[snafu(implicit)]
+        location: Location,
+    },
+
+    #[snafu(display("invalid destination label name in label_replace(): {}", label_name))]
```
fengjiachun:
```suggestion
    #[snafu(display("Invalid destination label name in label_replace(): {}", label_name))]
```

src/query/src/promql/planner.rs
```diff
@@ -1907,6 +1909,30 @@ impl PromPlanner {
         Ok((exprs, new_tags))
     }
 
+    /// Validate label name according to Prometheus specification.
+    /// Label names must match the regex: [a-zA-Z_][a-zA-Z0-9_]*
+    /// Additionally, label names starting with double underscores are reserved for internal use.
+    fn validate_label_name(label_name: &str) -> Result<()> {
+        // Check if label name matches the required pattern
+        let label_regex = regex::Regex::new(r"^[a-zA-Z_][a-zA-Z0-9_]*$").unwrap();
```
fengjiachun: Will it compile on every call? Can we use it as static regex?

src/query/src/promql/planner.rs
```diff
@@ -1907,6 +1909,30 @@ impl PromPlanner {
...
+            .fail();
+        }
+
+        // Check if label name starts with double underscores (reserved)
+        if label_name.starts_with("__") {
+            return InvalidDestinationLabelNameSnafu {
+                label_name: label_name.to_string(),
+            }
+            .fail();
+        }
```
fengjiachun: Maybe moving this check to the beginning of the method would avoid some unnecessary regex computation.

src/query/src/promql/planner.rs
```diff
@@ -1907,6 +1909,30 @@ impl PromPlanner {
...
+    /// Label names must match the regex: [a-zA-Z_][a-zA-Z0-9_]*
+    /// Additionally, label names starting with double underscores are reserved for internal use.
+    fn validate_label_name(label_name: &str) -> Result<()> {
+        // Check if label name matches the required pattern
+        let label_regex = regex::Regex::new(r"^[a-zA-Z_][a-zA-Z0-9_]*$").unwrap();
+        if !label_regex.is_match(label_name) {
+            return InvalidDestinationLabelNameSnafu {
+                label_name: label_name.to_string(),
+            }
+            .fail();
```
fengjiachun:
```suggestion
            return InvalidDestinationLabelNameSnafu { label_name }.fail();
```

yihong0618:
addressed

fengjiachun (approved):
LGTM

src/query/src/promql/planner.rs
```diff
@@ -1907,6 +1915,22 @@ impl PromPlanner {
         Ok((exprs, new_tags))
     }
 
+    /// Validate label name according to Prometheus specification.
+    /// Label names must match the regex: [a-zA-Z_][a-zA-Z0-9_]*
+    /// Additionally, label names starting with double underscores are reserved for internal use.
+    fn validate_label_name(label_name: &str) -> Result<()> {
```
killme2008, line 1921: Do we need to validate all labels? For example:

https://github.com/GreptimeTeam/greptimedb/blob/e80e4a9ed72827a534448eb5cc0cb7a1b17111a0/src/query/src/promql/planner.rs#L2021
https://github.com/GreptimeTeam/greptimedb/blob/e80e4a9ed72827a534448eb5cc0cb7a1b17111a0/src/query/src/promql/planner.rs#L2007

If this should be a general rule, we could add it to the PromQL parser:
https://github.com/GreptimeTeam/promql-parser

yihong0618: will check

yihong0618: seems no need we have already have the check for others
image

killme2008: Ok. So do we need to check the labels in the `label_join` function? Which adds the new labels too.

yihong0618: will check now.

yihong0618: @killme2008 
label_join do not need to check the name for promql


image

PR #6729, opened 2025-08-13
feat: predicate extractor (region prune part 1)

waynexia:

The first part of pruning unused regions in a distributed query.

It extracts all valid predicates from query plan and converts it to partition expr, for comparing against the table partition exprs to get real effective regions later.

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/predicate_extractor.rs
+++ b/src/query/src/dist_plan/predicate_extractor.rs
@@ -1,0 +1,1199 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Predicate extraction for partition pruning.
+//!
+//! [`PredicateExtractor`] extract a list of [`PartitionExpr`] from given [`LogicalPlan`].
+
+use std::collections::HashSet;
+
+use arrow::datatypes::DataType;
+use common_telemetry::debug;
+use datafusion_common::Result as DfResult;
+use datafusion_expr::{Expr, LogicalPlan, Operator};
+use datatypes::value::Value;
+use partition::expr::{Operand, PartitionExpr, RestrictedOp};
+
+/// Extracts a list of [`PartitionExpr`] from given [`LogicalPlan`]
+pub struct PredicateExtractor;
+
+impl PredicateExtractor {
+    /// Extract partition expressions for partition columns from logical plan  
+    /// This method returns PartitionExpr objects suitable for ConstraintPruner
+    pub fn extract_partition_expressions(
+        plan: &LogicalPlan,
+        partition_columns: &[String],
+    ) -> DfResult<Vec<PartitionExpr>> {
+        // Collect all filter expressions from the logical plan
+        let mut filter_exprs = Vec::new();
+        Self::collect_filter_expressions(plan, &mut filter_exprs)?;
+
+        if filter_exprs.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        // Convert each DataFusion filter expression to PartitionExpr
+        let mut partition_exprs = Vec::with_capacity(filter_exprs.len());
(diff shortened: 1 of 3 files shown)


src/query/src/dist_plan/predicate_extractor.rs
```diff
@@ -0,0 +1,1199 @@
...
+                        }
+                        ExpressionCheckResult::Drop => {
+                            debug!(
+                                "Dropping mixed expression for correctness: {}",
+                                partition_expr
+                            );
+                        }
+                    }
+                }
+                Err(err) => {
```
discord9, line 72: what should happen if a complex filter expression involve partition columns but can't be converted to restricted partition exprs?

waynexia: it will be ignored as this expr doesn't exist

discord9: > it will be ignored as this expr doesn't exist

is it ok to ignore it though? If the filter is something like `a>=30 or black_box(a)` would it be handled correctly?

waynexia: No, for OR it will cause the entire expr to be ignored. and this is ok. If it's AND, only the black box part will be ignored

src/query/src/dist_plan/predicate_extractor.rs
```diff
@@ -0,0 +1,1199 @@
...
+                    if Self::expr_only_involves_partition_columns(expr, partition_columns) {
+                        ExpressionCheckResult::UseAsIs(expr.clone())
+                    } else {
+                        // This shouldn't happen for simple comparisons, but handle defensively
+                        ExpressionCheckResult::Drop
+                    }
+                } else {
+                    ExpressionCheckResult::Drop
+                }
+            }
```
evenyag: Can we simplify this to
```suggestion
                if Self::expr_only_involves_partition_columns(expr, partition_columns) {
                    ExpressionCheckResult::UseAsIs(expr.clone())
                } else {
                    // This shouldn't happen for simple comparisons, but handle defensively
                    ExpressionCheckResult::Drop
                }
            }
```

src/query/src/dist_plan/predicate_extractor.rs
```diff
@@ -0,0 +1,1199 @@
...
+        match expr {
+            Expr::Column(col) => {
+                // Handle qualified column names (table.column) by extracting just the column name
+                // For partition pruning, we typically only care about the column name itself
+                let column_name = if let Some(relation) = &col.relation {
+                    debug!(
+                        "Using qualified column reference: {}.{}",
+                        relation, col.name
+                    );
+                    col.name.clone()
```
evenyag, line 395: Is it possible that the relation belongs to another table? For example, joining two tables.

src/query/src/dist_plan/predicate_extractor.rs
```diff
@@ -0,0 +1,1199 @@
...
+    ) {
+        if let RestrictedOp::And = expr.op() {
+            // Recursively process both sides of AND
+            Self::extract_constraints_from_operand(expr.lhs(), partition_columns, result);
+            Self::extract_constraints_from_operand(expr.rhs(), partition_columns, result);
+        } else {
+            // Non-AND expression: check if it involves partition columns
+            if Self::expr_involves_partition_columns(expr, partition_columns) {
+                result.push(expr.clone());
+            }
```
evenyag: Do we need to check if it only involves partition columns? What if this is an OR expr that involves both partition and non-partition columns?

waynexia: it will be dropped if the other side of OR is not partition column, see the second case "mixed_or_expression" of `test_mixed_partition_non_partition_expressions`

src/query/src/dist_plan/predicate_extractor.rs
waynexia: Add a checker to fallback to no pruning if there are multiple relations. Will handle this case in the follow up PR

src/query/src/dist_plan/predicate_extractor.rs
evenyag: I see, the rule became stricter in https://github.com/GreptimeTeam/greptimedb/pull/6729/commits/a97476c33cc588e7b1b871c8c146760cf74497b5

`mixed_or_expression` wraps the `OR` expr after `BETWEEN` so it might not call `extract_and_constraints()` because the root expr is `OR`.

Do we need a case that the root expr is `AND`? Although https://github.com/GreptimeTeam/greptimedb/pull/6729/commits/a97476c33cc588e7b1b871c8c146760cf74497b5 should address this.
```
(partition col < 10).or(no partition col < 5).and(partition col > 5)
```

waynexia: good case, will include in next patch

discord9 (approved):

PR #6731, opened 2025-08-13
chore: add u64 for EqualValue and set expr is true when filter is empty

paomian:

* add u64 for EqualValue and set expr is true when filter is empty

## PR Checklist


src/log-query/src/log_query.rs
```diff
@@ -336,6 +336,8 @@ pub enum EqualValue {
     Boolean(bool),
     /// Exact match with a number value.
     Int(i64),
+    /// Exact match with an unsigned integer value.
+    Uint(u64),
```
evenyag:
```suggestion
    UInt(u64),
```

evenyag (approved):

PR #6742, opened 2025-08-14
fix: truncate manifest action compat

discord9:

##6602

as title, previous pr failed to notice this incompatibility with region truncate action in manifest files, so if you truncate a table in old version, then immediately upgrade to new version, db wouldn't understand the manifest action files(luckily it didn't break manifest checkpoint files), ~so this fix make sure region truncate deser now first try v2(latest version) of itself, then try v1(old version) and convert it to v2~, added `#[serde(flatten)]` and `#[serde(untagged)]` instead

## PR Checklist

### DIFF
--- a/src/mito2/src/manifest/action.rs
+++ b/src/mito2/src/manifest/action.rs
@@ -62,11 +62,55 @@ pub struct RegionRemove {
     pub region_id: RegionId,
 }
 
-/// Last data truncated in the region.
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+#[derive(Serialize, Clone, Debug, PartialEq, Eq)]
 pub struct RegionTruncate {
     pub region_id: RegionId,
     pub kind: TruncateKind,
+}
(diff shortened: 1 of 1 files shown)


WenyXu (approved):

src/mito2/src/manifest/action.rs
```diff
@@ -62,13 +62,57 @@ pub struct RegionRemove {
     pub region_id: RegionId,
 }
 
-/// Last data truncated in the region.
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+#[derive(Serialize, Clone, Debug, PartialEq, Eq)]
 pub struct RegionTruncate {
     pub region_id: RegionId,
     pub kind: TruncateKind,
 }
```
evenyag: Can internal tagging solve this? We use the old kind as the default type.
https://serde.rs/enum-representations.html#internally-tagged

evenyag (approved):

PR #6750, opened 2025-08-15
feat: flownode grpc client to frontend tls option

discord9:

flow's grpc client to frontend allow tls option

## PR Checklist


sunng87 (approved):

src/common/grpc/src/channel_manager.rs
```diff
@@ -97,6 +98,7 @@ impl ChannelManager {
         }
     }
 
+    /// Read tls cert and key files and create a ChannelManager with TLS config.
     pub fn with_tls_config(config: ChannelConfig) -> Result<Self> {
```
fengys1996, line 102: nit: It looks like `with_tls_config` and `with_config` could be merged into a single method. And the semantics here are a bit confusing, we can refactor it later.

sunng87: That's why there was no enabled in tlsconfig, the API is designed for calling this function to create tls client, instead of following a config toggle.

PR #6751, opened 2025-08-15
feat: add event ts to region manifest

discord9:

add event timestamp to both region manifest file and manifest action files, so that later gc worker can tell which removed(from manifest) files should be deleted now
- add event ts to region manifest, time being when the manifest is updated
- add event ts to `RegionEdit` and `RegionTruncate` so the time of removing files is noted, which is also vital for gc to collect files that should be deleted

## PR Checklist


waynexia:
>add event ts to region manifest, time being when the manifest is updated

Is this naturally available as part of file metadata? As we won't actually "update" a manifest file, the time of creation is the timestamp of each action

src/mito2/src/sst/location.rs
```diff
@@ -54,6 +56,33 @@ pub fn index_file_path(
     util::join_path(&index_dir, &format!("{}.puffin", region_file_id.file_id()))
 }
 
+/// Get RegionFileId from sst or index filename
+pub fn parse_file_id_from_path(filepath: &str) -> crate::error::Result<FileId> {
```
evenyag: This file is unused in this PR.

src/mito2/src/manifest/action.rs
```diff
@@ -484,4 +595,186 @@ mod tests {
             }
         );
     }
+    /// Test if old version can still be deserialized then serialized to the new version.
+    #[test]
+    fn test_old_region_manifest_compat() {
```
evenyag, line 691: Also add a test for the new format.

src/mito2/src/manifest/action.rs
```diff
@@ -196,8 +260,55 @@ impl RegionManifestBuilder {
...
 }
 
+/// A record of removed files in the region manifest.
+/// This is used to keep track of files that have been removed from the manifest but may still
+/// be kept for a while
+#[derive(Serialize, Deserialize, Clone, Debug, Default, PartialEq, Eq)]
+pub struct RemovedFilesRecord {
+    /// a list of `(FileIds, timestamp)` pairs, where the timestamp is the time when
+    /// the files are removed from manifest. The timestamp is in milliseconds since unix epoch.
+    pub removed_files: Vec<(HashSet<FileId>, i64)>,
```
evenyag: Can we avoid using tuples here? It makes the deserialized JSON unclear. Use a struct to replace the tuple `(HashSet<FileId>, i64)`.
The RemovedFilesRecord itself could be replaced by a Vec.

src/mito2/src/manifest/action.rs
```diff
@@ -196,8 +260,55 @@ impl RegionManifestBuilder {
...
+        if total_removed_files <= opt.keep_count {
+            return Ok(());
+        }
+
+        let can_evict_until = chrono::Utc::now()
+            - chrono::Duration::from_std(opt.keep_ttl).context(DurationOutOfRangeSnafu {
+                input: opt.keep_ttl,
+            })?;
+
+        self.removed_files.sort_by_key(|(_, ts)| *ts);
```
evenyag: sort unstable by key?

src/mito2/src/manifest/action.rs
```diff
@@ -196,8 +260,55 @@ impl RegionManifestBuilder {
...
+            .filter_map(|(files, ts)| {
+                if ts < can_evict_until.timestamp_millis() {
+                    // can evict all files
+                    // TODO(discord9): maybe only evict to below keep_count? Maybe not, or the update might be too frequent.
+                    None
+                } else {
+                    Some((files, ts))
+                }
+            })
+            .collect();
```
evenyag: Maybe you can use retain without sorting the vec. Also consider truncating the vec by keep_count after retain.

src/mito2/src/config.rs
```diff
@@ -65,6 +65,15 @@ pub struct MitoConfig {
     /// Number of meta action updated to trigger a new checkpoint
     /// for the manifest (default 10).
     pub manifest_checkpoint_distance: u64,
+    /// Number of removed files to keep in manifest's `removed_files` field before also
+    /// remove them from `removed_files`. Mostly for debugging purpose.
+    pub experimental_manifest_keep_removed_file_count: usize,
```
evenyag, line 72: Do we need experimental prefix?

src/mito2/src/manifest/action.rs
```diff
@@ -196,8 +260,64 @@ impl RegionManifestBuilder {
...
+    /// a list of `(FileIds, timestamp)` pairs, where the timestamp is the time when
+    /// the files are removed from manifest. The timestamp is in milliseconds since unix epoch.
+    pub removed_files: Vec<RemovedFiles>,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, Default, PartialEq, Eq)]
+pub struct RemovedFiles {
+    /// The timestamp is the time when
+    /// the files are removed from manifest. The timestamp is in milliseconds since unix epoch.
+    pub at: i64,
```
evenyag:
```suggestion
    pub removed_at: i64,
```

discord9:
> > add event ts to region manifest, time being when the manifest is updated
> 
> Is this naturally available as part of file metadata? As we won't actually "update" a manifest file, the time of creation is the timestamp of each action

yes the problem is that "action" files have a limited number and need to keep track of timestamp in checkpoint file

src/mito2/src/config.rs
discord9: > Do we need experimental prefix?

yes? I mean this option could very possible change later when need to change how the manifest's `removed_files` field is updated?

src/mito2/src/manifest/action.rs
```diff
@@ -196,8 +260,72 @@ impl RegionManifestBuilder {
...
+                    // can evict all files
+                    // TODO(discord9): maybe only evict to below keep_count? Maybe not, or the update might be too frequent.
+                    cur_file_cnt -= f.file_ids.len();
+                    None
+                } else {
+                    Some(f)
+                }
+            })
+            .collect();
+        self.removed_files = updated;
```
evenyag, line 320: I think it would be better to hard limit the file num under keep_count otherwise this option is meaningless. We can take the last keep_count elements in `updated`.

discord9: > I think it would be better to hard limit the file num under keep_count otherwise this option is meaningless. We can take the last keep_count elements in `updated`.

hard limiting the number instead of time would have the problem of not tracking some files' delete time(if large amount of files are removed) hence causing larger amount of untracked removed files causing gc worker to delay their deletion and cause possible "leak"?

evenyag: Will it be a risk if we keeping too much file ids in the memory?

discord9: > Will it be a risk if we keeping too much file ids in the memory?

could be, but the risk should be fine, since those file ids originally come from removed files in manifest and should only be about similiar size(or at most 10x or 100x times large in extreme scene) so should be ok?

config/config.md
```diff
@@ -509,6 +509,8 @@
 | `region_engine.mito.worker_channel_size` | Integer | `128` | Request channel size of each worker. |
 | `region_engine.mito.worker_request_batch_size` | Integer | `64` | Max batch size for a worker to handle requests. |
 | `region_engine.mito.manifest_checkpoint_distance` | Integer | `10` | Number of meta action updated to trigger a new checkpoint for the manifest. |
+| `region_engine.mito.experimental_manifest_keep_removed_file_count` | Integer | `256` | Number of removed files to keep in manifest's `removed_files` field before also<br/>remove them from `removed_files`. Mostly for debugging purpose. |
```
waynexia: do we need to handle 0 value individually?

discord9: > do we need to handle 0 value individually?

changed that if keep_count is zero only use keep_ttl

PR #6752, opened 2025-08-15
feat: region prune part 2

waynexia:

follow #6729

Implement the region pruner itself

## PR Checklist

Signed-off-by: Ruihang Xia <waynestxia@gmail.com>Signed-off-by: Ruihang Xia <waynestxia@gmail.com>

### DIFF
--- a/tests/cases/standalone/common/partition.result
+++ b/tests/cases/standalone/common/partition.result
@@ -238,17 +238,23 @@ EXPLAIN ANALYZE SELECT * FROM valid_rule;
 +-+-+-+
 | stage | node | plan_|
 +-+-+-+
-| 0_| 0_|_MergeScanExec: REDACTED
+| 0_| 0_|_CooperativeExec REDACTED
+|_|_|_MergeScanExec: REDACTED
 |_|_|_|
-| 1_| 0_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
+| 1_| 0_|_CooperativeExec REDACTED
+|_|_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
 |_|_|_|
-| 1_| 1_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
+| 1_| 1_|_CooperativeExec REDACTED
+|_|_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
 |_|_|_|
-| 1_| 2_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
+| 1_| 2_|_CooperativeExec REDACTED
+|_|_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
 |_|_|_|
-| 1_| 3_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
+| 1_| 3_|_CooperativeExec REDACTED
+|_|_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
 |_|_|_|
-| 1_| 4_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
+| 1_| 4_|_CooperativeExec REDACTED
+|_|_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
 |_|_|_|
 |_|_| Total rows: 0_|
 +-+-+-+
@@ -266,9 +272,11 @@ WHERE a < 10;
 +-+-+-+
 | stage | node | plan_|
 +-+-+-+
-| 0_| 0_|_MergeScanExec: REDACTED
+| 0_| 0_|_CooperativeExec REDACTED
+|_|_|_MergeScanExec: REDACTED
 |_|_|_|
-| 1_| 0_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
+| 1_| 0_|_CooperativeExec REDACTED
+|_|_|_SeqScan: region=REDACTED, "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} REDACTED
 |_|_|_|
 |_|_| Total rows: 0_|
 +-+-+-+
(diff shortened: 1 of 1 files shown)


src/query/src/dist_plan/planner.rs
```diff
@@ -198,7 +209,99 @@ impl DistExtensionPlanner {
...
+        let partition_expressions = match PredicateExtractor::extract_partition_expressions(
+            logical_plan,
+            &partition_columns,
+        ) {
+            Ok(expressions) => expressions,
+            Err(err) => {
+                common_telemetry::debug!(
+                    "Failed to extract partition expressions, using all regions: {}",
+                    err
+                );
```
evenyag: Should we log the table id and table name?

```suggestion
                common_telemetry::debug!(
                    "Failed to extract partition expressions, using all regions: {:?}",
                    err
                );
```

src/query/src/dist_plan/planner.rs
```diff
@@ -198,7 +209,99 @@ impl DistExtensionPlanner {
...
+            .find_table_partitions(table.table_info().table_id())
+            .await
+        {
+            Ok(partitions) => partitions,
+            Err(err) => {
+                common_telemetry::debug!(
+                    "Failed to get partition information for table {}: {}, using all regions",
+                    table_name,
+                    err
+                );
```
evenyag: Do we need to print the err in debug format?

src/query/src/dist_plan/planner.rs
```diff
@@ -198,7 +209,99 @@ impl DistExtensionPlanner {
...
+            &partitions,
+            partition_column_types,
+        ) {
+            Ok(regions) => regions,
+            Err(err) => {
+                common_telemetry::debug!(
+                    "Failed to prune regions for table {}: {}, using all regions",
+                    table_name,
+                    err
+                );
```
evenyag: Do we need to log err using `{:?}`?

src/query/src/dist_plan/region_pruner.rs
```diff
@@ -0,0 +1,757 @@
...
+                    region_atomics.source_expr_index - query_expressions.len();
+                candidate_regions.insert(expression_to_partition[&partition_expr_index]);
+            }
+        }
+
+        debug!(
+            "Constraint pruning: {} -> {} regions",
+            partitions.len(),
+            candidate_regions.len()
+        );
```
evenyag: Also logs the prune cost?

src/query/src/dist_plan/region_pruner.rs
```diff
@@ -0,0 +1,757 @@
...
+
+        // Collect all partition expressions for unified normalization
+        let mut expression_to_partition = HashMap::default();
+        let mut all_partition_expressions = Vec::with_capacity(partitions.len());
+        for partition in partitions {
+            if let Some(expr) = &partition.partition_expr {
+                expression_to_partition.insert(all_partition_expressions.len(), partition.id);
+                all_partition_expressions.push(expr.clone());
+            }
+        }
```
evenyag: Looks like we can use a Vec for expression_to_partition since the key in the map is the index of all_partition_expressions.

src/query/src/dist_plan/region_pruner.rs
```diff
@@ -0,0 +1,757 @@
...
+
+        range
+    }
+}
+
+/// Represents a value range derived from a group of nucleons for the same column
+#[derive(Debug, Clone)]
+struct ValueRange {
+    min_value: Option<(OrderedF64, bool)>, // (value, inclusive)
+    max_value: Option<(OrderedF64, bool)>, // (value, inclusive)
```
evenyag: non-block: Maybe we can use `Bound`.
https://doc.rust-lang.org/stable/std/ops/enum.Bound.html

src/query/src/dist_plan/region_pruner.rs
```diff
@@ -0,0 +1,757 @@
...
+                    range.min_value = Some((value, true));
+                    range.max_value = Some((value, true));
+                    break; // Equality is the most restrictive, no need to check others
+                }
+                Lt => {
+                    // Less than: update max_value if it's more restrictive
+                    let new_max = (value, false);
+                    if range.max_value.is_none() || range.max_value.as_ref().unwrap().0 > value {
+                        range.max_value = Some(new_max);
+                    }
```
evenyag: non-blocking: if range.max_value is inclusive, we can also update max value. But the benefit is very little.

src/query/src/dist_plan/region_pruner.rs
```diff
@@ -0,0 +1,757 @@
...
+                    // Greater than or equal: update min_value if it's more restrictive
+                    let new_min = (value, true);
+                    if range.min_value.is_none() || range.min_value.as_ref().unwrap().0 < value {
+                        range.min_value = Some(new_min);
+                    }
+                }
+                NotEq => {
+                    // NotEq should be broke into two other AtomicExprs
+                    continue;
+                }
```
evenyag: Did Collider break NotEq? Now it only breaks `OR`.

waynexia: No, but it won't affect the correctness. I'm thinking of breaking it in some other places (not decided yet)

src/query/src/dist_plan/planner.rs
```diff
@@ -198,7 +209,101 @@ impl DistExtensionPlanner {
...
+            &partition_columns,
+        ) {
+            Ok(expressions) => expressions,
+            Err(err) => {
+                common_telemetry::debug!(
+                    "Failed to extract partition expressions for table {} (id: {}), using all regions: {}",
+                    table_name,
+                    table.table_info().table_id(),
+                    err
+                );
```
evenyag:
```suggestion
                common_telemetry::debug!(
                    "Failed to extract partition expressions for table {} (id: {}), using all regions: {:?}",
                    table_name,
                    table.table_info().table_id(),
                    err
                );
```

src/query/src/dist_plan/planner.rs
```diff
@@ -198,7 +209,101 @@ impl DistExtensionPlanner {
...
+            .find_table_partitions(table.table_info().table_id())
+            .await
+        {
+            Ok(partitions) => partitions,
+            Err(err) => {
+                common_telemetry::debug!(
+                    "Failed to get partition information for table {}: {:?}, using all regions",
+                    table_name,
+                    err
+                );
```
evenyag:
```suggestion
                common_telemetry::debug!(
                    "Failed to get partition information for table {}, using all regions: {:?}",
                    table_name,
                    err
                );
```

src/query/src/dist_plan/planner.rs
```diff
@@ -198,7 +209,101 @@ impl DistExtensionPlanner {
...
+            &partitions,
+            partition_column_types,
+        ) {
+            Ok(regions) => regions,
+            Err(err) => {
+                common_telemetry::debug!(
+                    "Failed to prune regions for table {}: {:?}, using all regions",
+                    table_name,
+                    err
+                );
```
evenyag:
```suggestion
                common_telemetry::debug!(
                    "Failed to prune regions for table {}, using all regions: {:?}",
                    table_name,
                    err
                );
```

src/query/src/dist_plan/region_pruner.rs
```diff
@@ -0,0 +1,754 @@
...
+            | (Bound::Included(cur), Bound::Excluded(new))
+            | (Bound::Excluded(cur), Bound::Excluded(new)) => {
+                if new < cur {
+                    self.upper = new_upper;
+                } else if new == cur {
+                    // prefer Excluded over Included for the same value (more restrictive)
+                    self.upper = new_upper;
+                    if matches!(new_upper, Bound::Excluded(_))
+                        && matches!(self.upper, Bound::Included(_))
+                    {}
```
evenyag: Looks like we should update upper inside the if block?
```suggestion
                    // prefer Excluded over Included for the same value (more restrictive)
                    if matches!(new_upper, Bound::Excluded(_))
                        && matches!(self.upper, Bound::Included(_))
                    {
                        self.upper = new_upper;
                    }
```

evenyag (approved):

discord9:
maybe add some sqlness testcases(that doesn't redact region ids in MergeScanExec) would be beneficial?

waynexia:
> maybe add some sqlness testcases(that doesn't redact region ids in MergeScanExec) would be beneficial?

https://github.com/GreptimeTeam/greptimedb/pull/6752/commits/21fdc72efc11777400c651d5ff62409d72e2ac1b

PR #6763, opened 2025-08-18
refactor: change plugin option type from `&[PluginOptions]` to `Option<&PluginOptions>` for understandability

zyy17:

The original type `&[PluginOptions]` is not accurate for describing the actual plugin configurations. In most scenarios, there is only one element for `[PluginOptions]`. Moreover, it's not easy to use environment variables to set up plugins.

After using `Option<&PluginOptions>`, the configs will look like:

```
...
[plugins]
[plugins.foo]
name = "foo"
[plugins.bar]
name = "bar"
...
```

For one plugin, it can design its own configs. Therefore, the `PluginOptions` will look like:

```rust
#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
pub struct PluginOptions {
    pub PluginFoo: Option<PluginFooOptions>,
    pub PluginBar: Option<PluginBarOptions>,
}
```

It will be simpler to write the setup code, for example:

```rust
pub async fn setup_datanode_plugins(
    plugins: &mut Plugins,
    plugin_options: Option<&PluginOptions>,
    dn_opts: &DatanodeOptions,
) -> Result<()> {
    if let Some(PluginFoo) = plugin_options.PluginFoo {
        // initialization of PluginFoo.
    }
}
```

## PR Checklist

### DIFF
--- a/src/cmd/src/datanode/builder.rs
+++ b/src/cmd/src/datanode/builder.rs
@@ -71,7 +71,7 @@ impl InstanceBuilder {
         maybe_activate_heap_profile(&dn_opts.memory);
         create_resource_limit_metrics(APP_NAME);
 
-        plugins::setup_datanode_plugins(plugins, &opts.plugins, dn_opts)
+        plugins::setup_datanode_plugins(plugins, opts.plugins.as_ref(), dn_opts)
             .await
             .context(StartDatanodeSnafu)?;
 
--- a/src/plugins/src/datanode.rs
+++ b/src/plugins/src/datanode.rs
@@ -22,7 +22,7 @@ use crate::options::PluginOptions;
 #[allow(unused_mut)]
 pub async fn setup_datanode_plugins(
     plugins: &mut Plugins,
-    plugin_options: &[PluginOptions],
+    plugin_options: Option<&PluginOptions>,
     dn_opts: &DatanodeOptions,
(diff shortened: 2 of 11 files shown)


src/cmd/src/datanode/builder.rs
```diff
@@ -71,7 +71,7 @@ impl InstanceBuilder {
         maybe_activate_heap_profile(&dn_opts.memory);
         create_resource_limit_metrics(APP_NAME);
 
-        plugins::setup_datanode_plugins(plugins, &opts.plugins, dn_opts)
+        plugins::setup_datanode_plugins(plugins, opts.plugins.clone(), dn_opts)
```
sunng87: can we use borrow here?

src/plugins/src/datanode.rs
```diff
@@ -22,7 +22,7 @@ use crate::options::PluginOptions;
 #[allow(unused_mut)]
 pub async fn setup_datanode_plugins(
     plugins: &mut Plugins,
-    plugin_options: &[PluginOptions],
+    plugin_options: Option<PluginOptions>,
```
sunng87: can we use `Option<&PluginOptions>` for this ?

paomian:
Do we need to mark this pr as breaking?

Issue #6721, opened 2025-08-12
Introduce internal metrics recording


WenyXu:
### What problem does the new feature solve?

Currently, there is no unified and efficient way to collect and track key metrics inside the cluster, such as region-level write rate and throughput, as well as topic-level write throughput and request count. 

This feature implements internal metrics recording that periodically collects and aggregates key operational data. The collected metrics improve cluster observability and provide reliable information for metasrv decisions.

### What does the feature do?

This feature introduces a unified internal metrics recording system to periodically collect cluster metrics, such as region write rate and topic write throughput.
- At the datanode, metrics are locally accumulated over a fixed interval (e.g., 30 seconds).
- These metrics are then reported to Metasrv through heartbeat messages.
- Metasrv aggregates and forwards the metrics to `Frontend` via write RPC calls.
- All internal metrics are stored in a private internal table (under `greptime_private` schema).

```mermaid
graph LR
    DN[Datanode]
    MS[Metasrv]
    FE[Frontend]

    DN -->|Heartbeat with metrics| MS
    MS -->|Write RPC with metrics| FE
```

This system provides lightweight and accurate data to support cluster decisions. It is designed to be distinct from external monitoring endpoints (e.g., `/metrics` HTTP API), focusing on internal decision-making rather than external observability.

PR #6768, opened 2025-08-19
feat: derive macro `ToRow`

WenyXu:

#6721

This PR introduces the `ToRow` derive macro, which enables structs to be automatically converted into `greptime_proto::v1::Row` and the corresponding `ColumnSchema`. This reduces boilerplate code and prevents potential field/type mismatches.

### Features
- New derive macro: `#[derive(ToRow)]` (exported from common_macro)
- Generated methods:
  - `to_row(&self) -> greptime_proto::v1::Row`
  - `schema(&self) -> Vec<greptime_proto::v1::ColumnSchema>`
- Field attribute `#[to_row(...)]`:
  - `name`: override column name
  - `datatype`: set column data type (case-insensitive)
  - `semantic`: set semantic type (field | tag | timestamp, case-insensitive)
- Type handling:
  - Auto-infers primitives: i8/i16/i32/i64/u8/u16/u32/u64/f32/f64/bool
  - Other types (e.g., String, time/date/complex) require explicit type
  - Supports Option<T>; `None` maps to `Value { value_data: None }`

### Usage
```rust
#[derive(ToRow)]
struct LogRecord {
    id: i32,
    #[col(name = "lvl", datatype = "string", semantic = "tag")]
    level: String,
    #[col(name = "greptime_timestamp", datatype = "TimestampMillisecond", semantic = "timestamp")]
    ts: i64,
}

let rec = LogRecord { id: 1, level: "INFO".into(), ts: 1718563200000 };
let row = rec.to_row();
let schema = rec.schema();
```

## PR Checklist


sunng87:
We can put this into rust ingester too

src/common/macro/src/to_row.rs
```diff
@@ -0,0 +1,494 @@
...
+}
+
+fn get_column_data_type(
+    infer_column_data_type: &Option<ColumnDataType>,
+    attribute: &ColumnAttribute,
+) -> Option<ColumnDataType> {
+    if let Some(column_data_type) = attribute.column_data_type {
+        return Some(column_data_type);
+    }
+    *infer_column_data_type
```
MichaelScofield:
```suggestion
    attribute.column_data_type.or(*infer_column_data_type)
```

src/common/macro/src/to_row.rs
```diff
@@ -0,0 +1,494 @@
...
+}
+
+fn convert_time_type_to_column_data_type(ident: &str) -> Option<ColumnDataType> {
+    match ident {
+        "timemillisecond" => Some(ColumnDataType::TimeMillisecond),
+        "timemicrosecond" => Some(ColumnDataType::TimeMicrosecond),
+        "timenanosecond" => Some(ColumnDataType::TimeNanosecond),
+        _ => None,
+    }
+}
```
evenyag: Don't need to check timesecond type?

src/common/macro/src/to_row.rs
```diff
@@ -0,0 +1,494 @@
...
+    if let Some(column_data_type) = attribute.column_data_type {
+        return Some(column_data_type);
+    }
+    *infer_column_data_type
+}
+
+fn convert_column_data_type_to_value_data_ident(column_data_type: &ColumnDataType) -> Ident {
+    let suffix = "Value";
+    match column_data_type {
+        ColumnDataType::Boolean => format_ident!("{}{}", "Bool", suffix),
```
MichaelScofield: I see no difference to simply write `ColumnDataType::Boolean => "BoolValue"`, and others below. What's the necessity to make `"Value"` a variable and `format_ident!` two string literals? Is it generated by AI?

src/common/macro/src/to_row.rs
WenyXu: Good catch!

WenyXu: Resolved in [cfa1cc4](https://github.com/GreptimeTeam/greptimedb/pull/6768/commits/cfa1cc471d6188dd7a097e0c9b4bcd53602b10d4)

evenyag (approved):

src/common/macro/src/to_row.rs
WenyXu: As considered, the suffix may change in the future, so I introduced a `suffix` here.

WenyXu: resolved in [3ec1441](https://github.com/GreptimeTeam/greptimedb/pull/6768/commits/3ec144153f9670ce833146c32db5cfeb537d81de)

src/common/macro/src/to_row.rs
WenyXu: resolved in [3ec1441](https://github.com/GreptimeTeam/greptimedb/pull/6768/commits/3ec144153f9670ce833146c32db5cfeb537d81de)

src/common/macro/src/to_row.rs
```diff
@@ -0,0 +1,491 @@
...
+        ));
+    };
+
+    let method_name = format_ident!("{}_to_value", ident);
+    let value_data = convert_column_data_type_to_value_data_ident(&column_data_type);
+    let body = if is_optional {
+        quote! {
+            match &self.#ident {
+                Some(v) => greptime_proto::v1::Value {
+                    value_data: Some(greptime_proto::v1::value::ValueData::#value_data(v.clone().into())),
```
fengjiachun:
```suggestion
                    value_data: Some(greptime_proto::v1::value::ValueData::#value_data(v.into())),
```

src/common/macro/src/to_row.rs
```diff
@@ -0,0 +1,491 @@
...
+                },
+                None => greptime_proto::v1::Value { value_data: None },
+            }
+        }
+    } else {
+        quote! {
+            greptime_proto::v1::Value {
+                value_data: Some(greptime_proto::v1::value::ValueData::#value_data(self.#ident.clone().into())),
```
fengjiachun:
```suggestion
                value_data: Some(greptime_proto::v1::value::ValueData::#value_data(self.#ident.into())),
```

src/common/macro/src/to_row.rs
fengjiachun: I also think this is an unnecessary string concatenation.

src/common/macro/src/to_row.rs
```diff
@@ -0,0 +1,491 @@
...
+        "i16" => Some(ColumnDataType::Int16),
+        "i32" => Some(ColumnDataType::Int32),
+        "i64" => Some(ColumnDataType::Int64),
+        "u8" => Some(ColumnDataType::Uint8),
+        "u16" => Some(ColumnDataType::Uint16),
+        "u32" => Some(ColumnDataType::Uint32),
+        "u64" => Some(ColumnDataType::Uint64),
+        "f32" => Some(ColumnDataType::Float32),
+        "f64" => Some(ColumnDataType::Float64),
+        "bool" => Some(ColumnDataType::Boolean),
```
fengjiachun: Can we use [phf_map](https://docs.rs/phf/latest/phf/macro.phf_map.html) to avoid this inefficient string one by one matching?

src/common/macro/src/to_row.rs
WenyXu: The clone might be unavoidable in this case, which is why we provide support for reference types.

src/common/macro/src/to_row.rs
WenyXu: Resolved in [e2d894d](https://github.com/GreptimeTeam/greptimedb/pull/6768/commits/e2d894d412fa521ad853e6deb6aaa8e28860275c)

PR #6771, opened 2025-08-19
feat: Implements FlatCompatBatch to adapt schema in flat format

evenyag:

Change CompatBatch to an enum
- Renames CompatBatch to PrimaryKeyCompatBatch
- Implements a new adapter FlatCompatBatch for flat format

FlatCompatBatch handles missing columns, different types, and different primary key encodings.

## PR Checklist


src/mito2/src/read/compat.rs
```diff
@@ -450,11 +665,251 @@ impl RewritePrimaryKey {
...
+                CompositeValues::Sparse(sparse_values) => {
+                    let column_id_values = self
+                        .metadata
+                        .primary_key
+                        .iter()
+                        .map(|id| {
+                            let value = sparse_values.get_or_null(*id);
+                            (*id, value.as_value_ref())
+                        })
+                        .collect::<Vec<_>>();
```
fengjiachun: The `column_id_values` is only borrowed, but a new Vec is allocated for each data row processed. Can we pre-allocate a buffer outside the loop and reuse it within the loop?

evenyag: If we want to reuse the vec, then we have to clone the value here. I changed to use clone now.

src/mito2/src/read/compat.rs
```diff
@@ -450,11 +665,247 @@ impl RewritePrimaryKey {
...
+            .downcast_ref::<BinaryArray>()
+            .unwrap();
+        let mut builder = BinaryBuilder::with_capacity(
+            old_pk_values_array.len(),
+            old_pk_values_array.value_data().len(),
+        );
+
+        // Binary buffer for the primary key.
+        let mut buffer =
+            Vec::with_capacity(old_pk_values_array.value_data().len() / old_pk_values_array.len());
```
WenyXu: `old_pk_values_array.len()` can't be zero

Issue #6749, opened 2025-08-15
Next things todo after DataFusion is upgraded


MichaelScofield:
### What type of enhancement is this?

Refactor

### What does the enhancement do?

to name a few:

- [ ] Move ORC file source implementation from datafusion-orc to greptimedb (otherwise we have to modify its datafusion dependency version everytime we upgrade datafusion)
- [ ] Use async UDF to improve "admin" function's implementation
- [ ] Implement greptimedb's "range" syntax in sqlparser. Best if it can be done in a `GreptimedbDialect`. The main block is **where** to rewrite the whole "select" to the "`range_fn`" in sqlparser
- [ ] Directly use Arrow Flight's official encoder/decoder. The implementation in greptimedb is for historical reasons (mainly to support "insert into" in `do_get`). But I think it's time to move on, given the official encoder/decoder has more important features like splitting large record batches and supporting more complex Arrow datatypes like dictionary and list
- [ ] Implement “Utf8View" datatype in Arrow
- [ ] Rewrite greptimedb's implementation of UDAF to adapt more of datafusion
- [ ] The entrypoint of datafusion query engine is kinda messy, time to refactor it
- [ ] upgrade opendal to v0.54 and beyond

killme2008:
> Now the ScalarFunctionArgs contains ConfigOptions, we may use it to pass time zone and rewrite our time-related functions 

After conducting research, I found that this refactoring is unnecessary. The extensions in ConfigOptions are typically simple primitive types by default, while our FunctionContext includes various handlers and complete query context information, which cannot be covered by the extension mechanism. The value of refactoring solely for timezone support is limited.

https://github.com/GreptimeTeam/greptimedb/blob/7e573e497c5c4675e3eee69ce5be668bdf22fc0d/src/common/function/src/function.rs#L28-L32

killme2008:
FYI.

But we should set the config options `datafusion.execution.time_zone` and `datafusion.format.timestamp_tz_format` for datafusion's internal UDFs. I'll implement them.

PR #6781, opened 2025-08-21
refactor: query config options

killme2008:

#6749 

Refactor query configuration options by introducing a centralized method to create DataFusion ConfigOptions from the QueryContext, ensuring consistent timezone configuration across the system.

## PR Checklist

### DIFF
--- a/src/common/query/src/columnar_value.rs
+++ b/src/common/query/src/columnar_value.rs
@@ -17,7 +17,7 @@ use datatypes::prelude::ConcreteDataType;
 use datatypes::vectors::{Helper, VectorRef};
 use snafu::ResultExt;
 
-use crate::error::{self, FromScalarValueSnafu, IntoVectorSnafu, Result};
+use crate::error::{self, GeneralDataFusionSnafu, IntoVectorSnafu, Result};
 use crate::prelude::ScalarValue;
 
 /// Represents the result from an expression
@@ -40,12 +40,16 @@ impl ColumnarValue {
 
     /// Convert a columnar value into an VectorRef
     pub fn try_into_vector(self, num_rows: usize) -> Result<VectorRef> {
-        match self {
-            ColumnarValue::Vector(v) => Ok(v),
+        Ok(match self {
+            ColumnarValue::Vector(v) => v,
             ColumnarValue::Scalar(s) => {
-                Helper::try_from_scalar_value(s, num_rows).context(FromScalarValueSnafu)
+                let v = s
+                    .to_array_of_size(num_rows)
+                    .context(GeneralDataFusionSnafu)?;
+                let data_type = v.data_type().clone();
+                Helper::try_into_vector(v).context(IntoVectorSnafu { data_type })?
             }
-        }
+        })
     }
 }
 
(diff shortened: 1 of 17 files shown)


src/common/query/src/columnar_value.rs
```diff
@@ -40,16 +40,12 @@ impl ColumnarValue {
...
-            ColumnarValue::Vector(v) => v,
+        match self {
+            ColumnarValue::Vector(v) => Ok(v),
             ColumnarValue::Scalar(s) => {
-                let v = s
-                    .to_array_of_size(num_rows)
-                    .context(GeneralDataFusionSnafu)?;
-                let data_type = v.data_type().clone();
-                Helper::try_into_vector(v).context(IntoVectorSnafu { data_type })?
+                Helper::try_from_scalar_value(s, num_rows).context(FromScalarValueSnafu)
```
evenyag: We can replace the implementation of try_from_scalar_value() with `to_array_of_size() + try_into_vector()` as it should support more types.

killme2008: I don't think so. `try_from_scalar_value` use constant vector, which is better than `to_array_of_size() + try_into_vector()`.

killme2008: Perhaps we can fall back to `to_array_of_size()` followed by `try_into_vector()` when the type isn’t currently supported.

killme2008: Looks like the unsupported list in these two methods are almost the same

https://github.com/GreptimeTeam/greptimedb/blob/main/src/datatypes/src/vectors/helper.rs#L237-L247

https://github.com/GreptimeTeam/greptimedb/blob/main/src/datatypes/src/vectors/helper.rs#L374-L386

I don't think we should change this behaviour.

evenyag: However, this PR has changed the behavior of this method. After this PR, this method may return a ConstantVector instead of a Vector with N values. It may affect all places using this method.

We should keep the behavior unchanged or document this if it returns constant vectors.

killme2008: `ColumnarValue::try_into_vector` is only used in UDF and admin clause execution.

And all the UDF tests passed, so I think it's fine.

killme2008: ConstantVector is still an N-length vector that stores only a single repeated value and its length.

https://github.com/GreptimeTeam/greptimedb/blob/main/src/datatypes/src/vectors/constant.rs

killme2008: We have many calls to `downcast_ref` on vectors that don't check if it's constant.

Looks like it's better to revert this change.

killme2008: Addressed in https://github.com/GreptimeTeam/greptimedb/pull/6781/commits/8dec53fcfd2e34f12af370d2fbc9b25c625e2ac6

tests/cases/standalone/common/promql/scalar.result
```diff
@@ -375,7 +375,17 @@ TQL EVAL (0, 15, '5s') clamp(host, 6 - 6, 6 + 6);
...
+[
+  12.0,
+  0.0,
+  0.0,
+  0.0,
+  12.0,
+  12.0,
+[
+] }, PrimitiveVector { array: PrimitiveArray<Float64>
+] }
```
waynexia, line 388: is this data change expected?

killme2008: Yes.

PR #6784, opened 2025-08-21
feat: frontend internal grpc port

discord9:

add internal grpc port for frontend which only for internal use

## PR Checklist

### DIFF
--- a/tests-integration/src/cluster.rs
+++ b/tests-integration/src/cluster.rs
@@ -445,10 +445,10 @@ impl GreptimeDbClusterBuilder {
     }
 
     fn build_frontend_options(&self) -> FrontendOptions {
-        let mut fe_opts = FrontendOptions::default();
-
-        // is distributed test
-        fe_opts.meta_client = Some(Default::default());
+        let mut fe_opts = FrontendOptions {
+            meta_client: Some(Default::default()),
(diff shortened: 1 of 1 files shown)


src/servers/src/grpc.rs
```diff
@@ -116,6 +116,8 @@ impl GrpcOptions {
 
 const DEFAULT_GRPC_ADDR_PORT: &str = "4001";
 
+const DEFAULT_INTERNAL_GRPC_ADDR_PORT: &str = "14001";
```
sunng87: better to use 4010 as default:

- easier to write port range if any config files requires

sunng87:
We still need an integration test to ensure this internal grpc server will skip authentication. However, our current grpc intergation server doesn't cover frontend's `build_grpc_server`

src/servers/src/grpc/builder.rs
```diff
@@ -62,6 +62,7 @@ macro_rules! add_service {
 }
 
 pub struct GrpcServerBuilder {
+    name: Option<String>,
```
shuiyisong, line 65: How about just use `String` and default to `GRPC_SERVER` if not manually set?

shuiyisong (approved):
Rest LGTM

PR #6788, opened 2025-08-22
fix: follow promql rule for hanndling label of aggr

waynexia:

for a query like `sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate5m{cluster="$cluster"}) by (namespace)`, it should not return `__name__` label in the result.

expect:

image

actual:

image

## PR Checklist


evenyag:
Most queries that are not a single vector/matrix selector don't return the `__name__` label based on my test.

For example, a metric container_memory_rss
image

binary expr:
image

unary expr:
image

aggregate expr with binary expr:
image

sunng87 (approved):

PR #6810, opened 2025-08-24
perf(metric-engine)!: tsid generator

v0y4g3r:

Optimize `TsidGenerator` by replacing murmur hash with fxhash to achieve 2x performance boost in typical prometheus remote write cases.

## Security Concerns

fxhash is vulnerable to deliberately constructed collision keys which may cause outage in DoS cases. But for Prometheus remote write cases the input comes from intranet. For observability cloud services we should be able to customize hash algorithms which is left open to discussions.

## Compatiblity

This will change the time-series ordering because the tsid value for the same label values combination is changed, so this is a breaking change.

## Benchmarks

```
$ cargo bench -p metric-engine -F testing --bench bench_tsid_generator

encode_ts_id/mur3-string
                        time:   [93.603 µs 93.952 µs 94.349 µs]
                        change: [+0.8519% +1.3230% +1.7482%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 100 measurements (1.00%)

encode_ts_id/mur3-bytes time:   [94.322 µs 94.983 µs 95.766 µs]
                        change: [+3.7723% +4.2515% +4.7567%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 16 outliers among 100 measurements (16.00%)

encode_ts_id/fxhash-string
                        time:   [42.574 µs 42.597 µs 42.626 µs]
                        change: [+0.0728% +0.1195% +0.1669%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 11 outliers among 100 measurements (11.00%)

encode_ts_id/fxhash-bytes
                        time:   [42.765 µs 42.801 µs 42.847 µs]
                        change: [−0.0912% −0.0219% +0.0477%] (p = 0.55 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
```

## PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2912,7 +2912,7 @@ dependencies = [
  "cast",
  "ciborium",
  "clap 3.2.25",
- "criterion-plot",
+ "criterion-plot 0.5.0",
  "futures",
  "itertools 0.10.5",
  "lazy_static",
@@ -2939,7 +2939,7 @@ dependencies = [
  "cast",
  "ciborium",
  "clap 4.5.40",
- "criterion-plot",
+ "criterion-plot 0.5.0",
  "is-terminal",
  "itertools 0.10.5",
  "num-traits",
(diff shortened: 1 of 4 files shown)


v0y4g3r:
@evenyag @fengjiachun @waynexia WDYT about the security and compatibility issue I mentioned above?

waynexia:
murmur for tsid is a very common usage among various time series storage implementations. From this perspective, I thought changing hash function might be the last and least step of performance optimization. Could we take a look at some engineering aspects to squeeze performance, like reducing hash calls, reusing hash value etc? This might be more promising and generic, and can also alleviate the performance difference between hash functions.

PR #6821, opened 2025-08-26
Fix: correct heartbeat stream handling logic 

WenyXu:

This PR fixes a critical logic error in heartbeat stream handling where `Ok(None)` responses would not re-establish heartbeat connections, leading to permanent heartbeat failures.

### Bug Fix
**Critical Logic Error Fixed:** Previously, when heartbeat response streams returned `Ok(None)` (indicating stream closure), the code would simply break the loop without attempting to re-establish the connection. This caused **permanent heartbeat failures** where the client would stop sending heartbeats entirely after any stream interruption.

**Root Cause:** The original logic treated stream closure `Ok(None)` as a normal termination condition rather than a recoverable error that requires connection retry.

## PR Checklist

### DIFF
--- a/src/meta-srv/src/service/heartbeat.rs
+++ b/src/meta-srv/src/service/heartbeat.rs
@@ -27,10 +27,9 @@ use snafu::OptionExt;
 use tokio::sync::mpsc;
 use tokio::sync::mpsc::Sender;
 use tokio_stream::wrappers::ReceiverStream;
-use tonic::{Request, Response, Streaming};
+use tonic::{Request, Response, Status, Streaming};
 
-use crate::error;
-use crate::error::Result;
+use crate::error::{self, Result};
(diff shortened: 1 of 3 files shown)


src/meta-srv/src/service/heartbeat.rs
```diff
@@ -109,6 +108,12 @@ impl heartbeat_server::Heartbeat for Metasrv {
 
                 if is_not_leader {
                     warn!("Quit because it is no longer the leader");
+                    let _ = tx
+                        .send(Err(Status::aborted(format!(
+                            "The requested meta node is not leader, node addr: {}",
```
fengjiachun:
```suggestion
                            "The requested metasrv node is not leader, node addr: {}",
```

discord9 (approved):
no need to terminate heartbeat stream by design, LGTM

PR #6827, opened 2025-08-26
fix: gRPC auth

shuiyisong:

This PR mainly
1. removes unused gRPC auth middleware
2. refactor auth logic in bulk ingest & move common code to utils
3. add auth logic to otel arrow API

## PR Checklist


src/servers/src/grpc/flight.rs
```diff
@@ -204,7 +199,7 @@ impl FlightCraft for GreptimeRequestHandler {
                 output,
                 TracingContext::from_current_span(),
                 flight_compression,
-                query_ctx,
+                QueryContext::arc(),
```
sunng87, line 202: Can we double check it's ok to use default query context?
Is catalog/schema correct in this way?

src/servers/src/grpc/utils.rs
```diff
@@ -0,0 +1,156 @@
...
+        AuthScheme::Token(_) => Err(UnsupportedAuthScheme {
+            name: "Token AuthScheme".to_string(),
+        }),
+    }
+    .inspect_err(|e| {
+        METRIC_AUTH_FAILURE
+            .with_label_values(&[e.status_code().as_ref()])
+            .inc();
+    })
+}
```
sunng87: it's better to rename this module, auth is not a "utility"

sunng87 (commented):
Also we need some test cases to ensure this.

src/client/src/database.rs
```diff
@@ -473,8 +473,8 @@ impl Database {
         }) = &self.ctx.auth_header
         {
             let encoded = BASE64_STANDARD.encode(format!("{username}:{password}"));
-            let value =
-                MetadataValue::from_str(&encoded).context(InvalidTonicMetadataValueSnafu)?;
+            let value = MetadataValue::from_str(&format!("Basic {encoded}"))
```
fengjiachun, line 476: It seems that we also need to update the java, go and rust ingester

src/servers/src/grpc/flight.rs
shuiyisong: Yes, the query context here is only for the option. https://github.com/GreptimeTeam/greptimedb/blob/1df605ec4b4769d6ed4625db0a6fcf5ae361cc7c/src/servers/src/grpc/flight/stream.rs#L111

src/servers/src/grpc/utils.rs
shuiyisong: Renamed to `context_auth`

src/client/src/database.rs
sunng87: Does it mean this will be a breaking change?

shuiyisong: Yes, if auth is used in client. Do we need to leave this part unchanged (skip `Basic` prefix in the value)?

sunng87: We need to keep it backward compatible. If it's really necessary to add `Basic` prefix, we need to support both cases.

shuiyisong: PTAL

Issue #6779, opened 2025-08-20
Page cache may cause unexpected high resident memory


evenyag:
### What type of bug is this?

Performance issue

### What subsystems are affected?

Storage Engine

### Minimal reproduce step

1. Ingests some data into a table `my_table`. Ensure the total storage size is large enough (100GB+)
2. Executes queries like `select min(greptime_timestamp) from my_table` that searching the whole table.

### What did you expect to see?

The database memory usage decreases after the query.

### What did you see instead?

The database's memory usage remains high after the query. The total increment of the memory is much higher than the page cache size.

The memory usage:

<img width="4502" height="2240" alt="Image" src="https://github.com/user-attachments/assets/f3d87478-438f-4e5e-8e01-d2228c46194a" />

The page cache size

<img width="4516" height="2328" alt="Image" src="https://github.com/user-attachments/assets/56a73a5d-7ca9-474c-b48e-801d86b8cdc2" />

Sets `page_cache_size = 0` manually:

<img width="4488" height="2240" alt="Image" src="https://github.com/user-attachments/assets/97750aac-a9bb-46a1-84c5-46d030393659" />

### What operating system did you use?

Ubuntu 24.04

### What version of GreptimeDB did you use?

v0.16.0

### Relevant log output and stack trace

```bash

```

evenyag:
There are several places that may make the returned page reference a larger buffer.

The `gap()` we use may merge column pages into one range. Then the two pages may share the same buffer.
https://github.com/apache/opendal/blob/v0.54.0/core/src/types/read/reader.rs#L166-L192

The `Fs` backend in opendal has a buffer pool. It may reuse the same buffer from the buffer pool.
https://github.com/apache/opendal/blob/v0.54.0/core/src/services/fs/reader.rs#L36-L46

So pages returned from `fetch()` may share the same buffer. We can't release the memory unless we evict all related pages. The page size may not reflect the actual memory usage of the shared buffer.

Issue #5716, opened 2025-03-14
Read Cache is ineffective due to the non-fixed read range


zhongzc:
### What type of bug is this?

Performance issue

### What subsystems are affected?

Datanode, Storage Engine

### Minimal reproduce step

1. Scan 16GB data on S3, read cache size becomes 16GB
2. Upon scanning for the second time, there is no caching effect observed, and the read cache size inreases to 30GB

### What did you expect to see?

Cache size does not change subsequently with the number of scans, and it begins to exhibit a caching effect.

### What did you see instead?

```
279da3db8561e896f6b3f3b887cf3914.cache-bytes=38862643-41850015
279da3db8561e896f6b3f3b887cf3914.cache-bytes=38862643-41852191
279da3db8561e896f6b3f3b887cf3914.cache-bytes=388647221-391635224
279da3db8561e896f6b3f3b887cf3914.cache-bytes=388647221-391635549
279da3db8561e896f6b3f3b887cf3914.cache-bytes=391635550-394627152
279da3db8561e896f6b3f3b887cf3914.cache-bytes=391635550-394629529
279da3db8561e896f6b3f3b887cf3914.cache-bytes=394629530-397615530
279da3db8561e896f6b3f3b887cf3914.cache-bytes=394629530-397617951
279da3db8561e896f6b3f3b887cf3914.cache-bytes=397617952-400606497
279da3db8561e896f6b3f3b887cf3914.cache-bytes=397617952-400608856
279da3db8561e896f6b3f3b887cf3914.cache-bytes=4-2989724
279da3db8561e896f6b3f3b887cf3914.cache-bytes=4-2991851
279da3db8561e896f6b3f3b887cf3914.cache-bytes=400608857-403591093
279da3db8561e896f6b3f3b887cf3914.cache-bytes=400608857-403593477
279da3db8561e896f6b3f3b887cf3914.cache-bytes=403593478-406584477
279da3db8561e896f6b3f3b887cf3914.cache-bytes=403593478-406586883
279da3db8561e896f6b3f3b887cf3914.cache-bytes=406586884-409577640
279da3db8561e896f6b3f3b887cf3914.cache-bytes=406586884-409578685
279da3db8561e896f6b3f3b887cf3914.cache-bytes=409578686-412571140
279da3db8561e896f6b3f3b887cf3914.cache-bytes=409578686-412573561
279da3db8561e896f6b3f3b887cf3914.cache-bytes=412573562-415563793
```

[...]

PR #6829, opened 2025-08-26
fix: use actual buf size as cache page value size

evenyag:

fixes https://github.com/GreptimeTeam/greptimedb/issues/6779

Use the buffer size in opendal side as the page value size.

This PR changes the page cache to cache the whole row group with projection because the pages fetched may share the same buffer.

It also takes the best efforts to compute the max buffer size that pages may actually occupy for the `PageValue`.
Unfortunately, this PR removes the uncompressed page cache because it is hard to only cache the uncompressed pages. It'd be wasteful if we cache both compressed and uncompressed pages.

## PR Checklist


v0y4g3r (approved):
LGTM

evenyag:
This PR may also fix https://github.com/GreptimeTeam/greptimedb/issues/5716 because we always cache the whole row group now.

PR #6834, opened 2025-08-27
refactor!: add `enable_read_cache` config to support disable read cache explicitly

zyy17:

1. Add `enable_read_cache` options: In some scenarios(for example, when using mito as a compactor service in remote compaction), we need to disable read cache explicitly.

2. Refactor `ObjectStorageCacheConfig` and `build_cache_layer()`:

   - Add `sanitize()` for `DatanodeOptions` and `StandaloneOptions` to set up the `cache_path`;
   - Simplify the `build_cache_layer()`;

### What's breaking

Let `enable_read_cache` control whether to enable read cache or not, instead of setting `cache_path` to an empty string.

If the original config uses the empty `cache_path` to disable read cache, it should turn to use `enable_read_cache = false`.

## PR Checklist

### DIFF
--- a/tests-integration/src/test_util.rs
+++ b/tests-integration/src/test_util.rs
@@ -214,8 +214,7 @@ pub fn get_test_store_config(store_type: &StorageType) -> (ObjectStoreConfig, Te
             if *store_type == StorageType::S3WithCache {
                 s3_config.cache.cache_path = "/tmp/greptimedb_cache".to_string();
             } else {
-                // An empty string means disabling.
-                s3_config.cache.cache_path = "".to_string();
+                s3_config.cache.enable_read_cache = false;
             }
 
             let builder = S3::from(&s3_config.connection);
--- a/tests-integration/tests/http.rs
+++ b/tests-integration/tests/http.rs
@@ -1594,6 +1594,7 @@ fn drop_lines_with_inconsistent_results(input: String) -> String {
         "max_background_flushes =",
         "max_background_compactions =",
         "max_background_purges =",
+        "enable_read_cache =",
     ];
 
     input


killme2008:
We already have an option:

https://github.com/GreptimeTeam/greptimedb/blob/200422313f3b7c70c1665f280f84c1a2c71f16cb/config/datanode.example.toml#L272

Set it to be empty sting means disabling read cache.

zyy17:
> We already have an option:
> 
> https://github.com/GreptimeTeam/greptimedb/blob/200422313f3b7c70c1665f280f84c1a2c71f16cb/config/datanode.example.toml#L272
> 
> Set it to be empty sting means disabling read cache.

IMO, it's not intuitive enough. Since we already have `enable_write_cache`, it might be consistent to have `enable_read_cache`.

killme2008:
> > We already have an option:
> > https://github.com/GreptimeTeam/greptimedb/blob/200422313f3b7c70c1665f280f84c1a2c71f16cb/config/datanode.example.toml#L272
> > 
> > Set it to be empty sting means disabling read cache.
> 
> IMO, it's not intuitive enough. Since we already have `enable_write_cache`, it might be consistent to have `enable_read_cache`.

I agree. But we must "fix" the old behaviour and make this PR as breaking change.

zyy17:
> > > We already have an option:
> > > https://github.com/GreptimeTeam/greptimedb/blob/200422313f3b7c70c1665f280f84c1a2c71f16cb/config/datanode.example.toml#L272
> > > 
> > > Set it to be empty sting means disabling read cache.
> > 
> > 
> > IMO, it's not intuitive enough. Since we already have `enable_write_cache`, it might be consistent to have `enable_read_cache`.
> 
> I agree. But we must "fix" the old behaviour and make this PR as breaking change.

Let `enable_read_cache` be the master switch of the read cache.

If the `enable_read_cache` is `true`(default) and the `cache_path` is empty, it will use the default cache dir.

src/object-store/src/config.rs
```diff
@@ -304,6 +317,8 @@ impl Default for HttpClientConfig {
 #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq)]
 #[serde(default)]
 pub struct ObjectStorageCacheConfig {
+    /// Whether to enable read cache. If not set, the read cache will be enabled by default.
+    pub enable_read_cache: Option<bool>,
```
killme2008: Why not just use a boolean? Is Option unnecessary?

config/datanode.example.toml
```diff
@@ -274,6 +274,9 @@ type = "File"
 ## @toml2docs:none-default
 #+ cache_path = ""
 
+## Whether to enable read cache. If not set, the read cache will be enabled by default.
```
killme2008:
```suggestion
## Whether to enable read cache. If not set, the read cache will be enabled by default when using object storage.
```

config/standalone.example.toml
```diff
@@ -361,6 +361,9 @@ data_home = "./greptimedb_data"
 ## - `Oss`: the data is stored in the Aliyun OSS.
 type = "File"
 
+## Whether to enable read cache. If not set, the read cache will be enabled by default.
```
killme2008:
```suggestion
## Whether to enable read cache. If not set, the read cache will be enabled by default when using object storage.
```

config/standalone.example.toml
```diff
@@ -361,6 +361,9 @@ data_home = "./greptimedb_data"
...
+enable_read_cache = true
```
killme2008:
```suggestion
#+ enable_read_cache = true
```

config/datanode.example.toml
```diff
@@ -274,6 +274,9 @@ type = "File"
...
+enable_read_cache = true
```
killme2008:
```suggestion
#+ enable_read_cache = true
```

killme2008:
@evenyag @WenyXu PTAL

src/datanode/src/store.rs
```diff
@@ -76,94 +80,37 @@ pub(crate) async fn new_object_store(
...
-
-        Ok(Some(cache_layer))
-    } else {
-        Ok(None)
-    }
+    let atomic_temp_dir = join_dir(&cache_config.cache_path, ATOMIC_WRITE_DIR);
+    clean_temp_dir(&atomic_temp_dir).context(error::ObjectStoreSnafu)?;
+
+    // Compatible code. Remove this after a major release.
+    let old_atomic_temp_dir = join_dir(&cache_config.cache_path, OLD_ATOMIC_WRITE_DIR);
```
killme2008: It can be removed currently

killme2008 (approved):
LGTM

WenyXu (approved):

src/datanode/src/store.rs
```diff
@@ -76,94 +78,33 @@ pub(crate) async fn new_object_store(
...
-            .root(path)
-            .atomic_write_dir(&atomic_temp_dir)
-            .build()
-            .context(error::BuildCacheStoreSnafu)?;
+    let cache_layer = LruCacheLayer::new(
+        Arc::new(cache_store),
+        cache_config.cache_capacity.0 as usize,
+    )
+    .context(error::BuildCacheStoreSnafu)?;
+    cache_layer.recover_cache(false).await;
```
evenyag, line 102: nitpick: We can add a log before recover_cache so we can know the datanode is recovering.

tests-integration/src/test_util.rs
```diff
@@ -212,10 +212,10 @@ pub fn get_test_store_config(store_type: &StorageType) -> (ObjectStoreConfig, Te
             let mut s3_config = s3_test_config();
 
             if *store_type == StorageType::S3WithCache {
-                s3_config.cache.cache_path = Some("/tmp/greptimedb_cache".to_string());
+                s3_config.cache.cache_path = "/tmp/greptimedb_cache".to_string();
             } else {
                 // An empty string means disabling.
```
evenyag: Is it outdated?

src/datanode/src/store.rs
zyy17: I think it's unnecessary. There are the logs in `recover_cache()`.

PR #6844, opened 2025-08-28
feat: file ref mgr

discord9:

file reference manager, which can produce a list of all referenced files by current datanode and return using heartbeat's mailbox, useful for gc worker so it wouldn't delete still in referenced files

The intended usage is like this:
1. metasrv try to determine which table need gc by checking it's to be deleted files number.
2. If needed to gc a certain table, metasrv first send a get file ref command to all datanode that have this table's region, and get all file refs in file id.
3. Then metasrv send a gc command to one of the datanodes with ref file ids, 
4. which datanode will
also read manifest and union with ref file ids then delete files that has been unused for a while(in case tmp ref files is a little bit late)

## PR Checklist


evenyag:
> file reference manager, which can produce a list of all referenced files by the current datanode and upload to `<table-id>/.purger-refs/` folder when use object storage, useful for gc worker so it wouldn't delete still in referenced files

I prefer to use a global directory for this. It'd be easier to inspect the reference by a human without checking each table directory. What's more, we don't have to use a hidden directory.

Maybe we can use a global directory under root:
```
/file-refs/<table-id>/xxx
```

We should also consider how to clean up the ref files. For example, a datanode may clean the reference after it restarts.

What do you think? @v0y4g3r @waynexia @WenyXu

src/mito2/src/compaction/compactor.rs
```diff
@@ -196,6 +197,7 @@ pub async fn open_compaction_region(
             purge_scheduler.clone(),
             access_layer.clone(),
             None,
+            Arc::new(FileReferenceManager::new(0)),
```
evenyag: Maybe we should pass None. A manager with node 0 may be confusing in the future.

src/mito2/src/engine.rs
```diff
@@ -319,7 +338,7 @@ impl MitoEngine {
         self.find_region(id)
     }
 
-    fn find_region(&self, region_id: RegionId) -> Option<MitoRegionRef> {
+    pub fn find_region(&self, region_id: RegionId) -> Option<MitoRegionRef> {
```
evenyag: I wonder if this has potential issues. I didn't pub this because it will be easy to modify the MitoRegionRef outside of the mito worker.

At least we should not pub it now.

src/mito2/src/sst/file_ref.rs
```diff
@@ -0,0 +1,377 @@
...
+
+/// File references for a table.
+/// It contains all files referenced by the table.
+#[derive(Debug, Clone, Default)]
+pub struct TableFileRefs {
+    /// (FileRef, Ref Count) meaning how many FileHandleInner is opened for this file.
+    pub files: HashMap<FileRef, usize>,
+}
+
+pub const REFS_PATH: &str = ".refs";
```
evenyag: Don't use `hidden` directories.

src/mito2/src/sst/file.rs
```diff
@@ -333,16 +333,16 @@ struct FileHandleInner {
...
     fn drop(&mut self) {
-        if self.deleted.load(Ordering::Relaxed) {
-            self.file_purger.send_request(PurgeRequest {
-                file_meta: self.meta.clone(),
-            });
-        }
+        self.file_purger.send_request(PurgeRequest {
+            file_meta: self.meta.clone(),
+            deleted: self.deleted.load(Ordering::Relaxed),
+        });
```
evenyag: I prefer not to do this via the file purger. Because it will create a future and send the request to the scheduler. The scheduler may drop requests if the channel is full.

Updating the file references is sync so we can do it in drop directly.

src/mito2/src/sst/file_ref.rs
```diff
@@ -0,0 +1,377 @@
...
+/// And periodically update the references to tmp file in object storage.
+/// This is useful for ensuring that files are not deleted while they are still in use by any
+/// query.
+#[derive(Debug)]
+pub struct FileReferenceManager {
+    /// Datanode id. used to determine tmp ref file name.
+    node_id: u64,
+    /// TODO(discord9): use no hash hasher since table id is sequential.
+    files_per_table: DashMap<TableId, TableFileRefs>,
+}
```
evenyag: Maybe we can add a metric for the referenced file count.

src/mito2/src/sst/file_purger.rs
```diff
@@ -143,6 +160,29 @@ impl FilePurger for LocalFilePurger {
...
+    }
+
+    fn add_new_file(&self, file_meta: &FileMeta) {
+        if self.is_local_fs {
+            // If the access layer is local file system, we don't need to track the file reference.
+            return;
+        }
+        self.file_ref_manager.add_file(file_meta);
+    }
+}
```
WenyXu: Should we implement two separate `FilePurgers` instead of mixing them together? For example, `LocalFsFilePurger` and `ObjectStoreFilePurger`.

src/mito2/src/metrics.rs
```diff
@@ -437,6 +437,12 @@ lazy_static! {
             "mito stalled write request in each worker",
             &[WORKER_LABEL]
         ).unwrap();
+    /// Number of ref files per table
+    pub static ref GC_TABLE_REF_FILE_CNT: IntGaugeVec = register_int_gauge_vec!(
+            "greptime_gc_table_ref_file_count",
+            "gc table ref file count",
+            &["table_id"]
```
evenyag: Will it be a problem if there are too many tables?

evenyag (approved):
Rest LGTM

src/mito2/src/metrics.rs
discord9: > Will it be a problem if there are too many tables?

probably, then only store one number or what?

evenyag: I prefer to store only one number.

waynexia (approved):

PR #6848, opened 2025-08-28
refactor: add stop methods for `LocalFilePurger` and `CompactionRegion`

zyy17:

1. Add `stop()` for `LocalFilePurger`;
2. Add `stop_file_purger()` for `CompactionRegion` by using `stop()` of `LocalFilePurger`

## PR Checklist

### DIFF
--- a/src/mito2/src/compaction/compactor.rs
+++ b/src/mito2/src/compaction/compactor.rs
@@ -247,10 +247,10 @@ impl CompactionRegion {
         self.file_purger.clone()
     }
 
-    /// Stop the file purger of the compaction region.
-    pub async fn stop_file_purger(&self) -> Result<()> {
+    /// Stop the file purger scheduler of the compaction region.
+    pub async fn stop_purger_scheduler(&self) -> Result<()> {
         if let Some(file_purger) = &self.file_purger {
-            file_purger.stop().await
+            file_purger.stop_scheduler().await
         } else {
             Ok(())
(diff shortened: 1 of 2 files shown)


src/mito2/src/sst/file_purger.rs
```diff
@@ -76,6 +77,11 @@ impl LocalFilePurger {
             cache_manager,
         }
     }
+
+    /// Stop the scheduler of the file purger.
+    pub async fn stop(&self) -> Result<()> {
+        self.scheduler.stop(true).await
+    }
```
evenyag: Maybe we should rename it to stop_scheduler because it stops the shared scheduler not the purger itself.

src/mito2/src/compaction/compactor.rs
```diff
@@ -242,9 +242,19 @@ pub async fn open_compaction_region(
...
     }
+
+    /// Stop the file purger of the compaction region.
+    pub async fn stop_file_purger(&self) -> Result<()> {
+        if let Some(file_purger) = &self.file_purger {
+            file_purger.stop().await
+        } else {
+            Ok(())
+        }
+    }
```
evenyag: Maybe rename to stop_purger_scheduler. We should use it carefully because it stops the shared scheduler, and we don't need to call it most time.

v0y4g3r (approved):
LGTM

Issue #6558, opened 2025-07-18
Tracking issue: repartition


waynexia:
### What does the feature do?

See this RFC https://github.com/GreptimeTeam/greptimedb/pull/6557

### Implementation challenges

- [x] Cross-region file reference https://github.com/GreptimeTeam/greptimedb/issues/6409
- [x] Global GC worker #6571 
- [ ] Rule checker https://github.com/GreptimeTeam/greptimedb/issues/6450
- Metasrv
  - [x] Procedure (2 weeks)
    - [ ] Scaffold https://github.com/GreptimeTeam/greptimedb/pull/7077
  - [ ] Allocate/Deallocate resources
- Frontend
  - [x] Grammar and parser (1 week) #7082
  - [ ] No ingestion switch (2 weeks)
- Region Server
  - [ ] Modify manifest
    - [x] Store PartitionRule https://github.com/GreptimeTeam/greptimedb/pull/6699
  - [ ] Region Edit (1 week)
  - [x] Compute manifest (2 weeks) #7071
  - [ ] Rollback manifest (1 week)
  - [ ] Stage changes
    - [x] Region Staging Mode https://github.com/GreptimeTeam/greptimedb/pull/6664
    - [x] Stage flush changes #6726
    - [x] Exit Staging Mode (1 week) https://github.com/GreptimeTeam/greptimedb/pull/6913
    - [x] Enter Staging Mode #7261
    - [x] Remap manifests interface #7265
    - [ ] Exit Staging Mode with manifest rollback
- [ ] Special compaction

PR #6849, opened 2025-08-28
feat: store partition expr per file in region manifest

waynexia:

#6558

Store partition expr in `FileMeta` and then region manifest.

## PR Checklist


src/mito2/Cargo.toml
```diff
@@ -59,6 +59,7 @@ paste.workspace = true
 pin-project.workspace = true
 prometheus.workspace = true
 prost.workspace = true
+partition.workspace = true
```
waynexia, line 63: make sure there is no dep issue, will use later

src/mito2/src/region.rs
```diff
@@ -511,7 +511,7 @@ impl MitoRegion {
             .flat_map(|level| {
                 level.files().map(|file| {
                     let meta = file.meta_ref();
-                    let region_id = meta.region_id;
+                    let region_id = self.region_id;
```
waynexia: previous impl is not correct. the region id from file's region meta is the id who created that file, which might be different from the region that currently owns this file

src/mito2/src/sst/file.rs
```diff
@@ -413,4 +471,117 @@ mod tests {
...
+                {"value": 0, "unit": "Millisecond"}
+            ],
+            "level": 0,
+            "file_size": 0,
+            "available_indexes": [],
+            "index_file_size": 0,
+            "num_rows": 0,
+            "num_row_groups": 0,
+            "sequence": null,
+            "partition_expr": null
```
zhongzc, line 583: Also test the scenario where this field is not present

src/mito2/src/sst/file.rs
```diff
@@ -167,6 +208,18 @@ pub struct FileMeta {
...
+    /// This is stored as a PartitionExpr object in memory for convenience,
+    /// but serialized as JSON string for manifest compatibility.
+    /// Compatibility behavior:
+    /// - None: no partition expr was set when the file was created (legacy files).
+    /// - Some(expr): partition expression from region metadata.
+    #[serde(
+        serialize_with = "serialize_partition_expr",
+        deserialize_with = "deserialize_partition_expr"
+    )]
+    pub partition_expr: Option<PartitionExpr>,
```
evenyag, line 222: I'm worried about whether it requires too much additional memory to keep a partition expr for each `FileMeta`. Also, it may be wasteful to store the same expr again and again in the manifest. We may have thousands of files for each region, and more.

Do you have any ideas? @zhongzc @v0y4g3r 

Can we only store the hash of the expr in the manifest and keep a full map from the hash to the expr in the `RegionManifest`?

I'm not sure if this is an over-optimization.

evenyag: To save memory, maybe we can arc the expr and find some place to optimize this after deserialization.

zhongzc: I actually don't understand why the file meta would have a partition expr. Different regions with different partitions would read the same file—that's what I've always thought.

v0y4g3r: I would also suggestion not to keep an instance of `PartitionExpr` for each `FileMeta` because of it's memory footprint and serialization/deserialization overhead.

waynexia: we need to be able to get the partition expr of a file. it's stored in parquet metadata but that would be much more consuming if we fetch exprs from every single file. 

>Different regions with different partitions would read the same file—that's what I've always thought.

That's true, but the expression stored in parquet metadata and file meta are the partition when that file is generated, not the expression of current region. So does the region id.

zhongzc: @evenyag The average size of `PartitionExpr` should be around a few dozen bytes, which is acceptable.

src/mito2/src/region.rs
evenyag: Should we also display the creator region of the file? @zhongzc

evenyag (approved):

src/mito2/src/sst/file.rs
waynexia: I tried to do so in https://github.com/GreptimeTeam/greptimedb/pull/6849/commits/84725b2e5ce0b1bd4dbc8ff943f94ef56c88defe. It's not quite viable. Reusing partition expr in file meta means you can't build and use a file meta alone, but have to pass that map container all around to get a reference to arc-ed expr

evenyag: It's fine to postpone this.

src/mito2/src/region.rs
zhongzc: It needs

zhongzc (approved):

src/mito2/src/region.rs
zhongzc: I'm on it.

evenyag:
One commit needs sigoff

PR #6852, opened 2025-08-29
feat: flow prom ql auto sink table is also promql-able

discord9:

as title, change val column name to `val DOUBLE NULL` and tag column `STRING null

## PR Checklist

### DIFF
--- a/tests/cases/distributed/flow-tql/flow_tql.result
+++ b/tests/cases/distributed/flow-tql/flow_tql.result
@@ -2,7 +2,7 @@ CREATE TABLE http_requests (
   ts timestamp(3) time index,
   host STRING,
   idc STRING,
-  val BIGINT,
+  val DOUBLE,
   PRIMARY KEY(host, idc),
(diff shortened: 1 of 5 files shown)


tests/cases/distributed/flow-tql/flow_tql.result
```diff
@@ -2,7 +2,7 @@ CREATE TABLE http_requests (
   ts timestamp(3) time index,
   host STRING,
   idc STRING,
-  val BIGINT,
+  val DOUBLE,
```
evenyag, line 5: Is this a breaking change?

evenyag (approved):

tests/cases/distributed/flow-tql/flow_tql.result
discord9: > Is this a breaking change?

not really, since previous testcase is just wrong

WenyXu (approved):

PR #6862, opened 2025-08-30
feat(mito): backfill partition expr on region open

zhongzc:

Older regions didn’t persist partition_expr in engine metadata; new regions do. To unify behavior and support online upgrade, we backfill partition_expr on region open and persist it to the manifest.

Added on-open backfill to persist partition_expr for legacy regions.

## PR Checklist


fengjiachun (approved):
LGTM

src/mito2/src/region/opener.rs
```diff
@@ -382,6 +399,30 @@ impl RegionOpener {
...
+                    let new_meta = Arc::new(builder.build().context(InvalidMetadataSnafu)?);
+
+                    let action = RegionMetaAction::Change(RegionChange { metadata: new_meta });
+                    // Use follower state during open.
+                    let _ = manifest_manager
+                        .update(
+                            RegionMetaActionList::with_action(action),
+                            RegionRoleState::Follower,
+                        )
+                        .await?;
```
evenyag: I wonder if this has the potential race condition in cluster mode.

Can we trigger a manifest checkpoint after the region becomes a leader intead?

src/mito2/src/test_util.rs
```diff
@@ -280,6 +296,7 @@ impl TestEnv {
                 log_store,
                 zelf.object_store_manager.as_ref().unwrap().clone(),
                 zelf.schema_metadata_manager.clone(),
+                zelf.partition_expr_fetcher.clone(),
```
WenyXu, line 303:
```suggestion
                self.partition_expr_fetcher.clone(),
```

src/mito2/src/region/opener.rs
WenyXu: +1, The datanode may open the readonly regions

src/datanode/src/datanode.rs
```diff
@@ -647,6 +657,16 @@ async fn open_all_regions(
...
+            } else {
+                // Finalize leadership: persist backfilled metadata.
+                if let SetRegionRoleStateResponse::InvalidTransition(err) = region_server
+                    .set_region_role_state_gracefully(region_id, SettableRegionRoleState::Leader)
+                    .await?
+                {
+                    error!(
+                        err; "failed to convert region {region_id} to leader"
+                    );
+                }
```
evenyag: @WenyXu Do you think we should set the state again here?

src/mito2/src/region.rs
```diff
@@ -423,6 +425,25 @@ impl MitoRegion {
             }
         }
 
+        // Hack(zhongzc): If we have just become leader (writable), persist any backfilled metadata
+        // and trigger a manifest checkpoint.
```
evenyag: We don't have to trigger the checkpoint now.

src/mito2/src/region.rs
```diff
@@ -423,6 +425,25 @@ impl MitoRegion {
...
+            if manifest_meta.partition_expr.is_none() && current_meta.partition_expr.is_some() {
+                let action = RegionMetaAction::Change(RegionChange {
+                    metadata: current_meta.clone(),
+                });
+                let _ = manager
+                    .update(
+                        RegionMetaActionList::with_action(action),
+                        RegionRoleState::Leader(RegionLeaderState::Writable),
+                    )
+                    .await;
```
evenyag: Maybe we should add a log when we backfill the metadata, including the manifest version.
If we can't backfill, we can log a warning.

src/mito2/src/worker/handle_catchup.rs
```diff
@@ -126,6 +126,10 @@ impl<S: LogStore> RegionWorkerLoop<S> {
 
         if request.set_writable {
             region.set_role(RegionRole::Leader);
+            // Finalize leadership: persist backfilled metadata.
+            region
+                .set_role_state_gracefully(SettableRegionRoleState::Leader)
+                .await?;
```
evenyag, line 132: @WenyXu Do we need to call this here?

src/datanode/src/datanode.rs
WenyXu: The `open_with_writable` always be false in distributed mode..

src/datanode/src/datanode.rs
```diff
@@ -655,12 +665,25 @@ async fn open_all_regions(
...
+                        )
+                        .await?
+                    {
+                        error!(err; "failed to convert region {region_id} to leader");
+                    }
+                }
+                Err(e) => {
+                    error!(e; "failed to convert region {region_id} to leader");
+                }
+            }
```
WenyXu, line 686: How about to call `set_region_role_state_gracefully` directly?

WenyXu (commented):
Rest LGTM

src/datanode/src/datanode.rs
zhongzc: Tried but follower cannot be transfered to leader via `set_region_role_state_gracefully `.

WenyXu (approved):

PR #6867, opened 2025-09-01
fix(flow): promql auto create table

discord9:

as title, fix flow use promql query without aggregation node auto created table will be wrong
- now if aggr node is not found using tql, it will determine first f64 column as val column, and all string column as tag

- also change the val column's name from "val" to datafusion's expr name
so that for case which tql eval might return multiple float64 values case(like `count_values`) there wouldn't be multiple same name "val" column

## PR Checklist

### DIFF
--- a/src/flow/src/batching_mode/engine.rs
+++ b/src/flow/src/batching_mode/engine.rs
@@ -29,10 +29,15 @@ use common_runtime::JoinHandle;
 use common_telemetry::tracing::warn;
 use common_telemetry::{debug, info};
 use common_time::TimeToLive;
+use datafusion_common::tree_node::{TreeNodeRecursion, TreeNodeVisitor};
+use datafusion_expr::LogicalPlan;
+use datatypes::prelude::ConcreteDataType;
 use query::QueryEngineRef;
+use session::context::QueryContext;
 use snafu::{ensure, OptionExt, ResultExt};
 use sql::parsers::utils::is_tql;
 use store_api::storage::{RegionId, TableId};
+use table::table_reference::TableReference;
 use tokio::sync::{oneshot, RwLock};
 
 use crate::batching_mode::frontend_client::FrontendClient;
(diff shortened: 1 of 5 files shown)


tests/cases/distributed/flow-tql/flow_tql.result
```diff
@@ -248,3 +248,78 @@ DROP TABLE rate_reqs;
...
+|           |   "ts" TIMESTAMP(3) NOT NULL,                             |
+|           |   "prom_rate(ts_range,byte,ts,Int64(60000))" DOUBLE NULL, |
+|           |   "host" STRING NULL,                                     |
+|           |   "job" STRING NULL,                                      |
+|           |   "instance" STRING NULL,                                 |
+|           |   TIME INDEX ("ts"),                                      |
+|           |   PRIMARY KEY ("host", "job", "instance")                 |
+|           | )                                                         |
+|           |                                                           |
+|           | ENGINE=mito                                               |
```
evenyag, line 339: Can users use tql to query rate_reqs?

src/flow/src/batching_mode/task.rs
```diff
@@ -809,15 +823,48 @@ fn create_table_with_expr(
...
+                None
+            }
+        })
+        .collect::<Vec<_>>();
+
+    Ok(TableDef {
+        ts_col: first_time_stamp,
+        pks: string_columns,
+    })
+}
```
WenyXu, line 871: Should we validate that the source table contains only a single value column?

src/flow/src/batching_mode/engine.rs
```diff
@@ -484,6 +493,131 @@ impl BatchingEngine {
...
+                get_table_info(self.table_meta.table_info_manager(), &table_id).await?;
+            // first check if it's only one f64 value column
+            let value_cols = table_info
+                .table_info
+                .meta
+                .schema
+                .column_schemas
+                .iter()
+                .filter(|col| col.data_type == ConcreteDataType::float64_datatype())
+                .collect::<Vec<_>>();
```
evenyag, line 566: You can use `.count()` to get value num.

evenyag (approved):

src/flow/src/batching_mode/engine.rs
discord9: > You can use `.count()` to get value num.

error msg need column schema so collect

PR #6877, opened 2025-09-02
chore: add some trace logs in fetching data from cache and object store

zyy17:

Add some trace logs in fetching data from the cache and object store.

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet/helper.rs
+++ b/src/mito2/src/sst/parquet/helper.rs
@@ -17,7 +17,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use bytes::Bytes;
-use common_telemetry::debug;
+use common_telemetry::trace;
 use object_store::ObjectStore;
 use parquet::basic::ColumnOrder;
 use parquet::file::metadata::{FileMetaData, ParquetMetaData, RowGroupMetaData};
@@ -116,8 +116,8 @@ pub async fn fetch_byte_ranges(
         .map(|buf| buf.to_bytes())
         .collect::<Vec<_>>();
 
-    debug!(
-        "Fetch {} bytes from {}, cost: {:?}",
+    trace!(
+        "Fetch {} bytes from '{}' in object store, cost: {:?}",
         total_size,
         file_path,
         start.elapsed()
--- a/src/object-store/src/layers/lru_cache/read_cache.rs
+++ b/src/object-store/src/layers/lru_cache/read_cache.rs
@@ -14,7 +14,7 @@
 
 use std::sync::Arc;
 
-use common_telemetry::debug;
+use common_telemetry::{debug, trace};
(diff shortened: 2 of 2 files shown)


src/mito2/src/sst/parquet/helper.rs
```diff
@@ -109,5 +114,14 @@ pub async fn fetch_byte_ranges(
         .await?
         .into_iter()
         .map(|buf| buf.to_bytes())
-        .collect::<Vec<_>>())
+        .collect::<Vec<_>>();
+
+    debug!(
```
evenyag: trace?
```suggestion
    trace!(
```

shuiyisong (approved):

src/mito2/src/sst/parquet/helper.rs
```diff
@@ -109,5 +114,14 @@ pub async fn fetch_byte_ranges(
...
         .map(|buf| buf.to_bytes())
-        .collect::<Vec<_>>())
+        .collect::<Vec<_>>();
+
+    trace!(
+        "Fetch {} bytes from '{}', cost: {:?}",
+        total_size,
+        file_path,
+        start.elapsed()
+    );
```
evenyag, line 124: We already have a timer in the caller of this method. Can we move this log to there?

evenyag: See, and we can log more info about this in the row group.
https://github.com/GreptimeTeam/greptimedb/blob/9175fa643d5345fa62a1dbbb8d7935c7de47c9da/src/mito2/src/sst/parquet/row_group.rs#L292-L304

zyy17:
```rust
                trace!(
                    "Write cache miss for key '{}' (region_id='{}', file_id='{}', row_group_idx='{}') and fetch {} bytes from file '{}' in read cache or object store, cost: {:?}",
                    key,
                    self.region_id,
                    self.file_id,
                    self.row_group_idx,
                    total_size,
                    self.file_path,
                    timer.stop_and_record()
                );
```

zyy17: Maybe it's complicated to add logs here. `fetch_byte_ranges()` is the final function that is used for reading files.

src/mito2/src/sst/parquet/helper.rs
```diff
@@ -109,5 +114,14 @@ pub async fn fetch_byte_ranges(
...
         .into_iter()
         .map(|buf| buf.to_bytes())
-        .collect::<Vec<_>>())
+        .collect::<Vec<_>>();
+
+    trace!(
+        "Fetch {} bytes from '{}' in object store, cost: {:?}",
+        total_size,
+        file_path,
+        start.elapsed()
```
waynexia, line 123: use `#[instrument]` if you want to know an unbiased time in async code

PR #6880, opened 2025-09-02
fix: flow full aggr only trigger on new data

discord9:

## PR Checklist


waynexia (approved):

src/flow/src/batching_mode/engine.rs
```diff
@@ -154,6 +154,7 @@ impl BatchingEngine {
                 for src_table_name in src_table_names {
                     if let Some((timestamps, unit)) = group_by_table_name.get(src_table_name) {
                         let Some(expr) = &task.config.time_window_expr else {
+                            task.state.write().unwrap().dirty_time_windows.set_dirty();
```
killme2008: Can we mark dirty outside the loop? Use a boolean and mark it once.

PR #6891, opened 2025-09-03
fix: move prune_region_dir to region drop

zhongzc:

* The `prune_region_dir` call was placed in the wrong spot.
* Invoke it during region drop to align with the region lifecycle.

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/index/indexer/abort.rs
+++ b/src/mito2/src/sst/index/indexer/abort.rs
@@ -21,6 +21,7 @@ impl Indexer {
         self.do_abort_inverted_index().await;
         self.do_abort_fulltext_index().await;
         self.do_abort_bloom_filter().await;
+        self.do_prune_intm_sst_dir().await;
         self.puffin_manager = None;
     }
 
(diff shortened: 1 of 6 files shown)


src/mito2/src/sst/index/indexer/abort.rs
```diff
@@ -21,7 +21,6 @@ impl Indexer {
         self.do_abort_inverted_index().await;
         self.do_abort_fulltext_index().await;
         self.do_abort_bloom_filter().await;
-        self.do_prune_intm_sst_dir().await;
```
evenyag: So we don't prune sst dir in abort now?

zhongzc: If want to actively delete, we can prune during abort. I tend to prefer not being that aggressive; aligning with the SST lifecycle is also acceptable.

waynexia (approved):

src/mito2/src/sst/index/indexer/abort.rs
evenyag: The problem is that if we only abort this while we have issues writing the SST/index. It returns an error and thus doesn't have a FileHandle to purge. That's why we clean manually here.

https://github.com/GreptimeTeam/greptimedb/blob/7cf47ccf5430321786204ac8eccec74623a84def/src/mito2/src/sst/parquet/writer.rs#L212-L232

evenyag (approved):

Issue #4856, opened 2024-10-18
High level framework for ergonomic UDF programming


sunng87:
### What type of enhancement is this?

Refactor

### What does the enhancement do?

The idea is to create a high level framework for UDF development (not UDAF), to
remove boilerplate code, and improve ergonomic.

The core responsibility of this framework is to provide:

- Automatic function argument extraction/coerce
- Declarative validation
- Document generation

## Current status

At the moment, a typical implementation of UDF looks like this one:
https://github.com/GreptimeTeam/greptimedb/blob/main/src/common/function/src/scalars/geo/h3.rs#L95

Basically we do following steps to generate the result vector:

1. Validate input columns: column size and length for each column
2. Initialise the result vector
3. Extract values by row from columns
4. Call the rust function and dealing with error if any
5. Fill the result vector

## Desired state

Because every implementation has do these 1/2/3/5 steps. An ergonomic solution
is to provide a declarative way to extract rust data types from column vectors,
and the user simply focus on calling rust function. The implementation of UDF
should be stateless, so until we have a real case, we don't need to provide any
type of context for execution except the original `FunctionContext`.

Inspired by how axum designed its web handler. The API looks like

```rust

trait Extract {
    fn extract(v: Value) -> Option<Self>;

    fn validate(&self) -> Result<()>;
}

struct Coordinate(f64);

impl Extract for Coordinate {

    fn extract(v: Value) -> Option<Self> {
        ...
    }

    fn validate(&self) -> Result<()> {
        Ok(())
    }
}

struct Resolution(i8);

impl Extract for Coordinate {

    fn extract(v: Value) -> Option<Self> {
        ...
    }

    fn validate(&self) -> Result<()> {
        ensure!(self.0 >=0 && self.0 < 18)
    }
}

trait FunctionExt1: Function {
    type T0: Extract;
    fn call(_ctx: FunctionContext, arg0: T0) -> R;
}

trait FunctionExt2: Function {
    type T0: Extract;
    type T1: Extract;
    fn call(_ctx: FunctionContext, arg0: T0, arg1: T1) -> R;
}

trait FunctionExt3: Function {
    type T0: Extract;
    type T1: Extract;
    type T2: Extract;
    fn call(_ctx: FunctionContext, arg0: T0, arg1: T1, arg2: T2) -> R;
}



...
```

`FunctionExtN` will provide default implementation for `Function::eval`.

TODO: think about how to detail with `R`

## Limitation

[...]

PR #6908, opened 2025-09-04
refactor: use DataFusion's Signature directly in UDF

MichaelScofield:

https://github.com/GreptimeTeam/greptimedb/issues/4856

Part 1 of gradually adapting DataFusion's implementation of our UDF.

## PR Checklist

### DIFF
--- a/src/common/function/src/scalars/ip/cidr.rs
+++ b/src/common/function/src/scalars/ip/cidr.rs
@@ -16,7 +16,8 @@ use std::net::{Ipv4Addr, Ipv6Addr};
 use std::str::FromStr;
 
 use common_query::error::{InvalidFuncArgsSnafu, Result};
-use datafusion_expr::{Signature, TypeSignature, Volatility};
+use datafusion_common::types;
+use datafusion_expr::{Coercion, Signature, TypeSignature, TypeSignatureClass, Volatility};
 use datatypes::arrow::datatypes::DataType;
 use datatypes::prelude::{ConcreteDataType, Value};
 use datatypes::scalars::ScalarVectorBuilder;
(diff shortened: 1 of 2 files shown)


src/common/function/src/scalars/geo/measure.rs
```diff
@@ -46,10 +46,7 @@ impl Function for STDistance {
 
     fn signature(&self) -> Signature {
         Signature::new(
-            TypeSignature::Exact(vec![
-                ConcreteDataType::string_datatype(),
-                ConcreteDataType::string_datatype(),
-            ]),
+            TypeSignature::Exact(vec![DataType::Utf8, DataType::Utf8]),
```
waynexia: it would be better and more generic to map string data type to {`utf8`|`large utf8`|`utf8 view`}, as well as bytes type

src/common/function/src/scalars/json/json_path_exists.rs
```diff
@@ -159,20 +146,20 @@ mod tests {
                          } if valid_types ==
             vec![
                 TypeSignature::Exact(vec![
-                    ConcreteDataType::json_datatype(),
-                    ConcreteDataType::string_datatype(),
+                    DataType::Binary,
```
waynexia, line 150: fwiw changing from json to binary doesn't change the underlying physical type, but loses a layer of verification on the type system level. maybe we need to ensure the input data is valid jsonb or be able to handle invalid data. just mark it. we can do this in the future as json type is in a very early experimental stage

waynexia (approved):
rest lgtm

src/common/function/src/scalars/geo/measure.rs
sunng87: Is there any shortcuts for this in datafusion? There will be a few combinations.

sunng87: Will datafusion do auto coerce for utf8/utf8view? I can call a udf defined in utf8 with a string literal

waynexia: maybe no as I know, they will be boiled down to different concrete arrow array types, a nightmare to udf

waynexia: there is something like string array accessor, but it can't handle everything

PR #6913, opened 2025-09-05
feat: exiting staging mode on success case

waynexia:

#6558

on success, we only need to make all manifests visible

## PR Checklist


src/mito2/src/manifest/storage.rs
```diff
@@ -661,6 +683,69 @@ impl ManifestObjectStore {
...
+
+        let delete_tasks = staging_entries.iter().map(|entry| async {
+            debug!("Deleting staging manifest file: {}", entry.path());
+            self.object_store
+                .delete(entry.path())
+                .await
+                .context(OpenDalSnafu)
+        });
+
+        try_join_all(delete_tasks).await?;
```
evenyag: Can we use the deleter api?
https://docs.rs/opendal/latest/opendal/struct.Operator.html#method.deleter

src/mito2/src/region.rs
```diff
@@ -297,19 +298,30 @@ impl MitoRegion {
...
-    pub(crate) fn set_staging(&self) -> Result<()> {
+    /// Cleans any existing staging manifests before entering staging mode.
+    pub(crate) async fn set_staging<'a, 'b>(
+        &'a self,
+        manager: RwLockWriteGuard<'b, RegionManifestManager>,
+    ) -> Result<()>
+    where
+        'a: 'b,
+    {
+        manager.store().clear_staging_manifests().await?;
```
evenyag, line 311: If we call this before setting staging, maybe we can use remove_all?
https://docs.rs/opendal/latest/opendal/struct.Operator.html#method.remove_all

src/mito2/src/region.rs
```diff
@@ -297,19 +298,30 @@ impl MitoRegion {
...
     /// You should call this method in the worker loop.
     /// Transitions from Writable to Staging state.
-    pub(crate) fn set_staging(&self) -> Result<()> {
+    /// Cleans any existing staging manifests before entering staging mode.
+    pub(crate) async fn set_staging<'a, 'b>(
+        &'a self,
+        manager: RwLockWriteGuard<'b, RegionManifestManager>,
+    ) -> Result<()>
+    where
+        'a: 'b,
```
evenyag: non-blocking: maybe we can pass `&mut RegionManifestManager` from the guard.
```suggestion
    pub(crate) async fn set_staging(
        &self,
        manager: &mut RegionManifestManager,
    ) -> Result<()>
```

src/mito2/src/region.rs
waynexia: looks like we can always `remove_all` in any case. it's much simplified

src/mito2/src/region.rs
waynexia: accept an owned guard can make it simple. it's dropped after the function. changing it to reference is more error-prone

evenyag (approved):
LGTM. There are some conflicts.

PR #6918, opened 2025-09-05
feat: upgraded pg_catalog support

sunng87:

This patch will pull in datafusion-postgres for an enhanced postgres support on datafusion.

## PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -335,9 +335,9 @@ dependencies = [
 
 [[package]]
 name = "arrow-array"
-version = "56.0.0"
+version = "56.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e2730bc045d62bb2e53ef8395b7d4242f5c8102f41ceac15e8395b9ac3d08461"
+checksum = "744109142cdf8e7b02795e240e20756c2a782ac9180d4992802954a8f871c0de"
 dependencies = [
  "ahash 0.8.12",
  "arrow-buffer",
@@ -352,9 +352,9 @@ dependencies = [
 
 [[package]]
 name = "arrow-buffer"
-version = "56.0.0"
+version = "56.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "54295b93beb702ee9a6f6fbced08ad7f4d76ec1c297952d4b83cf68755421d1d"
+checksum = "601bb103c4c374bcd1f62c66bcea67b42a2ee91a690486c37d4c180236f11ccc"
 dependencies = [
  "bytes",
  "half",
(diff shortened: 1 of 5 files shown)


sunng87:
This patch has not included all pg_catalog compatibility fixes from datafusion-postgres. We will update again when we upgraded to datafusion 50.

sunng87:
A summary of this patch:

- replaced all pg_catalog implementation with datafusion-postgres
- bridge between our API with datafusion-postgres: mostly for reading recordbatches from datafusion tables
- register UDFs and UDTFs
- fix an issue with sqlness runner on postgresql output
- update sqlness results for these tables
- use datafusion-postgres sql rewriter to rewrite sql into the form that works with datafusion
- update parser to support sql rewritten by postgres dialect (FETCH n IN CURSOR, show <var>)
- use `patch.crate-io` to include our datafusion fork, so it overrides all datafusion 49 in our dependency tree.

src/catalog/src/system_schema/pg_catalog.rs
```diff
@@ -69,30 +56,31 @@ impl SystemSchemaProvider for PGCatalogProvider {
...
+        let static_tables =
+            PgCatalogStaticTables::try_new().expect("Failed to initialize static tables");
+        let inner = PgCatalogSchemaProvider::try_new(
+            CatalogManagerWrapper {
+                catalog_name: catalog_name.clone(),
+                catalog_manager,
+            },
+            Arc::new(static_tables),
+        )
+        .expect("Failed to initialize PgCatalogSchemaProvider");
```
MichaelScofield, line 73: Are these two expects safe?

src/catalog/src/system_schema/pg_catalog.rs
```diff
@@ -129,24 +107,23 @@ impl SystemSchemaProviderInner for PGCatalogProvider {
...
+            let table = self.inner.build_table_by_name(name).expect(name);
+
+            if let Some(table) = table {
+                let system_table = DFTableProviderAsSystemTable::try_new(
+                    *table_id,
+                    table_name,
+                    table::metadata::TableType::Temporary,
+                    table,
+                )
+                .expect(name);
```
MichaelScofield: ditto

src/catalog/src/system_schema/pg_catalog.rs
sunng87: Yes, these errors are mostly loading schema from data file that bundled into the crate, which is a static behavior that can be verified by unit tests. 

But I will check again if we can use `try_new` here.

sunng87: Checked its usage, it doesn't allow us to return error. I think it's ok and safe for this.

src/catalog/src/system_schema/pg_catalog.rs
sunng87: Like above, the only potential error is for transforming schema.

But since this API is called at runtime, I can return `None` if it's failed to load, for the worse case.

sunng87:
image

yihong0618:
> image

amazing!

src/catalog/src/system_schema/pg_catalog.rs
MichaelScofield: better add to safety comment

MichaelScofield (approved):

src/catalog/src/system_schema/pg_catalog.rs
sunng87: done

waynexia (approved):

Issue #6756, opened 2025-08-18
Tracking Issue for async index build


SNC123:
### What problem does the new feature solve?

Previously, index construction only supported synchronous operations. This update introduces support for asynchronous index building.

### What does the feature do?

Introduces a framework for building SST file indexes asynchronously in the background to improve the performance and stability of foreground operations like flushing and compaction.

### Tasks

 - [x] Implement the basic framework for asynchronous index rebuilding.
  #6927 
 - [x] Implement construction and testing for four types of indexes.
 #7029 
 #7104
 - [x] Performance optimization
 #7034

killme2008:
Looks like we can close this issue right now.

Great work! Thanks a lot @SNC123

evenyag:
We still have some remaining works to do.

PR #6927, opened 2025-09-08
feat: introduce IndexBuildTask for async index build

SNC123:

https://github.com/GreptimeTeam/greptimedb/issues/6756

As title. 🐭  

Introduce some components:

- **(core)** IndexBuildTask: A task to build an index for a single SST file.
- IndexBuildMode: An enum (Sync, Async) to control the index building behavior.
- RegionBuildIndexRequest: A request dispatched to the region worker to trigger an index build.

## PR Checklist


src/mito2/src/test_util/scheduler_util.rs
```diff
@@ -51,7 +52,7 @@ pub(crate) struct SchedulerEnv {
 impl SchedulerEnv {
     /// Creates a new mocker.
     pub(crate) async fn new() -> SchedulerEnv {
-        let path = create_temp_dir("");
+        let path = create_temp_dir("114514");
```
zhongzc: 😨

SNC123: Sorry, this was a temporary addition during debugging that I forgot to revert. 
I will fix it later.😭

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -0,0 +1,205 @@
...
+        for file_handle in build_tasks {
+            let _ = self
+                .index_build_scheduler
+                .schedule_build(self.new_index_build_task(
+                    &region,
+                    file_handle,
+                    request.build_type.clone(),
+                    None,
+                ));
+        }
```
zhongzc, line 116: A prominent note is needed: Currently, rebuilding the index will reconstruct the index for all files in the region, which is a simplified approach and is not yet available for production use; further optimization is required.

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -0,0 +1,205 @@
...
+            write_cache.build_puffin_manager()
+        } else {
+            access_layer.build_puffin_manager()
+        };
+
+        let intermediate_manager = if let Some(write_cache) = self.cache_manager.write_cache() {
+            write_cache.intermediate_manager().clone()
+        } else {
+            access_layer.intermediate_manager().clone()
+        };
```
zhongzc, line 52: If I remember correctly, are these two intermediate_manager instances the same?

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        let indexer_builder_ref = Arc::new(IndexerBuilderImpl {
+            op_type: OperationType::Flush, //TODO(SNC123): Temporarily set to Flush, 4 BuildTypes will be introduced later.
+            metadata: version.metadata.clone(),
```
zhongzc: Perhaps adding a new OperationType::Async would be more reasonable, of course, along with the corresponding feature gate configuration item.

zhongzc (commented):
Almost LGTM

killme2008:
@SNC123 Thanks

There are some conflicts have to be resolved.

src/mito2/src/sst/index.rs
```diff
@@ -384,11 +396,208 @@ impl IndexerBuilderImpl {
...
+            )
+            .build_index_file_path(RegionFileId::new(region_id, file_id));
+            if let Err(e) = write_cache
+                .upload(puffin_key, &puffin_path, remote_store)
+                .await
+            {
+                err = Some(e);
+            }
+            upload_tracker.push_uploaded_file(puffin_path);
+            if err.is_some() {
```
fengjiachun: Maybe a warning log could be added here, or should we return the error? cc @zhongzc

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -0,0 +1,205 @@
...
+
+            // Check if the file exists in the object store.
+            let path = location::sst_file_path(
+                region.access_layer.table_dir(),
+                RegionFileId::new(file_meta.region_id, file_id),
+                PathType::Bare,
+            );
+            region.access_layer.object_store().exists(&path).await.map_err(|e| {
+                warn!(e; "SST file not found for index build, region: {}, file_id: {}", region_id, file_meta.file_id);
+            }).ok();
```
fengjiachun: Question: Why does it say "OK" when the file is missing? Is it correct to proceed with updating the manifest?

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        // Safety: The corresponding file exists.
```
fengjiachun: But the inspection code above indicates that the file may not exist.

src/mito2/src/worker/handle_rebuild_index.rs
SNC123: I rechecked the initialization of  [write_cache source](https://github.com/GreptimeTeam/greptimedb/blob/0717773f625abd35e3a75b38c49a9dd7d737e225/src/mito2/src/worker.rs#L173) and [access_layer source](https://github.com/GreptimeTeam/greptimedb/blob/0717773f625abd35e3a75b38c49a9dd7d737e225/src/mito2/src/worker.rs#L206) in `worker.rs`. Their sources are nearly identical in most cases, except for [compactor](https://github.com/GreptimeTeam/greptimedb/blob/0717773f625abd35e3a75b38c49a9dd7d737e225/src/mito2/src/compaction/compactor.rs#L149). Although both use `mito.config.index.aux_path` to initialize the filesystem, they are two distinct instances. Therefore, handling `intermediate_manager` separately seems more reasonable.

src/mito2/src/worker/handle_rebuild_index.rs
SNC123: I find `OperationType` mainly used for determining type of `IndexerBuilderImpl` (from SstWriteRequest). Replacing `OperationType` with `IndexBuildType` in `IndexerBuilderImpl` and adding corresponding feature gate configuration will be more reasonable?

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -0,0 +1,205 @@
...
+        let version_result = region
+            .manifest_ctx
+            .update_manifest(
+                RegionLeaderState::Writable,
+                RegionMetaActionList::with_action(RegionMetaAction::Edit(request.edit.clone())),
+            )
+            .await;
```
evenyag: We can update the manifest in the background task instead of in the worker loop. It avoids blocking the worker, so the worker only has to apply the region edit.

Currently, flush and compaction all use this pattern.
https://github.com/GreptimeTeam/greptimedb/blob/8bcf4a8ab518c0a4e283e18b110586bbfefdc6a7/src/mito2/src/flush.rs#L413-L423

cc @zhongzc

src/mito2/src/worker/handle_rebuild_index.rs
SNC123: Good catch. I forgot to handle this. Thanks!

src/mito2/src/sst/index.rs
SNC123: Thanks! Perhaps error better. Updated.

evenyag:
There are some conflicts. @SNC123

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -0,0 +1,225 @@
...
+            match version_result {
+                Ok(version) => {
+                    debug!(
+                        "Successfully update manifest version to {version}, region: {region_id}, reason : index build"
+                    );
+                    region.version_control.apply_edit(
+                        Some(request.edit.clone()),
+                        &[],
+                        region.file_purger.clone(),
+                    );
```
evenyag: We have to modify the `VersionControl` in the worker loop.
https://github.com/GreptimeTeam/greptimedb/blob/d777e8c52f59c9d3084abaa9119c9d3e814f7259/src/mito2/src/region.rs#L98-L102

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -0,0 +1,225 @@
...
+                region.access_layer.path_type(),
+            );
+            match region.access_layer.object_store().exists(&path).await {
+                Ok(found_in_object_store) => {
+                    if !found_in_object_store {
+                        warn!(
+                            "SST file not found for index build, region: {}, file_id: {}",
+                            region_id, file_meta.file_id
+                        );
+                        return;
```
evenyag: We should do most of these jobs in the IndexBuildTask, including updating the manifest. So `handle_index_build_finished` only has to apply the edit.

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -0,0 +1,225 @@
...
+            // Check if the file exists in the object store.
+            let path = location::sst_file_path(
+                region.access_layer.table_dir(),
+                RegionFileId::new(file_meta.region_id, file_id),
+                region.access_layer.path_type(),
+            );
+            match region.access_layer.object_store().exists(&path).await {
+                Ok(found_in_object_store) => {
+                    if !found_in_object_store {
+                        warn!(
```
evenyag: This is expensive to do in the worker loop and we should only do it in IndexBuildTask.

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -0,0 +1,225 @@
...
+            .flat_map(|level| level.files.iter())
+            .filter(|(_, handle)| !handle.is_deleted())
+            .map(|(id, handle)| (*id, handle.clone()))
+            .collect();
+
+        let build_tasks = if request.file_metas.is_empty() {
+            // NOTE: Currently, rebuilding the index will reconstruct the index for all
+            // files in the region, which is a simplified approach and is not yet available for
+            // production use; further optimization is required.
+            all_files.values().cloned().collect::<Vec<_>>()
```
evenyag, line 101: Another risk is that the write cache and object store read cache may have cached the old files/pages. If we read the file, we may get "corrupted result", especially we read an old page from the read cache.

src/mito2/src/sst/index.rs
```diff
@@ -384,11 +399,221 @@ impl IndexerBuilderImpl {
...
+        let index_output = indexer.finish().await;
+
+        if index_output.file_size > 0 {
+            // Upload index file if write cache is enabled.
+            self.maybe_upload_index_file(index_output.clone()).await?;
+
+            // Check SST file existence again after building index.
+            if !self.check_sst_file_exists().await {
+                return Ok(IndexBuildOutcome::Aborted("SST file not found".to_string()));
+            }
```
evenyag, line 548: We should clean the index file in this case.

src/mito2/src/sst/index.rs
```diff
@@ -384,11 +399,221 @@ impl IndexerBuilderImpl {
...
+            let region_id = self.file_meta.region_id;
+            let remote_store = self.access_layer.object_store();
+            let mut upload_tracker = UploadTracker::new(region_id);
+            let mut err = None;
+            let puffin_key = IndexKey::new(region_id, file_id, FileType::Puffin);
+            let puffin_path = RegionFilePathFactory::new(
+                self.access_layer.table_dir().to_string(),
+                self.access_layer.path_type(),
+            )
+            .build_index_file_path(RegionFileId::new(region_id, file_id));
```
evenyag, line 593: If write cache is enabled, do we need to update the write cache?

src/mito2/src/sst/index.rs
```diff
@@ -384,11 +399,221 @@ impl IndexerBuilderImpl {
...
+                    "Failed to check SST file existence during index build, skipping. region: {}, file_id: {}, error: {:?}",
+                    self.file_meta.region_id, self.file_meta.file_id, e
+                );
+                false
+            }
+        }
+    }
+
+    async fn index_build(&mut self) -> Result<IndexBuildOutcome> {
+        let mut indexer = self.indexer_builder.build(self.file_meta.file_id).await;
```
evenyag, line 514: We assume puffin/parquet files in the engine are immutable. It'd be better to assign a new id to the index if the file already has an index. We can consider it after this PR.

src/mito2/src/sst/index.rs
SNC123: Yes. This is because in `Async` build mode, the following code is not triggered (index_metadata.file_size is 0):

https://github.com/GreptimeTeam/greptimedb/blob/d777e8c52f59c9d3084abaa9119c9d3e814f7259/src/mito2/src/cache/write_cache.rs#L223-L235

Therefore, an additional update is required within IndexBuildTask.

src/mito2/src/sst/index.rs
```diff
@@ -384,11 +404,290 @@ impl IndexerBuilderImpl {
...
+            return false;
+        }
+
+        let sst_path = location::sst_file_path(
+            self.access_layer.table_dir(),
+            RegionFileId::new(self.file_meta.region_id, self.file_meta.file_id),
+            self.access_layer.path_type(),
+        );
+
+        match self.access_layer.object_store().exists(&sst_path).await {
```
evenyag: If we can find the file in the version, then it should exist in the object store. I think we don't have to check the object store to avoid an API call.

src/mito2/src/sst/index.rs
```diff
@@ -384,11 +404,290 @@ impl IndexerBuilderImpl {
...
+            indexer.update(batch).await;
+        }
+        let index_output = indexer.finish().await;
+
+        if index_output.file_size > 0 {
+            // Check SST file existence again after building index.
+            if !self.check_sst_file_exists(version).await {
+                // Calls abort to clean up index files.
+                indexer.abort().await;
+                return Ok(IndexBuildOutcome::Aborted("SST file not found".to_string()));
```
evenyag: Add the file id to the message.

src/mito2/src/sst/index.rs
```diff
@@ -384,11 +404,290 @@ impl IndexerBuilderImpl {
...
+    Aborted(String),
+}
+
+pub struct IndexBuildTask {
+    /// The file meta to build index for.
+    pub file_meta: FileMeta,
+    pub reason: IndexBuildType,
+    pub flushed_entry_id: Option<EntryId>,
+    pub flushed_sequence: Option<SequenceNumber>,
+    pub committed_sequence: Option<SequenceNumber>,
```
evenyag: Updating the index doesn't change entry id and sequence, we can always set them to None.

src/mito2/src/sst/index.rs
```diff
@@ -384,11 +404,290 @@ impl IndexerBuilderImpl {
...
+            .access_layer
+            .read_sst(FileHandle::new(
+                self.file_meta.clone(),
+                self.file_purger.clone(),
+            ))
+            .build()
+            .await?;
+
+        // TODO(SNC123): optimize index batch
+        while let Some(batch) = &mut parquet_reader.next_batch().await? {
```
evenyag: If next_batch returns an err, we won't abort the indexer.

src/mito2/src/sst/index.rs
```diff
@@ -384,11 +404,290 @@ impl IndexerBuilderImpl {
...
+                        notify: BackgroundNotify::IndexBuildFinished(index_build_finished),
+                    }
+                }
+                Err(e) => {
+                    let err = Arc::new(e);
+                    WorkerRequest::Background {
+                        region_id: self.file_meta.region_id,
+                        notify: BackgroundNotify::IndexBuildFailed(IndexBuildFailed { err }),
+                    }
+                }
```
evenyag, line 570: When should we send `BackgroundNotify::IndexBuildFailed` instead of returning `IndexBuildOutcome`?

src/mito2/src/sst/index.rs
```diff
@@ -384,11 +404,290 @@ impl IndexerBuilderImpl {
...
+            OperationType::Compact => IndexBuildType::Compact,
+        }
+    }
+}
+
+/// Outcome of an index build task.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum IndexBuildOutcome {
+    Finished,
+    Aborted(String),
```
evenyag, line 437: Why don't we keep an error in this variant?

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -0,0 +1,151 @@
...
+        );
+    }
+
+    pub(crate) async fn handle_index_build_failed(
+        &mut self,
+        region_id: RegionId,
+        request: IndexBuildFailed,
+    ) {
+        error!(request.err; "Index build failed for region: {}", region_id);
+    }
```
evenyag, line 148: The IndexBuildFailed notification seems unnecessary.

SNC123: Thanks for the great questions! I'll address all three points here.

> **Why don't we keep an error in this variant?**

I intentionally unify various failure scenarios into the `Aborted` variant instead of including a detailed `Error` type. There are two main reasons for this design:

1.  `IndexBuildOutcome` is designed as a simple receipt for the **caller**. The caller's primary concern is whether the task succeeded, not the specific internal details of the failure.
2.  Critical error information, which requires specific handling, is passed to the **region worker** via the `IndexBuildFailed` notification. This ensures that the component responsible for region state handles the error, rather than burdening the original caller with it.

> **When should we send `BackgroundNotify::IndexBuildFailed` instead of returning `IndexBuildOutcome`?**

We send `BackgroundNotify::IndexBuildFailed` when an error is fatal to the region's core state management or has a significant performance impact, suggesting a systemic issue.

A  example is when updating the manifest fails. This is a critical failure because it affects the region's metadata integrity. Such an error might indicate that other `IndexBuildTask` instances for the same region should also be halted.

> **The `IndexBuildFailed` notification seems unnecessary.**

For now, I'd like to keep this notification, as it's intended for more robust error handling that will be implemented as part of the `IndexBuildScheduler` enhancement coming in a future PR.

src/mito2/src/worker/handle_rebuild_index.rs
SNC123: Right. Maybe we can consider it in the next PR?

src/mito2/src/worker/handle_rebuild_index.rs
evenyag: If we have things to do in the worker loop then we might need the notification. My concern is that the IndexBuildOutcome may not provide enough information to debug a failure without an error log or other details. But we can discuss this later.

PR #6928, opened 2025-09-08
docs: laminar flow rfc

discord9:

as title, [rendered](https://github.com/GreptimeTeam/greptimedb/blob/luminar_flow_rfc/docs/rfcs/2025-09-08-luminar-flow.md)
redesign flow streaming mode to be more consistence and cleaner 

## PR Checklist

### DIFF
--- a/docs/rfcs/2025-09-08-laminar-flow.md
+++ b/docs/rfcs/2025-09-08-laminar-flow.md
@@ -1,11 +1,11 @@
 ---
-Feature Name: "luminar-flow"
+Feature Name: "laminar-flow"
 Tracking Issue: https://github.com/GreptimeTeam/greptimedb/issues/TBD
 Date: 2025-09-08
 Author: "discord9 <discord9@163.com>"
 ---
 
-# Luminar Flow
+# laminar Flow
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


docs/rfcs/2025-09-08-luminar-flow.md
```diff
@@ -0,0 +1,359 @@
...
+
+### Phase 3: Advanced Features
+
+1. **Load Balancing**: Implement intelligent frontend selection
+2. **Fault Tolerance**: Add retry mechanisms and error handling
+3. **Performance Tuning**: Optimize query batching and state management
+
+## Drawbacks
+
+### Increased Network Communication
```
killme2008: I have a proposal: 
could we, to some extent, merge the flownode and the frontend? In other words, the flownode would simply be an extended frontend that adds the flow distributed coordination capabilities introduced in this RFC, plus a bit of cache. 

This is similar to how we treat the datanode, which can play multiple roles such as compactor and indexer. In this way, we can largely address these drawbacks, and the frontend itself already supports grouping and horizontal scaling.

killme2008: * Frontend becomes multi-role, but we still logically abstract a flownode subsystem.
* Resource isolation is achieved through grouping.

discord9: > * Frontend becomes multi-role, but we still logically abstract a flownode subsystem.
> 
>     * Resource isolation is achieved through grouping.

currently flownode can remove this hop by using it's embedded frontend, but in the future flownode can become a special frontend for better load balancing and resource usage etc.

docs/rfcs/2025-09-08-luminar-flow.md
```diff
@@ -0,0 +1,351 @@
...
+2. **Historical Data (Batch Mode)**: For older time windows, flownode triggers batch computation directly and no need to refill state
+3. **Hybrid Approach**: Combines stream and batch processing based on data age and availability
+
+#### Mirror Write Optimization
+
+Mirror writes are simplified to only transmit timestamps to flownode:
+
+```rust
+struct MirrorWrite {
+    timestamp: Timestamp,
```
fengjiachun: `timestamps[]` or `time windows[]` ?

PR #6930, opened 2025-09-08
feat(pipeline): generate create table sql from pipeline config

shuiyisong:

This PR adds
1. `expr_to_create`: create `CreateTable` using `CreateTableExpr`, which is the opposite of `create_to_expr`
2. `/v1/pipelines/{pipeline_name}/create_table`: generate create table SQL from pipeline's transform config

## PR Checklist

### DIFF
--- a/src/servers/src/http.rs
+++ b/src/servers/src/http.rs
@@ -1028,6 +1028,10 @@ impl HttpServer {
                 routing::get(event::query_pipeline),
             )
             .route(
+                "/pipelines/{pipeline_name}/create_table",
+                routing::get(event::query_pipeline_create_table),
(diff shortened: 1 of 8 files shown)


src/servers/src/http.rs
```diff
@@ -1027,6 +1027,10 @@ impl HttpServer {
                 "/pipelines/{pipeline_name}",
                 routing::get(event::query_pipeline),
             )
+            .route(
+                "/pipelines/{pipeline_name}/create_table",
```
sunng87: I think `ddl` can be a better name to describe the behavior, this `create_table` feels like an action.

PR #6932, opened 2025-09-08
feat!: improve `greptime_identity` pipeline behavior

waynexia:

- remove `flatten_json_object` option. now flat unconditionally
- expose `max_nested_levels` option to control how many levels can exist
- don't return error when maxium nested levels reached. instead, store remaining parts as string
- use string to store array type

## PR Checklist


src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -65,31 +66,32 @@ pub struct GreptimePipelineParams {
     /// This should not be used directly, instead, use the parsed shortcut option values.
     options: HashMap<String, String>,
 
-    /// Parsed shortcut option values
-    pub flatten_json_object: OnceCell<bool>,
     /// Whether to skip error when processing the pipeline.
     pub skip_error: OnceCell<bool>,
+    /// Max nested levels when flattening JSON object. Defaults to
+    /// `DEFAULT_MAX_NESTED_LEVELS_FOR_JSON_FLATTENING` when not provided.
+    pub max_nested_levels: OnceCell<usize>,
```
shuiyisong, line 73: How about `max_flatten_levels`

src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -659,32 +658,52 @@ pub fn identity_pipeline(
...
     if !object.is_empty() {
         // it will use recursion to flatten the object.
-        do_flatten_object(&mut flattened, None, object, 1, max_nested_levels)?;
+        do_flatten_object(&mut flattened, None, object, 1, max_nested_levels);
     }
 
     Ok(VrlValue::Object(flattened))
 }
 
+fn vrl_value_to_serde_json(value: &VrlValue) -> serde_json_crate::Value {
```
shuiyisong, line 676:
```suggestion
fn vrl_value_to_serde_json_value(value: &VrlValue) -> serde_json_crate::Value {
```

shuiyisong (commented):
This is a big breaking change to the existing identity pipeline users. Should we make it default cc @sunng87

waynexia:
Rationale of these changes: (1) flat but only flat one layer looks weird and inconsistent (2) JSON type is in fact deprecated (3) limiting the max depth can achieve a similar functionality of "whether to flat"

sunng87:
@shuiyisong the goal is the deprecate json type as soon as possible from our built-in features. the breaking change is acceptable.

sunng87 (approved):

killme2008:
> @shuiyisong the goal is the deprecate json type as soon as possible from our built-in features. the breaking change is acceptable.

Should we consider naming this version `greptime_identity_v2`? There are already quite a few users using the current version—how significant would the impact be on them if we introduce breaking changes? @sunng87 @shuiyisong

sunng87:
@killme2008 The goal is to move user from current json type so I think it's ok to break it as early as possible. We need to emphasize this breaking change in release notes.

PR #6938, opened 2025-09-09
feat: support flat format for SeriesScan

evenyag:

This PR adds flat format support for the SeriesScan. Now all scanners can use the flat format.

## PR Checklist


src/mito2/src/read/series_scan.rs
```diff
@@ -627,3 +758,332 @@ fn new_partition_metrics(
...
+        let batch_pk_keys = batch_pk_array.keys();
+        let pk_indices = batch_pk_keys.values();
+        let mut change_offset = 0;
+        for (i, &key) in pk_indices.iter().enumerate() {
+            let batch_pk = batch_pk_values.value(key as usize);
+
+            if buffer_last_pk != batch_pk {
+                change_offset = i;
+                break;
+            }
```
fengjiachun, line 833: If all the primary keys are the same, should we expect the `change_offset` to be 0 as well?

evenyag: We have checked the last pk is not the same as the buffer pk in L814, so this is impossible.

fengjiachun: is this right?
```
let change_offset = pk_indices.iter()
      .enumerate()
      .find(|(_, &key)| {
          let batch_pk = batch_pk_values.value(key as usize);
          buffer_last_pk != batch_pk
      })
      .map(|(i, _)| i);
```

fengjiachun: got it

src/mito2/src/read/series_scan.rs
```diff
@@ -377,11 +381,100 @@ struct SeriesDistributor {
...
+            self.semaphore.clone(),
+        )
+        .await?;
+        let mut metrics = SeriesDistributorMetrics::default();
+        let mut fetch_start = Instant::now();
+
+        let mut divider = FlatSeriesBatchDivider::default();
+        while let Some(record_batch) = reader.try_next().await? {
+            metrics.scan_cost += fetch_start.elapsed();
+            fetch_start = Instant::now();
```
v0y4g3r: Updating `fetch_start` here may cause the metric to be inaccurate

src/mito2/src/read/series_scan.rs
```diff
@@ -627,3 +758,332 @@ fn new_partition_metrics(
...
+/// However, a [FlatSeriesBatch] may contain rows from multiple series.
+#[derive(Default)]
+struct FlatSeriesBatchDivider {
+    buffer: FlatSeriesBatch,
+}
+
+impl FlatSeriesBatchDivider {
+    /// Pushes a record batch into the divider.
+    ///
+    /// Returns a [FlatSeriesBatch] if the series in the buffer is exhausted.
```
v0y4g3r:
```suggestion
    /// Returns a [FlatSeriesBatch] if the series in the buffer is not exhausted.
```

evenyag: This may be wrong, we only return the batch when we ensure the batch contains all rows of the series.

evenyag: https://github.com/GreptimeTeam/greptimedb/pull/6938/commits/22158cb98340cd6857d4c7847ae0ec060692400f

src/mito2/src/read/series_scan.rs
evenyag: https://github.com/GreptimeTeam/greptimedb/pull/6938/commits/bd445fdfe6eeca728ce0b863ab97913d2bd03183

PR #6940, opened 2025-09-09
feat: datanode side local gc worker

discord9:

## PR Checklist


src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+            set.extend(files.file_ids.iter().cloned());
+        }
+
+        Ok(ret)
+    }
+
+    /// Concurrently list unused files in the region dir
+    /// because there may be a lot of files in the region dir
+    /// and listing them may take a long time.
+    pub async fn list_to_be_deleted_files(
```
fengjiachun: This method so hurge, can we split it to 2~3 methods， such as: `partition_region_files`, `list_region_files_concurrent` and `filter_deletable_files` ?

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+    pub lingering_time: Duration,
+    /// Lingering time before deleting unknown files(files with undetermine expel time).
+    /// expel time is the time when the file is considered as removed, as in removed from the manifest.
+    /// This should only occur rarely, as manifest keep tracks in `removed_files` field
+    /// unless something goes wrong.
+    #[serde(with = "humantime_serde")]
+    pub unknown_file_lingering_time: Duration,
+    /// Maximum concurrent list operations per GC job.
+    /// This is used to limit the number of concurrent listing operations and speed up listing.
+    pub max_concurrent_per_gc_job: usize,
```
evenyag: Do we have a limit for the overall GC jobs in a node?

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+}
+
+pub struct LocalGcWorker {
+    pub(crate) table_id: TableId,
+    pub(crate) access_layer: AccessLayerRef,
+    pub(crate) cache_manager: Option<CacheManagerRef>,
+    pub(crate) manifest_mgrs: HashMap<RegionId, RegionManifestManager>,
+    /// Lingering time before deleting files.
+    pub(crate) opt: FileGcOption,
+    pub(crate) mito_config: MitoConfig,
```
evenyag: non-blocking: Maybe we should only keep the necessary options instead of the whole mito config.

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+    pub(crate) file_ref_manifest: TableFileRefsManifest,
```
evenyag, line 101: nitpick: Can we document this field? I forgot what TableFileRefsManifest is while reviewing this. Maybe a document will be helpful for other readers.

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+                UnexpectedSnafu {
+                    reason: format!(
+                        "All regions should belong to the same table, found region {} and table {}",
+                        region_id, table_id
+                    ),
+                }
+            );
+            let mgr = zelf.open_mgr_for(region_id).await?;
+            zelf.manifest_mgrs.insert(region_id, mgr);
+        }
```
evenyag, line 164: Can we just take a HashSet as input?

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+    ///
+    /// Note that the files that are still in use or may still be kept for a while are not deleted
+    /// to avoid deleting files that are still needed.
+    pub async fn do_region_gc(
+        &self,
+        region_id: RegionId,
+        tmp_ref_files: &HashSet<FileId>,
+    ) -> Result<Vec<FileId>> {
+        info!("Doing gc for region {}", region_id);
+        // TODO(discord9): impl gc worker
```
evenyag: Should we remove the todo?

src/mito2/src/gc.rs
discord9: > Do we have a limit for the overall GC jobs in a node?

not yet, ~should set to how many?~ was thinking about let metasrv do this limit

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+            info!("Doing gc for region {}", region_id);
+            let tmp_ref_files = tmp_ref_files
+                .get(region_id)
+                .cloned()
+                .unwrap_or_else(HashSet::new);
+            let files = self.do_region_gc(*region_id, &tmp_ref_files).await?;
+            deleted_files.insert(*region_id, files);
+            info!("Gc for region {} finished", region_id);
+        }
+        info!("LocalGcWorker finished");
```
evenyag: nitpick: log the gc cost.

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+        let true_tmp_ref_files = tmp_ref_files
+            .iter()
+            .filter(|f| !current_files.contains_key(f))
+            .collect::<HashSet<_>>();
+
+        info!("True tmp ref files: {:?}", true_tmp_ref_files);
+
+        GC_TMP_REF_FILE_CNT_PER_TABLE
+            .with_label_values(&[&self.table_id.to_string()])
+            .set(true_tmp_ref_files.len() as i64);
```
evenyag: This may generate too many time series.

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+
+        info!(
+            "Successfully deleted {} unused files for region {}",
+            unused_len, region_id
+        );
+
+        Ok(unused_files)
+    }
+
+    async fn delete_files(&self, region_id: RegionId, file_ids: &[FileId]) -> Result<()> {
```
evenyag, line 322: This method has some duplicate code with `LocalFilePurger::delete_file`. Can we wrap the common logic into a function?

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+            .filter_map(|(ts, files)| {
+                if *ts < Timestamp::new_millisecond(may_linger_until.timestamp_millis()) {
+                    // if the expel time is before the may linger time, we can delete it
+                    return None;
+                }
+
+                Some(files)
+            })
+            .flatten()
+            .collect::<HashSet<_>>();
```
evenyag: Looks like we didn't use the sorted properties of the map. Do we still need a btree map?

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+
+            listers.push((lister, end));
+        }
+
+        let (tx, mut rx) = tokio::sync::mpsc::channel(1024);
+        let mut handles = vec![];
+        for (lister, end) in listers {
+            let tx = tx.clone();
+            let handle = tokio::spawn(async move {
+                let stream = lister.take_while(|e: &std::result::Result<Entry, _>| match e {
```
evenyag, line 436: Does the lister ensure the returned entries are sorted?

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,660 @@
...
+                .await
+            {
+                error!(e; "Failed to purge stager with index file, file_id: {}, region: {}",
+                    file_id, region_id);
+            }
+        }
+
+        GC_FILE_CNT_PER_TABLE
+            .with_label_values(&[&self.table_id.to_string()])
+            .add(file_ids.len() as i64);
```
evenyag: It may have too many series.

src/mito2/src/gc.rs
discord9: > Does the lister ensure the returned entries are sorted?

"List results are always returned in UTF-8 binary order." https://docs.aws.amazon.com/AmazonS3/latest/userguide/ListingKeysUsingAPIs.html

src/mito2/src/gc.rs
```diff
@@ -0,0 +1,657 @@
...
+            manifest_dir: new_manifest_dir(&region_dir_from_table_dir(
+                table_dir, region_id, path_type,
+            )),
+            object_store: self.access_layer.object_store().clone(),
+            compress_type: manifest_compress_type(mito_config.compress_manifest),
+            checkpoint_distance: mito_config.manifest_checkpoint_distance,
+            remove_file_options: RemoveFileOptions {
+                keep_count: mito_config.experimental_manifest_keep_removed_file_count,
+                keep_ttl: mito_config.experimental_manifest_keep_removed_file_ttl,
+            },
```
evenyag: Looks like we only use a few fields of the MitoConfig

evenyag (approved):

src/mito2/src/gc.rs
```diff
@@ -124,7 +144,7 @@ impl LocalGcWorker {
             cache_manager,
             manifest_mgrs: HashMap::new(),
             opt,
-            mito_config,
+            manifest_open_config: mito_config.into(),
```
evenyag: The `try_new()` can take `manifest_open_config` instead of `mito_config` as the argument.

PR #6947, opened 2025-09-10
chore: bump sequence on region edit

v0y4g3r:

This PR allows region worker to bump the region sequence while handling region edit requests. That sequence will also be assigned to the files to add in the `RegionEdit` request so that each ingested SST files will have a overriding sequence number assigned by the region worker.

## PR Checklist

### DIFF
--- a/src/mito2/src/engine/bump_committed_sequence_test.rs
+++ b/src/mito2/src/engine/bump_committed_sequence_test.rs
@@ -15,7 +15,7 @@
 use std::collections::HashMap;
 
 use api::v1::Rows;
-use store_api::region_engine::RegionEngine;
+use store_api::region_engine::{RegionEngine, RegionRole};
 use store_api::region_request::{PathType, RegionOpenRequest, RegionRequest};
 use store_api::storage::RegionId;
 
@@ -89,6 +89,37 @@ async fn test_bump_committed_sequence() {
             region_id,
             RegionRequest::Open(RegionOpenRequest {
                 engine: String::new(),
+                table_dir: table_dir.clone(),
+                path_type: PathType::Bare,
+                options: HashMap::default(),
+                skip_wal_replay: false,
+                checkpoint: None,
+            }),
+        )
+        .await
+        .unwrap();
+    let region = engine.get_region(region_id).unwrap();
+    region.set_role(RegionRole::Leader);
+    assert_eq!(region.version_control.current().version.flushed_sequence, 0);
+    assert_eq!(region.version_control.committed_sequence(), 43);
+
+    // Write another 2 rows after editing.
+    let column_schemas = rows_schema(&request);
+    let rows = Rows {
+        schema: column_schemas,
+        rows: build_rows(0, 2),
+    };
+    put_rows(&engine, region_id, rows).await;
+    assert_eq!(region.version_control.committed_sequence(), 45);
+    assert_eq!(region.version_control.current().version.flushed_sequence, 0);
+
+    // Reopen region.
+    let engine = env.reopen_engine(engine, MitoConfig::default()).await;
+    engine
+        .handle_request(
+            region_id,
+            RegionRequest::Open(RegionOpenRequest {
+                engine: String::new(),
                 table_dir,
                 path_type: PathType::Bare,
                 options: HashMap::default(),
@@ -99,9 +130,7 @@ async fn test_bump_committed_sequence() {
         .await
         .unwrap();
     let region = engine.get_region(region_id).unwrap();
+    region.set_role(RegionRole::Leader);
     assert_eq!(region.version_control.current().version.flushed_sequence, 0);
-    // After reopen, firstly committed_sequence will be set to 43 because during manifest replay,
-    // we assign the committed_sequence of RegionEdit to region manifest. And the WAL replays to
-    // memtable and another 42 rows are written to memtable. So the committed sequence will be 85
-    assert_eq!(region.version_control.committed_sequence(), 85);
+    assert_eq!(region.version_control.committed_sequence(), 45);
 }
--- a/src/mito2/src/engine/bump_committed_sequence_test.rs
+++ b/src/mito2/src/engine/bump_committed_sequence_test.rs
@@ -14,6 +14,7 @@
 
 use std::collections::HashMap;
 
+use api::v1::Rows;
 use store_api::region_engine::RegionEngine;
 use store_api::region_request::{PathType, RegionOpenRequest, RegionRequest};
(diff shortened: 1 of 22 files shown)


src/mito2/src/worker/handle_manifest.rs
```diff
@@ -203,9 +204,16 @@ impl<S> RegionWorkerLoop<S> {
...
+            mut edit,
             tx: sender,
         } = request;
+        let file_sequence = region.version().flushed_sequence + 1;
+        edit.flushed_sequence = Some(file_sequence);
+
+        // For every file added through region edit, we should fill the file sequence
+        for file in &mut edit.files_to_add {
+            file.sequence = NonZeroU64::new(file_sequence);
+        }
```
evenyag: Is it possible that the flushed_sequence is greater than the committed sequence if we bump it too many times?

v0y4g3r: > Is it possible that the flushed_sequence is greater than the committed sequence if we bump it too many times?

You are right, we need to assign the `committed_sequence+1` to files instead of `flushed_sequence`. I added a `committed_sequence` to `RegionEdit` so that we can bump the committed sequence once `edit_region` succeeds. 8e3df8a1305ed2015fea8808f7382df02ee861ec

src/store-api/src/region_engine.rs
```diff
@@ -746,25 +744,11 @@ pub trait RegionEngine: Send + Sync {
...
-    ) -> Result<HashMap<u64, u64>, BoxedError> {
-        let mut results = HashMap::with_capacity(seqs.region_ids.len());
-
-        for region_id in seqs.region_ids {
-            let seq = self.get_last_seq_num(region_id).await?.unwrap_or_default();
-            results.insert(region_id.as_u64(), seq);
-        }
-
-        Ok(results)
-    }
```
evenyag, line 767: Some places may use this method. Not sure if this is a breaking change. @discord9

evenyag (approved):

src/store-api/src/region_engine.rs
discord9: > Some places may use this method. Not sure if this is a breaking change. @discord9

Can delete now, flow will use region sequences in different way(get latest sequence in `OutputMeta` of a query) and this is not currently in use

src/mito2/src/manifest/action.rs
```diff
@@ -120,6 +120,7 @@ pub struct RegionManifest {
     pub flushed_entry_id: EntryId,
     /// Last sequence of flushed data.
     pub flushed_sequence: SequenceNumber,
+    pub committed_sequence: Option<SequenceNumber>,
```
evenyag, line 123: Maybe need to add `#[serde(default)]` here. Do we have a JSON compatibility test for this struct?

src/mito2/src/region/opener.rs
```diff
@@ -473,6 +473,16 @@ impl RegionOpener {
...
         let version_control = Arc::new(VersionControl::new(version));
+        if let Some(committed_sequence) = manifest.committed_sequence {
+            debug!(
+                "Overriding committed sequence, region: {}, flushed_sequence: {}, committed_sequence: {} -> {}",
+                self.region_id,
+                version_control.current().version.flushed_sequence,
+                version_control.committed_sequence(),
+                committed_sequence
+            );
+            version_control.set_committed_sequence(committed_sequence);
```
evenyag: Will we override the sequence to a smaller one after replay?

src/mito2/src/engine/bump_committed_sequence_test.rs
```diff
@@ -0,0 +1,94 @@
...
+use store_api::region_request::{PathType, RegionOpenRequest, RegionRequest};
+use store_api::storage::RegionId;
+
+use crate::config::MitoConfig;
+use crate::manifest::action::RegionEdit;
+use crate::sst::file::FileMeta;
+use crate::test_util::{CreateRequestBuilder, TestEnv};
+
+#[tokio::test]
+async fn test_bump_committed_sequence() {
```
evenyag, line 28: Also need a case that writes some rows into the memtable, then edits the region and reopens it.

src/mito2/src/manifest/action.rs
v0y4g3r: https://github.com/GreptimeTeam/greptimedb/blob/c9c1b0aafe656837f7a2b56a7501361402f88170/src/mito2/src/manifest/action.rs#L450

src/mito2/src/region/opener.rs
v0y4g3r: No, wal replay will build a `RegionWriteCtx` from current committed_sequence (as set here):

https://github.com/GreptimeTeam/greptimedb/blob/c9c1b0aafe656837f7a2b56a7501361402f88170/src/mito2/src/region/opener.rs#L724-L730

https://github.com/GreptimeTeam/greptimedb/blob/c9377e7c5a59eb37fdfc1ef38cd48f611434c262/src/mito2/src/region_write_ctx.rs#L114-L134

and each write to memtable will increase the `committed_sequence`.

Let me further explain why we need to set the `committed_sequence` according to `RegionEdit` at first.

Suppose there're two writer, W1 writes to memtable, and W2 directly ingests SST files, the sequence they are assigend are as follow:

```
W1 seq: 1, 2, 3,                 5, 6
W2 seq:            4(ingest)
```

If we set `committed_sequence` after the replay: during the replay, the sequence will be `1, 2, 3, 4, 5`, while the `committed_sequence` in `RegionEdit` is 4, in that case the `committed_sequence` jumps backward.

But if we set `committed_sequence` before replay: the `committed_sequence` will start at 4, and the sequence during wal replay will become: `5, 6, 7, 8, 9`. Sequences `1, 2, 3` are completely skipped, but we can still guratee that no sequence duplication and no backward jumps.

src/mito2/src/engine/bump_committed_sequence_test.rs
v0y4g3r: c59c9a4a1

src/mito2/src/region/opener.rs
evenyag: We have already stored the sequence in the Mutation.

src/mito2/src/region/opener.rs
```diff
@@ -719,6 +735,7 @@ where
             None,
         );
         for mutation in entry.mutations {
+            info!("=== Replay mutation sequence: {}", mutation.sequence);
```
evenyag: Do we still need to keep this?

PR #6950, opened 2025-09-10
feat: memtable seq range read

discord9:

add a sequence range to scan request for scanning memtable within given sequence range, useful for seq read for append only table's partial update query

## PR Checklist

### DIFF
--- a/src/mito2/src/memtable/bulk.rs
+++ b/src/mito2/src/memtable/bulk.rs
@@ -1380,7 +1380,7 @@ mod tests {
 
         memtable.write_bulk(part).unwrap();
 
-        let predicate_group = PredicateGroup::new(&metadata, &[]);
+        let predicate_group = PredicateGroup::new(&metadata, &[]).unwrap();
         let sequence_filter = Some(SequenceRange::LtEq { max: 400 }); // Filters out rows with sequence > 400
         let ranges = memtable
             .ranges(None, predicate_group, sequence_filter, false)


src/store-api/src/storage/types.rs
```diff
@@ -14,6 +14,63 @@
...
+    To {
+        /// Inclusive upper bound
+        max: SequenceNumber,
+    },
+    Range {
+        /// Exclusive lower bound
+        min: SequenceNumber,
+        /// Inclusive upper bound
+        max: SequenceNumber,
+    },
```
evenyag: non-blocking: Maybe we can use Gt/LtEq? WDYT
```suggestion
    Gt {
        /// Exclusive lower bound
        min: SequenceNumber,
    },
    LtEq {
        /// Inclusive upper bound
        max: SequenceNumber,
    },
    GtLtEq {
        /// Exclusive lower bound
        min: SequenceNumber,
        /// Inclusive upper bound
        max: SequenceNumber,
    },
```

src/store-api/src/storage/requests.rs
```diff
@@ -58,6 +58,9 @@ pub struct ScanRequest {
     /// If set, only rows with a sequence number lesser or equal to this value
     /// will be returned.
     pub sequence: Option<SequenceNumber>,
+    /// Optional constraint on the minimal sequence number in the memtable.
+    /// If set, only the memtables that contain sequences greater than this value will be scanned
+    pub memtable_min_sequence: Option<SequenceNumber>,
```
evenyag: Is it exclusive?

src/store-api/src/storage/requests.rs
```diff
@@ -58,6 +58,9 @@ pub struct ScanRequest {
     /// If set, only rows with a sequence number lesser or equal to this value
     /// will be returned.
     pub sequence: Option<SequenceNumber>,
```
evenyag: Do we need to rename this to max_sequence?

discord9: > Do we need to rename this to max_sequence?

still useful for refilling which might still need to limit the sequence range and such

evenyag: Will renaming this field break this?

src/store-api/src/storage/requests.rs
discord9: > Is it exclusive?

"greater than this value will be scanned" so yes

src/store-api/src/storage/requests.rs
discord9: > Will renaming this field break this?

changed to `memtable_max_sequence` as it's just doing that

src/mito2/src/read.rs
```diff
@@ -361,17 +361,50 @@ impl Batch {
...
+                let max = UInt64Array::new_scalar(max);
+                datafusion_common::arrow::compute::kernels::cmp::lt_eq(seqs, &max)
+                    .context(ComputeArrowSnafu)?
+            }
+            SequenceRange::GtLtEq { min, max } => {
+                let min = UInt64Array::new_scalar(min);
+                let max = UInt64Array::new_scalar(max);
+                let pred_min = datafusion_common::arrow::compute::kernels::cmp::gt(seqs, &min)
+                    .context(ComputeArrowSnafu)?;
+                let pred_max = datafusion_common::arrow::compute::kernels::cmp::lt_eq(seqs, &max)
```
waynexia: Can we reuse `SequenceRange::filter` here to reduce duplication?

Issue #6873, opened 2025-09-01
Poor performance when using db aliases


paomian:
### What type of bug is this?

Performance issue

### What subsystems are affected?

Standalone mode

### Minimal reproduce step

```sql
SELECT count(`timestamp`) as t_cnt     FROM greptime_cloud.greptime_cloud_op_logs    WHERE `timestamp` > 1756722459558000000 and `timestamp` <= 1756726059558000000     LIMIT 50;
+---------+
| t_cnt   |
+---------+
| 1010740 |
+---------+
1 row in set (0.219 sec)

SELECT count(`tt`.`timestamp`) as t_cnt     FROM greptime_cloud.greptime_cloud_op_logs AS tt     WHERE `tt`.`timestamp` > 1756722459558000000 and `tt`.`timestamp` <= 1756726059558000000;
+---------+
| t_cnt   |
+---------+
| 1010740 |
+---------+
1 row in set (22.995 sec)
```

### What did you expect to see?

same performance

### What did you see instead?

poor performance when use db alias

### What operating system did you use?

mac m1 15.6 (24G84)

### What version of GreptimeDB did you use?

246b832d79d81479d56572194dfebfbc4c87cc47

### Relevant log output and stack trace

```bash

```

evenyag:
https://github.com/GreptimeTeam/greptimedb/pull/6963 should fix this. @paomian

Issue #6380, opened 2025-06-24
Filters are not pushed down when selecting from aliased table


MichaelScofield:
### What type of bug is this?

Performance issue

### What subsystems are affected?

Query Engine

### Minimal reproduce step

First create the table:

```sql
CREATE TABLE IF NOT EXISTS `bar` (
  `ts` TIMESTAMP(3) NOT NULL,
  `i8` TINYINT NULL,
  `ui8` TINYINT UNSIGNED NULL,
  TIME INDEX (`ts`),
  PRIMARY KEY (`i8`, `ui8`)
)
```

Then compare these two queries:

```sql
-- Q1:
 explain analyze SELECT i8 FROM bar WHERE ts <= '2025-06-23 09:00:31.065' AND ts >= '2025-06-22 09:00:31.065';

-- Q2:
explain analyze SELECT a.i8 FROM bar as a WHERE ts <= '2025-06-23 09:00:31.065' AND ts >= '2025-06-22 09:00:31.065';
```

You can see from the query outputs and database logs that the timestamp filter is not pushed down to the table scan in Q2, which is not good for performance.

### What did you expect to see?

The filter can be pushed down in Q2, just like Q1:

```plain text
+-------+------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| stage | node | plan                                                                                                                                                                                                                                                                            |
+-------+------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|     0 |    0 |  MergeScanExec: peers=[4440996184064(1034, 0), ] metrics=[output_rows: 0, greptime_exec_read_cost: 0, finish_time: 8536335, first_consume_time: 8533040, ready_time: 7475125, ]
                                                                                                |
|     1 |    0 |  SeqScan: region=4440996184064(1034, 0), "partition_count":{"count":0, "mem_ranges":0, "files":0, "file_ranges":0} metrics=[output_rows: 0, mem_used: 0, build_parts_cost: 10, build_reader_cost: 10, elapsed_await: 10, elapsed_poll: 18122, scan_cost: 10, yield_cost: 10, ]
 |
|  NULL | NULL | Total rows: 0                                                                                                                                                                                                                                                                   |
+-------+------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

### What did you see instead?

[...]

PR #6963, opened 2025-09-12
feat: support `SubqueryAlias` pushdown

discord9:

https://github.com/GreptimeTeam/greptimedb/issues/6873

as title, support `SubqueryAlias` pushdown using proper alias tracker

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/analyzer.rs
+++ b/src/query/src/dist_plan/analyzer.rs
@@ -653,6 +653,9 @@ impl EnforceDistRequirementRewriter {
                         result_alias_mapping.insert((original_col.clone(), *level), cols);
                     } else {
                         // if no aliased column found in current node, there should be alias in child node as promised by enforce col reqs
+                        // because it should insert required columns in child node
+                        // so we can find the alias in child node
+                        // if not found, it's an internal error
                         let aliases_in_child = aliased_columns_for(
                             &[original_col.clone()].into(),
                             child,
@@ -667,7 +670,7 @@ impl EnforceDistRequirementRewriter {
                                 "EnforceDistRequirementRewriter: no alias found for required column {original_col} in child plan {child} from original plan {original}",
                             )));
                         };
-                        // TODO(discord9): fix the problem of it in first projection, but not found in second projection
+
                         result_alias_mapping.insert((original_col.clone(), *level), aliases);
                     }
                 }
--- a/tests/cases/distributed/explain/subqueries.result
+++ b/tests/cases/distributed/explain/subqueries.result
@@ -236,15 +236,27 @@ DROP TABLE integers;
 
 Affected Rows: 0
 
-CREATE TABLE t(ts timestamp time index, a INT, b INT);
+CREATE TABLE t(ts timestamp time index, a INT, b INT)PARTITION ON COLUMNS (a) (
+  a < 1000,
+  a >= 1000 AND a < 2000,
+  a >= 2000
+);
 
 Affected Rows: 0
 
-CREATE TABLE t1(ts timestamp time index, a INT);
+CREATE TABLE t1(ts timestamp time index, a INT)PARTITION ON COLUMNS (a) (
+  a < 1000,
+  a >= 1000 AND a < 2000,
+  a >= 2000
+);
 
 Affected Rows: 0
 
-CREATE TABLE t2(ts timestamp time index, a INT);
+CREATE TABLE t2(ts timestamp time index, a INT)PARTITION ON COLUMNS (a) (
+  a < 1000,
+  a >= 1000 AND a < 2000,
+  a >= 2000
+);
 
 Affected Rows: 0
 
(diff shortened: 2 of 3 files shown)


src/query/src/dist_plan/analyzer/utils.rs
```diff
@@ -87,59 +158,60 @@ fn original_column_for_inner(
 ///
 /// if `original_node` is None, it means original columns are from leaf node
 pub fn aliased_columns_for(
-    original_columns: &HashSet<Column>,
+    original_columns: &BTreeSet<Column>,
     aliased_node: &LogicalPlan,
     original_node: Option<&LogicalPlan>,
-) -> DfResult<HashMap<Column, HashSet<Column>>> {
-    let initial_aliases: HashMap<Column, HashSet<Column>> = {
+) -> DfResult<BTreeMap<Column, BTreeSet<Column>>> {
```
evenyag, line 168: Why do we change to use BTreeSet?

src/query/src/dist_plan/analyzer.rs
```diff
@@ -599,15 +609,69 @@ struct EnforceDistRequirementRewriter {
...
             column_requirements,
             cur_level,
+            plan_per_level: BTreeMap::new(),
         }
     }
+
+    fn get_current_applicable_column_requirements(
+        &self,
+        node: &LogicalPlan,
+    ) -> DfResult<BTreeMap<(Column, usize), BTreeSet<Column>>> {
```
evenyag, line 630: Add some documents for the method? Expecially the return value.

src/query/src/dist_plan/analyzer.rs
```diff
@@ -621,23 +685,17 @@ impl TreeNodeRewriter for EnforceDistRequirementRewriter {
                     .to_string(),
             ));
         }
+        self.plan_per_level.insert(self.cur_level, node.clone());
```
evenyag, line 694: Is it possible that we have multiple node on the same level?

src/query/src/dist_plan/analyzer.rs
```diff
@@ -621,23 +685,17 @@ impl TreeNodeRewriter for EnforceDistRequirementRewriter {
...
         self.cur_level += 1;
         Ok(Transformed::no(node))
     }
 
     fn f_up(&mut self, node: Self::Node) -> DfResult<Transformed<Self::Node>> {
         self.cur_level -= 1;
```
evenyag, line 700: So we don't have to pop the plan from plan_per_level?

src/query/src/dist_plan/analyzer.rs
```diff
@@ -599,15 +609,69 @@ struct EnforceDistRequirementRewriter {
...
+            .collect::<Vec<_>>();
+
+        // track alias for columns and use aliased columns instead
+        // aliased col reqs at current level
+        let mut result_alias_mapping = BTreeMap::new();
+        let Some(child) = node.inputs().first().cloned() else {
+            return Ok(Default::default());
+        };
+        for (col_req, level) in col_req_per_level {
+            if let Some(original) = self.plan_per_level.get(level) {
```
evenyag, line 644: Looks like we will call this for each level. Does it mean that we may compute the same thing multiple times?

src/query/src/dist_plan/analyzer/utils.rs
discord9: > Why do we change to use BTreeSet?

deterministic, sometimes only need one alias from alias set, so choosing first one deterministically

src/query/src/dist_plan/analyzer.rs
discord9: > Is it possible that we have multiple node on the same level?

no since this is already on a expanding plan branch, which promise to have <= 1 input

src/query/src/dist_plan/analyzer.rs
discord9: > So we don't have to pop the plan from plan_per_level?

no, since code on smaller level might need to look up plan in larger level

src/query/src/dist_plan/analyzer.rs
discord9: > Looks like we will call this for each level. Does it mean that we may compute the same thing multiple times?

not the same things, new alias might be introduced by current node

src/query/src/dist_plan/analyzer/utils.rs
evenyag: We didn't document this in the comment.

evenyag (approved):
Supporting alias is so complicated......

Can this PR also solve https://github.com/GreptimeTeam/greptimedb/issues/6380?

discord9:
> Supporting alias is so complicated......
> 
> Can this PR also solve #6380?

True it is complicated, and it should solve those issues as now that aliased table get pushed down so later filter can push down too

src/query/src/dist_plan/analyzer.rs
```diff
@@ -599,15 +609,75 @@ struct EnforceDistRequirementRewriter {
...
+                            child,
+                            Some(original),
+                        )?;
+                        let Some(aliases) = aliases_in_child
+                            .get(original_col)
+                            .cloned()
+                            .filter(|a| !a.is_empty())
+                        else {
+                            return Err(datafusion_common::DataFusionError::Internal(format!(
+                                "EnforceDistRequirementRewriter: no alias found for required column {original_col} in child plan {child} from original plan {original}",
```
fengjiachun, line 670: Nit: Printing the logical plan to error might cause a log explosion?

fengjiachun (approved):
Rest LGTM

src/query/src/dist_plan/analyzer.rs
discord9: > Nit: Printing the logical plan to error might cause a log explosion?

`Display` of `LogicalPlan` seems rather concise? So it's kind of ok?

PR #6967, opened 2025-09-12
refactor: rewrite some UDFs to DataFusion style (part 2)

MichaelScofield:

as title

## PR Checklist


src/common/function/src/scalars/geo/geohash.rs
```diff
@@ -17,22 +17,21 @@ use std::sync::Arc;
...
-use snafu::{ResultExt, ensure};
+use snafu::ResultExt;
 
-use crate::function::{Function, FunctionContext};
+use crate::function::Function;
+use crate::scalars::geo::helpers;
 
 macro_rules! ensure_resolution_usize {
-    ($v: ident) => {
+    ($v: expr) => {
```
sunng87: using `expr` here can be dangerous, the expression might be evaluated multiple times for each `$v`. I suggest to keep it `ident`, if we have an expr for this, we can do `let v = expr;` first.

sunng87 (approved):
Much clear and much better. LGTM

PR #6968, opened 2025-09-14
fix: shorten lock time

shuiyisong:

This PR mainly does shorten lock time in OTLP metrics endpoint.
On my local machine, the CPU usage is slightly lower than before.

## PR Checklist


sunng87:
According to https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html#method.entry concurrent access to `entry` may risk dead lock.

sunng87:
Had some search that the recommendation of using dashmap is to avoid holding a lock across an await. This patch should have fixed this issue.

sunng87 (approved):

PR #6971, opened 2025-09-15
fix: clean intm ignore notfound

zhongzc:

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/index/intermediate.rs
+++ b/src/mito2/src/sst/index/intermediate.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::io::ErrorKind;
 use std::path::PathBuf;
 
 use async_trait::async_trait;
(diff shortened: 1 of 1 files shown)


src/mito2/src/sst/index/intermediate.rs
```diff
@@ -58,14 +59,22 @@ impl IntermediateManager {
         let aux_pb = PathBuf::from(aux_path.as_ref());
         let intm_dir = aux_pb.join(INTERMEDIATE_DIR);
         let deleted_dir = intm_dir.with_extension(format!("deleted-{}", Uuid::new_v4()));
-        if let Err(err) = tokio::fs::rename(&intm_dir, &deleted_dir).await {
-            warn!(err; "Failed to rename intermediate directory");
-        }
-        tokio::spawn(async move {
-            if let Err(err) = tokio::fs::remove_dir_all(deleted_dir).await {
-                warn!(err; "Failed to remove intermediate directory");
+        match tokio::fs::rename(&intm_dir, &deleted_dir).await {
```
evenyag, line 62: Is it better to check the dir existence first? Then we only rename the dir when it exists.

src/mito2/src/sst/index/intermediate.rs
```diff
@@ -58,14 +59,22 @@ impl IntermediateManager {
...
+            Ok(_) => {
+                tokio::spawn(async move {
+                    if let Err(err) = tokio::fs::remove_dir_all(deleted_dir).await
+                        && err.kind() != ErrorKind::NotFound
```
evenyag: It is possible that the renamed dir doesn't exist?

evenyag (approved):

src/mito2/src/sst/index/intermediate.rs
zhongzc: What you said is too complicated :-\ . Checking not found errors isn't that bad.

PR #6972, opened 2025-09-15
feat: enlarge SeriesScan distributor batch size and reduce sleep time

evenyag:

This PR modifies the SeriesDistributor to collect more rows before sending the batch using the sender.

It also reduces the sleep time if it can't send the batch directly.

This should reduce the sleep time when we have a SortPreservingMergeExec after SeriesScan. Although I couldn't reproduce the timeout using a unit test.

## PR Checklist


evenyag:
I'm going to reimplement SeqScan and SeriesScan.

PR #6978, opened 2025-09-16
fix: deadlock in dashmap

shuiyisong:

As the title suggests.

## PR Checklist

### DIFF
--- a/src/frontend/src/instance.rs
+++ b/src/frontend/src/instance.rs
@@ -377,38 +377,15 @@ impl Instance {
     ) -> server_error::Result<bool> {
         let db_string = ctx.get_db_string();
         // fast cache check
-        {
-            let cache = self
-                .otlp_metrics_table_legacy_cache
-                .entry(db_string.clone())
-                .or_default();
-
-            let hit_cache = names
-                .iter()
-                .filter_map(|name| cache.get(*name))
-                .collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


src/frontend/src/instance.rs
```diff
@@ -396,6 +396,9 @@ impl Instance {
                 // add doc links in err msg later
                 ensure!(!(hit_legacy && hit_prom), OtlpMetricModeIncompatibleSnafu);
 
+                // drop hit_cache to release references before inserting
```
evenyag: Do we need to wrap the `fast cache check` into a method and add a unit test for it?

evenyag (approved):

src/frontend/src/instance.rs
evenyag:
```suggestion
                // drop hit_cache to release references before inserting to avoid deadlock
```

killme2008: Agree. We need a test to cover this case.

v0y4g3r (approved):
LGTM

src/frontend/src/instance.rs
shuiyisong: Added in https://github.com/GreptimeTeam/greptimedb/pull/6978/commits/f671d6ae083629cb6c0624a13725083c32519af3

PR #6985, opened 2025-09-17
feat: gc worker metasrv scheduler

discord9:

gc worker metasrv trigger, which trigger gc worker on datanode depending on `file_removal_rate` and `sst_num`

TODO: ~still need to change the way datanode update `removed_files` to reduce the number of delete operation it send to object store~ done, but maybe also need to persistent actually delete files in `RegionEdit` 
TODO:
- more unit tests&integration tests

## PR Checklist

### DIFF
--- a/src/meta-srv/src/gc/ctx.rs
+++ b/src/meta-srv/src/gc/ctx.rs
@@ -152,6 +152,13 @@ impl SchedulerCtx for DefaultGcSchedulerCtx {
                         .or_default()
                         .push(*region_id);
                 }
+            } else {
+                return error::UnexpectedSnafu {
+                    violated: format!(
+                        "region_routes{region_routes:?} does not contain region_id: {region_id}",
+                    ),
+                }
+                .fail();
             }
         }
 
--- a/src/meta-srv/src/gc/handler.rs
+++ b/src/meta-srv/src/gc/handler.rs
@@ -384,6 +384,12 @@ impl GcScheduler {
         // Continue retrying until all regions succeed or reach max retry limit
         while !current_retry_regions.is_empty() && retry_round < self.config.max_retries_per_region
         {
+            // Apply exponential backoff if there are still regions to retry
+            if !current_retry_regions.is_empty() && retry_round < self.config.max_retries_per_region
+            {
+                self.apply_retry_backoff(current_retry_regions.len(), retry_round)
+                    .await;
+            }
             retry_round += 1;
 
             info!(
@@ -426,13 +432,6 @@ impl GcScheduler {
                 .values()
                 .flat_map(|r| r.need_retry_regions.iter().copied())
                 .collect();
-
-            // Apply exponential backoff if there are still regions to retry
-            if !current_retry_regions.is_empty() && retry_round < self.config.max_retries_per_region
-            {
-                self.apply_retry_backoff(current_retry_regions.len(), retry_round)
-                    .await;
-            }
         }
 
         info!(
@@ -677,7 +676,7 @@ impl GcScheduler {
             "{} regions still need retry after round {}, waiting {} seconds before next round (exponential backoff)",
             remaining_regions,
             retry_round,
-            backoff_duration.as_secs()
+            backoff_duration.as_secs_f32()
         );
 
         // Wait for backoff period before next retry round
--- a/src/meta-srv/src/gc/scheduler.rs
+++ b/src/meta-srv/src/gc/scheduler.rs
@@ -53,20 +53,9 @@ pub struct GcJobReport {
 impl GcJobReport {
     pub fn merge(&mut self, mut other: GcJobReport) {
         // merge per_datanode_reports&failed_datanodes
-        let all_dn_ids = self
-            .per_datanode_reports
-            .keys()
-            .cloned()
-            .chain(other.per_datanode_reports.keys().cloned())
-            .collect::<std::collections::HashSet<_>>();
-        for dn_id in all_dn_ids {
+        for (dn_id, report) in other.per_datanode_reports {
             let mut self_report = self.per_datanode_reports.entry(dn_id).or_default();
-            self_report.merge(
-                other
-                    .per_datanode_reports
-                    .remove(&dn_id)
-                    .unwrap_or_default(),
-            );
+            self_report.merge(report);
         }
         let all_failed_dn_ids = self
             .failed_datanodes
@@ -80,6 +69,8 @@ impl GcJobReport {
                 entry.extend(other_errors);
             }
         }
+        self.failed_datanodes
+            .retain(|dn_id, _| !self.per_datanode_reports.contains_key(dn_id));
     }
 }
 
--- a/src/meta-srv/src/gc/mock.rs
+++ b/src/meta-srv/src/gc/mock.rs
@@ -22,7 +22,7 @@ mod integration;
 mod misc;
 mod retry;
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 
@@ -30,18 +30,30 @@ use common_meta::datanode::{RegionManifestInfo, RegionStat};
 use common_meta::key::table_route::PhysicalTableRouteValue;
 use common_meta::peer::Peer;
 use common_meta::rpc::router::{Region, RegionRoute};
+use common_telemetry::debug;
 use ordered_float::OrderedFloat;
 use store_api::region_engine::RegionRole;
 use store_api::storage::{FileRefsManifest, GcReport, RegionId};
 use table::metadata::TableId;
 use tokio::sync::mpsc::Sender;
 
-use crate::error::Result;
+use crate::error::{Result, UnexpectedSnafu};
 use crate::gc::candidate::GcCandidate;
 use crate::gc::ctx::SchedulerCtx;
 use crate::gc::handler::Region2Peers;
 use crate::gc::options::GcSchedulerOptions;
 use crate::gc::scheduler::{Event, GcScheduler};
+
+pub fn new_empty_report_with(region_ids: impl IntoIterator<Item = RegionId>) -> GcReport {
+    let mut deleted_files = HashMap::new();
+    for region_id in region_ids {
+        deleted_files.insert(region_id, vec![]);
+    }
+    GcReport {
+        deleted_files,
+        need_retry_regions: HashSet::new(),
+    }
+}
 
 #[allow(clippy::type_complexity)]
 #[derive(Debug, Default)]
@@ -124,8 +136,10 @@ impl MockSchedulerCtx {
 
     /// Set success after a specific number of retries for a region
     pub fn set_gc_regions_success_after_retries(&self, region_id: RegionId, retries: usize) {
-        *self.gc_regions_success_after_retries.lock().unwrap() =
-            HashMap::from([(region_id, retries)]);
+        self.gc_regions_success_after_retries
+            .lock()
+            .unwrap()
+            .insert(region_id, retries);
     }
 
     /// Get the retry count for a specific region
@@ -198,8 +212,8 @@ impl SchedulerCtx for MockSchedulerCtx {
 
     async fn get_file_references(
         &self,
-        _region_ids: &[RegionId],
-        _region_to_peer: &Region2Peers,
+        region_ids: &[RegionId],
+        region_to_peer: &Region2Peers,
         _timeout: Duration,
     ) -> Result<FileRefsManifest> {
         *self.get_file_references_calls.lock().unwrap() += 1;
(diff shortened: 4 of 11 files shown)


src/meta-srv/src/gc/mailbox.rs
```diff
@@ -0,0 +1,224 @@
...
+use store_api::storage::{FileRefsManifest, GcReport, RegionId};
+
+use crate::error::{self, Result};
+use crate::gc::scheduler::GcScheduler;
+use crate::handler::HeartbeatMailbox;
+use crate::service::mailbox::Channel;
+
+impl GcScheduler {
+    /// Send an instruction to a datanode and wait for the reply.
+    ///
```
MichaelScofield:
```suggestion
```

src/meta-srv/src/gc/options.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Interval for cleaning up stale region entries from the GC tracker.
+    /// This removes entries for regions that no longer exist (e.g., after table drops).
+    /// Set to a larger value (e.g., 6 hours) since this is just for memory cleanup.
+    pub tracker_cleanup_interval: Duration,
+}
+
+impl Default for GcSchedulerOptions {
+    fn default() -> Self {
+        Self {
+            enable: false,
```
MichaelScofield, line 67: But the comment says it defaults to "true"?

src/mito2/src/gc.rs
```diff
@@ -386,40 +362,30 @@ impl LocalGcWorker {
...
+        // TODO(discord9): implement this
+        debug!(
+            "Trying to update manifest removed files for region {}",
+            region.region_id()
+        );
+        let mut manager = region.manifest_ctx.manifest_manager.write().await;
+        let mut manifest = (*manager.manifest()).clone();
+        manifest.removed_files.clear_deleted_files(deleted_files);
+        manager.set_manifest(Arc::new(manifest));
+        drop(manager);
```
MichaelScofield:
```suggestion
```

src/mito2/src/gc.rs
```diff
@@ -386,40 +362,30 @@ impl LocalGcWorker {
...
+    ) -> Result<()> {
+        // TODO(discord9): implement this
+        debug!(
+            "Trying to update manifest removed files for region {}",
+            region.region_id()
+        );
+        let mut manager = region.manifest_ctx.manifest_manager.write().await;
+        let mut manifest = (*manager.manifest()).clone();
+        manifest.removed_files.clear_deleted_files(deleted_files);
+        manager.set_manifest(Arc::new(manifest));
```
fengjiachun: Shouldn't it be persistent? Otherwise, restarting will cause the already deleted files to be deleted again, right?

discord9: > Shouldn't it be persistent? Otherwise, restarting will cause the already deleted files to be deleted again, right?

it's consider ok as gc worker at most will need to delete those deleted files one more times after restarting which is fine for now per @evenyag

src/meta-srv/src/gc/candidate.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Calculate GC priority score for a region based on various metrics.
+    fn calculate_gc_score(&self, region_stat: &RegionStat) -> f64 {
+        let sst_count_score = region_stat.sst_num as f64 * self.config.sst_count_weight;
+
+        let file_removal_rate_score = match &region_stat.region_manifest {
+            RegionManifestInfo::Mito {
+                file_removal_rate, ..
+            } => *file_removal_rate as f64 * self.config.file_removal_rate_weight,
+            // Metric engine doesn't have file_removal_rate, also this should be unreachable since metrics engine doesn't support gc
+            RegionManifestInfo::Metric { .. } => 0.0,
```
killme2008, line 61: Better to panic?

src/meta-srv/src/gc/candidate.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    #[allow(unused)]
+    fn score_f64(&self) -> f64 {
+        self.score.into_inner()
+    }
+}
+
+impl GcScheduler {
+    /// Calculate GC priority score for a region based on various metrics.
```
killme2008, line 52: It's better to doc the alogorithm.

src/meta-srv/src/gc/ctx.rs
```diff
@@ -0,0 +1,134 @@
...
+
+pub(crate) struct DefaultGcSchedulerCtx {
+    /// The metadata manager.
+    pub(crate) table_metadata_manager: TableMetadataManagerRef,
+    /// For getting `RegionStats`.
+    pub(crate) meta_peer_client: MetaPeerClientRef,
+    /// The mailbox to send messages.
+    pub(crate) mailbox: MailboxRef,
+    /// The server address.
+    pub(crate) server_addr: String,
```
killme2008, line 76: Who's server address?

src/meta-srv/src/gc/ctx.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+#[async_trait::async_trait]
+impl SchedulerCtx for DefaultGcSchedulerCtx {
+    async fn get_table_to_region_stats(&self) -> Result<HashMap<TableId, Vec<RegionStat>>> {
+        let dn_stats = self.meta_peer_client.get_all_dn_stat_kvs().await?;
+        let mut table_to_region_stats: HashMap<TableId, Vec<RegionStat>> = HashMap::new();
+        for (_dn_id, stats) in dn_stats {
+            let mut stats = stats.stats;
```
killme2008, line 85: Aren’t the stats already sorted by timestamp? I’m not sure, but it seems like they should be.

src/meta-srv/src/gc/ctx.rs
```diff
@@ -0,0 +1,134 @@
...
+use table::metadata::TableId;
+
+use crate::cluster::MetaPeerClientRef;
+use crate::error;
+use crate::error::{Result, TableMetadataManagerSnafu};
+use crate::handler::HeartbeatMailbox;
+use crate::service::mailbox::{Channel, MailboxRef};
+
+#[async_trait::async_trait]
+pub(crate) trait SchedulerCtx: Send + Sync {
```
killme2008, line 41: nit: love to doc the trait and functions

src/meta-srv/src/gc/ctx.rs
```diff
@@ -0,0 +1,134 @@
...
+            &format!("Metasrv@{}", self.server_addr),
+            &format!("Datanode-{}@{}", peer.id, peer.addr),
+            common_time::util::current_time_millis(),
+            &instruction,
+        )
+        .with_context(|_| error::SerializeToJsonSnafu {
+            input: instruction.to_string(),
+        })?;
+
+        let mailbox_rx = self
```
killme2008, line 325: Better to add a debug log here for the instruction and destination.

src/meta-srv/src/gc/ctx.rs
```diff
@@ -0,0 +1,134 @@
...
+        match mailbox_rx.await {
+            Ok(reply_msg) => {
+                let reply = HeartbeatMailbox::json_reply(&reply_msg)?;
+                Ok(reply)
+            }
+            Err(e) => {
+                error!(
+                    "Failed to receive reply from datanode {} for {}: {}",
+                    peer, description, e
+                );
```
killme2008, line 339:
```suggestion
                error!(e;
                    "Failed to receive reply from datanode {} for {}",
                    peer, description
                );
```

src/meta-srv/src/gc/handler.rs
```diff
@@ -0,0 +1,365 @@
...
+
+impl GcScheduler {
+    /// Iterate through all region stats, find region that might need gc, and send gc instruction to
+    /// the corresponding datanode with improved parallel processing and retry logic.
+    pub(crate) async fn trigger_gc(&self) -> Result<()> {
+        let start_time = Instant::now();
+        info!("Starting GC cycle");
+
+        // Step 1: Get all region statistics
+        let table_to_region_stats = self.ctx.get_table_to_region_stats().await?;
```
killme2008, line 46: Not sure if performance is an issue with many tables in the database.

src/meta-srv/src/gc/handler.rs
```diff
@@ -0,0 +1,365 @@
...
+        if per_table_candidates.is_empty() {
+            info!("No GC candidates found, skipping GC cycle");
+            return Ok(());
+        }
+
+        // Step 3: Process tables concurrently with limited parallelism
+        let (processed_tables, successful_tables, total_regions_processed) =
+            self.process_tables_concurrently(per_table_candidates).await;
+
+        let duration = start_time.elapsed();
```
killme2008, line 75: I think we should add some metrics for gc tasks for better o11y.

src/meta-srv/src/gc/handler.rs
```diff
@@ -0,0 +1,365 @@
...
+            table_id,
+            candidates.len()
+        );
+
+        // Step 1: Get table route information
+        // if is logic table, can simply pass.
+        let (phy_table_id, table_peer) = self.ctx.get_table_route(table_id).await?;
+
+        if phy_table_id != table_id {
+            return Ok(0);
```
killme2008: When does it happen? Do we need to log it?

src/meta-srv/src/gc/handler.rs
```diff
@@ -0,0 +1,365 @@
...
+                        Err(e) => {
+                            failed_regions.fetch_add(1, Ordering::Relaxed);
+                            error!("Failed to GC region {} after all retries: {}", region_id, e);
+                        }
+                    }
+                }
+            })
+            .await;
+
+        let successful_regions = successful_regions.load(Ordering::Relaxed);
```
killme2008: Again, need some metrics to measure the gc tasks.

src/meta-srv/src/gc/mailbox.rs
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        // Send GetFileRefs instructions to each datanode
+        let mut all_file_refs = HashMap::new();
+        let mut all_manifest_versions = HashMap::new();
+
+        for (peer, regions) in datanode_regions {
+            match self.send_get_file_refs_instruction(&peer, &regions).await {
+                Ok(manifest) => {
+                    // TODO(discord9): if other regions provide file refs for one region on other datanode, and no version,
```
killme2008: The todo comment issue is unlikely.

src/meta-srv/src/gc/options.rs
```diff
@@ -0,0 +1,168 @@
...
+            max_concurrent_tables: 10,
```
killme2008, line 68: Concurrency seems high; let's be conservative in the initial implementation.

src/meta-srv/src/gc/options.rs
```diff
@@ -0,0 +1,168 @@
...
+            // Clean up stale tracker entries every 6 hours
+            tracker_cleanup_interval: Duration::from_secs(60 * 60 * 6),
+        }
+    }
+}
+
+impl GcSchedulerOptions {
+    /// Validates the configuration options.
+    pub fn validate(&self) -> Result<()> {
+        if self.max_concurrent_tables == 0 {
```
killme2008, line 89: Replacing all these validations with `ensure!` looks better.

src/meta-srv/src/gc/scheduler.rs
```diff
@@ -0,0 +1,193 @@
...
+        common_runtime::spawn_global(async move { self.run().await });
+        info!("GC trigger started");
+        Ok(())
+    }
+
+    async fn run(&mut self) {
+        while let Some(event) = self.receiver.recv().await {
+            match event {
+                Event::Tick => {
+                    info!("Received gc tick");
```
killme2008, line 154: Annoying log
```suggestion
```

src/mito2/src/gc.rs
```diff
@@ -124,10 +126,10 @@ impl Default for GcConfig {
     fn default() -> Self {
         Self {
             enable: false,
-            // expect long running queries to be finished within a reasonable time
-            lingering_time: Duration::from_secs(60 * 5),
-            // 6 hours, for unknown expel time, which is when this file get removed from manifest, it should rarely happen, can keep it longer
-            unknown_file_lingering_time: Duration::from_secs(60 * 60 * 6),
+            // expect long running queries to be finished(or at least be able to notify it's using a deleted file) within a reasonable time
+            lingering_time: Some(Duration::from_secs(60)),
```
killme2008, line 131: Why change lingering time smaller?

src/mito2/src/gc.rs
```diff
@@ -386,6 +361,9 @@ impl LocalGcWorker {
             "Successfully deleted {} unused files for region {}",
             unused_len, region_id
         );
+        // TODO(discord9): update region manifest about deleted files
```
killme2008, line 376: The todo is outdated.

killme2008:
I’m wondering whether running the GC worker concurrently with migration or repartition tasks could lead to any concurrency issues. This is something I’m particularly concerned about. @waynexia @WenyXu @fengjiachun 

I’ve noticed that the current logic relies on **certain assumptions** — for example, that a region resides on the local datanode. However, these assumptions might not always hold true. It’s possible that the conditions are valid at the time of verification but become invalid during execution, leading to constraint violations.

The core question is: if these assumptions or checks are invalidated during execution due to ongoing migration or repartition operations, what would be the resulting behavior?

If such consistency cannot be guaranteed, then it would be safer to execute these processes sequentially rather than in parallel.

PR #6990, opened 2025-09-18
refactor: rewrite some UDFs to DataFusion style (part 3)

MichaelScofield:

as title

also tries to pass the `FunctionContext` through DataFusion

## PR Checklist

### DIFF
--- a/src/common/function/src/scalars/geo/measure.rs
+++ b/src/common/function/src/scalars/geo/measure.rs
@@ -20,7 +20,6 @@ use common_query::error::{self, Result};
 use datafusion_common::arrow::array::{Array, AsArray, Float64Builder};
 use datafusion_common::arrow::compute;
 use datafusion_common::arrow::datatypes::DataType;
-use datafusion_common::utils;
 use datafusion_expr::{ColumnarValue, ScalarFunctionArgs, Signature, Volatility};
 use derive_more::Display;
 use geo::algorithm::line_measures::metric_spaces::Euclidean;
@@ -28,7 +27,7 @@ use geo::{Area, Distance, Haversine};
 use geo_types::Geometry;
 use snafu::ResultExt;
 
-use crate::function::Function;
+use crate::function::{Function, extract_args};
 use crate::scalars::geo::wkt::parse_wkt;
 
 /// Return WGS84(SRID: 4326) euclidean distance between two geometry object, in degree
@@ -53,8 +52,7 @@ impl Function for STDistance {
         &self,
         args: ScalarFunctionArgs,
     ) -> datafusion_common::Result<ColumnarValue> {
-        let args = ColumnarValue::values_to_arrays(&args.args)?;
-        let [arg0, arg1] = utils::take_function_args(self.name(), args)?;
+        let [arg0, arg1] = extract_args(self.name(), &args)?;
 
         let arg0 = compute::cast(&arg0, &DataType::Utf8View)?;
         let wkt_this_vec = arg0.as_string_view();
@@ -107,8 +105,7 @@ impl Function for STDistanceSphere {
         &self,
         args: ScalarFunctionArgs,
     ) -> datafusion_common::Result<ColumnarValue> {
-        let args = ColumnarValue::values_to_arrays(&args.args)?;
-        let [arg0, arg1] = utils::take_function_args(self.name(), args)?;
+        let [arg0, arg1] = extract_args(self.name(), &args)?;
 
         let arg0 = compute::cast(&arg0, &DataType::Utf8View)?;
         let wkt_this_vec = arg0.as_string_view();
@@ -171,8 +168,7 @@ impl Function for STArea {
         &self,
         args: ScalarFunctionArgs,
     ) -> datafusion_common::Result<ColumnarValue> {
-        let args = ColumnarValue::values_to_arrays(&args.args)?;
-        let [arg0] = utils::take_function_args(self.name(), args)?;
+        let [arg0] = extract_args(self.name(), &args)?;
 
         let arg0 = compute::cast(&arg0, &DataType::Utf8View)?;
         let wkt_vec = arg0.as_string_view();
--- a/src/common/function/src/scalars/uddsketch_calc.rs
+++ b/src/common/function/src/scalars/uddsketch_calc.rs
@@ -19,13 +19,13 @@ use std::fmt::Display;
 use std::sync::Arc;
 
 use common_query::error::Result;
+use datafusion_common::DataFusionError;
 use datafusion_common::arrow::array::{Array, AsArray, Float64Builder};
 use datafusion_common::arrow::datatypes::{DataType, Float64Type};
-use datafusion_common::{DataFusionError, utils};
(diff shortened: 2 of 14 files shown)


src/common/function/src/system/database.rs
```diff
@@ -164,13 +186,26 @@ impl Function for ConnectionIdFunction {
...
+    ) -> datafusion_common::Result<ColumnarValue> {
+        let func_ctx = find_function_context(&args)?;
         let pid = func_ctx.query_ctx.process_id();
 
-        Ok(Arc::new(UInt32Vector::from_slice([pid])) as _)
+        Ok(ColumnarValue::Scalar(ScalarValue::UInt64(Some(pid as u64))))
     }
 }
 
+fn find_function_context(args: &ScalarFunctionArgs) -> datafusion_common::Result<&FunctionContext> {
```
sunng87: We can extract this to a common module, this will be used for a lot cases.

src/query/src/datafusion.rs
```diff
@@ -568,6 +569,16 @@ impl QueryEngine for DatafusionQueryEngine {
...
         }
+
+        state
+            .config_mut()
+            .options_mut()
+            .extensions
+            .insert(FunctionContext {
+                query_ctx: query_ctx.clone(),
+                state: self.engine_state().function_state(),
+            });
```
sunng87, line 580: we have similar code at line 529, confirm it's not duplicated

MichaelScofield: The two are not quite the same, though they looked similar. Here is the place where `FunctionContext` can be passed down to UDF args.

MichaelScofield:
@sunng87 PTAL

src/common/function/src/scalars/geo/measure.rs
```diff
@@ -47,33 +49,39 @@ impl Function for STDistance {
         Signature::string(2, Volatility::Stable)
     }
 
-    fn eval(&self, _func_ctx: &FunctionContext, columns: &[VectorRef]) -> Result<VectorRef> {
-        ensure_columns_n!(columns, 2);
+    fn invoke_with_args(
+        &self,
+        args: ScalarFunctionArgs,
+    ) -> datafusion_common::Result<ColumnarValue> {
```
killme2008, line 54: I think `ensure_columns_n!(columns, 2);` is still needed.

src/query/src/dist_plan/analyzer.rs
```diff
@@ -100,8 +102,35 @@ impl AnalyzerRule for DistPlannerAnalyzer {
...
         config: &ConfigOptions,
     ) -> datafusion_common::Result<LogicalPlan> {
-        // preprocess the input plan
-        let optimizer_context = OptimizerContext::new();
+        let mut config = config.clone();
+        // Aligned with the behavior in `datafusion_optimizer::OptimizerContext::new()`.
+        config.optimizer.filter_null_join_keys = true;
+        let config = Arc::new(config);
+
+        struct OptimizerContext {
```
killme2008, line 117: What is the purpose of this change?

src/common/function/src/system/database.rs
```diff
@@ -164,10 +186,14 @@ impl Function for ConnectionIdFunction {
...
-    fn eval(&self, func_ctx: &FunctionContext, _columns: &[VectorRef]) -> Result<VectorRef> {
+    fn invoke_with_args(
+        &self,
+        args: ScalarFunctionArgs,
+    ) -> datafusion_common::Result<ColumnarValue> {
+        let func_ctx = find_function_context(&args)?;
         let pid = func_ctx.query_ctx.process_id();
 
-        Ok(Arc::new(UInt32Vector::from_slice([pid])) as _)
+        Ok(ColumnarValue::Scalar(ScalarValue::UInt64(Some(pid as u64))))
```
killme2008: Why change to uint64?

src/common/function/src/scalars/uddsketch_calc.rs
```diff
@@ -71,30 +70,36 @@ impl Function for UddSketchCalcFunction {
...
-            .with_context(|| DowncastVectorSnafu {
-                err_msg: format!("expect BinaryVector, got {}", columns[1].vector_type_name()),
-            })?;
+    fn invoke_with_args(
+        &self,
+        args: ScalarFunctionArgs,
+    ) -> datafusion_common::Result<ColumnarValue> {
+        let args = ColumnarValue::values_to_arrays(&args.args)?;
+        let [arg0, arg1] = utils::take_function_args(self.name(), args)?;
+
```
killme2008: We've seen this pattern repeatedly. Would a macro be beneficial?

```rust
let [arg0, arg1, arg2] = extract_args!(self, &args);
```

src/common/function/src/scalars/geo/measure.rs
MichaelScofield: No need, DataFusion's `take_function_args` already covered the length check:

```rust
pub fn take_function_args<const N: usize, T>(
    function_name: &str,
    args: impl IntoIterator<Item = T>,
) -> Result<[T; N]> {
    let args = args.into_iter().collect::<Vec<_>>();
    args.try_into().map_err(|v: Vec<T>| {
        _exec_datafusion_err!(
            "{} function requires {} {}, got {}",
            function_name,
            N,
            if N == 1 { "argument" } else { "arguments" },
            v.len()
        )
    })
}
```

src/query/src/dist_plan/analyzer.rs
MichaelScofield: The `ConstEvaluator` in `SimplifyExpressions` might evaluate some UDFs early in the planning stage, by executing them directly. For example, the `database()` function. So the `ConfigOptions` here (which is set from the session context) should be present in the UDF's `ScalarFunctionArgs`. However, the default implementation in DataFusion seems to lost track on it: the `ConfigOptions` is recreated with its default values again. So we create a custom `OptimizerConfig` with the desired `ConfigOptions` to walk around the issue.

(also included in the code comments)

killme2008: Not sure if this is the best way. Should we create a DataFusion issue? WDYT @waynexia

src/common/function/src/scalars/uddsketch_calc.rs
MichaelScofield: Maybe not, the `ColumnarValue::values_to_arrays` is not always used together with `take_function_args`. For example, in "clamp" functions and "vector" functions.

MichaelScofield:
@killme2008 PTAL

src/common/function/src/scalars/uddsketch_calc.rs
killme2008: In this PR, it repeated about 10 times more. I think it's worthy to refactor.

src/query/src/dist_plan/analyzer.rs
killme2008: @discord9

waynexia: This looks a bit strange to me. Maybe in some places we reset the exec context

PR #6998, opened 2025-09-22
fix: step aggr merge phase not order nor filter

discord9:

as title, to mimic df's aggr_mode=Final's behavior, also is not meaningful anyway for merge wrapper to have order by, fix is:
- add fix/unfix analyzer which set/unset state wrapper's order/filter to correct ones, to workaround that substrait can't en/decode udaf with extra info
- explicitly set filter/order_by to none for merge wrapper so that it would confuse datafusion when it need check logical plan for column refs
- also in ScanHintRule, previously it require all aggr to be `last_value`, now it also accept `__last_value_state` function as well to apply last row selector and optimize query perf.

context:
the original problem is that state wrapper must know it's ordering field when it's `return_type` method being called, but it's not available there, so a "Fixing" analyzer is needed to give state wrapper proper ordering field(which is already there but wasn't accessible by state wrapper before fixing.

but another problem emerge after "Fixing", that is state wrapper will now produce a different return type due to ordering field being added, but substrait can't encode/decode state wrapper's internal fields(state wrapper's `ordering` field for example) so encode state wrapper with ordering field will cause the decoded plan to have  mismatch schema and cause problem, so before encode to substrait it need to be "unfixed" which remove ordering field inside udaf impl, and re-fixed after decoded. This is a workaround until datafusion can support more complex udaf to be encoded

## PR Checklist


src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -118,6 +124,8 @@ impl StateMergeHelper {
     }
 
     /// Split an aggregate plan into two aggregate plans, one for the state function and one for the merge function.
+    ///
+    /// and upper plan's aggr input's
```
evenyag: The sentence looks incomplete.

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -179,9 +187,13 @@ impl StateMergeHelper {
             let arg = Expr::Column(Column::new_unqualified(lower_state_output_col_name));
             let expr = AggregateFunction {
                 func: Arc::new(merge_func.into()),
+                // notice how filter/order_by is not supported in the merge function, as it's not meaningful to have them in the merge phase.
```
evenyag:
```suggestion
                // notice now filter/order_by is not supported in the merge function, as it's not meaningful to have them in the merge phase.
```

evenyag (commented):
sqlness failed...

discord9:
last_value's test being move from common to separate result file for standalone/distributed becuase their physical plan have sight difference(due to order of apply for optimizer perhaps)

tests/cases/standalone/optimizer/first_value_advance.sql
evenyag, line 1: Looks like we can put them under the common directory.

evenyag (approved):
Didn't check test content......

src/common/function/src/aggrs/aggr_wrapper/fix_order.rs
```diff
@@ -0,0 +1,189 @@
...
+        if self.is_dirty {
+            let node = LogicalPlan::Aggregate(aggregate).recompute_schema()?;
+            debug!(
+                "FixStateUdafOrderingAnalyzer: plan schema's field changed to {:?}",
+                node.schema().fields()
+            );
+
+            Ok(Transformed::yes(node))
+        } else {
+            Ok(Transformed::yes(LogicalPlan::Aggregate(aggregate)))
```
zhongzc: Transformed::no?

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -57,6 +62,38 @@ pub fn aggr_merge_func_name(aggr_name: &str) -> String {
...
+/// As in if it can be split into multiple steps:
+/// i.e. on datanode first call `state(input)` then
+/// on frontend call `calc(merge(state))` to get the final result.
+pub fn is_all_aggr_exprs_steppable(aggr_exprs: &[Expr]) -> bool {
+    aggr_exprs.iter().all(|expr| {
+        if let Some(aggr_func) = get_aggr_func(expr) {
+            if aggr_func.params.distinct {
+                // Distinct aggregate functions are not steppable(yet).
+                return false;
+            }
```
zhongzc, line 76: Looking at the implementation below, there's a lot of handling for distinct, but it's excluded here. Is there something missing?

tests/cases/standalone/optimizer/first_value_advance.sql
discord9: > Looks like we can put them under the common directory.

see comment:
> last_value's test being move from common to separate result file for standalone/distributed becuase their physical plan have sight difference(due to order of apply for optimizer perhaps)

src/common/function/src/aggrs/aggr_wrapper.rs
discord9: > Looking at the implementation below, there's a lot of handling for distinct, but it's excluded here. Is there something missing?

intended, those handling are for when later support distinct, for now they should have no effect

zhongzc: Can we add a TODO here to indicate the plan?

src/common/function/src/aggrs/aggr_wrapper/tests.rs
```diff
@@ -545,6 +547,210 @@ async fn test_avg_udaf() {
...
+    // evaluate the state function
+    let input = Int64Array::from(vec![Some(1), Some(2), None, Some(3)]);
+    let values = vec![Arc::new(input) as arrow::array::ArrayRef];
+
+    state_accum.update_batch(&values).unwrap();
+
+    let state = state_accum.state().unwrap();
+
+    // FIXME(discord9): once datafusion fixes the issue that last_value udaf state fields are not correctly(missing ordering field if `last` field is part of ordering field)
+    // then change it back to 3 fielda
```
zhongzc:
```suggestion
    // then change it back to 3 fields
```

PR #7008, opened 2025-09-23
fix: group by expr not as column in step aggr

discord9:

fix a bug where group by exprs in upper aggr in step aggr failed to change to column refs, causing 
step aggr to silently failed and cause wrong query plan(sliently!), fix:
- fail explicitly when transformer failed
- upper group by expr use column ref instead of original exprs
- also correctly deduce aggregate's original return type

## PR Checklist


src/query/src/dist_plan/analyzer.rs
```diff
@@ -839,7 +843,7 @@ impl TreeNodeRewriter for PlanRewriter {
         let parent = parent.clone();
 
         // TODO(ruihang): avoid this clone
```
waynexia:
```suggestion
```

src/query/src/dist_plan/commutativity.rs
```diff
@@ -142,24 +146,27 @@ impl Categorizer {
...
-                    return Commutativity::TransformedCommutative {
+                    return Ok(Commutativity::TransformedCommutative {
                         transformer: Some(Arc::new(|plan: &LogicalPlan| {
                             debug!("Before Step optimize: {plan}");
                             let ret = step_aggr_to_upper_aggr(plan);
-                            ret.ok().map(|s| TransformerAction {
+                            ret.inspect_err(|err| {
+                                common_telemetry::error!("Failed to step aggregate plan: {err}");
+                            })
+                            .map(|s| TransformerAction {
```
waynexia, line 156: ~~if it's only for logging error, we can still keep the return type as `Option<T>` instead of `Result<T>` by calling `.ok()` after `inspect_err()` to minimize the change set?~~

ignore it

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -245,7 +257,20 @@ impl StateWrapper {
...
+            .iter()
+            .map(|e| e.return_field(acc_args.schema))
+            .collect::<Result<Vec<_>, _>>()?;
+        self.inner.return_field(&input_fields).inspect_err(|e| {
+            common_telemetry::error!(
+                "StateWrapper: {:#?}\nacc_args:{:?}\nerror:{:?} at {}",
+                &self,
+                &acc_args,
+                e,
+                snafu::location!()
```
evenyag: Is it helpful to log the location here? It will always be the current line in this file.

src/query/src/dist_plan/commutativity.rs
```diff
@@ -142,24 +146,27 @@ impl Categorizer {
...
                 if !matches_partition && is_all_steppable {
                     debug!("Plan is steppable: {plan}");
-                    return Commutativity::TransformedCommutative {
+                    return Ok(Commutativity::TransformedCommutative {
                         transformer: Some(Arc::new(|plan: &LogicalPlan| {
                             debug!("Before Step optimize: {plan}");
                             let ret = step_aggr_to_upper_aggr(plan);
-                            ret.ok().map(|s| TransformerAction {
+                            ret.inspect_err(|err| {
+                                common_telemetry::error!("Failed to step aggregate plan: {err}");
```
evenyag:
```suggestion
                                common_telemetry::error!("Failed to step aggregate plan: {err:?}");
```

evenyag:
> fix a bug where group by exprs in upper aggr in step aggr failed to change to column refs, causing
> step aggr to silently failed and cause wrong query plan(sliently!)

What's the wrong query plan generated without this fix?

src/common/function/src/aggrs/aggr_wrapper.rs
discord9: > Is it helpful to log the location here? It will always be the current line in this file.

~yes, since many other debug msg will have similiar msg, and it's good to know why the expand happen for all the reasons~changed per review

discord9:
> > fix a bug where group by exprs in upper aggr in step aggr failed to change to column refs, causing
> > step aggr to silently failed and cause wrong query plan(sliently!)
> 
> What's the wrong query plan generated without this fix?

i.e. for a group by date_bin('1s', ts) the upper plan (without this fix) will not change the group exprs,
but in fact it should use col("date_bin('1s', ts)") which is a column ref to the lower plan's output

PR #7015, opened 2025-09-24
fix: not step when aggr have order by/filter

discord9:

due to bad support in datafusion and substrait, aggr with order by/filter have some issues when using/serde-ing, so disable step aggr for those case for now until those issues get fixed&properly tested:
- disable step aggr for aggr func with order by/filter
- set state function's return type to all nullable for case when input is empty but need a state(which usually will be null)

## PR Checklist


waynexia (approved):
todo count +1 🥲

PR #7029, opened 2025-09-26
feat: implement four build types for async index build

SNC123:

https://github.com/GreptimeTeam/greptimedb/issues/6756 (Tracking Issue)

**Note: This PR will need to be updated after the two PRs below are merged.**

https://github.com/GreptimeTeam/greptimedb/pull/6927 (Dependency, basic framework)

https://github.com/GreptimeTeam/greptime-proto/pull/276  (Dependency, proto)

As title. 🐭
Implemented four build types for async index build:
- **Flush**: Triggered after a flush operation.
- **Compact**: Triggered after compaction if new files are created.
- **SchemaChange**:  Triggered by a schema change when a fast-path index build is required.
- **Manual**:  Triggered manually by an administrator via the build_index('table_name') function.

## PR Checklist


src/store-api/src/region_request.rs
```diff
@@ -1342,6 +1343,8 @@ impl Default for RegionCompactRequest {
         }
     }
 }
+#[derive(Debug, Clone, Default)]
+pub struct RegionBuildIndexRequest {}
```
evenyag, line 1359:
```suggestion

#[derive(Debug, Clone, Default)]
pub struct RegionBuildIndexRequest {}
```

src/mito2/src/worker/handle_flush.rs
```diff
@@ -246,9 +247,24 @@ impl<S: LogStore> RegionWorkerLoop<S> {
             return;
         }
 
+        let index_build_file_metas = request.edit.files_to_add.clone();
```
evenyag: You can take `files_to_add` to avoid cloning.

src/mito2/src/manifest/action.rs
```diff
@@ -53,6 +53,13 @@ pub struct RegionChange {
     /// Format of the SST.
     #[serde(default)]
     pub sst_format: FormatType,
+    /// A flag indicates whether the index has been changed.
+    ///
+    /// For backward compatibility, this field is default to `false` when
+    /// deserializing from old manifest.
+    /// Used for index build, to determine whether we need to build index files after schema change.
+    #[serde(default)]
+    pub need_index: bool,
```
evenyag: I didn't find the place that sets this flag back to `false`. How do we know we have built all the indexes?

If this flag indicates whether the index has been changed, should we name it `index_changed` instead?

src/mito2/src/manifest/manager.rs
```diff
@@ -178,6 +178,7 @@ impl RegionManifestManager {
             RegionChange {
                 metadata: metadata.clone(),
                 sst_format,
+                need_index: true,
```
evenyag: Why the initial value of `need_index` is true?

src/mito2/src/worker/handle_alter.rs
```diff
@@ -280,3 +282,16 @@ fn log_option_update<T: std::fmt::Debug>(
...
+    match kind {
+        // `SetIndexes` is a fast-path operation because it can build indexes for existing SSTs
+        // in the background, without needing to wait for a flush or compaction cycle.
+        AlterKind::SetIndexes { options: _ } => true,
+        // For AddColumns, DropColumns, UnsetIndexes and ModifyColumnTypes, we don't treat them as index changes.
+        // Index files still need to be rebuilt after schema changes,
+        // but this will happen automatically during flush or compaction.
+        _ => false,
+    }
+}
```
evenyag, line 296: The meaning of `fast path index` is a little confusing. Maybe `need change index` is more straightforward.

src/mito2/src/worker/handle_compaction.rs
```diff
@@ -80,9 +84,26 @@ impl<S> RegionWorkerLoop<S> {
             region.file_purger.clone(),
         );
 
+        let index_build_file_metas = request.edit.files_to_add.clone();
```
evenyag: We can take the `files_to_add`.

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -107,13 +134,48 @@ impl<S> RegionWorkerLoop<S> {
...
+                "Scheduling index build for region {}, file_id {}",
+                region_id,
+                file_handle.meta_ref().file_id
+            );
+            let task = self.new_index_build_task(
+                &region,
+                file_handle.clone(),
+                request.build_type.clone(),
+                ResultMpscSender::new(tx.clone()),
+            );
```
evenyag, line 171: non-blocking: We may need a way to abort the rebuild process when the region is closed or the engine is stopped.

src/mito2/src/sst/index.rs
```diff
@@ -436,6 +437,15 @@ pub enum IndexBuildOutcome {
...
 }
+/// Mpsc output result sender.
+#[derive(Debug)]
+pub struct ResultMpscSender(Sender<Result<IndexBuildOutcome>>);
+
+impl ResultMpscSender {
+    pub fn new(tx: Sender<Result<IndexBuildOutcome>>) -> Self {
+        Self(tx)
+    }
+}
```
evenyag: Looks like there is no need to wrap a new type?

src/mito2/src/sst/index.rs
```diff
@@ -512,6 +535,17 @@ impl IndexBuildTask {
     ) -> Result<IndexBuildOutcome> {
         let version = &version_data.version;
         let mut indexer = self.indexer_builder.build(self.file_meta.file_id).await;
+
+        // Check SST file existence before building index to avoid failure of parquet reader.
+        if !self.check_sst_file_exists(version).await {
```
evenyag: I found that the check_sst_file_exists is an O(N) operation, although the files are already indexed by the file id.

The file meta actually contains the level of the file.

src/mito2/src/engine/index_build_test.rs
```diff
@@ -0,0 +1,305 @@
...
+    let request = CreateRequestBuilder::new().build();
+    let column_schemas = rows_schema(&request);
+    engine
+        .handle_request(region_id, RegionRequest::Create(request))
+        .await
+        .unwrap();
+
+    // Flush and make sure there is no index file.
+    put_and_flush(&engine, region_id, &column_schemas, 10..20).await;
+    wait().await;
```
evenyag: Can we use wait_finish instead?

SNC123: That might not work, as the index build task isn't expected to complete, which would cause wait_finish() to block.

I also considered using something like wait_begin(), but that wouldn't distinguish this case from one where the index file simply hasn't been created yet due to timing.

src/mito2/src/manifest/action.rs
SNC123: It seems I misunderstood the semantics of `RegionChange` earlier. Upon reflection, adding a `need_index` field to it doesn't seem like a good idea. Perhaps it would be more appropriate to compare the old and new metadata directly in `handle_manifest_region_change_result()`?

evenyag: If we don't need to persist it, we can pass it as a method parameter like the sender to avoid comparing the metadata again.

src/mito2/src/engine/index_build_test.rs
evenyag: Can we reopen the region? We can check the index files after that.

src/mito2/src/manifest/action.rs
SNC123: Wouldn't this violate the Single Responsibility Principle? For example, adding a method parameter to `handle_manifest_region_change`([here](https://github.com/SNC123/greptimedb/blob/feat/async_index_build_types/src/mito2/src/worker/handle_alter.rs#L142)) seems unrelated to its core responsibility.

src/mito2/src/worker/handle_rebuild_index.rs
SNC123: Would a mechanism similar to `FlushScheduler::on_region_dropped` be what you have in mind? I've started to draft this approach in a follow-up PR [here](https://github.com/SNC123/greptimedb/commit/ba50e0d621b4f50a0aec2d76025375ade4a6ce81#diff-759f77af307bbbd62c33e50a5415763d8e64a545810861f1fd770d5efefa9667R836-R852).

evenyag: We can check the RegionRoleState before each task.

src/mito2/src/manifest/action.rs
evenyag: It should be fine. Whether the index is changed is the knowledge that the method needs. Index change is a kind of region change.

src/mito2/src/worker/handle_rebuild_index.rs
```diff
@@ -107,13 +134,64 @@ impl<S> RegionWorkerLoop<S> {
...
+            debug!(
+                "Scheduling index build for region {}, file_id {}",
+                region_id,
+                file_handle.meta_ref().file_id
+            );
+
+            let should_abort = region.state()
+                == RegionRoleState::Leader(RegionLeaderState::Dropping)
+                || region.state() == RegionRoleState::Leader(RegionLeaderState::Downgrading)
+                || region.state() == RegionRoleState::Leader(RegionLeaderState::Truncating);
```
evenyag: We should also abort when the state is staging or the region is no longer a leader.

Maybe you should wrap a method `should_abort_index` for RegionRoleState.

evenyag (approved):
Others look good to me.

src/mito2/src/worker/handle_rebuild_index.rs
SNC123: Updated.

PR #7033, opened 2025-09-27
feat: struct value

sunng87:

Adding the last piece of `Value` type...

## PR Checklist


src/api/src/helper.rs
```diff
@@ -159,6 +161,45 @@ impl From<ColumnDataTypeWrapper> for ConcreteDataType {
...
             }
+            ColumnDataType::List => {
+                if let Some(TypeExt::ListType(d)) = datatype_wrapper
+                    .datatype_ext
+                    .as_ref()
+                    .and_then(|datatype_ext| datatype_ext.type_ext.as_ref())
+                {
+                    let item_type = ColumnDataTypeWrapper {
+                        datatype: d.datatype(),
+                        datatype_ext: d.datatype_extension.clone().map(|d| *d),
```
killme2008, line 172: nitpick: use `cloned` or `copied` instead of `clone` and `map`.

src/api/src/helper.rs
```diff
@@ -159,6 +161,45 @@ impl From<ColumnDataTypeWrapper> for ConcreteDataType {
...
+                    };
+                    ConcreteDataType::list_datatype(item_type.into())
+                } else {
+                    // invalid state: type extension not found
+                    ConcreteDataType::null_datatype()
```
killme2008, line 177: Better to return an error instead of filling with null values?

src/api/src/helper.rs
```diff
@@ -249,6 +290,37 @@ impl ColumnDataTypeWrapper {
             }),
         }
     }
+
+    pub fn list_datatype(item_type: ColumnDataTypeWrapper) -> Self {
```
killme2008, line 295: Please add comments for the `list_datatype` and `struct_datatype` functions, especially for `struct_datatype`,which its arguments are a list of tuple.

src/datatypes/src/types/string_type.rs
```diff
@@ -90,7 +90,7 @@ impl DataType for StringType {
             Value::Decimal128(v) => Some(Value::String(StringBytes::from(v.to_string()))),
 
             // StringBytes is only support for utf-8, Value::Binary is not allowed.
```
killme2008: The comment need to be fixed too.

src/datatypes/src/vectors/null.rs
```diff
@@ -197,6 +237,29 @@ impl MutableVector for NullVectorBuilder {
...
+    }
+
+    fn push(&mut self, _value: Option<<Self::VectorType as ScalarVector>::RefItem<'_>>) {
+        self.length += 1;
+    }
+
+    fn finish(&mut self) -> Self::VectorType {
+        let result = NullVector::new(self.length);
+
+        self.length = 0;
```
killme2008, line 254: Seems unnecessary:

```suggestion
```

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -94,16 +99,36 @@ impl Vector for StructVector {
...
     }
 
-    fn get(&self, _: usize) -> Value {
-        unimplemented!("StructValue not supported yet")
+    fn get(&self, index: usize) -> Value {
+        if !self.array.is_valid(index) {
+            return Value::Null;
+        }
+
+        let mut values = Vec::new();
```
killme2008: 1. Prefer using an iterator instead of looping and collecting.
2. Repeatedly calling `Helper::try_into_vector(field_array)` seems costly; consider caching the result.

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -159,23 +182,288 @@ impl VectorOp for StructVector {
...
 impl Serializable for StructVector {
     fn serialize_to_json(&self) -> Result<Vec<serde_json::Value>> {
-        let mut result = serde_json::Map::new();
+        let mut vectors = BTreeMap::new();
         for (field, value) in self.array.fields().iter().zip(self.array.columns().iter()) {
             let value_vector = Helper::try_into_vector(value)?;
+            vectors.insert(field.name().clone(), value_vector);
+        }
+
+        let mut results = Vec::with_capacity(self.array().len());
```
killme2008: Nit: prefer iterator.

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -159,23 +182,288 @@ impl VectorOp for StructVector {
...
+                };
+
+                Some(Some(value))
+            }
+        } else {
+            None
+        }
+    }
+
+    #[inline]
```
killme2008:
```suggestion
```

src/datatypes/src/scalars.rs
```diff
@@ -187,6 +187,27 @@ impl_scalar_for_native!(i64, Int64Type);
 impl_scalar_for_native!(f32, Float32Type);
 impl_scalar_for_native!(f64, Float64Type);
 
+impl Scalar for () {
+    type VectorType = NullVector;
+    type RefType<'a> = ();
+
+    #[inline]
```
killme2008: We don't use `inline` anymore.

```suggestion
```

src/datatypes/src/value.rs
```diff
@@ -908,6 +943,60 @@ impl Ord for ListValue {
...
+    pub fn take_items(self) -> Vec<Value> {
+        self.items
+    }
+
+    pub fn struct_type(&self) -> &StructType {
+        &self.fields
+    }
+
+    fn estimated_size(&self) -> usize {
+        self.items
```
killme2008, line 995: Do we need to consider the `fields` size?

src/datatypes/src/value.rs
```diff
@@ -548,16 +587,11 @@ pub fn to_null_scalar_value(output_type: &ConcreteDataType) -> Result<ScalarValu
             IntervalType::DayTime(_) => ScalarValue::IntervalDayTime(None),
             IntervalType::MonthDayNano(_) => ScalarValue::IntervalMonthDayNano(None),
         },
-        ConcreteDataType::List(_) => ScalarValue::List(Arc::new(ListArray::new_null(
-            Arc::new(new_item_field(output_type.as_arrow_type())),
-            0,
-        ))),
+        ConcreteDataType::List(list_type) => {
```
killme2008, line 586: The original implementation uses a zero-length list, but here it becomes a single element. Is that expected?

src/datatypes/src/value.rs
```diff
@@ -1355,27 +1465,40 @@ pub fn transform_value_ref_to_json_value<'a>(
 /// Now comparison still requires some allocation (call of `to_value()`) and
 /// might be avoidable by downcasting and comparing the underlying array slice
 /// if it becomes bottleneck.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 pub enum ListValueRef<'a> {
     // TODO(yingwen): Consider replace this by VectorRef.
```
killme2008, line 1447: The TODO may be can removed. cc @evenyag

src/servers/src/postgres/types.rs
```diff
@@ -62,6 +62,14 @@ pub(super) fn schema_to_pg(origin: &Schema, field_formats: &Format) -> Result<Ve
         .collect::<Result<Vec<FieldInfo>>>()
 }
 
+fn encode_struct(
+    _query_ctx: &QueryContextRef,
+    _struct_value: &StructValue,
+    _builder: &mut DataRowEncoder,
+) -> PgWireResult<()> {
+    todo!("how to encode struct for postgres");
```
killme2008, line 70: Should we return an error instead of panicking?

src/datatypes/src/value.rs
```diff
@@ -824,14 +854,15 @@ impl TryFrom<Value> for serde_json::Value {
             Value::String(bytes) => serde_json::Value::String(bytes.into_string()),
             Value::Binary(bytes) => serde_json::to_value(bytes)?,
             Value::Date(v) => serde_json::Value::Number(v.val().into()),
-            Value::List(v) => serde_json::to_value(v)?,
+            Value::List(v) => serde_json::to_value(v.items())?,
```
evenyag: Is it a breaking change? Although it should be fine, as we didn't expose the list data type to users.

src/datatypes/src/value.rs
```diff
@@ -2183,8 +2503,7 @@ mod tests {
             to_json(Value::Duration(Duration::new_millisecond(1)))
         );
 
-        let json_value: serde_json::Value =
-            serde_json::from_str(r#"{"items":[{"Int32":123}],"datatype":{"Int32":{}}}"#).unwrap();
+        let json_value: serde_json::Value = serde_json::from_str(r#"[{"Int32":123}]"#).unwrap();
```
evenyag: Also need a test for the struct type.

src/servers/src/postgres/types.rs
discord9, line 70: maybe return a unsupported error? currently struct type is only use internally so postgres shouldn't have it anyway
p.s. maybe later can support encode struct type to pg's composite type if needed

src/datatypes/src/value.rs
```diff
@@ -1002,8 +1092,21 @@ impl TryFrom<ScalarValue> for Value {
...
                 .unwrap_or(Value::Null),
+            ScalarValue::Struct(struct_array) => {
+                let struct_type: StructType = (struct_array.fields()).try_into()?;
+                let items = struct_array
+                    .columns()
+                    .iter()
+                    .map(|array| {
+                        // we only take first element from each array
+                        let field_scalar_value = ScalarValue::try_from_array(array.as_ref(), 0)
+                            .context(ConvertArrowArrayToScalarsSnafu)?;
```
evenyag, line 1128: Is it correct to only take the first element?

src/datatypes/src/value.rs
```diff
@@ -908,6 +943,60 @@ impl Ord for ListValue {
...
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct StructValue {
+    items: Vec<Value>,
+    fields: StructType,
+}
+
+impl StructValue {
+    pub fn new(items: Vec<Value>, fields: StructType) -> Self {
+        Self { items, fields }
```
MichaelScofield: better `ensure` that `len`s of `items` and `fields` are equal

src/servers/src/postgres/types.rs
sunng87: This is implemented in next patch #7052

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -12,33 +12,34 @@
...
 }
 
-#[allow(unused)]
 impl StructVector {
-    pub fn new(array: StructArray) -> Result<Self> {
-        let fields = array.fields();
-        let data_type = ConcreteDataType::Struct(fields.try_into()?);
-        Ok(StructVector { array, data_type })
+    pub fn new(fields: StructType, array: StructArray) -> Self {
+        StructVector { array, fields }
```
MichaelScofield: better check that `fields` and `array` are compatible, so that the convention in `get` can be ensured `unreachable`

src/servers/src/postgres/types.rs
sunng87: This is implemented in next patch https://github.com/GreptimeTeam/greptimedb/pull/7052

src/datatypes/src/vectors/null.rs
sunng87: this is to reset builder state. this build api seems to be designed for multiple use.

src/common/macro/src/row/schema.rs
```diff
@@ -90,6 +90,12 @@ fn impl_schema_method(fields: &[ParsedField<'_>]) -> Result<TokenStream2> {
                                 Some(ColumnDataTypeExtension { type_ext: Some(TypeExt::VectorType(VectorTypeExtension { dim: #dim })) })
                             }
                         }
+                        Some(TypeExt::ListType(_ext)) => {
+                            todo!("macro for generating list item types")
+                        }
+                        Some(TypeExt::StructType(_ext)) => {
+                            todo!("macro for generating struct fields")
```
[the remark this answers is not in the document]
sunng87: this is something I missed

src/datatypes/src/value.rs
sunng87: This is a bug of original implementation, it requires 1 item in null list

src/datatypes/src/value.rs
sunng87: not 100% sure, but the field information should not be stored as data

src/api/src/helper.rs
sunng87: It's just type conversion, and it's a `From` impl without error case

src/api/src/helper.rs
```diff
@@ -1075,7 +1239,8 @@ pub fn value_to_grpc_value(value: Value) -> GrpcValue {
                 convert_month_day_nano_to_pb(v),
             )),
             Value::Decimal128(v) => Some(ValueData::Decimal128Value(convert_to_pb_decimal128(v))),
-            Value::List(_) | Value::Duration(_) => unreachable!(),
+            // TODO
+            Value::List(_) | Value::Duration(_) | Value::Struct(_) => unreachable!(),
```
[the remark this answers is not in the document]
sunng87: another TODO

src/api/src/helper.rs
sunng87: this is for take the content out of box

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -159,23 +182,288 @@ impl VectorOp for StructVector {
...
-        let mut result = serde_json::Map::new();
+        let mut vectors = BTreeMap::new();
         for (field, value) in self.array.fields().iter().zip(self.array.columns().iter()) {
             let value_vector = Helper::try_into_vector(value)?;
+            vectors.insert(field.name().clone(), value_vector);
+        }
+
+        let mut results = Vec::with_capacity(self.array().len());
+        for idx in 0..self.array().len() {
+            let mut result = serde_json::Map::new();
```
MichaelScofield: `serde_json::Map` does not preserve order unless its `preserve_order` feature is on. To make deserialize consistent with serialize, I think order on fields are required.

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -159,23 +182,288 @@ impl VectorOp for StructVector {
...
+            }
+
+            results.push(result.into());
+        }
+
+        Ok(results)
+    }
+}
+
+impl From<StructArray> for StructVector {
```
MichaelScofield: Do we really need this infallible `From`, instead of `TryFrom`?

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -159,23 +182,288 @@ impl VectorOp for StructVector {
...
+    fn push_struct_value(&mut self, struct_value: &StructValue) -> Result<()> {
+        for (index, value) in struct_value.items().iter().enumerate() {
+            self.value_builders[index].try_push_value_ref(&value.as_value_ref())?;
+        }
+        self.null_buffer.append_non_null();
+
+        Ok(())
+    }
+
+    fn push_null_struct_value(&mut self) -> Result<()> {
```
MichaelScofield: seems no need to return `Result`:

```suggestion
    fn push_null_struct_value(&mut self) {
```

src/api/src/helper.rs
```diff
@@ -622,6 +689,77 @@ pub fn pb_value_to_value_ref<'a>(
...
+            let list_datatype_ext = datatype_ext
+                .as_ref()
+                .and_then(|ext| {
+                    if let Some(TypeExt::ListType(l)) = &ext.type_ext {
+                        Some(l)
+                    } else {
+                        None
+                    }
+                })
+                .expect("list must contain datatype ext");
```
MichaelScofield, line 704: better return `Result`, because the input could be generated by clients, might be buggy

src/datatypes/src/vectors/struct_vector.rs
sunng87: We should never rely on the order of map entries...

src/datatypes/src/vectors/null.rs
```diff
@@ -112,6 +113,21 @@ impl Vector for NullVector {
...
+    type Builder = NullVectorBuilder;
+
+    fn get_data(&self, _idx: usize) -> Option<Self::RefItem<'_>> {
+        Some(())
+    }
+
+    fn iter_data(&self) -> Self::Iter<'_> {
+        NullIter::new(self)
+    }
+}
```
evenyag, line 129: Why do we need to implement ScalarVector for NullVector?

src/api/src/helper.rs
sunng87: this is going introduce massive refactors, will do it in next patch.

src/datatypes/src/value.rs
sunng87: Just changed to `try_new` but we still need some unwrap in our api

src/datatypes/src/vectors/struct_vector.rs
sunng87: Will check if we can get item from StructArray directly

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -159,23 +198,281 @@ impl VectorOp for StructVector {
...
         for (field, value) in self.array.fields().iter().zip(self.array.columns().iter()) {
             let value_vector = Helper::try_into_vector(value)?;
+            vectors.insert(field.name().clone(), value_vector);
+        }
+
+        (0..self.array.len())
+            .map(|idx| {
+                let mut result = serde_json::Map::with_capacity(vectors.len());
+                for field in vectors.keys() {
+                    let field_value = vectors.get(field).unwrap().get(idx);
```
evenyag: Looks like we can use a Vec instead of a map for `vectors`. We always iterate over all fields.

src/datatypes/src/value.rs
sunng87: Yes, list is not used in before we don't consider it's a breaking change. Also the previous implementation will return list as an object, which can be confusing.

Checked usage, this api is mostly used for influxdb return type

src/datatypes/src/value.rs
sunng87: Yes we assume they are all single-element arrays when converting to scalarvalue. Do you know if there is any exceptions? list?

src/datatypes/src/vectors/null.rs
sunng87: This is not required but nice to have. `NullVector` is the only vector missing `ScalarVector` implementation. I was trying to setup a connection between `Vector` trait and `ScalarVector` so we can use `iter_data` for any vector

src/datatypes/src/value.rs
sunng87: Let me know if we want to keep this @evenyag

sunng87:
@killme2008 @MichaelScofield @evenyag @discord9 most comments haven been addressed, please check again and we need to move on.

there is also related refactoring patch #7062 #7063 to simplify these

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -124,9 +159,7 @@ impl VectorOp for StructVector {
             column_arrays,
             self.array.nulls().cloned(),
         );
-        Arc::new(
-            StructVector::new(replicated_array).expect("Failed to create replicated StructVector"),
-        )
+        Arc::new(StructVector::try_new(self.fields.clone(), replicated_array).unwrap())
```
MichaelScofield, line 162: How are the above `unwrap`s safe?

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -159,23 +192,294 @@ impl VectorOp for StructVector {
 
 impl Serializable for StructVector {
     fn serialize_to_json(&self) -> Result<Vec<serde_json::Value>> {
-        let mut result = serde_json::Map::new();
-        for (field, value) in self.array.fields().iter().zip(self.array.columns().iter()) {
+        let mut vectors = Vec::new();
```
MichaelScofield:
```suggestion
        let mut vectors = Vec::with_capacity(self.array.columns().len());
```

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -159,23 +192,294 @@ impl VectorOp for StructVector {
...
+            assert_eq!(struct_value.struct_type(), &struct_type);
+            let mut items = struct_value.items().iter();
+            assert_eq!(items.next(), Some(&Value::Int32(1)));
+            assert_eq!(items.next(), Some(&Value::String("tom".into())));
+            assert_eq!(items.next(), Some(&Value::UInt8(25)));
+            assert_eq!(items.next(), Some(&Value::String("94038".into())));
+            assert_eq!(items.next(), None);
+        } else {
+            panic!("Expected a struct value");
+        }
```
discord9, line 484: maybe some tests with nested struct/list vector?

src/datatypes/src/vectors/struct_vector.rs
discord9: > How are the above `unwrap`s safe?

mostly because the `self` is already a valid struct vector?

src/datatypes/src/vectors/null.rs
evenyag: We may remove ScalarVector and switch to using arrow arrays directly in the future.

sunng87: sounds good. We should switch to plain arrow api as much as possible

src/datatypes/src/vectors/struct_vector.rs
sunng87: Will do it in #7052

src/datatypes/src/vectors/struct_vector.rs
sunng87: Yes, because the original vector is valid.

PR #7034, opened 2025-09-27
perf: some optimization for async index build

SNC123:

https://github.com/GreptimeTeam/greptimedb/issues/6756 (Tracking Issue)

**Note: This PR will need to be rebased after the PR below are merged.**

https://github.com/GreptimeTeam/greptimedb/pull/7029 (Dependency, build types implement)

Implemented three optimization for async index build:
- **Full SST Indexing Optimization**: Builds indexes for entire SST files at once, replacing the previous incremental method.
- **Task Resource Management**: Limits resource usage of background tasks to maintain system performance.
- **Task Priority Mechanism**: Prioritizes background tasks, allowing critical operations to execute sooner.

## PR Checklist


src/mito2/src/sst/index.rs
```diff
@@ -493,9 +540,30 @@ impl IndexBuildTask {
...
+                    region_id: self.file_meta.region_id,
+                    notify: BackgroundNotify::IndexBuildStopped(IndexBuildStopped {
+                        region_id: self.file_meta.region_id,
+                        file_id: self.file_meta.file_id,
+                    }),
+                };
+                let _ = self
+                    .request_sender
+                    .send(WorkerRequestWithTime::new(worker_request))
+                    .await;
```
zhongzc: Duplicate processing with the outer

zhongzc: And why not call `on_failure`?

SNC123: Maybe mistake when rebase... 🐁  
Will fix later.

evenyag:
Please fix the CI.

SNC123:
Fixed. @evenyag

src/mito2/src/gc.rs
```diff
@@ -309,7 +309,16 @@ impl LocalGcWorker {
...
-        self.delete_files(region_id, &unused_files).await?;
+        let unused_index_files: Vec<Option<FileId>> = unused_files
+            .iter()
+            .map(|file_id| {
+                current_files
+                    .get(file_id)
+                    .and_then(|meta| meta.index_file_id)
+            })
+            .collect();
+        self.delete_files(region_id, &unused_files, &unused_index_files)
```
evenyag: Why don't we call the `index_file_id()` method and filter all None. We can pass index_file_ids as `&[FileId]`.

src/mito2/src/sst/file.rs
```diff
@@ -381,6 +406,7 @@ pub async fn delete_files(
     region_id: RegionId,
     file_ids: &[FileId],
     delete_index: bool,
+    index_file_ids: &[Option<FileId>],
```
evenyag: Could we pass `&[File]`. We can delete SSTs first, then delete puffin files.

src/mito2/src/sst/file.rs
```diff
@@ -146,6 +146,11 @@ pub struct FileMeta {
     pub available_indexes: SmallVec<[IndexType; 4]>,
     /// Size of the index file.
     pub index_file_size: u64,
+    /// File ID of the index file.
+    ///
+    /// When this field is None, it means the index file id is the same as the file id.
+    /// Used for rebuilding index files.
+    pub index_file_id: Option<FileId>,
```
evenyag, line 154: https://github.com/GreptimeTeam/greptimedb/blob/main/src/store-api/src/sst_entry.rs
We may need to add this field to sst entries.

src/mito2/src/sst/index.rs
```diff
@@ -701,26 +775,198 @@ impl IndexBuildTask {
     }
 }
 
-#[derive(Clone)]
+impl PartialEq for IndexBuildTask {
+    fn eq(&self, other: &Self) -> bool {
+        self.reason == other.reason
+    }
+}
```
evenyag, line 762: Will two build type have the same priority in the future?

We don't have to implement PartialEq/Ord for IndexBuildType. We can use `reason.priority()` for comparision inside IndexBuildTask. So the PartialEq of IndexBuildType doesn't couple with IndexBuildTask.

src/mito2/src/sst/index.rs
```diff
@@ -701,26 +775,198 @@ impl IndexBuildTask {
...
+            .entry(task.file_meta.region_id)
+            .or_insert_with(|| IndexBuildStatus::new(task.file_meta.region_id));
+
+        if status.building_files.contains(&task.file_meta.file_id) {
+            // Index is already being built for this file, skip.
+            debug!(
+                "Index is already being built for region {}, file {}, skipping.",
+                task.file_meta.region_id, task.file_meta.file_id
+            );
+            return Ok(());
```
evenyag: We didn't notify the senders in the task.

src/mito2/src/sst/file.rs
SNC123: I noticed that there is already a field index_file_path for tracking the latest state of the index file, which can retrieve the latest information from FileMeta ([here](https://github.com/SNC123/greptimedb/blob/8d525ba523e53c16e49f201a16d83219a6950ad0/src/mito2/src/region.rs#L589-L592)). Could this field be unnecessary?

SNC123:
Thanks for so quick code review! Updated.

src/mito2/src/sst/file.rs
zhongzc: Can add it, the redundancy here doesn't have much impact.

src/mito2/src/sst/file.rs
```diff
@@ -381,6 +406,7 @@ pub async fn delete_files(
     region_id: RegionId,
     file_ids: &[FileId],
     delete_index: bool,
```
evenyag, line 409: @discord9 The delete_index flag here isn't very necessary as it only controls whether to remove the key from the cache. I think we can remove it later.

src/mito2/src/sst/file.rs
```diff
@@ -392,9 +418,13 @@ pub async fn delete_files(
...
 
-    for file_id in file_ids {
+    for (idx, file_id) in file_ids.iter().enumerate() {
         let region_file_id = RegionFileId::new(region_id, *file_id);
-        match access_layer.delete_sst(&region_file_id).await {
+        let index_file_id = RegionFileId::new(region_id, index_file_ids[idx]);
+        match access_layer
+            .delete_sst(&region_file_id, &index_file_id)
+            .await
+        {
```
evenyag, line 429: If we always ensure a file id has a corresponding index id, we can change `file_ids` to `&[(FileId, FileId)]` to avoid the implicit assumption that `index_file_ids.len() == file_ids.len()`.

src/mito2/src/sst/index.rs
```diff
@@ -701,26 +755,205 @@ impl IndexBuildTask {
...
+                RegionFileId::new(task.file_meta.region_id, task.file_meta.file_id);
+            debug!(
+                "Aborting index build task since index is already being built for region file {:?}",
+                region_file_id
+            );
+            task.on_success(IndexBuildOutcome::Aborted(format!(
+                "Index is already being built for region file {:?}",
+                region_file_id
+            )))
+            .await;
```
evenyag, line 841: non-blocking: Do we have a way to block and wait the previous task of this file? Returning OK here is fine.

src/mito2/src/gc.rs
```diff
@@ -309,7 +309,28 @@ impl LocalGcWorker {
...
+            "Found {} unused index files to delete for region {}",
+            unused_index_files.len(),
+            region_id
+        );
+
+        let file_pairs: Vec<(FileId, FileId)> = unused_files
+            .iter()
+            .zip(unused_index_files.iter())
+            .map(|(file_id, index_file_id)| (*file_id, *index_file_id))
+            .collect();
```
evenyag: We collect pairs in one run.
```suggestion
        let file_pairs: Vec<(FileId, FileId)> = unused_files
            .iter()
            .filter_map(|file_id| {
                current_files
                    .get(file_id)
                    .map(|meta| (meta.file_id().file_id(), meta.index_file_id().file_id()))
            })
            .collect();

        info!(
            "Found {} unused index files to delete for region {}",
            file_pairs.len(),
            region_id
        );
```

src/mito2/src/sst/index.rs
SNC123: We could mimic the merge mechanism in `RegionFlushTask` and adjust the result sender. The difference here is that we need to track running tasks instead of pending ones. This requires further careful consideration later.

PR #7036, opened 2025-09-28
refactor: make `Function` trait a simple shim of DataFusion UDF

MichaelScofield:

I was thinking of removing `Function` trait completely, but decide to keep it for future use maybe. At last, the UDFs are of greptimedb. However, to keep things simple, the `Function` trait is kept aligned with DataFusion's `ScalaUDFImpl`.

## PR Checklist

### DIFF
--- a/src/common/function/src/function_factory.rs
+++ b/src/common/function/src/function_factory.rs
@@ -52,9 +52,7 @@ impl From<ScalarUDF> for ScalarFunctionFactory {
 impl From<FunctionRef> for ScalarFunctionFactory {
     fn from(func: FunctionRef) -> Self {
         let name = func.name().to_string();
-        let func = Arc::new(move |ctx: FunctionContext| {
-            create_udf(func.clone(), ctx.query_ctx, ctx.state)
-        });
+        let func = Arc::new(move |_| create_udf(func.clone()));
         Self {
             name,
             factory: func,
--- a/src/common/function/src/system.rs
+++ b/src/common/function/src/system.rs
@@ -34,14 +34,35 @@ pub(crate) struct SystemFunction;
 
 impl SystemFunction {
     pub fn register(registry: &FunctionRegistry) {
-        registry.register_scalar(BuildFunction);
-        registry.register_scalar(VersionFunction);
-        registry.register_scalar(DatabaseFunction);
-        registry.register_scalar(ReadPreferenceFunction);
-        registry.register_scalar(PgBackendPidFunction);
-        registry.register_scalar(ConnectionIdFunction);
-        registry.register_scalar(TimezoneFunction);
+        registry.register_scalar(BuildFunction::default());
+        registry.register_scalar(VersionFunction::default());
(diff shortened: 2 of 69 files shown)


src/common/function/src/system.rs
```diff
@@ -34,14 +34,35 @@ pub(crate) struct SystemFunction;
...
+
+        impl Default for $name {
+            fn default() -> Self {
+                Self {
+                    signature: datafusion_expr::Signature::nullary(Volatility::Immutable),
+                }
+            }
+        }
+    };
+}
```
sunng87, line 66: This type of macro is going to be difficult to read and update. Do you know if it's possible to write it as derive macro like `derive(NullarySignature)`

sunng87: ~~Another idea, can we define this signature as a `&'static` constant, so it doesn't have to be member of all these functions.~~ this requires const fn

sunng87 (approved):

src/common/function/src/function_factory.rs
```diff
@@ -52,9 +52,7 @@ impl From<ScalarUDF> for ScalarFunctionFactory {
 impl From<FunctionRef> for ScalarFunctionFactory {
     fn from(func: FunctionRef) -> Self {
         let name = func.name().to_string();
-        let func = Arc::new(move |ctx: FunctionContext| {
-            create_udf(func.clone(), ctx.query_ctx, ctx.state)
-        });
+        let func = Arc::new(move |_| create_udf(func.clone()));
```
fengjiachun, line 55: Question: Is the `FunctionContext` transmission ignored here?

src/common/function/src/system.rs
MichaelScofield: > This type of macro is going to be difficult to read and update. Do you know if it's possible to write it as derive macro like `derive(NullarySignature)`

It's possible, but will make more complicate codes: two procedure macros have to be used together, one for rewriting(or redefining) the struct to add the "signature" field, the other one for implementing the `Default` on it. I think it'd be better to keep the simple `macro_rules` way.

src/common/function/src/function_factory.rs
MichaelScofield: `FunctionContext` is passed by DataFusion now. See here: 

https://github.com/GreptimeTeam/greptimedb/blob/c4a7cc0adbcb99cfa308f101e2b0015cf5fdd3be/src/query/src/datafusion.rs#L573-L586

fengjiachun: For example, the SessionState created by `DefaultPlanDecoder::decode` does not seem to call engine_context(), so FunctionContext cannot be obtained.

MichaelScofield: Seems the `FunctionContext` is not used in`DefaultPlanDecoder`, so it doesn't matter. If there are any UDFs in the future that would require retrieving the `FunctionContext`, I can add it.

PR #7046, opened 2025-09-30
chore: rename the default ts column name to `greptime_timestamp` for influxdb line protocol

shuiyisong:

As the title suggests. 
This is **NOT** a breaking change. For existing users, the handler will use the existing table's ts columns name.

## PR Checklist

### DIFF
--- a/src/servers/src/influxdb.rs
+++ b/src/servers/src/influxdb.rs
@@ -25,7 +25,6 @@ use crate::row_writer::{self, MultiTableData};
 
 const INFLUXDB_API_PATH_NAME: &str = "influxdb";
 const INFLUXDB_API_V2_PATH_NAME: &str = "influxdb/api/v2";
-// const INFLUXDB_TIMESTAMP_COLUMN_NAME: &str = "ts";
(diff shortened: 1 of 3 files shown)


sunng87:
Can we confirm the behavior? Will the handler check current timestamp column? It's not shown from the changeset

shuiyisong:
> Can we confirm the behavior? Will the handler check current timestamp column? It's not shown from the changeset

Yes, I've tested on my local machine.
1. Start a standalone instance of the old version, write data using line protocol. The column name is `ts`.
2. Stop and restart the instance using the new version, and write data using line protocol. The data can be correctly appended to the table.

killme2008:
> > Can we confirm the behavior? Will the handler check current timestamp column? It's not shown from the changeset
> 
> Yes, I've tested on my local machine.
> 
> 1. Start a standalone instance of the old version, write data using line protocol. The column name is `ts`.
> 2. Stop and restart the instance using the new version, and write data using line protocol. The data can be correctly appended to the table.

Yes, it's fixed in https://github.com/GreptimeTeam/greptimedb/pull/6126

PR #7049, opened 2025-10-02
refactor!: unify the API of getting total cpu and memory

zyy17:

1. Add `get_total_cpu_millicores()` / `get_total_cpu_cores()` / `get_total_memory_bytes()` / `get_total_memory_readable()` in common-stat to unify the API of getting total CPU and memory.

   - The original implementation by using `num_cpus::get()` can not handle the non-integer cgroups cpu limit(for example, `1500m`);
   
2. Breaking changes in `information_schema.cluster_info` table:

   - `cpus` -> `total_cpu_millicores`(use the `millicores` as the basic unit)
   - `memory_bytes` -> `total_memory_bytes`

## PR Checklist


sunng87:
@zyy17 I already have typo fixes in #7047 
We can merge that standalone pr first.

sunng87 (approved):

src/common/config/src/utils.rs
```diff
@@ -13,61 +13,22 @@
...
-}
+use common_stat::{get_total_cpu_millicores, get_total_memory_readable};
 
 /// `ResourceSpec` holds the static resource specifications of a node,
 /// such as CPU cores and memory capacity. These values are fixed
 /// at startup and do not change dynamically during runtime.
 #[derive(Debug, Clone, Copy)]
 pub struct ResourceSpec {
-    pub cpus: usize,
+    pub cpus: i64,
```
MichaelScofield, line 23: Why use `i64` for type of `cpus`? Can cpus be negative?

zyy17: Actually, for most scenarios, `i64` and `usize` **will both be fine**. It's big enough to store the `cpus`. However, using `i64` will be general and it's compatible with Prometheus IntGauge.

PR #7052, opened 2025-10-04
feat: conversion between struct and value

sunng87:

This is successor of #7033 

The idea is to add data conversion between serde_json's Value to our Value, vice versa. This path contains some AI generated code.

## PR Checklist


sunng87:
@MichaelScofield @discord9 PTAL

sunng87:
There are residual issues with this patch that I haven't been able to figure out a solution.

At the moment, we can assume `Struct` as json data, however, if the input is `List`, or even scalar values like number or boolean, it can be json, too. This requires us to encode them as json at protocol layer, especially for postgres, which has first-class type for json. However, currently we are in lack of a mechanism for this. `LogicalType` seems to be a good solution. Any concrete data type can be json.

src/datatypes/src/json.rs
```diff
@@ -0,0 +1,2348 @@
...
+            }
+        }
+        Json::Array(json_array) => {
+            let item_type = if let Some(ConcreteDataType::List(list_type)) = data_type {
+                Some(list_type.item_type())
+            } else {
+                None
+            };
+            let list_value = encode_json_array_with_context(json_array, item_type, context)?;
+            Ok(Value::List(list_value))
```
sunng87, line 182: Here we have a gap. If the input is a json array, it will be stored with `List` type. However, we cannot assume list is json because some udf can generate `List` (like geo). `List` is encoded as Array in postgres protocol.

src/datatypes/src/json.rs
```diff
@@ -0,0 +1,2363 @@
...
+                Ok(Value::Struct(struct_value))
+            } else if data_type.is_some() {
+                Err(error::InvalidJsonSnafu {
+                    value: "JSON object can only be encoded to Struct type".to_string(),
+                }
+                .build())
+            } else {
+                let struct_value = encode_json_object_with_context(json_object, None, context)?;
+                Ok(Value::Struct(struct_value))
+            }
```
MichaelScofield: maybe make codes look clearer?

```suggestion
            ensure!(
                matches!(data_type, Some(ConcreteDataType::Struct(_)) | None),
                error::InvalidJsonSnafu {
                    value: "JSON object can only be encoded to Struct type".to_string(),
                }
            )

            let data_type = data_type.and_then(|x| x.as_struct());
            let struct_value = encode_json_object_with_context(json_object, data_type, context)?;
            Ok(Value::Struct(struct_value))
```

src/datatypes/src/json.rs
MichaelScofield: Maybe we could create a new `Value` variant for json?

```rust
enum Value {
  ...
  Json(Value), // <- so that it could represent any values
}

File: src/datatypes/src/json.rs [hunk 3]
```diff
@@ -0,0 +1,2363 @@
...
+}
+
+fn encode_json_object_with_context<'a>(
+    mut json_object: Map<String, Json>,
+    fields: Option<&StructType>,
+    context: &JsonContext<'a>,
+) -> Result<StructValue, Error> {
+    let mut items = Vec::new();
+    let mut struct_fields = Vec::new();
+    let mut processed_keys = std::collections::HashSet::new();
```
MichaelScofield: Better pre-allocate the capacity:

```suggestion
    let fields_len = fields.map(|x| x.fields().len();
    let mut items = Vec::with_capacity(fields_len);
    let mut struct_fields = Vec::with_capacity(fields_len);
    let mut processed_keys = std::collections::HashSet::with_capacity(fields_len);
```

then after the "if":

```rust
items.reserve_exact(json_object.len());
struct_fields.reserve_exact(json_object.len());
```

src/datatypes/src/value.rs
```diff
@@ -979,6 +979,13 @@ impl StructValue {
         Ok(Self { items, fields })
     }
 
+    /// Create a new struct value.
+    ///
+    /// Panics if the number of items does not match the number of fields.
```
MichaelScofield, line 992: Why not simply `try_new`?

src/datatypes/src/json.rs
```diff
@@ -0,0 +1,2363 @@
...
+                // Field exists in schema but not in JSON - add null value
+                items.push(Value::Null);
+                struct_fields.push(field.clone());
+            }
+        }
+    }
+
+    // Then, process any remaining JSON fields that weren't in the schema
+    for (key, value) in json_object {
+        if processed_keys.contains(&key) {
```
MichaelScofield: Seems `processed_keys` is of no use? The `key`s here can not be in the `processed_keys` (otherwise they must be removed from `json_object` above), this "contains" always returns false.

src/datatypes/src/json.rs
```diff
@@ -0,0 +1,2363 @@
...
+    let struct_type = StructType::new(struct_fields);
+    StructValue::try_new(items, struct_type)
+}
+
+fn encode_json_array_with_context<'a>(
+    json_array: Vec<Json>,
+    item_type: Option<&ConcreteDataType>,
+    context: &JsonContext<'a>,
+) -> Result<ListValue, Error> {
+    let mut items = Vec::new();
```
MichaelScofield:
```suggestion
    let mut items = Vec::with_capacity(json_array.len());
```

src/datatypes/src/json.rs
```diff
@@ -0,0 +1,2363 @@
...
+        _ => decode_primitive_value(value),
+    }
+}
+
+/// Decode a structured value to JSON object
+fn decode_struct_with_context<'a>(
+    struct_value: StructValue,
+    context: &JsonContext<'a>,
+) -> Result<Json, Error> {
+    let mut json_object = Map::new();
```
MichaelScofield:
```suggestion
    let mut json_object = Map::with_capacity(fields.fields().len());
```

src/datatypes/src/json.rs
```diff
@@ -0,0 +1,2363 @@
...
+
+    Ok(Json::Object(json_object))
+}
+
+/// Decode a list value to JSON array
+fn decode_list_with_context<'a>(
+    list_value: ListValue,
+    context: &JsonContext<'a>,
+) -> Result<Json, Error> {
+    let mut json_array = Vec::new();
```
MichaelScofield:
```suggestion
    let mut json_array = Vec::with_capacity(data_items.len());
```

src/datatypes/src/json.rs
```diff
@@ -0,0 +1,2363 @@
...
+    match value {
+        // Handle expected format: StructValue with single _raw field
+        Value::Struct(struct_value) => {
+            if struct_value.struct_type().fields().len() == 1 {
+                let field = &struct_value.struct_type().fields()[0];
+                if field.name() == JsonStructureSettings::RAW_FIELD
+                    && let Some(Value::String(s)) = struct_value.items().first()
+                {
+                    let json_str = s.as_utf8();
+                    return serde_json::from_str(json_str).context(error::DeserializeSnafu {
```
MichaelScofield:
```suggestion
                    return serde_json::from_str(json_str).with_context(|_| error::DeserializeSnafu {
```

src/datatypes/src/json.rs
```diff
@@ -0,0 +1,2363 @@
...
+            Err(error::InvalidJsonSnafu {
+                value: "Unstructured value must be stored as struct with single _raw field"
+                    .to_string(),
+            }
+            .build())
+        }
+        // Handle old format: plain string (for backward compatibility)
+        Value::String(s) => {
+            let json_str = s.as_utf8();
+            serde_json::from_str(json_str).context(error::DeserializeSnafu {
```
MichaelScofield:
```suggestion
            serde_json::from_str(json_str).with_context(|_| error::DeserializeSnafu {
```

src/datatypes/src/json.rs
```diff
@@ -0,0 +1,2363 @@
...
+    }
+
+    // Check if the entire decoding should be unstructured (UnstructuredRaw case)
+    if matches!(context.settings, JsonStructureSettings::UnstructuredRaw) {
+        // For UnstructuredRaw, the entire struct should be reconstructed from _raw field
+        return decode_unstructured_raw_struct(struct_value);
+    }
+
+    let mut items = Vec::new();
+    let mut struct_fields = Vec::new();
```
MichaelScofield:
```suggestion
    let mut items = Vec::with_capacity(fields.fields().len());
    let mut struct_fields = Vec::with_capacity(fields.fields().len());
```

src/datatypes/src/json.rs
```diff
@@ -0,0 +1,2363 @@
...
+    let struct_type = StructType::new(struct_fields);
+    StructValue::try_new(items, struct_type)
+}
+
+/// Decode a ListValue that was encoded with current settings back into a fully structured ListValue
+fn decode_list_with_settings<'a>(
+    list_value: ListValue,
+    context: &JsonContext<'a>,
+) -> Result<ListValue, Error> {
+    let mut items = Vec::new();
```
MichaelScofield:
```suggestion
    let mut items = Vec::with_capacity(data_items.len());
```

src/datatypes/src/json.rs
```diff
@@ -0,0 +1,2363 @@
...
+fn decode_unstructured_raw_struct(struct_value: StructValue) -> Result<StructValue, Error> {
+    // For UnstructuredRaw, the struct must have exactly one field named "_raw"
+    if struct_value.struct_type().fields().len() == 1 {
+        let field = &struct_value.struct_type().fields()[0];
+        if field.name() == JsonStructureSettings::RAW_FIELD
+            && let Some(Value::String(s)) = struct_value.items().first()
+        {
+            let json_str = s.as_utf8();
+            let json_value: Json =
+                serde_json::from_str(json_str).context(error::DeserializeSnafu {
```
MichaelScofield:
```suggestion
                serde_json::from_str(json_str).with_context(|_| error::DeserializeSnafu {
```

src/datatypes/src/value.rs
sunng87: this provide a shortcut when it's ok to run unwrap like in test cases

src/datatypes/src/json.rs
sunng87: Good catch!

src/datatypes/src/json.rs
sunng87: Interesting, let me consider this. You can have an offline discussion with @evenyag too

evenyag: It's ok to have a json value.

sunng87:
I will be experimenting `Value::json(value)` in a new branch. We can merge this if it looks good

src/datatypes/src/json.rs
sunng87: Discuss in https://github.com/GreptimeTeam/greptimedb/issues/7084

PR #7056, opened 2025-10-09
fix: only skips auto convert when encoding is sparse

evenyag:

The skip_auto_convert flag should only take effect when the encoding is sparse.

Otherwise, we will have an incorrect schema during compaction when the encoding is dense.

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet/flat_format.rs
+++ b/src/mito2/src/sst/parquet/flat_format.rs
@@ -379,6 +379,12 @@ impl ParquetPrimaryKeyToFlat {
         column_ids: impl Iterator<Item = ColumnId>,
         skip_auto_convert: bool,
     ) -> ParquetPrimaryKeyToFlat {
+        assert!(if skip_auto_convert {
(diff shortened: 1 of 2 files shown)


src/mito2/src/sst/parquet/flat_format.rs
```diff
@@ -154,11 +156,18 @@ impl FlatReadFormat {
...
         let parquet_adapter = if is_legacy {
             // Safety: is_legacy_format() ensures primary_key is not empty.
-            ParquetAdapter::PrimaryKeyToFlat(ParquetPrimaryKeyToFlat::new(
-                metadata,
-                column_ids,
-                skip_auto_convert,
-            ))
+            if metadata.primary_key_encoding == PrimaryKeyEncoding::Sparse {
+                // Only skip auto convert when the primary key encoding is sparse.
+                ParquetAdapter::PrimaryKeyToFlat(ParquetPrimaryKeyToFlat::new(
```
discord9, line 161: maybe also safe guard and defensively check for primary_key_encoding inside ParquetPrimaryKeyToFlat::new as well?

PR #7057, opened 2025-10-09
feat: align influxdb line timestamp with table time index

MichaelScofield:

Sometimes the table is created before any influxdb lines were put, or the lines timestamp precision could change. Anyway, we should align the input to the table's schema.

## PR Checklist

### DIFF
--- a/src/frontend/src/instance/influxdb.rs
+++ b/src/frontend/src/instance/influxdb.rs
@@ -164,7 +164,7 @@ fn align_time_unit(value: &ValueData, target: TimeUnit) -> servers::error::Resul
         ValueData::TimestampNanosecondValue(x) => Timestamp::new_nanosecond(*x),
         _ => {
             return UnexpectedResultSnafu {
-                reason: format!("Timestamp value '{:?} is not of timestamp type!", value),
+                reason: format!("Timestamp value '{:?}' is not of timestamp type!", value),
             }
             .fail();
         }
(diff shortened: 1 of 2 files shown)


src/frontend/src/instance/influxdb.rs
```diff
@@ -64,3 +77,109 @@ impl InfluxdbLineProtocolHandler for Instance {
...
+        requests: RowInsertRequests,
+        query_context: &QueryContextRef,
+    ) -> servers::error::Result<RowInsertRequests> {
+        let mut inserts = requests.inserts;
+        for insert in inserts.iter_mut() {
+            let Some(rows) = &mut insert.rows else {
+                continue;
+            };
+
+            let Some(target_time_unit) = self
```
killme2008, line 99: I know the catalog manager caches data, but we might still need to fetch all table metadata upfront to avoid repeated calls.

Collect table names from `inserts` and retrieve table metadata initially.

MichaelScofield: no need, the influxdb lines are grouped into `RowInsertRequest`s by the table name, so the number of method calls of fetching table metadata are the same with or without collecting the table names upfront

sunng87:
just curious, as we can already auto detect timestamp column in #7046 , is it possible to do this in datanode? This feels like similar logic.

MichaelScofield:
@sunng87 Are you suggesting apply the same timestamp unit alignment to all protocols, not just for influxdb?

killme2008:
> @sunng87 Are you suggesting apply the same timestamp unit alignment to all protocols, not just for influxdb?

Don’t do that; it only applies to InfluxDB line protocol.

PR #7060, opened 2025-10-09
feat: table/column/flow COMMENT

waynexia:

- New clause `COMMENT ON [TABLE/COLUMN/FLOW]` to set/remove comment
- Display comment on corresponding `SHOW` clauses

## PR Checklist

### DIFF
--- a/src/common/meta/src/ddl/comment_on.rs
+++ b/src/common/meta/src/ddl/comment_on.rs
@@ -305,7 +305,7 @@ impl CommentOnProcedure {
         new_flow_info.comment = self.data.comment.clone().unwrap_or_default();
         new_flow_info.updated_time = Utc::now();
 
-        let raw_value = serde_json::to_vec(&new_flow_info).context(crate::error::SerdeJsonSnafu)?;
+        let raw_value = new_flow_info.try_as_raw_value()?;
 
         self.context
             .table_metadata_manager


src/operator/src/statement/comment.rs
```diff
@@ -0,0 +1,362 @@
...
+            CacheIdent::TableName(table_name),
+        ];
+        self.cache_invalidator
+            .invalidate(&Context::default(), &cache_ident)
+            .await
+            .context(InvalidateTableCacheSnafu)?;
+        Ok(())
+    }
+
+    async fn update_table_info(
```
MichaelScofield: Directly update the table metadata in the kvbackend, without using a procedure, seems dangerous. Will there be some concurrent modification issues on the table info? For example, multiple `COMMENT` statements are being executed at the same time on the same table, replacing the comments on each other. 

Other places I can think of could raise a conflict updating with here are the altering table procedure and the dropping table procedure. Essentially, you have to acquire the "lock" in the procedure to safely update the table info.

src/operator/src/statement/show.rs
```diff
@@ -120,7 +122,30 @@ impl StatementExecutor {
         table: TableRef,
         query_ctx: QueryContextRef,
     ) -> Result<Output> {
-        let table_info = table.table_info();
+        let mut table_info = table.table_info();
+        let partition_column_names: Vec<_> =
+            table_info.meta.partition_column_names().cloned().collect();
+
+        if let Some(latest) = self
```
MichaelScofield, line 129: Why is it necessary to retrieve the table info again here?

src/sql/src/parsers/comment_parser.rs
```diff
@@ -0,0 +1,187 @@
...
+        let _ = self.parser.next_token(); // consume COMMENT
+
+        if !self.parser.parse_keyword(Keyword::ON) {
+            return self.expected("ON", self.parser.peek_token());
+        }
+
+        let target_token = self.parser.next_token();
+        let comment = match target_token.token {
+            Token::Word(word) if word.keyword == Keyword::TABLE => {
+                let raw_table = self.parse_object_name().context(error::UnexpectedSnafu {
```
MichaelScofield:
```suggestion
                let raw_table = self.parse_object_name().with_context(|_| error::UnexpectedSnafu {
```

src/sql/src/parsers/comment_parser.rs
```diff
@@ -0,0 +1,187 @@
...
+                let table = Self::canonicalize_object_name(raw_table);
+                CommentObject::Table(table)
+            }
+            Token::Word(word) if word.keyword == Keyword::COLUMN => {
+                self.parse_column_comment_target()?
+            }
+            Token::Word(word)
+                if word.keyword == Keyword::NoKeyword && word.value.eq_ignore_ascii_case(FLOW) =>
+            {
+                let raw_flow = self.parse_object_name().context(error::UnexpectedSnafu {
```
MichaelScofield:
```suggestion
                let raw_flow = self.parse_object_name().with_context(|_| error::UnexpectedSnafu {
```

src/sql/src/parsers/comment_parser.rs
```diff
@@ -0,0 +1,187 @@
...
+        };
+
+        Ok(Statement::Comment(Comment {
+            object: comment,
+            comment: comment_value,
+        }))
+    }
+
+    fn parse_column_comment_target(&mut self) -> Result<CommentObject> {
+        let raw = self.parse_object_name().context(error::UnexpectedSnafu {
```
MichaelScofield:
```suggestion
        let raw = self.parse_object_name().with_context(|_| error::UnexpectedSnafu {
```

src/operator/src/statement/comment.rs
```diff
@@ -0,0 +1,362 @@
...
+            .context(TableMetadataManagerSnafu)?
+            .context(FlowNotFoundSnafu {
+                flow_name: &flow_name_str,
+            })?;
+
+        let mut new_flow_info = flow_info.get_inner_ref().clone();
+        new_flow_info.comment = comment.unwrap_or_default();
+        new_flow_info.updated_time = Utc::now();
+
+        self.update_flow_metadata(flow_id, new_flow_info).await?;
```
discord9: maybe here need to invalidity by flow id since flow info is updated?(And flow id being key to flow info)

src/operator/src/statement/comment.rs
```diff
@@ -0,0 +1,362 @@
...
+
+        Ok(())
+    }
+
+    async fn update_flow_metadata(
+        &self,
+        flow_id: FlowId,
+        new_flow_info: FlowInfoValue,
+    ) -> Result<()> {
+        let raw_value = serde_json::to_vec(&new_flow_info)
```
discord9: `new_flow_info.try_as_raw_value()` maybe?

src/operator/src/statement/comment.rs
discord9: or if procedure is too heavy, at least use `compare_and_put` and fail if compare fail(maybe could retry, idk)?

waynexia: Procedure looks heavy for this to me. But metasrv kvbackend doesn't support transaction, so I end up with this way. Or we should use a procedure as metasrv kvbackend is limited?

MichaelScofield: You can create a procedure for modifying the table info, not just for comment only. Then your procedure can be used for other purpose in the future, making it less heavy.

Or you can put the comment modification in alter table procedure.

src/operator/src/statement/show.rs
waynexia: For some reason I haven't figured out, table info from `table` sometimes lacks some info like partition

src/operator/src/statement/comment.rs
```diff
@@ -0,0 +1,362 @@
...
+    catalog: String,
+    schema: String,
+    table: String,
+    table_id: TableId,
+    table_name: TableName,
+    table_info: DeserializedValueWithBytes<TableInfoValue>,
+}
+
+impl StatementExecutor {
+    pub async fn comment(&self, stmt: Comment, query_ctx: QueryContextRef) -> Result<Output> {
```
killme2008, line 40:
```suggestion
    /// Adds a comment to a database object (table, column, or flow).
    ///
    /// # Arguments
    ///
    /// * `stmt`: A `Comment` struct containing the object to comment on and the comment text.
    /// * `query_ctx`: A `QueryContextRef` providing contextual information for the query.
    ///
    /// # Returns
    ///
    /// A `Result` containing the `Output` of the operation, or an error if the operation fails.
    pub async fn comment(&self, stmt: Comment, query_ctx: QueryContextRef) -> Result<Output> {
```

src/operator/src/statement/comment.rs
```diff
@@ -0,0 +1,362 @@
...
+    ) -> Result<Output> {
+        let ResolvedTable {
+            table_id,
+            table_name,
+            table_info,
+            ..
+        } = self.resolve_table_for_comment(&table, &query_ctx).await?;
+
+        let mut new_table_info = table_info.table_info.clone();
+        new_table_info.desc = comment;
```
killme2008: Not related to this PR; it just seems that our table metadata only has a created_on field and no updated_on.

src/operator/src/statement/comment.rs
killme2008: I think it’s another type of alter. Maybe we can handle it in the alter procedure.

src/sql/src/statements/comment.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use std::fmt::{self, Display, Formatter};
+
+use serde::Serialize;
+use sqlparser_derive::{Visit, VisitMut};
+
+use crate::ast::{Ident, ObjectName};
+
+#[derive(Debug, Clone, PartialEq, Eq, Visit, VisitMut, Serialize)]
+pub struct Comment {
```
killme2008, line 32: Better to document these pub strucutres.

src/operator/src/statement/comment.rs
waynexia: I (via claude) made a dedicated COMMENT ON procedure to replace current behavior. It only updates a fixed part of table info, so a general procedure for modifying table info would be a bit complex to express that "optional" on various info fields. And this can update table/column/flow in the same procedure, so the current alter procedure doesn't apply.

src/common/meta/src/ddl/comment_on.rs
```diff
@@ -0,0 +1,446 @@
...
+            })?
+            .table_id();
+
+        let table_info = self
+            .context
+            .table_metadata_manager
+            .table_info_manager()
+            .get(table_id)
+            .await?
+            .context(TableNotFoundSnafu {
```
MichaelScofield:
```suggestion
            .with_context(|| TableNotFoundSnafu {
```

src/common/meta/src/ddl/comment_on.rs
```diff
@@ -0,0 +1,446 @@
...
+                .table_info
+                .meta
+                .schema
+                .column_schemas
+                .iter()
+                .any(|col| &col.name == column_name);
+
+            ensure!(
+                column_exists,
+                UnexpectedSnafu {
```
MichaelScofield: I think it should be "column not found" error.

src/common/meta/src/ddl/comment_on.rs
```diff
@@ -0,0 +1,446 @@
...
+    pub fn from_json(json: &str, context: DdlContext) -> ProcedureResult<Self> {
+        let data = serde_json::from_str(json).context(FromJsonSnafu)?;
+
+        Ok(Self { context, data })
+    }
+
+    pub async fn on_prepare(&mut self) -> Result<Status> {
+        match self.data.object_type {
+            CommentObjectType::Table | CommentObjectType::Column => {
+                self.prepare_table_or_column().await?;
```
MichaelScofield, line 67: Better check whether the new and old comments are same in the preparation. And if they are same, skip to "finished succeeded" state of this procedure.

src/operator/src/statement/show.rs
MichaelScofield: better create an issue to track this

src/operator/src/statement/comment.rs
waynexia: 4d7d474

PR #7064, opened 2025-10-10
feat: add index cache eviction support

zhongzc:

Add `CacheManager::evict_puffin_cache` to purge bloom, inverted, result, metadata, and write caches for a rebuilt puffin file.

## PR Checklist

### DIFF
--- a/src/mito2/src/cache.rs
+++ b/src/mito2/src/cache.rs
@@ -170,6 +170,19 @@ impl CacheStrategy {
         }
     }
 
+    /// Calls [CacheManager::evict_puffin_cache()].
+    pub async fn evict_puffin_cache(&self, file_id: RegionFileId) {
+        match self {
+            CacheStrategy::EnableAll(cache_manager) => {
+                cache_manager.evict_puffin_cache(file_id).await
+            }
+            CacheStrategy::Compaction(cache_manager) => {
+                cache_manager.evict_puffin_cache(file_id).await
+            }
+            CacheStrategy::Disabled => {}
+        }
+    }
+
     /// Calls [CacheManager::get_selector_result()].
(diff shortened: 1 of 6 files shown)


src/mito2/src/cache.rs
```diff
@@ -374,6 +387,35 @@ impl CacheManager {
...
+            cache.invalidate_file(file_id.file_id());
+        }
+
+        if let Some(cache) = &self.inverted_index_cache {
+            cache.invalidate_file(file_id.file_id());
+        }
+
+        if let Some(cache) = &self.index_result_cache {
+            cache.invalidate_file(file_id.file_id());
+        }
```
evenyag, line 402: Does moka cache invalidate entries immediately or in background?

waynexia: moka [says](https://docs.rs/moka/latest/moka/future/struct.Cache.html#method.invalidate_entries_if) this call only set a marker, and the real removal happens in a later maintainance task:

>This method returns immediately by not actually removing the invalidated entries. Instead, it just sets the predicate to the cache with the time when this method was called. The actual removal of the invalidated entries is done as a maintenance task driven by a user thread. For more details, see [the Maintenance Tasks section](https://docs.rs/moka/latest/moka/index.html#maintenance-tasks) in the crate level documentation.

evenyag: @zhongzc Is that acceptable?

zhongzc: > Also the `get` and other retrieval methods will apply the closure to a cached entry to determine if it should have been invalidated. Therefore, it is guaranteed that these methods must not return invalidated values.

It's acceptable.

PR #7065, opened 2025-10-10
fix: Separate admin and normal function registries.

### LINKED ISSUE #7059: `Admin` statement can find normal(not admin) functions
### What type of bug is this?

User Experience

Query Engine

### Minimal reproduce step

Just try executing some functions that are not admin. For example, `admin mod(1, 2)`

should report a "not found" error like "'mod' admin function not found"

### What did you see instead?

"ERROR 1210 (HY000): (InvalidArguments): Failed to build admin function args: Function mod is not async"

macos

main

```bash

```

Shyamnatesan:

Fixes #7059 

1. In my previous PR (https://github.com/GreptimeTeam/greptimedb/pull/7061), I added a check to determine whether a function is an admin function. However, that check did not cover all cases.
2. Introduced a new `ADMIN_FUNCTION_REGISTRY` that registers **only admin functions**. The `execute_admin_command()` function now checks this registry instead of `FUNCTION_REGISTRY`, ensuring a clear separation between normal and admin functions.
3. All changes from the previous PR have been reverted, since this new approach fully addresses the issue.

## Implementation Details
- Added `ADMIN_FUNCTION_REGISTRY` as a dedicated registry for admin functions.
- Updated `execute_admin_command()` to use the new registry.
- Removed previous logic that attempted to detect admin functions from the mixed `FUNCTION_REGISTRY` inside `execute_admin_command()` .

## PR Checklist

### DIFF
--- a/src/common/function/src/function_registry.rs
+++ b/src/common/function/src/function_registry.rs
@@ -147,9 +147,8 @@ pub static FUNCTION_REGISTRY: LazyLock<Arc<FunctionRegistry>> = LazyLock::new(||
     MatchesFunction::register(&function_registry);
     MatchesTermFunction::register(&function_registry);
 
-    // System and administration functions
+    // System functions
     SystemFunction::register(&function_registry);
-    AdminFunction::register(&function_registry);
 
     // Json related functions
     JsonFunction::register(&function_registry);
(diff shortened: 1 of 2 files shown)


MichaelScofield:
Good to know the failed tests revealed two other hidden bugs: 1. `SELECT` can invoke admin functions, which might not be a wanted feature if we are to restrict the permissions in the future; 2. the `precedure_state` function is mistakenly registered as a system function. Can you fix those as well? @Shyamnatesan

killme2008:
> Good to know the failed tests revealed two other hidden bugs: 1. `SELECT` can invoke admin functions, which might not be a wanted feature if we are to restrict the permissions in the future; 2. the `precedure_state` function is mistakenly registered as a system function. Can you fix those as well? @Shyamnatesan

Permission isn't guaranteed by a separate registry but is checked within the `admin_fn` macro.

As i said in https://github.com/GreptimeTeam/greptimedb/issues/7059#issuecomment-3388457967, let's don't make things complex.

PR #7066, opened 2025-10-10
refactor: restructure sqlness to support multiple envs and extract common utils

WenyXu:

This PR refactors the sqlness runner to support for kubernetes-based testing env.

### Extracted Common Utilities 
- Added client.rs: Implements MultiProtocolClient that supports gRPC, PostgreSQL, and MySQL protocols
- Added formatter.rs: Provides formatters for different output types:
- ErrorFormatter for error display
- OutputFormatter for record batches and affected rows
- PostgresqlFormatter for PostgreSQL query results
- MysqlFormatter for MySQL query results

### Restructured Test Runner (tests/runner/src/)
- Command Layer (cmd.rs, cmd/bare.rs, cmd/kube.rs):
  - Introduced subcommand structure with bare and kube modes
  - BareCommand: Runs tests against bare metal/local GreptimeDB instances
  - KubeCommand: Runs tests against Kubernetes-deployed GreptimeDB clusters
  - Shared SqlnessConfig for common configuration options
- Environment Layer (env/bare.rs, env/kube.rs):
  - Split original env.rs into separate bare and Kubernetes implementations
  - env/bare.rs: Contains the original standalone/distributed test logic 
  - env/kube.rs: New Kubernetes environment implementation with:

### Breaking Changes
The command-line interface now requires specifying a subcommand:
- Before: `sqlness-runner --case-dir tests/cases`
- After: `sqlness-runner bare --case-dir tests/cases (or kube --case-dir tests/cases)`

## PR Checklist

### DIFF
--- a/tests/runner/src/cmd/bare.rs
+++ b/tests/runner/src/cmd/bare.rs
@@ -32,7 +32,6 @@ enum Wal {
 
 // add a group to ensure that all server addresses are set together
 #[derive(clap::Args, Debug, Clone, Default)]
-#[group(multiple = true, requires_all=["server_addr", "pg_server_addr", "mysql_server_addr"])]
 pub(crate) struct ServerAddr {
     /// Address of the grpc server.
     #[clap(short, long)]
@@ -199,6 +198,7 @@ impl BareCommand {
                 println!("Stopping etcd");
                 util::stop_rm_etcd();
             }
+            // TODO(weny): remove postgre and mysql containers
             println!("Removing state in {:?}", sqlness_home);
             tokio::fs::remove_dir_all(sqlness_home).await.unwrap();
         }


tests/runner/src/cmd/bare.rs
```diff
@@ -0,0 +1,206 @@
...
+#[derive(ValueEnum, Debug, Clone)]
+#[clap(rename_all = "snake_case")]
+enum Wal {
+    RaftEngine,
+    Kafka,
+}
+
+// add a group to ensure that all server addresses are set together
+#[derive(clap::Args, Debug, Clone, Default)]
+#[group(multiple = true, requires_all=["server_addr", "pg_server_addr", "mysql_server_addr"])]
```
fengjiachun: It seems that we only need to ensure that if `pg/mysql` is set, `server_addr` must be required, but why do we need `requires_all`?

tests/runner/src/cmd/bare.rs
```diff
@@ -0,0 +1,206 @@
...
+
+        // clean up and exit
+        if !self.preserve_state {
+            if self.setup_etcd {
+                println!("Stopping etcd");
+                util::stop_rm_etcd();
+            }
+            println!("Removing state in {:?}", sqlness_home);
+            tokio::fs::remove_dir_all(sqlness_home).await.unwrap();
+        }
```
fengjiachun, line 204: Do we need to provide `stop_rm_mssql` and `stop_rm_pg`?

tests/runner/src/cmd/bare.rs
```diff
@@ -0,0 +1,206 @@
...
+    extra_args: Vec<String>,
+}
+
+impl BareCommand {
+    pub async fn run(mut self) {
+        let temp_dir = tempfile::Builder::new()
+            .prefix("sqlness")
+            .tempdir()
+            .unwrap();
+        let sqlness_home = temp_dir.keep();
```
fengjiachun, line 113: The keep() function will prevent automatic cleanup. If a panic occurs or the process exits early, the directory will leak. Can we determine whether to keep it at the end of this method?

WenyXu: I think it's okay to leak the temp directory, as it will be delete when the host restarts

tests/runner/src/cmd/bare.rs
WenyXu: Left a TODO here, maybe these dependencies should be managed externally.

tests/runner/src/cmd/bare.rs
WenyXu: Removed in [26dfd56](https://github.com/GreptimeTeam/greptimedb/pull/7066/commits/26dfd568a92bf2236595440b97f7565f60bf7aac)

PR #7072, opened 2025-10-11
feat: add updated_on to tablemeta with a default of created_on

### LINKED ISSUE #7070: Added updated_on field to table metadata

API improvement, User experience

The TableMeta and RawTableMeta structs include a `created_on` field for the table creation timestamp but lack an `updated_on` field for modification timestamps.

https://greptimedb.rs/table/metadata/struct.TableMeta.html
https://greptimedb.rs/table/metadata/struct.RawTableMeta.html

I think we can add an `updated_on` field, update it on table alterations, and add it in `information_schema.tables`.

### Implementation challenges

- [ ] Add `updated_on` to `mdatadata` with a default of `created_on`.
- [ ] Update `updated_on` in the alter procedure.
- [ ] Include `updated_on` in `information_schema.tables`.

Standing-Man:

- Closes #7070.

- [x] Add `updated_on` to `mdatadata` with a default of `created_on`.
- [x] Update `updated_on` in the alter procedure.
- [x] Include `updated_on` in `information_schema.tables`.

## PR Checklist

### DIFF
--- a/src/common/meta/src/key/table_info.rs
+++ b/src/common/meta/src/key/table_info.rs
@@ -289,6 +289,7 @@ mod tests {
     fn test_deserialization_compatibility() {
         let s = r#"{"version":1,"table_info":{"ident":{"table_id":8714,"version":0},"name":"go_gc_duration_seconds","desc":"Created on insertion","catalog_name":"e87lehzy63d4cloud_docs_test","schema_name":"public","meta":{"schema":{"column_schemas":[{"name":"instance","data_type":{"String":null},"is_nullable":true,"is_time_index":false,"default_constraint":null,"metadata":{}},{"name":"job","data_type":{"String":null},"is_nullable":true,"is_time_index":false,"default_constraint":null,"metadata":{}},{"name":"quantile","data_type":{"String":null},"is_nullable":true,"is_time_index":false,"default_constraint":null,"metadata":{}},{"name":"greptime_timestamp","data_type":{"Timestamp":{"Millisecond":null}},"is_nullable":false,"is_time_index":true,"default_constraint":null,"metadata":{"greptime:time_index":"true"}},{"name":"greptime_value","data_type":{"Float64":{}},"is_nullable":true,"is_time_index":false,"default_constraint":null,"metadata":{}}],"timestamp_index":3,"version":0},"primary_key_indices":[0,1,2],"value_indices":[],"engine":"mito","next_column_id":5,"region_numbers":[],"engine_options":{},"options":{"write_buffer_size":null,"ttl":null,"extra_options":{}},"created_on":"1970-01-01T00:00:00Z"},"table_type":"Base"}}"#;
         let v = TableInfoValue::try_from_raw_value(s.as_bytes()).unwrap();
+        assert_eq!(v.table_info.meta.created_on, v.table_info.meta.updated_on);
         assert!(v.table_info.meta.partition_key_indices.is_empty());
     }
 
--- a/src/table/src/metadata.rs
+++ b/src/table/src/metadata.rs
@@ -26,7 +26,7 @@ use datatypes::schema::{
     SkippingIndexOptions,
 };
 use derive_builder::Builder;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use snafu::{OptionExt, ResultExt, ensure};
 use store_api::metric_engine_consts::PHYSICAL_TABLE_METADATA_KEY;
 use store_api::mito_engine_options::{COMPACTION_TYPE, COMPACTION_TYPE_TWCS};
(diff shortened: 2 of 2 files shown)


Standing-Man:
Hi @MichaelScofield and @evenyag, I updated `update_on` in the `ALTER PROCEDURE` based on the operations listed in [`builder_with_alter_kind`](https://greptimedb.rs/src/table/metadata.rs.html#241-261). Could you please confirm if this correctly covers all relevant `alter` operations?

src/common/meta/src/key/table_info.rs
```diff
@@ -287,7 +287,7 @@ mod tests {
...
...sion":0},"name":"go_gc_duration_seconds","desc":"Created on insertion","catalog_name":"e87lehzy63d4cloud_docs_test","schema_name":"public","meta":{"schema":{"column_schemas":[{"name":"instance","data_type":{"String":null},"is_nullable":true,"is_time_index":false,"default_constraint":null,"metadata":{}},{"name":"job","data_type":{"String":null},"is_nullable":true,"is_time_index":false,"default_constraint":null,"metadata":{}},{"name":"quantile","data_type":{"String":null},"is_nullable":true,"is_time_index":false,"default_constraint":null,"metadata":{}},{"name":"greptime_timestamp","data_type":{"Timestamp":{"Millisecond":null}},"is_nullable":false,"is_time_index":true,"default_constraint":null,"metadata":{"greptime:time_index":"true"}},{"name":"greptime_value","data_type":{"Float64":{}},"is_nullable":true,"is_time_index":false,"default_constraint":null,"metadata":{}}],"timestamp_index":3,"version":0},"primary_key_indices":[0,1,2],"value_indices":[],"engine":"mito","next_column_id":5,"region_numbers":[],"engine_options":{},"options":{"write_buffer_size":null,"ttl":null,"extra_options":{}},"created_on":"1970-01-01T00:00:00Z", "updated_on":"1970-01-01T00:00:00Z"},"table_type":"Base"}}"#;
```
killme2008: We should test the original metadata, which lacks `updated_on`, to ensure it deserializes correctly as null or defaults to `created_on`.

src/table/src/metadata.rs
```diff
@@ -287,6 +295,8 @@ impl TableMeta {
         let mut builder = self.new_meta_builder();
         builder.options(new_options);
 
+        let _ = builder.updated_on(Utc::now());
```
killme2008: We can move this into `builder_with_alter_kind`, which will handle the dispatch to the functions below, so we don’t need to repeat this code elsewhere.

Standing-Man: I’m a bit confused about where it’s appropriate to handle the table_meta update. we had test case for `ALTER TABLE` on new commit, but found that `update_on` was not updated.

src/common/meta/src/key/table_info.rs
Standing-Man: Thanks for the review! I added a custom deserializer for TableMeta, and the test that the original metadata locks `update_on` are passing now.

src/table/src/metadata.rs
killme2008: Thanks. Let me run the test locally and try to figure out.

src/common/meta/src/key/table_info.rs
killme2008: Looks like we don't need a custom deserializer. Just set this field with `#[serde(default)]`?

src/table/src/metadata.rs
killme2008: I found the cause: the `update_time` uses `ConcreteDataType::timestamp_second_datatype()`, which has second-level precision. Because the CREATE TABLE and ALTER TABLE operations ran too closely together, the timestamp didn’t update. Adding a delay of more than one second before the ALTER TABLE makes the test pass.

```rust
    std::thread::sleep(std::time::Duration::from_millis(1100));
```

Another choice is to change the data type of `create_time` and `update_time` in `tables.rs`.

But I think sleep is needed too.

Standing-Man: Awesome, thank you so much for your suggestion! I’ve updated the test cases for `ALTER TABLE` on new commit.

src/common/meta/src/key/table_info.rs
Standing-Man: Right, `#[serde(default)]` ensures a default value is set when update_on is absent in the input data.

killme2008: @Standing-Man I realized your original implementation (with the helper) is correct since `updated_on` should use the created time when absent, not the default `DateTime` value. Could you please revert that change? Also, please add an assertion for `updated_on` in `test_deserialization_compatibility` to ensure it deserializes to the `created_on` value.

Standing-Man: Got it.

Standing-Man:
Thanks to all reviewers for your great feedback!

PR #7076, opened 2025-10-13
chore: update datafusion to 50

MichaelScofield:

just trying to keep up with our dependency

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/commutativity.rs
+++ b/src/query/src/dist_plan/commutativity.rs
@@ -97,7 +97,7 @@ impl Categorizer {
         // Subquery is treated separately in `inspect_plan_with_subquery`. To avoid rewrite the
         // "maybe rewritten" plan, stop the check here.
         if has_subquery(plan)? {
-            return Ok(Commutativity::NonCommutative);
+            return Ok(Commutativity::Unimplemented);
         }
 
         let partition_cols = partition_cols.unwrap_or_default();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1804,7 +1804,6 @@ dependencies = [
  "session",
  "snafu 0.8.6",
  "store-api",
- "substrait 0.18.0",
  "table",
  "tempfile",
  "tokio",
@@ -1949,7 +1948,6 @@ dependencies = [
  "snafu 0.8.6",
  "standalone",
  "store-api",
- "substrait 0.18.0",
  "table",
  "temp-env",
  "tempfile",
@@ -2027,6 +2025,7 @@ dependencies = [
  "common-base",
  "common-error",
  "common-macro",
+ "common-stat",
  "common-telemetry",
  "common-test-util",
  "common-wal",
@@ -2034,13 +2033,11 @@ dependencies = [
  "datanode",
  "humantime-serde",
  "meta-client",
- "num_cpus",
  "object-store",
  "serde",
  "serde_json",
  "serde_with",
  "snafu 0.8.6",
- "sysinfo",
  "temp-env",
  "tempfile",
  "toml 0.8.23",
@@ -2475,6 +2472,7 @@ dependencies = [
  "common-macro",
  "common-telemetry",
  "common-time",
+ "criterion 0.7.0",
  "datafusion",
  "datafusion-common",
  "datatypes",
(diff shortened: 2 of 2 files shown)


src/catalog/src/system_schema/pg_catalog.rs
```diff
@@ -26,8 +26,9 @@ use datafusion::datasource::TableType;
 use datafusion::error::DataFusionError;
 use datafusion::execution::TaskContext;
 use datafusion::physical_plan::stream::RecordBatchStreamAdapter as DfRecordBatchStreamAdapter;
-use datafusion_pg_catalog::pg_catalog::catalog_info::CatalogInfo;
-use datafusion_pg_catalog::pg_catalog::{
+use datafusion_postgres::pg_catalog::catalog_info::CatalogInfo;
+use datafusion_postgres::pg_catalog::context::EmptyContextProvider;
+use datafusion_postgres::pg_catalog::{
```
sunng87: let me deal with this part

src/query/src/dist_plan/commutativity.rs
evenyag, line 1: @discord9 Please check this change

tests/cases/standalone/common/tql/tql-cte.result
```diff
@@ -775,8 +775,8 @@ LIMIT 5;
 |               | SubqueryAlias: l                                                                                                                                                       |
 |               |   TableScan: labels                                                                                                                                                    |
 |               | ]]                                                                                                                                                                     |
-| physical_plan | SortPreservingMergeExec: [ts@0 ASC NULLS LAST, host@2 ASC NULLS LAST, avg_value@1 ASC NULLS LAST], fetch=5                                                             |
```
evenyag, line 778: Is this expected?

MichaelScofield: @waynexia can you help confirm this?

evenyag: The sql has `ORDER BY t.ts, l.host, avg_value`. But the physical plan doesn't have this. cc @discord9

discord9: > The sql has `ORDER BY t.ts, l.host, avg_value`. But the physical plan doesn't have this. cc @discord9

seems right, the optimizer get smarter and optimize out `order by l.host` because `WHERE l.host = 'host1'` exist

src/query/src/dist_plan/commutativity.rs
discord9: > @discord9 Please check this change

lgtm

src/query/src/dist_plan/commutativity.rs
```diff
@@ -93,6 +94,12 @@ impl Categorizer {
         plan: &LogicalPlan,
         partition_cols: Option<AliasMapping>,
     ) -> DfResult<Commutativity> {
+        // Subquery is treated separately in `inspect_plan_with_subquery`. To avoid rewrite the
+        // "maybe rewritten" plan, stop the check here.
+        if has_subquery(plan)? {
+            return Ok(Commutativity::NonCommutative);
```
discord9: return UnImplement here might be more in line with below's `LogicalPlan::Subquery(_) =>Commutativity::Unimplemented`

src/servers/src/http/handler.rs
```diff
@@ -208,7 +208,7 @@ pub async fn sql_format(
 
     let mut parts: Vec<String> = Vec::with_capacity(stmts.len());
     for stmt in stmts {
-        let mut s = format!("{:#}", stmt);
```
discord9, line 211: will this cause different display result due to [alternate](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.alternate) become false?

MichaelScofield: No, instead, the original "{:#}" will cause the difference.

discord9: > No, instead, the original "{:#}" will cause the difference.

but isn't that sql format api so have pretty format is expected behavior?
https://docs.rs/sqlparser/0.58.0/sqlparser/index.html#pretty-printing

MichaelScofield: Pretty format was not expected. Actually the "pretty" was not working until this sqlparser version (v0.58.0): https://github.com/apache/datafusion-sqlparser-rs/pull/1847

Pretty format now will display too many new lines, breaking the test "test_sql_format_api". So I think it's better to keep the behavior.

PR #7083, opened 2025-10-13
feat: add Value::Json value type

sunng87:

This patch add a new variant in `Value` as discussed in previous patch, for marking a physical `Value` as logical json type.

In `ConcreteDataType`, I'm reusing its current `json` type, but introduce a new format called `Native`. Since the implementation of `Native` is quite different from original `Jsonb`, I try to avoid using `match`es in Jsonb specific functions. The `Native` variant also holds inner type information.

**There is no conversion from ScalarValue to `Value::Json` for lack of context information**.

**There is no vector for json.**

Protobuf values has been a big part too. `JsonValue` is added to column and row based enumeration, and with type of `Value` itself. A new datatype extension is added: `JsonNativeType` to hold inner type information. I tried to implement all conversion functions if I didn't miss any.

Prerequisite: https://github.com/GreptimeTeam/greptime-proto/pull/281

## PR Checklist

### DIFF
--- a/src/datatypes/src/value.rs
+++ b/src/datatypes/src/value.rs
@@ -1306,7 +1306,11 @@ impl<'a> ValueRef<'a> {
 
     /// Returns true if this is null.
     pub fn is_null(&self) -> bool {
-        matches!(self, ValueRef::Null)
+        match self {
+            ValueRef::Null => true,
+            ValueRef::Json(v) => v.is_null(),
+            _ => false,
+        }
     }
 
     /// Cast itself to binary slice.


MichaelScofield (approved):

src/datatypes/src/value.rs
```diff
@@ -81,8 +81,12 @@ pub enum Value {
     IntervalDayTime(IntervalDayTime),
     IntervalMonthDayNano(IntervalMonthDayNano),
 
+    // Collection types:
     List(ListValue),
     Struct(StructValue),
+
+    // Json Logical types:
+    Json(Box<Value>),
```
fengjiachun, line 89: ValueRef::is_null() still only checks for the plain Null variant. With the new native JSON support a database NULL is represented as ValueRef::Json(Box::new(ValueRef::Null)), so is_null() now returns false for JSON nulls. Is it by design?

PR #7090, opened 2025-10-14
fix: part cols not in projection

discord9:

as title, fix a problem when `projection` cause table scan's output schema doesn't actually contain the partition columns, this case should be considered acceptable not as a hard error

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/analyzer.rs
+++ b/src/query/src/dist_plan/analyzer.rs
@@ -546,7 +546,7 @@ impl PlanRewriter {
                                     let index =
                                         if let Some(c) = plan.schema().index_of_column_by_name(None, &c){
                                             c
-                                        }else{
+                                        } else {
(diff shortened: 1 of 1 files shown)


src/query/src/dist_plan/analyzer.rs
```diff
@@ -544,13 +544,13 @@ impl PlanRewriter {
...
-                                        plan.schema().index_of_column_by_name(None, &c).ok_or_else(|| {
-                                            datafusion_common::DataFusionError::Internal(
-                                                format!(
-                                                    "PlanRewriter: maybe_set_partitions: column {c} not found in schema of plan: {plan}"
-                                                ),
-                                            )
-                                        })?;
+                                        if let Some(c) = plan.schema().index_of_column_by_name(None, &c){
+                                            c
+                                        }else{
```
evenyag:
```suggestion
                                        } else {
```

fengjiachun (approved):
LGTM

PR #7092, opened 2025-10-14
feat: writer mem limiter for http and grpc service

fengjiachun:

Add memory limiting for concurrent HTTP and gRPC requests to prevent unbounded memory growth from large concurrent write operations.

## PR Checklist

### DIFF
--- a/src/servers/src/error.rs
+++ b/src/servers/src/error.rs
@@ -20,6 +20,7 @@ use axum::http::StatusCode as HttpStatusCode;
 use axum::response::{IntoResponse, Response};
 use axum::{Json, http};
 use base64::DecodeError;
+use common_base::readable_size::ReadableSize;
 use common_error::define_into_tonic_status;
 use common_error::ext::{BoxedError, ErrorExt};
 use common_error::status_code::StatusCode;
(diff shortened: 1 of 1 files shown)


config/frontend.example.toml
```diff
@@ -31,6 +31,10 @@ timeout = "0s"
 ## The following units are supported: `B`, `KB`, `KiB`, `MB`, `MiB`, `GB`, `GiB`, `TB`, `TiB`, `PB`, `PiB`.
 ## Set to 0 to disable limit.
 body_limit = "64MB"
+## Maximum total memory for all concurrent HTTP request bodies.
+## Set to 0 to disable the limit. Default: "0" (unlimited)
```
MichaelScofield, line 35: better set a reasonable default, like 100MB, for better out-of-the-box user experience

src/servers/src/request_limiter.rs
```diff
@@ -0,0 +1,213 @@
...
+        // Acquire exactly the limit
+        let _guard = limiter.try_acquire(1000).unwrap();
+        assert_eq!(limiter.current_usage(), 1000);
+
+        // Try to acquire 1 more byte - should fail
+        let result = limiter.try_acquire(1);
+        assert!(result.is_err());
+    }
+
+    #[tokio::test]
```
MichaelScofield:
```suggestion
    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
```

config/frontend.example.toml
fengjiachun: Providing a default number is challenging right now because the difference between protocol compression and uncompression can be substantial. For now, we must rely on user decisions; users can monitor metrics to determine the appropriate value for their system.

PR #7103, opened 2025-10-16
feat: new create table syntax for new json datatype

MichaelScofield:

please see the test in the new `json.rs`

## PR Checklist

### DIFF
--- a/tests/cases/standalone/common/alter/change_col_skipping_options.result
+++ b/tests/cases/standalone/common/alter/change_col_skipping_options.result
@@ -323,7 +323,7 @@ Error: 1002(Unexpected), Invalid skipping index option: Invalid false positive r
 
 ALTER TABLE test MODIFY COLUMN value SET SKIPPING INDEX WITH(granularity = 1024, type = 'BLOOM', false_positive_rate = -0.01);
 
-Error: 1004(InvalidArguments), Unrecognized table option key: false_positive_rate, value: -0.01
+Error: 1004(InvalidArguments), Invalid expr as option value, error: -0.01 not accepted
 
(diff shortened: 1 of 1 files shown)


waynexia (approved):
grammar LGTM, we can tune specific keys later.

src/sql/src/util.rs
```diff
@@ -42,29 +47,105 @@ pub fn format_raw_object_name(name: &ObjectName) -> String {
...
+        ensure!(
+            matches!(expr, Expr::Value(_) | Expr::Identifier(_) | Expr::Array(_)),
+            InvalidExprAsOptionValueSnafu {
+                error: format!("{expr} not accepted")
+            }
+        );
+        Ok(Self(expr))
+    }
+
+    fn expr_as_string(expr: &Expr) -> Option<&str> {
```
sunng87, line 64: option value is not always string, there is case for int literal like granularity in full text index

src/sql/src/util.rs
```diff
@@ -42,29 +47,105 @@ pub fn format_raw_object_name(name: &ObjectName) -> String {
...
-            ..
-        }) => v,
-        Expr::Identifier(v) => v.value,
-        Expr::Value(ValueWithSpan {
-            value: Value::Number(v, _),
-            ..
-        }) => v.clone(),
-        value => return InvalidTableOptionValueSnafu { key, value }.fail(),
-    };
+    let v = OptionValue::try_new(value)?;
```
sunng87, line 141: To match previous behavior, we'd better do `expr_as_string` in `try_new` and return error when expr not matched.

PR #7105, opened 2025-10-17
feat: introduce the Noop WAL provider for datanode

### LINKED ISSUE #7100: Introduce the Noop WAL provider for datanode

When the configured WAL provider (such as Kafka) becomes temporarily unavailable — for example, due to a cluster outage — datanodes currently fail to write WAL records and may become partially non-functional. This causes interruptions to data ingestion.

### What does the feature do?

The Noop WAL provider acts as a fallback WAL implementation that performs no actual persistence but allows datanodes to continue operating normally. With this feature, datanodes can temporarily bypass WAL writes and resume normal WAL-based operations once the primary WAL provider becomes available again.

### Implementation challenges

_No response_

WenyXu:

close #7100

This PR introduces a Noop WAL provider for the datanode. It is designed for situations where the configured WAL provider becomes temporarily unavailable.

In a GreptimeDB cluster, there are two places where the WAL provider should be configured:
- Metasrv — responsible for generating WAL provider metadata for new regions.
- Datanode — responsible for reading and writing WAL data.

This means that when the WAL provider is temporarily unavailable, users only need to configure the datanode’s WAL provider as Noop to keep the cluster functional.

## PR Checklist

### DIFF
--- a/src/log-store/src/noop/log_store.rs
+++ b/src/log-store/src/noop/log_store.rs
@@ -90,3 +90,27 @@ impl LogStore for NoopLogStore {
         Ok(0)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_append_batch() {
+        let log_store = NoopLogStore;
+        let entries = vec![Entry::Naive(NaiveEntry {
+            provider: Provider::noop_provider(),
+            region_id: RegionId::new(1, 1),
+            entry_id: 1,
+            data: vec![1],
+        })];
+
+        let last_entry_ids = log_store
+            .append_batch(entries)
+            .await
+            .unwrap()
+            .last_entry_ids;
+        assert_eq!(last_entry_ids.len(), 1);
+        assert_eq!(last_entry_ids[&(RegionId::new(1, 1))], 0);
+    }
+}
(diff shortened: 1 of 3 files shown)


src/log-store/src/noop/log_store.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    async fn stop(&self) -> Result<()> {
+        Ok(())
+    }
+
+    async fn append_batch(&self, entries: Vec<Entry>) -> Result<AppendBatchResponse> {
+        let last_entry_ids = entries
+            .iter()
+            .map(|entry| (entry.region_id(), 0))
+            .collect::<HashMap<RegionId, EntryId>>();
```
fengjiachun, line 40: Is this necessary, or can it simply return an empty collection?

fengjiachun (approved):

src/log-store/src/noop/log_store.rs
WenyXu: The return map will be unwrapped by the upper layer.
```
                        // Safety: the log store implementation ensures that either the `write_to_wal` fails and no
                        // response is returned or the last entry ids for each region do exist.
                        let last_entry_id = response.last_entry_ids.get(region_id).unwrap();
                        region_ctx.set_next_entry_id(last_entry_id + 1);
```

MichaelScofield (approved):

Issue #1168, opened 2023-03-14
Migrates more tests from duckdb


killme2008:
### What type of enhancement is this?

Tech debt reduction, Other

### What does the enhancement do?

We already migrate [some tests](https://github.com/GreptimeTeam/greptimedb/tree/develop/tests) from duckdb's SQL test: https://github.com/duckdb/duckdb/tree/master/test/sql. And it already helps us figure out some bugs such as #790 #791 #759  etc.

But there are still many tests that are not migrated. Let's create an issue to track them:

* [ ] [join](https://github.com/duckdb/duckdb/tree/master/test/sql/join)
* [ ] [[select]()](https://github.com/duckdb/duckdb/tree/master/test/sql/select)
* [ ] [subquery](https://github.com/duckdb/duckdb/tree/master/test/sql/subquery)
* [ ] [tpch](https://github.com/duckdb/duckdb/tree/master/test/sql/tpch)
* [ ] [copy](https://github.com/duckdb/duckdb/tree/master/test/sql/copy)
* [ ] [prepared](https://github.com/duckdb/duckdb/tree/master/test/sql/prepared)
* [ ] [parser](https://github.com/duckdb/duckdb/tree/master/test/sql/parser)

I just list the main tests. Of course, our current tests still miss many cases too, they need to be migrated too.

WenyXu:
Could we introduce the [Sqllogictest-rs](https://github.com/risinglightdb/sqllogictest-rs) framework to perform tests?

killme2008:
We already have sqlness https://docs.greptime.com/developer-guide/tests/sqlness-test

WenyXu:
> We already have sqlness https://docs.greptime.com/developer-guide/tests/sqlness-test

I see. I thought we could utilize the existing sqllogictest test files, but we still need to modify them.

Gump9:
Hello!I'm new to here.Can I have a try to migrate select tests?What test cases can I refer to that have been migrated? where are they?

MichaelScofield:
@Gump9 welcome! The test cases for select are [here](https://github.com/duckdb/duckdb/tree/master/test/sql/select)

Gump9:
> @Gump9 welcome! The test cases for select are [here](https://github.com/duckdb/duckdb/tree/master/test/sql/select)

tks a lot

Gump9:
[...]

PR #7107, opened 2025-10-17
feat: adds regex_extract function and more type tests

killme2008:

#1168

Main changes:

* Migrated type and related function tests from DuckDB.
* Implemented `regex_extract` functions.
* Added `Date64` support for `date_format`.

Remaining issues:
1. DataFusion does not correctly support sampling: https://github.com/apache/datafusion/issues/13563  
2. Subqueries are not supported, for example:

```sql
SELECT a, b,
    a - (SELECT MIN(a) FROM test WHERE a IS NOT NULL) AS diff_from_min
FROM test
WHERE a IS NOT NULL
ORDER BY a - (SELECT MIN(a) FROM test WHERE a IS NOT NULL);

Error: 1001 (Unsupported), This feature is not implemented: Physical plan does not support logical expression ScalarSubquery(<subquery>)
```

TODO:

* Contribute `format(fmt, arg1, arg2...)` and `regex_extract` functions to Datafusion and remove our custom implementations.

## PR Checklist


tests/cases/standalone/common/order/order_by_expressions.result
```diff
@@ -0,0 +1,137 @@
...
++---+----+----------+
+
+-- ORDER BY with subqueries in expressions
+SELECT a, b,
+    a - (SELECT MIN(a) FROM test WHERE a IS NOT NULL) as diff_from_min
+FROM test
+WHERE a IS NOT NULL
+ORDER BY a - (SELECT MIN(a) FROM test WHERE a IS NOT NULL);
+
+Error: 1001(Unsupported), This feature is not implemented: Physical plan does not support logical expression ScalarSubquery(<subquery>)
```
killme2008, line 132: It's known issue.

src/common/function/src/scalars/string/format.rs
```diff
@@ -0,0 +1,525 @@
...
+        if let Some(name) = name_opt
+            && needed_names.contains(name)
+        {
+            if i + 1 >= values.len() {
+                return Err(DataFusionError::Execution(format!(
+                    "FORMAT: named argument '{}' has no corresponding value",
+                    name
+                )));
+            }
+            named_map.insert(name.clone(), &values[i + 1]);
```
evenyag: What if the positional value has the same value as the needed_names? We may not be able to distinguish that because the needed_names only contains the name, without the position of the format name.

src/common/function/src/scalars/string/format.rs
```diff
@@ -0,0 +1,525 @@
...
+                    p += 2;
+                    continue;
+                }
+                let start = p + 1;
+                let mut q = start;
+                while q < bytes.len() && bytes[q] != b'}' {
+                    q += 1;
+                }
+                if q < bytes.len() {
+                    let name = &template[start..q];
```
evenyag: Is it safe to check the byte `{` to get the string boundary of a UTF-8 string?

killme2008: Yes.  
`String#as_bytes` returns UTF-8 bytes.

src/common/function/src/scalars/string/format.rs
killme2008: Good catch! We should require the name parameter to appear at the end of the parameter list, similar to Python or Rust macros.

Let me fix it.

killme2008: @evenyag I removed the `format` function for now, as it’s not the right time to add it before we consider it carefully. Let’s remove it temporarily.

PTAL.

evenyag (approved):
I didn't check the sqlness tests.

Issue #7102, opened 2025-10-16
Substrait failed to handle alias expr in projection


discord9:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Query Engine

### Minimal reproduce step

add a test in `src/query/src/dist_plan/analyzer/test.rs`
```rs
#[test]
fn test_substrait_decode_projection(){
      let plan = LogicalPlanBuilder::scan_with_filters("t", table_source.clone(), None, vec![])
        .unwrap()
        .project(vec![col("t.ts").alias("ts")])
        .unwrap()
        .build()
        .unwrap();

    try_encode_decode_substrait(&plan, ctx.state());
}
```
and it will fail report found `col("t.ts")` instead of the original `col("t.ts").alias("ts")`

### What did you expect to see?

original `col("t.ts").alias("ts")`

### What did you see instead?

`col("t.ts")`

### What operating system did you use?

all

### What version of GreptimeDB did you use?

0.17.2

### Relevant log output and stack trace

```bash

```

discord9:
fixed by #7109

PR #7109, opened 2025-10-17
fix: prom ql logical plan use column index not name

discord9:

#7102
https://github.com/GreptimeTeam/greptime-proto/pull/285 need merge first

to avoid substrait decode error, use index not name, which will prevent substrait from confuse column names since it will decoded to a slightly different plan(but project node's input schema should still in the same order)

**This pull request contains AI slop**

## PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5328,7 +5328,7 @@ dependencies = [
 [[package]]
 name = "greptime-proto"
 version = "0.1.0"
-source = "git+https://github.com/GreptimeTeam/greptime-proto.git?rev=d9541cd1116a5034526b6d78ba1642f8e643ddcb#d9541cd1116a5034526b6d78ba1642f8e643ddcb"
+source = "git+https://github.com/GreptimeTeam/greptime-proto.git?rev=72a0d22e0f5f716b2ee21bca091f87a88c36e5ca#72a0d22e0f5f716b2ee21bca091f87a88c36e5ca"
 dependencies = [
  "prost 0.13.5",
  "prost-types 0.13.5",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -147,7 +147,7 @@ etcd-client = { git = "https://github.com/GreptimeTeam/etcd-client", rev = "f62d
 fst = "0.4.7"
 futures = "0.3"
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "d9541cd1116a5034526b6d78ba1642f8e643ddcb" }
+greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "72a0d22e0f5f716b2ee21bca091f87a88c36e5ca" }
 hex = "0.4"
 http = "1"
 humantime = "2.1"


killme2008:
It looks like there’s a lot of duplicated code and structs. Can we refactor them?

discord9:
> It looks like there’s a lot of duplicated code and structs. Can we refactor them?

refactor some method, but the "unfix" struct seems necessary, as refactor to either enum or use macro does seems even more messy, this is due to datafusion provided extension deserializer didn't provide input logical plan hence cause this issue

src/promql/src/extension_plan/range_manipulate.rs
```diff
@@ -181,13 +191,25 @@ impl RangeManipulate {
...
+
+        let tag_column_indices = self
+            .field_columns
+            .iter()
+            .map(|name| serialize_column_index(self.input.schema(), name))
+            .collect::<Vec<u64>>();
+        debug!(
+            "RangeManipulate serialize time_index_idx: {}, tag_column_indices: {:?}",
+            time_index_idx, tag_column_indices
+        );
```
evenyag: Do we still need these logs? I found that some plans don't have these logs.

src/promql/src/extension_plan/range_manipulate.rs
```diff
@@ -209,10 +237,11 @@ impl RangeManipulate {
             end: pb_range_manipulate.end,
             interval: pb_range_manipulate.interval,
             range: pb_range_manipulate.range,
-            time_index: pb_range_manipulate.time_index,
-            field_columns: pb_range_manipulate.tag_columns,
             input: placeholder_plan,
             output_schema: empty_schema,
+            unfix: Some(unfix),
+            time_index: String::new(),
+            field_columns: Vec::new(),
```
evenyag: nitpick: We should move time_index and field_column before input.

src/promql/src/extension_plan/scalar_calculate.rs
```diff
@@ -166,17 +186,20 @@ impl ScalarCalculate {
...
+        let unfix = UnfixIndices {
+            time_index_idx: pb_scalar_calculate.time_index_idx,
+            tag_column_indices: pb_scalar_calculate.tag_column_indices.clone(),
+            field_column_idx: pb_scalar_calculate.field_column_idx,
+        };
+
         // TODO(Taylor-lagrange): Supports timestamps of different precisions
         let ts_field = Field::new(
-            &pb_scalar_calculate.time_index,
+            "placeholder_time_index",
```
evenyag, line 198: Why do we change to placeholder_time_index?

tests/cases/standalone/common/promql/encode_substrait.sql
```diff
@@ -0,0 +1,22 @@
...
+    increase(count_total{
+      tag_a="ffa",
+    }[1h])[12h:1h];
+
+tql eval (0, 100, '1s') 
+    increase(count_total{
+      tag_a="ffa",
+    }[1h])[12h:1h];
+
+drop table count_total;
```
evenyag:
```suggestion
drop table count_total;

```

tests/cases/standalone/common/promql/encode_substrait.sql
```diff
@@ -0,0 +1,22 @@
...
+    ts timestamp time index,
+    tag_a string,
+    tag_b string,
+    val double,
+    primary key (tag_a, tag_b),
+);
+
+-- if `RangeManipulate` can be encoded/decoded correctly in substrait, the following queries should pass(for now it simply wouldn't push down)
+-- SQLNESS REPLACE (peers.*) REDACTED
+-- SQLNESS REPLACE (partitioning.*) REDACTED
```
evenyag: Does it mean that we can't push down all promql now?

src/promql/src/extension_plan/scalar_calculate.rs
discord9: > Why do we change to placeholder_time_index?

the old "time_index"(String) is deprcated instead now use column index(`time_index_idx`), so the `pb.time_index` should always be empty while still in `deserialize`(as input plan is not available at current stage)

tests/cases/standalone/common/promql/encode_substrait.sql
discord9: > Does it mean that we can't push down all promql now?

fixed now, comment updated, push down is ok now

evenyag (approved):

killme2008 (approved):
LGTM

Issue #6989, opened 2025-09-18
Enhance copy from csv file command


killme2008:
### What type of enhancement is this?

User experience

### What does the enhancement do?

Enhance the `COPY [table] FROM '/tmp/test.csv' WITH (FORMAT = 'CSV')` command with the following optional parameters:

* `headers= 'true' | 'false'`: Enforce CSV header validation (default: true).
* `skip_bad_records = 'true' | 'false'`: Skip rows that fail parsing (default: false).

Raj-G07:
@killme2008 I believe the enhancement should look like ...
```
 COPY my_table
 FROM '/tmp/test.csv'
 WITH (FORMAT = 'CSV' , headers = 'true' , skip_bad_records = 'false');
```
If yes, then assign me and can you give the exact path of enhancement ?

QuakeWang:
@killme2008 Please assign this task to me, I am glad to do it : )

QuakeWang:
@killme2008 @WenyXu Hi, #8198 has implemented and merged `skip_bad_records`, and #8233 has implemented and merged the `headers` option plus headerless CSV import with positional mapping when `headers=false`.

However, this issue describes `headers=true` as “Enforce CSV header validation”. The current implementation does not perform strict header-name validation: CSV files with headers still use the existing name-based projection behavior, importing the intersection between CSV header names and target table columns. Unknown or missing columns are not rejected at a dedicated header-validation stage; they continue through the existing insert/default/null handling.

Could you confirm whether this current behavior is sufficient to consider #6989 completed and close it? Or should we add a follow-up PR to implement strict header validation for `headers=true`?

fengjiachun:
> [@killme2008](https://github.com/killme2008) [@WenyXu](https://github.com/WenyXu) Hi, [#8198](https://github.com/GreptimeTeam/greptimedb/pull/8198) has implemented and merged `skip_bad_records`, and [#8233](https://github.com/GreptimeTeam/greptimedb/pull/8233) has implemented and merged the `headers` option plus headerless CSV import with positional mapping when `headers=false`.
> 
> However, this issue describes `headers=true` as “Enforce CSV header validation”. The current implementation does not perform strict header-name validation: CSV files with headers still use the existing name-based projection behavior, importing the intersection between CSV header names and target table columns. Unknown or missing columns are not rejected at a dedicated header-validation stage; they continue through the existing insert/default/null handling.
> 
[...]

PR #7115, opened 2025-10-20
feat: add optional parameters for copy table from command

Standing-Man:

- #6989.

- Add optional parameters for copy table from command
  - [x] `header`(default: true): Enforce validation to ensure that the imported CSV file matches the table’s schema, including column names and types.
  - [x] `continue_on_error` (default: false): If a mismatch is detected between the imported csv file and the table schema, and `continue_on_error` is set to true, the file will be skipped and the next CSV file will be processed. Otherwise, the process will stop.

## PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2745,6 +2745,7 @@ dependencies = [
  "common-runtime",
  "common-telemetry",
  "lazy_static",
+ "nix 0.30.1",
  "num_cpus",
  "prometheus 0.14.0",
  "sysinfo",
@@ -8434,6 +8435,18 @@ dependencies = [
  "bitflags 2.9.1",
  "cfg-if",
  "cfg_aliases 0.1.1",
+ "libc",
+]
+
+[[package]]
+name = "nix"
+version = "0.30.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "74523f3a35e05aba87a1d978330aef40f67b0304ac79c1c00b294c9830543db6"
+dependencies = [
+ "bitflags 2.9.1",
+ "cfg-if",
+ "cfg_aliases 0.2.1",
  "libc",
 ]
 
--- a/src/common/stat/Cargo.toml
+++ b/src/common/stat/Cargo.toml
@@ -13,6 +13,8 @@ num_cpus.workspace = true
 prometheus.workspace = true
 sysinfo.workspace = true
 tokio.workspace = true
+[target.'cfg(target_os = "linux")'.dependencies]
+nix.workspace = true
 
 [lints]
 workspace = true


WenyXu:
Hi @Standing-Man, Thank you for your contribution! Could you add some [SQLness tests](https://docs.greptime.com/nightly/contributor-guide/tests/sqlness-test/) or integration tests under the tests-integration directory? You can find an example in `tests-integration/src/test/instance_test.rs`.

Standing-Man:
Hi @WenyXu, I’m a bit confused — should `header` and `has_header` be consistent? Once the header exists, can we directly validate it? Also, how do we determine if a record is corrupted?

https://github.com/GreptimeTeam/greptimedb/blob/main/src/operator/src/statement/copy_table_from.rs#L234

So far, I see how a stream connects to a CSV source, but I’m unsure where the CSV reading and processing logic happens. Could you point me in the right direction?

WenyXu:
> should header and has_header be consistent?

Yes
> how do we determine if a record is corrupted?

It depends on the CSV reader’s capabilities. If users attempt to copy data from multiple files and one of them has a schema that differs from the target table, we should consider skipping data from that file. It would be best to review the CSV reader’s source code.

Standing-Man:
Hi @WenyXu, no matter whether `has_header` is `true` or `false`, the schema compatibility of the target table will be checked. 

https://github.com/GreptimeTeam/greptimedb/blob/d8563ba56d91dde0375457f32b50203357887876/src/operator/src/statement/copy_table_from.rs#L405

Also, the CSV reader reads rows from the source in batches, with default batch size of `DEFAULT_BATCH_SIZE`.

https://github.com/GreptimeTeam/greptimedb/blob/d8563ba56d91dde0375457f32b50203357887876/src/operator/src/statement/copy_table_from.rs#L247

I don’t think it’s necessary to add these two optional parameters to enhance the `COPY TABLE FROM FILE  (FORMAT='CSV')` command.

WenyXu:
> Hi @WenyXu, no matter whether `has_header` is `true` or `false`, the schema compatibility of the target table will be checked.
> 
> https://github.com/GreptimeTeam/greptimedb/blob/d8563ba56d91dde0375457f32b50203357887876/src/operator/src/statement/copy_table_from.rs#L405

Yes, this is the expected behavior. The underlying CSV reader supports reading files without a header line — in such cases, the fields are named field_0, field_1, and so on. Therefore, we assume that the column order in the input CSV matches that of the target table, and align the CSV schema with the table schema before performing the compatibility check.

src/operator/src/statement/copy_table_from.rs
```diff
@@ -390,6 +408,12 @@ impl StatementExecutor {
                 .await?;
 
             let file_schema = file_metadata.schema();
+            if file_type == "CSV"
+                && has_header
+                && !file_schema.equivalent_names_and_types(&table_schema)
+            {
+                continue;
+            }
```
Standing-Man: Hi @WenyXu, files that do not meet the header check are currently skipped. It’s unclear whether the command should terminate and return an error in this case.

WenyXu: Maybe we can allow users to set a continue_on_error option, similar to what COPY DATABASE does.

```rust
        let continue_on_error = req
            .with
            .get(CONTINUE_ON_ERROR_KEY)
            .and_then(|v| bool::from_str(v).ok())
            .unwrap_or(false);
```

Standing-Man: Thank you, I’ve already added the `continue_on_error` option to skip broken CSV files.

src/common/datasource/src/file_format/csv.rs
```diff
@@ -36,6 +36,9 @@ use crate::share_buffer::SharedBuffer;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CsvFormat {
     pub has_header: bool,
+    pub header: bool,
+    pub skip_bad_records: bool,
```
Standing-Man: Hi @killme2008, I’m not sure how rows that can’t be parsed are defined. In the `infer_schema` method, the file’s schema — including column names, types, and metadata — is inferred using `schema_infer_max_record`.

https://github.com/GreptimeTeam/greptimedb/blob/7cc0439cc97523c276f50cfe613abba65ffd63cb/src/common/datasource/src/file_format/csv.rs#L115

Standing-Man:
@killme2008 and @WenyXu PTAL when you have time.

tests/cases/standalone/common/copy/copy_from_fs_csv.sql
WenyXu, line 1: Prefer adding new test cases instead of modifying existing ones.

src/common/datasource/src/file_format/csv.rs
WenyXu, line 1: Can you add unit tests to illustrate the expected behavior of the new feature?

tests/cases/standalone/common/copy/copy_from_fs_csv.sql
Standing-Man: Good point! I’ve added new sqlness tests to validate the optional parameters.

src/common/datasource/src/file_format/csv.rs
Standing-Man: I've added separate tests to validate the optional parameters.the test clearly illustrate the expected behavior of the `header` and `continue_on_error` options. I'm confused about how `skip_bad_records` is intended to be used. 
https://github.com/GreptimeTeam/greptimedb/pull/7115#discussion_r2522957692

src/operator/src/statement/copy_table_from.rs
```diff
@@ -498,6 +515,14 @@ fn can_cast_types_for_greptime(from: &ArrowDataType, to: &ArrowDataType) -> bool
 }
 
 fn ensure_schema_compatible(from: &SchemaRef, to: &SchemaRef) -> Result<()> {
+    if from.fields().len() != to.fields().len() {
+        return error::InvalidHeaderSnafu {
+            table_schema: to.to_string(),
+            file_schema: from.to_string(),
+        }
+        .fail();
+    }
```
WenyXu: We can’t simply determine schema incompatibility based on field length. If the source schema is a subset of the target schema, they should be considered compatible.

tests/cases/standalone/common/copy/copy_from_csv_parameters.result
WenyXu, line 1: I think it would be better to add some test cases in tests-integration to cover the scenarios when `header=false`, e.g.,:
- The order of CSV fields differs from the table’s.
- The number of CSV fields is fewer or greater than the table’s.

BTW, when testing with `header=false`, please make sure to use a CSV file without a header row.

src/operator/src/statement/copy_table_from.rs
Standing-Man: Should we implement a separate header checker method to verify whether the schema lengths of the file and the table match, and whether their names and types are consistent?

tests/cases/standalone/common/copy/copy_from_csv_parameters.result
Standing-Man: Good suggestions. If a CSV file doesn’t contain a header, can I skip checking the header?

WenyXu: > If a CSV file doesn’t contain a header, can I skip checking the header?

I think we should assume that the data fields in the CSV are in the same order as the table columns.

src/operator/src/statement/copy_table_from.rs
WenyXu: Yes, and I suggest adding more tests to cover this feature, as there are many cases we need to handle.

src/operator/src/statement/copy_table_from.rs
```diff
@@ -497,6 +514,18 @@ fn can_cast_types_for_greptime(from: &ArrowDataType, to: &ArrowDataType) -> bool
     can_cast_types(from, to)
 }
 
+fn check_file_header(from: &SchemaRef, to: &SchemaRef) -> Result<()> {
+    if from.fields().len() != to.fields().len() {
```
MichaelScofield, line 519:
```suggestion
    if from.fields().len() > to.fields().len() {
```

The file schema is guaranteed incompatible with table schema if `from.fields().len() > to.fields().len()`, but not if `<`. So `!=` isn't correct here.

src/common/datasource/src/file_format/csv.rs
```diff
@@ -100,6 +120,8 @@ impl Default for CsvFormat {
     fn default() -> Self {
         Self {
             has_header: true,
+            header: false,
+            continue_on_error: true,
```
MichaelScofield: I think the error (either schema incompatibility or data Incompatibilty) should be revealed early and explicitly by default, so:

```suggestion
            header: true,
            continue_on_error: false,
```

killme2008:
@Standing-Man Do you have time to address the comments? I think this PR is nearly ready to merge. Thanks.

@MichaelScofield @WenyXu Let's move forward.

src/operator/src/statement/copy_table_from.rs
Standing-Man: It primarily checks that the file and table have the same number of fields; type compatibility is verified only when the counts match.

If the file has five fields but the table only has four, return false. If the table and file have the same number of fields, proceed to check type compatibility for all fields.

src/common/datasource/src/file_format/csv.rs
Standing-Man: Should the default behavior be to validate headers and not skip invalid files?

tests/cases/standalone/common/copy/copy_from_csv_parameters.result
Standing-Man: In the Greptime documentation, I found that when importing a CSV file into the database, the CSV file must include a header for the import to succeed.

src/common/datasource/src/file_format/csv.rs
MichaelScofield: I believe by default the header should be checked, and the errors are not to be skipped either.

tests/cases/standalone/common/copy/copy_from_csv_parameters.result
```diff
@@ -0,0 +1,241 @@
...
+Copy enable_long_fields FROM '${SQLNESS_HOME}/demo/export/csv_header/demo_1.csv' with (format='csv', header='true');
+
+Affected Rows: 0
+
+SELECT count(*) FROM enable_long_fields;
+
++----------+
+| count(*) |
++----------+
+| 0        |
```
MichaelScofield, line 51: Intuitively, the data should be imported to the table successfully, so the `count(*)` should be 3. Is it because the `continue_on_error` is default to `true`?

MichaelScofield:
@Standing-Man @WenyXu It seems to me the current discussion is a little blurry, let's clarify it a bit:

1. Should the header of a csv file must be present for it to be imported?
  - If yes, then the option `header` might be useless, we have to check the header either way;
  - If no, then there are situations combined from 2 cases: the csv has header or not, and the `header` option is `true` or `false`. We should deal with them separately.
2. The `continue_on_error` can be kept, to determine whether the header validation failure ends in an error.
3. The `skip_bad_records` is needed, too. To not confuse with `continue_on_error` which is about header validation, this option is about csv **data** validation.

WDYT?

Standing-Man:
@MichaelScofield  @WenyXu 
> Should the header of a csv file must be present for it to be imported?

Currently, CSV files must include a header to be imported successfully. We use the `header` parameter to filter and import only CSV files whose schemas are consistent.

> The `skip_bad_records` is needed, too. To not confuse with `continue_on_error` which is about header validation, this option is about csv data validation.

For the `skip_bad_records` parameter, the reader does not process data row by row but reads multiple rows in batches. As a result, this parameter does not have much practical effect.

Standing-Man:
Does the SQLness framework have an expected-error mechanism, similar to SLT’s statement error?
```slt
statement error
SELECT ....
```

MichaelScofield:
> @MichaelScofield @WenyXu
> 
> > Should the header of a csv file must be present for it to be imported?
> 
> Currently, CSV files must include a header to be imported successfully. We use the `header` parameter to filter and import only CSV files whose schemas are consistent.

There might be a misunderstanding for `header` parameter here. I think it actually means "whether to skip the csv header checking or not" in the original issue. @killme2008 can you confirm?

> 
> > The `skip_bad_records` is needed, too. To not confuse with `continue_on_error` which is about header validation, this option is about csv data validation.
> 
> For the `skip_bad_records` parameter, the reader does not process data row by row but reads multiple rows in batches. As a result, this parameter does not have much practical effect.

I think the `skip_bad_records` is valuable, the problem is how to implement it, that might need to modify the upstream crate (namely, DataFusion or Arrow). We can do it in another PR, let's focus on csv header header.

MichaelScofield:
> Does the SQLness framework have an expected-error mechanism, similar to SLT’s statement error?
> 
> ```
> statement error
> SELECT ....
> ```

It has, just simply put the error in the result file.

MichaelScofield:
@Standing-Man To push this PR forward, I suggest:

1. Add only `header` and `continue_on_error`, put aside `skip_bad_records` for now.
2. Upon checking header, if `header = true`, go on checking. Otherwise lessen the requirement that csv file must have a header, deduce the file schema from its data, then continue to check.
3. If `continue_on_error = true`, ignore the checking error. Otherwise return the error to user.
4. `header` is defaulted to `true` and `continue_on_error` is defaulted to `false` to align with current behavior.

WDYT?

Standing-Man:
> 1. Add only `header` and `continue_on_error`, put aside `skip_bad_records` for now.
> 2. Upon checking header, if `header = true`, go on checking. Otherwise lessen the requirement that csv file must have a header, deduce the file schema from its data, then continue to check.
> 3. If `continue_on_error = true`, ignore the checking error. Otherwise return the error to user.
> 4. `header` is defaulted to `true` and `continue_on_error` is defaulted to `false` to align with current behavior.

@MichaelScofield The PR follows these rules now.

tests/cases/standalone/common/copy/copy_from_fs_csv.sql
```diff
@@ -15,7 +15,7 @@ Copy demo TO '${SQLNESS_HOME}/demo/export/csv/demo.csv' with (format='csv');
 
 CREATE TABLE with_filename(host string, cpu double, memory double, ts timestamp time index);
 
-Copy with_filename FROM '${SQLNESS_HOME}/demo/export/csv/demo.csv' with (format='csv', start_time='2022-06-15 07:02:37', end_time='2022-06-15 07:02:39');
+Copy with_filename FROM '${SQLNESS_HOME}/demo/export/csv/demo.csv' with (format='csv', header='false', start_time='2022-06-15 07:02:37', end_time='2022-06-15 07:02:39');
```
WenyXu, line 18: Better to keep it unchanged. We need tests to cover the cases where `header=true` or when it’s using the default value.

src/operator/src/statement/copy_table_from.rs
```diff
@@ -366,6 +371,18 @@ impl StatementExecutor {
...
+            if let Some(csv_format) = csv_format_opt
+                && csv_format.header
+                && let Err(e) = check_file_header(file_schema, &table_schema)
+            {
+                if csv_format.continue_on_error {
+                    continue;
+                } else {
+                    return Err(e);
+                }
+            };
```
WenyXu, line 385: I’m curious about how a CSV without a header is associated with the table’s fields — in other words, how each column in the CSV maps to a specific column in the table. This is also why I’d prefer adding unit tests to cover this scenario instead of a higher-level sqlness test. As I recall, if we let the underlying reader parse a CSV file without a header, it will label the columns as `field_0`, `field_1`, and so on.

tests/cases/standalone/common/copy/copy_from_fs_csv.sql
Standing-Man: @WenyXu, I think `copy_from_fs_csv.sql` should test the functionality of importing data from CSV files, and should not include tests for the header parameter. Keeping the SQL focused on data import is best.

src/operator/src/statement/copy_table_from.rs
Standing-Man: I’ve added a test case to cover copying from a CSV file without a header, but I noticed in [the GreptimeDB documentation](https://docs.greptime.com/reference/sql/copy/#copy-from) that imported CSV files are required to have a header row.

tests-integration/src/tests/instance_test.rs
```diff
@@ -1872,6 +1872,42 @@ async fn test_cast_type_issue_1594(instance: Arc<dyn MockInstance>) {
...
+        .data,
+        OutputData::AffectedRows(0)
+    ));
+
+    let filepath = find_testing_resource("/src/common/datasource/tests/csv/no_heaer.csv");
+
+    let output = try_execute_sql(
+        &instance,
+        &format!(
+            "copy no_header from '{}' WITH(FORMAT='csv', header='false');",
```
MichaelScofield: From what I see, that csv file should be imported successfully if `header = false`. What changed?

WenyXu: +1

tests-integration/src/tests/instance_test.rs
```diff
@@ -1872,6 +1872,42 @@ async fn test_cast_type_issue_1594(instance: Arc<dyn MockInstance>) {
...
+    .await;
+
+    assert!(output.is_err());
+    assert!(matches!(
+        output.unwrap_err(),
+        Error::TableOperation {
+            source: _,
+            location: _
+        }
+    ));
```
WenyXu: Question: Why does this case fail?

Standing-Man: The current implementation still matches by column names even when `has_header=false`. 
As a result, a CSV without a header is inferred as `column_1/2/3`, which leads to an empty projection and triggers the Arrow error you observed. 

I have already confirmed this in [another PR](https://github.com/GreptimeTeam/greptimedb/pull/7767).
cc @MichaelScofield

Standing-Man: I think this is a separate feature and could be addressed in [another PR](https://github.com/GreptimeTeam/greptimedb/pull/7767).

tests-integration/src/tests/instance_test.rs
Standing-Man: In [the GreptimeDB documentation](https://docs.greptime.com/reference/sql/copy/#copy-from), the imported CSV files are required to have a header row. 

Additionally, the `header` parameter is used to validate the CSV header against the table schema, whereas `has_header` indicates whether the first row of the file is a header.

> Note:
>    The CSV file must have a header row to be imported correctly. The header row should contain the column 
>     names of the table.

Therefore, `has_header` must be set to true for CSV files.

src/operator/src/statement/copy_table_from.rs
WenyXu: Yes, that's exactly why you're creating a pull request to implement this feature.

tests-integration/src/tests/instance_test.rs
WenyXu: If you're planning to separate this feature into another PR, please ensure that this PR does not include any changes related to that feature. In other words, keep this PR focused on its key features and tests.

Standing-Man: Got it. I have removed the related test validations and fixed the CI errors.

Standing-Man:
Hi @WenyXu and @MichaelScofield, The CI errors have been fixed. Please let me know if anything else is needed. 😄

tests/cases/standalone/common/copy/copy_from_csv_parameters.result
```diff
@@ -0,0 +1,279 @@
...
+insert into
+    demo_2(host, jsons, ts)
+values
+    ('host4', '{"host4":"4"}', 1755276557000),
+    ('host5', '{"host5":"5"}', 1755276558000),
+    ('host6', '{"host6":"6"}', 1822077263000);
+
+Affected Rows: 3
+
+Copy demo_1 TO '${SQLNESS_HOME}/demo/export/csv_header/demo_1.csv' with (format='csv');
```
MichaelScofield, line 29: The generated csv file has header. Does this affect the effectivity of the `header = false` test?

Standing-Man:
> So it breaks down to the following requirements: 
> 1. if the csv file has header line, import it as usual; 
> 2. if not, and if the `header = true` (the default), return an error; 
> 3. otherwise deduce the file schema and match it with table schema, then import. 

Not exactly. The current PR assumes that the CSV file **must have a header row**.
As for the `continue_on_error` parameter: if any errors occur during the import process, they will be ignored when `continue_on_error = true`; otherwise, the errors will be returned.

I will support importing data from CSV files without header rows in [another PR](https://github.com/GreptimeTeam/greptimedb/pull/7767).

tests/cases/standalone/common/copy/copy_from_csv_parameters.result
Standing-Man: The `header` parameter indicates whether the schema in the CSV file should be validated against the schema of the target table. If `header = false`, it means the schema of the CSV file will not be checked against the schema of the table.

WenyXu:
> Got it. I have removed the related test validations and fixed the CI errors.

Please take another look at this change. There are still some unrelated changes included.

tests/cases/standalone/common/copy/copy_from_csv_parameters.result
Standing-Man: For the case where `header=false`, I have already added [tests](https://github.com/Standing-Man/greptimedb/blob/cc02184afcb0b3d41b1b873c14cc22a0818feaa8/tests/cases/standalone/common/copy/copy_from_csv_parameters.result#L76-L132) covering schema mismatches between the table and the CSV file, including differences in column count and column order. 
More specifically, [no schema validation](https://github.com/Standing-Man/greptimedb/blob/cc02184afcb0b3d41b1b873c14cc22a0818feaa8/src/operator/src/statement/copy_table_from.rs#L376-L385) is performed between the table and the CSV file.

Standing-Man:
> Please take another look at this change. There are still some unrelated changes included.

Thanks for the review! I didn’t find any obvious unrelated changes on my side. Could you point out the specific parts so I can take another look?

src/operator/src/statement/copy_table_from.rs
```diff
@@ -498,6 +515,18 @@ fn can_cast_types_for_greptime(from: &ArrowDataType, to: &ArrowDataType) -> bool
     can_cast_types(from, to)
 }
 
+fn check_file_header(from: &SchemaRef, to: &SchemaRef) -> Result<()> {
```
WenyXu, line 518: I think there may be a misunderstanding about `generated_schema_projection_and_compatible_file_schema`. We actually allow copying a subset of columns from external files, and any missing columns can be populated with default values or set to null.

src/common/datasource/src/file_format/csv.rs
```diff
@@ -84,6 +86,24 @@ impl TryFrom<&HashMap<String, String>> for CsvFormat {
                 .build()
             })?;
         };
+        if let Some(header) = value.get(file_format::FORMAT_HEADER) {
```
WenyXu, line 89: We already have `FORMAT_HAS_HEADER`

tests/cases/standalone/common/copy/copy_from_fs_csv.sql
```diff
@@ -33,13 +33,13 @@ select host, cpu, memory, jsons, ts from demo where host != 'host3';
...
 
-Copy with_path FROM '${SQLNESS_HOME}/demo/export/csv/' with (format='csv', start_time='2023-06-15 07:02:37');
+Copy with_path FROM '${SQLNESS_HOME}/demo/export/csv/' with (format='csv', header='false', start_time='2023-06-15 07:02:37');
 
 select * from with_path order by ts;
 
 CREATE TABLE with_pattern(host string, cpu double, memory double, ts timestamp time index);
 
-Copy with_pattern FROM '${SQLNESS_HOME}/demo/export/csv/' WITH (pattern = 'demo.*', format='csv', end_time='2025-06-15 07:02:39');
+Copy with_pattern FROM '${SQLNESS_HOME}/demo/export/csv/' WITH (pattern = 'demo.*', format='csv', header='false', end_time='2025-06-15 07:02:39');
```
WenyXu, line 42: I don't want to change the tests for this flag. We also don't support copying data from CSV files without headers, so `header=false` is confusing.

src/common/datasource/src/file_format/csv.rs
Standing-Man: Got it, but I'm still a bit unclear about what exactly should be checked for the `has_header` flag. It might also be due to the issue description not being very clear.

Standing-Man:
After thinking more about this, I’m still unclear about the intended behavior of this flag. I’ll close this PR for now and continue to follow up on the issue. Thanks everyone for the time spent reviewing! ❤️

PR #7121, opened 2025-10-20
fix: fix index and tag filtering for flat format

evenyag:

When using sparse encoding under flat format, tag columns are not decoded during flush and compaction. So we have to decode them in the index creator manually.

The tag columns need exact filtering. However, when the tag column is not in the projection but present in the filter, the flat format can't get them from the converted batch. This PR tries to decode the tag from the primary key so we can filter the batch by tags.

This PR address these issues by:
1. decoding tag columns in the index creator
2. filtering tag columns explicitly

This PR also fixes another issue that the FlatCompatBatch didn't handle non-string tags correctly.

Limitation:
- If the filter has a field column that is not in the projection, we can't filter it here.

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/index/bloom_filter/creator.rs
+++ b/src/mito2/src/sst/index/bloom_filter/creator.rs
@@ -295,79 +295,78 @@ impl BloomFilterIndexer {
         let mut decoded_pks: Option<Vec<(CompositeValues, usize)>> = None;
 
         for (col_id, creator) in &mut self.creators {
-            if let Some(column_meta) = self.metadata.column_by_id(*col_id) {
-                let column_name = &column_meta.column_schema.name;
-                if let Some(column_array) = batch.column_by_name(column_name) {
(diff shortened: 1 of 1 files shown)


src/mito2/src/sst/index/bloom_filter/creator.rs
```diff
@@ -317,20 +319,55 @@ impl BloomFilterIndexer {
...
+                                .context(EncodeSnafu)?;
+                                Ok(buf)
+                            })
+                            .transpose()?;
+
+                        creator
+                            .push_n_row_elems(*count, elems)
+                            .await
+                            .context(PushBloomFilterValueSnafu)?;
+                    }
```
fengjiachun: Nit: Nested Hell, it would be better if the nesting were reduced.

evenyag: I reduced the block to get the column meta.

PR #7125, opened 2025-10-21
feat: prefix option for timestamp index and value column

shuiyisong:

Add an option for setting the default prefix for the timestamp column name and value column name on Frontend, Datanode and Standalone. If not set, it will use the default `greptime_timestamp` and `greptime_value`.
Note, this option will only affect the auto-creating-table protocols like Prometheus remote write.

## PR Checklist

### DIFF
--- a/config/config.md
+++ b/config/config.md
@@ -13,7 +13,7 @@
 | Key | Type | Default | Descriptions |
 | --- | -----| ------- | ----------- |
 | `default_timezone` | String | Unset | The default timezone of the server. |
-| `default_column_prefix` | String | Unset | The default column prefix for auto-created timestamp and value columns. |
+| `default_column_prefix` | String | Unset | The default column prefix for auto-created time index and value columns. |
 | `init_regions_in_background` | Bool | `false` | Initialize all regions in the background during the startup.<br/>By default, it provides services after all regions have been initialized. |
 | `init_regions_parallelism` | Integer | `16` | Parallelism of initializing regions. |
 | `max_concurrent_queries` | Integer | `0` | The maximum current queries allowed to be executed. Zero means unlimited. |
@@ -227,7 +227,7 @@
 | Key | Type | Default | Descriptions |
 | --- | -----| ------- | ----------- |
 | `default_timezone` | String | Unset | The default timezone of the server. |
-| `default_column_prefix` | String | Unset | The default column prefix for auto-created timestamp and value columns. |
+| `default_column_prefix` | String | Unset | The default column prefix for auto-created time index and value columns. |
 | `max_in_flight_write_bytes` | String | Unset | The maximum in-flight write bytes. |
 | `runtime` | -- | -- | The runtime options. |
 | `runtime.global_rt_size` | Integer | `8` | The number of threads to execute the runtime for global read operations. |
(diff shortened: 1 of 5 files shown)


config/config.md
```diff
@@ -13,6 +13,7 @@
 | Key | Type | Default | Descriptions |
 | --- | -----| ------- | ----------- |
 | `default_timezone` | String | Unset | The default timezone of the server. |
+| `default_timestamp_column_name` | String | Unset | The default timestamp column name. |
```
killme2008: Prefer `default_time_index_column_name` and clarify the description: the name of the time index column used when automatically creating tables, such as with Prometheus remote write or InfluxDB line protocol.

src/cmd/src/frontend.rs
```diff
@@ -332,6 +334,9 @@ impl StartCommand {
             .context(error::StartFrontendSnafu)?;
 
         set_default_timezone(opts.default_timezone.as_deref()).context(error::InitTimezoneSnafu)?;
+        set_greptime_timestamp(opts.default_timestamp_column_name.as_deref())
```
killme2008: The name `set_greptime_timestamp` isn’t clear; let’s rename it to `set_time_index_column_name`.

src/common/query/src/prelude.rs
```diff
@@ -12,12 +12,46 @@
...
+    };
+
+    ensure!(
+        NAME_PATTERN_REG.is_match(ts),
+        DefaultTSColNameSnafu {
+            name: ts,
+            message: format!("Invalid character in timestamp column name: {}", ts)
+        }
+    );
+
```
killme2008: Add an info log.

src/common/query/src/prelude.rs
```diff
@@ -12,12 +12,46 @@
...
+    GREPTIME_TIMESTAMP_CELL.get_or_init(|| ts.to_string());
+    Ok(())
+}
+
+/// Default timestamp column name constant for backward compatibility.
+const GREPTIME_TIMESTAMP: &str = "greptime_timestamp";
 /// Default value column name for Prometheus metrics.
 pub const GREPTIME_VALUE: &str = "greptime_value";
```
killme2008: Should `greptime_value` also be configurable?

shuiyisong: In that case, how about change the config option to something like `default_prefix`, so that it applies to time index column name and value column name. For example, set the option to `new_db`, and the column names would be `new_db_timestamp` and `new_db_value`

killme2008: I think so. It can also be empty.  

What do you think, @sunng87?

MichaelScofield:
Why not change the default to a lesser-likely-conflict-name like `__greptime_timestamp__`?

src/common/query/src/prelude.rs
sunng87: Sounds good

killme2008:
> Why not change the default to a lesser-likely-conflict-name like `__greptime_timestamp__`?

It’s not about conflict, it’s about branding.

src/common/query/src/prelude.rs
shuiyisong: @killme2008 PTAL

src/common/query/src/prelude.rs
```diff
@@ -12,15 +12,61 @@
...
+static GREPTIME_VALUE_CELL: OnceCell<String> = OnceCell::new();
+
+pub fn set_default_prefix(prefix: Option<&str>) -> Result<()> {
+    match prefix {
+        None => {
+            // use default greptime prefix
+            GREPTIME_TIMESTAMP_CELL.get_or_init(|| GREPTIME_TIMESTAMP.to_string());
+            GREPTIME_VALUE_CELL.get_or_init(|| GREPTIME_VALUE.to_string());
+        }
+        Some("") => {
```
killme2008: Looks more safe:
```suggestion
        Some(s) if s.trim().is_empty() => {
```

config/config.md
```diff
@@ -13,6 +13,7 @@
 | Key | Type | Default | Descriptions |
 | --- | -----| ------- | ----------- |
 | `default_timezone` | String | Unset | The default timezone of the server. |
+| `default_column_prefix` | String | Unset | The default column prefix for auto-created timestamp and value columns. |
```
killme2008:
```suggestion
| `default_column_prefix` | String | Unset | The default column prefix for auto-created time index and value columns. |
```

The same suggestions as below.

Issue #7021, opened 2025-09-25
enhance Region Migration Procedure to support multiple regions


WenyXu:
### What type of enhancement is this?

Performance

### What does the enhancement do?

At present, each region migration is handled by a separate procedure. It may become inefficient in large cases such as failover or rebalance, where a datanode may need to migrate many regions. Since creating and running a procedure is expensive, running one for each region causes high overhead. We propose to improve the migration mechanism so that a single procedure can handle multiple region migrations.

fengjiachun:
It looks like all the work is done. Can we close this issue? @WenyXu

PR #7130, opened 2025-10-22
refactor: refactor instruction handler and adds support for batch region downgrade operations

WenyXu:

#7021

This PR refactors the instruction handler of datanode and adds support for batch region downgrade operations, as preparatory work for batch region migration.

### Key Changes
- Renamed `Instruction::DowngradeRegion` to `Instruction::DowngradeRegions` with `Vec<DowngradeRegion>` support
- Renamed `InstructionReply::DowngradeRegion` to `InstructionReply::DowngradeRegions` with `Vec<DowngradeRegionReply>` return type
- Added `region_id` field to `DowngradeRegionReply` for identifying individual responses in batch operations
- Implemented parallel processing of multiple region downgrades using `join_all`
- Updated all related tests and mock utilities to work with the new batch format

## PR Checklist

### DIFF
--- a/src/common/meta/src/instruction.rs
+++ b/src/common/meta/src/instruction.rs
@@ -55,6 +55,8 @@ impl Display for RegionIdent {
 /// The result of downgrade leader region.
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
 pub struct DowngradeRegionReply {
+    /// The [RegionId].
+    /// For compatibility, it is defaulted to [RegionId::new(0, 0)].
     #[serde(default)]
     pub region_id: RegionId,
     /// Returns the `last_entry_id` if available.


src/common/meta/src/instruction.rs
```diff
@@ -460,7 +541,11 @@ pub enum InstructionReply {
     #[serde(alias = "close_region")]
     CloseRegions(SimpleReply),
     UpgradeRegion(UpgradeRegionReply),
-    DowngradeRegion(DowngradeRegionReply),
+    #[serde(
+        alias = "downgrade_region",
+        deserialize_with = "downgrade_regions_compat_from"
+    )]
+    DowngradeRegions(DowngradeRegionsReply),
```
fengjiachun, line 550: Why not use `Vec<DowngradeRegionReply>` directly?

src/common/meta/src/instruction.rs
```diff
@@ -452,6 +500,39 @@ impl Display for UpgradeRegionReply {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
+pub struct DowngradeRegionsReply {
+    pub replies: Vec<DowngradeRegionReply>,
+}
```
fengjiachun, line 508: It seems unnecessary to wrap this (DowngradeRegionsReply); can we use `Vec<DowngradeRegionReply>` directly?

src/common/meta/src/instruction.rs
```diff
@@ -55,6 +55,8 @@ impl Display for RegionIdent {
 /// The result of downgrade leader region.
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
 pub struct DowngradeRegionReply {
+    #[serde(default)]
```
fengjiachun, line 60: Why need `default`? Comments should be added to explain this point.

src/common/meta/src/instruction.rs
```diff
@@ -452,6 +500,39 @@ impl Display for UpgradeRegionReply {
...
+fn downgrade_regions_compat_from<'de, D>(deserializer: D) -> Result<DowngradeRegionsReply, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let helper = DowngradeRegionsCompat::deserialize(deserializer)?;
+    Ok(match helper {
+        DowngradeRegionsCompat::Single(x) => DowngradeRegionsReply::new(vec![x]),
+        DowngradeRegionsCompat::Multiple(reply) => reply,
+    })
+}
```
fengjiachun, line 536: Why is this needed? Can't `single_or_multiple_from` take care of it?

src/datanode/src/heartbeat/handler/downgrade_region.rs
```diff
@@ -12,209 +12,242 @@
...
+            warn!(
+                "Region: {region_id} is not writable, flush_timeout: {:?}",
+                flush_timeout
+            );
+            return ctx.downgrade_to_follower_gracefully(region_id).await;
+        }
+
+        // If flush_timeout is not set, directly convert region to follower.
+        let Some(flush_timeout) = flush_timeout else {
+            return ctx.downgrade_to_follower_gracefully(region_id).await;
```
MichaelScofield, line 63: Why does writable region can be directly downgraded to follower here without setting its role to `DowngradingLeader` first as below?

src/common/meta/src/instruction.rs
WenyXu: The `InstructionReply` is serialized with a tag field, so it cannot be serialized as `{ tag: '', [{...}] }`. We need to wrap the vector in a root field.

src/datanode/src/heartbeat/handler/downgrade_region.rs
WenyXu: The `None` stands for don't flush region before downgrading it.

src/common/meta/src/instruction.rs
WenyXu: Addressed in [42068d4](https://github.com/GreptimeTeam/greptimedb/pull/7130/commits/42068d48380cfe50017ff037c83a213cc9fbff1e).
For compatibility, it is defaulted to `[RegionId::new(0, 0)]`.

src/datanode/src/heartbeat/handler/downgrade_region.rs
MichaelScofield: It's not related to `None`, it's about maybe the role is not properly set.

WenyXu: The `DowngradingLeader` is designed to reject write operations while still allowing flushes. Therefore, it's ok to change the role from leader to follower directly.

fengjiachun (approved):
Rest LGTM

MichaelScofield (approved):

PR #7136, opened 2025-10-23
fix: memtable value push result was ignored

MichaelScofield:

as title

## PR Checklist

### DIFF
--- a/src/mito2/src/memtable/time_series.rs
+++ b/src/mito2/src/memtable/time_series.rs
@@ -924,7 +924,7 @@ impl ValueBuilder {
                     mutable_vector.push_nulls(num_rows - 1);
                     mutable_vector
                         .push(field_value)
-                        .unwrap_or_else(|e| panic!("unexpected: {e}"));
+                        .unwrap_or_else(|e| panic!("unexpected field value: {e:?}"));
                     self.fields[idx] = Some(mutable_vector);
                     MEMTABLE_ACTIVE_FIELD_BUILDER_COUNT.inc();
                 }


src/mito2/src/memtable/time_series.rs
```diff
@@ -922,14 +923,16 @@ impl ValueBuilder {
                             )
                         };
                     mutable_vector.push_nulls(num_rows - 1);
-                    let _ = mutable_vector.push(field_value);
+                    mutable_vector
+                        .push(field_value)
+                        .context(DataTypeMismatchSnafu)?;
```
MichaelScofield: key change

MichaelScofield:
> LGTM
> 
> But can we add a test that return an error in such a case?

Actually it's an illegal state, like `unreachable!`. So it's not required a test. I made it returned an error just out of personal preference.

killme2008:
> > LGTM
> > But can we add a test that return an error in such a case?
> 
> Actually it's an illegal state, like `unreachable!`. So it's not required a test. I made it returned an error just out of personal preference.

If so, panicking directly with `expect` is preferable, as the result chain impacts performance a little bit.

src/mito2/src/memtable/time_series.rs
```diff
@@ -892,7 +893,7 @@ impl ValueBuilder {
         sequence: u64,
         op_type: u8,
         fields: impl Iterator<Item = ValueRef<'a>>,
-    ) -> usize {
```
evenyag, line 895: https://github.com/GreptimeTeam/greptimedb/blob/b53a0b86fb8b2f3cede177025c8a9ffb4bb40e3a/src/mito2/src/memtable/time_series.rs#L887-L889

It was by design to not returning an error here. @v0y4g3r  WDYT?
I'm fine to change this to return a result, but you should also update the comment.

Also we can't panic here otherwise we can't recover the write worker task.

v0y4g3r: > https://github.com/GreptimeTeam/greptimedb/blob/b53a0b86fb8b2f3cede177025c8a9ffb4bb40e3a/src/mito2/src/memtable/time_series.rs#L887-L889
> 
> It was by design to not returning an error here. @v0y4g3r WDYT? I'm fine to change this to return a result, but you should also update the comment.

Yes, by design it's not expected to encounter an error in `Series::push`.

MichaelScofield: changed to simply panic, ptal @killme2008 @evenyag @v0y4g3r

src/mito2/src/memtable/time_series.rs
```diff
@@ -922,7 +922,9 @@ impl ValueBuilder {
                             )
                         };
                     mutable_vector.push_nulls(num_rows - 1);
-                    let _ = mutable_vector.push(field_value);
+                    mutable_vector
+                        .push(field_value)
+                        .unwrap_or_else(|e| panic!("unexpected: {e}"));
```
evenyag: We may print it in debug format or just use `unwrap()`.
```suggestion
                        .unwrap_or_else(|e| panic!("unexpected field value: {e:?}"));
```

PR #7151, opened 2025-10-28
feat: json vector builder

MichaelScofield:

vector builder for json type values

please see the tests in the new "builder.rs" file first

## PR Checklist

### DIFF
--- a/src/datatypes/src/types/json_type.rs
+++ b/src/datatypes/src/types/json_type.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use std::str::FromStr;
 use std::sync::Arc;
 
@@ -36,6 +36,7 @@ use crate::vectors::{BinaryVectorBuilder, MutableVector};
 
 pub const JSON_TYPE_NAME: &str = "Json";
 const JSON_PLAIN_FIELD_NAME: &str = "__plain__";
+const JSON_PLAIN_FIELD_METADATA_KEY: &str = "is_plain_json";
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize, Default)]
 pub enum JsonFormat {
@@ -65,17 +66,19 @@ impl JsonType {
     /// Make json type a struct type, by:
     /// - if the json is an object, its entries are mapped to struct fields, obviously;
     /// - if not, the json is one of bool, number, string or array, make it a special field called
-    ///   [JSON_PLAIN_FIELD_NAME] in a struct with only that field.
+    ///   [JSON_PLAIN_FIELD_NAME] with metadata [JSON_PLAIN_FIELD_METADATA_KEY] = `"true"` in a
+    ///   struct with only that field.
     pub(crate) fn as_struct_type(&self) -> StructType {
         match &self.format {
             JsonFormat::Jsonb => StructType::default(),
             JsonFormat::Native(inner) => match inner.as_ref() {
                 ConcreteDataType::Struct(t) => t.clone(),
-                x => StructType::new(Arc::new(vec![StructField::new(
-                    JSON_PLAIN_FIELD_NAME.to_string(),
-                    x.clone(),
-                    true,
-                )])),
+                x => {
+                    let mut field =
+                        StructField::new(JSON_PLAIN_FIELD_NAME.to_string(), x.clone(), true);
+                    field.insert_metadata(JSON_PLAIN_FIELD_METADATA_KEY, true);
+                    StructType::new(Arc::new(vec![field]))
+                }
             },
         }
     }
@@ -87,7 +90,11 @@ impl JsonType {
             return true;
         };
         let fields = t.fields();
-        fields.len() == 1 && fields[0].name() == JSON_PLAIN_FIELD_NAME
+        let Some((single, [])) = fields.split_first() else {
+            return false;
+        };
+        single.name() == JSON_PLAIN_FIELD_NAME
+            && single.metadata(JSON_PLAIN_FIELD_METADATA_KEY) == Some("true")
     }
 
     /// Try to merge this json type with others, error on datatype conflict.
@@ -104,6 +111,47 @@ impl JsonType {
             }
             .fail(),
         }
+    }
+
+    pub(crate) fn is_mergeable(&self, other: &JsonType) -> bool {
+        match (&self.format, &other.format) {
+            (JsonFormat::Jsonb, JsonFormat::Jsonb) => true,
+            (JsonFormat::Native(this), JsonFormat::Native(that)) => {
+                is_mergeable(this.as_ref(), that.as_ref())
+            }
+            _ => false,
+        }
+    }
+}
+
+fn is_mergeable(this: &ConcreteDataType, that: &ConcreteDataType) -> bool {
+    fn is_mergeable_struct(this: &StructType, that: &StructType) -> bool {
+        let this_fields = this.fields();
+        let this_fields = this_fields
+            .iter()
+            .map(|x| (x.name(), x))
+            .collect::<HashMap<_, _>>();
+
+        for that_field in that.fields().iter() {
+            if let Some(this_field) = this_fields.get(that_field.name())
+                && !is_mergeable(this_field.data_type(), that_field.data_type())
+            {
+                return false;
+            }
+        }
+        true
+    }
+
+    match (this, that) {
+        (this, that) if this == that => true,
+        (ConcreteDataType::List(this), ConcreteDataType::List(that)) => {
+            is_mergeable(this.item_type(), that.item_type())
+        }
+        (ConcreteDataType::Struct(this), ConcreteDataType::Struct(that)) => {
+            is_mergeable_struct(this, that)
+        }
+        (ConcreteDataType::Null(_), _) | (_, ConcreteDataType::Null(_)) => true,
+        _ => false,
     }
 }
 
@@ -243,10 +291,12 @@ mod tests {
             let result = json_type.merge(other);
             match (result, expected) {
                 (Ok(()), Ok(expected)) => {
-                    assert_eq!(json_type.name(), expected)
+                    assert_eq!(json_type.name(), expected);
+                    assert!(json_type.is_mergeable(other));
                 }
                 (Err(err), Err(expected)) => {
-                    assert_eq!(err.to_string(), expected)
+                    assert_eq!(err.to_string(), expected);
+                    assert!(!json_type.is_mergeable(other));
                 }
                 _ => unreachable!(),
             }
--- a/src/datatypes/src/types/struct_type.rs
+++ b/src/datatypes/src/types/struct_type.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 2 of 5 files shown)


src/datatypes/src/types/struct_type.rs
```diff
@@ -46,6 +46,15 @@ impl TryFrom<&Fields> for StructType {
     }
 }
 
+impl<const N: usize> From<[StructField; N]> for StructType {
+    fn from(value: [StructField; N]) -> Self {
+        let value: Box<[StructField]> = Box::new(value);
```
sunng87, line 52: is box required here?

src/datatypes/src/types/json_type.rs
```diff
@@ -54,28 +56,40 @@ impl JsonType {
...
-    // TODO(LFC): remove "allow unused"
-    #[allow(unused)]
+    /// Check if this json type is the special "plain" one.
+    /// See [JsonType::as_struct_type].
+    pub(crate) fn is_plain_json(&self) -> bool {
+        let JsonFormat::Native(box ConcreteDataType::Struct(t)) = &self.format else {
+            return true;
+        };
+        let fields = t.fields();
+        fields.len() == 1 && fields[0].name() == JSON_PLAIN_FIELD_NAME
```
sunng87: Do we have additional metadata for this plain type? There is a minor case that the field name can be same with user provided json.

src/datatypes/src/vectors/json/builder.rs
```diff
@@ -0,0 +1,379 @@
...
+
+        let ValueRef::Json(value) = value else {
+            // Safety: json datatype value must be the value of json.
+            unreachable!()
+        };
+        builder.push(value)
+    }
+
+    fn push_null(&mut self) {
+        let null_json_value = ValueRef::Json(Box::new(ValueRef::Null));
```
sunng87, line 279: can we use structvector's push null, it should use arrow's null semantics?

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -323,6 +323,96 @@ impl StructVectorBuilder {
...
+                .iter()
+                .position(|other| other.name() == field.name())
+            {
+                let other = &mut other.value_builders[j];
+                if field.data_type().is_struct() {
+                    let Some(builder) = builder.as_mut_any().downcast_mut::<StructVectorBuilder>()
+                    else {
+                        // Safety: a struct datatype field must be corresponding to a struct vector builder.
+                        unreachable!()
+                    };
```
sunng87: just use unwrap

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -323,6 +323,96 @@ impl StructVectorBuilder {
...
+        }
+
+        for (i, field) in self.fields.fields().iter().enumerate() {
+            let builder = &mut self.value_builders[i];
+
+            if let Some(j) = other
+                .fields
+                .fields()
+                .iter()
+                .position(|other| other.name() == field.name())
```
sunng87: is this nested for loop in a hot path? I wonder if we need to optimize this, for example, keep fields sorted by default

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -323,6 +323,96 @@ impl StructVectorBuilder {
...
+                }
+            } else {
+                builder.push_nulls(other.len());
+            }
+        }
+        Ok(())
+    }
+
+    /// Same as [StructVectorBuilder::try_merge], but not consume the other builder's data.
+    pub(crate) fn try_merge_cloned(&mut self, other: &StructVectorBuilder) -> Result<()> {
```
sunng87: any chance to avoid duplicated code here?

src/datatypes/src/types/struct_type.rs
MichaelScofield: Yes, the receiver of method `into_vec` in line 53 is `Box[T]`: https://doc.rust-lang.org/stable/std/primitive.slice.html#method.into_vec

Otherwise we have to use `to_vec` which requires memory copy, not good since we own the input value here.

src/datatypes/src/vectors/json/builder.rs
MichaelScofield: It **is** using struct vector's "push null". The inner `StructVectorBuilder`s of this `JsonVectorBuilder` are created lazily in `try_push_value_ref`, so we have to go through that method here.

src/datatypes/src/vectors/struct_vector.rs
MichaelScofield: Thought about that, but I'd like to keep it like this. For two reasons: 1. fields in struct type don't require sorted, by definition; 2. to be merged `StructVectorBuilder` generally does not have many fields, simply iterating the fields might achieve a balanced time and space costs, comparing to sort the fields first or put the fields into a hashmap.

src/datatypes/src/vectors/struct_vector.rs
MichaelScofield: Unfortunately can't. Tried but ended in complex macros, hurting codes readability very much. The difficulties hide in conditionally 1. unify `&mut` and `&`; 2. calling `downcast_mut` vs `downcast_ref` and `to_vector` vs `to_vector_cloned`. I'm all ears if you can guide me other ways.

src/datatypes/src/types/json_type.rs
MichaelScofield: Are you suggesting putting some metadata in the arrow field? Though I think the name collision (to `__plain__`) is minimum, I can do that.

MichaelScofield:
@sunng87 PTAL

src/datatypes/src/types/json_type.rs
sunng87: Yes, we can address this in next patch. Althrough it's super minor but once it hits the data is corrupted.

src/datatypes/src/vectors/struct_vector.rs
sunng87: OK. Sounds good. Never use macro for this.

sunng87 (approved):

sunng87:
@evenyag @waynexia You may want to check this too

src/datatypes/src/vectors/json/builder.rs
```diff
@@ -0,0 +1,379 @@
...
+                .unwrap_or_else(|e| panic!("failed to merge json builders, error: {e}"));
+        }
+        unified_jsons.to_vector()
+    }
+
+    fn try_push_value_ref(&mut self, value: &ValueRef) -> Result<()> {
+        let data_type = value.data_type();
+        let json_type = data_type.as_json().with_context(|| TryFromValueSnafu {
+            reason: format!("expected json value, got {value:?}"),
+        })?;
```
evenyag, line 258: Can we merge types successfully if the value is null?

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -323,6 +323,97 @@ impl StructVectorBuilder {
         }
         self.null_buffer.append_null();
     }
+
+    /// Try to merge (and consume the data of) other struct vector builder into this one.
+    pub(crate) fn try_merge(&mut self, other: &mut StructVectorBuilder) -> Result<()> {
+        if let Some(other) = other.null_buffer.finish_cloned() {
```
evenyag: Should we use `finish()`? The comment says it consume the builder.

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -323,6 +323,97 @@ impl StructVectorBuilder {
...
+            self.null_buffer.append_buffer(&other)
+        } else {
+            self.null_buffer.append_n_non_nulls(other.len());
+        }
+
+        for (i, field) in self.fields.fields().iter().enumerate() {
```
evenyag: Here we only check fields in self, what about fields in `other` but not in self?

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -323,6 +323,97 @@ impl StructVectorBuilder {
...
+                            reason: format!(
+                                "expecting struct vector builder for field '{}' in other",
+                                field.name()
+                            ),
+                        })?;
+                    builder.try_merge(other)?;
+                } else {
+                    let vector = other.to_vector();
+                    builder.extend_slice_of(vector.as_ref(), 0, vector.len())?;
+                }
```
evenyag: Is it possible that other and builder has different data type?

src/datatypes/src/vectors/struct_vector.rs
```diff
@@ -323,6 +323,97 @@ impl StructVectorBuilder {
...
+        for (i, field) in self.fields.fields().iter().enumerate() {
+            let builder = &mut self.value_builders[i];
+
+            if let Some(j) = other
+                .fields
+                .fields()
+                .iter()
+                .position(|other| other.name() == field.name())
+            {
+                let other = &mut other.value_builders[j];
```
evenyag: non-blocking: There are so many `other` here. Sometime it is hard to remember what `other` is.

src/datatypes/src/vectors/json/builder.rs
MichaelScofield: To push nulls into this builder, the function `push_null` is used. There the nulls are "wrapped" as null json values, have null json types(`Json<Struct<"__plain__": Null>>`), can be successfully merged.

src/datatypes/src/vectors/struct_vector.rs
MichaelScofield: This "merge" is for json type only, so there can't be. But since this function is defined in struct vector builder, I will complete it. Same for the following comment.

evenyag: We can document this and add a debug assertion. I don't think we have to implment all cases in this method. We just need to remind callers that the builder should has the unified type.

MichaelScofield: On second thought, I decide to move the "merge" functions to json vector builder. Struct vector builder stays "static" for struct values, by nature. Merging can be only happened for json values, then the restrictions can be ensured there together.

Standing-Man:
Hi @MichaelScofield, is this PR about the vector functions `JSON_TO_VEC` and `VEC_TO_JSON`?

MichaelScofield:
@Standing-Man No, it's about the new implementation of json type.

src/datatypes/src/types/json_type.rs
```diff
@@ -31,9 +31,12 @@ use crate::scalars::ScalarVectorBuilder;
 use crate::type_id::LogicalTypeId;
 use crate::types::{ListType, StructField, StructType};
 use crate::value::Value;
+use crate::vectors::json::builder::JsonVectorBuilder;
 use crate::vectors::{BinaryVectorBuilder, MutableVector};
 
 pub const JSON_TYPE_NAME: &str = "Json";
+const JSON_PLAIN_FIELD_NAME: &str = "__plain__";
```
evenyag, line 38: Do we need to rename it to a name that has lower conflict possibility? e.g. `__json_plain__`
```suggestion
const JSON_PLAIN_FIELD_NAME: &str = "__json_plain__";
```

evenyag (approved):

src/datatypes/src/types/json_type.rs
MichaelScofield: That's good naming. I'll do it in the next PR.

PR #7131, opened 2025-10-22
refactor: convert to postgres values directly from arrow

MichaelScofield:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

following https://github.com/GreptimeTeam/greptimedb/pull/7096
for postgres

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #7153, opened 2025-10-29
refactor: convert to prometheus values directly from arrow

MichaelScofield:

following https://github.com/GreptimeTeam/greptimedb/pull/7131
for prometheus values (http output)

also reduced some memory for storing intermediate prometheus series data

## PR Checklist

### DIFF
--- a/src/servers/src/http/prometheus.rs
+++ b/src/servers/src/http/prometheus.rs
@@ -762,8 +762,17 @@ fn record_batches_to_series(
     Ok(())
 }
 
+/// Writer from a row in the record batch to a Prometheus time series:
+///
+/// `{__name__="<metric name>", <label name>="<label value>", ...}`
+///
+/// The metrics name is the table name; label names are the column names and
+/// the label values are the corresponding row values (all are converted to strings).
 struct RowWriter {
+    /// The template that is to produce a Prometheus time series. It is pre-filled with metrics name
+    /// and label names, waiting to be filled by row values afterward.
     template: HashMap<Column, Option<String>>,
+    /// The current filling row.
     current: Option<HashMap<Column, Option<String>>>,
 }
 


src/servers/src/http/prometheus.rs
```diff
@@ -723,20 +756,344 @@ fn record_batches_to_series(
...
-        for row in batch.rows() {
-            let mut element: HashMap<String, String> = row
+        let mut writer = RowWriter::new(&batch.schema, table_name);
+        writer.write(batch, series)?;
+    }
+    Ok(())
+}
+
+struct RowWriter {
+    template: HashMap<Column, Option<String>>,
```
sunng87, line 774: provide some comments for the meaning of key and value of template

src/servers/src/http/prometheus.rs
```diff
@@ -723,20 +756,344 @@ fn record_batches_to_series(
...
+    }
+
+    fn insert_bytes(&mut self, column_schema: &ColumnSchema, bytes: &[u8]) -> Result<()> {
+        let column_name = column_schema.name.as_str().into();
+
+        if column_schema.data_type.is_json() {
+            let s = jsonb_to_string(bytes).context(ConvertScalarValueSnafu)?;
+            self.insert(column_name, s);
+        } else {
+            let hex = bytes
```
evenyag, line 812: The output labels in promql can only be string type. I think we can simplify the writer. We can return an error if the output label type isn't string.

src/servers/src/http/prometheus.rs
```diff
@@ -723,20 +756,344 @@ fn record_batches_to_series(
...
+    fn write(
+        &mut self,
+        record_batch: RecordBatch,
+        series: &mut Vec<HashMap<Column, String>>,
+    ) -> Result<()> {
+        let schema = record_batch.schema.clone();
+        let record_batch = record_batch.into_df_record_batch();
+        for i in 0..record_batch.num_rows() {
+            for (j, array) in record_batch.columns().iter().enumerate() {
+                let column = schema.column_name_by_index(j).into();
```
evenyag, line 841: It's unexpected to have other data types. We can only process strings.

PR #7154, opened 2025-10-29
fix: avoid filtering rows with delete op by fields under merge mode

evenyag:

https://github.com/GreptimeTeam/greptimedb/issues/7161

When scanning data with predicates on field columns, the current implementation always applies except last-non-null mode. However, when delete operations are present, filtering on field columns during the pruning phase can lead to incorrect result since deleted rows will be filtered out. Thus the engine may return rows already been deleted.

See the test in the `prune_test.rs` for more details.

This PR introduces a `PreFilterMode` to control the field filtering strategy to avoid delete markers being filtered.
- `All`: Apply filters on all columns (append mode)
- `SkipFields`: Always skip filtering on field columns during pruning (last non null mode)
- `SkipFieldsOnDelete`: Conditionally skip field filtering only when delete operations are detected in row groups (last row mode)

Filter Categorization and Split
- **Separated field filters from non-field filters**: Introduced `partition_by_field_filters()` method to categorize filters based on whether they reference field columns
- **Dual index appliers**: Created separate index appliers (inverted, bloom, fulltext) for field and non-field filters, allowing independent control
- **Removed aggressive filter removal**: Replaced the old `maybe_remove_field_filters()` that unconditionally removed all field filters in LastNonNull mode

It introduced `RangesOptions` struct to replace the simple `for_flush: bool` parameter in `Memtable::ranges()` method.

Note that if we use merge mode but sets fields as partition expressions, we may got unexpected result because rows with the same key may belongs to different regions. It is out of the scope of this PR.

## PR Checklist

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -117,7 +117,7 @@ jobs:
       - name: Build greptime binaries
         shell: bash
         # `cargo gc` will invoke `cargo build` with specified args
-        run: cargo build --bin greptime --bin sqlness-runner --features "pg_kvbackend,mysql_kvbackend"
+        run: cargo gc -- --bin greptime --bin sqlness-runner --features "pg_kvbackend,mysql_kvbackend"
       - name: Pack greptime binaries
         shell: bash
         run: |
@@ -808,7 +808,7 @@ jobs:
       - name: Setup external services
         working-directory: tests-integration/fixtures
         run: ../../.github/scripts/pull-test-deps-images.sh && docker compose up -d --wait
-
+        
       - name: Run nextest cases
         run: cargo llvm-cov nextest --workspace --lcov --output-path lcov.info -F dashboard -F pg_kvbackend -F mysql_kvbackend
         env:


src/mito2/src/read/scan_region.rs
```diff
@@ -466,12 +471,22 @@ impl ScanRegion {
...
         );
 
-        // Remove field filters for LastNonNull mode after logging the request.
-        self.maybe_remove_field_filters();
-
-        let inverted_index_applier = self.build_invereted_index_applier();
-        let bloom_filter_applier = self.build_bloom_filter_applier();
-        let fulltext_index_applier = self.build_fulltext_index_applier();
+        // // Remove field filters for LastNonNull mode after logging the request.
+        // self.maybe_remove_field_filters();
```
v0y4g3r:
```suggestion
```

src/mito2/src/memtable.rs
```diff
@@ -191,14 +227,13 @@ pub trait Memtable: Send + Sync + fmt::Debug {
...
     ///
-    /// The `for_flush` flag is true if the flush job calls this method for flush.
     /// The returned map contains the range id and the range after applying the predicate.
     fn ranges(
         &self,
         projection: Option<&[ColumnId]>,
         predicate: PredicateGroup,
         sequence: Option<SequenceRange>,
-        for_flush: bool,
+        options: RangesOptions,
```
v0y4g3r, line 256: Can we merge `projection`/`predicate`/`sequence` into `RangeOptions`?

src/mito2/src/sst/parquet/file_range.rs
```diff
@@ -222,40 +262,50 @@ impl FileRangeContext {
...
-            .context(DecodeStatsSnafu {
-                file_path: self.reader_builder.file_path(),
-            })
+        row_group_contains_delete(metadata, row_group_index, self.reader_builder.file_path())
     }
 }
 
+/// Mode to pre-filter columns in a range.
+#[derive(Debug, Clone, Copy)]
+pub enum PreFilterMode {
```
v0y4g3r, line 299: Looks likje `FilterApplyMode` would be more understandable?

v0y4g3r (approved):
LGTM

src/mito2/src/sst/parquet/file_range.rs
evenyag: We may do a post filter after the merge reader later so I named it as PreFilterMode.

evenyag:
I created an issue https://github.com/GreptimeTeam/greptimedb/issues/7161 for this case. I'll update the sqlness test in another PR.

src/mito2/src/memtable.rs
evenyag: Adding the projection will introduce a lifetime parameter to the options which is I'd like to avoid. I'm considering moving predicate and sequence to the options.

src/mito2/src/memtable/bulk/part_reader.rs
```diff
@@ -152,8 +176,14 @@ impl BulkPartRecordBatchIter {
         // Apply projection first.
         let projected_batch = self.apply_projection(record_batch)?;
         // Apply combined filtering (both predicate and sequence filters)
+        // For BulkPartRecordBatchIter, we don't have row group information.
+        let skip_fields = match self.context.pre_filter_mode() {
+            PreFilterMode::All => false,
+            PreFilterMode::SkipFields => true,
+            PreFilterMode::SkipFieldsOnDelete => true,
```
fengjiachun, line 183: Unable to check if the batch includes any delete operations?

evenyag: It's possible but requires iterating the op types. So I prefer to keep things simple here. Also, this PR is already large enough.

fengjiachun: Got it

PR #7164, opened 2025-10-31
feat: BulkMemtable stores small fragments in another buffer

evenyag:

This PR adds an `UnorderedPart` to the `BulkMemtable` to reduce the compaction overhead when the requests are too small. Once the part has enough rows, it compacts itself into a BulkPart in place.

During each query, the UnorderedPart sorts the parts in it and returns the ordered parts.

Ingest latency comparison under batch size 10:
image

This PR also updates the memtable scan metrics in the BulkMemtable reader.

## PR Checklist

### DIFF
--- a/src/mito2/src/memtable/bulk/part_reader.rs
+++ b/src/mito2/src/memtable/bulk/part_reader.rs
@@ -283,11 +283,15 @@ impl Iterator for BulkPartRecordBatchIter {
     type Item = error::Result<RecordBatch>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let record_batch = self.record_batch.take()?;
+        let Some(record_batch) = self.record_batch.take() else {
+            // `take()` should be cheap, we report the metrics directly.
+            self.report_mem_scan_metrics();
+            return None;
+        };
 
         let result = self.process_batch(record_batch).transpose();
 
-        // Report metrics when iteration is complete
+        // Reports metrics when iteration is complete
         if result.is_none() {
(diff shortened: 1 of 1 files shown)


src/mito2/src/memtable/bulk/part_reader.rs
```diff
@@ -198,7 +285,37 @@ impl Iterator for BulkPartRecordBatchIter {
     fn next(&mut self) -> Option<Self::Item> {
         let record_batch = self.record_batch.take()?;
 
-        self.process_batch(record_batch).transpose()
+        let result = self.process_batch(record_batch).transpose();
+
+        // Report metrics when iteration is complete
+        if result.is_none() {
+            self.report_mem_scan_metrics();
```
fengjiachun, line 296: It never reaches here because it will return at line 286.

src/mito2/src/memtable/bulk/part_reader.rs
```diff
@@ -198,7 +285,37 @@ impl Iterator for BulkPartRecordBatchIter {
     fn next(&mut self) -> Option<Self::Item> {
         let record_batch = self.record_batch.take()?;
```
fengjiachun: If `record_batch` is None, should we call `report_mem_scan_metrics`?

src/mito2/src/memtable/bulk/part_reader.rs
evenyag: This is possible if we have filtered all rows in the batch.

PR #7168, opened 2025-10-31
feat: add greptime's arrow json extension type

sunng87:

This patch adds greptime's arrow extension type `greptime.json`. This information can be stored as parquet metadata to indicate that a particular field is logical json.

I also migrated our json structure settings in standard arrow extension way. Adding arrow extension type data to our column definition.

## PR Checklist

### DIFF
--- a/src/datatypes/src/extension/json.rs
+++ b/src/datatypes/src/extension/json.rs
@@ -87,7 +87,9 @@ impl ExtensionType for JsonExtensionType {
             // boolean
             | DataType::Boolean
             // null
-            | DataType::Null=> Ok(()),
+            | DataType::Null
+            // legacy json type
+            | DataType::Binary => Ok(()),
             dt => Err(ArrowError::SchemaError(format!(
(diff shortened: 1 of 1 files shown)


sunng87:
They failing case is because the json type is still using `Binary` for underlying data type. I suggest to use `json2` for this new structure @MichaelScofield

MichaelScofield:
@sunng87 Seems making the new json type a new datatype `json2`, has to do much more things than making it compatible with the old? Because the datatype system is very deeply rooted in the codes, a lot of places need to adapt to a new datatype.

sunng87:
@MichaelScofield ok. I can make Binary a supported datatype of this extension. But we need to be careful with the state of legacy json in upcoming 0.18 and make sure it's at least not broken.

sunng87:
This extension information will be stored in `Field` of arrow, and should be accessible from `RecordBatch`.

waynexia (approved):

PR #7174, opened 2025-11-04
fix(mito): avoid shortcut in picking multi window files

v0y4g3r:

- Fix the issue where windows without files to compact would discontinue the whole compaciton process
- Limit compaction tasks to `max_background_compactions`
- Delete expired files before compaction.

## PR Checklist

### DIFF
--- a/src/mito2/src/compaction/task.rs
+++ b/src/mito2/src/compaction/task.rs
@@ -28,7 +28,7 @@ use crate::manifest::action::{RegionEdit, RegionMetaAction, RegionMetaActionList
 use crate::metrics::{COMPACTION_FAILURE_COUNT, COMPACTION_STAGE_ELAPSED};
 use crate::region::RegionLeaderState;
 use crate::request::{
-    BackgroundNotify, CompactionFailed, CompactionFinished, OutputTx, RegionEditRequest,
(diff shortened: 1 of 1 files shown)


evenyag (approved):

src/mito2/src/compaction/twcs.rs
```diff
@@ -88,7 +88,7 @@ impl TwcsPicker {
             // because after compaction there will be no overlapping files.
             let filter_deleted = !files.overlapping && found_runs <= 2 && !self.append_mode;
             if found_runs == 0 {
-                return output;
+                continue;
```
evenyag, line 93: We may need to limit the number of output. Otherwise, the region may have a large compaction task after fixing this.

waynexia (approved):

src/mito2/src/compaction/task.rs
```diff
@@ -78,9 +81,97 @@ impl CompactionTaskImpl {
...
+            return;
+        }
+
+        // 2. Notify region worker loop to remove expired files from region version.
+        self.send_to_worker(WorkerRequest::EditRegion(RegionEditRequest {
+            region_id: compaction_region.region_id,
+            edit,
+            tx: expire_delete_sender,
+        }))
+        .await;
```
evenyag, line 131: We should send RegionEditResult instead. The worker will spawn a task to update the manifest it handles the request.

PR #7175, opened 2025-11-04
feat: simple read write new json type values

MichaelScofield:

please see the sqlness test `json-structured.result` first

for compatibility consideration, the table created with datatype `JSON` is treated as the old json type

## PR Checklist


src/api/src/helper.rs
```diff
@@ -789,7 +776,7 @@ pub fn pb_value_to_value_ref<'a>(
             let items = struct_value
                 .items
                 .iter()
-                .zip(struct_datatype_ext.fields.iter())
+                .zip(struct_value.fields.iter())
```
sunng87: include field information in each struct value can be verbose, is it possible to reuse it from datatype ?

tests/cases/standalone/common/types/json/json-structured.result
```diff
@@ -0,0 +1,79 @@
...
+
+Affected Rows: 1
+
+INSERT INTO t VALUES (1762128010000, '{"int": -5}');
+
+Affected Rows: 1
+
+INSERT INTO t VALUES (1762128011000, '{}');
+
+Error: 1004(InvalidArguments), Invalid InsertRequest, reason: empty structs are not supported, consider adding a dummy field
```
sunng87: Can we use assume it's all null for known fields?

src/common/recordbatch/src/recordbatch.rs
```diff
@@ -249,6 +253,47 @@ impl RecordBatch {
...
+            if f.extension_type_name() == Some(JsonExtensionType::NAME) {
+                let field_type = f.data_type();
+                let array_type = a.data_type();
+                match (field_type, array_type) {
+                    (ArrowDataType::Struct(x), ArrowDataType::Struct(y)) if x != y => {
+                        common_telemetry::debug!(
+                            "Align field {} datatype: {field_type} => {array_type}",
+                            f.name(),
+                        );
+                        let field = Field::new_struct(f.name(), y.clone(), f.is_nullable())
```
sunng87: do we need to merge x and y here?

MichaelScofield:
successor of https://github.com/GreptimeTeam/greptimedb/pull/7248

MichaelScofield:
@sunng87 @fengys1996 PTAL

src/api/src/helper.rs
```diff
@@ -127,6 +125,7 @@ impl From<ColumnDataTypeWrapper> for ConcreteDataType {
                         };
                         ConcreteDataType::json_native_datatype(inner_type.into())
                     }
+                    None => ConcreteDataType::Json(JsonType::empty()),
```
fengys1996: nit: better name 

```suggestion
                    None => ConcreteDataType::Json(JsonType::null()),
```

src/api/src/helper.rs
```diff
@@ -441,19 +440,22 @@ impl TryFrom<ConcreteDataType> for ColumnDataTypeWrapper {
                         JsonFormat::Jsonb => Some(ColumnDataTypeExtension {
                             type_ext: Some(TypeExt::JsonType(JsonTypeExtension::JsonBinary.into())),
                         }),
-                        JsonFormat::Native(inner) => {
+                        JsonFormat::Native(inner) => (!ConcreteDataType::from(inner.as_ref())
+                            .is_null())
+                        .then(|| {
                             let inner_type = ColumnDataTypeWrapper::try_from(
                                 ConcreteDataType::from(inner.as_ref()),
                             )?;
```
fengys1996: Avoid duplicate calls `ConcreteDataType::from(inner.as_ref())`

src/api/src/helper.rs
```diff
@@ -882,116 +884,11 @@ pub fn is_column_type_value_eq(
     ColumnDataTypeWrapper::try_new(type_value, type_extension)
         .map(|wrapper| {
             let datatype = ConcreteDataType::from(wrapper);
-            expect_type == &datatype
+            expect_type.is_json() && datatype.is_json() || &datatype == expect_type
```
fengys1996: nit: add comment to function signature to explain json related.

src/api/src/helper.rs
fengys1996, line 883: Incorrect document

src/api/src/helper.rs
```diff
@@ -1872,7 +1754,7 @@ mod tests {
             Arc::new(ConcreteDataType::boolean_datatype()),
         ));
 
-        let pb_value = to_proto_value(value);
+        let pb_value = value_to_grpc_value(value);
```
fengys1996: nit: better function name
```suggestion
        let pb_value = to_grpc_value(value);
```

src/common/sql/src/convert.rs
```diff
@@ -231,13 +231,15 @@ pub fn sql_value_to_value(
         }
     }
 
-    if value.data_type() != *data_type {
+    let value_datatype = value.data_type();
+    // The datatype of json value is determined by its actual data, so we can't simply "cast" it here.
+    if value_datatype.is_json() || value_datatype == *data_type {
```
fengys1996, line 236: This logic might be problematic.

For example, when evaluating `parse_column_default_constraint`, the incoming value might be a number, while the column’s declared type is JSON. In this case, should we return an error?

MichaelScofield: Here we are just to avoid unnecessary cast for json values. The actual validation of input json value against column json type is later during the insertion.

As for default value, I choose to disallow it for json column, because it's meaningless. Note that disallow default value doesn't mean disallow null value. Users can just put "null"s for json column.

MichaelScofield:
@fengys1996 PTAL

src/common/recordbatch/src/recordbatch.rs
```diff
@@ -327,6 +330,94 @@ pub fn merge_record_batches(schema: SchemaRef, batches: &[RecordBatch]) -> Resul
...
+                json_array, schema_fields
+            )
+        }
+    );
+
+    let json_array =
+        StructArray::try_new(schema_fields.clone(), aligned, json_array.nulls().cloned())
+            .context(NewDfRecordBatchSnafu)?;
+    Ok(Arc::new(json_array))
+}
```
sunng87, line 405: Do we have unit tests to cover this function, especially for some edge cases.

src/datatypes/src/types/json_type.rs
```diff
@@ -232,6 +236,43 @@ impl JsonType {
...
+        (this, that) if this == that => true,
+        (JsonNativeType::Array(this), JsonNativeType::Array(that)) => {
+            is_include(this.as_ref(), that.as_ref())
+        }
+        (JsonNativeType::Object(this), JsonNativeType::Object(that)) => {
+            is_include_object(this, that)
+        }
+        (_, JsonNativeType::Null) => true,
+        _ => false,
+    }
```
sunng87, line 275: Consider covering this with unit tests

src/operator/src/req_convert/insert/stmt_to_region.rs
```diff
@@ -94,25 +106,64 @@ impl<'a> StatementToRegion<'a> {
...
+            insert_columns = find_insert_columns(&table, &column_names)?;
+        }
+
+        // Finally convert GreptimeDB values to GRPC values, ready to do insertion on Datanode.
+        for (i, row) in value_rows.into_iter().enumerate() {
+            for value in row {
+                let grpc_value = to_grpc_value(value);
+                rows[i].values.push(grpc_value);
+            }
+        }
```
sunng87, line 156: Possible to extract this logic into a new function for readability?
Also we may not need to create `JsonColummnTypeUpdater` for all cases?

src/operator/src/req_convert/insert/stmt_to_region.rs
```diff
@@ -194,6 +235,147 @@ impl<'a> StatementToRegion<'a> {
...
+
+                updater.merge_types(insert_column, &value)?;
+
+                value_row.push(value);
+            }
+            value_rows.push(value_row);
+        }
+        Ok(value_rows)
+    }
+}
```
sunng87, line 275: Would be helpful to include unit tests to demo and ensure usage.

src/operator/src/req_convert/insert/stmt_to_region.rs
```diff
@@ -237,9 +419,25 @@ fn sql_value_to_grpc_value(
...
+            ensure!(
+                !value.is_empty_object(),
+                InvalidInsertRequestSnafu {
+                    reason: "empty json object is not supported, consider adding a dummy field"
+                }
+            );
+            Ok(())
+        }
+        _ => Ok(()),
+    }
```
sunng87, line 440: can we assume empty object is null?

src/sql/src/statements.rs
```diff
@@ -269,7 +273,14 @@ pub fn sql_data_type_to_concrete_data_type(data_type: &SqlDataType) -> Result<Co
                 Ok(ConcreteDataType::decimal128_datatype(*p as u8, *s as i8))
             }
         },
-        SqlDataType::JSON => Ok(ConcreteDataType::json_datatype()),
+        SqlDataType::JSON => {
+            let format = if column_extensions.json_datatype_options.is_some() {
```
sunng87, line 277: in next patch, we'd better to provide an explicit toggle for new json type.

src/common/sql/src/convert.rs
fengys1996: Get

src/operator/src/req_convert/insert/stmt_to_region.rs
MichaelScofield: Tried further refactor it but ended with messier codes.

`JsonColummnTypeUpdater` is rather light-weight, no clones involved, so I think it's fine here.

src/operator/src/req_convert/insert/stmt_to_region.rs
MichaelScofield: I've specifically crafted the sqlness test cases to cover the codes here, so I think it's fine.

src/operator/src/req_convert/insert/stmt_to_region.rs
MichaelScofield: It can be treated as a null value. But currently the null value insertion in json column has some problems, I'm going to fix them as well as this one here in the following PRs.

src/sql/src/statements.rs
MichaelScofield: Now the column created with `JSON()` is using the new json type. It's kind of like a "toggle" already.

sunng87 (approved):

sunng87:
@fengys1996 PTAL

src/operator/src/lib.rs
```diff
@@ -14,6 +14,7 @@
 
 #![feature(assert_matches)]
 #![feature(if_let_guard)]
+#![feature(box_patterns)]
```
fengys1996: It seems this feature isn't being used.

PR #7177, opened 2025-11-04
feat: allow creating logical tabel with same partition rule with physical table's

waynexia:

Logical table with the same partition rule is also allowed (equal to no op)

## PR Checklist


tests/cases/standalone/common/create/metric_engine_partition.result
```diff
@@ -429,5 +458,5 @@ Affected Rows: 0
 
 drop table metric_engine_partition;
 
-Affected Rows: 0
+Error: 1004(InvalidArguments), Physical region 4398046511104(1024, 0) is busy, there are still some logical regions using it
```
WenyXu: We should drop the `logical_table_1` before dropping the physical table

WenyXu (approved):
Rest LGTM

PR #7180, opened 2025-11-05
feat: import backup data from local files

waynexia:

Hard code all connection info in the `COPY DATABASE` clause is inconvenient when back up from a moved files.

## PR Checklist

### DIFF
--- a/src/cli/src/data/import.rs
+++ b/src/cli/src/data/import.rs
@@ -282,6 +282,20 @@ impl Import {
     }
 }
 
+#[async_trait]
+impl Tool for Import {
+    async fn do_work(&self) -> std::result::Result<(), BoxedError> {
+        match self.target {
+            ImportTarget::Schema => self.import_create_table().await.map_err(BoxedError::new),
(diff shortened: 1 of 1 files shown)


discord9 (approved):
LGTM, could use some sort of integration test later(maybe in separate pr)

src/cli/src/data/export.rs
```diff
@@ -668,10 +668,26 @@ impl Export {
...
+                            r#"COPY DATABASE "{}"."{}" FROM '{}' WITH ({}){};"#,
+                            export_self.catalog, schema, path, with_options_clone, connection_part
+                        );
+
+                        format!(
+                            "-- {}\n{}",
+                            command_with_connection, command_without_connection
+                        )
+                    }
+                };
```
evenyag, line 690: Can we extract the replace logic into a function and test it? Same as the replace logic in import.

waynexia: [aa9ce7d](https://github.com/GreptimeTeam/greptimedb/pull/7180/commits/aa9ce7db6a23d37f1de1cade5eabefcb06e76998)

PR #7181, opened 2025-11-05
feat: tracks index files in another cache and preloads them

evenyag:

This PR implements an dedicated file cache for puffin files. It adds the following two options:
- `preload_index_cache`: Preload index (puffin) files into cache on region open. This is enabled by default.
- `index_cache_percent`: Percentage of write cache capacity allocated for index (puffin) files.

The `FileCache` adds another index to track puffin files under the write cache. When the engine opens a region, it creates a `RegionLoadCacheTask` to load the puffin files. The `WriteCache` runs the `RegionLoadCacheTask` after it finish recovery.

## PR Checklist

### DIFF
--- a/src/mito2/src/region/opener.rs
+++ b/src/mito2/src/region/opener.rs
@@ -961,10 +961,11 @@ fn can_load_cache(state: RegionRoleState) -> bool {
         RegionRoleState::Leader(RegionLeaderState::Writable)
         | RegionRoleState::Leader(RegionLeaderState::Staging)
         | RegionRoleState::Leader(RegionLeaderState::Altering)
-        | RegionRoleState::Leader(RegionLeaderState::Downgrading)
-        | RegionRoleState::Leader(RegionLeaderState::Editing) => true,
-        RegionRoleState::Leader(RegionLeaderState::Dropping)
-        | RegionRoleState::Leader(RegionLeaderState::Truncating) => true,
-        RegionRoleState::Follower => true,
+        | RegionRoleState::Leader(RegionLeaderState::Editing)
+        | RegionRoleState::Follower => true,
+        // The region will be closed soon if it is downgrading.
+        RegionRoleState::Leader(RegionLeaderState::Downgrading)
+        | RegionRoleState::Leader(RegionLeaderState::Dropping)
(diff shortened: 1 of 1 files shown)


src/mito2/src/region/opener.rs
```diff
@@ -809,3 +817,151 @@ where
...
+        let mut files_already_cached = 0;
+
+        {
+            let version = version_control.current().version;
+            for level in version.ssts.levels() {
+                for file_handle in level.files.values() {
+                    let file_meta = file_handle.meta_ref();
+                    if file_meta.exists_index() {
+                        let puffin_key =
+                            IndexKey::new(file_meta.region_id, file_meta.file_id, FileType::Puffin);
```
fengjiachun: Should we call `file_meta.index_file_id()` instead of `file_meta.file_id`?

fengjiachun: The part that downloads Puffin in `src/mito2/src/worker/handle_manifest.rs` is similar.

src/mito2/src/region/opener.rs
```diff
@@ -809,3 +817,151 @@ where
...
+            }
+
+            // Checks if adding this file would exceed capacity
+            if current_size + file_size > capacity {
+                info!(
+                    "Stopping index cache preload due to capacity limit, region: {}, file_id: {}, current_size: {}, file_size: {}, capacity: {}",
+                    region_id, puffin_key.file_id, current_size, file_size, capacity
+                );
+                files_skipped = (total_files - files_downloaded) as usize;
+                CACHE_FILL_PENDING_FILES.sub(total_files - files_downloaded);
```
fengjiachun, line 897: Every time the download, whether successful or failed, has already been decremented; there may have been repeated operations here.

evenyag: We break here so we decrease the remaining number of files.

src/mito2/src/region/opener.rs
evenyag: https://github.com/GreptimeTeam/greptimedb/pull/7181/commits/4066115cba35570288ecb651bbf93d588fa831ea

PR #7183, opened 2025-11-05
chore: print root cause in opendal logging interceptor

evenyag:

Get the root cause in the LoggingInterceptor in case of the error implementation doesn't display the source

## PR Checklist

### DIFF
--- a/src/object-store/src/util.rs
+++ b/src/object-store/src/util.rs
@@ -12,11 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::error::Error as StdError;
 use std::fmt::Display;
 use std::path;
 use std::time::Duration;
 
+use common_error::root_source;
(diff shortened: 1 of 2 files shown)


src/object-store/src/util.rs
```diff
@@ -174,19 +175,28 @@ impl LoggingInterceptor for DefaultLoggingInterceptor {
...
     ) {
         if let Some(err) = err {
+            let mut root = err.source();
+            while let Some(r) = root {
+                if let Some(s) = r.source() {
+                    root = Some(s);
+                } else {
+                    break;
+                }
+            }
```
sunng87: Would be helpful to make this an utility function in our commons

evenyag: Well, we can move it to common-error. Since common-error has already used the error_iter feature, maybe I can simplify the implementation here.

PR #7186, opened 2025-11-05
fix(mito): append mode in flat format not working

### LINKED ISSUE #7185: Append mode sqlness tests fails
### What type of bug is this?

Incorrect result

Other

### Minimal reproduce step

Enable `default_experimental_flat_format`, and run the test case in https://github.com/GreptimeTeam/greptimedb/blob/main/tests/cases/standalone/common/select/flush_append_only.sql

Duplicated rows are kept.

### What did you see instead?

Duplicated rows are removed.

NA

Latest

https://github.com/GreptimeTeam/greptimedb/actions/runs/19112438018/job/54613495314

v0y4g3r:

- fixes #7185
- coses #7189
- #7168 

Also fixes the issue introduced in https://github.com/GreptimeTeam/greptimedb/pull/7168, when column metadata contains extensions, we cannot assume the serialized bytes of column metadata remain the same across datanodes, so we need to decode the column metadata and compare.

https://github.com/GreptimeTeam/greptimedb/blob/fb92e4d0b2ebc1d4b81aacd95ac12eaf61baec77/src/common/meta/src/ddl/utils.rs#L458-L463

See the failures in https://github.com/GreptimeTeam/greptimedb/actions/runs/19119270322/job/54638226708

## PR Checklist

### DIFF
--- a/src/common/meta/src/ddl/utils.rs
+++ b/src/common/meta/src/ddl/utils.rs
@@ -442,7 +442,7 @@ pub fn extract_column_metadatas(
     results: &mut [RegionResponse],
     key: &str,
 ) -> Result<Option<Vec<ColumnMetadata>>> {
-    let schemas = results
+    let mut schemas = results
         .iter_mut()
         .map(|r| r.extensions.remove(key))
         .collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


WenyXu:
> we cannot assume the serialized bytes of column metadata remain the same across datanodes

Question: Why might the column metadata differ between datanodes?

sunng87:
Because it's backed by a hashmap, the order of entries are not guaranteed

PR #7203, opened 2025-11-07
feat: gc worker only local regions&test

discord9:

- make gc worker only do local regions(hence can update manifest)
- make gc's `enable` option actually work(shouldn't turn on for now)
- add a image for gc worker race condition with region-migration
- some unit tests for gc worker

## PR Checklist

### DIFF
--- a/src/mito2/src/manifest/action.rs
+++ b/src/mito2/src/manifest/action.rs
@@ -18,7 +18,6 @@ use std::collections::{HashMap, HashSet};
 use std::time::Duration;
 
 use chrono::Utc;
-use common_telemetry::warn;
 use serde::{Deserialize, Serialize};
 use snafu::{OptionExt, ResultExt};
 use store_api::ManifestVersion;
@@ -303,23 +302,6 @@ impl RemovedFilesRecord {
         }
 
         self.removed_files.retain(|fs| !fs.file_ids.is_empty());
-    }
-
-    /// Count the number of files removed after the given timestamp. Also return the minimum
-    /// timestamp of all removed files.
-    fn file_removed_cnt_after(&self, t_ms: i64) -> (u64, Option<i64>) {
-        let mut cnt = 0;
-        let mut min_ts_after: Option<i64> = None;
-        for record in &self.removed_files {
-            if record.removed_at >= t_ms {
-                cnt += record.file_ids.len();
-                min_ts_after = match min_ts_after {
-                    Some(ts) => Some(ts.min(record.removed_at)),
-                    None => Some(record.removed_at),
-                };
-            }
-        }
-        (cnt as u64, min_ts_after)
     }
 
     pub fn update_file_removed_cnt_to_stats(&self, stats: &ManifestStats) {
--- a/src/mito2/src/manifest/manager.rs
+++ b/src/mito2/src/manifest/manager.rs
@@ -54,16 +54,10 @@ pub struct RegionManifestOptions {
 
 /// Options for updating `removed_files` field in [RegionManifest].
 #[derive(Debug, Clone)]
+#[cfg_attr(any(test, feature = "test"), derive(Default))]
 pub struct RemoveFileOptions {
     /// Whether GC is enabled. If not, the removed files should always be empty when persisting manifest.
     pub enable_gc: bool,
-}
-
-#[cfg(any(test, feature = "test"))]
-impl Default for RemoveFileOptions {
-    fn default() -> Self {
-        Self { enable_gc: false }
-    }
 }
 
 // rewrite note:


src/mito2/src/compaction/compactor.rs
```diff
@@ -172,21 +172,20 @@ pub async fn open_compaction_region(
...
-            Default::default(),
-        )
-        .await?
-        .context(EmptyRegionDirSnafu {
-            region_id: req.region_id,
-            region_dir: &region_dir_from_table_dir(&req.table_dir, req.region_id, req.path_type),
-        })?
+        RegionManifestManager::open(region_manifest_options, &Default::default())
+            .await?
+            .context(EmptyRegionDirSnafu {
```
MichaelScofield:
```suggestion
            .with_context(|| EmptyRegionDirSnafu {
```

src/mito2/src/compaction/compactor.rs
```diff
@@ -172,21 +172,16 @@ pub async fn open_compaction_region(
             compress_type: manifest_compress_type(mito_config.compress_manifest),
             checkpoint_distance: mito_config.manifest_checkpoint_distance,
             remove_file_options: RemoveFileOptions {
-                keep_count: mito_config.experimental_manifest_keep_removed_file_count,
-                keep_ttl: mito_config.experimental_manifest_keep_removed_file_ttl,
+                gc_enabled: mito_config.gc.enable,
```
evenyag: Should we rename it to `enable_gc`?
```suggestion
                enable_gc: mito_config.gc.enable,
```

src/mito2/src/manifest/action.rs
```diff
@@ -294,6 +294,62 @@ pub struct RemovedFilesRecord {
...
+        self.removed_files.retain(|fs| !fs.file_ids.is_empty());
+    }
+
+    /// Count the number of files removed after the given timestamp. Also return the minimum
+    /// timestamp of all removed files.
+    fn file_removed_cnt_after(&self, t_ms: i64) -> (u64, Option<i64>) {
+        let mut cnt = 0;
+        let mut min_ts_after: Option<i64> = None;
+        for record in &self.removed_files {
+            if record.removed_at >= t_ms {
```
evenyag: nitpick: Maybe we can use `BTreeMap::range()`.

src/mito2/src/worker.rs
```diff
@@ -340,6 +340,8 @@ impl WorkerGroup {
         time_provider: TimeProviderRef,
         partition_expr_fetcher: PartitionExprFetcherRef,
     ) -> Result<WorkerGroup> {
+        use std::sync::Arc;
```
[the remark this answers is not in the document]
evenyag: Why do we need to import Arc here?

src/store-api/src/region_engine.rs
```diff
@@ -497,6 +497,8 @@ pub enum RegionManifestInfo {
     Mito {
         manifest_version: u64,
         flushed_entry_id: u64,
+        /// Number of files removed per hour.
+        file_removal_rate: u64,
```
evenyag: Should we just report the total number of files removed instead of computing a rate in code? Like other region statistics.

Also we may rename that to `file_removal_per_hour` for the rate.

src/mito2/src/manifest/manager.rs
```diff
@@ -54,21 +54,15 @@ pub struct RegionManifestOptions {
...
-    /// Duration to keep removed files in manifest's `removed_files` field before also
-    /// remove them from `removed_files`. Only remove files when both `keep_count` and `keep_duration` is reached.
-    pub keep_ttl: std::time::Duration,
+    /// Whether GC is enabled. If not, the removed files should always be empty when persisting manifest.
+    pub gc_enabled: bool,
 }
 
 #[cfg(any(test, feature = "test"))]
 impl Default for RemoveFileOptions {
+    #[allow(deprecated)]
```
evenyag: Do we need to add allow deprecated here?

src/mito2/src/gc.rs
```diff
@@ -401,40 +387,29 @@ impl LocalGcWorker {
...
-        )
-        .await?
-        .context(EmptyRegionDirSnafu {
-            region_id,
-            region_dir: &region_dir_from_table_dir(table_dir, region_id, path_type),
-        })
+        let mut manager = region.manifest_ctx.manifest_manager.write().await;
+        let mut manifest = (*manager.manifest()).clone();
+        manifest.removed_files.clear_deleted_files(deleted_files);
+        manager.set_manifest(Arc::new(manifest));
```
evenyag: Can we wrap a method in the manager for this?

src/mito2/src/gc.rs
```diff
@@ -401,40 +387,29 @@ impl LocalGcWorker {
...
+        debug!("Updated manifest to clear deleted files");
```
evenyag: We should log the region id at least.

src/mito2/src/manifest/action.rs
discord9: > nitpick: Maybe we can use `BTreeMap::range()`.

it's a `Vec` in manifest so difficult to use `BTreeMap::range()` here

src/mito2/src/manifest/manager.rs
```diff
@@ -534,6 +540,17 @@ impl RegionManifestManager {
         Ok(version)
     }
 
+    /// Clear deleted files from manifest's `removed_files` field without update version. Notice if datanode exit before checkpoint then new manifest by open region may still contain these deleted files, which is acceptable for gc process.
+    pub fn clear_deleted_files(&mut self, deleted_files: Vec<FileId>) {
+        let mut manifest = (*self.manifest()).clone();
+        manifest.removed_files.clear_deleted_files(deleted_files);
+        self.set_manifest(Arc::new(manifest));
+    }
```
evenyag, line 548: How often will we call this method? We cloning the whole RegionManifest so it may be costly if we call it too frequently.

evenyag (approved):
LGTM. I left one question.

src/mito2/src/manifest/manager.rs
discord9: > How often will we call this method? We cloning the whole RegionManifest so it may be costly if we call it too frequently.

currently only after a gc run is done, so very infrequent

PR #7206, opened 2025-11-07
feat: support altering sst format for a table

evenyag:

Supports changing the sst format type for a table from `primary_key` to `flat`.
Note that we can't convert the format from `flat` to `primary_key`.

## PR Checklist

### DIFF
--- a/src/mito2/src/memtable.rs
+++ b/src/mito2/src/memtable.rs
@@ -399,8 +399,6 @@ impl MemtableBuilderProvider {
     }
 
     pub(crate) fn builder_for_options(&self, options: &RegionOptions) -> MemtableBuilderRef {
-        common_telemetry::info!("Builder for options: {:?}", options);
-
         let dedup = options.need_dedup();
         let merge_mode = options.merge_mode();
         let flat_format = options


src/mito2/src/memtable.rs
```diff
@@ -399,6 +399,8 @@ impl MemtableBuilderProvider {
     }
 
     pub(crate) fn builder_for_options(&self, options: &RegionOptions) -> MemtableBuilderRef {
+        common_telemetry::info!("Builder for options: {:?}", options);
```
fengjiachun: Does it need to be at the info level? Will there be a lot of logs?

src/mito2/src/worker/handle_alter.rs
```diff
@@ -57,36 +58,38 @@ impl<S> RegionWorkerLoop<S> {
...
-                    keys.iter().map(Into::into).collect(),
-                ) {
-                    Ok(_) => sender.send(Ok(0)),
-                    Err(e) => sender.send(Err(e).context(InvalidMetadataSnafu)),
+                keys.iter().map(Into::into).collect()
+            }
+            _ => Vec::new(),
+        };
+        if !set_options.is_empty() {
+            match self.handle_alter_region_options_fast(&region, version, set_options) {
```
fengjiachun, line 76: Moment T1: ALTER request → fast path update options → trigger flush  
Moment T2: Flush completed → reschedule ALTER  
Moment T3: Second entry into fast path → options are already the same → return early  
          ↑  
          Is this a bug? It should continue executing `handle_alter_region_with_empty_memtable()`

evenyag: It only triggers flush when there are options (ALTER format) that need to be applied after flush.

src/mito2/src/memtable.rs
evenyag: Fixed in https://github.com/GreptimeTeam/greptimedb/pull/7206/commits/3ed2b8d6ec85ff2e4caded4f9fb061bff65eb145

PR #7223, opened 2025-11-12
fix: obtain system time after fetching lease values

WenyXu:

This PR fixes a bug in the `build_active_filter` function where the system time was acquired outside the closure, causing the filter to use a stale timestamp instead of the current time when evaluating items. This bug occasionally caused "The target peer is unavailable temporally" errors during region migration.

### Problem
The `build_active_filter` function had a critical timing bug:
1. The bug: The now timestamp was captured outside the closure when the filter was created
2. The issue: When `lookup_datanode_peer` uses this filter, it may list lease values from storage that have timestamps newer than the pre-captured now value
3. The consequence: The filter would incorrectly calculate elapsed time, treating active datanodes as inactive because their lease timestamps appeared to be "in the future" relative to the stale now

This race condition was particularly problematic in `lookup_datanode_peer`, where the time gap between filter creation and actual lease value listing could cause the system to incorrectly reject valid, active datanodes. In rare cases during manual region migration, this would manifest as "The target peer is unavailable temporally" errors even when the target datanode was fully operational.

### Changes
Refactored the filter interface:
- Changed `build_active_filter` signature to accept now as a parameter to the returned closure
- The filter now has signature `impl Fn(i64, &T) -> bool` instead of `impl Fn(&T) -> bool`
- This makes it explicit that callers must provide the timestamp at the point of filtering

## PR Checklist

### DIFF
--- a/src/meta-srv/src/discovery/utils.rs
+++ b/src/meta-srv/src/discovery/utils.rs
@@ -19,7 +19,7 @@ use common_meta::DatanodeId;
 use common_meta::cluster::NodeInfo;
 use common_meta::kv_backend::KvBackendRef;
 use common_meta::peer::Peer;
-use common_time::util::{DefaultSystemTimer, SystemTimer};
+use common_time::util::SystemTimer;
 use common_workload::DatanodeWorkloadType;
 use snafu::ResultExt;
 
@@ -59,6 +59,7 @@ pub fn build_active_filter<T: LastActiveTs>(active_duration: Duration) -> impl F
 
 /// Returns the alive datanodes.
 pub async fn alive_datanodes(
+    timer: &impl SystemTimer,
     accessor: &impl LeaseValueAccessor,
     active_duration: Duration,
     condition: Option<fn(&NodeWorkloads) -> bool>,
@@ -66,7 +67,7 @@ pub async fn alive_datanodes(
     let active_filter = build_active_filter(active_duration);
     let condition = condition.unwrap_or(|_| true);
     let lease_values = accessor.lease_values(LeaseValueType::Datanode).await?;
-    let now = DefaultSystemTimer.current_time_millis();
+    let now = timer.current_time_millis();
     Ok(lease_values
         .into_iter()
         .filter_map(|(peer_id, lease_value)| {
@@ -81,6 +82,7 @@ pub async fn alive_datanodes(
 
 /// Returns the alive flownodes.
 pub async fn alive_flownodes(
+    timer: &impl SystemTimer,
     accessor: &impl LeaseValueAccessor,
     active_duration: Duration,
     condition: Option<fn(&NodeWorkloads) -> bool>,
@@ -88,7 +90,7 @@ pub async fn alive_flownodes(
     let active_filter = build_active_filter(active_duration);
     let condition = condition.unwrap_or(|_| true);
     let lease_values = accessor.lease_values(LeaseValueType::Flownode).await?;
-    let now = DefaultSystemTimer.current_time_millis();
+    let now = timer.current_time_millis();
     Ok(lease_values
         .into_iter()
         .filter_map(|(peer_id, lease_value)| {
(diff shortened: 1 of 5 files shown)


src/meta-srv/src/discovery/utils.rs
```diff
@@ -49,16 +49,14 @@ pub trait LastActiveTs {
...
-/// so the "now" reference point is fixed when this function is called.
 pub fn build_active_filter<T: LastActiveTs>(
     timer: impl SystemTimer,
     active_duration: Duration,
 ) -> impl Fn(&T) -> bool {
-    let now = timer.current_time_millis();
-    let active_duration = active_duration.as_millis() as u64;
     move |item: &T| {
+        // The `now` must be constructed inside the closure to avoid using the stale timestamp.
+        let now = timer.current_time_millis();
```
fengjiachun: It's better to pass `now` as an argument through the closure.

WenyXu: [5116a6a](https://github.com/GreptimeTeam/greptimedb/pull/7223/commits/5116a6ad56f5835bf463fcbb6595e08d4417003e)

src/meta-srv/src/discovery/utils.rs
```diff
@@ -70,14 +63,15 @@ pub async fn alive_datanodes(
...
     condition: Option<fn(&NodeWorkloads) -> bool>,
 ) -> Result<Vec<Peer>> {
-    let active_filter = build_active_filter(DefaultSystemTimer, active_duration);
+    let active_filter = build_active_filter(active_duration);
     let condition = condition.unwrap_or(|_| true);
-    Ok(accessor
-        .lease_values(LeaseValueType::Datanode)
-        .await?
+    let lease_values = accessor.lease_values(LeaseValueType::Datanode).await?;
+    // The `now` should be obtained after lease_value is obtained to avoid using the stale timestamp.
```
fengjiachun: This comment seems unnecessary; can we delete it?

src/meta-srv/src/discovery/utils.rs
```diff
@@ -92,14 +86,15 @@ pub async fn alive_flownodes(
...
     condition: Option<fn(&NodeWorkloads) -> bool>,
 ) -> Result<Vec<Peer>> {
-    let active_filter = build_active_filter(DefaultSystemTimer, active_duration);
+    let active_filter = build_active_filter(active_duration);
     let condition = condition.unwrap_or(|_| true);
-    Ok(accessor
-        .lease_values(LeaseValueType::Flownode)
-        .await?
+    let lease_values = accessor.lease_values(LeaseValueType::Flownode).await?;
+    // The `now` should be obtained after lease_value is obtained to avoid using the stale timestamp.
```
fengjiachun: Same as above

src/meta-srv/src/discovery/utils.rs
```diff
@@ -113,13 +108,14 @@ pub async fn alive_frontends(
     lister: &impl NodeInfoAccessor,
     active_duration: Duration,
 ) -> Result<Vec<Peer>> {
-    let active_filter = build_active_filter(DefaultSystemTimer, active_duration);
-    Ok(lister
-        .node_infos(NodeInfoType::Frontend)
-        .await?
+    let active_filter = build_active_filter(active_duration);
+    let node_infos = lister.node_infos(NodeInfoType::Frontend).await?;
+    // The `now` should be obtained after lease_value is obtained to avoid using the stale timestamp.
```
fengjiachun: +1

src/meta-srv/src/discovery/utils.rs
```diff
@@ -134,11 +130,14 @@ pub async fn alive_datanode(
...
 ) -> Result<Option<Peer>> {
-    let active_filter = build_active_filter(DefaultSystemTimer, active_duration);
-    let v = lister
+    let active_filter = build_active_filter(active_duration);
+    let lease_value = lister
         .lease_value(LeaseValueType::Datanode, peer_id)
-        .await?
-        .filter(|(_, lease)| active_filter(lease))
+        .await?;
+    // The `now` should be obtained after lease_value is obtained to avoid using the stale timestamp.
```
fengjiachun: +1

fengjiachun (approved):
Rest LGTM

src/meta-srv/src/discovery/utils.rs
WenyXu: [f614766](https://github.com/GreptimeTeam/greptimedb/pull/7223/commits/f6147662acead07b508bee914f520f925c6d67f4)

PR #7226, opened 2025-11-14
chore: add `tls-watch` option in cmd

shuiyisong:

As the title suggests, add a `tls-watch` option in the standalone and the frontend's cmd option to support watch TLS file change and reload when addr is specified in the cmd option too.
Note, the option only works for MySQL and pg protocol.

## PR Checklist

### DIFF
--- a/src/servers/src/tls.rs
+++ b/src/servers/src/tls.rs
@@ -88,7 +88,7 @@ impl TlsOption {
             tls_option.key_path = key_path
         };
 
-        tls_option.watch = watch
+        tls_option.watch = watch;
 
         tls_option
     }


waynexia (approved):

src/servers/src/tls.rs
```diff
@@ -83,6 +88,10 @@ impl TlsOption {
             tls_option.key_path = key_path
         };
 
+        if watch {
+            tls_option.watch = watch
+        };
```
evenyag: Should we set it directly?
```suggestion
        tls_option.watch = watch
```

PR #7230, opened 2025-11-14
feat: reloadable tls client config

shuiyisong:

This patch introduces the ability for `channel_manager` to use a re-loadable client TLS config to automatically reload the TLS files when changed.
The `ReloadableTlsServerConfig` is now extracted into `reloadable_tls` for reuse.

## PR Checklist

### DIFF
--- a/src/common/grpc/src/reloadable_tls.rs
+++ b/src/common/grpc/src/reloadable_tls.rs
@@ -136,10 +136,10 @@ where
     info!("Spawning background task for watching TLS cert/key file changes");
     std::thread::spawn(move || {
         let _watcher = watcher;
-        while let Ok(res) = rx.recv() {
-            if let Ok(event) = res {
-                match event.kind {
-                    EventKind::Modify(_) | EventKind::Create(_) => {
+        loop {
+            match rx.recv() {
+                Ok(Ok(event)) => {
+                    if let EventKind::Modify(_) | EventKind::Create(_) = event.kind {
                         info!("Detected TLS cert/key file change: {:?}", event);
                         if let Err(err) = tls_config_for_watcher.reload() {
                             error!("Failed to reload TLS config: {}", err);
@@ -148,7 +148,12 @@ where
                             on_reload();
                         }
                     }
-                    _ => {}
+                }
+                Ok(Err(err)) => {
+                    error!("Failed to watch TLS cert/key file: {}", err);
+                }
+                Err(err) => {
+                    error!("TLS cert/key file watcher channel closed: {}", err);
                 }
             }
         }


src/common/grpc/src/reloadable_tls.rs
```diff
@@ -0,0 +1,157 @@
...
+    // Watch all paths returned by the TlsConfigLoader
+    for path in tls_config.get_tls_option().watch_paths() {
+        watcher
+            .watch(path, RecursiveMode::NonRecursive)
+            .with_context(|_| FileWatchSnafu {
+                path: path.display().to_string(),
+            })?;
+    }
+
+    std::thread::spawn(move || {
```
MichaelScofield, line 137: Spawning a thread is not trivial, better add an info log before it.

src/servers/src/tls.rs
```diff
@@ -149,96 +147,34 @@ impl TlsOption {
...
-                        }
-                    }
-                    _ => {}
-                }
-            }
+/// Convenience function for watching server TLS configuration
+pub fn maybe_watch_server_tls_config(
+    tls_server_config: Arc<ReloadableTlsServerConfig>,
+) -> Result<()> {
+    common_grpc::reloadable_tls::maybe_watch_tls_config(tls_server_config, || {}).map_err(|e| {
```
MichaelScofield, line 173: If server TLS config is changed, do we need to proactively close all connected client channels?

shuiyisong: The newly request would be served using the new TLS config in MySQL and PG, but the connected/alive connection stays unaffected. What do you think, should we actively close the connection to force refresh TLS change? @sunng87

src/common/grpc/src/reloadable_tls.rs
```diff
@@ -0,0 +1,158 @@
...
+            .watch(path, RecursiveMode::NonRecursive)
+            .with_context(|_| FileWatchSnafu {
+                path: path.display().to_string(),
+            })?;
+    }
+
+    info!("Spawning background task for watching TLS cert/key file changes");
+    std::thread::spawn(move || {
+        let _watcher = watcher;
+        while let Ok(res) = rx.recv() {
```
fengys1996: nit: Should we ignore the error here? Or add log

src/servers/src/tls.rs
sunng87: I think we don't need to. The previous connection should work seamlessly because all state is already in memory.

sunng87:
To be clear, in this patch we only changed intra-cluster grpc client to reload certificate. The grpc server is not covered, right?

shuiyisong:
> To be clear, in this patch we only changed intra-cluster grpc client to reload certificate. The grpc server is not covered, right?

Yes, this only change the gRPC client(channel manager) to be able to reload the TLS change. The gRPC server of the database is not changed.

sunng87:
This can introduce an issue if all clients reloads certs automatically but server doesn't. The updated client will not be able to connect to server without a restart.

MichaelScofield (approved):

src/servers/src/tls.rs
MichaelScofield: But doesn't this make TLS reloading less meaningful? Is the primary concern that the server reload TLS about potential certificate leak?

sunng87 (approved):
Confirmed with @shuiyisong , this will be used for communication between database and external services like authentication providers or something.

But we still need to ensure the cert reload will work if intra-cluster grpc is configured with tls.

PR #7248, opened 2025-11-18
refactor: make json value use json type

MichaelScofield:

instead of being backed by `ConcreteDataType` directly, which makes json value verbose to operate its type

also the new json type is precisely corresponding to the json value, making codes lesser error-prone

## PR Checklist

### DIFF
--- a/src/datatypes/src/types/json_type.rs
+++ b/src/datatypes/src/types/json_type.rs
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::{BTreeMap, HashMap};
+use std::collections::BTreeMap;
+use std::fmt::{Display, Formatter};
 use std::str::FromStr;
 use std::sync::Arc;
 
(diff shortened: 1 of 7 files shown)


src/datatypes/src/types/json_type.rs
```diff
@@ -57,22 +174,22 @@ impl JsonType {
...
 
-    pub(crate) fn new_native(native: ConcreteDataType) -> Self {
+    pub(crate) fn new_native(native: JsonNativeType) -> Self {
         Self {
             format: JsonFormat::Native(Box::new(native)),
         }
     }
 
-    pub(crate) fn native_type(&self) -> ConcreteDataType {
+    pub(crate) fn native_type(&self) -> &JsonNativeType {
```
sunng87, line 183: do we really need to change return type of this function ot `JsonNativeType`? I think this function is to describe what the underlying data type of json format.

MichaelScofield: Yes, the `JsonNativeType` is exactly the underlying data type of json value. Once needed, it can convert to `ConcreteDataType`.

sunng87 (approved):

PR #7252, opened 2025-11-18
feat: gc scheduler ctx&procedure

discord9:

as title, gc scheduler's ctx&procedure, provide all op that gc scheduler ever needed, and useful for mock test(the full gc scheduler's pr is too large(5000+loc) so break into smaller pieces)

## PR Checklist

### DIFF
--- a/src/meta-srv/src/gc/ctx.rs
+++ b/src/meta-srv/src/gc/ctx.rs
@@ -33,8 +33,7 @@ use store_api::storage::{FileId, FileRefsManifest, GcReport, RegionId};
 use table::metadata::TableId;
 
 use crate::cluster::MetaPeerClientRef;
-use crate::error;
-use crate::error::{Result, TableMetadataManagerSnafu};
+use crate::error::{self, Result, TableMetadataManagerSnafu, UnexpectedSnafu};
 use crate::gc::Region2Peers;
 use crate::gc::procedure::GcRegionProcedure;
 use crate::handler::HeartbeatMailbox;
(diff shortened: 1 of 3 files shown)


src/meta-srv/src/gc/ctx.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+#[async_trait::async_trait]
+impl SchedulerCtx for DefaultGcSchedulerCtx {
+    async fn get_table_to_region_stats(&self) -> Result<HashMap<TableId, Vec<RegionStat>>> {
+        let dn_stats = self.meta_peer_client.get_all_dn_stat_kvs().await?;
+        let mut table_to_region_stats: HashMap<TableId, Vec<RegionStat>> = HashMap::new();
+        for (_dn_id, stats) in dn_stats {
+            let mut stats = stats.stats;
+            stats.sort_by_key(|s| s.timestamp_millis);
```
MichaelScofield: Seems you can use the more efficient "max" to find the latest stat?

src/meta-srv/src/gc/ctx.rs
```diff
@@ -0,0 +1,347 @@
...
+                for follower in followers {
+                    datanode_regions
+                        .entry(follower.clone())
+                        .or_default()
+                        .push(*region_id);
+                }
+            } else {
+                return error::UnexpectedSnafu {
+                    violated: format!(
+                        "region_routes{region_routes:?} does not contain region_id: {region_id}",
```
MichaelScofield:
```suggestion
                        "region_routes: {region_routes:?} does not contain region_id: {region_id}",
```

src/mito2/src/gc.rs
```diff
@@ -316,38 +316,42 @@ impl LocalGcWorker {
         let region_id = manifest.metadata.region_id;
         let current_files = &manifest.files;
 
+        let in_manifest_file_cnt = current_files.len();
```
MichaelScofield: Seems most of the following new variables are only used in debug logging, better move them into the `debug!`.

src/meta-srv/src/gc/ctx.rs
```diff
@@ -0,0 +1,347 @@
...
+        let res = watcher::wait(&mut watcher)
+            .await
+            .context(error::WaitProcedureSnafu)?
+            .with_context(|| error::UnexpectedSnafu {
+                violated: format!(
+                    "GC procedure {id} successfully completed but no result returned"
+                ),
+            })?;
+
+        let gc_report = GcRegionProcedure::cast_result(res)?;
```
MichaelScofield, line 294: Why not simply return the concrete typed value?

discord9: > Why not simply return the concrete typed value?

res is `res: Arc<dyn Any + Send + Sync>` so need deal with it

MichaelScofield (approved):

src/meta-srv/src/gc/procedure.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// only write lock the regions not catatlog/schema because it can run concurrently with other procedures(i.e. drop database/table)
+    /// TODO:(discord9): integration test to verify this
+    fn lock_key(&self) -> LockKey {
+        let lock_key: Vec<_> = self
+            .data
+            .gc_regions
+            .regions
+            .iter()
+            .sorted() // sort to have a deterministic lock order
+            .map(|id| RegionLock::Write(*id).into())
```
WenyXu: A read lock should be fine, as it doesn't modify any metadata.

src/meta-srv/src/gc/procedure.rs
```diff
@@ -0,0 +1,184 @@
...
+use serde::{Deserialize, Serialize};
+use snafu::ResultExt as _;
+use store_api::storage::GcReport;
+
+use crate::error::{self, Result, SerializeToJsonSnafu};
+use crate::handler::HeartbeatMailbox;
+use crate::service::mailbox::{Channel, MailboxRef};
+
+/// TODO(discord9): another procedure which do both get file refs and gc regions.
+pub struct GcRegionProcedure {
```
WenyXu, line 39: For each procedure, we should invoke `ProcedureManager::register_loader` when starting the metasrv, so that the `ProcedureManager` can recover previously persisted procedures. For non-recoverable procedures, we may register a `NoopLoader` instead, which returns an error by default.

PR #7257, opened 2025-11-18
fix: dynamic reload tracing layer loses trace id

waynexia:

`tracing_subscriber::reload::Layer` prevents the trace layer from casting to the original item, which (I guess) will make the tracer not work as expected.

([source](https://docs.rs/tracing-subscriber/0.3.20/src/tracing_subscriber/reload.rs.html#187-204))

image

The proposed new wrapper is almost identical to the above one, but with a concrete inner type and can be downcast back. It also uses an `RwLock` as well as the above one.

## PR Checklist

### DIFF
--- a/src/common/telemetry/src/logging.rs
+++ b/src/common/telemetry/src/logging.rs
@@ -67,27 +67,18 @@ impl TraceReloadHandle {
         Self { inner }
     }
 
-    pub fn reload(&self, new_layer: Option<OtelTraceLayer>) -> Result<(), TraceReloadError> {
-        let mut guard = self.inner.write().map_err(|_| TraceReloadError)?;
+    pub fn reload(&self, new_layer: Option<OtelTraceLayer>) {
+        let mut guard = self.inner.write().unwrap();
         *guard = new_layer;
         drop(guard);
 
         callsite::rebuild_interest_cache();
-        Ok(())
     }
 }
 
-#[derive(Debug)]
-pub struct TraceReloadError;
-
-impl std::fmt::Display for TraceReloadError {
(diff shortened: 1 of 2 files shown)


src/common/telemetry/src/logging.rs
```diff
@@ -52,15 +54,185 @@ pub const DEFAULT_LOGGING_DIR: &str = "logs";
...
+            .unwrap_or_else(tracing::subscriber::Interest::always)
+    }
+
+    fn enabled(
+        &self,
+        metadata: &tracing::Metadata<'_>,
+        ctx: tracing_subscriber::layer::Context<'_, DynSubscriber>,
+    ) -> bool {
+        self.with_layer(|layer| layer.enabled(metadata, ctx))
+            .unwrap_or(true)
```
evenyag: Should we return `false` in `unwrap_or`?
```suggestion
            .unwrap_or(false)
```

src/common/telemetry/src/logging.rs
```diff
@@ -52,15 +54,185 @@ pub const DEFAULT_LOGGING_DIR: &str = "logs";
...
+        let _ = self.with_layer(|layer| layer.on_follows_from(span, follows, ctx));
+    }
+
+    fn event_enabled(
+        &self,
+        event: &tracing::Event<'_>,
+        ctx: tracing_subscriber::layer::Context<'_, DynSubscriber>,
+    ) -> bool {
+        self.with_layer(|layer| layer.event_enabled(event, ctx))
+            .unwrap_or(true)
```
evenyag: Should we return false? Looks like it will be true even when the inner is None.
```suggestion
            .unwrap_or(false)
```

discord9 (approved):

PR #7260, opened 2025-11-19
feat: gc get ref from manifest

discord9:

as title, support get file ref from other region's manifest to support repartition

## PR Checklist


waynexia (approved):
this patch LGTM, but depends on how the new param is going to be used, I'm not sure if this change is necessary.

discord9:
> this patch LGTM, but depends on how the new param is going to be used, I'm not sure if this change is necessary.

could be needing to get file ref from both follower region& child/parent regions in repartition so make a difference in api here, will see how useful once both gc scheduler&repartition got merged

Issue #7254, opened 2025-11-18
Failed to insert jdbc Boolean using postgres jdbc java.sql.PreparedStatement#setBoolean  method


donothangbagsfromthehandlebar:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Write Protocols

### Minimal reproduce step

1. Create table

```
create table if not exists jdbc_insert (
    LOGGED_TIME timestamp not null,
    ORIGINAL_TIME timestamp,
    LOG_CATEGORY varchar(30) INVERTED INDEX,
    LOG_ACTION varchar(30) INVERTED INDEX,
    SUCCESS boolean,
    TIME INDEX (ORIGINAL_TIME)   
) with ('append_mode'='true');
```

2. Run the following java program using 42.7.8 postgres jdbc driver and java 21.0.9+10

```
    String url = "jdbc:postgresql://localhost:4003/public";
    Connection conn = DriverManager.getConnection(url);

    TimeZone timeZone = TimeZone.getTimeZone("UTC");
    Calendar cal = Calendar.getInstance(timeZone);

    // conn = m_dataSource.getConnection();

    String insert = """
      INSERT INTO jdbc_insert (LOGGED_TIME, ORIGINAL_TIME, LOG_CATEGORY, LOG_ACTION, SUCCESS)
      VALUES(now() at time zone 'utc',?,?,?,?)
      """;

    PreparedStatement stmt = conn.prepareStatement(insert);

    // stmt.setTimestamp(1, new Timestamp(System.currentTimeMillis()), cal);
    stmt.setLong(1, System.currentTimeMillis());
    stmt.setString(2, "category1");
    stmt.setString(3, "action1");
    stmt.setBoolean(4, true);
    stmt.addBatch();

    stmt.executeBatch();
```

### What did you expect to see?

A successfully inserted row in specified table. 

Note that using the using http://localhost:4000/dashboard/#/dashboard/query and using the insert statement from the error one get a different error, changing to a timestamp string literal instead of `'1763460882420'::int8` also works similar to #7253

```
INSERT INTO jdbc_insert (LOGGED_TIME, ORIGINAL_TIME, LOG_CATEGORY, LOG_ACTION, SUCCESS)
VALUES(now() at time zone 'utc',('1763460882420'::int8),('category1'),('action1'),('TRUE'::boolean))
```

`Cast error: Cannot cast string '1763460882420' to value of Int8 type`

### What did you see instead?

```
[...]

Issue #7253, opened 2025-11-18
Failed to insert jdbc Timestamp using postgres jdbc java.sql.PreparedStatement#setTimestamp(int, java.sql.Timestamp, java.util.Calendar) method


donothangbagsfromthehandlebar:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Write Protocols

### Minimal reproduce step

1. Create table

```
create table if not exists jdbc_insert (
    LOGGED_TIME timestamp not null,
    ORIGINAL_TIME timestamp,
    LOG_CATEGORY varchar(30) INVERTED INDEX,
    LOG_ACTION varchar(30) INVERTED INDEX,
    SUCCESS boolean,
    TIME INDEX (ORIGINAL_TIME)   
) with ('append_mode'='true');
```

2. Run the following java program using 42.7.8 postgres jdbc driver and java 21.0.9+10

```
    String url = "jdbc:postgresql://localhost:4003/public";
    Connection conn = DriverManager.getConnection(url);

    TimeZone timeZone = TimeZone.getTimeZone("UTC");
    Calendar cal = Calendar.getInstance(timeZone);

    // conn = m_dataSource.getConnection();

    String insert = """
      INSERT INTO jdbc_insert (LOGGED_TIME, ORIGINAL_TIME, LOG_CATEGORY, LOG_ACTION, SUCCESS)
      VALUES(now() at time zone 'utc',?,?,?,?)
      """;

    PreparedStatement stmt = conn.prepareStatement(insert);

    stmt.setTimestamp(1, new Timestamp(System.currentTimeMillis()), cal);
    // stmt.setLong(1, System.currentTimeMillis());
    stmt.setString(2, "category1");
    stmt.setString(3, "action1");
    stmt.setBoolean(4, true);
    stmt.addBatch();

    stmt.executeBatch();
```

### What did you expect to see?

A successfully inserted row in specified table. 

Note that using the using http://localhost:4000/dashboard/#/dashboard/query and using the insert statement from the error below it works as expected.

```
INSERT INTO jdbc_insert (LOGGED_TIME, ORIGINAL_TIME, LOG_CATEGORY, LOG_ACTION, SUCCESS)
VALUES(now() at time zone 'utc',('2025-11-18 10:04:15.081+00'),('category1'),('action1'),('TRUE'::boolean))
```

### What did you see instead?

```
[...]

PR #7270, opened 2025-11-21
fix!: align numeric type aliases with PostgreSQL and MySQL

killme2008:

part of #7254 #7253 

But doesn't fix them.

As the title said.

added/updated mappings:

- INT2 → SMALLINT
- INT4 → INTEGER / INT
- INT8 → BIGINT
- FLOAT4 -> FLOAT
- FLOAT8 -> DOUBLE

PostgreSQL natively supports these aliases MySQL parser accepts them and maps internally to standard integer types.

## PR Checklist

### DIFF
--- a/src/sql/src/statements/transform/type_alias.rs
+++ b/src/sql/src/statements/transform/type_alias.rs
@@ -39,8 +39,8 @@ use crate::statements::{TimezoneInfo, sql_data_type_to_concrete_data_type};
 ///  - `INT2` for `smallint`
 ///  - `INT4` for `int`
 ///  - `INT8` for `bigint`
-///  - `float4` for `float`
-///  - `float8` for `double`
+///  - `FLOAT4` for `float`
+///  - `FLOAT8` for `double`
 ///
 /// Extended type aliases for Arrow types:
 ///  - `INT16` for `smallint`


src/servers/src/postgres/types.rs
```diff
@@ -898,7 +898,13 @@ pub(super) fn parameters_to_scalar_values(
             .and_then(|t| t.as_ref());
 
         let client_type = if let Some(client_given_type) = client_param_types.get(idx) {
-            client_given_type.clone()
+            match (client_given_type, server_type) {
+                (&Type::UNKNOWN, Some(server_type)) => {
```
killme2008, line 902: When the client type is `unknown`, try to use server type.

@sunng87

src/sql/src/statements/transform/type_alias.rs
```diff
@@ -153,20 +162,21 @@ fn replace_type_alias(data_type: &mut DataType) {
...
         DataType::Custom(name, tokens) if name.0.len() == 1 && tokens.is_empty() => {
             get_data_type_by_alias_name(name.0[0].to_string_unquoted().as_str())
         }
-        DataType::Int8(None) => Some(DataType::TinyInt(None)),
+        DataType::Int2(None) => Some(DataType::SmallInt(None)),
+        DataType::Int4(None) => Some(DataType::Int(None)),
+        DataType::Int8(None) => Some(DataType::BigInt(None)),
         DataType::Int16 => Some(DataType::SmallInt(None)),
         DataType::Int32 => Some(DataType::Int(None)),
         DataType::Int64 => Some(DataType::BigInt(None)),
```
sunng87, line 173: I think it will be less confusing by writing it as:

```
DataType::Int2(None) | DataType::Int16 => Some(DataType::SmallInt(None)),
...
```

Although it's still very confusing. It's better to include some comments here.

killme2008: We document the mapping at the above already and will modify the doc https://docs.greptime.com/reference/sql/data-types/#type-aliases

sunng87 (approved):

src/sql/src/statements/transform/type_alias.rs
sunng87: I mean in the code, when Int2 and Int16 appears in the same context, it's especially hard to tell whether Int8 stands for 8-bit or 8-byte

PR #7293, opened 2025-11-25
fix: implement bulk write for time partitions and bulk memtable

evenyag:

Fix incorrect implementation for bulk insert requests when using the bulk memtable.

It adds a method to convert the bulk part to the expected part that the bulk memtable can store.

One remaining issue is that the frontend can't fill the impure default value for the bulk request now.

## PR Checklist


fengjiachun (approved):
LGTM

Issue #7259, opened 2025-11-19
Invalidating read cache cause high cpu usage


evenyag:
### What type of bug is this?

Performance issue

### What subsystems are affected?

Distributed Cluster, Standalone mode

### Minimal reproduce step

This is observed from an environment that has around 100 regions and a large number of entries in the read cache.

The CPU usage is nearly 90% while executing a small number of queries.

### What did you expect to see?

A lower CPU usage.

### What did you see instead?

The cpu usage of datanode increased dramatically while executing queries. The flamegraph showed that the read cache is invalidating entries.

The flamegraph:

https://github.com/user-attachments/assets/6009346f-b67e-4d98-a58e-6b3e6bc3e3da

### What operating system did you use?

Unrelated

### What version of GreptimeDB did you use?

0.12

### Relevant log output and stack trace

```bash

```

killme2008:
Could we close this issue right now? @evenyag

evenyag:
closed by https://github.com/GreptimeTeam/greptimedb/pull/7294

PR #7294, opened 2025-11-25
feat!: download file to fill the cache on write cache miss

evenyag:

https://github.com/GreptimeTeam/greptimedb/issues/7259

This PR refactors the caching architecture by removing the generic LRU read cache layer from
  object storage and replacing it with more specialized caching mechanisms:

  - **Removes** the generic `LruCacheLayer` that previously cached all remote object storage reads
  - **Introduces** a dedicated manifest cache with configurable capacity (`manifest_cache_size`,
  default: 256MB)
  - **Enhances** the file cache with background download capabilities for data and index files
  - **Simplifies** configuration by removing legacy read cache options (`cache_path`,
  `enable_read_cache`, `cache_capacity`)

  ### Key Changes

  #### Caching Architecture

  The old approach used a transparent cache layer at the object store level that cached all reads.
  The new approach uses specialized caches:

  1. **Manifest Cache** (`src/mito2/src/cache/manifest_cache.rs`): Dedicated cache for manifest
  files with improved cleanup logic for empty directories
  2. **File Cache** (`src/mito2/src/cache/file_cache.rs`): Enhanced with background download worker
  that asynchronously fetches files without blocking queries

  #### Background Downloads

  Index appliers (bloom filter, fulltext, inverted) and parquet readers now trigger background
  downloads through `CacheStrategy::maybe_download_background()`. A dedicated worker thread
  processes these requests via a bounded channel, improving query performance by not blocking on
  cache misses.

  #### Configuration

  - **Added**: `region_engine.mito.manifest_cache_size` (default: 256MB)
  - **Removed**: `storage.cache_path`, `storage.enable_read_cache`, `storage.cache_capacity`

  #### Cleanup & Maintenance

  - Automatically cleans up old LRU cache directories on startup
  - Improved empty directory cleanup in manifest cache
  - Removed unused md5 0.7.0 dependency

  ### Testing

  - Updated integration and unit tests
  - Added tests for empty directory cleanup in manifest cache
  - All existing test suites pass

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/index/bloom_filter/applier.rs
+++ b/src/mito2/src/sst/index/bloom_filter/applier.rs
@@ -45,10 +45,10 @@ use crate::error::{
 };
 use crate::metrics::INDEX_APPLY_ELAPSED;
 use crate::sst::file::RegionIndexId;
-use crate::sst::index::{trigger_index_background_download, TYPE_BLOOM_FILTER_INDEX};
 use crate::sst::index::bloom_filter::INDEX_BLOB_TYPE;
 pub use crate::sst::index::bloom_filter::applier::builder::BloomFilterIndexApplierBuilder;
 use crate::sst::index::puffin_manager::{BlobReader, PuffinManagerFactory};
+use crate::sst::index::{TYPE_BLOOM_FILTER_INDEX, trigger_index_background_download};
 
 /// Metrics for tracking bloom filter index apply operations.
 #[derive(Default, Clone)]
--- a/src/mito2/src/sst/index/fulltext_index/applier.rs
+++ b/src/mito2/src/sst/index/fulltext_index/applier.rs
@@ -45,12 +45,12 @@ use crate::error::{
 };
 use crate::metrics::INDEX_APPLY_ELAPSED;
 use crate::sst::file::RegionIndexId;
-use crate::sst::index::{trigger_index_background_download, TYPE_FULLTEXT_INDEX};
 use crate::sst::index::fulltext_index::applier::builder::{FulltextRequest, FulltextTerm};
 use crate::sst::index::fulltext_index::{INDEX_BLOB_TYPE_BLOOM, INDEX_BLOB_TYPE_TANTIVY};
 use crate::sst::index::puffin_manager::{
     PuffinManagerFactory, SstPuffinBlob, SstPuffinDir, SstPuffinReader,
 };
+use crate::sst::index::{TYPE_FULLTEXT_INDEX, trigger_index_background_download};
 
 pub mod builder;
 
--- a/src/mito2/src/sst/index/inverted_index/applier.rs
+++ b/src/mito2/src/sst/index/inverted_index/applier.rs
@@ -41,9 +41,9 @@ use crate::error::{
 };
 use crate::metrics::{INDEX_APPLY_ELAPSED, INDEX_APPLY_MEMORY_USAGE};
 use crate::sst::file::RegionIndexId;
-use crate::sst::index::{trigger_index_background_download, TYPE_INVERTED_INDEX};
 use crate::sst::index::inverted_index::INDEX_BLOB_TYPE;
 use crate::sst::index::puffin_manager::{BlobReader, PuffinManagerFactory};
+use crate::sst::index::{TYPE_INVERTED_INDEX, trigger_index_background_download};
 
 /// Metrics for tracking inverted index apply operations.
 #[derive(Default, Clone)]


config/config.md
```diff
@@ -154,6 +151,8 @@
 | `region_engine.mito.write_cache_ttl` | String | Unset | TTL for write cache. |
 | `region_engine.mito.preload_index_cache` | Bool | `true` | Preload index (puffin) files into cache on region open (default: true).<br/>When enabled, index files are loaded into the write cache during region initialization,<br/>which can improve query performance at the cost of longer startup times. |
 | `region_engine.mito.index_cache_percent` | Integer | `20` | Percentage of write cache capacity allocated for index (puffin) files (default: 20).<br/>The remaining capacity is used for data (parquet) files.<br/>Must be between 0 and 100 (exclusive). For example, with a 5GiB write cache and 20% allocation,<br/>1GiB is reserved for index files and 4GiB for data files. |
+| `region_engine.mito.enable_refill_cache_on_read` | Bool | `true` | Enable refilling cache on read operations (default: true).<br/>When disabled, cache refilling on read won't happen. |
```
killme2008, line 154: I think we should explain what is refilling cache.

src/datanode/src/store.rs
```diff
@@ -47,66 +42,63 @@ pub(crate) async fn new_object_store_without_cache(
...
     );
+    if let Err(e) = clean_temp_dir(&cache_atomic_temp_dir) {
+        warn!(e; "Failed to clean old cache atomic temp directory {}", cache_atomic_temp_dir);
+    }
+}
 
-    Ok(Some(cache_layer))
+pub async fn new_object_store(store: ObjectStoreConfig, data_home: &str) -> Result<ObjectStore> {
+    // Cleans up old LRU read cache directories.
+    clean_old_read_cache(&store, data_home);
```
killme2008, line 90: TODO: Remove this line after the 1.0 release.

src/mito2/src/cache/manifest_cache.rs
```diff
@@ -412,11 +427,16 @@ impl ManifestCache {
...
                 if subdir_empty {
-                    if let Err(e) = std::fs::remove_dir(&path)
-                        && e.kind() != std::io::ErrorKind::NotFound
-                    {
-                        warn!(e; "Failed to remove empty directory {}", path.display());
-                        is_empty = false;
+                    if let Err(e) = std::fs::remove_dir(&path) {
+                        if e.kind() != std::io::ErrorKind::NotFound {
+                            warn!(e; "Failed to remove empty directory {}", path.display());
+                            is_empty = false;
```
killme2008, line 433: Not found doesn't means it's not empty.

src/mito2/src/cache/file_cache.rs
```diff
@@ -576,9 +638,42 @@ impl FileCache {
...
+        remote_path: String,
+        remote_store: ObjectStore,
+        file_size: u64,
+    ) {
+        // Do nothing if background worker is disabled (channel is None)
+        let Some(tx) = &self.download_task_tx else {
+            return;
+        };
+
+        let task = DownloadTask {
```
killme2008, line 662: Should we implement RTII for `DownloadTask` and remove it from the cache in `drop`?

src/mito2/src/sst/index/bloom_filter/applier.rs
```diff
@@ -378,12 +378,27 @@ impl BloomFilterIndexApplier {
         column_id: ColumnId,
         file_size_hint: Option<u64>,
     ) -> Result<BlobReader> {
+        let path_factory = RegionFilePathFactory::new(self.table_dir.clone(), self.path_type);
```
killme2008, line 381: What about extracting a `maybe_download_remote_file` function? This code block appears three times.

src/mito2/src/cache/manifest_cache.rs
evenyag: remove_dir only removes empty dir. If it returns error and the error code isn't not found, then we don't know whether the dir is empty so we set  is_empty  to false. We only set is_empty to true when we ensure the dir is empty.

src/mito2/src/cache/file_cache.rs
evenyag: The download() method will clean the file on failure.

waynexia:
### Code review

Found 4 issues:

1. Missing config check in background downloads (violates user configuration)

The `maybe_download_background` method is called without checking if `enable_refill_cache_on_read` is set to false. This means background downloads happen regardless of user settings, violating the expected behavior.

https://github.com/GreptimeTeam/greptimedb/blob/6138fbbba8e859576edf1c8712b6618a7212c32c/src/mito2/src/sst/index/inverted_index/applier.rs#L324
https://github.com/GreptimeTeam/greptimedb/blob/6138fbbba8e859576edf1c8712b6618a7212c32c/src/mito2/src/sst/index/bloom_filter/applier.rs#L391
https://github.com/GreptimeTeam/greptimedb/blob/6138fbbba8e859576edf1c8712b6618a7212c32c/src/mito2/src/sst/parquet/reader.rs#L337

2. Memory leak in background worker

The background download worker spawned with `tokio::spawn` is never joined or cancelled, causing it to continue running after the cache is dropped, leading to resource leaks.

https://github.com/GreptimeTeam/greptimedb/blob/6138fbbba8e859576edf1c8712b6618a7212c32c/src/mito2/src/cache/file_cache.rs#L332-L345

3. Loss of cache filtering logic

The old `ReadCache` excluded files ending with `_last_checkpoint` from caching, but the new implementation doesn't have equivalent filtering, causing checkpoint files to be cached unnecessarily.

https://github.com/GreptimeTeam/greptimedb/blob/6138fbbba8e859576edf1c8712b6618a7212c32c/src/mito2/src/cache/file_cache.rs#L650-L676

4. Removed TODO comment without proper fix

A TODO comment warning about unnecessary logging was removed along with the warning itself, instead of implementing the proper fix as suggested.

https://github.com/GreptimeTeam/greptimedb/blob/2f6bf8634b041ff5824553ce2b7c9ff15b6f7492/src/mito2/src/cache/write_cache.rs#L76

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #7296, opened 2025-11-25
feat: add batch gc procedure

discord9:

a batch gc regions procedure which include get file ref then call region gc, intended for admin function use only

## PR Checklist

### DIFF
--- a/src/meta-srv/src/gc/procedure.rs
+++ b/src/meta-srv/src/gc/procedure.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use std::any::Any;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -26,7 +26,7 @@ use common_procedure::{
     Context as ProcedureContext, Error as ProcedureError, LockKey, Procedure,
     Result as ProcedureResult, Status,
 };
-use common_telemetry::{debug, error, info};
+use common_telemetry::{debug, error, info, warn};
 use itertools::Itertools as _;
 use serde::{Deserialize, Serialize};
 use snafu::ResultExt as _;
@@ -409,8 +409,9 @@ impl BatchGcProcedure {
         })
     }
 
-    /// Send GC instruction to all datanodes that host the regions
-    async fn send_gc_instructions(&self) -> Result<()> {
+    /// Send GC instruction to all datanodes that host the regions,
+    /// returns regions that need retry.
+    async fn send_gc_instructions(&self) -> Result<Vec<RegionId>> {
         let regions = &self.data.regions;
         let region_routes = &self.data.region_routes;
         let file_refs = &self.data.file_refs;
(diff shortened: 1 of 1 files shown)


src/meta-srv/src/gc/procedure.rs
```diff
@@ -182,3 +248,284 @@ impl Procedure for GcRegionProcedure {
...
+                        error!("Failed to get file references: {}", e);
+                        Err(ProcedureError::external(e))
+                    }
+                }
+            }
+            State::Gcing => {
+                // Send GC instructions to all datanodes
+                match self.send_gc_instructions().await {
+                    Ok(()) => {
+                        debug!(
```
waynexia:
```suggestion
                        info!(
```

src/meta-srv/src/gc/procedure.rs
```diff
@@ -182,3 +248,284 @@ impl Procedure for GcRegionProcedure {
...
+                self.data.state = State::Acquiring;
+                Ok(Status::executing(true))
+            }
+            State::Acquiring => {
+                // Get file references from all datanodes
+                match self.get_file_references().await {
+                    Ok(file_refs) => {
+                        self.data.file_refs = file_refs;
+                        self.data.state = State::Gcing;
+                        Ok(Status::executing(true))
```
evenyag: If we don't want to persist the procedure, we can just set the persist flag to `false`. Then it won't be recovered.

src/meta-srv/src/gc/procedure.rs
```diff
@@ -182,3 +248,284 @@ impl Procedure for GcRegionProcedure {
...
+                timeout,
+                "Batch GC",
+            )
+            .await?;
+
+            // GC successful for this datanode
+            debug!(
+                "GC successful for datanode {} regions {:?}",
+                peer, regions_for_peer
+            );
```
evenyag: We can log in info level.

src/meta-srv/src/gc/procedure.rs
```diff
@@ -13,28 +13,139 @@
...
+                "Datanode {} reported error during GC for regions {:?}: {}",
+                peer, gc_regions, e
+            );
+            Err(error::UnexpectedSnafu {
+                violated: format!(
+                    "Datanode {} reported error during GC for regions {:?}: {}",
+                    peer, gc_regions, e
+                ),
+            }
+            .fail()?)
```
evenyag: Can we remove the outer `Err`?
```suggestion
            error::UnexpectedSnafu {
                violated: format!(
                    "Datanode {} reported error during GC for regions {:?}: {}",
                    peer, gc_regions, e
                ),
            }
            .fail()
```

src/mito2/src/gc.rs
```diff
@@ -209,9 +209,6 @@ impl LocalGcWorker {
 
     /// Get tmp ref files for all current regions
     ///
```
evenyag:
```suggestion
```

evenyag (approved):

PR #7297, opened 2025-11-25
feat: run histogram quantile in safe mode for incomplete data

waynexia:

Implement another safe mode, compared to the existing optimistic path, for PromQL histogram quantile, which would not assume all buckets are present and with the same length. The operator will switch to safe mode automatically when the input data failed with verfication.

## PR Checklist

### DIFF
--- a/src/promql/src/extension_plan/histogram_fold.rs
+++ b/src/promql/src/extension_plan/histogram_fold.rs
@@ -561,6 +561,8 @@ impl HistogramFoldStream {
         Ok(builders)
     }
 
+    /// Determines bucket count using buffered batches, concatenating them to
+    /// detect the first complete bucket that may span batch boundaries.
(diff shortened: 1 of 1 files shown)


src/promql/src/extension_plan/histogram_fold.rs
```diff
@@ -603,6 +677,170 @@ impl HistogramFoldStream {
...
+        &self,
+        vectors: &[VectorRef],
+        row: usize,
+        tag_values: &[ValueRef<'_>],
+    ) -> bool {
+        self.normal_indices
+            .iter()
+            .zip(tag_values.iter())
+            .all(|(idx, expected)| vectors[*idx].get_ref(row) == *expected)
+    }
```
evenyag, line 750: What is a group? Points from the same time series?

PR #7298, opened 2025-11-25
chore: optimize search traces from Grafana

shuiyisong:

For the trace search API, the Grafana Jaeger plugin and the Jaeger UI behave differently. Clicking the trace id on the serarch page 
- Grafana Jaeger plugin: Invoking a new query trace by id request 
- Jaeger UI: Expand the result carried by the search request. No new request is made.

The current implementation comply with the second approach, querying all related span records for the search API.
While large traces can have thousands of spans, using search API on the Grafana Jaeger plugin can easily result in timeout.
This PR introduce the user-agent for search API, optimizing searches from the Grafana Jaeger plugin to limit the number of spans within each trace.

## PR Checklist

### DIFF
--- a/src/frontend/src/instance/jaeger.rs
+++ b/src/frontend/src/instance/jaeger.rs
@@ -57,7 +57,7 @@ use table::table::adapter::DfTableProviderAdapter;
 use crate::instance::Instance;
 
 const DEFAULT_LIMIT: usize = 2000;
-const KEY_RN: &str = "rn";
+const KEY_RN: &str = "greptime_rn";
 
 #[async_trait]
 impl JaegerQueryHandler for Instance {
--- a/src/servers/src/http/jaeger.rs
+++ b/src/servers/src/http/jaeger.rs
@@ -351,21 +351,19 @@ pub struct QueryTraceParams {
 pub enum TraceUserAgent {
     Grafana,
     // Jaeger-UI does not actually send user agent
-    Jaeger,
+    // But it's a jaeger API, so let's treat it as jaeger
     #[default]
-    Unknown,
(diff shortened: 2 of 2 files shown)


sunng87 (approved):

src/frontend/src/instance/jaeger.rs
```diff
@@ -49,12 +50,14 @@ use servers::otlp::trace::{
...
 use session::context::QueryContextRef;
 use snafu::{OptionExt, ResultExt};
+use table::TableRef;
 use table::requests::{TABLE_DATA_MODEL, TABLE_DATA_MODEL_TRACE_V1};
 use table::table::adapter::DfTableProviderAdapter;
 
 use crate::instance::Instance;
 
 const DEFAULT_LIMIT: usize = 2000;
+const KEY_RN: &str = "rn";
```
evenyag: Do we need to use a rare name for this to avoid conflicts?

src/servers/src/http/jaeger.rs
```diff
@@ -340,6 +342,32 @@ pub struct QueryTraceParams {
...
+
+impl From<UserAgent> for TraceUserAgent {
+    fn from(value: UserAgent) -> Self {
+        let ua_str = value.as_str().to_lowercase();
+        if ua_str.contains("grafana") {
+            Self::Grafana
+        } else if ua_str.contains("jaeger") {
+            Self::Jaeger
+        } else {
+            warn!("Unknown user agent: {}", ua_str);
```
evenyag: Should we log it at the warning level now?

evenyag (approved):

PR #7301, opened 2025-11-25
chore: return meaningful message when content type mismatch in otel

killme2008:

Some OpenTelemetry SDKs still send telemetry data in JSON format, which GreptimeDB does not support.

Return a clearer, more informative error message in this case.

## PR Checklist

### DIFF
--- a/src/servers/src/http/loki.rs
+++ b/src/servers/src/http/loki.rs
@@ -43,7 +43,7 @@ use snafu::{OptionExt, ResultExt, ensure};
 use vrl::value::{KeyString, Value as VrlValue};
 
 use crate::error::{
-    DecodeOtlpRequestSnafu, InvalidLokiLabelsSnafu, InvalidLokiPayloadSnafu, ParseJsonSnafu,
+    DecodeLokiRequestSnafu, InvalidLokiLabelsSnafu, InvalidLokiPayloadSnafu, ParseJsonSnafu,
     PipelineSnafu, Result, UnsupportedContentTypeSnafu,
 };
 use crate::http::HttpResponse;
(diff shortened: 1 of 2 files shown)


src/servers/src/http/loki.rs
```diff
@@ -491,8 +491,11 @@ pub struct LokiPbParser {
 impl LokiPbParser {
     pub fn from_bytes(bytes: Bytes) -> Result<Self> {
         let decompressed = prom_store::snappy_decompress(&bytes).unwrap();
-        let req = loki_proto::logproto::PushRequest::decode(&decompressed[..])
-            .context(DecodeOtlpRequestSnafu)?;
+        let req = loki_proto::logproto::PushRequest::decode(&decompressed[..]).context(
+            DecodeOtlpRequestSnafu {
+                content_type: "application/x-protobuf",
+            },
```
shuiyisong: Nit: Loki req is not OTLP

src/servers/src/http/otlp.rs
```diff
@@ -56,16 +74,24 @@ pub async fn metrics(
     State(state): State<OtlpState>,
     Extension(mut query_ctx): Extension<QueryContext>,
     http_opts: OtlpMetricOptions,
+    content_type: Option<TypedHeader<ContentType>>,
     bytes: Bytes,
 ) -> Result<OtlpResponse<ExportMetricsServiceResponse>> {
+    if is_json_content_type(content_type.as_ref().map(|h| &h.0)) {
```
sunng87, line 80: why we check json specifically here? It can be any invalid content type like html or text or something

killme2008: Yes. If that happens, decoding will fail. I chose to test JSON because it’s one of the two formats OTEL supports. I don’t want to check for Protobuf since the MIME header isn’t standardized.

sunng87: Make sense :+1:

src/servers/src/http/loki.rs
killme2008: Yes, need another error.

PR #7302, opened 2025-11-26
refactor: load metadata using offical impl

discord9:

as title, use offical loader for metadata

## PR Checklist

### DIFF
--- a/src/mito2/src/error.rs
+++ b/src/mito2/src/error.rs
@@ -180,15 +180,6 @@ pub enum Error {
         error: ArrowError,
     },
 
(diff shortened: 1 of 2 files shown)


src/mito2/src/error.rs
```diff
@@ -180,6 +180,15 @@ pub enum Error {
         error: ArrowError,
     },
 
+    #[snafu(display("Failed to load parquet metadata, path: {}", path))]
+    LoadParquetMetadata {
```
evenyag: Maybe we can reuse the ReadParquet error?

PR #7305, opened 2025-11-26
feat: mem manager on compaction

fengjiachun:

As @Copilot said.

---

New ideas supplement:  
Based on the approach of streaming reading file contents during the Compaction process (which can be understood as reading and decompressing one RowGroup at a time), it is unreasonable to estimate memory solely based on file size. Therefore, when constructing SST files, I calculated the `max_row_group_uncompressed_size` of the file and wrote it into FileMeta (compatibility has been considered). Before executing Compaction, all related files are traversed to calculate their total `max_row_group_uncompressed_size`. This value is theoretically closely related to peak memory usage during Compaction but does not account for the portion of memory required for data to expand into Arrow format in memory. Theoretically, we need a coefficient to estimate this portion of memory; currently, this coefficient is lacking and may require more practice to obtain empirical values.

## PR Checklist

### DIFF
--- a/src/mito2/src/compaction/memory_manager.rs
+++ b/src/mito2/src/compaction/memory_manager.rs
@@ -19,8 +19,12 @@ use std::{fmt, mem};
 use common_telemetry::debug;
 use humantime::{format_duration, parse_duration};
 use serde::{Deserialize, Serialize};
+use snafu::ensure;
 use tokio::sync::{OwnedSemaphorePermit, Semaphore, TryAcquireError};
 
+use crate::error::{
+    CompactionMemoryLimitExceededSnafu, CompactionMemorySemaphoreClosedSnafu, Result,
+};
 use crate::metrics::{
     COMPACTION_MEMORY_IN_USE, COMPACTION_MEMORY_LIMIT, COMPACTION_MEMORY_REJECTED,
 };
@@ -153,20 +157,33 @@ impl CompactionMemoryManager {
     }
 
     /// Acquires memory, waiting if necessary until enough is available.
-    pub async fn acquire(&self, bytes: u64) -> CompactionMemoryGuard {
+    ///
+    /// # Errors
+    /// - Returns error if requested bytes exceed the total limit
+    /// - Returns error if the semaphore is unexpectedly closed
+    pub async fn acquire(&self, bytes: u64) -> Result<CompactionMemoryGuard> {
         match &self.quota {
-            None => CompactionMemoryGuard::unlimited(),
+            None => Ok(CompactionMemoryGuard::unlimited()),
             Some(quota) => {
                 let permits = bytes_to_permits(bytes);
+
+                // Fail-fast: reject if request exceeds total limit.
+                ensure!(
+                    permits <= quota.limit_permits,
+                    CompactionMemoryLimitExceededSnafu {
+                        requested_bytes: bytes,
+                        limit_bytes: permits_to_bytes(quota.limit_permits),
+                    }
+                );
 
                 let permit = quota
                     .semaphore
                     .clone()
                     .acquire_many_owned(permits)
                     .await
-                    .unwrap();
+                    .map_err(|_| CompactionMemorySemaphoreClosedSnafu.build())?;
                 quota.update_in_use_metric();
-                CompactionMemoryGuard::limited(permit, quota.clone())
+                Ok(CompactionMemoryGuard::limited(permit, quota.clone()))
             }
         }
     }
@@ -429,7 +446,7 @@ mod tests {
             let manager = manager.clone();
             tokio::spawn(async move {
                 // This will block until memory is available
-                let _guard = manager.acquire(bytes).await;
+                let _guard = manager.acquire(bytes).await.unwrap();
             })
         };
 
--- a/src/mito2/src/compaction/task.rs
+++ b/src/mito2/src/compaction/task.rs
@@ -124,9 +124,13 @@ impl CompactionTaskImpl {
                 )
                 .await
                 {
-                    Ok(guard) => {
+                    Ok(Ok(guard)) => {
                         timer.observe_duration();
                         Ok(guard)
+                    }
+                    Ok(Err(e)) => {
+                        timer.observe_duration();
+                        Err(e)
                     }
                     Err(_) => {
                         timer.observe_duration();
@@ -134,13 +138,13 @@ impl CompactionTaskImpl {
                             "Compaction for region {} waited {:?} for {} bytes but timed out",
                             region_id, wait_timeout, requested_bytes
                         );
-                        Err(CompactionMemoryExhaustedSnafu {
+                        CompactionMemoryExhaustedSnafu {
                             region_id,
                             required_bytes: requested_bytes,
                             limit_bytes,
                             policy: format!("wait_timeout({}ms)", wait_timeout.as_millis()),
                         }
-                        .build())
+                        .fail()
                     }
                 }
             }
--- a/src/mito2/src/error.rs
+++ b/src/mito2/src/error.rs
@@ -1035,6 +1035,24 @@ pub enum Error {
     },
 
     #[snafu(display(
+        "Requested compaction memory ({} bytes) exceeds total limit ({} bytes)",
+        requested_bytes,
+        limit_bytes
+    ))]
+    CompactionMemoryLimitExceeded {
+        requested_bytes: u64,
+        limit_bytes: u64,
+        #[snafu(implicit)]
+        location: Location,
+    },
+
+    #[snafu(display("Compaction memory semaphore unexpectedly closed"))]
+    CompactionMemorySemaphoreClosed {
+        #[snafu(implicit)]
+        location: Location,
+    },
+
+    #[snafu(display(
         "Incompatible WAL provider change. This is typically caused by changing WAL provider in database config file without completely cleaning existing files. Global provider: {}, region provider: {}",
         global,
         region
@@ -1300,6 +1318,10 @@ impl ErrorExt for Error {
             ManualCompactionOverride {} => StatusCode::Cancelled,
 
             CompactionMemoryExhausted { .. } => StatusCode::RuntimeResourcesExhausted,
+
+            CompactionMemoryLimitExceeded { .. } => StatusCode::RuntimeResourcesExhausted,
+
+            CompactionMemorySemaphoreClosed { .. } => StatusCode::Unexpected,
 
             IncompatibleWalProviderChange { .. } => StatusCode::InvalidArguments,
 


src/mito2/src/compaction.rs
```diff
@@ -429,31 +457,214 @@ impl CompactionScheduler {
...
+    fn handle_memory_exhausted(
+        &mut self,
+        region_id: RegionId,
+        requested_bytes: u64,
+        task: Box<CompactionTaskImpl>,
+    ) -> Result<()> {
+        let limit_bytes = self.memory_manager.limit_bytes();
+
+        // Check if request exceeds total capacity
+        if self.exceeds_memory_limit(requested_bytes) {
```
v0y4g3r:
```suggestion
        if limit_bytes > 0 && requested_bytes > limit_bytes
```

src/mito2/src/compaction.rs
```diff
@@ -429,31 +457,214 @@ impl CompactionScheduler {
...
+            warn!(
+                "Region {} compaction requires {} bytes but total limit is {} bytes; \
+                 request exceeds capacity, policy: {:?}",
+                region_id, requested_bytes, limit_bytes, self.memory_policy
+            );
+            // Handle based on policy even when exceeding limit
+            return match self.memory_policy {
+                OnExhaustedPolicy::Skip => self.skip_compaction_due_to_memory(
```
v0y4g3r: I'll draft an issue for that case to avoid creating compactions tasks which will never succeed due to memory limit.

src/mito2/src/compaction.rs
```diff
@@ -429,31 +457,214 @@ impl CompactionScheduler {
...
+        };
+
+        let failure = Arc::new(make_error().build());
+        let waiters = mem::take(&mut task.waiters);
+        Self::notify_waiters_error(waiters, region_id, failure.clone());
+        self.finish_region_status(region_id, failure);
+        make_error().fail()
+    }
+
+    fn spawn_memory_waiter(&self, region_id: RegionId, bytes: u64) {
```
v0y4g3r: Maybe use a hashset to dedup the regions waiting here

src/mito2/src/compaction/memory_manager.rs
```diff
@@ -0,0 +1,285 @@
...
+                return;
+            }
+            notified.await;
+        }
+    }
+}
+
+impl Inner {
+    fn on_acquire(&self, permits: u32) {
+        let bytes = permits_to_bytes(permits);
```
v0y4g3r: Can we change `Inner::used_bytes` to `Inner::used_permits` to avoid converting from permits to bytes everytime in `on_acquire` and `release`?

fengjiachun: We cannot do this because we need to use a u32 to acquire permits, but bytes is a u64. See https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.try_acquire_many.

src/mito2/src/compaction.rs
fengjiachun: https://github.com/GreptimeTeam/greptimedb/pull/7305/commits/b9a96d9696bc53bdc216f2b7047a175ece9325c7

src/mito2/src/compaction/memory_manager.rs
```diff
@@ -0,0 +1,285 @@
...
+/// Minimum bytes controlled by one semaphore permit.
+const PERMIT_GRANULARITY_BYTES: u64 = 1 << 20; // 1 MB
+
+/// Defines how the scheduler reacts when compaction cannot acquire enough memory.
+#[derive(Debug, Default, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum OnExhaustedPolicy {
+    /// Wait until enough memory is released.
+    #[default]
+    Wait,
```
evenyag: Is it possible to encounter a deadlock in the wait mode?

src/mito2/src/compaction/memory_manager.rs
```diff
@@ -0,0 +1,285 @@
...
+}
+
+fn bytes_to_permits(bytes: u64) -> u32 {
+    if bytes == 0 {
+        return 0;
+    }
+    bytes
+        .saturating_add(PERMIT_GRANULARITY_BYTES - 1)
+        .saturating_div(PERMIT_GRANULARITY_BYTES)
+        .min(u32::MAX as u64) as u32
```
evenyag, line 396: We should use MAX_PERMITS.
https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#associatedconstant.MAX_PERMITS

src/mito2/src/compaction/memory_manager.rs
```diff
@@ -0,0 +1,285 @@
...
+        self.inner
+            .as_ref()
+            .map(|inner| inner.used_bytes.load(Ordering::Relaxed))
+            .unwrap_or(0)
+    }
+
+    /// Returns available bytes.
+    pub fn available_bytes(&self) -> u64 {
+        self.limit_bytes().saturating_sub(self.used_bytes())
+    }
```
evenyag, line 157: Since we can get available permits from the semaphore, can we just use the aligned bytes as the semaphore's max_permits? We only need to align the bytes before acquiring. Then the available_permits always indicate the available memory.
https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.available_permits

src/mito2/src/compaction/memory_manager.rs
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+
+    /// Waits until the specified amount of memory becomes available.
+    ///
+    /// Note: This method does not reserve the memory. After this method returns,
+    /// you should call `try_acquire` to attempt to reserve the memory. However,
+    /// there is no guarantee the memory will still be available at that point,
+    /// as another task may acquire it concurrently.
+    pub async fn wait_for_available(&self, bytes: u64) {
```
evenyag: Why not use https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.acquire

src/mito2/src/sst/file.rs
```diff
@@ -145,18 +142,10 @@ pub struct FileMeta {
     pub level: Level,
     /// Size of the file.
     pub file_size: u64,
+    /// Maximum uncompressed row group size of the file. 0 means unknown.
+    pub max_row_group_uncompressed_size: u64,
```
evenyag, line 146: non-blocking: Is it possible to get this during compaction time so we don't need to add one more field to the FileMeta.

evenyag, line 146: non-blocking: Is it possible to get this during compaction time so we don't need to add one more field to the FileMeta.

src/mito2/src/compaction.rs
```diff
@@ -429,31 +457,211 @@ impl CompactionScheduler {
...
+                    memory_bytes: requested_bytes.max(1),
+                });
+
+                if is_new_pending {
+                    self.spawn_memory_waiter(region_id, requested_bytes.max(1));
+                    info!(
+                        "Region {} compaction waits for {} bytes to become available",
+                        region_id, requested_bytes
+                    );
+                }
```
evenyag: I prefer not to implement the wait policy here because it makes the compaction scheduler too complicated. The retry behavior also conflicts with how we handle other compaction errors. The compaction task will be picked again next time we schedule compaction for a region, or when we manually compact the region.

We can also compact regions periodically in the future.

src/mito2/src/compaction.rs
```diff
@@ -429,31 +457,211 @@ impl CompactionScheduler {
...
+        region_id: RegionId,
+        requested_bytes: u64,
+    ) -> Result<()> {
+        match self.memory_manager.try_acquire(requested_bytes) {
+            Some(guard) => {
+                task.memory_guard = Some(guard);
+                self.submit_compaction_task(task, region_id)
+            }
+            None => self.handle_memory_exhausted(region_id, requested_bytes, task),
+        }
```
evenyag: I think we should acquire the permits in the compaction task instead of in the CompactionScheduler. Because the compaction scheduler can schedule many regions, but only max_bg_compactions tasks will execute concurrently. Acquiring permits inside the task is more accurate than inside the CompactionScheduler.

We can also wait inside the task (maybe with a timeout), which won't block the region worker and act as a rate limit mechanism.

src/mito2/src/compaction/memory_manager.rs
evenyag: May not if the semaphore is fair.
https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.acquire

evenyag (commented):
I have some different options for the memory management.

src/mito2/src/compaction/memory_manager.rs
fengjiachun: The purpose here is only to detect, but it will not consume the permits. This is because the design only consumes permits in one place, which is before executing `submit_compaction_task`.

src/mito2/src/compaction.rs
fengjiachun: I plan to assess peak memory usage before starting a task. If it exceeds the limit, I won't proceed. To achieve this, I've decided to obtain permits at the beginning of scheduling to avoid concerns about blocking or timeouts during execution.

@v0y4g3r is testing this now, so we might see the actual effect before deciding whether to update the design.

src/mito2/src/compaction/memory_manager.rs
fengjiachun: [39ce3a3](https://github.com/GreptimeTeam/greptimedb/pull/7305/commits/39ce3a3bb3ccb800764c7132fbd50e0d16bdb117)
We now take the minimum of Semaphore::MAX_PERMITS and u32::MAX. This limit, measured in 1 MB units, is sufficient for our needs. Conversely, using usize for Semaphore::MAX_PERMITS would make type conversion complicated.

src/mito2/src/compaction/memory_manager.rs
fengjiachun: [39ce3a3](https://github.com/GreptimeTeam/greptimedb/pull/7305/commits/39ce3a3bb3ccb800764c7132fbd50e0d16bdb117)

src/mito2/src/compaction.rs
fengjiachun: WDYT @v0y4g3r

fengjiachun:
Added on-demand additional memory allocation via `AdditionalMemoryGuard`, so running compaction tasks can request extra permits and release them early; the base guard still releases any remaining additional permits on drop as a safety net.  cc @v0y4g3r

src/mito2/src/compaction/memory_manager.rs
```diff
@@ -0,0 +1,711 @@
...
+enum GuardState {
+    Unlimited,
+    Limited {
+        inner: Arc<Inner>,
+        /// Base quota allocated at task startup.
+        base_permits: u32,
+        /// Additional memory allocated during task execution.
+        additional_permits: Arc<AtomicI64>,
+    },
+}
```
evenyag, line 251: Can we use OwnedSemaphorePermit to implement the guard? The CompactionMemoryGuard/AdditionalMemoryGuard can hold an OwnedSemaphorePermit.

You can also merge two permits so you may not have to define a new type for additional guard.
https://docs.rs/tokio/latest/tokio/sync/struct.OwnedSemaphorePermit.html#method.merge

fengjiachun: [d6301c6](https://github.com/GreptimeTeam/greptimedb/pull/7305/commits/d6301c667a56d7c53e7995ef6301979cc81900a1)

src/mito2/src/compaction/memory_manager.rs
```diff
@@ -0,0 +1,597 @@
...
+
+    /// Acquires memory, waiting if necessary until enough is available.
+    pub async fn acquire(&self, bytes: u64) -> CompactionMemoryGuard {
+        match &self.quota {
+            None => CompactionMemoryGuard::unlimited(),
+            Some(quota) => {
+                let permits = bytes_to_permits(bytes);
+                if permits == 0 {
+                    return CompactionMemoryGuard::unlimited();
+                }
```
evenyag: Do we need to ensure the permits < limit_permits? Maybe use min(limit_permits) to avoid allocating out of limit permits.

evenyag (approved):

src/mito2/src/compaction/memory_manager.rs
fengjiachun: Good catch! Permits should be rounded up to a minimum of 1; otherwise, the subsequent request_additional becomes unrestricted, which is a bug.

fengjiachun: fixed in [e40a15b](https://github.com/GreptimeTeam/greptimedb/pull/7305/commits/e40a15bdb731ae7838c34db216b66055d6a3e888)

PR #7315, opened 2025-11-29
feat!: improve mysql/pg compatibility

### LINKED ISSUE #7278: GreptimeDB MySQL Protocol SHOW DATABASES Compatibility Issue
### What type of bug is this?

Incorrect result

Other

### Minimal reproduce step

**Current Behavior (GreptimeDB):**
```sql
SHOW DATABASES;

+---------+
| Schemas |
+---------+
| public  |
+---------+
```

**Expected Behavior (MySQL Standard):**
```sql
SHOW DATABASES;

+----------+
| Database |
+----------+
| public   |
+----------+
```

SHOW DATABASES;

+---------+
| Schemas |  ← Following standard MYSQL protocol specification : column should be | Database |
+---------+
| public  |
+---------+

It's 
### MySQL Protocol Specification
- **URL:** https://dev.mysql.com/doc/refman/8.3/en/show-databases.html
- **Evidence:** MySQL Reference Manual specifies column name as `Database`
- **Citation:** "SHOW DATABASES lists the databases on the MySQL server host... The output displays only those databases for which you have some kind of privilege"
- **Standard Column Name:** `Database` (confirmed in MariaDB docs at https://mariadb.com/docs/server/reference/sql-statements/administrative-sql-statements/show/show-databases showing output with `| Database |`)

### 3. MySQL Information Schema Standard
- **URL:** https://dev.mysql.com/doc/refman/8.3/en/information-schema.html
- **INFORMATION_SCHEMA.SCHEMATA Table:** Contains `SCHEMA_NAME` column (not `Schemas`)
- **Evidence:** Standard metadata tables use singular column names per database entry

### Error Trace
JDBC connection fails:
\`\`\`
ERROR 1064 [42000]: ERROR: Failed to describe statement
\`\`\`

This is only one example, but column names in other metadata queries (such as `SHOW TABLES`) could also be affected if they do not follow the MySQL specification. As a result, JDBC connectors using the MySQL driver fail to retrieve metadata correctly.

helm greptimedb-cluster repo

1.0.0-beta.1

```bash

```

### LINKED ISSUE #7313: Postico requires 'obj_description' function for listing tables

User experience

Postico needs the `obj_description` function for listing tables:

```
PG_DIAG_SEVERITY: ERROR
PG_DIAG_SQLSTATE: P0000
PG_DIAG_MESSAGE_PRIMARY: Failed to plan SQL: Error during planning: Invalid function 'obj_description'.
Did you mean 'list_position'?

SELECT
	oid
	,nspname
	,nspname = ANY (current_schemas(true)) AS is_on_search_path
	
	    ,obj_description(oid, 'pg_namespace') AS comment
	
FROM pg_namespace; SELECT
oid
,nspname
FROM pg_namespace
WHERE oid = pg_my_temp_schema();
```

Image

This would unlock using postico to debug/inspect greptime tables visually - it's the best postgres db admin app.

### Implementation challenges

Could be related to #3560 but that's closed and requests were for new issues.

killme2008:

Close #7278 
Close #7313 

Most of the code consists of SQLness test results, so don't worry 😄 

This PR enhances MySQL and PostgreSQL protocol compatibility by implementing several key improvements for better integration with JDBC connectors and client tools like StarRocks, DBeaver, and standard MySQL/PostgreSQL clients.

**Key Changes:**
- Implemented MySQL `IF()` function with proper truthiness semantics (NULL and 0 are falsy, non-zero numerics are truthy)
- Added PostgreSQL compatibility functions (`obj_description`, `col_description`, `shobj_description`, `pg_my_temp_schema`) that return placeholder values. @sunng87 
- Changed `SHOW TABLES` column naming from static "Tables" to MySQL-compatible "Tables_in_{schema_name}" format @sunchanglong 
- Introduced session-level warning management system with `SHOW WARNINGS` support for MySQL protocol
- Modified `SET` variable handling to emit warnings instead of errors for unsupported variables in MySQL/PostgreSQL protocols
- Enhanced `INFORMATION_SCHEMA.PARTITIONS` table with `partition_description` field for StarRocks compatibility

Remaining tasks:
* Implement `obj_description`, `col_description`, `shobj_description`, and `pg_my_temp_schema` that return placeholder values right now. Since OIDs are managed by datafusion-postgres, this is not easy to support directly in GreptimeDB. @sunng87  
* The `SHOW TABLES` result column has changed from `Tables` to `ables_in_{schema_name}`. Please verify that the dashboard still works after this breaking change. @sunchanglong It is preferable to query `information_schema.tables` instead of using `SHOW TABLES`.

## PR Checklist

### DIFF
--- a/src/common/function/src/scalars/expression/if_func.rs
+++ b/src/common/function/src/scalars/expression/if_func.rs
@@ -16,10 +16,11 @@ use std::fmt;
 use std::fmt::Display;
 
 use arrow::array::ArrowNativeTypeOp;
-use datafusion::arrow::array::{Array, ArrayRef, AsArray, BooleanArray};
+use arrow::datatypes::ArrowPrimitiveType;
+use datafusion::arrow::array::{Array, ArrayRef, AsArray, BooleanArray, PrimitiveArray};
 use datafusion::arrow::compute::kernels::zip::zip;
 use datafusion::arrow::datatypes::DataType;
-use datafusion_common::{DataFusionError, ScalarValue};
+use datafusion_common::DataFusionError;
 use datafusion_expr::type_coercion::binary::comparison_coercion;
 use datafusion_expr::{ColumnarValue, ScalarFunctionArgs, Signature, Volatility};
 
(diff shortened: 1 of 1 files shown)


src/session/src/lib.rs
```diff
@@ -156,4 +162,27 @@ impl Session {
...
+        let mut inner = self.mutable_inner.write().unwrap();
+        if inner.warnings.len() >= MAX_WARNINGS {
+            inner.warnings.pop_front();
+        }
+        inner.warnings.push_back(warning);
+    }
+
+    pub fn clear_warnings(&self) {
+        self.mutable_inner.write().unwrap().warnings.clear();
+    }
```
sunng87, line 195: We already have a warning system in QueryContext for postgres query scoped warnings. We need to merge these two mechanism together.

For postgres, we dump warnings to client automatically.

sunng87:
> Implement obj_description, col_description, shobj_description, and pg_my_temp_schema that return placeholder values right now. Since OIDs are managed by datafusion-postgres, this is not easy to support directly in GreptimeDB. @sunng87

should be good. We don't have OID system neither in greptime nor in datafusion-postgres for now. Would be nice to submit them to upstream datafusion-postgres repo too.

src/session/src/lib.rs
killme2008: Yes, MySQL uses `show warnings` to display warnings.

Let me find a way to merge them.

killme2008: Sorry, I didn’t think it through yesterday. Actually, there’s no problem here. 
* One is per‑request (stored in the QueryContext, like PostgreSQL does), and the other is per‑session (the MySQL approach).
* The per‑request one is placed into the session and destroyed after it’s used.
* The session‑level one will be cleared during the next query, and meanwhile it can be displayed via `SHOW WARNINGS`.

sunng87: > The session‑level one will be cleared during the next query, and meanwhile it can be displayed via SHOW WARNINGS.

if session-level warnings are cleared during next query, it's considered query scoped too. We'd better store it QueryContext, and copy to `Session` after query got processed in case that next query is `SHOW WARNINGS` to dump it.

killme2008: The current design adds the query context’s warnings in the session, which can then be retrieved with `SHOW WARNINGS`.

This approach works well and remains compatible with potential future support for executing multiple statements in a single call.

src/common/function/src/scalars/expression/if_func.rs
```diff
@@ -0,0 +1,441 @@
...
+            for i in 0..len {
+                result.push(!array.is_null(i) && typed_array.value(i) != 0);
+            }
+        }
+        DataType::UInt64 => {
+            let typed_array = array.as_primitive::<arrow::datatypes::UInt64Type>();
+            for i in 0..len {
+                result.push(!array.is_null(i) && typed_array.value(i) != 0);
+            }
+        }
```
fengjiachun: The code is repetitive; can we simplify it using macros?

src/common/function/src/scalars/expression/if_func.rs
```diff
@@ -0,0 +1,441 @@
...
+        DataType::Float64 => {
+            let typed_array = array.as_primitive::<arrow::datatypes::Float64Type>();
+            for i in 0..len {
+                result.push(
+                    !array.is_null(i)
+                        && (typed_array.value(i).is_nan() || !typed_array.value(i).is_zero()),
+                );
+            }
```
fengjiachun: Same as above

fengjiachun (approved):
Others rest LGTM

src/common/function/src/scalars/expression/if_func.rs
killme2008: It's in the method, not easy to refactor. And we currently discourage macro usage.

killme2008: Addressed in https://github.com/GreptimeTeam/greptimedb/pull/7315/commits/c35a5177f7f3f5daba30c0d870be56bc08c5fff7

src/common/function/src/scalars/expression/if_func.rs
killme2008: Addressed in https://github.com/GreptimeTeam/greptimedb/pull/7315/commits/c35a5177f7f3f5daba30c0d870be56bc08c5fff7

switz:
wow this is sick! thank you!

PR #7326, opened 2025-12-02
feat: implement a cache for manifest files

evenyag:

This is part of https://github.com/GreptimeTeam/greptimedb/issues/7259

Implement a cache `ManifestCache` for manifest files
- The cache acts as a write cache and stores manifest files under `cache/object/manifest/`
- Since manifest files are important, it doesn't hash the file path but stores the file in the same path under the cache directory
- Since we need to clean the directory, the manifest cache implements methods to remove empty directories.
  - After the manifest cache recovers itself, it will clean empty directories
  - If we drop a region, the cache will clean the empty table directory to avoid leaking the table directory
  - Metric engine and mito engine have different manifest paths, so we use a generic method to clean them.

The `ManifestCache` is disabled and we will enable it after we remove the read cache.

This PR also refactors the FileCache to make it easier to implement a read cache replacement.

## PR Checklist

### DIFF
--- a/src/mito2/src/cache/manifest_cache.rs
+++ b/src/mito2/src/cache/manifest_cache.rs
@@ -388,6 +388,10 @@ impl ManifestCache {
             }
             Err(e) => return Err(e),
         };
+
+        let mut is_empty = true;
+        // Iterates all entries under the directory.
+        // We have to check all entries to clean up all empty subdirectories.
         for entry in entries {
             let entry = entry?;
             let path = entry.path();
@@ -401,15 +405,18 @@ impl ManifestCache {
                     && elapsed < Duration::from_secs(3600)
                 {
                     common_telemetry::debug!("Skip directory by mtime, elapsed: {:?}", elapsed);
-                    // Only removes if not modified for at least 1 hour
-                    return Ok(false);
+                    // Only removes if not modified for at least 1 hour.
+                    is_empty = false;
+                    continue;
                 }
 
                 let subdir_empty = Self::remove_empty_dirs_recursive_sync(&path, check_mtime)?;
                 if subdir_empty {
-                    if let Err(e) = std::fs::remove_dir(&path) {
+                    if let Err(e) = std::fs::remove_dir(&path)
+                        && e.kind() != std::io::ErrorKind::NotFound
+                    {
                         warn!(e; "Failed to remove empty directory {}", path.display());
-                        return Ok(false);
+                        is_empty = false;
                     } else {
                         info!(
                             "Removed empty directory {} from manifest cache",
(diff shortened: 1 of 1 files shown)


src/mito2/src/manifest/storage.rs
```diff
@@ -719,6 +763,63 @@ impl ManifestObjectStore {
...
+            .await
+            .context(OpenDalSnafu)?;
+
+        // Put to cache if we cloned the data
+        if let Some(data) = cache_data {
+            self.put_to_cache(path.to_string(), &data, is_staging).await;
+        }
+
+        Ok(())
+    }
```
fengjiachun, line 813: It is better to clone the data only once and use the `put_to_cache` method to consume it.

fengjiachun:
```
if !is_staging && self.manifest_cache.is_some() {
    self.object_store.write(path, data.clone()).await.context(OpenDalSnafu)?;
    self.put_to_cache(path.to_string(), data, is_staging).await;
} else {
    self.object_store.write(path, data).await.context(OpenDalSnafu)?;
}
```

src/mito2/src/cache/manifest_cache.rs
```diff
@@ -0,0 +1,567 @@
...
+            if metadata.is_dir() {
+                // Checks if we should skip this directory based on modification time
+                if check_mtime
+                    && let Ok(modified) = metadata.modified()
+                    && let Ok(elapsed) = modified.elapsed()
+                    && elapsed < Duration::from_secs(3600)
+                {
+                    common_telemetry::debug!("Skip directory by mtime, elapsed: {:?}", elapsed);
+                    // Only removes if not modified for at least 1 hour
+                    return Ok(false);
```
fengjiachun: Will the sibling directories behind not be checked anymore?

src/mito2/src/cache/manifest_cache.rs
```diff
@@ -0,0 +1,567 @@
...
+                }
+
+                let subdir_empty = Self::remove_empty_dirs_recursive_sync(&path, check_mtime)?;
+                if subdir_empty {
+                    if let Err(e) = std::fs::remove_dir(&path) {
```
fengjiachun: Can `NotFound` be seen as a success?

src/mito2/src/manifest/storage.rs
evenyag: Some places require put_to_cache receiving the slice to reduce the clone.

src/mito2/src/manifest/storage.rs
```diff
@@ -501,69 +522,91 @@ impl ManifestObjectStore {
...
+                        );
+                        debug!(
+                            "Failed to load checkpoint from path: {}, fall back to path: {}",
+                            path, fall_back_path
+                        );
+
+                        // Try to get fallback from cache first
+                        if let Some(data) = self.get_from_cache(&fall_back_path, false).await {
+                            verify_checksum(&data, metadata.checksum)?;
+                            return Ok(Some((version, data)));
```
WenyXu, line 568: Do we need to log this?

evenyag: Maybe unnecessary, we didn't log this before.

PR #7329, opened 2025-12-02
refactor(servers): bulk insert service

v0y4g3r:

This PR refactors the bulk insert service to improve code organization, performance monitoring, and error handling. The changes move FlightData decoding earlier in the pipeline, implement stream-based processing with per-batch responses.

## PR Checklist

### DIFF
--- a/src/frontend/src/instance/grpc.rs
+++ b/src/frontend/src/instance/grpc.rs
@@ -319,11 +319,9 @@ impl GrpcQueryHandler for Instance {
                     None,
                 )
                 .await
-                .context(CatalogSnafu)
-                .and_then(|opt| {
-                    opt.with_context(|| TableNotFoundSnafu {
-                        table_name: table_name.to_string(),
-                    })
+                .context(CatalogSnafu)?
+                .with_context(|| TableNotFoundSnafu {
+                    table_name: table_name.to_string(),
                 })?;
 
             // Check permissions once for the stream
@@ -333,7 +331,7 @@ impl GrpcQueryHandler for Instance {
 
             // Process each request in the stream
             while let Some(request_result) = stream.next().await {
-                let request = request_result.map_err(|e|{
+                let request = request_result.map_err(|e| {
                     let error_msg = format!("Stream error: {:?}", e);
                     IncompleteGrpcRequestSnafu { err_msg: error_msg }.build()
                 })?;
--- a/src/operator/src/bulk_insert.rs
+++ b/src/operator/src/bulk_insert.rs
@@ -54,7 +54,7 @@ impl Inserter {
         }
 
         let body_size = raw_flight_data.data_body.len();
-        // TODO(yingwen): Fill record batch impure default values.
+        // TODO(yingwen): Fill record batch impure default values. Note that we should override `raw_flight_data` if we have to fill defaults.
         // notify flownode to update dirty timestamps if flow is configured.
         self.maybe_update_flow_dirty_window(table_info.clone(), record_batch.clone());
 
@@ -145,20 +145,6 @@ impl Inserter {
 
         let mut handles = Vec::with_capacity(mask_per_datanode.len());
 
-        // Encode the record batch once for reuse when mask.select_all()
-        let encode_timer = metrics::HANDLE_BULK_INSERT_ELAPSED
-            .with_label_values(&["encode"])
-            .start_timer();
-        let mut encoder = FlightEncoder::default();
-        let encoded = encoder.encode(FlightMessage::RecordBatch(record_batch.clone()));
-        let FlightData {
-            data_header,
-            data_body,
-            ..
-        } = encoded.into_iter().next().unwrap();
-        let raw_data_bytes = (data_header, data_body);
-        encode_timer.observe_duration();
-
         for (peer, masks) in mask_per_datanode {
             for (region_id, mask) in masks {
                 if mask.select_none() {
@@ -169,7 +155,10 @@ impl Inserter {
                 let node_manager = self.node_manager.clone();
                 let peer = peer.clone();
                 let raw_header_and_data = if mask.select_all() {
-                    Some(raw_data_bytes.clone())
+                    Some((
+                        raw_flight_data.data_header.clone(),
+                        raw_flight_data.data_body.clone(),
+                    ))
                 } else {
                     None
                 };


src/frontend/src/instance/grpc.rs
```diff
@@ -279,10 +281,102 @@ impl GrpcQueryHandler for Instance {
...
+
+            // Check permissions once for the stream
+            let interceptor_ref = plugins.get::<GrpcQueryInterceptorRef<Error>>();
+            let interceptor = interceptor_ref.as_ref();
+            if let Err(e) = interceptor.pre_bulk_insert(table_ref.clone(), ctx.clone()) {
+                yield Err(e);
+                return;
+            }
+
+            if let Err(e) = plugins
```
MichaelScofield: I think permission check should be the first thing in the stream, even before finding the table.

v0y4g3r: > I think permission check should be the first thing in the stream, even before finding the table.

7b88f2f132

src/frontend/src/instance/grpc.rs
```diff
@@ -279,10 +281,101 @@ impl GrpcQueryHandler for Instance {
...
+    ) -> Pin<Box<dyn Stream<Item = Result<DoPutResponse>> + Send>> {
+        // Resolve table once for the stream
+        // Clone all necessary data to make it 'static
+        let catalog_manager = self.catalog_manager().clone();
+        let plugins = self.plugins.clone();
+        let inserter = self.inserter.clone();
+        let table_name = stream.table_name().clone();
+        let ctx = ctx.clone();
+
+        Box::pin(stream! {
```
evenyag: non-blocking: Maybe you can use try_stream to simplify error handling.

src/frontend/src/instance/grpc.rs
```diff
@@ -279,10 +281,101 @@ impl GrpcQueryHandler for Instance {
...
+                return;
+            }
+
+            // Process each request in the stream
+            while let Some(request_result) = stream.next().await {
+                let request = match request_result {
+                    Ok(request) => request,
+                    Err(e) => {
+                        // Convert tonic::Status to Error
+                        let error_msg = format!("Stream error: {}", e);
```
evenyag: What's the type of the Error? If the error is defined by us, use `output_msg()` instead of Display to get the cause.

src/servers/src/grpc/greptime_handler.rs
```diff
@@ -144,37 +142,24 @@ impl GreptimeRequestHandler {
...
-                timer.observe_duration();
-                let result = result
-                    .map(|x| DoPutResponse::new(request_id, x))
-                    .map_err(Into::into);
-                if let Err(e)= result_sender.try_send(result)
-                    && let TrySendError::Closed(_) = e {
-                    warn!(r#""DoPut" client with request_id {} maybe unreachable, abort handling its message"#, request_id);
+                if let Err(e) =
+                    result_sender.try_send(result.map_err(|e| Status::from_error(Box::new(e))))
+                    && let TrySendError::Closed(_) = e
```
evenyag, line 160: If the channel is full, we won't send the response again. Is that expected?

src/operator/src/bulk_insert.rs
```diff
@@ -158,8 +157,20 @@ impl Inserter {
...
+            .start_timer();
+        let mut encoder = FlightEncoder::default();
+        let encoded = encoder.encode(FlightMessage::RecordBatch(record_batch.clone()));
+        let FlightData {
+            data_header,
+            data_body,
+            ..
+        } = encoded.into_iter().next().unwrap();
+        let raw_data_bytes = (data_header, data_body);
+        encode_timer.observe_duration();
```
evenyag: Looks like both the fast path and this path encode the record batch. Can we do it before checking `region_masks.len() == 1`?

src/servers/src/grpc/greptime_handler.rs
v0y4g3r: > If the channel is full, we won't send the response again. Is that expected?

[Yes](https://github.com/GreptimeTeam/greptimedb/pull/6488#discussion_r2194571366)

src/frontend/src/instance/grpc.rs
v0y4g3r: fixed in https://github.com/GreptimeTeam/greptimedb/pull/7329/commits/41eb0d0f25c20223f80bf95da11a3b8f5fddf450#diff-ed50ea67c31eb65e7b5b87feb777fbd80bde5f54229139a709983b6646c28fd6R307

src/frontend/src/instance/grpc.rs
v0y4g3r: TonicStatus

src/operator/src/bulk_insert.rs
v0y4g3r: We can avoid re-encode if we pass the raw flight data bytes down to here: https://github.com/GreptimeTeam/greptimedb/commit/41eb0d0f25c20223f80bf95da11a3b8f5fddf450#diff-5b4053c1f815b8bbe894a29d013e96c8337dd712be73fe8194e22cb1f9a6e9cbR105-R106

src/frontend/src/instance/grpc.rs
```diff
@@ -279,10 +281,79 @@ impl GrpcQueryHandler for Instance {
...
+                    &table_name.table_name,
+                    None,
+                )
+                .await
+                .context(CatalogSnafu)
+                .and_then(|opt| {
+                    opt.with_context(|| TableNotFoundSnafu {
+                        table_name: table_name.to_string(),
+                    })
+                })?;
```
evenyag:
```suggestion
                .context(CatalogSnafu)?
                .with_context(|| TableNotFoundSnafu {
                    table_name: table_name.to_string(),
                })?;
```

src/frontend/src/instance/grpc.rs
```diff
@@ -279,10 +281,79 @@ impl GrpcQueryHandler for Instance {
...
+
+            // Check permissions once for the stream
+            let interceptor_ref = plugins.get::<GrpcQueryInterceptorRef<Error>>();
+            let interceptor = interceptor_ref.as_ref();
+            interceptor.pre_bulk_insert(table_ref.clone(), ctx.clone())?;
+
+            // Process each request in the stream
+            while let Some(request_result) = stream.next().await {
+                let request = request_result.map_err(|e|{
```
evenyag:
```suggestion
                let request = request_result.map_err(|e| {
```

src/operator/src/bulk_insert.rs
```diff
@@ -40,32 +41,19 @@ impl Inserter {
...
-            return Ok(0);
-        };
-        decode_timer.observe_duration();
 
         if record_batch.num_rows() == 0 {
             return Ok(0);
         }
 
+        let body_size = raw_flight_data.data_body.len();
         // TODO(yingwen): Fill record batch impure default values.
```
evenyag: Maybe we have to leave a comment that if we fill default values here, then we can't reuse the `raw_flight_data`.
```suggestion
        // TODO(yingwen): Fill record batch impure default values. Note that we should override `raw_flight_data` if we have to fill defaults.
```

PR #7346, opened 2025-12-04
test: more gc worker integration test

discord9:

more gc worker integration test, including testing long listing time&long query

## PR Checklist

### DIFF
--- a/tests-integration/src/test_util.rs
+++ b/tests-integration/src/test_util.rs
@@ -149,8 +149,6 @@ fn s3_test_config() -> S3Config {
             bucket: env::var("GT_S3_BUCKET").unwrap(),
             region: Some(env::var("GT_S3_REGION").unwrap()),
             endpoint: env::var("GT_S3_ENDPOINT_URL").ok(),
-            allow_anonymous: true, // Enable anonymous access to bypass reqsign for local testing
-            disable_config_load: true, // Disable config load to bypass reqsign for local MinIO
             ..Default::default()
         },
         ..Default::default()
--- a/tests-integration/src/tests/gc.rs
+++ b/tests-integration/src/tests/gc.rs
@@ -21,9 +21,7 @@ use meta_srv::gc::Region2Peers;
 use store_api::storage::RegionId;
 use table::metadata::TableId;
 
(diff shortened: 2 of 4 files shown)


src/cli/src/common/object_store.rs
```diff
@@ -89,6 +89,10 @@ wrap_with_clap_prefix! {
         region: Option<String>,
         #[doc = "Enable virtual host style for the object store."]
         enable_virtual_host_style: bool = Default::default(),
+        #[doc = "Allow anonymous access (disable credential signing) for testing."]
+        allow_anonymous: bool = Default::default(),
+        #[doc = "Disable config load from environment and files for testing."]
+        disable_config_load: bool = Default::default(),
```
fengjiachun: It’s better to use the same style as `enable_virtual_host_style`.
```suggestion
        enable_config_load: bool = true,
```

src/object-store/src/config.rs
```diff
@@ -117,6 +117,10 @@ pub struct S3Connection {
     /// By default, opendal will send API to https://s3.us-east-1.amazonaws.com/bucket_name
     /// Enabled, opendal will send API to https://bucket_name.s3.us-east-1.amazonaws.com
     pub enable_virtual_host_style: bool,
+    /// Allow anonymous access (disable credential signing) - useful for local testing
+    pub allow_anonymous: bool,
+    /// Disable config load from environment and files - useful for local testing
+    pub disable_config_load: bool,
```
fengjiachun: I prefer the enable_xxx naming style because it appears to be used across all workspaces.

fengjiachun (approved):
Rest LGTM

src/cli/src/common/object_store.rs
discord9: > It’s better to use the same style as `enable_virtual_host_style`.

both is directly copy from S3Config 
https://docs.rs/opendal/latest/opendal/services/struct.S3Config.html#structfield.enable_virtual_host_style 
and 
https://docs.rs/opendal/latest/opendal/services/struct.S3Config.html#structfield.disable_config_load so I guess their default value is false&reasonable?

PR #7357, opened 2025-12-06
refactor: extract file watcher to common-config

shuiyisong:

As the title suggests, extract the function for later reuse.

## PR Checklist

### DIFF
--- a/src/auth/src/user_provider/watch_file_user_provider.rs
+++ b/src/auth/src/user_provider/watch_file_user_provider.rs
@@ -46,7 +46,7 @@ impl WatchFileUserProvider {
         FileWatcherBuilder::new()
             .watch_path(filepath)
             .context(FileWatchSnafu)?
-            .config(FileWatcherConfig::modify_and_create())
+            .config(FileWatcherConfig::new())
             .spawn(move || match load_credential_from_file(&filepath_owned) {
                 Ok(credential) => {
                     let mut users = users_clone.lock().expect("users credential must be valid");
--- a/src/common/config/src/file_watcher.rs
+++ b/src/common/config/src/file_watcher.rs
@@ -40,18 +40,18 @@ pub struct FileWatcherConfig {
 }
 
 impl FileWatcherConfig {
-    /// Create a config that watches for modify and create events only.
-    pub fn modify_and_create() -> Self {
-        Self {
-            include_remove_events: false,
-        }
+    pub fn new() -> Self {
+        Self::default()
     }
 
-    /// Create a config that also watches for remove events.
-    pub fn with_remove_events() -> Self {
-        Self {
-            include_remove_events: true,
-        }
+    pub fn with_modify_and_create(mut self) -> Self {
+        self.include_remove_events = false;
+        self
+    }
+
+    pub fn with_remove_events(mut self) -> Self {
+        self.include_remove_events = true;
+        self
     }
 }
 
@@ -240,7 +240,7 @@ mod tests {
         FileWatcherBuilder::new()
             .watch_path(&file_path)
             .unwrap()
-            .config(FileWatcherConfig::modify_and_create())
+            .config(FileWatcherConfig::new())
             .spawn(move || {
                 counter_clone.fetch_add(1, Ordering::SeqCst);
             })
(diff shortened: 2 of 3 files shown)


src/auth/src/user_provider/watch_file_user_provider.rs
```diff
@@ -41,61 +39,35 @@ impl WatchFileUserProvider {
...
-        let mut debouncer =
-            notify::recommended_watcher(tx).context(FileWatchSnafu { path: "<none>" })?;
-        let mut dir = Path::new(filepath).to_path_buf();
-        ensure!(
-            dir.pop(),
-            InvalidConfigSnafu {
-                value: filepath,
-                msg: "UserProvider path must be a file path",
-            }
-        );
```
waynexia, line 58: looks like this `dir.pop()` disappears in the new implementation (`watch_path()`), is this intended?

shuiyisong: Looks like the `dir.pop()` is to check whether the path point to a file. Adding the check in `watch_path` now https://github.com/GreptimeTeam/greptimedb/pull/7357/commits/e3cac52deff759e5f41840f78d43d2be951444f7

waynexia: it also modifies `dir`, which is a mutable reference

src/common/config/src/file_watcher.rs
```diff
@@ -0,0 +1,214 @@
...
+            include_remove_events: false,
+        }
+    }
+
+    /// Create a config that also watches for remove events.
+    pub fn with_remove_events() -> Self {
+        Self {
+            include_remove_events: true,
+        }
+    }
```
sunng87, line 55: better as

```rust
    /// Create a config that also watches for remove events.
    pub fn with_remove_events(mut self) -> Self {
        self.include_remove_events = true;
        self
    }
```

src/common/config/src/file_watcher.rs
```diff
@@ -0,0 +1,214 @@
...
+        let config = self.config;
+        let paths_display: Vec<_> = self.paths.iter().map(|p| p.display().to_string()).collect();
+
+        info!("Spawning file watcher for paths: {:?}", paths_display);
+
+        std::thread::spawn(move || {
+            // Keep watcher alive in the thread
+            let _watcher = watcher;
+
+            while let Ok(res) = rx.recv() {
```
sunng87, line 161: Check again if rx.recv is recoverable? do we want to quit the loop when there is an error?

src/common/config/src/file_watcher.rs
```diff
@@ -0,0 +1,214 @@
...
+                .watch(path, RecursiveMode::NonRecursive)
+                .with_context(|_| FileWatchSnafu {
+                    path: path.display().to_string(),
+                })?;
+        }
+
+        let config = self.config;
+        let paths_display: Vec<_> = self.paths.iter().map(|p| p.display().to_string()).collect();
+
+        info!("Spawning file watcher for paths: {:?}", paths_display);
```
sunng87: put paths_display inside macro to avoid unneeded compute

src/auth/src/user_provider/watch_file_user_provider.rs
sunng87: it's design to watch the directory of config file, instead of config file itself

shuiyisong: Should we still watch the parent directory in this refactor?

sunng87: I remember one of the reasons is if we watch the file, when the file get deleted and recreated, we will never get notification for new file.

src/common/config/src/file_watcher.rs
shuiyisong: `rx.recv` will return an `Err` if the sender is disconnected. I don't think it's recoverable 🤔

src/auth/src/user_provider/watch_file_user_provider.rs
shuiyisong: In https://github.com/GreptimeTeam/greptimedb/pull/7357/commits/81b83e88e1d28f8a114172a8cae18ccf8e67b02f, change to watch folder instead of the file. PTAL

PR #7353, opened 2025-12-05
docs: rfc for vector index

killme2008:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

Vector index based on usearch.

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #7366, opened 2025-12-09
feat(vector_index): adds the foundational types and SQL parsing support for vector index

killme2008:
This PR adds the foundational types and SQL parsing support for vector index:

- `VectorIndexEngineTyp`e enum (Usearch HNSW)
- `VectorDistanceMetric` enum (L2sq, Cosine, InnerProduct)
- `VectorIndexOptions` with HNSW parameters (m, ef_construct, ef_search)
- `VectorSearchRequest` for KNN search hints
- SQL parsing for `VECTOR INDEX` column options
- Distance computation utilities in index crate

The syntax:

```sql
-- Create a table with vector column and L2sq vector index
CREATE TABLE vectors_l2sq (
    vec_id INT,
    ts TIMESTAMP TIME INDEX,
    embedding VECTOR(4) VECTOR INDEX WITH (metric = 'l2sq', engine='usearch', connectivity='256', ...),
    PRIMARY KEY (vec_id)
);
```

This is part 1 of 4 PRs to implement vector index support. #7353 

## PR Checklist

### DIFF
--- a/src/index/Cargo.toml
+++ b/src/index/Cargo.toml
@@ -41,7 +41,7 @@ tantivy = { version = "0.24", features = ["zstd-compression"] }
 tantivy-jieba = "0.16"
 tokio.workspace = true
 tokio-util.workspace = true
-usearch = "2.21"
+usearch = { version = "2.21", default-features = false, features = ["fp16lib"] }
 uuid.workspace = true
 
 [dev-dependencies]


src/index/src/vector.rs
```diff
@@ -0,0 +1,137 @@
...
+/// Calculates the cosine distance between two vectors.
+fn cosine(lhs: &[f32], rhs: &[f32]) -> f32 {
+    let lhs_vec = DVectorView::from_slice(lhs, lhs.len());
+    let rhs_vec = DVectorView::from_slice(rhs, rhs.len());
+
+    let dot_product = lhs_vec.dot(&rhs_vec);
+    let lhs_norm = lhs_vec.norm();
+    let rhs_norm = rhs_vec.norm();
+
+    if dot_product.abs() < f32::EPSILON
```
v0y4g3r, line 73: Can we simply reuse https://github.com/GreptimeTeam/greptimedb/blob/c9377e7c5a59eb37fdfc1ef38cd48f611434c262/src/common/function/src/scalars/vector/distance/cos.rs#L20 ?

src/sql/src/statements/create.rs
```diff
@@ -233,6 +247,93 @@ impl ColumnExtensions {
...
+            })
+            .transpose()?
+            .unwrap_or(64);
+
+        Ok(Some(VectorIndexOptions {
+            engine,
+            metric,
+            connectivity,
+            expansion_add,
+            expansion_search,
```
v0y4g3r: Do we need to check if these parameters are not zero and within a reasonable bound?

src/index/src/vector.rs
```diff
@@ -0,0 +1,137 @@
...
+/// Computes distance between two vectors using the specified metric.
+///
+/// Uses SIMD-optimized implementations via nalgebra.
+///
+/// **Note:** The caller must ensure that the two vectors have the same length.
+pub fn compute_distance(v1: &[f32], v2: &[f32], metric: VectorDistanceMetric) -> f32 {
+    match metric {
+        VectorDistanceMetric::L2sq => l2sq(v1, v2),
+        VectorDistanceMetric::Cosine => cosine(v1, v2),
+        VectorDistanceMetric::InnerProduct => -dot(v1, v2),
```
v0y4g3r, line 48: nit: maybe doc the negation here

src/index/src/vector.rs
```diff
@@ -0,0 +1,137 @@
...
+        VectorDistanceMetric::InnerProduct => MetricKind::IP,
+    }
+}
+
+/// Computes distance between two vectors using the specified metric.
+///
+/// Uses SIMD-optimized implementations via nalgebra.
+///
+/// **Note:** The caller must ensure that the two vectors have the same length.
+pub fn compute_distance(v1: &[f32], v2: &[f32], metric: VectorDistanceMetric) -> f32 {
```
v0y4g3r, line 39: Is it expected to have different behavior when input vectors are empty?

For example:
```
l2sq(&[], &[]) == 0.0
cos(&[], &[]) == 1.0
-dot(&[], &[]) == -0.0
```

src/index/src/vector.rs
killme2008: Yes, we'll refactor the function in next PR.

src/sql/src/statements/create.rs
killme2008: Good point, I think so.

src/index/src/vector.rs
killme2008: This function requires vectors of equal length and not empty, but we can check upfront.

killme2008: Empty vectors return 0.0 for all metrics.

src/sql/src/statements/create.rs
killme2008: Fixed via https://github.com/GreptimeTeam/greptimedb/pull/7366/commits/631dd6e5600b6dffa9d7ed256c96313d7bb0deef

killme2008:
@lyang24 Please review this PR if you’re interested. It’s the first of four parts of #7353.

src/store-api/src/storage/requests.rs
```diff
@@ -14,11 +14,66 @@
...
+    pub keys: Vec<u64>,
+    /// Distances from the query vector.
+    pub distances: Vec<f32>,
+}
+
+/// Trait for vector index engines (HNSW implementations).
+///
+/// This trait defines the interface for pluggable vector index engines.
+/// Implementations (e.g., UsearchEngine) are provided by storage engines like mito2.
+pub trait VectorIndexEngine: Send + Sync {
```
lyang24, line 52: people often update the vectors because they ran through new models. Is the index append only or will we support some kind of upsert operations here?

would upsert be completed by adding the same key again?

src/sql/src/statements/create.rs
```diff
@@ -233,6 +247,89 @@ impl ColumnExtensions {
...
+        if let Some(s) = options_map.get("connectivity") {
+            let value = s.parse::<u32>().map_err(|_| {
+                InvalidSqlSnafu {
+                    msg: format!(
+                        "invalid VECTOR INDEX connectivity: {s}, expected positive integer"
+                    ),
+                }
+                .build()
+            })?;
+            if value == 0 {
```
lyang24: we should match validation to downstream lib must be > 2
https://github.com/unum-cloud/USearch/blob/a2f17599101729d667dc0260dd278852d9098183/include/usearch/index.hpp#L1383

maybe we can also add an reasonable upper bound as a safeguard to prevent mistakes
https://milvus.io/docs/hnsw.md#Index-building-params (milvus has 2048)

src/store-api/src/storage/requests.rs
lyang24, line 52: some thoughts here, maybe in v2 we can include batched ops like batch add or batch search (trade latency for higher throughput).

killme2008: Yes, the upsert behaves as it currently does by adding data with the same primary key and time index.

src/sql/src/statements/create.rs
killme2008: Good point! Let me add the vaidation.

src/store-api/src/storage/requests.rs
killme2008: Sure, it's the first prototype.

src/sql/src/statements/create.rs
killme2008: Addressed in https://github.com/GreptimeTeam/greptimedb/pull/7366/changes/c32d6c5e5bb0af534fd5444fc12caf732ed001e1

PR #7370, opened 2025-12-09
feat: suspend frontend and datanode

MichaelScofield:

Sometimes the frontend and datanode need to suspend handling requests. As to "suspend", only services that facing clients are made to return special "suspend" error, heartbeat (like region lease) are not affected.

## PR Checklist

### DIFF
--- a/src/common/error/src/status_code.rs
+++ b/src/common/error/src/status_code.rs
@@ -42,6 +42,8 @@ pub enum StatusCode {
     External = 1007,
     /// The request is deadline exceeded (typically server-side).
     DeadlineExceeded = 1008,
+    /// Service got suspended for various reason. For example, resources exceed limit.
+    Suspended = 1009,
     // ====== End of common status code ================
 
     // ====== Begin of SQL related status code =========
@@ -175,7 +177,8 @@ impl StatusCode {
             | StatusCode::AccessDenied
             | StatusCode::PermissionDenied
             | StatusCode::RequestOutdated
-            | StatusCode::External => false,
+            | StatusCode::External
+            | StatusCode::Suspended => false,
         }
     }
 
@@ -223,7 +226,8 @@ impl StatusCode {
             | StatusCode::InvalidAuthHeader
             | StatusCode::AccessDenied
             | StatusCode::PermissionDenied
-            | StatusCode::RequestOutdated => false,
+            | StatusCode::RequestOutdated
+            | StatusCode::Suspended => false,
         }
     }
 
@@ -347,7 +351,8 @@ pub fn status_to_tonic_code(status_code: StatusCode) -> Code {
         | StatusCode::RegionNotReady => Code::Unavailable,
         StatusCode::RuntimeResourcesExhausted
         | StatusCode::RateLimited
-        | StatusCode::RegionBusy => Code::ResourceExhausted,
+        | StatusCode::RegionBusy
+        | StatusCode::Suspended => Code::ResourceExhausted,
         StatusCode::UnsupportedPasswordType
         | StatusCode::UserPasswordMismatch
         | StatusCode::AuthHeaderNotFound
--- a/src/common/meta/src/heartbeat/handler/suspend.rs
+++ b/src/common/meta/src/heartbeat/handler/suspend.rs
@@ -1,0 +1,69 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+use async_trait::async_trait;
+use common_telemetry::{info, warn};
+
+use crate::error::Result;
+use crate::heartbeat::handler::{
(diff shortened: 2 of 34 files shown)


src/common/error/src/status_code.rs
```diff
@@ -42,6 +42,8 @@ pub enum StatusCode {
     External = 1007,
     /// The request is deadline exceeded (typically server-side).
     DeadlineExceeded = 1008,
+    /// Service got suspended for various reason. For example, resources exceed limit.
+    Suspend = 1009,
```
sunng87: better as `Suspended` just the comment and `DeadlineExceeded` above

src/common/meta/src/heartbeat/handler/suspend.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    async fn handle(&self, context: &mut HeartbeatResponseHandlerContext) -> Result<HandleControl> {
+        if let Some((_, Instruction::Suspend)) = context.incoming_message.take() {
+            self.suspend.store(true, atomic::Ordering::Relaxed);
+            warn!("set suspend state");
+        } else {
+            // Suspended components are made always tried to get rid of this state, we don't want
+            // an "un-suspend" instruction to resume them running. That can be error-prone.
+            // So if the "suspend" instruction is not found in the heartbeat, just unset the state.
+            if self.suspend.load(atomic::Ordering::Relaxed) {
```
sunng87: should we use compare_and_set for this?

src/datanode/src/region_server.rs
```diff
@@ -595,11 +600,21 @@ impl RegionServer {
             .handle_sync_region(engine_with_status.engine(), region_id, manifest_info)
             .await
     }
+
+    fn is_suspended(&self) -> bool {
+        self.suspend.load(Ordering::Relaxed)
+    }
+
+    pub(crate) fn suspend(&self) -> Arc<AtomicBool> {
```
sunng87: suspend feels like a verb here

src/frontend/src/error.rs
```diff
@@ -364,6 +364,12 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
+
+    #[snafu(display("Service suspended"))]
+    Suspended {
```
sunng87, line 369: I think `Suspended` is correct here. We should keep naming consistent

src/frontend/src/instance.rs
```diff
@@ -171,6 +173,14 @@ impl Instance {
     pub fn procedure_executor(&self) -> &ProcedureExecutorRef {
         self.statement_executor.procedure_executor()
     }
+
+    pub fn suspend(&self) -> Arc<AtomicBool> {
```
sunng87: sounds like a verb too

src/common/error/src/status_code.rs
MichaelScofield: renamed: `Suspend` -> `Suspended`

by mentioning `DeadlineExceeded`, do you mean to remove the `Suspended` status code and reuse the `DeadlineExceeded`?

sunng87: I mean we should use `ed` formal when naming

src/common/meta/src/heartbeat/handler/suspend.rs
MichaelScofield: I remember `compare_and_set` is heavier than just `load`. Most of the time it's not suspended, so it's better to `load` first here.

src/datanode/src/region_server.rs
MichaelScofield: renamed: `suspend` -> `suspend_state`

src/frontend/src/error.rs
MichaelScofield: it's already `Suspended`?

src/frontend/src/instance.rs
MichaelScofield: renamed: `suspend` -> `suspend_state`

src/common/meta/src/heartbeat/handler/suspend.rs
sunng87: if we don't use compare_and_set here, we lose the meaning of using a atomic for `self.suspend`. It just feels like to bypass the compiler check for a mutable value.

src/frontend/src/instance/grpc.rs
```diff
@@ -61,6 +61,8 @@ impl GrpcQueryHandler for Instance {
     type Error = Error;
 
     async fn do_query(&self, request: Request, ctx: QueryContextRef) -> Result<Output> {
+        ensure!(!self.is_suspended(), SuspendedSnafu);
```
fengjiachun: Can we implement this check in the interceptors for both HTTP and gRPC separately? This way, we can avoid making judgments in multiple places, reducing the risk of overlooking checks for certain modules.

fengjiachun (approved):
Others LGTM

src/common/meta/src/heartbeat/handler/suspend.rs
```diff
@@ -0,0 +1,69 @@
...
+    async fn handle(&self, context: &mut HeartbeatResponseHandlerContext) -> Result<HandleControl> {
+        let flip_state = |expect: bool| {
+            self.suspend
+                .compare_exchange(expect, !expect, Ordering::Relaxed, Ordering::Relaxed)
+                .is_ok()
+        };
+
+        if let Some((_, Instruction::Suspend)) = context.incoming_message.take() {
+            if flip_state(false) {
+                warn!("set suspend state");
```
sunng87:
```suggestion
                warn!("Suspend instruction received from meta, entering suspension state");
```

src/frontend/src/instance/grpc.rs
MichaelScofield: resolved in https://github.com/GreptimeTeam/greptimedb/pull/7370/commits/392d3780e356b5d8bf02c684378813d66f4ff32c

PR #7374, opened 2025-12-09
fix: part sort behavior

waynexia:

Correct part sort behavior on order by ... limit ... case

## PR Checklist

### DIFF
--- a/src/query/src/part_sort.rs
+++ b/src/query/src/part_sort.rs
@@ -284,7 +284,6 @@ struct PartSortStream {
     buffer: PartSortBuffer,
     expression: PhysicalSortExpr,
     limit: Option<usize>,
-    produced: usize,
     input: DfSendableRecordBatchStream,
     input_complete: bool,
     schema: SchemaRef,
@@ -340,7 +339,6 @@ impl PartSortStream {
             buffer,
             expression: sort.expression.clone(),
             limit,
-            produced: 0,
             input,
             input_complete: false,
             schema: sort.input.schema(),
@@ -565,7 +563,6 @@ impl PartSortStream {
             )
         })?;
 
-        self.produced += sorted.num_rows();
         drop(full_input);
         // here remove both buffer and full_input memory
         self.reservation.shrink(2 * total_mem);
@@ -624,7 +621,6 @@ impl PartSortStream {
             )
         })?;
 
-        self.produced += concat_batch.num_rows();
         Ok(concat_batch)
     }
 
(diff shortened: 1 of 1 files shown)


src/query/src/part_sort.rs
```diff
@@ -687,6 +694,12 @@ impl PartSortStream {
         cx: &mut Context<'_>,
     ) -> Poll<Option<datafusion_common::Result<DfRecordBatch>>> {
         loop {
+            // Early termination: if we've already produced enough rows,
+            // don't poll more input - just return
+            if matches!(self.limit, Some(0)) {
```
discord9, line 699: should this be like `self.limit.map(|limit|limit<=self.produced).unwrap_or(false)` ?

src/query/src/part_sort.rs
```diff
@@ -701,7 +714,24 @@ impl PartSortStream {
...
                 && evaluating_batch.num_rows() != 0
             {
+                // Check if we've already processed all partitions
+                if self.cur_part_idx >= self.partition_ranges.len() {
+                    // All partitions processed, discard remaining data
+                    if self.buffer.is_empty() {
+                        return Poll::Ready(None);
+                    } else {
+                        let sorted_batch = self.sort_buffer()?;
+                        self.limit = self
```
discord9, line 724: why dec limit instead of adding to `produced`?

src/query/src/part_sort.rs
```diff
@@ -666,6 +667,12 @@ impl PartSortStream {
         let sorted_batch = self.sort_buffer();
         // step to next proper PartitionRange
         self.cur_part_idx += 1;
+
+        // If we've processed all partitions, discard remaining data
+        if self.cur_part_idx >= self.partition_ranges.len() {
+            return sorted_batch.map(|x| if x.num_rows() == 0 { None } else { Some(x) });
+        }
```
evenyag, line 674: Maybe we can add a debug_assert to check `remaining_range.is_empty()`. But if our unit test case contains out-of-part data, please ignore this comment.

src/query/src/part_sort.rs
```diff
@@ -624,6 +624,7 @@ impl PartSortStream {
             )
         })?;
 
+        self.produced += concat_batch.num_rows();
```
evenyag: Looks like the `produced` is never used.

src/query/src/part_sort.rs
waynexia: I'm going to remove `produced` count, it's less descriptive as it mixes all output rows among different parts

discord9 (approved):

PR #7383, opened 2025-12-10
feat: mark index outdated

discord9:

mark index file as outdated for FileHandle, this allow `LocalFilePurger` to properly delete old index file, however `ObjectStorePurger` still need impl later, notice this require `FileHandleInner` with unique (FileId, IndexVersion) should only have one instance on a datanode

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/version.rs
+++ b/src/mito2/src/sst/version.rs
@@ -63,10 +63,8 @@ impl SstVersion {
                 .files
                 .entry(file.file_id)
                 .and_modify(|f| {
-                    let is_same = *f.meta_ref() == file;
-                    let is_index_up_to_date = f.meta_ref().exists_index() && file.exists_index() && f.index_id().version >= new_index_version;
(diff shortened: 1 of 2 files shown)


src/mito2/src/sst/version.rs
```diff
@@ -57,9 +57,15 @@ impl SstVersion {
...
+            let new_index_version = file.index_version;
+            if let Some(old_file_handle) = self.levels[level as usize]
                 .files
-                .insert(file.file_id, FileHandle::new(file, file_purger.clone()));
+                .insert(file.file_id, FileHandle::new(file, file_purger.clone()))
+                && old_file_handle.index_id().version < new_index_version
+            {
+                // now the old file handle's index is outdated
+                old_file_handle.set_index_outdated(true);
+            }
```
evenyag: If the file already exists, then we should only replace the handle when the index is outdated.

evenyag (approved):

PR #6728, opened 2025-08-13
fix: correct offset's symbol

waynexia:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

https://github.com/GreptimeTeam/greptimedb/issues/6703

## What's changed and what's your intention?

I reversed the symbol of offset for a long while 🥲

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #7392, opened 2025-12-11
fix: promql offset direction

### LINKED ISSUE #7361: Offset modifier behavior inverted: positive offsets query future instead of past
### What type of bug is this?

Incorrect result

Query Engine

### Minimal reproduce step

Insert time series data with known timestamps:

```sql
INSERT INTO metric (timestamp, value) VALUES 
  (940, 10),
  (1000, 20),
  (1060, 30);
```

Query at evaluation time T=1000 with positive offset:

```
metric offset 60s
```

Query at evaluation time T=1000 without offset:

```
metric
```

Compare the results

Expected behavior per Prometheus spec: metric offset 60s at T=1000 should return the value from T=940 (60 seconds in the past).
Actual behavior: metric offset 60s at T=1000 returns the value from T=1060 (60 seconds in the future).

  Test Data

  Timestamp: 940  -> Value: 10
  Timestamp: 1000 -> Value: 20
  Timestamp: 1060 -> Value: 30

  Query Results at Evaluation Time T=1000

  | Query              | Expected Value (Prometheus Spec) | Actual Value (GreptimeDB) | Source Timestamp |
  |--------------------|----------------------------------|---------------------------|------------------|
  | metric             | 20                               | 20 ✅                      | T=1000           |
  | metric offset 60s  | 10 (from T=940)                  | 30 ❌                      | T=1060 (FUTURE) |
  | metric offset -60s | 30 (from T=1060)                 | 10 ❌                      | T=940 (PAST)    |

  Analysis

  Per Prometheus specification:
  - offset 60s should look 60 seconds into the PAST (T=1000 - 60 = T=940)
  - Expected result: Value 10

  GreptimeDB actual behavior:
  - offset 60s looks 60 seconds into the FUTURE (T=1000 + 60 = T=1060)
  - Actual result: Value 30

According to the [Prometheus PromQL documentation](https://prometheus.io/docs/prometheus/latest/querying/basics/#offset-modifier):

"The offset modifier allows changing the time offset for individual instant and range vectors in a query. For example, the following expression returns the value of http_requests_total 5 minutes in the past relative to the current query evaluation time: http_requests_total offset 5m"

When querying metric offset 60s at evaluation time T=1000, the query should:

Fetch data from timestamp 940 (1000 - 60)
Return that data aligned to the query evaluation time

Positive offsets should retrieve data from the past, and negative offsets should retrieve data from the future (for temporal comparisons forward in time, as documented).

### What did you see instead?

The behavior is inverted:

Positive offsets (e.g., offset 60s) query data from the future
Negative offsets (e.g., offset -60s) query data from the past

This is backwards from the Prometheus specification.
Root cause: PR #6728 (commit 2992e703) attempted to fix offset behavior but inverted the logic:
In planner.rs, the changes were:

start - offset_duration → start + offset_duration
end - offset_duration → end + offset_duration

This causes the planner to query future timestamps when a positive offset is specified. While normalize.rs adjusts the timestamps back, the fundamental data selection is wrong—the system fetches data from the wrong time window entirely.
Workaround: Currently, users must use negative offsets to query historical data (e.g., offset -60s to look 60 seconds into the past), which is counterintuitive and incompatible with standard PromQL.

Linux x64

1.0.0-beta.2

```bash

```

waynexia:

Fixes #7361 

Thanks to @ctorrisi. I got :dizzy_face:  while working with this offset... I remember I was pretty sure the previous impl is wrong and https://github.com/GreptimeTeam/greptimedb/pull/6728 was the fix. Would look back into https://github.com/GreptimeTeam/greptimedb/issues/6703 again to make sure both sides are working as expected. 

And though it might be a bit duplicated with existing suites, I still add a new sqlness case dedicated to the direction case `offset_direction.sql`

## PR Checklist

### DIFF
--- a/src/promql/src/extension_plan/normalize.rs
+++ b/src/promql/src/extension_plan/normalize.rs
@@ -365,7 +365,7 @@ impl SeriesNormalizeStream {
             Arc::new(ts_column.clone()) as _
         } else {
             Arc::new(TimestampMillisecondArray::from_iter(
-                ts_column.iter().map(|ts| ts.map(|ts| ts - self.offset)),
+                ts_column.iter().map(|ts| ts.map(|ts| ts + self.offset)),
             ))
         };
         let mut columns = input.columns().to_vec();
@@ -518,11 +518,11 @@ mod test {
             "+---------------------+--------+------+\
             \n| timestamp           | value  | path |\
             \n+---------------------+--------+------+\
-            \n| 1970-01-01T00:00:59 | 0.0    | foo  |\
-            \n| 1970-01-01T00:01:59 | 1.0    | foo  |\
-            \n| 1969-12-31T23:59:59 | 10.0   | foo  |\
-            \n| 1970-01-01T00:00:29 | 100.0  | foo  |\
-            \n| 1970-01-01T00:01:29 | 1000.0 | foo  |\
+            \n| 1970-01-01T00:01:01 | 0.0    | foo  |\
+            \n| 1970-01-01T00:02:01 | 1.0    | foo  |\
+            \n| 1970-01-01T00:00:01 | 10.0   | foo  |\
+            \n| 1970-01-01T00:00:31 | 100.0  | foo  |\
+            \n| 1970-01-01T00:01:31 | 1000.0 | foo  |\
             \n+---------------------+--------+------+",
         );
 


evenyag (approved):
😵

evenyag:
Need to fix sqlness.

PR #7412, opened 2025-12-16
feat!: gc versioned index

discord9:

gc worker now able to gc outdated index:
- change to region manifest, add `RemovedFile` type which can distinguish between remove a file/index pair or just remove a outdated index file, note this change isn't consider breaking because this field is always emtpy when persistent to disk for now(since gc never get enabled)
- handle remove outdated index in `apply_edit`
- handle checking for file reference for outdated index file in gc worker(check if in manifest/in use by tmp ref/is still within lingering time i.e.) 

breaking: changed region manifest's `removed_files` field,  but it should stay empty when gc is disabled

## PR Checklist

### DIFF
--- a/src/mito2/src/gc.rs
+++ b/src/mito2/src/gc.rs
@@ -49,6 +49,9 @@ use crate::region::{MitoRegionRef, RegionRoleState};
 use crate::sst::file::{RegionFileId, RegionIndexId, delete_files, delete_index};
 use crate::sst::location::{self};
 
+#[cfg(test)]
+mod worker_test;
+
 /// Helper function to determine if a file should be deleted based on common logic
 /// shared between Parquet and Puffin file types.
 fn should_delete_file(
@@ -80,9 +83,6 @@ fn should_delete_file(
             is_unknown_linger_time_exceeded()
         }
 }
-
-#[cfg(test)]
-mod worker_test;
 
 /// Limit the amount of concurrent GC jobs on the datanode
 pub struct GcLimiter {


src/mito2/src/gc.rs
```diff
@@ -575,35 +629,87 @@ impl LocalGcWorker {
...
+                    crate::cache::file_cache::FileType::Parquet => {
+                        // notice this cause we don't track index version for parquet files
+                        // since entries comes from listing, we can't get index version from path
+                        RemovedFile::File(file_id, None)
+                    }
+                    crate::cache::file_cache::FileType::Puffin(version) => {
+                        RemovedFile::Index(file_id, version)
+                    }
+                };
+                ready_for_delete.push(removed_file);
```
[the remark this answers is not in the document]
discord9: Explain: entries include all entry under region dir, so every parquet/puffin files will be processed anyway

> When listing files from object store and creating `RemovedFile::File(file_id, None)` for parquet files, the index version is lost. This None value becomes 0 via `unwrap_or(0)` at line 376, which is then passed to `delete_sst`. The `delete_sst` method deletes index versions from 0 up to the provided version (inclusive). If a parquet file has an index at a higher version (e.g., version 5), only version 0 will be deleted, leaving versions 1-5 as orphaned files in object storage. Consider tracking the actual index version when listing parquet files, or modifying the deletion logic to handle this case.

src/mito2/src/gc.rs
```diff
@@ -535,38 +562,65 @@ impl LocalGcWorker {
...
+                    GC_SKIPPED_UNPARSABLE_FILES.inc();
                     continue;
                 }
             };
 
-            if may_linger_filenames.contains(&file_id) {
-                all_in_exist_linger_files.insert(file_id);
-            }
+            let should_delete = match file_type {
+                crate::cache::file_cache::FileType::Parquet => {
```
fengjiachun: I prefer to avoid using fully qualified names whenever possible.
```suggestion
                FileType::Parquet => {
```

src/mito2/src/gc.rs
```diff
@@ -575,35 +629,88 @@ impl LocalGcWorker {
...
+                    };
+                    !is_in_manifest
+                        && !is_in_tmp_ref
+                        && if is_known {
+                            is_eligible_for_delete
+                        } else {
+                            is_unknown_linger_time_exceeded()
+                        }
+                }
+                crate::cache::file_cache::FileType::Puffin(version) => {
```
fengjiachun: Same as above

src/mito2/src/gc.rs
```diff
@@ -575,35 +629,88 @@ impl LocalGcWorker {
...
+                    };
+                    !is_in_manifest
+                        && !is_in_tmp_ref
+                        && if is_known {
+                            is_eligible_for_delete
+                        } else {
+                            is_unknown_linger_time_exceeded()
+                        }
+                }
+            };
```
fengjiachun: The logic of the two branches is almost identical, differing only in how they handle version. Recommendation: extract the common logic.

src/mito2/src/manifest/action.rs
```diff
@@ -193,17 +193,27 @@ impl RegionManifestBuilder {
...
         }
-        self.removed_files.add_removed_files(
+        removed_files.extend(
             edit.files_to_remove
                 .iter()
-                .map(|meta| meta.file_id)
-                .collect(),
-            edit.timestamp_ms
-                .unwrap_or_else(|| Utc::now().timestamp_millis()),
+                .map(|f| RemovedFile::File(f.file_id, f.index_version())),
```
evenyag, line 210: Do we have compatiblity issue? Since the struct in files_to_remove changed.

src/mito2/src/sst/file.rs
```diff
@@ -312,6 +312,14 @@ impl FileMeta {
         !self.available_indexes.is_empty()
     }
 
+    pub fn index_version(&self) -> Option<IndexVersion> {
+        if self.exists_index() {
+            Some(self.index_version)
+        } else {
+            None
+        }
+    }
```
evenyag, line 321: question: Should we return 0 as index version?

src/mito2/src/sst/file_ref.rs
```diff
@@ -132,7 +139,15 @@ impl FileReferenceManager {
...
         {
-            let file_ref = FileRef::new(file_meta.region_id, file_meta.file_id);
+            let file_ref = FileRef::new(
+                file_meta.region_id,
+                file_meta.file_id,
+                if file_meta.exists_index() {
+                    Some(file_meta.index_version)
+                } else {
+                    None
+                },
```
evenyag: Does file_meta provide an index_version() method? Maybe we can wrap one.

src/store-api/src/storage/file.rs
```diff
@@ -77,38 +77,47 @@ impl FromStr for FileId {
 pub struct FileRef {
     pub region_id: RegionId,
     pub file_id: FileId,
+    pub index_version: Option<IndexVersion>,
```
evenyag, line 80: Maybe we can use 0 as default?

src/store-api/src/storage/file.rs
```diff
@@ -77,38 +77,47 @@ impl FromStr for FileId {
...
     /// Manifest version when this manifest is read for it's files
     pub manifest_version: HashMap<RegionId, ManifestVersion>,
 }
 
 #[derive(Clone, Default, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct GcReport {
     /// deleted files per region
+    /// TODO(discord9): change to `RemovedFile`?
     pub deleted_files: HashMap<RegionId, Vec<FileId>>,
+    pub deleted_indexes: HashMap<RegionId, Vec<(FileId, IndexVersion)>>,
```
evenyag, line 107: Do we need to consider compatibility?

src/mito2/src/gc.rs
```diff
@@ -28,25 +28,59 @@ use std::time::Duration;
...
+        && !is_in_tmp_ref
+        && if is_known {
+            is_eligible_for_delete
+        } else {
+            is_unknown_linger_time_exceeded()
+        }
+}
+
 #[cfg(test)]
 mod worker_test;
```
evenyag: nitpick: consider moving this to the head of the mod file.

src/mito2/src/gc.rs
```diff
@@ -247,7 +282,12 @@ impl LocalGcWorker {
                 .cloned()
                 .unwrap_or_else(HashSet::new);
             let files = self.do_region_gc(region.clone(), &tmp_ref_files).await?;
-            deleted_files.insert(*region_id, files);
+            let index_files = files
+                .iter()
+                .filter_map(|f| f.index_version().map(|v| (f.file_id(), v)))
```
evenyag, line 287: if index_version is None, we will skip the index file,  is it expected?

src/mito2/src/manifest/action.rs
discord9: > Do we have compatiblity issue? Since the struct in files_to_remove changed.

not yet as persisent to disk's manifest is always empty in field `removed_files` and the `edit.files_to_remove` is not changed anyway

src/mito2/src/sst/file.rs
discord9: > question: Should we return 0 as index version?

no? This differ between having a index/not having one?

src/store-api/src/storage/file.rs
discord9: > Do we need to consider compatibility?

this is not persistent, only need metasrv/datanode to be the same version when gc is enabled

src/mito2/src/gc.rs
discord9: > if index_version is None, we will skip the index file, is it expected?

yes, to distinguish between having a index file or not

PR #7420, opened 2025-12-16
feat: repartition map kv

discord9:

repartition map kv, repartition procedure will add new repartition src/dst regions mapping, and gc scheduler will gradually check that dst region still hold src region's file and remove those mapping 

## PR Checklist

### DIFF
--- a/src/common/meta/src/key/table_repart.rs
+++ b/src/common/meta/src/key/table_repart.rs
@@ -599,12 +599,12 @@ mod tests {
 
         let value = TableRepartValue { src_to_dst };
 
-        // Create the table part
+        // Create the table repart
         let (txn, _) = manager.build_create_txn(1024, &value).unwrap();
         let result = kv.txn(txn).await.unwrap();
         assert!(result.succeeded);
 
-        // Get the table part
+        // Get the table repart
         let retrieved = manager.get(1024).await.unwrap().unwrap();
         assert_eq!(retrieved, value);
     }
@@ -618,7 +618,7 @@ mod tests {
             src_to_dst: BTreeMap::new(),
         };
 
-        // Create initial table part
+        // Create initial table repart
         let (create_txn, _) = manager.build_create_txn(1024, &initial_value).unwrap();
         let result = kv.txn(create_txn).await.unwrap();
         assert!(result.succeeded);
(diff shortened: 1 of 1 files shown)


src/common/meta/src/key/table_repart.rs
```diff
@@ -0,0 +1,853 @@
...
+    /// Updates mappings from src region to dst regions.
+    /// Should be called once repartition is done.
+    pub async fn update_mappings(&self, src: RegionId, dst: &[RegionId]) -> Result<()> {
+        let table_id = src.table_id();
+
+        // Get current table repart with raw bytes for CAS operation
+        let current_table_repart = self
+            .get_with_raw_bytes(table_id)
+            .await?
+            .context(crate::error::TablePartitionNotFoundSnafu { table_id })?;
```
WenyXu: How about `TableRepartNotFoundSnafu `?

src/common/meta/src/error.rs
```diff
@@ -224,6 +224,13 @@ pub enum Error {
         location: Location,
     },
 
+    #[snafu(display("Failed to find table partition for table id {}", table_id))]
```
WenyXu:
```suggestion
    #[snafu(display("Failed to find table repartition metadata for table id {}", table_id))]
```

src/common/meta/src/key/table_repart.rs
```diff
@@ -0,0 +1,853 @@
...
+impl TableRepartValue {
+    /// Creates a new TableRepartValue with an empty src_to_dst map.
+    pub fn new() -> Self {
+        Default::default()
+    }
+    /// Update mapping from src region to dst regions. Should be called once repartition is done.
+    ///
+    /// If `dst` is empty, this method does nothing.
+    pub fn update_mappings(&mut self, src: RegionId, dst: &[RegionId]) {
+        self.src_to_dst.entry(src).or_default().extend(dst);
```
WenyXu, line 112: Should we check the `dst`, if it's empty?

waynexia (approved):

src/common/meta/src/key/table_repart.rs
```diff
@@ -0,0 +1,856 @@
...
+            crate::error::MetadataCorruptionSnafu {
+                err_msg: format!(
+                    "Failed to update mappings for table {}: CAS operation failed",
+                    table_id
+                ),
+            }
+        );
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
discord9: > The update_mappings method fails immediately if the CAS operation fails due to concurrent modification, returning a MetadataCorruption error. Consider implementing retry logic similar to other metadata update operations in the codebase to handle transient conflicts gracefully. The CAS operation can legitimately fail when multiple operations occur concurrently, which should be retried rather than treated as data corruption.

it should failed as anyone trying to update it must have acquired the lock for `src` region id

PR #7430, opened 2025-12-17
chore: remove `canonicalize`

shuiyisong:

In cases when the file is a link symbol, the exact file itself doesn't trigger a notify event. Now we make watching the directory the default behavior, which means any changes under the directory would trigger a callback reload.

Note: the modification root should still under the same parent of the link symbol, though it's OK to be file or folder.

## PR Checklist


sunng87:
Can we enable this by default?

shuiyisong:
Update: remove the option and make the watching dir behavior default

PR #7441, opened 2025-12-18
feat: file range dynamic filter

discord9:

try to filter file range by dynamic filter before read file, hence dynamic filter might reduce numbers of row group/files read, still need impl collect dynamic filter to function, but datafusion didn't handle alias in dynamic filter yet so not working

TODO(in next PR): (for `with_dyn_filters`)The caller of this method must also ensure the dynamic filters are for tags, fields or timestamps and only use some of them. For example, when skip fields is true, we should not use dyn filters for fields.

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet/file_range.rs
+++ b/src/mito2/src/sst/parquet/file_range.rs
@@ -157,7 +157,7 @@ impl FileRange {
             PreFilterMode::SkipFieldsOnDelete => {
                 // Check if this specific row group contains delete op
                 row_group_contains_delete(
-                    &self.context.reader_builder.parquet_metadata(),
+                    self.context.reader_builder.parquet_metadata(),
                     self.row_group_idx,
                     self.context.reader_builder.file_path(),
                 )


src/mito2/src/sst/parquet/file_range.rs
```diff
@@ -114,12 +118,43 @@ impl FileRange {
...
+        let pred = Predicate::new(vec![]).with_dyn_filters(self.context.base.dyn_filters.clone());
+
+        pred.prune_with_stats(&stats, prune_schema.arrow_schema())
+            .first()
+            .cloned()
+            .unwrap_or(true) // unexpected, not skip just in case
+    }
+
     /// Returns a reader to read the [FileRange].
     pub(crate) async fn reader(
```
evenyag, line 170: Also need to support this in flat_reader()

src/mito2/src/memtable/bulk/context.rs
```diff
@@ -85,7 +85,10 @@ impl BulkIterContext {
         Ok(Self {
             base: RangeBase {
                 filters: simple_filters,
+                dyn_filters: Arc::new(vec![]), // TODO(discord9): pass dynamic filters
                 read_format,
+                skip_fields: false,
```
evenyag: Is it safe to set skip_files to false?

src/mito2/src/sst/parquet/file_range.rs
```diff
@@ -114,12 +118,43 @@ impl FileRange {
...
+        let read_format = self.context.read_format();
+        let prune_schema = &self.context.base.prune_schema;
+        let stats = RowGroupPruningStats::new(
+            std::slice::from_ref(curr_row_group),
+            read_format,
+            None, // TODO(discord9): check if need expected metadata
+            self.context.base.skip_fields,
+        );
+
+        let pred = Predicate::new(vec![]).with_dyn_filters(self.context.base.dyn_filters.clone());
```
evenyag, line 145: Can we reuse the pred in the range?

src/table/src/predicate.rs
```diff
@@ -65,6 +70,15 @@ impl Predicate {
...
             exprs: Arc::new(exprs),
+            dyn_filters: Arc::new(vec![]),
+        }
+    }
+
+    /// Sets the dynamic filter physical exprs.
+    pub fn with_dyn_filters(self, dyn_filters: Arc<Vec<DynamicFilterPhysicalExpr>>) -> Self {
+        Self {
+            exprs: self.exprs,
+            dyn_filters,
```
evenyag, line 81: non-blocking: The caller of this method must also ensure the dynamic filters are for tags, fields or timestamps and only use some of them. For example, when skip fields is true, we should not use dyn filters for fields.

Issue #7457, opened 2025-12-23
Panic when sorting > 1800,000 rows


yihong0618:
### What type of bug is this?

Crash

### What subsystems are affected?

Query Engine

### Minimal reproduce step

from discuss https://github.com/orgs/GreptimeTeam/discussions/7448

60w is not 100% reproducer needs make it bigger like 180w

```


```

### What did you expect to see?

right answer 

### What did you see instead?

crash

### What operating system did you use?

mac

### What version of GreptimeDB did you use?

main

### Relevant log output and stack trace

```bash
2025-12-23T00:27:31.228208Z ERROR common_telemetry::panic_hook: panicked at src/query/src/part_sort.rs:562:43:
attempt to subtract with overflow backtrace=   0: backtrace::backtrace::libunwind::trace
             at /Users/hyi/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/backtrace-0.3.75/src/backtrace/libunwind.rs:117:9
      backtrace::backtrace::trace_unsynchronized
             at /Users/hyi/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/backtrace-0.3.75/src/backtrace/mod.rs:66:14
   1: backtrace::backtrace::trace
             at /Users/hyi/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/backtrace-0.3.75/src/backtrace/mod.rs:53:14
   2: backtrace::capture::Backtrace::create
             at /Users/hyi/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/backtrace-0.3.75/src/capture.rs:294:9
   3: backtrace::capture::Backtrace::new
             at /Users/hyi/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/backtrace-0.3.75/src/capture.rs:259:22
   4: common_telemetry::panic_hook::set_panic_hook::{{closure}}
             at /Users/hyi/prs/greptimedb/src/common/telemetry/src/panic_hook.rs:37:25
   5: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/fa3155a644dd62e865825087b403646be01d4cef/library/alloc/src/boxed.rs:1987:9
      std::panicking::panic_with_hook
             at /rustc/fa3155a644dd62e865825087b403646be01d4cef/library/std/src/panicking.rs:842:13
   6: std::panicking::panic_handler::{{closure}}
[...]

PR #7458, opened 2025-12-23
fix: close issue #7457 guard against empty buffer

yihong0618:

## PR Checklist

as diff and comments and reproducer

### DIFF
--- a/src/query/src/part_sort.rs
+++ b/src/query/src/part_sort.rs
@@ -1076,6 +1076,60 @@ mod test {
     use super::*;
     use crate::test_util::{MockInputExec, new_ts_array};
 
+    #[tokio::test]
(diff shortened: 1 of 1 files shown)


discord9:
This is a good guard logic, but doesn't necessary fix the bug, gonna merge it but keep the discussion open as I found some logic still need fix in `PartSort`(i.e. sort_top_buffer construct dyn filter wrong, and maybe other logic bug that need investigate)

waynexia (approved):

yihong0618:
> This is a good guard logic, but doesn't necessary fix the bug, gonna merge it but keep the discussion open as I found some logic still need fix in `PartSort`(i.e. sort_top_buffer construct dyn filter wrong, and maybe other logic bug that need investigate)

cool

PR #7460, opened 2025-12-23
fix: part sort share same topk dyn filter&early stop use dyn filter

discord9:

as title, previously it construct a new one without sort column as it's child, which could cause bug, ~but before part sort's stop early optimization `sort_top_buffer` is called only once so bug is hidden, but after early stop optimization `sort_top_buffer` might got called multiple times hence cause problem~ and also dyn filter need to be shared to properly prune stuff
- make sure dyn filter is shared among `PartSortExec`/`PartSortStream`/`TopK`
- Also due to the intereface of `TopK`, had to refactor to using dyn filter to determine early stop as it's the most clear way, in term of correctness, `TopK` also use this dyn filter to filter before `insert_batch` so the result should be correct since the logic of filtering is the same

## PR Checklist

### DIFF
--- a/src/query/src/part_sort.rs
+++ b/src/query/src/part_sort.rs
@@ -1157,7 +1157,7 @@ mod test {
 
         // The TopK result buffer is empty, so we cannot determine early-stop.
         // Ensure this path returns `Ok(false)` (and, importantly, does not panic).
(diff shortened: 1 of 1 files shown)


waynexia (approved):

fengjiachun (approved):
LGTM

PR #7461, opened 2025-12-23
fix(mito2): pass partition expr explicitly to flush task for region

WenyXu:

#6558 

When a region is in staging mode and performs a flush, the partition expression is read from `version.metadata.partition_expr` at flush time. However, in staging mode, the partition expression should come from `staging_partition_expr` which is set when entering staging. This can cause SST files generated during staging to have incorrect or missing partition expressions.

- The partition expression is now passed explicitly to `RegionFlushTask` when the task is created
- Added `get_partition_expr_str()` function that returns the correct partition expression based on whether the region is in staging mode:
 - Staging mode: returns `staging_partition_expr`
  - Normal mode: returns `version.metadata.partition_expr`
- Extracted parse_partition_expr() as a reusable helper function
- All flush task creation call sites updated to use `get_partition_expr_str()`

## PR Checklist

### DIFF
--- a/src/mito2/src/region.rs
+++ b/src/mito2/src/region.rs
@@ -727,11 +727,12 @@ impl MitoRegion {
         Ok(())
     }
 
-    /// Gets the partition expr for a region.
+    /// Returns the partition expression string for this region.
     ///
-    /// If the region is in staging mode, returns the partition expr from the staging partition expr.
-    /// Otherwise, returns the partition expr from the version metadata.
-    pub fn partition_expr_str(&self) -> Option<String> {
+    /// If the region is currently in staging state, this returns the partition expression held in
+    /// the staging partition field. Otherwise, it returns the partition expression from the primary
+    /// region metadata (current committed version).
+    pub fn maybe_staging_partition_expr_str(&self) -> Option<String> {
         let is_staging = self.is_staging();
         if is_staging {
             let staging_partition_expr = self.staging_partition_expr.lock().unwrap();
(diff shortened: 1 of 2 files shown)


src/mito2/src/worker/handle_flush.rs
```diff
@@ -220,14 +230,17 @@ impl<S: LogStore> RegionWorkerLoop<S> {
             }
             self.update_topic_latest_entry_id(region);
 
+            let is_staging = region.is_staging();
+            let partition_expr = get_partition_expr_str(region, is_staging);
```
evenyag: We should add a method to MitoRegion.

WenyXu: [9ec7e3e](https://github.com/GreptimeTeam/greptimedb/pull/7461/commits/9ec7e3eff1921eb07582bad0c475fec3b51aaa96)

src/mito2/src/region.rs
```diff
@@ -725,6 +726,27 @@ impl MitoRegion {
 
         Ok(())
     }
+
+    /// Gets the partition expr for a region.
+    ///
+    /// If the region is in staging mode, returns the partition expr from the staging partition expr.
+    /// Otherwise, returns the partition expr from the version metadata.
+    pub fn partition_expr_str(&self) -> Option<String> {
```
evenyag: If we only use it in flush, maybe we can use a new name.

evenyag (approved):

src/mito2/src/region.rs
WenyXu: [0ad6a9e](https://github.com/GreptimeTeam/greptimedb/pull/7461/commits/0ad6a9e9fadb4c71d0c1677fc2a78ec6a8a9fbd0)

PR #7462, opened 2025-12-23
refactor: explicitly define json struct to ingest jsonbench data

MichaelScofield:

create table with json struct defined to ingest json data

see the test `test_load_jsonbench_data`

## PR Checklist


src/datatypes/src/json.rs
```diff
@@ -144,12 +156,54 @@ impl<'a> JsonContext<'a> {
...
+    let Some(json_object) = json.as_object_mut() else {
+        return InvalidJsonSnafu {
+            value: "expect JSON object when struct is provided",
+        }
+        .fail();
+    };
+    let mut encoded = BTreeMap::new();
+
+    fn extract_field(json_object: &mut Map<String, Json>, field: &str) -> Result<Option<Json>> {
+        let (first, rest) = field.split_once('.').unwrap_or((field, ""));
```
sunng87, line 169: we can use serde_json Value pointer for now to access data more reliably: https://docs.rs/serde_json/latest/serde_json/enum.Value.html#method.pointer

sunng87 (approved):
Should be good for initial attempt to reproduce JSONBench with new data structure

src/datatypes/src/json.rs
MichaelScofield: I need to take the field out from the json object, so the readonly `pointer` method is not useful here. And I can't use the `pointer_mut` + `take` either, because `take` would leave a "null" value in place.

src/datatypes/src/json.rs
```diff
@@ -144,12 +156,54 @@ impl<'a> JsonContext<'a> {
...
+        if rest.is_empty() {
+            Ok(json_object.remove(first))
+        } else {
+            let Some(value) = json_object.get_mut(first) else {
+                return Ok(None);
+            };
+            let json_object = value.as_object_mut().with_context(|| InvalidJsonSnafu {
+                value: format!(r#"expect "{}" an object"#, first),
+            })?;
+            extract_field(json_object, rest)
```
fengys1996, line 180: nit: Should we consider the risk of stack overflow caused by recursion?

fengys1996 (approved):

src/datatypes/src/json.rs
MichaelScofield: it might require hundreds of nested layers of json to cause the overflow, I think we can ignore that for now

PR #7463, opened 2025-12-23
feat(standalone): support restoring metadata from snapshot on startup

WenyXu:

This PR adds support for restoring metadata from a snapshot file when starting a standalone GreptimeDB instance for the first time.

### New Configuration Options

Added two new options to `metadata_store` configuration:

| Option | Type | Default | Description |
|--------|------|---------|-------------|
| `init_metadata_path` | `Option<String>` | `None` | Path to a metadata snapshot file for restoration (relative to data home) |
| `ignore_metadata_snapshot_restore_error` | `bool` | `false` | Whether to ignore errors during restoration |

### Implementation Details

- Added `restore_metadata_from_snapshot` function in `standalone` crate
- The restoration only proceeds when the target metadata store is empty (checked via `check_target_source_clean`)
- If the store is not empty, a warning is logged and restoration is skipped
- Changed `KvBackendConfig` from `Copy` to `Clone` to support the new `Option<String>` field
- Updated `build_metadata_kvbackend` signature to take `&KvBackendConfig` instead of `KvBackendConfig`

### Safety Guarantees

- **No accidental overwrites**: Restoration is skipped if existing data is detected
- **Graceful error handling**: Optional `ignore_metadata_snapshot_restore_error` flag for fault-tolerant scenarios

## Example Configuration
```toml
[metadata_store]
init_metadata_path = "backup/metakv.metadata.fb"
ignore_metadata_snapshot_restore_error = false## Testing
```

Added unit tests covering:
- Successful restoration from snapshot
- Skip restoration when existing data is present
- Error handling with `ignore_error` flag

## PR Checklist

### DIFF
--- a/tests-integration/tests/http.rs
+++ b/tests-integration/tests/http.rs
@@ -1468,6 +1468,7 @@ prefill_log_files = false
 file_size = "64MiB"
 purge_threshold = "256MiB"
 purge_interval = "1m"
+ignore_metadata_snapshot_restore_error = false
 
 [procedure]
 max_retry_times = 3


killme2008:
Restoring backup or snapshot files in this manner is not ideal. A better approach would be to implement this feature as the command-line tool's subcommand for restoring metadata from these files. 

Configurations should be designed as long-term runtime values rather than temporary ones.

cc @fengjiachun @WenyXu

PR #7465, opened 2025-12-24
fix: load with latest

shuiyisong:

This PR adds the check whether loading pipeline into the cache should also write the `latest` cache entry.
Currently it doesn't. However, If the request doesn't carry the version (in most cases), the pipeline queried from the database is actually the `latest` one, so it should be put into the `latest` entry.

## PR Checklist


shuiyisong:
use #7491

PR #7469, opened 2025-12-24
feat: adjust some args to gc worker

discord9:

some small changes to gc parameters
- tmp ref files now exclude in manifest files
- gc worker will skip if manifest version in tmp ref doesn't match current manifest files
- change gc scheduler to even more tend to gc regions with more files to delete

## PR Checklist

### DIFF
--- a/src/mito2/src/gc.rs
+++ b/src/mito2/src/gc.rs
@@ -332,19 +332,16 @@ impl LocalGcWorker {
 
         let manifest = region.manifest_ctx.manifest().await;
         // If the manifest version does not match, skip GC for this region to avoid deleting files that are still in use.
-        if self
+        let file_ref_manifest_version = self
             .file_ref_manifest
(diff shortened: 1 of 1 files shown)


src/mito2/src/gc.rs
```diff
@@ -328,15 +329,37 @@ impl LocalGcWorker {
...
+            .get(&region.region_id())
+            .cloned()
+            != Some(manifest.manifest_version)
+        {
+            // should be rare enough(few seconds after leader update manifest version), just skip gc for this region
+            warn!(
+                "Tmp ref files manifest version {:?} does not match region {} manifest version {}, skip gc for this region",
+                self.file_ref_manifest
+                    .manifest_version
+                    .get(&region.region_id()),
```
evenyag: Should we assign this to a variable so we don't have to get it from the map here again.

PR #7476, opened 2025-12-24
fix: fix sqlness test error about double precision

Standing-Man:

## PR Checklist

### DIFF
--- a/tests/cases/standalone/common/join/join_with_aggregates.result
+++ b/tests/cases/standalone/common/join/join_with_aggregates.result
@@ -96,7 +96,7 @@ FROM (
     s."location",
     COUNT(DISTINCT s.sensor_id) as sensor_count,
     COUNT(r.reading_id) / COUNT(DISTINCT s.sensor_id) as avg_readings_per_sensor,
-    AVG(r."value") as location_avg_value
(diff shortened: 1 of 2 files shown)


Standing-Man:
@killme2008, PTAL

yihong0618:
sqlness failed

tests/cases/standalone/common/join/join_with_aggregates.result
evenyag: Can we trunc the avg value here to avoid the precision issue? @discord9 @killme2008

discord9: > Can we trunc the avg value here to avoid the precision issue? @discord9 @killme2008

reasonable, slight difference in different platform is expected

Standing-Man: How can we ensure that tests pass across different local platforms?

killme2008: We should truncate the result for cross-platform compatibility, `round(avg(value), 6)` eg.

Standing-Man: It works as expected. Thank you for your helpful tips.

PR #7486, opened 2025-12-25
feat: Implement per range stats for bulk memtable

evenyag:

Previously, all memtable ranges shared the same MemtableStats, which caused all flushed ranges have the same sequence number. However, we consider ranges with the same sequence as the same file group so the compaction picker won't compact these files into one file.

This PR moves the memtable stats to MemtableRange so each range can have its own sequence and won't be considered as the same file group.
```rust
pub struct MemtableRange {
    /// Shared context.
    context: MemtableRangeContextRef,
    /// Statistics for this memtable range.
    stats: MemtableStats,
}
```

## PR Checklist

### DIFF
--- a/src/mito2/src/memtable/simple_bulk_memtable.rs
+++ b/src/mito2/src/memtable/simple_bulk_memtable.rs
@@ -249,14 +249,7 @@ impl Memtable for SimpleBulkMemtable {
         let time_range = {
             let num_rows = self.num_rows.load(Ordering::Relaxed);
             if num_rows > 0 {
-                let ts_type = self
-                    .region_metadata
-                    .time_index_column()
-                    .column_schema
-                    .data_type
-                    .clone()
-                    .as_timestamp()
-                    .expect("Timestamp column must have timestamp type");
+                let ts_type = self.region_metadata.time_index_type();
                 let max_timestamp =
                     ts_type.create_timestamp(self.max_timestamp.load(Ordering::Relaxed));
                 let min_timestamp =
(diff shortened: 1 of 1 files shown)


src/mito2/src/memtable/bulk/part.rs
```diff
@@ -170,6 +170,28 @@ impl BulkPart {
...
 
+    /// Creates MemtableStats from this BulkPart.
+    pub fn to_memtable_stats(&self, region_metadata: &RegionMetadataRef) -> MemtableStats {
+        let ts_type = region_metadata
+            .time_index_column()
+            .column_schema
+            .data_type
+            .clone()
+            .as_timestamp()
+            .expect("Timestamp column must have timestamp type");
```
v0y4g3r:
```suggestion
         let ts_type = region_metadata.time_index_type();
```

src/mito2/src/memtable/simple_bulk_memtable.rs
```diff
@@ -243,6 +243,30 @@ impl Memtable for SimpleBulkMemtable {
         let sequence = options.sequence;
         let start_time = Instant::now();
         let projection = Arc::new(self.build_projection(projection));
+
+        // Use the memtable's overall time range and max sequence for all ranges
+        let max_sequence = self.max_sequence.load(Ordering::Relaxed);
+        let time_range = {
+            let num_rows = self.num_rows.load(Ordering::Relaxed);
+            if num_rows > 0 {
+                let ts_type = self
```
v0y4g3r: same here

v0y4g3r (approved):

PR #7489, opened 2025-12-26
feat: impl `json_get_string` with new json type

MichaelScofield:

and passed jsonbench q1 and q2

## PR Checklist

### DIFF
--- a/tests-integration/tests/jsonbench.rs
+++ b/tests-integration/tests/jsonbench.rs
@@ -62,13 +62,13 @@ SELECT \
     json_get_string(data, '$.commit.collection') AS event, count() AS count \
 FROM bluesky \
 GROUP BY event \
-ORDER BY event ASC, count DESC";
+ORDER BY count DESC, event ASC";
     let expected = r#"
 +-----------------------+-------+
 | event                 | count |
 +-----------------------+-------+
-| app.bsky.feed.post    | 3     |
 | app.bsky.feed.like    | 3     |
+| app.bsky.feed.post    | 3     |
 | app.bsky.graph.follow | 3     |
 | app.bsky.feed.repost  | 1     |
 +-----------------------+-------+"#;
@@ -85,13 +85,13 @@ WHERE \
     (json_get_string(data, '$.kind') = 'commit') AND \
     (json_get_string(data, '$.commit.operation') = 'create') \
 GROUP BY event \
-ORDER BY event ASC, count DESC";
+ORDER BY count DESC, event ASC";
     let expected = r#"
 +-----------------------+-------+-------+
 | event                 | count | users |
 +-----------------------+-------+-------+
-| app.bsky.feed.post    | 3     | 3     |
 | app.bsky.feed.like    | 3     | 3     |
+| app.bsky.feed.post    | 3     | 3     |
 | app.bsky.graph.follow | 3     | 3     |
 | app.bsky.feed.repost  | 1     | 1     |
 +-----------------------+-------+-------+"#;


src/common/function/src/scalars/json/json_get.rs
```diff
@@ -172,41 +175,211 @@ impl Function for JsonGetString {
...
+        let result = match arg0.data_type() {
+            DataType::Binary | DataType::LargeBinary | DataType::BinaryView => {
+                let arg0 = compute::cast(&arg0, &DataType::BinaryView)?;
+                let jsons = arg0.as_binary_view();
+                jsonb_get_string(jsons, paths)?
+            }
+            DataType::Struct(_) => {
+                let jsons = arg0.as_struct();
+                json_struct_get_string(jsons, paths)?
+            }
```
waynexia, line 201: Should we treat nonexistence or wrong type as NULL instead of error?

waynexia (approved):
rest LGTM

src/common/function/src/scalars/json/json_get.rs
sunng87: this wild card branch is data type error I think it should be good. postgres will also raise error when data type mismatch.

src/common/function/src/scalars/json/json_get.rs
```diff
@@ -172,41 +175,211 @@ impl Function for JsonGetString {
...
+
+    for i in 0..size {
+        if jsons.is_null(i) || paths.is_null(i) {
+            builder.append_null();
+            continue;
+        }
+        let path = paths.value(i);
+
+        // naively assume the JSON path is our kind of indexing to the field, by removing its "root"
+        let field_path = path.replace("$.", "");
```
sunng87: Do we have assumption `$.` will be at start of the path? It's better use something like `trim_start` if so

src/common/function/src/scalars/json/json_get.rs
```diff
@@ -172,41 +175,211 @@ impl Function for JsonGetString {
...
+            })?;
+            // the wanted field is not retrievable from the JSON struct columns directly, we have
+            // to combine everything (columns and the "_raw") into a complete JSON value to find it
+            let value = json_struct_to_value(raw, jsons, i)?;
+
+            match path.find(&value) {
+                Value::Null => builder.append_null(),
+                Value::Array(values) => match values.as_slice() {
+                    [] => builder.append_null(),
+                    [x] => {
```
sunng87, line 275: Any reference for this behaviour, treating single element array as its inner value

src/common/function/src/scalars/json/json_get.rs
MichaelScofield: nonexistence is treated as null when iterating the array

wrong type is better treated as error to fail fast (less surprise)

src/common/function/src/scalars/json/json_get.rs
MichaelScofield: json path must be started with "$."

I'm using `trim` in https://github.com/GreptimeTeam/greptimedb/pull/7489/commits/b104b692bf2e9e247e57f0c4e3c7bb2d97e4c269

src/common/function/src/scalars/json/json_get.rs
MichaelScofield: it's documented in the `JsonPath::find`: the returned value must be either "null" or a list of `Value`

MichaelScofield:
@sunng87 PTAL

sunng87 (approved):

PR #7491, opened 2025-12-29
fix: pipeline loading issue

shuiyisong:

When the user upgrade the database and meets these conditions:
1. Upgrading from v0.12 or before
2. Same-named pipeline under multiple schemas
3. Created a new schema and try to use the existing pipeline
The insertion would be failed because the pipeline engine don't know which pipeline to use.
By design, if the user creates a new pipeline with the same content and name, the pipeline would be inserted with an empty schema, and solve this case because the engine will always try to find the empty schema one.

However, the current implementation have an issue on loading the correct empty schema pipeline into the cache, causing the error to exist. This PR mainly fixes this issue by updating the loading cache logic.

## PR Checklist

### DIFF
--- a/src/pipeline/src/manager/table.rs
+++ b/src/pipeline/src/manager/table.rs
@@ -313,13 +313,10 @@ impl PipelineTable {
                         return self
                             .cache
                             .get_failover_cache(schema, name, input_version)?
-                            .ok_or(
-                                PipelineNotFoundSnafu {
-                                    name,
-                                    version: input_version,
-                                }
-                                .build(),
-                            );
+                            .context(PipelineNotFoundSnafu {
+                                name,
+                                version: input_version,
+                            });
                     }
                     _ => {
                         // if other error, we should return it


src/pipeline/src/manager/pipeline_cache.rs
```diff
@@ -33,24 +35,32 @@ const PIPELINES_CACHE_TTL: Duration = Duration::from_secs(10);
...
-    original_pipelines: Cache<String, (String, TimestampNanosecond)>,
+    original_pipelines: Cache<String, PipelineCacheValue>,
     /// If the pipeline table is invalid, we can use this cache to prevent failures when writing logs through the pipeline
     /// The failover cache never expires, but it will be updated when the pipelines cache is updated.
-    failover_cache: Cache<String, (String, TimestampNanosecond)>,
+    failover_cache: Cache<String, PipelineCacheValue>,
 }
 
+// pipeline cache content(str), pipeline version
+pub type PipelineCacheValue = (String, TimestampNanosecond);
```
sunng87: do we need to use a struct for this value and include schema cow?

shuiyisong: Introduce `PipelineContent` to include name, content, version and schema

src/pipeline/src/manager/table.rs
```diff
@@ -316,42 +330,40 @@ impl PipelineTable {
...
+            .or_else(|| pipeline_vec.iter().position(|v| v.schema == schema))
+            .map(|idx| pipeline_vec.remove(idx));
 
         // multiple pipeline with no empty or current schema
         // throw an error
-        let (pipeline_content, found_schema, version) =
-            pipeline.context(MultiPipelineWithDiffSchemaSnafu {
-                schemas: pipeline_vec.iter().map(|v| v.1.clone()).join(","),
-            })?;
+        let pipeline_content = pipeline.context(MultiPipelineWithDiffSchemaSnafu {
```
evenyag:
```suggestion
        let pipeline_content = pipeline.with_context(|| MultiPipelineWithDiffSchemaSnafu {
```

src/pipeline/src/manager/table.rs
```diff
@@ -304,8 +312,14 @@ impl PipelineTable {
...
-                            .get_failover_cache(schema, name, version)?
-                            .ok_or(PipelineNotFoundSnafu { name, version }.build());
+                            .get_failover_cache(schema, name, input_version)?
+                            .ok_or(
+                                PipelineNotFoundSnafu {
+                                    name,
+                                    version: input_version,
+                                }
+                                .build(),
+                            );
```
evenyag: Can we use context() to return the error?
```
.context(PipelineNotFoundSnafu { .. })?;
```

PR #7495, opened 2025-12-29
feat: impl `json_get_int` for new json type

MichaelScofield:

and test the last 3 jsonbench queries

## PR Checklist


src/common/function/src/scalars/json/json_get.rs
```diff
@@ -152,24 +145,73 @@ json_get!(
...
+    fn build(&mut self) -> ArrayRef;
+}
+
+struct JsonGet {
     signature: Signature,
 }
 
-impl JsonGetString {
-    pub const NAME: &'static str = "json_get_string";
+impl JsonGet {
```
sunng87, line 172: we can make this json_get a function, and let user to pass type as the third argument, we will try to coerce the get result into that type. this will also align with our goal to provide dynamic type.

MichaelScofield: will do in the next pr

shuiyisong (approved):

PR #7499, opened 2025-12-30
feat: export import v2 rfc

fengjiachun:

As the rfc said

## PR Checklist

### DIFF
--- a/docs/rfcs/2025-12-30-export-import-v2.md
+++ b/docs/rfcs/2025-12-30-export-import-v2.md
@@ -410,8 +410,7 @@ greptime export delete --snapshot s3://bucket/snapshots/old-snapshot
 # Unresolved Questions
 
 1. **Cross-version restore**: Should V2 support restoring to older GreptimeDB versions?
-2. **Schema evolution**: How to handle schema changes between export and import?
-3. **Partial schema export**: Should we support table-level filtering (not just schema-level)?
+2. **Partial schema export**: Should we support table-level filtering (not just schema-level)?
 
 # Future Possibilities
 


killme2008:
Looks like it's designed for GreptimeDB itself, and you can't export the data in other standard formats like csv, sql etc?

fengjiachun:
> Looks like it's designed for GreptimeDB itself, and you can't export the data in other standard formats like csv, sql etc?

Designed exclusively for GreptimeDB to replace the current `export/import` tool: https://docs.greptime.com/user-guide/deployments-administration/disaster-recovery/back-up-&-restore-data/

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,404 @@
...
+
+**Scenario 1**: Export to different paths ✅
+
+- Multiple exports can run simultaneously to different storage locations
+- No conflicts (only reads database, writes to different paths)
+
+**Scenario 2**: Export to same path ⚠️
+
+- Concurrent exports to the same path can corrupt the snapshot
+- Solution: Path existence check (reject if `manifest.json` exists)
```
evenyag: Maybe we can acquire a file lock.

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,404 @@
...
+**Chunk-level retry**:
+
+- Each chunk is an independent unit of work
+- Failed chunks are retried with exponential backoff (capped at 5 minutes)
+- Successful chunks are never re-exported
+
+**Resume capability**:
+
+- Manifest tracks chunk status (Pending, Completed, Failed)
+- `--resume` flag skips completed chunks and retries failed ones
```
evenyag: This should be true by default.

fengjiachun:
> Looks like it's designed for GreptimeDB itself, and you can't export the data in other standard formats like csv, sql etc?

We can support CSV because the dependent "copy database" already accepts it; I'll update the RFC.

fengjiachun:
PTAL @killme2008 @evenyag

src/query/src/dist_plan/commutativity.rs
waynexia, line 1: those typos are fixed in the latest main branch

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+1. **Cross-version restore**: Should V2 support restoring to older GreptimeDB versions?
+2. **Schema evolution**: How to handle schema changes between export and import?
+3. **Partial schema export**: Should we support table-level filtering (not just schema-level)?
+
+# Future Possibilities
+
+1. **Incremental backup**: Export only data changes since last backup (requires WAL integration)
+2. **Parallel chunk processing**: Export/import multiple chunks simultaneously (requires careful resource management)
+3. **Cloud-native snapshots**: Direct integration with S3 Glacier, Azure Archive for cold storage
+4. **Point-in-time recovery**: Combine full + incremental snapshots for PITR capability
```
waynexia: those two points seem to exceed the scope of an import/export tool. Become built-in db features should me more suitable

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+
+- Users unlikely to intentionally conflict
+- Path check prevents most accidents
+- Distributed lock adds significant complexity
+- YAGNI (You Aren't Gonna Need It)
+
+# Unresolved Questions
+
+1. **Cross-version restore**: Should V2 support restoring to older GreptimeDB versions?
+2. **Schema evolution**: How to handle schema changes between export and import?
```
waynexia: As long as the import command is implemented via the existing ingestion path like v1, this should not be a problem IMHO. But I haven't find a related states in this proposal about how import is going to be implemented.

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+## Why no distributed lock for concurrent exports?
+
+**Alternative**: Implement distributed lock via Meta service
+
+**Decision**: Path existence check only
+
+- Users unlikely to intentionally conflict
+- Path check prevents most accidents
+- Distributed lock adds significant complexity
+- YAGNI (You Aren't Gonna Need It)
```
waynexia: Both import/export work offline as a side path. So instead of YAGNI, I thought it's not even an option for this tool.

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+### Storage Types
+
+V2 supports two storage types:
+
+| Type               | Example                 | Use Case                 |
+| ------------------ | ----------------------- | ------------------------ |
+| **Remote Storage** | `s3://bucket/snapshots` | Production (recommended) |
+| **Server Path**    | `file:///data/backup`   | Local dev/testing        |
+
+**Important**: Local paths (e.g., `/tmp/export`, `./backup`) are **not supported** because schema export (CLI) and data export (server) run in different processes, which would split the snapshot across two machines.
```
waynexia, line 97: I believe this is a very necessary feature to export things to **client's path** or vise versa. The different processes problem looks like a tiny stumbling block. If we support export to another centralized storage like s3, the client's local path is conceptually the same thing.

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+├── manifest.json              # Snapshot metadata and chunk index
+├── schema/
+│   ├── schemas.json           # Schema definitions (JSON)
+│   ├── tables.json            # Table definitions (JSON)
+│   └── views.json             # View definitions (JSON)
+└── data/
+    ├── chunk_20240101_20240102/
+    │   ├── public.metrics.parquet
+    │   ├── public.logs.parquet
+    │   └── _chunk.meta
```
waynexia: What's this for? I only find one occurrence of that name here

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+- Immutable (content never changes after creation)
+- Verifiable (checksums at file, chunk, and snapshot levels)
+
+### Chunk
+
+A **chunk** is a time-range partition of data. Each chunk is independently exportable/importable and retryable.
+
+**Chunk properties**:
+
+- Time-aligned (e.g., [2024-01-01, 2024-01-02))
```
waynexia: If we have a dedicated metadata file to store chunks' information, we can try not requiring force alignment to provide more flexibility, to suit broader scenarios like large one-day volume deployment. We can track chunk's time range (or ranges in other axis, as discussed below). And change the chunk dir's naming to sequence like `1, 2, 3` or UUID to prepare for further non-timestamp-axis or shorter time granularity

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+
+- Multiple exports can run simultaneously to different storage locations
+- No conflicts (only reads database, writes to different paths)
+
+**Scenario 2**: Export to same path ⚠️
+
+- Concurrent exports to the same path can corrupt the snapshot
+- With default resume behavior: both processes will resume the same export (usually safe)
+- Race condition exists during initial manifest creation, but unlikely in practice
+- Recommendation: Include timestamp in snapshot path to avoid conflicts
```
waynexia, line 313: what about acquiring a file lock (or a group of file locks if exporting to multiple destinations) before starting?

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+
+**Why no distributed lock?**
+
+- Users unlikely to intentionally export to the same path
+- Path check prevents most accidents
+- Distributed lock adds complexity (Meta service dependency, timeouts, deadlocks)
```
waynexia: As above, this doesn't look like a sensible option for such an offline tool

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+  --to <LOCATION>                 Target storage location
+
+Optional:
+  --catalog <CATALOG>             Catalog name (default: greptime)
+  --schemas <SCHEMAS>             Comma-separated schema list (default: all)
+  --start-time <TIMESTAMP>        Time range start (default: earliest)
+  --end-time <TIMESTAMP>          Time range end (default: now)
+  --chunk-time-window <DURATION>  Chunk time window (default: 1d)
+  --parallelism <N>               Concurrency level (default: 1)
+  --format <FORMAT>               Export format: parquet (default) or csv
```
waynexia:
```suggestion
  --format <FORMAT>               Export format for data file: parquet (default) or csv
```

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+  - If snapshot doesn't exist: create new snapshot
+  - If snapshot exists: automatically resume export (skip completed chunks,
+    retry failed chunks, process pending chunks)
+  - If --force is specified: delete existing snapshot first, then create new one
+```
+
+### Import Command
+
+```
+greptime import restore [OPTIONS] --from <SNAPSHOT>
```
waynexia: is `restore` subcommand necessary here? Unlike `export`, `import` doesn't have other subcommands mentioned in this proposal

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+
+Key error types:
+
+- `UnsupportedLocalPath`: Bare path without URI scheme
+- `RemoteServerWithLocalPath`: file:/// used with remote server
+- `InvalidTimeRange`: start >= end
+- `ChecksumMismatch`: File/chunk/snapshot checksum verification failed
+- `SnapshotNotFound`: Resume requested but snapshot doesn't exist (import only)
+
+Errors include detailed hints for resolution.
```
waynexia:
```suggestion
```

This section looks verbose for an RFC. It's about implementation details

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+
+The manifest is a JSON file containing snapshot metadata and chunk index:
+
+**Key fields**:
+
+- `snapshot_id`: Unique identifier (UUID)
+- `catalog`, `schemas`: Catalog and schema list
+- `time_range`: Overall time range covered
+- `chunks[]`: Array of chunk metadata
+- `checksum`: Snapshot-level SHA256 checksum
```
waynexia, line 177: What about using Iceberg as manifest format for this tool. We may add a few extra information for features like resume and file lock etc on top of it

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+- Easy to debug and inspect
+- Simple hand-editing for small datasets
+
+**Use cases**: Data inspection, debugging, small-scale migration, third-party system integration
+
+**Limitations**:
+
+- Larger file size (no compression by default)
+- Type information loss (everything is text)
+- Less efficient for large datasets
```
waynexia: Are all formats supported by file-engine be used here? Looks like only a few things need to be done to be as such.

And by the way, Pros and Cons of each format should belong to the usage document. They look a bit verbose in a RFC here.

docs/rfcs/2025-12-30-export-import-v2.md
```diff
@@ -0,0 +1,457 @@
...
+
+V2 uses the existing `COPY DATABASE TO` SQL for data export:
+
+```sql
+COPY DATABASE <schema> TO '<path>' WITH (
+    START_TIME = '<start>',
+    END_TIME = '<end>',
+    FORMAT = 'parquet'  -- or 'csv'
+)
+```
```
waynexia: How will those new features be implemented if we are still based on the existing `COPY DATABASE TO` mechanism?

docs/rfcs/2025-12-30-export-import-v2.md
fengjiachun: The problem is that the COPY DATABASE statement does not support exporting data to a client-side/local directory.

docs/rfcs/2025-12-30-export-import-v2.md
fengjiachun: Outdated design — I'll remove it.

docs/rfcs/2025-12-30-export-import-v2.md
fengjiachun: Will remove this part

docs/rfcs/2025-12-30-export-import-v2.md
fengjiachun: will remove `restore`

docs/rfcs/2025-12-30-export-import-v2.md
fengjiachun: The Export/Import tool implements these functions on top of COPY DATABASE; each chunk invokes COPY DATABASE separately.

docs/rfcs/2025-12-30-export-import-v2.md
fengjiachun: The current design's default resume behavior has already resolved the main issues, so introducing a lock doesn't seem necessary for now.

docs/rfcs/2025-12-30-export-import-v2.md
fengjiachun: Yes, like v1, that's not a problem.

src/query/src/dist_plan/commutativity.rs
fengjiachun: rebased

docs/rfcs/2025-12-30-export-import-v2.md
fengjiachun: I wasn’t very familiar with Iceberg before, but after looking into it, I agree its snapshot and manifest model is relevant.

That said, Export/Import V2 is a tooling-level backup and restore mechanism, and adopting Iceberg would add additional complexity beyond our current needs. For now, we prefer a lightweight, purpose-built manifest, while keeping the design open to future integration.

docs/rfcs/2025-12-30-export-import-v2.md
sunng87: Maybe we can implement postgres COPY subprotocol which allows server to streaming byte streams to client

sunng87: Use postgres COPY subprotocol we may be able to use the TCP connection to streaming bytes from server to client. 

https://www.postgresql.org/docs/current/sql-copy.html

But the problem is it assumes the output is single file, so we may need to wrap the directory into a tarball or compressed format.

fengjiachun: One additional consideration is granularity: Postgres COPY is table/query-scoped, while Export/Import V2 targets a snapshot/database-level artifact (manifest + multiple chunks across schemas/tables).

Reusing the COPY subprotocol for streaming is an interesting direction, but it would likely require an extra snapshot-level protocol layer. This feels better suited as a future enhancement rather than something to fold into V2’s initial scope.

evenyag: Maybe we can use this in an export tool for tables?

evenyag (approved):

killme2008 (approved):

PR #7503, opened 2025-12-30
feat(http): improve error logging with client IP

### LINKED ISSUE #7328: HTTP handler error log improvement

User experience

Sometimes we may want to find out the reason for a non-200 HTTP response and we may not be able to get the error log from client side.

To collect the error logs in the HTTP handler, we added a log in our `IntoResponse` implementation for the Error.

https://github.com/GreptimeTeam/greptimedb/blob/69f0249039f46e1caf713636748d29a649ce9b65/src/servers/src/error.rs#L804-L816

https://github.com/GreptimeTeam/greptimedb/blob/69f0249039f46e1caf713636748d29a649ce9b65/src/servers/src/http/result/error_result.rs#L71-L90

It'd be better to have some logs for HTTP errors so we can debug later. We may need to check all `IntoResponse` implementations.

### Implementation challenges

Not all handler implementations convert errors into an `Error`. Then we won't be able to get the error message.

Case 1, the handler may returns an HTTPResponse directly:
https://github.com/GreptimeTeam/greptimedb/blob/69f0249039f46e1caf713636748d29a649ce9b65/src/servers/src/http/event.rs#L700-L708

Case 2, we can't catch the error raised by the axum itself. Unfortunately, the trace layer also can't get the error message and 400 is not considered a failure.
https://docs.rs/tower-http/latest/tower_http/trace/index.html#on_response
https://github.com/tokio-rs/axum/blob/main/examples/error-handling/src/main.rs

We can get the remote address, which may be useful in logging.
https://github.com/tokio-rs/axum/issues/43
https://docs.rs/axum/latest/axum/struct.Router.html#method.into_make_service_with_connect_info

maximk777:

Closes #7328

This PR improves HTTP error logging as requested in #7328.

1. ErrorResponse::from_error_message() now logs errors (error_result.rs)

> Previously, from_error_message() created error responses without logging, while from_error() did log. This meant handlers using from_error_message() directly bypassed the logging mechanism.
> Now both methods log consistently:
- error! level for server errors (where should_log_error() returns true)
- debug! level for client errors (e.g., InvalidArguments)

2. New middleware log_error_with_client_ip (client_ip.rs)

> Added middleware that logs all 4xx/5xx responses with client IP address.

3. Enabled ConnectInfo in HTTP server (http.rs)

> Changed into_make_service() to into_make_service_with_connect_info::<SocketAddr>() to make client IP available in request extensions.

## Limitations:

> Client IP may not be available in tests (middleware handles this gracefully by logging without IP)

## PR Checklist

   > Logging behavior was verified manually.
- [ ] This PR requires documentation updates. (Not required - internal changes only)


yihong0618:
I think we add a new function here, add a unittest for it would be better

maximk777:
Done! Added unit tests for the middleware.

src/servers/src/http/result/error_result.rs
```diff
@@ -32,17 +32,15 @@ pub struct ErrorResponse {
 impl ErrorResponse {
     pub fn from_error(error: impl ErrorExt) -> Self {
         let code = error.status_code();
-
-        if code.should_log_error() {
-            error!(error; "Failed to handle HTTP request");
-        } else {
-            debug!("Failed to handle HTTP request, err: {:?}", error);
-        }
```
evenyag, line 40: I'm afraid that we can't reuse from_error_message() here. We log the error in Debug format so we can get a more detailed message here. We may construct ErrorResponse directly and keep the log here.

src/servers/src/http/client_ip.rs
```diff
@@ -0,0 +1,103 @@
...
+        .map(|c| c.0);
+    let response = next.run(req).await;
+
+    if response.status().is_client_error() || response.status().is_server_error() {
+        if let Some(addr) = client_addr {
+            warn!(
+                "HTTP error response {} from client {}",
+                response.status(),
+                addr
+            );
```
evenyag: Can we also log the uri, method and matched path?
https://github.com/tokio-rs/axum/blob/051628c163b5d9f639dd88a8188f2df9b0552538/examples/error-handling/src/main.rs#L66-L75

src/servers/src/http/client_ip.rs
```diff
@@ -0,0 +1,103 @@
...
+        } else {
+            warn!("HTTP error response {}", response.status(),);
```
evenyag: Maybe we don't have to print the log if the client's address isn't available. We have metrics to collect the request count by status.

evenyag (commented):
I have also found one place missing the error log:
https://github.com/GreptimeTeam/greptimedb/blob/9343da7fe85c88f2402204eac4a61cdf17022438/src/servers/src/http/result/prometheus_resp.rs#L67-L76

Could you also add it for the `PrometheusJsonResponse`?

PR #7507, opened 2025-12-31
feat(meta-srv): fallback to non-TLS connection when etcd TLS prefer mode fail

WenyXu:

This PR fixes the etcd connection behavior when using TLS "prefer" mode. When TLS mode is set to Prefer, if the TLS connection to etcd fails, it will now automatically retry without TLS instead of failing immediately. This provides a more resilient connection strategy.

## PR Checklist

### DIFF
--- a/src/meta-srv/src/utils/etcd.rs
+++ b/src/meta-srv/src/utils/etcd.rs
@@ -40,47 +40,36 @@ pub async fn create_etcd_client_with_tls(
             client_options.keep_alive_timeout,
         );
 
-    let is_tls_prefer = matches!(tls_config.as_ref().map(|x| x.mode), Some(TlsMode::Prefer));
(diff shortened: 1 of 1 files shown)


src/meta-srv/src/utils/etcd.rs
```diff
@@ -38,16 +39,43 @@ pub async fn create_etcd_client_with_tls(
...
+    let client = Client::connect(&etcd_endpoints, Some(connect_options))
         .await
-        .context(error::ConnectEtcdSnafu)
+        .context(error::ConnectEtcdSnafu);
+
+    match client {
+        Ok(client) => Ok(client),
+        Err(e) => {
+            if is_tls_prefer {
+                error!(
```
fengjiachun: warn ?

WenyXu: resolved

MichaelScofield (approved):

PR #7512, opened 2026-01-04
fix: dist planner rm col req when rm sort

discord9:

as title,

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/commutativity.rs
+++ b/src/query/src/dist_plan/commutativity.rs
@@ -144,7 +144,7 @@ impl Categorizer {
                     }
                 }
                 // all group by expressions are partition columns can push down, unless
-                // another push down(including `Limit` or `Sort`) is already in progress(which will then prvent next cond commutative node from being push down).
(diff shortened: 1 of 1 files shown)


src/query/src/dist_plan/analyzer.rs
```diff
@@ -392,10 +392,11 @@ impl PlanRewriter {
                     && ext_b.node.name() == MergeSortLogicalPlan::name()
                 {
                     // revert last `ConditionalCommutative` result for Sort plan in this case.
-                    // `update_column_requirements` left unchanged because Sort won't generate
-                    // new columns or remove existing columns.
+                    // also need to remove any column requirements made by the Sort Plan
+                    // as it may refer to columns later no longer exist(rightfully) like by aggregate or projection
```
evenyag, line 396: Can we reproduce the issue in sqlness?

discord9: > Can we reproduce the issue in sqlness?

tql can't write that exact case it seems, also explained why this bug went undiscovered for so long

PR #7513, opened 2026-01-04
chore: add timestamp range to flight meta

v0y4g3r:

Add timestamp range to Flight DoPutMetadata to facilitate better ingestion performance.

## PR Checklist

### DIFF
--- a/src/common/grpc/src/flight/do_put.rs
+++ b/src/common/grpc/src/flight/do_put.rs
@@ -21,26 +21,26 @@ use crate::error::{Error, SerdeJsonSnafu};
 
 /// The metadata for "DoPut" requests and responses.
 ///
-/// Currently, there's only a "request_id", for coordinating requests and responses in the streams.
+/// Currently, there's a "request_id", for coordinating requests and responses in the streams.
 /// Client can set a unique request id in this metadata, and the server will return the same id in
 /// the corresponding response. In doing so, a client can know how to do with its pending requests.
 #[derive(Serialize, Deserialize)]
 pub struct DoPutMetadata {
     request_id: i64,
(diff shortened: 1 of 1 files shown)


src/common/grpc/src/flight/do_put.rs
```diff
@@ -27,16 +27,34 @@ use crate::error::{Error, SerdeJsonSnafu};
 #[derive(Serialize, Deserialize)]
 pub struct DoPutMetadata {
     request_id: i64,
+    /// Start timestamp of the batch in nanoseconds (optional, for time-windowed batches)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    start_timestamp: Option<i64>,
```
evenyag: Should we use Timestamp instead of nanosecond?

src/common/grpc/src/flight/do_put.rs
```diff
@@ -21,22 +21,40 @@ use crate::error::{Error, SerdeJsonSnafu};
...
 /// the corresponding response. In doing so, a client can know how to do with its pending requests.
 #[derive(Serialize, Deserialize)]
 pub struct DoPutMetadata {
     request_id: i64,
+    /// Start timestamp of the batch (optional, for time-windowed batches)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    start_timestamp: Option<i64>,
+    /// End timestamp of the batch (optional, for time-windowed batches)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    end_timestamp: Option<i64>,
```
evenyag: Is the end timestamp inclusive? Better to document that.

If it is inclusive, we should rename them to min/max timestamp.

PR #7519, opened 2026-01-05
refactor: remove `region_numbers` from `TableMeta` and `TableInfo`

WenyXu:

This PR removes non-orthogonal region information storage from `TableInfo` and `LogicalTableRouteValue`.

The region information was stored redundantly in multiple places:

1. **`region_numbers` in `TableMeta`/`TableInfo`**: This field was stored separately but should be derived from `TableRoute` which is the source of truth for region distribution.

2. **`region_ids` in `LogicalTableRouteValue`**: For logical tables (metric tables), the region information is already available in the physical table's route. Storing it again in logical table route creates data inconsistency during region operations (repartition split, merge).

Region information should always be retrieved from the physical table's `TableRoute`, which is updated atomically during region operations.

### Changes

1. **Remove `region_numbers` from `TableMeta` and `TableInfo`**
   - Remove `region_numbers` field from `TableMeta`, `TableMetaBuilder`, and `RawTableMeta`
   - Remove `region_ids()` method from `TableInfo`
   - Remove region_numbers synchronization logic in `create_table_metadata`

2. **Remove `region_ids` from `LogicalTableRouteValue`**
   - Simplify `LogicalTableRouteValue` to only store `physical_table_id`
   - Update `TableRouteValue::logical()` constructor
   - `region_numbers()` now returns empty vec for logical tables (callers should query physical table route)

3. **Derive region information from physical `TableRoute`**
   - Export `create_partitions_from_region_routes` function from `partition::manager`
   - Update `DistExtensionPlanner` to fetch region information from `physical_table_route`
   - Update `information_schema/tables.rs` to query region information from table routes

### Breaking Changes

- `TableMeta.region_numbers` field is removed
- `TableInfo.region_ids()` method is removed
- `LogicalTableRouteValue.region_ids` field is removed
- `TableRouteValue::logical()` no longer accepts `region_ids` parameter
- Code that relies on region information from `TableInfo` or `LogicalTableRouteValue` should now query from physical `TableRoute`

### ⚠️ Compatibility Notice: Removal of `TableMeta.region_numbers`
In the new version, the `TableMeta.region_numbers` field has been removed.
As older versions still depend on this field, downgrading from the new version to an older version may introduce compatibility issues. Specifically, an older version may fail or throw errors when reading data generated by the new version, since the region_numbers field is no longer present.

## PR Checklist

### DIFF
--- a/src/catalog/src/metrics.rs
+++ b/src/catalog/src/metrics.rs
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub(crate) const METRIC_DB_LABEL: &str = "db";
-
 use lazy_static::lazy_static;
 use prometheus::*;
 
@@ -25,7 +23,7 @@ lazy_static! {
     pub static ref METRIC_CATALOG_MANAGER_TABLE_COUNT: IntGaugeVec = register_int_gauge_vec!(
         "greptime_catalog_table_count",
         "catalog table count",
-        &[METRIC_DB_LABEL]
+        &["db"]
     )
     .unwrap();
     pub static ref METRIC_CATALOG_KV_REMOTE_GET: Histogram =


fengjiachun (approved):
Rest LGTM

PR #7527, opened 2026-01-06
feat: update special remote write label name

sunng87:

Because special labels starts with `__` will be removed by most agent as temporary label during relabelling. This patch adds new special label naming pattern for our remote write extension.

The new naming pattern follows how we extend http headers: `x_greptime_<label>`

## PR Checklist

### DIFF
--- a/src/servers/src/prom_store.rs
+++ b/src/servers/src/prom_store.rs
@@ -47,8 +47,10 @@ pub const DATABASE_LABEL_BYTES: &[u8] = b"x_greptime_database";
 pub const DATABASE_LABEL_ALT: &str = "__database__";
 pub const DATABASE_LABEL_ALT_BYTES: &[u8] = b"__database__";
 
-// deprecated
+/// deprecated, use DATABASE_LABEL instead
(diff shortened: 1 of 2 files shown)


src/servers/src/prom_store.rs
```diff
@@ -77,20 +110,12 @@ pub fn table_name(q: &Query) -> Result<String> {
...
-        }
-    }
-
-    // If no __schema__ found, look for __database__
-    for query in &request.queries {
-        for matcher in &query.matchers {
-            if matcher.name == DATABASE_LABEL && matcher.r#type == MatcherType::Eq as i32 {
+            if is_database_selection_label(&matcher.name)
+                && matcher.r#type == MatcherType::Eq as i32
+            {
```
sunng87, line 123: Here I removed this prioritized check for `__schema__` as I suggest to deprecate support for `__schema__` in favor of `__database__`. In almost all contexts, we don't expose `schema` as a user-facing concept.

waynexia (approved):

PR #7528, opened 2026-01-07
feat: impl vector index scan in storage

### LINKED ISSUE #7505: Workflow run 'GreptimeDB Nightly Build' failed
@GreptimeTeam/db-approver
New failure: https://github.com/GreptimeTeam/greptimedb/actions/runs/20608784979

killme2008:

Fixed #7505

Implements the read path for vector index - applies HNSW index during SST scanning for KNN queries.

- `VectorIndexApplier`: Loads vector index blob, executes KNN search, maps HNSW keys back to row offsets (handling nulls via binary search)
- `VectorIndexCache`: Caches deserialized HNSW indexes in memory (moka-based, weighted by size)
- Row group pruning: Integrates with ParquetReaderBuilder to prune row groups based on vector search results
- Over-fetching: When filters exist, k *= 2 to compensate for filtered rows
- Refactor encoding and decoding of blob with `VectorIndexBlobHeader`

 ### Architecture

```
  ┌─────────────────┐     ┌──────────────────────┐     ┌─────────────────┐
  │  ScanRegion     │────▶│ VectorIndexApplier   │────▶│ VectorIndexCache│
  │  (build applier)│     │ (apply KNN search)   │     │ (cache HNSW)    │
  └─────────────────┘     └──────────────────────┘     └─────────────────┘
                                    │
                                    ▼
                          ┌──────────────────────┐
                          │ ParquetReaderBuilder │
                          │ (row group pruning)  │
                          └──────────────────────┘
```

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/index/vector_index/applier.rs
+++ b/src/mito2/src/sst/index/vector_index/applier.rs
@@ -442,13 +442,6 @@ fn hnsw_key_to_row_offset(
         }
     }
 
-    if left >= total_rows {
-        return ApplyVectorIndexSnafu {
-            reason: "Failed to map HNSW key to row offset".to_string(),
-        }
-        .fail();
-    }
-
     Ok(left)
 }
 


src/mito2/src/read/scan_region.rs
```diff
@@ -964,15 +1036,22 @@ impl ScanInput {
...
             .cache(self.cache_strategy.clone())
             .inverted_index_appliers(self.inverted_index_appliers.clone())
             .bloom_filter_index_appliers(self.bloom_filter_index_appliers.clone())
-            .fulltext_index_appliers(self.fulltext_index_appliers.clone())
+            .fulltext_index_appliers(self.fulltext_index_appliers.clone());
+        #[cfg(feature = "vector_index")]
+        {
+            reader =
+                reader.vector_index_applier(self.vector_index_applier.clone(), self.vector_index_k);
+        }
```
evenyag: Can we use this way to avoid the `#[cfg_attr(not(feature = "vector_index"), allow(unused_mut))]` above?

```suggestion
        #[cfg(feature = "vector_index")]
        {
            let mut reader = reader;
            reader =
                reader.vector_index_applier(self.vector_index_applier.clone(), self.vector_index_k);
        }
```

waynexia:
@codex review

src/mito2/src/read/scan_util.rs
```diff
@@ -630,6 +646,9 @@ impl ScanMetricsSet {
         READ_ROW_GROUPS_TOTAL
             .with_label_values(&["bloom_filter_index_filtered"])
             .inc_by(self.rg_bloom_filtered as u64);
+        READ_ROW_GROUPS_TOTAL
+            .with_label_values(&["vector_index_filtered"])
+            .inc_by(self.rg_vector_filtered as u64);
```
evenyag, line 652: Only do that when rg_vector_filtered > 0? So we won't see the vector metrics when the vector index is disabled.

src/mito2/src/sst/index/vector_index/applier.rs
```diff
@@ -0,0 +1,650 @@
...
+            reason: format!("Total rows {} exceeds u32::MAX", total_rows),
+        }
+        .build()
+    })?;
+
+    let mut row_offsets = Vec::with_capacity(keys.len());
+    for key in keys {
+        let offset = hnsw_key_to_row_offset(null_bitmap, total_rows_u32, indexed_rows, key)?;
+        row_offsets.push(offset as u64);
+    }
```
evenyag, line 402: Why don't we use the key as the row offset?

src/mito2/src/sst/index/vector_index/format.rs
```diff
@@ -0,0 +1,324 @@
...
+    pub engine_type: VectorIndexEngineType,
+    pub dim: u32,
+    pub metric: VectorDistanceMetric,
+    pub connectivity: u16,
+    pub expansion_add: u16,
+    pub expansion_search: u16,
+    pub total_rows: u64,
+    pub indexed_rows: u64,
+    pub null_bitmap_len: u32,
+}
```
evenyag, line 39: Why not use protobuf like other indexes?
You can define it in the index crate if you don't want to put it in greptime-proto.
https://github.com/GreptimeTeam/greptimedb/blob/main/src/log-store/proto/logstore.proto

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -1073,6 +1190,9 @@ impl ReaderFilterMetrics {
         READ_ROWS_IN_ROW_GROUP_TOTAL
             .with_label_values(&["bloom_filter_index_filtered"])
             .inc_by(self.rows_bloom_filtered as u64);
+        READ_ROWS_IN_ROW_GROUP_TOTAL
+            .with_label_values(&["vector_index_filtered"])
+            .inc_by(self.rows_vector_filtered as u64);
```
evenyag, line 1195: Only report when rg_vector_filtered > 0.

src/mito2/src/read/scan_region.rs
killme2008: Good point! Let me try it.

src/mito2/src/sst/index/vector_index/format.rs
killme2008: Good point. I don't see that before.

evenyag (approved):
I think this is an experimental feature in a very early stage. It's fine to merge under a feature gate.

src/mito2/src/sst/index/vector_index/applier.rs
killme2008: We could choose to store the original row offset as the key, but the current design doesn’t. The index is built over indexed_rows (a filtered subset), and the build path can reorder or compact entries. Using a
  dense per‑index ID is simpler and more space‑efficient for sparse data, but it means we must map back to the original row offsets at read time. If we wanted key = row_offset, we’d need to guarantee no
  reordering, preserve sparse keys for filtered rows, and handle nulls consistently—i.e., a different build contract rather than a usearch limitation.

killme2008: key is the HNSW internal entry/node ID, not the table row offset. The index is built over indexed_rows (a subset) and can be reordered/dense; we also need null_bitmap to map back to the real row positions.

src/mito2/src/sst/index/vector_index/format.rs
killme2008: Let's leave this to another PR. I don't want to modify greptime-proto in this PR.

src/mito2/src/sst/parquet/reader.rs
killme2008: Use  #[cfg(feature = "vector_index")]

waynexia:
@codex review

src/mito2/src/sst/index/vector_index/applier.rs
```diff
@@ -0,0 +1,650 @@
...
+            left = mid + 1;
+        }
+    }
+
+    if left >= total_rows {
+        return ApplyVectorIndexSnafu {
+            reason: "Failed to map HNSW key to row offset".to_string(),
+        }
+        .fail();
+    }
```
v0y4g3r: nit: Looks like this check is unnecessary because `left` always falls in `[0, total_rows-1]`

src/mito2/src/sst/index/vector_index/format.rs
```diff
@@ -0,0 +1,324 @@
...
+        if version != VECTOR_INDEX_BLOB_VERSION {
+            return Err(VectorIndexBlobFormatError::UnsupportedVersion(version));
+        }
+
+        let engine_type = VectorIndexEngineType::try_from_u8(read_u8(data, &mut offset)?)
+            .ok_or_else(|| VectorIndexBlobFormatError::UnknownEngine(data[offset - 1]))?;
+        let dim = read_u32(data, &mut offset)?;
+        let metric = VectorDistanceMetric::try_from_u8(read_u8(data, &mut offset)?)
+            .ok_or_else(|| VectorIndexBlobFormatError::UnknownMetric(data[offset - 1]))?;
+        let connectivity = read_u16(data, &mut offset)?;
```
v0y4g3r, line 108: We may resort to other sophisticated crates to address the serialization/deserialization.

v0y4g3r (approved):
Rest LGTM

src/mito2/src/sst/index/vector_index/format.rs
killme2008: Will replace them with proto.

PR #7531, opened 2026-01-07
fix: precise filter time index if not in projection

evenyag:

Add a test case that time index is not in the projection but in the filter.

This corrects the behavior of precise filters.

## PR Checklist

### DIFF
--- a/tests/cases/standalone/common/select/prune.result
+++ b/tests/cases/standalone/common/select/prune.result
@@ -128,3 +128,68 @@ drop table demo;
 
 Affected Rows: 0
 
+CREATE TABLE test_time_filter(
+    host STRING,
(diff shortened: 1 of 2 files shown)


src/mito2/src/engine/projection_test.rs
```diff
@@ -100,3 +123,96 @@ async fn test_scan_projection_with_format(flat_format: bool) {
 +-------+---------+---------------------+";
     assert_eq!(expected, batches.pretty_print().unwrap());
 }
+
+#[tokio::test]
+async fn test_scan_projection_without_primary_key() {
```
evenyag, line 128: Better to add a sqlness test.

PR #7545, opened 2026-01-08
feat: use dyn filter

discord9:

### Performance(x700 speedup)
before:
```sql
--------------
SELECT start_time, end_time, run_type, status FROM langchain_traces ORDER BY end_time DESC LIMIT 10
--------------

+---------------------+----------------------------+----------+--------+
| start_time          | end_time                   | run_type | status |
+---------------------+----------------------------+----------+--------+
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
+---------------------+----------------------------+----------+--------+
10 rows in set (28.86 sec)
```
after:
```sql
--------------
SELECT start_time, end_time, run_type, status FROM langchain_traces ORDER BY end_time DESC LIMIT 10
--------------

+---------------------+----------------------------+----------+--------+
| start_time          | end_time                   | run_type | status |
+---------------------+----------------------------+----------+--------+
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
| 2025-07-27 23:59:59 | 2025-07-27 23:59:59.999000 | NULL     | NULL   |
+---------------------+----------------------------+----------+--------+
10 rows in set (0.21 sec)
```
Here is the comparison table in English, based on the analysis of the `topk_dyn_filter_test_*.result` files.

### Performance Comparison: `ORDER BY end_time DESC LIMIT 10`
*This query targets the **non-indexed** timestamp column (`end_time`), where the optimization is most prominent.*

| Metric | `main` Branch (Avg) | `dyn_filter` Branch (Avg) | Speedup |
| :--- | :--- | :--- | :--- |
| **Total Query Time (User Time)** | **28.70 s** | **0.20 s** | **~143x** |
| **Scan Node Cost (Total)** | 28.81 s | 0.20 s | **~144x** |
| **Sort Exec Compute Time** | 6.55 s | 0.009 s | **~720x** |
| **Scan Rows before Filter** | High (Full Scan) | Near 0 (Pruned) | **Significant** |

---

### Comparison by Column Type (Execution Time)
This table highlights why the Dynamic Filter is critical for non-indexed columns but redundant for the Time Index.

| Query Scenario | `main` (No Dyn Filter) | `dyn_filter` (With Dyn Filter) | Optimization Effect |
| :--- | :--- | :--- | :--- |
| **`ORDER BY start_time`** (Time Index) | ~0.20 s | ~0.20 s | **Native Index Pruning** (No change) |
| **`ORDER BY end_time`** (Simple Timestamp) | **~28.80 s** | **0.20 s** | **Dynamic Filter Pruning** (~140x) |

---

### Key Observations from Scan Nodes:
*   **For `start_time` (Index):** The Scan node shows `dyn_filters: ["DynamicFilter [ empty ]"]`. The system relies on existing SST time-range metadata to early-stop.
*   **For `end_time` (Non-index):** The Scan node shows active predicates like `DynamicFilter [ end_time > <threshold> ]`. This threshold is dynamically updated by the TopK operator and sent back to the Scan node to prune irrelevant SST files, eliminating massive IO wait times (`poll_duration`).
### What's changed

- Added dynamic filter pushdown plumbing from DataFusion to table scan, so runtime filters from parent operators can be propagated into region scanners.
- Extended the `RegionScanner` interface with `update_predicate_with_dyn_filter(...)` and implemented it in Mito scan paths (`seq_scan`, `series_scan`, `unordered_scan`) to apply pushed filters at scan time.
- Updated `RegionScanExec::handle_child_pushdown_result` to consume parent pushdown filters and report per-filter support back to DataFusion.
- Refactored predicate dynamic-filter storage to runtime-updatable state (`ArcSwap`) so filters can be safely refreshed while scans are running.
- Ensured dynamic filters are installed even when there is no pre-existing static predicate (important for plans like plain TopK/HashJoin-driven pruning).
- Added/updated sqlness coverage for dynamic-filter scenarios (`hash_join_dyn_filter`, `hash_join_topk_dyn_filter`, `topk_dyn_filter`) and refreshed expected explain/vector outputs accordingly.

### Why this change

The main intention is to make runtime pruning actually effective in Greptime's table scan layer, not only in upper DataFusion operators.

Before this change, runtime filters produced by operators like HashJoin/TopK were not fully reflected in region scans, which could still read unnecessary row groups/data blocks. By wiring these filters into scanner predicates, we can cut unnecessary reads earlier, reduce IO/CPU, and improve end-to-end query efficiency.

This also establishes a clearer extension point for future runtime pruning improvements by standardizing how scanners accept and acknowledge dynamic filters.
### Limitations
For now this optimization still only applies to topk query, to make it also works on join, need to add sort of a 
remote dyn filter which can update scan operator's dyn filter on datanode from join operator on frontend
### Alternative
For **TopK queries ordered by the time index**, the benchmark strongly suggests that **`PartSortExec` should not be combined with dynamic-filter pushdown**. On `SELECT * FROM langchain_traces ORDER BY start_time DESC LIMIT 10`, the `PartSortExec` path **without** dynamic filter  finished in **0.26s**, while `PartSortExec` **with** dynamic filter slowed down to **0.44s**. By contrast, the version that used **dynamic filter only** and fell back to plain `SortExec: TopK` ran in **0.29s**, which shows that the dynamic filter itself is not inherently bad. The problem is specifically the **combination** of `PartSortExec` and dynamic filter: `PartSortExec` already performs pruning based on partition ranges and time ordering, so adding dynamic-filter pushdown appears to introduce overlapping optimization and perturbs scan behavior enough to make the query slower overall. Therefore, the most reasonable conclusion is: **keep dynamic-filter pushdown for normal `SortExec: TopK`, keep `PartSortExec` for time-index-aware sorting in general, but disable dynamic-filter pushdown on the `PartSortExec` TopK path.**

## PR Checklist

### DIFF
--- a/src/mito2/src/read/scan_region.rs
+++ b/src/mito2/src/read/scan_region.rs
@@ -1619,8 +1619,7 @@ impl StreamContext {
         let filter_expr = filter_exprs
             .into_iter()
             .filter_map(|expr| {
-                let expr: Arc<dyn std::any::Any + Send + Sync + 'static> = expr;
-                if let Ok(dyn_filter) = expr
+                if let Ok(dyn_filter) = (expr as Arc<dyn std::any::Any + Send + Sync + 'static>)
                 .downcast::<datafusion::physical_plan::expressions::DynamicFilterPhysicalExpr>()
             {
                 supported.push(true);
--- a/src/query/src/query_engine/state.rs
+++ b/src/query/src/query_engine/state.rs
@@ -181,7 +181,8 @@ impl QueryEngineState {
         physical_optimizer
             .rules
             .push(Arc::new(WindowedSortPhysicalRule));
-        // explicitly not do filter pushdown for windowed sort&part sort as
+        // explicitly not do filter pushdown for windowed sort&part sort
+        // (notice that `PartSortExec` create another new dyn filter that need to be pushdown if want to use dyn filter optimization)
         // benchmark shows it can cause performance regression due to useless filtering and extra shuffle.
         // We can add a rule to do filter pushdown for windowed sort in the future if we find a way to avoid the performance regression.
         physical_optimizer
--- a/src/table/src/predicate.rs
+++ b/src/table/src/predicate.rs
@@ -175,7 +175,7 @@ impl Predicate {
                 },
                 Err(e) => {
                     // since dynamic filter exprs could be complex, it's possible that `PruningPredicate::try_new` fails to prove anything from it. In that case, we just log it and skip pruning with this expr.
-                    debug!("Failed to create predicate for expr: {e:?}");
+                    debug!("Failed to create pruning predicate for expr: {e:?}");
                 }
             }
         }
--- a/src/table/src/table/scan.rs
+++ b/src/table/src/table/scan.rs
@@ -445,8 +445,6 @@ impl ExecutionPlan for RegionScanExec {
         child_pushdown_result: ChildPushdownResult,
         _config: &datafusion::config::ConfigOptions,
     ) -> DfResult<FilterPushdownPropagation<Arc<dyn ExecutionPlan>>> {
-        // TODO(discord9): use the pushdown result to update the scanner's predicate
-
         let parent_filters = child_pushdown_result
             .parent_filters
             .into_iter()


evenyag:
@codex review

src/table/src/predicate.rs
```diff
@@ -53,14 +54,14 @@ macro_rules! return_none_if_utf8 {
...
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub struct Predicate {
     /// logical exprs
     exprs: Arc<Vec<Expr>>,
     /// dynamic filter physical exprs, only useful if dynamic filtering is enabled
     ///
     /// They are usually from `TopK` or `Join` operators, and can dynamically filter data during query execution by using current runtime information to further reduce data scanning
-    dyn_filters: Arc<Vec<DynamicFilterPhysicalExpr>>,
+    dyn_filters: Arc<ArcSwap<Vec<Arc<DynamicFilterPhysicalExpr>>>>,
```
evenyag, line 64: Is the outer Arc duplicated with the ArcSwap?

discord9: > Is the outer Arc duplicated with the ArcSwap?

The outer `Arc` is for making `Predicate` cloneable since `ArcSwap` is not clone-able(in previous version it can be clone, but will result in two separate `ArcSwap` that doesn't share data)

MichaelScofield:
EPIC!

src/mito2/src/read/seq_scan.rs
```diff
@@ -22,7 +22,7 @@ use async_stream::try_stream;
 use common_error::ext::BoxedError;
 use common_recordbatch::util::ChainedRecordBatchStream;
 use common_recordbatch::{RecordBatchStreamWrapper, SendableRecordBatchStream};
-use common_telemetry::tracing;
+use common_telemetry::{tracing, warn};
```
MichaelScofield: Seems no "warn" is used, how's clippy not complaining?

waynexia:
EPIC!

src/mito2/src/read/scan_region.rs
```diff
@@ -1598,6 +1608,32 @@ impl StreamContext {
...
+    /// Safe after stream creation; in-flight reads may still observe an older snapshot.
+    pub(crate) fn update_predicate_with_dyn_filter(
+        self: &Arc<Self>,
+        filter_exprs: Vec<Arc<dyn datafusion::physical_plan::PhysicalExpr>>,
+    ) -> Vec<bool> {
+        let mut supported = Vec::with_capacity(filter_exprs.len());
+        let filter_expr = filter_exprs
+            .into_iter()
+            .filter_map(|expr| {
+                let expr: Arc<dyn std::any::Any + Send + Sync + 'static> = expr;
```
evenyag: Is this line necessary?

src/mito2/src/read/scan_region.rs
```diff
@@ -1676,12 +1704,26 @@ impl PredicateGroup {
...
+            None
+        } else {
+            Some(&self.predicate_without_region)
+        }
+    }
+
+    /// Updates dynamic filters in the predicates.
+    pub(crate) fn update_dyn_filters(&self, dyn_filters: Vec<Arc<DynamicFilterPhysicalExpr>>) {
+        self.predicate_all.add_dyn_filters(dyn_filters.clone());
+        self.predicate_without_region.add_dyn_filters(dyn_filters);
```
evenyag, line 1725: Looks like add_dyn_filters() extends the filters. Will it be an issue if we call update_dyn_filters multiple times? They have different semantics.

src/mito2/src/read/scan_region.rs
discord9: > Is this line necessary?

as_any doesn't work since here need a Arc<dyn Any> not &dyn Any so need this

src/mito2/src/read/scan_region.rs
discord9: > Looks like add_dyn_filters() extends the filters. Will it be an issue if we call update_dyn_filters multiple times? They have different semantics.

They are only called in physical execution plan optimzer phase in `FilterPushdown(Post)`

src/mito2/src/read/scan_region.rs
```diff
@@ -1598,6 +1608,32 @@ impl StreamContext {
...
+    /// Safe after stream creation; in-flight reads may still observe an older snapshot.
+    pub(crate) fn add_dyn_filter_to_predicate(
+        self: &Arc<Self>,
+        filter_exprs: Vec<Arc<dyn datafusion::physical_plan::PhysicalExpr>>,
+    ) -> Vec<bool> {
+        let mut supported = Vec::with_capacity(filter_exprs.len());
+        let filter_expr = filter_exprs
+            .into_iter()
+            .filter_map(|expr| {
+                let expr: Arc<dyn std::any::Any + Send + Sync + 'static> = expr;
```
killme2008: Move the `Arc<dyn std::any::Any + Send + Sync + 'static>` into `| expr |` above?

src/query/src/query_engine/state.rs
```diff
@@ -181,6 +181,9 @@ impl QueryEngineState {
         physical_optimizer
             .rules
             .push(Arc::new(WindowedSortPhysicalRule));
+        // explicitly not do filter pushdown for windowed sort&part sort as
```
killme2008: Is the comment correctly placed?

src/table/src/table/scan.rs
```diff
@@ -445,7 +446,28 @@ impl ExecutionPlan for RegionScanExec {
         _config: &datafusion::config::ConfigOptions,
     ) -> DfResult<FilterPushdownPropagation<Arc<dyn ExecutionPlan>>> {
         // TODO(discord9): use the pushdown result to update the scanner's predicate
-        Ok(FilterPushdownPropagation::if_all(child_pushdown_result))
+
```
killme2008: Looks the above `TODO` could be removed.

src/table/src/predicate.rs
```diff
@@ -157,7 +174,8 @@ impl Predicate {
                     }
                 },
                 Err(e) => {
-                    error!(e; "Failed to create predicate for expr");
+                    // since dynamic filter exprs could be complex, it's possible that `PruningPredicate::try_new` fails to prove anything from it. In that case, we just log it and skip pruning with this expr.
+                    debug!("Failed to create predicate for expr: {e:?}");
```
killme2008: Not sure, but looks `warning` is better.

discord9: > Not sure, but looks `warning` is better.

complext predicate will always fail to create pruning predicate as it only support very simple filters so debug

src/query/src/query_engine/state.rs
discord9: > Is the comment correctly placed?

Correct, as since windowed sort replace `SortExec` with `PartSortExec` after `FilterPushdown(Post)` is done(which collect dyn filtes from old `SortExec` and put it to table scan) so for `PartSortExec` if one need to get its' dyn filter to table scan, had to re-run `FilterPushdown(Post)` again after windowed sort optimze rule, so the comment here is explain why not run `FilterPushdown(Post)` here again

src/mito2/src/read/scan_region.rs
discord9: > Move the `Arc<dyn std::any::Any + Send + Sync + 'static>` into `| expr |` above?

wrong closure type

killme2008:
@discord9 there is a sqlness failure.

PR #7546, opened 2026-01-08
chore: fix sqlness format error

Standing-Man:

As titled.

## PR Checklist

### DIFF
--- a/tests/cases/standalone/common/select/tz_encode.result
+++ b/tests/cases/standalone/common/select/tz_encode.result
@@ -30,9 +30,9 @@ SELECT
 FROM
     ngx_access_log;
 
-+--------------------------------+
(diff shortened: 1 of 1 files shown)


tests/cases/standalone/common/select/tz_encode.result
```diff
@@ -30,9 +30,9 @@ SELECT
 FROM
     ngx_access_log;
 
-+--------------------------------+
-| now()                          |
-+--------------------------------+
++-----------------------------+
+| now()                       |
++-----------------------------+
```
evenyag, line 35: This may be caused by the timestamp precision. We should find a better way to avoid the different lengths of the formatted timestamp or remove the whitespaces. @discord9

PR #7549, opened 2026-01-08
feat: flush region before close when skip-wal is enabled

v0y4g3r:

When closing a region with Noop WAL provider, the region is now flushed before closing to ensure data durability. This prevents data loss for regions configured with skip_wal.

Changes:
- Add `Closing` variant to `FlushReason` enum
- Modify `handle_close_request` to trigger flush for Noop WAL regions
- Pass flush reason through the flush pipeline
- Add test to verify data persistence after close with skip-wal

The flush-on-close flow completes the region cleanup after the flush finishes, ensuring the region is properly removed from all schedulers.

## PR Checklist

### DIFF
--- a/src/mito2/src/engine/skip_wal_test.rs
+++ b/src/mito2/src/engine/skip_wal_test.rs
@@ -135,13 +135,6 @@ async fn test_close_follower_region_skip_wal() {
         .unwrap();
     assert!(region.is_follower());
 
-    // Convert back to Leader before close, because Followers cannot update
-    // the manifest (which is required for flush to complete).
-    engine
-        .set_region_role(region_id, RegionRole::Leader)
-        .unwrap();
-    assert!(!region.is_follower());
-
     // Close the region. This should trigger a flush.
     engine
         .handle_request(region_id, RegionRequest::Close(RegionCloseRequest {}))


src/mito2/src/worker/handle_flush.rs
```diff
@@ -316,6 +320,10 @@ impl<S: LogStore> RegionWorkerLoop<S> {
         self.schedule_compaction(&region).await;
 
         self.listener.on_flush_success(region_id);
+        if flush_on_close {
+            self.remove_region(region_id).await;
+            info!("Region {} closed after flush", region_id);
```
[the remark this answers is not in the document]
v0y4g3r: fixed in 70fb5ea835

src/mito2/src/worker/handle_close.rs
```diff
@@ -15,36 +15,68 @@
...
+        if region.provider == Provider::Noop
+            && !region
+                .version_control
+                .current()
+                .version
+                .memtables
+                .is_empty()
+        {
+            info!("Region {} has pending data, waiting for flush", region_id);
+            self.handle_flush_request(
```
killme2008, line 51: Should we add a metric for this operation to facilitate performance troubleshooting?

src/mito2/src/worker/handle_flush.rs
```diff
@@ -295,26 +299,30 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
-            self.flush_scheduler.on_flush_success(region_id)
-        {
-            // Perform DDLs first because they require empty memtables.
-            self.handle_ddl_requests(&mut ddl_requests).await;
-            // Handle pending write requests, we don't stall these requests.
-            self.handle_write_requests(&mut write_requests, &mut bulk_writes, false)
-                .await;
+        if flush_on_close {
+            // Remove region from server for flush on closing,
+            // no need to handle requests and schedule compactions.
```
killme2008, line 304: I understand it doesn’t need to schedule compaction here, but why not handle pending requests?

v0y4g3r: > I understand it doesn’t need to schedule compaction here, but why not handle pending requests?

If we continue to handle pending write requests, then the memtable will have data that needs to be flushed again.

src/mito2/src/worker/handle_close.rs
v0y4g3r: > Should we add a metric for this operation to facilitate performance troubleshooting?

Flush task will print a log with flush reason = Closing that reflects region id, total rows, cost time etc.

https://github.com/GreptimeTeam/greptimedb/blob/7e97b50b5e6d583e10d802bc790c5d4018bf7176/src/mito2/src/flush.rs#L385-L395

src/mito2/src/worker/handle_flush.rs
killme2008: So in this situation, will the pending write requests fail?

v0y4g3r: > So in this situation, will the pending write requests fail?

Yes, pending writes will not be acked.

src/mito2/src/engine/skip_wal_test.rs
WenyXu, line 1: Add a test case to verify flushing a readonly region fails.

src/mito2/src/worker/handle_close.rs
```diff
@@ -15,36 +15,68 @@
...
-            return Ok(0);
+            sender.send(Ok(0));
+            return;
         };
 
         info!("Try to close region {}, worker: {}", region_id, self.id);
 
+        // If the region is using Noop WAL and has data in memtable,
+        // we should flush it before closing to ensure durability.
+        if region.provider == Provider::Noop
```
WenyXu, line 42: I think we should handle cases where the flush operation fails. e.g., a readonly region cannot be flushed.

WenyXu (commented):
BTW, in a pod env, k8s only sends `SIGTERM` during shutdown. If the process takes too long to exit, the pod may be forcefully killed, and a long region flush could still cause data loss. We might need to handle `SIGTERM` properly and consider adjusting `terminationGracePeriodSeconds` if the flush operation can be slow. cc @daviderli614

src/mito2/src/worker/handle_close.rs
v0y4g3r: Flushing unflushable regions (region not in Writable/Staging/Degrading state) should succeed immediately. This was fixed in 9f3cb5f153561a2092331ce73ce1389d34105f31

src/mito2/src/engine/skip_wal_test.rs
v0y4g3r: fixed in 9f3cb5f153561a2092331ce73ce1389d34105f31

WenyXu:
@codex review

src/mito2/src/engine/skip_wal_test.rs
```diff
@@ -105,3 +105,73 @@ async fn test_close_region_skip_wal(insert: bool) {
...
+        .await
+        .unwrap();
+
+    let region = engine.get_region(region_id).unwrap();
+
+    // Set the region to Follower state.
+    engine
+        .set_region_role(region_id, RegionRole::Follower)
+        .unwrap();
+    assert!(region.is_follower());
```
WenyXu, line 136: Should we test closing the follower region?

WenyXu (approved):

PR #7550, opened 2026-01-09
feat: Allow ALTER TABLE to modify the skip_wal option dynamically

OmCheeLin:

This PR adds support for dynamically modifying the `skip_wal` table option via `ALTER TABLE SET/UNSET` syntax, allowing users to toggle WAL on or off at runtime without recreating tables.

changes:
- **SQL Syntax**: `ALTER TABLE <table> SET 'skip_wal'='true'/'false'` and `ALTER TABLE <table> UNSET 'skip_wal'`
- **Region-level**: When `skip_wal` is altered, the change is immediately applied to all regions in memory via `handle_alter_region_options_fast()`, setting `wal_options` to `Noop` (disabled) or `RaftEngine` (enabled).
- **Metadata persistence**: The change is persisted in both `TableInfo` and `DatanodeTableValue` entries in metasrv, ensuring consistency across restarts.
- **SHOW CREATE TABLE**: Updated to display the `skip_wal` option in the `WITH` clause.

Limitations
- In distributed mode, when enabling WAL (`skip_wal=false`), the current implementation uses `RaftEngine` as default. The correct WAL options (e.g., Kafka topic) allocated by metasrv are persisted in `DatanodeTableValue` and will be loaded correctly on region restart, but the in-memory version may temporarily show `RaftEngine` even if the actual WAL is Kafka. This is documented with a TODO for future improvement.

## PR Checklist

### DIFF
--- a/src/mito2/src/worker/handle_alter.rs
+++ b/src/mito2/src/worker/handle_alter.rs
@@ -249,7 +249,7 @@ impl<S: LogStore> RegionWorkerLoop<S> {
                         }
                         current_options.wal_options = WalOptions::Noop;
                     } else {
-                        // Restore WAL options: priority order:
+                        // Restore WAL options: priority order
                         // 1. Provided wal_options from request
                         // 2. Saved original_wal_options
                         // 3. Fallback to RaftEngine


src/mito2/src/worker/handle_alter.rs
```diff
@@ -229,6 +230,24 @@ impl<S: LogStore> RegionWorkerLoop<S> {
                         );
                     }
                 }
+                SetRegionOption::SkipWal(skip_wal) => {
```
WenyXu: We can’t set the WAL options to `RaftEngine`; they could be `Kafka`. It would be better to let the MetaSrv send the detailed configuration.

src/mito2/src/worker/handle_alter.rs
```diff
@@ -229,6 +230,24 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+                        // Disable WAL by setting to Noop
+                        current_options.wal_options = WalOptions::Noop;
+                    } else {
+                        // Enable WAL: restore to default (RaftEngine)
+                        // TODO: In distributed mode, this should be allocated by metasrv,
+                        // but for simplicity, we use RaftEngine as default here.
+                        // The actual WAL options will be persisted in metasrv.
+                        // We should read the correct WAL options from DatanodeTableValue
+                        // or pass them through the AlterRegionRequest.
+                        current_options.wal_options = WalOptions::RaftEngine;
```
v0y4g3r: The WAL type could be Kafak

src/mito2/src/worker/handle_alter.rs
```diff
@@ -264,6 +283,13 @@ fn new_region_options_on_empty_memtable(
 
                 current_options.sst_format = Some(new_format);
             }
+            SetRegionOption::SkipWal(skip_wal) => {
```
WenyXu: If only the WAL options are configured, the underlying WAL provider is not affected, so the region continues writing data to the WAL.

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -283,13 +283,43 @@ impl AlterTableProcedure {
         };
 
         // Safety: region distribution is set in `submit_alter_region_requests`.
+        // Check if skip_wal changed and update WAL options if needed
+        let current_skip_wal = table_info_value.table_info.meta.options.skip_wal;
```
WenyXu: I prefer not to reallocate the WAL options, since a region might start writing to a different Kafka topic after reallocation, which could cause problems.

OmCheeLin:
Sorry, I accidentally closed the PR.

OmCheeLin:
@WenyXu Should I wait until [#7549](https://github.com/GreptimeTeam/greptimedb/pull/7549) to be merged?

WenyXu:
> @WenyXu Should I wait until [#7549](https://github.com/GreptimeTeam/greptimedb/pull/7549) to be merged?

No, go ahead whenever you’re ready.

OmCheeLin:
I have encountered some issues that need further confirmation at present. Should we modify the proto?

WenyXu:
> I have encountered some issues that need further confirmation at present. Should we modify the proto?

Could you explain the details of the issue?

PR #7554, opened 2026-01-09
feat: UDF json_get with user specified return type

### LINKED ISSUE #7500: new function: `json_get(column, path, datatype)`
we can make this json_get a function, and let user to pass type as the third argument, we will try to coerce the get result into that type. this will also align with our goal to provide dynamic type.

_Originally posted by @sunng87 in https://github.com/GreptimeTeam/greptimedb/pull/7495#discussion_r2651019233_

sunng87:

Fixes #7500

1. add `json_get` function that accept a return type, like `string`, `int`, 'float' and 'bool'
2. add datafusion `return_field_from_args` to our `Function` trait
3. port `json_get_bool` and `json_get_float` to new data type

## PR Checklist


waynexia:
How about use datatype itself as a parameter instead of parsing a string? like this https://github.com/waynexia/jsonfusion/blob/645f8eccf8eccf9d6139b848d1fffe9f70d95cdf/src/get_field_typed.rs#L79-L85

This can express a more complicated type, and reduce maintaince burden of the string to type mapping

sunng87:
@waynexia I'm afraid this will make the function difficult to understand, for example `json_get(a, 'a', true)` has exact same meaning of `json_get(a, 'a', false)`.

waynexia:
why use it like this, rather than `json_get(a, 'a')::boolean`. The third parameter should be for internal usage.

sunng87:
Even if we will be rewriting `json_get(a, 'a')::boolean` to the three-argument form, I still think using type name is more direct than using type instance for the hidden argument.

For example, if we want to support int array as return type, we can rewrite `json_get(a, 'a')::int4[]` to `json_get(a, 'a', 'int4[]')`, rather than generating a real int4[] literal value just to represent the type like `json_get(a, 'a', '{1}'::int4[])`

PR #7564, opened 2026-01-13
feat: impl vector index query

killme2008:

Main changes:

  - Added vector-search hinting inside `ScanHintRule` (new `scan_hint/vector_search.rs`), scoped to single-input paths and supporting `TopK` via `Sort.fetch`.
  - Dist plan sort commutativity now handles vector sorts in the unified Sort branch when the feature is enabled.
  - Improved safety/observability: empty vector inputs handled, vector scan metric added, and CI runs cargo check --all-features.

## PR Checklist

### DIFF
--- a/src/query/src/optimizer/scan_hint.rs
+++ b/src/query/src/optimizer/scan_hint.rs
@@ -106,8 +106,8 @@ impl ScanHintRule {
                             .take_vector_request_from_dummy(adapter, &table_scan.table_name)
                         {
                             adapter.with_vector_search_hint(vector_request);
-                            transformed = true;
                         }
+                        transformed = true;
                     }
                 }
                 if transformed {


src/query/src/optimizer/vector_search.rs
```diff
@@ -0,0 +1,720 @@
...
+    }
+
+    fn extract_distance_info(expr: &Expr) -> Option<VectorDistanceInfo> {
+        if let Expr::ScalarFunction(func) = expr {
+            let func_name = func.name().to_lowercase();
+            let metric = match func_name.as_str() {
+                VEC_L2SQ_DISTANCE => VectorDistanceMetric::L2sq,
+                VEC_COS_DISTANCE => VectorDistanceMetric::Cosine,
+                VEC_DOT_PRODUCT => VectorDistanceMetric::InnerProduct,
+                _ => return None,
```
v0y4g3r: Should we return an error here?

src/query/src/optimizer/vector_search.rs
```diff
@@ -0,0 +1,720 @@
...
+
+    fn parse_json_vector(s: &str) -> Option<Vec<f32>> {
+        let trimmed = s.trim();
+        if !trimmed.starts_with('[') || !trimmed.ends_with(']') {
+            return None;
+        }
+        let inner = &trimmed[1..trimmed.len() - 1];
+        if inner.trim().is_empty() {
+            return None;
+        }
```
v0y4g3r:
```suggestion
```

`parse_string_to_vector_type_value` already handles empty vector.

src/query/src/dist_plan/vector_scan.rs
```diff
@@ -0,0 +1,352 @@
...
+pub fn metric_to_u32(metric: VectorDistanceMetric) -> u32 {
+    match metric {
+        VectorDistanceMetric::L2sq => 0,
+        VectorDistanceMetric::Cosine => 1,
+        VectorDistanceMetric::InnerProduct => 2,
+    }
+}
+
+/// Converts a proto u32 to `VectorDistanceMetric`.
+pub fn u32_to_metric(value: u32) -> VectorDistanceMetric {
```
v0y4g3r: Why use u32 instead of protobuf enums in the proto definition so we can avoid these conversions?

src/query/src/dist_plan/vector_scan.rs
```diff
@@ -0,0 +1,352 @@
...
+                .iter()
+                .zip(other.query_vector.iter())
+                .all(|(a, b)| a.to_bits() == b.to_bits())
+    }
+}
+
+impl Eq for VectorScanLogicalPlan {}
+
+impl PartialOrd for VectorScanLogicalPlan {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
```
v0y4g3r: We can simplify this partial order comparison because `column_id`, `k`, `metric`, serialized float bits and `encoded_input` are all total order.

src/query/src/optimizer/vector_search.rs
killme2008: No—if we don’t detect a valid vector function in the `order by`, we simply don’t use the vector index, but should not return an error.

src/query/src/dist_plan/vector_scan.rs
killme2008: The `self.input.partial_cmp(&other.input)` would return `None`, it's not in total order.

I can simplify the code with tuple.

src/query/src/dist_plan/vector_scan.rs
killme2008: It’s not an enum in the proto, but I can convert it to one and implement the `From` trait.

We shouldn’t remove `VectorDistanceMetric` and use the proto enum directly, because it’s a core concept of vector search, and tying it to the proto doesn’t make sense.

src/query/src/dist_plan/vector_scan.rs
killme2008: Fixed in https://github.com/GreptimeTeam/greptimedb/pull/7564/changes/8ac4f7d96cdd34d032a5b8403d75313425d23ae3

src/query/src/dist_plan/vector_scan.rs
killme2008: Fixed in https://github.com/GreptimeTeam/greptimedb/pull/7564/changes/8ac4f7d96cdd34d032a5b8403d75313425d23ae3

.github/workflows/develop.yml
```diff
@@ -73,6 +73,8 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: Run cargo check
         run: cargo check --locked --workspace --all-targets
+      - name: Run cargo check (all features)
+        run: cargo check --locked --workspace --all-targets --all-features
```
evenyag, line 77: Why do we run the check twice?

evenyag, line 77: Better to push a branch to the main repo and manually run the workflow to ensure the action can pass. The CI won't use the yaml in this branch.

src/query/src/dist_plan/analyzer.rs
```diff
@@ -585,9 +597,29 @@ impl PlanRewriter {
...
+                on_node.clone(),
+                request.column_id,
+                request.query_vector,
+                request.k,
+                request.metric,
+            )
+            .into_logical_plan()
+        } else {
+            on_node.clone()
+        };
```
evenyag: The dist analyzer usually should not handle the vector scan plan.

You should check how the RangeSelectPlanner and PromPlanner handle new logical plans and how the PromExtensionPlanner handles physical plans.

The logical planner should add the vector scan node as the parent of the table scan.

The vector scan should have a physical plan. It may scan again if the table scan doesn't return enough rows. Otherwise, we don't need to add a new plan node. We can add `k` to the ScanRequest.

.github/workflows/develop.yml
killme2008: We should verify that the build succeeds with all features enabled or disabled.

killme2008: Already checked on local machine.

waynexia:
@codex review

src/query/src/optimizer/scan_hint/vector_search.rs
```diff
@@ -0,0 +1,896 @@
...
+            && table_name != hint_table
+        {
+            return None;
+        }
+
+        let metadata = provider.region_metadata();
+        let column = metadata.column_by_name(&info.column_name)?;
+
+        Some(VectorSearchRequest {
+            column_id: column.column_id,
```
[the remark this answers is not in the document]
killme2008: Fixed in https://github.com/GreptimeTeam/greptimedb/pull/7564/changes/999d50cb05844a880b5f6c84a21141a7b4426839

src/query/src/dist_plan/commutativity.rs
```diff
@@ -149,14 +188,19 @@ impl Categorizer {
...
                     return Ok(Commutativity::Commutative);
                 }
 
                 // sort plan needs to consider column priority
-                // Change Sort to MergeSort which assumes the input streams are already sorted hence can be more efficient
-                // We should ensure the number of partition is not smaller than the number of region at present. Otherwise this would result in incorrect output.
+                // Change Sort to MergeSort which assumes the input streams are already sorted hence can be more efficient.
+                #[cfg(feature = "vector_index")]
+                if is_vector_sort(_sort) {
+                    return Ok(Commutativity::TransformedCommutative {
```
discord9, line 200: Why need a extra check for vector sort? Is the original merge sort transformer not enough? Is Vector Sort must be push down?

src/query/src/optimizer/scan_hint.rs
```diff
@@ -93,6 +102,26 @@ impl ScanHintRule {
...
                         }
 
+                        #[cfg(feature = "vector_index")]
+                        if let Some(vector_request) = visitor
+                            .vector_search
+                            .take_vector_request_from_dummy(adapter, &table_scan.table_name)
+                        {
+                            adapter.with_vector_search_hint(vector_request);
+                        }
+                        transformed = true;
```
evenyag: Shouldn't we set transformed inside the `if-let`?
```suggestion
                        #[cfg(feature = "vector_index")]
                        if let Some(vector_request) = visitor
                            .vector_search
                            .take_vector_request_from_dummy(adapter, &table_scan.table_name)
                        {
                            adapter.with_vector_search_hint(vector_request);
                            transformed = true;
                        }
```

src/query/src/optimizer/scan_hint.rs
```diff
@@ -58,13 +64,16 @@ impl ScanHintRule {
         let _ = plan.visit(&mut visitor)?;
 
         if visitor.need_rewrite() {
-            plan.transform_down(&|plan| Self::set_hints(plan, &visitor))
+            plan.transform_down(&mut |plan| Self::set_hints(plan, &mut visitor))
```
evenyag, line 65: Prefer not to change this to `&mut` as this feature should not change the mutability.

src/query/src/optimizer/scan_hint.rs
```diff
@@ -93,6 +102,26 @@ impl ScanHintRule {
...
+                    #[cfg(feature = "vector_index")]
+                    if let Some(adapter) = source
+                        .table_provider
+                        .as_any()
+                        .downcast_ref::<DfTableProviderAdapter>()
+                        && let Some(vector_request) = visitor
+                            .vector_search
+                            .take_vector_request_from_adapter(adapter, &table_scan.table_name)
+                    {
+                        adapter.with_vector_search_hint(vector_request);
```
evenyag: We don't need to support DfTableProviderAdapter because other scan hints also don't support that.

src/query/src/dist_plan/commutativity.rs
killme2008: This extra check is to guarantee the vector index hint actually gets pushed down to the region server. The default MergeSort/Sort rewrites can fold or eliminate the Sort (e.g. TopK/Sort.fetch), which would drop the vector hint and degrade the query to a full scan. Vector sort is not just a performance tweak; it is the trigger for vector index search, so it must be preserved/pushed down. This is only enabled under `vector_index` feature to avoid affecting normal sort behavior.

src/query/src/optimizer/scan_hint.rs
killme2008: Good catch!

src/query/src/optimizer/scan_hint.rs
killme2008: ok

src/query/src/optimizer/scan_hint.rs
killme2008: I attempted to remove the `&mut`, but I couldn't find a good solution. I'm open to suggestions.

src/query/src/optimizer/scan_hint.rs
killme2008: The logic remains the same, so I don’t think we need to change it.

https://github.com/GreptimeTeam/greptimedb/blob/65f46b6a2f66510ff0e74fc69190c434e3b41193/src/query/src/optimizer/scan_hint.rs#L97

src/query/src/optimizer/scan_hint.rs
killme2008: Removed in https://github.com/GreptimeTeam/greptimedb/pull/7564/commits/e456b46540d7efcd3d8a503a0dd0a91171c36cd3

src/query/src/optimizer/scan_hint.rs
killme2008: If we modify it, we should set `transformed = true` in other hints as well. I can make those changes if you deem it necessary.

src/query/src/optimizer/scan_hint.rs
killme2008: I tried several options, but none are as simple as changing the `visitor` mutability.

MichaelScofield:
@gemini-code-assist review

src/query/src/optimizer/scan_hint.rs
evenyag: Can't we clone the vector hint instead of taking it? All other hints are cloned.

src/query/src/optimizer/scan_hint/vector_search.rs
```diff
@@ -0,0 +1,837 @@
...
+    }
+
+    pub(crate) fn take_vector_request_from_dummy(
+        &mut self,
+        provider: &DummyTableProvider,
+        table_name: &TableReference,
+    ) -> Option<VectorSearchRequest> {
+        let hint = self.take_vector_hint(table_name)?;
+        self.build_vector_request_from_dummy(provider, table_name, &hint)
+    }
```
evenyag, line 174: Maybe we can clone the hint to avoid the requirement of `&mut`.

killme2008: We can make take_vector_request_from_dummy take &self and clone, but that changes semantics. The current &mut consumes the queue so each hint is applied once. If we switch to &self, we need extra bookkeeping
(e.g. a consumed set) to avoid repeatedly applying the same hint to multiple scans. 

Alternatively, a mutex can protect the queue, maintaining internal mutability. If you think this is a good solution, I can implement it.

evenyag (approved):

MichaelScofield (approved):

PR #7590, opened 2026-01-19
feat: use tsid on promql planner

waynexia:

Probe if `__tsid` is available from current table, and use it if so

## PR Checklist

### DIFF
--- a/src/query/src/promql/planner.rs
+++ b/src/query/src/promql/planner.rs
@@ -182,6 +182,7 @@ impl PromPlannerContext {
     fn reset_table_name_and_schema(&mut self) {
         self.table_name = Some(String::new());
         self.schema_name = None;
+        self.use_tsid = false;
     }
 
     /// Check if `le` is present in tag columns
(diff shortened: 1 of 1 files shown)


waynexia:
@codex review

src/query/src/promql/planner.rs
```diff
@@ -1456,48 +1547,158 @@ impl PromPlanner {
...
+        }
 
         if !is_time_index_ms {
             // cast to ms if time_index not in Millisecond precision
             let expr: Vec<_> = self
                 .ctx
                 .field_columns
                 .iter()
-                .map(|col| DfExpr::Column(Column::new(Some(table_ref.clone()), col.clone())))
+                .map(|col| DfExpr::Column(Column::new(Some(scan_table_ref.clone()), col.clone())))
```
evenyag: Why do we use a way different from L1715 to get field exprs here? Just a question...
Maybe need a comment to document why we have to add the relation to the column expr.

src/query/src/promql/planner.rs
```diff
@@ -1581,6 +1798,7 @@ impl PromPlanner {
         self.ctx.reset_table_name_and_schema();
```
evenyag, line 1797: Do we need to reset the use_tsid in all places that call reset_table_name_and_schema()?

discord9:
@codex review

PR #7595, opened 2026-01-20
feat: add TTL filtering for expired rows at ingestion time

v0y4g3r:
## Summary

This PR adds TTL (Time To Live) filtering functionality to automatically filter out expired rows during data ingestion, improving storage efficiency and query performance by preventing expired data from being written to disk.

This PR avoids filtering out data when TTL is `instant` because that is specifically designed for flow tasks and we need the data, even if expired, stay in original table, to facilitate flow tasks in batching mode.

## Caveats

This PR introduces another iteration over the ingested data aside from data partition according to partition rules. We may need to merge these two iterations into single run for better performance in the future, if we find the performance degradation noticeable.

## Changes

- **Added TTL expiration check function**: `is_expired()` to determine if a row has exceeded its TTL
- **Implemented ingestion-time TTL filter**: New `ttl_filter.rs` module that filters expired rows before they reach storage
- **Added metrics**: Track number of rows filtered by TTL for observability
- **Integrated into partitioner**: TTL filtering happens after flownode mirroring to preserve instant TTL data
- **Added SQLness tests**: Comprehensive test coverage for TTL filtering behavior

## Test plan

- [x] Added SQLness test cases for TTL filtering (`ttl_ingestion_filter.sql`)
- [x] Updated existing test results affected by the new behavior
- [x] Verified clippy warnings are addressed

### DIFF
--- a/src/operator/src/req_convert/insert/row_to_region.rs
+++ b/src/operator/src/req_convert/insert/row_to_region.rs
@@ -112,8 +112,8 @@ pub fn filter_normal_requests_by_ttl(
             continue;
         };
 
-        // Skip filtering for instant TTL.
-        if matches!(ttl, TimeToLive::Instant) {
+        // Skip filtering for instant and forever TTL.
+        if matches!(ttl, TimeToLive::Instant | TimeToLive::Forever) {
             filtered_requests.push(request);
             continue;
         }
@@ -268,6 +268,7 @@ mod tests {
                 schema: build_column_schema(),
                 rows,
             }),
+            partition_expr_version: None,
         }
     }
 
@@ -479,6 +480,7 @@ mod tests {
             requests: vec![RegionInsertRequest {
                 region_id: region_id.as_u64(),
                 rows: None,
+                partition_expr_version: None,
             }],
         };
 
--- a/src/operator/src/req_convert/common/partitioner.rs
+++ b/src/operator/src/req_convert/common/partitioner.rs
@@ -14,7 +14,6 @@
 
 use api::v1::region::{DeleteRequest, InsertRequest};
 use api::v1::{PartitionExprVersion, Rows};
-use common_time::Timestamp;
 use partition::manager::PartitionRuleManager;
 use snafu::ResultExt;
 use store_api::storage::RegionId;


src/operator/src/req_convert/insert/row_to_region.rs
```diff
@@ -81,3 +85,70 @@ impl<'a> RowToRegion<'a> {
...
+        if matches!(ttl, TimeToLive::Instant) {
+            filtered_requests.push(request);
+            continue;
+        }
+
+        let Some(timestamp_index) = table_info.meta.schema.timestamp_index() else {
+            // No timestamp column, skip filtering (safe default)
+            filtered_requests.push(request);
+            continue;
+        };
```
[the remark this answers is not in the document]
v0y4g3r: fixed 9e57f0e9e775fe85a0240cac0261a64b528a1f09

fengjiachun:
@codex  review

src/operator/src/req_convert/insert/row_to_region.rs
```diff
@@ -81,3 +87,421 @@ impl<'a> RowToRegion<'a> {
...
 }
+
+/// Filter expired rows from normal insert requests based on TTL.
+/// This happens AFTER instant classification to preserve instant TTL data for flownodes.
+pub fn filter_normal_requests_by_ttl(
+    requests: RegionInsertRequests,
+    table_infos: &HashMap<TableId, Arc<TableInfo>>,
+) -> Result<RegionInsertRequests> {
+    let mut filtered_requests = Vec::with_capacity(requests.requests.len());
+    let now = Timestamp::current_millis();
```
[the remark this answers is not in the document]
v0y4g3r: That is intentional to use only one `now` for all rows in current request and avoid building one for each row for both efficiency and consistency. For those false negative results (expired, but not filtered out), we have futher expiration logic during compaction.

src/operator/src/req_convert/insert/row_to_region.rs
```diff
@@ -81,3 +87,421 @@ impl<'a> RowToRegion<'a> {
...
+            // Timestamp not found in row, simply fallback
+            filtered_requests.push(request);
+            continue;
+        };
+        let (filtered_rows, filtered_count) = filter_expired_rows(
+            std::mem::take(&mut rows_data.rows),
+            timestamp_index_in_rows,
+            ttl,
+            &now,
+        );
```
fengjiachun, line 147: We can add an optional `time_range` to `Rows` to indicate the batch's time span as a fast path, which will be used in most cases. The SDK may write the `time_range`; for protocols that require server-side parsing (e.g., Prometheus remote write), we can compute it during parsing. I think we can create an issue for now and handle this in the next PR.

src/operator/src/req_convert/insert/row_to_region.rs
```diff
@@ -81,3 +87,421 @@ impl<'a> RowToRegion<'a> {
...
+            })?;
+
+        let ttl = &table_info.meta.options.ttl;
+        let Some(ttl) = ttl else {
+            filtered_requests.push(request);
+            continue;
+        };
+
+        // Skip filtering for instant TTL.
+        if matches!(ttl, TimeToLive::Instant) {
```
evenyag: We can also skip `TimeToLive::Forever`.

tests/cases/standalone/common/flow/flow_advance_ttl.result
```diff
@@ -255,21 +255,13 @@ SELECT
 FROM
     out_distinct_basic;
 
-+-----+
-| dis |
-+-----+
-| 20  |
-| 22  |
-+-----+
```
evenyag, line 263: @discord9 Do you want to review these tests?

src/operator/src/req_convert/common/ttl_filter.rs
```diff
@@ -0,0 +1,337 @@
...
+use snafu::ResultExt;
+
+use crate::error;
+use crate::error::CheckTtlSnafu;
+
+/// Extracts the timestamp value from a row at the given column index.
+/// Returns None if the index is invalid or the value is not a timestamp type.
+pub fn extract_timestamp_from_row(row: &Row, timestamp_index: usize) -> Option<Timestamp> {
+    if timestamp_index >= row.values.len() {
+        return None;
```
killme2008, line 28: When does it happen?

tests-integration/tests/http.rs
```diff
@@ -6579,10 +6579,6 @@ pub async fn test_jaeger_query_api_for_trace_v1(store_type: StorageType) {
                 HeaderName::from_static("x-greptime-log-pipeline-name"),
                 HeaderValue::from_static(GREPTIME_INTERNAL_TRACE_PIPELINE_V1_NAME),
             ),
-            (
```
killme2008, line 6582: Why remove these options?

src/operator/src/insert.rs
```diff
@@ -405,9 +406,13 @@ impl Inserter {
         .await?;
         flow_mirror_task.detach(self.node_manager.clone())?;
 
+        // Filter expired rows from normal_requests AFTER flownode mirroring
+        // but BEFORE datanode writes. Instant requests are already excluded from datanodes.
+        let filtered_requests = filter_normal_requests_by_ttl(normal_requests, table_infos)?;
```
killme2008, line 411: I would like to add an option for this behaviour, and make it as false by default.

tests-integration/tests/http.rs
v0y4g3r: > Why remove these options?

Because this test always writes stale data and relies on incorrect behavior to work. Now these stale data will be immediately removed upon ingestion.

src/operator/src/insert.rs
v0y4g3r: I don't think so. Explicitly set TTL for a table and expect to be able to read the outdated data is self-contradictory. And adding dedicated option for that is not necessary.

Also see 
- VictoriaMetrics: https://github.com/VictoriaMetrics/VictoriaMetrics/blob/08755c83/lib/storage/storage.go#L1885-L1894
- Prometheus: https://github.com/prometheus/prometheus/blob/0279e14d/storage/remote/queue_manager.go#L655-L676

killme2008: Looks like it will hurt performance.

tests-integration/tests/http.rs
killme2008: A better approach is to correct the data and ensure the test still passes.

PR #7609, opened 2026-01-23
fix: use phy table when need fitler by tsid/table id

discord9:

as title, should use physical table if need to filter by internal columns, so that substrait wouldn't confuse and not found the internal column when decoding(since only physical table have those internal columns, but now optimizer will try to filter logical table with those columns (tsid&table id)

## PR Checklist

### DIFF
--- a/tests/cases/distributed/flow-tql/tsid_on_phy.sql
+++ b/tests/cases/distributed/flow-tql/tsid_on_phy.sql
@@ -52,9 +52,9 @@ CREATE TABLE IF NOT EXISTS test_tsid (
 INSERT INTO test_tsid
   (ts, v, tag1, tag2, le, tag4, tag5, tag8, tag6, tag7)
 VALUES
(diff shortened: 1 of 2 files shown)


tests/cases/distributed/flow-tql/tsid_on_phy.sql
```diff
@@ -0,0 +1,111 @@
...
+  )
+) ENGINE=metric WITH(
+  on_physical_table = 'phy'
+);
+
+INSERT INTO test_tsid
+  (ts, v, tag1, tag2, le, tag4, tag5, tag8, tag6, tag7)
+VALUES
+  ('2026-01-23T03:40:00Z', 10.0, 'istio-ingressgateway', 'outbound', '0.5',  'svc-a', 'prod', 'peer.example', 'svc-b', 'prod'),
+  ('2026-01-23T03:41:00Z', 5.0,  'istio-ingressgateway', 'outbound', '0.9',  'svc-a', 'prod', 'peer.example', 'svc-b', 'prod'),
```
waynexia: value of `0.5` bucket is greater than `0.9` bucket and `+INF` bucket, this input is incorrect. It's an accumulator kind of value that should be mono-increase

PR #7614, opened 2026-01-26
feat: ensure datanode plan's dist and ordering for tsid col

waynexia:

Make sure `__tsid` optimization will still follows series scan and distribution rules

## PR Checklist

### DIFF
--- a/src/table/src/table/scan.rs
+++ b/src/table/src/table/scan.rs
@@ -158,6 +158,9 @@ impl RegionScanExec {
                 // Metric engine physical tables keep tag columns alongside `__tsid`. When `__tsid`
                 // is present, the scan output is also ordered by `(__tsid, time index)` which can
                 // help eliminate redundant sorts for promql plans (e.g. SeriesDivide).
+                //
+                // `pk_time_ordering` and the potential `(__tsid, time index)` ordering are actually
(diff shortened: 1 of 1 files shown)


src/table/src/table/scan.rs
```diff
@@ -146,29 +147,56 @@ impl RegionScanExec {
...
+                        PhysicalSortExpr::new(
+                            Arc::new(tsid_col) as _,
+                            SortOptions {
+                                descending: false,
+                                nulls_first: true,
+                            },
+                        ),
+                        ts,
+                    ]);
+                }
```
evenyag, line 178: Does the order matter? We put tsid in the last column of the ordering.

waynexia:
@codex review

src/table/src/table/scan.rs
discord9: It is `OrderingEquivalenceClass` hence so it means this two ordering is the same:
1. [tag0 ASC, tag1 ASC ..., ts ASC]
2. [ts_id ASC, ts ASC]
so it's correct?

waynexia: that's `Vec<Vec<T>>` and in both cases ts col is the last

PR #7620, opened 2026-01-27
chore: add an initial internal sql-json-path query implementation

sunng87:

This patch uses [sql-json-path](https://github.com/risingwavelabs/sql-json-path/) to implement an early version of sql-json-path query on our current json data. We will keep improve this for the upcoming json data structures.

The sql-json-path query language is from postgres. More information: https://github.com/obartunov/sqljsondoc/blob/master/jsonpath.md

See sqlness files in the changeset for some details.

I'm not going to include this in our docs until our new json data structure is available. And we will need more effort to port this on it.

## PR Checklist


killme2008 (approved):
LGTM

src/common/function/src/scalars/json/json_path_query.rs
```diff
@@ -0,0 +1,127 @@
...
+            let path = paths.is_valid(i).then(|| paths.value(i));
+
+            let result = match (json, path) {
+                (Some(json), Some(path)) => {
+                    if !jsonb::is_null(json) {
+                        let jsonb_value = jsonb::from_slice(json).map_err(|e| {
+                            DataFusionError::Execution(format!("invalid jsonb binary: {e}"))
+                        })?;
+                        let mut json_str = jsonb_value.to_string();
+                        let json_value: simd_json::OwnedValue = unsafe {
```
MichaelScofield, line 84: Nit: best if we can unify the json string deserialization. Currently we have both serde_json and simd_json.

MichaelScofield (commented):
We have already used the jsonpath-rust for querying inside json data, why introduce another one?

src/common/function/src/scalars/json/json_path_query.rs
killme2008: @sunng87 PTAL

PR #7621, opened 2026-01-27
fix(repartition): improve physical region allocation and compaction read path correctness

WenyXu:

#6558 

This PR improves repartition correctness for physical tables and updates the mito2 compaction read path to handle sparse primary key encoding when partition expressions differ. It also expands repartition integration tests for format/encoding coverage and improves partition stability assertions.
- Meta DDL:
  - Add `build_template_from_raw_table_info_for_physical_table` for physical tables, including metric internal columns and deriving primary keys from tag semantics while preserving column IDs from metadata.
  - Remove `skip_internal_columns` flag from logical templates; logical table templates now always preserve original column order and PK indices.
- Mito2 Compaction & Read:
  - Introduce `CompactionProjectionMapper` to project flat-format compaction batches (fields + time index) and append internal columns, applied only when compaction is enabled, partition exprs differ, flat format is on, and PK encoding is sparse.
  - Fix `FlatReadFormat::format_projection` so `skip_auto_convert` uses a primary-key format projection for filter/prune index resolution.
  - Add `DfBatchAssembler` as a helper used by `CompactionProjectionMapper`; `FlatProjectionMapper` gains `project_vectors` to return projected vectors.
- Metric Engine:
  - Enter staging now skips metadata regions; callers are expected to pass data region IDs.
- Tests:
  - Repartition integration tests cover both flat and primary-key formats for mito, and the full format × PK-encoding matrix for metric.
  - Repartition validation now uses `information_schema.partitions` for stable partition ordering and IDs.
- Other changes:
  - Improve column metadata comparison diagnostics in DDL utils.
  - Add a `RecordBatch::to_df_record_batch` helper with Arrow view-array casting.
  - Make `MitoRegion::metadata` public for test helpers.
  - Add extra debug/logs in repartition allocation and region drop.

## PR Checklist

### DIFF
--- a/tests-integration/tests/repartition.rs
+++ b/tests-integration/tests/repartition.rs
@@ -450,13 +450,13 @@ pub async fn test_repartition_mito(store_type: StorageType, flat_format: bool) {
     )
     .await
     .unwrap();
-    let expected_create_table_after_merge = r#"+---------------+--------------+------------------------+----------------+----------------------+----------------------------------+-----------------------+----------------------------+
-| table_catalog | table_schema | table_name             | partition_name | partition_expression | partition_description            | greptime_partition_id | partition_ordinal_position |
-+---------------+--------------+------------------------+----------------+----------------------+----------------------------------+-----------------------+----------------------------+
-| greptime      | public       | repartition_mito_table | p0             | id                   | id < 5                           | 4398046511104         | 1                          |
-| greptime      | public       | repartition_mito_table | p1             | id                   | id >= 10 AND id < 20 OR id >= 20 | 4398046511105         | 2                          |
-| greptime      | public       | repartition_mito_table | p2             | id                   | id >= 5 AND id < 10              | 4398046511107         | 3                          |
-+---------------+--------------+------------------------+----------------+----------------------+----------------------------------+-----------------------+----------------------------+"#;
+    let expected_create_table_after_merge = r#"+---------------+--------------+------------------------+----------------+----------------------+-----------------------+-----------------------+----------------------------+
+| table_catalog | table_schema | table_name             | partition_name | partition_expression | partition_description | greptime_partition_id | partition_ordinal_position |
++---------------+--------------+------------------------+----------------+----------------------+-----------------------+-----------------------+----------------------------+
+| greptime      | public       | repartition_mito_table | p0             | id                   | id < 5                | 4398046511104         | 1                          |
+| greptime      | public       | repartition_mito_table | p1             | id                   | id >= 10              | 4398046511105         | 2                          |
+| greptime      | public       | repartition_mito_table | p2             | id                   | id >= 5 AND id < 10   | 4398046511107         | 3                          |
++---------------+--------------+------------------------+----------------+----------------------+-----------------------+-----------------------+----------------------------+"#;
     check_output_stream(result.data, expected_create_table_after_merge).await;
 
     let sql =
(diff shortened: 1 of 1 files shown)


WenyXu:
@codex review

src/mito2/src/sst/parquet/file_range.rs
```diff
@@ -628,10 +641,12 @@ impl RangeBase {
...
             }
 
-            // Get the column directly by its projected index
+            // Get the column directly by its projected index.
+            // If the column is missing and it's not a tag/time column, this filter is skipped.
+            // Assumes the projection indices align with the input batch schema.
             let column_idx = flat_format.projected_index_by_id(filter_ctx.column_id());
             if let Some(idx) = column_idx {
-                let column = &input.columns()[idx];
+                let column = &input.columns().get(idx).unwrap();
```
evenyag, line 649: This change looks like doing the same thing as `[idx]`.

src/mito2/src/sst/parquet/flat_format.rs
```diff
@@ -407,11 +418,20 @@ impl ParquetPrimaryKeyToFlat {
...
 
         let format = PrimaryKeyReadFormat::new(metadata.clone(), column_ids.iter().copied());
+        let primary_key_format_projection = if skip_auto_convert {
+            Some(FormatProjection {
+                projection_indices: format.projection_indices().to_vec(),
+                column_id_to_projected_index: format.field_id_to_projected_index().clone(),
+            })
+        } else {
+            None
+        };
```
evenyag: Can we just keep FormatProjection in PrimaryKeyReadFormat?

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -336,14 +339,45 @@ impl ParquetReaderBuilder {
...
         // Gets the metadata stored in the SST.
         let region_meta = Arc::new(Self::get_region_metadata(&file_path, key_value_meta)?);
+        let region_partition_expr = self
+            .expected_metadata
+            .as_ref()
+            .and_then(|meta| meta.partition_expr.as_ref());
+        let is_same_region_partition = Self::is_same_region_partition(
+            region_partition_expr.as_ref().map(|expr| expr.as_str()),
+            self.file_handle.meta_ref().partition_expr.as_ref(),
+        )?;
```
evenyag: We also do a similar check in build_partition_filter(). Can we optimize this?

src/mito2/src/worker/handle_drop.rs
```diff
@@ -275,6 +275,10 @@ pub(crate) async fn remove_region_dir_once(
     }
 
     if !has_parquet_file {
+        info!(
+            "Deleting region path: {} with files: {:?}",
+            region_path, files_to_remove_first
+        );
```
evenyag: Should we keep this log?

src/metric-engine/src/engine/create.rs
```diff
@@ -361,7 +361,7 @@ impl MetricEngineInner {
             .map(|(idx, metadata)| (metadata.column_schema.name.clone(), idx))
             .collect::<HashMap<String, usize>>();
 
-        // check if internal columns are not occupied
+        // // check if internal columns are not occupied
```
evenyag: ?

src/common/meta/src/ddl/utils.rs
```diff
@@ -457,17 +457,28 @@ pub fn extract_column_metadatas(
     let first_column_metadatas = schemas
         .swap_remove(0)
         .map(|first_bytes| ColumnMetadata::decode_list(&first_bytes).context(DecodeJsonSnafu))
-        .transpose()?;
+        .transpose()?
+        .map(|mut c| {
+            c.sort_unstable_by_key(|c| c.column_id);
+            c
+        });
```
evenyag: What does this fix?

src/metric-engine/src/engine/create.rs
```diff
@@ -361,15 +362,18 @@ impl MetricEngineInner {
...
             .collect::<HashMap<String, usize>>();
 
-        // check if internal columns are not occupied
+        let table_id_col_def = request.column_metadatas.iter().any(is_metric_name_col);
+        let tsid_col_def = request.column_metadatas.iter().any(is_tsid_col);
+
+        // // check if internal columns are not occupied or defined in the request
         ensure!(
-            !name_to_index.contains_key(DATA_SCHEMA_TABLE_ID_COLUMN_NAME),
+            !name_to_index.contains_key(DATA_SCHEMA_TABLE_ID_COLUMN_NAME) || table_id_col_def,
```
evenyag, line 370: Why do we need to check table_id_col_def?

src/common/meta/src/ddl/utils.rs
WenyXu: Removed

src/mito2/src/sst/parquet/file_range.rs
WenyXu: Good catch

src/metric-engine/src/engine/create.rs
WenyXu: This change allows MetaSrv to create a physical table whose column metadata order is consistent with the existing region.

PR #7631, opened 2026-01-28
feat: add function rewrite rule for json_get with cast

sunng87:

This patch provides support for sql like

```sql
SELECT json_get(parse_json('{"a":1}'), 'a')::text;

SELECT json_get(parse_json('{"a":1}'), 'a')::int8;
```

It also changes the third argument of `json_get` to accept value type instead of type name.

Note that we have a sql transform rewrite `type_alias` that rewrites `::int8` to `arrow_cast` function calls. So we will deal with both `arrow_cast` function call and cast expression here.

## PR Checklist


waynexia:
@codex review

sunng87:
Convert to draft for now. I'm going to remove the arrow_cast trasformation in `type_alias.rs` if possible. That will simplify the rewrite quite a bit.

Update: not removable. We still need arrow_cast for timestamps

tests/cases/standalone/common/function/json/json_get.result
```diff
@@ -293,70 +293,70 @@ SELECT json_get_int(json_get_object(j, '[9]'), 'a.i') FROM jsons;
...
-|                                            |
-|                                            |
-|                                            |
-+--------------------------------------------+
-
-SELECT json_get(j, '[3]', 'string') FROM jsons;
+SELECT json_get(j, '[0]')::int8 FROM jsons;
+
++---------------------------------------------------------+
+| arrow_cast(json_get(jsons.j,Utf8("[0]")),Utf8("Int64")) |
```
MichaelScofield, line 299: Why is the 2nd arg of `arrow_cast` int64 instead of int8?

sunng87: It's rewritten by type_alias.rs. 

This int8 meaning 8-byte integer, which is used in postgres (INT2, INT4, INT8). The Int64 is a arrow concept.

MichaelScofield (approved):

killme2008:
@sunng87  I think this feature could be implemented as a `TransformRule`:

https://github.com/GreptimeTeam/greptimedb/blob/279b009583f3811ec2d70c59cbea26e9a1c65f54/src/sql/src/statements/transform.rs#L38

Similar to how type aliasing is handled:

https://github.com/GreptimeTeam/greptimedb/tree/main/src/sql/src/statements/transform

It doesn't need to be implemented as an analyzer rule.

src/common/function/src/scalars/json/json_get.rs
```diff
@@ -507,20 +508,24 @@ fn json_struct_to_value(raw: &str, jsons: &StructArray, i: usize) -> Result<Valu
...
 impl JsonGetWithType {
-    const NAME: &'static str = "json_get";
+    pub(crate) const NAME: &'static str = "json_get";
 }
 
 impl Default for JsonGetWithType {
     fn default() -> Self {
         Self {
-            signature: Signature::any(3, Volatility::Immutable),
+            //signature: Signature::any(3, Volatility::Immutable),
```
waynexia:
```suggestion
```

sunng87:
@killme2008 Yes it's possible. This change is to use datafusion function rewriter API: https://docs.rs/datafusion/latest/datafusion/logical_expr/expr_rewriter/trait.FunctionRewrite.html and make this API work on our implementation.

killme2008:
@sunng87 Conflicts must be resolved.

Issue #7654, opened 2026-02-02
Upgrading version from `v1.0.0-beta.4` to `v1.0.0-rc.1` will OOM (Out of Memory) error


daviderli614:
### What type of bug is this?

Crash

### What subsystems are affected?

Standalone mode

### Minimal reproduce step

Related to [here](https://github.com/GreptimeTeam/greptime-config/commit/5a0b76d8a9f1766d0820c7274463b62700194ea7)

### What did you expect to see?

Pod status Running

### What did you see instead?

```
k get po -w
NAME                      READY   STATUS    RESTARTS       AGE
greptimedb-standalone-0   1/1     Running   5 (104s ago)   4m59s
greptimedb-standalone-0   0/1     OOMKilled   5 (105s ago)   5m
greptimedb-standalone-0   0/1     CrashLoopBackOff   5 (14s ago)    5m13s
```

### What operating system did you use?

AWS EC2 Linux x86_64

### What version of GreptimeDB did you use?

v1.0.0-rc.1

### Relevant log output and stack trace

```bash

```

daviderli614:
logs:
[logs.txt](https://github.com/user-attachments/files/25021610/logs.txt)

daviderli614:
![Image](https://github.com/user-attachments/assets/4fbfabe1-f938-47b3-9bb8-228859e6e7f4)

Maybe related to this PR: https://github.com/GreptimeTeam/greptimedb/pull/7312

PR #7656, opened 2026-02-03
fix: substract overflow in identity pipeline

shuiyisong:

fixes https://github.com/GreptimeTeam/greptimedb/issues/7654

As the title suggests. Adding the corresponding test as well.

## PR Checklist

### DIFF
--- a/tests-integration/tests/http.rs
+++ b/tests-integration/tests/http.rs
@@ -2535,7 +2535,7 @@ pub async fn test_identity_pipeline_with_null_column(store_type: StorageType) {
     // Test resolve_value function with various data types
     // Covers: VrlValue::Null, Integer, Float, Boolean, Bytes (string)
     // Note: Array and Object are flattened/stringified by identity pipeline before resolve_value
-    let body_all_types = r#"{"null_col":null,"int_col":42,"float_col":3.14,"bool_col":true,"str_col":"hello","array_col":[1,2,3],"obj_col":{"nested":"value"}}"#;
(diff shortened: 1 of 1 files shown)


src/pipeline/src/etl/transform/transformer/greptime.rs
```diff
@@ -820,6 +820,12 @@ fn identity_pipeline_inner(
     let column_count = schema_info.schema.len();
     for (_, row) in opt_map.iter_mut() {
         for row in row.iter_mut() {
+            assert!(
```
WenyXu, line 823: use `debug_assert` instead?

sunng87: debug_assert will be removed in production code, while assert will panic.
should we use `ensure` here?

shuiyisong: This (schema's count is less than the row's column count) case is **unexpected** and should never happen.
`debug_assert` will be removed in the production code. `ensure` is not strong enough, that the user might treat it as normal error and skip it, resulting all the data being discarded.
The `assert` used here is on purpose to trigger a panic so that we know the calculation code is broken.

tests-integration/tests/http.rs
```diff
@@ -2441,6 +2442,86 @@ pub async fn test_identity_pipeline(store_type: StorageType) {
     guard.remove_all().await;
 }
 
+/// Test for identity pipeline with null values for new columns.
+/// This test verifies that null values for columns not yet in the schema
+/// are handled correctly - they should NOT push to the row without updating schema.
+/// Regression test for: https://github.com/GreptimeTeam/greptimedb/issues/XXXX
```
fengys1996: nit: Write the specific issue number instead of `XXXX`?
```suggestion
/// Regression test for: https://github.com/GreptimeTeam/greptimedb/issues/7654
```

PR #7658, opened 2026-02-03
fix(copy): treat null inferred columns as Utf8 to preserve later values

WenyXu:

This PR:

- Normalize inferred file schemas so columns inferred as `Null` are treated as `Utf8`, preventing later non-null values from being dropped when `schema_infer_max_record` is too small.
- Ensure JSON export preserves explicit nulls so null-only columns remain visible during schema inference.
- Add CSV/JSON copy-from-fs tests where a column is null for the first N rows and non-null in the last row, validating import with `schema_infer_max_record = N-1`.

## PR Checklist


WenyXu:
@codex review

src/common/datasource/src/util.rs
```diff
@@ -27,9 +29,30 @@ pub fn find_dir_and_filename(path: &str) -> (String, Option<String>) {
...
+                f.clone()
+            }
+        })
+        .collect::<Vec<_>>();
+
+    arrow_schema::Schema {
+        fields: arrow_schema::Fields::from(fields),
+        metadata: schema.metadata,
+    }
+}
```
evenyag, line 52: Should we add an option to return error if can't infer a non null schema? Maybe we should allow users specifying the column schema in the future.

WenyXu: The upper layer will return an error if it attempts to insert a null value into a non-nullable column. Here, we only focus on normalizing the inferred schema.

PR #7659, opened 2026-02-03
test: sqlness compat

discord9:

run `cargo sqlness compat --from=1.0.0-beta.1 --to=current --preserve-state` to run compat test!
perhaps also need ci to test previous few versions automatically 

Still need more feature cases, but the general framework is this, the case dir might be a little to verbose, but that's the easiest way to impl this, and the general idea of compat test from rfc, and also `TEMPLATE` sqlness interceptor is also needed
due to metasrv need to restart, must use etcd to run which in turn need docker to run test

## PR Checklist


discord9:
@codex review

tests/compatibility/1.feature/standalone/common
```diff
@@ -0,0 +1 @@
+../distributed/common
```
fengjiachun, line 1: Add an empty line?

tests/compatibility/2.verify/standalone/common
```diff
@@ -0,0 +1 @@
+../distributed/common
```
fengjiachun, line 1: Add an empty line

tests/compatibility/3.cleanup/standalone/common
```diff
@@ -0,0 +1 @@
+../distributed/common
```
fengjiachun, line 1: Add an empty line

tests/runner/src/cmd/compat.rs
```diff
@@ -0,0 +1,92 @@
...
+            self.fail_fast,
+        )
+        .await
+        .context("Failed to create compatibility runner")?;
+
+        let run_result = runner.run().await;
+
+        if !self.preserve_state {
+            println!("Stopping etcd");
+            crate::util::stop_rm_etcd();
```
fengjiachun, line 75: CompatCommand always calls stop_rm_etcd() when --preserve-state is false, even if the selected cases never started a distributed environment or the run failed before etcd came up. That means a standalone-only compat run can still fail in cleanup on machines without Docker, and an earlier real test failure can be masked by this cleanup path. Cleanup should be conditional on whether this invocation actually launched etcd.

tests/runner/src/version.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    #[allow(unused)]
+    pub fn is_current(&self) -> bool {
+        matches!(self, Version::Current)
+    }
+
+    pub fn to_semantic(&self) -> Option<SemverVersion> {
+        match self {
+            Version::Semantic(v) => Some(v.clone()),
+            Version::Current => Self::current_version_from_cargo(),
```
fengjiachun, line 70: VERSION current is not the version of the binary under test" body="The VERSION interceptor resolves current from env!(\"CARGO_PKG_VERSION\"), which is just the workspace version baked into sqlness-runner at compile time.

discord9:
impl is too hacky, will return to first add more stuff to sqlness framework before retry this

PR #7669, opened 2026-02-05
feat: support changing table's append_mode to true

evenyag:

This PR supports changing the `append_mode` to true. Migrating a non append-only table to append-only isn't hard because we don't have to handle duplicated rows.
```
ALTER TABLE test_append_mode_true SET 'append_mode' = 'true';
```

Currently, we don't allow changing the `append_mode` from `true` to `false`.

## PR Checklist

### DIFF
--- a/src/table/src/metadata.rs
+++ b/src/table/src/metadata.rs
@@ -29,7 +29,7 @@ use serde::{Deserialize, Deserializer, Serialize};
 use snafu::{OptionExt, ResultExt, ensure};
 use store_api::metric_engine_consts::PHYSICAL_TABLE_METADATA_KEY;
 use store_api::mito_engine_options::{
-    APPEND_MODE_KEY, COMPACTION_TYPE, COMPACTION_TYPE_TWCS, SST_FORMAT_KEY,
+    APPEND_MODE_KEY, COMPACTION_TYPE, COMPACTION_TYPE_TWCS, MERGE_MODE_KEY, SST_FORMAT_KEY,
 };
 use store_api::region_request::{SetRegionOption, UnsetRegionOption};
 use store_api::storage::{ColumnDescriptor, ColumnDescriptorBuilder, ColumnId};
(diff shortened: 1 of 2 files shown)


WenyXu:
@codex review

tests/cases/standalone/common/alter/alter_append_mode.sql
```diff
@@ -0,0 +1,52 @@
...
+    cpu DOUBLE,
+    PRIMARY KEY(host)
+) ENGINE=mito WITH('append_mode'='true');
+
+-- Try to alter append_mode from true to false (should fail)
+-- SQLNESS REPLACE \d+\(\d+,\s+\d+\) REDACTED
+ALTER TABLE test_append_mode_true SET 'append_mode' = 'false';
+
+-- Altering to the same value should succeed
+ALTER TABLE test_append_mode_true SET 'append_mode' = 'true';
```
WenyXu, line 52: It would be better to run `SHOW CREATE TABLE` and restart all nodes to verify that the option is still available.

evenyag: fixed

src/mito2/src/worker/handle_open.rs
```diff
@@ -92,6 +92,7 @@ impl<S: LogStore> RegionWorkerLoop<S> {
             return;
         }
         info!("Try to open region {}, worker: {}", region_id, self.id);
+        sanitize_open_request_options(&mut request.options);
```
killme2008, line 95: why not move this function into `RegionOpenRequest`?

Than we can just call it via `request.sanitize_open_options();`

src/table/src/metadata.rs
```diff
@@ -364,6 +366,14 @@ impl TableMeta {
                         .extra_options
                         .insert(SST_FORMAT_KEY.to_string(), value.clone());
                 }
+                SetRegionOption::AppendMode(value) => {
+                    new_options
```
killme2008, line 370: Adds an assertion to ensure value is true?

killme2008 (approved):
Almost LGTM

src/mito2/src/worker/handle_open.rs
evenyag: This is mito-specific option and validation.

src/table/src/metadata.rs
evenyag: The option may be false here.

PR #7688, opened 2026-02-09
feat: adapt prometheus 3.x selector and lookback behavior

waynexia:

Prometheus 3.x also changes the instant selector behavior (start exclusive). And on the pair comparing test, I also find and fix a previous lookback handling bug that lookback won't take effect for `end` bound or matrix selector.

## PR Checklist

### DIFF
--- a/src/promql/src/extension_plan/instant_manipulate.rs
+++ b/src/promql/src/extension_plan/instant_manipulate.rs
@@ -467,7 +467,7 @@ impl Stream for InstantManipulateStream {
 impl InstantManipulateStream {
     // Refer to Prometheus `vectorSelectorSingle` / lookback semantics.
     //
-    // Prometheus `v3.9.1` (commit `9ec59ba`) uses a start-exclusive lookback window:
+    // Prometheus `v3.9.1` uses a start-exclusive lookback window:
     //   (eval_ts - lookback_delta, eval_ts]
     // i.e. a sample at exactly `eval_ts - lookback_delta` is considered too old.
     pub fn manipulate(&self, input: RecordBatch) -> DataFusionResult<RecordBatch> {


tests-integration/src/tests/promql_test.rs
```diff
@@ -236,9 +236,7 @@ async fn sql_insert_promql_query_ceil(instance: Arc<dyn MockInstance>) {
         \n| 1970-01-01T00:00:00 | 67.0      | 1024.0       | host1 |\
         \n| 1970-01-01T00:00:05 | 67.0      | 4096.0       | host1 |\
         \n| 1970-01-01T00:00:10 | 100.0     | 20480.0      | host1 |\
-        \n| 1970-01-01T00:00:50 | 12424.0   | 1334.0       | host1 |\
         \n| 1970-01-01T00:01:20 | 0.0       | 2334.0       | host1 |\
-        \n| 1970-01-01T00:01:40 | 49.0      | 3334.0       | host1 |\
```
sunng87, line 241: is these changes expected?

tests/cases/standalone/common/promql/set_operation.result
```diff
@@ -367,7 +367,6 @@ tql eval (0, 2000, '400') t1 or t2;
 | 1970-01-01T00:00:00 | 1.0            | a   |
 | 1970-01-01T00:06:40 | 0.0            |     |
 | 1970-01-01T00:13:20 | 0.0            |     |
-| 1970-01-01T00:13:20 | 2.0            | b   |
```
sunng87, line 370: is this expected?

waynexia: value at 800 second minus 300 lookback is >500 (exclusive) instead of >= 500 so this line disappear

tests-integration/src/tests/promql_test.rs
waynexia: here lookback is set to 1 second, so only value >99 will be considered

evenyag (approved):

sunng87 (approved):

PR #7694, opened 2026-02-09
fix(repartition): reject writes on deallocating regions during region merge

### LINKED ISSUE #7678: Data loss during concurrent inserts and repartition
### What type of bug is this?

Incorrect result

Distributed Cluster

### Minimal reproduce step

When running `ALTER TABLE ... MERGE/SPLIT PARTITION` (repartition) on a partitioned table, concurrent inserts during the repartition process may result in missing rows.
In fuzz testing, inserts with sequential timestamps were continuously generated, but after repartition finished, querying the table showed gaps in the time index, indicating that some inserted data was likely dropped during repartition.

1. Create a partitioned table:
```sql
CREATE TABLE et (
  voluptatibus DOUBLE,
  molestias TIMESTAMP(6) TIME INDEX,
  eT SMALLINT NOT NULL,
  consequaTur DOUBLE,
  ACcUSanTIUM DOUBLE,
  PRIMARY KEY(voluptatibus)
)
ENGINE=mito
PARTITION ON COLUMNS(voluptatibus) (...);
```
2. Start repartition by merging partitions:
```sql
ALTER TABLE et MERGE PARTITION (...);
```
3. While repartition is ongoing, continuously insert rows with sequential timestamps:
```sql
INSERT INTO et (...) VALUES
('2026-02-06 10:44:20.597', ...),
('2026-02-06 10:44:21.597', ...),
('2026-02-06 10:44:22.597', ...),
...
```

4. Query the timestamp:
```sql
SELECT molestias FROM et ORDER BY molestias;
```

- Inserts executed during repartition should not be lost.
- Query results should contain all inserted timestamps continuously.

### What did you see instead?

After repartition + concurrent inserts:
- Some timestamps are missing in query results.
- This suggests that writes happening during repartition may be dropped or not persisted correctly.

doesn't matter

main

```bash

```

WenyXu:

#6558 
close #7678

This PR implements write rejection for **deallocating regions** in the **repartition merge (multiple regions -> fewer regions)** scenario, so writes are no longer accepted by regions that are being deallocated.

- Add route-level write policy in table route metadata:
  - `RegionRoute.write_route_policy` with `IgnoreAllWrites` semantics.
  - Backward-compatible serde handling for existing route metadata.
- Normalize region partition compatibility handling:
  - `Region` compatibility path now keeps normalized `partition_expr` behavior while preserving deserialization compatibility for legacy partition payloads.
- Introduce enter-staging partition rule enum in instruction/request flow:
  - `PartitionExpr(String)` and `RejectAllWrites`.
  - Enter-staging path carries explicit rule semantics.
- Repartition metadata updates for merge flow:
  - Mark `pending_deallocate_region_ids` as reject-all during apply-staging metadata update.
  - Preserve reject-all policy on exit-staging path (and keep rollback behavior explicit).
- Partition write routing update:
  - Exclude reject-all regions from write partition construction to avoid routing writes to deallocating regions.
- Mito staging state and write gate:
  - Store staging partition rule in region staging state.
  - Enforce reject-all in write path (including staged bulk-write path).
- Tests updated across affected modules:
  - `meta-srv` repartition update metadata and enter-staging tests.
  - `mito2` staging/write tests.
  - related request/routing compatibility tests.

## PR Checklist

### DIFF
--- a/src/common/meta/src/instruction.rs
+++ b/src/common/meta/src/instruction.rs
@@ -538,7 +538,7 @@ impl Display for EnterStagingRegion {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum StagingPartitionDirective {
-    PartitionExpr(String),
+    UpdatePartitionExpr(String),
     RejectAllWrites,
 }
 
@@ -546,7 +546,7 @@ impl StagingPartitionDirective {
     /// Returns the partition expression carried by this directive, if any.
     pub fn as_partition_expr(&self) -> Option<&str> {
         match self {
-            Self::PartitionExpr(expr) => Some(expr),
+            Self::UpdatePartitionExpr(expr) => Some(expr),
             Self::RejectAllWrites => None,
         }
     }
@@ -555,7 +555,7 @@ impl StagingPartitionDirective {
 impl Display for StagingPartitionDirective {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
-            Self::PartitionExpr(expr) => write!(f, "PartitionExpr({})", expr),
+            Self::UpdatePartitionExpr(expr) => write!(f, "UpdatePartitionExpr({})", expr),
             Self::RejectAllWrites => write!(f, "RejectAllWrites"),
         }
     }
@@ -569,7 +569,7 @@ where
     S: Serializer,
 {
     match rule {
-        StagingPartitionDirective::PartitionExpr(expr) => serializer.serialize_str(expr),
+        StagingPartitionDirective::UpdatePartitionExpr(expr) => serializer.serialize_str(expr),
         StagingPartitionDirective::RejectAllWrites => {
             #[derive(Serialize)]
             struct RejectAllWritesSer<'a> {
@@ -598,7 +598,7 @@ where
     }
 
     match Compat::deserialize(deserializer)? {
-        Compat::Legacy(expr) => Ok(StagingPartitionDirective::PartitionExpr(expr)),
+        Compat::Legacy(expr) => Ok(StagingPartitionDirective::UpdatePartitionExpr(expr)),
         Compat::TypeTagged { r#type } if r#type == "reject_all_writes" => {
             Ok(StagingPartitionDirective::RejectAllWrites)
         }
@@ -1293,7 +1293,7 @@ mod tests {
         assert_eq!(enter.region_id, RegionId::new(1024, 1));
         assert_eq!(
             enter.partition_directive,
-            StagingPartitionDirective::PartitionExpr(
+            StagingPartitionDirective::UpdatePartitionExpr(
                 "{\"Expr\":{\"lhs\":{\"Column\":\"x\"},\"op\":\"GtEq\",\"rhs\":{\"Value\":{\"Int32\":0}}}}"
                     .to_string()
             )
--- a/src/mito2/src/region.rs
+++ b/src/mito2/src/region.rs
@@ -182,7 +182,9 @@ impl StagingPartitionInfo {
     /// Builds staging partition info from a directive and derives its version marker.
     pub(crate) fn from_partition_directive(partition_directive: StagingPartitionDirective) -> Self {
         let partition_rule_version = match &partition_directive {
-            StagingPartitionDirective::PartitionExpr(expr) => partition_expr_version(Some(expr)),
+            StagingPartitionDirective::UpdatePartitionExpr(expr) => {
+                partition_expr_version(Some(expr))
+            }
             StagingPartitionDirective::RejectAllWrites => 0,
         };
         Self {
--- a/src/mito2/src/worker/handle_enter_staging.rs
+++ b/src/mito2/src/worker/handle_enter_staging.rs
@@ -128,7 +128,9 @@ impl<S: LogStore> RegionWorkerLoop<S> {
         }
 
         let partition_expr = match partition_directive {
-            StagingPartitionDirective::PartitionExpr(partition_expr) => partition_expr.clone(),
+            StagingPartitionDirective::UpdatePartitionExpr(partition_expr) => {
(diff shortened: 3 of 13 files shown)


WenyXu:
@codex review

src/mito2/src/region.rs
```diff
@@ -168,8 +169,33 @@ pub type MitoRegionRef = Arc<MitoRegion>;
...
+        Self {
+            partition_directive,
+            partition_rule_version,
+        }
+    }
+}
+
+pub(crate) fn reject_all_writes_partition_expr_version() -> u64 {
+    partition_expr_version(Some("__greptime_reject_all_writes__"))
 }
```
evenyag: Why not just use 0?

src/common/meta/src/instruction.rs
```diff
@@ -518,19 +518,97 @@ impl Display for GcRegionsReply {
...
+            "EnterStagingRegion(region_id={}, partition_directive={})",
+            self.region_id, self.partition_directive
         )
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum StagingPartitionDirective {
+    PartitionExpr(String),
+    RejectAllWrites,
```
MichaelScofield: Can't talk myself out this is a good choice that putting these two together in an enum, they seem little relevant. That controlling write in the region route, too. Is it because there lack of a "`RegionInfo`" in the metasrv?

src/mito2/src/worker/handle_apply_staging.rs
```diff
@@ -79,14 +79,15 @@ impl<S: LogStore> RegionWorkerLoop<S> {
         // If the partition expr mismatch, return error.
         if staging_partition_info
             .as_ref()
-            .map(|info| &info.partition_expr)
+            .and_then(|info| info.partition_expr().map(ToString::to_string))
+            .as_ref()
             != Some(&request.partition_expr)
```
evenyag: I think we can avoid to_string here? Maybe call `request.partition_expr.as_str()`?

src/mito2/src/worker/handle_enter_staging.rs
```diff
@@ -123,6 +127,14 @@ impl<S: LogStore> RegionWorkerLoop<S> {
             );
         }
 
+        let partition_expr = match partition_directive {
+            StagingPartitionDirective::PartitionExpr(partition_expr) => partition_expr.clone(),
+            StagingPartitionDirective::RejectAllWrites => {
+                // Rejects all writes just a memory flag, no need to write new staging manifest.
+                return Ok(());
+            }
```
evenyag, line 141: Do we need a log here?

src/common/meta/src/instruction.rs
WenyXu: I’m trying to introduce a flag for a region to reject all writes. Imagine the following scenario: region A accepts the range [0..10), and region B accepts [10..15). If users attempt to merge A and B, and B is expanded to cover [0..15), then region A should implicitly enforce a “reject-all-writes” rule.

>  Is it because there lack of a "RegionInfo" in the metasrv?

There should be two flags: one in the metasrv, so that the frontend knows to stop routing any new writes to Region X, and another on Region X itself, so that it knows it must reject all write requests.

MichaelScofield: Can partition expr alone do the job? If `A: [0..10), B: [10..15)` becames `A: null, B: [0..15)`, the latter naturally implies "no writes to A`.

WenyXu: Currently, null or None means it accepts all writes.

src/mito2/src/worker/handle_enter_staging.rs
WenyXu: [4aaa253](https://github.com/GreptimeTeam/greptimedb/pull/7694/commits/4aaa253272bfd4cd09b3c3ad82ef4d0e5cb709a5)

src/mito2/src/worker/handle_apply_staging.rs
WenyXu: [4aaa253](https://github.com/GreptimeTeam/greptimedb/pull/7694/commits/4aaa253272bfd4cd09b3c3ad82ef4d0e5cb709a5)

src/mito2/src/region.rs
WenyXu: [4aaa253](https://github.com/GreptimeTeam/greptimedb/pull/7694/commits/4aaa253272bfd4cd09b3c3ad82ef4d0e5cb709a5)

src/common/meta/src/instruction.rs
WenyXu: Before this PR, we had an `Option<PartitionExpr>` in `RegionMetadata` & `Region` struct:
- `None` represented no partitioning / a single partition (accepting all writes).
- `PartitionExpr` was used to specify a partition.

With this PR, we’re introducing a new behavior: `rejecting all writes`.
Since we don’t need to persist the reject all writes state into the manifest, we continue to use `Option<PartitionExpr>` in `RegionMetadata` (the region manifest persisted to S3). For `EnterStagingRequest`, we introduce a new enum, `StagingPartitionDirective`, which only allows either a `PartitionExpr` or `RejectAllWrites`.

cc @fengjiachun @evenyag

evenyag: Maybe we can rename the variant from `PartitionExpr` to `UpdatePartitionExpr`.

MichaelScofield: Still think the partition expr that denotes "empty" is more adequate for this purpose. But to not postpone the review process, I'm letting it go.

MichaelScofield (approved):

src/common/meta/src/instruction.rs
WenyXu: The “empty” expression already implies a single partition.

PR #7708, opened 2026-02-12
perf(mito2): speed up parquet scan via minmax caches

waynexia:

Implement minmax filter cache and preload file metadata on region open.

For the basic filter case (query with only a few pk eq filters, without external index) over ~30 files, it achieves 4x code run (the first query of a standalone instance) latency improvement on my environment (from 0.8s to 0.2s).

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet/reader.rs
+++ b/src/mito2/src/sst/parquet/reader.rs
@@ -1473,7 +1473,7 @@ impl ReaderFilterMetrics {
 }
 
 #[cfg(all(test, feature = "vector_index"))]
-mod tests {
+mod vector_index_tests {
     use super::*;
 
     #[test]


src/mito2/src/cache/index/result_cache.rs
```diff
@@ -239,6 +247,28 @@ impl InvertedIndexKey {
...
+    exprs: Arc<Vec<String>>,
+    schema_version: u64,
+    skip_fields: bool,
+    mem_usage: usize,
+}
+
+impl MinMaxKey {
+    pub fn new(exprs: Arc<Vec<String>>, schema_version: u64, skip_fields: bool) -> Self {
+        let mem_usage =
+            exprs.iter().map(|s| s.len()).sum::<usize>() + size_of::<u64>() + size_of::<bool>();
```
evenyag: non-blocking: Should we add `size_of::<Self>() + size_of::<Vec<String>>()` instead of `u64 + bool`?

src/mito2/src/region/opener.rs
```diff
@@ -943,6 +946,110 @@ fn maybe_load_cache(
...
+    }
+
+    let region = region.clone();
+    let cache_manager = cache_manager.clone();
+
+    tokio::spawn(async move {
+        let region_id = region.region_id;
+        let table_dir = region.access_layer.table_dir().to_string();
+        let path_type = region.access_layer.path_type();
+        let object_store = region.access_layer.object_store().clone();
```
evenyag, line 1063: Can we add a way to limit the parallelism of this task? Maybe also provide a way to disable preloading without disabling the metadata cache.

src/mito2/src/region/opener.rs
```diff
@@ -943,6 +946,110 @@ fn maybe_load_cache(
...
+async fn preload_parquet_meta_cache_for_files(
+    region_id: RegionId,
+    table_dir: String,
+    path_type: PathType,
+    object_store: object_store::ObjectStore,
+    cache_manager: CacheManagerRef,
+    mut files: Vec<FileHandle>,
+) -> usize {
+    // Load older files first so the most recent files remain hot in the LRU cache.
+    files.sort_by(|a, b| a.meta_ref().time_range.1.cmp(&b.meta_ref().time_range.1));
```
evenyag: Loading all files' metadata may be costly (on S3). In some environments, most regions are rarely queried.

Maybe we can adjust the strategy:
- Limit the total number of files to preload?
- Only preload metadata from file cache?
- Trigger a task to load metadata on a cache miss of the file cache. e.g. Maybe via something like the DownloadTask of the file cache. We can preload N files in adjacent time ranges.

src/mito2/src/region/opener.rs
```diff
@@ -943,6 +946,110 @@ fn maybe_load_cache(
...
+            files,
+        )
+        .await;
+
+        if loaded > 0 {
+            info!(
+                "Preloaded parquet metadata for region {}, loaded_files: {}",
+                region_id, loaded
+            );
+        }
```
evenyag, line 1095: We can also log the load time.

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -1136,20 +1135,54 @@ impl ParquetReaderBuilder {
...
+            } else {
+                None
+            };
+
+        if let Some(index_result_cache) = index_result_cache
+            && let Some(predicate_key) = cached_minmax_key.as_ref()
+            && let Some(result) = index_result_cache.get(predicate_key, file_id)
+        {
+            let num_row_groups = parquet_meta.num_row_groups();
+            metrics.rg_minmax_filtered += num_row_groups.saturating_sub(result.row_group_count());
```
evenyag: Also add metrics for minmax cache hit/miss?

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -1136,20 +1135,54 @@ impl ParquetReaderBuilder {
...
+        let file_id = self.file_handle.file_id().file_id();
+        let index_result_cache = self.cache_strategy.index_result_cache();
+        let cached_minmax_key =
+            if index_result_cache.is_some() && predicate.dyn_filters().is_empty() {
+                // Cache min-max pruning results keyed by predicate expressions. This avoids repeatedly
+                // building row-group pruning stats for identical predicates across queries.
+                let mut exprs = predicate
+                    .exprs()
+                    .iter()
+                    .map(|expr| format!("{expr:?}"))
```
evenyag, line 1161: non-blocking: Since expr also implements `Display`, which is better for the cache key?

evenyag (commented):
A few questions about the preload strategy.

killme2008:
@waynexia Clippy warnings and check failures exist.

src/mito2/src/region/opener.rs
waynexia: it now only loads from non-remote storages

Issue #7706, opened 2026-02-12
ADBC-postgres tracking issue


sunng87:
### What type of enhancement is this?

API improvement

### What does the enhancement do?

This is a tracking issue of our support of ADBC-postgres clients. We hope to provide some level of support of ADBC-postgres clients to work with GreptimeDB, with a controllable cost of effort.

Current status:

- [x] SQL Queries with `COPY` optimization turned off. `adbc_driver_postgresql.StatementOptions.USE_COPY.value: False`
- [ ] SQL Queries with `COPY` optimization turned on: implement COPY TO STDOUT on our pg interface #7709 
- [x] Ingestion: NOT GOING TO SUPPORT
- [ ] Getting arrow-schema using `conn.adbc_get_table_schema("table_name")` #7709 
- [ ] pandas query: cannot configure the cursor when using `pd.read_sql` #7709 
- [x] polars query

sunng87:
The pg_catalog query we need to support

```
SELECT attname, atttypid FROM pg_catalog.pg_class AS cls INNER JOIN pg_catalog.pg_attribute AS attr ON cls.oid = attr.attrelid INNER JOIN pg_catalog.pg_type AS typ ON attr.atttypid = typ.oid WHERE attr.attnum >= 0 AND cls.oid = $1::regclass::oid ORDER BY attr.attnum
```

PR #7709, opened 2026-02-13
feat: implement postgres copy to stdout

sunng87:

#7706 

This patch implements `COPY (query) TO STDOUT WITH (format binary)`, which is required by ADBC-postgres for optimizations.

- This copy query will use postgrs PGCOPY protocol to send data to client
- This feature is implemented **only** for postgres interface
- The COPY TO STDOUT is parsed at protocol layer of postgres interface, to control the output encoding
- I shortcut our SQL parser to treat COPY (query) TO STDOUT just as query.
- Will switch to merged/released pgwire and arrow-pg when implementation confirmed.
- Refactored encoding utilities in postgres types to support CopyEncoder
- Related test: https://github.com/GreptimeTeam/greptimedb-tests/pull/7
- Updated datafusion-pg-catalog fork to address the schema query: https://github.com/GreptimeTeam/datafusion-postgres/commit/a75a3c55aba6c129f9d50bd39cb75d9e9004b304

## PR Checklist

### DIFF
--- a/src/sql/src/parsers/copy_parser.rs
+++ b/src/sql/src/parsers/copy_parser.rs
@@ -825,5 +825,11 @@ mod tests {
             ParserContext::create_with_dialect(sql, &GreptimeDbDialect {}, ParseOptions::default());
 
         assert!(result.is_err());
+
+        let sql = "COPY (SELECT * FROM test_table) TO STDOUT (FORMAT csv))";
+        let result =
+            ParserContext::create_with_dialect(sql, &GreptimeDbDialect {}, ParseOptions::default());
+
+        assert!(result.is_err());
     }
 }


src/servers/src/lib.rs
```diff
@@ -64,6 +64,8 @@ pub struct SqlPlan {
     statement: Option<Statement>,
     plan: Option<LogicalPlan>,
     schema: Option<Schema>,
+    // Cached result of check_copy_to_stdout for COPY TO STDOUT statements
+    copy_to_stdout_format: Option<String>,
```
MichaelScofield: Nit: since `copy to stdout` is pg only, it'd be better to be placed in a new `struct PgSqlPlan`.

src/sql/src/parsers/copy_parser.rs
```diff
@@ -147,21 +159,55 @@ impl ParserContext<'_> {
...
+            // we support both "WITH (FORMAT binary)" and "(FORMAT binary)"
+            // for PostgreSQL compatibility
+
+            // Check for optional WITH keyword or direct LParen (PostgreSQL syntax)
+            // Both "WITH (...)" and "(...)" are valid after STDOUT
+            let _ = self.parser.parse_keyword(Keyword::WITH);
+            if self.parser.peek_token().token == Token::LParen {
+                let _ = self.parser.next_token();
+                // consume all tokens until we find matching RParen
+                let mut depth = 1;
```
MichaelScofield, line 176: What does `depth` be used for?

sunng87: in case there are multiple `(` after `WITH`

fengys1996:
@codex review

src/sql/src/parsers/copy_parser.rs
```diff
@@ -147,21 +159,62 @@ impl ParserContext<'_> {
...
+                            return error::UnexpectedTokenSnafu {
+                                expected: ")",
+                                actual: "EOF",
+                            }
+                            .fail();
+                        }
+                        _ => {}
+                    }
+                }
+            }
```
MichaelScofield, line 191: Does token EOF need to be checked here? Otherwise "())" will be treated as legal.

sunng87: Let me try this. I thought the last `)` will be treated as next statement.

sunng87: Yes, additional `)` will be treated as illegal

PR #7710, opened 2026-02-14
refactor(mito): adjust process order, move compat before pruning

evenyag:

Part 1 of refactoring the mito scanner.

This PR
- Moves schema compatibility (compat) logic from the scan layer into `ReadFormat` (`PrimaryKeyReadFormat` and `FlatReadFormat`), so that batches are adapted to the expected schema at read time rather than after reading
- Removes `CompatReader`, `CompatBatch` enum, and the `compat_batch` field from `FileRangeContext`/`RangeBase`, simplifying the scan pipeline
- Updates `PrimaryKeyCompatBatch` and `FlatCompatBatch` to accept `expected_meta`, `actual_meta`, and `column_ids` directly instead of depending on `ProjectionMapper`/`FlatProjectionMapper`
- Uses the expected metadata (instead of actual SST metadata) when building `CompactionProjectionMapper`, since compat now transforms batches before projection

Previously, compat was applied externally in `scan_region.rs` and `scan_util.rs` after reading batches from SSTs. This created a coupling between the scan layer and the compat logic, and required threading `CompatBatch` through `FileRangeContext`.

We may be able to simplify or remove the MaybeFilter after we merge this PR.

## PR Checklist

### DIFF
--- a/src/mito2/src/read/compat.rs
+++ b/src/mito2/src/read/compat.rs
@@ -147,6 +147,7 @@ impl FlatCompatBatch {
     /// - `actual_format_projection` is the projection of the read format for the input parquet.
     /// - `column_ids` are the projected column ids to read.
     /// - `compaction` indicates whether the reader is for compaction.
+    ///
     /// Creates a [FlatCompatBatch] and returns it along with a mapping from column_id
     /// to projected index in the post-compat batch.
     ///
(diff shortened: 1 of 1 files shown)


evenyag:
I'm going to remove the old read path in https://github.com/GreptimeTeam/greptimedb/issues/7732. After that, we can simplify the refactoring.

PR #7712, opened 2026-02-17
feat: add a fallback parameter type inference by reading cast type

sunng87:

Datafusion can only inference parameter types for few cases. For example, it's not able to inference data type of scalar function parameter like `$1` in `SELECT parse_ident($1)`. If the client driver doesn't provide parameter type either, the server will end up with error.

This patch add a fallback approach to use cast information for data type, we will need user to provide type information with type cast: `SELECT parse_ident($1::TEXT)`. We will assume `$1` is `TEXT`(Utf8).

## PR Checklist

### DIFF
--- a/src/servers/src/mysql/handler.rs
+++ b/src/servers/src/mysql/handler.rs
@@ -51,7 +51,7 @@ use crate::error::{
 };
 use crate::metrics::METRIC_AUTH_FAILURE;
 use crate::mysql::helper::{
-    self, fix_placeholder_types, format_placeholder, replace_placeholders, transform_placeholders,
+    self, format_placeholder, replace_placeholders, transform_placeholders,
 };
 use crate::mysql::writer;
 use crate::mysql::writer::{create_mysql_column, handle_err};
@@ -209,7 +209,7 @@ impl MysqlInstanceShim {
         let describe_result = self
             .do_describe(statement.clone(), query_ctx.clone())
             .await?;
-        let (mut plan, schema) = if let Some(DescribeResult {
+        let (plan, schema) = if let Some(DescribeResult {
             logical_plan,
             schema,
         }) = describe_result
@@ -219,9 +219,7 @@ impl MysqlInstanceShim {
             (None, None)
         };
 
-        let params = if let Some(plan) = &mut plan {
-            fix_placeholder_types(plan)?;
-            debug!("Plan after fix placeholder types: {:#?}", plan);
+        let params = if let Some(plan) = &plan {
             let param_types = DfLogicalPlanner::get_inferred_parameter_types(plan)
                 .context(InferParameterTypesSnafu)?
                 .into_iter()
@@ -294,8 +292,7 @@ impl MysqlInstanceShim {
         };
 
         let outputs = match sql_plan.plan {
-            Some(mut plan) => {
-                fix_placeholder_types(&mut plan)?;
+            Some(plan) => {
                 let param_types = DfLogicalPlanner::get_inferred_parameter_types(&plan)
                     .context(InferParameterTypesSnafu)?
                     .into_iter()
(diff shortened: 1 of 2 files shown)


waynexia:
Is it possible to organize all type infer logic into one rule? like jsonfusion's "strategies"

sunng87:
> Is it possible to organize all type infer logic into one rule? like jsonfusion's "strategies"

This `get_infered_parameter_types` is designed to hold all inference logic as our wrapper of datafusion's original `get_parameter_types`. We can add more strategy into it.

src/query/src/planner.rs
```diff
@@ -405,6 +408,89 @@ impl DfLogicalPlanner {
...
+    /// - None if the parameter type could not be inferred
+    ///
+    /// This function first uses DataFusion's `get_parameter_types()` to infer types.
+    /// If any parameters have `None` values (i.e., DataFusion couldn't infer their types),
+    /// it falls back to using `extract_placeholder_cast_types()` to detect explicit casts.
+    ///
+    /// This is because datafusion can only infer types for a limited cases.
+    ///
+    /// Example: For query `WHERE $1::TEXT AND $2`, DataFusion may not infer `$2`'s type,
+    /// but this function will return `{"$1": Some(DataType::Utf8), "$2": None}`.
```
evenyag, line 465: Do we need an integration test for this fallback?

sunng87: We will do it later in the test repo.

waynexia:
@codex review

src/query/src/planner.rs
```diff
@@ -405,6 +408,89 @@ impl DfLogicalPlanner {
...
+                let _ = expr.apply(|e| {
+                    if let DfExpr::Cast(cast) = e
+                        && let DfExpr::Placeholder(ph) = &*cast.expr
+                    {
+                        placeholder_types.insert(ph.id.clone(), Some(cast.data_type.clone()));
+                        casted_placeholders.insert(ph.id.clone());
+                    }
+
+                    if let DfExpr::Placeholder(ph) = e
+                        && !casted_placeholders.contains(&ph.id)
```
MichaelScofield, line 438: Seems `casted_placeholders` is useless? If the placeholder is casted, it's in the `placeholder_types`; if not, and if it was not found casted before, it's not in the `placeholder_types`, thus the next line can just fulfill the demands.

src/query/src/planner.rs
```diff
@@ -405,6 +408,89 @@ impl DfLogicalPlanner {
...
+        } else {
+            let cast_types = Self::extract_placeholder_cast_types(plan)?;
+
+            let mut merged = param_types;
+
+            for (id, opt_type) in cast_types {
+                merged
+                    .entry(id)
+                    .and_modify(|existing| {
+                        if existing.is_none() {
```
MichaelScofield, line 484: nit:
```suggestion
                        if existing.is_none() && opt_type.is_none() {
```

MichaelScofield (approved):

discord9:
maybe add a sqlness test?

src/query/src/planner.rs
sunng87: A placeholder can be in the statement multiple times like

```
where $1 > 10 AND $1::INT < 20
```

So any of the usages is cast, we will assume it's the data type.

src/query/src/planner.rs
sunng87: Here we are using `opt_type` to override `existing` no matter `opt_type` is none or not.

This is the case when datafusion fails to infer parameter type, we will fallback to cast type if any.

sunng87:
> maybe add a sqlness test?

@discord9 we cannot simulate this in sqlness because 1. we may not have extended query in sqlness 2. most client will send parameter types that doesn't rely on the server to infer it type.

We have a test case in postgres-adbc to cover this. https://github.com/GreptimeTeam/greptimedb-tests/pull/7

PR #7727, opened 2026-02-26
feat: pull meta config in frontend during startup

shuiyisong:

This PR adds the initial mechanism for the Frontend to ask and pull configs from the Metasrv during startup.
This allows dispatches and overwrites of the configs from the Metasrv.

TODO: merge the PR in the proto repo first

## PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5725,7 +5725,7 @@ dependencies = [
 [[package]]
 name = "greptime-proto"
 version = "0.1.0"
-source = "git+https://github.com/shuiyisong/greptime-proto.git?rev=a7c1498725264946a772cdb0e79d6323aac19392#a7c1498725264946a772cdb0e79d6323aac19392"
+source = "git+https://github.com/GreptimeTeam/greptime-proto.git?rev=4010e25ea09bf0a3ef359b946376da67d0323c5d#4010e25ea09bf0a3ef359b946376da67d0323c5d"
 dependencies = [
  "prost 0.14.1",
  "prost-types 0.14.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -154,7 +154,7 @@ etcd-client = { version = "0.17", features = [
 fst = "0.4.7"
 futures = "0.3"
 futures-util = "0.3"
-greptime-proto = { git = "https://github.com/shuiyisong/greptime-proto.git", rev = "a7c1498725264946a772cdb0e79d6323aac19392" }
+greptime-proto = { git = "https://github.com/GreptimeTeam/greptime-proto.git", rev = "4010e25ea09bf0a3ef359b946376da67d0323c5d" }
 hex = "0.4"
 http = "1"
 humantime = "2.1"


src/common/options/src/lib.rs
```diff
@@ -14,3 +14,4 @@
 
 pub mod datanode;
 pub mod memory;
+pub mod meta_config;
```
WenyXu: prefer `plugin_options`

src/cmd/src/frontend.rs
```diff
@@ -370,6 +372,14 @@ impl StartCommand {
         .await
         .context(error::MetaClientInitSnafu)?;
 
+        let meta_config: Vec<PluginOptions> = meta_client
+            .pull_meta_config()
+            .await
+            .context(error::MetaClientInitSnafu)?;
+        process_meta_config(meta_config, &mut plugins)
```
WenyXu: Can we reuse `plugins::setup_frontend_plugins`?

src/cmd/src/frontend.rs
```diff
@@ -370,6 +372,14 @@ impl StartCommand {
         .await
         .context(error::MetaClientInitSnafu)?;
 
+        let meta_config: Vec<PluginOptions> = meta_client
+            .pull_meta_config()
+            .await
+            .context(error::MetaClientInitSnafu)?;
```
WenyXu, line 379: We can move `pull_meta_config()` before `plugins::setup_frontend_plugins`. We should also ensure that the pulled plugin options do not overlap with those configured in the frontend.

src/common/options/src/plugin_options.rs
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+/// A trait for serializing Metasrv config to a JSON string.
+/// So it can be used in the metasrv's crate instead of depending on the plugins' crate.
+pub trait PluginOptionsSerializer: Send + Sync {
+    fn serialize(&self) -> Result<Vec<String>, serde_json::Error>;
```
WenyXu: we can return `String` instead of `Vec<String>`

src/meta-client/src/client/heartbeat.rs
```diff
@@ -171,6 +174,12 @@ impl Client {
         inner.ask_leader().await?;
         inner.heartbeat().await
     }
+
+    pub async fn pull_meta_config(&self) -> Result<PullMetaConfigResponse> {
```
WenyXu: Document the public function

src/cmd/src/frontend.rs
shuiyisong: Not really, `setup_frontend_plugins` happens before the meta client is built, and `start_frontend_plugins` is too late since the frontend is already built.

src/cmd/src/frontend.rs
shuiyisong: Same above. The meta client is built after `setup_frontend_plugins` is invoked. The `plugins` has to been existing to build the meta client, which is the output of `setup_frontend_plugins`.

src/common/options/src/plugin_options.rs
shuiyisong: If we were to use `String` here, we have to wrap it with `Option` to indicate that the server doesn't return any options. While the `Vec` can simply **be empty** to indicate that.

WenyXu:
@codex review

src/meta-client/src/client.rs
```diff
@@ -584,6 +585,18 @@ impl MetaClient {
...
+    where
+        T: DeserializeOwned + Default,
+    {
+        let res = self.heartbeat_client()?.pull_config().await?;
+        if res.payload.is_empty() {
+            Ok(T::default())
+        } else {
+            serde_json::from_str(&res.payload).context(ConvertMetaConfigSnafu)
+        }
+    }
```
WenyXu:
```suggestion
    pub async fn pull_config<T, U>(&self, deserializer: T) -> Result<U>
    where
        T: PluginOptionsDeserializer<U>,
        U: DeserializeOwned,
    {
        let res = self.heartbeat_client()?.pull_config().await?;
        let v = deserializer
            .deserialize(&res.payload)
            .context(ConvertMetaConfigSnafu)?;
        Ok(v)
    }
```

src/common/options/src/plugin_options.rs
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+/// A trait for serializing Metasrv config to a JSON string.
+/// So it can be used in the metasrv's crate instead of depending on the plugins' crate.
+pub trait PluginOptionsSerializer: Send + Sync {
+    fn serialize(&self) -> Result<String, serde_json::Error>;
+}
```
WenyXu, line 23: We can also introduce a `Deserializer` and let `PluginOptions` handle serialization and deserialization through these interfaces. This would help improve maintainability.

```suggestion
/// A trait for serializing Metasrv config to a JSON string.
/// So it can be used in the metasrv's crate instead of depending on the plugins' crate.
pub trait PluginOptionsSerializer: Send + Sync {
    fn serialize(&self) -> Result<String, serde_json::Error>;
}

pub type PluginOptionsSerializerRef = Arc<dyn PluginOptionsSerializer>;

pub trait PluginOptionsDeserializer<T: DeserializeOwned>: Send + Sync {
    fn deserialize(&self, s: &str) -> Result<T, serde_json::Error>;
}
```

src/plugins/src/options.rs
```diff
@@ -21,3 +22,12 @@ pub struct DummyOptions;
...
 }
+
+#[allow(unused)]
+pub struct PluginOptionsList(pub Vec<PluginOptions>);
+
+impl PluginOptionsSerializer for PluginOptionsList {
+    fn serialize(&self) -> Result<String, serde_json::Error> {
+        serde_json::to_string(&self.0)
+    }
+}
```
WenyXu, line 47:
```suggestion
pub struct PluginOptionsList(pub Vec<PluginOptions>);

impl PluginOptionsSerializer for PluginOptionsList {
    fn serialize(&self) -> Result<String, serde_json::Error> {
        if self.0.is_empty() {
            Ok("".to_string())
        } else {
            serde_json::to_string(&self.0)
        }
    }
}

pub struct PluginOptionsDeserializerImpl;

impl PluginOptionsDeserializer<Vec<PluginOptions>> for PluginOptionsDeserializerImpl {
    fn deserialize(&self, s: &str) -> Result<Vec<PluginOptions>, serde_json::Error> {
        if s.is_empty() {
            Ok(vec![])
        } else {
            serde_json::from_str(s)
        }
    }
}
```

src/plugins/src/options.rs
```diff
@@ -21,3 +22,12 @@ pub struct DummyOptions;
 pub enum PluginOptions {
     Dummy(DummyOptions),
 }
+
+#[allow(unused)]
```
WenyXu: We can public the struct instead adding `#[allow(unused)]`

src/meta-client/src/client.rs
```diff
@@ -584,6 +585,18 @@ impl MetaClient {
         self.heartbeat_client()?.ask_leader().await
     }
 
+    pub async fn pull_config<T>(&self) -> Result<T>
+    where
+        T: DeserializeOwned + Default,
+    {
+        let res = self.heartbeat_client()?.pull_config().await?;
```
WenyXu: Nit: It works for me to add a pull_config interface to the heartbeat client. Any thoughts?
cc @fengjiachun @MichaelScofield

src/meta-client/src/client.rs
```diff
@@ -584,6 +585,18 @@ impl MetaClient {
...
+        if res.payload.is_empty() {
+            Ok(T::default())
+        } else {
+            serde_json::from_str(&res.payload).context(ConvertMetaConfigSnafu)
```
WenyXu: nit: `DeserializePluginOptions`

src/meta-client/src/client.rs
MichaelScofield: If we have to pull the plugin configuration from Metasrv, I don't think the heartbeat service is a good place to put this method. It has little relationship with heartbeat. Rather, make it a new service, or place it in the service "`Cluster`".

shuiyisong: Move the `pull_config` API to a new `config` service in https://github.com/GreptimeTeam/greptimedb/pull/7727/commits/211e577ec391c45b626d46484581f8cf8ae92f10, PTAL

PR #7733, opened 2026-02-28
feat: gc batch delete files

discord9:

previously it delete one by one and is somewhere slow, so batch delete to try to speed up(if storage support)

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/file.rs
+++ b/src/mito2/src/sst/file.rs
@@ -23,11 +23,9 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use common_base::readable_size::ReadableSize;
 use common_telemetry::{debug, error};
 use common_time::Timestamp;
-use itertools::Itertools;
 use partition::expr::PartitionExpr;
 use serde::{Deserialize, Serialize};
 use smallvec::SmallVec;
-use snafu::ResultExt as _;
 use store_api::metadata::ColumnMetadata;
 use store_api::region_request::PathType;
 use store_api::storage::{ColumnId, FileId, IndexVersion, RegionId};
@@ -35,7 +33,6 @@ use store_api::storage::{ColumnId, FileId, IndexVersion, RegionId};
 use crate::access_layer::AccessLayerRef;
 use crate::cache::CacheManagerRef;
 use crate::cache::file_cache::{FileType, IndexKey};
-use crate::error::{DeleteIndexesSnafu, DeleteSstsSnafu};
 use crate::sst::file_purger::FilePurgerRef;
 use crate::sst::location;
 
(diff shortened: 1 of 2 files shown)


src/mito2/src/sst/file.rs
```diff
@@ -600,21 +654,95 @@ pub async fn delete_index(
...
+                for index_id in index_ids {
+                    delete_index_and_purge(*index_id, access_layer, cache_manager).await?;
+                }
+
+                if let Err(e) = deleter.close().await {
+                    error!(e; "Failed to close object store deleter after fallback index deletion");
+                }
+
+                return Ok(());
+            }
```
v0y4g3r: Why we simply ignore the error returned by `deleter.close()` here , but propagate the error in line 607?

src/mito2/src/sst/file.rs
```diff
@@ -554,31 +557,58 @@ pub async fn delete_files(
...
+        paths.push(location::sst_file_path(
+            table_dir,
+            region_file_id,
+            path_type,
+        ));
+        for version in 0..=*index_version {
+            let index_id = RegionIndexId::new(region_file_id, version);
+            paths.push(location::index_file_path(table_dir, index_id, path_type));
+        }
+    }
```
WenyXu: Nit: Maybe we can extract the path-building logic into a separate function and reuse it in both `delete_files` and `delete_indexes`.

WenyXu (approved):
Rest LGTM

src/mito2/src/sst/file.rs
```diff
@@ -554,31 +557,52 @@ pub async fn delete_files(
...
+        .context(DeleteSstsSnafu {
+            region_id,
+            file_ids: attempted_files.clone(),
+        })?;
+    deleter
+        .delete_iter(paths.iter().map(String::as_str))
+        .await
+        .context(DeleteSstsSnafu {
+            region_id,
+            file_ids: file_ids.iter().map(|f| f.0).collect_vec(),
```
WenyXu:
```suggestion
            file_ids: attempted_files.clone(),
```

src/mito2/src/sst/file.rs
```diff
@@ -554,31 +557,52 @@ pub async fn delete_files(
...
+            &path_type,
+            (0..=*index_version).map(|version| RegionIndexId::new(region_file_id, version)),
+        );
     }
 
+    let mut deleter = access_layer
+        .object_store()
+        .deleter()
+        .await
+        .context(DeleteSstsSnafu {
```
WenyXu:
```suggestion
        .with_context(DeleteSstsSnafu {
```

src/mito2/src/sst/file.rs
```diff
@@ -554,31 +557,52 @@ pub async fn delete_files(
...
+        paths.push(location::sst_file_path(
+            table_dir,
+            region_file_id,
+            path_type,
+        ));
+        push_index_paths(
+            &mut paths,
+            table_dir,
+            &path_type,
+            (0..=*index_version).map(|version| RegionIndexId::new(region_file_id, version)),
```
evenyag, line 564: I think we should provide a batch delete_ssts method in the AcccessLayer instead of letting the user assemble delete logic.

Issue #7734, opened 2026-02-28
MetaSrv: PostgreSQL KV backend fails over Unix socket due to TLS handling (Disable mode error / InvalidDnsNameError)


LER0ever:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Metasrv

### Minimal reproduce step

1. Prepare PostgreSQL running locally with Unix socket enabled (e.g. socket dir /run/postgresql).
2. Start MetaSrv via Docker/Compose and mount the socket directory into container, e.g. /run/postgresql:/var/run/postgresql.
3. Use PostgreSQL KV backend with a libpq-style Unix socket URL:

```
backend = "postgres_store"
store_addrs = ["host=/var/run/postgresql dbname=greptime user=greptime password=secret"]
meta_table_name = "greptime_metakv"
auto_create_schema = true
```

4. Case A: set [backend_tls] mode = "disable" and start MetaSrv.
5. Case B: remove [backend_tls] (default is Prefer) and start MetaSrv.

---

### Summary

Using MetaSrv with PostgreSQL as the KV backend through Unix Domain Socket reveals **two separate TLS handling bugs**:

1. **`[backend_tls] mode = "disable"` still errors**
   MetaSrv tries to create a TLS connector and fails with: `Cannot create TLS connector for Disable mode` (instead of using `NoTls`).

2. **Default `Prefer` mode fails for Unix domain socket connections**
   With `host=/var/run/postgresql`, rustls treats the socket path as a DNS name for SNI/server-name validation and fails with `InvalidDnsNameError`.

I’m intentionally using PostgreSQL over a Unix domain socket to avoid the TCP/IP (loopback) network stack when MetaSrv and PostgreSQL are co-located. This reduces latency and CPU overhead compared to localhost TCP, and the connection security is already enforced by filesystem permissions on the socket.

### Environment

* GreptimeDB: v1.0.0-rc.1 (`greptime/greptimedb:v1.0.0-rc.1`)
* PostgreSQL: 17.x (Unix socket at `/run/postgresql`, mounted into container as `/var/run/postgresql`)

#### Case 1: explicitly disable TLS

metasrv.toml:

```toml
backend = "postgres_store"
store_addrs = ["host=/var/run/postgresql dbname=greptime user=greptime password=secret"]
meta_table_name = "greptime_metakv"
auto_create_schema = true

[backend_tls]
mode = "disable"
```

Observed:

```text
[...]

PR #7735, opened 2026-02-28
fix: handle TLS correctly for PostgreSQL Unix socket connections

LER0ever:
The `create_postgres_pool()` function unconditionally creates a TLS connector whenever `backend_tls` config is present, causing two failures:

1. Default TLS mode (Prefer) on Unix sockets triggers InvalidDnsNameError as rustls parses the socket path as an SNI hostname.
2. Explicit `mode = "disable"` errors because `create_postgres_tls_connector` rejects the Disable variant outright.

Fix by adding mode-aware branching (matching the etcd pattern from #7507):
- Disable mode uses NoTls directly.
- Unix socket connections skip TLS with a warning log.
- TCP connections with TLS enabled retain existing behavior.

Fixes: #7734 

Summary of changes:

- Update meta-srv/src/utils/postgres.rs::create_postgres_pool() to branch on backend_tls.mode and whether the connection URL indicates a Unix socket.
- Add a helper to detect PostgreSQL Unix socket URLs and add unit tests for the detection logic.
- Add a defensive guard in common_meta::kv_backend::rds::postgres::PgStore::with_url_and_tls() so Unix socket + TLS config won’t hit the same failure if that path is used in the future.

Why needed:

- backend_tls.mode="disable" should reliably disable TLS and use NoTls.
- Unix socket URLs cannot be used as a valid rustls server name for SNI/verification; attempting TLS here fails with InvalidDnsNameError.
- This enables the common co-located deployment pattern where MetaSrv talks to PostgreSQL over Unix sockets for lower overhead.

Compatibility:

- No API changes.
- No schema/data changes.
- Behavior changes only affect PostgreSQL KV backend connection establishment in the above edge cases; TCP+TLS behavior remains unchanged.

## PR Checklist

### DIFF
--- a/src/common/meta/src/kv_backend/rds/postgres.rs
+++ b/src/common/meta/src/kv_backend/rds/postgres.rs
@@ -404,6 +404,22 @@ impl ExecutorFactory<PgClient> for PgExecutorFactory {
 /// It uses [deadpool_postgres::Pool] as the connection pool for [RdsStore].
 pub type PgStore = RdsStore<PgClient, PgExecutorFactory, PgSqlTemplateSet>;
 
+/// Returns `true` if the PostgreSQL connection URL indicates a Unix socket connection.
+///
+/// Uses [`tokio_postgres::Config`] to properly parse the connection string (both
+/// key-value and URI formats) and check whether any host is a Unix socket path.
+pub fn is_pg_unix_socket(url: &str) -> bool {
+    use tokio_postgres::config::Host;
+
+    let Ok(config) = url.parse::<tokio_postgres::Config>() else {
+        return false;
+    };
+    config
+        .get_hosts()
+        .iter()
+        .any(|h| matches!(h, Host::Unix(_)))
+}
+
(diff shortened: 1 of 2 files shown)


sunng87:
In real world, using postgres with domain socket makes little sense because we need it as a centralized data store for metadata storage, as well as leader election. Local postgres won't work with multiple metasrv instances.

LER0ever:
Thanks @sunng87. That’s fair, and I understand your point.

My current use case is a single-machine deployment using the clustered Docker Compose setup for local/self-hosted testing. In that environment, PostgreSQL is co-located with MetaSrv, so Unix domain sockets are still useful to me in practice. I also haven’t needed to scale this setup out to multiple MetaSrv nodes yet, so that limitation hasn’t been an issue for me so far.

I opened this mainly because it fixes a real issue in that setup. If this use case feels too niche for upstream, feel free to close the PR.

WenyXu:
I’d prefer not to add this feature, since it isn’t a best practice in real-world applications.

PR #7738, opened 2026-03-01
chore(version): refresh build info on demand

v0y4g3r:

This PR optimizes the build process to improve incremental compilation performance and fixes a missing dependency in the CI multi-language tests.

  1. Optimize Incremental Compilation
  The build script in src/common/version/build.rs was previously refreshing build information (like timestamps) on every build. This caused shadow.rs to be regenerated even when no code changes occurred, invalidating incremental compilation.
   - Changes: 
       - Build information is now only refreshed by default in release builds.
       - Added DISABLE_BUILD_INFO environment variable to explicitly disable refreshing even in release builds.
       - In non-refresh scenarios, if shadow.rs already exists, the script skips regeneration entirely, preserving incremental build artifacts.

  2. Fix CI Multi-language Tests
  The run-multi-lang-tests.yml workflow was failing due to a missing uv installation, which is required for certain database operations.
   - Changes: Added a "Setup uv" step using astral-sh/setup-uv@v3 to the workflow.

## PR Checklist

### DIFF
--- a/.github/actions/build-linux-artifacts/action.yml
+++ b/.github/actions/build-linux-artifacts/action.yml
@@ -56,7 +56,7 @@ runs:
       uses: ./.github/actions/build-greptime-binary
       with:
         base-image: ubuntu
-        features: servers/dashboard,common-version/refresh-build-info
+        features: servers/dashboard
         cargo-profile: ${{ inputs.cargo-profile }}
         artifacts-dir: greptime-linux-${{ inputs.arch }}-${{ inputs.version }}
         version: ${{ inputs.version }}
@@ -75,7 +75,7 @@ runs:
       if: ${{ inputs.arch == 'amd64' && inputs.dev-mode == 'false' }} # Builds greptime for centos if the host machine is amd64.
       with:
         base-image: centos
-        features: servers/dashboard,common-version/refresh-build-info
+        features: servers/dashboard
         cargo-profile: ${{ inputs.cargo-profile }}
         artifacts-dir: greptime-linux-${{ inputs.arch }}-centos-${{ inputs.version }}
         version: ${{ inputs.version }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -219,10 +219,10 @@ jobs:
         include:
           - os: ${{ needs.allocate-runners.outputs.macos-runner }}
             arch: aarch64-apple-darwin
-            features: servers/dashboard,common-version/refresh-build-info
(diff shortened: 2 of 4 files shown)


sunng87:
Can we simplify this to only refresh the build info in release mode

v0y4g3r:
> Can we simplify this to only refresh the build info in release mode

~Looks reasonable, I will give it a try~

v0y4g3r:
> Can we simplify this to only refresh the build info in release mode

In some development cases we need to frequently rebuild release binaries but build time is not necessary. I think use a feature to opt-in build info refresh would be more convenient.

sunng87:
> In some development cases we need to frequently rebuild release binaries but build time is not necessary. I think use a feature to opt-in build info refresh would be more convenient.

we can make this opt-in for debug mode, and required for release mode, with which:

1. we won't need to add new features to github actions in this repo and downstream repos
2. we will benefit from improved build time for local development

v0y4g3r:
> > In some development cases we need to frequently rebuild release binaries but build time is not necessary. I think use a feature to opt-in build info refresh would be more convenient.
> 
> we can make this opt-in for debug mode, and required for release mode, with which:
> 
> 1. we won't need to add new features to github actions in this repo and downstream repos
> 2. we will benefit from improved build time for local development

Maybe also introduce an env var to disable build info refresh so that users can still have a choice when build info does not matter?

sunng87:
> Maybe also introduce an env var to disable build info refresh so that users can still have a choice when build info does not matter?

I guess it's unnecessary because release mode build is infrequent and already slow enough to notice the difference. Also if we disabled it by mistake it might bring us trouble with identifying a build in prod environment.

v0y4g3r:
> > Maybe also introduce an env var to disable build info refresh so that users can still have a choice when build info does not matter?
> 
> I guess it's unnecessary because release mode build is infrequent and already slow enough to notice the difference. Also if we disabled it by mistake it might bring us trouble with identifying a build in prod environment.

It will bring distinct build time difference when running benchmarks.
```
$ cargo bench -p servers --bench bench_prom -- decode_prom_request
...
    Finished `bench` profile [optimized + debuginfo] target(s) in 1m 47s
```

And it also holds for building release binaries from common-version vs from what has changed.

sunng87:
Should be fine. It uses `bench` profile for `cargo bench`. We can skip it. Overall, we will try to avoid rebuild common-version for iterative tests like cargo test, cargo build and cargo bench

PR #7750, opened 2026-03-04
ci: add nightly jsonbench test

MichaelScofield:
as title

### DIFF
--- a/.github/workflows/nightly-jsonbench.yaml
+++ b/.github/workflows/nightly-jsonbench.yaml
@@ -1,0 +1,162 @@
+name: Nightly JSONBench
+
+on:
+  schedule:
+    # Trigger at 00:00(Asia/Shanghai) on every weekday.
+    - cron: "0 16 * * 0-4"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  allocate-runner:
+    name: Allocate runner
+    if: ${{ github.repository == 'GreptimeTeam/greptimedb' }}
+    runs-on: ubuntu-latest
+    outputs:
+      linux-arm64-runner: ${{ steps.start-linux-arm64-runner.outputs.label }}
+
+      # The following EC2 resource id will be used for resource releasing.
+      linux-arm64-ec2-runner-label: ${{ steps.start-linux-arm64-runner.outputs.label }}
+      linux-arm64-ec2-runner-instance-id: ${{ steps.start-linux-arm64-runner.outputs.ec2-instance-id }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Allocate Linux ARM64 runner
+        uses: ./.github/actions/start-runner
+        id: start-linux-arm64-runner
+        with:
+          runner: ${{ vars.DEFAULT_ARM64_RUNNER }}
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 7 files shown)


MichaelScofield:
@gemini-code-assist review
@codex review

.github/workflows/nightly-jsonbench.yaml
```diff
@@ -0,0 +1,105 @@
+name: Nightly JSONBench
```
sunng87, line 1: we can name this as performance CI or something.

And using github shared runner for performance testing is not reliable, we'd better use our self-hosted runner.

sunng87:
add @fengjiachun to see if this will fit into our performance CI, or can we use this as the start point.

tests/cases/standalone/common/types/json/json2.result
```diff
@@ -151,6 +151,14 @@ select j.c, j.y from json2_table order by ts;
 |                                   | false                             |
 +-----------------------------------+-----------------------------------+
 
+select j from json2_table order by ts;
+
+Error: 3001(EngineExecuteQuery), Failed to align JSON array, reason: Invalid argument error: use StructArray::try_new_with_length or StructArray::new_empty_fields to create a struct array with no fields so that the length can be set correctly
+
+select * from json2_table order by ts;
+
+Error: 3001(EngineExecuteQuery), Failed to align JSON array, reason: Invalid argument error: use StructArray::try_new_with_length or StructArray::new_empty_fields to create a struct array with no fields so that the length can be set correctly
```
sunng87, line 160: is this error expected?

MichaelScofield: Yes, see [this](https://greptime.feishu.cn/wiki/O24Bwaohsi5DLRkyiXicxEZinOd#share-FRcodFIiXofCUdxaegwcWgIEnx7). It's not obvious in DataFusion, so I haven't started doing it.

.github/workflows/nightly-jsonbench.yaml
MichaelScofield: This jsonbench test can run very long time, self-hosted runner costs money.

sunng87: it's ok. we need fixed/reliable environment for reference. The shared compute environment just cannot tell the real results

MichaelScofield: I've changed it to our ec2 runner. I will alter this workflow to adapt our performance ci in the future.

MichaelScofield:
@sunng87 PTAL

src/datatypes/src/json/value.rs
```diff
@@ -623,35 +620,55 @@ pub enum JsonVariantRef<'a> {
...
+}
+
+fn json_object_native_type<I, K>(fields: I) -> JsonNativeType
+where
+    I: IntoIterator<Item = (K, JsonNativeType)>,
+    K: Into<String>,
+{
+    let mut fields = fields.into_iter().peekable();
+    if fields.peek().is_none() {
+        JsonNativeType::Null
```
sunng87, line 669: empty object is not null

src/datatypes/src/json.rs
```diff
@@ -305,33 +305,48 @@ fn encode_json_array_with_context<'a>(
...
+        }
+        Some(merged)
+    } else {
+        None
+    };
+    let unified_item_type = match (provided_item_type, merged_item_type) {
+        (Some(mut x), Some(y)) => {
+            x.merge(&y)?;
+            Some(x)
+        }
```
sunng87, line 337: _ => provided_item_type.or(merged_item_type)

sunng87 (approved):

src/datatypes/src/json/value.rs
MichaelScofield: Yes but Parquet does not allow writing empty Struct Array. If we don't store the empty object field, it's worse than storing it as Null Array; if we use a dummy value, it's not elegant. In practice, treating empty object as null is not big deal.

Issue #7732, opened 2026-02-27
Unify read paths for scanners in mito


evenyag:
### What type of enhancement is this?

Refactor, Tech debt reduction

### What does the enhancement do?

The mito engine has two parallel read pipelines:
1. **Old (primary_key format)**: `Batch` → `Source` → `MergeReader` → `DedupReader` → `ConvertBatchStream` → output
2. **New (flat format)**: `RecordBatch` → `FlatMergeReader` → `FlatDedupReader` → `ConvertBatchStream` → output

We want to remove the old read pipeline to reduce code duplication and maintenance burden, while **keeping the ability to write primary-key-format parquet files**.

This also reduces the work to refactor or optimize the scanner, as we don't have to support two different paths.

### Implementation challenges

The flat format read path already supports reading primary_key format SSTs. So we only need to bridge the old memtables to the new read path.

We need to support the last row cache and extension range in flat format to ensure we don't miss features.

Implementation history:
- [ ] Extension range support in flat format https://github.com/GreptimeTeam/greptimedb/pull/7730
- [ ] Last row cache support in flat format
- [ ] Bridge old memtables to flat format
- [ ] Support writing primary-key-format parquet from FlatSource
- [ ] Migrate flush/compact path to always use FlatSource
- [ ] Migrate query scan paths to always use flat
- [ ] Remove code

evenyag:
We have done most of the work. The remaining part is removing unused code.

PR #7757, opened 2026-03-04
feat: implement last row cache reader for flat format

evenyag:

https://github.com/GreptimeTeam/greptimedb/issues/7732

- Add `FlatRowGroupLastRowCachedReader` and `FlatLastTimestampSelector` to support last-row selection
and caching for the flat SST format, mirroring the existing primary-key format cache
- Extend `SelectorResultValue` with a `SelectorResult` enum to store either primary-key `Batch` or
flat `RecordBatch` cache entries
- Wire the flat last-row reader into the file range read path, enabling it when the row group is
PUT-only and all rows are selected

TSBS lastpoint without cache:
```
min:  2087.04ms, med:  2094.78ms, mean:  2111.82ms, max: 2264.96ms, stddev:    51.22ms, sum:  21.1sec, count: 10
all queries             :
min:  2087.04ms, med:  2094.78ms, mean:  2111.82ms, max: 2264.96ms, stddev:    51.22ms, sum:  21.1sec, count: 10
```

with cache
```
min:   385.46ms, med:   395.38ms, mean:   579.50ms, max: 2256.26ms, stddev:   558.91ms, sum:   5.8sec, count: 10
all queries             :
min:   385.46ms, med:   395.38ms, mean:   579.50ms, max: 2256.26ms, stddev:   558.91ms, sum:   5.8sec, count: 10
```

### Details
The existing last-row cache (`RowGroupLastRowCachedReader`) only works with the primary-key SST
format. This PR adds equivalent support for the flat format:

- **`FlatLastTimestampSelector`**: Processes sorted flat `RecordBatch`es, tracking primary key
boundaries to emit only the row with the latest timestamp per series. Handles primary keys spanning
multiple batches and deduplicates timestamps across batch boundaries.
- **`FlatRowGroupLastRowCachedReader`**: Cache-aware reader that checks for cached flat results on
hit, or reads through `FlatRowGroupReader` and populates the cache on miss.
- **`SelectorResult` enum**: The cache value type now distinguishes between `PrimaryKey(Vec<Batch>)`
and `Flat(Vec<RecordBatch>)` variants. Cache lookups validate both format and projection match before
returning hits.
- **`FileRange::flat_reader`** now accepts an optional `TimeSeriesRowSelector` and creates the
last-row reader when applicable (PUT-only, all rows selected).
- Primary key arrays are compacted during selection to reduce memory usage in cached results.

### Future Work
The ScanHint rule will ensure it only passes the last row hint when there are only tag filters. So the last row cache doesn't consider field filters now. We can support a more flexible last row cache in the future.

## PR Checklist

### DIFF
--- a/src/mito2/src/read/last_row.rs
+++ b/src/mito2/src/read/last_row.rs
@@ -167,7 +167,7 @@ impl LastRowCacheReader {
     async fn next_batch(&mut self) -> Result<Option<Batch>> {
         let batches = match &self.value.result {
             SelectorResult::PrimaryKey(batches) => batches,
-            SelectorResult::Flat(_) => return Ok(None),
+            SelectorResult::Flat(_) => unreachable!(),
         };
         if self.idx < batches.len() {
             let res = Ok(Some(batches[self.idx].clone()));
@@ -331,14 +331,6 @@ impl FlatRowGroupLastRowCachedReader {
         }
     }
 
-    /// Gets the underlying reader metrics if uncached.
-    pub(crate) fn metrics(&self) -> Option<&ReaderMetrics> {
-        match self {
-            FlatRowGroupLastRowCachedReader::Hit(_) => None,
-            FlatRowGroupLastRowCachedReader::Miss(reader) => Some(&reader.metrics),
-        }
-    }
-
     /// Returns the next RecordBatch.
     pub(crate) fn next_batch(&mut self) -> Result<Option<RecordBatch>> {
         match self {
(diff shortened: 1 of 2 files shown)


src/mito2/src/read/last_row.rs
```diff
@@ -156,8 +165,12 @@ pub(crate) struct LastRowCacheReader {
 impl LastRowCacheReader {
     /// Iterates cached last rows.
     async fn next_batch(&mut self) -> Result<Option<Batch>> {
-        if self.idx < self.value.result.len() {
-            let res = Ok(Some(self.value.result[self.idx].clone()));
+        let batches = match &self.value.result {
+            SelectorResult::PrimaryKey(batches) => batches,
+            SelectorResult::Flat(_) => return Ok(None),
```
discord9: is `unreachable()` acceptable here?

src/mito2/src/read/prune.rs
discord9, line 313: is this count twice with?
```src/mito2/src/read/last_row.rs:484
fn next_batch(&mut self) -> Result<Option<RecordBatch>> {
        if self.pending.is_full() {
            return self.flush_pending();
        }

        while let Some(batch) = self.reader.next_batch()? {
            self.metrics.num_rows += batch.num_rows();
            self.metrics.num_batches += 1;
            self.selector.on_next(batch, &mut self.pending)?;
            if self.pending.is_full() {
                return self.flush_pending();
            }
        }
```

PR #7761, opened 2026-03-05
perf(remote-write): optimize decode prom 

v0y4g3r:

Optimize decoding of Prometheus Remote Write requests:
- Label name validation: replace UTF-8 validation by table look up and loop unrolling to conform to the [spec](https://prometheus.io/docs/specs/prw/remote_write_spec/#labels)
- Replace Bytes clone with `&'static [u8]` since we can ensure the original buffer outlives the decoded data 

Improve strict validation by ~10%

```
decode_prom_request/validation_mode/Strict
                        time:   [604.45 µs 607.40 µs 611.08 µs]
                        change: [-10.818% -10.502% -10.102%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  11 (11.00%) high severe
Benchmarking decode_prom_request/validation_mode/Lossy: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 3.0s. You may wish to increase target time to 3.2s, enable flat sampling, or reduce sample count to 60.
decode_prom_request/validation_mode/Lossy
                        time:   [621.12 µs 621.57 µs 622.05 µs]
                        change: [-5.5518% -5.4335% -5.3087%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
decode_prom_request/validation_mode/Unchecked
                        time:   [575.40 µs 576.41 µs 577.60 µs]
                        change: [-5.2507% -5.0604% -4.8747%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
```

## PR Checklist

### DIFF
--- a/src/servers/src/prom_row_builder.rs
+++ b/src/servers/src/prom_row_builder.rs
@@ -23,7 +23,7 @@ use common_query::prelude::{greptime_timestamp, greptime_value};
 use pipeline::{ContextOpt, ContextReq};
 use prost::DecodeError;
 
-use crate::http::{validate_label_name, PromValidationMode};
+use crate::http::{PromValidationMode, validate_label_name};
 use crate::proto::PromLabel;
 use crate::repeated_field::Clear;
 
@@ -228,8 +228,8 @@ impl<'a> TableBuilder<'a> {
 #[cfg(test)]
 mod tests {
     use api::prom_store::remote::Sample;
-    use api::v1::value::ValueData;
     use api::v1::Value;
+    use api::v1::value::ValueData;
     use prost::DecodeError;
 
     use crate::http::PromValidationMode;


src/servers/src/http.rs
```diff
@@ -176,8 +176,62 @@ pub enum PromValidationMode {
...
+        return false;
+    }
+    let first = name[0];
+    if !(first.is_ascii_alphabetic() || first == b'_') {
+        return false;
+    }
+
+    let mut rest = &name[1..];
+    while rest.len() >= 8 {
+        let res = IS_VALID_LABEL_REST[rest[0] as usize] as u8
```
MichaelScofield, line 207: Nit: will it be faster if the `IS_VALID_LABEL_REST` is stored as 4 uint64 (each bit corresponds to the boolean in the original array)? So that no cast (`rest[i] as usize`) is needed, and some bit-shifting calculation is sufficient Instead of array lookup.

src/servers/src/prom_row_builder.rs
```diff
@@ -100,7 +100,7 @@ pub struct TableBuilder {
     /// Rows written.
     rows: Vec<Row>,
     /// Indices of columns inside `schema`.
-    col_indexes: HashMap<Vec<u8>, usize>,
+    col_indexes: HashMap<RawBytes, usize>,
```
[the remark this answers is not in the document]
MichaelScofield: Seems gemini has a point here. WDYT @v0y4g3r

src/servers/src/proto.rs
```diff
@@ -237,13 +237,15 @@ impl PromTimeSeries {
 
 #[derive(Default, Debug)]
 pub struct PromWriteRequest {
+    raw_data: Bytes,
```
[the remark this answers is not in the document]
v0y4g3r: fixed in 8df45f1cb9

src/servers/src/prom_row_builder.rs
```diff
@@ -153,14 +166,14 @@ impl TableBuilder {
                 row[*e].value_data = tag_value;
                 continue;
             }
-            let tag_name = prom_validation_mode.decode_string(raw_tag_name)?;
+            let tag_name = prom_validation_mode.decode_label_name(raw_tag_name)?;
```
evenyag: We have validated the name. Should we convert it to string directly?

v0y4g3r: fixed in d1d5110a328534f54d167ca0431e707e6cb1e82e

src/servers/src/prom_row_builder.rs
```diff
@@ -142,7 +142,12 @@ impl TableBuilder {
         let mut row = vec![Value { value_data: None }; self.col_indexes.len()];
 
         for PromLabel { name, value } in labels {
-            prom_validation_mode.validate_bytes(name)?;
+            if !validate_label_name(name) {
+                return Err(DecodeError::new(format!(
+                    "Invalid label name: `{}`",
+                    String::from_utf8_lossy(name)
+                )));
+            }
```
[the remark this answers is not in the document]
v0y4g3r: fixed in d1d5110a328534f54d167ca0431e707e6cb1e82e

src/servers/src/http.rs
v0y4g3r: Did a benchmark:
```rust
const IS_VALID_LABEL_REST_U64: [u64; 4] = {
    let mut table = [0u64; 4];
    let mut i = 0;
    while i < 256 {
        let b = i as u8;
        if b.is_ascii_alphanumeric() || b == b'_' {
            table[i / 64] |= 1 << (i % 64);
        }
        i += 1;
    }
    table
};

#[inline]
pub fn validate_label_name_bitmask(name: &[u8]) -> bool {
    if name.is_empty() { return false; }
    let first = name[0];
    if !(first.is_ascii_alphabetic() || first == b'_') { return false; }

    let mut rest = &name[1..];
    while rest.len() >= 8 {
        let res = ((IS_VALID_LABEL_REST_U64[(rest[0] >> 6) as usize] >> (rest[0] & 63)) & 1)
            & ((IS_VALID_LABEL_REST_U64[(rest[1] >> 6) as usize] >> (rest[1] & 63)) & 1)
            & ((IS_VALID_LABEL_REST_U64[(rest[2] >> 6) as usize] >> (rest[2] & 63)) & 1)
            & ((IS_VALID_LABEL_REST_U64[(rest[3] >> 6) as usize] >> (rest[3] & 63)) & 1)
            & ((IS_VALID_LABEL_REST_U64[(rest[4] >> 6) as usize] >> (rest[4] & 63)) & 1)
            & ((IS_VALID_LABEL_REST_U64[(rest[5] >> 6) as usize] >> (rest[5] & 63)) & 1)
            & ((IS_VALID_LABEL_REST_U64[(rest[6] >> 6) as usize] >> (rest[6] & 63)) & 1)
            & ((IS_VALID_LABEL_REST_U64[(rest[7] >> 6) as usize] >> (rest[7] & 63)) & 1);

        if res == 0 { return false; }
        rest = &rest[8..];
    }
    for &b in rest {
        if ((IS_VALID_LABEL_REST_U64[(b >> 6) as usize] >> (b & 63)) & 1) == 0 {
            return false;
        }
    }
    true
}
```
Result: 
```
┌───────────────────────┬──────────────────────────┬─────────────────────────────┐
│ Benchmark Case        │ Table Approach (Current) │ Bitmask Approach (Proposed) │
├───────────────────────┼──────────────────────────┼─────────────────────────────┤
│ valid_short (8 bytes) │ 1.69 ns                  │ 2.70 ns                     │
│ valid_long (60 bytes) │ 6.88 ns                  │ 10.90 ns                    │
│ invalid_short         │ 0.86 ns                  │ 1.66 ns                     │
│ invalid_long          │ 4.86 ns                  │ 6.90 ns                     │
└───────────────────────┴──────────────────────────┴─────────────────────────────┘
```

The core problem of bitmask is it requires calculating the `u64` index and the bit offset within that `u64`. Even with optimized bit manipulation, this involves more CPU cycles per byte than a simple table lookup.
As for current implementation, the size of `[bool; 256]` and `[u64; 32]` is negligible because they both fits into L1 cache line. The cast from `u8` to `u64` is eliminated in compile time.

v0y4g3r:
@gemini-code-assist  review

v0y4g3r:
@gemini-code-assist review

PR #7777, opened 2026-03-09
fix(mito2): avoid parquet redownload when write cache already contains file

WenyXu:

This PR prevents duplicate parquet downloads in the write-cache refill path and adds focused test coverage for the new behavior.

### Summary

- Added an idempotent write-cache API:
  - `WriteCache::download_if_absent(...) -> Result<bool>`
  - returns `Ok(false)` when the target parquet is already in write cache, and skips remote download
  - returns `Ok(true)` when download is performed
- Wired manifest edit path to use the idempotent API:
  - in `worker/handle_manifest.rs`, parquet download now uses `download_if_absent` instead of unconditional `download`
  - keeps parquet metadata warmup after successful cache availability
  - only emits `on_file_cache_filled` when a real download happened in this invocation
- Added unit tests for both cache-hit and cache-miss flows:
  - `test_download_if_absent_skips_when_cached`
  - `test_download_if_absent_downloads_when_missing`
- Added a debug log for the cache-hit skip path to improve observability.

## PR Checklist

### DIFF
--- a/src/mito2/src/cache/write_cache.rs
+++ b/src/mito2/src/cache/write_cache.rs
@@ -323,6 +323,10 @@ impl WriteCache {
             .await
     }
 
+    /// Downloads the target file into write cache only when it is not cached.
+    ///
+    /// Returns `Ok(true)` if this call performs a download, or `Ok(false)` if the
+    /// file is already present in write cache and download is skipped.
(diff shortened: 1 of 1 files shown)


WenyXu:
@codex review

PR #7779, opened 2026-03-09
feat(wal): add NATS JetStream WAL backend

jpikoulas:
## Summary

Adds NATS JetStream as a new WAL provider complementing the existing Kafka and RaftEngine backends.

## Motivation

NATS JetStream is significantly more lightweight than Kafka and better suited for small and edge cluster deployments. It enables a fully distributed WAL without operating a full Kafka cluster.

## Architecture

- One JetStream stream per cluster; subjects serve as WAL shards (analogous to Kafka topics).
- `NatsProvider` wraps a subject string, used the same way as `KafkaProvider`.
- `NatsSubjectAllocator` (metasrv): assigns subjects round-robin from a fixed pool; no network calls needed at allocation time since NATS subjects are virtual.
- `NatsBatchProducer`: per-subject background Tokio task that serialises publishes and returns committed JetStream sequence numbers.
- `PurgeWorker`: background task that calls `stream.purge()` for compacted WAL entries at a configurable interval.
- Large entries are split into First/Middle/Last fragments using the same convention as the Kafka WAL.
- TLS and all NATS auth methods (NKey, credentials file, user/password) are supported.

## Configuration example

```toml
# datanode.toml
[wal]
provider = "nats_jetstream"
servers = ["nats://nats.nats.svc.cluster.local:4222"]
cluster_name = "greptimedb"
num_subjects = 64
num_replicas = 3

# metasrv.toml
[wal]
provider = "nats_jetstream"
servers = ["nats://nats.nats.svc.cluster.local:4222"]
num_subjects = 64
```

## Production status

Running on a Ryzen 7 Linux machines cluster with 22M+ WAL entries. Fully backward compatible — existing Kafka and RaftEngine deployments are unaffected.

## Test plan

- [ ] Unit tests in `src/log-store/src/nats/record.rs` and `src/log-store/src/nats/util.rs` pass without a NATS server
- [ ] Integration tests in `src/log-store/src/nats/log_store.rs` can be run with `GT_NATS_SERVERS=nats://localhost:4222 cargo nextest run`
- [ ] Config serde roundtrip test in `src/common/wal/src/config/nats.rs` passes
- [ ] `WalOptions::NatsJetstream` serde roundtrip test in `src/common/wal/src/options.rs` passes

### DIFF
--- a/src/log-store/src/nats/log_store.rs
+++ b/src/log-store/src/nats/log_store.rs
@@ -1,0 +1,433 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! [`NatsLogStore`] — full [`LogStore`] implementation backed by NATS JetStream.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use common_telemetry::{debug, info};
+use common_wal::config::nats::DatanodeNatsConfig;
+use dashmap::DashMap;
+use futures::future::try_join_all;
+use snafu::OptionExt;
+use store_api::logstore::entry::{Entry, Id as EntryId};
+use store_api::logstore::provider::{NatsProvider, Provider};
+use store_api::logstore::{AppendBatchResponse, LogStore, SendableEntryStream, WalIndex};
+use store_api::storage::RegionId;
+
+use crate::error::{self, InvalidProviderSnafu, Result};
+use crate::nats::client::{NatsClient, NatsClientRef};
+use crate::nats::consumer::read_entries;
+use crate::nats::producer::{NatsBatchProducer, NatsBatchProducerRef};
+use crate::nats::purge_worker::PurgeWorker;
+use crate::nats::record::convert_to_nats_records;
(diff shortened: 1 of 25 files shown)


src/log-store/src/nats/log_store.rs
```diff
@@ -0,0 +1,433 @@
...
+        for (provider, seq) in &results {
+            self.subject_seq_cache
+                .entry(provider.clone())
+                .and_modify(|e| *e = (*e).max(*seq))
+                .or_insert(*seq);
+        }
+        // Map sequence back to region.
+        for (region_id, provider) in provider_for_region {
+            if let Some((_, seq)) = results.iter().find(|(p, _)| p == &provider) {
+                last_entry_ids.insert(region_id, *seq);
```
fengjiachun, line 202: Unlike the Kafka WAL, this implementation groups writes by shared subject and then assigns the subject’s final stream sequence to every region mapped to that subject. In GreptimeDB, last_entry_ids[region_id] is persisted as the region’s own WAL boundary and later drives flush, checkpoint, truncate, and replay. When multiple regions share one NATS subject in the same batch, earlier regions can inherit a later sequence that actually belongs to another region, which breaks the region-level WAL boundary that mito expects. If subjects are pooled across regions, this path needs region-granular committed sequence tracking rather than subject-granular tracking.

Do I understand this correctly, and will it work? @WenyXu @v0y4g3r

fengjiachun:
Thanks for raising this — this looks more like a design discussion than a code suggestion.
Since it may affect WAL semantics, we should clarify the intended behavior before accepting the change.

cc @WenyXu @v0y4g3r @killme2008

killme2008:
@jpikoulas Thanks for your contribution.

It would be better to propose an RFC for this implementation before submitting this PR:

https://github.com/GreptimeTeam/greptimedb/tree/main/docs/rfcs

We'd be happy to discuss the new WAL implementation.

killme2008:
Sorry, we can’t accept this PR. Please split it into smaller, reviewable PRs, and consider submitting an RFC first.

PR #7782, opened 2026-03-10
feat: replace shadow-rs with self-maintained version info

waynexia:

Shadow-rs watches itself and always triggers rebuild (no reproducible build). This patch removes shadow-rs and uses a self-maintained version info instead. This also removes the source-time field from version info, which seems unnecessary to me as we always include a commit hash

## PR Checklist

### DIFF
--- a/src/common/version/build.rs
+++ b/src/common/version/build.rs
@@ -18,7 +18,7 @@ use std::process::Command;
 use std::{env, fs, io};
 
 use cargo_manifest::Manifest;
-use git2::{Repository, RepositoryOpenFlags, StatusOptions};
+use git2::{ErrorCode, Repository, RepositoryOpenFlags, StatusOptions};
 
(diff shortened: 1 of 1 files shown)


sunng87 (approved):

src/common/version/build.rs
```diff
@@ -13,80 +13,283 @@
...
+fn open_repository(workspace_root: &Path) -> Option<Repository> {
+    Repository::open_ext(
+        workspace_root,
+        RepositoryOpenFlags::NO_SEARCH,
+        std::iter::empty::<&Path>(),
+    )
+    .ok()
+}
+
+fn git_branch(repository: Option<&Repository>) -> String {
```
discord9, line 258: why not simply fail fast when repo is None?

waynexia: this is for supporting building from flat source without git, like downloaded repo tarball etc

PR #7784, opened 2026-03-10
feat(servers): add Loki HTTP query API for Grafana 11 compatibility

jpikoulas:
## What this PR does

Adds a Loki-compatible HTTP query API to GreptimeDB's server layer, enabling
Grafana 11 to use GreptimeDB as a Loki datasource for log storage and querying.

## Motivation

GreptimeDB can ingest logs via the existing Loki push endpoint. This PR adds
the query-side API so Grafana can discover labels, run LogQL queries, and
display logs in Explore — without any changes to the Grafana UI.

## Endpoints added

- `GET /loki/api/v1/labels` — label name discovery
- `GET /loki/api/v1/detected_labels` — Grafana 11 structured metadata label discovery
- `GET /loki/api/v1/label/{name}/values` — label value discovery
- `GET /loki/api/v1/query` — instant LogQL query (used for Grafana datasource health check)
- `GET /loki/api/v1/query_range` — range LogQL query (log streams + metric queries)
- `GET /loki/api/v1/series` — series discovery
- `GET /loki/api/v1/index/stats` — index stats (returns stub, required by Grafana)

## Implementation notes

- Queries are translated to SQL against the `loki_logs` table
- LogQL stream selectors are parsed and mapped to SQL WHERE clauses
- Metric queries (e.g. `count_over_time`) return `matrix` result type; plain stream selectors return `streams`
- Label value queries do not apply time range filtering (avoids empty dropdowns with sparse data)
- Unit tests added for all major handlers

## Tested with

- Grafana 11 with `lokiStructuredMetadata` feature flag enabled
- Alloy log collector (`loki.source.kubernetes`) pushing to GreptimeDB standalone
- Labels, label values, and log results all verified working in Grafana Explore

### DIFF
--- a/.github/workflows/fork-build.yml
+++ b/.github/workflows/fork-build.yml
@@ -34,6 +34,11 @@ jobs:
             echo "tag=${GITHUB_REF_NAME}" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y protobuf-compiler
+
       - name: Set up Rust toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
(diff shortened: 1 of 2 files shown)


killme2008:
Sorry, we can’t accept this PR. Please split it into smaller, reviewable PRs, and consider submitting an RFC first.

PR #7785, opened 2026-03-10
feat: export import v2 pr1

fengjiachun:

## Summary

This PR introduces the foundation layer for export/import v2:

- snapshot manifest model
- schema extraction and schema metadata handling
- snapshot storage abstraction over URI-based backends
- initial DDL export/import support
- base command wiring for `export-v2` and `import-v2`

This PR is intentionally focused on the shared format and storage layer. It does **not** introduce the full data export/import pipeline yet.

## Out of Scope

Not included in this PR:
- explicit chunking/export coordinator flow
- full import pipeline
- resume/state handling
- management commands

## PR Checklist

### DIFF
--- a/src/cli/src/data/import_v2/command.rs
+++ b/src/cli/src/data/import_v2/command.rs
@@ -351,6 +351,14 @@ fn is_schema_scoped_statement(sql: &str) -> bool {
         };
         rest = next.trim_start();
     }
+
+    if starts_with_keyword(rest, "EXTERNAL") {
+        let Some(next) = rest.get("EXTERNAL".len()..) else {
+            return false;
+        };
+        rest = next.trim_start();
+    }
+
     starts_with_keyword(rest, "TABLE") || starts_with_keyword(rest, "VIEW")
 }
 
@@ -505,6 +513,16 @@ CREATE VIEW v AS SELECT 1;
         let stmt = ddl_statement_for_schema(
             "test_db",
             "CREATE OR REPLACE VIEW metrics_view AS SELECT * FROM metrics".to_string(),
+        );
+        assert_eq!(stmt.execution_schema.as_deref(), Some("test_db"));
+    }
+
+    #[test]
+    fn test_ddl_statement_for_schema_create_external_table_uses_execution_schema() {
+        let stmt = ddl_statement_for_schema(
+            "test_db",
+            "CREATE EXTERNAL TABLE IF NOT EXISTS ext_metrics (ts TIMESTAMP TIME INDEX) ENGINE=file"
+                .to_string(),
         );
         assert_eq!(stmt.execution_schema.as_deref(), Some("test_db"));
     }


src/cli/src/data/export_v2/schema.rs
```diff
@@ -0,0 +1,367 @@
...
+    /// Primary key column names (in order).
+    pub primary_keys: Vec<String>,
+    /// Time index column name.
+    pub time_index: String,
+    /// Table options.
+    pub options: TableOptions,
+    /// Table comment.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub comment: Option<String>,
+}
```
WenyXu: We should add partition definition?

src/cli/src/data/export_v2/schema.rs
```diff
@@ -0,0 +1,367 @@
...
+    pub ttl: Option<String>,
+    /// All other table options (write_buffer_size, skip_wal, compaction_type, etc.).
+    /// This ensures forward compatibility with future GreptimeDB versions.
+    #[serde(flatten)]
+    pub extra: HashMap<String, String>,
+}
+
+/// Table definition in JSON format.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TableDefinition {
```
WenyXu: Can we reuse the existing table metadata struct? Introducing a new struct would make maintenance more difficult.

src/cli/src/data/export_v2/command.rs
```diff
@@ -0,0 +1,467 @@
...
+            .iter()
+            .map(|s| s.name.clone())
+            .collect::<Vec<_>>();
+        schemas.sort();
+
+        let mut ddl_by_schema = Vec::with_capacity(schemas.len());
+        for schema in schemas {
+            let create_database = self.show_create("DATABASE", &schema, None).await?;
+
+            let mut physical_tables = self.get_metric_physical_tables(&schema).await?;
```
WenyXu: Why do we need to query the physical table?

src/cli/src/data/import_v2/ddl_generator.rs
```diff
@@ -0,0 +1,809 @@
...
+        Ok(sql)
+    }
+
+    fn build_physical_table_definition(
+        &self,
+        logical_table: &TableDefinition,
+        physical_name: &str,
+    ) -> TableDefinition {
+        let mut extra = logical_table.options.extra.clone();
+        extra.remove(LOGICAL_TABLE_METADATA_KEY);
```
WenyXu: why need to remove this key?

src/cli/src/data/import_v2/ddl_generator.rs
```diff
@@ -0,0 +1,809 @@
...
+
+use crate::data::export_v2::schema::{
+    ColumnDefinition, SchemaDefinition, SchemaSnapshot, SemanticType, TableDefinition,
+    ViewDefinition,
+};
+use crate::data::import_v2::error::{InvalidColumnDefinitionSnafu, Result};
+use crate::data::import_v2::executor::DdlStatement;
+
+/// Generates DDL statements from schema definitions.
+pub struct DdlGenerator<'a> {
```
WenyXu: Can we reuse the logic in `show_create_table.rs` instead of introducing duplicate code?

src/cli/src/data/export_v2/schema.rs
fengjiachun: After a series of ideological struggles, I decided to remove TableDefinition first.

src/cli/src/data/export_v2/extractor.rs
```diff
@@ -0,0 +1,237 @@
...
+        for schema_name in &schema_names {
+            let schema_def = self.extract_schema_definition(schema_name).await?;
+            snapshot.add_schema(schema_def);
+        }
+
+        Ok(snapshot)
+    }
+
+    /// Gets all non-system schemas in the catalog.
+    async fn get_all_schemas(&self) -> Result<Vec<String>> {
```
WenyXu, line 66: Nit: have we considered using the MySQL or PostgreSQL protocol? This would let us use the sqlx crate and avoid writing complex extraction logic.

src/cli/src/data/export_v2.rs
WenyXu, line 1: Styling: we don’t use `mod.rs` in this workspace.

WenyXu, line 1: Styling: we don’t use `mod.rs` in this workspace.

src/cli/src/data/export_v2/tests.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    database_client
+        .sql_in_public(&format!("DROP DATABASE IF EXISTS {schema}"))
+        .await?;
+
+    Ok(())
+}
+
+#[tokio::test]
+#[ignore]
```
WenyXu, line 232:
```suggestion

```

src/cli/src/data/import_v2/command.rs
```diff
@@ -0,0 +1,502 @@
...
+                    .into_iter()
+                    .map(|sql| ddl_statement_for_schema(schema, sql)),
+            );
+        }
+
+        Ok(statements)
+    }
+}
+
+fn parse_ddl_statements(content: &str) -> Vec<String> {
```
WenyXu, line 237: have we considered simplifying this logic by placing each table creation SQL in its own file and maintaining a manifest to define their execution order?

src/cli/src/data/path.rs
```diff
@@ -0,0 +1,76 @@
...
+pub(crate) fn ddl_path_for_schema(schema: &str) -> String {
+    format!(
+        "{}/{}/{}.sql",
+        SCHEMA_DIR,
+        DDL_DIR,
+        encode_path_segment(schema)
+    )
+}
+
+pub(crate) fn encode_path_segment(value: &str) -> String {
```
WenyXu, line 28: using https://crates.io/crates/sanitize-filename?

src/cli/src/data/export_v2/extractor.rs
fengjiachun: Good suggestion, but it feels too broad for PR1. After the DDL-only simplification, the remaining extraction logic is small, while switching to `sqlx` would be a much larger client-stack change. I’d prefer to revisit that only if the current `DatabaseClient` becomes a real maintenance bottleneck.

src/cli/src/data/import_v2/command.rs
fengjiachun: I think at this stage it’s appropriate to put everything in a single file; using separate files for each table would require maintaining the manifest, which would be cumbersome, but that could be a direction to consider for the future.

src/cli/src/data/export_v2/tests.rs
fengjiachun: It is a e2e test

src/cli/src/data/path.rs
fengjiachun: I’d prefer to keep the local helper here. We need a deterministic, symmetric path-segment encoding for both export and import, rather than a general filename sanitization policy, and the current `%XX` encoding keeps that contract explicit with very little code.

src/cli/src/data/snapshot_storage.rs
```diff
@@ -0,0 +1,639 @@
...
+    let bucket_or_container = url.host_str().unwrap_or("").to_string();
+    if bucket_or_container.is_empty() {
+        return InvalidUriSnafu {
+            uri: uri.to_string(),
+            reason: "URI must include bucket/container in host",
+        }
+        .fail();
+    }
+
+    let root = url.path().trim_start_matches('/').to_string();
```
discord9, line 86: `extract_remote_location()` currently accepts `s3://bucket`/`oss://bucket`/`gs://bucket` with an empty root. Combined with `delete_snapshot()` using `remove_all("/")`, `export-v2 create --force` can end up deleting unrelated objects from the whole bucket/container when the snapshot is pointed at bucket root. I think we should reject empty remote roots here (or otherwise force a non-empty snapshot prefix) before wiring them into the backend config.

src/cli/src/data/export_v2/extractor.rs
```diff
@@ -0,0 +1,237 @@
...
+            if !all_schemas.iter().any(|s| s.eq_ignore_ascii_case(schema)) {
+                return SchemaNotFoundSnafu {
+                    catalog: self.catalog,
+                    schema,
+                }
+                .fail();
+            }
+        }
+
+        Ok(schemas
```
discord9: I think we need to dedupe schema names after the case-insensitive lookup here. Right now `--schemas public,PUBLIC` survives as `[public, public]`, which means export can write duplicate entries into `manifest.schemas`, and import will replay the same DDL twice. That is especially easy to trip over for views because they are not always emitted as `IF NOT EXISTS`.

src/cli/src/data/import_v2/command.rs
```diff
@@ -0,0 +1,502 @@
...
+
+fn ddl_statement_for_schema(schema: &str, sql: String) -> DdlStatement {
+    if is_schema_scoped_statement(&sql) {
+        DdlStatement::with_execution_schema(sql, schema.to_string())
+    } else {
+        DdlStatement::new(sql)
+    }
+}
+
+fn is_schema_scoped_statement(sql: &str) -> bool {
```
discord9, line 331: `ddl_statement_for_schema()` only treats `CREATE TABLE` / `CREATE VIEW` as schema-scoped. That means an exported `CREATE EXTERNAL TABLE ...` statement falls through to `sql_in_public()` during import, so restoring a non-`public` schema can recreate the external table in `public` or hit wrong-name collisions. Can we extend the schema-scoped detection to cover external tables as well?

src/cli/src/data/snapshot_storage.rs
fengjiachun: [554c163](https://github.com/GreptimeTeam/greptimedb/pull/7785/commits/554c163dd58ea791fdea3864f58ebc5b1359bb70)

src/cli/src/data/export_v2/extractor.rs
fengjiachun: [8d056c7](https://github.com/GreptimeTeam/greptimedb/pull/7785/commits/8d056c774cead116df79e77ff50527d4ffc27c45)

src/cli/src/data/import_v2/command.rs
fengjiachun: [6be196b](https://github.com/GreptimeTeam/greptimedb/pull/7785/commits/6be196bb2415d56da5cddb84d308353f3a1b26b8)

src/cli/src/data/export_v2/extractor.rs
WenyXu, line 1: Nit: maybe we can move the extraction logic into utils for future reuse, and keep the core logic simple.

src/cli/src/data/snapshot_storage.rs
```diff
@@ -0,0 +1,649 @@
...
+            "gs" | "gcs" => Ok(Self::Gcs),
+            "azblob" => Ok(Self::Azblob),
+            "file" => Ok(Self::File),
+            scheme => UnsupportedSchemeSnafu { scheme }.fail(),
+        }
+    }
+}
+
+/// Extracts bucket/container and root path from a URI.
+fn extract_remote_location(uri: &str) -> Result<RemoteLocation> {
```
WenyXu, line 75: Can we reuse an existing util or create a new one to handle similar logic?
https://github.com/GreptimeTeam/greptimedb/blob/b17947b5d8027ed2b731320cf8cb1c72d7c5895b/src/common/datasource/src/object_store.rs#L44

Nit: also prefer moving this tricky logic into utils.

src/cli/src/data/snapshot_storage.rs
```diff
@@ -0,0 +1,649 @@
...
+
+    StorageScheme::from_uri(uri)
+}
+
+fn schema_index_path() -> String {
+    format!("{}/{}", SCHEMA_DIR, SCHEMAS_FILE)
+}
+
+/// Extracts the absolute filesystem path from a file:// URI.
+fn extract_file_path_from_uri(uri: &str) -> Result<String> {
```
WenyXu, line 131: ditto

src/cli/src/data/export_v2/extractor.rs
fengjiachun: I’d prefer to keep it local for now. After the DDL-only simplification, the extractor is already fairly small, and I don’t see a second concrete caller yet. If the same parsing/extraction logic shows up elsewhere later, we can extract it with a clearer reuse boundary.

src/cli/src/data/snapshot_storage.rs
fengjiachun: I looked at `common/datasource::object_store::parse_url`, but the semantics here are stricter than generic URL parsing: we also validate snapshot-specific constraints like non-empty remote roots and restricted `file://` forms. I’d prefer to keep this logic local for now rather than reusing a looser helper or introducing a new utility prematurely.

PR #7793, opened 2026-03-11
feat: pending flow

discord9:

This PR explores allowing `CREATE FLOW` to succeed even when source tables do not exist yet, then letting flow activation happen later after the source tables are created:

- allow flow metadata to be created in a pending state when source tables are missing
- add metasrv-side pending flow reconciliation to retry activation later
- activate pending flows through a dedicated procedure guarded by flow-level locking instead of mutating metadata directly from the background ticker
- keep pending flows out of runtime routing metadata until activation succeeds
- treat pending flows as batching-mode only for now

### Main logic changes

1. **Pending flow metadata**
   - extend flow metadata to persist pending status, unresolved source table names, and last activation error
   - keep backward compatibility for old serialized `FlowInfoValue` payloads

2. **Create flow path**
   - when source tables are missing, `CreateFlowProcedure` no longer fails immediately
   - the procedure creates pending flow metadata instead of writing runtime flow routing state
   - if the flow later becomes activatable through `OR REPLACE`, peer allocation is now performed explicitly so we do not produce an active flow without flownodes or routes

3. **Pending activation path**
   - add `ActivatePendingFlowProcedure`
   - lock with `CatalogLock::Read` + `FlowLock::Write`
   - resolve source tables again, validate batching-mode activation constraints, create the flow on flownode(s), update metadata, and invalidate caches
   - metasrv background reconcile only scans pending flows and submits this procedure; it no longer mutates flow metadata directly

4. **Safety / invariants**
   - entering `CreateFlows` with an empty peer list is now treated as an error instead of silently succeeding
   - add regression coverage for replacing a pending flow and promoting it to an active flow with proper route/flownode metadata

### Limitations in this PR

- pending flow activation is batching-first only
- this PR does not add create-table event driven immediate activation yet; activation still relies on periodic metasrv reconciliation

### Compatibility

- API compatibility: no intended breaking API change
- metadata compatibility: newly added flow metadata fields use serde defaults and are covered by backward-compatibility tests for old payloads

## Tests

Added / updated tests around:

- pending flow creation when source tables are missing
- pending flow activation after source table creation
- pending flow remaining pending when activation constraints are not satisfied
- `OR REPLACE` on a pending flow activating it with allocated peers and route metadata
- backward compatibility for `FlowInfoValue`

## PR Checklist


discord9:
@codex review

src/common/meta/src/ddl/tests/create_flow.rs
```diff
@@ -74,9 +107,23 @@ async fn test_create_flow_source_table_not_found() {
...
+    let flow_id = *output.downcast_ref::<FlowId>().unwrap();
+    let flow_info = ddl_context
+        .flow_metadata_manager
+        .flow_info_manager()
+        .get(flow_id)
+        .await
+        .unwrap()
+        .unwrap();
+    assert!(flow_info.is_pending());
+    assert_eq!(flow_info.unresolved_source_table_names().len(), 1);
```
killme2008, line 126: assert the table name?

src/common/meta/src/ddl/activate_flow.rs
```diff
@@ -0,0 +1,482 @@
...
+use crate::instruction::{CacheIdent, CreateFlow};
+use crate::key::flow::flow_info::{FlowInfoValue, FlowStatus};
+use crate::key::flow::flow_route::FlowRouteValue;
+use crate::key::table_name::TableNameKey;
+use crate::key::{DeserializedValueWithBytes, FlowId, FlowPartitionId};
+use crate::lock_key::{CatalogLock, FlowLock, FlowNameLock};
+use crate::metrics;
+use crate::peer::Peer;
+
+pub struct ActivatePendingFlowProcedure {
```
killme2008, line 52: Better to document the pub procedure

src/common/meta/src/ddl/activate_flow.rs
```diff
@@ -0,0 +1,482 @@
...
+
+    async fn on_prepare(&mut self) -> Result<Status> {
+        let Some(current_flow_info) = self
+            .context
+            .flow_metadata_manager
+            .flow_info_manager()
+            .get_raw(self.data.flow_id)
+            .await?
+        else {
+            return Ok(Status::done());
```
killme2008, line 116: Not quite sure when it's not found, but better to add some log when it's not found.

src/common/meta/src/ddl/activate_flow.rs
```diff
@@ -0,0 +1,482 @@
...
+            let routes = self
+                .context
+                .flow_metadata_manager
+                .flow_route_manager()
+                .routes(self.data.flow_id)
+                .await?;
+            self.data.peers = routes.into_iter().map(|(_, value)| value.peer).collect();
+            self.data.resolved_table_ids =
+                current_flow_info.get_inner_ref().source_table_ids.clone();
+            self.data.state = ActivatePendingFlowState::InvalidateFlowCache;
```
killme2008, line 129: Adds some debug logs

src/common/meta/src/ddl/activate_flow.rs
```diff
@@ -0,0 +1,482 @@
...
+    }
+
+    fn lock_key(&self) -> LockKey {
+        if let Some(flow_info) = &self.data.prev_flow_info_value {
+            // if previous flow info is available, it means the flow is already created
+            // we should lock the flow name to avoid concurrent modification with create/replace flow procedure
+            LockKey::new(vec![
+                CatalogLock::Read(&self.data.catalog_name).into(),
+                FlowLock::Write(self.data.flow_id).into(),
+                FlowNameLock::new(
```
killme2008, line 282: The order of lock could be `[catalog, flow_name, flow_id]`.

I think it's better to let the `flow_id` as the last lock.

src/common/meta/src/ddl/activate_flow.rs
```diff
@@ -0,0 +1,482 @@
...
+fn build_create_request(
+    flow_id: FlowId,
+    flow_info: &FlowInfoValue,
+    resolved_table_ids: &[TableId],
+) -> CreateRequest {
+    let mut flow_options = flow_info.options.clone();
+    flow_options.insert(
+        FlowType::FLOW_TYPE_KEY.to_string(),
+        FlowType::Batching.to_string(),
+    );
```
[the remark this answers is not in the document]
killme2008: +1

src/common/meta/src/ddl/activate_flow.rs
```diff
@@ -0,0 +1,482 @@
...
+                .as_ref()
+                .context(error::UnexpectedSnafu {
+                    err_msg: "missing previous flow info for activation",
+                })?;
+        let request = build_create_request(
+            self.data.flow_id,
+            flow_info.get_inner_ref(),
+            &self.data.resolved_table_ids,
+        );
+        create_flow_on_peers(
```
killme2008, line 196: Assert that `unresolved_source_table_names.is_empty()`.

Although it’s checked above, keeping the assertion is good practice.

src/common/meta/src/ddl/create_flow.rs
```diff
@@ -281,20 +346,20 @@ impl CreateFlowProcedure {
             })]);
         }
 
-        let (_flow_info, flow_routes) = (&self.data).into();
-        let flow_part2peers = flow_routes
-            .into_iter()
-            .map(|(part_id, route)| (part_id, route.peer))
-            .collect();
+        if self.data.is_active() {
```
killme2008, line 349: Looks like it must be `actived` here?

If not, it's an unexpected error.

So we can change this `if` into `ensure`?

src/meta-srv/src/flow.rs
```diff
@@ -0,0 +1,120 @@
...
+use crate::define_ticker;
+use crate::error::{PendingFlowReconcileSnafu, Result};
+
+const PENDING_FLOW_RECONCILE_INTERVAL: Duration = Duration::from_secs(10);
+
+pub enum Event {
+    Tick,
+}
+
+pub type PendingFlowReconcileTickerRef = Arc<PendingFlowReconcileTicker>;
```
killme2008, line 34: Could we resuse the `RepeatedTask`?

https://greptimedb.rs/common_runtime/repeated_task/struct.RepeatedTask.html

src/meta-srv/src/flow.rs
```diff
@@ -0,0 +1,120 @@
...
+                .get_raw(flow_id)
+                .await
+                .context(PendingFlowReconcileSnafu)?;
+            let Some(current_flow_info) = current_flow_info else {
+                continue;
+            };
+            if !current_flow_info.get_inner_ref().is_pending() {
+                continue;
+            }
+            let _ = self
```
killme2008: Should we verify the existence of `unresolved_source_table_names` before submitting the procedure? If these tables are missing, we can prevent unnecessary processing and avoid overloading the metasrv.

The checking is lightweight.

src/operator/src/statement/ddl.rs
```diff
@@ -700,6 +699,10 @@ impl StatementExecutor {
             }
         }
 
+        if has_missing_source_table {
+            return Ok(FlowType::Batching);
```
killme2008, line 760: Adds a log

evenyag:
I think this makes things complicated. I'd prefer to give users a clear error message and let them retry once they've created the source table.

A better approach may be adding a `WITH` option to the create flow statement, so that we create the flow on the source table created. Then we can register a callback to the metasrv by the table name. Once the source table is created, the metasrv can invoke the callback and submit the create flow task like the normal execution flow.

@MichaelScofield @fengjiachun @WenyXu WDYT

killme2008:
> A better approach may be adding a `WITH` option to the create flow statement, so that we create the flow on the source table created. Then we can register a callback to the metasrv by the table name. Once the source table is created, the metasrv can invoke the callback and submit the create flow task like the normal execution flow.

This is as complex as this PR, may be more.

evenyag:
> > A better approach may be adding a `WITH` option to the create flow statement, so that we create the flow on the source table created. Then we can register a callback to the metasrv by the table name. Once the source table is created, the metasrv can invoke the callback and submit the create flow task like the normal execution flow.
> 
> This is as complex as this PR, may be more.

I don't think so. It won't introduce new procedures.

killme2008:
> > > A better approach may be adding a `WITH` option to the create flow statement, so that we create the flow on the source table created. Then we can register a callback to the metasrv by the table name. Once the source table is created, the metasrv can invoke the callback and submit the create flow task like the normal execution flow.
> > 
> > 
> > This is as complex as this PR, may be more.
> 
> I don't think so. It won't introduce new procedures.

The procedure logic isn’t as complex as you think. If you add a callback for creating tables, you’ll need to design and implement a new mechanism, and also handle consistency and concurrency during table creation—the complexity doesn’t go away.

killme2008:
@evenyag 
Another drawback is the need to introduce new SQL syntax like `with`, which users must learn and we must implement.

The current implementation avoids adding new syntax, so users don’t need to learn anything new.

fengjiachun:
I suggest making this behavior explicit via a new option, for example:

`WITH ('defer_on_missing_source'='true')`

By default, this option is `false`, so `CREATE FLOW` keeps its original semantics: if a source table is missing, it returns an error.

Only when users explicitly set `defer_on_missing_source=true` do we allow deferred activation, which is essentially the behavior implemented in this PR now.

This way we can reuse the pending-flow logic from this PR without changing the original semantics of `CREATE FLOW`.

I think this is a reasonable compromise between usability and keeping the default behavior easy to understand.

WDYT @discord9 @evenyag @killme2008

discord9:
> I suggest making this behavior explicit via a new option, for example:
> 
> `WITH ('defer_on_missing_source'='true')`
> 
> By default, this option is `false`, so `CREATE FLOW` keeps its original semantics: if a source table is missing, it returns an error.
> 
> Only when users explicitly set `defer_on_missing_source=true` do we allow deferred activation, which is essentially the behavior implemented in this PR now.
> 
> This way we can reuse the pending-flow logic from this PR without changing the original semantics of `CREATE FLOW`.
> 
> I think this is a reasonable compromise between usability and keeping the default behavior easy to understand.
> 
> WDYT @discord9 @evenyag @killme2008

Fair enough, I will add a optional option in create flow like `'defer_on_missing_source'='true'` or something, drafting this pr first

PR #7810, opened 2026-03-13
fix: correct unicode representation for jsonb_to_string

### LINKED ISSUE #7808: Imcompatible unicode encoding in json_to_string output
### What type of bug is this?

Incorrect result

Frontend

### Minimal reproduce step

In psql:

```
public=> create table testtable(t timestamp time index, v json);
OK 0
public=> insert into testtable values (0, '{"a":123}');
OK 1
public=> insert into testtable values (0, '{"a":"\u2024"}');
OK 1
public=> insert into testtable values (2, '{"a":"\u2024"}');
OK 1
public=> select t, json_to_string(v) from testtable;
             t              | json_to_string(testtable.v)
----------------------------+-----------------------------
 1970-01-01 00:00:00.000000 | {"a":"․"}
 1970-01-01 00:00:00.002000 | {"a":"․"}
(2 rows)

public=> insert into testtable values (4, '{"a":"\u2028"}');
OK 1
public=> select t, json_to_string(v) from testtable;
             t              | json_to_string(testtable.v)
----------------------------+-----------------------------
 1970-01-01 00:00:00.000000 | {"a":"․"}
 1970-01-01 00:00:00.002000 | {"a":"․"}
 1970-01-01 00:00:00.004000 | {"a":"\u{2028}"}
(3 rows)

```

The unicode character should be in this form: `\u2028` 

### What did you see instead?

It's in the rust form `\u{2028}`

bun or python will not parse this:

```
SyntaxError: JSON Parse error: "\u{202" is not a valid unicode escape


```

any

HEAD

We had relevant fix in #7520 but it didn't fix all cases.

sunng87:

Fixes #7808

Correct unicode representation for json string. 

## PR Checklist

### DIFF
--- a/src/datatypes/src/types/json_type.rs
+++ b/src/datatypes/src/types/json_type.rs
@@ -405,18 +405,12 @@ pub fn jsonb_to_string(val: &[u8]) -> Result<String> {
 /// Converts a json type value to serde_json::Value
 pub fn jsonb_to_serde_json(val: &[u8]) -> Result<serde_json::Value> {
     let json_string = jsonb_to_string(val)?;
-    let normalized = fix_unicode_point(&json_string)?;
-    serde_json::Value::from_str(&normalized).context(DeserializeSnafu { json: normalized })
+    serde_json::Value::from_str(&json_string).context(DeserializeSnafu { json: json_string })
 }
 
 /// Normalizes a JSON string by converting Rust-style Unicode escape sequences to JSON-compatible format.
 ///
-/// This function is intended to be used on JSON strings produced from the internal
-/// JSONB representation (e.g. via [`jsonb_to_string`]). It first calls
-/// `serde_json::Value::from_str` to check if the string is valid. If that succeeds,
-/// the original string is returned as-is.
-///
-/// If the initial parse fails, the input is scanned for Rust-style Unicode code
+/// The input is scanned for Rust-style Unicode code
 /// point escapes of the form `\\u{H...}` (a backslash, `u`, an opening brace,
 /// followed by 1–6 hexadecimal digits, and a closing brace). Each such escape is
 /// converted into JSON-compatible UTF‑16 escape sequences:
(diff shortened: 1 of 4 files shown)


src/datatypes/src/types/json_type.rs
```diff
@@ -405,16 +405,16 @@ pub fn jsonb_to_string(val: &[u8]) -> Result<String> {
 /// Converts a json type value to serde_json::Value
 pub fn jsonb_to_serde_json(val: &[u8]) -> Result<serde_json::Value> {
     let json_string = jsonb_to_string(val)?;
-    jsonb_string_to_serde_value(&json_string)
+    let normalized = jsonb_string_to_serde_value(&json_string)?;
+    serde_json::Value::from_str(&normalized).context(DeserializeSnafu { json: normalized })
```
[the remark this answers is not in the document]
avifenesh: Same as metrics discussion — the LLM agent can inject the token value directly. Pre-check handles the missing token case.

src/datatypes/src/types/json_type.rs
```diff
@@ -396,7 +396,7 @@ pub fn jsonb_to_string(val: &[u8]) -> Result<String> {
     match jsonb::from_slice(val) {
         Ok(jsonb_value) => {
             let serialized = jsonb_value.to_string();
-            Ok(serialized)
+            fix_unicode_point(&serialized)
```
MichaelScofield, line 399: This makes every call to `jsonb_to_string` will do a regex `replace_all`, regardless of whether the `val` contains Rust-style Unicode code, which is less often. It occurs to my mind that jsonb value can directly convert to serde_json value, maybe jsonb already take care of the Rust-style Unicode code escape?

sunng87: The root cause of this issue is jsonb's `to_string` doesn't generate correct unicode style. It outputs `\u{xxxx}` instead of `\uxxxx`. So for any `jsonb.to_string()` we will use this patch function to fix before shipping to client.

MichaelScofield: Yes, I mean is it possible to not use the `jsonb.to_string()`? Let the `jsonb_value` convert to serde_json value directly, and convert to string thereafter.

sunng87: Just tested, when constructing serde_json value from jsonb_value, it copies the string value to serde_json value directly, so the `u{xxxx}` style was kept. Since there is no deserialization process, it won't cause any error.

```rust
#[test]
fn test_unicode_point_json_string() {
    let value = r#"{"a": "Tomcat \u2008"}"#;
    let jsonb = jsonb::parse_value(value.as_bytes()).unwrap().to_vec();
    let result = jsonb::to_serde_json(jsonb.as_ref()).unwrap().to_string();
    assert_eq!(result, value);
}

thread 'types::json_type::tests::test_unicode_point_json_string' (1223921) panicked at src/datatypes/src/types/json_type.rs:474:9:
    assertion `left == right` failed
      left: "{\"a\":\"Tomcat \u{2008}\"}"
     right: "{\"a\": \"Tomcat \\u2008\"}"
```

MichaelScofield (approved):

PR #7816, opened 2026-03-16
docs: flow inc query rfc

discord9:

- RFC draft for Flow incremental query planning

- Add a concise RFC plan at `docs/rfcs/2026-03-16-flow-inc-query.md`.
- Summarize the correctness-first design for Flow incremental query:
  - opt-in incremental read (`seq > checkpoint`)
  - stale detection + fallback to full mode
  - per-region correctness watermark for checkpoint advancement

## PR Checklist

### DIFF
--- a/docs/rfcs/2026-03-16-flow-inc-query.md
+++ b/docs/rfcs/2026-03-16-flow-inc-query.md
@@ -47,6 +47,12 @@ When incremental mode is enabled:
 - map `after_seq` to `memtable_min_sequence` (exclusive lower bound)
 - keep existing snapshot upper-bound behavior (`memtable_max_sequence`)
 
+Important limitation in v1:
+
+- incremental filtering is correctness-proven only for memtable rows
+- SST files do not preserve detailed row-level sequence metadata; they only expose coarser file-level sequence information
+- therefore `seq > checkpoint` must not assume precise incremental pruning across memtable->SST flush boundaries
+
 If required incremental parameters are missing or invalid, return argument error.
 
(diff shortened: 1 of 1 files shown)


docs/rfcs/2026-03-16-flow-inc-query.md
```diff
@@ -0,0 +1,166 @@
...
+
+These options are opt-in and only affect Flow incremental execution paths.
+
+## 2) Scan mapping
+
+When incremental mode is enabled:
+
+- map `after_seq` to `memtable_min_sequence` (exclusive lower bound)
+- keep existing snapshot upper-bound behavior (`memtable_max_sequence`)
+
```
fengjiachun, line 49: Could you clarify how seq > checkpoint stays correct after data is flushed from memtable to SST? My understanding is that they share the same sequence space, but memtable filtering is row-level while SST filtering is file-level. If after_seq is mapped only to memtable_min_sequence, what prevents rereading old SST data or missing newly flushed data?

discord9: > Could you clarify how seq > checkpoint stays correct after data is flushed from memtable to SST? My understanding is that they share the same sequence space, but memtable filtering is row-level while SST filtering is file-level. If after_seq is mapped only to memtable_min_sequence, what prevents rereading old SST data or missing newly flushed data?

if inc query can't determine a seq because it's in sst, then this inc query simply fail and retreat to full query instead of inc query

fengjiachun:
Others LGTM

PR #7821, opened 2026-03-17
feat: flow inc query

discord9:

## Summary

This PR adds a proof-of-concept for Flow incremental queries using a changed-rows-only incremental shape.

Instead of recomputing the full result from the full source input on every round, the POC aggregates only the delta window and joins it with sink state to emit only the rows that need to be updated in the sink.

## Why

The goal is to validate whether incremental queries can materially reduce Flow query cost before we invest further in end-to-end scheduling and batching work.

## Benchmark Result

The strongest result comes from a workload shaped like `src >> sink >> delta`:

- `small_delta_src_ultra_sink_mid`
  - full recompute: `707-716 ms`
  - changed-rows-only incremental (`delta-only left join update`): `124-126 ms`

This is roughly a 5.7x improvement in wall-clock latency.

## Profiling

For this workload, full recompute is dominated by scanning and aggregating the full source table, while the incremental path shifts the main cost to sink-side scan and join work.

That means the POC already demonstrates a meaningful win, and the next optimization target is clear: reduce sink-side read cost further.

## Caveats

- This is still a POC, not the final production shape.
- The gain is workload-dependent; the positive region is currently clearest when `src >> sink >> delta`.
- The current benchmark focuses on query-shape validation, not full end-to-end Flow scheduling behavior.

## Next Step

Continue optimizing the changed-rows-only incremental path, especially sink-side scan/repartition cost.

## PR Checklist

### DIFF
--- a/src/client/src/database.rs
+++ b/src/client/src/database.rs
@@ -89,11 +89,23 @@ impl OutputMetrics {
     }
 
     pub fn region_watermark_map(&self) -> Option<std::collections::HashMap<u64, u64>> {
-        self.get()?.region_latest_sequences.map(|sequences| {
-            sequences
+        Some(
+            self.get()?
+                .region_watermarks
                 .into_iter()
-                .collect::<std::collections::HashMap<_, _>>()
-        })
+                .filter_map(|entry| entry.watermark.map(|seq| (entry.region_id, seq)))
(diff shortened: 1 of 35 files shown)


discord9:
done

PR #7825, opened 2026-03-17
fix: prevent stale in-flight cache refill after invalidation in CacheContainer

### LINKED ISSUE #7815: Query fails after repartition with error Logical region ... not found
### What type of bug is this?

Unexpected error

Query Engine

### Minimal reproduce step

This issue is **not deterministically reproducible**. It occurs during fuzz testing with concurrent inserts and repartition operations. The following SQL sequence illustrates the typical workflow where the failure was observed.

```sql
-- Step 1: Split partition
ALTER TABLE fuzz_repartition_metric_physical
SPLIT PARTITION (
  host >= '5QVb' AND host < '7Mu3'
) INTO (
  host >= '5QVb' AND host < '7Czg',
  host >= '7Czg' AND host < '7Mu3'
);

-- Step 2: Concurrent inserts into multiple metric tables
INSERT INTO `AlIQUam` (ts, `acCusamus`)
VALUES ('2026-03-13 14:13:52.672+0000', NULL);

INSERT INTO `COnSEQUATUR` (`MaioReS`, host, `eST`, ts, `NESCiunt`)
VALUES
('ad', 'voluptate', NULL, '2026-03-13 14:13:53.672+0000', 'voluptatem'),
(NULL, 'et', 'error', '2026-03-13 14:13:54.672+0000', 'molestias');

INSERT INTO `DEleniti` (ts, val, host)
VALUES ('2026-03-13 14:13:55.672+0000', 0.024464213259292822, 'culpa');

INSERT INTO `Quia` (`noN`, ts, val)
VALUES
('sint', '2026-03-13 14:13:56.672+0000', 0.8916716695957527),
('et', '2026-03-13 14:13:57.672+0000', 0.8849525113630123);

-- Step 3: Validate row counts (normally succeeds)
SELECT COUNT(1) FROM `AlIQUam`;

-- Step 4: Merge partitions
ALTER TABLE fuzz_repartition_metric_physical
MERGE PARTITION (
  host >= '5QVb' AND host < '7Czg',
  host >= '7Czg' AND host < '7Mu3'
);

-- Step 5: Continue inserts
INSERT INTO `AlIQUam` (val, `acCusamus`, `Ut`, `Est`, ts)
VALUES
(0.33308660818704217, 'impedit', 'dolores', 'voluptates', '2026-03-13 14:14:37.672+0000');

-- Step 6: Validation query may fail
SELECT COUNT(1) AS count FROM `AlIQUam`;
```

Query result

### What did you see instead?

Failed to decode DataFusion plan:
Logical region 4415226380354(1028, 66) not found

doesn't matter

main

```bash

```

WenyXu:

fixes #7815

This PR fixes a cache consistency issue where table route cache could return stale/dirty values during concurrent `invalidate` + cache-miss initialization.  
The core intention is to preserve existing fast-path behavior while enabling strict latest-value semantics where required.

1. **Fix race-induced dirty refill in cache container**
   - Introduced version-aware retry initialization flow (`init_with_retry`) with exponential backoff and bounded retries.
   - Prevents returning values loaded from an outdated cache version during invalidate/load races.
2. **Add and evolve latest-read semantics**
   - Added latest-read capability, then refactored to **construction-time strategy selection** via `InitStrategy`:
     - `Unchecked`: keeps historical behavior (may return stale/dirty value under race).
     - `VersionChecked`: retries on version change and returns latest value.
   - Avoids runtime mode switching complexity and keeps behavior explicit per container instance.
3. **Apply strict strategy to table route cache**
   - Updated table route cache to use `InitStrategy::VersionChecked` so route reads prefer correctness under concurrent invalidation.

## PR Checklist

### DIFF
--- a/src/common/meta/src/cache/container.rs
+++ b/src/common/meta/src/cache/container.rs
@@ -71,8 +71,8 @@ pub struct CacheContainer<K, V, CacheToken> {
 
 fn latest_get_backoff() -> impl Iterator<Item = Duration> {
     ExponentialBuilder::default()
-        .with_min_delay(Duration::from_millis(100))
-        .with_max_delay(Duration::from_millis(1000))
+        .with_min_delay(Duration::from_millis(10))
+        .with_max_delay(Duration::from_millis(100))
         .with_max_times(3)
         .build()
 }
--- a/src/common/meta/src/error.rs
+++ b/src/common/meta/src/error.rs
@@ -1254,7 +1254,10 @@ impl Error {
 
     /// Determine whether it is a retry later type through [StatusCode]
     pub fn is_retry_later(&self) -> bool {
-        matches!(self, Error::RetryLater { .. })
+        matches!(
+            self,
+            Error::RetryLater { .. } | Error::GetLatestCacheRetryExceeded { .. }
+        )
     }
 
     /// Determine whether it needs to clean poisons.


WenyXu:
@codex review

src/common/meta/src/cache/container.rs
```diff
@@ -75,13 +78,31 @@ where
...
+}
+
 #[async_trait::async_trait]
 impl<K, V> CacheInvalidator for CacheContainer<K, V, CacheIdent>
 where
     K: Send + Sync,
     V: Send + Sync,
 {
     async fn invalidate(&self, _ctx: &Context, caches: &[CacheIdent]) -> Result<()> {
+        self.create_new_version();
```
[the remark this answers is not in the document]
fengjiachun: +1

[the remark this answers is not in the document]
WenyXu: [83bb092](https://github.com/GreptimeTeam/greptimedb/pull/7825/commits/83bb09264ad53bf15b53d1aab0ea763bbd5144c0)

src/common/meta/src/cache/container.rs
```diff
@@ -75,6 +78,23 @@ where
     }
 }
 
+impl<K, V, CacheToken> CacheContainer<K, V, CacheToken> {
+    #[inline]
```
killme2008: We don't use `inline` anymore.

src/common/meta/src/cache/container.rs
```diff
@@ -75,6 +78,23 @@ where
...
+    fn get_version(&self) -> usize {
+        self.version.load(Ordering::Relaxed)
+    }
+
+    #[inline]
+    fn create_new_version(&self) -> usize {
```
killme2008:
```suggestion
    fn inc_version(&self) -> usize {
```

Looks better

src/common/meta/src/cache/container.rs
```diff
@@ -85,11 +105,16 @@ where
...
             .filter(|token| (self.token_filter)(token))
-            .map(|token| (self.invalidator)(&self.cache, token));
-        join_all(tasks)
-            .await
-            .into_iter()
-            .collect::<Result<Vec<_>>>()?;
+            .map(|token| (self.invalidator)(&self.cache, token))
+            .collect::<Vec<_>>();
+        if !tasks.is_empty() {
+            self.create_new_version();
```
killme2008: Should we move this above? Increase the version before calling the `invalidator`?

src/common/meta/src/cache/container.rs
```diff
@@ -119,8 +144,15 @@ where
                 .context(error::ValueNotExistSnafu)?
         };
 
+        let pre_version = self.get_version();
+
         match self.cache.try_get_with(key, init).await {
-            Ok(value) => Ok(Some(value)),
+            Ok(value) => {
+                if !self.validate_version(pre_version) {
+                    self.cache.invalidate(&key).await;
```
killme2008: Which issue are we addressing: 

* cache invalidation misses 
* or stale values during invalidation?

The first cannot be fixed in this PR. To address the second, we likely need to call `try_get_with` again when version validation fails.

src/common/meta/src/cache/container.rs
WenyXu: [b512662](https://github.com/GreptimeTeam/greptimedb/pull/7825/commits/b51266215b6262d6117dcf8789489c0802287832)

src/common/meta/src/cache/container.rs
WenyXu: [b512662](https://github.com/GreptimeTeam/greptimedb/pull/7825/commits/b51266215b6262d6117dcf8789489c0802287832)

src/common/meta/src/cache/container.rs
WenyXu: [b512662](https://github.com/GreptimeTeam/greptimedb/pull/7825/commits/b51266215b6262d6117dcf8789489c0802287832)

src/common/meta/src/cache/container.rs
WenyXu: The version will be incremented before executing the invalidation futures.

src/common/meta/src/cache/container.rs
```diff
@@ -75,6 +89,67 @@ where
     }
 }
 
+impl<K, V, CacheToken> CacheContainer<K, V, CacheToken> {
+    fn inc_version(&self) -> usize {
```
MichaelScofield: Seems the returned value is not used?

src/common/meta/src/cache/container.rs
```diff
@@ -75,6 +89,67 @@ where
...
+            .inc();
+        let _timer = metrics::CACHE_CONTAINER_LOAD_CACHE
+            .with_label_values(&[cache_name])
+            .start_timer();
+        let value = init(&key)
+            .await
+            .transpose()
+            .context(error::ValueNotExistSnafu)??;
+
+        if pre_version == version.load(Ordering::Relaxed) {
```
MichaelScofield, line 183: Seems `version` is not associated with key. So if multiple keys are initializing in parallel, does `version` increment affects each other? This could effectively make valid init falsely treated as invalid.

src/common/meta/src/cache/container.rs
```diff
@@ -99,25 +179,40 @@ where
...
-                .with_label_values(&[&self.name])
-                .start_timer();
-            moved_init(&moved_key)
-                .await
-                .transpose()
-                .context(error::ValueNotExistSnafu)?
-        };
+        let init = init(self.initializer.clone(), key, &self.name);
+
+        match self.cache.try_get_with::<_, Error>(key, init).await {
```
MichaelScofield: Seems the concurrency issues are not resolved between `get` and `get_latest`. What if `get` got executed before `get_latest`? This effectively diminish the contract of the "getting the latest value" for `get_latest`. You know, `try_get_with` will not init again if cached value exists.

WenyXu: Good catch, fixed in [a55ad0d](https://github.com/GreptimeTeam/greptimedb/pull/7825/commits/a55ad0ded5f4e4fa855f36e151a5e66eb2308ffb)

src/common/meta/src/cache/container.rs
WenyXu: [1aec129](https://github.com/GreptimeTeam/greptimedb/pull/7825/commits/1aec1290f6b68a6c233e8df7ec5945ed4bae2a60)

src/common/meta/src/cache/container.rs
```diff
@@ -85,11 +202,16 @@ where
...
-            .map(|token| (self.invalidator)(&self.cache, token));
-        join_all(tasks)
-            .await
-            .into_iter()
-            .collect::<Result<Vec<_>>>()?;
+            .map(|token| (self.invalidator)(&self.cache, token))
+            .collect::<Vec<_>>();
+        if !tasks.is_empty() {
+            self.inc_version();
+            join_all(tasks)
```
fengjiachun: There’s a sense that the join parallelization here isn’t faster than sequential execution, because the cache invalidator is originally very small and fast.

src/common/meta/src/cache/container.rs
```diff
@@ -139,11 +273,16 @@ where
...
```
fengjiachun: Same as above

WenyXu: [6e9e1ef](https://github.com/GreptimeTeam/greptimedb/pull/7825/commits/6e9e1ef52adc47367504df44b799f43eb28cfbda)

src/common/meta/src/cache/container.rs
WenyXu: [6e9e1ef](https://github.com/GreptimeTeam/greptimedb/pull/7825/commits/6e9e1ef52adc47367504df44b799f43eb28cfbda)

src/common/meta/src/cache/container.rs
WenyXu: It only affects cases where the version changes before and after the init key’s future completes. In such situations. Note that the version is only incremented during cache invalidation.

src/common/meta/src/cache/container.rs
```diff
@@ -44,6 +65,16 @@ pub struct CacheContainer<K, V, CacheToken> {
...
     initializer: Initializer<K, V>,
     token_filter: fn(&CacheToken) -> bool,
+    version: Arc<AtomicUsize>,
+    init_strategy: InitStrategy,
+}
+
+fn latest_get_backoff() -> impl Iterator<Item = Duration> {
+    ExponentialBuilder::default()
+        .with_min_delay(Duration::from_millis(100))
+        .with_max_delay(Duration::from_millis(1000))
```
killme2008: I think the min delay could be smaller ,such as 10ms?

src/common/meta/src/cache/container.rs
```diff
@@ -75,21 +132,83 @@ where
...
+
+        if pre_version == version.load(Ordering::Relaxed) {
+            return Ok(value);
+        }
+
+        if let Some(duration) = backoff.next() {
+            sleep(duration).await;
+            attempts += 1;
+        } else {
+            return error::GetLatestCacheRetryExceededSnafu { attempts }.fail();
```
killme2008, line 191: If initialization fails, will the user receive an error or a stale value?

src/common/meta/src/error.rs
```diff
@@ -714,6 +714,16 @@ pub enum Error {
     #[snafu(display("Failed to get cache"))]
     GetCache { source: Arc<Error> },
 
+    #[snafu(display(
+        "Failed to get latest cache value after {} attempts due to concurrent invalidation",
+        attempts
+    ))]
+    GetLatestCacheRetryExceeded {
```
killme2008, line 721: It should be retryable?

WenyXu: Good catch

src/common/meta/src/cache/container.rs
WenyXu: Returns an error

src/common/meta/src/cache/container.rs
WenyXu: [99e5a3a](https://github.com/GreptimeTeam/greptimedb/pull/7825/commits/99e5a3a21f1db68f87c2013f1b0238df983ef024)

src/common/meta/src/error.rs
WenyXu: [99e5a3a](https://github.com/GreptimeTeam/greptimedb/pull/7825/commits/99e5a3a21f1db68f87c2013f1b0238df983ef024)

PR #7827, opened 2026-03-18
refactor: simplify scan memory tracking

fengjiachun:

- make scan memory tracking stream-local
- expose scan_memory_on_exhausted in mito config
- simplify QueryMemoryTracker and MemoryTrackedStream
- update config docs and examples

And let's see what Gemini has to say.

## PR Checklist

### DIFF
--- a/config/config.md
+++ b/config/config.md
@@ -18,7 +18,7 @@
 | `write_bytes_exhausted_policy` | String | Unset | Policy when write bytes quota is exhausted.<br/>Options: "wait" (default, 10s timeout), "wait(<duration>)" (e.g., "wait(30s)"), "fail" |
 | `init_regions_in_background` | Bool | `false` | Initialize all regions in the background during the startup.<br/>By default, it provides services after all regions have been initialized. |
 | `init_regions_parallelism` | Integer | `16` | Parallelism of initializing regions. |
-| `max_concurrent_queries` | Integer | `0` | The maximum current queries allowed to be executed. Zero means unlimited. |
+| `max_concurrent_queries` | Integer | `0` | The maximum concurrent queries allowed to be executed. Zero means unlimited. |
 | `enable_telemetry` | Bool | `true` | Enable telemetry to collect anonymous usage data. Enabled by default. |
 | `runtime` | -- | -- | The runtime options. |
 | `runtime.global_rt_size` | Integer | `8` | The number of threads to execute the runtime for global read operations. |
@@ -441,7 +441,7 @@
 | `require_lease_before_startup` | Bool | `false` | Start services after regions have obtained leases.<br/>It will block the datanode start if it can't receive leases in the heartbeat from metasrv. |
 | `init_regions_in_background` | Bool | `false` | Initialize all regions in the background during the startup.<br/>By default, it provides services after all regions have been initialized. |
 | `init_regions_parallelism` | Integer | `16` | Parallelism of initializing regions. |
-| `max_concurrent_queries` | Integer | `0` | The maximum current queries allowed to be executed. Zero means unlimited. |
+| `max_concurrent_queries` | Integer | `0` | The maximum concurrent queries allowed to be executed. Zero means unlimited. |
 | `enable_telemetry` | Bool | `true` | Enable telemetry to collect anonymous usage data. Enabled by default. |
 | `http` | -- | -- | The HTTP server options. |
 | `http.addr` | String | `127.0.0.1:4000` | The address to bind the HTTP server. |
--- a/config/datanode.example.toml
+++ b/config/datanode.example.toml
@@ -17,7 +17,7 @@ init_regions_in_background = false
 ## Parallelism of initializing regions.
 init_regions_parallelism = 16
 
-## The maximum current queries allowed to be executed. Zero means unlimited.
+## The maximum concurrent queries allowed to be executed. Zero means unlimited.
 max_concurrent_queries = 0
 
 ## Enable telemetry to collect anonymous usage data. Enabled by default.
--- a/config/standalone.example.toml
+++ b/config/standalone.example.toml
@@ -23,7 +23,7 @@ init_regions_in_background = false
 ## Parallelism of initializing regions.
 init_regions_parallelism = 16
 
-## The maximum current queries allowed to be executed. Zero means unlimited.
+## The maximum concurrent queries allowed to be executed. Zero means unlimited.
 max_concurrent_queries = 0
 
 ## Enable telemetry to collect anonymous usage data. Enabled by default.


src/common/recordbatch/src/lib.rs
```diff
@@ -627,152 +490,251 @@ impl QueryMemoryTracker {
...
+        if let Err(error) = tracker.try_track(additional) {
+            if !matches!(
+                tracker.tracker.on_exhausted_policy,
+                OnExhaustedPolicy::Wait { .. }
+            ) {
+                return Poll::Ready(Some(Err(error)));
+            }
+
+            self.enter_waiting(batch, additional);
+            return self.poll_waiting(cx);
```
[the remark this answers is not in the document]
fengjiachun: This is a design compromise, which will add annotations explaining all of this

fengjiachun:
@gemini-code-assist  What are you doing, have you finished the review?

waynexia:
@codex review

src/common/recordbatch/src/lib.rs
```diff
@@ -627,152 +490,257 @@ impl QueryMemoryTracker {
...
-
-                if self.limit == 0 {
-                    // Unlimited mode
-                    return Some(new_total);
-                }
+            .guard
+            .acquire_additional_with_policy(additional as u64, self.tracker.on_exhausted_policy)
+            .await;
+        if result.is_ok() {
+            self.tracked_bytes = self.tracked_bytes.saturating_add(additional);
```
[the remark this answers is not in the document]
fengjiachun: This is a valid limitation, but not a new one introduced here. The previous implementation also accumulated scan memory for the lifetime of the stream and only released it on stream drop. This refactor preserves that behavior; `wait` just makes the existing limitation more visible. Fixing it would require a larger redesign to release quota on batch lifetime / downstream consumption rather than stream lifetime.

src/common/recordbatch/src/lib.rs
```diff
@@ -627,152 +490,257 @@ impl QueryMemoryTracker {
...
+
+        if let Err(error) = tracker.try_track(additional) {
+            match tracker.tracker.on_exhausted_policy {
+                OnExhaustedPolicy::Fail => return Poll::Ready(Some(Err(error))),
+                // `Wait` is a deliberate tradeoff: the batch has already been materialized, so we
+                // keep it in memory while waiting for quota instead of failing immediately. Under
+                // contention, real memory usage can therefore exceed `scan_memory_limit` by up to
+                // one buffered batch per blocked stream.
+                OnExhaustedPolicy::Wait { .. } => {
+                    self.enter_waiting(batch, additional);
```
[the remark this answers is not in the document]
fengjiachun: I think this comes down to metric semantics rather than correctness.

In `Wait` mode we intentionally probe with `try_track()` first. If memory is not immediately available, we count that as an immediate rejection / contention event and then fall back to waiting. So `SCAN_REQUESTS_REJECTED_TOTAL` is currently closer to “immediate memory misses” than “terminally failed scans”.

If we want this metric to mean final failures only, then the implementation should change. But with the current interpretation, this behavior is expected rather than a bug.

src/common/recordbatch/src/lib.rs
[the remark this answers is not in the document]
fengjiachun: @codex what do you think

config/config.md
```diff
@@ -160,7 +160,8 @@
...
 | `region_engine.mito.allow_stale_entries` | Bool | `false` | Whether to allow stale WAL entries read during replay. |
-| `region_engine.mito.scan_memory_limit` | String | `50%` | Memory limit for table scans across all queries.<br/>Supports absolute size (e.g., "2GB") or percentage of system memory (e.g., "20%").<br/>Setting it to 0 disables the limit.<br/>NOTE: Works with max_concurrent_queries for tiered memory allocation.<br/>- If max_concurrent_queries is set: 70% of queries get full access, 30% get 70% access.<br/>- If max_concurrent_queries is 0 (unlimited): first 20 queries get full access, rest get 70% access. |
+| `region_engine.mito.scan_memory_limit` | String | `50%` | Memory limit for table scans across all queries.<br/>Supports absolute size (e.g., "2GB") or percentage of system memory (e.g., "20%").<br/>Setting it to 0 disables the limit. |
+| `region_engine.mito.scan_memory_on_exhausted` | String | `fail` | Behavior when scan memory tracking cannot immediately acquire memory.<br/>Supported values:<br/>- "fail"<br/>- "wait" (same as "wait(10s)", not unlimited)<br/>- "wait(3s)"<br/>- "wait(500ms)" |
```
killme2008: What does this mean for end users? I’m not sure.

src/common/recordbatch/src/lib.rs
```diff
@@ -627,152 +490,257 @@ impl QueryMemoryTracker {
...
+    fn ready_tracker_mut(&mut self) -> &mut StreamMemoryTracker {
+        debug_assert!(
+            self.waiting.is_none(),
+            "a ready tracker must not coexist with a waiting future"
+        );
+        self.tracker.as_mut().unwrap()
+    }
+
+    fn enter_waiting(&mut self, batch: RecordBatch, additional: usize) {
+        let tracker = self.tracker.take().unwrap();
```
killme2008, line 694: Is it safe to unwrap directly? I'm unsure whether it could call `enter_waiting` more than once.

config/config.md
killme2008: Is it meaningful for end users, or what's the best option for most users?

fengjiachun: 3842cdf

src/common/recordbatch/src/lib.rs
fengjiachun: It is safe under the current state machine because `enter_waiting()` is only called from the ready state, where `waiting` must be `None` and `tracker` must be `Some(...)`. I added `debug_assert!`s for both invariants here to make that explicit.

config/config.md
```diff
@@ -18,7 +18,7 @@
 | `write_bytes_exhausted_policy` | String | Unset | Policy when write bytes quota is exhausted.<br/>Options: "wait" (default, 10s timeout), "wait(<duration>)" (e.g., "wait(30s)"), "fail" |
 | `init_regions_in_background` | Bool | `false` | Initialize all regions in the background during the startup.<br/>By default, it provides services after all regions have been initialized. |
 | `init_regions_parallelism` | Integer | `16` | Parallelism of initializing regions. |
-| `max_concurrent_queries` | Integer | `0` | The maximum current queries allowed to be executed. Zero means unlimited.<br/>NOTE: This setting affects scan_memory_limit's privileged tier allocation.<br/>When set, 70% of queries get privileged memory access (full scan_memory_limit).<br/>The remaining 30% get standard tier access (70% of scan_memory_limit). |
+| `max_concurrent_queries` | Integer | `0` | The maximum current queries allowed to be executed. Zero means unlimited. |
```
evenyag:
```suggestion
| `max_concurrent_queries` | Integer | `0` | The maximum concurrent queries allowed to be executed. Zero means unlimited. |
```

src/common/recordbatch/src/lib.rs
```diff
@@ -627,152 +490,265 @@ impl QueryMemoryTracker {
...
+    {
+        *self.inner.on_update.lock().unwrap() = Some(Arc::new(on_update));
+    }
+
+    fn set_on_reject<F>(&self, on_reject: F)
+    where
+        F: Fn() + Send + Sync + 'static,
+    {
+        *self.inner.on_reject.lock().unwrap() = Some(Arc::new(on_reject));
+    }
```
evenyag: Do we really need mutex here? We only set it in intialization phase.

src/common/recordbatch/src/lib.rs
```diff
@@ -627,152 +490,265 @@ impl QueryMemoryTracker {
...
     inner: SendableRecordBatchStream,
-    permit: Arc<MemoryPermit>,
-    // Total tracked size, released when stream drops.
-    total_tracked: usize,
+    tracker: Option<StreamMemoryTracker>,
+    // Waiting stores a batch that has already been pulled from the inner stream but has not yet
+    // acquired additional quota. This keeps `poll_next()` non-blocking and allows bounded waits,
+    // at the cost of temporarily holding one untracked batch per blocked stream in memory.
+    waiting: Option<PendingTrackFuture>,
 }
```
evenyag, line 666: Can we rewrite this using the `stream!` or `try_stream` macro? Then it should simplify the stream implementation and avoid creating the boxed enter_waiting future each time.
```
fn memory_tracked_stream(inner: SendableRecordBatchStream, tracker: QueryMemoryTracker) -> impl Stream {}

// or

fn memory_tracked_stream(inner: SendableRecordBatchStream, tracker: QueryMemoryTracker) -> SendableRecordBatchStream {}
```

src/common/recordbatch/src/lib.rs
fengjiachun: 15ca1a2

src/common/recordbatch/src/lib.rs
fengjiachun: I looked into the `try_stream!` option, but I don't think it would simplify this case in practice.

The main reasons are:
1. the common path is still the synchronous `try_track()` fast path, so moving the whole implementation into an async generator is not an obvious win;
2. the ownership transition around the waiting path still exists because `track_with_policy()` needs to take the tracker across `await`;
3. this type also implements `RecordBatchStream` (`schema()` / `output_ordering()`), which `try_stream!` would not preserve directly, so we would likely still need an explicit wrapper.

So I think the current explicit wrapper is the clearer choice here.

config/config.md
fengjiachun: 286fdda

fengjiachun:
@killme2008 @evenyag PTAL

src/common/recordbatch/src/lib.rs
```diff
@@ -627,152 +519,246 @@ impl QueryMemoryTracker {
...
-    pub fn new(inner: SendableRecordBatchStream, permit: Arc<MemoryPermit>) -> Self {
+    pub fn new(inner: SendableRecordBatchStream, tracker: QueryMemoryTracker) -> Self {
         Self {
             inner,
-            permit,
-            total_tracked: 0,
+            tracker: Some(tracker.new_stream_tracker()),
+            waiting: None,
         }
     }
```
evenyag, line 675: Looks like we always create a tracker, even when the memory limit is unlimited. Is it expected?

fengjiachun: Yes, the tracker is always created because we want to collect metrics even in Unlimited mode. Later, when the user enables limits, the metrics serve as a reference.

PR #7831, opened 2026-03-19
feat: pending rows batching for metrics

v0y4g3r:

This PR introduces pending-rows batching for Prometheus remote write in the metric-engine write path to reduce per-request overhead and improve ingestion throughput.

- Add `PendingRowsBatcher` and wire it into Prometheus remote-write HTTP handling.
- Batcher usage is now gated by `prom_store.with_metric_engine`, so non-metric-engine deployments keep the existing direct write flow.
- Preserve authorization and interceptor behavior for batched writes by running the same pre-write checks before submitting requests to the batcher.
- Keep existing non-batched behavior unchanged when batching is disabled.

Compatibility: no API changes; data/schema compatibility is unchanged.

## PR Checklist

### DIFF
--- a/src/servers/src/pending_rows_batcher.rs
+++ b/src/servers/src/pending_rows_batcher.rs
@@ -311,10 +311,10 @@ impl PendingRowsBatcher {
     }
 
     fn get_or_spawn_worker(&self, key: BatchKey) -> PendingWorker {
-        if let Some(worker) = self.workers.get(&key) {
-            if !worker.tx.is_closed() {
-                return worker.clone();
-            }
+        if let Some(worker) = self.workers.get(&key)
+            && !worker.tx.is_closed()
+        {
+            return worker.clone();
         }
 
         let entry = self.workers.entry(key.clone());


evenyag:
@codex review

killme2008:
Do we have any benchmark results for this optimization?

tests-integration/tests/http.rs
```diff
@@ -1482,6 +1482,11 @@ enable = true
 [prom_store]
 enable = true
 with_metric_engine = true
+pending_rows_flush_interval = "2s"
+max_batch_rows = 100000
+max_concurrent_flushes = 256
+worker_channel_capacity = 65526
+max_inflight_requests = 3000
```
evenyag: Should we disable them by default?

src/servers/src/pending_rows_batcher.rs
```diff
@@ -0,0 +1,1098 @@
...
+                        payload,
+                    })),
+                })),
+            };
+
+            let datanode = node_manager.datanode(&datanode).await;
+            let _timer = PENDING_ROWS_BATCH_INGEST_STAGE_ELAPSED
+                .with_label_values(&["flush_write_region"])
+                .start_timer();
+            match datanode.handle(request).await {
```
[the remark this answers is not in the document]
v0y4g3r: Not planned in this PR

src/servers/src/pending_rows_batcher.rs
```diff
@@ -0,0 +1,1106 @@
...
+            };
+
+            let record_batch = align_record_batch_to_schema(record_batch, region_schema.as_ref())?;
+            aligned_batches.push((table_name, record_batch));
+        }
+
+        Ok(aligned_batches)
+    }
+
+    fn get_or_spawn_worker(&self, key: BatchKey) -> PendingWorker {
```
fengjiachun, line 313: This creates one long-lived worker per BatchKey, but there is no idle eviction path and nothing ever removes entries from workers. Since schema and physical_table can vary by request, a long-running frontend can accumulate an unbounded number of dormant worker tasks/channels until restart. Consider adding idle shutdown plus map cleanup for unused workers.

v0y4g3r: This was addressed in https://github.com/GreptimeTeam/greptimedb/pull/7831/commits/9b9830953aa365c4431c0256c540a158a1f5e57d

v0y4g3r:
@fengjiachun PTAL

PR #7840, opened 2026-03-20
fix: prom cast to f64

discord9:

- User-reported PromQL compatibility issue where
  `count(filtered_expr) / count(rate_expr) * 100`
  returned `0` in GreptimeDB while Prometheus returned a fractional percentage.

## Whats changed and whats your intention?

This PR fixes PromQL arithmetic type handling for binary operators in the planner.

### Root cause

PromQL arithmetic in GreptimeDB could preserve integer-style division semantics after aggregate expressions such as `count(...)`.
That meant expressions like `1 / 3 * 100` could collapse to `0` before the final multiplication, which diverged from Prometheus float-based arithmetic semantics.

### Fix

- Update `PromPlanner::prom_token_to_binary_expr_builder` to cast arithmetic operands to `Float64` before building PromQL arithmetic expressions.
- Avoid generating redundant nested `CAST(... AS Float64)` when the operand is already a `Float64` cast or a `Float64` literal.
- Keep comparison and set-operator behavior unchanged.

### Why this is done in the planner

DataFusions default optimizer does simplify some casts, but the relevant upstream logic is mainly focused on:

- literal cast simplification, and
- unwrapping casts in comparison-with-literal cases.

It does not reliably eliminate the redundant same-type casts introduced in this PromQL arithmetic path, so the planner now avoids generating unnecessary nested casts directly.

### Test coverage

This PR adds and updates coverage for the affected behavior:

- PromQL integration regression for the reproduced ratio case in `tests-integration/src/tests/promql_test.rs`
- sqlness regression in `tests/cases/standalone/common/promql/anon_promql_ratio_repro.sql`
- planner test expectation updates in `src/query/src/promql/planner.rs` for cast-aware logical plans

### Compatibility

- No API changes
- No schema or data compatibility changes
- Behavior change is limited to making PromQL arithmetic consistent with Prometheus semantics

## PR Checklist

### DIFF
--- a/src/query/src/promql/planner.rs
+++ b/src/query/src/promql/planner.rs
@@ -3324,10 +3324,21 @@ impl PromPlanner {
         token: TokenType,
     ) -> Result<Box<dyn Fn(DfExpr, DfExpr) -> Result<DfExpr>>> {
         let cast_float = |expr| {
-            DfExpr::Cast(Cast {
-                expr: Box::new(expr),
-                data_type: ArrowDataType::Float64,
-            })
+            if matches!(
+                &expr,
+                DfExpr::Cast(Cast {
+                    data_type: ArrowDataType::Float64,
+                    ..
+                })
+            ) || matches!(&expr, DfExpr::Literal(ScalarValue::Float64(_), _))
+            {
+                expr
+            } else {
+                DfExpr::Cast(Cast {
+                    expr: Box::new(expr),
+                    data_type: ArrowDataType::Float64,
+                })
+            }
         };
         match token.id() {
             token::T_ADD => Ok(Box::new(move |lhs, rhs| {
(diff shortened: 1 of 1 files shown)


src/query/src/promql/planner.rs
```diff
@@ -3323,28 +3323,44 @@ impl PromPlanner {
...
+            })),
+            token::T_MUL => Ok(Box::new(move |lhs, rhs| {
+                Ok(cast_float(lhs) * cast_float(rhs))
+            })),
+            token::T_DIV => Ok(Box::new(move |lhs, rhs| {
+                Ok(cast_float(lhs) / cast_float(rhs))
+            })),
+            token::T_MOD => Ok(Box::new(move |lhs: DfExpr, rhs| {
+                Ok(cast_float(lhs) % cast_float(rhs))
+            })),
```
evenyag, line 3358: Can the optimizer eliminate the cast?

discord9: > Can the optimizer eliminate the cast?

datafusion only do cast eliminate for literal/comparsion so added a new check for nested cast as f64 to avoid nest

PR #7862, opened 2026-03-25
feat: implement prefilter framework and primary key prefilter

evenyag:

This PR introduces a basic parquet prefilter framework for mito2 flat-format reads, using primary-key predicates to refine row selection before reading the full projection. It also enables partition-column predicates in the parquet primary-key filter path and avoids re-evaluating predicates that were already applied during prefiltering.

- Added a two-phase parquet prefilter flow in the flat reader:
    - build prefilter context from usable primary-key predicates
    - read only prefilter columns first
    - compute a refined RowSelection
    - read the full projection with the refined selection
- Refactored row-group reader construction to carry a dedicated RowGroupBuildContext, so prefilter execution can share row selection, filter, and metric state cleanly.
- Updated flat parquet filtering to skip primary-key predicates that were already applied by prefiltering, avoiding duplicate decode/evaluation work.
- Enabled partition-column predicates in parquet primary-key filters by making skip_partition_column configurable in the PK codec/filter path.
- Kept the old partition-tree memtable behavior unchanged by continuing to skip partition columns there.
- Added prefilter metrics to parquet fetch stats, including prefilter elapsed time and number of rows filtered.
- Added helper APIs needed by the new flow, including:
    - cloneable SimpleFilterEvaluator
    - flat-format detection for dictionary-backed primary key batches
    - small reader/path plumbing updates to pass prefilter capability through the read path

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet/prefilter.rs
+++ b/src/mito2/src/sst/parquet/prefilter.rs
@@ -33,7 +33,7 @@ use parquet::schema::types::SchemaDescriptor;
 use snafu::{OptionExt, ResultExt};
 use store_api::metadata::{RegionMetadata, RegionMetadataRef};
 
-use crate::error::{ReadParquetSnafu, Result, UnexpectedSnafu};
+use crate::error::{DecodeSnafu, ReadParquetSnafu, Result, UnexpectedSnafu};
 use crate::sst::parquet::flat_format::primary_key_column_index;
 use crate::sst::parquet::format::{PrimaryKeyArray, ReadFormat};
 use crate::sst::parquet::reader::{RowGroupBuildContext, RowGroupReaderBuilder};
@@ -74,7 +74,10 @@ pub(crate) fn matching_row_ranges_by_primary_key(
             end += 1;
         }
 
-        if pk_filter.matches(pk_values.value(key as usize)) {
+        if pk_filter
+            .matches(pk_values.value(key as usize))
+            .context(DecodeSnafu)?
+        {
             if let Some(last) = matched_row_ranges.last_mut()
                 && last.end == start
             {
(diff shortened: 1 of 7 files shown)


src/mito2/src/sst/parquet/prefilter.rs
```diff
@@ -210,210 +164,190 @@ impl PrimaryKeyFilter for CachedPrimaryKeyFilter {
...
+    } else {
+        let prefilter_selection =
+            row_selection_from_row_ranges(matched_row_ranges.into_iter(), total_rows);
 
-        assert!(filters.is_empty());
-    }
+        // Intersect with original row selection if present.
+        match &build_ctx.row_selection {
+            Some(original) => original.intersection(&prefilter_selection),
+            None => prefilter_selection,
```
fengjiachun: `matched_row_ranges` here are measured in the prefilter input stream, which already has `build_ctx.row_selection` applied, so they are subset-relative rather than row-group-relative. Converting them with `row_selection_from_row_ranges(..., total_rows)` and then intersecting with the original selection mixes two different coordinate spaces. A simple case is: original selection keeps global rows `[5, 6, 7]`, prefilter matches `1..2` in that subset, and this code keeps `[5, 6, 7]` instead of rebasing to global row 6. This needs rebasing/composition semantics (`and_then`) instead of intersection.

evenyag: Fixed in https://github.com/GreptimeTeam/greptimedb/pull/7862/commits/d64b1f47c66220a6adb7e5037875a8602bb23fb5

evenyag:
@codex review

src/mito2/src/sst/parquet/file_range.rs
```diff
@@ -674,6 +711,12 @@ impl RangeBase {
                 continue;
             }
 
+            // Flat parquet PK prefiltering already applied these tag predicates while refining
+            // row selection, so skip them here to avoid decoding/evaluating the same condition twice.
+            if skip_prefiltered_pk_filters && filter_ctx.usable_primary_key_filter() {
```
discord9, line 716: if  PrimaryKeyFilter somehow failed to eval and let some extra rows that should be filtered out be included in result, this skip here(due to `usable_primary_key_filter` is true) cause later exact reevaluate be skipped hence cause output to include rows that should be filtered?

evenyag: It is unexpected that the usable primary key filter failed, but the simple filter succeeded. Then it might be a bug.

waynexia:
@codex review

src/mito2/src/sst/parquet/file_range.rs
evenyag: Yield the error in https://github.com/GreptimeTeam/greptimedb/pull/7862/commits/c5f4a19f68e87d8b17f4ad353be76f92675adf05

evenyag:
@codex review

PR #7866, opened 2026-03-26
fix(datatypes): correct ConstantVector rhs comparison in vector equality

cuiweixie:
## Summary

Fixes a bug in `equal()` when either operand is a `ConstantVector`: the recursive call compared `lhs.inner()` to `lhs.inner()` instead of `rhs.inner()`, so equality was wrong when only the right-hand side was constant.

## Change

- Use `rhs` when downcasting the second `ConstantVector` inner for comparison.

### DIFF
--- a/src/datatypes/src/vectors/eq.rs
+++ b/src/datatypes/src/vectors/eq.rs
@@ -259,6 +259,22 @@ mod tests {
         ])));
     }
 
+    // Regression: second arm must downcast `rhs` (was `lhs`), or same-length ConstantVectors
+    // with different inners compare equal.
+    #[test]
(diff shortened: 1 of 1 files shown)


src/datatypes/src/vectors/eq.rs
```diff
@@ -67,7 +67,7 @@ fn equal(lhs: &dyn Vector, rhs: &dyn Vector) -> bool {
                 .downcast_ref::<ConstantVector>()
                 .unwrap()
                 .inner(),
-            &**lhs
+            &**rhs
```
evenyag, line 70: Can we add a unit test to cover this?

PR #7868, opened 2026-03-26
fix(mito2): inclusive range overlap for compaction Ranged

cuiweixie:
## Summary

`Ranged::overlap` is documented as using **inclusive** bounds. For inclusive ranges, two ranges overlap iff `max(start) <= min(end)`. The previous strict `<` treated ranges that only meet at a shared boundary as non-overlapping.

## Change

- Use `<=` in the overlap predicate in `src/mito2/src/compaction/run.rs`.


v0y4g3r:
This check is intentionally relaxed, please refer to this discussion: https://github.com/GrepTimeTeam/greptimedb/pull/6121#discussion_r2099441497

Issue #7860, opened 2026-03-25
OTLP ingestion rejects INT64 value for Float64 column (should coerce)


turisanapo:
## Description

When ingesting OpenTelemetry traces via the OTLP endpoint, GreptimeDB rejects writes where a span attribute value is encoded as `INT64` but the existing column type is `Float64`. This is a lossless widening conversion (every INT64 value is representable as Float64 within reasonable ranges) and should be handled automatically.

## Reproduction

1. Send an OTLP trace with a span attribute `gen_ai.request.temperature = 0.7` — this auto-creates the column as `Float64`
2. Later, send a trace where the same attribute is `gen_ai.request.temperature = 0` (integer zero)
3. The OTEL SDK encodes `0` as `INT64` (since `Number.isInteger(0) === true` in JavaScript)
4. GreptimeDB rejects the entire batch with:

```
Invalid request to region ..., reason: column span_attributes.gen_ai.request.temperature
expect type Float64(Float64Type), given: INT64(4)
```

## Impact

- The entire write batch containing the mismatched attribute is dropped — not just the single span
- This causes silent data loss for traces/metrics where a numeric attribute happens to be a whole number
- The issue is intermittent and depends on whether the client sends `0` vs `0.0`, `1` vs `1.0`, etc.
- In the OpenTelemetry ecosystem, attribute values are simply "numbers" — the INT64 vs DOUBLE distinction is a protobuf serialization detail that producers don't control

## Expected behavior

GreptimeDB should implicitly coerce INT64 values to Float64 when the target column is Float64, since this is a safe, lossless widening conversion. This is consistent with how most databases handle numeric type promotion (e.g., PostgreSQL, ClickHouse, DuckDB all accept integers in float columns).

## Environment

- GreptimeDB version: v1.0.0-rc.1
- Ingestion protocol: OTLP/HTTP
- OTEL SDK: `@opentelemetry/sdk-trace-node` (JavaScript)

## Workaround

None available on the client side. In JavaScript, `0` and `0.0` are the same value — there's no way to force the OTEL protobuf serializer to encode an integer as DOUBLE. The fix needs to be server-side.

killme2008:
Good catch. Allowing safe type coercion here makes sense. We'll investigate this and evaluate the best fix.

turisanapo:
One follow-up question: is it expected that a single span with a mismatched attribute type causes the **entire write batch** to be rejected? 

[...]

PR #7870, opened 2026-03-26
fix: allow auto type upscale conversion in trace ingestion

shuiyisong:

#7860 

This this the first PR trying to solve issue 7860.

What we do is
- During the row-writer phase, we use `write_field_unchecked` to write the raw datatype and data value
- In the end, use `reconcile_trace_column_types` to decide what datatype should be and do the coerce

There are several cases where we can do the coerce
1. In- and cross-batch(request) data type mismatch.
2. Incoming request data with the existing table schema's data type

## PR Checklist

### DIFF
--- a/src/frontend/src/instance/otlp.rs
+++ b/src/frontend/src/instance/otlp.rs
@@ -210,6 +210,10 @@ impl OpenTelemetryProtocolHandler for Instance {
 }
 
 impl Instance {
+    /// Picks the final datatype for one trace column.
+    ///
+    /// Existing table schema is authoritative when present. Otherwise we resolve the
+    /// request-local observed types using the shared trace coercion rules.
     fn choose_trace_target_type(
         observed_types: &[ColumnDataType],
         existing_type: Option<ColumnDataType>,
@@ -269,6 +273,8 @@ impl Instance {
                 let mut observed_types = Vec::new();
                 push_observed_trace_type(&mut observed_types, current_type);
 
+                // Scan the full request first so the final type decision is not affected
+                // by row order inside the batch.
                 for row in &rows.rows {
                     let Some(value) = row
                         .values
@@ -304,6 +310,8 @@ impl Instance {
                     continue;
                 }
 
+                // Decide the final type once per column, then rewrite all affected cells
+                // together in one row pass below.
                 let Some(target_type) =
                     Self::choose_trace_target_type(&observed_types, existing_type).map_err(
                         |_| {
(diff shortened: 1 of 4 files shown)


src/frontend/src/instance/otlp.rs
```diff
@@ -200,3 +205,261 @@ impl OpenTelemetryProtocolHandler for Instance {
...
+        .collect()
+}
+
+fn is_supported_trace_coercion(request_type: ColumnDataType, target_type: ColumnDataType) -> bool {
+    matches!(
+        (request_type, target_type),
+        (ColumnDataType::Int64, ColumnDataType::Float64)
+            | (ColumnDataType::String, ColumnDataType::Int64)
+            | (ColumnDataType::String, ColumnDataType::Float64)
+            | (ColumnDataType::String, ColumnDataType::Boolean)
```
sunng87: I think it's possible to add int64/float64/boolean->string too. There is chance the initial input is string.

src/frontend/src/instance/otlp.rs
```diff
@@ -200,3 +205,261 @@ impl OpenTelemetryProtocolHandler for Instance {
...
     }
 }
+
+/// Try to coerce a single `ValueData` to a target `ColumnDataType`.
+/// Returns the coerced value on success, or `None` if coercion is not possible.
+fn coerce_value_data(
+    value: &Option<ValueData>,
+    target: ColumnDataType,
+    request_type: ColumnDataType,
+) -> Option<Option<ValueData>> {
```
sunng87: is it really necessary for nested Option?

shuiyisong: The Inner `Option<ValueData>` is the item in the `RowInsertRequest`.
The outer `Option` indicates there's a error during the coerce, which would stop the batch and return an error to the outer.
We can use something like `Result<Option<ValueData>, ()>`, which is basically the same thing.

src/servers/src/otlp/trace/v1.rs
```diff
@@ -247,13 +310,25 @@ fn write_attributes(
         let key = format!("{}.{}", prefix, key_suffix);
         match attr.value.and_then(|v| v.value) {
             Some(OtlpValue::StringValue(v)) => {
+                let (datatype, value_data) = if let Some(existing) = writer.column_datatype(&key) {
+                    let incoming_value = ValueData::StringValue(v.clone());
```
sunng87: can we avoid this clone?

fengys1996:
@codex review

fengys1996 (approved):

fengys1996:
@codex review

PR #7873, opened 2026-03-26
feat: use partition range cache in scan

evenyag:

Uses partition range cache in SeqScan and SeriesScan read paths.

Improves how the range cache buffers results.

## PR Checklist

### DIFF
--- a/src/mito2/src/read/range_cache.rs
+++ b/src/mito2/src/read/range_cache.rs
@@ -374,11 +374,8 @@ impl CacheConcatState {
         })?;
 
         let compacted = compact_record_batches(batches)?;
-        self.estimated_size += compacted
-            .iter()
-            .map(|batch| batch.batch.get_array_memory_size())
-            .sum::<usize>();
-        self.cached_batches.extend(compacted);
+        self.estimated_size += compacted.batch.get_array_memory_size();
+        self.cached_batches.push(compacted);
         Ok(())
     }
 
@@ -387,13 +384,11 @@ impl CacheConcatState {
     }
 }
 
-fn compact_record_batches(batches: Vec<RecordBatch>) -> Result<Vec<CachedBatchSlice>> {
-    if batches.is_empty() {
-        return Ok(Vec::new());
-    }
+fn compact_record_batches(batches: Vec<RecordBatch>) -> Result<CachedBatchSlice> {
+    debug_assert!(!batches.is_empty());
 
     let slice_lengths = batches.iter().map(RecordBatch::num_rows).collect();
-    Ok(vec![build_cached_batch_slice(batches, slice_lengths)?])
+    build_cached_batch_slice(batches, slice_lengths)
 }
 
 fn build_cached_batch_slice(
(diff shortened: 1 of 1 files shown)


src/mito2/src/read/range_cache.rs
```diff
@@ -297,135 +323,251 @@ fn query_time_range_covers_partition_range(
...
+        cache_strategy: CacheStrategy,
+        part_metrics: PartitionMetrics,
+        result_tx: Option<oneshot::Sender<Option<Arc<RangeScanCacheValue>>>>,
+    ) {
+        let Some(sender) = self.sender.take() else {
+            return;
+        };
+
+        if sender
+            .send(CacheConcatCommand::Finish {
```
discord9, line 537: it is ok to send finish and return, and let background async task to actually call `put_range_result` right?

evenyag: Yes, we compact and put the cache in background

discord9:
would it be useful to add a test with multiple partition range(each with multiple sources) and with both permit being small like 1 or 2?

waynexia:
@codex review

src/mito2/src/read/range_cache.rs
```diff
@@ -297,135 +323,251 @@ fn query_time_range_covers_partition_range(
...
-        }
-        self.batches.push(batch);
+    fn finish(self) -> RangeScanCacheValue {
+        RangeScanCacheValue::new(self.cached_batches, self.estimated_size)
     }
+}
 
-    fn init_first_batch(&mut self, batch: &RecordBatch) {
-        self.total_size += batch.get_array_memory_size();
+fn compact_record_batches(batches: Vec<RecordBatch>) -> Result<Vec<CachedBatchSlice>> {
```
v0y4g3r: Can we simply return `Result<CachedBatchSlice>` here? Looks like the vec can only have one element.

src/mito2/src/read/range_cache.rs
```diff
@@ -297,135 +323,251 @@ fn query_time_range_covers_partition_range(
...
+    if batches.is_empty() {
+        return Ok(Vec::new());
+    }
```
v0y4g3r: We've checked this in the caller site.

src/mito2/src/read/range_cache.rs
```diff
@@ -297,135 +323,251 @@ fn query_time_range_covers_partition_range(
...
+            self.notify_compact();
         }
 
-        self.total_size += batch_size;
+        Ok(())
     }
 
-    fn estimated_batches_size(&self) -> usize {
-        self.total_size
+    fn notify_compact(&mut self) {
```
v0y4g3r, line 508: What is the necessity of doing compaction on the go?

killme2008:
@evenyag Some comments need to be resolved.

src/mito2/src/read/range_cache.rs
evenyag: In order to deep clone the batches and reduce the number of small batches.
Otherwise, the cached batches may be referencing more memory than we expected.

PR #7877, opened 2026-03-27
feat: auto-align Prometheus schemas in pending rows batching

v0y4g3r:

- N/A

This PR improves the Prometheus ingest batching path by aligning logical-table schemas to physical-table expectations in `PendingRowsBatcher`, reducing redundant table/partition work, and consolidating flush behavior for better throughput and simpler write flow.

Key changes:
- Refactors pending-rows flush planning/execution to batch by physical table and streamline worker/flush logic.
- Improves schema accommodation/alignment for Prometheus rows in `prom_row_builder` and related batching helpers, with clearer error handling.
- Adds and updates metrics around flush stages and bulk-insert timing for better observability.
- Adds/updates tests around batch transformation, schema handling, and batching behavior.
- Includes accompanying docs/plans and related config/doc updates touched during this branch.

Compatibility intent:
- No intentional API breakage.
- No intentional incompatible schema/data format change; behavior remains backward compatible while improving ingest path handling.

## Performance 
With batching mode enabled, we can expect about 2x performance boost with 4 regions.

image

image

## PR Checklist

### DIFF
--- a/src/metric-engine/src/batch_modifier.rs
+++ b/src/metric-engine/src/batch_modifier.rs
@@ -37,7 +37,15 @@ pub struct TagColumnInfo {
     pub column_id: ColumnId,
 }
 
-/// Computes `__tsid` values for each row.
+/// Computes the TSID for each row in a [RecordBatch].
+///
+/// The TSID is a stable hash of the set of labels (tags) present in each row.
+/// It accounts for both the names and values of all non-null tag columns.
+///
+/// # Logic
+/// - If a row has no nulls across all `sorted_tag_columns`, it uses a precomputed hash of all label names.
+/// - If a row has nulls, it dynamically computes a hash of the names of labels that are present (non-null).
+/// - In both cases, it then hashes the values of those present labels in the order specified by `sorted_tag_columns`.
 pub fn compute_tsid_array(
     batch: &RecordBatch,
     sorted_tag_columns: &[TagColumnInfo],
@@ -109,12 +117,30 @@ fn build_tag_arrays<'a>(
         .collect()
 }
 
-/// Modifies a RecordBatch for sparse primary key encoding.
+/// Modifies a [RecordBatch] to include a sparse primary key column.
+///
+/// This function transforms the input `batch` into a new `RecordBatch` where the first column
+/// is the generated primary key (named [PRIMARY_KEY_COLUMN_NAME]), followed by columns
+/// indicated by `extra_column_indices`.
+///
+/// The primary key uses a "sparse" encoding, which compactly represents the row's identity
+/// by only including non-null tag values. The encoding, handled by [SparsePrimaryKeyCodec],
+/// consists of:
+/// 1. The `table_id`.
+/// 2. A `tsid` (Time Series ID), which is a hash of the present tags.
+/// 3. The actual non-null tag values paired with their `column_id`.
+///
+/// # Parameters
+/// - `batch`: The source [RecordBatch].
+/// - `table_id`: The ID of the table.
+/// - `sorted_tag_columns`: Metadata for tag columns, used for both TSID computation and PK encoding.
+/// - `extra_column_indices`: Indices of columns from the original batch to keep in the output
+///   (typically the timestamp and value fields).
 pub fn modify_batch_sparse(
     batch: RecordBatch,
     table_id: u32,
     sorted_tag_columns: &[TagColumnInfo],
-    non_tag_column_indices: &[usize],
+    extra_column_indices: &[usize],
 ) -> Result<RecordBatch> {
     let num_rows = batch.num_rows();
     let codec = SparsePrimaryKeyCodec::schemaless();
@@ -150,7 +176,7 @@ pub fn modify_batch_sparse(
     ))];
     let mut columns: Vec<Arc<dyn Array>> = vec![Arc::new(pk_array)];
 
-    for &idx in non_tag_column_indices {
+    for &idx in extra_column_indices {
         fields.push(batch.schema().fields()[idx].clone());
         columns.push(batch.column(idx).clone());
     }


src/servers/src/prom_row_builder.rs
```diff
@@ -0,0 +1,564 @@
...
+    ArrayRef, Float64Builder, StringBuilder, TimestampMicrosecondBuilder,
+    TimestampMillisecondBuilder, TimestampNanosecondBuilder, TimestampSecondBuilder,
+    new_null_array,
+};
+use arrow::compute::cast;
+use arrow::datatypes::{DataType as ArrowDataType, Schema as ArrowSchema};
+use arrow::record_batch::RecordBatch;
+use arrow_schema::TimeUnit;
+use common_query::prelude::{greptime_timestamp, greptime_value};
+use datatypes::data_type::DataType;
```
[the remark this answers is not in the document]
v0y4g3r: This is false positive.

src/servers/src/pending_rows_batcher.rs
```diff
@@ -260,51 +314,342 @@ impl PendingRowsBatcher {
...
+        let mut seen = HashSet::new();
+
+        for (table_name, rows) in table_rows {
+            if seen.insert(table_name.as_str()) {
+                unique_tables.push((table_name.as_str(), &rows.schema));
+            }
+        }
+
+        unique_tables
+    }
```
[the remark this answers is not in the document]
v0y4g3r: @copilot We can omit this because `RowInsertRequest`s are decoded from remote write request, and we must ensure that there's no duplicate table data with different table schema.

fengjiachun:
@codex review

src/frontend/src/instance/prom_store.rs
```diff
@@ -267,6 +456,61 @@ impl PromStoreProtocolHandler for Instance {
...
+        physical_table: &str,
+        ctx: QueryContextRef,
+    ) -> ServerResult<()> {
+        let table = self
+            .catalog_manager()
+            .table(catalog, schema, physical_table, Some(ctx.as_ref()))
+            .await
+            .map_err(BoxedError::new)
+            .context(error::ExecuteGrpcQuerySnafu)?;
+        if table.is_some() {
```
fengjiachun, line 473: Existing physical tables are accepted without validating compatibility with the new flush path
`create_metric_physical_table_if_missing()` only enforces `sst_format=flat` when it creates a new physical table. If the physical table already exists, this function returns `Ok(()) ` immediately without checking whether that table is actually compatible with the new pending-rows physical flush path. On upgraded clusters, a pre-existing metric physical table created before flat-format enforcement can pass this check unchanged, and remote-write requests will proceed until they fail later during flush.

src/servers/src/pending_rows_batcher.rs
```diff
@@ -579,6 +920,219 @@ async fn spawn_flush(
...
+    let expected_essential_names = [
+        PRIMARY_KEY_COLUMN_NAME,
+        greptime_timestamp(),
+        greptime_value(),
+    ];
+    let batch_schema = batch.schema();
+    for (index, expected_name) in expected_essential_names.iter().enumerate() {
+        let actual_name = batch_schema.field(index).name();
+        ensure!(
+            actual_name == *expected_name,
```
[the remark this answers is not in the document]
v0y4g3r: fixed in ab0d309477

src/frontend/src/instance/prom_store.rs
v0y4g3r: We intentionally fails these requests if table already exists because this will cause performance degradation. Maybe we can trigger altering to set `sst_format='flat'`, but it may not be included in this pull request. I will create a todo for this.

src/metric-engine/src/batch_modifier.rs
```diff
@@ -39,7 +39,7 @@ pub(crate) struct TagColumnInfo {
 
 /// Computes `__tsid` values for each row.
 #[allow(dead_code)]
-pub(crate) fn compute_tsid_array(
+pub fn compute_tsid_array(
```
evenyag, line 49: Are they still dead_code?

src/servers/src/prom_row_builder.rs
```diff
@@ -0,0 +1,564 @@
...
+                rows,
+                *src_idx,
+                &rows.schema[*src_idx].column_name,
+                src_arrow_type.clone(),
+                row_count,
+            )?;
+            if array.data_type() != target_field.data_type() {
+                columns.push(
+                    cast(array.as_ref(), target_field.data_type()).context(error::ArrowSnafu)?,
+                );
```
evenyag: When will they have different types?

src/servers/src/pending_rows_batcher.rs
```diff
@@ -260,51 +297,347 @@ impl PendingRowsBatcher {
...
+    /// first-seen schema for duplicate table names.
+    fn collect_unique_table_schemas(
+        table_rows: &[(String, Rows)],
+    ) -> Result<Vec<(&str, &[ColumnSchema])>> {
+        let mut unique_tables: Vec<(&str, &[ColumnSchema])> = Vec::with_capacity(table_rows.len());
+        let mut seen = HashSet::new();
+
+        for (table_name, rows) in table_rows {
+            if seen.insert(table_name.as_str()) {
+                unique_tables.push((table_name.as_str(), &rows.schema));
```
evenyag, line 380: What if rows with the same table name have different schemas?

src/servers/src/pending_rows_batcher.rs
```diff
@@ -594,236 +1121,383 @@ async fn flush_batch(
...
             }
         };
 
-        let table = {
-            let _timer = PENDING_ROWS_BATCH_INGEST_STAGE_ELAPSED
-                .with_label_values(&["flush_resolve_table"])
-                .start_timer();
-            match catalog_manager
-                .table(
-                    &catalog,
```
evenyag, line 641: Maybe we can strip the column before filter_record_batch to reduce the number of columns to take.

src/servers/src/pending_rows_batcher.rs
```diff
@@ -594,236 +1121,383 @@ async fn flush_batch(
...
+        let table_id = table_batch.table_id;
+
+        // Transform each chunk to physical format directly, avoiding an
+        // intermediate concat_batches per logical table.
+        for batch in &table_batch.batches {
+            // Identify tag columns and non-tag columns from the logical batch schema.
+            // Chunks within a table_batch may have different schemas if new tag columns
+            // are added between submits.
+            // In prom batches, Float64 = value, Timestamp = timestamp, Utf8 = tags.
+            let batch_schema = batch.schema();
```
evenyag, line 1275: Maybe we can wrap this for loop into a helper function to simplify this function.

src/metric-engine/src/batch_modifier.rs
evenyag: We will pass partition_columns_set to this function. Maybe we should rename non_tag_column_indices to avoid misleading.

src/servers/src/pending_rows_batcher.rs
v0y4g3r: > What if rows with the same table name have different schemas?

This function handles single `RowInsertRequests` which is decoded from `WriteRequest`, and we ensure it already grouped table data by their names, so `table_rows` won't have duplicated table names, hence no conflicting table schema.

src/metric-engine/src/batch_modifier.rs
v0y4g3r: fixed in a7b78ae3de

src/servers/src/prom_row_builder.rs
v0y4g3r: fixed in a7b78ae3de

src/servers/src/pending_rows_batcher.rs
v0y4g3r: fixed in dac9bf63c6638ab7cc40a4f1437519aee04066fc

src/metric-engine/src/batch_modifier.rs
v0y4g3r: fixed in 36c86a66bded1dc1aba13d7cd59caeb30bcc55f7

PR #7848, opened 2026-03-23
feat: add incremental read context and scan boundaries

discord9:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

Add the foundational query context and scan request fields for
sequence-based incremental reads.

This wires flow incremental extensions, snapshot sequence transport,
and scan boundary fields without bringing in Flow state-machine or
rewrite POC logic yet.

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #7879, opened 2026-03-27
feat(flow): inc query scan bind seq

discord9:

Add mito-side scan boundary semantics for sequence-based incremental reads.

This PR is follow up on #7848. It wires memtable/SST sequence bounds through the scan path, binds effective snapshot information on scan open, and introduces stale incremental-read detection.

Key files:
- `mito2/src/read/scan_region.rs` - sequence-aware scan handling
- `mito2/src/region.rs` - stale boundary detection
- `mito2/src/engine/scan_test.rs` - sequence scan tests
- `query/src/dummy_catalog.rs` - query-to-mito scan binding

Does NOT include (follow-up PRs):
- Flow checkpoint handling
- Terminal metrics propagation
- Incremental rewrite logic

## PR Checklist

### DIFF
--- a/src/mito2/src/engine.rs
+++ b/src/mito2/src/engine.rs
@@ -129,8 +129,8 @@ use crate::cache::{CacheManagerRef, CacheStrategy};
 use crate::config::MitoConfig;
 use crate::engine::puffin_index::{IndexEntryContext, collect_index_entries_from_puffin};
 use crate::error::{
-    InvalidRequestSnafu, JoinSnafu, MitoManifestInfoSnafu, RecvSnafu, RegionNotFoundSnafu, Result,
-    SerdeJsonSnafu, SerializeColumnMetadataSnafu,
+    IncrementalQueryStaleSnafu, InvalidRequestSnafu, JoinSnafu, MitoManifestInfoSnafu, RecvSnafu,
+    RegionNotFoundSnafu, Result, SerdeJsonSnafu, SerializeColumnMetadataSnafu,
 };
 #[cfg(feature = "enterprise")]
 use crate::extension::BoxedExtensionRangeProviderFactory;
@@ -1013,11 +1013,29 @@ impl EngineInner {
 
     /// Handles the scan `request` and returns a [ScanRegion].
     #[tracing::instrument(skip_all, fields(region_id = %region_id))]
-    fn scan_region(&self, region_id: RegionId, request: ScanRequest) -> Result<ScanRegion> {
+    fn scan_region(&self, region_id: RegionId, mut request: ScanRequest) -> Result<ScanRegion> {
         let query_start = Instant::now();
         // Reading a region doesn't need to go through the region worker thread.
         let region = self.find_region(region_id)?;
-        let version = region.version();
+        let version_data = region.version_control.current();
+        let version = version_data.version;
+
+        if request.snapshot_on_scan && request.memtable_max_sequence.is_none() {
+            request.memtable_max_sequence = Some(version_data.committed_sequence);
+        }
+
+        if let Some(given_seq) = request.memtable_min_sequence {
+            let min_readable_seq = version.flushed_sequence;
+            ensure!(
+                given_seq >= min_readable_seq,
+                IncrementalQueryStaleSnafu {
+                    region_id,
+                    given_seq,
+                    min_readable_seq,
+                }
+            );
+        }
+
         // Get cache.
         let cache_manager = self.workers.cache_manager();
 
--- a/src/mito2/src/region.rs
+++ b/src/mito2/src/region.rs
@@ -318,6 +318,10 @@ impl MitoRegion {
         self.version_control.committed_sequence()
     }
 
+    pub fn flushed_sequence(&self) -> SequenceNumber {
+        self.version_control.current().version.flushed_sequence
+    }
+
     /// Returns whether the region is readonly.
     pub fn is_follower(&self) -> bool {
         self.manifest_ctx.state.load() == RegionRoleState::Follower
--- a/src/mito2/src/region.rs
+++ b/src/mito2/src/region.rs
@@ -318,6 +318,11 @@ impl MitoRegion {
         self.version_control.committed_sequence()
     }
 
+    /// Returns the latest sequence that has already been persisted into SSTs.
+    ///
+    /// Incremental memtable-only reads must use a cursor greater than or equal to
+    /// this boundary; older cursors are stale because the corresponding updates may
+    /// already have been flushed out of memtables.
     pub fn flushed_sequence(&self) -> SequenceNumber {
         self.version_control.current().version.flushed_sequence
     }
--- a/src/query/src/dummy_catalog.rs
+++ b/src/query/src/dummy_catalog.rs
@@ -197,7 +197,8 @@ impl TableProvider for DummyTableProvider {
             && let Some(query_ctx) = &self.query_ctx
             && let Some(snapshot_sequence) = scanner.snapshot_sequence()
         {
-            bind_snapshot_bound_region_seq(query_ctx, self.region_id, snapshot_sequence);
+            bind_snapshot_bound_region_seq(query_ctx, self.region_id, snapshot_sequence)
+                .map_err(|e| DataFusionError::External(Box::new(e)))?;
         }
 
         let query_memory_tracker = self.engine.query_memory_tracker();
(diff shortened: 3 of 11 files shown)


evenyag:
@codex review

src/mito2/src/engine.rs
```diff
@@ -1009,11 +1009,29 @@ impl EngineInner {
...
+        if let Some(given_seq) = request.memtable_min_sequence {
+            let min_readable_seq = version.flushed_sequence;
+            ensure!(
+                given_seq >= min_readable_seq,
+                IncrementalQueryStaleSnafu {
+                    region_id,
+                    given_seq,
+                    min_readable_seq,
+                }
+            );
```
evenyag, line 1036: Should we allow given_seq == min_readable_seq?

src/query/src/dummy_catalog.rs
```diff
@@ -314,6 +324,72 @@ impl DummyTableProviderFactory {
...
+
+fn reuse_bound_snapshot_sequence(
+    query_ctx: &QueryContext,
+    region_id: RegionId,
+    scan_request: &mut ScanRequest,
+) {
+    if let Some(snapshot_sequence) = query_ctx.get_snapshot(region_id.as_u64()) {
+        scan_request.memtable_max_sequence = Some(snapshot_sequence);
+        scan_request.snapshot_on_scan = false;
+    }
```
evenyag, line 452: Why do we set snapshot_on_scan to false?

src/query/src/dummy_catalog.rs
evenyag, line 1: Need more documents about the sequence relationship and handling.
There are so many variables:

- sst_min_sequence
- snapshot_on_scan
- snapshot_sequence
- memtable_min_sequence
- memtable_max_sequence
- query_ctx.snapshot
- flow_extensions.incremental_after_seqs

When should we set these fields? What sequences do the incremental flow require? In what order should we set these fields?

src/query/src/dummy_catalog.rs
```diff
@@ -314,6 +324,72 @@ impl DummyTableProviderFactory {
...
+    region_id: RegionId,
+    query_ctx: &QueryContext,
+) -> Result<ScanRequest> {
+    let mut scan_request = ScanRequest {
+        sst_min_sequence: query_ctx.sst_min_sequence(region_id.as_u64()),
+        snapshot_on_scan: query_requires_snapshot_bound(query_ctx),
+        ..Default::default()
+    };
+
+    reuse_bound_snapshot_sequence(query_ctx, region_id, &mut scan_request);
```
evenyag: Both query_requires_snapshot_bound and reuse_bound_snapshot_sequence take the same query_ctx but set snapshot_on_scan to different values. This is a little confusing. These functions also lack docs.

src/mito2/src/engine.rs
discord9: > Should we allow given_seq == min_readable_seq?

equal is ok here since in ScanRequest it require only the memtables that contain sequences greater than `memtable_min_sequence` will be scanned, so if request.memtable_min_sequence == version.flushed_sequence it still a valid, non-stale query, since `all seq in memtables > request.memtable_min_sequence == version.flushed_sequence`

src/query/src/dummy_catalog.rs
discord9: > Why do we set snapshot_on_scan to false?

because the snapshot is already cached in qeury_ctx, so want to reuse that instead, and prevent accidentally acquire a newer seq num

src/query/src/dummy_catalog.rs
discord9: > Need more documents about the sequence relationship and handling. There are so many variables:
> 
>     * sst_min_sequence
> 
>     * snapshot_on_scan
> 
>     * snapshot_sequence
> 
>     * memtable_min_sequence
> 
>     * memtable_max_sequence
> 
>     * query_ctx.snapshot
> 
>     * flow_extensions.incremental_after_seqs
> 
> 
> When should we set these fields? What sequences do the incremental flow require? In what order should we set these fields?

refactor to three step `decide_flow_scan/build_scan_request/apply_cached_snapshot` and add docs explain what they do

src/query/src/dummy_catalog.rs
discord9: > Both query_requires_snapshot_bound and reuse_bound_snapshot_sequence take the same query_ctx but set snapshot_on_scan to different values. This is a little confusing. These functions also lack docs.

refactored

evenyag (approved):

src/mito2/src/engine.rs
```diff
@@ -1013,11 +1013,29 @@ impl EngineInner {
...
+        let version_data = region.version_control.current();
+        let version = version_data.version;
+
+        if request.snapshot_on_scan && request.memtable_max_sequence.is_none() {
+            request.memtable_max_sequence = Some(version_data.committed_sequence);
+        }
+
+        if let Some(given_seq) = request.memtable_min_sequence {
+            let min_readable_seq = version.flushed_sequence;
+            ensure!(
```
killme2008, line 1029: When will this occur?

src/mito2/src/region.rs
```diff
@@ -318,6 +318,10 @@ impl MitoRegion {
         self.version_control.committed_sequence()
     }
```
killme2008, line 320: Adds a document for this function

src/query/src/dummy_catalog.rs
```diff
@@ -314,6 +324,140 @@ impl DummyTableProviderFactory {
...
+}
+
+fn bind_snapshot_bound_region_seq(
+    query_ctx: &QueryContext,
+    region_id: RegionId,
+    snapshot_sequence: u64,
+) -> u64 {
+    if let Some(existing) = query_ctx.get_snapshot(region_id.as_u64()) {
+        if existing != snapshot_sequence {
+            common_telemetry::warn!(
```
killme2008: Should we return error in such a case?

src/query/src/options.rs
```diff
@@ -162,6 +162,13 @@ impl FlowQueryExtensions {
         Ok(self.incremental_after_seqs.is_some())
     }
```
killme2008, line 178: Adds a doc.

Looks like it's easy to forget updating this function when you add more flow query options.

killme2008:
@discord9 Please review these comments.

src/mito2/src/engine.rs
discord9: > When will this occur?

this happens when an incremental cursor lags behind the region’s flushed boundary, once data has been flushed out of memtables, an older memtable-only cursor becomes stale

src/query/src/dummy_catalog.rs
```diff
@@ -314,6 +325,138 @@ impl DummyTableProviderFactory {
...
+fn build_scan_request(
+    query_ctx: &QueryContext,
+    region_id: RegionId,
+    decision: &FlowScanDecision,
+) -> ScanRequest {
+    // Build the initial scan request from the final decision known at provider creation
+    // time. A later scan may still refresh `memtable_max_sequence` if another source scan
+    // has bound a snapshot into `query_ctx` after this provider was created.
+    ScanRequest {
+        sst_min_sequence: query_ctx.sst_min_sequence(region_id.as_u64()),
```
fengjiachun, line 416: `plain_scan()` above says sink-table scans should bypass flow scan semantics, but this still unconditionally carries over `sst_min_sequence` from the query context. If the checkpoint ever includes a bound for the sink region, the sink scan can skip older SST data and stop behaving like a plain read. Should this be gated on `!decision.is_sink_scan`?

src/query/src/dummy_catalog.rs
```diff
@@ -314,6 +325,138 @@ impl DummyTableProviderFactory {
...
+        ..Default::default()
+    }
+}
+
+fn apply_cached_snapshot_to_request(
+    query_ctx: &QueryContext,
+    region_id: RegionId,
+    scan_request: &mut ScanRequest,
+) {
+    if let Some(snapshot_sequence) = query_ctx.get_snapshot(region_id.as_u64()) {
```
fengjiachun, line 429: This refresh applies cached snapshot bounds to every scan, including sink-table scans. That seems inconsistent with the earlier “bypass all flow scan semantics” contract for sink scans: a sink scan can still inherit `memtable_max_sequence` from `query_ctx` and get a bounded view instead of a plain read. Should sink scans skip this path as well?

fengjiachun (approved):
LGTM

Issue #7863, opened 2026-03-25
Refactor view_info field in ViewInfoValue


killme2008:
![medium](https://www.gstatic.com/codereviewagent/medium-priority.svg)

The use of `view_info.view_info.clone()` results in cloning the `Vec<u8>` of the serialized plan every time a view is resolved. On a frequently used view, this could become a performance concern.

To avoid this allocation and copy, consider changing the type of the `view_info` field in `ViewInfoValue` from `Vec<u8>` to `bytes::Bytes`. This would make cloning a cheap reference count operation.

_Originally posted by @gemini-code-assist[bot] in https://github.com/GreptimeTeam/greptimedb/pull/7857#discussion_r2982337409_

ynachi:
This is a great issue to onboard on the project. I started to look in case someone is not already on it.

killme2008:
> This is a great issue to onboard on the project. I started to look in case someone is not already on it.

Sounds great. Happy hacking.

PR #7882, opened 2026-03-29
fix: avoid cloning serialized view plans on resolve

ynachi:

close https://github.com/GreptimeTeam/greptimedb/issues/7863

Change `ViewInfoValue.view_info` from `Vec<u8>` to `common_base::bytes::Bytes`
so resolving a view no longer clones the full serialized plan buffer on every decode. 
Cloning the full vector is expensive and unnecessary.

To keep the change narrow, the metadata write boundary still accepts `Vec<u8>`
 and converts once when constructing/updating `ViewInfoValue`. The hot read path
 now uses a cheap clone of the stored bytes.  

Add some benchmarks to catch before and after performance improvement/regression.
The bench, run locally, gave the following results.
- view_resolution/256: ~4.60 us -> ~4.48 us (~7% faster)
- view_resolution/4096: ~4.64 us -> ~4.42 us (~7% faster)
- view_resolution/65536: ~6.69 us -> ~4.42 us (~34% faster)
- view_resolution/1048576: ~26.67 us -> ~4.45 us (~83% faster)

## PR Checklist

### DIFF
--- a/src/catalog/src/table_source.rs
+++ b/src/catalog/src/table_source.rs
@@ -15,7 +15,6 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use bytes::Bytes;
 use common_catalog::format_full_table_name;
 use common_query::logical_plan::{SubstraitPlanDecoderRef, rename_logical_plan_columns};
 use datafusion::common::{ResolvedTableReference, TableReference};
@@ -151,11 +150,7 @@ impl DfTableSourceProvider {
         let catalog_list = Arc::new(DummyCatalogList::new(self.catalog_manager.clone()));
         let logical_plan = self
             .plan_decoder
-            .decode(
-                Bytes::from(view_info.view_info.clone()),
-                catalog_list,
-                false,
-            )
+            .decode(view_info.view_info.clone().into(), catalog_list, false)
             .await
             .context(DecodePlanSnafu {
                 name: &table.table_info().name,
--- a/src/common/meta/src/key/view_info.rs
+++ b/src/common/meta/src/key/view_info.rs
@@ -16,6 +16,7 @@ use std::collections::{HashMap, HashSet};
 use std::fmt::Display;
 use std::sync::Arc;
 
+use common_base::bytes::Bytes;
 use serde::{Deserialize, Serialize};
 use snafu::OptionExt;
 use table::metadata::TableId;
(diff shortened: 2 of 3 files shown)


src/catalog/src/table_source.rs
```diff
@@ -152,7 +151,7 @@ impl DfTableSourceProvider {
         let logical_plan = self
             .plan_decoder
             .decode(
-                Bytes::from(view_info.view_info.clone()),
+                view_info.view_info.clone().into(),
```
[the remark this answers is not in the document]
ynachi: `view_info.view_info` is not of type `bytes::Bytes`. Instead, it is of the existing wrapper's `common_base::bytes::Bytes` type so `.into()` converts that wrapper into bytes::Bytes. It's not redundant.

src/catalog/benches/view_resolution.rs
```diff
@@ -0,0 +1,181 @@
...
+            &state,
+            |b, state| {
+                b.iter(|| runtime.block_on(state.resolve_view_once()));
+            },
+        );
+    }
+    group.finish();
+}
+
+criterion_group!(benches, bench_view_resolution);
```
killme2008: This is a very good test, but I think it’s enough to just describe the test results in the PR. It feels like this benchmark doesn’t need to be included in the patch. What do you think?

src/common/meta/src/key/view_info.rs
```diff
@@ -86,7 +84,7 @@ impl MetadataKey<'_, ViewInfoKey> for ViewInfoKey {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct ViewInfoValue {
     // The encoded logical plan
-    pub view_info: RawViewLogicalPlan,
+    pub view_info: Bytes,
```
killme2008, line 87: Add a compatibility test to deserialize existing `ViewInfoValue` objects that use `Vec<u8>` for `view_info`, ensuring no breakage of existing views.

src/catalog/benches/view_resolution.rs
ynachi: Fair point. This could indeed be a one shot test, not to be in the patch.

ynachi: Removed

src/common/meta/src/key/view_info.rs
ynachi: Compatibility test added.

PR #7885, opened 2026-03-30
feat: cancel local compaction for enter staging

WenyXu:

This PR adds cooperative cancellation for active local compaction so `EnterStaging` no longer has to wait for the whole local compaction task to finish naturally before taking effect.

The main changes are:
- split local compaction execution into a cancellable pre-commit phase and a non-cancellable manifest update phase
- reuse `CancellationHandle` / `CancellableFuture` to stop the active local compaction before manifest publish
- track active local compaction state in the scheduler and let `EnterStaging` request cancellation before enqueueing its pending DDL
- make merge task fan-out use `JoinSet` on the compact runtime so spawned merge subtasks are aborted when the cancellable phase is dropped
This PR is intentionally scoped to local compaction. Remote compaction is still treated as non-cancellable in this iteration.

One limitation of the current design is that a cancelled compaction may leave uncommitted compaction output files behind. This is acceptable for the current repartition-related use case because those files are not referenced by manifest updates, and GC will eventually clean up the orphan files.

## PR Checklist

### DIFF
--- a/src/mito2/src/compaction/compactor.rs
+++ b/src/mito2/src/compaction/compactor.rs
@@ -519,6 +519,7 @@ where
                 .collect();
 
             while let Some((inputs, handle)) = spawned.pop() {
+                let abort_handle = handle.abort_handle();
                 match CancellableFuture::new(handle, self.cancel_handle.clone()).await {
                     Ok(Ok(Ok(files))) => {
                         output_files.extend(files);
@@ -541,6 +542,7 @@ where
                         // If the cancel handle is cancelled,
                         // cancel the remaining tasks before returns the error.
                         if self.cancel_handle.is_cancelled() {
+                            abort_handle.abort();
                             for (_, handle) in spawned {
                                 handle.abort();
                             }
@@ -553,6 +555,7 @@ where
                             region_id,
                             spawned.len(),
                         );
+                        abort_handle.abort();
                         for (_, handle) in spawned {
                             handle.abort();
                         }
@@ -921,7 +924,7 @@ mod tests {
                 .await
         });
 
-        sleep(Duration::from_millis(50)).await;
+        sleep(Duration::from_millis(100)).await;
         cancel_handle.cancel();
 
         let merge_output = task
@@ -933,7 +936,6 @@ mod tests {
 
         assert!(merge_output.files_to_add.is_empty());
         assert!(merge_output.files_to_remove.is_empty());
-        assert!(started >= 1);
-        assert!(started <= 2);
+        assert_eq!(started, 2);
     }
 }


src/mito2/src/compaction/compactor.rs
```diff
@@ -422,6 +424,49 @@ impl DefaultCompactor {
...
+            let next = join_set.join_next().await;
+
+            let Some(join_result) = next else {
+                break;
+            };
+            match join_result.context(JoinSnafu)? {
+                Ok(result) => output.push(result),
+                Err(err) => {
+                    join_set.abort_all();
+                    while join_set.join_next().await.is_some() {}
```
v0y4g3r: Why we need to join the remaining tasks when `abort_all` is already called?

evenyag:
@codex review

src/mito2/src/compaction/compactor.rs
```diff
@@ -511,28 +518,28 @@ where
...
                     let handle = common_runtime::spawn_compact(fut);
                     (inputs, handle)
                 })
                 .collect();
 
-            for (inputs, handle) in spawned {
-                match handle.await {
-                    Ok(Ok(files)) => {
+            while let Some((inputs, handle)) = spawned.pop() {
+                match CancellableFuture::new(handle, self.cancel_handle.clone()).await {
```
evenyag, line 523: Why do we change it from for loop to while loop?

src/mito2/src/compaction/task.rs
```diff
@@ -41,6 +43,8 @@ use crate::{error, metrics};
 pub const MAX_PARALLEL_COMPACTION: usize = 1;
 
 pub(crate) struct CompactionTaskImpl {
+    /// Shared local-compaction state for cooperative cancellation.
+    pub(crate) local_compaction: RunningLocalCompactionRef,
```
evenyag: Why don't we call it LocalCompactionState or CompactionState?

src/mito2/src/compaction/compactor.rs
WenyXu: It used to cancel the remaining tasks.
```
for (_, handle) in spawned {
    handle.abort();
}
```

src/mito2/src/compaction.rs
```diff
@@ -599,6 +644,57 @@ impl CompactionScheduler {
...
+    }
+}
+
+pub(crate) type RunningLocalCompactionRef = Arc<Mutex<RunningLocalCompaction>>;
+
+#[derive(Debug)]
+pub(crate) struct RunningLocalCompaction {
+    cancel_handle: Arc<CancellationHandle>,
+    commit_started: bool,
+}
```
evenyag: Can we only use the mutex to guard commit_started? Then we change cancel_handle and mark_commit_started to methods of the struct.

src/mito2/src/compaction.rs
```diff
@@ -719,9 +817,54 @@ impl CompactionStatus {
...
+            } => {
+                let local_compaction = local_compaction.lock().unwrap();
+                if local_compaction.commit_started {
+                    return RequestCancelResult::TooLateToCancel;
+                }
+                if local_compaction.cancel_handle.is_cancelled() {
+                    return RequestCancelResult::AlreadyCancelling;
+                }
+
+                local_compaction.cancel_handle.cancel();
```
evenyag: We can wrap into a method.

src/mito2/src/compaction.rs
```diff
@@ -297,10 +313,7 @@ impl CompactionScheduler {
             return Vec::new();
         };
 
-        // Notify all waiters that compaction is finished.
-        for waiter in std::mem::take(&mut status.waiters) {
-            waiter.send(Ok(0));
-        }
+        finish_compaction_waiters(std::mem::take(&mut status.waiters));
```
evenyag: Why don't we implement this as a method of status?

src/mito2/src/compaction.rs
```diff
@@ -719,9 +817,54 @@ impl CompactionStatus {
...
+                local_compaction.cancel_handle.cancel();
+                RequestCancelResult::CancelIssued
+            }
+            ActiveCompaction::Remote => RequestCancelResult::TooLateToCancel,
         }
     }
 
+    fn clear_running_task(&mut self) -> bool {
+        self.active_compaction.take().is_some()
+    }
```
evenyag, line 860: Do we need to cancel the handle?

src/mito2/src/request.rs
```diff
@@ -991,6 +993,24 @@ pub(crate) struct CompactionFinished {
...
+    pub(crate) senders: Vec<OutputTx>,
+}
+
+impl CompactionCancelled {
+    pub(crate) fn on_success(self) {
+        for sender in self.senders {
+            sender.send(Ok(0));
+        }
+        info!("Compaction cancelled for region: {}", self.region_id);
+    }
```
evenyag, line 1011: Should we send ok to the senders on cancellation?

src/mito2/src/compaction.rs
```diff
@@ -352,6 +369,25 @@ impl CompactionScheduler {
...
+        // follow-up compaction request and hand pending DDLs back to the worker to resume.
+        status.finish_pending_request_with_error(Arc::new(ManualCompactionOverrideSnafu.build()));
+        let pending_ddl_requests = std::mem::take(&mut status.pending_ddl_requests);
+        self.region_status.remove(&region_id);
+        pending_ddl_requests
+    }
+
     /// Notifies the scheduler that the compaction job is failed.
     pub(crate) fn on_compaction_failed(&mut self, region_id: RegionId, err: Arc<Error>) {
         error!(err; "Region {} failed to compact, cancel all pending tasks", region_id);
```
evenyag, line 384: Do we need to clear running task on failure?

WenyXu: No, it will remove the status instead.

src/mito2/src/request.rs
WenyXu: good catch, now it returns errors [410c49d](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/410c49da30f79198f727fe13e127e968701e2299)

src/mito2/src/compaction.rs
WenyXu: No, just clean the state

src/mito2/src/compaction.rs
WenyXu: [410c49d](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/410c49da30f79198f727fe13e127e968701e2299)

src/mito2/src/compaction.rs
WenyXu: [410c49d](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/410c49da30f79198f727fe13e127e968701e2299)

src/mito2/src/compaction.rs
WenyXu: [410c49d](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/410c49da30f79198f727fe13e127e968701e2299)

src/mito2/src/compaction/task.rs
WenyXu: [410c49d](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/410c49da30f79198f727fe13e127e968701e2299)

src/mito2/src/compaction/task.rs
```diff
@@ -239,6 +243,16 @@ impl CompactionTaskImpl {
...
 
+        Ok(compaction_result)
+    }
+
+    async fn update_manifest(
+        &self,
+        compaction_result: crate::compaction::compactor::MergeOutput,
+    ) -> error::Result<RegionEdit> {
+        // Stop accepting cancellation once we are about to publish the compaction edit.
+        LocalCompactionState::mark_commit_started(&self.state);
```
evenyag: ?

src/mito2/src/compaction.rs
```diff
@@ -764,6 +894,22 @@ impl CompactionStatus {
         }
     }
 
+    fn on_cancel(mut self) -> Vec<SenderDdlRequest> {
```
evenyag, line 905: Do we need to add `#[must_use]`?

src/mito2/src/compaction.rs
```diff
@@ -597,6 +632,69 @@ impl CompactionScheduler {
...
+
+        status.on_cancel()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct LocalCompactionState {
+    cancel_handle: Arc<CancellationHandle>,
+    commit_started: Arc<AtomicBool>,
+}
```
evenyag, line 648: We still need a mutex to protect the cancellation.

WenyXu: [136f14a](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/136f14a830c0524d83affd5fee80a5fa41a63d39)

src/mito2/src/compaction.rs
WenyXu: [136f14a](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/136f14a830c0524d83affd5fee80a5fa41a63d39)

src/mito2/src/compaction/task.rs
WenyXu: [136f14a](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/136f14a830c0524d83affd5fee80a5fa41a63d39)

src/mito2/src/compaction.rs
```diff
@@ -671,13 +670,14 @@ impl LocalCompactionState {
     /// which means the compaction task is in the final stage and is about to update region version and manifest.
     /// It will reject cancellation request after this method is called.
     pub(crate) fn mark_commit_started(&self) {
-        self.commit_started
-            .store(true, std::sync::atomic::Ordering::SeqCst);
+        *self.commit_started.lock().unwrap() = true;
     }
```
evenyag: We should check cancel here.

WenyXu: [20ebc46](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/20ebc468c471be7d2e0239f916f8d48395fdf3d0)

src/mito2/src/compaction.rs
```diff
@@ -669,8 +669,15 @@ impl LocalCompactionState {
...
-        *self.commit_started.lock().unwrap() = true;
+    ///
+    /// Returns true if this is the first time to mark commit started, false otherwise.
+    pub(crate) fn mark_commit_started(&self) -> bool {
+        let commit_started = *self.commit_started.lock().unwrap();
+        if !commit_started {
+            *self.commit_started.lock().unwrap() = true;
+            return true;
+        }
+        false
```
evenyag: We should check whether the handle is cancel in the lock. If not, we mark it started. This avoids the ABA problem.
```
let mut commit_started = self.commit_started.lock().unwrap();
// check if handle is cancel
if self.handle.is_cancel() {
    return false;
}
*commit_started = true;

true
```

fengjiachun:
@codex review

src/mito2/src/compaction.rs
```diff
@@ -597,6 +631,77 @@ impl CompactionScheduler {
...
+
+    /// Marks the compaction task as started to commit,
+    /// which means the compaction task is in the final stage and is about to update region version and manifest.
+    /// It will reject cancellation request after this method is called.
+    ///
+    /// Returns true if this is the first time to mark commit started, false otherwise.
+    pub(crate) fn mark_commit_started(&self) -> bool {
+        let commit_started = *self.commit_started.lock().unwrap();
+        if !commit_started {
+            *self.commit_started.lock().unwrap() = true;
```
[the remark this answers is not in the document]
WenyXu: [11a140a](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/11a140abd8dd402ce46dff7c4d1706338498aed8)

src/mito2/src/compaction.rs
WenyXu: [11a140a](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/11a140abd8dd402ce46dff7c4d1706338498aed8)

src/mito2/src/compaction/compactor.rs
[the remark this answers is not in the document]
WenyXu: [4adf1d5](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/4adf1d589807983c052d4bac8ec4c0375876cbd8)

src/mito2/src/compaction/compactor.rs
```diff
@@ -503,38 +510,61 @@ where
...
+                    Err(_) => {
+                        debug!(
+                            "Compaction merge cancelled for region: {}, aborting remaining {} spawned tasks",
+                            region_id,
+                            spawned.len(),
+                        );
+                        for (_, handle) in spawned {
+                            handle.abort();
+                        }
+                        break;
```
fengjiachun, line 562: Once this handle has been popped out of spawned, can the in-flight task still be cancelled at all if cancellation happens while we are awaiting it? `CancellableFuture` seems to cancel the await, but I don’t see a path here that aborts the underlying JoinHandle. Would it make sense to switch this fan-out to `JoinSet` and call `abort_all()` on cancellation so the currently running merge task is still cancellable as well?

src/mito2/src/compaction/task.rs
```diff
@@ -296,14 +307,61 @@ impl CompactionTask for CompactionTaskImpl {
...
-            Ok(edit) => BackgroundNotify::CompactionFinished(CompactionFinished {
-                region_id: self.compaction_region.region_id,
-                senders: std::mem::take(&mut self.waiters),
-                start_time: self.start_time,
-                edit,
-            }),
-            Err(e) => {
+        // Marks files compacting before compaction and unmark after compaction (even if compaction is cancelled or failed), so that they won't be picked by other compaction tasks.
+        self.mark_files_compacting(true);
+        self.handle_expiration().await;
```
fengjiachun, line 312: Should this expiration-cleanup phase be cancellable as well? Right now cancellation is not observed until after `handle_expiration()` returns, so EnterStaging may still wait for the expired-SST cleanup round trip before compaction stops.

WenyXu: The handle expiration will update the manifests. For maintainability and to simplify the code, we prefer to make it uncancelable. The time required for manifest updating is acceptable.

src/mito2/src/compaction/compactor.rs
WenyXu: Good catch, [3a52700](https://github.com/GreptimeTeam/greptimedb/pull/7885/commits/3a5270094c38a6bf1c6ea0f84111afb1ffbbe12d)

PR #7892, opened 2026-03-31
feat: partial success in trace ingestion

### LINKED ISSUE #7875: OTLP ingestion: single span error should not reject the entire write batch
## Description

When GreptimeDB receives an OTLP trace export batch and a single span within that batch triggers a write error (e.g., a type mismatch on one attribute), the **entire batch is rejected**. All other spans in the batch — which are valid and belong to completely unrelated traces — are silently lost.

## Context

I originally reported a specific trigger for this in #7860 (INT64 value for a Float64 column), which is being addressed. However, the batch rejection behavior is a separate, broader concern that will apply to any future write error on a single span.

## The problem

The OTEL SDK's `BatchSpanProcessor` batches spans by time, not by trace or user. A typical batch contains **hundreds to over a thousand spans from dozens of unrelated traces, users, and API requests** that simply happen to be flushed together.

When one span in a batch causes a write error:
- The datanode returns an error for the entire gRPC request
- The frontend propagates the error back to the OTLP endpoint
- The OTEL SDK receives a 500 and has no mechanism to retry individual spans
- **All spans in the batch are permanently lost**

This means a single problematic span from one user causes silent data loss for many other users whose spans had no issues at all.

## Expected behavior

GreptimeDB should accept valid spans from a batch even when some spans fail to write. This is commonly referred to as **partial success** in the OTLP specification.

The [OTLP specification explicitly supports partial success](https://opentelemetry.io/docs/specs/otlp/#partial-success) — the server can return a `partial_success` response indicating how many data points were rejected, while still accepting the rest. This allows the client SDK to log/report the partial failure without losing valid data.

## Impact

In our production environment:
- Our `BatchSpanProcessor` is configured with `maxExportBatchSize: 1024`, so each export can contain up to **1,024 spans**
- A single type mismatch on one span was causing ~9 batch rejections per day
- Each rejection drops an entire batch of unrelated traces from different users
- We rely on trace data for usage tracking and debugging, so this silent data loss has a direct business impact

## Proposed solution

1. Process each span (or at least each resource/scope group) independently during OTLP ingestion
2. Return an OTLP `partial_success` response when some spans fail, indicating the number of rejected items
3. Successfully persist all valid spans from the batch

This aligns with the OTLP spec and how other backends (e.g., Jaeger, Tempo) handle partial write failures.

shuiyisong:

close #7875 

This PR changes OTLP trace ingestion to preserve valid spans when part of a request fails, instead of failing the entire batch.
  - add OTLP trace partial success responses with rejected span counts and summarized error messages
  - ingest traces by OTLP resource/scope group, then subchunk each group for bounded writes
  - retry deterministic chunk failures span-by-span to salvage valid spans
  - discard ambiguous chunk failures without retrying and count those spans as rejected
  - split trace row building so main trace rows and auxiliary trace tables are handled independently
  - derive auxiliary trace tables only from spans whose main-table writes succeeded

## PR Checklist

### DIFF
--- a/src/frontend/src/instance/otlp.rs
+++ b/src/frontend/src/instance/otlp.rs
@@ -902,11 +902,7 @@ mod tests {
         let mut messages = Vec::new();
 
         for idx in 0..=4 {
-            Instance::push_trace_failure_message(
-                &mut messages,
-                label,
-                format!("failure-{idx}"),
-            );
+            Instance::push_trace_failure_message(&mut messages, label, format!("failure-{idx}"));
         }
 
         assert_eq!(messages.len(), 4);


fengys1996:
@codex review

src/frontend/src/instance/otlp.rs
```diff
@@ -210,6 +222,264 @@ impl OpenTelemetryProtocolHandler for Instance {
...
+            | StatusCode::Unsupported
+            | StatusCode::TableNotFound
+            | StatusCode::TableColumnNotFound
+            | StatusCode::TableColumnExists
+            | StatusCode::DatabaseNotFound => ChunkFailureReaction::RetryPerSpan,
+            _ => ChunkFailureReaction::DiscardChunk,
+        }
+    }
+
+    fn push_trace_failure_message(&self, messages: &mut Vec<String>, message: String) {
```
sunng87: I suggest to add a counter metric here to monitor the error rate. Because from now on the error will affect ingestion rate.

shuiyisong: https://github.com/GreptimeTeam/greptimedb/pull/7892/changes#diff-3f9441ac9d612bc1a227bb24a49452da7e54964e6f6fa5ea2469eb95996a9a12R470

Added a metrics here. We use counter for both `discard_chunk` and `span_failure`. One `discard_chunk` means a batch of records are lost while one `span_failure` just means one record is lost.

Should we set the lost record number here, or stay with the event counter.

sunng87: I this this counter should be good. We just want to monitor a rough rate

src/frontend/src/instance/otlp.rs
```diff
@@ -210,6 +233,270 @@ impl OpenTelemetryProtocolHandler for Instance {
...
+            table_name: &table_name,
+            is_trace_v1_model,
+        };
+        let mut ingest_state = TraceIngestState {
+            aux_data: TraceAuxData::default(),
+            outcome: TraceIngestOutcome::default(),
+            failure_messages: Vec::new(),
+        };
+
+        for group in groups {
```
killme2008, line 264: Looks we can refactor it by iterator:


```rust
use itertools::Itertools;

for group in groups {
    for chunk in &group.spans.into_iter().chunks(TRACE_INGEST_CHUNK_SIZE) {
        self.ingest_trace_chunk(
            &ingest_ctx,
            chunk.collect(),
            ctx.clone(),
            &mut ingest_state,
        )
        .await?;
    }
}
```

src/frontend/src/instance/otlp.rs
```diff
@@ -210,6 +233,270 @@ impl OpenTelemetryProtocolHandler for Instance {
...
+                .context(error::ExecuteGrpcQuerySnafu)
+        }
+    }
+
+    // TODO(shuiyisong): review this
+    fn classify_trace_chunk_failure(status: StatusCode) -> ChunkFailureReaction {
+        match status {
+            StatusCode::InvalidArguments
+            | StatusCode::InvalidSyntax
+            | StatusCode::Unsupported
```
killme2008, line 492: Retrying on "Unsupported" seems incorrect.

src/frontend/src/instance/otlp.rs
```diff
@@ -210,6 +233,270 @@ impl OpenTelemetryProtocolHandler for Instance {
...
+                .map_err(BoxedError::new)
+                .context(error::ExecuteGrpcQuerySnafu)
+        }
+    }
+
+    // TODO(shuiyisong): review this
+    fn classify_trace_chunk_failure(status: StatusCode) -> ChunkFailureReaction {
+        match status {
+            StatusCode::InvalidArguments
+            | StatusCode::InvalidSyntax
```
killme2008, line 491: Retrying on "InvalidSyntax" and `InvalidArguments`  seems incorrect.

src/servers/src/otlp/trace/v0.rs
```diff
@@ -38,56 +38,46 @@ const APPROXIMATE_COLUMN_COUNT: usize = 24;
...
 const MAX_TIMESTAMP: i64 = 4102444800000000000;
 
-/// Convert SpanTraces to GreptimeDB row insert requests.
-/// Returns `InsertRequests` and total number of rows to ingest
-pub fn v0_to_grpc_insert_requests(
-    request: ExportTraceServiceRequest,
-    _pipeline: PipelineWay,
-    _pipeline_params: GreptimePipelineParams,
-    table_name: String,
+pub fn v0_to_grpc_main_insert_requests(
```
killme2008, line 45: Why remove the docs?

src/servers/src/otlp/trace/v1.rs
```diff
@@ -41,64 +40,46 @@ const APPROXIMATE_COLUMN_COUNT: usize = 30;
...
-///    `resource_attributes.service_name`
-///
-/// For other compound data structures like span_links and span_events here we
-/// are still using `json` data structure.
-pub fn v1_to_grpc_insert_requests(
-    request: ExportTraceServiceRequest,
-    _pipeline: PipelineWay,
-    _pipeline_params: GreptimePipelineParams,
-    table_name: String,
+pub fn v1_to_grpc_main_insert_requests(
```
killme2008, line 47: The same question as above.

src/servers/src/otlp/trace.rs
```diff
@@ -65,27 +67,49 @@ pub const SPAN_STATUS_PREFIX: &str = "STATUS_CODE_";
...
 pub const SPAN_STATUS_ERROR: &str = "STATUS_CODE_ERROR";
 
-/// Convert SpanTraces to GreptimeDB row insert requests.
-/// Returns `InsertRequests` and total number of rows to ingest
-pub fn to_grpc_insert_requests(
-    request: ExportTraceServiceRequest,
-    pipeline: PipelineWay,
-    pipeline_params: GreptimePipelineParams,
-    table_name: String,
+#[derive(Debug, Default)]
```
killme2008, line 76: Better to document the struct and its functions.

src/frontend/src/instance/otlp.rs
```diff
@@ -210,6 +233,270 @@ impl OpenTelemetryProtocolHandler for Instance {
...
+            _ => ChunkFailureReaction::DiscardChunk,
+        }
+    }
+
+    fn push_trace_failure_message(messages: &mut Vec<String>, label: &str, message: String) {
+        OTLP_TRACES_FAILURE_COUNT.with_label_values(&[label]).inc();
+
+        if messages.len() < TRACE_FAILURE_MESSAGE_LIMIT {
+            messages.push(message);
+        }
```
killme2008, line 524: Better to add a debug log:
```suggestion
        } else {
             debug!()..
        }       
```

PR #7900, opened 2026-04-01
feat: add parquet nested leaf projection

fengys1996:

This PR adds nested parquet leaf projection support in mito2 parquet reads.

Instead of building the projection mask from parquet root indices, it introduces a read-column abstraction and converts projections into parquet leaf indices before calling `ProjectionMask::leaves(...)`.

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet/read_columns.rs
+++ b/src/mito2/src/sst/parquet/read_columns.rs
@@ -110,8 +110,8 @@ pub fn build_projection_mask(
     parquet_schema_desc: &SchemaDescriptor,
 ) -> ProjectionMask {
     if parquet_read_cols.has_nested() {
-        let leaf_indices = build_parquet_leaves_indices(parquet_schema_desc, &parquet_read_cols);
-        ProjectionMask::leaves(parquet_schema_desc, leaf_indices.into_iter())
+        let leaf_indices = build_parquet_leaves_indices(parquet_schema_desc, parquet_read_cols);
+        ProjectionMask::leaves(parquet_schema_desc, leaf_indices)
     } else {
         ProjectionMask::roots(parquet_schema_desc, parquet_read_cols.root_indices_iter())
     }
--- a/src/mito2/src/sst/parquet/reader.rs
+++ b/src/mito2/src/sst/parquet/reader.rs
@@ -425,7 +425,7 @@ impl ParquetReaderBuilder {
             read_format.projection_indices().iter().copied(),
         );
 
-        let projection_mask = build_projection_mask(&parquet_read_cols, &parquet_schema_desc);
+        let projection_mask = build_projection_mask(&parquet_read_cols, parquet_schema_desc);
         let selection = self
             .row_groups_to_read(&read_format, &parquet_meta, &mut metrics.filter_metrics)
             .await;


fengys1996:
@codex review

src/mito2/src/sst/parquet/read_columns.rs
```diff
@@ -0,0 +1,223 @@
...
+pub struct ParquetReadColumn {
+    /// Root field index in the parquet schema.
+    root_index: usize,
+    /// Nested paths to read under this root column.
+    ///
+    /// Each path includes the root column itself. For example, for a root
+    /// column `j`, path `["j", "a", "b"]` refers to `j.a.b`.
+    ///
+    /// If empty, the whole root column is read.
+    nested_paths: Vec<ParquetNestedPath>,
```
MichaelScofield, line 80: Am I understand it correctly that `nested_paths` can only be pointed to the very last leaves? What if I want to read the middle level, for example, read the whole field "b" in "{a: {b: {c: 1, d: 2}}}"?

fengys1996: Yes, it can. The code here performs matching based on prefixes. The code as follows
```rust
        if nested_paths
            .iter()
            .any(|nested_path| leaf_path.starts_with(nested_path))
        {
            leaf_indices.push(leaf_idx);
        }
```
And I can also add some tests to cover this case.

src/mito2/src/sst/parquet/read_columns.rs
```diff
@@ -0,0 +1,285 @@
...
+        &self.nested_paths
+    }
+}
+
+/// Builds parquet leaf-column indices from parquet read columns.
+pub fn build_parquet_leaves_indices(
+    parquet_schema_desc: &SchemaDescriptor,
+    projection: &ParquetReadColumns,
+) -> Vec<usize> {
+    let mut map = HashMap::new();
```
MichaelScofield:
```suggestion
    let mut map = HashMap::with_capacity(projection.cols.len());
```

sunng87:
@evenyag PTAL. 

And shall we merge this after 1.0 merge window?

evenyag:
> @evenyag PTAL.
> 
> And shall we merge this after 1.0 merge window?

I plan to merge it after we release 1.0.

sunng87:
@evenyag sounds good. We need to create a maintenance branch for 1.0

src/mito2/src/sst/parquet/read_columns.rs
```diff
@@ -0,0 +1,285 @@
...
+
+/// Builds parquet leaf-column indices from parquet read columns.
+pub fn build_parquet_leaves_indices(
+    parquet_schema_desc: &SchemaDescriptor,
+    projection: &ParquetReadColumns,
+) -> Vec<usize> {
+    let mut map = HashMap::with_capacity(projection.cols.len());
+    for col in &projection.cols {
+        map.insert(col.root_index, &col.nested_paths);
+    }
```
evenyag, line 128: Since root_index is an array index, can we use a `Vec<Option<>>` for the lookup map?

src/mito2/src/sst/parquet/read_columns.rs
```diff
@@ -0,0 +1,285 @@
...
+
+    let mut leaf_indices = Vec::new();
+    for (leaf_idx, leaf_col) in parquet_schema_desc.columns().iter().enumerate() {
+        let root_idx = parquet_schema_desc.get_column_root_idx(leaf_idx);
+        let Some(nested_paths) = map.get(&root_idx) else {
+            continue;
+        };
```
evenyag, line 135: We may have many leaf columns. Can we do a quick check to see if there is a nested path first in the projection? If not, we can return early in build_parquet_leaves_indices.

fengys1996: Added a fast path for without nested paths in https://github.com/GreptimeTeam/greptimedb/pull/7900/changes/551ddecbbfbffe6046a688ed9bc0115cd7f936d9

src/mito2/src/sst/parquet/read_columns.rs
fengys1996: In cases where the root indices are relatively sparse, it may not perform as well as a HashMap.

killme2008:
@evenyag PTAL

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -406,10 +421,17 @@ impl ParquetReaderBuilder {
...
+            read_format.projection_indices().iter().copied(),
+        );
+
+        let projection_mask = if parquet_projection.has_nested() {
+            let leaf_indices =
+                build_parquet_leaves_indices(parquet_schema_desc, &parquet_projection);
+            ProjectionMask::leaves(parquet_schema_desc, leaf_indices.iter().copied())
+        } else {
+            ProjectionMask::roots(parquet_schema_desc, parquet_projection.root_indices_iter())
+        };
```
evenyag: non-blocking: We can wrap this into a method of ParquetReadColumns.
```rust
parquet_projection.compute_projection_mask(parquet_schema_desc);
```

PR #7902, opened 2026-04-01
refactor(metric-engine): Refactor PendingRowsBatcher for better testability and benchmarking

v0y4g3r:

This PR refactors the `PendingRowsBatcher` to improve its maintainability, testability, and performance tracking.

### Key Changes:

- **Enhanced Error Handling**:
    - Introduced `Arc<Error>` to share error instances across all waiters of a batch, reducing redundant error string conversions and providing more structured error information.
    - Improved error propagation using `Snafu` context for better diagnostics in the write path.
- **Cleaned State Management**:
    - Refactored `PendingBatch` to use `Option<PendingBatch>` within the worker loop, simplifying initialization and draining logic using idiomatic `Option` methods.
- **Improved Testability and Benchmarking**:
    - Refactored `flush_batch_physical` to be more testable and exported it along with related internal types (`PhysicalTableMetadata`, `TableBatch`, etc.) to allow external benchmarking.
    - Added a new Criterion benchmark `flush_batch_physical` to measure the performance of physical batch flushing under various loads (number of logical tables and rows).
    - Added unit tests for `transform_logical_batches_to_physical` and `drain_batch`.

## PR Checklist

### DIFF
--- a/src/servers/src/pending_rows_batcher.rs
+++ b/src/servers/src/pending_rows_batcher.rs
@@ -1022,7 +1022,6 @@ async fn spawn_flush(
 }
 
 struct FlushRegionWrite {
-    row_count: usize,
     datanode: Peer,
     request: RegionRequest,
 }
@@ -1032,6 +1031,7 @@ struct PlannedRegionBatch {
     batch: RecordBatch,
 }
 
+#[cfg(test)]
 impl PlannedRegionBatch {
     fn num_rows(&self) -> usize {
         self.batch.num_rows()
@@ -1176,8 +1176,6 @@ async fn flush_region_writes_concurrently(
                 .with_label_values(&["flush_write_region"])
                 .start_timer();
             node_manager.handle(&write.datanode, write.request).await?;
-            FLUSH_TOTAL.inc();
-            FLUSH_ROWS.observe(write.row_count as f64);
         }
         return Ok(());
     }
@@ -1188,8 +1186,6 @@ async fn flush_region_writes_concurrently(
             .start_timer();
 
         node_manager.handle(&write.datanode, write.request).await?;
-        FLUSH_TOTAL.inc();
-        FLUSH_ROWS.observe(write.row_count as f64);
         Ok::<_, Error>(())
     });
 
@@ -1237,6 +1233,9 @@ async fn flush_batch(
     if result.is_err() {
         FLUSH_FAILURES.inc();
         FLUSH_DROPPED_ROWS.inc_by(total_row_count as u64);
+    } else {
+        FLUSH_TOTAL.inc();
+        FLUSH_ROWS.observe(total_row_count as f64);
     }
 
     debug!(
@@ -1505,7 +1504,6 @@ fn encode_region_write_requests(
     let mut region_writes = Vec::with_capacity(resolved_batches.len());
     for resolved in resolved_batches {
         let region_id = resolved.planned.region_id;
-        let row_count = resolved.planned.num_rows();
         let (schema_bytes, data_header, payload) = {
             let _timer = PENDING_ROWS_BATCH_FLUSH_STAGE_ELAPSED
                 .with_label_values(&["flush_physical_encode_ipc"])
@@ -1530,7 +1528,6 @@ fn encode_region_write_requests(
         };
 
         region_writes.push(FlushRegionWrite {
-            row_count,
             datanode: resolved.datanode,
             request,
         });
@@ -2039,7 +2036,6 @@ mod tests {
 
         let writes = vec![
             FlushRegionWrite {
-                row_count: 10,
                 datanode: Peer {
                     id: 1,
                     addr: "node1".to_string(),
@@ -2047,7 +2043,6 @@ mod tests {
                 request: RegionRequest::default(),
             },
             FlushRegionWrite {
-                row_count: 12,
                 datanode: Peer {
                     id: 2,
                     addr: "node2".to_string(),
@@ -2637,7 +2632,6 @@ mod tests {
         let writes = encode_region_write_requests(vec![resolved_batch]).unwrap();
 
         assert_eq!(1, writes.len());
-        assert_eq!(1, writes[0].row_count);
         assert_eq!(1, writes[0].datanode.id);
         let Some(region_request::Body::BulkInsert(request)) = &writes[0].request.body else {
             panic!("expected bulk insert request");


fengjiachun:
@codex review

src/servers/src/pending_rows_batcher.rs
```diff
@@ -1119,177 +1209,148 @@ async fn flush_batch(
...
     // into physical format and write them together.
     let physical_table_name = ctx
         .extension(PHYSICAL_TABLE_KEY)
         .unwrap_or(GREPTIME_PHYSICAL_TABLE)
         .to_string();
-    flush_batch_physical(
+    let partition_provider = PartitionManagerPhysicalFlushAdapter { partition_manager };
+    let node_requester = NodeManagerPhysicalFlushAdapter { node_manager };
+    let catalog_provider = CatalogManagerPhysicalFlushAdapter { catalog_manager };
+    let result = flush_batch_physical(
```
fengjiachun, line 1220: This refactor also changes the failure-isolation behavior of pending flushes.

Before this patch, failures during physical flush were handled at table/region granularity: a bad logical table or a failed region write would be recorded, but the rest of the batch could still continue. After switching this path to `Result<()>` + `?`, any error in transform / leader resolution / IPC encoding / region write aborts the whole batch immediately.

Since one pending batch can aggregate multiple logical tables, this turns a local failure or transient single-region issue into a full-batch failure, which looks like a behavior regression rather than a pure refactor.

Is that semantic change intentional?

src/servers/src/pending_rows_batcher.rs
```diff
@@ -1119,177 +1209,148 @@ async fn flush_batch(
...
+        &partition_provider,
+        &node_requester,
+        &catalog_provider,
     )
     .await;
 
     let elapsed = start.elapsed().as_secs_f64();
     FLUSH_ELAPSED.observe(elapsed);
+
+    if result.is_err() {
```
fengjiachun, line 1233: This changes the dropped-row accounting semantics in a problematic way.

`flush_region_writes_concurrently()` now increments `FLUSH_TOTAL/FLUSH_ROWS` immediately for each successful region write, but `flush_batch()` increments `FLUSH_DROPPED_ROWS` by `total_row_count` whenever the overall result is `Err`.

That means a partially successful flush will count some rows as both successfully flushed and dropped. The previous implementation accounted dropped rows per failed sub-write/table, which kept the metrics accurate.

src/servers/src/pending_rows_batcher.rs
```diff
@@ -86,6 +88,115 @@ pub trait PendingRowsSchemaAlterer: Send + Sync {
 
 pub type PendingRowsSchemaAltererRef = Arc<dyn PendingRowsSchemaAlterer>;
 
+#[derive(Clone)]
+pub struct PhysicalTableMetadata {
+    pub table_info: TableInfoRef,
+    pub name_to_ids: Option<HashMap<String, u32>>,
```
evenyag: Are they logical table id or column id?

src/servers/src/pending_rows_batcher.rs
```diff
@@ -747,16 +860,15 @@ fn start_worker(
                         Some(WorkerCommand::Submit { table_batches, total_rows, ctx, response_tx, _permit }) => {
                             idle_timer.as_mut().reset(tokio::time::Instant::now() + worker_idle_timeout);
 
-                            if batch.total_row_count == 0 {
-                                batch.created_at = Some(Instant::now());
-                                batch.ctx = Some(ctx);
+                            let pending_batch =batch.get_or_insert_with(||{
```
evenyag:
```suggestion
                            let pending_batch = batch.get_or_insert_with(||{
```

src/servers/src/pending_rows_batcher.rs
```diff
@@ -914,15 +1021,20 @@ async fn spawn_flush(
...
     request: RegionRequest,
 }
 
-enum FlushWriteResult {
-    Success { row_count: usize },
-    Failed { row_count: usize, message: String },
+struct PlannedRegionBatch {
+    region_id: RegionId,
+    row_count: usize,
+    batch: RecordBatch,
```
evenyag: Why don't we get the row count from the record batch?

src/servers/src/pending_rows_batcher.rs
```diff
@@ -1119,177 +1209,148 @@ async fn flush_batch(
...
-    let physical_table_info = physical_table.table_info();
-    let name_to_ids = match physical_table_info.name_to_ids() {
-        Some(ids) => ids,
-        None => {
-            record_failure!(
-                total_row_count,
-                format!(
-                    "Physical table '{}' has no column IDs for pending flush",
+            .await?
+            .context(error::InternalSnafu {
```
evenyag: Use with_context

src/servers/src/pending_rows_batcher.rs
```diff
@@ -1119,177 +1209,148 @@ async fn flush_batch(
...
-                )
-            );
-            return;
-        }
+                ),
+            })?
     };
 
+    let physical_table_info = physical_table.table_info;
+    let name_to_ids = physical_table.name_to_ids.context(error::InternalSnafu {
```
evenyag: with_context

src/servers/src/pending_rows_batcher.rs
```diff
@@ -1333,147 +1380,133 @@ async fn flush_batch_physical(
...
+
+    Ok(Some(PlannedRegionBatch {
+        region_id: RegionId::new(physical_table_id, region_number),
+        row_count,
+        batch: region_batch,
+    }))
+}
 
-    let mut region_writes = Vec::new();
+#[allow(clippy::too_many_arguments)]
```
evenyag: Do we still need this allow? The function only receives 4 arguments.

src/servers/src/pending_rows_batcher.rs
v0y4g3r: Yes. The previous approach records the first error and if that error present, we will send failed response to all pending clients, so every client will resend the request, even if some of them would have been successfully written. This behavior change will fail fast and cancel following writes to avoid duplication.

src/servers/src/pending_rows_batcher.rs
v0y4g3r: fixed in 4116726c03ffb65f1880b42e2bf5d5768889ed16

src/servers/src/pending_rows_batcher.rs
v0y4g3r: fixed in 4116726c03ffb65f1880b42e2bf5d5768889ed16

src/servers/src/pending_rows_batcher.rs
v0y4g3r: fixed in cdb63a53ddbf391e3887e0999aa0398213bcac26

fengjiachun (approved):
LGTM

PR #7903, opened 2026-04-01
perf: reduce full reshuffle to partition coalesce op

waynexia:

Works when two partitions do not match, but can be transformed with merely a coalesce operation.

## PR Checklist


src/query/src/optimizer/reduce_aggregate_repartition.rs
```diff
@@ -0,0 +1,505 @@
...
+
+    fn contains_promql_exec_deep(plan: &Arc<dyn ExecutionPlan>) -> bool {
+        if Self::is_promql_vector_exec(plan) {
+            return true;
+        }
+
+        plan.children()
+            .into_iter()
+            .any(Self::contains_promql_exec_deep)
+    }
```
evenyag: They look almost the same. Is it still correct if we always use contains_promql_vector_exec?

evenyag (approved):

src/query/src/optimizer/reduce_aggregate_repartition.rs
```diff
@@ -0,0 +1,505 @@
...
+
+            Ok(Transformed::yes(Arc::new(new_agg)))
+        })
+        .data()
+    }
+
+    fn can_reduce_repartition(repartition_exec: &RepartitionExec) -> bool {
+        let has_direct_promql_input =
+            Self::has_direct_promql_partial_input(repartition_exec.input());
+        if Self::contains_promql_exec_deep(repartition_exec.input()) {
```
discord9: should we make the condition stricter here, and add a few focused regression tests for PromQL/TQL cases as well?

discord9 (approved):
rest LGTM

waynexia:
this may affects other unexpected cases, close for now

PR #7909, opened 2026-04-02
feat!: switch default sst format to flat

evenyag:

- https://github.com/GreptimeTeam/greptimedb/issues/7732
- https://github.com/GreptimeTeam/greptimedb/pull/7901

This PR updates Mito's SST format behavior in two ways:

1. Allow altering a region from flat back to primary_key.
2. Make flat the default experimental SST format for new regions.

As a follow-up to https://github.com/GreptimeTeam/greptimedb/pull/7901, it also removes the static parallel_scan_channel_size config and derives the parallel scan channel size dynamically from the estimated split batch size during flat scans, keeping buffering aligned with the actual batch shape.

Changes
- Allow ALTER TABLE ... SET 'sst_format' = 'primary_key' for regions currently using flat.
- Persist and reopen regions correctly after flat -> primary_key format changes.
- Change region_engine.mito.default_experimental_flat_format default from false to true.
- Remove region_engine.mito.parallel_scan_channel_size from config, examples, and docs.
- Compute parallel scan channel size dynamically from the estimated split batch size for flat scans.
- Add unit tests for split estimation and channel-size calculation.
- Add engine and SQL integration coverage for flat -> primary_key alter flows.

Testing
- Added engine tests for flat -> primary_key alter with and without flush.
- Added SQL tests covering alter-format behavior and reads after subsequent writes/flushes.
- Added unit tests for split heuristics and computed channel size.

## PR Checklist

### DIFF
--- a/src/mito2/src/engine/create_test.rs
+++ b/src/mito2/src/engine/create_test.rs
@@ -401,7 +401,7 @@ async fn test_engine_create_with_format_one_case(create_format: &str, default_fl
     let mut env = TestEnv::new().await;
     let engine = env
         .create_engine(MitoConfig {
-            default_flat_format: default_flat_format,
+            default_flat_format,
             ..Default::default()
         })
         .await;


src/mito2/src/region/opener.rs
```diff
@@ -626,8 +626,10 @@ pub(crate) fn sanitize_region_options(manifest: &RegionManifest, options: &mut R
             manifest.sst_format,
             manifest.metadata.region_id,
         );
-        options.sst_format = Some(manifest.sst_format);
     }
+    // Always set sst_format from manifest to ensure it's explicitly stored,
+    // even when the default matches the manifest value.
+    options.sst_format = Some(manifest.sst_format);
```
killme2008, line 632: Should we add a debug log, or are there any existing debug logs?

tests/cases/standalone/common/alter/alter_format.result
```diff
@@ -116,11 +116,48 @@ SELECT i, h FROM test_alt_format;
 | 23 | 13 |
 +----+----+
 
--- not allow to change from flat to primary_key
--- SQLNESS REPLACE \d+\(\d+,\s+\d+\) REDACTED
+-- allow to change from flat to primary_key
 ALTER TABLE test_alt_format SET 'sst_format' = 'primary_key';
```
killme2008, line 141: Adds a `show create table` to ensure option is set.

src/mito2/src/region/opener.rs
evenyag: There is a warning log above.

src/mito2/src/read/seq_scan.rs
```diff
@@ -332,8 +342,11 @@ impl SeqScan {
                     file_scan_semaphore.clone(),
                 ).await?;
 
+                let channel_size = compute_parallel_channel_size(
+                    split_batch_size.unwrap_or(DEFAULT_READ_BATCH_SIZE),
+                );
```
fengjiachun, line 347: When `split_batch_size` is `None`, this falls back to `DEFAULT_READ_BATCH_SIZE`, which makes `compute_parallel_channel_size()` return `2`. That means we are shrinking the channel for normal flat scans that do not split batches, not just tuning the split case. Previously the default buffer was much larger, so this looks like a behavior/perf change outside the stated scope of "derive the channel size from the estimated split batch size". Should we keep the old default when splitting is not applied, and only use the dynamic size for the split path?

src/mito2/src/config.rs
```diff
@@ -230,7 +225,7 @@ impl Default for MitoConfig {
             vector_index: VectorIndexConfig::default(),
             memtable: MemtableConfig::default(),
             min_compaction_interval: Duration::from_secs(0),
-            default_experimental_flat_format: false,
+            default_experimental_flat_format: true,
```
fengjiachun: Changing the global default to flat also affects callers that do not explicitly set `sst_format`. In particular, metric-engine data region creation still injects `memtable.type = partition_tree` but does not set `sst_format`, and under flat format `builder_for_options()` will override the memtable config and force `BulkMemtable`. Is that behavior change intentional for metric-engine as well? If yes, I think we need an explicit test for that path; otherwise we may want to keep those regions on `primary_key` unless they opt in.

src/mito2/src/read/seq_scan.rs
evenyag: In https://github.com/GreptimeTeam/greptimedb/pull/7901 we changed the channel size for large batch from 2 to default. In this PR, we adjust it according to the estimated batch size.

src/mito2/src/config.rs
evenyag: Yes, this is expected. We'll deprecate partition_tree soon.

evenyag:
Removed experimental from the option in https://github.com/GreptimeTeam/greptimedb/pull/7909/commits/96f84b4fea6cdcf55754377e6f22e27fd26c181a.

Flat is no longer experimental now.

v0y4g3r (approved):

PR #7918, opened 2026-04-07
fix(repartition): harden repartition rollback paths

WenyXu:

This PR improves rollback handling in `meta-srv` repartition procedures and adds regression coverage for rollback and retry paths.
- restore repartition group metadata on rollback
- deallocate allocated regions when repartition fails
- add tests for group rollback, parent rollback, and retry scenarios

## PR Checklist

### DIFF
--- a/src/meta-srv/src/procedure/repartition/group/update_metadata/rollback_staging_region.rs
+++ b/src/meta-srv/src/procedure/repartition/group/update_metadata/rollback_staging_region.rs
@@ -34,7 +34,7 @@ impl UpdateMetadata {
     fn rollback_staging_region_routes(
         group_id: GroupId,
         sources: &[RegionDescriptor],
-        target_routes: &[RegionRoute],
+        original_target_routes: &[RegionRoute],
         pending_deallocate_region_ids: &[RegionId],
         current_region_routes: &[RegionRoute],
     ) -> Result<Vec<RegionRoute>> {
@@ -50,13 +50,16 @@ impl UpdateMetadata {
                     region_id: source.region_id,
                 },
             )?;
+            // Clean leader staging state for source regions.
             region_route.clear_leader_staging();
             if pending_deallocate_region_ids.contains(&source.region_id) {
+                // Clean ignore all writes state for source regions if it's pending to be deallocated,
+                // which means the source region is merged into the target region.
                 region_route.clear_ignore_all_writes();
             }
         }
 
-        for target in target_routes {
+        for target in original_target_routes {
             let region_route = region_routes_map.get_mut(&target.region.id).context(
                 error::RepartitionTargetRegionMissingSnafu {
                     group_id,
@@ -64,9 +67,11 @@ impl UpdateMetadata {
                 },
             )?;
 
+            // Revert the partition expression and write route policy to the original value for the target region.
             region_route.region.partition_expr = target.region.partition_expr.clone();
             region_route.write_route_policy = target.write_route_policy;
 
+            // Clean leader staging state for target regions.
             region_route.clear_leader_staging();
         }
 
--- a/src/meta-srv/src/procedure/repartition/group.rs
+++ b/src/meta-srv/src/procedure/repartition/group.rs
@@ -370,7 +370,7 @@ impl Context {
 pub struct GroupPrepareResult {
     /// The validated source region routes.
     pub source_routes: Vec<RegionRoute>,
-    /// The validated target region routes.
+    /// Validated target region routes used for metadata rollback (logical rollback).
     pub target_routes: Vec<RegionRoute>,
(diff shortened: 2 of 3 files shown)


WenyXu:
@codex review

WenyXu:
@codex review

WenyXu:
@codex review

WenyXu:
@codex review

src/meta-srv/src/procedure/repartition.rs
```diff
@@ -456,6 +481,133 @@ impl RepartitionProcedure {
...
+            self.context
+                .update_table_route(&table_route_value, new_region_routes, HashMap::new())
+                .await
+                .map_err(BoxedError::new)
+                .with_context(|_| error::RetryLaterWithSourceSnafu {
+                    reason: format!(
+                        "Failed to rollback allocated region routes for repartition table: {}",
+                        table_id
+                    ),
+                })?;
```
fengjiachun, line 602: it logs and ignores deallocate_regions() failures, then still removes the allocated region routes from table metadata. That can leave newly allocated physical regions alive on datanodes but permanently orphaned from metadata, and a later rollback retry cannot discover them anymore because the routes have already been deleted.  Is that by design?

src/meta-srv/src/procedure/repartition/collect.rs
```diff
@@ -104,7 +104,19 @@ impl State for Collect {
...
+            ctx.persistent_ctx
+                .unknown_procedures
+                .extend(self.unknown_procedures.iter());
+            return crate::error::UnexpectedSnafu {
+                violated: format!(
+                    "Repartition groups failed or became unknown, table_id: {}, failed: {}, unknown: {}",
+                    table_id, failed, unknown
+                ),
+            }
+            .fail();
```
fengjiachun, line 118: `Collect` now treats `procedure_state_receiver()==None` as unknown and fails the whole repartition, but the procedure framework removes finished child procedures during manager recovery. After a restart, a parent recovered in `Collect` can therefore lose the in-memory receiver for an already-finished child and incorrectly roll back a repartition that had succeeded. Is that correct?

WenyXu: Good catch [a5c8f8a](https://github.com/GreptimeTeam/greptimedb/pull/7918/commits/a5c8f8a26077454f2455cf3bba63f74fd3c1bd91)

src/meta-srv/src/procedure/repartition.rs
WenyXu: Yes, during rollback it might not retry later (the rollback may be triggered after the retry limit is reached), so we do our best to drop the region and eventually remove the metadata.

fengjiachun (approved):
Rest LGTM

src/meta-srv/src/procedure/repartition.rs
```diff
@@ -74,6 +76,10 @@ pub struct PersistentContext {
     pub table_name: String,
     pub table_id: TableId,
     pub plans: Vec<RepartitionPlanEntry>,
+    #[serde(default)]
+    pub failed_procedures: Vec<ProcedureMeta>,
+    #[serde(default)]
+    pub unknown_procedures: Vec<ProcedureMeta>,
```
evenyag, line 84: Document these fields.

src/meta-srv/src/procedure/repartition/dispatch.rs
```diff
@@ -106,7 +106,7 @@ impl State for Dispatch {
 
         Ok((
             Box::new(Collect::new(procedure_metas)),
-            Status::suspended(procedures, true),
```
evenyag, line 109: Why do we change it to false?

src/meta-srv/src/procedure/repartition/deallocate_region.rs
```diff
@@ -88,6 +88,8 @@ impl State for DeallocateRegion {
             &ctx.persistent_ctx.schema_name,
             &ctx.persistent_ctx.table_name,
         );
+        let _operating_guards =
+            Context::register_operating_regions(&ctx.memory_region_keeper, &deallocating_regions)?;
```
evenyag: Why do we need guards here? Maybe need some documents.

src/meta-srv/src/procedure/repartition/group/update_metadata/rollback_staging_region.rs
```diff
@@ -60,6 +63,10 @@ impl UpdateMetadata {
                     region_id: target.region.id,
                 },
             )?;
+
+            region_route.region.partition_expr = target.region.partition_expr.clone();
+            region_route.write_route_policy = target.write_route_policy;
```
evenyag, line 72: Add some comments.

PR #7929, opened 2026-04-08
docs: rfc for partial eval with stats

discord9:

as title, partial eval using stats for some simple query

## PR Checklist

### DIFF
--- a/docs/rfcs/2026-04-08-aggregate-stats-physical-pass.md
+++ b/docs/rfcs/2026-04-08-aggregate-stats-physical-pass.md
@@ -1,0 +1,363 @@
+---
+Feature Name: Aggregate Statistics Physical Optimizer Pass
+Tracking Issue: TBD
+Date: 2026-04-08
+Author: @OpenAgent
+---
+
+# Summary
+
+This RFC proposes a new physical-plan optimization pass for aggregate queries in GreptimeDB.
+Instead of relying primarily on DataFusion's relation-level `Statistics/ColumnStatistics` to optimize whole-query planning, the new pass opportunistically consumes per-file statistics during physical optimization and execution planning.
+
+If some files have usable statistics for an aggregate, GreptimeDB will avoid scanning those files and synthesize partial aggregate state from their metadata. Files whose statistics are missing, incompatible, or unsafe to use will still be scanned normally. The two paths are then merged with the existing state/merge aggregate wrapper mechanism so the query remains correct while extracting as much value as possible from available statistics.
+
+# Motivation
+
+Today GreptimeDB's aggregate optimization is still mostly constrained by DataFusion's `ColumnStatistics/Statistics` model.
+That model is useful for planner-level estimation and some coarse optimizations, but it is not a good fit for GreptimeDB's desired behavior:
+
+1. We want to optimize at the **physical file level**, not only at the relation level.
+2. We want to support **mixed execution**:
+   - files with usable statistics -> answer from stats
+   - files without usable statistics -> fall back to scan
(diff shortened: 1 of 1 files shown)


docs/rfcs/2026-04-08-aggregate-stats-physical-pass.md
```diff
@@ -0,0 +1,370 @@
...
+
+# Non-Goals
+
+1. This RFC does **not** attempt to replace all DataFusion statistics usage.
+2. This RFC does **not** try to optimize arbitrary expressions above aggregates.
+3. This RFC does **not** guarantee support for every aggregate function in v1.
+4. This RFC does **not** support ordinary `GROUP BY` aggregation in v1.
+5. This RFC does **not** attempt tag-column aggregate optimization in v1.
+6. This RFC does **not** require statistics reconstruction for files whose metadata has already lost the needed semantics.
+7. This RFC does **not** require support for non-append-only tables in v1.
```
killme2008, line 44: Can we support non-append-only tables in the future?

We could detect whether SST files contain deletions or updates during creation and use that information to determine if the statistics state can be applied.

killme2008 (approved):
LGTM

Let's go

killme2008:
@waynexia @evenyag PTAL

docs/rfcs/2026-04-08-aggregate-stats-physical-pass.md
evenyag: Not only deletion and updates, we also need to consider overlapping files. There may be duplicated rows in the file, too.

docs/rfcs/2026-04-08-aggregate-stats-physical-pass.md
```diff
@@ -0,0 +1,370 @@
...
+
+So the realistic MVP should be framed as a support matrix, not as a blanket promise for all `MIN/MAX/COUNT` forms.
+
+| Query shape | v1 status | Notes |
+|---|---|---|
+| `COUNT(*)` without `GROUP BY` | Supported candidate | Requires exact `FileMeta.num_rows`; old files with unknown row counts fall back to scan |
+| `MIN/MAX(time_index)` without `GROUP BY` | Supported candidate | Can rely on file/range metadata or equivalent parquet time stats when available |
+| `MIN/MAX(field_col)` without `GROUP BY` | Supported candidate | Only when all needed row-group min/max stats are available and semantically safe |
+| `COUNT(field_col)` / `COUNT(time_index)` without `GROUP BY` | Supported candidate | Only when exact null-count stats are available for all needed row groups |
+| `MIN/MAX/COUNT` on tag columns | Not supported in v1 | Current tag stats are incomplete for this feature |
```
evenyag, line 162: I think we can support count without filters for tags.

docs/rfcs/2026-04-08-aggregate-stats-physical-pass.md
```diff
@@ -0,0 +1,370 @@
...
+
+```sql
+select max(v) from t;
+```
+
+the optimized execution is conceptually equivalent to:
+
+1. compute state from scanned files: `__max_state(v)`
+2. compute state from stats-only files: `__max_state(file_max(v))`
+3. merge them with `__max_merge(...)`
```
evenyag, line 88: A problem is that file statistics may not be available in the optimization phase. We may need to fetch them from files if they are not in cache.

discord9:
draft as found out the impl should widen range to get more query optimizable like simple aggr but with certain filter (i.e. for time range or partition filter)

Issue #7943, opened 2026-04-09
Inconsistent Full-Text Search Results Within Single Log Entry


zillaer:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Datanode

### Minimal reproduce step

index configration：
`FULLTEXT INDEX WITH(analyzer = 'Chinese', backend = 'bloom', case_sensitive = 'true', false_positive_rate = '0.01', granularity = '10240')`

log content：
`[2026/04/09/ 13:56:11.031]2026-04-09 13:56:11.031 - [ trace_id=340a6a44b0bd8e37bb7697ss7da61ff0 span_id=085ff5ttf1e0a23b trace_flags=01] - [http-nio-8081-exec-16] INFO c.h.p.xx.web.service.impl.CCCXForwardKKKServiceImpl.pushout(188) - 登录手机号18888888888的动态key：829889AC8`

Support for result-returning SQL queries：
`SELECT "real_log_time" as "timestamp", "message" as "body" FROM "public"."kafka_logs_table" WHERE ( "real_log_time" >= '2026-04-08T08:21:38.348Z' AND "real_log_time" <= '2026-04-09T08:21:38.348Z' ) AND ( "message" @@ '00179696f06a114d2aa98b8ace1c4462' ) ORDER BY timestamp DESC LIMIT 1000;`

Not support for result-returning SQL queries：
`SELECT "real_log_time" as "timestamp", "message" as "body" FROM "public"."kafka_logs_table" WHERE ( "real_log_time" >= '2026-04-08T08:21:38.348Z' AND "real_log_time" <= '2026-04-09T08:21:38.348Z' ) AND ( "message" @@ '手机' ) ORDER BY timestamp DESC LIMIT 1000;`

`SELECT "real_log_time" as "timestamp", "message" as "body" FROM "public"."kafka_logs_table" WHERE ( "real_log_time" >= '2026-04-08T08:21:38.348Z' AND "real_log_time" <= '2026-04-09T08:21:38.348Z' ) AND ( "message" @@ '18888888888' ) ORDER BY timestamp DESC LIMIT 1000;`

### What did you expect to see?

The system should support searching by any arbitrary keyword found in the logs. The search index needs to cover all terms to ensure that any keyword, rather than just partial ones, yields the correct results.

### What did you see instead?

Inconsistent Full-Text Search Results Within Single Log Entry

### What operating system did you use?

centos7

### What version of GreptimeDB did you use?

1.0rc2

[...]

PR #7952, opened 2026-04-10
fix: match term zh

discord9:

#7943 

as title, a more intuitive way when match terms with chinese character, this partially solve the issue of bad index and query experience of full text index, but more change for full text index might be needed for analyzer=chinese

## PR Checklist

### DIFF
--- a/src/common/function/src/scalars/matches_term.rs
+++ b/src/common/function/src/scalars/matches_term.rs
@@ -20,6 +20,8 @@ use datafusion_common::arrow::compute;
 use datafusion_common::arrow::datatypes::DataType;
 use datafusion_common::{DataFusionError, ScalarValue};
 use datafusion_expr::{ColumnarValue, ScalarFunctionArgs, Signature, Volatility};
+use icu_properties::props::Script;
+use icu_properties::{CodePointMapData, CodePointMapDataBorrowed};
 use memchr::memmem;
 
 use crate::function::Function;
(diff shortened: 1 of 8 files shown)


src/common/function/src/scalars/matches_term.rs
```diff
@@ -43,9 +44,8 @@ use crate::function_registry::FunctionRegistry;
 /// BooleanVector where each element indicates if the corresponding text
 /// contains an exact match of the term, following these rules:
 /// 1. Exact substring match found (case-sensitive)
-/// 2. Match boundaries are either:
-///    - Start/end of text
-///    - Any non-alphanumeric character (including spaces, hyphens, punctuation, etc.)
+/// 2. For ASCII-only terms, adjacent ASCII word characters block the match
+/// 3. For Han-containing terms, contiguous substring match is sufficient
```
evenyag, line 50: Should we extend this to non-alphanumeric terms?

discord9: > Should we extend this to non-alphanumeric terms?

I don't see the necessarily of extend this to non-alphanumeric term. This fix is purely because word-boundary checks based on alphanumeric characters are a poor fit for Chinese text. Non-alphanumeric terms are much more heterogeneous, and relaxing them the same way would broaden matching behavior without a clear product need.

src/common/function/src/scalars/matches_term.rs
```diff
@@ -215,28 +217,113 @@ impl Function for MatchesTermFunction {
...
-    ends_with_non_alnum: bool,
+    term_kind: TermKind,
+    starts_with_other: bool,
+    ends_with_other: bool,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum CharClass {
+    AsciiWord,
+    Han,
```
killme2008, line 237: Should we include all CJK terms?

killme2008 (approved):
LGTM

src/common/function/src/scalars/matches_term.rs
discord9: > Should we include all CJK terms?

I would prefer not to. The change is meant to address the fact that the current boundary logic is a poor fit for Han text specifically. Extending it to all CJK scripts would broaden behavior quite a bit, and scripts like Kana or Hangul may need different treatment. If we want that, I think it should be a separate follow-up with concrete cases.

killme2008: Yes, please create an issue to track it, we must support CJK characters(CJK stands for Chinese, Japanese, and Korean).

evenyag (approved):

PR #7954, opened 2026-04-11
feat(mito2): add PK-range-aware TWCS overlap handling

v0y4g3r:

- This PR adds primary-key range awareness to mito2 compaction so TWCS does not treat files as overlapping when they only overlap in time but are disjoint by primary key. The intent is to reduce unnecessary compaction work while preserving correctness.
- It extracts `__primary_key` min/max range from parquet metadata, stores that range on `FileHandle`, and propagates it through flush/compaction/open paths (including cache warm/preload paths) so compaction has PK range information available consistently.
- It updates overlap logic in compaction run selection:
  - keep strict overlap semantics for sorted-run construction (`<`) to preserve existing run behavior,
  - use inclusive boundary semantics in `find_overlapping_items` (`<=`) to preserve original boundary-touching behavior,
  - and apply PK-range checks for `FileGroup` overlap decisions.
- It updates TWCS window overlap handling to avoid false transitive overlap marking, and ensures append-mode behavior remains stable with the new PK-aware overlap logic.
- It extends/updates unit and integration coverage around PK-disjoint overlap, append-mode compaction behavior, and GC scenarios.
- Compatibility: no external API contract changes are introduced. Data/schema compatibility is preserved; changes are internal compaction-selection behavior and metadata propagation.

## PR Checklist


v0y4g3r:
@codex review

evenyag:
@codex review

src/mito2/src/sst/file.rs
```diff
@@ -36,6 +38,33 @@ use crate::cache::file_cache::{FileType, IndexKey};
...
+/// Custom serde functions for Bytes fields serialized as base64 strings.
+fn serialize_bytes_option<S>(bytes: &Option<Bytes>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    match bytes {
+        None => serializer.serialize_none(),
+        Some(b) => serializer.serialize_some(&BASE64_STANDARD.encode(b)),
+    }
+}
```
evenyag, line 50: Can we serde using hex?

src/mito2/src/sst/version.rs
```diff
@@ -45,11 +45,6 @@ impl SstVersion {
         &self.levels
     }
 
-    /// Add files to the version. If a file with the same `file_id` already exists,
-    /// it will be overwritten with the new file.
-    ///
-    /// # Panics
-    /// Panics if level of [FileMeta] is greater than [MAX_LEVEL].
```
evenyag, line 52: Why do we remove this?

src/mito2/src/sst/version.rs
```diff
@@ -75,10 +71,10 @@ impl SstVersion {
                         }
                     } else {
                         // include case like old file have no index or index is outdated
-                        *f = FileHandle::new(file.clone(), file_purger.clone());
+                        *f = build_handle();
                     }
                 })
-                .or_insert_with(|| FileHandle::new(file.clone(), file_purger.clone()));
+                .or_insert_with(build_handle);
```
evenyag: Looks like this change is unnecessary.

src/mito2/src/region/version.rs
```diff
@@ -475,7 +474,6 @@ impl VersionBuilder {
         self
     }
 
-    /// Add files to the builder.
```
evenyag, line 478: non-blocking: Should we modify them?

src/mito2/src/region/opener.rs
```diff
@@ -1021,11 +1023,14 @@ async fn preload_parquet_meta_cache_for_files(
...
-        if cache_manager
+        if let Some(metadata) = cache_manager
             .get_parquet_meta_data(file_id, &mut cache_metrics, Default::default())
             .await
-            .is_some()
         {
+            if let Some(primary_key_range) = extract_primary_key_range(&metadata, &region_metadata)
+            {
+                file_handle.set_primary_key_range(primary_key_range);
+            }
```
evenyag, line 1035: We should only extract ranges when there are no primary key ranges in the meta of the file handle.

src/mito2/src/compaction/twcs.rs
```diff
@@ -103,7 +104,6 @@ impl TwcsPicker {
                 if run.items().len() < self.trigger_file_num {
                     continue;
                 }
-                // no overlapping files, try merge small files
                 merge_seq_files(run.items(), self.max_output_file_size)
```
evenyag, line 107: Keep the comments?

src/mito2/src/compaction/run.rs
```diff
@@ -281,21 +335,19 @@ where
...
+            if current_run.items.is_empty() {
+                // current run is empty, just add current_item
+                selected.set(true);
+                current_run.push_item(item.clone());
+            } else {
+                // the current item does not overlap with any item in current run,
+                // then it belongs to current run.
+                let overlaps_any = current_run.items.iter().any(|i| i.overlap(item));
+                if !overlaps_any {
+                    // does not overlap, push to current run
```
evenyag, line 348: Can't we only compare the last item? Due to primary key ranges?

src/mito2/src/sst/file.rs
v0y4g3r: > Can we serde using hex?

HEX encoded bytes is 200% of the size of original data, while base64 is about 130%.

src/mito2/src/compaction/twcs.rs
v0y4g3r: fixed in 5939e9e931b2a34d1424d6ab43d7c830b86d5b57

src/mito2/src/region/version.rs
v0y4g3r: fixed in 5939e9e931b2a34d1424d6ab43d7c830b86d5b57

src/mito2/src/sst/version.rs
v0y4g3r: fixed in 5939e9e931b2a34d1424d6ab43d7c830b86d5b57

src/mito2/src/sst/version.rs
v0y4g3r: fixed in 5939e9e931b2a34d1424d6ab43d7c830b86d5b57

src/mito2/src/region/opener.rs
v0y4g3r: 5787dc488c62b78ba613657db369e2da67d030db

src/mito2/src/compaction/run.rs
v0y4g3r: Yes. I added some docs for that in ee738f5d1c8560560ab349ba99946f0923983bff

src/mito2/src/region/version.rs
```diff
@@ -444,7 +444,6 @@ impl VersionBuilder {
         self
     }
 
-    /// Apply edit to the builder.
```
evenyag, line 447: Keep this?

src/mito2/src/sst/version.rs
```diff
@@ -78,7 +78,9 @@ impl SstVersion {
                         *f = FileHandle::new(file.clone(), file_purger.clone());
                     }
                 })
-                .or_insert_with(|| FileHandle::new(file.clone(), file_purger.clone()));
+                .or_insert_with(||{
+                    FileHandle::new(file.clone(), file_purger.clone())
+                });
```
evenyag: Strange format

evenyag (approved):

src/mito2/src/region/opener.rs
```diff
@@ -1021,11 +1023,16 @@ async fn preload_parquet_meta_cache_for_files(
...
+        if let Some(metadata) = cache_manager
             .get_parquet_meta_data(file_id, &mut cache_metrics, Default::default())
             .await
-            .is_some()
         {
+            if file_handle.primary_key_range().is_none()
+                && let Some(primary_key_range) =
+                    extract_primary_key_range(&metadata, &region_metadata)
+            {
+                file_handle.set_primary_key_range(primary_key_range);
```
fengjiachun, line 1034: Question: Should PK-range backfill depend on index-cache preload?

This new legacy-SST PK-range backfill runs only through `preload_parquet_meta_cache_for_files()`. Is it intentional that it is therefore skipped whenever `preload_index_cache` is false in `maybe_preload_parquet_meta_cache()`? That flag appears to control index-cache preload, but with this change it also decides whether pre-existing SSTs ever get PK ranges populated. If that coupling is unintended, PK-aware TWCS would only become effective for newly flushed/compacted files in that configuration.

src/mito2/src/sst/file.rs
```diff
@@ -498,6 +576,14 @@ impl FileHandle {
     pub fn is_deleted(&self) -> bool {
         self.inner.deleted.load(Ordering::Relaxed)
     }
+
+    pub fn primary_key_range(&self) -> Option<(Bytes, Bytes)> {
+        self.inner.primary_key_range.read().unwrap().clone()
+    }
+
+    pub(crate) fn set_primary_key_range(&self, primary_key_range: (Bytes, Bytes)) {
```
fengjiachun, line 584: This only updates the runtime FileHandle state, while `meta_ref()` still returns the original `FileMeta`. Should callers that clone or persist `FileMeta` also observe the backfilled PK range?

fengjiachun (approved):
Others LGTM

src/mito2/src/region/opener.rs
v0y4g3r: Normally newly flushed files will bring pk range inside their `FileMeta` instead of `FileHandle`. So the preload here is just a best-effort attempt to recover pk range if the parquet file metadata is present in cache.

src/mito2/src/sst/file.rs
v0y4g3r: > This only updates the runtime FileHandle state, while `meta_ref()` still returns the original `FileMeta`. Should callers that clone or persist `FileMeta` also observe the backfilled PK range?

Setting pk range into `FileMeta` here does not affect the persisted file meta in manifest. The `primary_key_range` in `FileHandle` serves just as a fallback path for legacy files.

PR #7955, opened 2026-04-11
fix: randomize standalone test ports in cli export test

v0y4g3r:

- Replace the hard-coded standalone test ports in `test_export_create_table_with_quoted_names` with dynamically selected free ports.
- Add a small helper that probes a random port offset and only uses it after all standalone service ports can be bound locally.
- This reduces flaky failures caused by port collisions when the `cmd` crate tests run in parallel or share CI hosts with other processes.
- This PR does not change runtime behavior, public APIs, or persisted data formats; it only makes the test setup more reliable.

## PR Checklist

### DIFF
--- a/src/cmd/src/cli.rs
+++ b/src/cmd/src/cli.rs
@@ -102,31 +102,79 @@ impl Command {
 
 #[cfg(test)]
 mod tests {
+    use std::net::TcpListener;
+    use std::ops::RangeInclusive;
+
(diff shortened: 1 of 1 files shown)


MichaelScofield (approved):
There's a same utility function in sqlness-runner: `server_mode::get_unique_random_port`, better reuse it.

evenyag (approved):

PR #7959, opened 2026-04-13
feat: support nested projection in mito2 read path

fengys1996:

This PR adds nested projection support in the Mito read path.

Previously, the scan path only propagated root column ids to memtables and SST readers. While this was sufficient for root-level projection, nested projection information was lost before reaching the Parquet reader, preventing precise pruning of nested leaf columns.

### The execution flow is as follows:

- Build `ReadColumns` from the pushed-down projection, including nested paths.
- Merge in columns referenced by pushed-down predicates, ensuring filter columns are still read even if they are not projected.
- Propagate `ReadColumns` through `ScanInput`, `FlatProjectionMapper`, `FlatReadFormat`, and `ParquetReaderBuilder`.
- Convert logical `ReadColumns` into `ParquetReadColumns` during format projection planning.
- Build the Parquet projection mask from `ParquetReadColumns`, enabling nested leaf projection pushdown.

### Limitations / Not covered in this PR:

- Nested-path projection is currently applied only to Parquet data pruning. The output schema is still not pruned accordingly, so after reading SST batches we must align struct arrays to the expected output schema. 
- Nested projection is not applied to memtables.
- For pushed-down predicates, only column ids are extracted; nested paths are not yet considered.

## PR Checklist

### DIFF
--- a/src/mito2/src/read/flat_projection.rs
+++ b/src/mito2/src/read/flat_projection.rs
@@ -33,6 +33,7 @@ use store_api::storage::ColumnId;
 use crate::cache::CacheStrategy;
 use crate::error::{InvalidRequestSnafu, RecordBatchSnafu, Result};
 use crate::read::projection::{read_column_ids_from_projection, repeated_vector_with_cache};
+use crate::read::read_columns::ReadColumns;
 use crate::sst::parquet::flat_format::sst_column_id_indices;
 use crate::sst::parquet::format::FormatProjection;
 use crate::sst::{
@@ -48,11 +49,11 @@ pub struct FlatProjectionMapper {
     metadata: RegionMetadataRef,
     /// Schema for converted [RecordBatch] to return.
     output_schema: SchemaRef,
-    /// Ids of columns to read from memtables and SSTs.
+    /// The columns to read from memtables and SSTs.
     /// The mapper won't deduplicate the column ids.
     ///
     /// Note that this doesn't contain the `__table_id` and `__tsid`.
-    read_column_ids: Vec<ColumnId>,
+    read_cols: ReadColumns,
     /// Ids and DataTypes of columns of the expected batch.
     /// We can use this to check if the batch is compatible with the expected schema.
     ///
@@ -73,9 +74,9 @@ impl FlatProjectionMapper {
     /// empty `RecordBatch` and only use its row count in this query.
     pub fn new(
         metadata: &RegionMetadataRef,
-        projection: impl Iterator<Item = usize>,
+        projection: impl IntoIterator<Item = usize>,
     ) -> Result<Self> {
-        let projection: Vec<_> = projection.collect();
+        let projection: Vec<_> = projection.into_iter().collect();
         let read_column_ids = read_column_ids_from_projection(metadata, &projection)?;
         Self::new_with_read_columns(metadata, projection, read_column_ids)
     }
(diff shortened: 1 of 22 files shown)


fengys1996:
@codex review

fengys1996:
@codex review

src/mito2/src/read/flat_projection.rs
```diff
@@ -84,39 +85,38 @@ impl FlatProjectionMapper {
     pub fn new_with_read_columns(
         metadata: &RegionMetadataRef,
         projection: Vec<usize>,
-        read_column_ids: Vec<ColumnId>,
+        read_cols: impl Into<ReadColumns>,
     ) -> Result<Self> {
+        let projection: Vec<_> = projection.into_iter().collect();
```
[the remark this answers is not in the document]
fengys1996: Good catch.

fengys1996:
@codex review

evenyag:
@codex review

src/mito2/src/read/read_columns.rs
```diff
@@ -106,12 +103,21 @@ impl ReadColumns {
...
 }
 
+impl<I> From<I> for ReadColumns
+where
+    I: IntoIterator<Item = ColumnId>,
+{
+    fn from(col_ids: I) -> Self {
+        ReadColumns::from_deduped_column_ids(col_ids)
+    }
+}
```
evenyag: I'd prefer not to provide the From implementation and let the user call from_deduped_column_ids() directly. It reminds the caller to check the column ids are unique.

src/mito2/src/sst/parquet/reader/stream.rs
```diff
@@ -146,6 +151,83 @@ fn fill_missing_cols(
...
+{
+    type Item = Result<RecordBatch>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+
+        match Pin::new(&mut this.inner).poll_next(cx) {
+            Poll::Ready(Some(Ok(rb))) => {
+                Poll::Ready(Some(align_batch_to_schema(rb, &this.output_schema)))
+            }
```
evenyag: We should avoid invoking align_batch_to_schema or the NestedSchemaAligner if we're reading non-nested schema or the schema matches the expected one.

src/mito2/src/sst/parquet/reader/stream.rs
```diff
@@ -45,7 +49,8 @@ pub struct MissingColFiller<S> {
     all_matched: bool,
 }
 
-pub(crate) type ProjectedRecordBatchStream = MissingColFiller<ParquetErrorAdapter>;
+pub(crate) type ProjectedRecordBatchStream =
+    NestedSchemaAligner<MissingColFiller<ParquetErrorAdapter>>;
```
evenyag: Should we use a boxed stream for the record batch stream instead? We can avoid creating the nested adapters in the normal path.

What's more, NestedSchemaAligner and MissingColFiller both only handle a part of the schema alignment for nested schema. I think we should implement them as one adapter.

src/mito2/src/sst/parquet/format.rs
```diff
@@ -596,50 +598,73 @@ impl FormatProjection {
...
-                    .get(&column_id)
+                    .get(&col.column_id)
                     .copied()
-                    .map(|index| (column_id, index))
+                    .map(|index_of_sst| (col.column_id, index_of_sst, col.nested_paths))
             })
             .collect();
-        // Sorts columns by their indices in the SST. SST uses a bitmap for projection.
-        // This ensures the schema of `projected_schema` is the same as the batch returned from the SST.
-        projected_schema.sort_unstable_by_key(|x| x.1);
```
evenyag, line 614: I think we should not remove the comments. We can update them.

src/mito2/src/sst/parquet/format.rs
```diff
@@ -596,50 +598,73 @@ impl FormatProjection {
...
+        sst_column_num: usize,
+    ) {
+        for index in sst_column_num - FIXED_POS_COLUMN_NUM..sst_column_num {
+            let needs_append = parquet_read_cols
+                .last()
+                .map(|col| col.root_index() != index)
+                .unwrap_or(true);
+            if needs_append {
+                parquet_read_cols.push(ParquetReadColumn::new(index));
+            }
```
evenyag: When will the needs_append be false?

MichaelScofield: Could be: the "`last()`" is the time index column?

fengys1996: Example: with `sst_column_num = 10` and `FIXED_POS_COLUMN_NUM = 4`, the fixed range is `6..9`, and `time index = 6` (i.e., sst_column_num - FIXED_POS_COLUMN_NUM).

If projection already includes time index 6, then at index = 6 we get `needs_append = false`, so we skip appending to avoid duplication.

fengys1996:
@evenyag @MichaelScofield ptal

MichaelScofield (approved):

src/mito2/src/sst/parquet/reader/stream.rs
```diff
@@ -99,18 +111,22 @@ where
...
                         this.expected_input_col_num,
                     )?
                 };
-                Poll::Ready(Some(Ok(rb)))
+
+                if rb.schema() == this.output_schema {
+                    Poll::Ready(Some(Ok(rb)))
+                } else {
+                    Poll::Ready(Some(align_batch_to_schema(rb, &this.output_schema)))
+                }
```
evenyag: Can we cache the schema check result in the stream so that the next time we don't need to check it again?

src/mito2/src/sst/parquet/format.rs
evenyag: Document this in the comment? And Is it possible that ts is not the last column? If we only handle the time index, we should check it explicitly outside of the loop.

src/mito2/src/sst/parquet/reader/stream.rs
fengys1996: Good suggestion!

fengys1996: Fixed in https://github.com/GreptimeTeam/greptimedb/pull/7959/changes/dc36e7041c9d17781cd380f981d64fa9ce637197

fengys1996:
https://github.com/GreptimeTeam/greptimedb/pull/7959#discussion_r3199964166

fixed in https://github.com/GreptimeTeam/greptimedb/pull/7959/changes/a6cbf7e947c8915398f344c48add5edf928ca79c

evenyag (approved):

PR #7963, opened 2026-04-13
fix: mysql prepare correctly returns error instead of panic

### LINKED ISSUE #7956: MySQL: server drops connection in two distinct error scenarios instead of sending ERR packet
### What type of bug is this?

Unexpected error

Frontend

## Summary

Two separate code paths cause GreptimeDB's MySQL frontend to close the TCP connection instead of sending a MySQL `ERR` packet. The MySQL protocol requires the server to send an `ERR` packet and keep the connection alive. Dropping the connection means:

- The client receives `ERR_MYSQL_CONNECTION_CLOSED` with no SQL error detail
- The connection is lost from the pool (Bun recovers by opening a new one, but the latency spike and missing error message are both problems)

---

### What happens

When a `COM_STMT_PREPARE` request contains invalid SQL, GreptimeDB closes the TCP connection. The same query sent via `COM_QUERY` (text protocol) correctly returns an `ERR` packet.

### Reproducer

```ts
import { SQL } from 'bun';

const sql = new SQL('mysql://greptime@localhost:4002/public', { max: 1, ssl: false });

// ── COM_QUERY (text protocol via sql.unsafe) — works correctly ──────────
try {
  await sql.unsafe(`SELECT \`timestamp\` FROM t WHERE \`timestamp\` > NOW() - INTERVAL '1 hour'`);
} catch (e: any) {
  console.log('unsafe:', e.code);
  // ERR_MYSQL_SERVER_ERROR  ← correct: ERR packet received, connection stays alive
}

// ── COM_STMT_PREPARE (template literal, same SQL) — drops connection ────
try {
  await sql`SELECT \`timestamp\` FROM t WHERE \`timestamp\` > NOW() - INTERVAL '1 hour'`;
} catch (e: any) {
  console.log('template:', e.code);
  // ERR_MYSQL_CONNECTION_CLOSED  ← bug: connection dropped instead of ERR packet
}
```

Start GreptimeDB:

```bash
docker run --rm -p 4002:4002 greptime/greptimedb:v1.0.0 standalone start --mysql-addr=0.0.0.0:4002
```

### Verified triggers (COM_STMT_PREPARE path)

All of the following drop the connection when sent via template literal (COM_STMT_PREPARE) but return a proper ERR packet via `sql.unsafe()` (COM_QUERY):

| Query | Parse error |
|-------|-------------|
| `SELECT … WHERE … > NOW() - INTERVAL '1 hour'` | PG-style interval literal |
| `SELECT … WHERE … + INTERVAL '30 minutes' AS ts` | PG-style interval literal |
| `` `SELCT 1` `` | Unknown keyword |

### Expected behaviour

`COM_STMT_PREPARE` with invalid SQL should respond with a `COM_STMT_PREPARE` error response (first byte `0xFF`, error code + message). Connection must remain open.

### Actual behaviour

TCP connection is closed. Client receives `ERR_MYSQL_CONNECTION_CLOSED` with no error code or message.

---

### What happens

When `sql.unsafe()` sends a COM_QUERY containing two or more valid SQL statements separated by `;`, GreptimeDB closes the TCP connection without sending any response. This was confirmed by a raw MySQL protocol inspector using Bun's exact capability flags (`0x011b8208`): the connection drops immediately upon receiving the multi-statement query, with zero bytes returned.

This is the same failure mode as Bug A — a silent TCP drop instead of an `ERR` packet — but triggered by a different code path.

### Root cause — capability flag mismatch

The `CLIENT_MULTI_STATEMENTS` bit (`0x00010000`) in the MySQL handshake controls whether the server should execute multi-statement queries. The protocol requires that a client only set a capability bit if the server also advertised that bit in its initial greeting.

Inspecting the actual handshake bytes reveals a two-sided mismatch:

| Side | `CLIENT_MULTI_STATEMENTS` | `CLIENT_MULTI_RESULTS` |
|------|--------------------------|------------------------|
| **GreptimeDB server greeting** | **0 — not advertised** | 0 — not advertised |
| **Bun.SQL client response** | **1 — set anyway** | 1 — set anyway |

GreptimeDB never advertises multi-statement support, yet Bun.SQL sets the bit unconditionally. This causes GreptimeDB to receive multi-statement queries it was never intended to handle, exposing a code path that drops the TCP connection instead of returning an error.

The correct behaviour at each side:

- **GreptimeDB (server)**: a `COM_QUERY` containing more than one statement should either be rejected with a single `ERR` packet (if multi-statement support is not implemented) or executed correctly with `SERVER_MORE_RESULTS_EXISTS` set in intermediate EOF/OK packets. The connection must remain open in either case.
- **Bun.SQL (client)**: should not set `CLIENT_MULTI_STATEMENTS` in the client capabilities unless the server advertised it first (separate issue: [TICKET_BUN_MULTI_STATEMENTS.md](TICKET_BUN_MULTI_STATEMENTS.md)).

### Reproducer

```ts
import { SQL } from 'bun';

const sql = new SQL('mysql://greptime@localhost:4002/public', { max: 1, ssl: false });

// Two valid statements — connection drops
try {
  await sql.unsafe('SELECT 1 AS a; SELECT 2 AS b');
} catch (e: any) {
  console.log(e.code); // ERR_MYSQL_CONNECTION_CLOSED
}

// Valid + invalid — connection stays alive (ERR packet returned for the second statement)
try {
  await sql.unsafe('SELECT 1 AS a; SELCT 2 AS b');
} catch (e: any) {
  console.log(e.code); // ERR_MYSQL_SERVER_ERROR  ← correct
}
```

### Distinguishing behaviour

| Statements | Both valid | First valid, second invalid |
|------------|------------|-----------------------------|
| Result     | Connection dropped | Proper ERR packet, connection alive |

A trailing semicolon with no second statement (`SELECT 1;`) and a doubled semicolon (`SELECT 1;;`) do not trigger the bug. The valid+invalid case (`SELECT 1; SELCT 2`) returns a proper `ERR` packet — the parse error is caught before the execution path that drops the connection is reached.

### Expected behaviour

There are two conformant fix paths:

**Path 1 — Implement multi-statement support**

Execute each statement in sequence and send a result set for each. Set the `SERVER_MORE_RESULTS_EXISTS` status flag (`0x0008`) in the EOF/OK packet that follows each result set *except the last*, so the client knows to keep draining. Also advertise `CLIENT_MULTI_STATEMENTS | CLIENT_MULTI_RESULTS` in the server greeting.

Note: advertising the capability without also setting `SERVER_MORE_RESULTS_EXISTS` in intermediate packets is not sufficient — clients gate their drain loop on that flag.

**Path 2 — Reject multi-statements when the capability was not negotiated (minimal fix)**

Keep the server greeting as-is. When processing a COM_QUERY, check whether `CLIENT_MULTI_STATEMENTS` appears in the negotiated capability intersection (server-advertised ∩ client-requested). If it does not, and the query string contains more than one statement, return a single `ERR` packet (error code 1149 / SQLSTATE `42000`) and keep the connection alive.

Path 2 is the smaller change and eliminates the connection drop immediately. Path 1 requires correctly implementing the full multi-result protocol state machine but enables the feature for clients that want it.

### Actual behaviour

GreptimeDB closes the TCP connection immediately upon receiving the multi-statement COM_QUERY, without sending any bytes. The client receives `ERR_MYSQL_CONNECTION_CLOSED` with no error code or message — identical to Bug A's failure mode.

---

## PostgreSQL path is not affected

The same error scenarios behave correctly on GreptimeDB's PostgreSQL frontend:

| Scenario | MySQL result | PostgreSQL result |
|----------|-------------|-------------------|
| Parse error via prepared statement | Connection dropped | Proper error, connection alive |
| Parse error via text query | Proper error | Proper error |
| Multiple valid statements (`; `) | Connection dropped | Executes all, connection alive |

Both bugs are MySQL-frontend-specific.

## Severity

**Scenario A**: Medium. Affects any invalid SQL sent via the standard prepared-statement path (template literals). Prevents the client from receiving the actual SQL error message, making debugging significantly harder.

**Scenario B**: Low. Only triggered by deliberately sending multiple statements in one call, which standard MySQL clients avoid. The pool recovers automatically. However, if triggered, no error detail is available.

MacOS

1.0.0

MichaelScofield:

fix https://github.com/GreptimeTeam/greptimedb/issues/7956

wait https://github.com/databendlabs/opensrv/pull/81

also in this pr: mysql writer supports multiple statement execution

## PR Checklist

### DIFF
--- a/src/servers/src/mysql/writer.rs
+++ b/src/servers/src/mysql/writer.rs
@@ -47,7 +47,7 @@ use crate::metrics::*;
 
 /// Try to write multiple output to the writer if possible.
 pub async fn write_output<W: AsyncWrite + Send + Sync + Unpin>(
-    w: QueryResultWriter<'_, W>,
+    mut writer: QueryResultWriter<'_, W>,
     query_context: QueryContextRef,
     session: SessionRef,
     outputs: Vec<Result<Output>>,
@@ -56,21 +56,87 @@ pub async fn write_output<W: AsyncWrite + Send + Sync + Unpin>(
         session.add_warning(warning);
     }
 
-    let mut writer = Some(MysqlResultWriter::new(
-        w,
-        query_context.clone(),
-        session.clone(),
-    ));
-    for output in outputs {
-        let result_writer = writer.take().context(error::InternalSnafu {
-            err_msg: "Sending multiple result set is unsupported",
-        })?;
-        writer = result_writer.try_write_one(output).await?;
+    enum Response {
+        ResultSet {
+            columns: Vec<Column>,
+            stream: SendableRecordBatchStream,
+        },
+        AffectedRows(usize),
     }
 
-    if let Some(result_writer) = writer {
-        result_writer.finish().await?;
+    let mut responses = Vec::with_capacity(outputs.len());
+    for output in outputs {
+        match output {
+            Ok(x) => {
+                let output = match x.data {
+                    OutputData::Stream(stream) => either::Left(stream),
+                    OutputData::RecordBatches(record_batches) => {
+                        either::Left(record_batches.as_stream())
+                    }
+                    OutputData::AffectedRows(rows) => either::Right(rows),
+                };
+                responses.push(match output {
+                    either::Left(stream) => {
+                        let schema = stream.schema();
+                        let columns = match create_mysql_column_def(&schema) {
+                            Ok(columns) => columns,
+                            Err(e) => {
+                                MysqlResultWriter::write_query_error(
+                                    e,
+                                    writer,
+                                    query_context.clone(),
+                                )
+                                .await?;
+                                return Ok(());
+                            }
+                        };
+                        Response::ResultSet { columns, stream }
+                    }
+                    either::Right(rows) => Response::AffectedRows(rows),
+                });
+            }
+            Err(e) => {
+                MysqlResultWriter::write_query_error(e, writer, query_context.clone()).await?;
+                return Ok(());
+            }
+        }
     }
+
+    for response in &mut responses {
+        writer = match response {
+            Response::ResultSet { columns, stream } => {
+                let mut row_writer = writer.start(columns).await?;
+                while let Some(record_batch) = stream.next().await {
+                    match record_batch {
+                        Ok(record_batch) => {
+                            if let Err(e) = MysqlResultWriter::write_recordbatch(
+                                &mut row_writer,
+                                record_batch,
+                                query_context.clone(),
+                            )
+                            .await
+                            {
+                                let (kind, err) = handle_err(e, query_context);
+                                row_writer.finish_error(kind, &err.as_bytes()).await?;
+                                return Ok(());
+                            }
+                        }
+                        Err(e) => {
+                            let (kind, err) = handle_err(e, query_context);
+                            row_writer.finish_error(kind, &err.as_bytes()).await?;
+                            return Ok(());
+                        }
+                    }
+                }
+                row_writer.finish_one().await?
+            }
+            Response::AffectedRows(rows) => {
+                MysqlResultWriter::write_affected_rows(writer, *rows, &session).await?
+            }
+        }
+    }
+
+    writer.no_more_results().await?;
     Ok(())
 }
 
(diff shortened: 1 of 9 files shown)


src/servers/src/mysql/writer.rs
```diff
@@ -56,21 +56,87 @@ pub async fn write_output<W: AsyncWrite + Send + Sync + Unpin>(
...
+                            if let Err(e) = MysqlResultWriter::write_recordbatch(
+                                &mut row_writer,
+                                record_batch,
+                                query_context.clone(),
+                            )
+                            .await
+                            {
+                                let (kind, err) = handle_err(e, query_context);
+                                row_writer.finish_error(kind, &err.as_bytes()).await?;
+                                return Ok(());
```
sunng87, line 121: can we use and_then to reduce duplicate error processing code

MichaelScofield: seems can't because the `write_recordbatch` is async, not chainable in `Result`'s `and_then`

sunng87 (approved):

fengjiachun:
@codex review

fengjiachun (approved):
LGTM

PR #7966, opened 2026-04-14
refactor: update SqlPlan for more cleaner variants

sunng87:

Our `SqlPlan` has been unclear with its members:

- query string
- `Option<LogicalPlan>`
- `Option<Statement>`
- `Option<Schema>`

in this patch, I refactored it into several fixed variants:

- Empty query
- Short-cut query: not processed by query engine
- LogicalPlan: processed by datafusion query engine 
- Statement: processed by our own handler, eg CREATE TABLE/SHOW TABLES

In this case, we won't store 3 forms (string, parsed statement, plan) of every query. 

## PR Checklist


src/frontend/src/instance.rs
```diff
@@ -638,24 +629,21 @@ impl Instance {
 
     async fn do_exec_plan_inner(
         &self,
-        stmt: Option<Statement>,
         plan: LogicalPlan,
+        query: String,
```
MichaelScofield, line 633: Seems this `query` can be derived from `plan` inside this function and only when needed? So that you don't have to do `let query = logical_plan.display_indent().to_string();` outside every time, which is costly.

src/servers/src/mysql/handler.rs
```diff
@@ -230,49 +206,43 @@ impl MysqlInstanceShim {
...
+                .transpose()?
+                .unwrap_or_default();
+
+        match plan {
+            Some(plan) if params.len() == param_num - 1 => {
+                self.save_plan(SqlPlan::Plan(plan, query.clone()), stmt_key)
+                    .map_err(|e| {
+                        error!(e; "Failed to save prepared statement");
+                        e
+                    })?;
```
MichaelScofield:
```suggestion
                    .inspect_err(|e| {
                        error!(e; "Failed to save prepared statement");
                    })?;
```

src/servers/src/mysql/handler.rs
```diff
@@ -230,49 +206,43 @@ impl MysqlInstanceShim {
...
+            }
+            _ => {
+                self.save_plan(SqlPlan::Statement(statement, query), stmt_key)
+                    .map_err(|e| {
+                        error!(e; "Failed to save prepared statement");
+                        e
+                    })?;
```
MichaelScofield: ditto

src/servers/src/lib.rs
```diff
@@ -54,14 +53,19 @@ mod row_writer;
...
-    statement: Option<Statement>,
-    plan: Option<LogicalPlan>,
-    schema: Option<Schema>,
+pub enum SqlPlan {
+    /// Empty Query
+    Empty,
+    /// Hardcoded SQL shortcuts
+    Shortcut(String),
+    /// Datafusion parsed execution plan with the original query string
+    Plan(LogicalPlan, String),
```
MichaelScofield, line 64: The original query string can be calculated from `LogicalPlan`, is it needed to store it here as well?

src/frontend/src/instance.rs
sunng87: In the generated plan we lost the original query string, the string from `logical_plan.display_indent().to_string()` is actually the explain view of plan, instead of the original one.

The only case we need to use `logical_plan.display_indent().to_string()` is for the grpc query interface when no query string is provided. Actually I think that interface is not in use.

src/servers/src/mysql/handler.rs
```diff
@@ -315,18 +283,17 @@ impl MysqlInstanceShim {
...
-                    self.do_exec_plan(
-                        &sql_plan.query,
-                        sql_plan.statement.clone(),
-                        plan,
-                        query_ctx.clone(),
-                    )
-                    .await,
+                    self.query_handler
+                        .do_exec_plan(replaced_plan, query, query_ctx.clone())
+                        .await,
```
shuiyisong, line 303: This now seems bypass the `federated::check`, is this intentional?

shuiyisong (commented):
Rest LGTM

MichaelScofield (approved):

src/frontend/src/instance.rs
MichaelScofield: I remember it's saved for 3rd party query engines like Apache Trino/Calcite/Ballista that can talk in Substrait. Now that is not among our primary concerns, I think it can be removed.

sunng87: It's still used for our slow query logger. I feel it's still necessary but would be nice if there is a better container to carry that.

shuiyisong (approved):

PR #7971, opened 2026-04-15
fix(meta): renew operating region leases from keeper roles

WenyXu:

This PR fixes a metasrv lease-renew bug for operating regions.
Previously, `MemoryRegionKeeper` only recorded `(datanode_id, region_id)`, and operating-region lease renewal reused the region role reported by datanode heartbeats. This could break leader-only operations such as drop table, drop database, repartition, and region migration: if a leader region was temporarily reported as follower during an ongoing operation, metasrv could keep renewing the lease as follower, and the operation would get stuck in retries.
To fix this, this PR makes metasrv use an in-memory authoritative role for operating regions:
- refactor `MemoryRegionKeeper` from a `HashSet<(DatanodeId, RegionId)>` to a `HashMap<(DatanodeId, RegionId), RegionRole>`
- make operating-region registration explicitly write the intended role at all related call sites
- renew operating-region leases from the role stored in `MemoryRegionKeeper`, instead of the transient role reported by heartbeat

## PR Checklist

### DIFF
--- a/src/common/meta/src/region_keeper.rs
+++ b/src/common/meta/src/region_keeper.rs
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, RwLock};
 
+use store_api::region_engine::RegionRole;
 use store_api::storage::RegionId;
 
 use crate::DatanodeId;
@@ -24,7 +25,7 @@ use crate::DatanodeId;
 pub struct OperatingRegionGuard {
     datanode_id: DatanodeId,
     region_id: RegionId,
-    inner: Arc<RwLock<HashSet<(DatanodeId, RegionId)>>>,
+    inner: Arc<RwLock<HashMap<(DatanodeId, RegionId), RegionRole>>>,
 }
 
 impl Drop for OperatingRegionGuard {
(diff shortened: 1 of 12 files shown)


evenyag:
@codex review

src/meta-srv/src/region/lease_keeper.rs
```diff
@@ -154,13 +148,21 @@ impl RegionLeaseKeeper {
...
+        // so looking up by `region_id` is sufficient here.
+        if let Some(role) = operating_regions.get(&region_id) {
+            let region_lease_info = RegionLeaseInfo::operating(region_id, *role);
+            if *role != reported_role {
+                info!(
+                    "The region {} on datanode {} is operating with role {:?}, but reported as {:?}",
+                    region_id, datanode_id, role, reported_role
+                );
+            }
             return Some(region_lease_info);
```
fengjiachun: This new `operating_regions` fast path now always trusts the in-memory keeper role when a region is marked as operating. That works for create/drop/repartition, but region migration registers the candidate in `open_candidate_region` as Follower and only clears `opening_region_guards` after upgrade_candidate_region has already committed the route change and awaited deregister_failure_detectors(). A heartbeat arriving in that window will still hit this branch and renew the newly promoted region as Follower, even though metadata already says it is the leader. If leader promotion still depends on the lease grant, the migration can remain stuck in exactly the kind of retry loop this PR is trying to eliminate.

WenyXu: Good catch, [1d487c6](https://github.com/GreptimeTeam/greptimedb/pull/7971/commits/1d487c69a0c85bf653e17a7d5119c2c1cb88e9f3)

src/common/meta/src/region_keeper.rs
```diff
@@ -59,40 +61,48 @@ impl MemoryRegionKeeper {
...
-        if inner.insert((datanode_id, region_id)) {
-            Some(OperatingRegionGuard {
-                datanode_id,
-                region_id,
-                inner: self.inner.clone(),
-            })
-        } else {
-            None
+        match inner.entry((datanode_id, region_id)) {
+            Entry::Occupied(_) => None,
```
evenyag, line 73: Why do we return None?

WenyXu: Only return `OperatingRegionGuard` if the key does not exist.

fengjiachun (approved):
LGTM

evenyag (approved):

PR #7972, opened 2026-04-15
feat: support prefiltering any columns in flat format

evenyag:

Previously, parquet prefiltering only supported primary key (PK) columns — tag predicates were evaluated by decoding the encoded primary key column to compute a refined row selection before reading the remaining columns.

This PR generalizes prefiltering to support **all column types** (tags, fields, and timestamps), not just primary keys. Key changes:

**New filter planning layer** (`prefilter.rs`):
- Introduces `build_reader_filter_plan()` and `build_bulk_filter_plan()` that categorize predicates into three groups:
  - **PK filters**: tag predicates evaluated via encoded primary key decoding (existing path)
  - **Simple prefilters**: tag/field/timestamp predicates on columns directly readable from parquet
  - **Physical prefilters**: more complex expressions (`IN`, `IS NULL`, `IS NOT NULL`, `BETWEEN`) compiled to physical exprs for single-column evaluation
- A `PreFilterMode` (`All` vs `SkipFields`) controls whether field column predicates participate in prefiltering
- Prefiltering is enabled only when the estimated selectivity meets a configurable threshold

**Unified prefilter execution** (`prefilter.rs`):
- `PrefilterContext` now holds a unified projection covering PK + simple + physical filter columns
- `execute_prefilter()` reads the projected columns once, then applies all three filter types to produce a combined `BooleanBuffer` row mask
- Filters already applied during prefiltering are removed from the post-read filter list to avoid redundant evaluation

**Refactored reader/filter path** (`reader.rs`, `file_range.rs`):
- `SimpleFilterContext` no longer tracks `usable_primary_key_filter`; filter routing is handled by the plan builder
- New `PhysicalFilterContext` wraps a `PhysicalExpr` for single-column prefilter evaluation
- `RowGroupBuildContext` is simplified — prefilter-related fields (`filters`, `skip_fields`) are removed
- `precise_filter_flat()` no longer needs to skip prefiltered PK filters since they're excluded from the remaining filter list

**Other changes**:
- Extracts `Predicate::to_physical_expr()` as a public method for building a single physical expr
- Adds tests covering generalized prefilter behavior and verifying physical filters are not re-evaluated post-filter

**Performance**
4.5x faster in TSBS cpu-max-all-8 query.

<details>
<summary>Before</summary>

```
Run complete after 100 queries with 1 workers (Overall query rate 5.51 queries/sec):
Influx max of all CPU metrics, random    8 hosts, random 8h0m0s by 1h:
min:   128.09ms, med:   182.73ms, mean:   181.21ms, max:  239.89ms, stddev:    29.43ms, sum:  18.1sec, count: 100
all queries                                                          :
min:   128.09ms, med:   182.73ms, mean:   181.21ms, max:  239.89ms, stddev:    29.43ms, sum:  18.1sec, count: 100
wall clock time: 18.143011sec
```

</details>

<details>
<summary>After</summary>

```
Run complete after 100 queries with 1 workers (Overall query rate 25.21 queries/sec):
Influx max of all CPU metrics, random    8 hosts, random 8h0m0s by 1h:
min:    26.60ms, med:    39.41ms, mean:    39.60ms, max:   55.67ms, stddev:     5.73ms, sum:   4.0sec, count: 100
all queries                                                          :
min:    26.60ms, med:    39.41ms, mean:    39.60ms, max:   55.67ms, stddev:     5.73ms, sum:   4.0sec, count: 100
wall clock time: 3.973266sec
```

</details>

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet/prefilter.rs
+++ b/src/mito2/src/sst/parquet/prefilter.rs
@@ -18,13 +18,13 @@
 //! (the prefilter phase), applying filters to compute a refined row selection,
 //! then reading the remaining columns with the refined selection.
 
-use std::collections::{HashMap, HashSet};
+use std::collections::HashSet;
 use std::ops::{BitAnd, Range};
 use std::sync::Arc;
 
 use api::v1::SemanticType;
 use common_recordbatch::filter::SimpleFilterEvaluator;
-use datatypes::arrow::array::{ArrayRef, BinaryArray, BooleanArray, BooleanBufferBuilder};
+use datatypes::arrow::array::{BinaryArray, BooleanArray, BooleanBufferBuilder};
 use datatypes::arrow::buffer::BooleanBuffer;
 use datatypes::arrow::record_batch::RecordBatch;
 use futures::StreamExt;
@@ -33,24 +33,19 @@ use parquet::arrow::ProjectionMask;
 use parquet::arrow::arrow_reader::RowSelection;
 use parquet::schema::types::SchemaDescriptor;
 use snafu::{OptionExt, ResultExt};
-use store_api::codec::PrimaryKeyEncoding;
 use store_api::metadata::{RegionMetadata, RegionMetadataRef};
-use store_api::storage::ColumnId;
 use store_api::storage::consts::PRIMARY_KEY_COLUMN_NAME;
 
 use crate::error::{
     ComputeArrowSnafu, DecodeSnafu, EvalPartitionFilterSnafu, NewRecordBatchSnafu,
     ReadParquetSnafu, RecordBatchSnafu, Result, UnexpectedSnafu,
 };
-use crate::sst::parquet::flat_format::{
-    DecodedPrimaryKeys, FlatReadFormat, decode_primary_keys, primary_key_column_index,
-};
+use crate::sst::parquet::flat_format::{FlatReadFormat, primary_key_column_index};
 use crate::sst::parquet::format::PrimaryKeyArray;
 use crate::sst::parquet::reader::{
     MaybeFilter, MaybePhysicalFilter, PhysicalFilterContext, RowGroupBuildContext,
     RowGroupReaderBuilder, SimpleFilterContext,
 };
-use crate::sst::parquet::row_selection::row_selection_from_row_ranges_exact;
 
 pub(crate) fn matching_row_ranges_by_primary_key(
     input: &RecordBatch,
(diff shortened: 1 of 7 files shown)


src/mito2/src/sst/parquet/prefilter.rs
```diff
@@ -216,18 +184,196 @@ impl PrimaryKeyFilter for CachedPrimaryKeyFilter {
...
+    if prefilter_builder.is_some() {
+        ReaderFilterPlan {
+            remaining_simple_filters,
+            prefilter_builder,
+        }
+    } else {
+        // If prefilter setup is not worthwhile, keep the original simple
+        // filters on the normal path so behavior is unchanged.
+        remaining_simple_filters.extend(prefilter_simple_filters);
+        remaining_simple_filters.extend(pk_filter_contexts);
```
discord9, line 395: could add explain that physical filters are best effort based and drop them on None is ok

src/mito2/src/sst/parquet/reader.rs
discord9, line 1715: maybe fast path of prefilter_result.refined_selection.row_count() == 0 then return empty stream?

src/mito2/src/sst/parquet/prefilter.rs
```diff
@@ -216,18 +184,196 @@ impl PrimaryKeyFilter for CachedPrimaryKeyFilter {
...
+
+pub(crate) fn build_reader_filter_plan(
+    predicate: Option<&Predicate>,
+    expected_metadata: Option<&RegionMetadata>,
+    pre_filter_mode: PreFilterMode,
+    read_format: &FlatReadFormat,
+    parquet_schema: &SchemaDescriptor,
+    codec: &Arc<dyn PrimaryKeyCodec>,
+) -> ReaderFilterPlan {
+    let Some(predicate) = predicate else {
```
killme2008, line 291: Adds metrics for building plan, duration, times, filters number etc.

src/mito2/src/sst/parquet/prefilter.rs
killme2008: At this layer, the goal is to filter as much as possible, right? Ultimately, we still fall back to DataFusion’s FilterExec, correct?

src/mito2/src/sst/parquet/prefilter.rs
evenyag: We already have metrics for prefilter, so I'm not going to add more metrics here.

src/mito2/src/sst/parquet/prefilter.rs
evenyag: Tags and time index filters that SimpleFilterEvaluator supports are filtered precisely in the engine.

src/mito2/src/sst/parquet/reader.rs
evenyag: This requires modifying the callers a lot. Parquet already has a fast path, so we don't need to do this.
https://github.com/apache/arrow-rs/blob/4fa8d2ff5f18f2d773f9642631715509f844a062/parquet/src/arrow/push_decoder/reader_builder/mod.rs#L536-L545

evenyag:
@codex review

killme2008 (approved):

v0y4g3r (approved):

PR #7978, opened 2026-04-15
fix: allow ipv4_num_to_string to accept Int32/Int64/UInt64 inputs

### LINKED ISSUE #7977: Make ipv4_num_to_string function accepts int32
```sql
mysql> select ipv4_num_to_string(3232235521::uint32);
+------------------------------------------------------------------+
| ipv4_num_to_string(arrow_cast(Int64(3232235521),Utf8("UInt32"))) |
+------------------------------------------------------------------+
| 192.168.0.1                                                      |
+------------------------------------------------------------------+
1 row in set (0.01 sec)

mysql> select ipv4_num_to_string(3232235521);
ERROR 1815 (HY000): (PlanQuery): Failed to plan SQL: Error during planning: Failed to coerce arguments for 'ipv4_num_to_string': cannot coerce Int64 to Exact([UInt32]). No matching function for 'ipv4_num_to_string(Int64)'. You may need an explicit cast.
    Candidate functions:
    ipv4_num_to_string(UInt32)
```

We could update the `ipv4_num_to_string` signature to accept `Int32` and automatically coerce the input type.

https://github.com/GreptimeTeam/greptimedb/blob/8a3bd374a89a74fc2dff9ef24d55a71b30c4e1f0/src/common/function/src/scalars/ip/ipv4.rs#L47

Ai-chan-0411:
## Summary

Fixes #7977

I ran into this when trying to use `ipv4_num_to_string(3232235521)` directly — SQL integer literals default to `Int64`, but the function only accepted `UInt32`, so it would fail with a coercion error.

The fix expands the accepted types from just `UInt32` to also include `Int32`, `Int64`, and `UInt64`, and adds a `compute::cast` to `UInt32` at the start of `invoke_with_args` (the same pattern used by other functions in this module like `ipv4_string_to_num`). Tests for `Int32` and `Int64` inputs are included.

## Changes

- `TypeSignature::Exact([UInt32])` → `TypeSignature::OneOf([UInt32, Int32, Int64, UInt64])`
- Add `compute::cast(&arg0, &DataType::UInt32)?` before processing
- Add unit tests `test_ipv4_num_to_string_int32` and `test_ipv4_num_to_string_int64`

### DIFF
--- a/src/common/function/src/scalars/ip/ipv4.rs
+++ b/src/common/function/src/scalars/ip/ipv4.rs
@@ -44,7 +44,12 @@ impl Default for Ipv4NumToString {
     fn default() -> Self {
         Self {
             signature: Signature::new(
-                TypeSignature::Exact(vec![DataType::UInt32]),
+                TypeSignature::OneOf(vec![
+                    TypeSignature::Exact(vec![DataType::UInt32]),
+                    TypeSignature::Exact(vec![DataType::Int32]),
+                    TypeSignature::Exact(vec![DataType::Int64]),
+                    TypeSignature::Exact(vec![DataType::UInt64]),
+                ]),
                 Volatility::Immutable,
             ),
             aliases: ["inet_ntoa".to_string()],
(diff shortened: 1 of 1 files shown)


Ai-chan-0411:
Thank you for the detailed code review. I see the PR was closed before I could address your feedback about adding UInt64Array imports and tests for UInt64 inputs with out-of-range handling. I appreciate the specific suggestions — they align with proper test coverage for edge cases. I'll incorporate these improvements in a follow-up PR.

PR #7979, opened 2026-04-16
feat: remote dyn filter basics

discord9:

https://github.com/GreptimeTeam/greptime-proto/pull/314

## Summary
- define the Phase 1 remote dynamic filter wire ABI, including `remote_query_id`, `DynFilterUpdate`, and shared payload encode/decode helpers
- add the region unary RPC scaffolding for remote dynamic filter `update` / `unregister` handling between frontend and datanode
- keep this PR scoped to scaffolding only, leaving frontend producer and datanode apply/runtime work to follow-up changes

## Why
This splits out the protocol and control-plane groundwork into a smaller reviewable PR before the larger end-to-end remote dynamic filter implementation. It gives us a stable ABI and RPC entrypoint first, so later work can build on a reviewed contract instead of mixing transport and runtime changes together.

## Included in this PR
- wire ABI groundwork
- region RPC scaffolding

## Not included
- frontend producer registration/build-link logic
- datanode apply/runtime wrapper integration
- lifecycle cleanup / observability / validation follow-ups

## PR Checklist

### DIFF
--- a/src/common/query/src/request.rs
+++ b/src/common/query/src/request.rs
@@ -12,9 +12,161 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::sync::Arc;
+
 use api::v1::region::RegionRequestHeader;
+use datafusion::arrow::datatypes::Schema;
+use datafusion::execution::TaskContext;
+use datafusion::physical_expr::expressions::Column;
+use datafusion::physical_plan::PhysicalExpr;
+use datafusion::physical_plan::joins::HashTableLookupExpr;
+use datafusion_common::tree_node::{TreeNode, TreeNodeRecursion};
+use datafusion_common::{DataFusionError, Result as DataFusionResult};
 use datafusion_expr::LogicalPlan;
+use datafusion_proto::physical_plan::DefaultPhysicalExtensionCodec;
+use datafusion_proto::physical_plan::from_proto::parse_physical_expr;
+use datafusion_proto::physical_plan::to_proto::serialize_physical_expr;
+use datafusion_proto::protobuf::PhysicalExprNode;
+use prost::Message;
+use serde::{Deserialize, Serialize};
 use store_api::storage::RegionId;
+
+pub const DYN_FILTER_PROTOCOL_VERSION: u32 = 1;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[non_exhaustive]
+#[serde(tag = "kind", content = "payload", rename_all = "snake_case")]
+pub enum DynFilterPayload {
+    Datafusion(Vec<u8>),
+}
+
+impl DynFilterPayload {
+    pub fn from_datafusion_expr(
+        expr: &Arc<dyn PhysicalExpr>,
+        max_payload_bytes: usize,
+    ) -> DataFusionResult<Self> {
+        validate_supported_payload_expr(expr)?;
+
+        let codec = DefaultPhysicalExtensionCodec {};
+        let proto = serialize_physical_expr(expr, &codec)?;
+        let mut bytes = Vec::new();
+        proto.encode(&mut bytes).map_err(|e| {
+            DataFusionError::Internal(format!("Failed to encode PhysicalExprNode: {e}"))
+        })?;
+
+        validate_payload_size(bytes.len(), max_payload_bytes)?;
+
+        Ok(Self::Datafusion(bytes))
+    }
+
+    pub fn decode_datafusion_expr(
+        &self,
+        task_ctx: &TaskContext,
+        input_schema: &Schema,
+        max_payload_bytes: usize,
+    ) -> DataFusionResult<Arc<dyn PhysicalExpr>> {
+        let Self::Datafusion(bytes) = self;
+        validate_payload_size(bytes.len(), max_payload_bytes)?;
+        let codec = DefaultPhysicalExtensionCodec {};
+        let proto = PhysicalExprNode::decode(bytes.as_slice()).map_err(|e| {
+            DataFusionError::Internal(format!("Failed to decode PhysicalExprNode: {e}"))
+        })?;
+
+        let expr = parse_physical_expr(&proto, task_ctx, input_schema, &codec)?;
+        validate_supported_payload_expr(&expr)?;
+        validate_decoded_payload_expr(&expr, input_schema)?;
+        Ok(expr)
+    }
+}
+
+fn validate_payload_size(
+    payload_size_bytes: usize,
+    max_payload_bytes: usize,
+) -> DataFusionResult<()> {
+    if payload_size_bytes > max_payload_bytes {
+        return Err(DataFusionError::Plan(format!(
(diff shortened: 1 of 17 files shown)


src/servers/src/grpc/greptime_handler.rs
```diff
@@ -214,7 +217,11 @@ pub(crate) fn create_query_context(
         .current_catalog(catalog)
         .current_schema(schema)
         .timezone(timezone)
-        .channel(channel);
+        .channel(channel)
+        .set_extension(
+            REMOTE_QUERY_ID_EXTENSION_KEY.to_string(),
```
fengjiachun, line 222: `remote_query_id` is being introduced here as the internal correlation key for remote dynamic filters, but this builder still applies every client-supplied hint after generating it. Because `x-greptime-hints` accepts arbitrary keys, an external gRPC caller can set remote_query_id=... and force collisions with another query's control-plane traffic. The same overwrite path exists for HTTP via http::hints::extract_hints(). Once follow-up PRs start using this ID to register/update/unregister filters, this becomes a spoofable internal identifier.

src/common/query/src/request.rs
```diff
@@ -12,10 +12,162 @@
...
+use prost::Message;
+use serde::{Deserialize, Serialize};
 use store_api::storage::RegionId;
 
+pub const DYN_FILTER_PROTOCOL_VERSION: u32 = 1;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[non_exhaustive]
+#[serde(tag = "kind", content = "payload", rename_all = "snake_case")]
+pub enum DynFilterPayload {
```
killme2008, line 49: Always document the public structs and functions.

src/datanode/src/region_server.rs
```diff
@@ -697,6 +698,70 @@ impl RegionServer {
...
+        if request.query_id.is_empty() {
+            return error::MissingRequiredFieldSnafu { name: "query_id" }.fail();
+        }
+
+        match request
+            .action
+            .as_ref()
+            .context(error::MissingRequiredFieldSnafu { name: "action" })?
+        {
+            api::v1::region::remote_dyn_filter_request::Action::Update(update) => {
```
killme2008: `use api::v1::region::remote_dyn_filter_request::Action`

Make the code clean.

src/servers/src/grpc/greptime_handler.rs
killme2008: Agree

src/session/src/lib.rs
```diff
@@ -106,6 +110,10 @@ impl Session {
             .channel(self.conn_info.channel)
             .process_id(self.process_id)
             .conn_info(self.conn_info.clone())
+            .set_extension(
+                REMOTE_QUERY_ID_EXTENSION_KEY.to_string(),
+                generate_remote_query_id(),
+            )
```
evenyag: Normally, we should have a trace id in the RegionRequestHeader::tracing_context. Should we reuse that as the query id?

src/session/src/context.rs
```diff
@@ -483,6 +526,10 @@ impl QueryContext {
 impl QueryContextBuilder {
     pub fn build(self) -> QueryContext {
         let channel = self.channel.unwrap_or_default();
+        let mut extensions = self.extensions.unwrap_or_default();
+        extensions
+            .entry(REMOTE_QUERY_ID_EXTENSION_KEY.to_string())
+            .or_insert_with(generate_remote_query_id);
```
evenyag, line 532: We set extension when we build the QueryContextBuilder and then we check it again in QueryContextBuilder::build(). So is it necessary to set_extension every time we create the QueryContextBuilder?

Or should we provide a constructor for QueryContextBuilder that always creates a remote query id?

src/datanode/src/region_server.rs
```diff
@@ -697,6 +698,70 @@ impl RegionServer {
...
+                self.handle_remote_dyn_filter_update(&request.query_id, update)
+                    .await
+            }
+            api::v1::region::remote_dyn_filter_request::Action::Unregister(unregister) => {
+                self.handle_remote_dyn_filter_unregister(&request.query_id, unregister)
+                    .await
+            }
```
evenyag: question: Do we have a mechanism to remove the dyn filter if the query is aborted?

src/common/query/src/request.rs
```diff
@@ -12,10 +12,162 @@
...
+    Datafusion(Vec<u8>),
```
evenyag: What does binary in the serialized json look like? Is it an array of u8?

src/session/src/lib.rs
discord9: > Normally, we should have a trace id in the RegionRequestHeader::tracing_context. Should we reuse that as the query id?

I’d prefer not to reuse the tracing trace id as the canonical remote dynamic filter query id.

`remote_query_id` is intended to be an internal query-lifecycle/control-plane identifier. In the current branch it is generated internally as a `QueryId`/UUID v7 and protected as a reserved extension key, so external hints cannot spoof it.

`RegionRequestHeader::tracing_context` is W3C tracing propagation metadata and may come from upstream/external callers. Reusing it as `remote_query_id` would couple observability correlation with query lifecycle ownership, and could weaken the anti-spoofing boundary we added for `remote_query_id`.

I think we should keep `remote_query_id` independent and continue propagating tracing_context separately for observability/correlation. If needed, we can log both ids together.

src/datanode/src/region_server.rs
discord9: > question: Do we have a mechanism to remove the dyn filter if the query is aborted?

yes, cleanup is in later pr, which basically checks for usage of this filter then clean it up if no one is using it periodically

src/common/query/src/request.rs
discord9: > What does binary in the serialized json look like? Is it an array of u8?

changed to base64 now

src/session/src/context.rs
discord9: > We set extension when we build the QueryContextBuilder and then we check it again in QueryContextBuilder::build(). So is it necessary to set_extension every time we create the QueryContextBuilder?
> 
> Or should we provide a constructor for QueryContextBuilder that always creates a remote query id?

changed to gen id in builder

fengjiachun:
@codex review

killme2008 (approved):
LGTM

src/common/query/src/request.rs
```diff
@@ -12,10 +12,205 @@
...
+    Ok(())
+}
+
+fn validate_decoded_payload_expr(
+    expr: &Arc<dyn PhysicalExpr>,
+    input_schema: &Schema,
+) -> DataFusionResult<()> {
+    expr.apply(|node| {
+        if let Some(column) = node.as_any().downcast_ref::<Column>()
+            && input_schema.fields().get(column.index()).is_none()
```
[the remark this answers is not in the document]
killme2008: @discord9 PTAL

PR #7980, opened 2026-04-16
feat(flow): parse defer on miss src table

discord9:

## Summary
- add `WITH (...)` support to `CREATE FLOW`
- propagate flow options into operator expressions
- validate and normalize user-provided flow options
- preserve flow options in `SHOW CREATE FLOW` and `information_schema.flows.flow_definition`

## Why
This extracts the SQL parsing/validation surface from the larger pending-flow change so it can be reviewed independently.

## Out of Scope
- pending-flow runtime/meta semantics
- activation/reconcile behavior
- pending-flow sqlness lifecycle coverage

## Tests
- `cargo test -p sql create_flow`
- `cargo test -p operator validate_and_normalize_flow_options`

## PR Checklist

### DIFF
--- a/src/operator/src/expr_helper.rs
+++ b/src/operator/src/expr_helper.rs
@@ -1047,8 +1047,20 @@ pub fn to_create_flow_task_expr(
         eval_interval: eval_interval.map(|seconds| api::v1::EvalInterval { seconds }),
         comment: create_flow.comment.unwrap_or_default(),
         sql: create_flow.query.to_string(),
-        flow_options: create_flow.flow_options.into_map(),
+        flow_options: stringify_flow_options(create_flow.flow_options)?,
     })
+}
+
+fn stringify_flow_options(flow_options: OptionMap) -> Result<HashMap<String, String>> {
+    let options_len = flow_options.len();
+    let flow_options = flow_options.into_map();
+    ensure!(
+        flow_options.len() == options_len,
+        InvalidSqlSnafu {
+            err_msg: "flow options only support scalar string-compatible values".to_string(),
+        }
+    );
+    Ok(flow_options)
 }
 
 /// sanitize the flow name, remove possible quotes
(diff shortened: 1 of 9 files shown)


src/query/src/sql.rs
```diff
@@ -78,6 +79,15 @@ const VIEWS_COLUMN: &str = "Views";
...
 const TABLE_TYPE_COLUMN: &str = "Table_type";
+
+fn user_visible_flow_options(options: &HashMap<String, String>) -> OptionMap {
+    OptionMap::from(
+        options
+            .iter()
+            .filter(|(key, _)| key.as_str() != FlowType::FLOW_TYPE_KEY)
+            .map(|(key, value)| (key.clone(), value.clone())),
+    )
+}
```
evenyag, line 90: duplicated with the method in src/catalog/src/system_schema/information_schema/flows.rs

evenyag (approved):

src/operator/src/expr_helper.rs
```diff
@@ -1047,7 +1047,7 @@ pub fn to_create_flow_task_expr(
         eval_interval: eval_interval.map(|seconds| api::v1::EvalInterval { seconds }),
         comment: create_flow.comment.unwrap_or_default(),
         sql: create_flow.query.to_string(),
-        flow_options: Default::default(),
+        flow_options: create_flow.flow_options.into_map(),
```
fengjiachun, line 1050: `CreateFlow.flow_options` is converted with `OptionMap::into_map()`, which keeps only values that can be flattened to strings. The parser still accepts array/struct option values, so input like `WITH (defer_on_missing_source = [true])` will reach this point, get dropped from the map, and then pass `validate_and_normalize_flow_options()` as if no option was provided. That means malformed flow options are silently ignored instead of rejected.

src/query/src/sql.rs
discord9: > duplicated with the method in src/catalog/src/system_schema/information_schema/flows.rs

refactor to sql crate

WenyXu:
@codex review

src/operator/src/statement/ddl.rs
```diff
@@ -629,17 +681,16 @@ impl StatementExecutor {
         expr: CreateFlowExpr,
         query_context: QueryContextRef,
     ) -> Result<SubmitDdlTaskResponse> {
+        let mut expr = expr;
```
killme2008: why not just changing the function signature?

killme2008 (approved):
LGTM

NIT: add sqlness tests for this new option?

PR #7985, opened 2026-04-17
fix: upgrade mysql metadata value limit to mediumblob

WenyXu:

Previously, the MySQL metadata table stored the `v` column as `BLOB`, which limits a single value to about 64 KiB. This is too small for some metadata values and makes MySQL more restrictive than the other metadata backends.

This PR changes new MySQL metadata tables to use `MEDIUMBLOB` for the `v` column, and adds a best-effort startup upgrade for existing tables that still use `BLOB`.
The upgrade path works as follows:
- create new tables with `MEDIUMBLOB`
- inspect existing table DDL through `SHOW CREATE TABLE`
- if `v` is still `BLOB`, try to run `ALTER TABLE ... MODIFY COLUMN v MEDIUMBLOB`

To keep backward compatibility, inspection or upgrade failure does not block startup.
Tests were added for:
- new table schema creation
- legacy `BLOB` table upgrade
- large value read/write
- parser robustness for `SHOW CREATE TABLE`

## PR Checklist

### DIFF
--- a/src/common/meta/src/kv_backend/rds/mysql.rs
+++ b/src/common/meta/src/kv_backend/rds/mysql.rs
@@ -18,13 +18,15 @@ use std::sync::Arc;
 use common_telemetry::{debug, info, warn};
 use lazy_static::lazy_static;
 use regex::Regex;
-use snafu::ResultExt;
+use snafu::{OptionExt, ResultExt};
 use sqlx::mysql::MySqlRow;
 use sqlx::pool::Pool;
 use sqlx::{MySql, MySqlPool, Row, Transaction as MySqlTransaction};
 use strum::AsRefStr;
 
-use crate::error::{CreateMySqlPoolSnafu, MySqlExecutionSnafu, MySqlTransactionSnafu, Result};
+use crate::error::{
+    CreateMySqlPoolSnafu, MySqlExecutionSnafu, MySqlTransactionSnafu, Result, UnexpectedSnafu,
+};
 use crate::kv_backend::KvBackendRef;
 use crate::kv_backend::rds::{
     Executor, ExecutorFactory, ExecutorImpl, KvQueryExecutor, RDS_STORE_OP_BATCH_DELETE,
@@ -587,13 +589,11 @@ impl MySqlStore {
         sql_template_set: &MySqlTemplateSet,
     ) -> Result<()> {
         let table_name = &sql_template_set.table_name;
-        let Some(create_table_sql) = Self::fetch_create_table_sql(pool, sql_template_set).await?
-        else {
-            warn!(
-                "Missing MySQL metadata table definition for `{table_name}`, skip automatic MEDIUMBLOB upgrade"
-            );
-            return Ok(());
-        };
+        let create_table_sql = Self::fetch_create_table_sql(pool, sql_template_set)
+            .await?
+            .context(UnexpectedSnafu {
+                err_msg: format!("Failed to fetch CREATE TABLE SQL for `{table_name}`"),
+            })?;
 
         match Self::parse_value_column_blob_type(&create_table_sql) {
             Some(ValueBlobType::Blob) => {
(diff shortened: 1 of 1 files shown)


WenyXu:
@codex review

fengjiachun:
@codex review

fengjiachun (approved):
LGTM

src/common/meta/src/kv_backend/rds/mysql.rs
```diff
@@ -161,7 +175,11 @@ impl<'a> MySqlTemplateFactory<'a> {
             table_name: table_name.to_string(),
             create_table_statement: format!(
                 // Cannot be more than 3072 bytes in PRIMARY KEY
-                "CREATE TABLE IF NOT EXISTS `{table_name}`(k VARBINARY(3072) PRIMARY KEY, v BLOB);",
+                "CREATE TABLE IF NOT EXISTS `{table_name}`(k VARBINARY(3072) PRIMARY KEY, v MEDIUMBLOB);",
+            ),
+            show_create_table_statement: format!("SHOW CREATE TABLE `{table_name}`"),
+            alter_value_column_statement: format!(
+                "ALTER TABLE `{table_name}` MODIFY COLUMN v MEDIUMBLOB;"
```
MichaelScofield, line 184: Actually I don't think we should do this (alter table) automatically for users, it might cause surprise. WDYT @fengjiachun ? Besides, does old codes (read against `BLOB`) able to read the new `MEDIUMBLOB` column? This relates to downgrade compatible.

WenyXu: > Besides, does old codes (read against BLOB) able to read the new MEDIUMBLOB column? This relates to downgrade compatible.

This is compatible. BLOB and MEDIUMBLOB only differ in size, so the read path won’t be affected.
Just note that if we ever downgrade back to BLOB, data >64KB could cause issues.

WenyXu: > Actually I don't think we should do this (alter table) automatically for users, it might cause surprise. 

We can document this change in the changelog. Some users may not allow automatic schema DDL (create/alter), so they would need to manually update the metadata table. For others, metasrv can handle the ALTER TABLE automatically.

src/common/meta/src/kv_backend/rds/mysql.rs
```diff
@@ -558,6 +642,11 @@ impl MySqlStore {
             .with_context(|_| MySqlExecutionSnafu {
                 sql: sql_template_set.create_table_statement.clone(),
             })?;
+        if let Err(err) =
+            Self::maybe_upgrade_value_column_to_mediumblob(&pool, &sql_template_set).await
+        {
+            warn!(err; "Failed to auto-upgrade MySQL metadata value column, continue startup");
```
fengjiachun: This path currently continues startup when `SHOW CREATE TABLE` or `ALTER TABLE` fails. That may leave an existing `BLOB` schema unchanged while the backend still appears to initialize successfully, and the incompatibility would only show up later when a value larger than 64 KiB is written. Since this patch is meant to make larger metadata values work on existing MySQL deployments as well, it may be safer to handle the “legacy schema detected but upgrade failed” case more explicitly instead of only logging a warning.

src/common/meta/src/kv_backend/rds/mysql.rs
fengjiachun: I think a clearer approach here would be to keep new tables on MEDIUMBLOB, but require an explicit manual migration for existing BLOB tables. We can still detect the legacy schema at startup, but instead of issuing ALTER TABLE automatically, return a clear error with the migration SQL. That keeps startup semantics predictable and avoids surprising schema DDL on the metadata store.

MichaelScofield: reasonable

evenyag: Can we expose an option for migrating it automatically? If users have many deployments, it's too verbose to alter the columns for each deployment manually.

src/common/meta/src/kv_backend/rds/mysql.rs
WenyXu: [c57d3b1](https://github.com/GreptimeTeam/greptimedb/pull/7985/commits/c57d3b143c380cc0c21ab4de19a145b9f5c4ea67)

src/common/meta/src/kv_backend/rds/mysql.rs
WenyXu: I changed the behavior to automatically alter the metadata table during startup, and to stop startup if the upgrade alter fails. By default, the MySQL store automatically creates the metadata table, so automatically altering the metadata table also seems reasonable.

fengjiachun (approved):
Rest LGTM

MichaelScofield (approved):

PR #7994, opened 2026-04-19
fix: allow ipv4_num_to_string to accept valid integers

### LINKED ISSUE #7977: Make ipv4_num_to_string function accepts int32

JoeS51:

Closes [issue 7977](https://github.com/GreptimeTeam/greptimedb/issues/7977)

Previously, the `ipv4_num_to_string` function would reject any integers that were not of type `uint32` even if the passed in integer could be casted into a valid `uint32`. This PR broadens the `ipv4_num_to_string` signature to take in any of `UInt32`, `UInt64`, `Int32`, or `Int64, and then strictly casts the input to `UInt32` before formatting it as an IPv4 string.

This removes the need for explicit `::uint32` casts for valid integers while still rejecting invalid values, such as negative numbers or integers larger than `UInt32::MAX`. This PR also has unit tests and a sqlness regression test for the `Int64` case reported in the issue.

## PR Checklist

### DIFF
--- a/src/common/function/src/scalars/ip/ipv4.rs
+++ b/src/common/function/src/scalars/ip/ipv4.rs
@@ -20,7 +20,10 @@ use common_query::error::InvalidFuncArgsSnafu;
 use datafusion_common::arrow::array::{Array, AsArray, StringViewBuilder, UInt32Builder};
 use datafusion_common::arrow::compute;
 use datafusion_common::arrow::datatypes::{DataType, UInt32Type};
-use datafusion_expr::{ColumnarValue, ScalarFunctionArgs, Signature, TypeSignature, Volatility};
+use datafusion_expr::{
+    Coercion, ColumnarValue, ScalarFunctionArgs, Signature, TypeSignature, TypeSignatureClass,
+    Volatility,
+};
 use derive_more::Display;
 
 use crate::function::{Function, extract_args};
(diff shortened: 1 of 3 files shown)


killme2008 (approved):
LGTM

Thanks

src/common/function/src/scalars/ip/ipv4.rs
```diff
@@ -43,8 +43,13 @@ pub struct Ipv4NumToString {
...
         Self {
-            signature: Signature::new(
-                TypeSignature::Exact(vec![DataType::UInt32]),
+            signature: Signature::one_of(
+                vec![
+                    TypeSignature::Exact(vec![DataType::UInt32]),
+                    TypeSignature::Exact(vec![DataType::UInt64]),
+                    TypeSignature::Exact(vec![DataType::Int32]),
+                    TypeSignature::Exact(vec![DataType::Int64]),
+                ],
```
fengjiachun: Could we make this signature coercible over the integer family instead of enumerating four exact types?

This fixes the `Int64` literal case, but it looks like other valid integer inputs such as `SMALLINT`, `TINYINT`, `UInt8`, and `UInt16` would still be rejected during signature matching. Since the implementation already casts to `UInt32` at runtime, a coercible integer signature seems closer to the intended behavior here.

JoeS51: Hi @fengjiachun, thank you for pointing out this cleaner solution! I've changed the signature to use a coercible `TypeSignatureClass::Integer` instead of using exact integer types. I ran the tests and verified locally that inputs such as `TINYINT` and `UInt8` are now accepted by this function.

Let me know if this looks good to you.

fengjiachun (approved):
LGTM

PR #8007, opened 2026-04-21
feat(cli): add retry and import state fundations

fengjiachun:

This PR adds the shared retry/backoff helper and the local import state foundation for export/import v2. It also adds the related import-state error variants and atomic state file persistence helpers. Resume/coordinator wiring will be done in the next PR.

## PR Checklist


src/cli/src/data/retry.rs
```diff
@@ -0,0 +1,194 @@
...
+            max_retries: 3,
+            initial_backoff: Duration::from_secs(1),
+            max_backoff: Duration::from_secs(300),
+            multiplier: 2,
+        }
+    }
+}
+
+impl RetryConfig {
+    pub(crate) fn backoff_delay(&self, attempt: usize) -> Duration {
```
WenyXu: nit: Should we use `backon` to stay consistent with the repository’s existing code style?

src/cli/src/data/import_v2/state.rs
```diff
@@ -0,0 +1,370 @@
...
+            Some(PathBuf::from(drive).join(path))
+        })
+}
+
+fn default_state_path_with(
+    home: Option<&Path>,
+    cwd: Option<&Path>,
+    snapshot_id: &str,
+) -> Option<PathBuf> {
+    let file_name = import_state_file_name(snapshot_id);
```
discord9: if user want to import one snapshot to two different target_addr, would here also need target_addr also be in state file key to prevent later resume having different progress for different address?

src/cli/src/data/import_v2/error.rs
```diff
@@ -117,10 +149,15 @@ impl ErrorExt for Error {
             | Error::IncompleteSnapshot { .. }
             | Error::EmptyChunkManifest { .. }
             | Error::MissingChunkData { .. } => StatusCode::InvalidArguments,
+            Error::ImportStateUnknownChunk { .. } => StatusCode::Internal,
```
discord9: should this be Unexpected since I thought internal is retryable in our error def?

src/cli/src/data/import_v2/state.rs
```diff
@@ -0,0 +1,370 @@
...
+        .await
+        .context(ImportStateIoSnafu {
+            path: path.display().to_string(),
+        })?;
+    sync_parent_dir(path).await?;
+    Ok(())
+}
+
+pub(crate) async fn delete_import_state(path: &Path) -> Result<()> {
+    match tokio::fs::remove_file(path).await {
```
discord9, line 275: does delete import state also need a `sync_parent_dir` just in case?

src/cli/src/data/retry.rs
fengjiachun: Yes — I switched this over to `backon` so the retry foundation now follows the same pattern we already use elsewhere in the repository, while keeping only a small CLI-specific policy helper.

src/cli/src/data/import_v2/state.rs
fengjiachun: Good catch. I changed the default local state key to include both `snapshot_id` and `target_addr`, so imports of the same snapshot to different targets no longer alias to the same state file or lock.

src/cli/src/data/import_v2/state.rs
fengjiachun: 40a961c

src/cli/src/data/import_v2/error.rs
fengjiachun: 40a961c

src/cli/src/data/import_v2/state.rs
```diff
@@ -0,0 +1,370 @@
...
+        tokio::fs::create_dir_all(parent)
+            .await
+            .context(ImportStateIoSnafu {
+                path: parent.display().to_string(),
+            })?;
+    }
+
+    let bytes = serde_json::to_vec_pretty(state).context(ImportStateSerializeSnafu)?;
+    let tmp_path = path.with_extension("tmp");
+    let mut file = tokio::fs::File::create(&tmp_path)
```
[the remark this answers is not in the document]
fengjiachun: [9e44005](https://github.com/GreptimeTeam/greptimedb/pull/8007/commits/9e44005cd23846e871b27e9eb1047030d9d7dba9)

discord9 (approved):

WenyXu (approved):

Issue #7986, opened 2026-04-17
Failed to delete manifest files in object store results in incorrect checkpoint state


MichaelScofield:
### What type of bug is this?

Unexpected error, Other

### What subsystems are affected?

Storage Engine

### Minimal reproduce step

Make deletion failed on S3 (some verbose information are removed):

```text
{"timestamp":"2026-04-17T06:31:55.210054Z","level":"ERROR","fields":{"message":"service=s3 name=greptimedb queued=1000 deleted=1000: delete failed","target":"opendal::services"}
{"timestamp":"2026-04-17T06:31:55.249733Z","level":"ERROR","fields":{"message":"Failed to delete manifest actions until version 961922 for region 9302899163136(2166, 0)","err":"0: OpenDAL operator failed, at greptimedb/src/mito2/src/manifest/[storage.rs:305](http://storage.rs:305/):14\n1: Unexpected (persistent) at delete => S3Error { code: \"InappropriateJSON\", deleted: 1000\n"},"target":"mito2::manifest::checkpointer"}
```

The manifest files are not deleted in S3 (which is actually ok, I think), then checkpoint files and last checkpoint are not updated correctly, finally turned into a long Datanode startup time.

### What did you expect to see?

Failed to delete the obsoleted manifest files should not hinder checkpoint file update.

### What did you see instead?

the oppsite

### What operating system did you use?

ubuntu

### What version of GreptimeDB did you use?

1.0

### Relevant log output and stack trace

```bash

```

PR #8009, opened 2026-04-21
fix: manifest recovery scans after last version if possible

evenyag:

https://github.com/GreptimeTeam/greptimedb/issues/7986

When `RegionManifestManager::has_update` (and other manifest scans) listed the manifest directory, it pulled every entry and then filtered by version in memory. For regions with long histories the object-store `list` returned many files that were immediately discarded, which made the check unnecessarily expensive on object stores.

This PR pushes a version lower bound into the `list` request via OpenDAL's `start_after`, so older manifest files are skipped at the storage layer instead of being listed and filtered client-side.

Main changes:
- Add `list_start_after(version)` helper that returns the 20-digit bare key used as a strict-exclusive lower bound. Manifest files are named `{version:020}.{json,checkpoint}[.gz]`; because `.` (ASCII 46) sorts before `'0'` (48), the bare 20-digit key is strictly less than any `{version:020}.*` file, so entries with `v < version` are skipped while entries with `v >= version` are kept.
- `DeltaStorage::manifest_lister` / `get_paths` now accept `start_after: Option<&str>` and forward it to the lister. `ManifestObjectStore::manifest_lister` plumbs this through.
- `DeltaStorage::scan(start, end)` uses `list_start_after(start)` when `start > 0` (skipped when `start == 0` because there is nothing to exclude).
- `RegionManifestManager::has_update` uses `list_start_after(last_version)`. Files for `v == last_version` may still be returned (the bare key sorts before `{v:020}.json`) but are filtered out by the existing `version > last_version` check.
- The deletion path (`delete_until`) intentionally passes `start_after = None`: a previously interrupted deletion may have left stale files below the current checkpoint, and the lister must surface them so cleanup can finish.
- Staging storage keeps its previous behavior by passing `None`.

Tests:
- Unit tests `test_scan_with_start_after_{uncompress,compress}` cover `start > 0`, `start == 0`, an exclusive upper bound, and `start` beyond any existing version, with a checkpoint sharing the directory.
- Integration test `test_object_list_start_after` verifies OpenDAL's `start_after` semantics against FS / S3 / OSS / Azblob / GCS backends, including the case where the bound matches an existing key exactly (which excludes that key).

Also downgrades the "Empty root dir not found before removal" log from `warn!` to `info!`, since the condition is benign.

## PR Checklist

### DIFF
--- a/src/mito2/src/manifest/manager.rs
+++ b/src/mito2/src/manifest/manager.rs
@@ -35,8 +35,8 @@ use crate::manifest::action::{
 };
 use crate::manifest::checkpointer::Checkpointer;
 use crate::manifest::storage::{
-    ManifestObjectStore, file_version, is_checkpoint_file, is_delta_file, manifest_compress_type,
-    manifest_dir,
+    ManifestObjectStore, file_version, is_checkpoint_file, is_delta_file, list_start_after,
+    manifest_compress_type, manifest_dir,
 };
 use crate::metrics::MANIFEST_OP_ELAPSED;
 use crate::region::{ManifestStats, RegionLeaderState, RegionRoleState};
@@ -652,13 +652,17 @@ impl RegionManifestManager {
     pub async fn has_update(&self) -> Result<bool> {
         let last_version = self.last_version();
 
-        let streamer =
-            self.store
-                .manifest_lister(false)
-                .await?
-                .context(error::EmptyManifestDirSnafu {
-                    manifest_dir: self.store.manifest_dir(),
-                })?;
+        // Skip older files at the object-store layer. Files for `v == last_version`
+        // may still appear (the 20-digit bare key sorts before `{v:020}.json`) but
+        // they are filtered out below by the `version > last_version` check.
+        let start_after = list_start_after(last_version);
+        let streamer = self
+            .store
+            .manifest_lister(false, Some(&start_after))
+            .await?
+            .context(error::EmptyManifestDirSnafu {
+                manifest_dir: self.store.manifest_dir(),
+            })?;
 
         let need_update = streamer
             .try_any(|entry| async move {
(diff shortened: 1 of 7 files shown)


MichaelScofield:
@codex review
@evenyag ci failed

src/mito2/src/manifest/storage/delta.rs
```diff
@@ -90,16 +100,22 @@ impl<T: Tracker> DeltaStorage<T> {
...
+        F: FnMut(Entry) -> Option<R>,
     {
-        let Some(streamer) = self.manifest_lister().await? else {
+        let Some(streamer) = self.manifest_lister(start_after).await? else {
             return Ok(vec![]);
         };
 
         streamer
-            .try_filter_map(|e| async { Ok(filter(e)) })
+            .try_filter_map(|e| {
```
WenyXu, line 115: ?

WenyXu (approved):

WenyXu (approved):
Rest LGTM

src/mito2/src/manifest/storage/delta.rs
evenyag: We can't put FnMut inside the async block.

src/mito2/src/manifest/manager.rs
```diff
@@ -652,13 +652,17 @@ impl RegionManifestManager {
...
-                })?;
+        // Skip older files at the object-store layer. Files for `v == last_version`
+        // may still appear (`{path}{v:020}` sorts before `{path}{v:020}.json`) but
+        // they are filtered out below by the `version > last_version` check.
+        let start_after = list_start_after(self.store.manifest_dir(), last_version);
+        let streamer = self
+            .store
+            .manifest_lister(false, Some(&start_after))
+            .await?
+            .context(error::EmptyManifestDirSnafu {
```
MichaelScofield, line 663:
```suggestion
            .with_context(|_| error::EmptyManifestDirSnafu {
```

MichaelScofield (approved):

PR #8011, opened 2026-04-21
feat: json2 flush

MichaelScofield:

implement flush for json2 

## PR Checklist

### DIFF
--- a/src/datatypes/src/types/json_type.rs
+++ b/src/datatypes/src/types/json_type.rs
@@ -317,6 +317,12 @@ fn merge(this: &JsonNativeType, that: &JsonNativeType) -> JsonNativeType {
     }
 }
 
+impl From<&ArrowDataType> for JsonType {
+    fn from(t: &ArrowDataType) -> Self {
+        JsonType::new_json2(JsonNativeType::from(&ConcreteDataType::from_arrow_type(t)))
+    }
+}
+
 impl DataType for JsonType {
     fn name(&self) -> String {
         match &self.format {
--- a/src/datatypes/src/vectors/json/array.rs
+++ b/src/datatypes/src/vectors/json/array.rs
@@ -1,0 +1,352 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::cmp::Ordering;
+use std::sync::Arc;
+
+use arrow::compute;
+use arrow::util::display::{ArrayFormatter, FormatOptions};
+use arrow_array::cast::AsArray;
+use arrow_array::{Array, ArrayRef, GenericListArray, ListArray, StructArray, new_null_array};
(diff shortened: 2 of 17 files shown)


MichaelScofield:
@codex review
@gemini-code-assist review

src/datatypes/src/types/json_type.rs
```diff
@@ -317,6 +317,12 @@ fn merge(this: &JsonNativeType, that: &JsonNativeType) -> JsonNativeType {
     }
 }
 
+impl From<&ArrowDataType> for JsonType {
```
fengys1996, line 320: nitpick: Can all Arrow data types be converted to native JSON types? Do we need to use TryFrom here?

src/datatypes/src/vectors/json/array.rs
```diff
@@ -0,0 +1,342 @@
...
+                    .fail();
+                }
+            }
+        }
+        if i < expect_fields.len() {
+            for field in &expect_fields[i..] {
+                aligned.push(new_null_array(field.data_type(), struct_array.len()));
+            }
+        }
+        if j < array_fields.len() {
```
fengys1996, line 127: nitpick: use `snafu::ensure!`

src/datatypes/src/vectors/json/array.rs
```diff
@@ -0,0 +1,342 @@
...
+        let formatter = ArrayFormatter::try_new(&self.inner, &FormatOptions::default())
+            .context(ArrowComputeSnafu)?;
+        let values = (0..self.inner.len())
+            .map(|i| {
+                self.inner
+                    .is_valid(i)
+                    .then(|| formatter.value(i).to_string())
+            })
+            .collect::<Vec<_>>();
+        Ok(Arc::new(StringArray::from(values)))
```
fengys1996, line 169: nitpick
We currently use `serde_json` for serialization, which produces a valid UTF-8 string. However, since the data will later be converted to a Variant, it would be more appropriate to use a binary representation here.

src/datatypes/src/vectors/json/array.rs
```diff
@@ -0,0 +1,342 @@
...
+                            }
+                            (DataType::List(expect_item), DataType::List(array_item)) => {
+                                let list_array = array_columns[j].as_list::<i32>();
+                                let item_aligned =
+                                    match (expect_item.data_type(), array_item.data_type()) {
+                                        (DataType::Struct(_), DataType::Struct(_)) => {
+                                            JsonArray::from(list_array.values())
+                                                .try_align(expect_item.data_type())?
+                                        }
+                                        _ => JsonArray::from(list_array.values())
```
fengys1996, line 90: It seems this only recursively handles List<Struct>. Would it still work correctly for schema evolution within nested arrays, such as List<List<Struct<a>>> -> List<List<Struct<a, b>>>? When the item is still a List, it appears to go straight to try_cast().

src/datatypes/src/types/json_type.rs
MichaelScofield: Yes, the JSON type is created by ourselves, we can ensure that. `TryFrom` is more robust indeed, I have tried to use it here, only found out that more and more codes need to be changed as well. I gave up finally when it came to `pb_value_to_value_ref`. Just can't do it easily.

src/datatypes/src/vectors/json/array.rs
MichaelScofield: It's actually needed to do some special "cast" according to the `to_type`. I've leave a todo here, upcoming PRs will resolve it.

MichaelScofield:
@fengys1996 PTAL

fengys1996:
It seems that there is an error when executing FLUSH TABLE for jsonv2 in sqlness.

sunng87 (approved):

fengys1996 (approved):
LGTM

Issue #7913, opened 2026-04-03
bug: PromQL time filter is not pushed down when time index is not millisecond


discord9:
## What type of bug is this?
- Performance issue
- Incorrect result risk in planning/pushdown

## What subsystems are affected?
- Query Engine
- Storage Engine

## Minimal reproduce step
1. Create or query a metric table whose time index uses a non-millisecond precision, for example `Timestamp(ns)` or `Timestamp(us)`.
2. Run a PromQL query with a bounded time range, such as an instant or range selector that should only read a small time window.
3. Inspect the logical plan or the scan request filters.

Current non-ms plans can look like:

```text
Filter: timestamp >= TimestampMillisecond(...) AND timestamp <= TimestampMillisecond(...)
  Projection: CAST(metrics.timestamp AS Timestamp(ms)) AS timestamp
    TableScan: metrics [timestamp:Timestamp(ns)]
```

Because the filter sits above a projection that casts the time index, the time predicate is not pushed down into `TableScan`. As a result, `ScanRequest.filters` misses the time range predicate, and storage-side pruning falls back to the full range.

## What did you expect to see?
PromQL time predicates should be pushed down to the scan even when the underlying time index is not millisecond precision, so SST or file pruning can use a compact time range predicate.

## What did you see instead?
The planner keeps the time filter above a cast projection for non-ms time indexes. Storage receives no usable time-range filter for pruning, which can degrade bounded PromQL queries into full scans across SST files.

## What operating system did you use?
Linux x64

## What version of GreptimeDB did you use?
main (observed on current source checkout)

## Relevant log output and stack trace
No crash. The issue is visible from the generated plan shape and missing scan-level time filter pushdown for non-ms PromQL queries.

Possible fix direction: build the scan predicate on the raw time-index unit first, then project or cast the output timestamp back to millisecond precision for the rest of the PromQL plan.

discord9:
closed by #7926

discord9:
reopen with https://github.com/GreptimeTeam/datafusion/pull/11

PR #8019, opened 2026-04-22
fix: window sort off by one precision TimeRange&better alias track

discord9:

This PR fixes two window sort issues around time-index sorting:

1. Window sort could fail to recognize aliased or wrapped time-index expressions, so some queries that should use the `WindowedSortExec` path would fall back to a normal sort plan.
2. Precision conversion between partition ranges and sort expressions could introduce an off-by-one boundary issue when timestamp precision changes, especially around exclusive range ends.

To address this, this PR:

- improves time-index expression tracking in `WindowedSortPhysicalRule`
- allows wrapped time-index expressions such as direct aliases, casts, and supported `to_timestamp_*` wrappers to be recognized as sortable time-index expressions
- fixes precision conversion for partition/time ranges used by window sort
- ensures exclusive end bounds are rounded correctly during timestamp unit conversion
- adds targeted unit tests for alias tracking and precision-boundary behavior
- adds sqlness regression coverage for the alias/wrapper cases in both standalone and distributed modes

A notable testing detail is that standalone and distributed produce slightly different `EXPLAIN ANALYZE` plans for wrapped expressions like `to_timestamp_millis(ts)`, so the sqlness cases are split per mode instead of sharing a common expected result.

## Why this matters

Queries such as the following should still benefit from window sort optimization:

- `SELECT ts AS ts ... ORDER BY ts`
- `SELECT to_timestamp_millis(ts) AS ts ... ORDER BY ts`

Without this fix, these queries may miss the optimized window sort path. Also, incorrect precision projection at range boundaries can cause rows near the end of a range to be sliced incorrectly.

## Testing

- added unit tests for:
  - alias tracking through projection
  - wrapped/casted time-index expression tracking
  - exclusive-end precision conversion
  - boundary preservation after precision conversion
- added sqlness regression tests for `windowed_sort_advance`

## PR Checklist

### DIFF
--- a/src/query/src/optimizer/windowed_sort.rs
+++ b/src/query/src/optimizer/windowed_sort.rs
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashSet;
 use std::sync::Arc;
 
+use arrow_schema::DataType;
 use datafusion::physical_optimizer::PhysicalOptimizerRule;
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
@@ -26,7 +26,8 @@ use datafusion::physical_plan::sorts::sort::SortExec;
 use datafusion::physical_plan::sorts::sort_preserving_merge::SortPreservingMergeExec;
 use datafusion_common::Result as DataFusionResult;
 use datafusion_common::tree_node::{Transformed, TreeNode};
-use datafusion_physical_expr::expressions::Column as PhysicalColumn;
+use datafusion_physical_expr::expressions::{CastExpr, Column as PhysicalColumn};
+use datafusion_physical_expr::{PhysicalExpr, ScalarFunctionExpr};
 use store_api::region_engine::PartitionRange;
 use table::table::scan::RegionScanExec;
 
@@ -91,9 +92,11 @@ impl WindowedSortPhysicalRule {
                         .expr
                         .as_any()
                         .downcast_ref::<PhysicalColumn>()
-                        && scanner_info
-                            .time_index
-                            .contains(input_schema.field(column_expr.index()).name())
+                        && matches!(
+                            input_schema.field(column_expr.index()).data_type(),
+                            DataType::Timestamp(_, _)
+                        )
+                        && is_time_index_expr(sort_input.clone(), first_sort_expr.expr.clone())?
                         && sort_exec.fetch().is_none()
                     // skip if there is a limit, as dyn filter along is good enough in this case
                     {
@@ -154,14 +157,11 @@ impl WindowedSortPhysicalRule {
 #[derive(Debug)]
 struct ScannerInfo {
     partition_ranges: Vec<Vec<PartitionRange>>,
-    time_index: HashSet<String>,
     tag_columns: Vec<String>,
 }
 
 fn fetch_partition_range(input: Arc<dyn ExecutionPlan>) -> DataFusionResult<Option<ScannerInfo>> {
     let mut partition_ranges = None;
-    let mut time_index = HashSet::new();
-    let mut alias_map = Vec::new();
     let mut tag_columns = None;
 
     input.transform_up(|plan| {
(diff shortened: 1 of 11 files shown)


waynexia:
@codex review

src/query/src/optimizer/windowed_sort.rs
```diff
@@ -219,14 +205,90 @@ fn fetch_partition_range(input: Arc<dyn ExecutionPlan>) -> DataFusionResult<Opti
...
+    is_time_index_expr(child, Arc::new(column_expr.clone()))
+}
+
+fn passthrough_child(plan: &dyn ExecutionPlan) -> Option<Arc<dyn ExecutionPlan>> {
+    if plan.as_any().is::<FilterExec>()
+        || plan.as_any().is::<CoalescePartitionsExec>()
+        || plan.as_any().is::<RepartitionExec>()
+        || plan.as_any().is::<CooperativeExec>()
+    {
+        return plan.children().first().cloned().cloned();
```
evenyag: Why do we need to clone twice?

src/query/src/optimizer/windowed_sort.rs
```diff
@@ -219,14 +205,90 @@ fn fetch_partition_range(input: Arc<dyn ExecutionPlan>) -> DataFusionResult<Opti
...
+        };
+        return is_time_index_expr(projection.input().clone(), projection_expr.expr.clone());
+    }
+
+    if let Some(region_scan_exec) = plan.as_any().downcast_ref::<RegionScanExec>() {
+        return Ok(matches!(
+            plan.schema().field(column_expr.index()).data_type(),
+            DataType::Timestamp(_, _)
+        ) && plan.schema().field(column_expr.index()).name().as_ref()
+            == region_scan_exec.time_index());
```
evenyag: Is it safe to get the field using `column_expr.index`?

src/query/src/optimizer/windowed_sort.rs
discord9: > Why do we need to clone twice?

`first()` return `Option<&&Arc<dyn ExecutionPlan>>` so two cloned to deref twice

evenyag (approved):

src/query/src/optimizer/windowed_sort.rs
```diff
@@ -219,14 +205,91 @@ fn fetch_partition_range(input: Arc<dyn ExecutionPlan>) -> DataFusionResult<Opti
...
+    }
+
+    let Some(child) = passthrough_child(plan.as_ref()) else {
+        return Ok(false);
+    };
+    is_time_index_expr(child, Arc::new(column_expr.clone()))
+}
+
+fn passthrough_child(plan: &dyn ExecutionPlan) -> Option<Arc<dyn ExecutionPlan>> {
+    if plan.as_any().is::<FilterExec>()
```
evenyag: Can we ensure FilterExec always has the same schema as its child?

discord9: > Can we ensure FilterExec always has the same schema as its child?

add support for project inside filter instead

src/query/src/optimizer/windowed_sort.rs
discord9: > Is it safe to get the field using `column_expr.index`?

fix by add guard to make sure schema is the same before passthrough

killme2008:
Could it solve #7913 ?

src/query/src/optimizer/windowed_sort.rs
```diff
@@ -219,14 +205,110 @@ fn fetch_partition_range(input: Arc<dyn ExecutionPlan>) -> DataFusionResult<Opti
...
-            time_index,
             tag_columns: tag_columns?,
         }
     };
 
     Ok(result)
 }
 
+fn is_time_index_expr(
+    plan: Arc<dyn ExecutionPlan>,
```
killme2008: Looks like we can use `&Arc<...>` here? Avoid the cloning externally.

src/query/src/optimizer/windowed_sort.rs
```diff
@@ -219,14 +205,110 @@ fn fetch_partition_range(input: Arc<dyn ExecutionPlan>) -> DataFusionResult<Opti
...
+
+fn schema_preserving_child(plan: &dyn ExecutionPlan) -> Option<Arc<dyn ExecutionPlan>> {
+    let child = plan.children().first().cloned().cloned()?;
+    (plan.schema().as_ref() == child.schema().as_ref()).then_some(child)
+}
+
+fn is_supported_time_index_wrapper(expr: &ScalarFunctionExpr) -> bool {
+    matches!(
+        expr.name(),
+        "to_timestamp"
```
killme2008: Consider case sensitivity?

killme2008 (approved):
LGTM

discord9:
> Could it solve #7913 ?

it's a bit different, could just make TableScan accept a wider range of filter or even simplify them in TableScan, but we could also use a public helper for determine if a column is or derived from time index column since it's quite common

killme2008 (approved):
LGTM

PR #8025, opened 2026-04-23
feat: use SemanticType::FollowSchema for pre-defined tables

sunng87:

Rationale:

For created table, this semantic type in our grpc protocol becomes verbose information. For example, when we adjust table schema in #8008 , we will also need to adjust semantic types when constructing `InsertRequest`, by updating tag/field for certain columns. This will make the server layer code a breaking change, those created table won't work even if column names didn't change at all.

In this patch, I just add a new semantic type called `FollowSchema`. As its name suggests, it will follow the table schema for the column's semantic type. 

1. when table is not found, the server will throw an error
2. when column with follow schema is not found, it treats the column as field

Related change: https://github.com/GreptimeTeam/greptime-proto/pull/316

## PR Checklist

### DIFF
--- a/src/api/src/helper.rs
+++ b/src/api/src/helper.rs
@@ -877,8 +877,10 @@ pub fn pb_value_to_value_ref<'a>(
 }
 
 /// Returns true if the pb semantic type is valid.
+/// If the input semantic type is [SemanticType::FollowSchema], it always returns true
+/// as the caller should use the table's semantic type instead.
 pub fn is_semantic_type_eq(type_value: i32, semantic_type: SemanticType) -> bool {
-    type_value == semantic_type as i32
+    type_value == semantic_type as i32 || type_value == SemanticType::FollowSchema as i32
 }
 
 /// Returns true if the pb type value is valid.
--- a/src/mito2/src/request.rs
+++ b/src/mito2/src/request.rs
@@ -176,22 +176,27 @@ impl WriteRequest {
     ///
     /// If column with default value is missing, it returns a special [FillDefault](crate::error::Error::FillDefault)
     /// error.
-    pub(crate) fn check_schema(&self, metadata: &RegionMetadata) -> Result<()> {
+    /// If a column's semantic type is FollowSchema, it means we will use table semantic type.
+    pub(crate) fn check_schema(&mut self, metadata: &RegionMetadata) -> Result<()> {
         debug_assert_eq!(self.region_id, metadata.region_id);
 
         let region_id = self.region_id;
         // Index all columns in rows.
-        let mut rows_columns: HashMap<_, _> = self
+        let mut name_to_index: HashMap<_, _> = self
             .rows
             .schema
             .iter()
-            .map(|column| (&column.column_name, column))
+            .enumerate()
+            .map(|(i, column)| (&column.column_name, i))
             .collect();
 
         let mut need_fill_default = false;
+        let mut follow_schema_fixups: Vec<(usize, i32)> = vec![];
         // Checks all columns in this region.
         for column in &metadata.column_metadatas {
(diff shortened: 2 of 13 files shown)


v0y4g3r:
> when column with follow schema is not found, it treats the column as field

Shouldn't we also treat this as an error case when column is not found and request asks for "follow the column semantics in table"?

sunng87:
> Shouldn't we also treat this as an error case when column is not found and request asks for "follow the column semantics in table"?

There is case like adding new field for existing table. For example, trace span may carry new attribute columns.

src/api/src/helper.rs
```diff
@@ -877,8 +877,10 @@ pub fn pb_value_to_value_ref<'a>(
 }
 
 /// Returns true if the pb semantic type is valid.
+/// If the input semantic type is [SemanticType::FollowSchema], it always returns true
+/// as the caller should use the table's semantic type instead.
 pub fn is_semantic_type_eq(type_value: i32, semantic_type: SemanticType) -> bool {
```
killme2008, line 882: Love to change this function name too, such as `is_semantic_type_compatible`?

src/mito2/src/request.rs
```diff
@@ -264,6 +274,12 @@ impl WriteRequest {
         // If we need to fill default values, return a special error.
         ensure!(!need_fill_default, FillDefaultSnafu { region_id });
 
+        // Apply FollowSchema fixups: replace with the table's semantic type.
+        // This must happen after `name_to_index` is dropped to release the borrow on `self.rows.schema`.
+        for (idx, semantic_type) in follow_schema_fixups {
```
killme2008, line 279: Better to add debug logs for `follow_schema_fixups` with table id or name etc.

src/mito2/src/request.rs
```diff
@@ -264,6 +274,12 @@ impl WriteRequest {
...
+            self.rows.schema[idx].semantic_type = semantic_type;
```
killme2008, line 280: I suggest keeping `check_schema` side-effect free, returning `follow_schema_fixups`. We can then update the request externally.

Semantic schema checks should not modify the schema.

killme2008: And better to add a unit test for this fixing.

src/operator/src/insert.rs
```diff
@@ -565,6 +565,21 @@ impl Inserter {
...
                 None => {
+                    if let Some(rows) = &req.rows
+                        && rows
+                            .schema
+                            .iter()
+                            .any(|col| col.semantic_type == SemanticType::FollowSchema as i32)
+                    {
+                        return InvalidInsertRequestSnafu {
+                                reason: format!(
+                                    "Table `{}` does not exist, cannot use semantic type FollowSchema for auto-created table",
```
killme2008, line 576: Is `FollowSchema` solely for schema compatibility during schema changes?

src/operator/src/insert.rs
evenyag, line 549: I think we should fix the semantic type in the frontend.

src/mito2/src/sst/parquet/format.rs
```diff
@@ -386,7 +385,7 @@ impl PrimaryKeyReadFormat {
         };
         match column.semantic_type {
             SemanticType::Tag => self.tag_values(row_groups, column, true),
-            SemanticType::Field => {
+            SemanticType::Field | SemanticType::FollowSchema => {
```
evenyag, line 388: It doesn't make sense to treat FollowSchema as Field in the datanode. We may expect semantic type to be specific in datanode and fix it in frontend.

Maybe we should decouple the grpc semantic type and the semantic type in the ColumnMetadata.

MichaelScofield:
@codex review

PR #8035, opened 2026-04-24
fix(server): describe EXPLAIN statements so bind parameters work

### LINKED ISSUE #8029: # EXPLAIN / EXPLAIN ANALYZE rejects bind parameters
### What type of bug is this?

Unexpected error

Query Engine

## Summary

Both `EXPLAIN` and `EXPLAIN ANALYZE` fail with `unsupported_parameter_type` (SQLSTATE 22023) when the inner query uses a bind parameter over the Postgres wire protocol. The same SELECT without `EXPLAIN` executes successfully with the same parameter binding, and the same `EXPLAIN ANALYZE` with a literal in place of the parameter also works.

Validated against Postgres 17 and MySQL 8 — both accept bind parameters in `EXPLAIN` and `EXPLAIN ANALYZE`. The issue is specific to GreptimeDB.

## Environment

- GreptimeDB: `greptime/greptimedb:v1.0.0` (standalone mode in Docker)
- Postgres: `postgres:17-alpine` (control)
- MySQL: `mysql:8` (control)
- Client: Bun.sql (Postgres wire) + `mysql2` (MySQL)
- Observed: 2026-04-24

## Minimal reproduction

Start a standalone GreptimeDB, connect over the Postgres wire protocol, and run four probes against a trivial 3-row table. No benchmark data required.

```bash
docker run -d --rm --name gt -p 4003:4003 greptime/greptimedb:v1.0.0 \
  standalone start --postgres-addr 0.0.0.0:4003 --http-addr 0.0.0.0:4000
```

```ts
import { SQL } from 'bun';
const sql = new SQL('postgres://greptime@localhost:4003/public', { prepare: false });

await sql`CREATE TABLE t (k VARCHAR(36) NOT NULL, ts TIMESTAMP(3) NOT NULL, TIME INDEX(ts))`;
await sql`INSERT INTO t (k, ts) VALUES ('a', 1), ('b', 2), ('c', 3)`;

// Probe 1 — control: normal SELECT with bind param (expected: works)
await sql`SELECT k FROM t WHERE k = ${'a'}`;

// Probe 2 — EXPLAIN with bind param (observed: fails)
await sql`EXPLAIN SELECT k FROM t WHERE k = ${'a'}`;

// Probe 3 — EXPLAIN ANALYZE with bind param (observed: fails)
await sql`EXPLAIN ANALYZE SELECT k FROM t WHERE k = ${'a'}`;

// Probe 4 — control: EXPLAIN ANALYZE with literal (expected: works)
await sql.unsafe(`EXPLAIN ANALYZE SELECT k FROM t WHERE k = 'a'`);
```

Probes 2 and 3 throw `unsupported_parameter_type`. Probes 1 and 4 succeed. The same four probes run against `postgres:17-alpine` and `mysql:8` all pass.

## Results

| Probe | GreptimeDB v1.0.0 | Postgres 17 | MySQL 8 |
|---|---|---|---|
| 1. Plain SELECT with bound param     | PASS | PASS | PASS |
| 2. `EXPLAIN` with bound param        | **FAIL** (`unsupported_parameter_type`) | PASS | PASS |
| 3. `EXPLAIN ANALYZE` with bound param| **FAIL** (`unsupported_parameter_type`) | PASS | PASS |
| 4. `EXPLAIN ANALYZE` with literal    | PASS | PASS | PASS |

Only GreptimeDB rejects the parameterized EXPLAIN variants. The bind machinery clearly works for regular SELECTs (probe 1 passes), and the EXPLAIN ANALYZE plan node works when the value is inlined (probe 4 passes) — the failure is specifically at the intersection of EXPLAIN-level plan rewriting and parameter binding.

## Error detail

```
PostgresError: unsupported_parameter_type
    errno: "22023"
    detail: "unknown"
    severity: "ERROR"
    code: "ERR_POSTGRES_SERVER_ERROR"
```

Returned to the client at the Bind step of the extended query protocol.

## Workaround

Inline parameter values as SQL literals inside `EXPLAIN` / `EXPLAIN ANALYZE`:

```ts
await sql.unsafe(`EXPLAIN ANALYZE SELECT k FROM t WHERE k = '${value}'`);
```

This bypasses the bind path. Requires careful escaping for user-controlled values; safe for internal diagnostic code that only ever inlines known-shape values like UUIDs.

## Expected behavior

Standard Postgres accepts bind parameters in `EXPLAIN` and `EXPLAIN ANALYZE`:

```sql
PREPARE foo AS SELECT 1 FROM t WHERE k = $1;
EXPLAIN ANALYZE EXECUTE foo('value');   -- works on Postgres, validated
```

MySQL 8 likewise accepts bind parameters via the `?` placeholder inside `EXPLAIN` / `EXPLAIN ANALYZE` (validated). GreptimeDB should match this behaviour for Postgres-wire-protocol compatibility.

n/a

### What did you see instead?

n/a

n/a

1.0.0

```bash

```

BootstrapperSBL:

Fixes #8029

Over the Postgres extended query protocol, `EXPLAIN` and `EXPLAIN ANALYZE` against a query with a bind parameter returned `unsupported_parameter_type` (SQLSTATE 22023) at the Bind step, while the same `SELECT` with a bind parameter and the same `EXPLAIN` with a literal both work. Postgres 17 and MySQL 8 both accept bind parameters in `EXPLAIN`, so this was a GreptimeDB-specific gap.

### Root cause

`Instance::do_describe_inner` (`src/frontend/src/instance.rs`) only planned `Statement::Insert`, `Statement::Query`, and `Statement::Delete`. For anything else it returned `Ok(None)`, which caused the Postgres handler to store the statement in `SqlPlan::Statement(..)` instead of `SqlPlan::Plan(..)`. Only the `SqlPlan::Plan(..)` arm in `do_describe_statement` runs `get_inferred_parameter_types`, so `EXPLAIN`-wrapped queries lost parameter type inference entirely, and the handler rejected the Bind.

### Fix

Recognize `Statement::Explain` when its inner statement is itself plannable (`Insert` / `Query` / `Delete`) and route it through the same describe path. The planner already supports `EXPLAIN` end-to-end via `explain_to_plan` and produces a `LogicalPlan::Explain` / `LogicalPlan::Analyze` that wraps the inner plan, so the existing DataFusion-based `get_parameter_types` traversal picks up the placeholders. `check_permission` (top level) already lets `Statement::Explain` through, so no permission change is needed. The fix benefits the MySQL extended protocol path as well, since it goes through the same `do_describe` hook.

### Before / after

```sql
CREATE TABLE t (k VARCHAR(36) NOT NULL, ts TIMESTAMP(3) NOT NULL, TIME INDEX(ts));
INSERT INTO t (k, ts) VALUES ('a', 1), ('b', 2), ('c', 3);

-- Before: fails at Bind with unsupported_parameter_type (22023).
-- After:  succeeds, returns the plan.
EXPLAIN SELECT k FROM t WHERE k = $1;

-- Before: same failure.
-- After:  succeeds, returns the analyzed plan.
EXPLAIN ANALYZE SELECT k FROM t WHERE k = $1;
```

### Test

Adds `test_postgres_explain_bind_parameter` in `tests-integration/tests/sql.rs`, which reproduces the original issue: it creates the table from the bug report, runs a plain parameterized `SELECT` as a sanity check, then runs `EXPLAIN` and `EXPLAIN ANALYZE` with a bind parameter using `tokio-postgres` (extended query protocol).

## PR Checklist

### DIFF
--- a/src/frontend/src/instance.rs
+++ b/src/frontend/src/instance.rs
@@ -710,16 +710,14 @@ impl Instance {
         // EXPLAIN / EXPLAIN ANALYZE wrap an inner statement; describe them when the
         // wrapped statement is something we already plan (so that bind parameters
         // in the inner query get their types inferred). See #8029.
-        let plannable = matches!(
-            stmt,
-            Statement::Insert(_) | Statement::Query(_) | Statement::Delete(_)
-        ) || matches!(
-            &stmt,
-            Statement::Explain(explain) if matches!(
-                explain.statement.as_ref(),
+        let is_inner_plannable = |s: &Statement| {
+            matches!(
+                s,
                 Statement::Insert(_) | Statement::Query(_) | Statement::Delete(_)
             )
-        );
+        };
+        let plannable = is_inner_plannable(&stmt)
+            || matches!(&stmt, Statement::Explain(explain) if is_inner_plannable(explain.statement.as_ref()));
 
         if plannable {
             self.plugins
--- a/src/frontend/src/instance.rs
+++ b/src/frontend/src/instance.rs
@@ -707,10 +707,21 @@ impl Instance {
     ) -> Result<Option<DescribeResult>> {
         ensure!(!self.is_suspended(), error::SuspendedSnafu);
 
-        if matches!(
+        // EXPLAIN / EXPLAIN ANALYZE wrap an inner statement; describe them when the
+        // wrapped statement is something we already plan (so that bind parameters
+        // in the inner query get their types inferred). See #8029.
+        let plannable = matches!(
             stmt,
             Statement::Insert(_) | Statement::Query(_) | Statement::Delete(_)
-        ) {
+        ) || matches!(
+            &stmt,
+            Statement::Explain(explain) if matches!(
+                explain.statement.as_ref(),
+                Statement::Insert(_) | Statement::Query(_) | Statement::Delete(_)
+            )
+        );
+
+        if plannable {
             self.plugins
                 .get::<PermissionCheckerRef>()
                 .as_ref()
--- a/tests-integration/tests/sql.rs
+++ b/tests-integration/tests/sql.rs
@@ -83,6 +83,7 @@ macro_rules! sql_tests {
                 test_postgres_intervalstyle,
                 test_postgres_parameter_inference,
                 test_postgres_uint64_parameter,
+                test_postgres_explain_bind_parameter,
                 test_postgres_array_types,
                 test_mysql_prepare_stmt_insert_timestamp,
                 test_mysql_federated_prepare_stmt,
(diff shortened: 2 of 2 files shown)


BootstrapperSBL:
Pushed 634d97b adopting the closure suggestion — the dedup reads cleaner. Quick note on the first point: the original `matches!(stmt, ...)` didn't actually move `stmt` because the patterns only use `_` bindings (no by-value capture), which is why the build went through. Still, the refactor makes the duplication fix unambiguous. Thanks for the review.

killme2008 (approved):
Thanks!

sunng87 (approved):

PR #8036, opened 2026-04-25
feat: allow detailed index config in pipeline

### LINKED ISSUE #7951: Allow detailed index configuration in pipeline config

User experience

GreptimeDB allows user to specify index columns in [pipeline configuration](https://docs.greptime.com/reference/pipeline/pipeline-config/#the-index-field), however it only accepts basic index type, like `fulltext`/`skipping `. We should allow users to specify detailed index config in that config.

### Implementation challenges

This feature should be rather easy to implement once we reach agreement on the syntax.

QuakeWang:

close: #7951 

This PR adds detailed index config support to pipeline YAML for `fulltext` and `skipping`, while keeping the existing `index: <type>` syntax backward compatible.

It introduces object-style index config with optional `options`, reuses existing SQL/Data Index option names and validation rules, rejects unsupported `options` on `inverted` / `time` / `tag`, and aligns pipeline auto-create behavior so only `fulltext` remains string-only.

## PR Checklist

### DIFF
--- a/src/pipeline/src/etl/transform.rs
+++ b/src/pipeline/src/etl/transform.rs
@@ -162,6 +162,7 @@ impl TransformIndexOptions {
         }
     }
 
+    #[cfg(test)]
     pub(crate) fn as_fulltext(&self) -> Option<&FulltextOptions> {
         match self {
             TransformIndexOptions::Fulltext(options) => Some(options),
@@ -169,6 +170,7 @@ impl TransformIndexOptions {
         }
     }
 
+    #[cfg(test)]
     pub(crate) fn as_skipping(&self) -> Option<&SkippingIndexOptions> {
         match self {
             TransformIndexOptions::Skipping(options) => Some(options),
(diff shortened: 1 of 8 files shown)


src/pipeline/src/etl/transform.rs
```diff
@@ -131,10 +143,40 @@ pub struct Transform {
...
+            TransformIndexOptions::Skipping(_) => None,
+        }
+    }
+
+    pub(crate) fn as_skipping(&self) -> Option<&SkippingIndexOptions> {
+        match self {
+            TransformIndexOptions::Skipping(options) => Some(options),
+            TransformIndexOptions::Fulltext(_) => None,
+        }
+    }
```
shuiyisong, line 179: Let's annotate these functions with `#[cfg(test)]` as they're meant to be used only in tests.

src/pipeline/src/etl/transform.rs
```diff
@@ -231,13 +273,174 @@ fn get_default_for_type(ty: &ColumnDataType) -> Result<ValueData> {
     Ok(v)
 }
 
+fn parse_transform_index(
```
shuiyisong, line 278: From this function all the way down to `lower_transform_index_options`, it seems to be a lot of functions to parse the yaml. Can we try to simplify these functions.

shuiyisong (commented):
First of all, good job for implementing! 👍 

The correctness of the code is okay, but the implementation seems can be more concise.
Also, please add tests in the `integration-tests/http` to actually test the creation of index with options.

src/pipeline/src/etl/transform.rs
QuakeWang: Done. Added `#[cfg(test)]` to these test-only helpers.

src/pipeline/src/etl/transform.rs
QuakeWang: Done. Simplified this path by folding the small helpers into the parsing/lowering code while keeping YAML parsing and typed option lowering separate.

QuakeWang:
> First of all, good job for implementing! 👍
> 
> The correctness of the code is okay, but the implementation seems can be more concise. Also, please add tests in the `integration-tests/http` to actually test the creation of index with options.

Added an HTTP integration test that creates the table through pipeline ingest and verifies the fulltext/skipping index options with `SHOW CREATE TABLE`.

shuiyisong (approved):
LGTM 🚀

killme2008 (approved):
Good job!

Clean code and design.

PR #8038, opened 2026-04-27
feat: add read column abstraction

fengys1996:

The part of https://github.com/GreptimeTeam/greptimedb/pull/7959.

This commit introduces a new read_columns abstraction to define what the scan layer should read.

  - Added `ReadColumns` / `ReadColumn` to represent root columns (column_id) plus optional nested field paths (nested_paths, empty means read full column).
  - Added `merge(a, b)` with normalization semantics:
      - merge by column_id
      - deduplicate/fold nested paths by prefix
      - full-column read overrides nested-path reads
  - Added parsers/builders:
      - `read_columns_from_projection(...)` from ProjectionInput (+ default time-index read when projection is empty)
      - `read_columns_from_predicate(...)` from predicate-referenced root columns

## PR Checklist

### DIFF
--- a/src/mito2/src/read/read_columns.rs
+++ b/src/mito2/src/read/read_columns.rs
@@ -1,0 +1,538 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO(fys): remove it until this module is used
+#![allow(dead_code)]
+
+use std::collections::{BTreeMap, HashSet};
+use std::mem;
+
+use datafusion_common::HashMap;
+use datafusion_expr::utils::expr_to_columns;
+use snafu::OptionExt;
+use store_api::metadata::RegionMetadataRef;
+use store_api::storage::{ColumnId, NestedPath, ProjectionInput};
+
+use crate::error::{InvalidRequestSnafu, Result};
+use crate::read::scan_region::PredicateGroup;
+
+/// Logical columns to read from a region.
+///
+/// Read columns describe which logical columns and nested fields should be read
+/// from storage. Each read column is identified by its [`ColumnId`],
+/// which represents the root column in the storage schema.
+///
+/// Nested fields under the column are specified by [`NestedPath`] entries.
+/// Each path includes the root column name as its first element.
+///
+/// For example, assume column id `9` corresponds to a root column named `j`
+/// with nested fields:
+///
+/// ```text
+/// j
+/// ├── a
+/// └── b
+///     └── c
+/// ```
+///
+/// The following SQL:
+///
+/// SELECT j.a, j.b.c FROM t
+///
+/// may produce read columns like:
+///
+/// ```text
+/// ReadColumn {
+///     column_id: 9,
+///     nested_paths: [
+///         ["j", "a"],
+///         ["j", "b", "c"],
+///     ]
+/// }
+/// ```
+///
+/// If `nested_paths` is empty, the whole column will be read.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
+pub struct ReadColumns {
+    cols: Vec<ReadColumn>,
+}
+
+impl ReadColumns {
+    pub fn from_column_ids<I>(column_ids: I) -> Self
+    where
+        I: IntoIterator<Item = ColumnId>,
+    {
+        let cols = column_ids
+            .into_iter()
+            .map(|col_id| ReadColumn::new(col_id, vec![]))
+            .collect();
+        ReadColumns { cols }
+    }
+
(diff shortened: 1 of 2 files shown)


fengys1996:
@codex review

fengys1996:
@codex review

evenyag (approved):

src/mito2/src/read/read_columns.rs
```diff
@@ -0,0 +1,563 @@
...
+    }
+
+    pub fn estimated_size(&self) -> usize {
+        self.nested_paths.capacity() * mem::size_of::<NestedPath>()
+            + self
+                .nested_paths
+                .iter()
+                .map(|path| {
+                    path.capacity() * mem::size_of::<String>()
+                        + path.iter().map(|node| node.capacity()).sum::<usize>()
```
MichaelScofield, line 141: Better make a new "`estimated_size`" method in `NestedPath`.

src/mito2/src/read/read_columns.rs
```diff
@@ -0,0 +1,563 @@
...
+
+    pub fn column_id(&self) -> ColumnId {
+        self.column_id
+    }
+
+    pub fn nested_paths(&self) -> &[NestedPath] {
+        &self.nested_paths
+    }
+
+    pub fn estimated_size(&self) -> usize {
```
MichaelScofield, line 133: Missing `column_id`?

src/mito2/src/read/read_columns.rs
```diff
@@ -0,0 +1,563 @@
...
+    }
+}
+
+pub fn merge(a: ReadColumns, b: ReadColumns) -> ReadColumns {
+    let mut merged = BTreeMap::<ColumnId, Vec<NestedPath>>::new();
+
+    for col in a.cols.into_iter().chain(b.cols) {
+        if let Some(nested_paths) = merged.get_mut(&col.column_id) {
+            if nested_paths.is_empty() || col.nested_paths.is_empty() {
+                *nested_paths = vec![];
```
MichaelScofield, line 153: Why reset the old `nested_paths`?

src/mito2/src/read/read_columns.rs
```diff
@@ -0,0 +1,563 @@
...
+            })
+            .collect(),
+    }
+}
+
+fn normalize_nested_paths(nested_paths: Vec<NestedPath>) -> Vec<NestedPath> {
+    let mut normalized = Vec::with_capacity(nested_paths.len());
+    merge_nested_paths(&mut normalized, nested_paths);
+    normalized
+}
```
MichaelScofield, line 178: Seems the returned value is same to the `nested_paths`? If so, there's no need to make this function.

src/mito2/src/read/read_columns.rs
```diff
@@ -0,0 +1,563 @@
...
+            .get(root_idx)
+            .with_context(|| InvalidRequestSnafu {
+                region_id: metadata.region_id,
+                reason: format!("projection index {} is out of bounds", root_idx),
+            })?;
+        let col_id = col.column_id;
+
+        let nested_paths = paths_by_col
+            .remove(&col.column_schema.name)
+            .unwrap_or_default();
```
MichaelScofield, line 245: No need to use `seen`:

```suggestion
    for root_idx in root_indices {
        let col = metadata
            .column_metadatas
            .get(root_idx)
            .with_context(|| InvalidRequestSnafu {
                region_id: metadata.region_id,
                reason: format!("projection index {} is out of bounds", root_idx),
            })?;
        let col_id = col.column_id;

        let Some(nested_paths) = paths_by_col
            .remove(&col.column_schema.name) else {
            continue;
        };
```

fengys1996: `seen` is necessary: if root indices contain duplicates and seen is removed, these duplicates are often computed as `nested_paths = []` (which means **reading the entire column**) during subsequent processing, unintentionally widening partial reads into full-column reads.

src/mito2/src/read/read_columns.rs
fengys1996: We reset to [] intentionally because, in this model, empty nested_paths means **read the whole root column**; once either side requests full-column read, that is the least upper bound of the merge, so any previously collected sub-paths become redundant and must be cleared to preserve correct semantics.

src/mito2/src/read/read_columns.rs
fengys1996: normalize_nested_paths performs canonicalization (for example, it normalizes `["a"]` and `["a","b"]` to keep only `["a"]`, and removes paths covered by a shorter prefix), so it is not a no-op.

src/mito2/src/read/read_columns.rs
fengys1996: `NestedPath` is currently a type alias (`pub type NestedPath = Vec<String>`), so we cannot add an inherent `estimated_size()` method directly on it;

src/mito2/src/read/read_columns.rs
fengys1996: Good catch!

src/mito2/src/read/read_columns.rs
MichaelScofield: Won't the duplicate computation be skipped by the "continue" in line 240?

fengys1996: Sorry, I didn’t read it carefully at first. This change still has an issue: it breaks reading columns without nested paths. Example: projection = [1, 0], where only field_0 has nested paths and tag_0 has none; tag_0 would hit remove(...) == None and be skipped, so it would be incorrectly dropped.

MichaelScofield (approved):

PR #8015, opened 2026-04-22
feat: add seq watermark in record batch metrics

discord9:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

as title, part of the incremental query feature

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [ ] I have written the necessary rustdoc comments.
- [ ] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #8045, opened 2026-04-28
feat: flow inc query terminal metrics transport

discord9:

This PR wires terminal record-batch metrics through the Flight/client/Flow consumer path.

It builds on #8015, which introduced `RecordBatchMetrics.region_watermarks` and query-side terminal metric collection.

## Changes

- Adds Flight transport support for optional trailing terminal `Metrics` messages.
- Adds client-side `OutputWithMetrics` / `OutputMetrics` helpers so callers can consume terminal metrics without breaking existing `Output` APIs.
- Adds Flow frontend-client support for requesting and reading terminal region watermarks in both standalone and distributed modes.
- Validates Flow query extensions at the Flight boundary before dispatch.
- Keeps normal query/affected-row behavior backward-compatible when terminal metrics are absent.

## Scope

This PR is limited to terminal metrics transport and consumption:

- `src/servers/src/grpc/flight.rs`
- `src/client/src/database.rs`
- `src/flow/src/batching_mode/frontend_client.rs`
- related error/test updates

It intentionally does not include later stale-cursor, incremental-after-seq, benchmark.

## Compatibility

Existing client APIs (`sql`, `query`, `create`, `alter`, etc.) continue to return plain `Output`.
Terminal metrics are opt-in through the new metrics-aware helper path.

Malformed terminal metrics are rejected as transport/parsing errors instead of being silently ignored.

## Tests

- `cargo test -p client terminal_metrics --lib`
- `cargo test -p flow query_with_terminal_metrics --lib`

Coverage includes:

- stream terminal metrics roundtrip
- affected-rows terminal metrics roundtrip
- invalid terminal metrics rejection
- invalid Flow extension rejection
- standalone and distributed Flow consumer paths
## PR Checklist


src/client/src/database.rs
```diff
@@ -57,6 +61,172 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
 
 type DoPutResponseStream = Pin<Box<dyn Stream<Item = Result<DoPutResponse>>>>;
 
+#[derive(Debug, Clone, Default)]
+pub struct OutputMetrics {
+    metrics: Arc<ArcSwapOption<RecordBatchMetrics>>,
```
evenyag: I think RwLock is fine.

src/servers/src/grpc/flight.rs
```diff
@@ -191,6 +195,13 @@ impl FlightCraft for GreptimeRequestHandler {
         let ticket = request.into_inner().ticket;
         let request =
             GreptimeRequest::decode(ticket.as_ref()).context(error::InvalidFlightTicketSnafu)?;
+        let query_ctx =
+            create_query_context(Channel::Grpc, request.header.as_ref(), hints.clone())?;
+        // Validate flow hint syntax at the transport boundary before dispatching the request.
+        // This does not authorize or execute anything; `handle_request()` below still performs
+        // the normal frontend handling and auth checks before query execution.
+        FlowQueryExtensions::parse_flow_extensions(&query_ctx.extensions())
+            .map_err(|e| Status::invalid_argument(e.to_string()))?;
```
evenyag: Should we use output_msg?
```suggestion
            .map_err(|e| Status::invalid_argument(e.output_msg()))?;
```

src/servers/src/grpc/flight.rs
```diff
@@ -538,12 +549,22 @@ fn to_flight_data_stream(
             Box::pin(stream) as _
         }
         OutputData::AffectedRows(rows) => {
-            let stream = tokio_stream::iter(
-                FlightEncoder::default()
-                    .encode(FlightMessage::AffectedRows(rows))
-                    .into_iter()
-                    .map(Ok),
-            );
+            let affected_rows = FlightEncoder::default().encode(FlightMessage::AffectedRows(rows));
```
evenyag: FlightMetadata itself has a metrics field. Can we extend FlightMessage::AffectedRows to carry the metrics so we don't have to encode and send twice?

discord9: > FlightMetadata itself has a metrics field. Can we extend FlightMessage::AffectedRows to carry the metrics so we don't have to encode and send twice?

adda metrics field for affected rows

src/client/src/database.rs
```diff
@@ -57,6 +61,291 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
 
 type DoPutResponseStream = Pin<Box<dyn Stream<Item = Result<DoPutResponse>>>>;
 
+const FLOW_EXTENSIONS_METADATA_KEY: &str = "x-greptime-flow-extensions";
+
+#[derive(Debug, Clone, Default)]
+pub struct OutputMetrics {
+    metrics: Arc<RwLock<Option<RecordBatchMetrics>>>,
+    ready: Arc<AtomicBool>,
```
killme2008: Could we introduce `OutputMetricsInner` ?

```rust
struct OutputMetricsInner {
    metrics: RwLock<Option<RecordBatchMetrics>>,
    ready: AtomicBool,
}

pub struct OutputMetrics {
    inner: Arc<OutputMetricsInner>,
}
```

src/client/src/database.rs
```diff
@@ -57,6 +61,291 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
 
 type DoPutResponseStream = Pin<Box<dyn Stream<Item = Result<DoPutResponse>>>>;
 
+const FLOW_EXTENSIONS_METADATA_KEY: &str = "x-greptime-flow-extensions";
+
+#[derive(Debug, Clone, Default)]
```
killme2008, line 68: Alaways document the pub structs

src/client/src/database.rs
```diff
@@ -57,6 +61,291 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
...
+    fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn update(&self, metrics: Option<RecordBatchMetrics>) {
+        *self.metrics.write().expect("metrics lock poisoned") = metrics;
+    }
+
+    pub fn mark_ready(&self) {
+        self.ready.store(true, Ordering::Release);
```
killme2008: Do we need to use `compare_and_swap` to avoid multiple markings? Or assert it's not in ready state.

src/client/src/database.rs
```diff
@@ -57,6 +61,291 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
...
+                    ensure!(
+                        flight_message_stream.next().await.is_none(),
+                        IllegalFlightMessagesSnafu {
+                            reason: "Expect 'AffectedRows' Flight messages to be followed by at most one Metrics message"
+                        }
+                    );
+                }
+                Some(FlightMessage::Metrics(_)) => {
+                    return IllegalFlightMessagesSnafu {
+                        reason: "'AffectedRows' Flight metadata already carries Metrics and cannot be followed by another Metrics message".to_string(),
```
killme2008:
```suggestion
                        reason: "'AffectedRows' Flight metadata already carries Metrics and cannot be followed by another Metrics message",
```

src/client/src/database.rs
```diff
@@ -333,41 +641,94 @@ impl Database {
         let request = Request::Query(QueryRequest {
             query: Some(Query::Sql(sql.as_ref().to_string())),
         });
-        self.do_get(request, hints).await
+        self.do_get(request, hints, &[])
+            .await
+            .map(OutputWithMetrics::into_output)
+    }
+
+    pub async fn sql_with_terminal_metrics<S>(
```
killme2008, line 672: Document the new pub function.

src/client/src/database.rs
```diff
@@ -57,6 +61,291 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
...
+        self.ready.store(true, Ordering::Release);
+    }
+
+    pub fn is_ready(&self) -> bool {
+        self.ready.load(Ordering::Acquire)
+    }
+
+    pub fn get(&self) -> Option<RecordBatchMetrics> {
+        self.metrics.read().expect("metrics lock poisoned").clone()
+    }
```
evenyag, line 107: unwrap() should be sufficient.
```suggestion
    pub fn update(&self, metrics: Option<RecordBatchMetrics>) {
        *self.metrics.write().unwrap() = metrics;
    }

    pub fn mark_ready(&self) {
        self.ready.store(true, Ordering::Release);
    }

    pub fn is_ready(&self) -> bool {
        self.ready.load(Ordering::Acquire)
    }

    pub fn get(&self) -> Option<RecordBatchMetrics> {
        self.metrics.read().unwrap().clone()
    }
```

src/common/grpc/src/flight.rs
```diff
@@ -41,7 +41,10 @@ use crate::error::{DecodeFlightDataSnafu, InvalidFlightDataSnafu, Result};
 pub enum FlightMessage {
     Schema(SchemaRef),
     RecordBatch(DfRecordBatch),
-    AffectedRows(usize),
+    AffectedRows {
```
killme2008, line 47: Is it a breaking change in protocol?

src/client/src/database.rs
```diff
@@ -57,6 +61,291 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
...
+}
+
+impl Stream for StreamWithMetrics {
+    type Item = common_recordbatch::error::Result<RecordBatch>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let polled = Pin::new(&mut self.stream).poll_next(cx);
+        match &polled {
+            Poll::Ready(Some(_)) => self.sync_terminal_metrics(),
```
evenyag: Do we need to sync_terminal_metrics() every time?

src/datanode/src/region_server.rs
```diff
@@ -749,6 +769,83 @@ impl RegionServer {
...
+    }
+}
+
+impl Stream for RegionWatermarkStream {
+    type Item = common_recordbatch::error::Result<RecordBatch>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        match Pin::new(&mut self.stream).poll_next(cx) {
+            Poll::Ready(None) => {
+                self.finished.store(true, Ordering::Relaxed);
```
killme2008: Since we have `&mut self` here, do we really need an `AtomicBool` for `finished`?

src/client/src/database.rs
```diff
@@ -57,6 +61,291 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
...
+                None => terminal_metrics.mark_ready(),
+                Some(FlightMessage::Metrics(s)) if terminal_metrics.get().is_none() => {
+                    terminal_metrics.update(Some(parse_terminal_metrics(&s)?));
+                    terminal_metrics.mark_ready();
+                    ensure!(
+                        flight_message_stream.next().await.is_none(),
+                        IllegalFlightMessagesSnafu {
+                            reason: "Expect 'AffectedRows' Flight messages to be followed by at most one Metrics message"
+                        }
+                    );
```
evenyag: > Once a stream has finished (returned Ready(None) from poll_next), calling its poll_next method again may panic, block forever, or cause other kinds of problems.

Better not to call next again.

src/flow/src/batching_mode/frontend_client.rs
```diff
@@ -342,6 +343,78 @@ impl FrontendClient {
...
+        match self {
+            FrontendClient::Distributed {
+                query, batch_opts, ..
+            } => {
+                let query_parallelism = query.parallelism.to_string();
+                let hints = vec![
+                    (QUERY_PARALLELISM_HINT, query_parallelism.as_str()),
+                    (READ_PREFERENCE_HINT, batch_opts.read_preference.as_ref()),
+                ];
+                let db = self.get_random_active_frontend(catalog, schema).await?;
```
killme2008: Leave a TODO: we should support more fine-grained load balancing in the future, such as AZ awareness.

src/client/src/database.rs
```diff
@@ -57,6 +61,291 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
...
+                                    metrics_ref.swap(Some(Arc::new(m)));
+                                }
+                                Err(e) => {
+                                    yield Err(BoxedError::new(e)).context(ExternalSnafu);
+                                }
+                            };
+                            break;
+                        }
+                        FlightMessage::AffectedRows { .. } | FlightMessage::Schema(_) => {
+                            yield IllegalFlightMessagesSnafu {reason: format!("A Schema message must be succeeded exclusively by a set of RecordBatch messages, flight_message: {:?}", flight_message)}
```
evenyag: Format the code.

src/query/src/datafusion.rs
```diff
@@ -544,6 +547,9 @@ impl QueryExecutor for DatafusionQueryEngine {
         plan: &Arc<dyn ExecutionPlan>,
     ) -> Result<SendableRecordBatchStream> {
         let explain_verbose = ctx.query_ctx().explain_verbose();
+        let should_collect_region_watermark =
```
killme2008, line 554: Monkey patch almost everywhere, but I’m not sure if there’s a better way.

src/query/src/metrics.rs
```diff
@@ -230,13 +230,52 @@ fn collect_region_watermarks(plan: Arc<dyn ExecutionPlan>) -> Vec<RegionWatermar
...
             );
         }
-
         stack.extend(plan.children().into_iter().cloned());
     }
 
     finalize_region_watermarks(merged)
 }
 
+fn merge_region_watermark_entries(
```
killme2008, line 294: Better to comment the merge logic.

src/client/src/database.rs
```diff
@@ -57,6 +61,291 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
...
+                        FlightMessage::Metrics(s) => {
+                            match parse_terminal_metrics(&s) {
+                                Ok(m) => {
+                                    metrics_ref.swap(Some(Arc::new(m)));
+                                }
+                                Err(e) => {
+                                    yield Err(BoxedError::new(e)).context(ExternalSnafu);
+                                }
+                            };
+                            break;
```
fengjiachun: Is a `Metrics` frame guaranteed to mean the record batch stream is complete? FlightRecordBatchStream can still emit metrics after each record batch when `query_ctx.explain_verbose()` is true, so treating the first metrics frame as terminal here will stop decoding before later record batches. If metrics are not explicitly terminal in the wire protocol, this branch should update the cached metrics and continue reading, or the protocol needs a way to distinguish partial metrics from terminal metrics.

src/client/src/database.rs
```diff
@@ -333,41 +641,94 @@ impl Database {
...
+    }
+
+    pub async fn query_with_terminal_metrics(
+        &self,
+        request: QueryRequest,
+        hints: &[(&str, &str)],
+    ) -> Result<OutputWithMetrics> {
+        self.query_with_terminal_metrics_and_flow_extensions(request, hints, &[])
+            .await
+    }
```
evenyag: Maybe we don't need to add a new method query_with_terminal_metrics(). We can always call query_with_terminal_metrics_and_flow_extensions(). The only difference is passing `&[]`.

src/client/src/database.rs
```diff
@@ -57,6 +61,291 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
 
 type DoPutResponseStream = Pin<Box<dyn Stream<Item = Result<DoPutResponse>>>>;
 
+const FLOW_EXTENSIONS_METADATA_KEY: &str = "x-greptime-flow-extensions";
+
+#[derive(Debug, Clone, Default)]
+pub struct OutputMetrics {
+    metrics: Arc<RwLock<Option<RecordBatchMetrics>>>,
```
fengjiachun: Does this need to be an `RwLock`?

src/common/grpc/src/flight.rs
```diff
@@ -41,7 +41,10 @@ use crate::error::{DecodeFlightDataSnafu, InvalidFlightDataSnafu, Result};
...
+        rows: usize,
+        metrics: Option<String>,
```
evenyag, line 49: Our metrics are strings, right?

src/flow/src/batching_mode/frontend_client.rs
```diff
@@ -342,6 +343,78 @@ impl FrontendClient {
...
+                let database_client = {
+                    database_client
+                        .handler
+                        .lock()
+                        .map_err(|e| {
+                            UnexpectedSnafu {
+                                reason: format!("Failed to lock database client: {e}"),
+                            }
+                            .build()
+                        })?
```
evenyag: Is it too verbose to handle the mutex lock error? We can unwrap.

src/query/src/options.rs
```diff
@@ -177,10 +177,35 @@ impl FlowQueryExtensions {
...
+
+    should_collect_region_watermark(return_region_seq, has_incremental_after_seqs)
+}
+
+fn should_collect_region_watermark(
+    return_region_seq: bool,
+    has_incremental_after_seqs: bool,
+) -> bool {
+    return_region_seq || has_incremental_after_seqs
+}
```
evenyag, line 207: Is it necessary to wrap this into a function?

discord9: > Is it necessary to wrap this into a function?

it's used twice so kind of necessary?

src/servers/src/grpc/flight.rs
```diff
@@ -538,13 +578,60 @@ fn to_flight_data_stream(
...
-                FlightEncoder::default()
-                    .encode(FlightMessage::AffectedRows(rows))
-                    .into_iter()
-                    .map(Ok),
-            );
+            let terminal_metrics = terminal_recordbatch_metrics_from_plan_if_requested(
+                output.meta.plan,
+                should_emit_terminal_metrics,
+            )
+            .and_then(|metrics| serde_json::to_string(&metrics).ok());
```
evenyag: Does it affect the correctness if we return None when we can't serialize the metrics?

evenyag (approved):

src/common/grpc/src/flight.rs
discord9: > Is it a breaking change in protocol?

added common-grpc codec compat test to make sure it doesn't break stuff

killme2008 (approved):
LGTM

PR #8057, opened 2026-04-30
perf(mito-codec): optimize SparseValues decode and lookup

evenyag:

Optimizes `SparsePrimaryKeyCodec::has_column` and `SparseValues` for the common case where a primary key carries only a handful of tag columns. The previous implementation paid two avoidable costs on every primary key match: it eagerly decoded the entire pk into a `HashMap` on the first column lookup, and `SparseValues` itself stored decoded rows in a `HashMap`. Both lose to a small linear scan when the row has just a few tags, which is the dominant shape in practice.

Changes:

- **Lazy, resumable offset cache** (`SparseOffsetsCache`). `has_column` now walks the pk only as far as needed to answer the current query and remembers where it stopped. The next lookup resumes from that cursor instead of re-decoding from the start. Misses on a fully scanned pk are O(1) via a `finished` flag.
- **Fixed offsets for reserved columns**. `table_id` and `tsid` live at known byte positions in the sparse encoding, so `has_column` returns their offsets directly without touching the deserializer.
- **Small-vec fast path for the offset cache**. The cache keeps the first 32 `(column_id, offset)` entries inline in a pre-reserved `Vec` (linear scan, no hashing) and only spills into a `HashMap` beyond that. 32 covers the vast majority of primary keys we see.
- **`SparseValues` switched from `HashMap` to `Vec`**. Lookup is now a linear scan, which is faster for the typical small row and avoids per-decode hashmap allocation. The container is documented as append-only — callers must not insert a duplicate column id, since a later duplicate would shadow the newer value on lookup. `extend` on `CompositeValues` now carries the same contract.
- **Benchmarks**. Added `has_column` benches covering `table_id`, `tsid`, first-tag, and last-tag lookups across 5 / 10 / 50 / 100 tag counts, plus sparse encode/decode benches, so future regressions are visible.

Before
```
~/greptime-main datanode scanbench --config ~/greptime/scanbench/config-k8s.toml --region-id 1024:0 --table-dir data/greptime/k8s/1024/ --parallelism 16 --scan-config ~/greptime/scanbench/scanconfig-etcd-proj.json --path-type data --scanner series --iterations 10

Summary Average: 1983150 rows in 185.103873ms over 10 iterations
```

After
```
~/greptime datanode scanbench --config ~/greptime/scanbench/config-k8s.toml --region-id 1024:0 --table-dir data/greptime/k8s/1024/ --parallelism 16 --scan-config ~/greptime/scanbench/scanconfig-etcd-proj.json --path-type data --scanner series --iterations 10

Summary Average: 1983150 rows in 122.405785ms over 10 iterations
```

## PR Checklist

### DIFF
--- a/src/mito-codec/benches/bench_sparse_encoding.rs
+++ b/src/mito-codec/benches/bench_sparse_encoding.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
 use std::hint::black_box;
 
 use bytes::Bytes;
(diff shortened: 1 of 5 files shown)


evenyag:
@codex review

src/mito-codec/benches/bench_sparse_encoding.rs
```diff
@@ -82,5 +86,424 @@ fn encode_sparse(c: &mut Criterion) {
...
+            "hashmap",
+            &column_ids,
+            last_id,
+            missing_id,
+            || build_sparse_values(&pairs),
+        );
+
+        run_lookup_workloads(&mut group, "vec", &column_ids, last_id, missing_id, || {
+            build_vec_values(&pairs)
+        });
```
v0y4g3r, line 477: These two runs don't seem to compare hashmap-based vs.  vec-based implementations, because `SparseValues` is now already vec-backed.

WenyXu (approved):

src/mito-codec/src/row_converter/sparse.rs
```diff
@@ -111,6 +131,91 @@ pub const RESERVED_COLUMN_ID_TABLE_ID: ColumnId = ReservedColumnId::table_id();
...
+    /// Returns the cached offset for `column_id`, if any.
+    fn get(&self, column_id: ColumnId) -> Option<usize> {
+        for entry in &self.inline {
+            if entry.0 == column_id {
+                return Some(entry.1);
+            }
+        }
+        if self.overflow.is_empty() {
+            return None;
+        }
```
v0y4g3r: Looks like there's no need to check if overflow is empty here, afterall `self.overflow.get` will check that again

v0y4g3r (approved):
Rest LGTM

PR #8060, opened 2026-05-03
fix: revoke meta kv writes outside metasrv leader

### LINKED ISSUE #7585: Revoke KV write access outside metasrv leader

Tech debt reduction, Refactor

## Background
MetaKvBackend currently implements full `KvBackend` write APIs and is used by frontend/flownode/datanode builders. Meta-client store client also exposes write APIs. On the server side, metasrv has a leader-only KV cache and leader-only write semantics. We should align all non-leader/non-metasrv paths to be read-only.

## Evidence
- `src/catalog/src/kvbackend/client.rs` implements full write APIs in `MetaKvBackend`.
- `src/cmd/src/frontend.rs`, `src/cmd/src/flownode.rs`, `src/cmd/src/datanode/builder.rs` construct catalog backends with `MetaKvBackend`.
- `src/meta-client/src/client/store.rs` exposes store `put/batch_put/compare_and_put/delete_range/batch_delete` and is used outside metasrv.
- metasrv has leader-only cache semantics in `src/meta-srv/src/service/store/cached_kv.rs` (`LeaderCachedKvBackend`), assuming only leader writes.

## Problem
Write APIs are exposed to non-leader components via MetaKvBackend and meta-client store. This violates leader-only write assumptions and can conflict with leader cache/consistency semantics. Additionally, metasrv store service should reject write requests from non-leader contexts.

## Proposal
Revoke write permission for non-leader paths at three layers:

1. **Catalog client (MetaKvBackend):**
   - Introduce a read-only wrapper (e.g. `ReadOnlyKvBackend`) that rejects write operations.
   - Use it in frontend/flownode/datanode catalog builders.

2. **Meta-client store:**
   - Provide a read-only store client or a mode flag that disables write calls and returns a clear error.
   - Update call sites to use read-only store client where applicable.

3. **Metasrv store service:**
   - Enforce leader-only writes server-side.
   - Reject write requests when metasrv is not leader (return `IsNotLeader` or a permission error).

## Scope / Files
- `src/catalog/src/kvbackend/client.rs`
- `src/cmd/src/frontend.rs`
- `src/cmd/src/flownode.rs`
- `src/cmd/src/datanode/builder.rs`
- `src/meta-client/src/client/store.rs`
- `src/meta-srv/src/service/store/*` (enforce leader-only write checks)

## Acceptance Criteria
- Non-metasrv components cannot write to meta KV (client-side write attempts fail fast).
- Metasrv store service rejects write requests when not leader.
- Unit/integration tests cover write rejection paths.

## Risks / Notes
- Identify any legitimate write paths from non-metasrv components and migrate them.
- Error types should be consistent with existing error handling (`Error::IsNotLeader` or permission error).

### Implementation challenges

_No response_

QuakeWang:

close: #7585 

This PR revokes direct meta KV write access from non-metasrv leader paths.

  - Adds `ReadOnlyKvBackend`, which forwards read operations and rejects write operations, including write txns.
  - Wraps frontend, datanode, and flownode meta catalog KV backends with the read-only wrapper.
  - Makes meta-client Store writes read-only by default, while keeping an explicit admin/test writable mode.
  - Enforces leader-only Store writes on metasrv and routes accepted writes through `LeaderCachedKvBackend`.
  - Fixes leader-cache state transition so `enable_leader_cache` is updated after leader cache initialization.

## PR Checklist

### DIFF
--- a/src/common/meta/src/key.rs
+++ b/src/common/meta/src/key.rs
@@ -1660,6 +1660,7 @@ mod tests {
     };
     use crate::kv_backend::KvBackend;
     use crate::kv_backend::memory::MemoryKvBackend;
+    use crate::kv_backend::read_only::ReadOnlyKvBackend;
     use crate::peer::Peer;
     use crate::rpc::router::{LeaderState, Region, RegionRoute, region_distribution};
     use crate::rpc::store::{PutRequest, RangeRequest};
@@ -1946,6 +1947,46 @@ mod tests {
             "new-a2"
         );
         assert_eq!(region_routes[0].follower_peers[0].addr, "new-a3");
+    }
+
+    #[tokio::test]
+    async fn test_get_full_table_info_with_read_only_kv_backend() {
+        let mem_kv = Arc::new(MemoryKvBackend::default());
+        let writable_manager = TableMetadataManager::new(mem_kv.clone());
+
+        let region_routes = vec![new_test_region_route()];
+        let table_info = new_test_table_info();
+        let table_id = table_info.ident.table_id;
+
+        create_physical_table_metadata(
+            &writable_manager,
+            table_info.clone(),
+            region_routes.clone(),
+            HashMap::new(),
+        )
+        .await
+        .unwrap();
+
+        let read_only_kv = Arc::new(ReadOnlyKvBackend::new(mem_kv));
+        let read_only_manager = TableMetadataManager::new(read_only_kv);
+
+        let (remote_table_info, remote_table_route) = read_only_manager
+            .get_full_table_info(table_id)
+            .await
+            .unwrap();
+
+        assert_eq!(
+            remote_table_info.unwrap().into_inner().table_info,
+            table_info
+        );
+        assert_eq!(
+            remote_table_route
+                .unwrap()
+                .into_inner()
+                .region_routes()
+                .unwrap(),
+            &region_routes
+        );
     }
 
     #[tokio::test]
--- a/src/meta-client/src/client.rs
+++ b/src/meta-client/src/client.rs
@@ -91,6 +91,7 @@ pub struct MetaClientBuilder {
     role: Role,
     enable_heartbeat: bool,
     enable_store: bool,
+    enable_direct_store_writes: bool,
     enable_procedure: bool,
     enable_access_cluster_info: bool,
     region_follower: Option<RegionFollowerClientRef>,
@@ -143,9 +144,26 @@ impl MetaClientBuilder {
         }
     }
 
+    /// Enables the Store client in read-only mode.
+    ///
+    /// Store write methods fail fast by default. Metadata writes from production
+    /// frontend/datanode/flownode clients should go through metasrv procedures.
     pub fn enable_store(self) -> Self {
         Self {
             enable_store: true,
+            ..self
+        }
+    }
+
+    /// Enables direct Store write RPCs for tests, examples, or controlled admin tooling.
+    ///
+    /// This is not a leader-aware production write client: writable Store requests
+    /// are sent to a randomly selected metasrv peer and may receive `NotLeader`.
+    /// Production metadata writes should use metasrv-owned write paths instead.
+    pub fn enable_direct_store_writes_for_admin(self) -> Self {
+        Self {
+            enable_store: true,
+            enable_direct_store_writes: true,
             ..self
         }
     }
@@ -216,9 +234,13 @@ impl MetaClientBuilder {
         let config = self
             .enable_heartbeat
             .then(|| ConfigClient::new(self.id, self.role, mgr.clone()));
-        let store = self
-            .enable_store
-            .then(|| StoreClient::new(self.id, self.role, mgr.clone()));
+        let store = self.enable_store.then(|| {
+            if self.enable_direct_store_writes {
+                StoreClient::new_writable(self.id, self.role, mgr.clone())
+            } else {
+                StoreClient::new(self.id, self.role, mgr.clone())
+            }
+        });
         let procedure = self.enable_procedure.then(|| {
             let mgr = self.ddl_channel_manager.unwrap_or(mgr.clone());
             ProcedureClient::new(
@@ -983,6 +1005,19 @@ mod tests {
         meta_client.start(urls).await.unwrap();
         let res = meta_client.put(PutRequest::default()).await;
         assert!(matches!(res.err(), Some(error::Error::NotStarted { .. })));
+    }
+
+    #[tokio::test]
+    async fn test_store_writes_are_read_only_by_default() {
+        let meta_client = MetaClientBuilder::new(0, Role::Datanode)
+            .enable_store()
+            .build();
+
+        let res = meta_client.put(PutRequest::default()).await;
+        assert!(matches!(
+            res.err(),
+            Some(error::Error::ReadOnlyKvBackend { .. })
+        ));
     }
 
     #[tokio::test]
--- a/src/meta-client/src/client.rs
+++ b/src/meta-client/src/client.rs
@@ -91,6 +91,7 @@ pub struct MetaClientBuilder {
     role: Role,
     enable_heartbeat: bool,
     enable_store: bool,
+    #[cfg(test)]
     enable_direct_store_writes: bool,
     enable_procedure: bool,
     enable_access_cluster_info: bool,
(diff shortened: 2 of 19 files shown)


QuakeWang:
@WenyXu I have addressed the Gemini and Copilot review comments: removed the unnecessary Txn clone and redacted write request logging. PTAL.

WenyXu:
> @WenyXu I have addressed the Gemini and Copilot review comments: removed the unnecessary Txn clone and redacted write request logging. PTAL.

Nice work. I'm thinking we could separate the KvBackend trait like this:
```
KvBackend: ReadonlyKvBackend + WritableKvBackend
WritableKvBackend: Txn
ReadonlyKvBackend: Txn
```
This might improve the development experience and make the trait boundaries clearer. However, we may need to find a balance between making large code changes and keeping the code well-organized. WDYT?

QuakeWang:
> > @WenyXu I have addressed the Gemini and Copilot review comments: removed the unnecessary Txn clone and redacted write request logging. PTAL.
> 
> Nice work. I'm thinking we could separate the KvBackend trait like this:
> 
> ```
> KvBackend: ReadonlyKvBackend + WritableKvBackend
> WritableKvBackend: Txn
> ReadonlyKvBackend: Txn
> ```
> 
> This might improve the development experience and make the trait boundaries clearer. However, we may need to find a balance between making large code changes and keeping the code well-organized. WDYT?

I agree with the direction. Splitting read-only and writable KV capabilities would make the boundary much clearer and would avoid relying on runtime guards in read-only paths.

However, I think it is better to handle this in a follow-up PR. The tricky part is `TxnService`: the current `Txn` can contain `Get`, `Put`, and `Delete`, while some read paths such as `get_full_table_info()` still use txn for multi-key reads. So a clean split likely needs either a read-only txn abstraction or moving those read paths to explicit read APIs.

For this PR, I’d prefer to keep the scope focused on enforcing the write guard and leader-only Store writes. I can open/follow up with a separate refactor for `ReadonlyKvBackend` / `WritableKvBackend` after this lands.

src/common/meta/src/key.rs
```diff
@@ -1948,6 +1949,46 @@ mod tests {
         assert_eq!(region_routes[0].follower_peers[0].addr, "new-a3");
     }
 
+    #[tokio::test]
+    async fn test_get_full_table_info_with_read_only_kv_backend() {
```
WenyXu, line 1953: question: why add this test?

src/meta-client/src/client/store.rs
```diff
@@ -39,11 +39,30 @@ pub struct Client {
...
     pub fn new(id: Id, role: Role, channel_manager: ChannelManager) -> Self {
+        Self::new_with_read_only(id, role, channel_manager, true)
+    }
+
+    /// Builds a writable direct Store RPC client.
+    ///
+    /// Writable requests are sent to a randomly selected metasrv peer. This is
+    /// intended for tests, examples, or controlled admin tooling, not production
+    /// metadata write paths that require leader-aware routing.
+    pub fn new_writable(id: Id, role: Role, channel_manager: ChannelManager) -> Self {
```
WenyXu: If it’s only used in tests, please add a #[cfg(test)] gate

src/meta-client/src/client.rs
```diff
@@ -143,13 +144,30 @@ impl MetaClientBuilder {
...
             ..self
         }
     }
 
+    /// Enables direct Store write RPCs for tests, examples, or controlled admin tooling.
+    ///
+    /// This is not a leader-aware production write client: writable Store requests
+    /// are sent to a randomly selected metasrv peer and may receive `NotLeader`.
+    /// Production metadata writes should use metasrv-owned write paths instead.
+    pub fn enable_direct_store_writes_for_admin(self) -> Self {
```
WenyXu: ditto

src/meta-srv/src/metasrv.rs
WenyXu, line 1: nit: unrelated changes?

src/meta-srv/src/state.rs
WenyXu, line 1: ditto

src/common/meta/src/key.rs
QuakeWang: This covers a real compatibility path. `get_full_table_info()` reads table info and route through a read-only txn, not plain `get`. Since this PR adds txn validation in `ReadOnlyKvBackend`, this test ensures catalog reads still work when the backend is wrapped as read-only.

src/meta-client/src/client.rs
QuakeWang: This is not test-only by design. It is an explicit escape hatch for tests, examples, and controlled admin tooling. Production default clients still use read-only Store writes through `enable_store()`, and the method name/docs make the non-default writable path explicit.

src/meta-srv/src/metasrv.rs
QuakeWang: This is related to the Store write guard. Store write RPCs need to check whether the leader cache is ready and route accepted writes through `leader_cached_kv_backend()` instead of the raw kv backend, so these accessors are required by the new write gating path.

src/meta-srv/src/state.rs
QuakeWang: This is required for the new leader-cache readiness check. `on_leader_start()` first marks leader cache as disabled, then enables it after cache initialization. The old Leader -> Leader transition kept the old flag, so `become_leader(true)` could fail to update `enable_leader_cache`, causing Store writes to remain rejected.

src/meta-client/src/client/store.rs
QuakeWang: Good point. I narrowed it to `pub(super)` since it is only needed by the parent builder internally. I didn’t add `#[cfg(test)]` because the writable path is also used by the explicit admin/example builder path, not only unit tests.

src/meta-srv/src/service/store.rs
WenyXu, line 1: I’d prefer not to change this file.

src/meta-client/src/client.rs
WenyXu: I think we can add a #[cfg(test)] gate for now, and remove it later if we really need to expose it outside of tests.

QuakeWang:
@WenyXu  Done. I removed the extra request-summary logging changes from this file and kept only the Store write guard path: leader check, leader-cache readiness check, and routing accepted writes through `leader_cached_kv_backend()`.

src/meta-srv/src/service/store.rs
WenyXu: @QuakeWang I think there’s some misunderstanding here. I’d prefer to keep the change minimal and avoid writing data through the leader-cached KvBackend

src/meta-srv/src/metasrv.rs
WenyXu, line 1: I’d prefer not to change this file

src/meta-srv/src/state.rs
WenyXu: I’d prefer not to change this file

QuakeWang:
@WenyXu  Thanks for your review. I’ll remove the leader-cached KvBackend write path and the related `metasrv.rs` / `state.rs` changes.

One thing to confirm: should this PR still keep a minimal follower-side Store write rejection in `src/meta-srv/src/service/store.rs`, while continuing to write through the original `kv_backend()`? Or should we avoid changing `store.rs` entirely and leave server-side Store enforcement for a follow-up?

WenyXu:
> @WenyXu Thanks for your review. I’ll remove the leader-cached KvBackend write path and the related `metasrv.rs` / `state.rs` changes.
> 
> One thing to confirm: should this PR still keep a minimal follower-side Store write rejection in `src/meta-srv/src/service/store.rs`, while continuing to write through the original `kv_backend()`? Or should we avoid changing `store.rs` entirely and leave server-side Store enforcement for a follow-up?

Yes, I prefer not to change the server side. Adding a readonly wrapper on the client side should be enough for now

QuakeWang:
> > @WenyXu Thanks for your review. I’ll remove the leader-cached KvBackend write path and the related `metasrv.rs` / `state.rs` changes.
> > One thing to confirm: should this PR still keep a minimal follower-side Store write rejection in `src/meta-srv/src/service/store.rs`, while continuing to write through the original `kv_backend()`? Or should we avoid changing `store.rs` entirely and leave server-side Store enforcement for a follow-up?
> 
> Yes, I prefer not to change the server side. Adding a readonly wrapper on the client side should be enough for now

@WenyXu   Thanks for clarifying. I reverted the server-side Store changes and kept this PR scoped to the client-side readonly wrapper.

Specifically, Store write RPCs now continue to use the original `kv_backend()` path, and I removed the `leader_cached_kv_backend()` routing, leader/cache readiness checks, and the related `metasrv.rs` / `state.rs` changes.

WenyXu (approved):

MichaelScofield (approved):

PR #8066, opened 2026-05-06
feat: json expr planner

MichaelScofield:

Rewrites `select a.b.c` to `select json_get(a, "b.c")`;

and pushes an "expected data type" argument to the `json_get` function.

## PR Checklist

### DIFF
--- a/src/query/src/datafusion/json_expr_planner.rs
+++ b/src/query/src/datafusion/json_expr_planner.rs
@@ -1,0 +1,286 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::{Arc, LazyLock};
+
+use arrow_schema::Field;
+use arrow_schema::extension::ExtensionType;
+use common_function::scalars::json::json_get::JsonGetWithType;
+use common_function::scalars::udf::create_udf;
+use datafusion_common::arrow::datatypes::DataType;
+use datafusion_common::{Column, DFSchema, Result, ScalarValue, TableReference};
+use datafusion_expr::expr::{BinaryExpr, ScalarFunction};
+use datafusion_expr::planner::{ExprPlanner, PlannerResult, RawBinaryExpr};
+use datafusion_expr::{Expr, ExprSchemable, Operator, ScalarUDF};
+use datatypes::extension::json::JsonExtensionType;
+use either::Either;
+use sqlparser::ast::BinaryOperator;
+
+/// Rewrites JSON-aware SQL expressions into DataFusion expressions.
+///
+/// This planner handles two cases:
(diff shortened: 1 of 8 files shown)


MichaelScofield:
@codex review

sunng87:
sounds good. But will this conflict with fully qualified table name like `catalog.schema.table`?

MichaelScofield:
> sounds good. But will this conflict with fully qualified table name like `catalog.schema.table`?

It won't. DataFusion has clear rules to separate the identifier to generate the arguments of `plan_compound_identifier`. Check this out: 

https://github.com/GreptimeTeam/datafusion/blob/6713439497561fa74a94177e5b8632322fb7cea5/datafusion/sql/src/expr/identifier.rs#L340

src/query/src/datafusion/json_expr_planner.rs
```diff
@@ -0,0 +1,286 @@
...
+use datafusion_expr::{Expr, ExprSchemable, Operator, ScalarUDF};
+use datatypes::extension::json::JsonExtensionType;
+use either::Either;
+use sqlparser::ast::BinaryOperator;
+
+/// Rewrites JSON-aware SQL expressions into DataFusion expressions.
+///
+/// This planner handles two cases:
+/// - Rewrites compound identifiers on JSON extension columns into `json_get` function.
+///   For example, `select a.b.c` => `select json_get(a, "b.c")`.
```
sunng87, line 34: how about `a.b.c::BOOL` => `json_get(a, "b.c", NULL::BOOL)`

MichaelScofield: `a.b.c::BOOL` will be rewritten as `json_get(a, "b.c")::BOOL` in this `JsonExprPlanner`, and then be rewritten again as `json_get(a, "b.c", NULL::BOOL)` in the `JsonGetRewriter`. `::BOOL` is actually `Expr::Cast` in DataFusion, won't go though expr planner here.

sunng87 (approved):
LGTM. Would be helpful to include some sqlness test

fengys1996:
> > sounds good. But will this conflict with fully qualified table name like `catalog.schema.table`?
> 
> It won't. DataFusion has clear rules to separate the identifier to generate the arguments of `plan_compound_identifier`. Check this out:
> 
> https://github.com/GreptimeTeam/datafusion/blob/6713439497561fa74a94177e5b8632322fb7cea5/datafusion/sql/src/expr/identifier.rs#L340

I reproduced a planning failure with JSON2 that appears unrelated to fully-qualified name conflicts, and instead seems to come from how the table provider exposes schema information.

Repro:

```sql
CREATE TABLE json_data (
  a INT,
  b INT,
  data JSON2,
  ts TIMESTAMP TIME INDEX
) WITH ('sst_format'='flat');

EXPLAIN SELECT data.day FROM greptime.public.json_data;
-- Error: Field day not found in struct
```

From DataFusion’s identifier resolution path, `plan_compound_identifier` is only reached after `DFSchema` successfully resolves a base `(qualifier, column)` pair.

In this JSON2 path, the table provider appears to expose the JSON schema as an empty struct (or at least in a form that is not resolvable as expected). As a result, the planner cannot derive:

* the base JSON column
* the nested path

and therefore the rewrite to `json_get(...)` never happens.

So the issue is likely in the table-provider schema construction for JSON2, rather than in the qualified-name splitting logic.

MichaelScofield:
@fengys1996 I've fixed it in https://github.com/GreptimeTeam/greptimedb/pull/8066/commits/e32cdc1569068302feb640b72b8e1fb280f9659a. It's because it got rewritten by the "get field expr planner" first.

fengys1996 (approved):
Great, LGTM

PR #8067, opened 2026-05-06
test: cover standalone user provider config

### LINKED ISSUE #5853: Set --user-provider in config.toml

Couldn't find ability to do so.

### What does the feature do?

Add ability to set command line option that's already present

### Implementation challenges

_No response_

Detachm:
Closes #5853

## What
Add coverage for loading `user_provider` from standalone config files, and document the config usage in the standalone example and generated config docs.

## Why
Standalone already supports propagating `user_provider` to frontend auth setup, but the config-file usage was not covered or visible in examples.

## Tests
- `cargo fmt --check`
- `cargo test -p cmd load_standalone --quiet`


WenyXu:
Should we also add an integration test for this as well?

Detachm:
Added an integration test to cover config-driven standalone `user_provider` auth through the HTTP path.

shuiyisong (approved):
LGTM 🚀

WenyXu (approved):

PR #8078, opened 2026-05-07
fix: sanitize store addrs in logs

### LINKED ISSUE #7525: Remove sensitive information from store_addrs in logs

User experience

The command line arguments `store_addrs` may contain sensitive information.

One of the cmd struct:
https://github.com/GreptimeTeam/greptimedb/blob/6f86a22e6fc22e0271b62679d6454b859a116519/src/cmd/src/metasrv.rs#L151-L152

We should not log them.

### Implementation challenges

Should we distinguish different backends and only escape the sensitive connection string for PostgreSQL/MySQL?

Detachm:
Closes #7525

Sanitizes `store_addrs` before logging kvbackend construction and strengthens connection string sanitization for URL and key-value formats.

The fix keeps useful host/database information in logs while redacting credentials such as URL userinfo and password-like key-value fields.

Validation:
- `cargo test -p common-meta test_sanitize_connection_string --quiet`
- `cargo test -p cli test_sanitize_store_addrs --quiet`
- `cargo test -p cmd test_start_command_debug_sanitizes_store_addrs --quiet`
- `cargo test -p meta-srv test_metasrv_options_debug_sanitizes_store_addrs --quiet`
- `git diff --check`

### DIFF
--- a/src/common/meta/src/kv_backend/util.rs
+++ b/src/common/meta/src/kv_backend/util.rs
@@ -12,46 +12,69 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::sync::LazyLock;
+
+use regex::{Captures, Regex};
+use url::Url;
+
+const REDACTED: &str = "***";
+
+static SENSITIVE_KV_RE: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r#"(?i)(^|\s)(password|pass|pwd|token|secret)\s*=\s*('[^']*'|"[^"]*"|\S+)"#).unwrap()
+});
+
 /// Removes sensitive information like passwords from connection strings.
-///
-/// This function sanitizes connection strings by removing credentials:
-/// - For URL format (mysql://user:password@host:port/db): Removes everything before '@'
-/// - For parameter format (host=localhost password=secret): Removes the password parameter
-/// - For URL format without credentials (mysql://host:port/db): Removes the protocol prefix
-///
-/// # Arguments
-///
-/// * `conn_str` - The connection string to sanitize
-///
-/// # Returns
-///
-/// A sanitized version of the connection string with sensitive information removed
 pub fn sanitize_connection_string(conn_str: &str) -> String {
-    // Case 1: URL format with credentials (mysql://user:password@host:port/db)
-    // Extract everything after the '@' symbol
-    if let Some(at_pos) = conn_str.find('@') {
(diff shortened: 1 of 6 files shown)


Detachm:
Updated the sanitizer to also redact `token` and `secret` in both URL query parameters and key-value connection strings, with test coverage added for both formats.

src/common/meta/src/kv_backend/util.rs
```diff
@@ -12,46 +12,69 @@
...
+                let value = if is_sensitive_key(&key) {
+                    REDACTED.into()
+                } else {
+                    value
+                };
+                (key.into_owned(), value.into_owned())
+            })
+            .collect::<Vec<_>>();
+
+        url.query_pairs_mut().clear().extend_pairs(pairs);
```
MichaelScofield, line 56: Is it possible to redact in place? Like this:

```suggestion
        url
            .query_pairs_mut()
            .for_each_mut(|(key, value)| {
                if is_sensitive_key(&key) {
                    *value = REDACTED.into()
                }
            })
```

MichaelScofield (commented):
Please fix the CI.

src/common/meta/src/kv_backend/util.rs
Detachm: Thanks for the suggestion. I tried the in-place redaction approach, but `url::Url::query_pairs_mut()` returns a serializer and does not expose a mutable iterator over existing query pairs, so `for_each_mut` is not available in the current `url` API.

The current implementation keeps using `url` to parse the query pairs, redacts sensitive values, then clears and writes the pairs back via `extend_pairs`. This avoids hand-rolling query-string parsing while preserving the same redaction behavior.

I also checked the visible failing checks. They seem to come from an older cancelled workflow run, while the latest CI run for the current head commit has passed.

killme2008:
@Detachm The CI needs to be fixed.

src/common/meta/src/kv_backend/util.rs
killme2008: I think it's good. cc @MichaelScofield

fengjiachun:
@Detachm Thanks for the contribution. This PR has been inactive for a while after the CI issue was raised, so I’m going to close it for now to keep the queue manageable.

If you’d like to continue working on it, please feel free to leave a comment after fixing the CI, and we can reopen/review it again.

PR #8079, opened 2026-05-07
fix(mito): queue writes during region edit

v0y4g3r:

This PR fixes the issue that `RegionEdit` requests will render region into `Editing` state, which rejects all insert/bulk insert requests and returns an error immediately.

This PR queues normal writes and bulk inserts that arrive while a region leader is in the transient `Editing` state, instead of rejecting them with a region-state error.

Behavior change: writes and bulk inserts received during an in-flight region edit now wait for that edit to finish. After the edit completion restores the region state, the worker drains those same-region stalled writes before starting the next queued region edit, preserving sequence/version ordering between direct SST edits and concurrent WAL/memtable writes.

It also refactors `WorkerRequest::BulkInserts` to wrap `BulkInsertRequest`, so bulk insert requests can be categorized and buffered during worker receive batching without matching a full `WorkerRequest` inside the buffering helper.

Added regression coverage for both normal writes and bulk inserts arriving during region editing.

## PR Checklist

## Test Plan

- [x] `cargo fmt --all`
- [x] `cargo nextest run -p mito2 edit_region`

### DIFF
--- a/src/mito2/src/worker.rs
+++ b/src/mito2/src/worker.rs
@@ -1021,33 +1021,12 @@ impl<S: LogStore> RegionWorkerLoop<S> {
         general_requests: &mut Vec<WorkerRequest>,
         bulk_requests: &mut Vec<SenderBulkRequest>,
     ) {
+        // Convert bulk insert requests into `bulk_requests` before background notifications run.
+        // This lets an edit-completion notification drain same-batch bulk writes for its region
+        // before it starts the next queued edit.
+        let mut pending_general_requests = Vec::with_capacity(general_requests.len());
         for worker_req in general_requests.drain(..) {
             match worker_req {
-                WorkerRequest::Write(_) | WorkerRequest::Ddl(_) => {
-                    // These requests are categorized into write_requests and ddl_requests.
-                    continue;
-                }
-                WorkerRequest::Background { region_id, notify } => {
-                    // For background notify, we handle it directly.
-                    self.handle_background_notify(region_id, notify).await;
-                }
-                WorkerRequest::SetRegionRoleStateGracefully {
-                    region_id,
-                    region_role_state,
-                    sender,
-                } => {
-                    self.set_role_state_gracefully(region_id, region_role_state, sender)
-                        .await;
-                }
-                WorkerRequest::EditRegion(request) => {
-                    self.handle_region_edit(request);
-                }
-                WorkerRequest::Stop => {
-                    debug_assert!(!self.running.load(Ordering::Relaxed));
-                }
-                WorkerRequest::SyncRegion(req) => {
-                    self.handle_region_sync(req).await;
-                }
                 WorkerRequest::BulkInserts {
                     metadata,
                     request,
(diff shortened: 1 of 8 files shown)


v0y4g3r:
@codex review

src/mito2/src/worker.rs
```diff
@@ -1024,10 +1063,25 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+                            if edit_result.update_region_state
+                    ) {
+                        // The current batch's write buffers may contain writes that arrived while
+                        // this edit was in flight. Classify them before restoring edit progress.
+                        self.handle_buffered_region_write_requests(
+                            &region_id,
+                            write_requests,
+                            bulk_requests,
+                        )
+                        .await;
```
evenyag: Shouldn't we handle the request after we handle the notify?

src/mito2/src/worker.rs
```diff
@@ -935,6 +936,11 @@ impl<S: LogStore> RegionWorkerLoop<S> {
                             match request_with_time.request {
                                 WorkerRequest::Write(sender_req) => write_req_buffer.push(sender_req),
                                 WorkerRequest::Ddl(sender_req) => ddl_req_buffer.push(sender_req),
+                                WorkerRequest::BulkInserts(bulk_insert) => {
+                                    bulk_insert_req_num += 1;
+                                    self.buffer_bulk_insert_request(bulk_insert, &mut bulk_req_buffer)
+                                        .await;
```
evenyag, line 942: Looks like it handles the request directly. But the method name is buffer.

src/mito2/src/worker/handle_write.rs
```diff
@@ -269,10 +333,14 @@ impl<S> RegionWorkerLoop<S> {
...
+                        // before starting the next queued edit.
                         debug!(
-                            "Region {} is altering, add request to pending writes",
-                            region.region_id
+                            "Region {} is {:?}, add request to pending writes",
+                            region.region_id,
+                            region.state()
                         );
                         self.stalling_count.add(1);
                         WRITE_STALL_TOTAL.inc();
```
evenyag, line 346: non-blocking: Should we add a label to the stall metrics so we can distinguish the stall reason?

src/mito2/src/worker/handle_write.rs
```diff
@@ -193,15 +193,79 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+                }
+                .fail(),
+            );
+        }
+    }
+
     /// Handles a specific region's stalled requests.
-    pub(crate) async fn handle_region_stalled_requests(&mut self, region_id: &RegionId) {
+    ///
+    /// `allow_stall` should remain true for ordinary backpressure retry paths. Region-edit
```
evenyag: allow_stall is false for ordinary backpressure retry paths. It is true for non-backpressure retries, like alter and staging.

src/mito2/src/worker/handle_close.rs
```diff
@@ -47,6 +49,18 @@ impl<S: LogStore> RegionWorkerLoop<S> {
                 .memtables
                 .is_empty()
         {
+            if !region.is_flushable() {
+                sender.send(
+                    RegionStateSnafu {
+                        region_id,
+                        state: region.state(),
+                        expect: RegionRoleState::Leader(RegionLeaderState::Writable),
+                    }
```
evenyag: Should we do this check and return the error in handle_flush_request?

src/mito2/src/worker.rs
v0y4g3r: > Looks like it handles the request directly. But the method name is buffer.

`buffer_bulk_insert_request` method simply converts `BulkInsertRequest` into `BulkPart` and push it into `pending_bulk_request`, so I think `buffer` is more precise.

src/mito2/src/worker.rs
v0y4g3r: > Shouldn't we handle the request after we handle the notify?

Here `handle_buffered_region_write_requests` only classifies the pending request in the channel of region worker loop and move the requests corresponding to current editing region into the stall queue instead of handling these writes immediately.

While inside `handle_background_notify`, when edit result is processed, it will try to handle stalled write/bulk write request:

https://github.com/GreptimeTeam/greptimedb/blob/6f06f509b9e953fd281c0dd20498ffb806ece969/src/mito2/src/worker/handle_manifest.rs#L373

evenyag (approved):

WenyXu (approved):

PR #8080, opened 2026-05-08
refactor(mito2)!: remove PartitionTreeMemtable

evenyag:

This PR removes PartitionTreeMemtable.

## PR Checklist


src/metric-engine/src/engine/options.rs
```diff
@@ -66,16 +71,25 @@ pub fn set_data_region_options(
...
     // Set memtable options for the data region.
-    options.insert("memtable.type".to_string(), "partition_tree".to_string());
-    if sparse_primary_key_encoding_if_absent
-        && !options.contains_key(MEMTABLE_PARTITION_TREE_PRIMARY_KEY_ENCODING)
-    {
-        options.insert(
-            MEMTABLE_PARTITION_TREE_PRIMARY_KEY_ENCODING.to_string(),
-            "sparse".to_string(),
-        );
+    options.insert("memtable.type".to_string(), "bulk".to_string());
```
fengjiachun, line 82: Could we also force the data region SST format to `flat` when switching it to bulk/sparse here? `BulkMemtable` can return encoded ranges, and flush writes those ranges through `put_sst()` directly instead of going through `write_sst()`. If a deployment still has `default_flat_format = false`, region creation records `FormatType::PrimaryKey`, but sparse primary key encoding forces `BulkMemtable`; after memtable compaction, the flushed encoded range is already flat-encoded while the region is still described as primary-key format. I think we should either set/reject `sst_format` consistently for this path, or re-encode encoded ranges through the primary-key writer when the target format is not flat.

fengjiachun, line 82: This still leaves the sparse metric data-region path without a flat SST format. `set_data_region_options()` now emits `memtable.type=bulk` and, when enabled, `primary_key_encoding=sparse`, but it never sets `sst_format=flat`. The new flat override in `RegionOptions::try_from_options()` only runs for legacy `memtable.type=partition_tree`, so this `bulk + sparse` path will not trigger it. With `default_flat_format=false`, `RegionOpener` will persist `FormatType::PrimaryKey` while `MemtableBuilderProvider` still forces `BulkMemtable` for sparse PK. That keeps the same encoded-BulkMemtable/primary-key-SST mismatch for newly created metric data regions.

fengjiachun (approved):
Others LGTM

v0y4g3r (approved):

PR #8091, opened 2026-05-09
refactor: store the schema of flat source

MichaelScofield:

Store the schema of the record batches in the `FlatSource` as well, so that we can handle the schema directly before consuming the iter/source.

## PR Checklist

### DIFF
--- a/config/config.md
+++ b/config/config.md
@@ -238,9 +238,6 @@
 | `runtime` | -- | -- | The runtime options. |
 | `runtime.global_rt_size` | Integer | `8` | The number of threads to execute the runtime for global read operations. |
 | `runtime.compact_rt_size` | Integer | `4` | The number of threads to execute the runtime for global write operations. |
-| `heartbeat` | -- | -- | The heartbeat options. |
-| `heartbeat.interval` | String | `18s` | Interval for sending heartbeat messages to the metasrv. |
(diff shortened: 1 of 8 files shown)


MichaelScofield:
@codex review
@gemini-code-assist review

MichaelScofield:
@codex review

evenyag (approved):

waynexia:
what about extending `BoxedRecordBatchStream` or `...Iterator` traits to provide a schema method, similar to how df defines record batch stream. the current change let a constructor to provide schema itself outside of the stream, which may lead to inconsistent.

MichaelScofield:
@waynexia A stream can be derived from an iterator, so I think we can simpy use DataFusion's record batch stream. WDYT @evenyag ?

waynexia:
error type might be a problem

MichaelScofield:
@waynexia Luckily we have our own `SendableRecordBatchStream`, let's use that.

MichaelScofield:
@waynexia I will do it in the following PRs.

evenyag:
> @waynexia Luckily we have our own `SendableRecordBatchStream`, let's use that.

SendableRecordBatchStream uses common_recordbatch::RecordBatch and its error type. I'd prefer to define a new one or still use a struct.

PR #8095, opened 2026-05-12
feat: expose node info for placement selectors

v0y4g3r:

N/A

This PR exposes richer node information to placement selectors so placement decisions can use datanode metadata such as reported environment variables and the originating frontend context.

Main changes:

- Change `PeerDiscovery::active_*` to return `NodeInfo` instead of only `Peer`, while adapting existing call sites that only need peer addresses.
- Add `PeerAllocContext` and thread DDL query-context extensions through table region allocation.
- Propagate the frontend peer address that submitted create-table DDLs via the reserved query-context extension `__greptime_origin_frontend.addr`.
- Extend metasrv/meta-client discovery helpers so active datanode/flownode/frontend lookups preserve `NodeInfo` metadata, including env vars.
- Keep non-table allocation paths such as flow creation, repartition, and failover on default allocation context.

This is intended to support env-based region placement selectors without changing table metadata formats or user-facing SQL syntax.

Validation:

- `cargo check --locked -p common-meta -p meta-client -p meta-srv -p operator -p frontend --lib`

## PR Checklist


src/meta-srv/src/discovery/utils.rs
```diff
@@ -124,6 +124,84 @@ pub async fn alive_frontends(
...
+                NodeStatus::Flownode(status) => NodeWorkloads::Flownode(status.workloads.clone()),
+                _ => return None,
+            };
+
+            if active_filter(now, &node_info) && condition(&workloads) {
+                Some(node_info)
+            } else {
+                None
+            }
+        })
```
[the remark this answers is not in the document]
v0y4g3r: Addressed in bddab9527145aaf4e9d187b94b14bb9de024c13e.

fengjiachun:
@codex review

src/meta-srv/src/discovery.rs
```diff
@@ -44,8 +45,8 @@ impl PeerDiscovery for MetaPeerClient {
     async fn active_datanodes(
         &self,
         filter: Option<for<'a> fn(&'a NodeWorkloads) -> bool>,
-    ) -> Result<Vec<Peer>> {
-        utils::alive_datanodes(
+    ) -> Result<Vec<NodeInfo>> {
+        utils::alive_datanode_infos(
             &DefaultSystemTimer,
             self,
             default_distributed_time_constants().datanode_lease,
```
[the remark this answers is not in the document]
v0y4g3r: Fixed in 13e9dcb7e0d4da944e342f801df6f0b952647525. `active_datanodes` now uses datanode lease timestamps/workloads for liveness and filtering, then returns the corresponding `NodeInfo` so reported env vars are preserved.

src/meta-srv/src/discovery/utils.rs
```diff
@@ -80,45 +81,158 @@ pub async fn alive_datanodes(
...
+                (&mut node_info.status, &lease_value.workloads)
+            {
+                status.workloads = workloads.clone();
+            }
+            Some(node_info)
+        })
+        .collect::<Vec<_>>())
+}
+
+fn datanode_node_info_from_lease(peer: Peer, lease_value: &LeaseValue) -> NodeInfo {
```
WenyXu, line 117: The returned `NodeInfo` is generated from `LeaseValue`, but many of its fields are left empty. Maybe we could return a tuple instead of using the `NodeInfo` struct?

src/operator/src/utils.rs
```diff
@@ -34,6 +34,18 @@ pub fn to_meta_query_context(
     }
 }
 
+pub fn to_meta_query_context_with_origin_frontend(
```
WenyXu, line 37: Question: why do we need to carry the original frontend address?

src/meta-srv/src/discovery/utils.rs
```diff
@@ -80,45 +81,158 @@ pub async fn alive_datanodes(
...
+pub async fn alive_datanode_infos(
+    timer: &impl SystemTimer,
+    accessor: &(impl LeaseValueAccessor + NodeInfoAccessor),
+    active_duration: Duration,
+    condition: Option<fn(&NodeWorkloads) -> bool>,
+) -> Result<Vec<NodeInfo>> {
+    let active_filter = build_active_filter(active_duration);
+    let condition = condition.unwrap_or(|_| true);
+    let lease_values = accessor.lease_values(LeaseValueType::Datanode).await?;
+    let mut node_infos = accessor
```
WenyXu, line 86: Can we use `node_infos` directly instead of querying twice?

src/meta-srv/src/discovery/utils.rs
```diff
@@ -80,45 +81,158 @@ pub async fn alive_datanodes(
...
+/// Returns the alive datanode node infos.
+pub async fn alive_datanode_infos(
+    timer: &impl SystemTimer,
+    accessor: &(impl LeaseValueAccessor + NodeInfoAccessor),
+    active_duration: Duration,
+    condition: Option<fn(&NodeWorkloads) -> bool>,
+) -> Result<Vec<NodeInfo>> {
+    let active_filter = build_active_filter(active_duration);
+    let condition = condition.unwrap_or(|_| true);
+    let lease_values = accessor.lease_values(LeaseValueType::Datanode).await?;
```
WenyXu, line 85: Non-blocking: maybe we can refactor this in the future by removing LeaseKey and LeaseValue, since they duplicate NodeInfoKey/NodeInfoValue. cc @fengjiachun

src/operator/src/utils.rs
v0y4g3r: > Question: why do we need to carry the original frontend address?

To ensure the metasrv distributes regions on the datanodes within the same AZ, we need to known where the create table request orginates from.

src/meta-srv/src/discovery/utils.rs
v0y4g3r: > Can we use `node_infos` directly instead of querying twice?

We need to filter nodes according to lease infos

WenyXu (approved):

PR #8103, opened 2026-05-12
feat: compact json2 data

MichaelScofield:

as title

## PR Checklist


MichaelScofield:
@codex review
@gemini-code-assist review

fengjiachun (commented):
I found a blocker in the compaction path.

1. The JSON alignment in `CompactionSstReaderBuilder::build_flat_sst_reader()` uses `scan_input.mapper.output_schema()`, which is the logical region output schema. However, `SeqScan::build_flat_reader_for_compaction()` returns flat SST batches, including internal columns such as `__primary_key`, `__sequence`, and `__op_type`. For JSON tables, rebuilding the batch with this logical schema can misalign or truncate the flat columns before the SST writer sees them. The alignment should use the actual flat schema, or happen at a boundary that preserves the flat layout.

2. Even after fixing the schema above, the overlapping-SST merge path still looks too late. `FlatMergeReader` builds/interleaves batches using the flat input schema from `FlatProjectionMapper::input_arrow_schema(compaction = true)`, but that schema is still derived from region metadata and does not include the merged JSON concrete types collected here. If two input SSTs have different JSON2 concrete Arrow types and their key ranges overlap, the merge reader can fail before the post-`SeqScan` alignment runs. This needs either pre-merge schema alignment or a compaction input schema that carries the merged JSON types, plus an overlapping JSON2 compaction regression test.

3. `collect_arrow_schemas_from_parquet()` also bypasses the existing compaction missing-file tolerance. The later scan input sets `with_ignore_file_not_found(true)`, but JSON compaction now reads every input file metadata before building that scan input. If one input SST has already disappeared, this new pre-read returns an error instead of preserving the existing skip semantics.

src/mito2/src/compaction.rs
```diff
@@ -995,19 +1007,63 @@ impl CompactionSstReaderBuilder<'_> {
...
+                        continue;
+                    };
+
+                    let json_type = JsonNativeType::try_from(field.data_type())
+                        .context(DataTypeMismatchSnafu)?;
+                    merged.merge(&json_type);
+                }
+            }
+
+            scan_region::concretize_json_types(&mut mapper, &json_type_hint);
```
fengjiachun: This concretizes only the mapper output schema. For overlapping SSTs, `FlatMergeReader` will interleave input batches before the post-`SeqScan` alignment runs, and its compaction input schema still comes from `FlatProjectionMapper::input_arrow_schema(true)` without these merged JSON concrete types. If two input SSTs have different JSON2 concrete Arrow types and overlapping key ranges, the merge can fail before this later alignment. We should align before merge, or carry the merged JSON types into the compaction flat input schema, and add an overlapping JSON2 compaction regression test.

src/mito2/src/compaction.rs
```diff
@@ -1027,6 +1083,38 @@ impl CompactionSstReaderBuilder<'_> {
...
+
+        for file_handle in self.inputs {
+            let file_path =
+                file_handle.file_path(self.sst_layer.table_dir(), self.sst_layer.path_type());
+            let file_size = file_handle.meta_ref().file_size;
+            let parquet_metadata = self
+                .sst_layer
+                .read_sst(file_handle.clone())
+                .cache(CacheStrategy::Compaction(self.cache.clone()))
+                .read_parquet_metadata(
```
fengjiachun, line 1089: This metadata pre-read bypasses the existing compaction missing-file tolerance. The later `ScanInput` sets `with_ignore_file_not_found(true)`, but JSON compaction now opens every input SST here before that scan path can skip object-not-found errors. If an input SST has already disappeared, this new step will fail the compaction instead of preserving the existing skip semantics.

src/mito2/src/compaction.rs
```diff
@@ -995,19 +1007,63 @@ impl CompactionSstReaderBuilder<'_> {
...
-            .await
-            .map(|stream| FlatSource::new_stream(schema.clone(), stream))
+            .await?
+            .and_then({
+                let schema = schema.clone();
+                let has_json_extension_field = schema.has_json_extension_field();
+                move |record_batch| {
+                    future::ready(if has_json_extension_field {
+                        let (_, columns, _) = record_batch.into_parts();
+                        maybe_align_json_array_with_schema(&schema, columns)
```
fengjiachun: This is the logical region output schema, but the stream below is still flat SST layout. Rebuilding the flat batch with this schema can misalign or truncate internal flat columns such as `__primary_key`, `__sequence`, and `__op_type` before the SST writer sees them. The alignment needs to use the actual flat schema, or be applied at a boundary that preserves the flat layout.

evenyag: Shouldn't we do this inside build_flat_reader_for_compaction?

MichaelScofield:
@fengjiachun @evenyag PTAL

MichaelScofield:
@gemini-code-assist review
@codex review
@copilot review

src/datatypes/src/types/struct_type.rs
```diff
@@ -28,22 +28,21 @@ pub struct StructType {
...
+impl From<&Fields> for StructType {
+    fn from(value: &Fields) -> Self {
         let fields = value
             .iter()
             .map(|field| {
-                Ok(StructField::new(
+                StructField::new(
                     field.name().clone(),
-                    ConcreteDataType::try_from(field.data_type())?,
+                    ConcreteDataType::from_arrow_type(field.data_type()),
```
evenyag, line 38: Can we ensure the data type is supported by ConcreteDataType?

MichaelScofield: It's ensured, because the Struct datatype data must be generated from our JSON values, at least for now. I'm planning to use Arrow's `StructField` directly here, then there'll no longer be any conversion in the future.

evenyag (approved):

MichaelScofield:
@copilot resolve the merge conflicts in this pull request

PR #8105, opened 2026-05-13
fix(mito2): drop unsound time-filter cache-key stripping

evenyag:

Drop the unsound time-filter cache-key stripping in the flat-format range cache.

`build_range_cache_key()` used to call `fingerprint.without_time_filters()` whenever the query's `TimestampRange` fully covered the partition's `FileTimeRange`, so different queries that all enclosed the same partition could share a cache entry. The cover check turned out to be too coarse: `extract_time_range_from_expr` can return a range wider than the predicate (e.g. `ts = a OR ts = b` widens to `[min(a, b), max(a, b) + 1)`) and does not analyze AND/OR shapes, so the dropped time predicates would still have filtered rows. This let the cache replay a previous scan's rows that should have been filtered out.

Until we have a sound per-predicate cover check, always carry the full fingerprint (both `time_filters` and `filters`) in the cache key. The `time_filters` / `filters` split and `without_time_filters()` are left in place behind a TODO for when an accurate check is available.

- `src/mito2/src/read/range_cache.rs`: always use the full fingerprint; remove the now-dead `query_time_range_covers_partition_range` helper; update the two affected unit tests to assert filters are preserved.
- `src/mito2/src/read/scan_region.rs`: keep the `time_filters` / `filters` split with a TODO explaining it is currently inert.
- `src/mito2/src/engine/scan_test.rs`: add a regression test where two `ts = a OR ts = b` scans with different `a` both cover the partition; the second scan must not replay the first scan's cached row.

## PR Checklist

### DIFF
--- a/src/mito2/src/engine/scan_test.rs
+++ b/src/mito2/src/engine/scan_test.rs
@@ -15,6 +15,7 @@
 use std::collections::BTreeMap;
 
 use api::v1::Rows;
+use common_base::readable_size::ReadableSize;
 use common_error::ext::ErrorExt;
 use common_error::status_code::StatusCode;
 use common_recordbatch::RecordBatches;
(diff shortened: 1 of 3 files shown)


v0y4g3r (approved):
LGTM

PR #8109, opened 2026-05-13
fix: remap batch table route addresses

WenyXu:

This PR fixes stale peer addresses returned from batch table route reads.
`TableRouteStorage::batch_get()` previously decoded table routes directly from KV and skipped address remapping, while single-route reads and `batch_get_with_raw_bytes()` already remapped peer addresses through the latest `NodeAddress` records. As a result, callers such as `information_schema.region_peers` could expose outdated datanode peer addresses.
The fix makes `batch_get()` reuse `batch_get_with_raw_bytes()`, so batch route reads share the same remap path as other table route read APIs. A regression test is added to verify both leader and follower peer addresses are remapped.

## PR Checklist


fengjiachun (approved):
LGTM

PR #8115, opened 2026-05-14
fix: avoid stale route update during repartition allocation

WenyXu:

This PR fixes a stale table-route update issue in the repartition allocation procedure.

During `AllocateRegion::ExecutePlan`, the procedure now refreshes the table route after acquiring the table write lock, and builds the new route list from that refreshed value. This ensures both the CAS old value and the generated new value are based on the latest table-route metadata.

Without this, a concurrent repartition-related metadata update could be lost: even if CAS used a refreshed `table_route_value`, the new route value could still be generated from an older `region_routes` snapshot.

This PR also adds a regression test that simulates a table-route update during lock acquisition and verifies the allocation path preserves that concurrent route while appending the newly allocated region.

## PR Checklist


WenyXu:
@codex review

evenyag (approved):

PR #8123, opened 2026-05-16
feat: look up cache with range calculation

waynexia:

Use cached content with range calculation instead of precise key matching to improve cache efficiency. Test scenario:

```sql
SELECT pod, container, avg(greptime_value) AS avg_v
FROM container_cpu_usage_seconds_total
WHERE greptime_timestamp >= '2025-10-20 00:00:00'
  AND greptime_timestamp <  '2025-10-21 00:00:00'
  AND namespace = 'kube-system'
GROUP BY pod, container
ORDER BY avg_v DESC
LIMIT 20;
```

Changes:

| Version | page hits | page misses | cache bytes added |
  |---|---:|---:|---:|
  | before | 0 | +63 | +132,136,272 |
  | after | +32 | +63 | +24,335,439 | 

- query gets 32 page-cache hits where old gets 0, meaning fragment reuse is happening inside the same SQL workload.
- The cache footprint is much smaller: about 24MB added vs 132MB added for the same query.

## PR Checklist


waynexia:
@codex review

waynexia:
@codex review

src/mito2/src/cache.rs
```diff
@@ -1131,73 +1150,298 @@ impl SstMetaKey {
...
+        for key in keys {
+            if self.cache.contains_key(&key) {
+                continue;
+            }
+
+            if self.remove_index_entry(key) && self.cache.contains_key(&key) {
+                self.insert_index_entry(key);
+            }
+        }
+    }
```
evenyag: remove_stale_index_entries() is invoked from the eviction listener and calls self.cache.contains_key(&key). Is there a potential deadlock issue?

src/mito2/src/cache.rs
```diff
@@ -1371,7 +1357,11 @@ impl PageRangeCache {
             let size = page_cache_weight(&key, &bytes);
             CACHE_BYTES.with_label_values(&[PAGE_TYPE]).add(size.into());
             self.cache.insert(key, bytes);
-            self.insert_index_entry(key);
+            let mut index = self.index.write().unwrap();
+            index
+                .entry(key.group_key())
+                .or_default()
+                .insert((key.start, key.end), key);
```
evenyag, line 1554: Should we insert the index before the cache? So we can ensure the key is in the index when the cache evicts items.

evenyag (approved):

killme2008:
@waynexia There is a conflict need to be fixed.

src/mito2/src/cache.rs
waynexia: this seems acceptable, otherwise the state would become more complex

PR #8124, opened 2026-05-18
feat: support pending flow metadata with defer_on_missing_source

discord9:
## Summary
- Add `defer_on_missing_source` flow option for metadata-only pending flow creation when source tables are missing
- Preserve `defer_on_missing_source` in metadata for `SHOW CREATE FLOW` / `information_schema.flows`, while stripping it from runtime/flownode `CreateRequest`
- Allow pending flows to be dropped safely even when they have no flownode routes

## Scope
This draft is intentionally reduced to the metadata-only MVP after complexity review.

Included:
- flow option validation: only `defer_on_missing_source` is supported
- missing source without defer => clear error
- missing source with defer => pending flow metadata
- `FlowStatus::PendingSources`, source names, unresolved source names
- batching semantics for missing-source pending flows
- reject cross-state pending<->active `OR REPLACE` transitions for now
- metadata/round-trip/drop sqlness coverage

Deferred follow-ups:
- automatic pending-flow activation/reconcile
- activation procedure and failure persistence
- cross-state pending<->active replacement behavior
- distributed activation lifecycle sqlness

## Tests
- `cargo test -p operator determine_flow_type_for_source_state`
- `cargo test -p common-meta create_flow`
- `cargo test -p common-meta drop_flow`
- `cargo sqlness bare --test-filter flow_pending --bins-dir /mnt/nvme_rust/rust-targets/pending_defer/debug`


src/common/meta/src/ddl/create_flow.rs
```diff
@@ -365,6 +409,60 @@ pub fn get_flow_type_from_options(flow_task: &CreateFlowTask) -> Result<FlowType
     }
 }
```
killme2008, line 411:
```suggestion
// The metadata key for a flow with non-existent source tables
```

src/common/meta/src/key/flow/flow_info.rs
```diff
@@ -27,12 +29,24 @@ use crate::FlownodeId;
...
 use crate::kv_backend::KvBackendRef;
 use crate::kv_backend::txn::{Compare, CompareOp, Txn, TxnOp};
+use crate::range_stream::{DEFAULT_PAGE_SIZE, PaginationStream};
+use crate::rpc::KeyValue;
+use crate::rpc::store::RangeRequest;
 
 pub const FLOW_INFO_KEY_PREFIX: &str = "info";
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub enum FlowStatus {
```
killme2008, line 45: Better to doc

tests/cases/standalone/common/flow/flow_pending.sql
```diff
@@ -0,0 +1,24 @@
...
+
+SELECT
+    flow_definition,
+    source_table_ids,
+    source_table_names,
+    flownode_ids,
+    options LIKE '%"defer_on_missing_source":"true"%' AS has_defer_option,
+    options LIKE '%"flow_type":"batching"%' AS has_flow_type_option
+FROM INFORMATION_SCHEMA.FLOWS
+WHERE flow_name = 'pending_with_defer';
```
killme2008, line 20: Add a test case where the table does not exist when creating the flow, then create the table afterward and verify that the flow works correctly.

killme2008 (approved):
LGTM almostly

killme2008:
@discord9 Please take a look at comments.

src/operator/src/statement/ddl.rs
```diff
@@ -204,6 +204,39 @@ fn validate_and_normalize_flow_options(
...
+                    "missing source tables for flow '{}'; use WITH ({DEFER_ON_MISSING_SOURCE_KEY} = true) to create a pending flow",
+                    flow_name
+                )
+            }
+        );
+        info!(
+            "Flow `{}` defaults to batching because defer_on_missing_source=true and some source tables are not available yet",
+            flow_name
+        );
+        return Ok(Some(FlowType::Batching));
```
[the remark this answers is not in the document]
discord9: Thanks for the callout. This is intentional for the reduced metadata-only MVP scope now.

Automatic activation/reconcile has been removed from this PR, so there is no activation-time re-evaluation in the current change. For missing-source flows we currently persist a deterministic batching-pending metadata state, and keep `flow_type` as an internal option rather than exposing it through user `WITH` options.

Streaming support for pending flows should be handled together with the future activation lifecycle, where we can either re-evaluate the flow type after sources resolve or add an explicit supported override with the right validation. I also updated the log wording to make this less ambiguous.

tests/cases/standalone/common/flow/flow_pending.sql
discord9: This test would require the automatic activation/reconcile lifecycle, which is intentionally deferred from this PR.

The current PR is scoped to a metadata-only MVP: it allows creating pending flow metadata when sources are missing, preserves the option for `SHOW CREATE FLOW` / `information_schema.flows`, and supports dropping the pending flow. It does not yet activate the flow after the source table is later created, so a “create table afterward and verify the flow works” test would either fail today or force activation/reconcile back into this PR.

I think that scenario is the right acceptance test for the follow-up activation PR. For this PR, `flow_pending.sql` keeps coverage limited to metadata / round-trip / drop behavior.

killme2008:
@discord9 There is a test failure.

PR #8130, opened 2026-05-18
feat: initial implementation for range cache with time filters

evenyag:

Revives the range-cache optimization that strips time-only predicates from the cache key when the query covers a partition's file-time range, so that different queries with different-but-equally-covering time bounds can share a cache entry. The previous version was disabled because its cover check (`extract_time_range_from_expr` + a `TimestampRange` containment test) was not a sound implication check and could let the cache serve rows that should have been filtered out.

- Adds `implied_time_range_from_exprs` in `read::range_cache`, a per-expression walker over time-only `Expr` nodes that returns a `TimestampRange` in which every predicate is provably true. It handles `=`, `<`, `<=`, `>`, `>=`, `BETWEEN`, and AND-composition; lower bounds round up and upper bounds round down on unit conversion so containment of the partition's `FileTimeRange` is a sound implication. It bails (returns `None`) on unsupported shapes — `OR`, `NOT`, `NOT BETWEEN`, `IN`, non-literal RHS, column-name mismatch, string-timestamp literals, `=` literals not exactly representable in the column unit, or overflow.
- `build_range_cache_key` now calls `without_time_filters()` only when the scan's implied range contains both the partition's `file_min` and `file_max`; otherwise it keeps the full fingerprint. Asserts that the file timestamps and the implied range share the time index column's `TimeUnit`.
- `build_scan_fingerprint` returns a new `ScanFingerprintBundle` (`{ fingerprint, implied_time_range }`). `StreamContext` carries `scan_implied_time_range` alongside `scan_fingerprint`. Time-only exprs that the legacy extractor recognizes are routed into the implication walker; unrecognized ones stay in `filters` and are never stripped.
- `table::predicate` exports `is_string_timestamp_literal`, which now covers `Utf8`, `LargeUtf8`, and `Utf8View`. The existing `return_none_if_utf8!` macro and the new walker both use it to reject string-timestamp literals consistently.
- Replaces the prior "preserve filters" tests with parameterized cases for stripping vs. preserving time filters, two distinct queries sharing the same cache key when both cover a partition, and `OR` disabling the optimization. Adds unit tests for `implied_time_range_from_exprs` covering supported shapes, unsupported shapes, and cross-unit literal conversion.

## PR Checklist

### DIFF
--- a/src/mito2/src/read/range_cache.rs
+++ b/src/mito2/src/read/range_cache.rs
@@ -19,13 +19,20 @@ use std::sync::Arc;
 
 use async_stream::try_stream;
 use common_telemetry::warn;
+use common_time::Timestamp;
+use common_time::range::TimestampRange;
+use common_time::timestamp::TimeUnit;
+use datafusion_expr::expr::Expr;
+use datafusion_expr::{Between, BinaryExpr, Operator};
 use datatypes::arrow::compute::concat_batches;
 use datatypes::arrow::record_batch::RecordBatch;
 use datatypes::prelude::ConcreteDataType;
+use datatypes::value::scalar_value_to_timestamp;
 use futures::TryStreamExt;
 use snafu::ResultExt;
 use store_api::region_engine::PartitionRange;
 use store_api::storage::{FileId, RegionId, TimeSeriesRowSelector};
+use table::predicate::is_string_timestamp_literal;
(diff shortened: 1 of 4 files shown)


v0y4g3r:
@codex review

v0y4g3r (approved):

PR #8143, opened 2026-05-20
fix: skip flush when closing follower region

v0y4g3r:

This PR fixes closing a region that is already in follower state when the region uses Noop WAL and still has pending memtable data.

- Skip the close-time pending-data flush unless the region is flushable, so follower regions can be removed without attempting an invalid flush.
- Return an explicit `RegionNotReady` error when manual flush requests target regions that are not in a flushable state.
- Add a regression test covering a Noop WAL follower region with pending memtable data.

## PR Checklist


v0y4g3r:
@codex review

WenyXu (approved):

PR #8145, opened 2026-05-21
feat: flush hook extension point

sunng87:

This patch adds extension points and plugin resgistration for flush/compaction.

## PR Checklist


sunng87:
@evenyag please check if this hook has covered all paths that with generate/update our parquet and manifests.

sunng87:
@v0y4g3r will this work with our direct bulk ingestion?

evenyag:
Took a closer look at which write paths this hook actually fires on, and I think the coverage is narrower than the "flush/compaction extension point" framing suggests. Sharing what I found in case it matters for the use case.

**Where the hook fires today:** flush, and local compaction (`CompactionTaskImpl`).

**Where it doesn't fire, but we still write parquet or touch the manifest:**
- **Remote compaction** — `open_compaction_region` sets `plugins: Plugins::new()`, so the standalone/remote compactor path produces SSTs without firing the hook.
- **Async index build** (`sst/index.rs`) — writes index files and commits a manifest edit.
- **Region copy / import** (`handle_copy_region.rs`) — copies parquet into a region and commits them.
- **Apply staging** (`handle_apply_staging.rs`) — publishes staged SSTs into the live manifest.
- **Generic `edit_region`** (`handle_manifest.rs`, Editing state) — external edits (migration catch-up, metric engine, etc.) that can add files.
- Plus manifest-only mutations: truncate, alter, GC removals, partition-expr change.

The reason most of these slip through is that the hooks are wired in at `RegionFlushTask` / `CompactionTaskImpl`, so every writer that takes a different entry point has to remember to call them — and several already don't.

Two options depending on intent:
1. If you genuinely only care about flush + local compaction output, that's fine — I'd just suggest documenting the hook as flush/local-compaction-only so nobody assumes it's a general "any SST/manifest change" hook.
2. If you want full coverage, the cleaner choke point is `ManifestContext::update_manifest` (every manifest mutation funnels through it) for `on_manifest_updated`, and `AccessLayer::write_sst` for `on_sst_files_written`. Hooking there covers all the paths above in one spot, and as a bonus you'd get the manifest version directly instead of re-reading `last_version()` afterward (which is slightly racy in the compaction path).

v0y4g3r:
> @v0y4g3r will this work with our direct bulk ingestion?

No. Bulk ingestion bypasses flush on datanode, instead, SSTs are written from frontend and then it will send a `RegionEditRequest` to datanode. So you will also need to hook `handle_region_edit`:

https://github.com/GreptimeTeam/greptimedb/blob/4668dd43bdddba47978942772d59cf22008a1e31/src/mito2/src/worker/handle_manifest.rs#L285

src/mito2/src/engine/flush_hook.rs
```diff
@@ -0,0 +1,70 @@
...
+/// ```ignore
+/// use std::sync::Arc;
+/// use common_base::Plugins;
+/// use mito2::engine::flush_hook::{FlushHook, FlushHookRef};
+///
+/// plugins.insert(Arc::new(MyHook) as FlushHookRef);
+/// ```
+#[async_trait]
+pub trait FlushHook: Send + Sync {
+    /// Called after SST files are written during flush.
```
fengjiachun: Could we make the scope explicit here? The hook currently only fires from `RegionFlushTask` and local `CompactionTaskImpl`; paths like direct `RegionEditRequest`/bulk ingestion, copy-region, apply-staging, async index builds, and remote/standalone compaction still update SSTs/manifests without invoking it. If the intended contract is broader than flush/local compaction, this likely needs to move to `AccessLayer::write_sst` / manifest update choke points; otherwise the docs should say the hook is scoped to these two paths.

src/mito2/src/compaction/task.rs
```diff
@@ -282,6 +283,42 @@ impl CompactionTaskImpl {
...
+                &files,
+            )
+            .await;
+        }
+    }
+
+    async fn invoke_manifest_hook(&self, edit: &RegionEdit) {
+        let hook: Option<FlushHookRef> = self.compaction_region.plugins.get();
+        if let Some(hook) = hook {
+            let manifest_version = self
```
fengjiachun: Can we pass through the `ManifestVersion` returned by `update_manifest_for_compaction()` instead of re-reading `last_version()` here? This second read can observe a later manifest update and report the wrong version for this `RegionEdit`, while the update API already returns the exact version.

src/mito2/src/region.rs
```diff
@@ -1185,6 +1202,13 @@ impl ManifestContext {
             );
         }
 
+        if let Some(hook) = hook
+            && let Some(action_list) = action_list_for_hook
+        {
+            hook.on_manifest_updated(manifest.metadata.region_id, &action_list, version)
```
fengjiachun: Could we avoid awaiting the hook while holding the manifest write lock? `manager` is still in scope here after `manager.update(...)`, so any hook implementation that reads the manifest or sends a region request that needs this lock can deadlock, and slow hooks will block all manifest updates. We can capture `region_id`, `version`, and the cloned action list, drop `manager`, then invoke the hook.

src/mito2/src/engine/region_hook.rs
```diff
@@ -0,0 +1,132 @@
...
+//! plugins.insert(Arc::new(MyHook) as RegionHookRef);
+//! ```
+//!
+//! ## Coverage
+//!
+//! | Scenario                     | `on_sst_files_written` | `on_manifest_updated` |
+//! |------------------------------|:----------------------:|:---------------------:|
+//! | Flush (memtable → SST)       | ✅ Yes                 | ✅ Yes                |
+//! | Local compaction             | ✅ Yes                 | ✅ Yes                |
+//! | Remote compaction            | ✅ (on compactor node)  | ✅ (on compactor node) |
```
fengjiachun: Is remote compaction actually covered here? The standalone `open_compaction_region()` path builds `ManifestContext::new(..., None)` and `plugins: Plugins::new()`, so there is no hook on that compactor node. Also `MergeOutput::sst_infos` is `#[serde(skip)]`, so the requester side cannot reconstruct the SST hook payload either. If remote compaction is out of scope, this row should say so; otherwise the hook/plugins and payload need to be wired through.

src/mito2/src/engine/region_hook.rs
```diff
@@ -0,0 +1,132 @@
...
+//! | RegionEdit / bulk ingestion  | ❌ (files pre-written)  | ✅ Yes                |
+//! | Copy region                  | ❌ (object-store copy)  | ✅ Yes                |
+//! | Apply staging                | ❌ (delegates to edit)  | ✅ Yes                |
```
fengjiachun: Apply staging still seems only partially covered. The edit request path can trigger `on_manifest_updated`, but after it succeeds `exit_staging_on_success()` writes the merged staged actions with `manager.update(merged_actions, false)` directly, bypassing `ManifestContext::update_manifest_with_state_check()`. That means the final partition-expression/merged manifest update is invisible to the hook despite this coverage table saying yes.

src/mito2/src/region.rs
```diff
@@ -851,14 +852,29 @@ impl MitoRegion {
...
         info!(
             "Successfully submitted merged staged manifests for region {}, new version: {}",
             self.region_id, new_version
         );
 
+        // Invoke the on_manifest_updated hook for the merged staging actions.
+        if let Some(hook) = hook
+            && let Some(action_list) = action_list_for_hook
+        {
+            hook.on_manifest_updated(self.region_id, &action_list, new_version)
```
fengjiachun: [P1] This still awaits the region hook while holding the manifest write lock. The apply-staging caller acquires `manifest_manager.write()` and passes that guard into `exit_staging_on_success`, and this method continues to use the same guard after the hook for staging cleanup. If a hook implementation reads the manifest, or calls back into a path that needs the manifest lock, apply staging can deadlock here. The "staging exit is infrequent" argument only bounds latency; it does not address the lock re-entry risk. Please move the hook notification outside the manifest write-lock scope, or return the cloned action list/version to the caller so it can notify after dropping the guard.

src/mito2/src/region.rs
```diff
@@ -573,6 +578,14 @@ impl MitoRegion {
 
         drop(manager);
 
+        // Invoke the hook outside the manifest write lock to avoid deadlock.
+        if let Some((action_list, version)) = hook_payload
```
fengjiachun: [P2] This hook payload only covers the staging-merge update, but the backfilled-metadata path just above still performs a real manifest update without notifying `on_manifest_updated`. The PR documents the manifest hook as firing after any manifest write, yet that direct `manager.update(...)` neither goes through `ManifestContext::update_manifest_with_state_check` nor contributes to this `hook_payload`. If the compatibility path runs, hook consumers miss the `RegionChange`; if it runs right after staging exit, they may only receive the earlier staging-merge version while the manifest has already advanced again. Please either add the backfill update to this lock-free hook payload flow as well, or route it through the same manifest-update helper used by the other hooked writes.

src/mito2/src/compaction/compactor.rs
```diff
@@ -250,11 +260,13 @@ impl CompactionRegion {
...
 
 /// `[MergeOutput]` represents the output of merging SST files.
-#[derive(Default, Clone, Debug, Serialize, Deserialize)]
+#[derive(Default, Debug, Serialize, Deserialize)]
 pub struct MergeOutput {
     pub files_to_add: Vec<FileMeta>,
     pub files_to_remove: Vec<FileMeta>,
     pub compaction_time_window: Option<i64>,
+    #[serde(skip)]
+    pub sst_infos: Vec<SstInfo>,
```
evenyag, line 269: Will it be an issue if we do the compaction remotely?

src/mito2/src/worker/handle_copy_region.rs
```diff
@@ -228,13 +230,13 @@ impl<S> RegionWorkerLoop<S> {
...
         let action_list = RegionMetaActionList::with_action(RegionMetaAction::Edit(edit.clone()));
-        info!("Applying {edit:?} to region {target_region_id}, reason: CopyRegionFrom");
+        info!(
+            "Applying {:?} to region {target_region_id}, reason: CopyRegionFrom",
+            edit
+        );
         let version = region
             .manifest_ctx
-            .manifest_manager
-            .write()
```
evenyag, line 235: @WenyXu Can we ensure the state is Writable here?

WenyXu: yes

fengjiachun (approved):
Rest LGTM

src/mito2/src/compaction/compactor.rs
sunng87: this piece of data is only for hook, and remote compaction calls hook in its own process so we don't need to carry this piece of data on wire.

sunng87:
Let me get this merge first and there may be further refactoring for this part.

PR #8151, opened 2026-05-21
test: verify KILL cancels INSERT SELECT

### LINKED ISSUE #6334: Currently `KILL` cannot terminate queries like `INSERT INTO SELECT`

QuakeWang:

Closes #6334

This PR adds tests to verify that tracked `INSERT INTO ... SELECT` statements can be cancelled through the process manager.

The tests cover both supported kill forms:

  - `KILL QUERY <connection_id>`
  - `KILL '<frontend>/<process_id>'`

The test uses a pending source table scan to keep `INSERT INTO target SELECT * FROM source` running, then verifies that `KILL` makes the original insert task return `Cancelled` instead of only disappearing from `SHOW PROCESSLIST`.

This does not change runtime logic. It only strengthens coverage for the existing insert-select process tracking and cancellation path.

## PR Checklist

### DIFF
--- a/src/frontend/src/instance.rs
+++ b/src/frontend/src/instance.rs
@@ -1215,7 +1215,7 @@ mod tests {
     use api::v1::meta::{ProcedureDetailResponse, ReconcileRequest, ReconcileResponse};
     use catalog::process_manager::ProcessManager;
     use common_base::Plugins;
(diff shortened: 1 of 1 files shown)


killme2008 (approved):
Good job

Issue #8125, opened 2026-05-18
tracking: flow incremental read


discord9:
## Summary

Support incremental read in the flow engine — when a flow runs, process only newly appended source data since the last safe checkpoint instead of reprocessing the entire source table.

The core plumbing and execution path have landed, but the feature is still guarded as an experimental optimization:

- incremental read is default-off and requires explicit per-flow opt-in;
- currently limited to supported SQL aggregate shapes over append-only source tables;
- correctness-sensitive fallback / transition handling remains tracked as follow-up work.

Remote dynamic filter work is related scan-pruning infrastructure, but its detailed rollout is tracked separately in #8127.

## Related PRs

- [x] #7848 — feat: add incremental read context and scan boundaries
- [x] #7879 — feat(flow): inc query scan bind seq
- [x] #7979 — feat: remote dyn filter basics
- [x] #8015 — feat: add seq watermark in record batch metrics
- [x] #8045 — feat: flow inc query terminal metrics transport
- [x] #8108 — feat: inc query join rewrite helper
- [x] #8154 — feat: add flow query-context plumbing for terminal watermarks
- [x] #8179 — feat(flow): support incremental read checkpoints
- [x] #8196 — fix(flow): harden incremental read correctness
- [x] #8242 — fix(flow): support complex SQL full-query flows
- [ ] #8248 — fix: restore flow dirty signal after fallback (draft)
- [ ] *full -> incremental transition safety follow-up* (WIP / link when submitted)

Related tracking:

- #8127 — remote dynamic filter tracking

## Component Breakdown

| Component | Description | Status |
|-----------|-------------|--------|
| Scan boundaries / context | Defines `IncrementalReadContext` and scan range boundaries so the flow knows what data to scan | ✅ #7848 |
| Scan bind sequence | Binds scan sequence numbers to flow queries for incremental processing | ✅ #7879 |
| Seq watermark metrics | Records sequence watermark in record batch metrics for progress tracking | ✅ #8015 |
| Terminal metrics transport | Transports flow incremental query terminal metrics through query execution paths | ✅ #8045, #8154 |
| Join rewrite helper | Rewrites aggregate delta queries to LEFT JOIN with sink table for incremental merge | ✅ #8108 |
| Flownode checkpoint / execution path | Maintains full-snapshot vs incremental checkpoint state, gates incremental mode, and sends incremental query extensions | ✅ #8179 |
[...]

PR #8154, opened 2026-05-22
feat: add flow query-context plumbing for terminal watermarks

discord9:

Refs #8125

This PR narrows PR3a to **Flow query-context / terminal-watermark plumbing only**. It prepares the lower-level paths needed by a follow-up PR to enable incremental Flow reads, but it does **not** activate those paths from normal Flow batching execution.

Core plumbing:

- Propagate `QueryContext` through frontend Substrait decode and dummy catalog table lookup.
- Add `FrontendClient::query_with_terminal_metrics(...)` so a future Flow path can forward flow extensions and collect terminal region watermark metrics.
- Keep datanode/query scan plumbing able to expose terminal region watermark metadata when explicitly requested through query context extensions.
- Add low-level tests for terminal-watermark source scan semantics and sink-scan exclusion.

Reviewability / independent correctness cleanup:

- Split batching task table-creation helpers and task tests into focused modules.
- Make batching task replace/remove lifecycle safer.
- Restore dirty-window state correctly when scoped or filterless query planning/execution fails.
- Return Flow batching affected-row counts as `usize` to avoid narrowing `OutputRows`.

Scope boundaries:

- Normal Flow batching execution still uses `FrontendClient::handle(...)`.
- Flow tasks do not emit `FLOW_RETURN_REGION_SEQ`, `FLOW_SINK_TABLE_ID`, `FLOW_INCREMENTAL_MODE`, or `FLOW_INCREMENTAL_AFTER_SEQS` in this PR.
- `TaskState` does not maintain checkpoint maps or checkpoint mode.
- Checkpoint state machine, incremental extension gating, aggregate rewrite wiring, sink dirty-window filtering, stale cursor handling, and standalone SQL incremental-read behavior are left to follow-up changes.

Validation:

- `cargo fmt --all --check`
- `cargo test -p flow --lib` (130 passed)
- `cargo test -p query terminal_watermark_context --lib`
- `cargo check -p flow -p frontend -p datanode -p catalog -p query`

Note: `.cargo/config.toml` was modified locally by build tooling and is intentionally not included in this PR.

## PR Checklist

### DIFF
--- a/src/flow/src/batching_mode/engine.rs
+++ b/src/flow/src/batching_mode/engine.rs
@@ -521,17 +521,63 @@ impl BatchingEngine {
         // check execute once first to detect any error early
         task.check_or_create_sink_table(&engine, &frontend).await?;
 
+        let (start_tx, start_rx) = oneshot::channel();
+
         // TODO(discord9): use time wheel or what for better
         let handle = common_runtime::spawn_global(async move {
-            task_inner.start_executing_loop(engine, frontend).await;
+            if start_rx.await.is_ok() {
+                task_inner.start_executing_loop(engine, frontend).await;
+            }
         });
         task.state.write().unwrap().task_handle = Some(handle);
+        let task_for_rollback = task.clone();
 
-        // only replace here not earlier because we want the old one intact if something went wrong before this line
-        let replaced_old_task_opt = self.tasks.write().await.insert(flow_id, task);
-        drop(replaced_old_task_opt);
+        // Only replace here, not earlier, because we want the old one intact if
+        // something went wrong before this line. Hold both registry locks while
+        // re-checking existence and updating the two maps so create/remove can't
+        // observe a task without its matching shutdown sender, or vice versa.
+        let (replaced_old_task_opt, replaced_old_shutdown_tx) = {
+            let mut tasks = self.tasks.write().await;
+            let mut shutdown_txs = self.shutdown_txs.write().await;
 
-        self.shutdown_txs.write().await.insert(flow_id, tx);
+            let is_exist = tasks.contains_key(&flow_id);
+            match (create_if_not_exists, or_replace, is_exist) {
+                (_, true, true) => {
+                    info!(
+                        "Replacing flow with id={} after final registry check",
+                        flow_id
+                    );
+                }
+                (false, false, true) => {
+                    abort_flow_task(flow_id, Some(task), "unregistered");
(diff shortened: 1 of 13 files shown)


src/flow/src/batching_mode/engine.rs
```diff
@@ -521,17 +521,63 @@ impl BatchingEngine {
...
-        // only replace here not earlier because we want the old one intact if something went wrong before this line
-        let replaced_old_task_opt = self.tasks.write().await.insert(flow_id, task);
-        drop(replaced_old_task_opt);
+        // Only replace here, not earlier, because we want the old one intact if
+        // something went wrong before this line. Hold both registry locks while
+        // re-checking existence and updating the two maps so create/remove can't
+        // observe a task without its matching shutdown sender, or vice versa.
+        let (replaced_old_task_opt, replaced_old_shutdown_tx) = {
+            let mut tasks = self.tasks.write().await;
+            let mut shutdown_txs = self.shutdown_txs.write().await;
```
evenyag: Is it easy to cause a deadlock if other places lock in a different order?

Can we wrap tasks and shutdown_txs with one lock instead?

src/flow/src/batching_mode/table_creator.rs
```diff
@@ -0,0 +1,377 @@
...
+        .map(|f| f.name().clone())
+        .collect();
+    // auto create table use first timestamp column in group by clause as time index
+    let first_time_stamp = fields
+        .iter()
+        .find(|f| {
+            all_pk_cols.contains(&f.name().clone())
+                && ConcreteDataType::from_arrow_type(f.data_type()).is_timestamp()
+        })
+        .map(|f| f.name().clone());
```
evenyag: Is it possible that the timestamp column is not in group by?

src/flow/src/batching_mode/task.rs
```diff
@@ -133,6 +123,13 @@ pub struct TaskArgs<'a> {
 pub struct PlanInfo {
     pub plan: LogicalPlan,
     pub filter: Option<FilterExprInfo>,
+    /// Dirty windows consumed while building a filterless query.
```
evenyag: What is a filterless query?

discord9: full query like tql or sql without dirty time window filter

evenyag:
@codex review

evenyag (approved):

PR #8160, opened 2026-05-25
fix: reregister missing heartbeat pusher

WenyXu:

as title

## PR Checklist

### DIFF
--- a/src/meta-srv/src/handler.rs
+++ b/src/meta-srv/src/handler.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, HashSet};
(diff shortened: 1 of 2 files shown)


fengjiachun:
I think this PR should be closed rather than adding re-registration on the heartbeat hot path.

The root issue is that stepdown currently clears the shared pusher map while existing heartbeat streams still keep their local `pusher_id` state. That creates an inconsistent lifecycle: the stream still believes it is registered, but the map entry has been removed. Re-registering on the next heartbeat only papers over that inconsistency, and can even re-add a pusher after stepdown before the stream is actually closed.

A better fix would be to close the response channel/pusher during stepdown, let the heartbeat stream observe that and exit, and then remove the pusher through the normal stream-owned deregistration path. Also, `clear` should not be a recommended production API for this lifecycle anymore; it should be deprecated or restricted to exceptional/test cleanup paths.

PR #8174, opened 2026-05-26
fix: reset procedure manager state on stop

WenyXu:

This PR resets `ProcedureManager` runtime state when stopping the local procedure manager.
Previously, `LocalManager::stop()` only stopped the background outdated-meta task and marked the manager as not running. In-memory procedure metadata, running/finished procedure sets, key locks, and spawned runner tasks could remain after metasrv leader stepdown. If the same metasrv process became leader again, recovery could be blocked by stale in-memory procedure state, for example by duplicate procedure IDs.

This PR changes the stop lifecycle to best-effort terminate the in-memory execution lifecycle:
- Track spawned root and subprocedure runner tasks by `ProcedureId`.
- Mark the manager as stopped before cleanup.
- Stop the outdated-meta background task. If stopping it fails, log the error and continue cleanup.
- Abort and await active procedure runner tasks.
- Reset runtime-only state:
  - procedure metadata
  - running procedure set
  - finished procedure queue
  - runner task registry
  - procedure key locks
  - dynamic key locks
- Keep persistent procedure store data intact so unfinished procedures can be recovered on the next `start()`.
This does not introduce API or data compatibility changes.

## PR Checklist

### DIFF
--- a/src/common/procedure/src/local.rs
+++ b/src/common/procedure/src/local.rs
@@ -336,7 +336,15 @@ impl ManagerContext {
         self.procedures.write().unwrap().clear();
         self.running_procedures.lock().unwrap().clear();
         self.finished_procedures.lock().unwrap().clear();
-        self.runner_tasks.lock().unwrap().clear();
+        for handle in self
+            .runner_tasks
+            .lock()
+            .unwrap()
+            .drain()
+            .map(|(_, handle)| handle)
+        {
+            handle.abort();
+        }
         self.key_lock.clear();
         self.dynamic_key_lock.clear();
     }
@@ -351,12 +359,16 @@ impl ManagerContext {
         }
 
         let handle = spawn();
-        if handle.is_finished() {
-            return false;
-        }
-
         let _ = tasks.insert(procedure_id, handle);
         true
+    }
+
+    fn remove_procedure(&self, procedure_id: ProcedureId) {
+        self.procedures.write().unwrap().remove(&procedure_id);
+        self.running_procedures
+            .lock()
+            .unwrap()
+            .remove(&procedure_id);
     }
 
     pub(crate) fn remove_runner_task(&self, procedure_id: ProcedureId) {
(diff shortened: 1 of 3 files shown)


WenyXu:
@codex review

WenyXu:
@codex review

WenyXu:
@codex review
/gemini review

WenyXu:
@codex review
/gemini review

evenyag:
@codex review

src/common/procedure/src/local.rs
```diff
@@ -675,16 +736,18 @@ impl LocalManager {
...
+                "LocalManager::submit_root_procedure",
+                    procedure_name = %runner.meta.type_name,
+                    procedure_id = %runner.meta.id,
+                ));
+                // Run the root procedure.
+                // The task was moved to another runtime for execution.
+                // In order not to interrupt tracing, a span needs to be created to continue tracing the current task.
+                runner.run().trace(span).await;
+            })
         });
```
evenyag: The bool returned by `spawn_runner_task` is ignored here. The `ensure!(self.manager_ctx.running(), ManagerNotStartSnafu)` check at the top of `submit_root` can pass and then `running` can flip to `false` before this call (e.g. concurrent `stop()` after a leader stepdown — exactly the scenario this PR targets). When that happens:

- `try_insert_procedure` has already inserted the meta into `procedures` / `running_procedures`
- `spawn_runner_task` returns `false` and does not spawn anything
- the caller still receives `Ok(watcher)` for a procedure that will never run, and the `Watcher` silently hangs

The next `stop()`/`start()` cycle will clear the orphaned meta via `reset_runtime_state`, but in the meantime callers see a procedure stuck in `Running` with no executor. Suggest: when `spawn_runner_task` returns `false`, roll back the `try_insert_procedure` (remove from `procedures` and `running_procedures`) and return `ManagerNotStartSnafu` so the caller sees an explicit failure.

src/common/procedure/src/local.rs
```diff
@@ -329,6 +332,64 @@ impl ManagerContext {
...
+    {
+        let mut tasks = self.runner_tasks.lock().unwrap();
+        if !self.running() {
+            return false;
+        }
+
+        let handle = spawn();
+        if handle.is_finished() {
+            return false;
+        }
```
evenyag: The `handle.is_finished()` check here is semantically confusing. A freshly-spawned task essentially can't be finished before the next instruction — the closure body does `runner.run().trace(span).await` which yields on its first `.await`. If it ever were finished, the task did in fact execute its body, but returning `false` here tells the caller "not started," which then (for `submit_root`/`submit_subprocedure`) leaks the meta with no tracked handle.

Either drop the check, or add a comment explaining the scenario it guards against. As written it looks like defensive coding for a case that can't occur.

src/common/procedure/src/local.rs
```diff
@@ -329,6 +332,64 @@ impl ManagerContext {
         self.running.store(false, Ordering::Relaxed);
     }
 
+    fn reset_runtime_state(&self) {
+        self.procedures.write().unwrap().clear();
+        self.running_procedures.lock().unwrap().clear();
+        self.finished_procedures.lock().unwrap().clear();
+        self.runner_tasks.lock().unwrap().clear();
```
evenyag: Worth flagging: `HashMap::clear()` here just **drops** the `JoinHandle`s — in tokio that does NOT abort the underlying tasks (only `abort()` or dropping the future itself does). This is currently safe in the `stop()` path because `abort_runner_tasks().await` drains and awaits all handles before `reset_runtime_state()` runs, so this clear is a no-op on an already-empty map.

But `start()` also calls `reset_runtime_state()`, and if a previous `stop()` was ever cut short (panic, future cancellation) leaving stray handles, those tasks would silently keep running and race with the freshly-started manager. Consider either:
- a `debug_assert!(self.runner_tasks.lock().unwrap().is_empty())` here to enforce the invariant, or
- iterating and `.abort()`-ing handles before clearing as belt-and-suspenders.

evenyag (commented):
Posting a few correctness concerns inline. Overall the stop/reset design is sound and tests are solid — these notes are about edge cases worth tightening before merge.

evenyag (approved):

PR #8179, opened 2026-05-26
feat(flow): support incremental read checkpoints

discord9:

Follow-up to #8154.

This PR adds the PR3b Flow incremental read checkpoint path.

Main changes:

- Add Flow batching checkpoint state for full-snapshot vs incremental mode.
- Track region watermarks returned by terminal queries and only advance checkpoints when returned participation/watermarks prove it is safe.
- Send query extensions for Flow terminal metrics, sink table id, and incremental after-seqs when the Flow is in a safe incremental state.
- Rewrite supported incremental aggregate SQL plans so delta aggregate results are merged with existing sink aggregate state.
- Drain all dirty windows for incremental SQL runs before checkpoint advancement, while keeping partial dirty-window protection for full-snapshot scoped runs.
- Add metrics for checkpoint decisions and query mode.
- Split incremental planning helpers and checkpoint bookkeeping helpers into `task/inc.rs` and `task/ckpt.rs` to keep `task.rs` focused on orchestration.

Limitations / follow-ups:

- Incremental read is only enabled for supported SQL aggregate shapes; unsupported expressions permanently disable incremental mode for that Flow.
- Stale incremental cursor handling is represented in labels/reasons but end-to-end recovery plumbing is not part of this PR.
- runtime observable(i.e. can check is in inc mode in system table)  incremental aggregate behavior is being developed separately and is not included in this draft yet.

## PR Checklist

### DIFF
--- a/src/flow/src/batching_mode/checkpoint.rs
+++ b/src/flow/src/batching_mode/checkpoint.rs
@@ -26,9 +26,11 @@ pub(super) enum FlowQueryFallbackReason {
     /// Some participating regions could not prove safe advancement against
     /// both the returned watermarks and the checkpoint map.
     IncompleteRegionWatermark,
-    /// The query only covered a scoped subset of dirty time windows while
-    /// additional dirty backlog remains, so global checkpoints cannot advance.
-    ScopedDirtyBacklogPending,
+    /// The query only covered part of the dirty backlog, so global checkpoints
+    /// cannot advance yet. Incremental SQL drains all dirty windows before
+    /// checkpoint advancement; this primarily protects scoped full-snapshot
+    /// runs capped by the per-query dirty-window limit.
+    DirtyBacklogPending,
     /// The datanode detected a stale incremental cursor and the Flow
     /// must recompute from scratch.
     StaleCursor,
@@ -45,7 +47,7 @@ impl FlowQueryFallbackReason {
         match self {
             Self::MissingRegionWatermark => "missing_region_watermark",
             Self::IncompleteRegionWatermark => "incomplete_region_watermark",
-            Self::ScopedDirtyBacklogPending => "scoped_dirty_backlog_pending",
+            Self::DirtyBacklogPending => "dirty_backlog_pending",
             Self::StaleCursor => "stale_cursor",
             Self::IncrementalQueryFailure => "incremental_query_failure",
             Self::IncrementalDisabled => "incremental_disabled",
--- a/src/flow/src/batching_mode/checkpoint.rs
+++ b/src/flow/src/batching_mode/checkpoint.rs
@@ -12,12 +12,114 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[allow(dead_code)]
+use crate::batching_mode::state::CheckpointMode;
+
+pub(super) const CHECKPOINT_DECISION_ADVANCE: &str = "advance";
+pub(super) const CHECKPOINT_DECISION_FALLBACK: &str = "fallback";
+pub(super) const CHECKPOINT_REASON_NONE: &str = "none";
+
+/// Why the task fell back to full snapshot mode.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub(super) enum CheckpointMode {
-    /// Full-snapshot reads over the source tables.
-    FullSnapshot,
-    /// Incremental reads driven by explicitly emitted incremental scan
-    /// extensions.
-    Incremental,
+pub(super) enum FlowQueryFallbackReason {
+    /// The query result did not include a region-watermark map at all.
+    MissingRegionWatermark,
+    /// Some participating regions could not prove safe advancement against
+    /// both the returned watermarks and the checkpoint map.
+    IncompleteRegionWatermark,
+    /// The query only covered a scoped subset of dirty time windows while
+    /// additional dirty backlog remains, so global checkpoints cannot advance.
+    ScopedDirtyBacklogPending,
+    /// The datanode detected a stale incremental cursor and the Flow
+    /// must recompute from scratch.
+    StaleCursor,
+    /// A non-stale-cursor query failure; the Flow resets to full snapshot
+    /// to avoid cascading errors.
+    IncrementalQueryFailure,
+    /// Incremental mode has been permanently disabled for this Flow
+    /// (e.g. because the query shape is not incrementally safe).
+    IncrementalDisabled,
+}
+
+impl FlowQueryFallbackReason {
+    pub(super) fn as_label(self) -> &'static str {
+        match self {
+            Self::MissingRegionWatermark => "missing_region_watermark",
+            Self::IncompleteRegionWatermark => "incomplete_region_watermark",
+            Self::ScopedDirtyBacklogPending => "scoped_dirty_backlog_pending",
+            Self::StaleCursor => "stale_cursor",
+            Self::IncrementalQueryFailure => "incremental_query_failure",
+            Self::IncrementalDisabled => "incremental_disabled",
+        }
+    }
+}
+
+/// Decision produced by `BatchingTask::apply_query_result_to_state` after
+/// each Flow query execution. Describes whether the task advanced its
+/// checkpoint state or fell back to full snapshot, and why.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) enum FlowCheckpointDecision {
+    /// FullSnapshot → Incremental transition.
+    ///
+    /// The query exercised every participating region, all returned valid
+    /// watermarks, and the checkpoint map was populated from scratch.
+    /// Subsequent executions will use incremental after-seqs.
+    AdvancedFromFullSnapshot {
+        participating_regions: usize,
+        watermarks: usize,
+    },
+    /// Existing Incremental → Incremental (in-place advancement).
+    ///
+    /// A subset of participating regions advanced their watermarks. The
+    /// task stays in incremental mode with an updated checkpoint map.
+    AdvancedIncremental {
+        participating_regions: usize,
+        watermarks: usize,
+    },
+    /// Any mode → FullSnapshot.
+    ///
+    /// Watermark information was incomplete, a participating region was
+    /// absent from the existing checkpoint map, the task has permanently
+    /// disabled incremental mode, or the query itself failed. The task
+    /// resets to full snapshot semantics for the next execution.
+    FallbackToFullSnapshot {
+        previous_mode: CheckpointMode,
+        reason: FlowQueryFallbackReason,
+    },
+}
+
+impl FlowCheckpointDecision {
+    pub(super) fn mode_label(self) -> &'static str {
+        match self {
+            Self::AdvancedFromFullSnapshot { .. } => {
+                checkpoint_mode_label(CheckpointMode::FullSnapshot)
+            }
+            Self::AdvancedIncremental { .. } => checkpoint_mode_label(CheckpointMode::Incremental),
+            Self::FallbackToFullSnapshot { previous_mode, .. } => {
+                checkpoint_mode_label(previous_mode)
+            }
+        }
+    }
+
+    pub(super) fn decision_label(self) -> &'static str {
+        match self {
+            Self::AdvancedFromFullSnapshot { .. } | Self::AdvancedIncremental { .. } => {
+                CHECKPOINT_DECISION_ADVANCE
+            }
+            Self::FallbackToFullSnapshot { .. } => CHECKPOINT_DECISION_FALLBACK,
+        }
+    }
+
+    pub(super) fn reason_label(self) -> &'static str {
+        match self {
+            Self::FallbackToFullSnapshot { reason, .. } => reason.as_label(),
+            _ => CHECKPOINT_REASON_NONE,
+        }
+    }
+}
+
+pub(super) fn checkpoint_mode_label(mode: CheckpointMode) -> &'static str {
+    match mode {
+        CheckpointMode::FullSnapshot => "full_snapshot",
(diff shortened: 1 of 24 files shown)


src/flow/src/batching_mode/state.rs
```diff
@@ -120,7 +214,23 @@ impl TaskState {
...
+                // existing max-timeout cap. This is best-effort mitigation for
+                // memtable-only / flush-driven stale cursor exposure, not a
+                // correctness guarantee.
+                let next_duration = self.last_query_duration.max(min_refresh_duration);
+                let next_duration = if let Some(max_timeout) = max_timeout {
+                    next_duration.min(max_timeout)
+                } else {
+                    next_duration
+                };
+                self.last_update_time + next_duration
```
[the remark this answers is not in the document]
discord9: Thanks, but the short incremental cadence intentionally uses min_refresh_duration instead of the time-window lower bound, so I will avoid style-only churn here.

src/flow/src/batching_mode/task/inc.rs
```diff
@@ -0,0 +1,205 @@
...
+            )
+            .await
+            .map_err(BoxedError::new)
+            .context(ExternalSnafu)?
+        {
+            extensions.push((
+                FLOW_SINK_TABLE_ID,
+                table.table_info().table_id().to_string(),
+            ));
+        }
```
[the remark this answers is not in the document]
discord9: Thanks, but this is a follow-up optimization and threading cached sink metadata through PlanInfo would add coupling to correctness-sensitive FLOW_SINK_TABLE_ID plumbing.

evenyag:
@codex review

src/flow/src/batching_mode/state.rs
```diff
@@ -50,6 +50,12 @@ pub struct TaskState {
     /// Dirty Time windows need to be updated
     /// mapping of `start -> end` and non-overlapping
     pub(crate) dirty_time_windows: DirtyTimeWindows,
+    checkpoint_mode: CheckpointMode,
+    checkpoints: BTreeMap<u64, u64>,
```
evenyag, line 56: What's the kv of `checkpoints`?

src/flow/src/batching_mode/state.rs
```diff
@@ -120,7 +214,23 @@ impl TaskState {
...
+            if prefer_short_incremental_cadence {
+                // Short incremental cadence: avoid the normal time-window-based
+                // lower bound, but still back off to at least the last query
+                // duration so sink-side work can self-throttle. Preserve the
+                // existing max-timeout cap. This is best-effort mitigation for
+                // memtable-only / flush-driven stale cursor exposure, not a
+                // correctness guarantee.
+                let next_duration = self.last_query_duration.max(min_refresh_duration);
+                let next_duration = if let Some(max_timeout) = max_timeout {
+                    next_duration.min(max_timeout)
```
evenyag: Can we simplify or polish the comment? It's not easy to understand now.

src/flow/src/batching_mode/task.rs
```diff
@@ -411,66 +461,77 @@ impl BatchingTask {
...
-            );
-            METRIC_FLOW_ROWS
-                .with_label_values(&[format!("{}-out-batching", flow_id).as_str()])
-                .inc_by(*affected_rows as _);
-        } else if let Err(err) = &res {
+        if let Err(err) = &res {
             warn!(
-                "Failed to execute Flow {flow_id} on frontend {:?}, result: {err:?}, elapsed: {:?} with query: {}",
-                peer_desc, elapsed, &plan
+                "Failed to execute Flow {flow_id}, result: {err:?}, elapsed: {:?} with query: {}",
```
evenyag: Regression: this warn used to include `on frontend {:?}` from `peer_desc` populated by `frontend_client.handle(...)`. The switch to `query_with_terminal_metrics` dropped that information. During incident response "which frontend served the failing query" is one of the first things you want — consider plumbing peer info back out of `query_with_terminal_metrics` (or returning a richer result struct) so this log line can keep it.

src/flow/src/batching_mode/task.rs
```diff
@@ -411,66 +461,77 @@ impl BatchingTask {
...
-                peer_desc, elapsed, &plan
+                "Flow {flow_id} executed for {:?} before complete, query: {}",
+                elapsed, &plan
             );
             METRIC_FLOW_BATCHING_ENGINE_SLOW_QUERY
-                .with_label_values(&[
-                    flow_id.to_string().as_str(),
-                    &peer_desc.unwrap_or_default().to_string(),
-                ])
+                .with_label_values(&[flow_id.to_string().as_str(), "watermark-path"])
```
evenyag: Same regression as the failure log above, but worse here because it's a metric label: the second label used to be the frontend peer (`peer_desc.unwrap_or_default()`), now it's a hardcoded `"watermark-path"`. That collapses the per-frontend slow-query histogram into a single bucket. If `"watermark-path"` is meant to distinguish from a future non-watermark path, OK — but please add peer info back as a third label rather than dropping it.

src/flow/src/batching_mode/task.rs
```diff
@@ -799,20 +876,32 @@ impl BatchingTask {
...
+            let mut state = self.state.write().unwrap();
+            let window_cnt = if state.checkpoint_mode() == CheckpointMode::Incremental
+                && !state.is_incremental_disabled()
+                && matches!(self.config.query_type, QueryType::Sql)
+            {
+                // Incremental scans are bounded by region sequence checkpoints,
+                // so the dirty-window filter only narrows sink-side/time-window
+                // work. Consume all pending dirty windows in one incremental
+                // query to keep checkpoint advancement globally safe.
+                i32::MAX as usize
```
evenyag: `i32::MAX as usize` is effectively unbounded. After a long downtime the dirty backlog could be huge and the resulting filter `OR` chain blows up substrait encoding size or trips the query timeout — which then forces a fallback to FullSnapshot, costing more than chunking would have. Consider a sane upper cap (a few thousand?) and, when over the cap, just leave `can_advance_checkpoints = false` so checkpoints stay put for this round and advance on the next drained iteration.

src/flow/src/batching_mode/task/inc.rs
```diff
@@ -0,0 +1,249 @@
...
+            } else {
+                None
+            }
+        };
+
+        if let Some(checkpoints_json) = incremental_checkpoints_json {
+            let table = self
+                .config
+                .catalog_manager
+                .table(
```
evenyag: Nit: the sink table id is invariant for the life of the task but this catalog lookup runs on every incremental query. Could be resolved once at task creation and stashed in `TaskConfig` (or memoized on `BatchingTask`). Minor — catalog reads are typically cached, but it's an avoidable async call on the hot path.

src/query/src/dist_plan/merge_scan.rs
```diff
@@ -481,6 +482,17 @@ impl MergeScanExec {
         &self.regions
     }
 
+    pub fn is_flow_sink_scan(&self) -> bool {
+        let Some(region_id) = self.regions.first() else {
```
evenyag: This only checks the first region's `table_id`. In practice a single `MergeScanExec` scans one table so the regions all share a table_id, but that invariant is implicit. If it ever breaks (multi-table merge scan), a partially-matching scan would either incorrectly skip or include terminal metrics. Either `debug_assert!` that all regions share a table_id, or check `self.regions.iter().all(|r| r.table_id() == sink_table_id)`.

src/flow/src/batching_mode/utils.rs
```diff
@@ -315,10 +315,38 @@ fn collect_output_projection_info(plan: &LogicalPlan) -> OutputProjectionInfo {
...
+    }
+
     projection_info.output_aliases = output_aliases;
     projection_info
 }
 
+fn is_passthrough_output_column(alias_name: &str, expr: &Expr) -> bool {
+    matches!(expr, Expr::Literal(_, _))
+        || match alias_name {
+            AUTO_CREATED_UPDATE_AT_TS_COL => expr.to_string().eq_ignore_ascii_case("now()"),
```
evenyag: Brittle: comparing `Expr::to_string()` against `"now()"` will silently break if a future DataFusion pretty-prints `now()` differently (e.g. `current_timestamp()` or `now ( )`), and that'd quietly stop classifying the column as literal — which prevents incremental rewrites without any error. Pattern-matching `Expr::ScalarFunction { func, .. }` against the `now` UDF would be more robust.

src/flow/src/batching_mode/checkpoint.rs
```diff
@@ -12,12 +12,116 @@
...
+    /// both the returned watermarks and the checkpoint map.
+    IncompleteRegionWatermark,
+    /// The query only covered part of the dirty backlog, so global checkpoints
+    /// cannot advance yet. Incremental SQL drains all dirty windows before
+    /// checkpoint advancement; this primarily protects scoped full-snapshot
+    /// runs capped by the per-query dirty-window limit.
+    DirtyBacklogPending,
+    /// The datanode detected a stale incremental cursor and the Flow
+    /// must recompute from scratch.
+    StaleCursor,
```
evenyag, line 36: The PR description notes that end-to-end stale-cursor recovery isn't wired up in this PR. Nothing currently produces this variant — please add a `// TODO(PR3c):` comment or similar so future readers don't think the path is live and grep-able.

evenyag (commented):
Looks solid overall — typed state machine, conservative fallbacks, good test matrix. A few comments inline, the peer-info regression in the failure log/slow-query metric and the `i32::MAX` dirty-window cap are the two worth addressing before merge; the rest are nits.

src/flow/src/batching_mode/task/inc.rs
discord9: Kept this uncached intentionally after review: the sink table id can become stale if the table is dropped and recreated while the flow task is running. Catalog reads are cached, and this path is correctness-sensitive.

src/flow/src/batching_mode/checkpoint.rs
discord9: This path is now live: query_failure_reason() maps StatusCode::RequestOutdated to FlowQueryFallbackReason::StaleCursor, with unit coverage for the state reset/metric reason.

evenyag (commented):
Three inline notes — overall the state machine and metric labels look solid, LGTM after minor cleanups.

src/flow/src/batching_mode/task.rs
```diff
@@ -62,6 +64,11 @@ use crate::metrics::{
 };
 use crate::{Error, FlowId};
 
+mod ckpt;
+mod inc;
+
+const MAX_INCREMENTAL_DIRTY_WINDOW_FILTERS: usize = 4096;
```
evenyag, line 74: Consider exposing this via `BatchingModeOptions` (or at least adding a doc comment explaining why 4096). A 4096-way OR filter on a single SQL has nontrivial planner/executor cost downstream, and operators may want to tune this without recompiling once they hit the cap.

src/flow/src/batching_mode/task/ckpt.rs
```diff
@@ -0,0 +1,189 @@
...
+        let checkpoint_mode = state.checkpoint_mode();
+        if !can_advance_checkpoints {
+            state.mark_full_snapshot();
+            return FlowCheckpointDecision::FallbackToFullSnapshot {
+                previous_mode: checkpoint_mode,
+                reason: FlowQueryFallbackReason::DirtyBacklogPending,
+            };
+        }
+
+        if checkpoint_mode == CheckpointMode::Incremental && state.is_incremental_disabled() {
```
evenyag: This branch looks unreachable in steady state: `disable_incremental()` already calls `mark_full_snapshot()`, and `advance_checkpoints` / `advance_incremental_checkpoints_with_participation` only transition to `Incremental` when `!incremental_disabled`. If this is defensive against a concurrent `disable_incremental()` racing between `prepare_plan_for_incremental` and here, a one-line comment would make that intent clear — otherwise it can be dropped.

src/flow/src/batching_mode/incremental_filter.rs
```diff
@@ -0,0 +1,222 @@
...
+                "Flow {} cannot infer sink dirty-window filter column: no timestamp group key in {:?}",
+                flow_id, analysis.group_key_names
+            );
+            None
+        }
+        _ => {
+            warn!(
+                "Flow {} cannot infer sink dirty-window filter column: ambiguous timestamp group keys {:?}",
+                flow_id, candidates
+            );
```
evenyag: Both `warn!` calls will fire on every query cycle for flows that permanently match these shapes (no timestamp group key / ambiguous timestamp group keys), since the inference runs every round. Consider downgrading to `debug!`, or logging once per flow (e.g., a one-shot flag on `TaskState`) to avoid log spam.

evenyag (approved):

src/flow/src/batching_mode/task.rs
discord9: Added a doc comment explaining why this cap exists: it bounds generated OR filters to keep Substrait encoding/planning predictable, and larger backlogs drain a capped batch without checkpoint advancement until a later run.

src/flow/src/batching_mode/task/ckpt.rs
discord9: Kept the branch and added a short comment. It is a defensive fallback in case incremental is disabled between plan preparation and result application; in that state we should not advance checkpoints.

src/flow/src/batching_mode/incremental_filter.rs
discord9: Downgraded these repeated inference messages from warn to debug to avoid log spam for flows that permanently match these shapes.

PR #8187, opened 2026-05-27
perf: read primary key as binary if it overflows the dictionary

evenyag:

When the `__primary_key` chunk in an SST is large, the parquet writer falls back from dictionary encoding to plain encoding. Reading such a chunk back into a `DictionaryArray<UInt32, Binary>` (as the flat-format reader's target schema requests) forces Arrow to rebuild the dictionary on the fly, which is expensive for wide / high-cardinality PKs.

This PR adds a small reader-side heuristic and adapter so we read those chunks as plain `Binary` and wrap the result back into an identity `DictionaryArray` ourselves, while keeping the downstream batch layout unchanged.

- `should_read_pk_as_binary` (in `sst/parquet/reader.rs`) samples the first 4 row groups and returns `true` if any of them has a `__primary_key` chunk whose uncompressed size exceeds `DEFAULT_DICTIONARY_PAGE_SIZE_LIMIT` — the same threshold the writer uses to bail out of dictionary encoding.
- When that fires, `ParquetReaderBuilder` overrides the schema given to `ArrowReaderMetadata` so the `__primary_key` field is `Binary` instead of `Dictionary<UInt32, Binary>` (`override_pk_field_to_binary` in `sst.rs`), and `FlatReadFormat::set_pk_as_binary` arms `convert_batch` to wrap the column back via `wrap_pk_binary_to_dict` (identity keys `0..n`, no dedup), so consumers still see the canonical dictionary layout.
- `matching_row_ranges_by_primary_key` in `sst/parquet/prefilter.rs` now accepts either `PrimaryKeyArray` (existing dict path) or `BinaryArray` (new plain path), sharing the matched-range coalescing via `push_matched_range`. Without this, prefiltering would fail with "Primary key column is not a dictionary array" on the same files this optimization enables.

Unit tests cover the heuristic on small vs large chunks, the dict-wrap conversion (with a repeated value to verify identity keys / no dedup), and `eval_pk_group_mask` against a binary PK batch.

On the same dataset that has L0 files.

Before
```
Summary Average: 1983150 rows in 84.362665ms over 10 iterations
```

After
```
Summary Average: 1983150 rows in 73.115644ms over 10 iterations
```

## PR Checklist

### DIFF
--- a/src/mito2/src/sst.rs
+++ b/src/mito2/src/sst.rs
@@ -270,6 +270,26 @@ pub(crate) fn internal_fields() -> [FieldRef; 3] {
     ]
 }
 
+/// Returns a copy of `schema` with the `__primary_key` field replaced by a plain `Binary` field.
+pub(crate) fn override_pk_field_to_binary(schema: &SchemaRef) -> SchemaRef {
+    let new_fields = schema
+        .fields()
+        .iter()
+        .map(|field| {
+            if field.name() == PRIMARY_KEY_COLUMN_NAME {
+                Arc::new(Field::new(
+                    PRIMARY_KEY_COLUMN_NAME,
+                    ArrowDataType::Binary,
+                    field.is_nullable(),
+                ))
+            } else {
+                field.clone()
+            }
+        })
+        .collect::<Vec<_>>();
+    Arc::new(Schema::new(new_fields))
+}
+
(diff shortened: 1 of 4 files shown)


evenyag:
@codex review

v0y4g3r (approved):

Issue #8171, opened 2026-05-26
Require capability check when opening regions on datanode heartbeat


WenyXu:
### What type of enhancement is this?

User experience

### What does the enhancement do?

When `datanode` handles `open_region` instructions from heartbeat, it should validate whether the target region can actually be opened on the current datanode.

Currently, an `open_region` request may be issued to a datanode even when the datanode is not backed by object storage. In this case, opening the region on another datanode may become inaccessible, because the region data may only exist on the original node's local storage.

We should make the `open_region` flow more explicit and safer.

## Expected behavior

Before opening a new region from heartbeat instructions, the datanode should ensure that the region is eligible to be opened on this node.

Possible approaches:

1. Add a `reason` field to `open_region`, for example:
   - region migration
   - region failover
   - region follower
   - normal open

2. Add a capability check before opening the region, for example:
   - require the region to be backed by object storage
   - reject or skip opening regions that depend on local-only storage

PR #8194, opened 2026-05-28
feat: check open region requirements

WenyXu:

Related to #8171

This PR adds explicit region open requirements to prevent region migration/failover from opening regions whose data is not backed by object storage.
Main changes:
- Add `OpenRegionReason` to meta `OpenRegion` instructions.
  - `RegionMigration`
  - `RegionFailover`
  - `RegionFollower` behind the `enterprise` feature
- Add `RegionRequirements` to describe requirements for region operations.
  - Currently supports `object_storage`.
  - The type is defined in `store-api` so it can be reused by future request paths, such as create region.
- Make `OpenRegion::new()` require both `reason` and `requirements` explicitly.
  - This avoids missing requirement propagation when adding new open-region paths.
- Pass requirements from metasrv open-region instructions through datanode heartbeat handling into `RegionOpenRequest`.
- Require object storage for region migration/failover candidate opens.
- Validate open requirements in Mito before opening the region.
  - The object-store check is centralized in `object_store::util::is_object_storage()`.
  - The core requirement check is extracted into a pure function and covered by unit tests.
- Log open-region reason when datanode receives an open-region instruction.

## PR Checklist

### DIFF
--- a/src/meta-srv/src/procedure/region_migration/open_candidate_region.rs
+++ b/src/meta-srv/src/procedure/region_migration/open_candidate_region.rs
@@ -104,7 +104,7 @@ impl OpenCandidateRegion {
                 region_options,
                 region_wal_options,
                 true,
-                reason,
+                Some(reason),
                 RegionRequirements::object_storage(),
             ));
         }
@@ -253,7 +253,7 @@ mod tests {
             Default::default(),
             Default::default(),
             true,
-            OpenRegionReason::RegionMigration,
+            Some(OpenRegionReason::RegionMigration),
             RegionRequirements::object_storage(),
         )])
     }
@@ -302,7 +302,10 @@ mod tests {
             .new_context(persistent_context.clone());
         let instruction = state.build_open_region_instruction(&mut ctx).await.unwrap();
         let open_regions = instruction.into_open_regions().unwrap();
-        assert_eq!(OpenRegionReason::RegionMigration, open_regions[0].reason);
+        assert_eq!(
+            Some(OpenRegionReason::RegionMigration),
+            open_regions[0].reason
+        );
         assert_eq!(
             RegionRequirements::object_storage(),
             open_regions[0].requirements
@@ -312,7 +315,10 @@ mod tests {
         let mut ctx = env.context_factory().new_context(persistent_context);
         let instruction = state.build_open_region_instruction(&mut ctx).await.unwrap();
         let open_regions = instruction.into_open_regions().unwrap();
-        assert_eq!(OpenRegionReason::RegionFailover, open_regions[0].reason);
+        assert_eq!(
+            Some(OpenRegionReason::RegionFailover),
+            open_regions[0].reason
+        );
         assert_eq!(
             RegionRequirements::object_storage(),
             open_regions[0].requirements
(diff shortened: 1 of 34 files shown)


WenyXu:
@codex review

src/mito2/src/region/opener.rs
```diff
@@ -206,6 +207,29 @@ impl RegionOpener {
...
+        let object_store = get_object_store(&options.storage, &self.object_store_manager)?;
+
+        ensure!(
+            supports_open_region_object_storage_requirement(&object_store),
+            error::OpenRegionRequirementSnafu {
+                region_id: self.region_id,
+                requirement: "object storage",
+                reason: "region data must be accessible from another datanode",
+            }
+        );
```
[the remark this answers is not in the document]
WenyXu: [140e95a](https://github.com/GreptimeTeam/greptimedb/pull/8194/commits/140e95a8366df9fd2fc9c4c720802144f2fa1652)

src/meta-srv/src/procedure/region_migration/open_candidate_region.rs
```diff
@@ -97,6 +104,8 @@ impl OpenCandidateRegion {
                 region_options,
                 region_wal_options,
                 true,
+                Some(reason),
+                RegionRequirements::object_storage(),
```
fengjiachun, line 108: This requirement failure is permanent, but the open-region reply path still reduces datanode failures to `SimpleReply` and `open_candidate_region()` wraps every `result=false` as `RetryLater`. On an FS-backed cluster this will keep retrying until the procedure deadline instead of aborting with the `InvalidArguments` requirement error. Can we preserve the status code in the reply, or classify `OpenRegionRequirement` as non-retryable here?

src/store-api/src/region_request.rs
```diff
@@ -581,6 +604,8 @@ pub struct RegionOpenRequest {
     pub skip_wal_replay: bool,
     /// Replay checkpoint.
     pub checkpoint: Option<ReplayCheckpoint>,
+    /// Requirements for opening the region.
+    pub requirements: RegionRequirements,
```
fengjiachun, line 608: Since this field is now part of the generic `RegionOpenRequest` contract, every engine that accepts `RegionRequest::Open` needs to honor it or explicitly reject unsupported requirements. `mito2` checks it, but `file-engine` currently ignores `request.requirements`, while region migration can still build open instructions from non-mito engine metadata. Otherwise the object-storage guard is bypassed for those engines.

WenyXu: Good catch, [3345cb4](https://github.com/GreptimeTeam/greptimedb/pull/8194/commits/3345cb4e96d5ab531a1b492b79f9b9241775eff0)

src/meta-srv/src/procedure/region_migration/open_candidate_region.rs
WenyXu: Good point. This is a real limitation in the current open-region reply path.

Right now `OpenRegions` still uses `SimpleReply`, so the datanode-side error is reduced to a boolean plus a string. By the time `open_candidate_region()` handles the reply, the original `StatusCode` / retryability information has already been lost, so it can only classify `result = false` as `RetryLater`.

I’d prefer not to fix this by matching error strings or special-casing `OpenRegionRequirement` in the procedure. The cleaner fix is to refactor the instruction reply format to carry structured error metadata, e.g. status code and/or retryability, so metasrv can make the retry/abort decision based on the original error semantics.

I’ll address this in a follow-up PR, since it changes the reply contract beyond this open-region requirement check.

fengjiachun (approved):

PR #8198, opened 2026-05-28
feat: support CSV copy skip bad records

QuakeWang:

Refs #6989

Previously, COPY FROM CSV with `skip_bad_records=true` skipped errors after the stream had already gone through DataFusion `FileStream`. This was unsafe for real CSV parse errors because `FileStream` stops scanning after returning an error, so COPY could silently stop reading the rest of the file while still succeeding.

This PR moves CSV bad-record recovery into the CSV reader path. When `skip_bad_records=true`, COPY FROM CSV uses a tolerant CSV stream that consumes skippable Arrow CSV parse/cast errors inside the decoder loop and continues reading following records. The existing outer skip wrapper is kept only for later adapter cast/filter errors.

The tolerant stream intentionally uses one-record batches. This has a performance cost compared with normal CSV scanning, but it keeps each parse/cast failure isolated to a single record. Arrow's CSV decoder clears buffered rows before type parsing, so a failed multi-row flush cannot be safely retried row by row without replaying input bytes.

It also adds a real SQLness regression case with `good row -> bad parse row -> good row`, asserting both valid rows are imported.

This change is backward compatible. The default behavior remains fail-fast unless `skip_bad_records=true` is set.

## PR Checklist

### DIFF
--- a/src/common/datasource/src/file_format.rs
+++ b/src/common/datasource/src/file_format.rs
@@ -59,7 +59,6 @@ use crate::share_buffer::SharedBuffer;
 
 pub const FORMAT_COMPRESSION_TYPE: &str = "compression_type";
 pub const FORMAT_DELIMITER: &str = "delimiter";
-pub const FORMAT_HEADERS: &str = "headers";
 pub const FORMAT_SCHEMA_INFER_MAX_RECORD: &str = "schema_infer_max_record";
 pub const FORMAT_HAS_HEADER: &str = "has_header";
 pub const FORMAT_SKIP_BAD_RECORDS: &str = "skip_bad_records";
--- a/src/common/datasource/src/file_format.rs
+++ b/src/common/datasource/src/file_format.rs
@@ -59,8 +59,10 @@ use crate::share_buffer::SharedBuffer;
 
 pub const FORMAT_COMPRESSION_TYPE: &str = "compression_type";
 pub const FORMAT_DELIMITER: &str = "delimiter";
+pub const FORMAT_HEADERS: &str = "headers";
 pub const FORMAT_SCHEMA_INFER_MAX_RECORD: &str = "schema_infer_max_record";
 pub const FORMAT_HAS_HEADER: &str = "has_header";
+pub const FORMAT_SKIP_BAD_RECORDS: &str = "skip_bad_records";
 pub const FORMAT_TYPE: &str = "format";
 pub const FILE_PATTERN: &str = "pattern";
 pub const TIMESTAMP_FORMAT: &str = "timestamp_format";
--- a/src/common/datasource/src/file_format/csv.rs
+++ b/src/common/datasource/src/file_format/csv.rs
@@ -14,14 +14,21 @@
 
 use std::collections::HashMap;
 use std::str::FromStr;
+use std::sync::Arc;
+use std::task::Poll;
 
 use arrow::csv::reader::Format;
 use arrow::csv::{self, WriterBuilder};
+use arrow::error::ArrowError;
 use arrow::record_batch::RecordBatch;
-use arrow_schema::Schema;
+use arrow_schema::{Schema, SchemaRef};
 use async_trait::async_trait;
+use bytes::{Buf, Bytes};
 use common_runtime;
+use common_telemetry::warn;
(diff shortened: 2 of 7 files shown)


WenyXu:
@codex review

src/common/datasource/src/file_format.rs
```diff
@@ -59,8 +59,10 @@ use crate::share_buffer::SharedBuffer;
 
 pub const FORMAT_COMPRESSION_TYPE: &str = "compression_type";
 pub const FORMAT_DELIMITER: &str = "delimiter";
+pub const FORMAT_HEADERS: &str = "headers";
```
WenyXu: Maybe this is a duplicate of FORMAT_HAS_HEADER.

src/common/datasource/src/file_format/csv.rs
```diff
@@ -189,10 +228,110 @@ impl DfRecordBatchEncoder for csv::Writer<SharedBuffer> {
     }
 }
 
+pub async fn tolerant_csv_stream(
+    store: &ObjectStore,
+    path: &str,
+    schema: SchemaRef,
+    projection: Vec<usize>,
+    format: &CsvFormat,
+) -> Result<SendableRecordBatchStream> {
```
WenyXu, line 222: Can we move this into a utility function and reduce its input dependencies? For example, it could take a CSV reader directly as input. That would make the function easier to test and help it focus on a single responsibility.

src/operator/src/statement/copy_table_from.rs
```diff
@@ -547,6 +690,27 @@ fn generated_schema_projection_and_compatible_file_schema(
     )
 }
 
+/// Generates a maybe compatible schema by mapping CSV fields to table fields by position.
+fn generated_positional_schema_projection_and_compatible_file_schema(
```
WenyXu: Maybe we can split this into two PRs: one focused on skipping bad rows, and another focused on importing CSV data without headers.

QuakeWang: @WenyXu Addressed.

I removed the `headers` alias and the headerless positional projection from this PR, so the scope is now focused on `skip_bad_records`.

I also moved the duplicated skippable Arrow error check into `common_datasource` and reused it in `copy_table_from`.

For `tolerant_csv_stream`, I split the decoder loop into a helper that works from a byte stream directly. The public function now only handles object-store reading and compression setup.

I kept the tolerant path using one-record batches and added a note explaining why: Arrow CSV clears buffered rows before type parsing, so after a multi-row flush error we cannot safely retry the failed batch row by row without replaying input bytes.

WenyXu: Hi @QuakeWang Please update the PR title and description

src/common/datasource/src/file_format/csv.rs
```diff
@@ -189,10 +228,110 @@ impl DfRecordBatchEncoder for csv::Writer<SharedBuffer> {
...
+
+    let projected_schema = Arc::new(
+        schema
+            .project(&projection)
+            .context(error::InferSchemaSnafu)?,
+    );
+    let mut decoder = csv::ReaderBuilder::new(schema)
+        .with_header(format.has_header)
+        .with_delimiter(format.delimiter)
+        .with_batch_size(SKIP_BAD_RECORDS_BATCH_SIZE)
```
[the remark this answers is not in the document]
WenyXu: +1

src/operator/src/statement/copy_table_from.rs
QuakeWang: @WenyXu Done. I updated the PR title/description and changed `Fixes #6989` to `Refs #6989` since this PR now only covers `skip_bad_records`.

src/common/datasource/src/file_format/csv.rs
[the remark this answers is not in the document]
QuakeWang: @WenyXu Thanks, documented the `batch_size=1` performance tradeoff in Rustdoc and the PR description. A real batching optimization is better handled in a follow-up PR to keep this PR scoped.

WenyXu (approved):

PR #8204, opened 2026-06-01
chore: update opendal to 0.57

shuiyisong:

Update the opendal version to 0.57.
Change log here: https://github.com/apache/opendal/releases/tag/v0.57.0

## PR Checklist


WenyXu:
@codex review

src/common/datasource/src/object_store/oss.rs
```diff
@@ -61,18 +63,23 @@ pub fn build_oss_backend(
...
-                    ALLOW_ANONYMOUS, allow_anonymous, e
-                ),
+                msg: format!("failed to parse the option {}={}, {}", key, value, e),
             }
             .build()
         })?;
-        if allow {
-            builder = builder.allow_anonymous();
+        if skip_signature {
+            builder = builder.skip_signature();
```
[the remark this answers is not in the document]
shuiyisong: I checked this against the OpenDAL 0.57 OSS implementation, and I do not think this comment is valid.

In `opendal-service-oss-0.57.0`, `allow_anonymous()` is deprecated and implemented as `self.skip_signature()`; the builder also computes `skip_signature = self.config.skip_signature || self.config.allow_anonymous`. The OSS core then skips signing when that flag is set. So for OSS in 0.57, keeping the legacy Greptime key on `builder.allow_anonymous()` would end up with the same unsigned-request behavior as calling `builder.skip_signature()` directly.

The 0.57 upgrade notes also state that, for OSS, `allow_anonymous` is kept as a deprecated compatibility alias and new code should use `skip_signature()`. Pre-0.57 OSS likewise skipped signing when `allow_anonymous=true`, so the PR is not introducing a new downgrade from signed to anonymous requests for existing `allow_anonymous` configs.

WenyXu (approved):

src/object-store/src/layers/mock.rs
```diff
@@ -222,15 +223,24 @@ impl<A: Access> LayeredAccess for MockAccessor<A> {
...
+        to: &str,
+        args: OpCopy,
+        opts: OpCopier,
+    ) -> Result<(RpCopy, Self::Copier)> {
+        if let Some(result) = self
+            .copy_interceptor
+            .as_ref()
+            .and_then(|copy_interceptor| copy_interceptor(from, to, args.clone()))
+        {
+            return result.map(|rp_copy| (rp_copy, Box::new(()) as oio::Copier));
```
evenyag, line 238: Is it correct to return () as Copier?

shuiyisong: Yes, this is correct for this mock path.

In OpenDAL 0.57, `()` implements `oio::Copy`: `next()` returns `Ok(None)`, `close()` returns default metadata, and `abort()` is a no-op. The interceptor branch has already returned the mocked `RpCopy`, so the returned copier only needs to represent an already-completed copy operation.

I also verified the current usage in `test_engine_copy_region_failure`; the targeted test passes with this implementation.

PR #8211, opened 2026-06-02
feat: join simplifier for promql binary op

waynexia:

Add one planner-phase expr simplifier optimization to reduce duplicate vector selector execution for promql-generated join plan. Only the planner has enough context of whether a join can be avoided or reused, otherwise it will need lots of conditions to implement this as a general optimizer rule.

## PR Checklist


fengjiachun:
@codex review

src/query/src/promql/planner.rs
```diff
@@ -607,11 +777,323 @@ impl PromPlanner {
...
+            return Ok(None);
+        }
+
+        let mut planned_leaves = Vec::with_capacity(collect_env.leaves.len());
+        for (idx, leaf) in collect_env.leaves.iter().enumerate() {
+            let plan = self
+                .prom_vector_selector_to_plan(&leaf.selector, false)
+                .await?;
+            let ctx = self.ctx.clone();
+            let alias = TableReference::bare(format!("prom_v{idx}"));
```
evenyag: Do we need a more special name (like with `__` prefix) to avoid name collisions?

evenyag (approved):

PR #8230, opened 2026-06-03
feat: identify noneffective binary modifiers

waynexia:

identify noneffective binary modifiers to help narrow join rule.

Representative join metrics on my local test env:

  - After: elapsed_compute=90ms, build_time=13ms, join_time=77ms, build_mem_used=39.0MiB
  - Before: elapsed_compute=2.09s, build_time=1.15s, join_time=601ms, build_mem_used=266.3MiB

## PR Checklist


src/query/src/promql/planner.rs
```diff
@@ -782,6 +782,8 @@ impl PromPlanner {
                     // under this case we only join on time index
                     left_context.tag_columns.is_empty() || right_context.tag_columns.is_empty(),
                     modifier,
+                    &left_context,
+                    &right_context,
```
discord9, line 1279: This binary join path aligns field columns to the zipped prefix above, but the comparison-filter path later projects the original side context via `project_binary_join_side`. For uneven multi-field inputs, only the aligned field is compared while extra fields from the preserved side can still be emitted unfiltered. Please pass the aligned output field list into `project_binary_join_side` (or truncate the projected context) and add a regression such as `two_field_metric > one_field_metric`.

src/query/src/promql/planner.rs
```diff
@@ -3494,18 +3496,25 @@ impl PromPlanner {
 
     fn binary_join_key_columns(
         &self,
-        left: &LogicalPlan,
-        right: &LogicalPlan,
+        left_schema: &DFSchemaRef,
+        right_schema: &DFSchemaRef,
+        left_context: &PromPlannerContext,
+        right_context: &PromPlannerContext,
```
discord9, line 4009: These per-side contexts are added here, but the non-TSID fallback below still builds both join-key sets from `self.ctx.tag_columns` (usually the RHS context). When TSID is disabled/unavailable and the two sides have different label sets, default PromQL matching can incorrectly join on the RHS subset. Please derive fallback keys from `left_context.tag_columns` and `right_context.tag_columns` independently, and add a TSID-disabled mismatched-label regression (for example, `timestamp(left_host_job) / right_by_job`).

evenyag (approved):

PR #8231, opened 2026-06-03
fix: run eval-interval flow without time window

discord9:

N/A

This PR fixes batching flows whose SQL query has an `EVAL INTERVAL` but no recognized time-window expression.

Previously, such SQL flows still depended on dirty-window signals before generating a full-query plan. This could skip interval-triggered evaluations when the query cannot be scoped by a time-window expression. Derived subquery sources could also be missed by source table extraction, so inserts into the actual source tables might not mark the flow dirty.

Changes:

- Recursively extract source tables from SQL derived subqueries, nested joins, SQL CTEs, and wrapped table factors.
- Make no-time-window SQL batching flows with `EVAL INTERVAL` generate an unscoped full-query plan on each interval tick, independent of dirty signals.
- Keep non-eval-interval no-time-window SQL flows dirty-signal driven.
- Add unit tests for recursive source extraction and eval-interval no-time-window plan generation.
- Add an anonymized sqlness case covering an eval-interval batching flow whose aggregate subqueries are joined by `LEFT JOIN`, including source extraction and sink output after `ADMIN FLUSH_FLOW`.

## PR Checklist

- [x] I have written the necessary rustdoc comments. (N/A: no public API/rustdoc change)

Validation:

- `cargo test -p sql util::tests --lib`
- `cargo test -p flow test_no_time_window_sql_with_eval_interval_generates_plan_without_dirty_signal --lib`
- `cargo sqlness bare -t flow_batch_join_subquery --preserve-state --bins-dir /mnt/nvme_rust/rust-targets/flow-batch-join-sqlness-20260603/debug`

### DIFF
--- a/src/flow/src/batching_mode/task.rs
+++ b/src/flow/src/batching_mode/task.rs
@@ -865,11 +865,19 @@ impl BatchingTask {
                 }
                 // no new data, sleep for some time before checking for new data
                 Ok(None) => {
-                    debug!(
-                        "Flow id = {:?} found no new data, sleep for {:?} then continue",
-                        self.config.flow_id, min_refresh
-                    );
-                    tokio::time::sleep(min_refresh).await;
+                    if let Some(eval_interval) = &mut interval {
+                        debug!(
+                            "Flow id = {:?} found no new data, wait for eval interval tick then continue",
+                            self.config.flow_id
+                        );
+                        eval_interval.tick().await;
+                    } else {
+                        debug!(
+                            "Flow id = {:?} found no new data, sleep for {:?} then continue",
+                            self.config.flow_id, min_refresh
+                        );
+                        tokio::time::sleep(min_refresh).await;
+                    }
                     continue;
                 }
                 // TODO(discord9): this error should have better place to go, but for now just print error, also more context is needed
@@ -966,6 +974,27 @@ impl BatchingTask {
                         "Flow id = {:?}, no time window, using the same query",
                         self.config.flow_id
                     );
+                    if self.config.flow_eval_interval.is_some()
+                        && self.config.time_window_expr.is_none()
+                    {
+                        // Evaluation-interval SQL without a recognized time-window expression is
+                        // an unscoped full-query flow. Its cadence is controlled by the interval,
+                        // not by source dirty-window signals, so each interval tick should execute
+                        // the original query even when no dirty signal has arrived.
+                        let (_, dirty_windows_to_restore) = self.drain_dirty_windows_signal();
+                        let plan_info = self
+                            .gen_unfiltered_plan_info(
+                                engine,
+                                query_ctx,
+                                sink_table_schema.clone(),
+                                primary_key_indices,
+                                allow_partial,
+                                dirty_windows_to_restore,
+                                None,
+                            )
+                            .await?;
+                        return Ok(Some(plan_info));
+                    }
                     // clean dirty time window too, this could be from create flow's check_execute
                     return self
                         .gen_unfiltered_plan_info_if_dirty(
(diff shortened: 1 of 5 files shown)


src/flow/src/batching_mode/task.rs
```diff
@@ -865,11 +865,19 @@ impl BatchingTask {
                 }
                 // no new data, sleep for some time before checking for new data
                 Ok(None) => {
-                    debug!(
-                        "Flow id = {:?} found no new data, sleep for {:?} then continue",
-                        self.config.flow_id, min_refresh
-                    );
-                    tokio::time::sleep(min_refresh).await;
+                    if let Some(eval_interval) = &mut interval {
```
evenyag: This `Ok(None)` change applies to **every** flow with `flow_eval_interval` set, not just the no-time-window case this PR targets. Existing time-windowed eval-interval flows previously re-polled every `min_refresh` after an empty poll; now they wait a full interval tick. That's plausibly the intended cadence, but it's a broader behavioral change than the PR description implies — can you confirm it doesn't regress those flows?

src/flow/src/batching_mode/task.rs
```diff
@@ -966,6 +974,27 @@ impl BatchingTask {
                         "Flow id = {:?}, no time window, using the same query",
                         self.config.flow_id
                     );
+                    if self.config.flow_eval_interval.is_some()
```
evenyag: This block's body is essentially identical to the `_ =>` catch-all arm just below (drain dirty windows → `gen_unfiltered_plan_info(...)` → `Ok(Some(...))`). Consider collapsing the two: adjust the match guard so the eval-interval-no-window case falls into the same `_` arm, leaving only the *non*-eval-interval `(None, Sql)` case on `gen_unfiltered_plan_info_if_dirty`. Reduces the risk of the two paths drifting.

Minor: `self.config.time_window_expr.is_none()` is always true in this arm (it's only reached when `expire_time_window_bound` is `None`, i.e. `time_window_expr` is `None`), so the second condition is redundant — fine as defensive code, but worth a note.

evenyag (approved):

src/sql/src/util.rs
```diff
@@ -339,21 +340,34 @@ pub fn location_to_index(sql: &str, location: &sqlparser::tokenizer::Location) -
...
+    if let Some(with) = &query.with {
+        for cte in &with.cte_tables {
+            cte_names.insert(ParserContext::canonicalize_identifier(cte.alias.name.clone()).value);
+            extract_tables_from_sql_query(&cte.query, names);
+        }
+    }
+
+    extract_tables_from_set_expr(&query.body, names);
+    remove_cte_names(names, &cte_names);
+}
```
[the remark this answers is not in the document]
discord9: Fixed by extracting CTE sources in isolated scoped sets and added coverage for shadowing and chained CTEs.

src/sql/src/util.rs
```diff
@@ -210,8 +210,9 @@ fn extract_tables_from_hybrid_cte_query(query: &Query, sql_names: &mut HashSet<O
     if let Some(hybrid_cte) = &query.hybrid_cte {
         for cte in &hybrid_cte.cte_tables {
             cte_names.insert(ParserContext::canonicalize_identifier(cte.name.clone()).value);
-            if let CteContent::Tql(tql) = &cte.content {
-                extract_tables_from_tql(tql, &mut tql_names);
+            match &cte.content {
+                CteContent::Sql(query) => extract_tables_from_sql_query(query, sql_names),
```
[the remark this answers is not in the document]
discord9: Renamed the CTE SQL binding to `cte_query` in the updated extraction path.

src/flow/src/batching_mode/task.rs
discord9: Reverted this broader Ok(None) behavior change; empty polls now keep the previous min_refresh sleep.

src/flow/src/batching_mode/task.rs
discord9: Collapsed this into the existing unfiltered catch-all by guarding only the non-eval SQL no-window arm.

PR #8233, opened 2026-06-04
feat: support headerless CSV copy from

QuakeWang:

Refs #6989.
Follow-up to #8198.

`COPY FROM` for CSV previously exposed the internal `has_header` option, but the public `headers` option from #6989 was not supported. More importantly, setting `has_header=false` only changed CSV parsing; the copy pipeline still used name-based schema mapping, so Arrow-generated headerless column names like `column_1` could not map to table columns correctly.

This PR adds `headers` as a public alias for `has_header`, keeps `has_header` for compatibility, and rejects conflicting values when both are provided.

For `headers=false`, CSV import now maps file columns to table columns by position, following the target table schema order. Users can inspect the table column order with `SHOW CREATE TABLE <table>`. CSV columns beyond the target table columns are ignored, and missing file columns are handled by the existing insert/default/null logic. CSV files with headers keep the existing name-based projection behavior.

This PR also updates the `skip_bad_records` reader schema path so headerless CSV uses table types by position when isolating bad records.

Strict header-name validation is intentionally not included in this PR.

## PR Checklist


tests/cases/standalone/common/copy/copy_from_fs_csv.result
```diff
@@ -201,6 +201,42 @@ select * from csv_skip_bad_records order by ts;
 | 2       | Bob       | 30.5          | 2024-01-01T00:00:02 |
 +---------+-----------+---------------+---------------------+
 
+CREATE TABLE csv_headerless(host_id int, host_name string, reading_value double, ts timestamp time index);
+
+Affected Rows: 0
+
+-- SQLNESS ENV PWD
```
WenyXu: We realized this may fail on Windows (see #8232). It might be worth adding these scenarios to the integration test suite to prevent regressions.

src/operator/src/statement/copy_table_from.rs
```diff
@@ -938,4 +986,126 @@ mod tests {
...
+            format: CsvFormat {
+                has_header,
+                ..CsvFormat::default()
+            },
+            path: "test.csv".to_string(),
+        }
+    }
+
+    #[test]
+    fn test_headerless_csv_schema_projection_is_positional() {
```
WenyXu, line 1008: Assert compatible schema column names and types.

WenyXu (commented):
Could you also document the rules for headerless CSV imports? It would help us keep the import documentation up to date.

docs/copy-from-csv.md
WenyXu, line 1: I think it's fine to include this in the PR description, and our docs agent can take care of updating the documentation afterward.

docs/copy-from-csv.md
```diff
@@ -0,0 +1,32 @@
...
+
+For headerless CSV files, set `HEADERS = 'false'`:
+
+```sql
+COPY metrics FROM '/data/metrics.csv' WITH (FORMAT = 'csv', HEADERS = 'false');
+```
+
+When `HEADERS = 'false'`, columns are mapped by position:
+
+- The first CSV column maps to the first table column, the second CSV column maps
```
WenyXu: Maybe we should also tell users how to determine the column order in the table, for example by using `SHOW CREATE TABLE {TABLE}`.

docs/copy-from-csv.md
```diff
@@ -0,0 +1,32 @@
...
+  to the second table column, and so on.
+- Extra CSV columns are ignored.
```
WenyXu: Do you mean that the CSV contains more columns than the target table?

tests-integration/src/tests/instance_test.rs
WenyXu, line 1: Would you like to add more cases to cover the scenarios you mentioned in the doc?

QuakeWang:
@WenyXu Updated. I removed the standalone docs file and kept the import rules in the PR description. I also added integration coverage for both column-count mismatch cases: CSV files with extra columns and CSV files with fewer columns than the target table.

WenyXu (approved):

PR #8234, opened 2026-06-04
fix: reset region detector after migration

WenyXu:

This PR resets region failure detectors for target/candidate regions after a region migration successfully upgrades metadata to the new leader.

Previously, the migration success path deregistered the old `from_peer` detector, but did not explicitly reset the `to_peer` detector. If the target detector already existed from the candidate/open phase with stale heartbeat state, the region supervisor could detect the newly migrated region as failed shortly after migration completed and trigger an unnecessary failover back.

## PR Checklist


fengjiachun (approved):
LGTM

PR #8238, opened 2026-06-04
fix: timestamp display precision should respect column schema (#8227)

### LINKED ISSUE #8227: Timestamp display precision should respect column schema

User experience

Currently, when querying GreptimeDB through the MySQL CLI, timestamp rendering does not always respect the timestamp precision defined by the column schema. (Maybe pg cli also has this problem)

For example

create table
```sql
CREATE TABLE ts_precision_test (
    ts_s  TIMESTAMP(0),
    ts_ms TIMESTAMP(3),
    ts_us TIMESTAMP(6),
    ts_ns TIMESTAMP(9),
    v     INT32,
    TIME INDEX (ts_ms),
    PRIMARY KEY (v)
);
```

insert data
```sql
INSERT INTO ts_precision_test VALUES (
    '2026-06-02 03:50:00',
    '2026-06-02 03:50:00.195',
    '2026-06-02 03:50:00.195123',
    '2026-06-02 03:50:00.195123456',
    1
);
```

query data
```sql
SELECT * FROM ts_precision_test;
```

actual output
```text
+---------------------+----------------------------+----------------------------+----------------------------+------+
| ts_s                | ts_ms                      | ts_us                      | ts_ns                      | v    |
+---------------------+----------------------------+----------------------------+----------------------------+------+
| 2026-06-02 03:50:00 | 2026-06-02 03:50:00.195000 | 2026-06-02 03:50:00.195123 | 2026-06-02 03:50:00.195123 |    1 |
+---------------------+----------------------------+----------------------------+----------------------------+------+
```

expected output
```text
+---------------------+----------------------------+----------------------------+-------------------------------+------+
| ts_s                | ts_ms                      | ts_us                      | ts_ns                         | v    |
+---------------------+----------------------------+----------------------------+-------------------------------+------+
| 2026-06-02 03:50:00 | 2026-06-02 03:50:00.195    | 2026-06-02 03:50:00.195123 | 2026-06-02 03:50:00.195123456 |    1 |
+---------------------+----------------------------+----------------------------+-------------------------------+------+
```

### Implementation challenges

_No response_

divyansh-1009:

Fixes #8227

### Summary

Fix timestamp display precision in the MySQL wire protocol output — timestamps were always rendered with 6 fractional-second digits regardless of the column's declared precision (`TIMESTAMP(0/3/6/9)`).

### Root Cause

In `src/servers/src/mysql/writer.rs`, the `Timestamp` value was first converted to a `NaiveDateTime` via `to_chrono_datetime_with_timezone()`, then handed to `write_col()`:

```rust
// Before (broken)
let v = datatypes::arrow_array::timestamp_array_value(column, i);
let v = v.to_chrono_datetime_with_timezone(Some(&query_context.timezone()));
row_writer.write_col(v)?;
```

`opensrv-mysql` serializes `NaiveDateTime` using its `Display` implementation, which hard-codes **6 fractional-second digits**. This caused two classes of bugs:

| Column type | Stored value | Actual output | Expected output |
|---|---|---|---|
| `TIMESTAMP(3)` | `03:50:00.195` | `03:50:00.195000` | `03:50:00.195` |
| `TIMESTAMP(9)` | `03:50:00.195123456` | `03:50:00.195123` | `03:50:00.195123456` |

### Fix

Use `Timestamp::to_timezone_aware_string()` directly on the `Timestamp` struct **before** the precision information is lost:

```rust
// After (fixed)
let v = datatypes::arrow_array::timestamp_array_value(column, i);
let s = v.to_timezone_aware_string(Some(&query_context.timezone()));
row_writer.write_col(s)?;
```

`to_timezone_aware_string()` uses chrono's `%.f` specifier, which:
- Strips trailing zeros automatically.
- Preserves full nanosecond fidelity when the unit requires it.
- Emits no fractional part at all for second-precision timestamps.

The fix is a single two-line change with no API or data compatibility impact.

### Tests Added

Added comprehensive unit tests in `src/servers/tests/mysql/mysql_writer_test.rs`:

- **Column type mapping** — every `ConcreteDataType` variant maps to the correct `ColumnType`.
- **`UNSIGNED_FLAG` propagation** — set for `u8/u16/u32/u64`, absent for all other types.
- **Timestamp precision (the bug fix)** — all four units (`Second/Millisecond/Microsecond/Nanosecond`) produce the correct number of fractional digits.
- **Edge cases**: zero subseconds, trailing-zero stripping (e.g. `100 ms → .1`), Unix epoch, negative timestamps (pre-1970), and timezone offset shifts.
- **Column-def ordering** — schema column order is preserved.
- **Empty schema** — produces an empty column list without panic.
- **Decimal variants** — `Decimal128` with different precision/scale pairs.

## PR Checklist


fengjiachun:
@divyansh-1009 
Hi, thanks for your contribution!

Is this PR still being worked on? It is currently blocked by multiple CI failures. If you plan to continue, could you please fix them first?

src/servers/src/mysql/writer.rs
```diff
@@ -265,8 +269,23 @@ impl MysqlResultWriter {
...
+                            let _ = write!(
+                                &mut format_buf,
+                                "{}",
+                                datetime.format("%Y-%m-%d %H:%M:%S%.f")
+                            );
+                        } else {
+                            let _ =
+                                write!(&mut format_buf, "[Timestamp{}: {}]", v.unit(), v.value());
+                        }
+                        row_writer.write_col(&*format_buf)?;
```
fengjiachun: This now passes a string into the row writer while the column is still advertised as MYSQL_TYPE_TIMESTAMP. That works for text queries, but COM_STMT_EXECUTE uses QueryResultWriter::new(..., true, ...), so RowWriter calls ToMysqlValue::to_mysql_bin; the string/[u8] implementation only accepts string-like MySQL column types and returns InvalidData for MYSQL_TYPE_TIMESTAMP. A prepared statement such as SELECT ts FROM a timestamp table would therefore fail before writing the row. We need to keep the binary timestamp encoding path working too, or change the advertised column type consistently.

fengys1996 (approved):
LGTM, Thanks for your contribution! @divyansh-1009

PR #8241, opened 2026-06-05
feat: fan out remote dynamic filter updates from FE

discord9:

Part of the remote dynamic filter work. Builds on the initial FE registration / filter-id plumbing that has landed on `main`.

This PR adds FE-side producer fanout for remote dynamic filters.

Summary:

- Watches FE-side `DynamicFilterPhysicalExpr` producers and fans out stable snapshots as `RemoteDynFilterUpdate` to region subscribers.
- Registers query-scoped remote dynamic filter entries and region subscribers from `MergeScanExec`.
- Starts fanout watchers only after region `do_get()` succeeds, so updates do not race ahead of initial registrations in the query context.
- Keeps registry lifecycle based on the existing `Arc` / `Weak` model:
  - `DynFilterRegistryManager` keeps only a weak index.
  - stream-owned `RemoteDynFilterRegistryLease` owns the registry `Arc`.
  - watcher tasks do not hold registry `Arc`.
  - registry drop closes a lifecycle watch channel, causing watchers to unregister and exit.
- Keeps producer filter ownership weak via `Weak<DynamicFilterPhysicalExpr>` so fanout does not keep producers alive.
- Sends unregister at most once per entry when registry lifecycle ends or the producer weak handle is gone.
- Adds a reconcile tick to catch producer updates that happen while an update RPC is in flight.
- Adds lifecycle-aware / timeout bounded RDF control RPCs:
  - update fanout can be cancelled by registry lifecycle close and has a 30s timeout;
  - unregister is cleanup, so it is not cancelled by lifecycle close, but it has a 30s timeout;
  - unregister timeout for one subscriber does not block later subscribers.

RDF remains an optimization. Update/unregister failures only degrade pruning and do not affect query correctness.

Limitations / follow-ups:

- This PR does not add DN runtime apply / scan wrapper integration.
- This PR does not reintroduce target-peer pinning; subscribers are still keyed by `region_id`.
- Large subscriber sets still fan out sequentially; bounded concurrency can be considered later.
- Timeout metrics and richer log context can be added later.

## PR Checklist

Validation:

- `cargo fmt --check`
- `cargo test -p query remote_dyn_filter_registry --lib`
- `cargo test -p query remote_dyn_filter --lib`
- `cargo check -p query`
- `cargo check -p frontend`

### DIFF
--- a/src/common/query/src/request.rs
+++ b/src/common/query/src/request.rs
@@ -19,10 +19,10 @@ use std::sync::Arc;
 
 use api::v1::region::RegionRequestHeader;
 use datafusion::execution::TaskContext;
-use datafusion::physical_expr::expressions::Column;
+use datafusion::physical_expr::expressions::{Column, InListExpr, lit};
 use datafusion::physical_plan::PhysicalExpr;
 use datafusion::physical_plan::joins::HashTableLookupExpr;
-use datafusion_common::tree_node::{TreeNode, TreeNodeRecursion};
+use datafusion_common::tree_node::{Transformed, TreeNode, TreeNodeRecursion};
 use datafusion_common::{DataFusionError, Result as DataFusionResult};
 use datafusion_expr::LogicalPlan;
 use datafusion_proto::physical_plan::DefaultPhysicalExtensionCodec;
@@ -31,6 +31,7 @@ use datafusion_proto::physical_plan::to_proto::serialize_physical_expr;
 use datafusion_proto::protobuf::PhysicalExprNode;
 use prost::Message;
 use serde::{Deserialize, Serialize};
+use snafu::ensure;
 use store_api::storage::RegionId;
 
 /// Current wire-format version for remote dynamic filter payload updates.
@@ -39,6 +40,7 @@ pub use self::initial_remote_dyn_filter_reg::{
     INITIAL_REMOTE_DYN_FILTER_REGS_MAX_TOTAL_PROTO_BYTES, InitialDynFilterReg,
     InitialDynFilterRegs, InitialDynFilterSnapshot,
 };
+use crate::error::{DynFilterPayloadTooLargeSnafu, Error as CommonQueryError};
 
 pub const DYN_FILTER_PROTOCOL_VERSION: u32 = 1;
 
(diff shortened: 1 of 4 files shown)


src/query/src/dist_plan/remote_dyn_filter_registry.rs
```diff
@@ -70,20 +85,24 @@ struct QueryDynFilterRegistryInner {
...
-    pub fn new(filter_id: FilterId, alive_dyn_filter: Arc<DynamicFilterPhysicalExpr>) -> Self {
+    pub fn new(filter_id: FilterId, producer_filter: Arc<DynamicFilterPhysicalExpr>) -> Self {
         Self {
             filter_id,
-            alive_dyn_filter: Arc::downgrade(&alive_dyn_filter),
+            producer_filter: Arc::downgrade(&producer_filter),
             subscribers: RwLock::new(HashSet::new()),
+            last_sent_generation: AtomicU64::new(0),
+            unregistered: AtomicBool::new(false),
+            fanout_started: AtomicBool::new(false),
```
evenyag: I think we can use a mutex to guard them. It should be more robust.

src/query/src/dist_plan/remote_dyn_filter_registry.rs
```diff
@@ -13,15 +13,29 @@
...
 use store_api::storage::RegionId;
+use tokio::sync::{Notify, watch};
 
 use crate::dist_plan::FilterId;
+use crate::region_query::RegionQueryHandlerRef;
 
-/// Routing metadata for a remote dynamic filter subscriber.
+const REMOTE_DYN_FILTER_UPDATE_PAYLOAD_MAX_BYTES: usize = 64 * 1024;
+const REMOTE_DYN_FILTER_RECONCILE_INTERVAL: Duration = Duration::from_secs(1);
+const REMOTE_DYN_FILTER_CONTROL_RPC_TIMEOUT: Duration = Duration::from_secs(30);
```
evenyag: question: Will it be too long?

evenyag:
@codex review

evenyag:
@codex review

src/query/src/dist_plan/remote_dyn_filter_registry.rs
discord9: changed to 10s

src/common/query/src/request.rs
```diff
@@ -118,13 +119,41 @@ fn encode_physical_expr_to_bytes(expr: &Arc<dyn PhysicalExpr>) -> DataFusionResu
...
+
+fn portable_remote_dyn_filter_expr(
+    expr: Arc<dyn PhysicalExpr>,
+    bounds_only: bool,
+) -> DataFusionResult<Arc<dyn PhysicalExpr>> {
+    expr.transform_up(|node| {
+        if node.as_any().is::<HashTableLookupExpr>()
+            || (bounds_only && node.as_any().is::<InListExpr>())
+        {
+            Ok(Transformed::yes(lit(true)))
```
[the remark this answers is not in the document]
discord9: This shape should not be produced by the remote dynamic-filter producer, and validating producer-side predicate semantics belongs before payload encoding rather than in the payload codec, so I'll keep the fallback scoped as-is.

evenyag (approved):

PR #8242, opened 2026-06-05
fix(flow): support complex SQL full-query flows

discord9:

Follow-up to #8231.

This PR makes batching flows handle complex SQL full-query plans safely.

Previously, batching flow time-window-expression (TWE) detection could infer a TWE from one local scope inside a complex logical plan. For plans with joins, window functions, set operations, nested aggregates, or subqueries, applying dirty-window pruning from that local TWE can be unsafe. Those plans should be treated as no-TWE and, when using `EVAL INTERVAL`, executed as full-query batching flows.

This PR also fixes the follow-up execution path needed by those full-query flows:

- Allow pure SQL CTEs in `CREATE FLOW ... AS WITH ...` while keeping the existing TQL/mixed CTE restrictions.
- Avoid setting an empty `hybrid_cte` for pure SQL CTEs so query display/round-trip stays valid.
- Execute unscoped SQL `EVAL INTERVAL` full-query snapshots as SQL text instead of internal Substrait `InsertIntoPlan`, avoiding logical-plan round-trip issues around complex joins/unions/CTEs and duplicate field aliases.
- Keep ordinary SQL full snapshots, TQL, and incremental SQL flows on the existing plan transport path.
- Fall back to the previous `InsertIntoPlan` transport if SQL unparsing does not support a plan shape.

The TWE change is intentionally conservative: some query shapes may be theoretically safe to prune, but this prioritizes correctness over optimization.

## Validation

- `cargo test -p sql test_parse_create_flow_with`
- `cargo test -p sql test_query_display`
- `cargo sqlness bare -t flow_basic --preserve-state --bins-dir <target-dir>/debug`
- `cargo sqlness bare -t flow_batch_join_subquery --preserve-state --bins-dir <target-dir>/debug`
- Locally validated an `EVAL INTERVAL` SQL CTE full-query flow in standalone and distributed mode, including scheduled execution.

## PR Checklist

### DIFF
--- a/src/flow/src/batching_mode/task.rs
+++ b/src/flow/src/batching_mode/task.rs
@@ -16,7 +16,7 @@ use std::collections::{BTreeSet, HashMap, HashSet};
 use std::sync::{Arc, RwLock};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-use api::v1::CreateTableExpr;
+use api::v1::{CreateTableExpr, TableName};
 use catalog::CatalogManagerRef;
 use common_error::ext::BoxedError;
 use common_query::logical_plan::breakup_insert_plan;
@@ -25,8 +25,8 @@ use common_telemetry::{debug, info};
 use common_time::Timestamp;
 use datafusion::datasource::DefaultTableSource;
 use datafusion::sql::unparser::expr_to_sql;
-use datafusion_common::DFSchemaRef;
 use datafusion_common::tree_node::{Transformed, TreeNode};
+use datafusion_common::{DFSchemaRef, TableReference};
 use datafusion_expr::{DmlStatement, LogicalPlan, WriteOp, col, lit};
 use datatypes::schema::Schema;
 use query::QueryEngineRef;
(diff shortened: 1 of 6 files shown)


evenyag:
@codex review

evenyag (approved):

evenyag:
@codex review

src/flow/src/batching_mode/time_window.rs
```diff
@@ -59,6 +61,80 @@ use crate::expr::error::DataTypeSnafu;
...
+}
+
+impl TimeWindowPlanShape {
+    fn should_skip_time_window_expr(&self) -> bool {
+        self.has_unsupported_pruning_node || self.table_scan_count != 1 || self.aggregate_count > 1
+    }
+
+    fn should_stop_inspection(&self) -> bool {
+        self.has_unsupported_pruning_node || self.table_scan_count > 1 || self.aggregate_count > 1
+    }
```
evenyag, line 82: Looks like they have the same condition.

src/flow/src/batching_mode/task.rs
```diff
@@ -552,16 +579,49 @@ impl BatchingTask {
...
+                    // Evaluation-interval SQL flows without a time-window
+                    // expression execute as full-query snapshots. Send these
+                    // as SQL text instead of Substrait to avoid logical-plan
+                    // round-trip issues around complex joins/unions/CTEs and
+                    // duplicate field aliases. Keep ordinary SQL full snapshots
+                    // on the existing InsertIntoPlan path because SQL unparsing
+                    // is not valid for every planned aggregate shape yet.
+                    // If the local SQL unparser does not support this plan,
+                    // keep the previous InsertIntoPlan transport as a fallback.
+                    match df_plan_to_sql(&insert_plan) {
```
evenyag: Is this insert plan a select statement, not an insert statement?

evenyag (approved):

PR #8248, opened 2026-06-05
fix: restore flow dirty signal after fallback

discord9:
## What
- Restore consumed unscoped dirty signal when an incremental flow query succeeds but falls back to full snapshot due to missing/incomplete watermarks.
- Add regression coverage for successful incremental fallback and normal incremental advancement.

## Why
Without restoring the dirty signal, the next full snapshot round can be skipped because the dirty windows were drained before the incremental query.

## Tests
- cargo test -p flow successful_incremental --lib
- cargo test -p flow --lib


discord9:
Close instead of #8277

PR #8251, opened 2026-06-06
feat(security): add password verifier formats

killme2008:

Stop requiring plaintext passwords at rest in the static/watch file user provider.

A credential password now accepts an explicit verifier format:

- `plain:<password>` — plaintext (the default when no prefix is given)
- `pbkdf2_sha256:<iterations>:<hex_salt>:<hex_hash>` — at-rest hash, used by cleartext-capable paths (HTTP/gRPC Basic, PostgreSQL cleartext, MySQL clear password)
- `mysql_native_password:<hex_sha1_sha1_password>` — at-rest verifier that still serves the MySQL `mysql_native_password` handshake

Parsing is hardened:

- PBKDF2 iterations are bounded.
- PBKDF2 salt length is bounded.
- PBKDF2 hashes must be exactly 32 bytes.
- Password verification uses constant-time comparison.
- Verifier bytes are redacted from debug output.
- Non-plain verifiers are not exported through `get_one_user_pwd()`.

### Capability note

A single verifier format does not serve every wire protocol:

| Verifier | HTTP/gRPC Basic | PostgreSQL cleartext | MySQL clear password | MySQL `mysql_native_password` |
| --- | --- | --- | --- | --- |
| `plain:<password>` or legacy `user=password` | yes | yes | yes | yes |
| `pbkdf2_sha256:<iterations>:<hex_salt>:<hex_hash>` | yes | yes | yes | no |
| `mysql_native_password:<hex_sha1_sha1_password>` | no | no | no | yes |

`pbkdf2_sha256` protects passwords at rest. It does not change wire security; cleartext-capable protocols still need TLS in production.

### Breaking change

Passwords are now prefix-parsed. A legacy plaintext password that literally starts with `plain:`, `pbkdf2_sha256:`, or `mysql_native_password:` changes meaning.

Use `plain:` to make that intent explicit. For example, the literal password `plain:secret` should be configured as:

```text
user=plain:plain:secret
```

## PR Checklist


sunng87:
I suggest to keep this file provider simple. It's not designed for security or any production usage. Users with this concern should switch to the enterprise edition.

killme2008:
> I suggest to keep this file provider simple. It's not designed for security or any production usage. Users with this concern should switch to the enterprise edition.

This seems more like a **baseline** security improvement than a commercial-only feature. Support for non-plaintext password verifiers and separation of internal auth from user credential handling should be part of the open-source core, while enterprise editions can differentiate with SSO, LDAP, audit logs, and advanced access control, which we already have.

MichaelScofield:
@codex review

src/auth/src/user_provider/static_user_provider.rs
```diff
@@ -56,16 +56,24 @@ impl StaticUserProvider {
...
+            .find_map(|(username, (verifier, _))| {
+                verifier
+                    .as_plain_text()
+                    .map(|pwd| (username.clone(), pwd.to_string()))
+            })
+            .context(InvalidConfigSnafu {
+                value: "",
+                msg: "No plain-text credential to export; configure `frontend_auth` or add a plain-text user",
+            })
     }
```
shuiyisong: We can delete this function now, for the internal communication should go with internal ports.

shuiyisong:
`mysql_native_password:<hex_sha1_sha1_password>`
One nit question is, for normal users, how can they get the correct `sha`ed hex string 🤔

killme2008:
> `mysql_native_password:<hex_sha1_sha1_password>` One nit question is, for normal users, how can they get the correct `sha`ed hex string 🤔

I'll implement a command like `greptime password mysql_native_password <password>` to generate the hashed password.

src/auth/src/user_provider/static_user_provider.rs
killme2008: Deleted in https://github.com/GreptimeTeam/greptimedb/pull/8251/changes/0aeb390ec5c9d95e6ef0d24ce081119c5676c00b

shuiyisong (approved):
LGTM

evenyag:
Is it a breaking change?

killme2008:
> Is it a breaking change?

Yes, but it won't break anything actually unless your password starts with `plain:` or `mysql_native_password`.

PR #8254, opened 2026-06-08
perf(mito): cached-size single-pass WAL entry encoder

lyang24:

there are two type of work in wal writing (encoding and io) - this pr focuses on encoding.  The key observation is that prost crate has no size caching (unlike proto c++), so WalEntry::encode_to_vec recomputes the encoded_len of every nested message for each length delimiter. In the deep WAL tree (WalEntry/Mutation/Rows/Row/Value) a leaf Value's length is recomputed once per ancestor level 

so we cache byte length out side of encoding_raw()
We add WalEntryEncoder: one size pass caches every message body length into a flat vector (pre-order), one encode pass writes bytes reading the cached lengths back via a cursor, so each length is computed exactly once. Leaf messages (Value, ColumnSchema, WriteHint, BulkWalEntry) delegate their body encoding to prost but have their single computed length cached.

  ## Correctness

  Output is **byte-for-byte identical** to `encode_to_vec` (required for WAL
  replay compatibility), asserted in tests covering: normal puts, multi-mutation,
  null values, `op_type=Delete` (proto3 default-skip), empty entry, encoder reuse.
 

Measured on a single-region saturation ramp (RaftEngine WAL, local SATA SSD),
back-to-back A/B of clean `main` vs this change, identical load:

| writers | main rows/s | this PR rows/s | throughput | encode time |
|---|---|---|---|---|
| 8  | 619,147 | 724,897 | **+17.1%** | −54% |
| 16 | 750,716 | 885,244 | **+17.9%** | −53% |
| 32 | 826,834 | 979,924 | **+18.5%** | −36% |
| 64 | 840,964 | 960,822 | **+14.3%** | −21% |

Can be removed if prost crate adds size caching upstream.
  

## PR Checklist

### DIFF
--- a/src/mito2/src/wal.rs
+++ b/src/mito2/src/wal.rs
@@ -14,6 +14,7 @@
 
 //! Write ahead log of the engine.
 
+pub mod encoder;
 pub(crate) mod entry_distributor;
 pub(crate) mod entry_reader;
 pub(crate) mod raw_entry_reader;
@@ -25,10 +26,10 @@ use std::sync::Arc;
 use api::v1::WalEntry;
 use common_error::ext::BoxedError;
 use common_telemetry::debug;
+use encoder::WalEntryEncoder;
 use entry_reader::NoopEntryReader;
 use futures::future::BoxFuture;
 use futures::stream::BoxStream;
-use prost::Message;
 use snafu::ResultExt;
 use store_api::logstore::entry::Entry;
 use store_api::logstore::provider::Provider;
@@ -79,6 +80,7 @@ impl<S: LogStore> Wal<S> {
             store: self.store.clone(),
             entries: Vec::new(),
             providers: HashMap::new(),
+            encoder: WalEntryEncoder::new(),
         }
     }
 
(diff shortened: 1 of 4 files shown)


evenyag:
@codex review

src/mito2/src/wal.rs
evenyag, line 1: Is it expected? We rarely use `mod.rs`.

src/mito2/src/wal/encoder.rs
```diff
@@ -0,0 +1,514 @@
...
+    key_len(tag) + encoded_len_varint(body_len as u64) + body_len
+}
+
+/// A reusable encoder that caches message body sizes between its size pass and
+/// its encode pass.
+#[derive(Default)]
+pub(crate) struct WalEntryEncoder {
+    /// Cached body lengths of message nodes, in pre-order.
+    sizes: Vec<usize>,
+}
```
evenyag, line 85: Is it possible to add a benchmark for the encoder to compare against the old encoding method?

lyang24: bench added in /benches

src/mito2/src/wal.rs
lyang24: right refactored it back

v0y4g3r:
@lyang24  Impressive work, thanks!

v0y4g3r (approved):

PR #8256, opened 2026-06-08
feat: introduce plugin setup functions with richer context

sunng87:

The problem is our `setup_*_plugins` has no access to the instance it belongs to. Although we have changed `start_*_plugins` to accept instance, some of the work still needs to be done in setup phase, for example, configuring the router.

This patch breaks setup_*_plugins into `pre_build` and `post_build` versions, with access to `Builder` in the `post_build` version. 

I also merged `setup_dynamic_frontend_plugins` into `pre_build`.

## PR Checklist

### DIFF
--- a/src/cmd/src/frontend.rs
+++ b/src/cmd/src/frontend.rs
@@ -52,7 +52,6 @@ use plugins::PluginOptions;
 use plugins::frontend::context::{
     CatalogManagerConfigureContext, DistributedCatalogManagerConfigureContext,
 };
-use plugins::frontend::setup_frontend_dynamic_plugins;
 use plugins::options::PluginOptionsDeserializerImpl;
 use servers::addrs;
 use servers::grpc::GrpcOptions;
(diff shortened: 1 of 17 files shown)


src/cmd/src/frontend.rs
```diff
@@ -375,6 +370,29 @@ impl StartCommand {
...
         let cache_tti = meta_client_options.metadata_cache_tti;
 
+        let meta_config: Vec<PluginOptions> = meta_client::create_meta_client(
+            MetaClientType::Frontend,
+            meta_client_options,
+            None,
+            None,
+        )
+        .await
+        .context(error::MetaClientInitSnafu)?
```
sunng87, line 380: here I use a one-shot meta_client for fetching configuration. this client is initialized without any plugin.

src/meta-srv/src/bootstrap.rs
shuiyisong, line 425: Should we keep the `.plugins(plugins)` here for the MetasrvBuilder?

sunng87: plugins will be set later here: https://github.com/GreptimeTeam/greptimedb/pull/8256/changes#diff-42c93e62d446f517ff64b4ab377897fe97a375c8e18eeeebf89c54ac6ed428c4R358

This builder now takes a read-only reference to plugins just for reading from it. It won't be able to call set_plugins without cloning it.

shuiyisong (approved):
LGTM

PR #8259, opened 2026-06-08
feat: support remote WAL logical pruning

WenyXu:

This PR adds a logical-delete mode for automatic remote WAL pruning with Kafka.

Previously, automatic WAL pruning always issued Kafka `DeleteRecords` after determining the prunable entry id. Some Kafka deployments, such as AWS MSK, do not support `DeleteRecords` and can only rely on Kafka TTL/retention for physical cleanup. This PR adds a logical-delete mode that skips `DeleteRecords` while still advancing GreptimeDB's metadata marker for pruned entries.

When auto_prune_logical_delete is enabled, the WAL prune procedure skips the physical Kafka DeleteRecords call and only updates the topic metadata marker with the pruned entry id. The default remains false, so existing behavior is unchanged.

## PR Checklist


fengjiachun (approved):
LGTM

Issue #8127, opened 2026-05-18
tracking: remote dyn filter


discord9:
## Summary

Support remote dynamic filter propagation for distributed queries — let frontend-side dynamic filters produced by joins be registered, serialized, and propagated to remote datanode scans so they can prune data earlier. This is an optimization-only feature: failures must safely degrade to local-only dynamic filtering and must not affect query correctness.

## Related PRs

- [x] #7931 — docs: rfc for remote dyn filter
- [x] #7979 — feat: remote dyn filter basics
- [x] #8148 — frontend producer / bridge / bounded initial-register handoff
- [x] #8241 — FE producer fanout and payload encoding fallback
- [x] #8262 — datanode remote filter state / apply runtime
- [ ] *observability, fallback, and lifecycle protection* (follow-up; partially covered by #8241 and #8262)
- [ ] *end-to-end validation and performance baseline* (follow-up; local benchmark/e2e work exists but not yet submitted)
- [ ] *large build-side membership payload, e.g. Bloom / clustered ranges / custom compact representation* (future enhancement)

## Component Breakdown

| Component | Description | Status |
|-----------|-------------|--------|
| RFC / design | Defines the high-level remote dyn filter propagation model and phased rollout | ✅ |
| Wire ABI / query identity | Defines `query_id`, `filter_id`, epoch / completion semantics, payload boundaries, and safe downgrade rules | ✅ |
| Region RPC control plane | Adds a unary frontend -> datanode control-plane entry for remote dyn filter `update` / `unregister` messages | ✅ |
| Frontend producer / bridge | Identifies distributed join dynamic filters, generates stable query-local `filter_id`s, stores minimal query-scoped frontend state, and attaches bounded initial-register metadata to the first remote read | ✅ |
| Frontend fanout / payload encoding | Watches producer snapshots, fans out remote updates to subscribers, and keeps payload serialization bounded with safe downgrade behavior | #8241 |
| Datanode initial registration | Receives initial-register metadata and prepares query-scoped handoff state for later consumer/runtime installation | ✅ |
| Datanode apply runtime | Maintains `query_id + filter_id` state, applies ordered updates, installs remote dynamic filter wrappers into scan predicates, handles remap, and owns successful-path cleanup | #8262 |
| Unregister / lifecycle cleanup | Cleans up frontend and datanode state on end-of-use, query finish, cancel, stream drop, or TTL fallback | 🔄 Partially covered by #8241 and #8262 |
[...]

PR #8262, opened 2026-06-08
feat: apply remote dynamic filters on datanode scans

discord9:

- Tracking issue: #8127
- Builds on #8241 — FE producer fanout and payload encoding fallback.
- datafusion backport, support more join type: https://github.com/GreptimeTeam/datafusion/pull/3
- datafusion aggr exec filter push down fix: https://github.com/GreptimeTeam/datafusion/pull/5

This PR implements the datanode apply path for Remote Dynamic Filter (RDF).

Main changes:

- Add a DN-side remote dynamic filter registry keyed by `query_id + filter_id`.
- Register initial remote dynamic filter metadata when a remote read starts.
- Buffer early RDF updates until the receiver-side scan schema is available.
- Create DN-local `DynamicFilterPhysicalExpr` runtime filters and apply update generations with stale/idempotent/complete handling.
- Inject remote dynamic filters as `FilterExec` predicates, then reuse DataFusion `FilterPushdown` so eligible predicates can reach `RegionScanExec`.
- Add cleanup for stream-scoped registrations on EOF/drop and handle explicit RDF unregister RPCs.
- Support initial snapshot payloads from Task03b by converting them into pending DN updates and applying them when the runtime filter is created.
- Use a default-feature DataFusion `TaskContext` when decoding RDF payloads, because payloads may contain built-in scalar functions such as `struct(...) IN (...)`.
- Extend physical plan wrapper infrastructure to compose multiple wrappers instead of replacing the existing wrapper.

Note on RDF explain output: RDF is now installed through a logical `RemoteDynFilterReceiver` before physical planning, and standalone in-process region reads register the same initial RDF metadata as remote reads. As a result, shared sqlness explain cases may now show `FilterExec: DynamicFilter [...]` and scan-level `"dyn_filters"` in both standalone and distributed modes; the tests redact the concrete dynamic-filter payloads because set/hash ordering is not semantically stable.

The remote dynamic filter remains an optimization only: missing/stale/oversized/decode-failed updates degrade to no extra pruning rather than affecting query correctness.

## PR Checklist

Validated locally:

```text
cargo test -p datanode remote_dyn_filter --lib
cargo test -p query dyn_filter --lib
cargo test -p common-query initial_dyn_filter --lib
cargo fmt --all -- --check
cargo check -p common-query -p query -p table -p datanode
cargo clippy -p datanode --lib --tests -- -D warnings
```


src/datanode/src/region_server/registrations.rs
```diff
@@ -0,0 +1,682 @@
...
+    pub(super) fn new(value: impl Into<String>) -> Self {
+        Self(value.into())
+    }
+}
+
+impl Display for RemoteDynFilterId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
```
evenyag, line 101: Looks like we can use type alias?

src/datanode/src/region_server/registrations.rs
```diff
@@ -0,0 +1,682 @@
...
+    fn from_initial_reg(reg: &InitialDynFilterReg) -> Option<Self> {
+        let snapshot = reg.initial_snapshot.as_ref()?;
+        match &snapshot.payload {
+            DynFilterPayload::Datafusion(payload) => Some(Self {
+                payload: payload.clone(),
+                generation: snapshot.generation,
+                is_complete: snapshot.is_complete,
+            }),
+            #[allow(unreachable_patterns)]
+            _ => None,
```
evenyag: Do we need to allow this?

src/datanode/src/region_server.rs
```diff
@@ -751,31 +862,96 @@ impl RegionServer {
...
+                    "Remote dynamic filter update had no active registration, query_id: {}, filter_id: {}",
+                    query_id, filter_id
+                );
+            }
+            RemoteDynFilterUpdateOutcome::Buffered => {
+                debug!(
+                    "Buffered early remote dynamic filter update, query_id: {}, filter_id: {}",
+                    query_id, filter_id
+                );
+            }
```
evenyag: Can we just match the outcome to a reason str and then log it in one line?

src/datanode/src/region_server.rs
```diff
@@ -2794,27 +3304,475 @@ mod tests {
...
+            mock_region_server
+                .inner
+                .initial_remote_dyn_filter_registrations
+                .get(&query_id)
+                .is_none()
+        );
+    }
+
+    #[test]
+    fn test_remote_dyn_filter_cleanup_stream_removes_on_drop() {
```
evenyag: We added so many tests. Is it possible to reduce the number of test code?

evenyag (commented):
I only reviewed a few files. Will review the remaining later.

src/datanode/src/region_server/registrations.rs
discord9: > Looks like we can use type alias?

I prefer keeping `RemoteDynFilterId` as a newtype for a bit more type safety and clearer intent.

src/datanode/src/region_server/registrations.rs
```diff
@@ -0,0 +1,681 @@
...
+        query_id: &QueryId,
+        expected: &Arc<Mutex<QueryRemoteDynFilterRegs>>,
+    ) {
+        // Protect against stale cleanup of an old per-query state: remove the
+        // outer entry only if it still points to the same inner mutex and that
+        // inner map is still empty.
+        self.queries.remove_if(query_id, |_, query_regs| {
+            Arc::ptr_eq(query_regs, expected) && query_regs.lock().unwrap().is_empty()
+        });
+    }
```
evenyag, line 75: I wonder if it will cause a leak. Is it possible that we didn't clean up a query at last?

evenyag (commented):
A few inline notes from review. All optimization-only / non-blocking, but worth a look — especially the kill-switch gap.

src/datanode/src/region_server/remote_dyn_filter.rs
```diff
@@ -0,0 +1,1039 @@
...
+use crate::error::{self, Result, UnexpectedSnafu};
+use crate::region_server::registrations::{
+    RemoteDynFilterId, RemoteDynFilterUpdateOutcome, apply_remote_dyn_filter_update,
+    initial_dyn_filter_regs_from_query_ctx, register_initial_dyn_filter_regs,
+    remote_dyn_filter_exprs_for_initial_regs, remove_initial_dyn_filter_regs,
+    unregister_remote_dyn_filter,
+};
+use crate::region_server::{RegionServer, RegionServerInner};
+
+fn remote_dyn_filter_receiver_plan(
```
evenyag, line 48: The receiver injection (and `register_initial_remote_dyn_filter_cleanup`) gates on `remote_query_id_value()` and the presence of initial regs, but not on `remote_dyn_filter_pushdown_enabled_from_extensions`. Only `MergeScanExec` (the FE→DN side) honors the flag, so `query.enable_remote_dynamic_filter_pushdown=false` won't suppress the DN receiver-side `FilterExec` injection — notably in standalone mode where there's no MergeScan. Is the kill-switch meant to cover this path too? If so, it should also check the enable flag here.

src/query/src/dist_plan/remote_dyn_filter_receiver.rs
```diff
@@ -0,0 +1,187 @@
...
+        LogicalPlan::Extension(Extension {
+            node: Arc::new(self),
+        })
+    }
+
+    fn dyn_filters(&self) -> &[Arc<dyn PhysicalExpr>] {
+        &self.dyn_filters
+    }
+
+    fn ord_key(&self) -> String {
```
evenyag, line 89: `ord_key` derives ordering from `format!("{self:?}")`, and `Debug` routes through `fmt_for_explain`, which only prints `filters={len}`. So two nodes with the same input-debug and same filter count but *different* `dyn_filters` compare `Ordering::Equal` under `PartialOrd`, while `PartialEq`/`Hash` (which include `dyn_filters`) treat them as distinct — an `Ord`/`Eq` inconsistency. Unlikely to bite today since these aren't kept in ordered/dedup'd sets, but consider basing `ord_key` on the same fields as `Hash`/`Eq` to avoid a latent footgun.

src/query/src/lib.rs
```diff
@@ -30,7 +30,6 @@ pub mod optimizer;
 pub mod options;
 pub mod parser;
 mod part_sort;
-pub mod physical_wrapper;
```
evenyag, line 33: Removing `pub mod physical_wrapper` (and `PhysicalPlanWrapperRef`) drops a `Plugins`-registered public trait — exactly the kind of seam out-of-tree/enterprise builds hook into. The PR checklist marks "API changes are backward compatible," which isn't quite accurate for this trait. Please confirm no enterprise/closed-source consumer registers a `PhysicalPlanWrapper`; if any do, they'll need a migration path.

evenyag (commented):
A few inline notes from review. All optimization-only / non-blocking, but worth a look — especially the kill-switch gap.

src/datanode/src/region_server/registrations.rs
```diff
@@ -0,0 +1,681 @@
...
+            }
+        }
+
+        let should_remove_query = locked.is_empty();
+        (removed_filters, should_remove_query)
+    };
+
+    for registered in removed_filters {
+        registered.deactivate();
+    }
```
evenyag: Can we deactivate on drop to avoid forgetting this?

src/datanode/src/region_server/registrations.rs
```diff
@@ -0,0 +1,681 @@
...
+pub(super) const REMOTE_DYN_FILTER_PAYLOAD_MAX_BYTES: usize = 64 * 1024;
+
+type QueryRemoteDynFilterRegs = HashMap<RemoteDynFilterId, RegisteredDynFilter>;
+
+#[derive(Debug, Default)]
+pub(super) struct RemoteDynFilterRegistry {
+    // Keep cross-query concurrency while making each query's RDF state machine a
+    // single critical section. RDF count per query is small
+    queries: DashMap<QueryId, Arc<Mutex<QueryRemoteDynFilterRegs>>>,
+}
```
evenyag, line 44: Can we keep a task context in the registry or somewhere and reuse it? Now we have to build TaskContext by remote_dyn_filter_task_context() every time.

src/datanode/src/region_server/registrations.rs
discord9: The query-scoped registry is cleaned up by the stream guard on EOF/drop and by unregister, and `remove_query_if_empty` removes the outer query entry once the last filter/subscriber for that query is gone.

src/query/src/lib.rs
discord9: added it back

src/frontend/src/instance.rs
```diff
@@ -340,9 +341,14 @@ impl Instance {
         query_interceptor: Option<&SqlQueryInterceptorRef<Error>>,
         tql: Tql,
     ) -> Result<Output> {
+        let mut tql_query_ctx = query_ctx.as_ref().clone();
+        // tql usually know the exactly data range so no need for remote dyn filters
+        tql_query_ctx.set_extension(QUERY_ENABLE_REMOTE_DYNAMIC_FILTER_PUSHDOWN, "false");
+        let query_ctx = Arc::new(tql_query_ctx);
```
evenyag: Is it possible to enable dyn filter on the necessary path instead of requring disable it explicitly?

For example, can we check the plan in remote_dyn_filter_receiver_plan() and disable dyn filters when there are tql-related plans?

discord9: changed to check in remote_dyn_filter_receiver_plan

src/datanode/src/region_server/registrations.rs
discord9: already using fn remote_dyn_filter_task_context & once lock

src/datanode/src/region_server/registrations.rs
discord9: done, call deactiv in drop

evenyag:
@codex review

evenyag (approved):

fengjiachun (commented):
Retracted.

PR #8277, opened 2026-06-11
fix!: fence scoped flow repair snapshots

discord9:

Refs https://github.com/GreptimeTeam/greptimedb/issues/8125

Depends on https://github.com/GreptimeTeam/greptimedb/pull/8279, which has been merged into `main`.

This PR implements the Flow state-machine side of snapshot-fenced scoped full-snapshot repair for incremental reads.

PR #8279 has been merged into `main` and contains the lower-level snapshot read-bound transport/storage plumbing. The intended review focus for this PR is the Flow batching logic on top of those read bounds.

Main changes:

- Add `QueryCoverage` to distinguish explicit full-query snapshots, scoped base repair, fenced repair chunks, and incremental deltas.
- Keep fenced repair as a `FullSnapshot` sub-state via `pending_fenced_repair` instead of adding a new top-level checkpoint mode.
- Use `QueryCoverage::FencedRepairChunk { high }` to send snapshot read bounds only for fenced repair chunks.
- Advance from scoped full-snapshot repair to incremental mode only when actual returned terminal watermarks prove full coverage.
- Abandon fenced repair on stale fence or watermark mismatch, restore dirty windows, and retry via scoped base repair with a fresh `H`.
- Prevent scoped repair or unsafe incremental rewrite from silently falling back to `UnfilteredFull`.
- Keep `UnfilteredFull` only for explicit full-query paths: TQL and SQL with eval interval.
- Reject SQL flow creation when there is no usable TWE and no eval interval. Complex SQL is treated as having no usable TWE, so it requires eval interval to run as full-query.

## API / behavior compatibility note

This PR intentionally changes SQL batching flow creation behavior for queries without a usable time-window expression (including complex SQL where no safe TWE can be derived): such flows now fail at creation time unless they explicitly specify `EVAL INTERVAL`.

Previously, these flows could be accepted and implicitly run as unfiltered full-query flows. To keep that behavior, users should opt in explicitly, for example:

```sql
CREATE FLOW ... EVAL INTERVAL '1m' AS
SELECT COUNT(*) FROM input_basic;
```

This is a safety break: it avoids silently running unfiltered full queries for ambiguous SQL batching flows, while preserving explicit full-query behavior for TQL and SQL flows with `EVAL INTERVAL`.

## PR Checklist

- [ ] API changes are backward compatible. See the API / behavior compatibility note above.

### DIFF
--- a/src/flow/src/batching_mode/frontend_client.rs
+++ b/src/flow/src/batching_mode/frontend_client.rs
@@ -15,7 +15,7 @@
 //! Frontend client to run flow as batching task which is time-window-aware normal query triggered every tick set by user
 
 use std::collections::HashMap;
-use std::sync::{Arc, Mutex, Weak};
+use std::sync::{Arc, Mutex, RwLock, Weak};
 
 use api::v1::greptime_request::Request;
 use api::v1::query_request::Query;
@@ -25,7 +25,6 @@ use common_error::ext::BoxedError;
 use common_grpc::channel_manager::{ChannelConfig, ChannelManager, load_client_tls_config};
 use common_meta::peer::{Peer, PeerDiscovery};
 use common_query::{Output, OutputData};
-use common_recordbatch::adapter::{RecordBatchMetrics, RegionWatermarkEntry};
 use common_telemetry::warn;
 use futures::stream::{FuturesUnordered, StreamExt};
 use meta_client::client::MetaClient;
@@ -394,6 +393,7 @@ impl FrontendClient {
         schema: &str,
         request: QueryRequest,
         extensions: &[(&str, &str)],
+        snapshot_seqs: &HashMap<u64, u64>,
         peer_desc: &mut Option<PeerDesc>,
     ) -> Result<OutputWithMetrics, Error> {
         let flow_extensions = build_flow_extensions(extensions)?;
@@ -415,7 +415,7 @@ impl FrontendClient {
                         request,
                         &hints,
                         extensions,
-                        &Default::default(),
+                        snapshot_seqs,
                     )
                     .await
                     .map_err(BoxedError::new)
@@ -437,6 +437,7 @@ impl FrontendClient {
                     .current_catalog(catalog.to_string())
                     .current_schema(schema.to_string())
                     .extensions(extensions_map)
+                    .snapshot_seqs(Arc::new(RwLock::new(snapshot_seqs.clone())))
                     .build();
                 let ctx = Arc::new(ctx);
                 let database_client = {
@@ -462,7 +463,7 @@ impl FrontendClient {
                     .do_query(Request::Query(request), ctx.clone())
                     .await
                     .map(|output| {
-                        wrap_standalone_output_with_terminal_metrics(output, &flow_extensions, &ctx)
+                        wrap_standalone_output_with_terminal_metrics(output, &flow_extensions)
                     })
                     .map_err(BoxedError::new)
                     .context(ExternalSnafu)
@@ -570,7 +571,6 @@ fn build_flow_extensions(extensions: &[(&str, &str)]) -> Result<FlowQueryExtensi
 fn wrap_standalone_output_with_terminal_metrics(
     output: Output,
     flow_extensions: &FlowQueryExtensions,
-    query_ctx: &QueryContextRef,
 ) -> OutputWithMetrics {
     let should_collect_region_watermark = flow_extensions.should_collect_region_watermark();
     let terminal_metrics =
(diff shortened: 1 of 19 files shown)


evenyag:
@codex review

evenyag (approved):
😵

src/flow/src/batching_mode/task/ckpt.rs
```diff
@@ -24,73 +24,185 @@ use crate::batching_mode::checkpoint::{
...
+                    }
+
+                    if state.is_incremental_disabled() {
+                        return FlowCheckpointDecision::FallbackToFullSnapshot {
+                            previous_mode: CheckpointMode::FullSnapshot,
+                            reason: FlowQueryFallbackReason::IncrementalDisabled,
+                        };
+                    }
+
+                    if state.dirty_time_windows.is_empty() {
```
fengjiachun, line 123: Could this still advance past rows that were committed before the source scan opens but whose dirty-window notification has not reached `dirty_time_windows` yet? In that case this scoped query filters those rows out by timestamp, the returned watermark can still include them, and `advance_checkpoints(watermark_map)` makes the next incremental scan start after them. The old unfiltered seeding path avoided this race by not checkpointing from a dirty-window-filtered full snapshot. I think this needs to either keep unfiltered seeding for checkpoint advancement, or add a synchronization/fence that proves no dirty signal `<= H` can still arrive before advancing.

discord9: Yes, this is a real theoretical race. Today we do not have an ordering guarantee that a dirty-window notification for a committed row must reach `dirty_time_windows` before that row becomes visible to the scoped repair scan, nor a fence proving that all dirty signals `<= H` have been delivered before advancing checkpoints.

This PR is based on the same best-effort dirty-window assumption: scoped repair is meant to reduce repair cost for the normal case where dirty notifications arrive promptly. If we need a strict correctness guarantee here, we would need either to keep unfiltered seeding for checkpoint advancement, or introduce an explicit dirty-notification/source-snapshot synchronization fence.

I think a real strict fix would require a larger protocol change: the insert/write path would need to return the committed source seqs, and the dirty-window notification would need to carry the relevant region seq/seq range. Then Flow could prove whether notifications have covered `<= H` before publishing `H` as checkpoint. Without seq-bearing dirty notifications, the Flow task state machine itself cannot close this race locally.

Given the current scope, I would treat this as a known limitation/tradeoff of this best-effort scoped repair rather than solving it in this PR.

src/flow/src/batching_mode/task/ckpt.rs
```diff
@@ -103,8 +169,48 @@ impl BatchingTask {
...
+                                reason: FlowQueryFallbackReason::IncrementalDisabled,
+                            }
+                        } else {
+                            FlowCheckpointDecision::AdvancedFromFullSnapshot {
+                                participating_regions: participating_region_count,
+                                watermarks: watermark_count,
+                            }
+                        }
+                    } else {
+                        if checkpoint_mode == CheckpointMode::Incremental {
```
evenyag: `UnfilteredFull` is only produced for TQL / eval-interval SQL, both of which preclude incremental mode, so this branch can never observe `checkpoint_mode == Incremental`. If it ever did, the returned `FallbackToFullSnapshot { previous_mode: Incremental, reason: IncompleteRegionWatermark }` matches the guard in `restore_unscoped_dirty_signal_after_successful_incremental_fallback`, which would then restore the same unscoped windows a second time (the restore above already ran). Suggest dropping this dead `if` or replacing it with `debug_assert!(checkpoint_mode != CheckpointMode::Incremental)` to make the invariant explicit and avoid a latent double-restore.

src/flow/src/batching_mode/frontend_client.rs
```diff
@@ -580,7 +582,6 @@ fn wrap_standalone_output_with_terminal_metrics(
                 .plan
                 .clone()
                 .and_then(terminal_recordbatch_metrics_from_plan)
```
evenyag, line 584: Removing the `.or_else(terminal_recordbatch_metrics_from_snapshots)` fallback (and the function) leaves `terminal_recordbatch_metrics_from_plan` as the only watermark source for the standalone path. The updated test asserts `region_watermark_map() == None`, but that's only because the mock returns an `Output` with no plan, so it doesn't actually confirm the real standalone path still produces watermarks. Can you confirm standalone surfaces region watermarks via the plan (from #8279)? Otherwise standalone flows would never advance past full-snapshot. A non-mock standalone test asserting a populated map would close this gap.

PR #8282, opened 2026-06-11
feat: pass Kafka pruned entry id when creating regions

WenyXu:

This PR passes Kafka WAL topic prune state through region WAL options so a newly created region can initialize its metadata/version `flushed_entry_id` correctly.

Main changes:

- Refactor region WAL options from stringly-typed JSON values to typed `WalOptions`, while keeping compatibility with the existing `RegionCreateRequest.options["wal_options"]` JSON string boundary.
- Add `initial_pruned_entry_id: Option<u64>` to `KafkaWalOptions`.
  - This is a create-time hint from metasrv to datanode.
  - It is not the authoritative source of topic prune state.
- When metasrv allocates Kafka WAL options, it reads the selected topic’s current `pruned_entry_id` and includes it in `KafkaWalOptions`.
- When mito creates a region, it initializes `flushed_entry_id` with:

  ```rust
  max(topic_latest_entry_id, initial_pruned_entry_id.unwrap_or(0))
  This preserves the existing behavior of considering the latest WAL entry while also ensuring the region does not start below the topic’s pruned boundary.


### Compatibility note:

This PR changes the persisted/instruction representation of `region_wal_options` from the legacy map value form:

```json
{
  "region_wal_options": {
    "1": "{\"wal.provider\":\"raft_engine\"}",
    "2": "{\"wal.provider\":\"kafka\",\"wal.kafka.topic\":\"topic_a\"}"
  }
}
```
to structured JSON objects:
```json
{
  "region_wal_options": {
    "1": { "wal.provider": "raft_engine" },
    "2": { "wal.provider": "kafka", "wal.kafka.topic": "topic_a" }
  }
}
```

The new code can read both the legacy string form and the structured form. However, old binaries only understand the legacy HashMap<RegionNumber, String> format. Therefore, after upgrading to this version, if a new metasrv creates/updates table metadata or sends open-region instructions with structured region_wal_options, rolling back to an older binary may fail to read those metadata records/messages.

## PR Checklist

### DIFF
--- a/src/common/meta/src/ddl/create_table.rs
+++ b/src/common/meta/src/ddl/create_table.rs
@@ -15,8 +15,6 @@
 pub mod executor;
 pub mod template;
 
-use std::collections::BTreeSet;
-
 use api::v1::CreateTableExpr;
 use async_trait::async_trait;
 use common_error::ext::BoxedError;
@@ -26,7 +24,6 @@ use common_procedure::error::{
 use common_procedure::local::DynamicKeyLockGuard;
 use common_procedure::{Context as ProcedureContext, LockKey, Procedure, ProcedureId, Status};
 use common_telemetry::info;
-use common_wal::options::WalOptions;
 use serde::{Deserialize, Serialize};
 use snafu::{OptionExt, ResultExt};
 use store_api::metadata::ColumnMetadata;
@@ -42,14 +39,16 @@ use crate::ddl::utils::map_to_procedure_error;
 use crate::ddl::{DdlContext, TableMetadata};
 use crate::error::{self, Result};
 use crate::key::table_route::PhysicalTableRouteValue;
-use crate::key::topic_name::TopicNameKey;
-use crate::lock_key::{CatalogLock, RemoteWalLock, SchemaLock, TableNameLock};
+use crate::lock_key::{CatalogLock, SchemaLock, TableNameLock};
 use crate::metrics;
 use crate::peer::PeerAllocContext;
 use crate::region_keeper::OperatingRegionGuard;
 use crate::rpc::ddl::{CreateTableTask, QueryContext};
 use crate::rpc::router::{RegionRoute, operating_leader_region_roles};
-use crate::wal_provider::{RegionWalOptions, optional_region_wal_options_serde};
+use crate::wal_provider::{
+    RegionWalOptions, acquire_remote_wal_read_locks, optional_region_wal_options_serde,
+    refresh_initial_pruned_entry_ids,
+};
 
 pub struct CreateTableProcedure {
     pub context: DdlContext,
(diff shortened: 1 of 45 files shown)


evenyag:
@codex review

src/common/meta/src/wal_provider.rs
```diff
@@ -75,19 +129,21 @@ impl WalProvider {
...
                 let options_batch = topic_manager
-                    .select_batch(num_regions)?
+                    .select_batch_with_pruned_entry_id(num_regions)
+                    .await?
                     .into_iter()
-                    .map(|topic| {
+                    .map(|(topic, initial_pruned_entry_id)| {
                         WalOptions::Kafka(KafkaWalOptions {
                             topic: topic.clone(),
+                            initial_pruned_entry_id: Some(initial_pruned_entry_id),
```
fengjiachun: This snapshot can still go stale before the region is actually created and its topic-region metadata is published. `WalPruneProcedure` holds `RemoteWalLock::Write(topic)`, but create-table only holds catalog/schema/table-name locks, so a prune can advance this topic's `pruned_entry_id` after allocation. The new region is also invisible to prune's topic-region scan until metadata is created. Can we hold the per-topic remote-WAL read lock across allocation/create/metadata publish, or refresh the prune boundary immediately before building the create request?

src/common/meta/src/ddl/create_table.rs
```diff
@@ -376,7 +374,7 @@ pub struct CreateTableData {
     /// None stands for not allocated yet.
     pub(crate) table_route: Option<PhysicalTableRouteValue>,
     /// None stands for not allocated yet.
-    pub region_wal_options: Option<HashMap<RegionNumber, String>>,
+    pub region_wal_options: Option<RegionWalOptions>,
```
fengjiachun, line 422: `CreateTableData` is persisted by the procedure framework via `dump()`/`from_json()`. Before this PR, pending procedures stored `region_wal_options` as `Option<HashMap<RegionNumber, String>>`, so legacy procedure JSON contains string values like `"{\"wal.provider\":...}"`. This field now deserializes directly as `Option<RegionWalOptions>` and does not use `region_wal_options_serde`, unlike `OpenRegion` and `RegionInfo`. After upgrade, an in-flight create-table procedure that already passed `on_prepare` can fail recovery when `from_json()` reads the legacy dump. Can we apply the same compatibility serde here?

WenyXu: [4d129b9](https://github.com/GreptimeTeam/greptimedb/pull/8282/commits/4d129b92cb9f95f2f8336eaa484fb7b81fa9bcde)

src/common/meta/src/wal_provider.rs
WenyXu: [ef30cd0](https://github.com/GreptimeTeam/greptimedb/pull/8282/commits/ef30cd0d24ed2ab0a162414c7e6af493a51f2861)

src/common/meta/src/wal_provider.rs
```diff
@@ -85,11 +177,7 @@ impl WalProvider {
                 let options_batch = topic_manager
                     .select_batch(num_regions)?
                     .into_iter()
-                    .map(|topic| {
-                        WalOptions::Kafka(KafkaWalOptions {
-                            topic: topic.clone(),
-                        })
-                    })
+                    .map(|topic| WalOptions::Kafka(KafkaWalOptions::new(topic.clone())))
```
fengjiachun, line 252: Could we keep populating `initial_pruned_entry_id` for allocator-generated Kafka options, or move the create-table refresh into a shared helper? `CreateTableProcedure` refreshes the hint before creating regions, but repartition allocates `WalOptions` via this same allocator and sends them to datanodes directly, so a repartition-created region on a pruned topic can still initialize below the topic pruned boundary.

WenyXu: Good catch [400717c](https://github.com/GreptimeTeam/greptimedb/pull/8282/commits/400717cd5ee73f88a23ce62e90fdda18eb63df94)

fengjiachun (approved):
LGTM

PR #8291, opened 2026-06-12
feat: add repartition column hint

WenyXu:

This PR adds `repartition.column.hint` as a table option for future auto repartition on unpartitioned tables.

The option lets users specify the preferred column for auto repartition. When auto repartition is triggered for an unpartitioned table, this hint should take priority over automatic column inference.

This PR only records and validates the hint in table metadata. It does not trigger repartition by itself.

## PR Checklist

### DIFF
--- a/src/common/meta/src/ddl/alter_table.rs
+++ b/src/common/meta/src/ddl/alter_table.rs
@@ -127,7 +127,7 @@ impl AlterTableProcedure {
 
         // Safety: Checked in `AlterTableProcedure::new`.
         let alter_kind = self.data.task.alter_table.kind.as_ref().unwrap();
-        if matches!(alter_kind, Kind::RenameTable { .. }) || is_metadata_only_alter(alter_kind) {
+        if is_metadata_only_alter(alter_kind) {
             self.data.state = AlterTableState::UpdateMetadata;
         } else {
             self.data.state = AlterTableState::SubmitAlterRegionRequests;
@@ -343,6 +343,7 @@ impl AlterTableProcedure {
 
 fn is_metadata_only_alter(alter_kind: &Kind) -> bool {
     match alter_kind {
+        Kind::RenameTable { .. } => true,
         Kind::SetTableOptions(SetTableOptions { table_options }) => {
             table_options.len() == 1 && table_options[0].key.as_str() == REPARTITION_COLUMN_HINT_KEY
         }
--- a/src/common/meta/src/ddl/alter_table.rs
+++ b/src/common/meta/src/ddl/alter_table.rs
@@ -19,8 +19,8 @@ mod region_request;
 use std::vec;
 
 use api::region::RegionResponse;
-use api::v1::RenameTable;
 use api::v1::alter_table_expr::Kind;
+use api::v1::{RenameTable, SetTableOptions, UnsetTableOptions};
 use async_trait::async_trait;
 use common_error::ext::BoxedError;
 use common_procedure::error::{FromJsonSnafu, Result as ProcedureResult, ToJsonSnafu};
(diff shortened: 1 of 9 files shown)


evenyag:
@codex review

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -126,7 +127,7 @@ impl AlterTableProcedure {
 
         // Safety: Checked in `AlterTableProcedure::new`.
         let alter_kind = self.data.task.alter_table.kind.as_ref().unwrap();
-        if matches!(alter_kind, Kind::RenameTable { .. }) {
+        if matches!(alter_kind, Kind::RenameTable { .. }) || is_metadata_only_alter(alter_kind) {
```
evenyag: Can we treat RenameTable as metadata only alter?

src/operator/src/statement/ddl.rs
```diff
@@ -2338,6 +2346,61 @@ pub fn create_table_info(
...
+            reason: format!("{REPARTITION_COLUMN_HINT_KEY} expects exactly one column name"),
+        }
+    );
+
+    ensure!(
+        !column_name.contains(','),
+        InvalidPartitionRuleSnafu {
+            reason: format!("{REPARTITION_COLUMN_HINT_KEY} expects exactly one column name"),
+        }
+    );
```
evenyag, line 2375: Will we support multiple columns in the future?

src/table/src/requests.rs
```diff
@@ -307,6 +309,10 @@ pub enum AlterKind {
     UnsetTableOptions {
         keys: Vec<UnsetRegionOption>,
     },
+    SetRepartitionColumnHint {
+        column_name: String,
+    },
```
evenyag, line 314: Should we use a vec for future extension?

src/table/src/metadata.rs
```diff
@@ -387,6 +391,72 @@ impl TableMeta {
...
+            ensure!(
+                column_index != time_index,
+                error::InvalidAlterRequestSnafu {
+                    table: table_name,
+                    err: format!(
+                        "cannot set {REPARTITION_COLUMN_HINT_KEY} to the time index column"
+                    ),
+                }
+            );
+        }
```
evenyag, line 436: Do we need to check the column data type?

src/table/src/metadata.rs
```diff
@@ -387,6 +391,72 @@ impl TableMeta {
...
+
+        let mut new_options = self.options.clone();
+        new_options.extra_options.insert(
+            REPARTITION_COLUMN_HINT_KEY.to_string(),
+            column_name.to_string(),
+        );
```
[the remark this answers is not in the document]
WenyXu: Partition columns are already protected from DROP COLUMN. repartition.column.hint is only an advisory option for future auto-repartition, and downstream auto-repartition should validate the column existence before using it. A stale hint should be treated as invalid and ignored there.

src/operator/src/statement/ddl.rs
WenyXu: Maybe, but this PR intentionally keeps the option single-column only. The current option name is singular (`repartition.column.hint`) and the auto-repartition strategy only needs one preferred column for now. If we need multi-column hints later, we can extend the option semantics then.

src/table/src/requests.rs
WenyXu: I'd keep `String` for now because the current API only supports one hint column. Using `Vec<String>` now may imply multi-column support before the selection/repartition semantics are defined. We can change it to a vec or add a new variant when multi-column hints are actually supported.

src/table/src/metadata.rs
WenyXu: I don't think we need an extra data type check here. This PR only records an advisory hint and validates that it refers to an existing non-time-index column. Explicit `PARTITION ON COLUMNS` also does not have a separate partition-column data type allowlist today; it validates column existence and partition expressions. If a future auto-repartition strategy has algorithm-specific type restrictions, it should validate them when consuming the hint.

src/common/meta/src/ddl/alter_table.rs
WenyXu: Good catch [0f5c26d](https://github.com/GreptimeTeam/greptimedb/pull/8291/commits/0f5c26d49ec7782e1e5530e181403d9e37246f15)

evenyag (approved):

PR #8304, opened 2026-06-16
feat: support SCRAM auth for Postgres

killme2008:

Implement SCRAM-SHA-256 authentication for the PostgreSQL protocol, so Postgres clients can authenticate without sending the password in cleartext.

**What changed**

- `src/auth`: add `PgScramSha256Verifier` (`common.rs`) holding `iterations`, `salt`, `StoredKey`, `ServerKey`, with a `pg_scram_sha256:<iterations>:<salt>:<stored_key>:<server_key>` credential format and a `format_pg_scram_sha256_password_verifier` helper. `Debug` redacts all secret fields and all comparisons are constant-time.
- `UserProvider::postgres_auth_info()`: a new method that decides, per connection, whether to offer SCRAM or fall back to cleartext. Implemented for the static and watch-file providers; the default impl returns `Cleartext` so external providers keep working.
- Credential conversion: a stored `plain:` credential derives a verifier on the fly, and a `pbkdf2_sha256:` credential reuses its hash directly as the SCRAM `SaltedPassword` (both are PBKDF2-HMAC-SHA256 / 32 bytes), so existing users can authenticate via SCRAM without re-hashing.
- `src/servers/src/postgres/auth_handler.rs`: implement the server side of the SASL exchange (`SASL` → `SASLContinue` → `SASLFinal`) with a per-connection auth state machine, client-first / client-final parsing, and client-proof verification.

**How it works**

PostgreSQL negotiates a single auth method at connection startup. To avoid leaking user or verifier existence, the server offers SCRAM only when **every** configured user can do SCRAM (i.e. none is `mysql_native_password`); otherwise it falls back to cleartext for all users. Unknown usernames are answered with a random throwaway verifier and still run the full handshake, so a failed login is indistinguishable from a wrong password.

**Limitations**

- Channel binding (`SCRAM-SHA-256-PLUS`) is not supported; the `n`/`y` gs2 flags are accepted and bound across messages, but `authzid` is not.
- When any user is configured with `mysql_native_password`, Postgres auth falls back to cleartext for the whole instance.

**Compatibility**

- No API or data-format breakage. The new credential format is additive; existing `plain:`, `pbkdf2_sha256:`, and `mysql_native_password:` credentials keep working. Cleartext behavior is preserved as a fallback.

## PR Checklist


src/auth/src/user_provider.rs
```diff
@@ -140,9 +157,55 @@ impl PasswordVerifier {
...
     }
 
+    fn supports_pg_scram_sha256(&self) -> bool {
+        !matches!(self, PasswordVerifier::MysqlNativePassword { .. })
+    }
+
+    fn to_pg_scram_sha256_verifier(&self) -> Option<PgScramSha256Verifier> {
+        match self {
+            PasswordVerifier::PlainText(password) => {
+                let salt = rand::random::<[u8; PG_SCRAM_SHA256_KEY_LEN]>();
```
fengjiachun: Plaintext-backed users still look different from unknown users here. For an existing `plain:` or unprefixed credential, this path runs PBKDF2 and generates a fresh `s=` value on every startup, while `mock_for_unknown_user()` is cheap and stable per username. Both the server-first timing and salt are observable before proof verification, so the common `static_user_provider:cmd:user=password` case still exposes a username oracle and lets a guessed valid username force PBKDF2 per connection. Could we either derive/cache a stable plaintext SCRAM verifier per user, or keep plaintext-backed credentials on the cleartext path?

src/auth/src/common.rs
```diff
@@ -200,6 +210,203 @@ pub fn format_pbkdf2_sha256_password_verifier(
...
+
+    /// Builds a throwaway verifier for an unknown user. The salt is derived
+    /// deterministically from the username and a process-wide secret, so the
+    /// SCRAM `server-first-message` (salt + iterations) stays stable per
+    /// username and indistinguishable from a real user across reconnects. No
+    /// PBKDF2 is run (avoids a CPU-exhaustion DoS keyed on unknown usernames),
+    /// and the random keys guarantee the client proof never matches.
+    pub fn mock_for_unknown_user(username: &[u8]) -> Self {
+        Self {
+            iterations: DEFAULT_PBKDF2_SHA256_ITERATIONS,
```
fengjiachun, line 309: `i=` is also observable in the SCRAM server-first message, but the mock verifier always uses `DEFAULT_PBKDF2_SHA256_ITERATIONS` while stored `pbkdf2_sha256` and `pg_scram_sha256` credentials can use any accepted iteration count. A candidate username with a non-default stored verifier is therefore distinguishable from an unknown username before the client proves the password. Could the SCRAM path require a uniform iteration count for all configured users, or derive the mock iteration from provider-wide policy instead of the default?

killme2008:
@fengjiachun @sunng87  Please take a look. It's read to be merged.

config/standalone.example.toml
```diff
@@ -13,8 +13,13 @@ default_column_prefix = "greptime"
...
 ## Examples: "static_user_provider:file:/path/to/users", "static_user_provider:cmd:greptime_user=greptime_pwd"
 ## Password verifier formats: "plain:<password>", "pbkdf2_sha256:<iterations>:<hex_salt>:<hex_hash>",
-## "mysql_native_password:<hex_sha1_sha1_password>"
-## "pbkdf2_sha256" protects passwords at rest but is not compatible with mysql_native_password.
+## "mysql_native_password:<hex_sha1_sha1_password>",
+## "pg_scram_sha256:<iterations>:<hex_salt>:<hex_stored_key>:<hex_server_key>"
+## "pbkdf2_sha256" and "pg_scram_sha256" protect passwords at rest.
+## "mysql_native_password" is MySQL-specific and falls back to cleartext password auth for PostgreSQL.
+## For PostgreSQL SCRAM, keep all users on the default iteration count (4096); a non-default
+## "pbkdf2_sha256"/"pg_scram_sha256" iteration count is observable in the SCRAM handshake and weakens
```
fengjiachun: This no longer matches the implementation: `pbkdf2_sha256` is excluded by `supports_pg_scram_sha256()`, so any such user forces PostgreSQL back to cleartext and its iteration count is never exposed in a SCRAM handshake. The block also dropped the existing warning that hash-only verifiers cannot authenticate via MySQL (`pg_scram_sha256` is rejected there too). Please update both example TOMLs to state the actual per-protocol fallbacks/incompatibilities, then regenerate `config/config.md`.

src/servers/src/postgres/auth_handler.rs
```diff
@@ -205,32 +307,220 @@ impl StartupHandler for PostgresServerHandlerInner {
...
-                let auth_result = self.login_verifier.auth(&login_info, &pwd.password).await;
+                let auth_message = format!(
+                    "{},{},{}",
+                    client_first_bare, server_first, client_final.without_proof
+                );
+                let Ok(client_proof) = BASE64.decode(client_final.proof.as_bytes()) else {
+                    return Ok(PgAuthenticationResult::Failed);
+                };
+                let Ok(Some(server_signature)) =
+                    verifier.verify_client_proof(auth_message.as_bytes(), &client_proof)
```
fengjiachun, line 431: SCRAM proof failures (including unknown users) return `Failed` here without going through the existing `METRIC_AUTH_FAILURE` increments. Once SCRAM is enabled, wrong-password and unknown-user attempts therefore disappear from `greptime_servers_auth_failure_count`. Please record one generic mismatch failure for this path so the metric remains useful without revealing whether the user exists.

fengjiachun (approved):
Rest LGTM

sunng87 (approved):

PR #8312, opened 2026-06-16
test(cli): harden import-v2 schema filter e2e

fengjiachun:
## Summary

- harden the ignored import-v2 schema-filter E2E with deterministic two-chunk data in both schemas
- verify the exported multi-schema snapshot before importing only the selected schema
- import with explicit task parallelism, progress disabled, and a state path outside the snapshot

## Stack

Dependency #8311 has merged; this PR is now rebased onto `main`.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p cli data::export_v2::tests::import_v2_schema_filter_data_e2e --lib`
- `GREPTIME_ADDR=127.0.0.1:4000 cargo test -p cli data::export_v2::tests::import_v2_schema_filter_data_e2e --lib -- --ignored --nocapture`
- `cargo test -p cli data::export_v2::command --lib`
- `cargo test -p cli data::import_v2::command --lib`
- `cargo clippy -p cli --lib -- -D warnings`
- Rebase validation: `cargo test -p cli data::export_v2::tests::import_v2_schema_filter_data_e2e --lib --no-run`
- Rebase validation: `cargo clippy -p cli --lib --tests -- -D warnings`
- `git diff --check`

Refs #7786

Issue #7786, opened 2026-03-10
tracking: split export/import v2 into 5 stacked PRs


fengjiachun:
### What type of enhancement is this?

Tech debt reduction

## Summary

This issue tracks the delivery of export/import v2 as a stack of 5 PRs.

The original implementation is functionally complete, but the full diff is too large to review effectively as a single PR.  
To improve review quality and reduce merge risk, the work is split by capability boundary.

Related design:
- `docs/rfcs/2025-12-30-export-import-v2.md`

## Goals

Deliver export/import v2 with:
- URI-based snapshot storage
- export-v2 create
- import-v2 pipeline
- resume/state hardening
- management commands: `list`, `verify`, `delete`

## PR Plan

- [x] PR1: `feat(cli): introduce export/import v2 foundations` #7785 
- [x] PR2: `feat(cli): add explicit chunking to export-v2 create` #7841 
- [ ] PR3: `feat(cli): implement import-v2 data pipeline` #7898 
- [ ] PR4: `feat(cli): add resume state and shared copy-retry for export/import v2`
- [ ] PR5: `feat(cli): add export-v2 snapshot management commands`

fengjiachun:
Export/import v2 tracking is now complete and the final validation PR has landed.

Implemented and merged:

- RFC/design: #7499
- Foundations: #7785
- Chunked export-v2 create: #7841
- import-v2 data pipeline: #7898
- Resume state and shared copy-retry: #8007, #8070
- Snapshot management:
  - list: #8096
  - verify: #8111
  - delete: #8162
- Follow-up hardening/usability:
  - verify plan/scan/reconcile refactor: #8206
  - streamed verify data listing: #8268
  - shared progress abstraction: #8270
  - import-v2 progress mode/UI: #8283, #8289
  - export-v2 chunk parallelism/progress: #8292, #8294
  - import-v2 task parallelism/state path override: #8300, #8302
- End-to-end and regression coverage:
  - file/local roundtrip hardening: #8310
  - import-v2 resume coverage: #8311
  - schema filter coverage: #8312
  - incomplete snapshot coverage: #8313
  - MinIO/object-store roundtrip coverage: #8314

The original scope in this tracking issue is covered:

- URI-based snapshot storage
- `export-v2 create`
- `import-v2` pipeline
- resume/state hardening
- snapshot `list`, `verify`, and `delete`
- progress and parallelism controls
- local and object-store E2E validation coverage

[...]

PR #8313, opened 2026-06-16
test(cli): harden incomplete import-v2 snapshot e2e

fengjiachun:
## Summary

- harden the ignored incomplete-snapshot E2E with deterministic two-chunk export data
- verify the snapshot succeeds before manifest corruption and fails after marking a chunk failed
- assert import-v2 rejects the corrupted snapshot with an explicit state path outside the snapshot

## Stack

Stacked on #8312 (`feat/import-v2-schema-filter-e2e`). Review this PR against `feat/import-v2-schema-filter-e2e`.

## Validation

- `cargo test -p cli data::export_v2::tests::import_v2_fails_on_incomplete_snapshot_e2e --lib`
- `GREPTIME_ADDR=127.0.0.1:4000 cargo test -p cli data::export_v2::tests::import_v2_fails_on_incomplete_snapshot_e2e --lib -- --ignored --nocapture`
- `cargo test -p cli data::export_v2::command --lib`
- `cargo test -p cli data::import_v2::command --lib`
- `cargo clippy -p cli --lib -- -D warnings`
- `git diff --check`

Refs #7786

### DIFF
--- a/src/cli/src/data/export_v2/tests.rs
+++ b/src/cli/src/data/export_v2/tests.rs
@@ -799,10 +799,15 @@ async fn import_v2_fails_on_incomplete_snapshot_e2e() -> Result<()> {
         .await?;
     database_client
         .sql(
-            "INSERT INTO metrics (ts, host, cpu) VALUES ('2025-01-01T00:00:00Z', 'h1', 1.0)",
+            "INSERT INTO metrics (ts, host, cpu) VALUES \
+             ('2025-01-01T00:00:00Z', 'h1', 1.0), \
+             ('2025-01-01T01:00:00Z', 'h2', 2.0)",
             schema,
         )
         .await?;
+
+    let expected_rows = query_count(&database_client, schema, "metrics").await?;
+    assert_eq!(expected_rows, 2);
 
     let src_dir = tempdir().context(FileIoSnafu)?;
     let src_uri = path_to_uri(src_dir.path())?;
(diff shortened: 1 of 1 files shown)


fengjiachun:
@codex review

fengjiachun:
@codex review

fengjiachun:
@codex review

fengjiachun:
@codex review

fengjiachun:
@codex review

src/cli/src/data/export_v2/tests.rs
```diff
@@ -863,6 +909,10 @@ async fn import_v2_fails_on_incomplete_snapshot_e2e() -> Result<()> {
         &catalog,
         "--schemas",
         schema,
+        "--state-path",
+        &import_state_path,
```
discord9: This gives the test an isolated state path, but the negative assertion below only checks the error text. It would still pass if a regression created the target schema or wrote this import state before returning `Incomplete snapshot`. Could we add post-failure assertions that the target `schema` does not exist and that `import_state_path` was not created? That would lock down the intended preflight/no-side-effects behavior.

fengjiachun:
@codex review

src/cli/src/data/export_v2/tests.rs
fengjiachun: Fixed in d133d54dc9 by keeping the import state path as a PathBuf and adding post-failure assertions that the target schema still does not exist and the import state file was not created.

Verified with:
- cargo fmt --all -- --check
- cargo test -p cli data::export_v2 --lib
- cargo clippy -p cli --lib --tests -- -D warnings
- git diff --check

discord9 (approved):

Issue #8216, opened 2026-06-02
Support Splunk HEC-compatible ingestion endpoint


killme2008:
### What problem does the new feature solve?

Consider adding support for a Splunk HEC-compatible ingestion endpoint in GreptimeDB.

# Motivation

[Splunk HEC](https://dev.splunk.com/enterprise/docs/dataapps/httpeventcollector/) is a common HTTP-based protocol for log and event ingestion. Supporting it would make it easier for users to:

* migrate from Splunk
* reuse existing agents, scripts, and pipelines
* integrate GreptimeDB with tools that already support HEC

### What does the feature do?

Support a basic compatible endpoint such as:
```
/v1/splunk/services/collector/event
```
With:

* token-based authentication
* JSON event ingestion
* batched payloads
* basic field mapping for time, host, source, sourcetype, and event
* pipeline supporting

agrawalx:
Hi @killme2008, I'd like to work on this. 👋

I've gone through the existing log-ingestion endpoints (Loki, Elasticsearch, InfluxDB) and the pipeline machinery. A Splunk HEC endpoint fits the same adapter pattern so I'd like to implement it.

## Proposed scope

- `POST /v1/splunk/services/collector/event` (and accept `/services/collector` as an alias)
- `GET /services/collector/health` 
- Batched JSON events
- HEC-shaped responses, e.g. `{"text":"Success","code":0}`

A typical event I'd handle:

```json
{
  "time": 1686931200,
  "host": "web-01",
  "source": "/var/log/nginx/access.log",
  "sourcetype": "nginx",
  "index": "web_logs",
  "event": "GET /api/users 200 12ms"
}
```

Mapping I have in mind:
- `time` -> timestamp column
- `host` / `source` / `sourcetype` -> tag columns
- `event` -> log field (string, or inferred when it's a nested object)
- `index` -> target table name (falling back to a `?table=` query param, then a default)

I'd plan to follow up with `/services/collector/raw` and indexer acknowledgment (`/ack` + channels) in later PRs

## One question before I start : auth

HEC sends `Authorization: Splunk <token>`, but GreptimeDB authenticates `username:password` against a `UserProvider`. For the first PR I'd like to reuse the existing InfluxDB approach.

```
Authorization: Splunk teamA:secretA   ->   user "teamA", password "secretA"
```

The only cost is operators format the token value as `user:pass` . Another approach could be to have a `WatchFileUserProvider` style user -> token mapping file 

[...]

PR #8321, opened 2026-06-17
feat: Add support for splunk HEC endpoints

agrawalx:

Part of #8216.

Adds a **Splunk HTTP Event Collector (HEC) compatible ingestion endpoint**, so existing
Splunk shippers (Vector's `splunk_hec` sink, the OTel Collector `splunk_hec` exporter,
Fluent Bit, etc.) can send logs to GreptimeDB by just changing the URL and token.

**Endpoints** (base path `/v1/splunk`, same convention as `/v1/loki` and `/v1/influxdb`;
the client appends the collector path):

- `POST /v1/splunk/services/collector/event`  events (aliases: `/services/collector`, `.../event/1.0`)
- `GET  /v1/splunk/services/collector/health` health probe (alias: `.../health/1.0`)

gzip request bodies are handled via `RequestDecompressionLayer` (same as Loki).

**How it works**

- **Parsing.** A HEC body is one or more JSON event objects, either concatenated with
  any/no separator or wrapped in a top-level array. A streaming deserializer handles both
  and flattens to a single list of events.
- **Field mapping** per event: `time` -> timestamp, `host`/`source`/`sourcetype` and the
  flat `fields` keys -> tags, `event` (+ any other keys) -> data, `index` ->  table name.
- **Pipeline.** Each event runs through `run_pipeline` with the `greptime_identity`
  default (which flattens a nested `event` object). Overridable per request via the
  `pipeline_name` query param or `x-greptime-pipeline-name` header; a custom pipeline owns
  its own schema.
- **Dynamic tagging.** Identity marks every column `Field`, so we override the metadata
  columns to `Tag` before insert. The insert auto-alter path then adds them to the primary
  key — including brand-new `fields` keys on later writes.
- **Timestamps.** HEC `time` is epoch seconds (optionally fractional); values past ~`1e12`
  are treated as milliseconds; missing/invalid falls back to ingest time.
- **Index -> table name.** Sanitized to a valid table name against the existing
  `NAME_PATTERN` (keeps `-`/`.`/`:`, replaces invalid chars with `_`, fixes the first char);
  empty falls back to `?table=`, then the default `splunk_logs`.
- **Auth.** HEC sends `Authorization: Splunk <token>`; we map the token as `user:pass`
  (split on `:`), reusing the existing InfluxDB `Token`-scheme approach (`Basic` also
  accepted). 
- **HEC response codes:** `0` success, `2` token required, `4` invalid token, `5` no data,
  `6` invalid data, `7` incorrect index, `8` internal; HTTP status mapped from the error.

**Testing**

- Unit tests for the parser (both batch forms), timestamp formats, index sanitization,
  field/tag mapping, and the `Splunk` auth scheme.
- A parser test built from **real captured payloads** of two clients - Vector's
  `splunk_hec` sink and the OTel Collector `splunk_hec` exporter (run in Docker against a
  recording server). They differ on separator, `event` type, and target path, and both
  parse correctly validating the format against real wire output.
- An integration test: real instance + auth -> POST a HEC batch -> query back and assert
  values, assert `host`/`source`/`sourcetype`/`fields` are tags, assert a new
  `fields` key on a later write becomes a tag (dynamic), and assert an invalid `?table=`
  returns code `7`.

**Limitations / out of scope (follow-ups)**

- `/services/collector/raw` (raw-text events).
- Indexer acknowledgment (`X-Splunk-Request-Channel`, `/services/collector/ack`).
- Opaque-token -> identity mapping. Real clients default to a single opaque token (no
  colon), so for now operators must format the token as `user:pass`, or auth fails.

**Compatibility:** purely additive, a new endpoint and new auto-created tables; no
changes to existing APIs, schemas, or data.

## PR Checklist

### DIFF
--- a/src/servers/src/http/authorize.rs
+++ b/src/servers/src/http/authorize.rs
@@ -148,7 +148,11 @@ fn splunk_hec_err(status: StatusCode, code: u32) -> Response {
         4 => "Invalid token",
         _ => "Unauthorized",
     };
-    (status, axum::Json(serde_json::json!({ "text": text, "code": code }))).into_response()
+    (
+        status,
+        axum::Json(serde_json::json!({ "text": text, "code": code })),
+    )
+        .into_response()
 }
 
 fn err_response(err: impl ErrorExt) -> Response {
--- a/src/servers/src/http/authorize.rs
+++ b/src/servers/src/http/authorize.rs
@@ -41,6 +41,7 @@ use crate::http::result::error_result::ErrorResponse;
 use crate::http::splunk::is_splunk_request;
 use crate::http::{AUTHORIZATION_HEADER, HTTP_API_PREFIX, PUBLIC_API_PREFIX};
 use crate::influxdb::{is_influxdb_request, is_influxdb_v2_request};
+
 /// AuthState is a holder state for [`UserProviderRef`]
 /// during [`check_http_auth`] function in axum's middleware
 #[derive(Clone)]
@@ -69,6 +70,7 @@ pub async fn inner_auth<B>(
 
     let query_ctx = query_ctx_builder.build();
     let need_auth = need_auth(&req);
+    let is_splunk = is_splunk_request(&req);
 
     // 2. check if auth is needed
     let user_provider = if let Some(user_provider) = user_provider.filter(|_| need_auth) {
@@ -87,6 +89,14 @@ pub async fn inner_auth<B>(
             crate::metrics::METRIC_AUTH_FAILURE
                 .with_label_values(&[e.status_code().as_ref()])
                 .inc();
+            if is_splunk {
+                // HEC: missing header -> 2 ("token is required"), else 4 ("invalid token").
+                let (status, code) = match &e {
+                    error::Error::NotFoundAuthHeader { .. } => (StatusCode::UNAUTHORIZED, 2),
+                    _ => (StatusCode::FORBIDDEN, 4),
+                };
+                return Err(splunk_hec_err(status, code));
+            }
             return Err(err_response(e));
         }
     };
@@ -111,6 +121,10 @@ pub async fn inner_auth<B>(
             crate::metrics::METRIC_AUTH_FAILURE
                 .with_label_values(&[e.status_code().as_ref()])
                 .inc();
+            // HEC: bad credentials -> 4 ("invalid token", 403).
+            if is_splunk {
+                return Err(splunk_hec_err(StatusCode::FORBIDDEN, 4));
+            }
             Err(err_response(e))
         }
     }
@@ -125,6 +139,16 @@ pub async fn check_http_auth(
         Ok(req) => next.run(req).await,
         Err(resp) => resp,
     }
+}
+
+/// HEC-shaped auth error (`{"text","code"}`) so Splunk clients can branch on `code`.
+fn splunk_hec_err(status: StatusCode, code: u32) -> Response {
+    let text = match code {
+        2 => "Token is required",
+        4 => "Invalid token",
+        _ => "Unauthorized",
+    };
+    (status, axum::Json(serde_json::json!({ "text": text, "code": code }))).into_response()
 }
 
 fn err_response(err: impl ErrorExt) -> Response {
--- a/src/servers/src/http/authorize.rs
+++ b/src/servers/src/http/authorize.rs
@@ -33,13 +33,14 @@ use snafu::{OptionExt, ResultExt, ensure};
 
 use crate::error::{
     self, InvalidAuthHeaderInvisibleASCIISnafu, InvalidAuthHeaderSnafu, InvalidParameterSnafu,
-    NotFoundInfluxAuthSnafu, Result, UnsupportedAuthSchemeSnafu, UrlDecodeSnafu,
+    NotFoundAuthHeaderSnafu, NotFoundInfluxAuthSnafu, Result, UnsupportedAuthSchemeSnafu,
+    UrlDecodeSnafu,
 };
 use crate::http::header::{GREPTIME_TIMEZONE_HEADER_NAME, GreptimeDbName};
 use crate::http::result::error_result::ErrorResponse;
+use crate::http::splunk::is_splunk_request;
 use crate::http::{AUTHORIZATION_HEADER, HTTP_API_PREFIX, PUBLIC_API_PREFIX};
 use crate::influxdb::{is_influxdb_request, is_influxdb_v2_request};
-
 /// AuthState is a holder state for [`UserProviderRef`]
 /// during [`check_http_auth`] function in axum's middleware
 #[derive(Clone)]
@@ -203,10 +204,33 @@ fn get_influxdb_credentials<B>(request: &Request<B>) -> Result<Option<(Username,
     }
 }
 
+fn get_splunk_credentials<B>(request: &Request<B>) -> Result<Option<(Username, Password)>> {
+    let Some(header) = request.headers().get(http::header::AUTHORIZATION) else {
+        return Ok(None);
+    };
+    let (auth_scheme, credential) = header
+        .to_str()
+        .context(InvalidAuthHeaderInvisibleASCIISnafu)?
+        .split_once(' ')
+        .context(InvalidAuthHeaderSnafu)?;
+
+    let (username, password) = match auth_scheme.to_lowercase().as_str() {
+        "splunk" => {
+            let (u, p) = credential.split_once(':').context(InvalidAuthHeaderSnafu)?;
+            (u.to_string(), p.to_string().into())
+        }
+        "basic" => decode_basic(credential)?,
+        _ => UnsupportedAuthSchemeSnafu { name: auth_scheme }.fail()?,
+    };
+    Ok(Some((username, password)))
+}
+
 pub fn extract_username_and_password<B>(request: &Request<B>) -> Result<(Username, Password)> {
     Ok(if is_influxdb_request(request) {
         // compatible with influxdb auth
         get_influxdb_credentials(request)?.context(NotFoundInfluxAuthSnafu)?
+    } else if is_splunk_request(request) {
+        get_splunk_credentials(request)?.context(NotFoundAuthHeaderSnafu)?
     } else {
         // normal http auth
         let scheme = auth_header(request)?;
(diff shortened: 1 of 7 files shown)


src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,691 @@
...
+        _ => return None,
+    };
+    const MILLIS_THRESHOLD: f64 = 1e12;
+    let nanos = if n >= MILLIS_THRESHOLD {
+        (n * 1000000.0) as i64
+    } else {
+        (n * 1000000000.0) as i64
+    };
+    Some(DateTime::from_timestamp_nanos(nanos))
+}
```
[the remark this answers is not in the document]
agrawalx: Resolved this in next commit. the suggestion is applied directly.

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,691 @@
...
+pub(crate) const SPLUNK_API_PATH_NAME: &str = "/v1/splunk";
+/// HEC response code for a healthy collector. Splunk returns
+/// `{"text":"HEC is healthy","code":17}`.
+const HEC_HEALTHY_CODE: u32 = 17;
+
+/// HEC response body `{"text", "code"}`; clients branch on `code`.
+/// `use<>`: the response owns `text`, so capture no lifetimes.
+fn hec_response(status: StatusCode, code: u32, text: &str) -> impl IntoResponse + use<> {
+    (status, axum::Json(json!({ "text": text, "code": code })))
+}
```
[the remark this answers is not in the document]
agrawalx: applied.

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,691 @@
...
+            continue;
+        };
+        for col in &mut rows.schema {
+            if tag_columns.contains(&col.column_name) {
+                col.semantic_type = SemanticType::Tag as i32;
+            }
+        }
+    }
+    ContextReq::default_opt_with_reqs(reqs)
+}
```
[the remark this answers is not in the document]
agrawalx: applied the fix and added a regression test to exercise exactly this edge case.

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,691 @@
...
+    if !first_ok {
+        out.insert(0, '_');
+    }
+
+    NAME_PATTERN_REG.is_match(&out).then_some(out)
+}
+
+pub(crate) fn is_splunk_request<B>(req: &axum::extract::Request<B>) -> bool {
+    req.uri().path().contains(SPLUNK_API_PATH_NAME)
+}
```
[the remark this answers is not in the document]
agrawalx: Not sure if this is needed

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,691 @@
...
+/// Like `ingest_logs_inner`, but retags metadata columns (identity default) before insert.
+async fn ingest_events(
+    handler: PipelineHandlerRef,
+    pipeline: PipelineDefinition,
+    requests: Vec<PipelineIngestRequest>,
+    query_ctx: QueryContextRef,
+    pipeline_params: GreptimePipelineParams,
+    tag_columns: HashSet<String>,
+    apply_tags: bool,
+) -> Result<HttpResponse> {
```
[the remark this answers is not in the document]
agrawalx: applied

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,691 @@
...
+    let query_table = params.table.as_deref();
+    let mut by_table: HashMap<String, Vec<VrlValue>> = HashMap::new();
+    let mut tag_columns: HashSet<String> = HashSet::new();
+    for event in events {
+        if let Some((table, map, tags)) = hec_event_to_map(event, DEFAULT_SPLUNK_TABLE, query_table)
+        {
+            tag_columns.extend(tags);
+            by_table.entry(table).or_default().push(map);
+        }
+    }
```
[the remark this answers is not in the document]
agrawalx: done.

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,753 @@
...
+        DateTime::from_timestamp(secs, nsecs)
+    }
+}
+
+/// Maps one HEC event to `(table, per-event map, tag names)`: `time`->timestamp,
+/// `index`->table, host/source/sourcetype/`fields`->tags, `event`+rest->data.
+/// `None` if the event isn't a JSON object.
+fn hec_event_to_map(
+    event: VrlValue,
+    default_table: &str,
```
sunng87: Is there a chance we will customize this `default_table`? We can inline `DEFAULT_SPLUNK_TABLE` directly

killme2008:
@agrawalx Thanks. Just got back to work, we'll take a look.

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,753 @@
...
+        }
+    }
+
+    // host / source / sourcetype are tags.
+    for key in ["host", "source", "sourcetype"] {
+        if let Some(v) = obj.remove(key) {
+            tag_names.push(key.to_string());
+            map.insert(KeyString::from(key), v);
+        }
+    }
```
sunng87, line 183: I will suggest to put `host`, `source` and `sourcetype` at the beginning of all tags, followed by those customized `fields`

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,753 @@
...
+            tag_columns.entry(table.clone()).or_default().extend(tags);
+            by_table.entry(table).or_default().push(map);
+        }
+    }
+    let requests: Vec<PipelineIngestRequest> = by_table
+        .into_iter()
+        .map(|(table, values)| PipelineIngestRequest { table, values })
+        .collect();
+
+    // Bad table name (e.g. invalid `?table=`) -> HEC code 7 ("incorrect index").
```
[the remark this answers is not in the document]
agrawalx: added this in next commit.

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,753 @@
...
+#[axum_macros::debug_handler]
+pub async fn handle_event(
+    State(log_state): State<LogState>,
+    Query(params): Query<LogIngesterQueryParams>,
+    Extension(query_ctx): Extension<QueryContext>,
+    headers: HeaderMap,
+    payload: Bytes,
+) -> impl IntoResponse {
+    let events = match parse_hec_events(&payload) {
+        Ok(events) => events,
```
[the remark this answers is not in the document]
agrawalx: added a channel for Splunk

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,753 @@
...
+    if !first_ok {
+        out.insert(0, '_');
+    }
+
+    NAME_PATTERN_REG.is_match(&out).then_some(out)
+}
+
+pub(crate) fn is_splunk_request<B>(req: &axum::extract::Request<B>) -> bool {
+    req.uri().path().contains(SPLUNK_API_PATH_NAME)
+}
```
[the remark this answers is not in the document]
agrawalx: changed now. this is still not totally fixed as something like /v1/splunkijk will still pass.

src/servers/src/http/splunk.rs
agrawalx: yep, addressed this in latest commit. its inline now and removed the default_table argument.

src/servers/src/http/splunk.rs
agrawalx: Is this only for better readability? the `ingest_events` calls `run_pipeline` which returns column name in sorted order as `values_to_row` iterates over ObjectMap(a BTreeMap) . 

If host/source/SourceType has to be added in beginning then it would require a change after calling run_pipeline(). changing order here would get overridden when run_pipeline() is called later

sunng87: It's also for performance consideration. The order of tags will decide the physical order of data in parquet files. So we prefer to put most queried tag in the front.

src/servers/src/http/splunk.rs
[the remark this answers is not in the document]
sunng87: Can we match for `/v1/splunk/` ? I guess all splunk endpoints have their own subpath?

src/servers/src/http/authorize.rs
```diff
@@ -86,6 +89,14 @@ pub async fn inner_auth<B>(
             crate::metrics::METRIC_AUTH_FAILURE
                 .with_label_values(&[e.status_code().as_ref()])
                 .inc();
+            if is_splunk {
```
killme2008: What about moving `is_splunk` into this code block? Only detect it when errors happen.


```suggestion
            if is_splunk_request(&req) {
```

src/servers/src/http/authorize.rs
```diff
@@ -110,6 +121,10 @@ pub async fn inner_auth<B>(
             crate::metrics::METRIC_AUTH_FAILURE
                 .with_label_values(&[e.status_code().as_ref()])
                 .inc();
+            // HEC: bad credentials -> 4 ("invalid token", 403).
+            if is_splunk {
```
killme2008: The same suggestion as above.

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,758 @@
...
+    transform_ndjson_array_factory(values, false)
+}
+
+/// HEC `time`: epoch seconds (optionally fractional); values past ~1e12 are read as
+/// milliseconds. `None` if absent/unparseable (caller falls back to ingest time).
+fn parse_hec_time(value: &VrlValue) -> Option<DateTime<Utc>> {
+    let n: f64 = match value {
+        VrlValue::Integer(i) => *i as f64,
+        VrlValue::Float(f) => f.into_inner(),
+        VrlValue::Bytes(b) => std::str::from_utf8(b).ok()?.trim().parse().ok()?,
```
killme2008, line 81: Is it possible to be a string?

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,758 @@
...
+
+/// Maps one HEC event to `(table, per-event map, tag names)`: `time`->timestamp,
+/// `index`->table, host/source/sourcetype/`fields`->tags, `event`+rest->data.
+/// `None` if the event isn't a JSON object.
+fn hec_event_to_map(
+    event: VrlValue,
+    query_table: Option<&str>,
+) -> Option<(String, VrlValue, Vec<String>)> {
+    let VrlValue::Object(mut obj) = event else {
+        return None;
```
killme2008: Love to add a debug log when it's not a object type.

It's friendly for debugging.

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,758 @@
...
+    };
+
+    // Timestamp: HEC `time` is honored first, else ingest time.
+    let ts = obj
+        .remove("time")
+        .as_ref()
+        .and_then(parse_hec_time)
+        .unwrap_or_else(Utc::now);
```
killme2008, line 148: Should we return an error if time can't be parsed?

Current implementation just use `Utc::now` instead.

tests-integration/tests/http.rs
```diff
@@ -1404,6 +1407,268 @@ pub async fn test_metrics_api(store_type: StorageType) {
...
+        false,
+    )
+    .await;
+    assert_eq!(StatusCode::OK, res.status());
+    assert!(res.text().await.contains("\"code\":0"));
+
+    // 2. Rows landed with the right values (proves route + auth + parse + insert,
+    //    and that the default table name is `splunk_logs`).
+    let rows = get_rows_from_output(
+        &query(
```
killme2008, line 1530: Better to select greptime_timestamp too and check its value.

src/servers/src/http/splunk.rs
```diff
@@ -0,0 +1,758 @@
...
+    // host / source / sourcetype are tags.
+    for key in ["host", "source", "sourcetype"] {
+        if let Some(v) = obj.remove(key) {
+            tag_names.push(key.to_string());
+            map.insert(KeyString::from(key), v);
+        }
+    }
+
+    // `event` and any remaining keys are data columns.
+    for (k, v) in obj {
```
killme2008, line 186: Should we verify that `event` exists and isn't empty?

Looks like splunk will not accept event missing:

* `code 12「Event field is required」` when event is not present
* `code 13「Event field cannot be blank」` when event is empty

killme2008 (commented):
Almostly LGTM

Thanks a lot.

src/servers/src/http/splunk.rs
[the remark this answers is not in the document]
agrawalx: Thank you for this suggestion. I will apply

src/servers/src/http/authorize.rs
agrawalx: sure, I will apply the suggestion

src/servers/src/http/splunk.rs
agrawalx: Thanks, I will add a debug log.

src/servers/src/http/splunk.rs
agrawalx: oh, I will add this. I missed these error codes.

src/servers/src/http/splunk.rs
agrawalx: sure, I will change this to a error when time can't be parsed

src/servers/src/http/splunk.rs
agrawalx: actually the String case is already handled in the `Bytes` arm. This is checked in `parse_time_string_number` test.

tests-integration/tests/http.rs
agrawalx: I added assert for greptime_timestamp and turns out there was a bug. the pipeline overrides the time if a `custom_time_index` is not set and all timestamps were getting overriden to Utc::now. Fixed it now by adding a custom_time_index instead of passing none.

src/servers/src/http/splunk.rs
agrawalx: Hi, I have made the change for this but I am not sure if this was the preferred approach. Can you please take a look at my last commit, the added test for this reordering passes but I Have added a function inside operator crate in `insert.rs` which only runs for `splunk` requests. I think this is a better approach than making changes in my `apply_tags` function as then the reordering would run for every batch and it will be O(rows x column) operation. 

my approach -> run the reorder once at the time of table creation. all the subsequent requests which come will be reordered correctly by the mito engine.

killme2008:
@agrawalx There are some typos

https://github.com/GreptimeTeam/greptimedb/actions/runs/28205630897/job/83573295728?pr=8321

src/servers/src/http/splunk.rs
killme2008: @sunng87 Please take a look. I think it's ready to be merged.

killme2008 (approved):

killme2008:
@agrawalx 
Thanks for your contribution!

PR #8322, opened 2026-06-17
fix: guard structured json alignment

### LINKED ISSUE #8320: json2 flush fails with "Failed to align JSON array, reason: expect struct array" when a JSON column has inconsistent native types across bulk parts
## What type of bug is this?

Crash / panic during flush.

Storage Engine (mito2), JSON v2 data type.

## What happened

Flushing a table with a JSON (json2) column can fail with:

```
Failed to align JSON array, reason: expect struct array
```

The error originates in `JsonArray::try_align` at `src/datatypes/src/vectors/json/array.rs:117-119`, where `self.inner.as_struct_opt()` returns `None`. Because the JSON vector builder unwraps the build result (`to_vector` -> `try_build().unwrap_or_else(|e| panic!(...))`), this surfaces as a hard failure during flush.

## Root cause

A json2 column's **physical** Arrow type depends on the inferred native type (`src/datatypes/src/types/json_type.rs:130` `JsonNativeType::as_arrow_type`):

| JSON native type | Arrow physical type |
|---|---|
| `Object{...}` | `Struct` |
| `Null` | `Null` (NullArray) |
| `Bool` / `Number` | `Boolean` / `Int64` / `UInt64` / `Float64` |
| `String` | `Utf8View` |
| `Array` | `List` |
| `Variant` | `Binary` |

So a JSON column is only a `StructArray` when its inferred native type is `Object`.

`JsonArray::try_align` (`array.rs:106`) only knows how to align **struct -> struct**: it calls `as_struct_opt()` and bails with `"expect struct array"` if the source array is not a struct. It cannot promote a non-struct array (NullArray / scalar / `Variant` Binary) into a merged object schema.

During flush, `align_parts` (`src/mito2/src/memtable/bulk/part.rs:483`) merges each JSON column's `JsonType` across all bulk parts into one unified type (line 500-507), then calls `JsonArray::from(column).try_align(merged.as_arrow_type())` on **every** part's column (line 530-535). When parts infer different native types for the same column, the merged type is typically promoted to `Object` (struct), but at least one part's physical column is still non-struct -> `try_align` fails.

## Reproduce conditions

Same JSON column, across bulk parts (or rows within a flush), inferred as inconsistent native types. Most likely triggers:

1. One part has the column **entirely NULL** -> inferred `JsonNativeType::Null` -> `NullArray`; another part has objects -> merged type `Object` -> aligning the NullArray to the struct target fails.
2. Mixed **scalar and object** JSON values across parts (e.g. one part is a string/number/`true`/`Variant`, another is an object).

To confirm a specific occurrence, enable the `common_telemetry::trace!` at `array.rs:111` ("Try aligning JSON array {from} to {to}") to see the non-struct source physical type.

## Affected versions

- **v1.1.0+** (and nightlies after 2026-04-23, commit `209880b991` "feat: json2 flush (#8011)").
- **Not** affected: v1.0.0 / v1.0.1 / v1.0.2 (released before #8011; do not contain this alignment path).

## Suggested fix directions

1. Make `try_align` handle non-struct sources: at minimum, when the source is a `NullArray`, build an all-null `StructArray` matching the target fields and row count. For scalar/`Variant` sources where the target is a struct, this indicates the merge step should have converged to `Variant` instead.
2. Fix `JsonType::merge` so that irreconcilable native types (`Object` vs non-`Object`) converge to `Variant` (Binary), so the alignment target is not a struct and `try_align` takes the `try_cast` branch.

Needs a closer look at `JsonType::merge`'s rules for `Object` vs `Null` / scalar to decide whether the fix belongs in merge or in `try_align`.

killme2008:

close https://github.com/GreptimeTeam/greptimedb/issues/8320

This PR fixes a compatibility issue for legacy JSON columns created before JSON v2.

Older table metadata may contain the `greptime.json` extension on a legacy JSON column whose physical Arrow type is still `Binary`. The current structured JSON paths only checked the extension metadata, so those legacy JSONB arrays could be treated as structured JSON and sent into JSON alignment. That fails with `expect struct array` because the array is not a `StructArray`.

The change makes structured JSON detection require both conditions:

- the field has the JSON extension metadata
- the physical Arrow type is `Struct`

It applies the guard across read, write, compaction, memtable bulk merge, record batch conversion, and parquet reader/writer paths. It also prevents `SHOW CREATE TABLE` from rendering legacy JSON columns as `JSON(format = 'structured')` only because stale extension metadata exists.

Added regression coverage for:

- legacy JSON with JSON extension metadata not entering structured JSON alignment
- JSON type hints not concretizing legacy JSON columns
- `SHOW CREATE TABLE` rendering legacy JSON as plain `JSON`

## PR Checklist

### DIFF
--- a/src/mito2/src/read/scan_region.rs
+++ b/src/mito2/src/read/scan_region.rs
@@ -31,7 +31,7 @@ use datafusion::physical_plan::expressions::DynamicFilterPhysicalExpr;
 use datafusion_common::Column;
 use datafusion_expr::Expr;
 use datafusion_expr::utils::expr_to_columns;
-use datatypes::schema::ext::ArrowSchemaExt;
+use datatypes::extension::json::is_structured_json_field;
 use datatypes::types::json_type::JsonNativeType;
 use futures::StreamExt;
 use itertools::Itertools;
@@ -443,7 +443,9 @@ impl ScanRegion {
             .metadata
             .schema
             .arrow_schema()
-            .has_json_extension_field()
+            .fields()
+            .iter()
+            .any(is_structured_json_field)
             .then_some(&self.request.json_type_hint)
             .inspect(|json_type_hint| {
                 debug!(
--- a/src/common/recordbatch/src/recordbatch.rs
+++ b/src/common/recordbatch/src/recordbatch.rs
@@ -21,7 +21,7 @@ use datafusion_common::arrow::array::ArrayRef;
 use datafusion_common::arrow::compute;
 use datafusion_common::arrow::datatypes::{DataType as ArrowDataType, SchemaRef as ArrowSchemaRef};
 use datatypes::arrow::array::{Array, AsArray, RecordBatchOptions};
-use datatypes::extension::json::is_json_extension_type;
+use datatypes::extension::json::is_structured_json_field;
 use datatypes::prelude::DataType;
 use datatypes::schema::SchemaRef;
 use datatypes::vectors::json::array::JsonArray;
@@ -359,13 +359,13 @@ fn maybe_align_json_array_with_schema(
     schema: &ArrowSchemaRef,
     arrays: Vec<ArrayRef>,
 ) -> Result<Vec<ArrayRef>> {
-    if schema.fields().iter().all(|f| !is_json_extension_type(f)) {
+    if schema.fields().iter().all(|f| !is_structured_json_field(f)) {
         return Ok(arrays);
     }
 
     let mut aligned = Vec::with_capacity(arrays.len());
     for (field, array) in schema.fields().iter().zip(arrays) {
-        if !is_json_extension_type(field) {
+        if !is_structured_json_field(field) {
             aligned.push(array);
             continue;
         }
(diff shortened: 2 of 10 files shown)


MichaelScofield (approved):

src/mito2/src/read/scan_region.rs
```diff
@@ -443,7 +443,9 @@ impl ScanRegion {
             .metadata
             .schema
             .arrow_schema()
-            .has_json_extension_field()
+            .fields()
+            .iter()
+            .any(is_structured_json_field)
```
fengjiachun, line 448: This guard also needs to be applied before `narrow_read_columns_by_json_type_hint()` mutates `read_cols` above. Legacy JSON columns can still have the `greptime.json` extension, so `json_get(j, "a")` produces a type hint; the unconditional narrowing turns a whole Binary root read into nested paths like `j.a`. Those paths do not match a Binary Parquet column, so the root is marked missing and later filled with nulls. Please only narrow read columns when the table has structured JSON fields, matching the mapper guard here.

MichaelScofield:
superceded by https://github.com/GreptimeTeam/greptimedb/pull/8323 which fixes the clippy by copilot

PR #8333, opened 2026-06-18
refactor(meta): centralize backend retry classification

WenyXu:

This PR centralizes retry hint classification for metadata backend errors.

Previously, `common-meta` and `meta-srv` mostly used coarse retry hints for backend failures. This PR introduces backend-specific retry classifiers for:

- etcd errors
- MySQL/sqlx errors
- PostgreSQL/tokio-postgres errors
- PostgreSQL pool errors

The retry classification now uses structured backend error information where available, such as:

- etcd gRPC status codes
- MySQL error numbers
- PostgreSQL SQLSTATE codes
- SQLx error variants
- deadpool PostgreSQL pool error variants

It also refines transaction retry detection by moving MySQL/PostgreSQL serialization checks into the retry hint modules, so `Error::is_serialization_error()` can reuse the same backend-specific classification helpers instead of duplicating message or SQLSTATE checks.

Additional details:

- `SqlExecutionTimeout` is marked retryable for SQL metadata backends.
- PostgreSQL serialization retry detection includes `serialization_failure` and `deadlock_detected`.
- MySQL serialization retry detection keeps the existing message-based behavior.
- `sqlx::Error::Tls` and `sqlx::Error::Protocol` are treated as non-retryable because SQLx does not expose stable structured details for these variants; TLS errors are often certificate/configuration failures, and protocol errors may indicate protocol mismatch or driver issues.

## PR Checklist

### DIFF
--- a/src/common/meta/src/error/retry_hint/mysql.rs
+++ b/src/common/meta/src/error/retry_hint/mysql.rs
@@ -133,23 +133,38 @@ fn retry_hint_from_mysql_database_error(number: Option<u16>, message: &str) -> R
         _ => {}
     }
 
-    if matches!(
-        message,
-        "Deadlock found when trying to get lock; try restarting transaction"
-            | "can't serialize access for this transaction"
-    ) {
+    if is_mysql_serialization_database_error(message) {
         RetryHint::Retryable
     } else {
         RetryHint::NonRetryable
     }
 }
 
-#[cfg(feature = "mysql_kvbackend")]
+fn is_mysql_serialization_database_error(message: &str) -> bool {
+    matches!(
+        message,
+        "Deadlock found when trying to get lock; try restarting transaction"
+            | "can't serialize access for this transaction"
+    )
+}
+
+pub fn is_mysql_serialization_error(error: &sqlx::Error) -> bool {
+    match error {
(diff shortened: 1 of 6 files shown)


src/common/meta/src/error/retry_hint/mysql.rs
```diff
@@ -0,0 +1,308 @@
...
+
+pub fn is_mysql_serialization_error(error: &sqlx::Error) -> bool {
+    match error {
+        sqlx::Error::Database(error) => {
+            let mysql_error = error
+                .as_error()
+                .downcast_ref::<sqlx::mysql::MySqlDatabaseError>();
+            matches!(
+                mysql_error.map(|error| error.number()),
+                Some(ER_LOCK_DEADLOCK)
```
fengjiachun: Could we include `ER_LOCK_WAIT_TIMEOUT` in this helper too? The retry classifier above marks errno 1205 as retryable, but `RdsStore::txn()` only performs the local transaction retry loop when `is_serialization_error()` returns true. As written, a MySQL lock wait timeout inside a metadata transaction bypasses that retry loop and returns immediately instead of using the configured backoff.

fengjiachun (approved):
Others LGTM

PR #8334, opened 2026-06-18
test: add sqlness compatibility runner

discord9:

Related to #7659.

This PR adds the first skeleton for sqlness-based compatibility testing:

- Add `cargo sqlness compat` to run compatibility cases against a `from` binary and then a `to` binary on preserved local state.
- Add an independent compatibility case format under `tests/compatibility/cases/<case>/`:
  - `case.toml`
  - `setup.sql`
  - `verify.sql`
  - `verify.result` (optional on first run; generated if missing)
- Add initial cases:
  - `basic_table`: basic table create/insert/alter/select compatibility.
  - `legacy_jsonb`: legacy JSONB data compatibility inspired by #8323; it flushes old JSON data and verifies new-version reads using `json_to_string`, typed JSON extraction via `json_get_int`, and `SHOW CREATE TABLE`.
  - `merge_mode`: verifies old-version data written with `merge_mode='last_non_null'` is still merged correctly by the new-version read path, including a filtered scan and persisted table option check.
- Add compat case discovery, metadata validation, namespace isolation, version-range filtering, binary path validation, and an etcd cleanup guard.
- Add same-job bare distributed lifecycle support in the sqlness runner:
  - start old-version distributed cluster
  - run setup SQL
  - restart with new-version binary on preserved state
  - reconnect MySQL/Postgres protocol clients after distributed frontend restart
  - run verify SQL and compare with expected output
- Support ordinary sqlness interceptor comments (`-- SQLNESS ...`) per statement, using the same interceptor registry as the normal runner plus GreptimeDB `PROTOCOL`.
- Use the full `distributed` topology for compat runs, including `flownode`.
- Fix `tests/conf/metasrv-test.toml.template` to use `backend = "etcd_store"`.

Why add a separate compatibility case format instead of reusing ordinary sqlness cases:

- Compatibility cases have a different lifecycle: `setup.sql` runs on the old/from binary, while `verify.sql` runs on the new/to binary against preserved state.
- Ordinary sqlness cases are regression tests for one binary and should not need to carry compatibility metadata, version ranges, old/new phase semantics, or xfail policy.
- Keeping `case.toml` next to `setup.sql`/`verify.sql` makes ownership, version scope, topology scope, and feature tags machine-readable without embedding a large amount of control metadata into SQL comments.
- This avoids making existing sqlness files serve two purposes at once, which would be hard to maintain as the compatibility matrix and generated cases grow.
- A separate corpus also makes future generator/migration tooling easier: tools can produce compact compatibility case directories without rewriting normal sqlness regression files.

Implementation note on sqlness reuse:

- The normal `sqlness::Runner` owns a monolithic lifecycle: environment discovery, case discovery, parse, execute, compare, and teardown.
- Its useful middle-layer types/functions (`TestCase`, `Query`, `from_file`, `execute`, compare) are not public APIs.
- This PR keeps compat's custom old→new lifecycle and implements the minimal per-file execution needed for `setup.sql`/`verify.sql`, while applying the same sqlness interceptors per statement.

Current limitations of this PR1 skeleton:

- No CI workflow is added in this PR.
- No release-version matrix is added in this PR.
- No case generator or migration tooling is added in this PR.
- No cross-job distributed state artifact restore is attempted; distributed compatibility is same-job from→to only.
- Flow-specific semantic compatibility cases are left to a follow-up PR.
- The compat runner does not add compatibility-specific configuration in SQL comments; `-- SQLNESS ...` keeps ordinary sqlness meaning.
- Version-range filtering is intentionally best-effort for local binaries: explicit `--from-version` is used directly, and local/default binaries are inferred via `greptime --version`. Non-wildcard ranges are skipped if the relevant version cannot be determined.

Local validation:

- `cargo test -p sqlness-runner -- cmd::compat`
- `cargo check -p sqlness-runner`
- `cargo check -p query`
- `cargo check -p datanode -p client`
- `cargo clippy -p sqlness-runner --all-targets -- -D warnings`
- Full default-topology smoke with `flownode`:
  - `cargo run -p sqlness-runner -- compat --from-bins-dir /mnt/nvme_rust/rust-targets/compat_framework_retry/debug --to-bins-dir /mnt/nvme_rust/rust-targets/compat_framework_retry/debug --test-filter basic_table`
- Different local debug-binary from→to smoke after the binary-switch fix:
  - from commit: `503f0aac68a2747c346a69ca7f1fc952c87bbfb2`
  - to commit: `e34ace3301eb6b53cb82a02688360d4457a8488d`
  - `cargo run -p sqlness-runner -- compat --from-bins-dir /mnt/nvme_rust/rust-targets/neon-falcon-task04-rdf-decode/debug --to-bins-dir /mnt/nvme_rust/rust-targets/neon-falcon/debug --test-filter basic_table --preserve-state`
  - Preserved logs show the first startup used `503f0aac...` and the restart used `e34ace330...`.
- Manual release old→new validation with downloaded clean release binaries:
  - from: `v1.1.0` (`c9062fe0425a6fa112242067795494c6a297895e`, clean)
  - to: `v1.1.1` (`126c212546dcc16cf595d9b0f0c76887b5e974d6`, clean)
  - `cargo run -p sqlness-runner -- compat --from-bins-dir /tmp/opencode/pr8334-release-verify/v1.1.0 --to-bins-dir /tmp/opencode/pr8334-release-verify/v1.1.1 --test-filter 'basic_table|legacy_jsonb|merge_mode'`
- Version-range skip validation:
  - `cargo run -p sqlness-runner -- compat --from-bins-dir /tmp/opencode/pr8334-release-verify/v1.1.1 --to-bins-dir /tmp/opencode/pr8334-release-verify/v1.1.1 --test-filter legacy_jsonb`
  - This skips `legacy_jsonb` because `from_range = ["<=v1.1.0"]` does not match `v1.1.1`.
- Sqlness interceptor smoke:
  - temporary compat case with `-- SQLNESS SORT_RESULT 1`
  - `cargo run -p sqlness-runner -- compat --case-dir /tmp/opencode/pr8334-sqlness-command-case --from-bins-dir /mnt/nvme_rust/rust-targets/neon-falcon-task04-rdf-decode/debug --to-bins-dir /mnt/nvme_rust/rust-targets/neon-falcon/debug --test-filter sort_result_case`
- Protocol reconnect / missing `verify.result` smoke:
  - temporary compat case with no `verify.result`, `-- SQLNESS PROTOCOL mysql`, and `-- SQLNESS PROTOCOL postgres`
  - `cargo run -p sqlness-runner -- compat --case-dir /tmp/opencode/pr8334-protocol-reconnect-case --from-bins-dir /mnt/nvme_rust/rust-targets/neon-falcon/debug --to-bins-dir /mnt/nvme_rust/rust-targets/neon-falcon/debug --test-filter protocol_reconnect`
  - This generated the initial `verify.result` and verified both protocol clients work after the old→new distributed restart.

Note: an earlier local probe tried `json_get(j, 'a') = 10` to exercise inferred type-hint filtering, but release `v1.1.0` legacy JSONB data returns `NULL` for untyped `json_get(j, 'a')` even in `v1.1.0` self-compat. The committed case uses `json_get_int(j, 'a') = 10` instead so it is a stable old→new compatibility assertion for the released versions. A stricter inferred `json_get` compatibility check can be added later once its expected legacy-JSONB semantics are clarified.

## PR Checklist


MichaelScofield:
@codex review

evenyag:
@codex review

fengjiachun (dismissed):

tests/conf/metasrv-test.toml.template
```diff
@@ -5,7 +5,7 @@ heartbeat_interval = "1s"
 store_addrs = [{store_addrs | unescaped}]
 
 ## The datastore for meta server.
-backend = "EtcdStore"
+backend = "etcd_store"
```
fengjiachun: This template is also used to start the `from` binary. `etcd_store` works for current code, but older releases such as `v0.9.5` deserialize `BackendImpl` without `#[serde(rename_all = "snake_case")]` and expect `EtcdStore`. Since this PR documents `--from-version v0.9.5`, the compat runner can fail before setup by starting the old metasrv with a config value it cannot parse.

Please make this backward-compatible, for example by omitting `backend` and relying on the default etcd backend for this test template, or by rendering a version-aware value for the old/new binary.

tests/runner/src/cmd/compat.rs
```diff
@@ -0,0 +1,716 @@
...
+        let to_bins_dir =
+            resolve_bins(self.to_bins_dir.as_ref(), None, self.pull_version_on_need).await;
+        let to_version = try_infer_version(&to_bins_dir).map(|v| v.to_string());
+        let to_ver_parsed = to_version
+            .as_deref()
+            .and_then(|s| compat_case::Version::parse(s).ok());
+
+        // ---- 5b. Filter by version range ----
+        let pre_filter_count = cases.len();
+        cases.retain(|c| {
```
fengjiachun, line 230: Invalid version constraints should fail validation before filtering. Right now `version_matches_range()` only logs a warning for parse errors and returns `false`; because `validate_cases(&cases)` runs after these retain calls, a selected case with a typo like `from_range = [">=not-a-version"]` is silently filtered out and the runner exits successfully with “No compat cases remaining after version-range filtering”. Please validate the selected metadata before version filtering, or make range matching return an error instead of treating parse errors as non-matches.

tests/compatibility/cases/basic_table/verify.result
```diff
@@ -0,0 +1,13 @@
...
+| ts                  | i | s        | f   |
++---------------------+---+----------+-----+
+| 2024-02-01T17:00:00 | 1 | my_tag_1 |     |
+| 2024-02-01T18:00:00 | 2 | my_tag_2 |     |
+| 2024-02-01T19:00:00 | 3 | my_tag_3 |     |
+| 2024-02-01T20:00:00 | 4 | my_tag_4 | 4.4 |
+| 2024-02-01T21:00:00 | 5 | my_tag_5 | 5.5 |
+| 2024-02-01T22:00:00 | 6 | my_tag_6 | 6.6 |
++---------------------+---+----------+-----+
+
```
fengjiachun: `git diff --check origin/main...HEAD` reports this generated snapshot as having a new blank line at EOF. The same issue appears in `legacy_jsonb/verify.result` and `merge_mode/verify.result`, so this will keep whitespace checks red until the trailing blank lines are removed.

discord9:
Addressed the latest follow-ups in 8f54d85eee:

- Removed the explicit `backend = "etcd_store"` from the metasrv test template so old from-binaries can rely on the default Etcd backend.
- Delayed compat temp dir / `SQLNESS_HOME` creation until after case filtering and validation.
- Made case discovery order deterministic by sorting cases by directory name.
- Made the namespace prelude protocol-aware: `CREATE DATABASE` goes through gRPC, Postgres uses `SET search_path` instead of `USE`; README documents the remaining qualified-table-name expectation for Postgres cases.

discord9:
Addressed the two latest threads in 5c773ead37:

- Split compat validation so selected metadata/version constraints are validated before version filtering, while namespace duplicate checks still run only on the final selected batch.
- Normalized generated `verify.result` output to avoid trailing blank lines at EOF and updated existing snapshots.

Local validation passed: fmt, `cargo test -p sqlness-runner -- cmd::compat` (25 tests), `git diff --check upstream/main...HEAD`, check, clippy, and release `v1.1.0 -> v1.1.1` compat smoke for all committed cases.

discord9:
Follow-up in eac7f71b84:

- Kept `verify.result` authoring-friendly: discovery no longer requires it.
- If `verify.result` is missing, the verify phase writes the actual output and fails once, so authors can review/commit/rerun instead of CI silently passing a missing baseline.
- Kept the tightened metadata behavior: unknown fields such as `isolation` are rejected and namespaces must be unique.

Also filed #8359 for the separate PostgreSQL `pg_*` unqualified table rewrite issue discovered while checking `PROTOCOL POSTGRES` behavior.

fengjiachun (approved):
LGTM

PR #8335, opened 2026-06-18
feat: support structured instruction reply errors

### LINKED ISSUE #8207: Preserve structured error status in instruction replies
Instruction replies currently often use `SimpleReply`, which only carries `result: bool` and an optional stringified error. This loses structured error information such as status code and retryability.

As a result, metasrv procedures may be unable to distinguish permanent failures from transient ones and may retry until deadline instead of aborting early.

This enhancement should make instruction replies preserve structured error metadata so callers can make correct retry/abort decisions.

_Originally posted by @WenyXu in https://github.com/GreptimeTeam/greptimedb/pull/8194#discussion_r3334053697_

WenyXu:

close #8207

This PR introduces structured instruction reply errors and updates procedure-side handling to respect retry hints from instruction errors.

Main changes:

- Add `InstructionError` / `InstructionResult` as structured instruction error representation.
- Add serde helpers for `StatusCode` and `RetryHint` so instruction errors can carry machine-readable error code and retry semantics.
- Refactor datanode instruction replies to return structured errors instead of plain strings.
- Keep legacy string error compatibility by mapping old string errors to `Internal + Retryable`.
- Update metasrv procedures to respect `InstructionError.retry_hint`:
  - retryable instruction errors return retryable procedure errors;
  - non-retryable instruction errors fail fast.
- Preserve existing whole-operation retry behavior; partial retry is intentionally not introduced because these operations are idempotent.

## PR Checklist

### DIFF
--- a/src/common/meta/src/instruction.rs
+++ b/src/common/meta/src/instruction.rs
@@ -16,6 +16,8 @@ use std::collections::{HashMap, HashSet};
 use std::fmt::{Display, Formatter};
 use std::time::Duration;
 
+use common_error::ext::{ErrorExt, RetryHint};
+use common_error::status_code::StatusCode;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use store_api::region_engine::SyncRegionFromRequest;
 use store_api::region_request::{RegionFlushReason, RegionRequirements};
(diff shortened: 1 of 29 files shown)


src/common/meta/src/instruction.rs
```diff
@@ -85,7 +175,7 @@ impl Display for DowngradeRegionReply {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
 pub struct SimpleReply {
     pub result: bool,
-    pub error: Option<String>,
+    pub error: Option<InstructionError>,
```
fengjiachun, line 178: Changing this field from `Option<String>` to `Option<InstructionError>` changes the mailbox reply wire shape for every datanode instruction error: new datanodes will serialize `error` as an object, while older metasrv binaries only know how to deserialize a string here. The manual `Deserialize` impl makes new metasrv accept legacy string errors, but it does not help the reverse direction during mixed-version or rolling upgrades. If component-level compatibility is expected, please keep the old `error: Option<String>` shape and add structured fields alongside it, or gate structured serialization by peer version.

WenyXu: Good point. For operator-managed rolling upgrades, metasrv is upgraded before datanode, so the problematic direction (old metasrv deserializing new datanode replies) should not happen. The manual Deserialize covers the supported mixed-version phase where new metasrv receives legacy string errors from old datanodes.

fengjiachun (approved):
LGTM

PR #8337, opened 2026-06-20
fix: redact Kafka SASL password in debug output

v0y4g3r:

This PR prevents Kafka SASL passwords from being exposed through debug-formatted WAL configuration logs.

- Add a custom `Debug` implementation for `KafkaClientSaslConfig` that preserves usernames and redacts password fields as `<REDACTED>`.
- Cover `MetasrvWalConfig` debug output with a regression test to ensure configured Kafka passwords are not printed.

## PR Checklist

Test plan:

- `rtk cargo test -p common-wal`

### DIFF
--- a/src/common/wal/src/config/kafka/common.rs
+++ b/src/common/wal/src/config/kafka/common.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::fmt;
 use std::io::Cursor;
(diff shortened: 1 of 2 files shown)


src/common/wal/src/config/kafka/common.rs
evenyag, line 72: We can also use SecretString

evenyag (approved):

Issue #6980, opened 2025-09-16
`v1/labels` returns slowly, causing grafana pending


tangyouzzz:
### What type of bug is this?

User Experience

### What subsystems are affected?

Frontend

### Minimal reproduce step

1. Use the grafana prometheus plugin
2. edit dashboard panel
3. When the prometheus plugin edit pane;, it will request the /v1/labels according to the number of query. 
3. When labels are a lot, it is very slow to return(about 2m), which will cause grafana request hangup

### What did you expect to see?

api/v1/labels return to become faster

### What did you see instead?

After looking at the interface logic now is to get all tables, and then loop to get all labels in the table. If I have many tables, this will be slow. Is it possible to query all labels of a specified schema from the `information_schema.columns` table?

### What operating system did you use?

linux

### What version of GreptimeDB did you use?

v0.15.3

### Relevant log output and stack trace

```bash

```

killme2008:
Yes, this is a known issue and will be addressed in the following releases.

We may need to design label storage like that to make it fast.

PR #8341, opened 2026-06-21
fix: stream tables for prometheus label discovery

RitwijParmar:
## What

This changes Prometheus `/api/v1/labels` discovery to iterate `CatalogManager::tables()` instead of first listing table names and then doing one table lookup per name. The KV catalog implementation already batches table metadata through that stream, so the label discovery path can reuse it directly.

## Why

Grafana calls `/api/v1/labels` during metric exploration. With many metric tables, the previous code performed serial catalog table lookups after listing table names, which makes the endpoint scale poorly. This is related to #6980.

## Tests

- `cargo fmt --check`
- `git diff --check`
- `PROTOC=/Users/ritwij/.local/protoc-25.3/bin/protoc CARGO_INCREMENTAL=0 CARGO_BUILD_JOBS=2 RUSTFLAGS='-C debuginfo=0' cargo test -p servers --lib test_get_all_column_names_uses_tag_columns`

### DIFF
--- a/src/servers/src/http/prometheus.rs
+++ b/src/servers/src/http/prometheus.rs
@@ -43,8 +43,8 @@ use datafusion_common::ScalarValue;
 use datatypes::prelude::ConcreteDataType;
 use datatypes::schema::{ColumnSchema, SchemaRef};
 use datatypes::types::jsonb_to_string;
-use futures::StreamExt;
 use futures::future::join_all;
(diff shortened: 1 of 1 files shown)


killme2008 (approved):
LGTM

killme2008:
Good catch! Thank you.

PR #8352, opened 2026-06-23
perf(mito): prune files by manifest time range

discord9:

N/A

This PR adds conservative file-level pruning for Mito scans before loading Parquet metadata.

When scanning an SST, `ScanInput::prune_file()` now first builds lightweight pruning statistics from the manifest-level `FileHandle::time_range()` and evaluates the current predicate snapshot, including dynamic filters. If the predicate proves that the whole SST file cannot match, the scan returns an empty `FileRangeBuilder` before calling `read_sst(...).build_reader_input(...)`, avoiding Parquet metadata loading and row-group pruning for that file.

This is intended to reduce metadata work for `ORDER BY time_index DESC LIMIT ...` / TopK dynamic-filter queries on large append-mode regions. After TopK establishes a timestamp threshold, older SST files can be skipped using manifest time ranges without reading their Parquet row-group metadata.

The pruning is conservative:

- only the time index column exposes file-level min/max/null-count statistics;
- unsupported columns or timestamp conversion failures keep the file;
- existing row-group pruning remains unchanged for files that cannot be skipped at file level.

Validation:

- `cargo test -p mito2 test_file_level_pruning`
- `cargo test -p mito2 scan_region`
- `cargo check -p mito2`

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet/reader.rs
+++ b/src/mito2/src/sst/parquet/reader.rs
@@ -1370,6 +1370,9 @@ fn all_required_row_groups_searched(
 /// Metrics of filtering rows groups and rows.
 #[derive(Debug, Default, Clone)]
 pub(crate) struct ReaderFilterMetrics {
+    /// Number of files filtered by manifest time-range pruning.
+    pub(crate) files_pruned_by_manifest_time_range: usize,
+
     /// Number of row groups before filtering.
     pub(crate) rg_total: usize,
     /// Number of row groups filtered by fulltext index.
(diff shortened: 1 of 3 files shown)


waynexia:
@codex review

discord9:
@codex review

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -1370,6 +1370,9 @@ fn all_required_row_groups_searched(
 /// Metrics of filtering rows groups and rows.
 #[derive(Debug, Default, Clone)]
 pub(crate) struct ReaderFilterMetrics {
+    /// Number of files filtered by manifest time-range pruning.
+    pub(crate) files_pruned_by_manifest_time_range: usize,
```
evenyag: Maybe move this to the end of the struct? Should we rename it to files_dyn_filtered, which is shorter?

discord9:
@codex review

evenyag:
@codex review

src/mito2/src/sst/parquet/reader.rs
discord9: Moved it near the other pruner metrics and shortened the name to `files_time_range_pruned`. I kept `time_range` instead of `dyn` because this counter can be incremented by any predicate evaluated against the manifest time range, not only TopK dynamic filters.

evenyag (approved):

Issue #8340, opened 2026-06-21
Support table-level auto_flush_interval


killme2008:
### What type of enhancement is this?

Configuration / User experience

### What does the enhancement do?

Make `auto_flush_interval` configurable at the table (region) level, in addition to the existing global setting.

Today `auto_flush_interval` is a global Mito engine config (`src/mito2/src/config.rs`, default 30min). It controls how long a region can stay un-flushed before the periodic flush job flushes it (see `flush_periodically` and `flush_regions_on_engine_full` in `src/mito2/src/worker/handle_flush.rs`, both read `self.config.auto_flush_interval`).

A single global value doesn't fit every workload. In practice you often want a different flush cadence for a few specific tables while leaving the rest on the default:

- Low-traffic tables that should flush sooner so data lands in object storage / becomes queryable with bounded latency.
- High-traffic or throwaway tables where you want a longer interval to reduce small-file pressure and compaction load.

Proposed behavior:

- Add an `auto_flush_interval` table option (e.g. `WITH ('auto_flush_interval' = '5m')`), stored in the region options.
- The periodic flush logic uses the per-region value when set, and falls back to the global engine config when unset.
- Setting it via `ALTER TABLE` should take effect without restart, consistent with how other region options behave.

This mirrors options like `ttl`, which already support both a global default and a per-table override.

### Implementation challenges

- Add `auto_flush_interval: Option<Duration>` to `RegionOptions` (`src/mito2/src/region/options.rs`), with parsing/validation consistent with existing duration options.
- In `handle_flush.rs`, resolve the effective interval per region: `region_options.auto_flush_interval.unwrap_or(self.config.auto_flush_interval)` instead of always using the global config.
- Wire the option through SQL table options -> region create/alter requests so it can be set at `CREATE TABLE` and changed via `ALTER TABLE`.
- Make sure the option is round-tripped in `SHOW CREATE TABLE` and covered by sqlness tests.

PR #8357, opened 2026-06-25
feat: support table-level auto_flush_interval

raphaelroshan:

Part of #8340 (the `CREATE TABLE` path; `ALTER TABLE` support to follow).

`auto_flush_interval` is currently a single global Mito engine config, so every region shares one flush cadence. This makes it configurable per table (region), mirroring how `ttl` supports a global default with a per-table override.

- Add `auto_flush_interval: Option<Duration>` to `RegionOptions`, parsed from the table option using the same `humantime` format as the global config.
- The periodic flush logic (`flush_periodically` and `flush_regions_on_engine_full`) resolves the effective interval per region as `region.options.auto_flush_interval.unwrap_or(self.config.auto_flush_interval)`, so unset regions keep the global behavior. Changes take effect on the next flush cycle since the option is read from the region's current version each time.
- Accept the key in `is_mito_engine_option_key` so it can be set at creation, e.g. `CREATE TABLE ... WITH ('auto_flush_interval' = '5m')`. Like other non-`ttl` options it round-trips through the region options map and `SHOW CREATE TABLE`.

Scope: this PR covers setting the option at `CREATE TABLE`. Changing it via `ALTER TABLE` is a natural follow-up (a `SetRegionOption::AutoFlushInterval` variant + the apply path in `handle_alter`); happy to fold it in here if preferred.

## PR Checklist

### DIFF
--- a/src/mito2/src/compaction/window.rs
+++ b/src/mito2/src/compaction/window.rs
@@ -243,6 +243,7 @@ mod tests {
             ssts: Arc::new(ssts),
             options: RegionOptions {
                 ttl: ttl.map(|t| t.into()),
+                auto_flush_interval: None,
                 compaction: Default::default(),
                 compaction_override: false,
                 storage: None,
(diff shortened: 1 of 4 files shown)


evenyag:
@codex review

evenyag (approved):
LGTM. Thank you!

PR #8366, opened 2026-06-26
perf(mito): skip manifest-pruned file ranges

discord9:

Follow-up to #8352.

This PR skips manifest-pruned SST file ranges before entering `scan_flat_file_ranges()` / `PartitionPruner::build_file_ranges()`.

After #8352, manifest time-range pruning can prove some SST files cannot match the current predicate without reading Parquet metadata. However, append-mode scans may already have expanded those SSTs into per-row-group file ranges, and each range would still go through the file-pruner / empty-stream path.

Main changes:

- Reuse manifest-level file pruning via `ScanInput::can_manifest_prune_file()`.
- Move the fast-skip decision into `PartitionPruner`, shared by `UnorderedScan` and `SeqScan` (`SeriesScan` is covered through the SeqScan path).
- Keep compaction out of this fast path.
- Keep manifest-pruned state inside `PrunerInner`, reset it when adding partition ranges, and update it from fast-skip, direct prune, and worker prune paths.
- Do not cache manifest-pruned empty builders; the manifest-pruned flag is enough.
- Keep builder cache metrics balanced when cached builders are created or cleared.

Local synthetic benchmark:

```text
Workload: empty scan where all SST file ranges are manifest-pruned.

Before this PR: ~23.9s
After this PR:  ~1.2s
```

The benchmark uses generated local metadata only and is not included in this PR.

Validation run locally:

```text
cargo fmt -p mito2 -- --check
cargo test -p mito2 --lib --features test pruner::tests::
cargo test -p mito2 --lib --features test test_manifest_pruning_observes_dynamic_filter_update
cargo test -p mito2 --lib --features test test_file_level_pruning
cargo check -p mito2 --features test
```

## PR Checklist


evenyag:
@codex review

src/mito2/src/read/pruner.rs
```diff
@@ -83,6 +88,7 @@ impl PartitionPruner {
             pruner,
             file_indices,
             pre_filter_modes,
+            manifest_pruned_files: (0..num_files).map(|_| AtomicBool::new(false)).collect(),
```
evenyag: We should reset manifest_pruned_files in add_partition_ranges().

src/mito2/src/read/seq_scan.rs
```diff
@@ -665,6 +666,18 @@ pub(crate) async fn build_flat_sources(
...
+            if !compaction
+                && try_skip_manifest_pruned_file_range(
+                    stream_ctx,
+                    *index,
+                    part_metrics,
+                    &partition_pruner,
+                )
+            {
+                continue;
+            }
```
evenyag, line 674: I think we can make try_skip_manifest_pruned_file_range a method of the partition pruner. The pruner actually has the stream ctx as its field.

src/mito2/src/read/pruner.rs
```diff
@@ -466,13 +518,19 @@ impl Pruner {
...
                     let arc_builder = Arc::new(builder);
-                    entry.builder = Some(arc_builder.clone());
-                    PRUNER_ACTIVE_BUILDERS.inc();
+                    let is_background = response_tx.is_none();
+
+                    // Only cache the builder if the file still has remaining ranges.
+                    // If remaining_ranges == 0, a concurrent `skip_file_range` (e.g. from a
+                    // dynamic filter tightening via manifest-prune fast-skip) already consumed
+                    // all ranges and may have cleared a previously cached builder.
+                    let did_cache = cache_builder_if_needed(&mut entry, &arc_builder, &mut metrics);
```
evenyag: Can we check can_manifest_prune_file() and set manifest_pruned_files inside the pruner? (In the worker loop and prune_file_directly()).

This maintains consistency, and we don't need to worry about concurrent `skip_file_range` calls.

src/mito2/src/read/pruner.rs
discord9: Done in 3b2f301e01. Moved the cache into PrunerInner and reset the whole manifest_pruned_files cache at the beginning of add_partition_ranges(), so each fresh scan sees the current dynamic filters.

src/mito2/src/read/seq_scan.rs
discord9: Done in 3b2f301e01. Removed the scan_util helper and made this a PartitionPruner method, so callers no longer need to pass StreamContext explicitly.

src/mito2/src/read/pruner.rs
discord9: Done in 3b2f301e01. The manifest-pruned state now lives in PrunerInner and is checked/updated inside both prune_file_directly() and the worker loop before Parquet metadata reads. Pruner paths call prune_file_after_manifest_check() afterward to avoid a duplicate manifest check.

src/mito2/src/read/pruner.rs
```diff
@@ -117,6 +117,42 @@ impl PartitionPruner {
         Ok(ranges)
     }
 
+    /// Skips a file range that was pruned before entering [`Pruner::build_file_ranges`].
+    #[allow(dead_code)]
+    pub fn skip_file_range(&self, index: RowGroupIndex, reader_metrics: &mut ReaderMetrics) {
+        self.pruner.skip_file_range(index, reader_metrics);
+    }
```
evenyag: Do we use it? Can we avoid `allow(dead_code)`?

src/mito2/src/read/pruner.rs
```diff
@@ -117,6 +117,42 @@ impl PartitionPruner {
...
+    /// Checks whether the file range at `index` can be skipped because the
+    /// current predicate definitively prunes it at manifest level (no I/O).
+    ///
+    /// Returns `true` if the range was skipped. When skipped, this method
+    /// balances the pruner's per-file reference count and merges the resulting
+    /// reader metrics into `part_metrics`.
+    ///
+    /// Uses the shared `PrunerInner::manifest_pruned_files` cache so that a
+    /// single CAS-per-file costs one predicate evaluation regardless of how many
+    /// row groups the file has.
```
evenyag: Is this comment a little verbose?

src/mito2/src/read/scan_region.rs
evenyag, line 1: Can we re-run the benchmark to make sure there is no regression?

evenyag (approved):

src/mito2/src/read/pruner.rs
discord9: Removed the unused `PartitionPruner::skip_file_range()` wrapper, so the `#[allow(dead_code)]` is gone.

src/mito2/src/read/pruner.rs
```diff
@@ -277,6 +364,16 @@ impl Pruner {
...
 
+    /// Skips a file range that has been pruned before entering the file pruner.
+    ///
+    /// This keeps the pruner's per-file reference counts balanced with
+    /// `add_partition_ranges()`. It may also clear a cached builder when this was the
+    /// last remaining range for the file.
+    pub fn skip_file_range(&self, index: RowGroupIndex, reader_metrics: &mut ReaderMetrics) {
+        let file_index = index.index - self.inner.stream_ctx.input.num_memtables();
+        self.decrement_and_maybe_clear(file_index, reader_metrics);
+    }
```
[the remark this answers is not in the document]
discord9: Added a defensive `is_file_range_index()` guard before converting `RowGroupIndex` to file index.

src/mito2/src/read/pruner.rs
```diff
@@ -225,8 +306,14 @@ impl Pruner {
...
     }
 
-    /// Adds reference counts for all partitions' ranges.
+    /// Adds reference counts for all partitions' ranges and resets the
+    /// manifest-prune cache so that dynamic-filter updates are visible to the
+    /// fresh scan.
     pub fn add_partition_ranges(&self, partition_ranges: &[PartitionRange]) {
+        for pruned in &self.inner.manifest_pruned_files {
+            pruned.store(false, Ordering::Relaxed);
+        }
```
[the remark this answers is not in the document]
discord9: Kept the full reset intentionally: `add_partition_ranges()` is the fresh-scan boundary for dynamic-filter updates, and full reset avoids stale positive manifest-prune state from earlier partition batches. I updated the comment to explicitly say full reset so it no longer suggests referenced-file-only reset.

src/mito2/src/read/pruner.rs
discord9: Shortened the comment while keeping the important shared-state behavior.

src/mito2/src/read/scan_region.rs
discord9: Re-ran the local synthetic benchmark on latest head `42eea87a1d6f547041c7eec4bfe1e34070cd641d` using generated local metadata only; the benchmark module remains local and is not included in this PR.

Command shape:
`MITO_EMPTY_PRUNE_FILES=<N> MITO_EMPTY_PRUNE_ROW_GROUPS=100 cargo test -p mito2 --features test bench_unordered_scan_empty_pruned_file_ranges -- --ignored --nocapture`

Results:
- 200 files × 100 row groups = 20,000 file ranges: `total_elapsed = 61.67787ms`, `output_rows = 0`
- 2,000 files × 100 row groups = 200,000 file ranges: `total_elapsed = 409.627239ms`, `output_rows = 0`
- 5,000 files × 100 row groups = 500,000 file ranges: `total_elapsed = 1.015168449s`, `output_rows = 0`

This is in line with the previous post-optimization result (~1.2s on the larger synthetic workload), so I do not see a regression.

discord9: Correction / comparable rerun: I found the original customer-sized benchmark scale. It was release mode with `133,173 files × 10 row groups = 1,331,730 file ranges`, with baseline before optimization around `23.9s` and prior optimized result around `1.27s`.

I reran that exact scale on latest head `42eea87a1d6f547041c7eec4bfe1e34070cd641d` (benchmark module temporarily registered locally only, not included in this PR):

```text
MITO_EMPTY_PRUNE_FILES=133173 MITO_EMPTY_PRUNE_ROW_GROUPS=10 \
  cargo test -p mito2 --lib --release --features test \
  bench_unordered_scan_empty_pruned_file_ranges -- --ignored --nocapture

files:               133173
row_groups_per_file: 10
expected_file_ranges:1331730
new_cost:            84.3082ms
stream_cost:         70.267519ms
drain_cost:          1.073093642s
total_elapsed:       1.227669361s
output_rows:         0
```

So the latest code is consistent with the previous optimized result (`~1.27s`) and shows no regression at the original customer-sized scale.

PR #8375, opened 2026-06-29
refactor: extract region hook function for extension

sunng87:

This is a small refactoring to extract a method for reuse

## PR Checklist

### DIFF
--- a/src/mito2/src/compaction/compactor.rs
+++ b/src/mito2/src/compaction/compactor.rs
@@ -267,16 +267,18 @@ impl CompactionRegion {
             return;
         };
 
-        // Local compaction fills `sst_infos` 1:1 with `files_to_add`. Remote
-        // compaction deserializes `MergeOutput` over gRPC, where `sst_infos`
-        // is `#[serde(skip)]`; substitute empty defaults — the hook still reads
-        // `record_count` from [`FileMeta::num_rows`].
-        let defaults: Vec<SstInfo>;
+        // Remote compaction deserializes `MergeOutput` over gRPC, where
+        // `sst_infos` is `#[serde(skip)]`. Rebuild each `SstInfo` from the
+        // matching `FileMeta` so the hook observes real ids/rows/sizes rather
+        // than zeros — see [`sst_info_from_file_meta`] for what stays default.
+        let synthesized: Vec<SstInfo>;
         let infos: &[SstInfo] = if merge_output.sst_infos.is_empty() {
-            defaults = (0..merge_output.files_to_add.len())
-                .map(|_| SstInfo::default())
+            synthesized = merge_output
+                .files_to_add
+                .iter()
+                .map(sst_info_from_file_meta)
                 .collect();
-            &defaults
+            &synthesized
         } else {
             // `sst_infos` and `files_to_add` are documented as 1:1. If they
             // ever diverge, `zip` below would silently truncate; warn so the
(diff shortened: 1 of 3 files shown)


src/mito2/src/compaction/compactor.rs
```diff
@@ -257,6 +257,42 @@ impl CompactionRegion {
...
+        // `record_count` from [`FileMeta::num_rows`].
+        let defaults: Vec<SstInfo>;
+        let infos: &[SstInfo] = if merge_output.sst_infos.is_empty() {
+            defaults = (0..merge_output.files_to_add.len())
+                .map(|_| SstInfo::default())
+                .collect();
+            &defaults
+        } else {
+            &merge_output.sst_infos
+        };
```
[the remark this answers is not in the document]
shuiyisong: There is a separate correctness issue here: for the remote/deserialized path this builds `SstFileInfo` with real `FileMeta` but `SstInfo::default()`. `SstFileInfo` exposes `sst_info_ref` as part of the hook contract, so hook consumers can observe fake file ids, zero row counts, zero file sizes, and empty index metadata for real SSTs. Deriving the available fields from `FileMeta` would be safer than passing default `SstInfo` values.

src/mito2/src/compaction/compactor.rs
```diff
@@ -257,6 +257,90 @@ impl CompactionRegion {
...
+/// fields that [`FileMeta`] also carries.
+///
+/// Remote compaction runs off-process and ships `MergeOutput` over the wire;
+/// [`MergeOutput::sst_infos`] is `#[serde(skip)]` because the parquet footer
+/// (`ParquetMetaData`) is not serde-serializable. To avoid feeding the hook
+/// default (zero) values for real SSTs, this rebuilds the seven fields `FileMeta`
+/// and `SstInfo` share: `file_id`, `time_range`, `file_size`,
+/// `max_row_group_uncompressed_size`, `num_rows`, `num_row_groups`, `num_series`.
+///
+/// Two fields stay default and **cannot be recovered on the datanode**:
```
shuiyisong, line 321: This branch now fires `on_sst_files_written` for remote compaction with a reconstructed `SstInfo` where `file_metadata` and `index_metadata` stay default. The public hook docs still promise the richest available metadata, including Parquet/index metadata, and say remote compaction hooks run on the compactor node. Hook implementors that rely on those fields can now silently build incomplete artifacts. Please update the hook contract/docs to call out the degraded remote metadata, or avoid firing this hook unless rich metadata is available.

src/mito2/src/compaction/compactor.rs
```diff
@@ -257,6 +257,90 @@ impl CompactionRegion {
...
+                    self.region_id
+                );
+            }
+            &merge_output.sst_infos
+        };
+
+        let files: Vec<SstFileInfo<'_>> = merge_output
+            .files_to_add
+            .iter()
+            .zip(infos)
```
shuiyisong, line 300: This still truncates hook input when `sst_infos` is non-empty but its length differs from `files_to_add`: `zip` drops the extra side, so `on_sst_files_written` can see only a prefix while the later manifest update publishes all files. Since this branch already has the `FileMeta` -> `SstInfo` fallback, on any length mismatch please synthesize all infos from `files_to_add` instead of zipping partial data.

shuiyisong (commented):
Review comments from branch pass.

src/mito2/src/compaction/compactor.rs
sunng87: this should already be guarded from previous step:

1. if sst_info is empty, we will generate infos from files_to_add
2. if sst_info is not empty, we have a check and warning for that case.

src/mito2/src/compaction/compactor.rs
sunng87: comment updated

shuiyisong (approved):

PR #8380, opened 2026-06-30
feat: stream explain analyze metrics over http

discord9:

N/A

This PR adds an experimental HTTP SSE endpoint for long-running `EXPLAIN ANALYZE VERBOSE` queries so clients can observe complete partial metrics snapshots before a query finishes or is canceled.

Summary:

- Adds config-gated route `POST /v1/sql/analyze/stream`, enabled by `http.experimental_enable_explain_analyze_stream` and disabled by default.
- Streams SSE events: `metrics`, `final`, `canceled`, and `error`.
- Requires a single explicit `EXPLAIN ANALYZE VERBOSE` statement. `FORMAT JSON` is accepted but not required; non-JSON formats are rejected.
- Reuses the existing analyze JSON metrics shape and returns final explain rows with the existing HTTP records output shape.
- Adds adaptive/coalesced complete metrics snapshots: large snapshots are sent less frequently, but emitted snapshots are not truncated.
- Cancels the underlying query on SSE stream drop before normal EOF.
- Updates config examples and generated `config/config.md`.
- Adds route-level and helper tests, including:
  - fast final SSE smoke test;
  - delayed-stream test verifying a timer-driven `metrics` event is emitted before `final` when the result stream is pending;
  - route-level negative tests for non-EXPLAIN, non-VERBOSE, non-JSON format, and multi-statement inputs;
  - explicit `FORMAT JSON` route-level acceptance test.

Protocol and limitations:

- This is a POST-only SSE endpoint. Browser clients should use fetch streaming or a compatible library instead of native `EventSource`, which only supports GET.
- `metrics` events are complete snapshots, not deltas. Clients should keep the latest snapshot.
- Live metrics snapshots are best-effort diagnostic snapshots from the running physical plan, not transactionally consistent snapshots. Values may change while metrics are being collected.
- Large metrics snapshots are throttled/coalesced by frequency. They are not truncated.
- `final`, `canceled`, and `error` are terminal events.
- If the client disconnects, it cannot receive a `canceled` event; the server best-effort cancels the underlying query on stream drop.
- No resume, reconnect, or detached query lifecycle is introduced in this PR.

Compatibility:

- Existing `/v1/sql` behavior is unchanged.
- The new API is opt-in behind an experimental config, so existing HTTP API compatibility is preserved.

Validation run locally:

- `cargo fmt --all --check`
- `git diff --check`
- `cargo check -p servers -p frontend -p query -p common-recordbatch`
- `cargo test -p frontend stream_wrapper`
- `cargo test -p frontend test_validate_analyze_stream_statement_strictness`
- `cargo test -p servers --lib test_analyze_stream_route_config_gate`
- `cargo test -p servers --lib test_final_analyze_event_uses_error_event_for_conversion_error`
- `cargo test -p servers --features testing --test mod http_handler_test`
- `cargo clippy -p servers --features testing --test mod -- -D warnings`
- `make config-docs`

## PR Checklist

### DIFF
--- a/src/frontend/src/instance.rs
+++ b/src/frontend/src/instance.rs
@@ -224,34 +224,37 @@ fn validate_analyze_stream_statement(stmt: &mut Statement) -> Result<()> {
 }
 
 impl Instance {
+    fn statement_slow_query_timer(
+        &self,
+        stmt: &Statement,
+        schema_name: String,
+    ) -> Option<SlowQueryTimer> {
+        if !stmt.is_readonly() || !self.slow_query_options.enable {
+            return None;
+        }
+
+        self.event_recorder.clone().map(|event_recorder| {
+            SlowQueryTimer::new(
+                CatalogQueryStatement::Sql(stmt.clone()),
+                schema_name,
+                self.slow_query_options.threshold,
+                self.slow_query_options.sample_ratio,
+                self.slow_query_options.record_type,
+                event_recorder,
+            )
+        })
+    }
+
     async fn query_statement(&self, stmt: Statement, query_ctx: QueryContextRef) -> Result<Output> {
         check_permission(self.plugins.clone(), &stmt, &query_ctx)?;
 
         let query_interceptor = self.plugins.get::<SqlQueryInterceptorRef<Error>>();
         let query_interceptor = query_interceptor.as_ref();
 
-        let is_readonly_stmt = stmt.is_readonly();
         if should_track_statement_process(&stmt) {
             let catalog_name = query_ctx.current_catalog().to_string();
             let schema_name = query_ctx.current_schema();
-            let slow_query_timer = if is_readonly_stmt {
-                self.slow_query_options
-                    .enable
-                    .then(|| self.event_recorder.clone())
-                    .flatten()
-                    .map(|event_recorder| {
-                        SlowQueryTimer::new(
-                            CatalogQueryStatement::Sql(stmt.clone()),
-                            schema_name.clone(),
-                            self.slow_query_options.threshold,
-                            self.slow_query_options.sample_ratio,
-                            self.slow_query_options.record_type,
-                            event_recorder,
-                        )
-                    })
-            } else {
-                None
-            };
+            let slow_query_timer = self.statement_slow_query_timer(&stmt, schema_name.clone());
 
             let ticket = self.process_manager.register_query(
                 catalog_name,
@@ -610,13 +613,16 @@ impl Instance {
             .check_permission(query_ctx.current_user(), PermissionReq::SqlStatement(&stmt))
             .context(PermissionSnafu)?;
         check_permission(self.plugins.clone(), &stmt, &query_ctx)?;
+        let catalog_name = query_ctx.current_catalog().to_string();
+        let schema_name = query_ctx.current_schema();
+        let slow_query_timer = self.statement_slow_query_timer(&stmt, schema_name.clone());
         let ticket = self.process_manager.register_query(
-            query_ctx.current_catalog().to_string(),
-            vec![query_ctx.current_schema()],
+            catalog_name,
+            vec![schema_name],
             stmt.to_string(),
             query_ctx.conn_info().to_string(),
             Some(query_ctx.process_id()),
-            None,
+            slow_query_timer,
         );
         let query_fut =
             self.exec_statement_with_timeout(stmt, query_ctx.clone(), query_interceptor);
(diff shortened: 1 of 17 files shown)


src/servers/src/http/handler.rs
```diff
@@ -155,6 +179,268 @@ pub async fn sql(
...
+        partial,
+        elapsed_ms: start.elapsed().as_millis() as u64,
+        metrics,
+        output,
+        reason,
+        code,
+    };
+    let payload_string = serde_json::to_string(&payload).unwrap_or_else(|e| {
+        format!(r#"{{"seq":{seq},"state":"error","partial":false,"reason":"{e}"}}"#)
+    });
```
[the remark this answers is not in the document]
discord9: Fixed in 2c3e6b2eff: fallback payload now uses serde_json::json! before serialization so error messages are escaped correctly.

src/servers/src/http/handler.rs
```diff
@@ -155,6 +179,268 @@ pub async fn sql(
...
+            mut stream,
+            schema,
+            plan,
+            mut batches,
+            mut seq,
+            start,
+            requested_interval_ms,
+            current_interval_ms,
+            done,
+        )| async move {
```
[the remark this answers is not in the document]
discord9: Fixed in 2c3e6b2eff: replaced the large unfold tuple with an AnalyzeStreamState struct.

src/servers/tests/mod.rs
```diff
@@ -80,6 +80,44 @@ impl SqlQueryHandler for DummyInstance {
...
+        query_ctx.set_explain_format("JSON".to_string());
+
+        self.query_engine
+            .planner()
+            .plan(&QueryStatement::Sql(statement), query_ctx.clone())
+            .and_then(|plan| self.query_engine.execute(plan, query_ctx.clone()))
+            .await
+            .map_err(BoxedError::new)
+            .context(ExecuteQuerySnafu)
+    }
```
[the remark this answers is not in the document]
discord9: Fixed in 2c3e6b2eff: the test DummyInstance now validates single EXPLAIN ANALYZE VERBOSE, rejects non-JSON formats, clears explicit JSON format, and propagates parse errors instead of unwrap().

src/frontend/src/instance.rs
```diff
@@ -549,6 +577,59 @@ fn attach_timeout(output: Output, mut timeout: Duration) -> Result<Output> {
...
+            .check_permission(query_ctx.current_user(), PermissionReq::SqlStatement(&stmt))
+            .context(PermissionSnafu)?;
+        check_permission(self.plugins.clone(), &stmt, &query_ctx)?;
+        let ticket = self.process_manager.register_query(
+            query_ctx.current_catalog().to_string(),
+            vec![query_ctx.current_schema()],
+            stmt.to_string(),
+            query_ctx.conn_info().to_string(),
+            Some(query_ctx.process_id()),
+            None,
```
fengjiachun: Could we pass the same slow-query timer that `query_statement` builds for readonly statements here? This endpoint is specifically for long-running analyze requests, but registering with `None` means slow-query recording will not fire when the SSE stream finishes or is dropped, unlike `/v1/sql` for the same `EXPLAIN ANALYZE` statement.

discord9: Fixed in da882671c2: analyze stream now uses the same readonly SQL slow-query timer helper as query_statement and passes it to register_query, so the timer is kept by the stream Ticket and fires when the SSE stream finishes or is dropped.

src/frontend/src/instance.rs
```diff
@@ -195,32 +195,59 @@ fn parse_stmt(sql: &str, dialect: &(dyn Dialect + Send + Sync)) -> Result<Vec<St
...
+}
+
 impl Instance {
+    fn statement_slow_query_timer(&self, stmt: &Statement) -> Option<SlowQueryTimer> {
+        if !stmt.is_readonly() || !self.slow_query_options.enable {
+            return None;
+        }
+
+        self.event_recorder.clone().map(|event_recorder| {
+            SlowQueryTimer::new(
```
fengjiachun, line 237: This helper needs one more adjustment when rebasing onto current `main`: `SlowQueryTimer::new` now carries the schema name, and merging this head with `origin/main` currently conflicts in `query_statement`. Could we pass the current schema into this helper and use it for both `/v1/sql` and analyze-stream registration, so the new stream path preserves the same slow-query schema context after the rebase?

discord9: Handled in the rebased head a43ce3301a: `statement_slow_query_timer` now takes the current schema name and passes it to `SlowQueryTimer::new`; both `/v1/sql` and analyze-stream registration call the helper with `query_ctx.current_schema()`.

fengjiachun (approved):
LGTM

PR #8382, opened 2026-06-30
feat: persist Prometheus remote write v2 native histograms

shuiyisong:

This pull request adds write-path persistence for Prometheus remote write v2 native histograms.

- Decodes PRW v2 native histograms and converts them into write requests.
- Splits samples and native histograms into separate `ContextReq`s.
- Keeps samples on the existing metric-engine/batcher path.
- Writes native histograms to the same metric table name as one field column: `greptime_native_histogram` Struct.
- Lets native histogram rows use the same metric-engine switch and physical-table path as samples when enabled, while bypassing the pending rows batcher.
- Adds metric-engine support for native histogram logical tables, including Struct field recognition, physical-column alter handling, and sparse/dense batch alignment.
- Stores common scalar fields, spans, buckets, custom values, and separate int/float count families inside the Struct; the populated family is the histogram type signal.
- Adds histogram written headers and `PROM_STORE_REMOTE_WRITE_HISTOGRAMS`.
- Preserves partial-write accounting for v2 error responses.
- Adds module docs, unit tests, HTTP handler coverage, binary fixture coverage, and File-backed integration coverage.

**Note: This PR only covers write path. We'll add query path in a separate PR.**

## Native Histogram Data Flow

```mermaid
flowchart TD
    A["HTTP /v1/prometheus/write"] --> B["remote_write_v2"]
    B --> C["decode_remote_write_v2_request"]
    C --> D["into_write_requests"]

    D --> E["samples ContextReq"]
    D --> F["histograms ContextReq"]

    E --> G["write_prometheus_rows_with_progress"]
    G --> H["metric engine / pending batcher when enabled"]

    F --> I["histogram ContextOpt"]
    I --> J["write_prometheus_rows_with_progress"]
    J --> K["same metric-engine flag as samples, no batcher"]
    K --> L["table: <metric>"]

    L --> M["field: greptime_native_histogram Struct"]
    M --> N["struct children: counts, spans, buckets, sum, schema"]

    H --> P["written headers + counters"]
    K --> P
```

## PR Checklist


src/servers/src/prom_remote_write/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+    H --> P["written headers and counters"]
+    K --> P
+```
+
+The conversion step splits one v2 request into two `ContextReq`s:
+
+- samples keep the existing sample table name and can use the metric-engine
+  physical table path and pending rows batcher;
+- native histograms are written to ordinary `<metric>_native_histogram` tables.
```
sunng87: I'm afraid this will generate a big number of tables which breaks our effort for using metric engine.
Have you considered store native histogram as a blob in metric engine tables? It will be a type of of time-series with binary value type instead of metrics' float value.

waynexia: what about writing naive histogram as is into metric engine

shuiyisong: Native histogram have something like 19 fields if we flat out all the fields in the struct. We may be able to merge some of them but it's still way more then just a f64 field 🤔

shuiyisong: Implemented, PTAL @sunng87 @waynexia

waynexia:
@codex review

src/operator/src/insert.rs
```diff
@@ -915,12 +922,22 @@ impl Inserter {
...
+            let request_is_native_histogram = request_is_native_histogram(request_schema);
+            let table_is_native_histogram = table_is_native_histogram(table);
+            ensure!(
+                request_is_native_histogram == table_is_native_histogram,
+                InvalidInsertRequestSnafu {
+                    reason: format!(
+                        "Table `{table_name}` cannot mix native histogram and float sample fields"
+                    ),
+                }
+            );
```
sunng87, line 934: do we still need this logic?

shuiyisong: Yes, I think we still need it. The v2 converter only rejects sample and histogram payloads for the same metric within one request, while this check protects an existing metric table from changing shape in a later request, from float sample field to native histogram struct or the reverse. It also keeps the Prom insert path error explicit before we build an alter request, so I would keep this guard.

src/common/query/src/native_histogram.rs
```diff
@@ -0,0 +1,106 @@
...
+        ),
+        COUNT_U64_FIELD | ZERO_COUNT_U64_FIELD => Some(ConcreteDataType::uint64_datatype()),
+        POSITIVE_BUCKETS_I64_FIELD | NEGATIVE_BUCKETS_I64_FIELD => Some(
+            ConcreteDataType::list_datatype(Arc::new(ConcreteDataType::int64_datatype())),
+        ),
+        _ => None,
+    }
+}
+
+pub fn native_histogram_value_type() -> ConcreteDataType {
```
sunng87: can we cache this with lazy_static or once_cell

src/metric-engine/src/engine/alter.rs
```diff
@@ -209,6 +213,54 @@ impl MetricEngineInner {
...
+            };
+            let added_fields = columns
+                .iter()
+                .filter(|col| col.column_metadata.semantic_type == SemanticType::Field)
+                .collect::<Vec<_>>();
+            let Some(first_added_field) = added_fields.first() else {
+                continue;
+            };
+            let first_added_field_name =
+                first_added_field.column_metadata.column_schema.name.clone();
```
sunng87: this clone is only used in error

sunng87 (approved):

src/common/query/src/native_histogram.rs
```diff
@@ -0,0 +1,111 @@
...
```
sunng87: nit: return `ConcreteDataType` and clone on demand

PR #8387, opened 2026-06-30
fix: record catalog and schema in slow queries

v0y4g3r:
## Summary
- Add catalog and schema fields to slow query events and the `slow_queries` system table rows.
- Capture query context in `SlowQueryTimer` for SQL, logical plan, and PromQL slow query paths.
- Keep new columns appended after existing slow-query fields and cover them in MySQL/PostgreSQL integration assertions.

## Test Plan
- `make fmt`
- `make clippy`
- `cargo nextest run`
- `make check`
- `cargo test -p tests-integration slow_query`

## Notes
- `make check-udeps` was attempted but could not run because `cargo-udeps` is not installed in this environment.

### DIFF
--- a/src/common/frontend/src/slow_query_event.rs
+++ b/src/common/frontend/src/slow_query_event.rs
@@ -24,6 +24,8 @@ pub const SLOW_QUERY_TABLE_NAME: &str = "slow_queries";
 pub const SLOW_QUERY_TABLE_COST_COLUMN_NAME: &str = "cost";
 pub const SLOW_QUERY_TABLE_THRESHOLD_COLUMN_NAME: &str = "threshold";
 pub const SLOW_QUERY_TABLE_QUERY_COLUMN_NAME: &str = "query";
+pub const SLOW_QUERY_TABLE_CATALOG_NAME_COLUMN_NAME: &str = "catalog_name";
+pub const SLOW_QUERY_TABLE_SCHEMA_NAME_COLUMN_NAME: &str = "schema_name";
 pub const SLOW_QUERY_TABLE_TIMESTAMP_COLUMN_NAME: &str = "timestamp";
 pub const SLOW_QUERY_TABLE_IS_PROMQL_COLUMN_NAME: &str = "is_promql";
 pub const SLOW_QUERY_TABLE_PROMQL_START_COLUMN_NAME: &str = "promql_start";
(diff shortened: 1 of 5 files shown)


src/common/frontend/src/slow_query_event.rs
```diff
@@ -118,6 +134,8 @@ impl Event for SlowQueryEvent {
                 ValueData::U64Value(self.promql_step.unwrap_or(0)).into(),
                 ValueData::TimestampMillisecondValue(self.promql_start.unwrap_or(0)).into(),
                 ValueData::TimestampMillisecondValue(self.promql_end.unwrap_or(0)).into(),
+                ValueData::StringValue(self.catalog_name.to_string()).into(),
+                ValueData::StringValue(self.schema_name.to_string()).into(),
```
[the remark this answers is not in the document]
v0y4g3r: Fixed by using `clone()` for the existing `String` fields.

src/common/frontend/src/slow_query_event.rs
```diff
@@ -104,6 +108,18 @@ impl Event for SlowQueryEvent {
                 semantic_type: SemanticType::Field.into(),
                 ..Default::default()
             },
+            ColumnSchema {
+                column_name: SLOW_QUERY_TABLE_CATALOG_NAME_COLUMN_NAME.to_string(),
+                datatype: ColumnDataType::String.into(),
+                semantic_type: SemanticType::Field.into(),
+                ..Default::default()
+            },
```
sunng87: do we need to persistent the catalog? it should be the catalog of slow query table itself because we don't allow cross-catalog query.

having catalog in the timer makes sense because we use it in the slow log.

src/common/frontend/src/slow_query_event.rs
```diff
@@ -104,6 +108,18 @@ impl Event for SlowQueryEvent {
                 semantic_type: SemanticType::Field.into(),
                 ..Default::default()
             },
+            ColumnSchema {
```
WenyXu: Is it nullable?

v0y4g3r: fixed in c1a56d64f8867d8911cbcac423568364b325c5ba

src/common/frontend/src/slow_query_event.rs
v0y4g3r: removed in c1a56d64f8867d8911cbcac423568364b325c5ba

sunng87 (approved):

PR #8389, opened 2026-07-01
fix(flow): bind scheduled now in dist plan

discord9:

This PR targets `main` and contains the query-side fix plus synthetic analyzer/sqlness regression coverage.

This PR fixes scheduled Flow queries whose distributed remote sub-plans can evaluate `now()` / `current_timestamp()` too early with wall-clock time.

Previously, scheduled Flow execution carried a logical scheduled timestamp in query extensions, but the distributed planner created fresh optimizer contexts while simplifying expressions for remote sub-plans. Those contexts did not know the scheduled timestamp, so `SimplifyExpressions` could fold `now()` into a wall-clock timestamp literal before the plan was sent to remote regions.

The fix:

- Parses and validates `flow.scheduled_time_millis` through the scheduled-time helper path, separate from `FlowQueryExtensions`.
- Carries the scheduled timestamp into DataFusion `ConfigOptions.extensions` as `ScheduledTimeExtension`.
- Makes the distributed planner's patched optimizer context return the scheduled timestamp from `query_execution_start_time()` when present.
- Applies the same scheduled timestamp to both pre-simplification optimizer contexts used by the distributed planner on main.
- Keeps normal non-Flow queries on the existing wall-clock fallback behavior.
- Adds analyzer regression coverage for `now()` and a SQL-level `current_timestamp()` filter folding before remote pushdown.
- Adds synthetic common sqlness coverage under `tests/cases/standalone/common/flow/` for scheduled `now()` / `current_timestamp()` boundary behavior. The distributed suite reaches the same cases through the `distributed/common` symlink.
- Adds a synthetic common sqlness case for late-arriving rows with an overlap lookback window. It uses generic `ts` / `series` / `reading` columns only.

This PR intentionally does not include any customer SQL, customer table schema, or customer-derived sqlness case.

Compatibility:

- No public API break is intended.
- Existing non-Flow query behavior is preserved.
- Malformed `flow.scheduled_time_millis` errors at scheduled-time parsing call sites instead of being silently ignored.

Validation run locally:

- `cargo test -p query scheduled_now_yields_stable_literal_in_remote_plan`
- `cargo test -p query scheduled_current_timestamp_filter_folds_before_remote_pushdown`
- `cargo test -p query scheduled_none_falls_back_to_wall_clock`
- `cargo test -p query scheduled_time`
- `cargo test -p query test_parse_flow_extensions`
- `cargo sqlness bare -t standalone:flow_scheduled_now_boundary -f --bins-dir /mnt/nvme_rust/rust-targets/flow-scheduled-now-main/debug`
- `cargo sqlness bare -t distributed:flow_scheduled_now_boundary -f --bins-dir /mnt/nvme_rust/rust-targets/flow-scheduled-now-main/debug`
- `cargo sqlness bare -t standalone:flow_scheduled_late_overlap -f --bins-dir /mnt/nvme_rust/rust-targets/flow-scheduled-now-main/debug`
- `cargo sqlness bare -t distributed:flow_scheduled_late_overlap -f --bins-dir /mnt/nvme_rust/rust-targets/flow-scheduled-now-main/debug`

Note: the final push used `--no-verify` to avoid rerunning the long pre-push hook after local targeted validation.

## PR Checklist

### DIFF
--- a/src/common/meta/src/ddl/alter_logical_tables.rs
+++ b/src/common/meta/src/ddl/alter_logical_tables.rs
@@ -58,13 +58,13 @@ pub struct AlterLogicalTablesProcedure {
 fn build_validator_from_alter_table_data<'a>(
     data: &'a AlterTablesData,
 ) -> AlterLogicalTableValidator<'a> {
-    let phsycial_table_id = data.physical_table_id;
+    let physical_table_id = data.physical_table_id;
     let alters = data
         .tasks
         .iter()
         .map(|task| &task.alter_table)
         .collect::<Vec<_>>();
-    AlterLogicalTableValidator::new(phsycial_table_id, alters)
+    AlterLogicalTableValidator::new(physical_table_id, alters)
 }
 
 /// Builds the executor from the [`AlterTablesData`].
--- a/src/query/src/datafusion.rs
+++ b/src/query/src/datafusion.rs
@@ -67,6 +67,7 @@ use crate::metrics::{
     OnDone, QUERY_STAGE_ELAPSED, maybe_attach_region_watermark_metrics,
     should_collect_region_watermark_from_query_ctx,
 };
+use crate::options::ScheduledTimeExtension;
 use crate::physical_wrapper::PhysicalPlanWrapperRef;
 use crate::planner::{DfLogicalPlanner, LogicalPlanner};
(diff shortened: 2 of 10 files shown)


evenyag:
@codex review

evenyag (approved):

PR #8390, opened 2026-07-01
fix: invalidate comment DDL cache and lock by object ID

### LINKED ISSUE #8379: Workflow run 'Nightly CI' failed
@GreptimeTeam/db-approver
New failure: https://github.com/GreptimeTeam/greptimedb/actions/runs/28409419536

WenyXu:

close #8379

This PR fixes `COMMENT ON` metadata visibility and locking behavior.

There are two main changes:

1. Invalidate local frontend cache after `COMMENT ON`

   `COMMENT ON TABLE/COLUMN/FLOW` updates metadata through metasrv, but the frontend may receive the cache invalidation message asynchronously. A following statement such as `SHOW CREATE TABLE` can therefore read stale table metadata from the local cache.

   This PR enriches `CommentOnTask` with the target table/flow ID before submitting the DDL task, and invalidates the local frontend cache immediately after the DDL finishes.

2. Use object ID based lock keys for `COMMENT ON` procedures

   `CommentOnProcedure` now includes `TableLock` or `FlowLock` in its lock keys when the target object ID is available, while still keeping the existing name-based locks.

   The task is also enriched on metasrv before constructing the procedure, so the procedure can acquire ID-based locks before `prepare`. For compatibility with older serialized procedures, the procedure still falls back to resolving the table/flow ID by name during `prepare` when the ID is missing.

## PR Checklist

### DIFF
--- a/src/common/meta/src/ddl/alter_logical_tables.rs
+++ b/src/common/meta/src/ddl/alter_logical_tables.rs
@@ -58,13 +58,13 @@ pub struct AlterLogicalTablesProcedure {
 fn build_validator_from_alter_table_data<'a>(
     data: &'a AlterTablesData,
 ) -> AlterLogicalTableValidator<'a> {
-    let phsycial_table_id = data.physical_table_id;
+    let physical_table_id = data.physical_table_id;
     let alters = data
         .tasks
         .iter()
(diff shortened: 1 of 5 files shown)


fengjiachun (approved):
LGTM

PR #8392, opened 2026-07-01
feat(flow): add information_schema.flow_statistics and SHOW FLOW STATUS (#7987)

### LINKED ISSUE #7987: feat: add flow_statistics system table and SHOW FLOW STATUS for flow runtime observability
## Summary

It would be useful to expose flow runtime status through both a system table and a SQL command, so users can quickly inspect flow health and basic runtime statistics.

## Proposed Feature

Add:

- a system table such as `flow_statistics`
- a SQL command such as `SHOW FLOW STATUS`

to expose information like:

- flow start time
- uptime / running duration
- last 3 errors
- total amount of data processed since startup
- if feasible, basic resource usage such as memory and CPU

## Motivation

Today it is hard to quickly understand whether a flow is healthy and how it has behaved since startup. A lightweight built-in status view would make debugging and daily operations much easier.

This could be a good first issue if the initial implementation is scoped to core runtime metadata first, with resource metrics included when feasible.

## Checklist

- [ ] Add a system table (for example `flow_statistics`) that exposes per-flow runtime statistics such as start time, uptime, processed data volume, and recent errors.
- [ ] Add a `SHOW FLOW STATUS` SQL command that provides a user-friendly way to inspect the same runtime status, with optional CPU/memory metrics when available.

onepizzateam:

Closes #7987

Implements the flow runtime observability design approved in #7987: a system table
`information_schema.flow_statistics` plus a `SHOW FLOW STATUS` SQL command, so users
can inspect per-flow runtime status directly through SQL.

**Summary of the change**
- Add the `information_schema.flow_statistics` system table (table id 44) exposing
  per-flow runtime stats with columns: `flow_id`, `flow_name`, `start_time`,
  `last_execution_time`, `uptime_seconds`, `state_size`.
- Add the `SHOW FLOW STATUS [LIKE ...]` statement (parser + AST + Display, operator
  dispatch, frontend permission handling) as a user-friendly view over the same table.
- Track each flow's first-execution `start_time` in both the streaming and batching
  engines and surface it through the existing `FlowStat` pipeline (extends
  `FlowStat`/`FlowStateValue` with a `start_time_map`).
- Add sqlness coverage (`flow/flow_status`) and regenerate the affected golden
  results (`information_schema`, `show_databases_tables`, `view/create`).

**How it works**
`SHOW FLOW STATUS` is resolved to a projection over `information_schema.flow_statistics`.
The table builder joins flow metadata (`flow_name_manager`) with runtime stats obtained
via `information_extension.flow_stats()`. `uptime_seconds` is derived as
`(now - start_time) / 1000`.

**Limitations**
- flow_id and flow_name return real data in both standalone and distributed modes. last_execution_time and state_size reflect a last-reporter snapshot in distributed mode — each flownode heartbeat overwrites the shared __flow/state key, so values show only the most recent reporter and may alternate across nodes. Aggregation across flownodes is tracked in #8454.
- `start_time` / `uptime_seconds` are populated in standalone mode only. In distributed
  mode they return NULL because the cross-node heartbeat wire type
  (`api::v1::meta::FlowStat`) does not yet carry start-time information — consistent with
  the proposal's "handle missing values as NULL". Propagating it through the heartbeat is
  a documented follow-up (`TODO(#7987-followup)`).
- `processed_rows` and `last_errors` from the proposal are intentionally deferred to a
  follow-up PR (incremental delivery per the issue).

-**Follow-up tracked separately:** #8454 

**API / data compatibility**
- No breaking API changes; `SHOW FLOW STATUS` and the new table are additive.
- `FlowStateValue.start_time_map` is annotated `#[serde(default)]`, so existing persisted
  flow-state metadata deserializes unchanged.

**Docs**
- This PR introduces user-facing surfaces (`SHOW FLOW STATUS` and the
  `information_schema.flow_statistics` table) that need documentation on the GreptimeDB
  docs site. I'll open a follow-up docs PR against the docs repo to cover the new command
  and table schema.

## PR Checklist


src/flow/src/adapter/stat.rs
```diff
@@ -43,11 +44,22 @@ impl FlowStatProvider for StreamingEngine {
...
+
+            match worker.get_flow_exec_stats().await {
+                Ok(stats) => {
+                    start_time_map
+                        .extend(stats.start_time_map.into_iter().map(|(k, v)| (k as u32, v)));
+                }
+                Err(err) => {
+                    common_telemetry::error!(err; "Get flow exec stats error");
+                }
+            }
```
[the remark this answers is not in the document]
onepizzateam: agreed this path could be tightened. I'm going to keep it as-is for this PR and handle it as a follow-up, for a couple of reasons:

Combining `get_state_size` / `get_last_exec_time_map` / `get_flow_exec_stats` into a single `QueryFlowExecStats` round-trip means changing the worker message protocol and handle API, which is a bigger change than fits the scope here (this PR is the observability surface, not a stats-pipeline refactor). The concurrent `join_all` fan-out across workers is reasonable on its own, but I'd rather land it together with the request-combining so the loop only gets restructured once.

In practice this only runs when someone queries `flow_statistics`, and the worker count is small, so the sequential overhead is bounded for now. I'll open a follow-up to batch the execution-stat queries into one request and fan out across workers concurrently. Keeping this thread linked for reference.

onepizzateam:
I restored the golden result to match the sqlness test harness: the runner unconditionally appends a double newline after every query result, so removing that blank line caused the Sqlness checks to fail. I reverted the change to keep CI green. Want me to open an issue/PR against the pinned sqlness dependency to discuss removing the extra newline, or should we keep matching the runner?

discord9:
Could we also add `Statement::ShowFlowStatus(_)` to the PostgreSQL extended-query describe path (`src/servers/src/postgres/handler.rs`)? `describe_fields` currently special-cases `SHOW TABLES`/`SHOW FLOWS`/`SHOW VIEWS`, but `SHOW FLOW STATUS` appears to fall through to `NoData` even though execution returns rows. Prepared/extended-protocol clients may not receive the expected row description. It would be good to return the six fields from `information_schema.flow_statistics` here, or derive them from that schema.

onepizzateam:
> Could we also add `Statement::ShowFlowStatus(_)` to the PostgreSQL extended-query describe path (`src/servers/src/postgres/handler.rs`)? `describe_fields` currently special-cases `SHOW TABLES`/`SHOW FLOWS`/`SHOW VIEWS`, but `SHOW FLOW STATUS` appears to fall through to `NoData` even though execution returns rows. Prepared/extended-protocol clients may not receive the expected row description. It would be good to return the six fields from `information_schema.flow_statistics` here, or derive them from that schema.

Pushed a fix for that, added Statement::ShowFlowStatus(_) to describe_fields, returns the 6 flow_statistics fields now (flow_id, flow_name, start_time, last_execution_time, uptime_seconds, state_size). I think that should do it? ptal

onepizzateam:
Filed the remaining scope from the proposal as a tracked follow-up so it's explicit
rather than just prose in the description: #8454 

This PR's scope is now just the standalone-mode core table + SHOW FLOW STATUS, matching
what was approved in #7987. Ready for review whenever you have time
@discord9 @waynexia @MichaelScofield @WenyXu @evenyag

src/servers/src/postgres/handler.rs
```diff
@@ -600,6 +600,53 @@ fn describe_fields(
...
+                None,
+                None,
+                Type::INT8,
+                format.format_for(4),
+            ),
+            FieldInfo::new(
+                "state_size".to_string(),
+                None,
+                None,
+                Type::INT8,
```
fengjiachun: Could we use Type::NUMERIC for state_size here (or derive these fields from InformationSchemaFlowStatistics::schema())? The column is UInt64, and type_gt_to_pg maps UInt64 to NUMERIC. arrow-pg encodes a non-NULL UInt64 as Decimal, so advertising INT8 in the extended-protocol RowDescription can make prepared clients decode the row with the wrong wire type. Streaming flows always populate state_size, so this path is reachable.

src/flow/src/batching_mode/state.rs
```diff
@@ -97,6 +100,9 @@ impl TaskState {
         self.exec_state = ExecState::Idle;
         self.last_query_duration = elapsed;
         self.last_update_time = Instant::now();
+        if self.start_time_millis.is_none() {
```
fengjiachun: Could we record start_time_millis immediately before the first frontend query is dispatched instead of in after_query_exec? For the first long-running batch, start_time remains NULL while it runs and is then set to the completion time, so uptime_seconds permanently excludes the duration of that first execution. after_query_exec should keep the last-execution/completion updates, while the start timestamp should be captured at dispatch.

src/meta-srv/src/handler/flow_state_handler.rs
```diff
@@ -55,7 +55,11 @@ impl HeartbeatHandler for FlowStateHandler {
                 .iter()
                 .map(|(k, v)| (*k, *v))
                 .collect();
-            let value: FlowStateValue = FlowStateValue::new(state_size, last_exec_time_map);
+            // TODO(#7987-followup): start_time_map is not yet propagated through the heartbeat
+            // wire format (`api::v1::meta::FlowStat`); it will always be empty in distributed
+            // mode until a follow-up PR adds heartbeat propagation.
+            let value: FlowStateValue =
+                FlowStateValue::new(state_size, last_exec_time_map, Default::default());
```
discord9, line 62: Could we clarify or track the multi-flownode behavior here? Each flownode heartbeat contains only node-local maps, but this handler writes each report wholesale to the single `__flow/state` key. With multiple flownodes, a later heartbeat replaces the earlier snapshot, so `flow_statistics` can only populate `state_size` and `last_execution_time` from the last reporter, and the visible values may alternate.

This overwrite behavior is pre-existing, but the new table relies on it and the PR body currently says these fields return real distributed-mode data; #8454 only tracks distributed `start_time`/uptime. If aggregation is intentionally out of scope here, could we at least correct that claim and track this snapshot issue in #8454 or a separate follow-up?

src/servers/src/postgres/handler.rs
onepizzateam: fixed, switched to Type::NUMERIC. also considered deriving from InformationSchemaFlowStatistics::schema() but kept it explicit for now since the describe path only needs these six field happy to change if you'd prefer that approach

src/flow/src/batching_mode/state.rs
onepizzateam: moved it to just before the frontend query is dispatched in execute_logical_plan_unlocked, extracted into record_start_time_if_first on TaskState. after_query_exec no longer touches it, so start_time now reflects when the first execution actually began rather than when it completed

src/meta-srv/src/handler/flow_state_handler.rs
onepizzateam: updated the PR description: last_execution_time and state_size are now documented as last-reporter only in distributed mode with the alternating-values caveat. added aggregation as point 4 in #8454. aggregation's out of scope for this PR but tracked now

src/flow/src/batching_mode/state.rs
discord9: Thanks—moving this out of `after_query_exec` fixes the completion-time issue. One edge remains: the current call is before the fallible request construction below, where `encode_insert_plan_request` or Substrait encoding can return via `?`. That can permanently set `start_time` even though no frontend query was dispatched. Could we move it after `req` and `snapshot_seqs` have been built, immediately before `query_with_terminal_metrics`?

tests/cases/standalone/common/system/information_schema.result
```diff
@@ -26,6 +26,7 @@ order by table_schema, table_name;
 |greptime|information_schema|engines|LOCALTEMPORARY|5|0|0|0|0|0||11|Fixed|0|0|0|DATETIME|DATETIME||utf8_bin|0|||Y|
 |greptime|information_schema|events|LOCALTEMPORARY|13|0|0|0|0|0||11|Fixed|0|0|0|DATETIME|DATETIME||utf8_bin|0|||Y|
 |greptime|information_schema|files|LOCALTEMPORARY|14|0|0|0|0|0||11|Fixed|0|0|0|DATETIME|DATETIME||utf8_bin|0|||Y|
+|greptime|information_schema|flow_statistics|LOCALTEMPORARY|44|0|0|0|0|0||11|Fixed|0|0|0|DATETIME|DATETIME||utf8_bin|0|||Y|
```
discord9: `INFORMATION_SCHEMA_FLOW_STATISTICS_TABLE_ID` was bumped to 45 to avoid the new `recycle_bin = 44`, but this generated result still expects `flow_statistics = 44`, so both rows now have ID 44. Could you rerun the corresponding sqlness case to regenerate this file after the ID change? Otherwise this should fail once CI gets past the current compile error.

src/flow/src/batching_mode/state.rs
onepizzateam: moved it past req/snapshot_seqs, right before query_with_terminal_metrics. ptal

onepizzateam:
ptal @discord9 @fengjiachun applied all recommended changes, all checks passing now

discord9 (commented):
Thanks for the PR! As discussed, here's a small review-feedback patch covering two follow-ups:

## 1. Persistence compatibility test
`src/common/meta/src/key/flow/flow_state.rs` — adds two unit tests:
- `test_deserialize_legacy_flow_state_value`: old persisted `FlowStateValue` JSON (only `state_size` + `last_exec_time_map`, no `start_time_map`) still deserializes, with `start_time_map` defaulting to empty via `#[serde(default)]`.
- `test_flow_state_value_roundtrip_includes_start_time_map`: the new field survives a serialization round-trip.

## 2. sqlness happy-path coverage
`tests/cases/standalone/common/flow/flow_status.{sql,result}` — adds a stable projection (`SELECT flow_id, flow_name ...` with `SQLNESS REPLACE` masking the flow id), complementing the existing `SHOW FLOW STATUS LIKE 'no_such_flow'` which only checks the six-column header with no rows.

Notes:
- sqlness `-- SQLNESS REPLACE <pattern> <replacement>` splits on the first space, so patterns must not contain literal spaces (`\s+` instead of ` ` / `[ T]`). Added a comment in the case.
- `tests/cases/distributed/common` is a symlink to `../standalone/common`, so the case must pass identically in both environments. Full `SHOW FLOW STATUS` rows differ (standalone populated vs distributed all-NULL until the heartbeat propagates), so only stable columns are asserted here.
- Observed while verifying: in distributed mode `last_execution_time`/`state_size` were also NULL right after `ADMIN FLUSH_FLOW` in the sqlness run (heartbeat report period > test execution time). The PR body says these reflect a last-reporter snapshot — the observed behavior may be worth clarifying in the PR description or tracking as a follow-up.

Verified locally against the PR head: `cargo test -p common-meta flow_state` (2 passed) and `cargo sqlness bare -t flow_status` (standalone + distributed) all pass.

```diff
diff --git a/src/common/meta/src/key/flow/flow_state.rs b/src/common/meta/src/key/flow/flow_state.rs
index 1d4443c..b468e21 100644
--- a/src/common/meta/src/key/flow/flow_state.rs
+++ b/src/common/meta/src/key/flow/flow_state.rs
@@ -183,3 +183,56 @@ impl From<FlowStat> for FlowStateValue {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::collections::BTreeMap;
+
+    use serde_json;
+
+    use crate::key::flow::flow_state::FlowStateValue;
+    use crate::key::FlowId;
+
+    #[test]
+    fn test_deserialize_legacy_flow_state_value() {
+        // Legacy format: only state_size and last_exec_time_map are present,
+        // without the start_time_map field added in PR #8392.
+        let legacy_json = r#"{"state_size":{"1":1024,"2":2048},"last_exec_time_map":{"1":1700000000000}}"#;
+        let value: FlowStateValue = serde_json::from_str(legacy_json).unwrap();
+
+        let mut expected_state_size = BTreeMap::new();
+        expected_state_size.insert(FlowId::from(1u32), 1024usize);
+        expected_state_size.insert(FlowId::from(2u32), 2048usize);
+        assert_eq!(value.state_size, expected_state_size);
+
+        let mut expected_last_exec_time_map = BTreeMap::new();
+        expected_last_exec_time_map.insert(FlowId::from(1u32), 1700000000000i64);
+        assert_eq!(value.last_exec_time_map, expected_last_exec_time_map);
+
+        // serde(default) kicks in: old persisted data must not break,
+        // and the new field defaults to empty.
+        assert!(value.start_time_map.is_empty());
+    }
+
+    #[test]
+    fn test_flow_state_value_roundtrip_includes_start_time_map() {
+        let mut state_size = BTreeMap::new();
+        state_size.insert(FlowId::from(1u32), 1024usize);
+        let mut last_exec_time_map = BTreeMap::new();
+        last_exec_time_map.insert(FlowId::from(1u32), 1700000000000i64);
+        let mut start_time_map = BTreeMap::new();
+        start_time_map.insert(FlowId::from(1u32), 1700000000000i64);
+
+        let value = FlowStateValue {
+            state_size,
+            last_exec_time_map,
+            start_time_map,
+        };
+
+        let json = serde_json::to_string(&value).unwrap();
+        assert!(json.contains("start_time_map"));
+
+        let decoded: FlowStateValue = serde_json::from_str(&json).unwrap();
+        assert_eq!(decoded, value);
+    }
+}
diff --git a/tests/cases/standalone/common/flow/flow_status.result b/tests/cases/standalone/common/flow/flow_status.result
index 7473b92..7b7ff8a 100644
--- a/tests/cases/standalone/common/flow/flow_status.result
+++ b/tests/cases/standalone/common/flow/flow_status.result
@@ -44,6 +44,18 @@ SELECT flow_name FROM information_schema.flow_statistics WHERE flow_name = 'test
 | test_flow_status |
 +------------------+
 
+-- SHOW FLOW STATUS is exercised below via the empty-result form (six-column header).
+-- Full row values differ between standalone (populated) and distributed (heartbeat
+-- has not yet propagated start_time), so assert only the stable columns here.
+-- SQLNESS REPLACE (\|\s+[0-9]+\s+\|) | ID |
+SELECT flow_id, flow_name FROM information_schema.flow_statistics WHERE flow_name = 'test_flow_status';
+
++---------+------------------+
+| flow_id | flow_name        |
++---------+------------------+
+| ID | test_flow_status |
++---------+------------------+
+
 -- the like filter matches against flow_name; no matching flow returns empty.
 SHOW FLOW STATUS LIKE 'no_such_flow';
 
diff --git a/tests/cases/standalone/common/flow/flow_status.sql b/tests/cases/standalone/common/flow/flow_status.sql
index bf14b01..6421604 100644
--- a/tests/cases/standalone/common/flow/flow_status.sql
+++ b/tests/cases/standalone/common/flow/flow_status.sql
@@ -26,6 +26,12 @@ ADMIN FLUSH_FLOW('test_flow_status');
 -- flow_name is deterministic; the rest of the columns are runtime dependent.
 SELECT flow_name FROM information_schema.flow_statistics WHERE flow_name = 'test_flow_status';
 
+-- SHOW FLOW STATUS is exercised below via the empty-result form (six-column header).
+-- Full row values differ between standalone (populated) and distributed (heartbeat
+-- has not yet propagated start_time), so assert only the stable columns here.
+-- SQLNESS REPLACE (\|\s+[0-9]+\s+\|) | ID |
+SELECT flow_id, flow_name FROM information_schema.flow_statistics WHERE flow_name = 'test_flow_status';
+
 -- the like filter matches against flow_name; no matching flow returns empty.
 SHOW FLOW STATUS LIKE 'no_such_flow';
 
```

discord9:
Thanks for applying the patch! Two things:

## 1. ⚠️ Please remove accidentally committed local tool output files
Commit `015f210086` includes three files that look like local scratch output from your Windows machine and should not be in the repo:
- `_fmt.txt` — `cargo fmt` diff output (`Diff in \\?\C:\Users\palak_uge27\greptimedb\...`)
- `_fmt2.txt` — just `EXIT:0`
- `_test.txt` — PowerShell `cargo test` output (618 lines)

They'd fail CI/license checks and shouldn't land in main. A `git rm` (or just deleting them) in a follow-up commit would be great.

## 2. Everything else looks good ✅
- The persistence compatibility tests (`flow_state.rs`) and the sqlness happy-path case (`flow_status.{sql,result}`) match what I verified locally — `cargo test -p common-meta flow_state` (2 passed) and `cargo sqlness bare -t flow_status` (standalone + distributed) both pass on this head.
- Nice touch merging `get_state_size` / `get_last_exec_time_map` / `get_flow_exec_stats` into a single `get_full_flow_stat` round-trip — that addresses the earlier inter-thread IPC feedback.
- The `record_start_time_if_first` comment clarifying pre/post-execution semantics is helpful.

One caveat from my verification worth keeping in mind: in distributed mode, right after `ADMIN FLUSH_FLOW`, `SHOW FLOW STATUS` returns NULL for all runtime columns (heartbeat report period > test execution time), so the PR description's claim that `last_execution_time`/`state_size` reflect a "last-reporter snapshot" in distributed mode may need a caveat (or a follow-up) — the sqlness case deliberately asserts only the stable columns.

onepizzateam:
> Thanks for applying the patch! Two things:
> 
> ## 1. ⚠️ Please remove accidentally committed local tool output files
> Commit `015f210086` includes three files that look like local scratch output from your Windows machine and should not be in the repo:
> 
> * `_fmt.txt` — `cargo fmt` diff output (`Diff in \\?\C:\Users\palak_uge27\greptimedb\...`)
> * `_fmt2.txt` — just `EXIT:0`
> * `_test.txt` — PowerShell `cargo test` output (618 lines)
> 
> They'd fail CI/license checks and shouldn't land in main. A `git rm` (or just deleting them) in a follow-up commit would be great.
> 
> ## 2. Everything else looks good ✅
> * The persistence compatibility tests (`flow_state.rs`) and the sqlness happy-path case (`flow_status.{sql,result}`) match what I verified locally — `cargo test -p common-meta flow_state` (2 passed) and `cargo sqlness bare -t flow_status` (standalone + distributed) both pass on this head.
> * Nice touch merging `get_state_size` / `get_last_exec_time_map` / `get_flow_exec_stats` into a single `get_full_flow_stat` round-trip — that addresses the earlier inter-thread IPC feedback.
> * The `record_start_time_if_first` comment clarifying pre/post-execution semantics is helpful.
> 
> One caveat from my verification worth keeping in mind: in distributed mode, right after `ADMIN FLUSH_FLOW`, `SHOW FLOW STATUS` returns NULL for all runtime columns (heartbeat report period > test execution time), so the PR description's claim that `last_execution_time`/`state_size` reflect a "last-reporter snapshot" in distributed mode may need a caveat (or a follow-up) — the sqlness case deliberately asserts only the stable columns.

Removed the three local output files and updated branch. On the distributed-mode caveat: added a note to #8454 to track the heartbeat timing / NULL behavior separately from the aggregation issue. ptal @discord9 @fengjiachun

discord9 (approved):

discord9:
LGTM, Thanks!

PR #8398, opened 2026-07-02
fix: handle PromQL time binary aggregation

### LINKED ISSUE #8397: PromQL query_range returns 500 for min/max/avg/sum by (...) (time() - metric)
### What type of bug is this?

Unexpected error

Frontend

### Minimal reproduce step

upgrading from `v1.0.0-rc.2` to `v1.1.1`.

```promql 
min( time() -
  container_start_time_seconds{}
) by (namespace, project,app)
```

The query should return a valid Prometheus matrix result instead of HTTP 500.

### What did you see instead?

unexpected response with status code 500: {"status":"error","error":"Internal error: 1003","errorType":"Internal"}

This also fails with `max`, `avg`, and `sum`.

However:
- removing `time()` makes the query work
- removing the aggregation makes the query work

linux

v1.1.1

```bash

```

fengjiachun:
## What changed

- Skip non-Greptime table scans when collecting PromQL row-key tag metadata.
- Add regression coverage for sum/min/max/avg by (...) over time() - metric.

## Why

Issue #8397 showed query_range returning 500 for aggregation over a binary expression involving time(). In v1.1.x, binary projections can retain __tsid for metric-engine optimization. The outer aggregate then inspects row-key metadata and may traverse the EmptyMetric dummy scan generated for time(), which is not backed by a Greptime table provider.

## Validation

- cargo fmt --check
- cargo test -p query aggregate_over_binary_time_function_expr --lib -- --nocapture
- cargo test -p query tsid --lib

Fixes #8397


fengjiachun:
@codex review

src/query/src/promql/planner.rs
```diff
@@ -2465,8 +2465,10 @@ impl PromPlanner {
             plan: &LogicalPlan,
             out: &mut BTreeSet<String>,
         ) -> Result<()> {
-            if let LogicalPlan::TableScan(scan) = plan {
-                let table = planner.table_from_source(&scan.source)?;
+            // Derived PromQL plans may contain non-Greptime scans without row-key metadata.
+            if let LogicalPlan::TableScan(scan) = plan
+                && let Ok(table) = planner.table_from_source(&scan.source)
+            {
```
evenyag, line 2471: Can we add some sqlness tests?

fengjiachun:
@codex review

evenyag (approved):

waynexia (approved):

PR #8403, opened 2026-07-02
feat: support ALTER TABLE SET auto_flush_interval

### LINKED ISSUE #8394: Support ALTER TABLE SET auto_flush_interval
### What type of request is this?

Follow-up enhancement for table-level `auto_flush_interval`.

Distributed Cluster, Storage Engine, Table Engine, Frontend, Datanode

### Background

PR #8357 added table-level `auto_flush_interval` support for the `CREATE TABLE` path:

```sql
CREATE TABLE t (
    ts TIMESTAMP TIME INDEX
) WITH ('auto_flush_interval' = '5m');
```

The PR description explicitly scoped the change to creation time and noted that `ALTER TABLE` support should follow later:

> Scope: this PR covers setting the option at `CREATE TABLE`. Changing it via `ALTER TABLE` is a natural follow-up (a `SetRegionOption::AutoFlushInterval` variant + the apply path in `handle_alter`); happy to fold it in here if preferred.

### Requested follow-up

Support changing the per-table auto flush interval on existing tables:

```sql
ALTER TABLE t SET 'auto_flush_interval' = '5m';
```

This is useful for old/existing tables after upgrade: users can opt in to the new per-table flush cadence without recreating the table.

### Current behavior

The statement is currently rejected because the ALTER TABLE SET option plumbing does not include `auto_flush_interval` yet:

```text
Error: 1004(InvalidArguments), Invalid set table option request: Invalid set region option request, key: auto_flush_interval, value: 5m
```

Relevant implementation areas:

- `src/store-api/src/region_request.rs`: add a `SetRegionOption::AutoFlushInterval` variant and parse/validate positive `humantime` durations.
- `src/table/src/metadata.rs`: persist the table option to `TableOptions.extra_options`.
- `src/mito2/src/worker/handle_alter.rs`: apply the new interval to the current region options, similar to other fast region options.

### Version

Current main / development build after PR #8357.

srivtx:

Closes #8394. Follow-up to #8357.

## Summary

Lets users change the per-table `auto_flush_interval` after CREATE:

```sql
ALTER TABLE t SET 'auto_flush_interval' = '5m';
```

- `src/store-api/src/region_request.rs` — new `SetRegionOption::AutoFlushInterval(Option<Duration>)` variant. `TryFrom<&PbOption>` parses with `humantime` and rejects zero/negative durations.
- `src/table/src/metadata.rs` — on `Some(d)`, write the humantime string into `TableOptions.extra_options`; on `None`, drop the key. Same path `append_mode` already uses.
- `src/mito2/src/worker/handle_alter.rs` — apply in `handle_alter_region_options_fast` and group with `Ttl`/`Twsc` in `new_region_options_on_empty_memtable` (no memtable flush needed; the flush worker re-reads the interval each cycle).

Tests: two unit tests in `metadata.rs`, plus a sqlness case covering create-then-alter, re-altering, altering a table that already had the option at CREATE, invalid duration, and zero duration.

## PR Checklist

### DIFF
--- a/src/store-api/src/region_request.rs
+++ b/src/store-api/src/region_request.rs
@@ -1398,6 +1398,12 @@ impl TryFrom<&PbOption> for SetRegionOption {
                 Ok(Self::AppendMode(append_mode))
             }
             AUTO_FLUSH_INTERVAL_KEY => {
+                if value.is_empty() {
+                    // SET 'auto_flush_interval' = NULL comes through as an empty
+                    // string; treat it as clearing the override (fall back to
+                    // the global default), same as Ttl.
+                    return Ok(Self::AutoFlushInterval(None));
+                }
                 let interval = humantime::parse_duration(value)
                     .map_err(|_| InvalidSetRegionOptionRequestSnafu { key, value }.build())?;
                 if interval <= Duration::ZERO {
@@ -1687,6 +1693,44 @@ mod tests {
             location: None,
         })
         .unwrap_err();
+    }
+
+    #[test]
+    fn test_set_region_option_auto_flush_interval_try_from() {
+        use std::time::Duration;
+
+        // Valid duration
+        let pb = PbOption {
+            key: "auto_flush_interval".to_string(),
+            value: "5m".to_string(),
+        };
+        let opt = SetRegionOption::try_from(&pb).unwrap();
+        assert_eq!(
+            opt,
+            SetRegionOption::AutoFlushInterval(Some(Duration::from_secs(300)))
+        );
+
+        // Empty value clears the override (mirrors Ttl's behaviour for `SET key = NULL`).
+        let pb = PbOption {
+            key: "auto_flush_interval".to_string(),
+            value: String::new(),
+        };
+        let opt = SetRegionOption::try_from(&pb).unwrap();
+        assert_eq!(opt, SetRegionOption::AutoFlushInterval(None));
+
+        // Zero is rejected up front (engine invariant).
+        let pb = PbOption {
+            key: "auto_flush_interval".to_string(),
+            value: "0s".to_string(),
+        };
+        assert!(SetRegionOption::try_from(&pb).is_err());
+
+        // Garbage value is rejected.
+        let pb = PbOption {
+            key: "auto_flush_interval".to_string(),
+            value: "not_a_duration".to_string(),
+        };
+        assert!(SetRegionOption::try_from(&pb).is_err());
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


srivtx:
Good catch — added a `> Duration::ZERO` check in the `TryFrom<&PbOption>` arm in the second commit, plus a '0s' error case in the sqlness test. Now fails at the ALTER TABLE statement instead of at next flush.

tests/cases/standalone/common/alter/alter_auto_flush_interval.sql
```diff
@@ -0,0 +1,63 @@
+-- Test altering auto_flush_interval on a mito table
```
fengjiachun, line 1: This new sqlness case needs a checked-in `alter_auto_flush_interval.result`. The current PR head has the `.sql` only, and all sqlness jobs are failing on this head, so the case is not mergeable as committed. Please run/update `cargo sqlness bare -t alter_auto_flush_interval` and commit the expected result.

src/store-api/src/region_request.rs
```diff
@@ -1394,6 +1397,14 @@ impl TryFrom<&PbOption> for SetRegionOption {
                     .map_err(|_| InvalidSetRegionOptionRequestSnafu { key, value }.build())?;
                 Ok(Self::AppendMode(append_mode))
             }
+            AUTO_FLUSH_INTERVAL_KEY => {
+                let interval = humantime::parse_duration(value)
```
fengjiachun, line 1407: Should this handle the empty value as `AutoFlushInterval(None)` or drop the `Option`/None test? `ALTER TABLE ... SET 'key'=NULL` is parsed as `value == ""` (TTL uses that to clear the override), and this PR adds the None persistence path, but this arm always calls `parse_duration`, so users cannot clear an auto_flush_interval override through SET NULL.

srivtx:
@fengjiachun done. .result checked in (ran cargo sqlness-runner bare, both standalone + distributed pass locally). Empty value now maps to None so SET ... = NULL clears the override, same as Ttl. Added a unit test and extended the sqlness case for the NULL path.

fengjiachun (approved):
LGTM

src/store-api/src/region_request.rs
```diff
@@ -1394,6 +1397,20 @@ impl TryFrom<&PbOption> for SetRegionOption {
                     .map_err(|_| InvalidSetRegionOptionRequestSnafu { key, value }.build())?;
                 Ok(Self::AppendMode(append_mode))
             }
+            AUTO_FLUSH_INTERVAL_KEY => {
+                if value.is_empty() {
+                    // SET 'auto_flush_interval' = NULL comes through as an empty
+                    // string; treat it as clearing the override (fall back to
+                    // the global default), same as Ttl.
```
fengjiachun, line 1404: Non-blocking: could we tweak this note? `SET ... = NULL` clears `auto_flush_interval`, but TTL maps an empty value to an explicit `TimeToLive::Forever` rather than removing the table option. This is closer to the empty-value clearing convention used by the TWCS options.

PR #8411, opened 2026-07-03
fix(meta): configure heartbeat message size

discord9:

N/A

This PR configures the heartbeat gRPC message size limits to use GreptimeDB's existing gRPC max send/receive settings instead of tonic's default limit.

The heartbeat stream carries mailbox messages between metasrv and datanodes. Some control-plane replies, such as GC `GetFileRefs` / `GcRegions` replies, can be larger than tonic's default 4MiB message limit. When that happens, metasrv may fail to decode a heartbeat request and the waiting mailbox receiver observes the datanode channel as closed.

Changes:

- Configure `HeartbeatClient` with `max_decoding_message_size` and `max_encoding_message_size` from its `ChannelManager` config.
- Configure metasrv `HeartbeatServer` with `max_decoding_message_size` and `max_encoding_message_size` from `MetasrvOptions.grpc.as_config()`.
- Keep the existing compression settings unchanged.

This is backward-compatible and only makes the heartbeat service honor the same gRPC message-size configuration already used by other gRPC services/clients.

Validation:

- `cargo fmt --check`
- `cargo check -p meta-client -p meta-srv`
- `git diff --check`

## PR Checklist

### DIFF
--- a/src/meta-client/src/client/heartbeat.rs
+++ b/src/meta-client/src/client/heartbeat.rs
@@ -280,10 +280,16 @@ impl Inner {
             .get(addr)
             .context(error::CreateChannelSnafu)?;
 
+        let config = self.channel_manager.config();
+        let max_decoding_message_size = config.max_recv_message_size.as_bytes() as usize;
+        let max_encoding_message_size = config.max_send_message_size.as_bytes() as usize;
+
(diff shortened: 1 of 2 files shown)


fengjiachun (approved):
LGTM

PR #8412, opened 2026-07-03
fix: global limit for distributed inspect streams

### LINKED ISSUE #8189: Distributed SST inspect tables can return incorrect results with LIMIT across datanodes
### What type of bug is this?

Incorrect result

Distributed Cluster

### Minimal reproduce step

In a distributed cluster with multiple datanodes, create a partitioned table that produces SST inspect rows on multiple regions and datanodes:
```
CREATE TABLE ssts_limit_case (
  a INT PRIMARY KEY INVERTED INDEX,
  b STRING SKIPPING INDEX,
  c STRING FULLTEXT INDEX,
  ts TIMESTAMP TIME INDEX,
)
PARTITION ON COLUMNS (a) (
  a < 1000,
  a >= 1000 AND a < 2000,
  a >= 2000
);
INSERT INTO ssts_limit_case VALUES
  (500, 'a', 'a', 1),
  (1500, 'b', 'b', 2),
  (2500, 'c', 'c', 3);
ADMIN FLUSH_TABLE('ssts_limit_case');
```
Then query SST inspect tables with LIMIT:

```
SELECT COUNT(*)
FROM (
  SELECT region_id
  FROM information_schema.ssts_manifest
  WHERE region_id != origin_region_id
  LIMIT 1
);
```

```
SELECT COUNT(*)
FROM (
  SELECT region_id
  FROM information_schema.ssts_manifest
  WHERE region_id != origin_region_id
  LIMIT 1
);
```
should return:
1

### What did you see instead?

return up to: min(region_num, datanode_count) * 10 rows.

doesn't matter

main

```bash

```

fengjiachun:
## What changed

- Add `EnsureGlobalLimitForFetch`, a physical optimizer rule that compensates multi-partition `FilterExec(fetch = N)` with a global fetch when its direct parent does not already provide one.
- Preserve ordering-sensitive parents by using `SortPreservingMergeExec(fetch = N)` with the child plan's actual output ordering instead of `CoalescePartitionsExec(fetch = N)` when the parent requires ordered input.
- Keep `LimitedRecordBatchStream` after distributed inspect streams as an early-stop optimization and drop the inner stream when the limit is reached.
- Add distributed sqlness coverage for `information_schema.ssts_*` with `WHERE ... LIMIT 1`.

## Why

DataFusion can absorb SQL `LIMIT` into a multi-partition `FilterExec`. In that shape, `fetch` is applied per partition, and the upper global limit is gone. For filtered distributed inspect queries, `ScanRequest.limit` can also be `None`, so scan-side limiting alone does not fix the issue.

## Validation

- `cargo fmt --check`
- `cargo test -p query global_limit --lib`
- `cargo test -p common-recordbatch limited --lib`
- `cargo test -p catalog test_datanode_inspect_region_info_build_plan --lib`
- `cargo sqlness bare -c tests/cases -t ssts_limit`

Fixes #8189.


fengjiachun:
@codex review

src/query/src/optimizer/global_limit.rs
```diff
@@ -0,0 +1,277 @@
...
+        parent: ParentContext,
+    ) -> DfResult<Arc<dyn ExecutionPlan>> {
+        let children = plan.children();
+        let plan = if children.is_empty() {
+            plan
+        } else {
+            let required_input_ordering = plan.required_input_ordering();
+            let child_parent = ParentContext {
+                has_global_fetch: provides_global_fetch(&plan),
+                required_ordering: None,
```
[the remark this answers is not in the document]
fengjiachun: Addressed in the latest commit. The rule now propagates inherited ordering requirements through order-preserving operators via `maintains_input_order()`, so a shape like `SortPreservingMergeExec -> ProjectionExec -> FilterExec(fetch)` wraps the fetched filter with `SortPreservingMergeExec(fetch)` instead of `CoalescePartitionsExec`. Added `preserves_inherited_ordering_requirement_through_projection` to cover this case, and verified with `cargo test -p query global_limit --lib` plus `cargo sqlness bare -c tests/cases -t tql-explain-analyze/explain`.

fengjiachun:
@codex review

src/query/src/optimizer/global_limit.rs
```diff
@@ -0,0 +1,312 @@
...
+                })
+                .collect::<DfResult<Vec<_>>>()?;
+            plan.with_new_children(children)?
+        };
+
+        let Some(fetch) = plan.fetch() else {
+            return Ok(plan);
+        };
+
+        if parent.has_global_fetch
```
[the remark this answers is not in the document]
fengjiachun: Addressed in commit `d6be665b9`. `ParentContext` now carries the parent global fetch value, and the rule skips compensation only when the parent global fetch is tighter than or equal to the filter fetch. If the parent fetch is looser, the filter still gets its own global fetch wrapper. Added `adds_tighter_global_fetch_under_looser_parent_fetch` to cover this case. Verified with `cargo fmt --check`, `cargo test -p query global_limit --lib`, and `cargo sqlness bare -c tests/cases -t tql-explain-analyze/explain`.

fengjiachun:
@codex review

src/query/src/optimizer/global_limit.rs
discord9: Skipping when `parent.has_global_fetch` is true seems only safe if the parent fetch is equal to or tighter than this `FilterExec` fetch. A looser parent global fetch can still allow more rows than the inner/filter fetch intended when the filter is multi-partitioned.

Could we track the parent global fetch value instead of a boolean, and skip only when the parent fetch is `<=` the current fetch? A nested-limit test would cover this.

src/query/src/optimizer/global_limit.rs
```diff
@@ -0,0 +1,312 @@
...
+            || !plan.as_any().is::<FilterExec>()
+            || plan.output_partitioning().partition_count() <= 1
+        {
+            return Ok(plan);
+        }
+
+        Ok(add_global_fetch(plan, fetch, parent.required_ordering))
```
discord9: This can change the child distribution after distribution enforcement has already run. If this `FilterExec(fetch)` is feeding a parent with a required input distribution, inserting `CoalescePartitionsExec` / `SortPreservingMergeExec` here turns that child into a single-partition output without re-satisfying the parent distribution requirement.

Can we either run/re-run distribution enforcement after this rule, or carry `required_input_distribution` in `ParentContext` and restore the required distribution after adding the global fetch? It would also be good to add a regression test where a limited/filter subquery feeds a hash-distributed operator, e.g. a join.

src/catalog/src/information_extension.rs
```diff
@@ -140,6 +141,12 @@ impl InformationExtension for DistributedInformationExtension {
 
         let chained =
             ChainedRecordBatchStream::new(streams).context(crate::error::CreateRecordBatchSnafu)?;
-        Ok(Box::pin(chained))
+        match limit {
+            Some(limit) => Ok(Box::pin(LimitedRecordBatchStream::new(
```
discord9, line 145: Stepping back, I think enforcing the SQL global `LIMIT` in a hidden stream wrapper still makes this path hard to reason about. The distributed inspect fan-in should keep the semantic/global limit visible in the logical/physical plan, while any limit sent to datanodes is treated only as a local pushdown/early-stop optimization.

Could we model this as an explicit distributed inspect scan/exec instead? It could borrow a simplified version of the DistPlanner / `MergeScanExec` idea: a visible fan-out/fan-in plan node that declares its partitioning/order semantics and keeps the outer global `Limit` in the plan, rather than chaining streams directly and then compensating with `LimitedRecordBatchStream` or a late physical optimizer rule.

src/query/src/optimizer/global_limit.rs
fengjiachun: Addressed in `b30e2f3f4`. `ParentContext` now carries the parent's required input distribution. After adding the global fetch wrapper, the rule restores hash distribution with `RepartitionExec` using the original partition count. Added `restores_parent_hash_distribution_after_global_fetch`, which builds a partitioned `HashJoinExec` over a fetched multi-partition filter.

Verified with `cargo fmt --check`, `cargo test -p query global_limit --lib`, `cargo sqlness bare -c tests/cases -t tql-explain-analyze/explain`, and `cargo sqlness bare -c tests/cases -t ssts_limit`.

src/catalog/src/information_extension.rs
fengjiachun: Agreed that an explicit distributed inspect exec would be a cleaner long-term shape. I would keep that out of this bug-fix PR because it is a larger planner/execution refactor. The semantic fix for #8189 is now the visible physical optimizer rule; the `LimitedRecordBatchStream` cap only applies when `scan.limit` is already pushed to datanodes as a local early-stop optimization.

We can track the explicit inspect exec as follow-up work separately.

fengjiachun:
@codex review

src/query/src/optimizer/global_limit.rs
fengjiachun: Addressed earlier in `d6be665b9`. `ParentContext` now carries the parent global fetch value, and the rule skips compensation only when the parent fetch is tighter than or equal to the current `FilterExec` fetch. If the parent fetch is looser, the filter still gets its own global fetch wrapper. Added `adds_tighter_global_fetch_under_looser_parent_fetch` to cover this nested-limit case.

src/catalog/src/information_extension.rs
fengjiachun: One more note on why the optimizer rule is not just a compensation for the stream-chaining design: an explicit inspect exec alone would not fix #8189. The broken shape was `FilterExec(fetch) -> RepartitionExec -> scan`, and DataFusion's physical `LimitPushdown` absorbs the global limit into the multi-partition `FilterExec` (marking the global state `satisfied` without checking partition count) regardless of how the scan node below is modeled. So a plan-level global fetch guard — either this rule or a root-cause fix in our DataFusion fork's `limit_pushdown.rs` — is required either way, and the exec refactor is an orthogonal cleanliness improvement.

I will file two follow-ups: (1) the explicit distributed inspect exec, and (2) the fork-side `limit_pushdown` fix (a non-partition-combining plan with multiple output partitions should not mark the limit `satisfied` when absorbing fetch). Once (2) lands, `EnsureGlobalLimitForFetch` can be retired.

fengjiachun: Filed the two follow-ups: #8415 (explicit distributed inspect exec) and #8416 (fork-side `LimitPushdown` root-cause fix, after which `EnsureGlobalLimitForFetch` can be retired).

discord9:
One remaining edge case from the latest distribution-restoration change: the direct `HashJoinExec -> FilterExec(fetch)` case is covered, but distribution-preserving intermediates can still be missed. `required_distribution` is taken only from the immediate parent `required_input_distribution()`, while ordering has inherited propagation through `maintains_input_order()`. For example, `HashJoinExec -> ProjectionExec -> FilterExec(fetch)` could still rewrite the filter to `CoalescePartitionsExec` without restoring the hash distribution required by the join, because `ProjectionExec` itself has no required input distribution.

Could you add a regression test with a projection or similar unary node between the hash join and the filtered child, and either propagate/restore the parent hash requirement through that node or otherwise constrain the rule?

fengjiachun:
@discord9 addressed in `5d2165316`.

The rule now carries the partitioning that must be restored, instead of passing the parent hash expression blindly through intermediate nodes. For a direct parent requirement we restore the child's existing hash partitioning when it already satisfies the parent distribution. For distribution-preserving unary intermediates such as `ProjectionExec`, if the intermediate node's current output partitioning satisfies the inherited parent hash requirement, we restore the child's original hash partitioning after adding the global fetch wrapper.

Added `restores_inherited_hash_distribution_through_projection` for the `HashJoinExec -> ProjectionExec -> FilterExec(fetch)` case. Verified with `cargo fmt --check`, `cargo test -p query global_limit --lib`, `cargo sqlness bare -c tests/cases -t tql-explain-analyze/explain`, and `cargo sqlness bare -c tests/cases -t ssts_limit`.

fengjiachun:
@codex review

fengjiachun:
@discord9 addressed in `828c9bef7`.

The inherited hash-preservation signal now propagates through multi-hop distribution-preserving unary chains without forwarding schema-specific partitioning expressions. Each hop recomputes the restore partitioning from its own child output partitioning, so `HashJoinExec -> ProjectionExec -> ProjectionExec -> FilterExec(fetch)` keeps the upstream hash distribution after the global-fetch wrapper.

Added `restores_inherited_hash_distribution_through_multiple_projections`. Verified with `cargo fmt --check`, `cargo test -p query global_limit --lib`, `cargo sqlness bare -c tests/cases -t ssts_limit`, and `cargo sqlness bare -c tests/cases -t tql-explain-analyze/explain`.

fengjiachun:
@codex review

discord9 (approved):

PR #8414, opened 2026-07-03
feat(mito): add metric value encoding modes

discord9:

## Benchmark result

This PR now exposes three MetricEngine physical value-column encoding modes. For
new MetricEngine physical data regions, an omitted option is materialized as
`no_dictionary` at create time. Existing/opened regions whose persisted options
are missing the key still parse as `plain` for compatibility.

- `plain`: keep Parquet writer behavior, including dictionary when Parquet chooses it. Best for low-cardinality/repeated values.
- `no_dictionary`: disable dictionary for `greptime_value`. Best read-path result in the high-cardinality synthetic metric suite.
- `byte_stream_split`: use Parquet `BYTE_STREAM_SPLIT` and disable dictionary for `greptime_value`. Best space result in the high-cardinality synthetic metric suite.

### Manual 100M MetricEngine-value encoding suite

The new manual suite generates direct-readable Mito SST fixtures that mimic MetricEngine physical data regions, validates actual Parquet metadata for `greptime_value`, then runs:

- `greptime datanode parquetbench --reader direct`
- `greptime datanode scanbench --scanner seq`

Run shape:

- rows per SST: `100,000,000`
- series: `256`
- source batch rows: `65,536`
- row group size: `65,536`
- row groups: `1,526`
- iterations: `1`
- value distribution: deterministic `metric_signal` with per-series baseline/phase, trend, periodic terms, and pseudo-noise; it intentionally avoids a tiny exact-value cycle.

| mode | total bytes | `greptime_value` bytes | value encodings | parquet value avg | parquet ts+value avg | scan value avg | scan ts+value avg |
| --- | ---: | ---: | --- | ---: | ---: | ---: | ---: |
| `plain` | 951,381,930 | 935,840,286 | `PLAIN, RLE, RLE_DICTIONARY` | 12,482.738 ms | 14,546.685 ms | 28,154.400 ms | 28,708.422 ms |
| `no_dictionary` | 763,338,574 | 747,819,042 | `PLAIN, RLE` | 2,373.333 ms | 4,022.212 ms | 14,009.447 ms | 14,081.303 ms |
| `byte_stream_split` | 691,715,759 | 676,197,516 | `BYTE_STREAM_SPLIT, RLE` | 11,101.376 ms | 13,812.737 ms | 24,824.253 ms | 24,882.864 ms |

Space savings:

| comparison | total savings | `greptime_value` savings |
| --- | ---: | ---: |
| `no_dictionary` vs `plain` | 19.77% | 20.09% |
| `byte_stream_split` vs `plain` | 27.29% | 27.74% |
| `byte_stream_split` vs `no_dictionary` | 9.38% | 9.58% |

Interpretation:

- If values are highly repeated or low cardinality, explicitly use `plain` so Parquet dictionary can help.
- If the goal is faster scans on high-cardinality metric values, use `no_dictionary`.
- If the goal is smaller SST footprint on high-cardinality float metric values, use `byte_stream_split`; it saves additional space beyond disabling dictionary. In this 100M run BSS is also faster than `plain`, but still much slower than `no_dictionary`, so the main reason to choose it is footprint, not scan speed.

The suite is manual and intentionally not part of the default query-regression cases.

### Write-path size checks

Separate real MetricEngine/Mito write-path checks also compare the three modes on generated metric rows that are flushed to physical data-region SSTs.

| shape | mode | total bytes | `greptime_value` bytes | value savings vs `plain` |
| --- | --- | ---: | ---: | ---: |
| 1,048,576 rows / 64 series / batch 4,096 / row group 65,536 | `plain` | 2,999,317 | 2,958,613 | - |
| 1,048,576 rows / 64 series / batch 4,096 / row group 65,536 | `no_dictionary` | 2,330,181 | 2,289,730 | 22.61% |
| 1,048,576 rows / 64 series / batch 4,096 / row group 65,536 | `byte_stream_split` | 1,584,215 | 1,543,811 | 47.82% |
| 1,048,576 rows / 64 series / batch 65,536 / row group 65,536 | `plain` | 9,436,276 | 2,284,302 | - |
| 1,048,576 rows / 64 series / batch 65,536 / row group 65,536 | `no_dictionary` | 9,384,409 | 2,232,578 | 2.26% |
| 1,048,576 rows / 64 series / batch 65,536 / row group 65,536 | `byte_stream_split` | 8,648,574 | 1,496,765 | 34.48% |

The large-batch shape is important because it exercises the Mito bulk pre-encoded SST path. It shows that BSS savings are not just dictionary-off savings.

N/A

This PR changes the original default-off BSS flag into a persisted table/region option enum for MetricEngine physical data-region value encoding:

```sql
experimental_metric_engine_value_encoding = 'plain' | 'no_dictionary' | 'byte_stream_split'
```

Summary:

- Adds `MetricEngineValueEncoding::{Plain, NoDictionary, ByteStreamSplit}` to Mito `RegionOptions`; missing/serde default remains `Plain` for existing-region compatibility.
- Exposes the option through table options as `experimental_metric_engine_value_encoding`.
- Defaults newly-created MetricEngine physical data regions to `no_dictionary` when the option is omitted, while open/recovery preserves missing options as missing/`plain`.
- Removes the old global Mito config flag `experimental_enable_metric_engine_value_byte_stream_split`.
- Applies the selected encoding only to MetricEngine physical data-region `greptime_value` columns when the type is `Float32` or `Float64`.
- Hard-gates the selector to MetricEngine physical data regions by requiring:
  - metric data-region group;
  - MetricEngine dense primary-key marker;
  - reserved marker columns `__table_id` and `__tsid`;
  - target column name `greptime_value`;
  - target column type `Float32` or `Float64`.
- Threads the same selected-column policy through flush, compaction, bulk memtable, and bulk pre-encoded SST generation.
- Preserves the option for MetricEngine data regions and strips it from MetricEngine metadata regions.
- Adds a manual non-default perf suite under `tests/perf/metric_value_encoding/`.
- Adds an upgrade compatibility case under
  `tests/compatibility/cases/metric_engine_value_encoding_default/` for old
  MetricEngine regions missing the new option.
- Extends `query_perf_fixture` with:
  - `metric_physical = true`;
  - `metric_engine_value_encoding = "plain" | "no_dictionary" | "byte_stream_split"`;
  - `source_batch_rows`;
  - per-column Parquet metadata in `files.jsonl`;
  - deterministic `metric_signal` value distribution for realistic synthetic metric values.

Known limitations / compatibility notes:

- The option affects newly written/generated SSTs. Existing SSTs are not rewritten automatically.
- `BYTE_STREAM_SPLIT` is a standard Parquet encoding, not a custom GreptimeDB format.
- Existing/opened regions with no persisted value-encoding key keep compatibility by parsing as `plain`; open/recovery does not synthesize `no_dictionary`.
- New MetricEngine physical data regions with no explicit option persist `no_dictionary` as the create-time default.
- Explicit `plain`, `no_dictionary`, and `byte_stream_split` values are preserved.
- If `byte_stream_split` is enabled, newly written SSTs may not be readable by older GreptimeDB/parquet readers that do not support `BYTE_STREAM_SPLIT`.
- Disabling the option later stops new BSS writes but does not rewrite existing BSS SSTs.

Validation:

- `cargo fmt --check -p cmd -p mito2 -p metric-engine -p store-api -p table -p tests-integration`
- `cargo check -p cmd --bin query_perf_fixture`
- `cargo check -p metric-engine`
- `cargo check -p mito2`
- `cargo test -p cmd metric_signal --bin query_perf_fixture -- --nocapture`
- `cargo test -p metric-engine data_region_options -- --nocapture`
- `cargo test -p metric-engine metadata_region_strips_metric_value_encoding -- --nocapture`
- `cargo test -p metric-engine test_create_request -- --nocapture`
- `cargo test -p metric-engine open_data_region_options -- --nocapture`
- `cargo test -p mito2 metric_engine_value -- --nocapture`
- `cargo test -p mito2 no_dictionary -- --nocapture`
- `cargo test -p mito2 byte_stream_split -- --nocapture`
- `cargo test -p mito2 region_options -- --nocapture`
- `cargo test -p mito2 metric_engine_float64_value_encoding_modes_readback_compatible -- --nocapture`
- `cargo test -p mito2 test_bulk_merge_preencode_metric_value -- --nocapture`
- `cargo test -p mito2 test_bulk_memtable_builder_metric_value -- --nocapture`
- `cargo test -p metric-engine metric_value_encoding -- --nocapture`
- `cargo run -p sqlness-runner -- compat --dry-run --test-filter metric_engine_value_encoding_default --from-bins-dir <debug-bins-dir> --to-bins-dir <debug-bins-dir>` (current→current is skipped by version range)
- `cargo run -p sqlness-runner -- compat --dry-run --test-filter metric_engine_value_encoding_default --from-version v1.1.0 --to-bins-dir <debug-bins-dir>`
- `cargo run -p sqlness-runner -- compat --test-filter metric_engine_value_encoding_default --from-version v1.1.0 --to-bins-dir <debug-bins-dir>`
- `uv run --no-project python -m py_compile tests/perf/metric_value_encoding/run.py tests/perf/query_regression_runner.py`
- `make check-toml`
- `git diff --check`
- Manual perf suite smoke: `uv run --no-project python tests/perf/metric_value_encoding/run.py --rows-per-sst 1024 --source-batch-rows 256 --row-group-size 512 --series-count 16 --iterations 1 --no-build`
- Manual 100M suite: `uv run --no-project python tests/perf/metric_value_encoding/run.py --rows-per-sst 100000000 --source-batch-rows 65536 --row-group-size 65536 --series-count 256 --iterations 1 --no-build`

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet.rs
+++ b/src/mito2/src/sst/parquet.rs
@@ -17,9 +17,13 @@
 use std::sync::Arc;
 
 use common_base::readable_size::ReadableSize;
+use datatypes::prelude::ConcreteDataType;
 use parquet::file::metadata::ParquetMetaData;
+use store_api::metadata::RegionMetadata;
+use store_api::metric_engine_consts::METRIC_DATA_REGION_GROUP;
 use store_api::storage::FileId;
 
+use crate::region::options::RegionOptions;
 use crate::sst::DEFAULT_WRITE_BUFFER_SIZE;
 use crate::sst::file::FileTimeRange;
 use crate::sst::index::IndexOutput;
@@ -65,6 +69,8 @@ pub struct WriteOptions {
     /// Note: This is not a hard limit as we can only observe the file size when
     /// ArrowWrite writes to underlying writers.
     pub max_file_size: Option<usize>,
+    /// Metric engine value column to encode with Parquet BYTE_STREAM_SPLIT.
+    pub metric_engine_value_byte_stream_split_column: Option<String>,
 }
 
 impl Default for WriteOptions {
@@ -73,8 +79,35 @@ impl Default for WriteOptions {
             write_buffer_size: DEFAULT_WRITE_BUFFER_SIZE,
             row_group_size: DEFAULT_ROW_GROUP_SIZE,
             max_file_size: None,
+            metric_engine_value_byte_stream_split_column: None,
         }
     }
+}
+
+/// Returns the metric engine value column name to encode with BYTE_STREAM_SPLIT.
+pub(crate) fn metric_engine_value_byte_stream_split_column(
+    enabled: bool,
+    region_metadata: &RegionMetadata,
+    region_options: &RegionOptions,
+) -> Option<String> {
+    if !enabled
+        || region_metadata.region_id.region_group() != METRIC_DATA_REGION_GROUP
+        || region_options.primary_key_encoding.is_none()
+    {
+        return None;
+    }
+
+    let value_column = common_query::prelude::greptime_value();
+    region_metadata
+        .field_columns()
+        .any(|column| {
+            column.column_schema.name == value_column
+                && matches!(
+                    column.column_schema.data_type,
+                    ConcreteDataType::Float32(_) | ConcreteDataType::Float64(_)
+                )
+        })
+        .then(|| value_column.to_string())
 }
 
 /// Parquet SST info returned by the writer.
@@ -117,8 +150,8 @@ mod tests {
     use datafusion_expr::{BinaryExpr, Expr, Literal, Operator, col, lit};
     use datatypes::arrow;
     use datatypes::arrow::array::{
-        ArrayRef, BinaryDictionaryBuilder, RecordBatch, StringArray, StringDictionaryBuilder,
-        TimestampMillisecondArray, UInt8Array, UInt64Array,
+        Array, ArrayRef, BinaryDictionaryBuilder, Float32Array, Float64Array, RecordBatch,
+        StringArray, StringDictionaryBuilder, TimestampMillisecondArray, UInt8Array, UInt64Array,
     };
     use datatypes::arrow::datatypes::{DataType, Field, Schema, UInt32Type};
     use datatypes::arrow::util::pretty::pretty_format_batches;
@@ -131,6 +164,7 @@ mod tests {
     use parquet::file::properties::WriterProperties;
     use store_api::codec::PrimaryKeyEncoding;
     use store_api::metadata::{ColumnMetadata, RegionMetadata, RegionMetadataBuilder};
+    use store_api::metric_engine_consts::{DATA_SCHEMA_TSID_COLUMN_NAME, METRIC_DATA_REGION_GROUP};
     use store_api::region_request::PathType;
     use store_api::storage::{ColumnSchema, RegionId};
     use table::predicate::Predicate;
@@ -143,7 +177,7 @@ mod tests {
     use crate::cache::{CacheManager, CacheStrategy};
     use crate::config::IndexConfig;
     use crate::read::FlatSource;
-    use crate::region::options::{IndexOptions, InvertedIndexOptions};
+    use crate::region::options::{IndexOptions, InvertedIndexOptions, RegionOptions};
     use crate::sst::file::{FileHandle, FileMeta, RegionFileId, RegionIndexId};
     use crate::sst::file_purger::NoopFilePurger;
     use crate::sst::index::bloom_filter::applier::BloomFilterIndexApplierBuilder;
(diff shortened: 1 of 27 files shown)


discord9:
Separate benchmark note for reviewer context (not part of this PR diff):

I ran a true MetricEngine-path synthetic workload that writes logical metric rows through `MetricEngine`, flushes the physical data region to Mito SST, and compares data-region Parquet file sizes with the current PLAIN+ZSTD path vs. BYTE_STREAM_SPLIT+ZSTD for the metric value column.

| rows | series | row group size | PLAIN+ZSTD data SST | BSS+ZSTD data SST | ratio | savings | BSS verified |
|---:|---:|---:|---:|---:|---:|---:|:---:|
| 65,536 | 16 | 16,384 | 243,077 B | 132,346 B | 0.544461 | 45.554% | true |
| 1,048,576 | 64 | 65,536 | 3,378,577 B | 1,894,635 B | 0.560779 | 43.922% | true |

Caveats:

- This is a realistic synthetic workload, not a production trace.
- It only compares data-region SST file sizes; it does not measure write CPU, read CPU, query latency, or compaction CPU.
- The PR remains experimental/default-off; older readers may not read SSTs written with `BYTE_STREAM_SPLIT`, so the flag should only be enabled after all possible readers are upgraded.

src/mito2/src/sst/parquet.rs
```diff
@@ -65,6 +69,8 @@ pub struct WriteOptions {
     /// Note: This is not a hard limit as we can only observe the file size when
     /// ArrowWrite writes to underlying writers.
     pub max_file_size: Option<usize>,
+    /// Metric engine value column to encode with Parquet BYTE_STREAM_SPLIT.
+    pub metric_engine_value_byte_stream_split_column: Option<String>,
```
fengjiachun: This new required field breaks the `query_perf_fixture` binary: `cargo check -p cmd --bin query_perf_fixture` fails because `src/cmd/src/bin/query_perf_fixture.rs` still initializes `WriteOptions` without `metric_engine_value_byte_stream_split_column`. Please update that call site (or use `..Default::default()`) before merging.

fengjiachun (dismissed):
The benchmark data looks good.

waynexia:
we can always enable this encoding, and remove those compatibility/config things

discord9:
Updated this PR after the option rewrite and the larger manual benchmark. I don't think we should always enable BSS now: the 100M high-cardinality metric-signal suite shows three distinct tradeoffs. Default `plain` keeps dictionary for repeated/low-cardinality values; `no_dictionary` is clearly the fastest scan mode; `byte_stream_split` is the smallest-footprint mode and still saves another ~9.6% on `greptime_value` beyond disabling dictionary, but it is not the fastest read mode. So the table/region option keeps those choices explicit while the default remains backward-compatible.

discord9:
Closing for now while we revisit the encoding/default strategy and gather more benchmark evidence.

PR #8406, opened 2026-07-02
feat: add query regression perf harness

discord9:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

Related to #7913.

## What's changed and what's your intention?

This PR adds an opt-in query performance regression harness for comparing a base GreptimeDB build with a candidate build, then extends it with multi-table SQL optimizer coverage.

### Query regression harness

- Adds `tests/perf/query_regression_runner.py`, a base-vs-candidate runner that starts a local distributed cluster for each target (`metasrv` + one `datanode` + `frontend`), seeds table metadata through frontend SQL, discovers assigned regions via `information_schema`, stops only the owning datanode for direct SST/manifest injection, restarts it, then measures queries through frontend.
- Adds `src/cmd/src/bin/query_perf_fixture.rs`, a generic direct-readable-SST fixture generator for deterministic Mito fixtures. The generator is case-driven and is not hard-coded to #7913.
- Extends the generator and runner to support multi-table cases with one region per table. The generator now accepts `--table`; the runner creates/discovers/generates/materializes fixtures per table.
- Adds a workload extension shape:

  ```toml
  [workload]
  kind = "direct_readable_sst"

  [workload.direct_readable_sst]
  # currently no required fields
  ```

  Old cases without `[workload]` still default to `direct_readable_sst`. This leaves room for future `write_then_query` and `write_while_query` workloads.

### Query cases

- Keeps the initial smoke case and `promql_pushdown_7913` case.
- Adds SQL optimizer regression cases:
  - `sql_topk_order_by`: TopK / `ORDER BY ... LIMIT` over a DOUBLE field with time/tag filters.
  - `sql_aggregate_order_by`: grouped aggregate ordered by aggregate value with `LIMIT`.
  - `sql_join_filter_order`: true two-table JOIN with filters, aggregate ordering, and `LIMIT`.

### Workflow and reporting

- Adds `.github/workflows/query-regression.yml`, an opt-in PR/dispatch workflow that self-builds base and candidate binaries with the `nightly` Cargo profile by default, runs the candidate runner, uploads artifacts, and writes a workflow summary. `workflow_dispatch` can override the profile with `nightly`, `release`, or `dev`.
- Adds `.github/scripts/query-regression-summary.py` to format runner JSON into Markdown, including multi-table region display.
[...]

PR #8423, opened 2026-07-06
ci: run query regression on self-hosted runners

discord9:

Follow-up to #8406 and #8413.

This PR moves the `Query Regression` workflow onto a dedicated self-hosted ARC runner scale set and hardens how PR-triggered benchmark results are built, reported, and commented.

What changed:

- Uses the dedicated runner scale set label `perf-regression-8-cores` instead of generic hosted-runner-like labels.
- Adds the ARC runner scale-set configuration and documentation under `.github/runner-scale-sets/query-regression/`.
- Uses a digest-pinned custom runner image with the workflow prerequisites preinstalled (`wget`, `uv`, `mold`, build tools, protobuf tooling, etc.).
- Builds base and candidate in the same `src` checkout with a shared `CARGO_TARGET_DIR`, copying the base binary aside before resetting the checkout to the candidate ref.
- Keeps PR query-regression runs explicitly label-triggered: remove and re-add `query-regression` to rerun; `synchronize` does not automatically rerun expensive benchmarks.
- Keeps the Rust cache restore-only for PR/dispatch runs (`save-if: false`).
- Hardens ARC runner pods with digest-pinned image, no service account token, `RuntimeDefault` seccomp, no privilege escalation, and dropped Linux capabilities.
- Adds a trusted `workflow_run` comment path that validates PR metadata and renders JSON reports from trusted default-branch code before using `pull-requests: write`.

Validation performed:

- Query Regression run `28788083210` succeeded on head `d868ba6bd22ca42db097cc7470d9435f3b3875df`.
- All 5 default distributed query-regression cases completed with status `ok` and zero validation errors.

| Case | Query | Base median | Candidate median | Delta |
| --- | --- | ---: | ---: | ---: |
| `promql_pushdown_7913` | `narrow_window` | 24.18ms | 25.27ms | +4.5% |
| `smoke_direct_sst` | `count_all` | 3.52ms | 3.59ms | +2.0% |
| `sql_aggregate_order_by` | `aggregate_service_top` | 7.73ms | 7.99ms | +3.4% |
| `sql_join_filter_order` | `join_host_aggregate_top` | 116.83ms | 117.43ms | +0.5% |
| `sql_topk_order_by` | `topk_value_desc` | 6.49ms | 6.04ms | -6.8% |

Notes:

- The sticky comment workflow changes cannot be fully live-validated by this PR itself because `workflow_run` executes the workflow definition from the repository default branch. After this PR is merged, the updated trusted comment workflow can be validated by a later query-regression PR run.
- The current result was posted manually in this PR because the default-branch comment workflow still has the old fork-PR association logic.

## PR Checklist


.github/workflows/query-regression.yml
```diff
@@ -75,7 +83,28 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - uses: rui314/setup-mold@v1
```
fengjiachun: The current default ARC runner image still looks too minimal for this step. The real `Query Regression` run on `201bbc4` failed here because `rui314/setup-mold` shells out to `wget`, but `ghcr.io/actions/actions-runner:latest` did not have `wget` installed (`line 4: wget: command not found`). The latest commit only removes the extra runner sizes, so the default `perf-regression-8-cores` path should still hit the same failure. Could we either use/provide a runner image with the workflow tool prerequisites, or install/move those setup steps before invoking `setup-mold`? Moving `setup-uv` earlier would also keep the always-run PR metadata step from failing with `uv: command not found` after an early setup failure.

waynexia (approved):

discord9:
## Query regression report

Manual report for Query Regression run `28788083210` on head `d868ba6bd22ca42db097cc7470d9435f3b3875df`.

- **Workflow run:** https://github.com/GreptimeTeam/greptimedb/actions/runs/28788083210
- **Status:** success
- **Case set:** default distributed query-regression cases
- **Validation errors:** 0

| Case | Query | Base median | Candidate median | Delta |
| --- | --- | ---: | ---: | ---: |
| `promql_pushdown_7913` | `narrow_window` | 24.18ms | 25.27ms | +4.5% |
| `smoke_direct_sst` | `count_all` | 3.52ms | 3.59ms | +2.0% |
| `sql_aggregate_order_by` | `aggregate_service_top` | 7.73ms | 7.99ms | +3.4% |
| `sql_join_filter_order` | `join_host_aggregate_top` | 116.83ms | 117.43ms | +0.5% |
| `sql_topk_order_by` | `topk_value_desc` | 6.49ms | 6.04ms | -6.8% |

Note: the automatic sticky comment path is included in this PR, but `workflow_run` uses the workflow definition from the repository default branch. This PR therefore cannot validate its own updated trusted comment workflow until after it is merged; the default-branch workflow still skipped the automatic comment for this run.

PR #8409, opened 2026-07-03
fix(flow): rebind stale snapshot fence

discord9:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

N/A

## What's changed and what's your intention?

This PR fixes a Flow batching-mode liveness issue in fenced full-snapshot repair.

When a fenced repair query fails because its snapshot upper bound has become stale, Mito returns a stale snapshot fence error with a `REBIND_SNAPSHOT_FENCE` retry hint. In distributed execution this error can be wrapped by frontend/client layers so the outer Flow error no longer has `StatusCode::RequestOutdated`. Flow then classified it as a generic query failure, kept the existing fenced repair state, and retried the same stale upper bound repeatedly.

The fix keeps the existing structured `RequestOutdated` behavior, and adds a narrow fenced-repair-only liveness guard that recognizes explicit stale-fence markers in wrapped errors:

- `STALE_SNAPSHOT_FENCE`
- `REBIND_SNAPSHOT_FENCE`
- `snapshot upper bound stale`

For `QueryCoverage::FencedRepairChunk`, such wrapped errors are classified as `SnapshotFenceExpired`, so the existing `abandon_fenced_repair()` path runs and the next repair can rebind a fresh upper fence.

Regression tests cover:

- wrapped stale-fence marker classification;
- no misclassification for generic fenced errors;
- no `SnapshotFenceExpired` classification for non-fenced incremental coverage;
- unchanged structured `RequestOutdated` behavior;
- state transition from wrapped stale fence to a fresh `ScopedBaseRepair` plan.

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [x] API changes are backward compatible.
- [x] Schema or data changes are backward compatible.

Validation:

- `cargo test -p flow stale_fence`
- `cargo test -p flow fence`
- `cargo fmt -p flow -- --check`
- `cargo check -p flow`
- `git diff --check`

PR #8428, opened 2026-07-06
fix(flow): notify flow after source writes

discord9:

Related to #8409.

**Draft status:** this PR is not merge-ready. The change fixed the original 100k rows/s 2-minute sinkstress mismatch, but a longer 100k rows/s 5-minute validation still exposed a remaining Flow correctness gap. Keeping this as a draft while investigating the residual race.

This PR fixes one Flow dirty/mirror notification ordering race in the frontend/operator write path.

Previously, normal inserts mirrored source rows to Flow before the corresponding datanode write had succeeded, and bulk inserts marked Flow dirty windows before source writes completed. Under high write pressure, Flow could consume a dirty window while the source rows were not yet visible, clear/advance the dirty state, and miss rows that became visible afterward.

The fix changes the write paths to notify Flow only after source write success:

- normal insert:
  - precomputes all per-peer Flow mirror tasks before spawning any datanode write;
  - delays instant-TTL mirror dispatch until normal mirror precompute succeeds;
  - dispatches each normal peer's mirror only after that peer's datanode write returns `Ok`;
  - preserves partial-success behavior by mirroring successful peers even when another peer fails and the final response is an error.
- bulk insert:
  - precomputes dirty-window notification metadata before source writes, so timestamp/cache errors happen before commit;
  - dispatches dirty notifications only after the successful write boundary;
  - waits for all multi-region write attempts, marks dirty if any region succeeded, then returns the first error if any attempt failed.

Flow notification RPCs remain best-effort/fire-and-forget after source commit, so a committed source write is not reported as failed only because Flow notification fails.

Validation performed:

- `cargo fmt -p operator -- --check`
- `cargo test -p operator`
- `cargo clippy -p operator --all-targets -- -D warnings`
- `git diff --check`
- pre-push hook completed successfully

Live validation with frontend built from this PR and flownode built from #8409:

- 100k rows/s for 2 minutes inserted `12,002,000` rows with 0 client errors.
- After source/flow/sink flushes, source and sink both reported `12,002,000` rows.
- Key/window diff reported 0 mismatched keys and 0 extra sink keys.

Remaining failure under investigation:

- Continuing the same setup at 100k rows/s for 5 minutes inserted another `22,123,000` rows with 0 client errors.
- After source/flow/sink flushes, source reported `34,125,000` rows while sink reported `34,102,000` rows, lagging by `23,000`.
- Key/window diff showed `23,000` mismatched keys, each lagging by exactly 1 row (`max_abs_lag = 1`, no sink extra keys).
- Inserting one sentinel row in the same window and retrying `ADMIN FLUSH_FLOW` repaired the lag, ending at `34,125,001 = 34,125,001`.

Limitations:

- This PR currently addresses only the source-write-before-notify ordering issue. It may still leave a race between asynchronous/best-effort Flow notification delivery and `ADMIN FLUSH_FLOW`/dirty-window/checkpoint handling.
- Bulk insert has focused helper tests for precompute/dispatch behavior. Full `handle_bulk_insert` partial-failure integration tests would require heavier table/FlightData/partition setup and can be added later if needed.

## PR Checklist

### DIFF
--- a/src/operator/src/bulk_insert.rs
+++ b/src/operator/src/bulk_insert.rs
@@ -26,6 +26,9 @@ use bytes::Bytes;
 use common_base::AffectedRows;
 use common_grpc::FlightData;
 use common_grpc::flight::{FlightEncoder, FlightMessage};
+use common_meta::cache::TableFlownodeSetCacheRef;
+use common_meta::node_manager::NodeManagerRef;
+use common_meta::peer::Peer;
 use common_telemetry::error;
 use common_telemetry::tracing_context::TracingContext;
 use snafu::{OptionExt, ResultExt, ensure};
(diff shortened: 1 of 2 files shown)


discord9:

PR #8444, opened 2026-07-08
test(perf): add remote write storage inspection

discord9:

Follow-up for the query regression Prom remote-write scenario.

This PR extends the Prometheus remote-write query regression scenario so it can generate controlled sample-value distributions and validate the landed Parquet/SST storage output.

Summary:

- Add value-pattern CLI options to `prom_remote_write_fixture`:
  - `linear` (default, preserves the previous formula)
  - `constant`
  - `modulo`
  - `unique`
  - `seeded_random`
  - `run_length`
  - `quantized_signal`
  - `signal_with_sporadic_stalls`
  - `mixed_signal_repeated`
- Add supporting knobs: `--value-base`, `--value-step`, `--value-cardinality`, `--value-seed`, `--value-run-length`, `--value-stall-every`, `--value-stall-length`, and `--value-mixed-every`.
- Preserve global sample ordinals across chunked remote-write runs through runner-provided `--value-sample-offset` and `--value-total-samples-per-series`.
- Plumb optional `[scenario.remote_write.value]` from `tests/perf/query_regression_runner.py` to the fixture helper.
- Add `parquet_footer_inspector`, a small Rust helper that reads local Parquet footers without requiring Python `pyarrow`.
- Add optional `[scenario.remote_write.storage]` support so a case can run remote-write → flush → visibility check → Parquet footer/size/encoding inspection → query.
- Add storage checks for file counts, target column presence, required/forbidden encodings, absolute byte-size limits, and base-vs-candidate storage-size regression percentages.
- Document the new case configuration in `tests/perf/README.md`.

The added distribution families cover the main synthetic value shapes needed by storage encoding experiments: run-heavy low-cardinality data, quantized signals, mostly-continuous signals with sporadic stalls, and mixed signal/repeated-default data.

The default behavior is preserved: existing cases that do not configure `[scenario.remote_write.value]` do not pass new fixture args, and the default `linear` pattern keeps the previous value formula.

Storage checks are generic footer and byte-size assertions. They do not encode product-specific storage-policy heuristics, but they let storage/encoding experiments assert the resulting Parquet layout and value-column size.

Validation:

- `cargo fmt -p cmd`
- `cargo test -p cmd --bin prom_remote_write_fixture`
- `cargo test -p cmd --bin parquet_footer_inspector`
- `uv run --no-project python -m py_compile tests/perf/query_regression_runner.py`
- `git diff --check`
- `cargo clippy -p cmd --bin prom_remote_write_fixture -- -D warnings`
- `cargo clippy -p cmd --bin parquet_footer_inspector -- -D warnings`
- `cargo build -p cmd --bin prom_remote_write_fixture`
- `cargo build -p cmd --bin parquet_footer_inspector`
- Dry-run query regression with temporary `[scenario.remote_write.value]` cases to verify value args and chunk offsets in generated commands, including the extended distribution knobs.
- Dry-run query regression with `[scenario.remote_write.storage]` and `root_suffix` to verify inspector command planning and default storage thresholds.

## PR Checklist


src/cmd/src/bin/parquet_footer_inspector.rs
waynexia, line 1: prefer making this a test bin so it won't show up as a general binary entry

waynexia (approved):
rest LGTM

PR #8446, opened 2026-07-08
feat: make parquet row group size configurable

evenyag:

This PR adds a `max_row_group_row_count` region option for the Mito engine and wires it through region options, alter handling, flush, and Parquet bulk memtable writing so SST row group sizing can be controlled per region.

It also updates inverted index row selection logic so selected row ranges are split correctly across Parquet row group boundaries, and adds SQL/engine tests for creating and flushing regions with the new option.

Validation: not run in this session; this PR was created from the existing branch state.

## PR Checklist


evenyag:
@codex review

evenyag:
@codex review

evenyag:
@codex review

src/mito2/src/worker/handle_alter.rs
```diff
@@ -288,6 +308,9 @@ fn new_region_options_on_empty_memtable(
                 current_options.append_mode = true;
                 current_options.merge_mode = None;
             }
+            SetRegionOption::MaxRowGroupRowCount(new_row_count) => {
```
discord9, line 332: This can panic for a valid multi-option ALTER. If `max_row_group_row_count` changes while the request also contains an idempotent `append_mode` value (for example, the current value is `false` and the request sets it to `false`), the row-group change sends the request through the slow path. `new_region_options_on_empty_memtable()` then reaches the unconditional assertion above, even though the append-mode option was already accepted as unchanged by the fast-path validation. Please only apply the append-mode transition when its value actually changes instead of asserting that every entry here is a `false -> true` transition.

src/mito2/src/worker/handle_alter.rs
```diff
@@ -246,6 +247,23 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+                            store_api::metadata::InvalidRegionRequestSnafu {
+                                region_id: region.region_id,
+                                err: format!(
+                                    "max_row_group_row_count must be in (0, \
+                                     {MAX_ROW_GROUP_ROW_COUNT_LIMIT}], got {row_count}"
+                                ),
+                            }
+                        );
+                    }
+                    if new_row_count != current_options.max_row_group_row_count {
```
discord9, line 278: A mixed fast/slow option ALTER can leave a partial update after returning an error. TTL/TWCS/auto-flush changes are installed into `version_control` here before the flush and manifest work required by a changed `max_row_group_row_count`. If flush scheduling or the later manifest update fails, the catalog ALTER fails but those in-memory options remain changed. Please stage the complete option set and publish it only after the slow path succeeds (or otherwise roll back the fast changes).

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -953,6 +955,7 @@ impl ParquetReaderBuilder {
                 Ok(apply_output) => RowGroupSelection::from_inverted_index_apply_output(
                     row_group_size,
                     num_row_groups,
+                    total_row_count,
```
discord9: Please validate `apply_output.total_row_count` against the Parquet row count before applying this selection. If an incomplete/stale index covers fewer rows than the SST, replacing its row count with the Parquet count makes the unrepresented tail look searched-but-empty after `fill_missing_row_groups`, which can silently prune matching rows. On a mismatch, the safe behavior is to skip this index and fall back to scanning.

evenyag:
@codex review

fengjiachun (approved):
LGTM

src/store-api/src/mito_engine_options.rs
```diff
@@ -99,6 +103,7 @@ pub fn is_mito_engine_option_key(key: &str) -> bool {
         APPEND_MODE_KEY,
         MERGE_MODE_KEY,
         SST_FORMAT_KEY,
+        MAX_ROW_GROUP_ROW_COUNT,
```
discord9, line 106: Please reject this option on logical Metric tables, or define explicit propagation/conflict semantics for their shared physical region. Adding the key to the generic Mito allowlist lets a logical table accept, persist, and display `max_row_group_row_count`, but `MetricEngine::create_logical_regions()` never validates or applies logical-region options to the physical data region. As a result, even an invalid value such as `0` can be accepted and shown by `SHOW CREATE TABLE` while having no effect.

src/mito2/src/worker/handle_alter.rs
```diff
@@ -297,11 +322,15 @@ fn new_region_options_on_empty_memtable(
...
-                current_options.merge_mode = None;
+                if *new_append_mode != current_options.append_mode {
+                    // Safety: handle_alter_region_options_fast() has validated that the only
+                    // supported transition is from false to true.
+                    current_options.append_mode = *new_append_mode;
+                    current_options.merge_mode = None;
+                }
+            }
+            SetRegionOption::MaxRowGroupRowCount(new_row_count) => {
+                current_options.max_row_group_row_count = *new_row_count;
```
discord9, line 333: This altered value is not propagated to already-open follower regions. ALTER is sent only to the leader, while the replicated `RegionChange` contains metadata, SST format, and append mode but not `max_row_group_row_count`; follower sync therefore keeps its old runtime options. During failover, `OpenRegion` returns success without reapplying catalog options when the follower is already open, and promotion carries no options, so the promoted region can continue writing with the old/default row-group size. Please replicate this setting through a backward-compatible manifest field or refresh catalog options during follower sync/promotion.

evenyag: This problem is out of this PR's scope. @WenyXu

src/store-api/src/mito_engine_options.rs
evenyag: Out of this PR's scope.

PR #8449, opened 2026-07-08
fix: compare all LoggingOptions fields in PartialEq

raphaelroshan:

The hand-written `impl PartialEq for LoggingOptions` only compared 7 of the struct's 11 fields, silently omitting four real fields:

- `log_format`
- `max_log_files`
- `otlp_export_protocol`
- `otlp_headers`

As a result, two `LoggingOptions` that differ only in those fields compared as equal. This masks config-parsing regressions: the config round-trip test in `src/cmd/tests/load_config_test.rs` asserts with `assert_eq!`, so a bug that mis-parses any of these four fields would still pass CI unnoticed.

### Fix

Replace the manual `PartialEq` impl with `#[derive(PartialEq)]` on the struct, so the comparison always covers every field (and stays correct as fields are added). Every field implements `PartialEq` (`String`, `Option<_>`, `usize`, `bool`, `HashMap`, `LogFormat` derives it, `OtlpExportProtocol` derives it, and `TracingSampleOptions` has a manual `PartialEq`), so the derive compiles.

I also removed the manual `impl Eq for LoggingOptions {}`. `Eq` cannot be derived here (`TracingSampleOptions` transitively contains an `f64`), and it turned out to be unused: `LoggingOptions` is never used as a `HashMap`/`HashSet` key, no code has a `T: Eq` bound over it, and its only container `GreptimeOptions<T>` (see `src/cmd/src/options.rs`) derives just `PartialEq`, not `Eq`. So dropping `Eq` is safe rather than keeping a redundant impl.

### Test

Added `test_logging_options_partial_eq_all_fields` in `logging.rs`, which starts from `LoggingOptions::default()` and asserts that mutating each of the four previously-ignored fields makes the value `!=` the default. This test fails on `main` (the old impl reports equality) and passes after the fix.

## PR Checklist

### DIFF
--- a/src/common/telemetry/src/logging.rs
+++ b/src/common/telemetry/src/logging.rs
@@ -237,7 +237,7 @@ enum TraceState {
 }
 
 /// The logging options that used to initialize the logger.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(default)]
 pub struct LoggingOptions {
     /// The directory to store log files. If not set, logs will be written to stdout.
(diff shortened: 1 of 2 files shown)


src/common/telemetry/src/logging.rs
```diff
@@ -237,7 +237,7 @@ enum TraceState {
 }
 
 /// The logging options that used to initialize the logger.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
```
fengjiachun, line 240: This now makes `cmd::load_config_test` fail because several expected `LoggingOptions` still leave `otlp_export_protocol` as `None`, while the example configs parse it as `Some(Http)`. Could you update the datanode/frontend/metasrv/standalone expected blocks to match the flownode case?

fengjiachun (approved):
LGTM

PR #8466, opened 2026-07-10
feat: skip oversized compaction tasks

### LINKED ISSUE #7325: Compaction picker should be aware of memory limit
I'll draft an issue for that case to avoid creating compactions tasks which will never succeed due to memory limit.

_Originally posted by @v0y4g3r in https://github.com/GreptimeTeam/greptimedb/pull/7305#discussion_r2579190040_

fengjiachun:
## What's changed

- Check the estimated local compaction memory against the configured compaction memory limit before creating and submitting the local task.
- Skip picker outputs that can never fit in the total compaction budget, notify waiters with `Ok(0)`, and leave the region without a running compaction status.
- Keep the oversized skip observable via `greptime_mito_compaction_memory_rejected_total{type="oversized"}`.
- Add targeted scheduler coverage with mock SST `max_row_group_uncompressed_size` values.

## Why

The compaction memory manager currently gates execution at task start. If the selected `PickerOutput` is already larger than the total limit, the local task can never acquire memory, so creating and submitting it only produces a doomed task.

This keeps the existing wait/fail policy for cases where the total request can fit but memory is temporarily unavailable. It only skips the impossible `estimated_bytes > limit_bytes` case.

Assumption: remote compaction keeps the existing behavior. This guard applies to local compaction and local fallback, because the configured local memory manager is what enforces this budget.

The skip path intentionally follows the existing scheduler no-task result and notifies waiters with `Ok(0)`. Returning an explicit manual-compaction error would require splitting "picker found nothing to compact" from "picker found an impossible local task" in the scheduler result; this PR keeps that API behavior unchanged and relies on the warning plus the `type="oversized"` rejected counter for the operational signal.

Fixes #7325

## Validation

- `cargo fmt --package mito2`
- `git diff --check`
- `cargo nextest run -p mito2 test_schedule_compaction_skips_task_exceeding_memory_limit test_schedule_compaction_returns_true_when_task_scheduled`
- `cargo nextest run -p mito2 test_schedule_on_finished test_on_compaction_finished_next_schedule_noop_removes_status test_on_compaction_finished_replays_pending_ddl_after_manual_noop test_schedule_compaction_does_not_publish_status_when_schedule_fails`

### DIFF
--- a/src/mito2/src/compaction.rs
+++ b/src/mito2/src/compaction.rs
@@ -64,7 +64,9 @@ use crate::error::{
     RegionClosedSnafu, RegionDroppedSnafu, RegionTruncatedSnafu, RemoteCompactionSnafu, Result,
     TimeRangePredicateOverflowSnafu, TimeoutSnafu,
 };
-use crate::metrics::{COMPACTION_STAGE_ELAPSED, INFLIGHT_COMPACTION_COUNT};
+use crate::metrics::{
+    COMPACTION_MEMORY_REJECTED, COMPACTION_STAGE_ELAPSED, INFLIGHT_COMPACTION_COUNT,
+};
 use crate::read::FlatSource;
 use crate::read::flat_projection::FlatProjectionMapper;
 use crate::read::read_columns::ReadColumns;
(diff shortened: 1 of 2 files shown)


fengjiachun:
@codex review

evenyag (approved):

PR #8467, opened 2026-07-10
feat: more region lifecycle hooks

sunng87:

This patch adds more region lifecycle hooks for downstream extensions, including open/close/drop.

## PR Checklist

### DIFF
--- a/src/mito2/src/engine/region_hook.rs
+++ b/src/mito2/src/engine/region_hook.rs
@@ -27,7 +27,7 @@
 //!   (file additions/removals, schema changes, truncation, partition expression changes, etc.).
 //!
 //! - [`on_region_closed`] / [`on_region_dropped`] / [`on_region_files_removed`]:
-//!   Region **lifecycle** callbacks for close, logical drop, and physical file removal.
+//!   Region **lifecycle** callbacks for open, close, logical drop, and physical file removal.
 //!   See [Region lifecycle](#region-lifecycle) below.
 //!
 //! Hook implementations are registered via the [`Plugins`](common_base::Plugins) system:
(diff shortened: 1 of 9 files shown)


src/mito2/src/engine/region_hook.rs
```diff
@@ -26,6 +26,10 @@
 //!   Receives the full [`RegionMetaActionList`] so consumers can inspect what changed
 //!   (file additions/removals, schema changes, truncation, partition expression changes, etc.).
 //!
+//! - [`on_region_closed`] / [`on_region_dropped`] / [`on_region_files_removed`]:
```
shuiyisong: This lifecycle list omits `on_region_opened`, and the overview above still says the trait provides two methods even though it now exposes six. That makes the public hook inventory internally inconsistent. Please include `on_region_opened` here and update the overview count/wording.

```suggestion
//! - [`on_region_opened`] / [`on_region_closed`] / [`on_region_dropped`] / [`on_region_files_removed`]:
```

src/mito2/src/engine/region_hook.rs
```diff
@@ -228,6 +271,79 @@ pub trait RegionHook: Send + Sync + Debug {
     ) {
         let _ = (region_id, action_list, manifest_version);
     }
+
+    /// Called after a region is **opened** (created or reopened) and registered
```
shuiyisong: `on_region_opened` is invoked before `insert_region` on both the create and open paths, so the region is not registered as active when this callback runs. This also contradicts the later `before the region is registered/acknowledged` wording. Please describe the callback as running after open/create succeeds but before registration.

src/mito2/src/engine/region_hook.rs
```diff
@@ -228,6 +271,79 @@ pub trait RegionHook: Send + Sync + Debug {
...
+
+    /// Called after a region is **closed** via a close request.
+    ///
+    /// The region is removed from the engine's active set, but its data files,
+    /// manifest, and WAL state are **preserved**; the region may be reopened
+    /// later. Fires once per successful close, after the region's background
+    /// tasks (flush/compaction) have been stopped.
+    ///
+    /// Does **not** fire when a region is dropped (see [`on_region_dropped`])
+    /// or for follower/catchup regions.
```
shuiyisong: This follower exclusion is not enforced: `remove_region` invokes `on_region_closed` unconditionally for any registered region, and follower regions are closed through that same path. If follower closes are intended to be observable, remove this clause; otherwise the callback needs an explicit role gate.

shuiyisong (commented):
Non-blocking documentation findings; the lifecycle implementation otherwise looks consistent.

shuiyisong (approved):
Rest LGTM

PR #8469, opened 2026-07-10
ci: gate releases on compat and query regression

discord9:

Follow-up to the recent compatibility CI and query-regression workflow work.

This PR makes the release workflow run compatibility and query performance regression validation before publishing artifacts externally.

The release workflow now:

- prepares immutable validation refs for release checks;
- runs the SQL compatibility test against the built linux-amd64 release artifact;
- reuses the persistent-cache query regression workflow through `workflow_call`;
- gates DockerHub image publishing and GitHub release publishing on validation results;
- triggers linux artifact builds automatically for manual `publish_github_release` / `release_images` runs.

Automatic tag and scheduled releases always run all validation. Manual dispatches have one auditable `release_validation` policy with these options:

- `all`
- `skip-compat`
- `skip-query-regression`
- `skip-all`

Publishing only accepts a skipped validation job when that exact check was explicitly disabled by the selected policy. Failed, cancelled, and unexpectedly skipped validation still block publishing. The effective policy, actors, event, candidate SHA, and baseline are written to the workflow summary.

For compatibility validation, the job uses the linux-amd64 release artifact as the candidate `greptime` binary, builds `sqlness-runner`, then runs:

```text
python3 .github/scripts/run-compat.py --preserve-state
```

For query regression validation, reusable calls must pass an immutable full candidate SHA. The existing fail-closed candidate resolver verifies that exact SHA while preserving the latest persistent cache, pinned runner image, and PR merge identity checks. Trusted helper scripts come from the verified base SHA for PRs and the verified candidate SHA for non-PR runs.

If an automatic formal-tag release fails validation, it can be manually dispatched on the same tag with an explicit skip policy. The retry reuses and validates the existing tag, does not recreate it, blocks retries of older-than-latest formal tags, and runs the normal downstream version bumps after publishing.

Validation performed:

```text
git diff --check
bash -n .github/scripts/create-version.sh
uv run python -m py_compile .github/scripts/run-compat.py .github/scripts/query-regression-run.py .github/scripts/query-regression-summary.py .github/scripts/query-regression-pr-metadata.py
```

Focused shell checks cover push tags, manual exact-tag recovery, manual non-tag versions, shared dev/nightly behavior without release opt-in, and invalid tags. `actionlint` was not available locally.

## PR Checklist

- [x] I have written the necessary rustdoc comments. (not applicable, workflow-only)
- [x] I have added the necessary unit tests and integration tests. (workflow/shell validation)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,7 @@ name: Release
 # There are two kinds of formal release:
 # 1. The tag('v*.*.*') push release: the release workflow will be triggered by the tag push event.
 # 2. The scheduled release(the version will be '${{ env.NEXT_RELEASE_VERSION }}-nightly-YYYYMMDD'): the release workflow will be triggered by the schedule event.
+# A failed tag push release can be manually dispatched on the same tag to recover it without changing the formal release version.
 on:
   push:
     tags:
@@ -11,7 +12,7 @@ on:
     # At 00:00 on Monday.
     - cron: '0 0 * * 1'
   workflow_dispatch: # Allows you to run this workflow manually.
-    # Notes: The GitHub Actions ONLY support 10 inputs, and it's already used up.
+    # Notes: The GitHub Actions ONLY support 10 inputs; this final input is the tenth.
     inputs:
       linux_amd64_runner:
         type: choice
@@ -74,6 +75,16 @@ on:
         description: Build and push images to DockerHub and ACR
         required: false
         default: false
+      release_validation:
+        type: choice
+        description: Manual policy for accepted perf regressions or runner failures; automatic releases always run all validation
+        required: true
+        default: all
+        options:
+          - all
+          - skip-compat
+          - skip-query-regression
+          - skip-all
 
 # Use env variables to control all the release process.
 env:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -292,13 +292,129 @@ jobs:
...
+    needs: [
+      allocate-runners,
+      prepare-release-validation,
+    ]
+    uses: ./.github/workflows/query-regression.yml
+    secrets: inherit
+    with:
+      case: all
+      base_ref: ${{ needs.prepare-release-validation.outputs.previous-release-tag }}
+      candidate_ref: ${{ needs.prepare-release-validation.outputs.candidate-ref }}
```
fengjiachun, line 464: After rebasing onto current main (#8474), this input is ignored for tag and scheduled callers because `query-regression.yml` only sets `CANDIDATE_REF` when `github.event_name == 'workflow_dispatch'`. It then fetches the remote default branch, so query regression may validate a different commit than the release candidate. Please make every non-PR reusable invocation consume `inputs.candidate_ref`.

.github/workflows/release.yml
```diff
@@ -292,13 +292,129 @@ jobs:
         run: |
           echo "build-windows-result=success" >> $Env:GITHUB_OUTPUT
 
+  prepare-release-validation:
+    name: Prepare release validation
+    if: ${{ inputs.publish_github_release || inputs.release_images || github.event_name == 'push' || github.event_name == 'schedule' }}
+    needs: [
+      allocate-runners,
+    ]
```
evenyag, line 313: Do we have a way to skip this?

.github/workflows/release.yml
discord9: Fixed in 31f33636be. After rebasing onto the latest persistent-cache/fail-closed runner, every non-PR invocation now consumes `inputs.candidate_ref || github.ref`. Reusable calls are detected by comparing the documented `job.workflow_ref` and `github.workflow_ref`, require a full immutable SHA, and verify that `FETCH_HEAD` exactly matches it. Direct manual dispatch still supports an explicit branch/ref fallback. Trusted helper scripts are extracted from the verified candidate SHA for non-PR runs.

.github/workflows/release.yml
discord9: Added in 31f33636be. Manual release dispatch now has a single auditable `release_validation` choice: `all`, `skip-compat`, `skip-query-regression`, or `skip-all`. Automatic tag/scheduled releases always force `all`; publishing accepts `skipped` only when that exact check was explicitly disabled, while failures/cancellations/unexpected skips remain blocking. A failed formal-tag run can be manually retried on the same tag with a skip policy; the workflow validates/reuses the tag, avoids recreating it, prevents older tags from moving `latest` backward, and runs downstream bumps.

fengjiachun (approved):
LGTM

PR #8471, opened 2026-07-10
refactor: suppress region hook notifications for staging-only manifest writes

sunng87:
## Summary

Only fire `on_manifest_updated` for writes to the live (normal) manifest directory. Writes to the staging directory (enter staging, operations during staging, the intermediate apply-staging `RegionEdit`) are suppressed — their effects accumulate and are delivered in a single notification when `exit_staging_on_success` promotes all staged actions to the live manifest.

This simplifies hook implementations that only care about "live" state, since they no longer receive intermediate staging notifications they cannot act on, and no file-list information is lost (the promote step carries all accumulated files from staging operations).

## Key changes

- **`PendingManifestHook`** gains an `is_staging` field; `fire()` is a no-op when `is_staging` is `true`
- **`ManifestContext::update_locked`** threads its `is_staging` parameter through to `PendingManifestHook`
- **Updated docs**: module-level docs and coverage table now clearly state which paths fire the hook and which are suppressed

## Motivation

Hook implementations that track "live" manifest state (e.g. Iceberg manifest generation) previously received intermediate staging notifications they had to ignore. These changes are not visible to readers until `exit_staging_on_success` promotes them, so the hook should not observe them individually. The promote notification includes the complete merged action list with all file additions/removals from staging, so no information is lost.

## Verification

- `cargo check -p mito2` passes
- All existing tests unchanged
- No behavioral change for non-staging operations

### DIFF
--- a/src/mito2/src/engine/region_hook.rs
+++ b/src/mito2/src/engine/region_hook.rs
@@ -22,9 +22,13 @@
 //!   Provides per-file [`SstInfo`] + [`FileMeta`]; metadata richness varies by path
 //!   (see [`SstFileInfo`] and the coverage footnote).
 //!
-//! - [`on_manifest_updated`]: Fires after **any** manifest write is successfully committed.
-//!   Receives the full [`RegionMetaActionList`] so consumers can inspect what changed
-//!   (file additions/removals, schema changes, truncation, partition expression changes, etc.).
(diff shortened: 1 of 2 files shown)


shuiyisong (approved):
LGTM

PR #8472, opened 2026-07-10
feat: split metrics value into integer and float columns

waynexia:
Signed-off-by: Ruihang Xia <waynestxia@gmail.com>

TL;DR: store actual integer value as integer, in `__metric_int_<original-column-name>` col, and other non-integer float values remain float type. This can bring ~40% storage size reduction with my test data (real data generated by nodeexporter). From `12,705,806,208` MB to `7,428,214,444` MB.

In my test data, >94% are actual integers:

image

This makes sense because from Prometheus semantics, metrics like counter, histogram bucket etc can only generate actual integer, only a few types of metrics can possibly be non-integer. 

We only need to split actual-integer and float into two physical columns to make use of this priori knowledge. Integer can easily achieve way much better encoding efficiency than floats.

For write performance, this newly introduced split way has almost the same encoding time with original plain float64 manner:

image

And for query performance, it also has no regression:

image

Notice this is not a forward-compatible change, once upgraded in this manner, new generated data files can't be handled by previous greptimedb.

## PR Checklist


discord9:
Two points seem worth addressing before merging:

1. **Rolling upgrade / rollback can silently turn integer samples into `NULL`.** The split is enabled by default for every new non-`last_non_null` metric region, and flush nulls the original Float64 cell once the value is stored in `__metric_int_*`. A pre-feature binary has no coalescing path, so if the region is migrated to or reopened by an older node, it will serve `NULL` rather than fail explicitly. This can also be triggered by a downgrading flush immediately before handoff. Since this changes persisted SST semantics, could we gate the first split write on a persisted cluster capability/format version and prevent assignment to unsupported nodes afterward? We should also have old→new mixed-SST and new→old rollback coverage; rollback must either preserve values or reject ownership before serving reads.

2. **The `primary_key` SST writer does not appear to perform the split.** `FlatWriteFormat::convert_batch()` calls `split_metric_value_columns()`, while `PrimaryKeyWriteFormat::convert_flat_batch()` is reached directly from `write_all_flat_as_primary_key()` and never calls it. A supported metric table with `sst_format = 'primary_key'` therefore keeps fresh integral samples in the Float64 column and leaves the companion column null, so the feature/storage benefit silently depends on SST format. Could we apply the split before both converters and add physical-column assertions for flat, primary-key, and format-switch/compaction cases?

discord9:
I ran a preliminary, bounded investigation to check the integer-heavy premise behind this PR. The current results are:

- **Scraped values:** six infrastructure-oriented golden/mock sources contain 3,096 scalar samples, of which **93.86% are exact integers**. Two public Prometheus demo environments contain 2,321 sampled scalars, of which **94.79% are exact integers**. In the live sample, ordinary (non-derived) scalars are 59.72% of rows and **92.86% of those are exact integers**. Giving each source equal weight produces similar means: 96.32% for fixtures and 95.10% for the two demos.
- **Source semantics:** I also audited 15 pinned OSS projects across host/Kubernetes infrastructure, databases/TSDBs, observability, networking and exporters. Taking ten consecutive traceable metric families per project, **113 of 150 families (75.3%) had integer-constrained logical source semantics**.

This is strong evidence that integer values are a useful storage prior for infrastructure-oriented Prometheus workloads, and it is consistent with the size reduction reported here. It supports a transparent per-value/per-chunk integer path with lossless Float64 fallback.

There are important boundaries: this was a purposive, infrastructure-heavy survey—not an estimate of all Prometheus workloads. Duration, CPU seconds, load, ratios, quantiles, histogram sums and special Float64 values remain essential counterexamples; classic histogram bucket/count rows can also inflate the observed integer share. Also, Prometheus counters are not inherently integer-valued: client APIs generally allow non-negative fractional increments, so the integer decision still has to come from the actual value/source rather than the metric type alone.

I have prepared a broader 186-project candidate registry, but its family-level source audit is not complete yet, so I am not treating that registry as additional distribution evidence.

discord9:
I ran the PR #8444 remote-write fixture over a normalized nine-distribution matrix using release `dev-tools` builds of base `e6472fd1` and tested head `06631be8`. The current head `9f8ffd7c` only adds fuzz expectations, so it does not change these storage-path results.

Each base/head target wrote and observed exactly **29,491,200 rows**, completed **10/10 flushes**, and passed query, footer, and direct-read execution checks.

## Write path and storage

| Distribution | Remote-write elapsed | Effective rows/s | Flush median | Total SST delta |
|---|---:|---:|---:|---:|
| constant, integer-only | 16.72s → 15.86s (**-5.14%**) | 1.764M → 1.860M (**+5.41%**) | 203.20ms → 190.76ms (**-6.12%**) | **+8.11%** |
| linear, step 0.125 | 17.39s → 17.60s (**+1.20%**) | 1.696M → 1.676M (**-1.18%**) | 204.49ms → 221.53ms (**+8.33%**) | **+249.85%** |
| mixed signal/repeated | 16.63s → 16.42s (**-1.27%**) | 1.773M → 1.796M (**+1.28%**) | 203.07ms → 215.67ms (**+6.21%**) | **+78.08%** |
| modulo, four integer values | 15.01s → 15.82s (**+5.43%**) | 1.965M → 1.864M (**-5.15%**) | 187.16ms → 196.57ms (**+5.03%**) | **+1.55%** |
| quantized signal | 15.49s → 16.48s (**+6.38%**) | 1.904M → 1.790M (**-6.00%**) | 191.12ms → 200.73ms (**+5.03%**) | -0.33% |
| run length, integer | 15.48s → 16.31s (**+5.35%**) | 1.905M → 1.808M (**-5.08%**) | 199.54ms → 193.01ms (**-3.28%**) | **-34.74%** |
| seeded random | 17.13s → 17.98s (**+4.99%**) | 1.722M → 1.640M (**-4.75%**) | 246.56ms → 240.26ms (**-2.56%**) | **-32.26%** |
| sporadic stalls | 16.10s → 17.05s (**+5.89%**) | 1.832M → 1.730M (**-5.56%**) | 193.24ms → 215.25ms (**+11.39%**) | **-14.19%** |
| unique integer counter | 16.36s → 17.16s (**+4.88%**) | 1.802M → 1.718M (**-4.66%**) | 237.31ms → 196.92ms (**-17.02%**) | **-97.14%** |

Remote-write elapsed is `remote_write.aggregate.elapsed_seconds`; it excludes the later SQL/footer/read-bench phases. Throughput is `29,491,200 / elapsed`. Flush is the median of all ten explicit flushes.

## Read path

| Distribution | SQL/TQL median | Parquet direct¹ | Direct scan median |
|---|---:|---:|---:|
| constant, integer-only | 293.95ms → 358.05ms (**+21.81%**) | 2.56ms → 2.33ms (**-8.95%**) | 150.25ms → 149.76ms (**-0.32%**) |
| linear, step 0.125 | 260.70ms → 283.44ms (**+8.72%**) | 5.92ms → 23.56ms (**+298.13%**) | 186.30ms → 388.95ms (**+108.77%**) |
| mixed signal/repeated | 269.14ms → 250.12ms (**-7.07%**) | 8.41ms → 22.17ms (**+163.73%**) | 193.68ms → 324.65ms (**+67.62%**) |
| modulo, four integer values | 231.15ms → 248.77ms (**+7.63%**) | 5.59ms → 2.42ms (**-56.68%**) | 173.05ms → 147.55ms (**-14.74%**) |
| quantized signal | 251.09ms → 325.69ms (**+29.71%**) | 7.80ms → 13.45ms (**+72.38%**) | 198.50ms → 244.59ms (**+23.22%**) |
| run length, integer | 222.63ms → 240.26ms (**+7.92%**) | 5.65ms → 3.25ms (**-42.50%**) | 175.53ms → 144.79ms (**-17.51%**) |
| seeded random | 247.84ms → 378.85ms (**+52.86%**) | 37.17ms → 36.02ms (**-3.08%**) | 456.38ms → 453.84ms (**-0.56%**) |
| sporadic stalls | 256.61ms → 299.08ms (**+16.55%**) | 6.68ms → 15.97ms (**+139.04%**) | 187.36ms → 268.40ms (**+43.25%**) |
| unique integer counter | 290.46ms → 260.97ms (**-10.15%**) | 37.95ms → 1.98ms (**-94.78%**) | 483.75ms → 149.09ms (**-69.18%**) |

¹ Mean of four SST-level medians; each SST median uses seven iterations. Direct scan is the median of seven iterations.

## Interpretation

- **Unique integer counter is the clear win:** storage -97%, parquet -95%, scan -69%, SQL -10%. Remote-write throughput is about 5% lower, while flush median improves 17%.
- **Run length also benefits overall:** storage -35% and direct reads improve 18–43%, at the cost of about 5% lower write throughput and an 8% SQL regression in this run.
- **Seeded random saves 32% storage** with nearly unchanged direct reads, but write throughput drops about 5% and SQL regresses 53% in this run.
- **Linear and mixed-repeated are major adverse cases:** storage grows 250% and 78%, parquet grows 298% and 164%, and scan time grows 109% and 68%. Their write time changes little, so the primary cost is persisted layout/reconstruction.
- Constant, modulo, and quantized provide little or no net storage benefit and mixed read/write results.

The split is extremely effective for high-cardinality integral data, but not uniformly beneficial. A ratio/cost-based fallback should estimate the combined integer column + float column + validity/metadata cost and keep the original layout when splitting is not expected to win.

This is **one sequential run**. Write, flush, SQL, and read deltas are directional and need repeated/interleaved measurements before stable performance claims. Exact row counts, successful flushes, encoding metadata, and SST byte sizes are deterministic for these generated artifacts.

There are also correctness limits: the fixture verifies row counts and successful scans, not value-by-value Float64 equality. It has no strictly all-fractional case, special values (`NaN`, infinities, signed zero), integer-boundary cases, old-reader/new-SST rollback, restart/compaction exact round trips, or primary-key writer-path coverage. The earlier compatibility and primary-key-path concerns therefore remain open.

At minimum I would add exact-value round trips, all-fractional/special-value cases, old/new mixed-SST rollback gating, flat/primary-key writer assertions, and storage thresholds over both physical value columns for all distributions.

discord9:
## Follow-up: why the two Parquet regressions are so large

I reran the two cases with release binaries against base `e6472fd1` and candidate `06631be8`, preserving and inspecting both the original `greptime_value` column and the new `__metric_int_greptime_value` companion column.

| Case | Base SST bytes | Candidate SST bytes | Change |
|---|---:|---:|---:|
| linear, step = 0.125 | 8,657,588 | 30,812,362 | **+255.90%** |
| mixed signal with repeated zero | 20,970,434 | 37,782,402 | **+80.17%** |

Within each case, base and candidate had the same file count, physical row count, and row-group count, so the byte decomposition is comparable.

### 1. The linear case is actually highly repetitive and dictionary-friendly

The generated value is:

```text
value = series_idx % 97 + local * 0.125
      = (8 * (series_idx % 97) + local) / 8
```

For one 1,440-sample flush chunk, all 2,949,120 rows therefore contain at most:

```text
1,440 + 8 * 96 = 2,208 distinct values
```

That is roughly 1,336 occurrences per value. Despite the name "linear", this is a low-cardinality 1/8 lattice across the whole chunk.

The footer sizes match the dictionary math almost exactly. For a 102,400-row group, at most 2,208 dictionary entries need 12-bit indices:

```text
102,400 * 12 / 8 + 2,208 * 8 ~= 171 KB
```

The base reports about 171.5 KB uncompressed per row group with `RLE_DICTIONARY`. After the split, 7/8 of rows remain in the float column, but dictionary is disabled:

```text
102,400 * 7/8 * 8 ~= 716.8 KB
```

With definition-level/page overhead, the candidate reports about 730 KB per row group. The full result is:

- float compressed bytes: 6,826,190 -> 28,398,062
- float uncompressed bytes: 54,249,614 -> 231,252,283
- new int companion compressed bytes: only 440,979

Of the 22,154,774-byte total increase, **97.37% comes from the float column losing dictionary**, while the integer companion contributes only 1.99%.

So this case is useful as a dictionary-friendly regression sentinel, but it should not be interpreted as a generic high-cardinality continuously increasing signal.

### 2. The mixed case is the more interesting trade-off case

The mixed generator writes zero every fifth sample and otherwise uses the same 1/8 signal. Exact integers occur when `local` is divisible by 5 or 8:

```text
1/5 + 1/8 - 1/40 = 30%
```

Its 16,811,968-byte increase decomposes as:

- float dictionary loss: 4,267,327 bytes (**25.38%**)
- new integer companion: 12,414,294 bytes (**73.84%**)
- footer/index/other residual: 130,347 bytes (0.78%)

Here the integer stream repeatedly alternates between zero and increasing integer spikes. This is a poor shape for a fixed `DELTA_BINARY_PACKED` policy because it produces large positive and negative deltas. By comparison, the linear case's sparse integer stream is close to monotonic and DELTA compresses it to only 0.44 MB.

### Interpretation

The split itself can still be useful for integer-heavy monotonic counters. The issue exposed by these cases is the unconditional policy in the writer:

```text
float -> PLAIN, dictionary disabled
int   -> DELTA_BINARY_PACKED, dictionary disabled
```

- linear demonstrates that unconditionally disabling float dictionary can destroy a large existing compression win;
- mixed demonstrates that a fixed DELTA encoding is unsuitable for reset-heavy/repeated integer distributions.

The safer direction seems to be:

1. keep or restore dictionary eligibility for the float half;
2. choose the integer encoding from the actual distribution (DELTA for monotonic/small-delta data, dictionary for repeated/reset-heavy data);
3. ideally estimate `float + int + validity/metadata` cost per row group and split only when it is expected to beat the original float representation.

Without such gating, this remains a distribution-sensitive storage regression rather than a generally safe optimization.

src/mito2/src/sst/parquet/metric_value_split.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    for row in 0..float_array.len() {
+        let Some(value) = logical_value(float_array, int_array, row) else {
+            float_builder.append_null();
+            int_builder.append_null();
+            continue;
+        };
+
+        if let Some(int_value) = integer_value(value) {
+            float_builder.append_null();
```
discord9: **[blocker] Preserve backward readability during rolling upgrades**

For an integral sample, this nulls the original Float64 cell and stores the value only in the new Int64 companion. A pre-PR binary has no coalescing logic and projects only the logical Float64 column, so it reads the sample as `NULL` after a new node flushes or compacts it. PromQL may then drop the sample during rolling upgrade, failover, or rollback. Please add version/capability gating, or retain a backward-readable Float64 representation until old readers are unsupported; this also needs a mixed-version compatibility case.

src/mito2/src/flush.rs
```diff
@@ -876,7 +881,7 @@ fn memtable_flat_sources(
         };
 
         for (_range_id, range) in ranges {
-            if let Some(encoded) = range.encoded() {
+            if !disable_encoded_ranges && let Some(encoded) = range.encoded() {
```
discord9: **[bug] Include decoded encoded ranges in `total_rows`**

When value splitting sets `disable_encoded_ranges`, this condition sends encoded ranges through the normal iterator path. However, `total_rows` above still filters those ranges out. The subsequent `rows_remaining -= range_rows` therefore underflows as soon as an encoded range is processed: overflow-checked builds panic, while release builds wrap to a huge value and break SST grouping. When encoded ranges are disabled, `total_rows` must include all ranges. Please add a multi-range encoded + non-encoded regression test.

src/store-api/src/metric_engine_consts.rs
```diff
@@ -83,7 +84,19 @@ pub const MANIFEST_INFO_EXTENSION_KEY: &str = "MANIFEST_INFO";
...
+}
+
+/// Returns the physical integer companion column name for a metric value column.
+pub fn metric_engine_value_int_column_name(value_column_name: &str) -> String {
+    format!("{DATA_SCHEMA_VALUE_INT_COLUMN_PREFIX}{value_column_name}")
+}
+
+/// Returns true if the column is a physical integer companion for a metric value column.
+pub fn is_metric_engine_value_int_column(name: &str) -> bool {
+    name.starts_with(DATA_SCHEMA_VALUE_INT_COLUMN_PREFIX)
```
discord9: **[compatibility] Do not identify companion columns by prefix alone**

Before this PR, `__metric_int_` was not reserved, so existing physical metric tables may legally contain fields with this prefix. This broad `starts_with` check now causes such columns to be treated as internal and hidden by the table adapter/read paths after upgrade. Please require a validated Float64/Int64 companion pair or an explicit persisted metadata marker, and cover an existing-schema collision in compatibility tests.

discord9 (commented):
Codex-assisted review on the current head. Codex already raised the signed-zero round-trip issue in discussion_r3558901627; I confirmed it is still present. I found three additional correctness/compatibility issues below.

discord9:
### Architectural concern

Beyond the inline correctness findings from the Codex-assisted review, I think the current design leaks an SST encoding optimization into too many layers. The companion column is treated as a real schema column, so create/alter, row writes, bulk inserts, flush, reads, filter rewriting, the DataFusion table adapter, and the PromQL planner all need special handling for `__metric_int_*`. The rolling-upgrade, prefix-collision, encoded-range, and signed-zero issues are symptoms of this cross-layer contract.

Ideally, the region should continue exposing the original logical `Float64` schema, while the split is represented as an explicit, versioned Mito SST encoding policy. The Mito writer would own the physical split and the Mito reader would reconstruct the logical value; WAL/memtable, Metric Engine APIs, DataFusion, and PromQL would only see the logical column. This would also provide one place for capability/version gating during rolling upgrade and rollback.

If changing the physical schema is unavoidable, I still think it needs a centralized virtual-column abstraction with explicit persisted metadata, rather than recognition by the `__metric_int_` naming convention and per-layer patches. Given the compatibility impact, it would also be safer to ship behind an `experimental_` option with mixed-version compatibility tests before making it the default.

src/store-api/src/metric_engine_consts.rs
waynexia: `__` prefix is reserved

src/mito2/src/engine.rs
```diff
@@ -1281,7 +1290,9 @@ impl RegionEngine for MitoEngine {
         &self,
         region_id: RegionId,
     ) -> std::result::Result<RegionMetadataRef, BoxedError> {
-        self.inner.get_metadata(region_id).map_err(BoxedError::new)
+        let metadata = self.get_physical_metadata(region_id)?;
+        let split_columns = metric_value_columns(&metadata);
+        visible_region_metadata(&metadata, &split_columns).map_err(BoxedError::new)
```
fengjiachun: This metadata API is also used by `RegionMetadataLister` during metasrv table reconciliation. Create/alter responses publish the physical column metadata (including the `__metric_int_*` companions) through `TABLE_COLUMN_METADATA_EXTENSION_KEY`, but this path removes those columns. Reconciliation can therefore rewrite `TableInfo` from the visible schema and, more importantly, treat regions with different physical companion columns as consistent. Please keep reconciliation on physical metadata (or make create/alter/catalog use the same canonical schema) and add a metric-table reconciliation test.

src/mito2/src/metric_value.rs
```diff
@@ -0,0 +1,105 @@
...
+use store_api::region_request::AlterKind;
+use store_api::storage::consts::ReservedColumnId;
+
+use crate::error::{InvalidMetadataSnafu, Result};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct MetricValueColumn {
+    pub(crate) value_index: usize,
+    pub(crate) int_index: usize,
+}
```
evenyag: Document that they are the index in RegionMetadata.

src/metric-engine/src/engine/create.rs
```diff
@@ -305,17 +306,17 @@ impl MetricEngineInner {
...
-                    .collect::<HashMap<_, _>>(),
+                request.column_metadatas.iter().map(|metadata| {
+                    // Safety: previous steps ensure the physical region exist.
+                    let mut column_metadata = (*physical_schema_map
+                        .get(metadata.column_schema.name.as_str())
+                        .unwrap())
+                    .clone();
+                    if metadata.semantic_type == SemanticType::Field {
+                        column_metadata.column_schema = metadata.column_schema.clone();
+                    }
```
evenyag, line 330: Document why we need to override this.

src/metric-engine/src/engine/create.rs
```diff
@@ -563,6 +580,93 @@ impl MetricEngineInner {
...
+            || (metadata.semantic_type == SemanticType::Field
+                && metadata.column_schema.data_type == ConcreteDataType::float64_datatype())
+        {
+            metadata.column_schema.set_nullable();
+        }
+        value_split_enabled || !is_int_column
+    });
+    if !value_split_enabled {
+        return;
+    }
```
evenyag: Should we turn the column to nullable silently here? It means we won't be able to know whether a column is not null after that.

Maybe we can only set int column when the float column is nullable.

src/metric-engine/src/engine/create.rs
```diff
@@ -563,6 +580,93 @@ impl MetricEngineInner {
...
+        .map(|metadata| metadata.column_id)
+        .filter(|column_id| !ReservedColumnId::is_reserved(*column_id))
+        .max()
+        .unwrap_or(0)
+        + 1;
+    let existing_names = column_metadatas
+        .iter()
+        .map(|metadata| metadata.column_schema.name.as_str())
+        .collect::<HashSet<_>>();
+    let mut int_columns = Vec::new();
```
evenyag: Can we allocate ResevedColumnId for these field columns? The metric int column always has `3 << (32 - 2) | column_id` as its column id. So we don't have to maintain next id and can get the int column id easier.

src/mito2/src/metric_value.rs
```diff
@@ -0,0 +1,105 @@
...
+        .enumerate()
+        .filter(|(_, column)| {
+            column.semantic_type == SemanticType::Field
+                && column.column_schema.data_type == ConcreteDataType::float64_datatype()
+        })
+        .filter_map(|(value_index, value_column)| {
+            let int_name = metric_engine_value_int_column_name(&value_column.column_schema.name);
+            let int_index = metadata.column_index_by_name(&int_name)?;
+            let int_column = &metadata.column_metadatas[int_index];
+            if int_column.semantic_type != SemanticType::Field
```
evenyag: non-blocking: If we use a special column id for int columns, we can find the column by that id without generating the int name.

src/mito2/src/metric_value.rs
```diff
@@ -0,0 +1,105 @@
...
+                .name
+                .clone()
+        })
+        .collect::<Vec<_>>();
+
+    let mut builder = RegionMetadataBuilder::from_existing((**metadata).clone());
+    builder
+        .alter(AlterKind::DropColumns { names })
+        .context(InvalidMetadataSnafu)?;
+    builder.build().map(Arc::new).context(InvalidMetadataSnafu)
```
evenyag: Can we make these columns visible for simplicity? Since `__table_id` and `__tsid` are also visible. Maybe promql planner needs to handle the schema specially.

src/mito2/src/engine.rs
discord9: @WenyXu PTAL

src/metric-engine/src/engine/bulk_insert.rs
```diff
@@ -68,15 +71,25 @@ impl MetricEngineInner {
...
+        let metadata = self
+            .mito
+            .get_physical_metadata(region_id)
+            .context(error::MitoReadOperationSnafu)?;
+        let original_num_columns = request.payload.num_columns();
+        request.payload =
+            Self::append_missing_metric_value_int_fields(region_id, &metadata, request.payload)?;
+        if request.payload.num_columns() != original_num_columns {
+            request.raw_data = record_batch_to_ipc(&request.payload)?;
+        }
```
evenyag: Mito has support to fill missing columns for a bulk part. Maybe we don't have to fill them here?

src/mito2/src/sst/parquet/metric_value_split.rs
```diff
@@ -0,0 +1,378 @@
...
+use store_api::metadata::RegionMetadata;
+
+use crate::error::{InvalidRecordBatchSnafu, NewRecordBatchSnafu, Result};
+use crate::metric_value::metric_value_columns;
+
+#[derive(Debug, Clone)]
+pub(crate) struct MetricValueSplitColumn {
+    pub(crate) float_index: usize,
+    pub(crate) int_index: usize,
+}
```
evenyag: Is it duplicated with the MetricValueColumn in src/mito2/src/metric_value.rs?

src/mito2/src/sst/parquet/writer.rs
```diff
@@ -392,18 +398,33 @@ where
...
-            .set_column_compression(op_type_col, Compression::UNCOMPRESSED)
+            .set_column_compression(op_type_col, Compression::UNCOMPRESSED);
+
+        metric_value_split_columns(region_metadata, schema)
+            .into_iter()
+            .fold(builder, |builder, split_column| {
+                let float_col =
+                    ColumnPath::new(vec![schema.field(split_column.float_index).name().clone()]);
+                let int_col =
+                    ColumnPath::new(vec![schema.field(split_column.int_index).name().clone()]);
```
evenyag: Why don't we get the name from region_metadata directly?

src/mito2/src/flush.rs
```diff
@@ -644,11 +645,14 @@ impl RegionFlushTask {
         );
         let field_column_start =
             flat_format::field_column_start(&version.metadata, batch_schema.fields().len());
+        let disable_encoded_ranges =
+            !metric_value_split_columns(&version.metadata, &batch_schema).is_empty();
```
evenyag: We should support value split in the bulk memtable.

Do we need to handle encoded ranges specially here? If we initialize int columns to nulls, then we should also be able to read the float columns. It's still a valid file without splitting.

src/metric-engine/src/value_split.rs
```diff
@@ -0,0 +1,704 @@
...
+
+pub(crate) fn prepare_value_split_scan(
+    region_id: RegionId,
+    request: &mut ScanRequest,
+    physical_metadata: &RegionMetadataRef,
+) -> Result<Option<ValueSplitProjectionMapper>> {
+    let split_columns = metric_value_columns(physical_metadata);
+    if split_columns.is_empty() {
+        return Ok(None);
+    }
```
evenyag, line 240: Can we put the value split behind our row group reader or FlatReadFormat?

Batches returned by the parquet reader are already coalesced. We build index before the split, so the index should work. For statistics, we can merge them for int and float columns in FlatReadFormat. We apply the remaining filters after converting batches, so we don't have to transform filters. (We may skip prefilter for split columns).

All remaining read paths should continue to work as before.

src/metric-engine/src/engine/create.rs
waynexia: Not viable. We may have lots of value columns (in design, at least)

src/mito2/src/metric_value.rs
waynexia: same as above

src/mito2/src/metric_value.rs
waynexia: Sounds good, but it depends on how we're going to use it. I'd categorize it as a storage-level special encoding method, hence all self-contained at the storage level and not leaked outside

src/metric-engine/src/engine/bulk_insert.rs
waynexia: Then we need to change WAL-related paths I guess?

evenyag: Maybe unrelated.

src/metric-engine/src/engine/create.rs
evenyag: More than 2^30? Even in that case, we also don't have enough IDs for all columns.

src/mito2/src/metric_value.rs
evenyag: Then maybe we should hide it behind the SST format, like sequence and op types. But now we have already leaked it in the region metadata and desc table.

src/metric-engine/src/engine/create.rs
```diff
@@ -560,6 +591,44 @@ impl MetricEngineInner {
...
+    let int_columns = column_metadatas
+        .iter_mut()
+        .filter(|metadata| {
+            metadata.semantic_type == SemanticType::Field
+                && metadata.column_schema.data_type == ConcreteDataType::float64_datatype()
+        })
+        .map(|metadata| {
+            // Every split row leaves either the float or integer half null. Logical metadata
+            // retains the original nullability constraint.
+            metadata.column_schema.set_nullable();
```
evenyag, line 611: Should we ensure the column is nullable when creating the physical table?

src/metric-engine/src/value_split.rs
```diff
@@ -0,0 +1,965 @@
...
+    /// Index of the Float64 value in `RegionMetadata::column_metadatas`.
+    pub(crate) value_index: usize,
+    /// Index of its Int64 companion in `RegionMetadata::column_metadatas`.
+    pub(crate) int_index: usize,
+}
+
+/// Derives the internal companion ID without consuming an ID from the user space.
+pub(crate) const fn metric_value_int_column_id(value_column_id: ColumnId) -> ColumnId {
+    value_column_id | INTERNAL_COLUMN_ID_MASK
+}
```
evenyag, line 75: Looks like this occupied the same id as reserved ids like sequence/op type.

The `INTERNAL_COLUMN_ID_MASK` should be `3 << (32 - 2)`. We can wrap some helper functions in ReservedColumnId.

src/metric-engine/src/engine/read.rs
```diff
@@ -53,16 +55,38 @@ impl MetricEngineInner {
...
+            .get_metadata(region_id)
+            .await
+            .context(MitoReadOperationSnafu)?;
+        let visible_metadata = visible_region_metadata(&physical_metadata)?;
+        let mapper = prepare_value_split_scan(
+            region_id,
+            &mut request,
+            &physical_metadata,
+            &visible_metadata,
+        )?;
```
evenyag, line 75: We can check enable_metric_value_split first to avoid getting visible metadata and building the mapper.

src/metric-engine/src/value_split.rs
```diff
@@ -0,0 +1,965 @@
...
+                .and_then(|value| value.value_data.as_ref())
+            else {
+                continue;
+            };
+            let Some(int_value) = integer_value(*value) else {
+                continue;
+            };
+
+            row.values[value_index].value_data = None;
+            row.values[int_index].value_data = Some(ValueData::I64Value(int_value));
```
evenyag, line 213: If the schema doesn't have the int column, we won't add an int column to it. Will it be a problem?

src/metric-engine/src/value_split.rs
```diff
@@ -0,0 +1,965 @@
...
+            )
+            .finish()
+    }
+}
+
+impl DisplayAs for MetricRegionScanner {
+    fn fmt_as(&self, t: DisplayFormatType, f: &mut fmt::Formatter) -> fmt::Result {
+        self.inner.fmt_as(t, f)
+    }
+}
```
evenyag, line 564: Future work: we may expose some metrics of MetricRegionScanner here.

src/metric-engine/src/value_split.rs
```diff
@@ -0,0 +1,965 @@
...
+
+        let mut df_record_batch =
+            DfRecordBatch::try_new(self.working_schema.arrow_schema().clone(), arrays)
+                .context(NewDfRecordBatchSnafu)?;
+        for (column_index, column) in self.columns.iter().enumerate() {
+            for evaluator in &column.simple_filters {
+                df_record_batch = apply_value_filter(df_record_batch, column_index, evaluator)?;
+            }
+        }
+        let batch = RecordBatch::from_df_record_batch(self.working_schema.clone(), df_record_batch);
```
evenyag, line 466: It's possible to produce an empty record batch here. Will it cause a problem later in Execs like PartSort?

PR #8485, opened 2026-07-11
refactor: reconcile OTLP trace schemas request-wide

shuiyisong:

## Description

This pull request improves OTLP trace v1 ingestion by resolving schema changes across the complete trace request before inserting individual chunks.

### Background

Trace requests are split into chunks for ingestion, but each chunk can contain different flattened attribute columns and value types.

Previously, every chunk was converted and reconciled independently:

1. Convert one chunk using only its local schema.
2. Compare that schema with the current table.
3. Create or alter the table as needed.
4. Coerce and insert the chunk.
5. Repeat for subsequent chunks.

As a result, later chunks could discover additional columns or wider types and issue more DDL. Schema selection could also depend on chunk order.

### New workflow

The trace v1 path now:

1. Converts every chunk once while recording scalar types, sparse-column presence, and Binary/JSON logical types during the same row-building pass.
2. Aggregates those observations into one request-global schema without padding chunks to the global column set.
3. Resolves the desired schema against the existing table, keeping compatible existing types authoritative and widening `Int64` columns to `Float64` when required.
4. Prepares coercions before DDL so invalid values cannot cause schema changes based on data that cannot be written.
5. Excludes only the failing column and batch observation while allowing unrelated safe columns to participate in global planning.
6. Creates the table or adds all compatible missing columns through one request-level schema operation, followed by a separate type-modification operation when necessary.
7. Re-reads the table schema after DDL to handle concurrent table creation or schema changes and recomputes prepared coercions when the resolved target changes.
8. Inserts the original unpadded chunks, using sparse per-span fallback when a deterministic failure requires isolation.

The fallback path preserves each row’s original scalar or Binary/JSON logical schema and updates auxiliary service and operation tables only for spans whose main-table writes succeeded.

This change is limited to OTLP trace v1. Trace v0 behavior, the OTLP wire format, and external configuration remain unchanged. In race or incompatible-data paths, ingestion can still fall back to the existing chunk or per-span reconciliation behavior.

## PR Checklist


fengys1996:
@codex review

src/frontend/src/instance/otlp/trace_ingest.rs
```diff
@@ -0,0 +1,1793 @@
...
+                    let (request, batch_schema) =
+                        otlp::trace::v1::v1_to_grpc_main_insert_requests_with_schema(
+                            chunk,
+                            ingest_ctx.table_name,
+                        )?;
+                    let span_fallback_only = batch_schema.has_incompatible_logical_types();
+                    if !span_fallback_only {
+                        request_schema.observe_batch_schema(batch_index, &request, &batch_schema);
+                    }
+                    chunk_inserts.push(TraceChunkInsert {
```
fengjiachun: This retains every converted, padded chunk until the whole request has been planned, so `trace_ingest_chunk_size` no longer bounds v1 frontend memory. `MultiTableData::into_row_insert_requests` pads each row to the chunk-wide union of dynamic columns; with disjoint keys, a 128-span chunk can therefore expand the cell count by roughly 128x, and `chunk_inserts` keeps that expansion for every chunk. The HTTP limit is 64 MiB by default and can be disabled, while the request memory limiter only charges the body length, so a valid request can allocate multiple GiB and OOM the frontend. Please retain only lightweight observations until planning is complete and materialize/write chunks afterward, or add an explicit request-wide memory bound.

src/frontend/src/instance/otlp/trace_ingest.rs
```diff
@@ -0,0 +1,1793 @@
...
+
+            let target_type = if needs_reconcile {
+                let decision = match choose_trace_reconcile_decision(
+                    &column.schema.column_name,
+                    &observed_types,
+                    existing_type,
+                ) {
+                    Ok(decision) => decision,
+                    Err(_) => {
+                        return TraceRequestSchemaPlan::IncompatibleObservations(HashMap::from([
```
fengjiachun, line 549: A new column observed as `String`, `Int64`, and `Float64` reaches this branch because the current resolver only accepts up to two distinct types, even though all three are coercible to `Float64`. Excluding every batch then restores order-dependent per-chunk creation: String-first persists a String column, while Float64-first persists Float64. That is the schema instability this request-wide planning is meant to remove. Please resolve compatible multi-type sets globally and cover this three-chunk case.

src/frontend/src/instance/otlp/trace_ingest.rs
```diff
@@ -0,0 +1,1793 @@
...
+            .values()
+            .flatten()
+            .copied()
+            .collect::<HashSet<_>>();
+        for batch_index in &batch_indexes {
+            if let Some(chunk_insert) = chunk_inserts.get_mut(*batch_index) {
+                chunk_insert.schema_prepared = false;
+            }
+        }
+        request_schema.remove_batches(&batch_indexes);
```
fengjiachun: `exclusions` is keyed by column, but flattening it to batch indexes makes one bad column erase every observation from that chunk. For a new table, a coercion failure in one attribute can also remove an otherwise valid `service_name` and unrelated attributes from the create plan. If no other chunk has `service_name`, fallback can only add it later, after the trace create path's skipping-index setup has already been missed permanently. Please remove only the `(column, batch)` observations represented by `exclusions` (and skip rewrite preparation for chunks already marked for fallback), rather than dropping the whole batch globally.

fengjiachun (commented):
I found three issues in the request-wide trace schema planning path.

fengys1996 (approved):
Huge PR, almost LGTM.

src/frontend/src/instance/otlp/trace_ingest.rs
shuiyisong: Fixed in 604e111. Exclusions now remain keyed by column and batch; remove_observations deletes only those exact observations while preserving unrelated columns from the same chunk, including service_name. Chunks marked for per-chunk reconciliation are skipped by request-wide rewrite prevalidation. Regression coverage verifies safe columns remain in the create plan.

src/frontend/src/instance/otlp/trace_ingest.rs
shuiyisong: Fixed in 604e111. Each v1 chunk is now converted once into unpadded table rows and immediately compacted to retain only fixed fields and present dynamic cells. Request-wide planning keeps this sparse representation, and dense rows are reconstructed one chunk at a time for validation or write. This removes request-wide padded retention and repeated span and JSON conversion. The compact-to-dense path has an equality regression test.

src/frontend/src/instance/otlp/trace_ingest.rs
shuiyisong: Fixed in 604e111. New-column resolution now evaluates all observed scalar types against supported target types, so String plus Int64 plus Float64 resolves globally to Float64 regardless of chunk order. Existing rewrite prevalidation still rejects non-parseable String values safely. Tests cover the three-type set in both encounter orders.

src/frontend/src/instance/otlp/trace_ingest.rs
```diff
@@ -0,0 +1,1826 @@
...
+        let mut schemas = Vec::<(ColumnDataType, ConcreteDataType, bool)>::new();
+        for observation in self.batches.values() {
+            if schemas.iter().any(|(observed_datatype, observed_type, _)| {
+                *observed_datatype == observation.schema_type
+                    && observed_type == &observation.concrete_type
+            }) {
+                continue;
+            }
+
+            let mut conflicts = false;
```
sunng87, line 774: Minor: an observation is marked conflicting if it conflicts with *any* other observation, so the minimal-exclusion goal isn't quite met here. Concretely, a **new** column observed as `String` (batch 0), `Int64` (batch 1), and `Binary` (batch 2) marks all three as conflicting — `String`/`Int64` each conflict with `Binary` via `trace_logical_types_incompatible` — so all three batches get excluded, even though `String`+`Int64` are mutually reconcilable and only the `Binary` batch needs to fall back.

Net effect: a column with a mixed scalar/binary logical type across chunks falls back to order-dependent per-chunk creation for *all* of its chunks, rather than just the incompatible one — the exact instability this PR aims to remove. It's a narrow edge case (same attribute key seen as a scalar in some spans and bytes in others), but worth narrowing to exclude the minimal set of batches whose removal leaves a mutually-compatible remainder.

sunng87 (commented):
Minor / edge-case notes from review (over-exclusion, double rewrite, shadowing).

src/frontend/src/instance/otlp/trace_ingest.rs
```diff
@@ -0,0 +1,1826 @@
...
+                ChunkFailureReaction::RetryPerSpan.as_metric_label(),
+                "Chunk fallback triggered by incompatible binary and JSON values".to_string(),
+            );
+            return self
+                .ingest_trace_v1_rows_span_by_span(retry, ctx, ingest_state)
+                .await;
+        }
+
+        let mut request = retry.to_request()?;
+        let _ = Self::prepare_trace_v1_chunk_rewrite(
```
sunng87, line 1117: Minor perf: the rewrite for a prepared chunk is computed twice — once during `prevalidate_trace_v1_chunk_rewrites` (to gate DDL) and again here at write time (to actually apply it). The prepared result from prevalidation isn't carried forward, so all cell coercions are recomputed for every prepared chunk. Behavior is correct as-is; just flagging the duplicated coercion work, which scales with chunk size × number of rewritten columns.

src/frontend/src/instance/otlp/trace_ingest.rs
```diff
@@ -0,0 +1,1826 @@
...
+                        return TraceRequestSchemaPlan::IncompatibleObservations(HashMap::from([
+                            (
+                                column.schema.column_name.clone(),
+                                column.batches.keys().copied().collect(),
+                            ),
+                        ]));
+                    }
+                };
+                decision
+                    .map(|decision| {
```
sunng87, line 558: Nit: the closure parameter `decision` shadows the outer `decision` bound on the match above. Renaming the closure param (e.g. `d`) avoids the shadowing and is a touch clearer. (Gemini flagged this earlier; still present.)

shuiyisong: Thanks. I am leaving this unchanged: the closure parameter is the unwrapped value of the immediately preceding Option<TraceReconcileDecision>, its scope is only the map body, and the current name makes that relationship explicit. There is no behavioral or project lint/style issue; renaming it to d would be a style-only change.

src/frontend/src/instance/otlp/trace_ingest.rs
shuiyisong: Fixed in 67d1cb0. For a new column, schema conflict detection now builds the distinct scalar reconciliation candidates first and excludes only observations incompatible with that set, so String + Int64 remains request-wide and only the Binary batch falls back. Existing-table handling and binary-vs-JSON fallback are unchanged. Regression coverage exercises both encounter orders and verifies the remaining scalar plan resolves stably; all 82 frontend library tests and clippy pass.

src/frontend/src/instance/otlp/trace_ingest.rs
shuiyisong: Thanks. The duplicate work is intentional in the current design, so I am leaving it unchanged. Prevalidation runs only on the DDL path and must prove coercions before schema mutation; after DDL the table is read and planned again, so a concurrent create or alter can change the target type. Carrying prepared values across that boundary would either retain request-wide rewrite state, undoing the sparse-memory design, or risk applying stale targets. We can revisit this with profiling and a plan-versioned sparse representation if the cost becomes measurable.

PR #8486, opened 2026-07-12
fix: prevent credential leaks in sanitize_connection_string

raphaelroshan:

`sanitize_connection_string` is the single redaction point for metasrv `store_addrs` before they're logged (via `sanitize_store_addrs` → `Debug`). It missed three realistic input forms and leaked credentials:

1. **Password as a URL query parameter** — `postgresql://user@host:5432/db?password=secret`: splitting on the userinfo `@` kept the `?password=secret` query.
2. **Password containing `@`** — `mysql://user:p@ss@host:3306/db`: `find('@')` matched the first `@` inside the password, leaking the tail.
3. **Quoted keyword password with spaces** — `password='my secret'`: `split_whitespace` dropped `password='my` but kept `secret'`.

This reworks the function to (a) strip URL userinfo at the *last* `@` in the authority (before the path), so an `@` in the password no longer leaks, and (b) redact every `password=` value — query or keyword, including quoted/space-containing — to `***`. The existing keyword case now shows `password=***` instead of dropping the token.

## Checklist

By submitting this pull request, I confirm that my contribution is made under the terms of the GreptimeDB CLA.


killme2008:
Good catch!

@raphaelroshan Could we use the `url` crate to parse the URL and redact the password, etc.?

https://crates.io/crates/url

v0y4g3r:
As per the libpq doc, all the following formats are legitimate connection parameters:

- `password = secret`
- `password='password with spaces'`
- `password='pa\\'ss'`

Given these edge cases, I'd also suggest using crates like `url` and `tokio-postgres` for credential redaction.

raphaelroshan:
Thanks for the feedback, updated!
 Now uses the url crate to parse URL-form addresses (mysql://, postgresql://, http:// etcd) and redact the userinfo password plus any `password` query parameter.

For the libpq keyword DSN form (host=… password=…) I kept a small redactor rather than tokio-postgres: this function runs unconditionally but tokio-postgres is an optional/feature-gated dep here, and store_addrs also covers mysql/etcd, which it can't parse. The redactor now handles `password = x` (spaces around =), single/double-quoted values with spaces, and \-escaped quotes, each with a test. Happy to adjust the split if you'd prefer.

raphaelroshan:
accidentally closed this from a depth 1 clone and force push 😓 , let me try to reopen this

PR #8490, opened 2026-07-13
fix(flow): recover stale incremental reads

discord9:

Follow-up to #8409.

Flow batching incremental reads can fall behind the source retention boundary under sustained write pressure. The storage layer then returns `STALE_CURSOR` / `FALLBACK_FULL_RECOMPUTE`, but those markers may be wrapped by frontend/client errors. The previous recovery path could treat the failure as generic, perform only dirty-window-scoped repair, advance checkpoints, and leave source-visible rows missing from the sink. `ADMIN FLUSH_FLOW` could also sample a stale zero window cap before waiting for an in-flight repair, or return the wrapped stale error instead of retrying the repaired state.

This PR makes stale recovery correctness-first:

- classify stale cursor and stale snapshot-fence markers through the full error source chain;
- force a one-shot unfiltered full recompute after a stale incremental cursor, clearing the guard only after a complete watermark proof;
- compute the explicit-flush dirty-window cap while holding the task execution lock;
- use a conservative flush cap that drains separated/sub-window ranges;
- when pending fenced repair and post-fence live dirty coverage coexist, process them in two bounded serialized rounds without merging post-`H` writes into the old fence;
- retry `ADMIN FLUSH_FLOW` up to three times for structured or wrapped stale errors.

The change does not alter public APIs or persisted metadata. A forced full recompute can be more expensive than scoped repair, but only occurs after storage explicitly reports that the incremental cursor is no longer readable.

Validation:

- `cargo fmt -p flow -- --check`
- `cargo test -p flow` (230 passed)
- `cargo clippy -p flow --all-targets -- -D warnings`
- live 100k rows/s, 5-minute sinkstress validation: 29,928,000 rows written with zero client errors; wrapped stale recovery retried inside `ADMIN FLUSH_FLOW`; source and sink converged to 154,432,508 rows with diff 0.

## PR Checklist

### DIFF
--- a/src/flow/src/batching_mode.rs
+++ b/src/flow/src/batching_mode.rs
@@ -22,6 +22,7 @@ use session::ReadPreference;
 
 mod checkpoint;
 pub(crate) mod engine;
+mod error_chain;
 mod eval_schedule;
 pub(crate) mod frontend_client;
 mod state;
--- a/src/flow/src/batching_mode/engine.rs
+++ b/src/flow/src/batching_mode/engine.rs
@@ -20,7 +20,8 @@ use std::time::Duration;
 
 use api::v1::flow::DirtyWindowRequests;
 use catalog::CatalogManagerRef;
-use common_error::ext::BoxedError;
+use common_error::ext::{BoxedError, ErrorExt};
+use common_error::status_code::StatusCode;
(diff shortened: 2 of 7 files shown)


discord9:

PR #8491, opened 2026-07-13
feat: support splunk HEC raw endpoint

agrawalx:

Part of #8216. Follow-up to #8321.

Adds the **Splunk HEC raw-text endpoint**, completing the second ingestion path of the HEC protocol:

- `POST /v1/splunk/services/collector/raw` (+ `/raw/1.0` alias)

**How it works**

- Body is plain text, one event per line (`\n` / `\r\n`, blank lines skipped). Each line is stored **as it is** in a `message`
  column; no content parsing. Structure, if wanted, comes from a custom pipeline reading `message`.
- Metadata is request-level, via query params: `host`/`source`/`sourcetype` -> tag columns on every event, `index` -> table name (sanitized; fallsback to `?table=`, then `splunk_logs`), `time` -> epoch default timestamp for all
  events (else ingest time).
- `channel` (query param or `x-splunk-request-channel` header) is accepted and **ignored**, indexer acknowledgment is a separate follow-up.
- Auth, gzip, HEC response codes (0/5/6/7/8), tag promotion, and metadata-first primary-key ordering are all shared with `/event`.
- The pipeline-resolution part of `/event` is factored into a shared helper; both endpoints now also honor `?version=` to pin a custom pipeline version (previously ignored on `/event`).

**Testing**

- Unit tests: line splitting, metadata mapping, 
- Integration test: raw ingest end-to-end (values, tags as primary key, verbatim
  `message`), `?time=`, gzip on the `/raw/1.0` alias, channel ignored, error codes 5/6/7, and a **replay** of a captured Vector request (percent-encoded query string decodes correctly).
- Capture note: Vector's default raw batching concatenates events with no separator; real Splunk's default line breaking would also read that as one event. Operators wanting one event per line should set `batch.max_events: 1` (or use `/event`).

**Out of scope (follow-ups per #8216):** indexer acknowledgment (`/ack`, channels), per-event timestamp extraction from line content(workarounds: `?time=` or a pipeline `date` processor).

**Compatibility:** purely additive, a new route on the existing `/v1/splunk` base; no changes to existing APIs, schemas, or data.

## PR Checklist

### DIFF
--- a/src/servers/src/http/splunk.rs
+++ b/src/servers/src/http/splunk.rs
@@ -441,6 +441,9 @@ pub async fn handle_raw(
     query_ctx.set_channel(Channel::Splunk);
 
     let body = String::from_utf8_lossy(&payload);
+    if matches!(body, std::borrow::Cow::Owned(_)) {
+        debug!("splunk raw body contains invalid UTF-8; bytes replaced with U+FFFD");
+    }
     let lines = parse_raw_lines(&body);
     if lines.is_empty() {
         // HEC code 5 == "No data".
@@ -532,7 +535,10 @@ async fn resolve_pipeline_and_ingest(
     let custom_time_index = Some((format!("{};epoch;ns", greptime_timestamp()), false));
     let pipeline = match PipelineDefinition::from_name(&pipeline_name, version, custom_time_index) {
         Ok(pipeline) => pipeline,
-        Err(_) => return hec_response(StatusCode::INTERNAL_SERVER_ERROR, 8, "pipeline error"),
+        Err(e) => {
+            error!(e; "failed to resolve splunk pipeline definition: {pipeline_name}");
+            return hec_response(StatusCode::INTERNAL_SERVER_ERROR, 8, "pipeline error");
+        }
     };
     let pipeline_params =
         GreptimePipelineParams::from_map(extract_pipeline_params_map_from_headers(headers));
@@ -681,6 +687,24 @@ mod tests {
             panic!("expected object");
         };
         assert_eq!(m.len(), 2);
+    }
+
+    #[test]
+    fn parses_real_raw_client_payloads() {
+        // Shapes captured from Vector's `splunk_hec_logs` sink with
+        // `endpoint_target = "raw"`
+
+        // --- batch.max_events = 1: one event per request, no trailing newline. ---
+        let single = "190.79.85.36 - b0rnc0nfused [13/Jul/2026:05:10:25 +0000] \"GET /money HTTP/2.0\" 300 29099";
+        assert_eq!(parse_raw_lines(single), vec![single]);
+
+        // --- Vector's default batching concatenates events with NO separator.
+        //     so the lines form just one event here.
+        let concatenated = concat!(
+            "214.106.15.209 - AnthraX [13/Jul/2026:05:08:12 +0000] \"GET /secret-info HTTP/1.0\" 401 47103",
+            "245.115.225.188 - KarimMove [13/Jul/2026:05:08:12 +0000] \"PUT /apps/deploy HTTP/2.0\" 302 18226",
+        );
+        assert_eq!(parse_raw_lines(concatenated).len(), 1);
     }
 
     // ---- parse_hec_time ----
--- a/src/servers/src/http/splunk.rs
+++ b/src/servers/src/http/splunk.rs
@@ -33,6 +33,7 @@ use common_error::ext::ErrorExt;
 use common_query::prelude::greptime_timestamp;
 use common_telemetry::{debug, error};
 use operator::insert::SPLUNK_PK_METADATA_ORDER_KEY;
+use pipeline::util::to_pipeline_version;
 use pipeline::{
     ContextReq, GREPTIME_INTERNAL_IDENTITY_PIPELINE_NAME, GreptimePipelineParams, PipelineContext,
     PipelineDefinition,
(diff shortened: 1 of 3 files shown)


src/servers/src/http/splunk.rs
```diff
@@ -59,6 +60,70 @@ const DEFAULT_SPLUNK_TABLE: &str = "splunk_logs";
...
+        VrlValue::Bytes(Bytes::from(line.to_string())),
+    );
+    for (key, value) in metadata {
+        map.insert(
+            KeyString::from(*key),
+            VrlValue::Bytes(Bytes::from(value.to_string())),
+        );
+    }
+    VrlValue::Object(map)
+}
```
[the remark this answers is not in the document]
agrawalx: applied

src/servers/src/http/splunk.rs
```diff
@@ -59,6 +60,72 @@ const DEFAULT_SPLUNK_TABLE: &str = "splunk_logs";
...
+    pub time: Option<String>,
+    pub table: Option<String>,
+    pub pipeline_name: Option<String>,
+    pub version: Option<String>,
+}
+
+/// Splits a raw HEC body into events: one event per line (`\n`, tolerating `\r\n`),
+/// skipping blank/whitespace-only lines.
+fn parse_raw_lines(body: &str) -> Vec<&str> {
+    body.lines()
```
fengjiachun: `body.lines()` irreversibly splits valid multiline HEC events such as stack traces. Splunk `/raw` applies timestamp/sourcetype-specific event-breaking rules and permits one event to span multiple lines within a request; after this split, `run_pipeline` processes each line separately, so a custom pipeline cannot reconstruct it. Please preserve the body as one message or make event breaking explicit/configurable, and cover a multiline event.

src/servers/src/http/splunk.rs
```diff
@@ -349,14 +416,116 @@ pub async fn handle_event(
...
+pub async fn handle_raw(
+    State(log_state): State<LogState>,
+    Query(params): Query<SplunkRawQueryParams>,
+    Extension(mut query_ctx): Extension<QueryContext>,
+    headers: HeaderMap,
+    payload: Bytes,
+) -> impl IntoResponse {
+    query_ctx.set_channel(Channel::Splunk);
+
+    let body = String::from_utf8_lossy(&payload);
```
fengjiachun: The enclosing Splunk router has `pass_through_unaccepted(true)`, so an unsupported or mistyped `Content-Encoding` reaches this line still compressed. `/event` rejects such bytes as invalid JSON, but `/raw` lossy-decodes them and can return code 0 after storing corrupted text. Please reject remaining non-identity encodings (or disable pass-through for this route) and add a request-level regression test.

src/servers/src/http/splunk.rs
killme2008: Agree.

The safest default is to preserve the entire request body as one message. Newline framing can be provided as an explicit pipeline or decoder option. FYI. Please add a multiline regression test

src/servers/src/http/splunk.rs
```diff
@@ -349,14 +416,116 @@ pub async fn handle_event(
...
+    // Table routing: `?index=` (sanitized) -> `?table=` -> default.
+    let table = params
+        .index
+        .as_deref()
+        .and_then(sanitize_index)
+        .or_else(|| params.table.clone())
+        .unwrap_or_else(|| DEFAULT_SPLUNK_TABLE.to_string());
+    // Bad table name (e.g. invalid `?table=`) -> HEC code 7 ("incorrect index").
+    if !NAME_PATTERN_REG.is_match(&table) {
+        let msg = format!("incorrect index: {table}");
```
killme2008:
```suggestion
        let msg = format!("Invalid index name: {table}");
```

src/servers/src/http/splunk.rs
killme2008: Assuming UTF-8 is reasonable because this endpoint ingests log text. The issue is specifically String::from_utf8_lossy: it silently changes invalid input and may still return HEC code 0. Please validate the body with ·std::str::from_utf8· and return HEC code 6 for invalid UTF-8. After validation, the original Bytes can still be passed to the pipeline without copying or modifying it.

```rust
  if std::str::from_utf8(&payload).is_err() {
      return hec_response(
          StatusCode::BAD_REQUEST,
          6,
          "Invalid data format",
      );
  }
```

killme2008 (commented):
Great job!

However, there are a few issues that need attention, especially `body.lines()`.

src/servers/src/http/splunk.rs
agrawalx: Applied this default, I have also added a optional `linebreaker` query param that takes percent-encoded values and splits based on those. added a multiline regression test `multiline_raw_body_is_one_event` and a section in integration-test which also sends a multiline request.

src/servers/src/http/splunk.rs
agrawalx: Fixed this by removing `from_utf8_lossy` and adding a check before running `from_utf8`. Tested it by sending a request with `snappy` content-encoding error which fails with error code 6.

src/servers/src/http/splunk.rs
agrawalx: applied this.

src/servers/src/http/splunk.rs
agrawalx: applied this as well.

agrawalx:
@killme2008  @fengjiachun thanks for taking the time to review this! I've pushed changes to address your comments. Please let me know what you think!

killme2008 (approved):
LGTM

PR #8504, opened 2026-07-13
fix(promql): handle missing labels in or matching

discord9:

Related to #8502, but this is an independent PromQL matching fix.

Fix PromQL `or` matching when an `on(...)` label is absent from one or both operands.

Prometheus treats an absent label and an empty label value as the same matching value. The previous planner passed requested label names directly to `UnionDistinctOn`, so physical planning failed with `Column ... not found` when a label was absent.

This PR:

- materializes hidden, normalized comparison columns for OR matching while preserving the original visible labels and left-side precedence;
- keeps comparison identity positional through `UnionDistinctOn`, including Substrait round trips where intermediate aliases are not preserved;
- merges child nullability into the declared OR output schema so unmatched RHS rows with absent labels can be emitted safely;
- aligns mismatched supported numeric value fields to `Float64`, consistent with PromQL's sample model;
- preserves the exact visible output shape and left-owned planner context without leaking internal columns;
- adds real-batch coverage for absent, NULL, empty, and nonempty labels, matching modifiers, collisions, empty operands, nested expressions, malformed indices, and repeated Substrait round trips.

The existing protobuf remains unchanged and already encodes `UnionDistinctOn` columns by index. This PR does not change persisted data, public APIs, or wire structure. Metric-name (`__name__`) provenance and hash-collision hardening are outside this PR's scope.

Validation:

- `cargo test -p query -- --skip merge_scan_reports_query_load_with_metrics_region_id` — 506 passed, 7 ignored; the skipped pre-existing global-metrics test passes in isolation
- `cargo test -p promql` — 148 passed
- `cargo sqlness bare -t '(set_operation|tsid_binary_join_regression)' --preserve-state` — standalone and distributed passed
- `make fmt`
- `make clippy`
- `make check-udeps` — not available locally (`cargo-udeps` is not installed)

## PR Checklist


waynexia:
@codex review

src/query/src/promql/planner.rs
```diff
@@ -4537,24 +4604,168 @@ impl PromPlanner {
...
+                let expr = if let Some((qualifier, _)) = field {
+                    DfExpr::ScalarFunction(ScalarFunction {
+                        func: coalesce(),
+                        args: vec![
+                            DfExpr::Column(Column::new(qualifier, label.clone())),
+                            DfExpr::Literal(empty.clone(), None),
+                        ],
+                    })
+                } else {
+                    DfExpr::Literal(empty.clone(), None)
```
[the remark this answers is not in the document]
discord9: Fixed in 2fc747ebe1. UnionDistinctOn now stores every RHS row hash separately and only uses the LHS signatures to suppress matching RHS rows, so an empty LHS preserves all RHS series even when they share the same normalized key. Added regression coverage for both empty-LHS preservation and LHS masking of all matching RHS rows. `cargo test -p promql`, the query tests (with the known global-metrics flaky case verified separately), and `make clippy` pass.

fengjiachun (approved):
LGTM

src/query/src/promql/planner.rs
```diff
@@ -4537,24 +4604,168 @@ impl PromPlanner {
...
+                };
+                expr.alias(internal_name.clone())
+            };
+            left_match_exprs.push(normalize(left_field));
+            right_match_exprs.push(normalize(right_field));
+        }
+
+        let left_augmented = LogicalPlanBuilder::from(left_projected)
+            .project(visible_left_exprs.into_iter().chain(left_match_exprs))
```
fengjiachun, line 4712: Not blocking, but could you share a focused before/after benchmark for this path? It now materializes one normalized string column per match label on both inputs, including ordinary `or` queries where the labels are already present. Together with `UnionDistinctOn` retaining every RHS row and its hash, high-cardinality inputs or low-cardinality matching such as `on()` may increase CPU and peak memory noticeably. A benchmark covering the default matching case and an `or on()` case with a large RHS would help quantify the cost.

discord9: Thanks for calling this out. The first focused run confirmed the concern: retaining/interleaving the full RHS made the 1M-row low-cardinality `on()` case 10.6x slower with 2.22x peak RSS. I replaced that implementation in `a8e86abebd` with LHS-first streaming: emit/hash LHS first, then filter and emit RHS one batch at a time. Memory is now `O(unique LHS signatures + one RHS batch)` rather than retaining all RHS rows/hashes.

**Build/method**

- Base: `56e9158819`; final head: `a8e86abebd`.
- The measured production candidate was parent `2fc747ebe1` plus production diff SHA-256 `84d39c0f2975f336afd17896bb498257c2e395d4892f3e883717ccaab637dd4b`; the subsequent changes before `a8e86abebd` were test-only cleanup.
- Base/head binary SHA-256: `4cffabcb0a43894abadb393e14b5efaf5057af2d97abbc60b860b0a37f333f83` / `e03a0ab1c1c8f46976a5ce5fc4f1d1180d5ea497b5da255a963ba349e32049b5`.
- `cargo build --locked --release -p promql --example pr8504_or_bench`; release profile (`debug=1`, unstripped), default features, jemalloc, Rust 1.96.0-nightly.
- One physical core without an SMT sibling, batch size 8192. Timing uses 30 samples/revision in 15 ABBA blocks; peak RSS uses five fresh processes/cell plus input-only baselines.
- `perf` was unavailable (`perf_event_paranoid=4`), so these are elapsed-time results, not cycle/instruction claims. DataFusion pool metrics do not cover the operator-owned `Vec`/`HashSet`/Arrow allocations, so peak RSS is from GNU time.

**Equal-output results**

| Case | Base median / p95 | Head median / p95 | ABBA head/base ratio (95% bootstrap CI) | Base -> head peak RSS |
|---|---:|---:|---:|---:|
| Default `or`: 500k LHS, 1M RHS, two present nullable UTF-8 labels, 500k overlap, 1M output | 243.2 / 292.1 ms | 67.2 / 80.1 ms | 0.280 (0.262-0.292) | 197.8 -> 109.8 MiB |
| `or on()` masked: 1 LHS, 100k same-timestamp RHS, 1 output | 1.02 / 1.18 ms | 0.87 / 0.99 ms | 0.851 (0.792-0.895) | 15.6 -> 15.2 MiB |
| `or on()` masked: 1 LHS, 1M same-timestamp RHS, 1 output | 7.17 / 7.77 ms | 6.12 / 6.35 ms | 0.846 (0.834-0.865) | 60.6 -> 60.0 MiB |

For the masked case, head execution-minus-input RSS is about 2.5 MiB at both 100k and 1M RHS, so it no longer scales with retained RHS data.

**Corrected-behavior case**

With empty LHS and 1M same-timestamp RHS rows, base emits one incorrect row; head correctly emits all 1M. Head takes 0.32 ms in the count-only timing sink and peaks at 60.0 MiB. I am not presenting a before/after ratio for this unequal-output case. Full-checksum verification separately confirms all 1M rows and checksum `bec684638bd857cc`; the low elapsed time comes from zero-copy batch streaming plus a count-only sink.

**Normalized-column decomposition**

A four-cell base/head x direct/normalized-plan run shows that, on the streaming head, the hidden `coalesce(label, "")` projections change median elapsed time from 64.6 to 67.1 ms (+3.9%; block ratio 1.029, 95% CI 0.996-1.053) and peak RSS from 103.9 to 109.8 MiB (+5.7%). The projections still have a measurable cost, but removing full-RHS retention eliminated the prior multiplier and leaves both below the predeclared 5% runtime / 10% RSS review thresholds.

All focused correctness tests, full PromQL tests, query regressions, clippy, and CI on `a8e86abebd` pass.

discord9:
Post-merge follow-up: I compared the final streaming implementation with Prometheus's current [`VectorOr`](https://github.com/prometheus/prometheus/blob/91e9d1c1a38d714e9f5c0b129b264003a5abc3e7/promql/engine.go#L3111-L3136), rather than interpreting `UnionDistinctOn` as SQL `UNION DISTINCT` / PostgreSQL `DISTINCT ON`.

## Core semantics match

For one evaluation timestamp, the reference behavior is:

```text
output = every LHS sample
       + every RHS sample whose matching signature is absent from the LHS
```

The final implementation has the same behavior:

- LHS rows are all emitted; they are **not** deduplicated by matching signature.
- RHS rows are also **not** deduplicated among themselves. If a signature is absent from the LHS, every RHS row with that signature is retained.
- If a signature occurs on the LHS, all matching RHS rows are suppressed.
- Empty-LHS and empty-RHS behavior matches the reference implementation.
- Missing labels, `NULL`, and `label=""` are normalized to the same hidden comparison value while preserving the visible output representation. This matches Prometheus label-matching semantics.

This also explains the two hash fields in the stream:

- `lhs_signatures: HashSet<u64>` is persistent cross-batch membership state. It deduplicates only the stored LHS **signatures**, because RHS filtering only needs an existence test; it does not deduplicate LHS output rows.
- `hashes: Vec<u64>` is a row-aligned scratch buffer for the current batch and is cleared/reused for each batch. A clearer name would be `batch_hashes` or `hash_buffer`.

Prometheus runs `VectorOr` once per evaluation step, so timestamp is not part of its label signature. GreptimeDB batches multiple evaluation steps together, so adding the time-index column to the key is equivalent as long as that column is the evaluation timestamp, which is the planner invariant here.

## Remaining differences / risks

These do not invalidate the missing-label fix, but they are worth recording as follow-ups:

1. **Collision-safe equality**

   Prometheus constructs a canonical, lossless label signature and ultimately compares the complete bytes. GreptimeDB stores only the `u64` produced by `create_hashes` and does not perform a key-equality check after a hash match. A 64-bit collision could therefore incorrectly suppress an unrelated RHS row. The probability is very small, but this is a real correctness difference.

2. **Duplicate complete labelsets**

   Prometheus separately rejects a result vector containing the same complete labelset more than once (`vector cannot contain metrics with the same labelset`; see [`rangeEval`](https://github.com/prometheus/prometheus/blob/91e9d1c1a38d714e9f5c0b129b264003a5abc3e7/promql/engine.go#L1537-L1568)). `UnionDistinctOn` itself has no equivalent validation. Equivalence therefore relies on an upstream uniqueness invariant; we should confirm/document that invariant or validate it at the appropriate evaluator layer.

3. **Explicit `on(__name__)`**

   Prometheus can include `__name__` in an `on(...)` signature. GreptimeDB derives comparison columns from tag columns, while metric/table name is not a tag column; an explicit `on(__name__)` therefore appears to be treated as a missing/empty label and may diverge. We should either support this case or reject/document it explicitly.

4. **Execution/data model differences**

   Prometheus evaluates materialized instant vectors. GreptimeDB aligns two Arrow schemas, widens nullability, and streams the LHS before lazily starting the RHS. These are implementation/resource-model differences, not normal-result differences.

So the merged implementation is compatible with the reference `or` behavior for ordinary valid `on`/`ignoring` queries. The main correctness follow-ups are collision-safe matching, the complete-labelset uniqueness invariant, and `on(__name__)` behavior.

PR #8508, opened 2026-07-14
fix: preserve nulls in timestamp arrays

discord9:

N/A.

PostgreSQL `TIMESTAMP_ARRAY` parameters targeting nanosecond timestamps used `filter_map`, which silently dropped SQL `NULL` elements and non-null timestamps outside the nanosecond range. Dropping elements changed the array length and shifted later values to the wrong positions.

This PR replaces the lossy conversion with a fallible, position-preserving map:

- SQL `NULL` elements become typed null nanosecond timestamp scalars at the same positions.
- In-range non-null timestamps retain the existing conversion behavior.
- Out-of-range non-null timestamps return the existing PostgreSQL `22023` / `numeric_value_out_of_range` error instead of being silently removed.

Focused regression tests cover an interior null between two valid values and the controlled out-of-range error.

This change does not alter the wire format, schema format, public API, or other timestamp units.

## PR Checklist

### DIFF
--- a/src/servers/src/postgres/types.rs
+++ b/src/servers/src/postgres/types.rs
@@ -1055,17 +1055,22 @@ pub(super) fn parameters_to_scalar_values(
                                 TimestampType::Nanosecond(_) => {
                                     let values = data
                                         .into_iter()
-                                        .filter_map(|ts| {
-                                            ts.and_then(|ts| {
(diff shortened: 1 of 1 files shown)


killme2008 (approved):
Good catch!

PR #8514, opened 2026-07-14
fix(log-store): deduplicate Kafka WAL multipart records

### LINKED ISSUE #8506: Kafka WAL duplicate multipart record breaks region recovery
### What type of bug is this?

Data corruption, Unexpected error

Datanode

### Minimal reproduce step

Replay a Kafka WAL stream containing a duplicated multipart part:
First → ... → Middle(9) → Middle(9) → ...
Restart a DataNode or trigger region catchup.

Recovery should handle an exact duplicate WAL part safely, or report a recoverable error with Kafka offset details. A duplicated record should not leave the region unavailable.

### What did you see instead?

Kafka WAL replay fails:
Failed to read WAL, provider: Kafka(topic=prod_greptimedb_3_14)
Illegal sequence of a middle record,
last record: Middle(9), incoming record: Middle(9)

doesn't matter

main

```bash

```

WenyXu:

Fixes #8506.

This PR makes Kafka WAL multipart replay tolerate exact duplicate `Middle` and `Last` records.

A duplicate `Middle` is ignored only when it has the same sequence number and payload as the buffered part. A duplicate `Last` is ignored only when its payload matches the preceding `Last`. Other malformed or out-of-order multipart sequences continue to return `IllegalSequence`.

`Last` records are now buffered and emitted when the next `First` for the same region arrives, or when `remaining_entries()` drains the trailing buffer. This prevents a duplicate `Last` from creating an extra incomplete entry.

The Kafka record format and public APIs are unchanged. WAL data remains backward compatible.

## PR Checklist


v0y4g3r:
@codex review

src/log-store/src/kafka/util/record.rs
```diff
@@ -284,21 +293,22 @@ pub(crate) fn maybe_emit_entry(
...
+                    IllegalSequenceSnafu {
+                        error: format!(
+                            "Illegal sequence of a last record, last record: {:?}, incoming record: {:?}",
+                            last_record.meta.tp, record.meta.tp
+                        )
+                    }
+                );
+                if last_record.meta.tp == RecordType::Last {
+                    return Ok(None);
+                }
```
v0y4g3r: We no longer emits a record when reaches Last. That can break the order of log entries. Imagine the following entries:
```
offset 1: First(seq=1)
offset 2: Last(seq=1)
offset 3: Full(seq=2)
```
The replayed entries will be offset `3, 1, 2`

WenyXu: Good catch

WenyXu: [9764433](https://github.com/GreptimeTeam/greptimedb/pull/8514/commits/976443318f3ebb63a678c35c7b53b2a391666b50)

v0y4g3r (approved):

PR #8522, opened 2026-07-14
fix(query): avoid unsafe count wildcard rewrites

discord9:

Prevent the count-wildcard optimizer from rewriting `count(*)` to a column count when the source time-index identity is not proven at the aggregate's immediate input.

The rule now falls back to `count(1)` across projections, multi-input plans, missing fields, and nullable fields. This avoids wrong results when a projection or another schema-changing node exposes a nullable column whose name collides with the source time index.

Regression tests cover direct and aliased scans, projection rename/reorder and name collision, multi-input plans, and a nullable field produced by an inner aggregate.

This change does not alter public APIs, persisted schemas, or data formats.

## PR Checklist

### DIFF
--- a/src/query/src/optimizer/count_wildcard.rs
+++ b/src/query/src/optimizer/count_wildcard.rs
@@ -93,9 +93,7 @@ impl CountWildcardToTimeIndexRule {
             if plan.inputs().len() > 1 {
                 return None;
             }
-            let Some(input) = plan.inputs().first().copied() else {
-                return None;
-            };
+            let input = plan.inputs().first().copied()?;
             let Ok((_, field)) = input.schema().qualified_field_from_column(col) else {
                 return None;
             };
@@ -155,7 +153,8 @@ impl TreeNodeVisitor<'_> for TimeIndexFinder {
         }
 
         if let LogicalPlan::SubqueryAlias(subquery_alias) = node {
-            self.table_alias = Some(subquery_alias.alias.clone());
+            self.table_alias
+                .get_or_insert_with(|| subquery_alias.alias.clone());
         }
 
         if let LogicalPlan::TableScan(table_scan) = &node
@@ -354,6 +353,20 @@ mod test {
             .analyze(simple_alias, &config)
             .unwrap();
         assert_count_argument_column(&simple_alias, "projected", "ts");
+
+        let nested_alias = count_star(
+            LogicalPlanBuilder::from(qp_026_source_plan("source"))
+                .alias("inner")
+                .unwrap()
+                .alias("outer")
+                .unwrap()
+                .build()
+                .unwrap(),
+        );
+        let nested_alias = CountWildcardToTimeIndexRule
+            .analyze(nested_alias, &config)
+            .unwrap();
+        assert_count_argument_column(&nested_alias, "outer", "ts");
 
         let nested_rename = count_star(
             LogicalPlanBuilder::from(qp_026_source_plan("source"))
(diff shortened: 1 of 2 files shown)


src/query/src/optimizer/count_wildcard.rs
```diff
@@ -140,12 +141,19 @@ impl CountWildcardToTimeIndexRule {
...
     fn f_down(&mut self, node: &Self::Node) -> DataFusionResult<TreeNodeRecursion> {
+        // A projection may rename or replace the time index. We only know the
+        // source metadata, so do not rewrite through an unproven projection.
+        if matches!(node, LogicalPlan::Projection(_)) {
+            self.has_projection = true;
+        }
+
         if let LogicalPlan::SubqueryAlias(subquery_alias) = node {
             self.table_alias = Some(subquery_alias.alias.clone());
         }
```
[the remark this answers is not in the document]
discord9: Fixed in c477244343. The top-down visitor now preserves the first (outermost) SubqueryAlias with get_or_insert_with, and the shape-matrix regression covers source -> inner -> outer -> count(*) and verifies count(outer.ts). Also applied the Clippy ? suggestion. Focused QP-026 tests pass 3/3, query Clippy with -D warnings passes, fmt/diff checks pass, and the tql-cte SQLness case passes in both standalone and distributed modes.

src/query/src/optimizer/count_wildcard.rs
```diff
@@ -85,20 +85,19 @@ impl CountWildcardToTimeIndexRule {
...
                 return None;
             }
-            for input in plan.inputs() {
-                if input.schema().has_column(col) {
-                    is_valid = true;
-                    break;
-                }
-            }
-            if !is_valid {
+            let input = plan.inputs().first().copied()?;
```
killme2008, line 100: We used to check all `inputs`, but now only the first—is that expected?

src/query/src/optimizer/count_wildcard.rs
```diff
@@ -319,6 +333,216 @@ mod test {
         );
     }
 
+    #[test]
+    fn qp_026_count_wildcard_shape_matrix() {
```
killme2008: Don't add any `qa_` prefix in tests.

src/query/src/optimizer/count_wildcard.rs
```diff
@@ -140,14 +139,22 @@ impl CountWildcardToTimeIndexRule {
...
 }
 
 impl TreeNodeVisitor<'_> for TimeIndexFinder {
     type Node = LogicalPlan;
 
     fn f_down(&mut self, node: &Self::Node) -> DataFusionResult<TreeNodeRecursion> {
+        // A projection may rename or replace the time index. We only know the
+        // source metadata, so do not rewrite through an unproven projection.
+        if matches!(node, LogicalPlan::Projection(_)) {
+            self.has_projection = true;
```
killme2008: I don’t think we need to check `has_projection`.

The row count is correct regardless of whether a projection exists.

tests/cases/standalone/common/tql/tql-cte.result
```diff
@@ -528,7 +528,7 @@ HAVING count(*) > 1;
 | logical_plan  | MergeScan [is_placeholder=false, remote_input=[                                                                                                       |
 |               | Projection: date_trunc(Utf8("minute"),tql_grouped.ts) AS minute, count(Int64(1)) AS count(*) AS point_count                                           |
 |               |   Filter: count(Int64(1)) > Int64(1)                                                                                                                  |
-|               |     Aggregate: groupBy=[[date_trunc(Utf8("minute"), tql_grouped.ts)]], aggr=[[count(tql_grouped.ts) AS count(Int64(1))]]                              |
+|               |     Aggregate: groupBy=[[date_trunc(Utf8("minute"), tql_grouped.ts)]], aggr=[[count(Int64(1))]]                                                       |
```
killme2008: Looks like this change is not expected.

evenyag (approved):

PR #8526, opened 2026-07-14
feat(meta): add retention GC for soft-dropped tables

v0y4g3r:

Part of GreptimeTeam/greptimedb-enterprise#851. Follow-up to GreptimeTeam/greptimedb#8458.

This PR adds the metadata and scheduling foundations for retaining and eventually garbage-collecting soft-dropped tables. The user-facing feature remains disabled while the remaining WAL-pruning safety work is addressed.

- Persist the soft-drop timestamp and a fixed retention deadline alongside tombstoned table metadata. To maintain these lifecycle markers safely, extend the tombstone move/restore path with extra marker operations instead of issuing unrelated writes at the caller. This reuses the existing idempotent retry flow, ensures a retry still completes marker writes or deletions after the source metadata has already moved, and accounts for marker operations when chunking transactions under the backend `max_txn_ops` limit. Legacy tombstones without markers remain compatible.
- Preserve the selected drop mode and retention deadline in procedure state so retries and recovery do not change an in-flight operation. The runtime gate prevents new soft drops, while recovery preserves the persisted mode and uses configured retention only as a fallback for legacy procedure snapshots.
- Extend metasrv GC to scan expired tombstones and submit the existing purge-table procedure with bounded concurrency, in-flight deduplication, panic/cancellation-safe reservation release, and round-robin candidate ordering so repeatedly failing tables cannot starve later candidates.
- Revalidate automatic purge eligibility after acquiring `TableLock` and whenever the procedure recovers, preventing a stale scan from purging a freshly re-dropped table. Automatic purge uses a distinct procedure type so older binaries cannot interpret it as an unconditional manual purge.
- Reserve `gc.experimental_soft_drop.enable` and `gc.experimental_soft_drop.retention` while keeping user-triggered DDL soft drop and automatic purge disabled until the known Kafka WAL pruning visibility risk is fixed. The retention value is only carried into recovery as a compatibility fallback for legacy persisted soft-drop procedures. Ordinary GC is unaffected.
- Add focused common-meta and metasrv tests for marker lifecycle, procedure recovery, stale purge rejection, reservation cleanup, bounded scheduling, deduplication, and fairness. Regenerate the configuration reference from the example TOML.

Limitations:

- Soft drop is not user-accessible in this PR; `gc.experimental_soft_drop.*` cannot enable new soft drops or automatic purge.
- The Kafka WAL pruning safety issue must be fixed before enabling the runtime gate.
- Periodic tickers in no-election metasrv mode remain a pre-existing general limitation and are outside this PR.

Compatibility:

- Persisted procedure fields use serde defaults for older snapshots.
- Legacy tombstones without retention markers remain readable, restorable, and manually purgeable.
- The new tombstone markers and reserved configuration are additive; no existing API or schema is repurposed.

Verification:

- `cargo nextest run -p common-meta drop_table`
- `cargo nextest run -p meta-srv gc`
- `cargo nextest run -p cmd test_load_metasrv_example_config`
- `cargo clippy -p common-meta -p meta-srv -p cmd --all-targets --all-features -- -D warnings`
- `make fmt` and `make fmt-toml`
- `make config-docs`

## PR Checklist


src/common/meta/src/ddl/drop_table.rs
```diff
@@ -71,7 +72,19 @@ impl DropTableProcedure {
     }
 
     pub fn from_json(json: &str, context: DdlContext) -> ProcedureResult<Self> {
-        let data: DropTableData = serde_json::from_str(json).context(FromJsonSnafu)?;
+        let mut data: DropTableData = serde_json::from_str(json).context(FromJsonSnafu)?;
+        if !context.soft_drop_enabled {
```
fengjiachun: Overwriting the persisted mode here makes recovery destructive. If a soft-drop procedure was persisted after `DeleteMetadata` and metasrv restarts with this switch disabled (or is downgraded to this build), it resumes as a hard drop: it sends `Drop` instead of `Close` and then deletes the tombstone. The gate can prevent creating new soft drops, but recovery should preserve `data.soft_drop_enabled` (or stop with an explicit incompatibility error) rather than silently changing an in-flight operation's meaning.

src/meta-srv/src/metasrv/builder.rs
```diff
@@ -219,6 +219,7 @@ impl MetasrvBuilder {
         } = self;
 
         let options = options.unwrap_or_default();
+        options.gc.validate()?;
```
fengjiachun: This makes every GC-only setting startup-fatal even when `gc.enable = false`. Previously `GcScheduler::new_with_config` performed this validation only inside the enabled branch, so an existing deployment with disabled GC and an unused value such as `max_concurrent_tables = 0` will now fail to start; the new `experimental_soft_drop` settings are explicitly ignored and do not require this. Please keep the existing validation boundary, or validate only fields that are actually consumed while GC is disabled.

src/common/meta/src/key/tombstone.rs
```diff
@@ -308,7 +322,10 @@ impl TombstoneManager {
                 )
             }
         );
-        let chunk_size = max_txn_ops / txn_ops_per_key;
+        let merge_extra_ops =
+            !extra_ops.is_empty() && max_txn_ops.saturating_sub(txn_ops_per_key) >= extra_ops.len();
+        let reserved_ops = if merge_extra_ops { extra_ops.len() } else { 0 };
+        let chunk_size = (max_txn_ops - reserved_ops) / txn_ops_per_key;
         if keys.len() > chunk_size {
```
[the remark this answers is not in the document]
v0y4g3r: Fixed in 6a97fb445b. The combined key/extra operation count now forces chunking, and marker-only operations are also chunked by `max_txn_ops`. Added regressions for both a key move followed by oversized marker ops and an empty-key marker-only transaction.

src/common/meta/src/ddl/drop_table.rs
v0y4g3r: Fixed in 6a97fb445b. Recovery now preserves the persisted `soft_drop_enabled` mode, so an in-flight soft drop continues with `Close` and retains its tombstone. The runtime gate only prevents creating new soft-drop procedures; configured retention remains available as a fallback for legacy snapshots.

src/meta-srv/src/metasrv/builder.rs
v0y4g3r: Fixed in 6a97fb445b. Removed unconditional builder validation; GC settings are again validated only when `gc.enable = true` through `GcScheduler::new_with_config`. Added a builder regression for disabled GC with `max_concurrent_tables = 0`.

src/meta-srv/src/gc/scheduler.rs
```diff
@@ -229,6 +222,49 @@ impl GcScheduler {
...
+    async fn purge_expired_soft_dropped_tables(&self, now_millis: i64) {
+        // The scheduler is only constructed after GcSchedulerOptions::validate().
+        let retention_millis = self.config.experimental_soft_drop.retention.as_millis() as i64;
+        let dropped_tables = match self.ctx.list_dropped_tables().await {
+            Ok(dropped_tables) => dropped_tables,
+            Err(error) => {
+                error!(error; "Failed to list soft-dropped tables for GC");
+                return;
+            }
+        };
```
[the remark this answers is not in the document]
v0y4g3r: Fixed in 6a97fb445b using checked conversion with saturation to `i64::MAX`; oversized retention can no longer wrap and trigger an immediate legacy purge.

src/meta-srv/src/gc/scheduler.rs
```diff
@@ -229,6 +222,49 @@ impl GcScheduler {
         Ok(report)
     }
 
+    async fn purge_expired_soft_dropped_tables(&self, now_millis: i64) {
+        // The scheduler is only constructed after GcSchedulerOptions::validate().
+        let retention_millis = self.config.experimental_soft_drop.retention.as_millis() as i64;
```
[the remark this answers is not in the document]
v0y4g3r: Fixed in 6a97fb445b. The conversion now uses `i64::try_from(...).unwrap_or(i64::MAX)`, with a regression covering retention beyond `i64::MAX` milliseconds.

v0y4g3r:
@codex review

src/common/meta/src/ddl/drop_table/executor.rs
```diff
@@ -110,7 +114,7 @@ impl DropTableExecutor {
             }
         );
 
-        if ctx.soft_drop_enabled
+        if soft_drop_enabled
             && let Some(dropped_table) = ctx
                 .table_metadata_manager
                 .get_dropped_table(&self.table)
```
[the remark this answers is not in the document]
v0y4g3r: Fixed in 0cb63233dd. Hard drops now reject a different-ID tombstone when the existing tombstone has a `dropped_at` marker, while still allowing unmarked orphan hard-drop tombstones. Drop procedures also acquire `TableNameLock` so same-name recreation cannot observe the chunked metadata-move window before the marker is published. Added regressions for the disabled-gate hard-drop scenario and the drop lock set.

v0y4g3r:
@codex review

src/common/meta/src/ddl/purge_dropped_table.rs
```diff
@@ -129,10 +155,19 @@ impl PurgeDroppedTableProcedure {
...
+        } else {
+            Self::TYPE_NAME
+        }
     }
 
     fn recover(&mut self) -> ProcedureResult<()> {
+        // Recovery reacquires TableLock, so automatic purges must re-read the
+        // tombstone in case it was undropped and dropped again while unlocked.
+        if self.data.check_expired {
+            self.data.state = PurgeDroppedTableState::Prepare;
```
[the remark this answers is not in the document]
v0y4g3r: Fixed in 582dae537d. Recovery now rewinds only pre-cleanup automatic purges persisted at `DropRegions`. A recovered `DeleteTombstone` remains post-cleanup and continues deleting the tombstone, avoiding recoverable-looking metadata after region files may already be gone. Added comments documenting this boundary; manual purge behavior is unchanged.

v0y4g3r:
@WenyXu @fengjiachun PTAL

v0y4g3r:
@codex review

src/common/meta/src/ddl/purge_dropped_table.rs
```diff
@@ -129,10 +155,21 @@ impl PurgeDroppedTableProcedure {
...
+            Self::EXPIRED_TYPE_NAME
+        } else {
+            Self::TYPE_NAME
+        }
     }
 
     fn recover(&mut self) -> ProcedureResult<()> {
+        // Recovery reacquires TableLock, so a pre-cleanup automatic purge must re-read the
+        // tombstone in case it was undropped and dropped again while unlocked.
+        if self.data.check_expired && self.data.state == PurgeDroppedTableState::DropRegions {
```
fengjiachun: A persisted `DropRegions` state is also potentially post-cleanup: the procedure runtime calls `execute()` first and only persists `Status::executing(true)` afterward. If `on_cleanup_regions_offline()` succeeds and metasrv crashes before `DeleteTombstone` is stored, recovery still sees `DropRegions` and this branch rewinds to `Prepare`; a disabled gate or replacement tombstone can then leave recoverable-looking metadata after the region files were cleaned. Please add a durable pre-cleanup/recheck boundary (or fence the exact tombstone generation) so recovery cannot cancel once cleanup may have run.

src/common/meta/src/ddl_manager.rs
```diff
@@ -272,6 +272,18 @@ impl DdlManager {
                 .context(RegisterProcedureLoaderSnafu { type_name })?;
         }
 
+        let type_name = PurgeDroppedTableProcedure::EXPIRED_TYPE_NAME;
```
WenyXu, line 275: Should we register both `PurgeDroppedTableProcedure::EXPIRED_TYPE_NAME` and `PurgeDroppedTableProcedure::TYPE_NAME`?

src/common/meta/src/ddl/purge_dropped_table.rs
v0y4g3r: Fixed in 47c67bc. The purge procedure now writes a durable `__tombstone/__purging/<table_id>` claim before region cleanup, and binds that claim to a UUID `drop_generation` stored with the tombstone. Recovery revalidates the exact generation under `TableLock` before claiming it; once a matching claim exists, cleanup continues even if the runtime gate changes, while foreign or replacement generations are rejected. UNDROP uses the same claim/generation fence so it cannot restore metadata after cleanup has started.

src/common/meta/src/ddl_manager.rs
v0y4g3r: `PurgeDroppedTableProcedure::TYPE_NAME` is already registered inside the macro in line 259


https://github.com/GreptimeTeam/greptimedb/blob/e9920f7b548ac2c6d0a29d04497d03e3276f9cc6/src/common/meta/src/ddl_manager.rs#L259

WenyXu (approved):

PR #8527, opened 2026-07-15
perf: reduce parquet metadata cache footprint

waynexia:

Reduce parquet metadata cache cap requirement by dropping unused column metadata and introducing a compressed tier. This patch brings a previous 1GB meta cache cap requirement down to 64MB without cache churn.

## PR Checklist

### DIFF
--- a/src/mito2/src/access_layer.rs
+++ b/src/mito2/src/access_layer.rs
@@ -16,11 +16,13 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use async_stream::try_stream;
+use common_telemetry::warn;
 use common_time::Timestamp;
 use futures::{Stream, TryStreamExt};
 use object_store::services::Fs;
 use object_store::util::{join_dir, with_instrument_layers};
 use object_store::{ATOMIC_WRITE_DIR, ErrorKind, OLD_ATOMIC_WRITE_DIR, ObjectStore};
+use parquet::file::metadata::PageIndexPolicy;
 use smallvec::SmallVec;
 use snafu::ResultExt;
 use store_api::metadata::RegionMetadataRef;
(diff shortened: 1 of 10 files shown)


waynexia:
@codex review

src/mito2/src/cache/file_cache.rs
```diff
@@ -619,35 +619,30 @@ impl FileCache {
...
-                            key
-                        );
-                        None
-                    }
-                }
-            })
+        let metadata = self
+            .get_parquet_meta_data(key, cache_metrics, page_index_policy)
+            .await?;
+        match prepare_sst_meta(&file_path, metadata, None, page_index_policy).await {
```
fengjiachun, line 635: When `sst_meta_cache_size` is zero but the write cache is enabled, this path still serializes and zstd-compresses every file-cache metadata hit, then drops the prepared entry because neither in-memory tier exists (`test_read_metadata_from_write_cache` covers this configuration). Please keep this path decode-only when the metadata cache is disabled.

src/mito2/src/access_layer.rs
```diff
@@ -410,14 +412,45 @@ impl AccessLayer {
...
+                    );
+                    let page_index_policy = if parquet_metadata.offset_index().is_some() {
+                        PageIndexPolicy::Optional
+                    } else {
+                        PageIndexPolicy::Skip
+                    };
+                    let parquet_metadata = parquet_metadata.clone();
+                    let region_metadata = region_metadata.clone();
+                    let cache_manager = cache_manager.clone();
+                    common_runtime::spawn_global(async move {
```
v0y4g3r: Can we spare `common_runtime::spawn_global` here since the task is spawned inside `prepare_sst_meta`?

fengjiachun (approved):
LGTM

src/mito2/src/cache.rs
```diff
@@ -152,11 +160,235 @@ impl RangeResultMemoryLimiter {
...
+#[derive(Debug)]
+pub(crate) enum SstMetaPreparation {
+    /// Both cache representations are ready for admission.
+    Prepared(PreparedSstMeta),
+    /// The metadata is usable by the reader, but its compact cache encoding failed.
+    DecodedOnly {
+        decoded: Arc<CachedSstMeta>,
+        encoding_error: crate::error::Error,
+    },
+}
```
evenyag, line 206: Did we compare the space efficiency and the time used to decode them?
- raw struct
- encoded without zstd
- zstd compressed

I wonder if it takes more time to decode the CompactSstMeta than to load it from the disk.

waynexia: yes, without this I can only avoid cache churn with 128MB cap on test set, after it the cap can be cut by half and  query latency is merely unchanged (within the noise band)

PR #8528, opened 2026-07-15
refactor: add field id and extension type to histogram

sunng87:

This patch add parquet field ids and extension type to histogram.

## PR Checklist

### DIFF
--- a/src/common/query/src/native_histogram.rs
+++ b/src/common/query/src/native_histogram.rs
@@ -115,31 +115,46 @@ pub fn is_native_histogram_value_schema(name: &str, data_type: &ConcreteDataType
 //
 // External readers resolve nested struct fields by `PARQUET:field_id`, so each
 // sub-field (and list element) needs a stable positive id. The struct schema
-// is fixed (always the same 18 fields), so ids are derived from a reserved
-// range, disjoint from user column ids and mito2 internal ids (`1 << 30`).
+// is fixed (always the same 18 fields). Each histogram column owns a block of
+// ids (offset from a reserved base by the column's id), so several histogram
+// columns in one table get disjoint sub-field ids. The reserved base is
+// disjoint from user column ids and mito2 internal ids (`1 << 30`).
 // ---------------------------------------------------------------------------
 
 /// Reserved base for native-histogram struct sub-field ids.
 pub const NATIVE_HISTOGRAM_SUBFIELD_ID_BASE: i32 = 0x5000_0000;
 
-/// Reserved base for the `element-id` of list-typed native-histogram
-/// sub-fields.
-pub const NATIVE_HISTOGRAM_LIST_ELEMENT_ID_BASE: i32 = 0x5000_0100;
+/// Number of ids reserved per histogram column (18 sub-fields + headroom for
+/// list element ids), so multiple histogram columns get disjoint ids.
+pub const NATIVE_HISTOGRAM_SUBFIELD_ID_STRIDE: i32 = 64;
 
-/// Returns the stable field id for a native-histogram struct sub-field by
-/// name, or `None` if `name` is not a known sub-field.
-pub fn native_histogram_subfield_id(name: &str) -> Option<i32> {
-    NATIVE_HISTOGRAM_FIELD_NAMES
-        .iter()
-        .position(|n| *n == name)
-        .map(|i| NATIVE_HISTOGRAM_SUBFIELD_ID_BASE + i as i32)
+/// Offset of list element ids within a column's id block.
+pub const NATIVE_HISTOGRAM_LIST_ELEMENT_OFFSET: i32 = 32;
+
+/// Returns the stable field id for a native-histogram struct sub-field,
+/// namespaced by its parent `column_id`, or `None` if `name` is not a known
+/// sub-field.
+pub fn native_histogram_subfield_id(column_id: i32, name: &str) -> Option<i32> {
+    let idx = subfield_index(name)?;
+    Some(NATIVE_HISTOGRAM_SUBFIELD_ID_BASE + column_id * NATIVE_HISTOGRAM_SUBFIELD_ID_STRIDE + idx)
 }
 
 /// Returns the stable list `element-id` for a list-typed native-histogram
-/// sub-field by its name, or `None` if `name` is not a known sub-field.
-pub fn native_histogram_list_element_id(name: &str) -> Option<i32> {
+/// sub-field, namespaced by its parent `column_id`, or `None` if `name` is not
+/// a known sub-field.
+pub fn native_histogram_list_element_id(column_id: i32, name: &str) -> Option<i32> {
+    let idx = subfield_index(name)?;
+    Some(
+        NATIVE_HISTOGRAM_SUBFIELD_ID_BASE
+            + column_id * NATIVE_HISTOGRAM_SUBFIELD_ID_STRIDE
+            + NATIVE_HISTOGRAM_LIST_ELEMENT_OFFSET
+            + idx,
+    )
+}
+
+fn subfield_index(name: &str) -> Option<i32> {
     NATIVE_HISTOGRAM_FIELD_NAMES
         .iter()
         .position(|n| *n == name)
-        .map(|i| NATIVE_HISTOGRAM_LIST_ELEMENT_ID_BASE + i as i32)
+        .map(|i| i as i32)
 }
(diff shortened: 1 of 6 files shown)


shuiyisong (commented):
Revalidated against the live PR head after the local branch merged `main`.

src/mito2/src/sst.rs
```diff
@@ -64,14 +64,102 @@ pub enum FormatType {
...
+/// recognized. Other struct columns are left untouched. mito2 reads SST columns
+/// by schema position, never by field metadata, so this only affects external
+/// readers.
+fn stamp_native_histogram_subfield_ids(field: &mut Field) {
+    use arrow_schema::extension::{EXTENSION_TYPE_NAME_KEY, ExtensionType};
+    use common_query::native_histogram::{
+        native_histogram_list_element_id, native_histogram_subfield_id, native_histogram_value_type,
+    };
+    use datatypes::extension::histogram::HistogramExtensionType;
+    if ConcreteDataType::from_arrow_type(field.data_type()) != *native_histogram_value_type() {
```
shuiyisong: **[P3] Keep histogram detection aligned with the canonical schema predicate.** `is_native_histogram_value_schema` defines a native histogram by both `NATIVE_HISTOGRAM_FIELD` and the exact value type, but this check uses only the type. A regular Struct column with the same 18-field shape would therefore be tagged `greptime.histogram`; two such columns would also receive identical nested field IDs, although [Iceberg requires field IDs to be unique across the schema](https://iceberg.apache.org/spec/#schemas-and-data-types). Please reuse the existing helper here:

```rust
if !is_native_histogram_value_schema(
    field.name(),
    &ConcreteDataType::from_arrow_type(field.data_type()),
) {
    return;
}
```

This keeps recognition consistent and ensures only the canonical field is stamped.

src/mito2/src/sst/parquet/writer.rs
```diff
@@ -457,7 +458,7 @@ where
                 self.bytes_written.clone(),
             );
             let arrow_writer =
-                AsyncArrowWriter::try_new(writer, schema.clone(), Some(writer_props))
+                AsyncArrowWriter::try_new(writer, maybe_wrap_schema(schema), Some(writer_props))
```
shuiyisong: **[P2] Add persisted-format coverage for this writer path.** This changes SST/Parquet schema metadata, but the current tests only exercise the extension helper; none calls `maybe_wrap_schema` or inspects the written footer. The repository invariant also requires a compatibility case when persisted metadata changes. Please add one focused SST writer/footer test that asserts the extension plus child/list-element IDs, verifies a same-shaped noncanonical Struct remains untouched, and add the corresponding compatibility read case.

src/mito2/src/sst.rs
```diff
@@ -64,14 +64,102 @@ pub enum FormatType {
...
+    );
+    let ArrowDataType::Struct(children) = field.data_type() else {
+        return;
+    };
+    let new_children: Fields = children
+        .iter()
+        .map(|child| {
+            let mut c = (**child).clone();
+            // Stamp the sub-field's own id.
+            if let Some(id) = native_histogram_subfield_id(c.name()) {
```
fengjiachun: These IDs are derived only from the child name, while `maybe_wrap_schema()` applies them to every type-matching top-level field. If a schema contains two histogram-shaped struct columns, both receive identical nested field IDs, but Iceberg field IDs must be unique within a table schema, making external projection ambiguous. Please either restrict stamping to the single reserved histogram column or derive nested IDs from the parent field ID, and add a regression test that covers multiple matches and inspects the emitted Parquet schema.

sunng87: good catch

sunng87: but currently histogram has a fixed column name, so there will be only one such column it a table.

this may change in future if we make histogram a datatype and loose the requirement of field name.

sunng87:
@fengjiachun @shuiyisong PTAL again

src/common/query/src/native_histogram.rs
```diff
@@ -109,3 +109,52 @@ pub fn native_histogram_value_type() -> &'static ConcreteDataType {
...
+
+/// Offset of list element ids within a column's id block.
+pub const NATIVE_HISTOGRAM_LIST_ELEMENT_OFFSET: i32 = 32;
+
+/// Returns the stable field id for a native-histogram struct sub-field,
+/// namespaced by its parent `column_id`, or `None` if `name` is not a known
+/// sub-field.
+pub fn native_histogram_subfield_id(column_id: i32, name: &str) -> Option<i32> {
+    let idx = subfield_index(name)?;
+    Some(NATIVE_HISTOGRAM_SUBFIELD_ID_BASE + column_id * NATIVE_HISTOGRAM_SUBFIELD_ID_STRIDE + idx)
```
fengjiachun: ColumnId is u32, but this expression uses unchecked i32 arithmetic. At column_id = 12_582_912, BASE + column_id * 64 exceeds i32::MAX, so debug builds panic and optimized builds can emit a negative or reused field ID. Please either keep fixed nested IDs now that matching is restricted to the single canonical histogram column, or use checked arithmetic and enforce a representable parent-ID range.

src/mito2/src/sst/parquet/writer.rs
fengjiachun: The new tests exercise maybe_wrap_schema() directly, but they still do not pass the schema through AsyncArrowWriter or inspect the emitted Parquet schema/footer. That leaves the on-disk contract this change is meant to provide unverified, and the repository requires a compatibility case for persisted-format changes. Please add a focused writer/footer assertion plus the corresponding compatibility read case.

src/mito2/src/sst.rs
```diff
@@ -64,14 +70,109 @@ pub enum FormatType {
 /// Iceberg-compatible column field ID key stored in Parquet column metadata.
 pub const PARQUET_FIELD_ID_KEY: &str = "PARQUET:field_id";
 
-/// Adds `PARQUET:field_id` metadata to an Arrow field.
+/// Adds `PARQUET:field_id` metadata to a top-level Arrow field.
+///
+/// Native-histogram sub-field ids are stamped separately at parquet-write
+/// time by [`stamp_native_histogram_schema_ids`], not here.
```
fengjiachun: This link still points to the nonexistent stamp_native_histogram_schema_ids function; the actual function is stamp_native_histogram_subfield_ids, so rustdoc reports an unresolved link.

src/common/query/src/native_histogram.rs
sunng87: Addressed in 971bd68. The sub-field id is now derived with checked arithmetic, and `maybe_wrap_schema` is fallible: when a sub-field id cannot be resolved (unknown name or i32 overflow) the SST write returns a new `InvalidNativeHistogramSubfield` error instead of wrapping, panicking, or silently dropping the id. Added unit tests pinning the overflow boundary (`None` at the helper, error at the writer).

src/mito2/src/sst/parquet/writer.rs
sunng87: Addressed in 971bd68. Added `parquet_footer_arrow_schema`, which writes a schema through `maybe_wrap_schema` + a real parquet `ArrowWriter` and reads the footer back, with two tests: (1) the `greptime.histogram` extension and every nested `PARQUET:field_id` (incl. list elements) survive on disk, and (2) a same-shaped non-canonical struct is left untouched.

Re the cross-version compat case: the repo compat suite runs `setup.sql` on an *old* binary, but native histograms are experimental and absent from prior releases, so a cross-version SQL case is not feasible there; the on-disk read-back contract is covered at the unit level instead. Happy to add a SQL-level current-version round-trip if you prefer.

src/mito2/src/sst.rs
sunng87: Addressed in 971bd68 — renamed the broken link to `stamp_native_histogram_subfield_ids`.

sunng87:
@fengjiachun Please check again

src/mito2/src/sst.rs
```diff
@@ -64,14 +71,130 @@ pub enum FormatType {
...
+    }
+    // Namespace sub-field ids by the parent column's field id (its
+    // `PARQUET:field_id`, stamped earlier by `with_field_id`) so several
+    // histogram columns in one table get disjoint ids. Falls back to 0 when
+    // the parent id is absent.
+    let column_id = field
+        .metadata()
+        .get(PARQUET_FIELD_ID_KEY)
+        .and_then(|s| s.parse::<i32>().ok())
+        .unwrap_or(0);
```
fengjiachun: with_field_id serializes this metadata from a u32, but parsing it as i32 and falling back to 0 means a missing or malformed ID, and every valid ID above i32::MAX, silently reuses column zero nested IDs. That bypasses the checked overflow path just added. Please parse as u32 and propagate a checked conversion or derivation error instead of unwrap_or(0), with a regression for u32::MAX.

fengjiachun (approved):
LGTM

PR #8441, opened 2026-07-08
fix: serialize repartitioned index rebuilds

discord9:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

GC/GetFileRefs changes were split out and merged separately in #8445. This PR now keeps only the rebuild-index-after-repartition fix.

## What's changed and what's your intention?

This PR fixes rebuild-index after repartition/remap when multiple current regions refer to the same physical SST file:

- Use the current/referring region for scheduling, worker notification, and manifest apply, while using the SST file's origin physical region for Puffin path, cache key, and listener region-file IDs.
- Add a WorkerGroup-level physical index build coordinator keyed by physical `RegionFileId`, so duplicate physical rebuilds are singleflighted across workers.
- Publish a current-agnostic `PhysicalIndexPatch` and let each current region apply that patch to its own manifest.
- Keep active `Publishing` state counted until all current-region manifest applies finish, and promote live waiters if a builder aborts before publishing.
- Keep a bounded/TTL completed-patch cache so stale current manifests can reuse a recently published physical index instead of rewriting physical v0 Puffin files.
- Add a defensive guard to avoid blindly overwriting an existing physical v0 Puffin when cached metadata is unavailable.

Tests run:

- `cargo fmt --package mito2`
- `git diff --check`
- `cargo nextest run -p mito2 test_index_build_type_manual_duplicate_in_flight`
- `cargo nextest run -p mito2 index_build`

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [ ] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [x] API changes are backward compatible.
- [x] Schema or data changes are backward compatible.

PR #8531, opened 2026-07-15
fix: harden repartition gc cleanup

discord9:

Follow-up hardening after #8445.

#8441 is related to repartitioned index rebuild serialization, but intentionally out of scope here.

This PR hardens `table_repart` cleanup to be fail-closed under partial GC file-reference observations.

Before this change, `cleanup_region_repartition()` rewrote a source region's destination set from exactly the destinations observed in the current GC file-ref snapshot. If a GC cycle had partial or retrying observations, that could shrink or remove the only persisted `table_repart` mapping used by later dropped-region GC cycles.

The new cleanup logic:

- starts from the existing source→destination mapping as the baseline;
- unions positive observed `cross_region_refs` instead of replacing existing mappings;
- prunes destinations only when the destination manifest was observed in this GC cycle, the source GC completed without retry, and there are no tmp refs;
- preserves empty tombstones when tmp refs or retry/no-success conditions mean the source should remain discoverable;
- handles source self-destinations when the source manifest was observed.

This PR also adds regression coverage:

- unit tests for preserving unobserved destinations, pruning only observed stale destinations, keeping tmp-ref tombstones, preserving positive refs on retry, and pruning observed self-destinations;
- a File-only integration test for plain non-index Mito data SSTs across multi-hop repartition, explicit `full_file_listing=true` table GC, full GC, query correctness, and manifest data path survival in storage.

Limitations / scope:

- This does not add broader GC/repartition locking.
- This does not enforce related manifest versions during deletion.
- `FileRefsManifest.manifest_version` is used only as proof that a normal manifest was observed for safe cleanup pruning.
- #8445's missing-related-target fail-closed behavior is a prerequisite, not reimplemented here.
- Rebuild-index serialization remains out of scope for this PR.

Compatibility:

- No API, wire-format, or persisted-format changes.

Validation:

- `cargo fmt -p meta-srv -p tests-integration -- --check`
- `git diff --check`
- `cargo nextest run -p meta-srv --features mock test_reconcile_repartition_mapping`
- `cargo test -p tests-integration --test main test_repartition_multi_hop_plain_gc_full_file_listing_file -- --nocapture`

## PR Checklist

### DIFF
--- a/src/meta-srv/src/gc/procedure.rs
+++ b/src/meta-srv/src/gc/procedure.rs
@@ -47,6 +47,52 @@ use crate::metrics::{METRIC_META_GC_DATANODE_CALLS_TOTAL, METRIC_META_GC_FAILED_
 use crate::procedure::utils::{instruction_error_result, instruction_to_error};
 use crate::service::mailbox::{Channel, MailboxReceiver, MailboxRef};
 
+fn reconcile_region_repartition_mapping(
+    repartition_value: &mut TableRepartValue,
+    src_region: RegionId,
+    observed_manifest_dst_regions: &HashSet<RegionId>,
+    observed_dst_regions: Option<&HashSet<RegionId>>,
+    has_tmp_ref: bool,
+    source_gc_succeeded: bool,
+) {
+    let had_existing_entry = repartition_value.src_to_dst.contains_key(&src_region);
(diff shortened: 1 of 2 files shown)


discord9:

PR #8538, opened 2026-07-15
fix: count Postgres SCRAM auth failures and correct auth config docs

killme2008:
Follow-up to #8304 (already merged), addressing the two remaining review comments from @fengjiachun.

## What changed

**1. Count SCRAM auth failures in the metric** (`src/servers/src/postgres/auth_handler.rs`)

The SCRAM SASL paths returned `Failed` without incrementing `METRIC_AUTH_FAILURE`, so once SCRAM was enabled, wrong-password and unknown-user attempts vanished from `greptime_servers_auth_failure_count`. Every SCRAM rejection now funnels through `record_scram_failure`, recording one failure with a uniform `UserPasswordMismatch` label — the counter stays useful without revealing whether the user exists.

**2. Correct per-protocol auth notes in the config examples** (`config/*.example.toml`, regenerated `config/config.md`)

The `user_provider` note no longer matched the implementation:
- `pbkdf2_sha256` is excluded from Postgres SCRAM, so its iteration count is never exposed in a SCRAM handshake (the enumeration caveat now applies only to `pg_scram_sha256`).
- Restored the warning that `pbkdf2_sha256` / `pg_scram_sha256` cannot authenticate over MySQL’s native password handshake, and that `mysql_native_password` cannot authenticate over Postgres.

## Checklist
- [x] `make fmt`, lib `cargo clippy` clean
- [x] Regenerated `config/config.md` via `make config-docs`
- [x] No API/wire/format changes


src/servers/src/postgres/auth_handler.rs
```diff
@@ -409,51 +429,54 @@ impl PostgresServerHandlerInner {
...
-                    return Ok(PgAuthenticationResult::Failed);
-                }
+                    drop(state);
+                    if self
+                        .login_verifier
+                        .authorize(login_info, &user_info)
+                        .await
+                        .is_err()
+                    {
+                        break 'sasl PgAuthenticationResult::Failed;
```
fengjiachun: `authorize()` already increments `METRIC_AUTH_FAILURE` when it returns an error. Breaking to `Failed` here then runs `record_scram_failure()` below, so one authorization rejection is counted twice and the second increment is mislabeled as `UserPasswordMismatch`. Please bypass the generic recorder for this already-recorded path (or centralize metric ownership), and add a regression test that asserts one increment per rejected attempt.

killme2008: good catch!

Fixed in https://github.com/GreptimeTeam/greptimedb/pull/8538/changes/13867d4b9f4bb2792198bf52803fc32eabce6a7c

fengjiachun (approved):
Rest LGTM

PR #8539, opened 2026-07-15
fix: prevent credential leaks in sanitize_connection_string

raphaelroshan:
Reopens #8486, which was accidentally closed (and whose branch history I've since repaired, so it couldn't be reopened in place).

## What's changed and why

`sanitize_connection_string` redacts `store_addrs` before they're logged, but the previous hand-rolled parsing missed several forms and could leak passwords.

Per the review on #8486 (@killme2008 suggested the `url` crate; @v0y4g3r listed additional libpq formats):

- **URL-form addresses** (`mysql://`, `postgresql://`, `http://` etcd) are now parsed with the **`url` crate**; the userinfo password and any `password` query parameter are redacted.
- **libpq keyword DSNs** (`host=… password=…`) are handled by a small redactor covering `password = x` (spaces around `=`), single/double-quoted values with spaces, and `\`-escaped quotes.

I kept a dedicated redactor for the keyword-DSN form rather than `tokio-postgres`, since `tokio-postgres` is an optional/feature-gated dependency in `common-meta` (this function runs unconditionally) and `store_addrs` also covers mysql/etcd, which it can't parse.

## Checklist

- [x] I have added the necessary unit tests.

By submitting this pull request, I confirm that my contribution is made under the terms of the GreptimeDB CLA.

### DIFF
--- a/src/common/meta/src/kv_backend/util.rs
+++ b/src/common/meta/src/kv_backend/util.rs
@@ -12,46 +12,180 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// Removes sensitive information like passwords from connection strings.
+use url::Url;
+
+/// Placeholder substituted for any password found in a connection string.
+const REDACTED: &str = "***";
+
+/// Substituted for a whole URI-shaped connection string that no parser accepted.
+const REDACTED_URI: &str = "<redacted connection string>";
+
+/// Removes sensitive information (passwords) from a connection string before it
+/// is logged.
 ///
-/// This function sanitizes connection strings by removing credentials:
-/// - For URL format (mysql://user:password@host:port/db): Removes everything before '@'
-/// - For parameter format (host=localhost password=secret): Removes the password parameter
-/// - For URL format without credentials (mysql://host:port/db): Removes the protocol prefix
+/// `store_addrs` entries can be a PostgreSQL DSN (a `postgres(ql)://` URL or a
+/// libpq keyword string), a MySQL/etcd URL, or an etcd host list.
 ///
-/// # Arguments
-///
-/// * `conn_str` - The connection string to sanitize
-///
-/// # Returns
-///
-/// A sanitized version of the connection string with sensitive information removed
+/// The function fails closed: it never returns a URI-shaped input verbatim. In
+/// order:
+/// 1. PostgreSQL DSNs are parsed with `tokio_postgres::Config` — the backend's
+///    own parser — and logged via its `Debug`, which redacts the password. This
+///    matches the grammar exactly (multi-host URIs, `\`-escapes, any Unicode
+///    whitespace, percent-encoded query keys, `&`/`;`/`://` inside values).
+/// 2. Any other URL is redacted with the `url` crate.
+/// 3. A URI-shaped input that neither parser accepted (e.g. a malformed
+///    multi-host Postgres URI, with or without leading whitespace) is redacted
+///    in full — credentials in an unparsable authority or query cannot be split
+///    out reliably, so the whole string is replaced rather than reparsed.
+/// 4. Anything else is treated as a keyword string and best-effort redacted.
 pub fn sanitize_connection_string(conn_str: &str) -> String {
-    // Case 1: URL format with credentials (mysql://user:password@host:port/db)
-    // Extract everything after the '@' symbol
-    if let Some(at_pos) = conn_str.find('@') {
-        return conn_str[at_pos + 1..].to_string();
+    // Redact explicit keyword-style password assignments before parsing. A
+    // malformed string such as `host=x;password=secret` can otherwise parse as
+    // a host value, causing `Config`'s Debug output to retain the password text.
+    if !is_uri_like(conn_str) {
+        let redacted = redact_keyword_password(conn_str);
+        if redacted != conn_str {
+            return redacted;
+        }
     }
 
-    // Case 2: Parameter format with password (host=localhost password=secret dbname=mydb)
-    // Filter out any parameter that starts with "password="
-    if conn_str.contains("password=") {
-        return conn_str
-            .split_whitespace()
-            .filter(|param| !param.starts_with("password="))
-            .collect::<Vec<_>>()
-            .join(" ");
+    #[cfg(feature = "pg_kvbackend")]
+    if let Ok(config) = conn_str.parse::<tokio_postgres::Config>() {
+        // `Config`'s Debug prints the password as `Some("REDACTED")`.
+        return format!("{config:?}");
     }
 
-    // Case 3: URL format without credentials (mysql://host:port/db)
-    // Extract everything after the protocol prefix
-    if let Some(host_part) = conn_str.split("://").nth(1) {
-        return host_part.to_string();
+    if let Ok(url) = Url::parse(conn_str) {
+        return redact_url(url);
     }
 
-    // Case 4: Already sanitized or unknown format
-    // Return as is
-    conn_str.to_string()
+    if is_uri_like(conn_str) {
+        return REDACTED_URI.to_string();
+    }
+
+    redact_keyword_password(conn_str)
+}
+
+/// Redacts the userinfo password and any `password` query parameter of a parsed URL.
+fn redact_url(mut url: Url) -> String {
+    if url.password().is_some() {
+        let _ = url.set_password(Some(REDACTED));
+    }
+    if url.query_pairs().any(|(key, _)| key == "password") {
+        let redacted: Vec<(String, String)> = url
+            .query_pairs()
+            .map(|(key, value)| {
+                let value = if key == "password" {
+                    REDACTED.to_string()
+                } else {
+                    value.into_owned()
+                };
+                (key.into_owned(), value)
+            })
+            .collect();
+        url.query_pairs_mut().clear().extend_pairs(redacted);
+    }
+    url.to_string()
+}
+
+/// True if `s`, ignoring leading whitespace, begins with a valid URI scheme
+/// followed by `://` (`scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`).
+/// Leading whitespace (including Unicode, e.g. U+2003) does not stop a value from
+/// being a URI, so it is trimmed first; a bare `contains("://")` would instead
+/// misfire on a `://` inside a keyword value.
+fn is_uri_like(s: &str) -> bool {
+    let s = s.trim_start();
+    let Some(pos) = s.find("://") else {
+        return false;
+    };
+    let mut chars = s[..pos].chars();
+    matches!(chars.next(), Some(c) if c.is_ascii_alphabetic())
+        && chars.all(|c| c.is_ascii_alphanumeric() || matches!(c, '+' | '-' | '.'))
+}
+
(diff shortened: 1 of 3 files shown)


src/common/meta/src/kv_backend/util.rs
```diff
@@ -12,74 +12,190 @@
...
+/// keyword DSNs are handled by [`redact_keyword_password`], which understands
+/// optional spaces around `=`, single/double quoting, and backslash escapes.
 pub fn sanitize_connection_string(conn_str: &str) -> String {
-    // Case 1: URL format with credentials (mysql://user:password@host:port/db)
-    // Extract everything after the '@' symbol
-    if let Some(at_pos) = conn_str.find('@') {
-        return conn_str[at_pos + 1..].to_string();
+    match Url::parse(conn_str) {
+        Ok(url) => redact_url(url),
+        Err(_) => redact_keyword_password(conn_str),
```
fengjiachun: `Url::parse` rejects valid PostgreSQL multi-host URIs such as `postgresql://user:secret@host1:1234,host2,host3:5678/db` with `InvalidPort`, while `tokio-postgres` accepts this format. This fallback then treats it as a keyword DSN and returns the URL unchanged, so `MetasrvOptions::Debug` logs the full password. Please use a conservative URL fallback or a parser compatible with the backend so parse failures cannot retain userinfo credentials.

src/common/meta/src/kv_backend/util.rs
```diff
@@ -12,74 +12,190 @@
...
+            }
+            if b[k] == quote {
+                return k + 1; // include the closing quote
+            }
+            k += 1;
+        }
+        return b.len(); // unterminated quote
+    }
+    let mut k = start;
+    while k < b.len() && !matches!(b[k], b' ' | b'\t' | b'\n' | b'\r' | b'&') {
```
fengjiachun: Unquoted libpq values end only at whitespace, and `\` escapes the next character; `&` is part of the value. With `password=secret&suffix`, this emits `password=***&suffix`; with `password=secret\ suffix`, it emits `password=*** suffix`, leaking credential suffixes. Please mirror the unquoted-value grammar used by the backend parser and cover both cases.

src/common/meta/src/kv_backend/util.rs
```diff
@@ -12,74 +12,190 @@
...
+        // Not a `password` key: copy one character and continue.
+        let ch = s[i..].chars().next().unwrap();
+        out.push(ch);
+        i += ch.len_utf8();
+    }
+    out
+}
+
+fn skip_spaces(bytes: &[u8], mut i: usize) -> usize {
+    while i < bytes.len() && (bytes[i] == b' ' || bytes[i] == b'\t') {
```
fengjiachun: `tokio-postgres` accepts any `char::is_whitespace` around `=`, but this helper only skips spaces and tabs. A valid multiline DSN such as `password\n=\nsecret` is therefore returned with the full password intact. Please match the backend whitespace grammar for both token boundaries and spaces around `=`.

src/common/meta/src/kv_backend/util.rs
```diff
@@ -12,74 +12,190 @@
...
+            }
+            k += 1;
+        }
+        return b.len(); // unterminated quote
+    }
+    let mut k = start;
+    while k < b.len() && !matches!(b[k], b' ' | b'\t' | b'\n' | b'\r' | b'&') {
+        k += 1;
+    }
+    k
```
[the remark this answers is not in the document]
fengjiachun: `;` is not a delimiter in libpq keyword DSNs. `tokio-postgres` parses `password=secret;host=localhost` with the entire `secret;host=localhost` as the password, so adding `;` to `skip_value` would leak part of the credential. The correct fix is to remove `;` from the boundary set in `is_keyword_at`. Similarly, `&` should not terminate an unquoted libpq value.

src/common/meta/src/kv_backend/util.rs
```diff
@@ -12,74 +12,284 @@
...
+/// keyword DSNs are handled by [`redact_keyword_password`].
 pub fn sanitize_connection_string(conn_str: &str) -> String {
-    // Case 1: URL format with credentials (mysql://user:password@host:port/db)
-    // Extract everything after the '@' symbol
-    if let Some(at_pos) = conn_str.find('@') {
-        return conn_str[at_pos + 1..].to_string();
+    if let Ok(url) = Url::parse(conn_str) {
+        return redact_url(url);
+    }
+    if conn_str.contains("://") {
```
fengjiachun: Checking `contains("://")` is too broad: valid keyword DSNs can contain `://` inside any unquoted value. `tokio-postgres` accepts `host=localhost password=secret application_name=http://client` (and `password=http://secret`), but this branch sends those strings to `redact_url_shaped`, which returns the password unchanged. Please only take this fallback when the input actually starts with a valid URI scheme followed by `://`.

src/common/meta/src/kv_backend/util.rs
```diff
@@ -12,74 +12,284 @@
...
+/// Redacts a `password` value in the query portion of a URL tail (`?a=b&password=x`).
+fn redact_query_password(tail: &str) -> String {
+    let Some(q) = tail.find('?') else {
+        return tail.to_string();
+    };
+    let (path, query) = tail.split_at(q + 1); // keep the '?'
+    let redacted = query
+        .split('&')
+        .map(|pair| match pair.split_once('=') {
+            Some(("password", _)) => format!("password={REDACTED}"),
```
fengjiachun: Fallback query matching needs percent-decoding. `tokio-postgres` accepts `postgresql://user@host1:1,host2/db?pass%77ord=qsecret` and decodes the key to `password`, while this raw match leaves `qsecret` in the logged string. Please decode query keys with the same form-urlencoding rules before deciding whether to redact.

src/common/meta/src/kv_backend/util.rs
```diff
@@ -12,74 +12,284 @@
...
+            }
+        }
+        let ch = s[i..].chars().next().unwrap();
+        out.push(ch);
+        i += ch.len_utf8();
     }
+    out
+}
+
+fn is_ascii_ws(b: u8) -> bool {
```
fengjiachun: `tokio-postgres` uses `char::is_whitespace`, but this ASCII-only predicate still misses valid separators such as U+2003. Both `host=localhost\u{2003}password=secret` and `password\u{2003}=\u{2003}secret` parse successfully and are logged unchanged here. Please scan chars with the backend whitespace rule for token boundaries and around `=`.

raphaelroshan:
Thank you for reviewing this in such detail, let me try to find a good way to fix this

raphaelroshan:
Pivoted to tokio_postgres::Config for postgres DSNs — parsing with the backend's own grammar and logging its Debug (which redacts the password). That covers all the cases you flagged: multi-host URIs, percent-encoded query keys, any Unicode whitespace, `\`-escapes, and `&`/`;`/`://` inside values. Non-postgres URLs still go through the url crate. Trade-off: a pg DSN now logs as a redacted `Config { … }` rather than a DSN string — happy to reshape if you'd prefer.

src/common/meta/src/kv_backend/util.rs
```diff
@@ -12,74 +12,195 @@
...
-/// A sanitized version of the connection string with sensitive information removed
+/// Other URLs go through the `url` crate; anything else falls back to a
+/// best-effort keyword redactor.
 pub fn sanitize_connection_string(conn_str: &str) -> String {
-    // Case 1: URL format with credentials (mysql://user:password@host:port/db)
-    // Extract everything after the '@' symbol
-    if let Some(at_pos) = conn_str.find('@') {
-        return conn_str[at_pos + 1..].to_string();
+    #[cfg(feature = "pg_kvbackend")]
+    if let Ok(config) = conn_str.parse::<tokio_postgres::Config>() {
```
fengjiachun, line 53: Parsing with the backend only on success still leaks credentials for invalid multi-host PostgreSQL URIs. For example, `postgresql://user:secret@host1:1234,host2:5678/db?unknown=x` is rejected by `tokio-postgres` (`UnknownOption`) and by `Url` (`InvalidPort`), so this falls through to the keyword redactor and returns the original string, including `secret`. Since this function protects logs even for malformed configuration, please fail closed for `postgres(ql)://` inputs (redact userinfo or the entire DSN on parse failure) and add this case as a regression test.

raphaelroshan:
Thanks so much for the detailed review! Have updated the PR:
  
  1. Fail closed on unparseable URIs. Previously a scheme://… string that neither tokio_postgres::Config nor the
  url crate accepted (e.g. a malformed multi-host URI) fell through to the keyword path and could be logged
  verbatim. It now goes through redact_uri_userinfo: the userinfo password is redacted (user:***@…) even when
  the host section is malformed, and any password query parameter is redacted too — with percent-decoding of the
  key, so pass%77ord=… is caught. starts_with_uri_scheme validates the scheme grammar (ALPHA *( ALPHA / DIGIT /
  "+" / "-" / "." )) so a :// inside a keyword value isn't mistaken for a URI.

  2. Unicode whitespace. The keyword fallback now tokenizes on char::is_whitespace rather than ASCII space, so
  separators like U+2003 no longer leak the value.

  The real DSN grammar is still delegated entirely to tokio_postgres::Config; the hand-rolled logic only runs as
  a fallback when pg_kvbackend is disabled. Regression tests cover both (test_sanitize_fail_closed_uri,
  test_sanitize_keyword) and pass with pg_kvbackend on and off.

src/common/meta/src/kv_backend/util.rs
```diff
@@ -12,74 +12,319 @@
...
+
+/// True if `s` begins with a valid URI scheme followed by `://`
+/// (`scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`). Unlike a bare
+/// `contains("://")`, this does not misfire on a `://` inside a keyword value.
+fn starts_with_uri_scheme(s: &str) -> bool {
+    let Some(pos) = s.find("://") else {
+        return false;
+    };
+    let mut chars = s[..pos].chars();
+    matches!(chars.next(), Some(c) if c.is_ascii_alphabetic())
```
fengjiachun, line 102: `starts_with_uri_scheme` still fails open when an otherwise URI-shaped value has leading whitespace. For example, ` postgresql://user:secret@host1:1234,host2:5678/db?unknown=x` is rejected by `tokio-postgres` and `Url` (`InvalidPort`); this predicate is then false because the prefix begins with a space, so the keyword fallback returns the original string including `secret`. U+2003 has the same result. Since malformed input is exactly the threat model for this fallback, please either account for leading whitespace during scheme detection or, more robustly, redact the entire unparseable URI-like input instead of reparsing it.

raphaelroshan:
I have updated it further to adress the feedback, hopefully it works - Do let me know if you would rather take over this change since I'm not super familiar with this stuff. Thank you for the detailed feedback!

src/common/meta/src/kv_backend/util.rs
v0y4g3r: resolve c9320d0d

v0y4g3r:
@raphaelroshan thanks! There's still a typo. With that fix I'm ok to merge this pull request

v0y4g3r (approved):
LGTM

raphaelroshan:
Thank you for the help @v0y4g3r and @fengjiachun

PR #8541, opened 2026-07-16
perf: preserve dictionary-encoded query labels

waynexia:

use dictionary array for in-memory pk columns to alleviate series key expansion overhead. gain ~24% end-to-end performance improvement on my local 200+ query test set.

## PR Checklist


src/query/src/dummy_catalog.rs
```diff
@@ -316,11 +327,13 @@ impl DummyTableProviderFactory {
                     region_id,
                 })?;
 
-        let scan_request = if let Some(ctx) = query_ctx.as_ref() {
+        let mut scan_request = if let Some(ctx) = query_ctx.as_ref() {
             scan_request_from_query_context(region_id, ctx)?
         } else {
             ScanRequest::default()
         };
+        scan_request.preserve_pk_dictionary_encoding =
```
discord9, line 335: **P1 — Preserve rolling-upgrade compatibility before enabling dictionary scans.** This unconditionally asks a new datanode to return dictionary-encoded tags. During a rolling upgrade, an old frontend/client treats the resulting Arrow Flight dictionary message (`FlightDecoder::try_decode() == None`) as an illegal "none message", so distributed queries against string PKs fail until every consumer is upgraded. Please negotiate this capability in the region query request, or keep dictionary preservation disabled for clients that have not advertised support. An old-client/new-server compatibility test would also catch this.

src/common/recordbatch/src/lib.rs
```diff
@@ -163,6 +163,91 @@ pub fn map_json_type_to_string_schema(schema: SchemaRef) -> (SchemaRef, bool) {
...
+        })
+        .collect();
+
+    if !apply_mapper {
+        return (schema, false);
+    }
+
+    let builder = schema.metadata().iter().fold(
+        SchemaBuilder::try_from(columns)
+            .unwrap()
```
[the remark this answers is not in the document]
discord9: The panic is valid, but the duplicate-alias example is rejected earlier by DataFusion (`DuplicateUnqualifiedField`). A reachable trigger is qualified same-name output, e.g. `SELECT a.host, b.host FROM t a JOIN t b ON a.ts = b.ts`: DataFusion accepts `a.host`/`b.host`, Arrow drops the qualifiers, and this `SchemaBuilder::try_from(...).unwrap()` then rejects the two `host` fields. The requested fix remains the same: preserve the existing Arrow schema shape or propagate a normal error.

src/mito2/src/read/flat_projection.rs
```diff
@@ -297,7 +332,12 @@ impl FlatProjectionMapper {
         for (output_idx, index) in self.batch_indices.iter().enumerate() {
             let mut array = batch.column(*index).clone();
             // Cast dictionary values to the target type.
-            if let ArrowDataType::Dictionary(_key_type, value_type) = array.data_type() {
+            if let ArrowDataType::Dictionary(_key_type, value_type) = array.data_type()
```
discord9, line 335: **P2 — Preserve logical nulls when dictionaries leave this path encoded.** A dictionary key can be valid while the referenced dictionary value is null. `DictionaryVector::is_null()` checks only key validity, so direct Prometheus consumers that bypass `map_query_output()` stringify that logical null as the literal label value `"Null"` instead of omitting it. Please check `get_ref(i).is_null()`/the child value, or decode dictionaries before those consumers; add a nullable-tag label-values/remote-read regression test.

src/common/grpc/src/flight.rs
```diff
@@ -99,15 +99,15 @@ impl FlightEncoder {
...
-                        self.dictionary_tracker.next_dict_id();
-                    }
-                });
-
-                vec1![self.encode_schema(schema.as_ref())]
-            }
+            FlightMessage::Schema(schema) => vec1![
+                self.data_gen
+                    .schema_to_bytes_with_dictionary_tracker(
+                        schema.as_ref(),
```
discord9: **P2 — Keep both public schema-encoding paths on the same dictionary tracker.** `encode(FlightMessage::Schema)` now seeds `self.dictionary_tracker`, but the public `encode_schema(&self, ...)` path still uses a temporary tracker. Existing callers use `encode_schema()` followed by `encode(RecordBatch)`; for a dictionary batch the latter has no dictionary IDs and the `.expect()` panics. Please make `encode_schema` mutate/populate the shared tracker (or otherwise unify the two paths) and test the direct schema-then-dictionary-batch API.

src/query/src/dummy_catalog.rs
waynexia: our rolling upgrade operator always upgrades the frontend cluster first, so no need to bother

fengjiachun: The current GreptimeDB Operator does the opposite: its deployer order is Meta -> Datanode -> Frontend, and reconciliation returns as soon as a deployer is not ready ([setup](https://github.com/GreptimeTeam/greptimedb-operator/blob/387b52b9170af07b8b15f32a370953ddcb1e94fc/controllers/greptimedbcluster/controller.go#L78-L84), [sync loop](https://github.com/GreptimeTeam/greptimedb-operator/blob/387b52b9170af07b8b15f32a370953ddcb1e94fc/controllers/greptimedbcluster/controller.go#L205-L228)). Therefore the supported rollout does create a new-datanode/old-frontend window, where the old frontend rejects dictionary Flight batches as `none message`. This compatibility issue still needs to be addressed.

discord9: Agreed. The Operator source confirms the supported rollout is Meta → Datanode → Frontend, so this is not a hypothetical upgrade order: it creates exactly the new-datanode/old-frontend window described above. The old frontend rejects dictionary IPC messages before it can consume the record batch, so the current head is not rolling-upgrade compatible. Please gate dictionary output on an explicit client capability/version (or keep producer-side decoding for legacy clients) and add a new-server/old-client compatibility test before resolving this thread.

discord9 (approved):

src/query/src/dummy_catalog.rs
discord9: Withdrawing my blocker assessment here. I was assuming the cluster upgrade contract requires query availability throughout the rollout. If the supported upgrade procedure already treats the cluster as unavailable during that window, the mixed-version query failure is not a regression this PR needs to solve. Sorry for the noise.

src/mito2/src/read/flat_projection.rs
waynexia: fixed

PR #8542, opened 2026-07-16
test: cover MySQL out-of-range timestamps

discord9:

## What's changed and what's your intention?\n\n- Adds MySQL text-protocol regression coverage for a non-nullable `TimestampSecond(i64::MAX)` value that Chrono cannot convert.\n- Accepts an explicit MySQL server error or a non-null text representation, while rejecting a successful SQL NULL result.\n- Adds independently selectable in-range and Arrow-NULL controls.\n- The out-of-range target intentionally fails on the baseline because it is emitted as SQL NULL.\n\n## PR Checklist\n- [ ] I have written the necessary rustdoc comments.\n- [x] I have added the necessary unit tests and integration tests.\n- [ ] This PR requires documentation updates.\n- [x] API changes are backward compatible.\n- [x] Schema or data changes are backward compatible.

### DIFF
--- a/src/servers/tests/mysql/mysql_server_test.rs
+++ b/src/servers/tests/mysql/mysql_server_test.rs
@@ -493,7 +493,7 @@ fn timestamp_table(
     )]));
     let recordbatch = RecordBatch::new(
         schema.clone(),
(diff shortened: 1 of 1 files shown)


discord9:
Closing this draft because it was opened prematurely during local regression-test validation.

PR #8544, opened 2026-07-16
feat: update flow windows after metric batch flush

v0y4g3r:
## Summary

- Add post-flush dirty-window notifications for metric writes through `PendingRowsBatcher`, so batching-mode flows are updated when Prometheus remote write uses the pending-row batcher path.
- Notifications fire only after the entire physical flush succeeds; failed flushes (including partial region failures) send no notification. Notification errors are best-effort and never affect the persisted write result.
- Optimize timestamp extraction with a null-free fast path (`extend_from_slice(values())`) and a nullable fallback (`iter().flatten()`).
- Refactor notifier to take ownership of `Vec<TableBatch>`, eliminating the per-flush vector clone.
- Key `PendingBatch` by logical `TableId` to prevent same-name drop/recreate merging within one flush interval.

## Known follow-ups (from self-review)

- Per-peer `Vec<i64>` clone in notifier could use `Arc<Vec<i64>>`; intentionally deferred as the cost is negligible for typical peer counts.
- No integration/sqlness test for end-to-end flow notification through the batcher.
- Cache lookup `Err` path and multi-table flush path lack dedicated unit tests.

## Test Plan

- [x] `cargo nextest run -p servers --lib --retries 0` — 293 tests pass
- [x] `cargo fmt --all -- --check`
- [x] `cargo check -p servers -p frontend -p operator -p tests-integration`
- [x] `git diff --check`

### DIFF
--- a/src/servers/src/pending_rows_batcher.rs
+++ b/src/servers/src/pending_rows_batcher.rs
@@ -37,6 +37,7 @@ use common_telemetry::tracing_context::TracingContext;
 use common_telemetry::{debug, error, warn};
 use dashmap::DashMap;
 use dashmap::mapref::entry::Entry;
+use futures::StreamExt;
 use metric_engine::batch_modifier::{TagColumnInfo, modify_batch_sparse};
 use partition::manager::PartitionRuleManagerRef;
 use partition::partition::PartitionRuleRef;
@@ -64,6 +65,7 @@ const PHYSICAL_TABLE_KEY: &str = "physical_table";
 const PENDING_ROWS_BATCH_SYNC_ENV: &str = "PENDING_ROWS_BATCH_SYNC";
 const WORKER_IDLE_TIMEOUT_MULTIPLIER: u32 = 3;
 const PHYSICAL_REGION_ESSENTIAL_COLUMN_COUNT: usize = 3;
+const MAX_CONCURRENT_FLOW_NOTIFICATIONS: usize = 8;
 
 #[async_trait]
 pub trait PendingRowsSchemaAlterer: Send + Sync {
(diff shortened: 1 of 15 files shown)


src/servers/src/pending_rows_batcher.rs
```diff
@@ -1227,39 +1288,114 @@ async fn flush_batch(
...
+                        })
+                        .await
+                    {
+                        error!(e; "Failed to mark timestamps as dirty, table: {}", table_id);
+                    }
+                });
+            }
+        }
+    });
 }
```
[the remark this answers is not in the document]
v0y4g3r: Fixed in 8625eb92ed. `table_batches` is already deduplicated by `TableId` when draining `PendingBatch`, so no additional grouping is needed. The notifier now resolves table-to-flownode mappings with bounded concurrency and feeds a separate bounded notification stage, avoiding head-of-line blocking without creating unbounded per-table tasks or peer RPC fan-out. Added a regression test that blocks the first table cache lookup and verifies a later cached table is still notified.

src/servers/src/pending_rows_batcher.rs
```diff
@@ -1227,39 +1335,138 @@ async fn flush_batch(
...
+fn extract_timestamps(table_batch: &TableBatch) -> Vec<i64> {
+    let mut timestamps = Vec::with_capacity(table_batch.row_count);
+    for batch in &table_batch.batches {
+        let timestamp_column = batch.batch.column(batch.timestamp_index);
+        let Some((timestamp_values, _)) =
+            datatypes::timestamp::timestamp_array_to_primitive(timestamp_column)
+        else {
+            error!("Failed to extract timestamps from record batch");
+            continue;
+        };
```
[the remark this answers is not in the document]
v0y4g3r: Fixed in e52f0ed646. The timestamp extraction failure log now includes the logical `table_id` and the validated `timestamp_index`, so an invalid timestamp array can be traced back to the affected table and column.

src/servers/src/pending_rows_batcher.rs
```diff
@@ -1227,39 +1335,138 @@ async fn flush_batch(
...
+            .handle_mark_window_dirty(DirtyWindowRequests {
+                requests: vec![DirtyWindowRequest {
+                    table_id,
+                    timestamps: timestamps.clone(),
+                }],
+            })
+            .await
+        {
+            error!(e; "Failed to mark timestamps as dirty, table: {}", table_id);
+        }
```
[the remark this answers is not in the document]
v0y4g3r: Fixed in ccef9b0705. The dirty-window notification error now records `table_id`, `peer_id`, and `peer_addr`, making partial failures attributable to the exact flownode.

v0y4g3r:
@gemini review

src/servers/src/pending_rows_batcher.rs
```diff
@@ -1227,39 +1335,147 @@ async fn flush_batch(
...
+    }
+}
+
+fn notify_flow_dirty_windows_after_flush(
+    table_batches: Vec<TableBatch>,
+    table_flownode_set_cache: TableFlownodeSetCacheRef,
+    node_manager: NodeManagerRef,
+) {
+    common_runtime::spawn_global(async move {
```
discord9, line 1499: Could we bound notification work across flushes? This starts one detached task for every successful flush, while both `for_each_concurrent(MAX_CONCURRENT_FLOW_NOTIFICATIONS, ...)` calls only limit concurrency inside that individual task. The flush permit is released once `flush_batch_with_managers` returns, so slow cache lookups or flownode RPCs allow later flushes to keep spawning tasks that retain `table_batches` and timestamp vectors. Under sustained ingestion, that backlog can grow without bound and cause memory pressure/OOM plus increasingly delayed notifications. A shared bounded queue or semaphore, preferably with per-table coalescing, would provide a global limit; since delivery is best-effort, explicitly dropping and logging when saturated would also be safer than retaining unlimited work.

v0y4g3r: Addressed in c78d0863aa and 610fbad09a. The first commit compacts dense per-batch timestamps into `time_ranges` with a raw-timestamp fallback for sparse/overflow cases. The second replaces per-flush detached notification work with one batcher-wide bounded queue (default capacity 1024), globally limits processing to 8 table notifications, and logs/counts best-effort drops when the queue is full or closed.

v0y4g3r: This remains addressed by 610fbad09a: all flushes share one bounded queue (default capacity 1024), processing is globally limited to 8 table notifications, and full/closed drops are logged and counted. Follow-up 5ece40bf77 only removed timestamp-to-range conversion and now queues raw timestamps; it did not change the global queue bound.

src/servers/src/pending_rows_batcher.rs
```diff
@@ -1227,39 +1335,147 @@ async fn flush_batch(
...
+                        timestamps,
+                        peers,
+                        node_manager.clone(),
+                    )
+                },
+            );
+
+        futures::future::join(resolve_flownodes, notify_flownodes).await;
+    });
 }
```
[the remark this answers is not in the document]
v0y4g3r: No code change needed: this thread is outdated and the referenced per-flush two-loop/channel implementation no longer exists. Commit 610fbad09a replaced it with one batcher-scoped bounded receiver stream and a single `for_each_concurrent` worker, so the suggested pipeline rewrite no longer applies.

src/servers/src/pending_rows_batcher.rs
```diff
@@ -1227,39 +1341,181 @@ async fn flush_batch(
...
+                tx.max_capacity()
+            );
+            false
+        }
+    }
+}
+
+fn enqueue_flow_notifications(table_batches: Vec<TableBatch>, tx: &mpsc::Sender<FlowNotification>) {
+    for table_batch in table_batches {
+        let timestamps = extract_timestamps(&table_batch);
```
discord9, line 1440: Non-blocking: could we reserve queue capacity before extracting timestamps? The queue is bounded by item count, but each item owns a potentially large `Vec<i64>`; with the defaults, 1,024 notifications at 100k timestamps each retain about 781 MiB, and `max_batch_rows` is only a soft threshold. Using `try_reserve()` before `extract_timestamps`, plus chunking or a byte budget for oversized items, would make the memory bound tighter. This can be a follow-up.

discord9 (approved):

src/servers/src/pending_rows_batcher.rs
v0y4g3r: Noted and intentionally deferred per your own non-blocking/follow-up marking. Current implementation bounds the queue by item count (default 1024) and drops/logs when full, so the global backlog is bounded. Extract-before-reserve keeps the hot path simple and avoids holding a reserved slot while doing O(rows) extraction work. Chunking or a byte budget can be explored later if profiling shows the count-based bound is insufficient.

PR #8547, opened 2026-07-17
feat: add a region hook for gc cleanup

sunng87:

This patch is to add a extension point to datanode GC procedure to include some custom region gc logic.

## PR Checklist

### DIFF
--- a/src/mito2/src/engine/region_hook.rs
+++ b/src/mito2/src/engine/region_hook.rs
@@ -391,21 +391,28 @@ pub trait RegionHook: Send + Sync + Debug {
     /// Called after the datanode's global GC worker (`LocalGcWorker`) finishes a
     /// GC pass for a region — live regions (periodic GC) or dropped/repartitioned
     /// regions (the global reclamation path). Lets extensions with sidecar files
-    /// outside mito2's region dir clean up their own residual files.
+    /// outside mito2's region dir clean up residual files.
     ///
-    /// Authorization is scoped to stay safe against cross-region / tmp refs mito
-    /// honors but an extension can't see:
-    /// - [`RegionGcInfo::terminal`] `false`: clean only sidecar artifacts for the
-    ///   files in [`RegionGcInfo::removed_files`] (plus orphans on a full-listing
-    ///   pass). [`RegionGcInfo::is_region_dropped`] is context, not authorization.
-    /// - [`RegionGcInfo::terminal`] `true`: all refs released; may remove the
-    ///   whole sidecar dir.
+    /// Always scoped to [`RegionGcInfo::removed_files`]: clean only sidecar
+    /// artifacts for the files mito deleted this pass. On a
+    /// [`RegionGcInfo::full_file_listing`] pass you may also reconcile
+    /// sidecar-only orphans (e.g. detect a fully-reaped region by cross-checking
+    /// the region dir against your own manifest). This callback never authorizes
+    /// blind whole-directory removal — derive "fully reaped" from your own state
+    /// so a stale mito snapshot can't cause accidental deletion.
+    /// [`RegionGcInfo::is_region_dropped`] is context, not authorization.
     ///
-    /// Returns `Err` to keep the region un-acknowledged (`need_retry_regions`)
-    /// so the next pass replays this callback.
+    /// # Retry
+    ///
+    /// Returning `Err` keeps the region un-acknowledged (`need_retry_regions`)
+    /// for a future replay. **Dropped/repartitioned**: guaranteed — metasrv keeps
+    /// the `table_repart` tombstone until `Ok`, so full-listing passes continue
+    /// until cleanup finishes. **Live**: best-effort — not expedited through
+    /// candidate selection, and a later full-listing pass may not reconstruct the
+    /// same `removed_files`; treat live cleanup as opportunistic.
     ///
     /// `region_metadata` is `None` for dropped regions; use `region_id` +
-    /// `access_layer`. Runs on the background GC task; must be idempotent.
+    /// `access_layer`. Idempotent; runs on the background GC task.
     async fn on_region_gc(
         &self,
         region_id: RegionId,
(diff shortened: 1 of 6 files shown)


src/mito2/src/engine/region_hook.rs
```diff
@@ -382,6 +386,44 @@ pub trait RegionHook: Send + Sync + Debug {
...
+    /// extensions that keep sidecar files outside mito2's region dir — e.g. an
+    /// Iceberg manifest tree under a separate warehouse root — to clean up their
+    /// own residual files: the whole sidecar dir for a dropped region, or stale
+    /// files for a live one.
+    ///
+    /// `region_metadata` is `None` for dropped regions; locate files from
+    /// `region_id` + `access_layer` instead. Runs on the background GC task, so
+    /// it may do non-trivial I/O, and must be idempotent (global GC re-runs on a
+    /// cooldown).
+    async fn on_region_gc(
```
discord9, line 416: Could this hook return a `Result` and participate in GC completion semantics?

For a dropped region, a transient sidecar cleanup failure is currently invisible to `LocalGcWorker`, so the region is reported as processed and metasrv may remove the `table_repart` tombstone—the only trigger for future cleanup. For live regions, the removed-file state is also updated before this callback, so the cleanup may not be replayed.

We need to preserve a durable retry trigger and avoid acknowledging the region as fully processed until extension cleanup succeeds.

src/mito2/src/engine/region_hook.rs
```diff
@@ -382,6 +386,44 @@ pub trait RegionHook: Send + Sync + Debug {
...
         let _ = (region_id, region_metadata);
     }
+
+    /// Called after the datanode's global GC worker (`LocalGcWorker`) finishes a
+    /// GC pass for a region — for both live regions (periodic GC) and dropped /
+    /// repartitioned regions (the global reclamation path, where
+    /// [`RegionGcInfo::is_region_dropped`] is `true`). This is the hook point for
+    /// extensions that keep sidecar files outside mito2's region dir — e.g. an
+    /// Iceberg manifest tree under a separate warehouse root — to clean up their
+    /// own residual files: the whole sidecar dir for a dropped region, or stale
```
discord9: I don’t think `is_region_dropped` is sufficient authorization to remove the whole sidecar directory. It is currently derived from a previously captured `Option<MitoRegionRef>`, so it means ‘absent from the active map at lookup time,’ not ‘all artifacts are now unreferenced.’

In particular, a repartitioned source may still have SSTs protected by cross-region/tmp references. Mito respects those references, while an extension following this contract could delete sidecar artifacts still required by the destination. Open/drop races can also make the captured state stale.

Could we expose an authoritative terminal-cleanup signal only after all references are released, and limit earlier callbacks to artifacts corresponding to `removed_files`?

src/mito2/src/gc.rs
```diff
@@ -539,6 +545,29 @@ impl LocalGcWorker {
                 .await?;
         }
 
+        // Notify extensions (e.g. an Iceberg manifest tree under a separate
+        // warehouse root) that GC reclaimed these files, so they can clean up
+        // their own residual sidecar files. Skip the invocation when there is
+        // nothing to report — a live region's periodic pass that deleted no
+        // files — to avoid unnecessary hook overhead/I/O. A dropped region
+        // always notifies (the reclamation path extensions act on).
+        if let Some(hook) = &self.hook
```
discord9: Should a full-listing pass invoke the hook even when Mito itself deleted no files? An extension may have sidecar-only orphans from failed writes or previous cleanup failures. With the current condition, it never gets a chance to scan them.

Perhaps the condition should include `self.full_file_listing`, or full-pass notification should be a separate callback.

discord9 (changes requested):
The new extension point needs stronger completion and lifecycle semantics before it is safe for sidecar cleanup. In particular, callback failures currently cannot retain a retry trigger, and `is_region_dropped` is not sufficient authorization for terminal directory removal. Also, DCO is currently failing because the commits lack Signed-off-by trailers.

src/mito2/src/engine/region_hook.rs
discord9: Thanks, this addresses the dropped-region tombstone path and defers live manifest cleanup. One edge case still seems unresolved: a failed live-region cleanup is only added to `need_retry_regions` in this report, while `BatchGcProcedure` still completes and future live-region selection goes through the normal size/cooldown/top-N filters. Since the Mito files are physically deleted before the hook runs, a later full-listing pass may also be unable to reconstruct the same `removed_files` payload.

Could we either guarantee that failed live regions bypass normal candidate filtering and retain the failed payload until acknowledgement, or narrow/document the retry guarantee to dropped-region cleanup? This may be follow-up territory if live cleanup is intentionally best-effort, but the current contract says the callback will be replayed on the next pass.

src/mito2/src/engine/region_hook.rs
discord9: Thanks for adding `terminal`; separating context from authorization is the right direction. However, the authorization still appears to come from point-in-time snapshots: `is_region_dropped` is based on a region-map lookup captured when the worker is created, and `terminal` only checks that captured state plus the supplied `FileRefsManifest`. There is no lifecycle/generation fence or revalidation immediately before the callback, so a concurrent reopen/recreate—or a reference not represented in that snapshot—can still produce `terminal = true` while the sidecar is in use.

Because extensions are explicitly allowed to delete the whole sidecar directory when this bit is true, this retains a real accidental-deletion risk. I think terminal cleanup needs authoritative fenced dropped-state/ref authorization, revalidated immediately before invoking the hook; otherwise this should remain scoped to `removed_files` only.

discord9 (approved):
The accidental-deletion risk is resolved by removing terminal whole-directory authorization and scoping cleanup to removed files. Dropped/repartitioned cleanup retains durable retry semantics; live cleanup is now explicitly documented as best-effort. The remaining retry edge case is non-blocking.

PR #8550, opened 2026-07-17
feat: allow unknown PluginOptions with a warning message

sunng87:

Before this patch, any unknown PluginOption from config file will lead to a hard error on greptimedb startup. This makes sharing configuration between oss and other version a little bit annoying. 

This patch just prints warning message for unknown `PluginOption` and let it pass.

## PR Checklist

### DIFF
--- a/src/cmd/src/options.rs
+++ b/src/cmd/src/options.rs
@@ -14,9 +14,12 @@
 
 use clap::Parser;
 use common_config::Configurable;
+use common_options::plugin_options::PluginOptionsDeserializer;
 use common_runtime::global::RuntimeOptions;
 use plugins::PluginOptions;
+use plugins::options::PluginOptionsDeserializerImpl;
 use serde::{Deserialize, Serialize};
+use serde_json::Value;
 
 #[derive(Parser, Default, Debug, Clone)]
 pub struct GlobalOptions {
@@ -42,7 +45,12 @@ pub struct GreptimeOptions<T> {
     /// The runtime options.
     pub runtime: RuntimeOptions,
     /// The plugin options.
-    #[serde(default, deserialize_with = "plugins::deserialize_plugin_options")]
+    ///
+    /// Deserialized leniently via [`deserialize_plugin_options`]: plugin
+    /// options not recognized by the current build are dropped (with a deferred
+    /// warning) instead of aborting startup, while malformed payloads for known
+    /// variants still error.
+    #[serde(default, deserialize_with = "deserialize_plugin_options")]
     pub plugins: Vec<PluginOptions>,
 
     /// The options of each component (like Datanode or Standalone) of GreptimeDB.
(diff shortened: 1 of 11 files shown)


src/cmd/src/options.rs
```diff
@@ -42,6 +42,7 @@ pub struct GreptimeOptions<T> {
     /// The runtime options.
     pub runtime: RuntimeOptions,
     /// The plugin options.
+    #[serde(default, deserialize_with = "plugins::deserialize_plugin_options")]
```
shuiyisong: This introduces a new required root symbol on the replaceable `plugins` crate. The current enterprise workspace maps `plugins` to `ent-plugins`, which exports `PluginOptions` but not `deserialize_plugin_options`, so advancing its GreptimeDB submodule to this commit will fail to compile. Please land the matching enterprise export/implementation before advancing the submodule, or keep the helper in a shared crate used by both implementations.

src/plugins/src/options.rs
```diff
@@ -41,7 +43,78 @@ impl PluginOptionsDeserializer<Vec<PluginOptions>> for PluginOptionsDeserializer
...
+/// unrecognized ones with a warning.
+///
+/// An unrecognized option usually belongs to a plugin that is not compiled into
+/// the current build (for example, an enterprise plugin option seen by an
+/// open-source build). Rather than aborting startup, the unknown option is
+/// ignored with a warning so a shared config file can be used across builds.
+fn filter_known_plugin_options(values: Vec<Value>) -> Vec<PluginOptions> {
+    values
+        .into_iter()
+        .filter_map(|value| match PluginOptions::deserialize(&value) {
```
shuiyisong: `PluginOptions::deserialize` can fail because the enum tag is unknown or because a known variant's payload is malformed, but this branch maps both cases to `None`. For example, `[{"Dummy": 1}]` now succeeds as an empty list; with real plugin variants, a wrong field or type can silently disable the intended plugin. Please skip only tags absent from this build and propagate payload errors for recognized variants, with a regression test.

src/plugins/src/options.rs
```diff
@@ -41,7 +43,78 @@ impl PluginOptionsDeserializer<Vec<PluginOptions>> for PluginOptionsDeserializer
...
+            Ok(options) => Some(options),
+            Err(err) => {
+                warn!(
```
shuiyisong: On the normal file-based startup path, this runs during `load_options`, before the server's `build` method initializes global logging. This tracing event therefore has no subscriber and the option is dropped silently despite the promised warning. Please use an early startup diagnostic or defer the warning until after logging initialization; the current test only checks the returned list.

src/plugins/src/options.rs
```diff
@@ -41,7 +43,78 @@ impl PluginOptionsDeserializer<Vec<PluginOptions>> for PluginOptionsDeserializer
...
+                    "Ignoring unrecognized plugin option in the config: {err}. \
+                     This usually means the option belongs to a plugin that is not \
+                     compiled into this build. Raw value: {value}"
```
shuiyisong: `value` is the complete arbitrary plugin configuration. Logging it verbatim can expose tokens, passwords, or database DSNs, especially on the frontend meta-config path where logging is already initialized. Please log only the unknown variant/tag and a safe error, not the raw payload.

shuiyisong (commented):
Review findings validated against the live PR head and final diff.

shuiyisong (approved):
LGTM 🚀

sunng87:
Let me test from downstream project again to ensure it will work.

PR #8552, opened 2026-07-17
fix: enforce table-aware permissions across query and ingest protocols

shuiyisong:

This pull request fixes table-level authorization gaps where several query and ingestion paths checked the operation type without reliably passing every concrete catalog, schema, and table target to ACL-aware permission checkers.

### What changes

- Adds resolved and unresolved table-target states. Ambiguous targets fail closed, while target-independent permission checkers retain their existing behavior.
- Extracts every statically referenced SQL, TQL, and PromQL table, including sources, destinations, CTEs, rename targets, and cross-schema selectors.
- Applies table authorization to Prometheus queries and metadata, remote read/write, OTLP, OpenTSDB, InfluxDB, and log ingestion.
- Rejects internal physical metric tables as user query targets and filters inaccessible logical metrics from metadata enumeration.
- Preflights every normalized batch before writing anything, preventing an unauthorized later batch from leaving partial data behind.
- Deduplicates targets, bounds concurrent catalog resolution, and avoids redundant metadata lookups.

### Behavior examples

- `INSERT INTO dst SELECT * FROM src` now requires access to both `src` and `dst`.
- PromQL metric discovery resolves logical metric candidates before execution; physical or unsafe unresolved targets are rejected.
- A remote-write, OTLP, or log request spanning multiple tables or schemas authorizes every final target before the first batch is persisted.

### Compatibility

Custom implementations of `PermissionChecker`, `PromStoreProtocolHandler`, `PipelineHandler`, or `OpentsdbProtocolHandler` must implement the new target-aware or preflight methods. This PR does not change configuration, persisted-data, or wire formats.

## PR Checklist

### DIFF
--- a/src/frontend/src/instance.rs
+++ b/src/frontend/src/instance.rs
@@ -70,18 +70,20 @@ use prometheus::HistogramTimer;
 use promql_parser::label::Matcher;
 use query::QueryEngineRef;
 use query::metrics::OnDone;
-use query::parser::{PromQuery, QueryLanguageParser, QueryStatement};
+use query::parser::{PromQuery, QueryStatement};
 use query::query_engine::DescribeResult;
 use query::query_engine::options::{QueryOptions, validate_catalog_and_schema};
 use servers::error::{
     self as server_error, AuthSnafu, CommonMetaSnafu, ExecuteQuerySnafu,
-    OtlpMetricModeIncompatibleSnafu, ParsePromQLSnafu, UnexpectedResultSnafu,
+    OtlpMetricModeIncompatibleSnafu, UnexpectedResultSnafu,
 };
 use servers::interceptor::{
     PromQueryInterceptor, PromQueryInterceptorRef, SqlQueryInterceptor, SqlQueryInterceptorRef,
 };
 use servers::otlp::metrics::legacy_normalize_otlp_name;
-use servers::prometheus_handler::{PrometheusHandler, resolve_schema_from_matchers};
+use servers::prometheus_handler::{
+    ParsedPromQuery, PrometheusHandler, resolve_schema_from_matchers,
+};
 use servers::query_handler::sql::SqlQueryHandler;
 use session::context::{Channel, QueryContextRef};
 use session::table_name::table_idents_to_full_name;
@@ -1082,19 +1084,14 @@ impl Instance {
 
     fn prom_queries_permission_targets(
         &self,
-        queries: &[PromQuery],
+        queries: &[ParsedPromQuery],
         query_ctx: &QueryContextRef,
     ) -> server_error::Result<PermissionTableTargets> {
         let mut targets = Vec::new();
         let mut resolved = true;
 
         for query in queries {
-            let stmt = QueryLanguageParser::parse_promql(query, query_ctx).with_context(|_| {
-                ParsePromQLSnafu {
-                    query: query.clone(),
-                }
-            })?;
-            let QueryStatement::Promql(eval_stmt, _) = stmt else {
+            let QueryStatement::Promql(eval_stmt, _) = query.statement() else {
                 unreachable!("query is parsed from promql");
             };
 
@@ -1123,27 +1120,32 @@ impl PrometheusHandler for Instance {
         query: &PromQuery,
         query_ctx: QueryContextRef,
     ) -> server_error::Result<Output> {
+        let query = ParsedPromQuery::parse(query.clone(), &query_ctx)?;
+        self.do_query_parsed(query, query_ctx).await
+    }
+
+    #[tracing::instrument(skip_all)]
+    async fn do_query_parsed(
+        &self,
+        query: ParsedPromQuery,
+        query_ctx: QueryContextRef,
+    ) -> server_error::Result<Output> {
         let interceptor = self
             .plugins
             .get::<PromQueryInterceptorRef<server_error::Error>>();
 
         self.check_prom_query_privilege(&query_ctx)?;
 
-        let stmt = QueryLanguageParser::parse_promql(query, &query_ctx).with_context(|_| {
-            ParsePromQLSnafu {
-                query: query.clone(),
-            }
-        })?;
+        let targets =
+            self.prom_queries_permission_targets(std::slice::from_ref(&query), &query_ctx)?;
+        self.check_query_target_permission(targets, &query_ctx)
+            .await?;
+
+        let (query, stmt) = query.into_parts();
 
         let QueryStatement::Promql(eval_stmt, _) = &stmt else {
             unreachable!("query is parsed from promql");
         };
-        let targets = match self.prom_expr_permission_targets(&eval_stmt.expr, &query_ctx)? {
-            Some(targets) => PermissionTableTargets::resolved(targets),
-            None => PermissionTableTargets::Unresolved,
-        };
-        self.check_query_target_permission(targets, &query_ctx)
-            .await?;
 
         let plan = self
             .statement_executor
@@ -1152,7 +1154,7 @@ impl PrometheusHandler for Instance {
             .map_err(BoxedError::new)
             .context(ExecuteQuerySnafu)?;
 
-        interceptor.pre_execute(query, &eval_stmt.expr, Some(&plan), query_ctx.clone())?;
+        interceptor.pre_execute(&query, &eval_stmt.expr, Some(&plan), query_ctx.clone())?;
 
         // Take the EvalStmt from the original QueryStatement and use it to create the CatalogQueryStatement.
         let query_statement = if let QueryStatement::Promql(eval_stmt, alias) = stmt {
@@ -1215,6 +1217,20 @@ impl PrometheusHandler for Instance {
     async fn check_query_permission(
         &self,
         queries: &[PromQuery],
+        query_ctx: &QueryContextRef,
+    ) -> server_error::Result<()> {
+        let queries = queries
+            .iter()
+            .cloned()
+            .map(|query| ParsedPromQuery::parse(query, query_ctx))
+            .collect::<server_error::Result<Vec<_>>>()?;
+        self.check_query_permission_parsed(&queries, query_ctx)
+            .await
+    }
+
+    async fn check_query_permission_parsed(
+        &self,
+        queries: &[ParsedPromQuery],
         query_ctx: &QueryContextRef,
     ) -> server_error::Result<()> {
         self.check_prom_query_privilege(query_ctx)?;
--- a/src/frontend/src/instance.rs
+++ b/src/frontend/src/instance.rs
@@ -26,6 +26,7 @@ mod promql;
 mod region_query;
 pub mod standalone;
 
+use std::collections::HashSet;
 use std::pin::Pin;
 use std::sync::atomic::AtomicBool;
 use std::sync::{Arc, atomic};
@@ -58,7 +59,7 @@ use common_telemetry::logging::SlowQueryOptions;
 use common_telemetry::{debug, error, tracing};
 use dashmap::DashMap;
 use datafusion_expr::LogicalPlan;
-use futures::{Stream, StreamExt};
+use futures::{Stream, StreamExt, future};
 use lazy_static::lazy_static;
 use operator::delete::DeleterRef;
 use operator::insert::InserterRef;
@@ -1013,28 +1014,66 @@ impl Instance {
             .transpose()
     }
 
+    async fn is_physical_query_permission_target(
+        &self,
+        target: &PermissionTableTarget,
+        query_ctx: &QueryContextRef,
+    ) -> server_error::Result<bool> {
+        self.catalog_manager
+            .table(
+                &target.catalog,
+                &target.schema,
+                &target.table,
+                Some(query_ctx),
+            )
+            .await
+            .map(|table| table.is_some_and(|table| table.table_info().is_physical_table()))
+            .map_err(BoxedError::new)
+            .context(ExecuteQuerySnafu)
+    }
+
     async fn resolve_query_permission_targets(
         &self,
         targets: PermissionTableTargets,
         query_ctx: &QueryContextRef,
     ) -> server_error::Result<PermissionTableTargets> {
-        let PermissionTableTargets::Resolved(targets) = targets else {
+        const CONCURRENCY: usize = 8;
+
+        let checker = self.plugins.get::<PermissionCheckerRef>();
+        if !checker.as_ref().uses_table_targets() {
+            return Ok(targets);
+        }
+
+        let PermissionTableTargets::Resolved(mut targets) = targets else {
             return Ok(PermissionTableTargets::Unresolved);
         };
-        for target in &targets {
-            let table = self
-                .catalog_manager
-                .table(
-                    &target.catalog,
-                    &target.schema,
-                    &target.table,
-                    Some(query_ctx),
-                )
-                .await
-                .map_err(BoxedError::new)
-                .context(ExecuteQuerySnafu)?;
-            if table.is_some_and(|table| table.table_info().is_physical_table()) {
-                return Ok(PermissionTableTargets::Unresolved);
+        if targets.len() > 1 {
+            let mut seen = HashSet::with_capacity(targets.len());
+            targets.retain(|target| seen.insert(target.clone()));
+        }
+        if let [target] = targets.as_slice() {
+            return if self
+                .is_physical_query_permission_target(target, query_ctx)
+                .await?
+            {
+                Ok(PermissionTableTargets::Unresolved)
+            } else {
+                Ok(PermissionTableTargets::resolved(targets))
+            };
+        }
+
+        // Bound catalog load and inspect results in target order to preserve serial semantics.
+        for chunk in targets.chunks(CONCURRENCY) {
+            let results = future::join_all(
+                chunk
+                    .iter()
+                    .map(|target| self.is_physical_query_permission_target(target, query_ctx)),
+            )
+            .await;
+            for result in results {
+                if result? {
+                    return Ok(PermissionTableTargets::Unresolved);
+                }
             }
         }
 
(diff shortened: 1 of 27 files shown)


sunng87 (commented):
## Review — table-aware permissions across query/ingest protocols

Overall this is a well-structured change. The core abstraction (`PermissionTableTarget` / `PermissionTableTargets` with fail-closed `Unresolved`, the `uses_table_targets()` opt-out so target-independent checkers skip catalog resolution, and the `fork()`-based context isolation) is clean and the test coverage is genuinely good. Check/clippy/compat are green. Below are findings worth addressing before merge, ordered by impact.

---

### 🔴 High impact — Native gRPC write/DDL paths are not table-checked

The PR headline is "enforce table-aware permissions across … ingest protocols," but the **primary SDK write path is not covered.** In `src/frontend/src/instance/grpc.rs`, `GrpcQueryHandler::do_query` only runs the operation-level check and then dispatches to handlers that have no target-aware check:

```rust
self.plugins.get::<PermissionCheckerRef>().as_ref()
    .check_permission_with_context(ctx.current_user(), PermissionReq::GrpcRequest(&request), …)?;
let output = match request {
    Request::Inserts(requests)   => self.handle_inserts(requests, ctx.clone()).await?,
    Request::RowInserts(requests)=> self.handle_row_inserts(…).await?,
    Request::Deletes(requests)   => self.handle_deletes(requests, ctx.clone()).await?,
    Request::RowDeletes(requests)=> self.handle_row_deletes(…).await?,
    Request::Ddl(expr)           => { … create/alter/drop/truncate … }
    Query::InsertIntoPlan(insert)=> self.handle_insert_plan(insert, ctx).await?,
    Query::LogicalPlan(plan)     => self.do_exec_plan_inner(…).await?,
};
```

`handle_inserts` / `handle_row_inserts` / `handle_deletes` / `handle_row_deletes` / `handle_insert_plan` contain no `check_row_insert_permission` / `check_table_permission` call, and `PermissionReq::GrpcRequest` is mode-only. **Impact:** a user with write access to one table/schema can, via gRPC, write to / delete from / drop / alter **any** `catalog.schema.table` embedded in the request. gRPC `RowInserts` is the high-throughput path used by the Python/Go/Java SDKs.

Flight bulk-insert is partially OK — it passes `PermissionReq::BulkInsert { catalog, schema, table }` — but via the **old** `check_permission`, not the new target-aware path, so it depends entirely on whether the deployed checker special-cases `BulkInsert`.

**Recommendation:** Either (a) extend target-aware checks to `Request::{Inserts,RowInserts,Deletes,RowDeletes,InsertIntoPlan}` and `Request::Ddl`, or (b) explicitly document the limitation in the PR description and link a follow-up issue. As written the PR title overstates the coverage — would want a maintainer answer here before approving.

---

### 🟠 Medium impact — `/api/v1/series` now hard-errors for regex / `or` selectors (regression)

`series_query` now requires a single statically-resolvable target:

```rust
let Some(target) = resolved_promql_targets(std::slice::from_ref(&prom_query), &query_ctx)
    .and_then(|targets| match targets.as_slice() {
        [target] => Some(target.clone()), _ => None,
    })
else {
    return PrometheusJsonResponse::error(StatusCode::InvalidArguments,
        "series selector must resolve exactly one metric table");
};
```

For `{__name__=~"cpu.*"}` or any selector with `or`-matchers / an unnamed vector selector, `collect_static_vector_target` increments `unresolved_selectors`, so this returns `InvalidArguments`. **Impact:** Previously the code used `retrieve_metric_name_from_promql(…).unwrap_or_default()` and still called `handler.do_query(…)`, so regex selectors **returned data**. Grafana commonly issues `/api/v1/series?match[]={__name__=~"…"}` — likely user-visible regression.

Inconsistency worth noting: `labels_query` handles the same unresolved case by **falling back** to `get_all_column_names` rather than erroring. `series_query` should fall back to result-based label extraction too; the `check_query_permission` call above already authorizes the operation, so the hard error isn't required for safety.

---

### 🟠 Medium impact — `label_values` for `__field__` fails the whole request if any table is inaccessible

In `label_values_query`, `FIELD_NAME_LABEL` with no matchers (`enumerate_all`) calls `retrieve_field_names` over **every** table in the schema, then `check_query_target_permission(current_schema_metric_targets(…, &table_names))` authorizes the full list. **Impact:** if the user lacks read on even one table, the entire field-name listing returns `PermissionDenied`. The sibling `__name__` path avoids this by calling `filter_metadata_metric_names` first. Apply the same filter-then-resolve pattern here for graceful multi-tenant behavior. (Secure as-is, just coarse.)

---

### 🟡 Low impact / nits

1. **InfluxDB double-check.** `InfluxdbLineProtocolHandler::exec` calls `check_row_insert_permission` twice — once before the aligner/post-conversion (against the un-forked `ctx`) and once after `fork()`. The first check is redundant given the post-conversion check is authoritative; if kept for defense-in-depth, fork before it for consistency with the other handlers.

2. **`extract_tables_from_statement_checked` is optimistic.** The `_ => true` arm marks unhandled variants as "complete" with empty targets, so any statement kind referencing user tables that isn't enumerated is silently under-checked rather than fail-closed. Two specifics to verify:
   - `CREATE TABLE … AS SELECT` / CTAS: `CreateTable` and `CreateExternalTable` structs have no query field, so the **source** read tables may not be extracted. Please confirm the source tables get authorized.
   - Worth a comment listing which variants intentionally produce empty targets vs. which were overlooked.

3. **Empty-target contract is implicit.** Several metadata endpoints call `check_query_permission(&[], ctx)` / `check_query_target_permission(Resolved(vec![]), …)`, relying on ACL checkers treating an empty resolved list as "operation allowed, nothing to authorize." This contract is documented but not enforced; a stricter custom checker that rejects empty targets would break every metadata endpoint. Consider a dedicated operation-only check or a test pinning the contract against the in-tree `TargetAwarePermissionChecker`.

4. **`debug_assert_eq!(outcome.samples_written, sample_count)` in `remote_write_v2`.** Release builds skip this, but the recorded Prometheus metric uses `sample_count` directly. If batching ever changed row counts, the metric would be silently wrong. Either make it a real assertion or compute the metric from `outcome`.

5. **Breaking trait changes.** `PrometheusHandler`, `PromStoreProtocolHandler`, `PipelineHandler`, `OpentsdbProtocolHandler` gained **non-defaulted** required methods, and `query_metric_names` gained a `schema` parameter. The PR body acknowledges this — just confirm the enterprise crate's implementations are updated in lockstep (`PrometheusHandler` especially, since it has the most new methods).

---

### Things that look right (worth keeping)

- `resolve_query_permission_targets` rejecting targets whose resolved table `is_physical_table()` as `Unresolved` is correct and matches the intent; `is_physical_table()` is gated on `PHYSICAL_TABLE_METADATA_KEY`, so regular mito tables are unaffected.
- OTLP trace target logic correctly gates `_services`/`_operations` on `span.service_name.is_some()`, matching `TraceAuxData::observe_span` exactly — no under/over-checking.
- The `write_all`/`insert_all` preflight-then-execute split genuinely preserves Prometheus v2 partial-progress headers (per-batch row counts accumulated before the failing batch). `test_mixed_v2_preflights_all_then_batches_only_samples` pins sample-batcher vs. histogram-direct routing.
- `prom_store::table_name`'s stricter validation (exactly one non-empty `Eq` `__name__`) is **not** a functional regression for remote read — `handle_remote_query` already required a concrete name; it just produces a clearer error now.
- `QueryContext::fork()` giving the clone an independent `mutable_session_data` `RwLock` is the right tool for "the checked schema can't change before this batch is written."

### Suggested priority
1. Resolve the gRPC native-write/DDL scope question (🔴) — affects the PR's core claim.
2. Fix the `/api/v1/series` regression (🟠) by falling back instead of erroring.
3. Filter-before-authorize for `__field__` enumeration (🟠).
4. Nits as time allows.

src/servers/src/http/prometheus.rs
```diff
@@ -1294,35 +1568,100 @@ pub async fn label_values_query(
...
+        );
+
+        let (field_columns, table_names) = if !enumerate_all && metric_names.is_empty() {
+            (HashSet::new(), Vec::new())
+        } else {
+            handle_schema_err!(
+                retrieve_field_names(&query_ctx, handler.catalog_manager(), metric_names).await
+            )
+            .unwrap_or_default()
+        };
```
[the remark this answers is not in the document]
shuiyisong: Thanks. This behavior is intentional. An unscoped __field__ request targets every non-physical table, and the table-aware permission contract rejects the whole request when any target is unauthorized. Per-result filtering is limited to __name__ logical-table metadata enumeration. Applying it here would also change established __field__ semantics because retrieve_table_names() excludes non-metric tables, while an empty filtered list makes retrieve_field_names() enumerate all tables again. Commit aff0945087 adds test_field_name_enumeration_fails_closed to pin the mixed allowed/denied logical-metric case to HTTP 403. I also verified that the exact suggested patch makes this regression test fail by returning success instead of PermissionDenied.

shuiyisong:
@sunng87 Thanks for the detailed review. Follow-up after commit 9bfdb58725:

- Native gRPC: confirmed that these request variants still use the existing operation-level GrpcRequest check. This is a real, broader scope gap covering row and column inserts, deletes, DDL, and encoded plans. I intentionally did not fold that protocol-wide change into this patch; it needs separate target extraction and behavioral coverage.
- /api/v1/series: fixed for safe non-equality metric-name selectors. It now uses the same bounded metric discovery as instant and range queries, expands to exact metric queries, and authorizes every expanded target before execution. Ambiguous or OR selector shapes remain fail-closed because their table targets cannot be resolved safely. The endpoint test pins regex expansion.
- __field__: unchanged intentionally. An unscoped field request targets all non-physical tables, so mixed authorization rejects the request rather than filtering results. The resolved inline thread contains the detailed rationale and regression test.
- InfluxDB: keeping both checks. The first blocks unauthorized targets before catalog alignment and post-conversion interceptor work; the second authorizes the final transformed requests on a forked context before writing. They protect different stages.
- SQL extraction: replaced the wildcard arm with exhaustive Statement variants, including explicit table-targetless cases, so future variants fail compilation until classified. COPY DATABASE was also classified explicitly. CTAS is not supported by the current CreateTable or CreateExternalTable AST, so there is no hidden source query to extract today.
- Empty targets: documented the contract that Resolved(empty) still checks the operation privilege but has no table ACLs, and added tests for both an allowed read operation and a rejected write operation.
- Remote write v2: success metrics and written-count headers now use outcome.samples_written and outcome.histograms_written rather than decoded request counts.

Validation on the pushed head: frontend unit tests 88 passed; auth, SQL, and servers unit tests 636 passed; SQL enterprise target-extraction test passed; targeted clippy, formatting, and diff checks passed.

src/frontend/src/instance.rs
```diff
@@ -965,6 +982,139 @@ pub fn attach_timer(output: Output, timer: HistogramTimer) -> Output {
...
+    fn prom_queries_permission_targets(
+        &self,
+        queries: &[PromQuery],
+        query_ctx: &QueryContextRef,
+    ) -> server_error::Result<PermissionTableTargets> {
+        let mut targets = Vec::new();
+        let mut resolved = true;
+
+        for query in queries {
+            let stmt = QueryLanguageParser::parse_promql(query, query_ctx).with_context(|_| {
```
sunng87: Minor perf suggestion (optional): this parses every PromQL query to extract targets, and then `do_query` parses the same string again to execute it. On the `/api/v1/series` and `labels` regex-expansion paths the same query is parsed yet again by `resolved_promql_targets` / `static_promql_targets`, and the whole thing is multiplied by N discovered metrics — so a single request can parse each metric query 3-4x.

If it's worth it, threading the already-parsed `QueryStatement`/`PromqlExpr` into an internal `check_query_permission_stmts(&[QueryStatement])` so the permission check and execution share one parse would cut this to a single parse per query. This is the highest-value optimization in the changeset.

src/servers/src/http/prometheus.rs
```diff
@@ -1195,23 +1331,161 @@ where
...
+    expr: &PromqlExpr,
+    metric_names: Vec<String>,
+) -> Vec<PromQuery> {
+    metric_names
+        .into_iter()
+        .map(|metric| {
+            let mut query = query.clone();
+            let mut expr = expr.clone();
+            update_metric_name_matcher(&mut expr, &metric);
+            query.query = expr.to_string();
```
sunng87: Minor perf suggestion (optional): for metric discovery that fans out to N metrics, this clones the full PromQL AST (`expr.clone()` above) and re-serializes it (`expr.to_string()` on this line) once per metric — that clone + round-trip is the expensive part of discovery. Pre-existing behavior, just moved into this helper.

Alternatives: do the `__name__` substitution at the string level on the original query, or carry the parsed `expr` + a name override into `do_query` instead of re-stringifying.

src/servers/src/http/prom_store.rs
```diff
@@ -370,27 +348,68 @@ struct PromWriteError {
...
+
+#[async_trait]
+impl PromWriteBatcher for PendingRowsBatcher {
+    async fn submit(&self, requests: RowInsertRequests, ctx: QueryContextRef) -> Result<u64> {
+        PendingRowsBatcher::submit(self, requests, ctx).await
+    }
+}
+
+fn into_prom_write_batches(req: ContextReq, query_ctx: QueryContextRef) -> Vec<PromWriteBatch> {
+    req.as_req_iter(query_ctx).collect()
```
sunng87, line 378: Heads-up on memory profile (not necessarily a change request): `into_prom_write_batches` now materializes the entire decoded request (`as_req_iter().collect()`) and `preflight_prometheus_rows` then walks every batch before the first write. Previously this streamed batch-by-batch. For large multi-schema remote writes, peak memory (all `RowInsertRequests` resident at once) and time-to-first-write both increase.

This is a deliberate tradeoff for partial-write safety (preflight before persist), but worth confirming it's acceptable for the largest writes — or whether a two-pass streaming preflight is feasible.

src/frontend/src/instance.rs
```diff
@@ -965,6 +982,139 @@ pub fn attach_timer(output: Output, timer: HistogramTimer) -> Output {
...
+        if !checker.as_ref().uses_table_targets() {
+            return Ok(targets);
+        }
+
+        let PermissionTableTargets::Resolved(mut targets) = targets else {
+            return Ok(PermissionTableTargets::Unresolved);
+        };
+        if targets.len() > 1 {
+            let mut seen = HashSet::with_capacity(targets.len());
+            targets.retain(|target| seen.insert(target.clone()));
```
sunng87, line 1054: Minor nit: `seen.insert(target.clone())` allocates 3 Strings per target just for deduplication. Permission checking is order-independent, so a sort-then-dedup (derive `Ord` on `PermissionTableTarget`) or a borrow-based pass would avoid the clones. Low impact in practice — few targets per query, and this whole path is skipped for the default checker via `uses_table_targets()`.

Related (same area): `ctx.current_schema()` returns an owned `String` (a lock + clone per call) while `current_catalog()` returns `&str`, so every target built via `check_row_insert_permission` / `current_schema_metric_targets` pays the schema clone. A `current_schema_str() -> &str` accessor would help these hot callers.

src/frontend/src/instance/influxdb.rs
```diff
@@ -77,6 +79,10 @@ impl InfluxdbLineProtocolHandler for Instance {
             .post_lines_conversion(requests, ctx.clone())
             .await?;
 
+        let ctx = Arc::new(ctx.fork());
+        self.check_row_insert_permission(&requests, &ctx, PermissionReq::LineProtocol)
```
sunng87, line 83: Minor perf note (you already explained the two-stage rationale, so just for awareness): running `check_row_insert_permission` twice means target construction (3 Strings x inserts) plus a checker invocation happens twice per InfluxDB write. Fine if the checker is cheap; flagging the duplicate work since the first check's targets are discarded after alignment / post-conversion.

src/frontend/src/instance/otlp.rs
```diff
@@ -65,16 +101,17 @@ impl OpenTelemetryProtocolHandler for Instance {
...
             .get::<OpenTelemetryProtocolInterceptorRef<servers::error::Error>>();
         interceptor_ref.pre_execute(ctx.clone())?;
+        let ctx = Arc::new(ctx.fork());
 
         let input_names = request
             .resource_metrics
             .iter()
             .flat_map(|r| r.scope_metrics.iter())
-            .flat_map(|s| s.metrics.iter().map(|m| &m.name))
+            .flat_map(|s| s.metrics.iter().map(|m| m.name.clone()))
```
sunng87, line 110: Minor: every metric name is cloned here because `request` is moved into `to_grpc_insert_requests` afterward (references would dangle). Necessary given the ordering, but it's new per-metric allocation on every OTLP metrics request. Since `check_otlp_legacy` needs the names before conversion, the ordering likely forces it — but worth a glance to see if `input_names` could be derived from the converted `requests`' `table_name`s instead.

sunng87 (commented):
Optional performance observations (separate from the correctness review — none of these are blockers). Inline below. The highest-value one is deduplicating the PromQL parsing on the metadata/discovery paths.

sunng87 (approved):
Added some non-blocking performance related issue

src/frontend/src/instance.rs
shuiyisong: Thanks, confirmed. `check_query_permission`, `do_query`, and the metadata helpers can parse the same generated query 3-4 times. This is a valid optimization, but sharing parsed statements crosses the `PrometheusHandler` boundary and is larger than this ACL fix. I am leaving this thread open for a measured follow-up.

src/servers/src/http/prometheus.rs
shuiyisong: Thanks. Confirmed this clone and serialization cost is real, but the instant and range paths already performed it inline before this PR; this helper centralizes the same behavior. String-level replacement is too fragile for nested selectors and escaping. I would fold any improvement into the parsed-statement follow-up above rather than expand this PR.

src/servers/src/http/prom_store.rs
shuiyisong: Thanks. The all-batch preflight intentionally delays the first write to prevent authorization or interceptor failures after partial persistence. `ContextReq` already owns the fully decoded `RowInsertRequest`s before `as_req_iter`, so the new peak is mainly retained batch and context wrapper overhead rather than a second copy of all rows. I am leaving this open until large multi-schema writes are measured; a two-pass design is only warranted if that shows a real regression.

src/frontend/src/instance.rs
shuiyisong: Thanks. The allocations are real but bounded to the target-aware path and typically small. Sort and dedup would change the preserved target and catalog-error order, while a plain `current_schema_str() -> &str` cannot safely borrow through the mutable `RwLock`; targets also require owned strings. I am keeping the current simple implementation.

src/frontend/src/instance/influxdb.rs
shuiyisong: Thanks. Confirmed. Both checks are intentional: the first authorizes targets before the timestamp aligner performs catalog reads, and the second authorizes the final requests after `post_lines_conversion` may change them. Skipping either weakens the authorization boundary, so I am keeping the duplicate small cost.

src/frontend/src/instance/otlp.rs
shuiyisong: Thanks. The converted request table names cannot replace `input_names`: legacy mode must be selected before conversion, conversion translates and fans out names based on that mode, and the cache must use raw input names only after final target authorization. Retaining raw names across the move is therefore intentional.

PR #8555, opened 2026-07-17
feat!: upgrade DataFusion to 55

### LINKED ISSUE #8463: PostgreSQL compat: SELECT array[null] returns incorrect value
### What type of bug is this?

Incorrect result

Standalone mode

### Minimal reproduce step

The PostgreSQL query `SELECT array[null]` returns on the wire a value which is different from PostgreSQL, with the incorrect oid. 

Here is the tshark trace for the same query in PostgreSQL: note that the oid is different (1009 meaning `_text` or array of text) and the content spells out `{NULL}`. 

```
PostgreSQL
    Type: Row description
    Length: 30
    Field count: 1
        Column name: array
            Table OID: 0
            Column index: 0
            Type OID: 1009
            Column length: -1
            Type modifier: -1
            Format: Text (0)
PostgreSQL
    Type: Data row
    Length: 16
    Field count: 1
        Column length: 6
        Data: 7b4e554c4c7d
PostgreSQL
    Type: Command completion
    Length: 13
    Tag: SELECT 1
```

### What did you see instead?

Below is the tshark trace for that query in GreptimeDB v1.1.2

```
PostgreSQL
    Type: Simple query
    Length: 24
    Query: select array[null];

PostgreSQL
    Type: Row description
    Length: 41
    Field count: 1
        Column name: make_array(NULL)
            Table OID: 0
            Column index: 0
            Type OID: 705
            Column length: 0
            Type modifier: 0
            Format: Text (0)
PostgreSQL
    Type: Data row
    Length: 10
    Field count: 1
        Column length: -1
PostgreSQL
    Type: Command completion
    Length: 13
    Tag: SELECT 1
```

Debian

1.1.2

```bash

```

discord9:

- DataFusion 55.0.0: [`apache/datafusion@d5552342`](https://github.com/apache/datafusion/commit/d5552342012888b7d1a3ab88d92e3d292fc0cde0)
- Global `LIMIT` across multiple partitions: [apache/datafusion#23800](https://github.com/apache/datafusion/pull/23800)
- Safe cast predicate preimages: [apache/datafusion#22906](https://github.com/apache/datafusion/pull/22906)
- Chained projection regression: [apache/datafusion#24686](https://github.com/apache/datafusion/pull/24686)
- DataFusion PostgreSQL integration: [`GreptimeTeam/datafusion-postgres@3c77e6c3`](https://github.com/GreptimeTeam/datafusion-postgres/commit/3c77e6c32b8db80635a0d2f4b318a36b31170bc3)
- Greptime SQL parser 0.62 range-query port: [`GreptimeTeam/sqlparser-rs@9e9019bb`](https://github.com/GreptimeTeam/sqlparser-rs/commit/9e9019bb1c7040ed956f654e39378dd42ab17884)
- Existing bugs reproduced during validation, tracked separately: #9150, #9151, #9152.

This PR upgrades GreptimeDB from the DataFusion 53.1 dependency line to DataFusion `55.0.0`, Arrow `59.2`, and Parquet `59.2`, directly rather than retaining an intermediate DF54 upgrade.

The upgrade adapts GreptimeDB planning and execution integrations to DataFusion 55 while retaining the behaviors that are material to GreptimeDB: distributed ordering via `EnforceSorting`, custom physical plans and expression traversal, recursive `MergeScan` fallback handling, global Top-K through `MergeSort`, table-function planning, PostgreSQL compatibility, generated range-query validation, `count_hash` compatibility, JSON field ordering, and topology-specific SQLness plan expectations.

### Dependency versions and pins

| Dependency | Revision / version | Purpose |
| --- | --- | --- |
| DataFusion | exact `55.0.0`; fork [`b66dca3260c8e04f314c8036cdfbb08f1a564d95`](https://github.com/discord9/datafusion/commit/b66dca3260c8e04f314c8036cdfbb08f1a564d95) | DataFusion 55 plus the retained GreptimeDB correctness and repair patches below |
| Arrow / Parquet | exact `59.2.0` | DataFusion 55 dependency line |
| `datafusion-orc` | released `0.10.0` | DataFusion 55 ORC integration |
| `orc-rust` | released `0.9.0` | Arrow 59-compatible ORC reader |
| `datafusion-pg-catalog` / `arrow-pg` | [`3c77e6c32b8d`](https://github.com/GreptimeTeam/datafusion-postgres/commit/3c77e6c32b8db80635a0d2f4b318a36b31170bc3) | DataFusion 55 PostgreSQL catalog and Arrow integration |
| Substrait | `0.63` at [`91ec978b`](https://github.com/GreptimeTeam/substrait-rs/commit/91ec978b0649417ad3da8390e7a515baec723b1b) | Local named-struct ordering and exact-spec tagless checkout |
| `sqlparser` | `0.62.0` at [`9e9019bb`](https://github.com/GreptimeTeam/sqlparser-rs/commit/9e9019bb1c7040ed956f654e39378dd42ab17884) | Preserves GreptimeDB range-query extensions |

All 15 DataFusion `[patch.crates-io]` entries and 35 corresponding lockfile source blocks use the same immutable fork revision shown above; no local path patches are used. The released ORC packages replace the temporary ORC git patches, and `datafusion-pg-catalog` and `arrow-pg` remain pinned together at their DataFusion 55-compatible commit.

### DataFusion thin fork: patches on top of official 55.0.0

The fork starts at official DataFusion `55.0.0` commit [`d555234201`](https://github.com/apache/datafusion/commit/d5552342012888b7d1a3ab88d92e3d292fc0cde0). The retained base patches serve these purposes:

| Topic | Thin-fork commit | Purpose |
| --- | --- | --- |
| Global multi-partition fetch / `LIMIT` | [`96625f97aa`](https://github.com/GreptimeTeam/datafusion/commit/96625f97aab176370754f695eecae2456b6fa16f) | Preserve global fetch semantics for multi-partition input; ports [apache/datafusion#23800](https://github.com/apache/datafusion/pull/23800) (upstream PR remains open) |
| Duplicate Substrait root names | [`d1dd7b9ad6`](https://github.com/GreptimeTeam/datafusion/commit/d1dd7b9ad68bf0efccdb9c2fc682e3ad67504956) | Preserve duplicate root names with deterministic internal qualifiers; retained from [GreptimeTeam/datafusion#20](https://github.com/GreptimeTeam/datafusion/pull/20) (`cee35d94c`) |
| Safe cast predicate preimages | [`da6529d542`](https://github.com/GreptimeTeam/datafusion/commit/da6529d54273f179fb7ee964a7ebf990ea360599) | Rewrite safe-cast predicates through exact or range preimages; [apache/datafusion#22906](https://github.com/apache/datafusion/pull/22906) (open), retained via [GreptimeTeam/datafusion#11](https://github.com/GreptimeTeam/datafusion/pull/11) |
| Ordered timestamp widening | [`e47063d1e4`](https://github.com/GreptimeTeam/datafusion/commit/e47063d1e4c2a8277c48eee5f9ebbfe978a74387) | Preserve ordered timestamp-widening predicate rewrites under the overflow policy below; port of [GreptimeTeam/datafusion#23](https://github.com/GreptimeTeam/datafusion/pull/23) |
| Chained logical projections | [`6922ddce3`](https://github.com/GreptimeTeam/datafusion/commit/6922ddce3c81c11a5c6dcd141b808f1b26a23899) | Restore one-layer-per-pass projection merging and prevent incorrect nested-projection results; [apache/datafusion#24686](https://github.com/apache/datafusion/pull/24686) (merged) |
| Custom float hash representations | [`dec6af133d`](https://github.com/GreptimeTeam/datafusion/commit/dec6af133d6a5e0063407c1f477d9a7b67649a29) | Preserve floating-point hash representations required by `count_hash` state compatibility. The custom-hasher API comes from merged [apache/datafusion#21820](https://github.com/apache/datafusion/pull/21820); this fork patch separately restores signed-zero representation sensitivity, not the API itself |

The scalar-planner hook restoration [`aab310aac36`](https://github.com/discord9/datafusion/commit/aab310aac36eaea11ab7414c34d1aba7bd07ba59) is retained for JSON2: [apache/datafusion#24846](https://github.com/apache/datafusion/pull/24846) (merged), with the Greptime port in [GreptimeTeam/datafusion#31](https://github.com/GreptimeTeam/datafusion/pull/31). Apache links identify upstream PRs; GreptimeTeam links identify source PRs in the maintained fork, not Apache merges. The fork also includes: [`72872aedb`](https://github.com/discord9/datafusion/commit/72872aedb) (lint-only timestamp conversion cleanup), [`a17899ec6`](https://github.com/discord9/datafusion/commit/a17899ec6) (recursive-CTE schema metadata), [`85fc18b51`](https://github.com/discord9/datafusion/commit/85fc18b51) (Date32 predicate widening), and [`7ac19906c`](https://github.com/discord9/datafusion/commit/7ac19906c191e51433c0c3d8810d4fd47ab1529a) (cross-timezone regression correction).

#### Additional retained correctness fixes

The final fork contains **20 commits** beyond official DataFusion 55 ([complete comparison](https://github.com/discord9/datafusion/compare/d5552342012888b7d1a3ab88d92e3d292fc0cde0...b66dca3260c8e04f314c8036cdfbb08f1a564d95)); all are described here, including timestamp equality/IN restoration in the compatibility section below. The following patches retain ordering, error-classification and dictionary behavior:

| Commit | Why needed / what changed |
| --- | --- |
| [`05fe011bd7`](https://github.com/discord9/datafusion/commit/05fe011bd7a4bf2bb4486de162fec034ee6be59e) | Ports [apache/datafusion#25065](https://github.com/apache/datafusion/pull/25065). Distribution enforcement must not remove fetched global boundaries or downgrade fetched ordered merges. It now preserves those nodes, including fetch zero, and protects ordered descendants before recursion. This is distinct from the existing LimitPushdown patch. |
| [`68fac3de71`](https://github.com/discord9/datafusion/commit/68fac3de71f3be55517d6445e237dfc825af78a0) | Required ordering was read too late for an inner LocalLimit or a removed no-op GlobalLimit. Accumulate it before early rewrites so global selection retains a sort-preserving merge. |
| [`84e55f0c6b`](https://github.com/discord9/datafusion/commit/84e55f0c6b5d9ea0c035c5482b135daeb74fefe7) | Restores the old fork's execution-error classification when every OneOf function signature fails; ordinary invalid arguments must not be described as a DataFusion internal bug. |
| [`e2b6246f37`](https://github.com/discord9/datafusion/commit/e2b6246f3764537db0d25b133f4c7d779e9ad3e1) | Restores Dictionary literal serialization: encode the underlying value and preserve the dictionary expression type through a cast. Existing dictionary schema support alone does not provide this. |
| [`579d2c9b9a`](https://github.com/discord9/datafusion/commit/579d2c9b9a18b30fd0ccb025b74648c4eecef312) | Restores dictionary-specialized grouping columns using DF55's existing factory, with cross-batch value deduplication, NULL handling, collision checks and typed output. Also clears drained group indexes on full emission so reuse is safe. |

Already-upstream retention includes finer timestamp coercion and safe overflow handling, preserved-side join dynamic filters, aggregate filter ordering, SPM statistics/cardinality, dictionary-regex NULL handling, and join-estimation overflow protection. Aligned timestamp widening equality/IN is restored under the explicitly accepted extreme-range overflow policy; unsafe dictionary casts remain restricted. Ordered timestamp widening retains its inherited overflow policy, not full-domain CAST/TRY_CAST equivalence.

DB-local sorting, opaque MergeSort global Top-K handling, and post-optimizer Filter-fetch repairs remain separate from the fork patches. `count_hash` preservation requires both the DB's fixed-seed custom-hasher API and the fork's raw floating-point representation; this is not a claim that all DataFusion hashing APIs are interchangeable.

#### DF55 correctness backports and local JSON ordering

- [`ca00e5c7f`](https://github.com/discord9/datafusion/commit/ca00e5c7f8b72896551802cd0e51eb7eeb04a1fa) backports [apache/datafusion#24817](https://github.com/apache/datafusion/pull/24817): disable aggregate dynamic filtering for the entire aggregate when any MIN/MAX argument is unsupported. Filtering only the supported subset can discard rows required by `MAX(ABS(v))`. All-direct-column MIN/MAX OR filtering remains enabled.
- [`b66dca326`](https://github.com/discord9/datafusion/commit/b66dca3260c8e04f314c8036cdfbb08f1a564d95) replaces the global `serde_json/preserve_order` requirement with local ordered serialization for physical EXPLAIN pgjson. Together with the separate Substrait change below, this prevents feature unification from changing unrelated DB JSON ordering.

Fork commit [`ad5c4be23`](https://github.com/discord9/datafusion/commit/ad5c4be23b614a9cb7f9e56dbadda8a474371c2b) backports three merged upstream correctness fixes:
- [#24248](https://github.com/apache/datafusion/pull/24248): preserve NULL semantics when cancelling repeated operands in complex bitwise XOR expressions.
- [#24380](https://github.com/apache/datafusion/pull/24380): preserve NULL rather than returning false for projected `col ~ '.*'`.
- [#24763](https://github.com/apache/datafusion/pull/24763): normalize literal-left comparisons before merging predicates, avoiding incorrect contradictions.

This is not a general collection of post-DF55 upstream fixes. The separately considered Parquet statistics, DISTINCT/fetch/OFFSET, legacy aggregate spill and window-argument fixes are not included.

#### JSON2 scalar-function planning

- **Why:** DataFusion 55 no longer called the scalar `ExprPlanner` hook GreptimeDB uses to supply JSON2 extraction types. Removing the obsolete GreptimeDB method would compile but makes expressions such as `abs(j.metrics.value)` fail during planning before analyzer rules can repair them.
- **What:** The retained scalar-planner patch restores the existing `RawScalarExpr` / `plan_scalar` extension point after scalar argument resolution. Modified `Original` arguments continue through later planners and normal function construction. GreptimeDB keeps its production JSON2 type injection; this does not introduce a test-only path or a new rewrite framework.

#### PromQL temporality join-key encoding

- **Why:** Stored primary-key labels use Arrow dictionary encoding. When a PromQL operand lacks `otlp_aggregation_temporality`, the planner supplies a NULL marker so that it matches cumulative (NULL) rows but not delta rows. Remote planning can decode this synthetic literal to plain UTF8 while the stored marker remains dictionary-encoded. DataFusion 55's multi-key hash-join comparator rejects those mixed array types, breaking arithmetic and `AND` in the existing `delta_temporality` SQLness case.
- **What:** Commit `ffba905e4cd` decodes only the present dictionary marker to its string value type and supplies the missing marker as a typed NULL of that same type. Qualified column names, NULL matching, and the left operand's visible columns and values for `AND`/`UNLESS` are preserved; padding is not exposed for a left operand lacking the temporality marker. This is a local PromQL alignment fix, not a global storage conversion or another fork patch. Tests cover both operand orders, both dictionary NULL representations, exact arithmetic/set results, and a nested consumer. The original stored-primary-key SQLness expectations remain unchanged.

#### Recursive CTE schema metadata

- **Why:** The stream-backed-scan reopening fix in GreptimeDB #9039 addresses execution-time reuse across recursive iterations. Its query over `information_schema.tables` also exposes a distinct DataFusion 55 physical-planning failure: `Cannot project plan to expected schema: schema metadata differ`. A literal seed and table-backed recursive branch may have different top-level metadata even though the declared recursive output takes the seed metadata; the query must continue to return depths `0, 1, 2`.
- **What:** Commit [`a17899ec6`](https://github.com/discord9/datafusion/commit/a17899ec6b9b031ea284bb680bcfc32678f50cff) makes the existing `project_plan_to_schema` helper construct `ProjectionExec` with the declared output schema metadata rather than reject only a top-level metadata difference. The exact-schema fast path remains. Column-count, data-type, nullability-narrowing, and field-metadata mismatches remain rejected. It neither strips scan metadata nor changes #9039's stream-reopening behavior.

#### Date32 predicate optimization restoration

- **Why:** The safe-cast preimage allowlist omitted lossless, order-preserving `Date32 → Date64`, so existing `simplify_expr.slt` EXPLAIN cases could not eliminate those casts. Every `i32` day value scaled by 86,400,000 fits in `i64`.
- **What:** Commit [`85fc18b51`](https://github.com/discord9/datafusion/commit/85fc18b51d2b21234bca67f376416c54c99024c2) allows only `Date32 → Date64`. Inverse-literal whole-day alignment and Date32-domain checks remain, and `Date64 → Date32` remains rejected. Coverage includes boundaries, ordered comparisons, nonaligned and out-of-range literals, and reverse-direction rejection.

#### Cross-timezone equality regression

- **Why:** A prior SLT positive case treated Brussels `2024-02-01 00:00:01 +01` and Los Angeles `2024-01-31 16:00:01 -08` as equal, although they differ by one hour. Because the view explicitly casts the input to `timestamp` before assigning Brussels, its matching Los Angeles wall clock is `15:00:01`.
- **What:** Commit [`7ac19906c`](https://github.com/discord9/datafusion/commit/7ac19906c191e51433c0c3d8810d4fd47ab1529a) changes the positive literal to 15:00, retains the expected positive row, and preserves 16:00 as a no-match regression. It changes neither timezone implementation nor the positive result semantics.

The lint-only commit [`72872aedb`](https://github.com/discord9/datafusion/commit/72872aedb2a78db72fb5b3cbc335200cac7486c5) removes four redundant `i128::from(timestamp_unit_scale(...))` conversions: the helper already returns `i128`, so strict clippy rejects these identity conversions. Timestamp arithmetic and guards are unchanged.

`EnforceSorting` remains conservatively preserved to retain the existing GreptimeDB sorting pass and distributed plan shaping during this upgrade. This does not establish that an extra pass is intrinsically required beyond DataFusion’s default `EnsureRequirements`.

### User-visible changes and compatibility limits

This upgrade is not a blanket SQL/API compatibility guarantee:

- Integer-input `median`, `approx_median`, and `approx_percentile_cont` produce floating-point results. Exact median interpolates even-sized inputs (for example, `4` becomes `4.5` in the existing regression); approximate quantile values also change. Consumers with fixed output schemas, including views and Flow sinks, need compatibility assessment; those schema migrations have not been exhaustively tested here.
- `REPLACE(s, '', replacement)` now returns `s` unchanged instead of interleaving the replacement string.
- Struct `unnest` output column names drop the `__unnest_placeholder(...)` wrapper. Clients depending on generated names should use explicit aliases.
- Approximate cardinality estimates can change: the existing 2,000-value `approx_distinct` case changes from `2000` to `1991`. This does not promise stable estimates across versions.
- Empty pretty-printed query results retain column headers rather than rendering as `++`; this is a display change, not added result rows.
- PostgreSQL result metadata honors OID-alias annotations (`OID`, `REGCLASS`, `REGTYPE`, and related types) instead of reporting only their former INT4/VARCHAR representations.
- Global `serde_json/preserve_order` is no longer enabled in the verified production graph. Both sources were removed: Substrait named-struct ordinals use a local IndexMap, and physical EXPLAIN JSON uses local ordered serialization. General DB JSON object ordering remains consistent with pre-upgrade main; SST metadata expectations were regenerated accordingly. The focused raw-byte CAS/checksum audit found no issue in the inspected paths, but is not a universal byte-stability claim.

Timestamp precision-widening equality and IN preimages are restored by fork commit [`605effa378`](https://github.com/discord9/datafusion/commit/605effa378c7ec6bb934263b66a273023f4aa295), reusing the existing literal conversion and round-trip alignment checks. Same-timezone widening is eligible; narrowing, nonaligned literals and unsafe dictionary casts remain guarded. The explicitly accepted policy favors normal-range time pruning even when rewriting changes CAST/TRY_CAST overflow behavior at extreme timestamps; no full-domain equivalence is claimed. Date32-to-Date64 remains eligible as well.

Standalone and distributed SQLness both verify `ts = TimestampMillisecond(5000)` reaches scan filters without CAST; aligned IN becomes source-column equalities. Exact/nonexact normal-range results and the accepted extreme-range policy are covered. Direct `arrow_try_cast` returns NULL on the tested positive overflow. The pre-existing SQL `TRY_CAST(... AS TIMESTAMP(9))` type-alias-lowering limitation remains outside this upgrade's scope.

Substrait 0.63 is patched via `GreptimeTeam/substrait-rs` revision `91ec978b0649417ad3da8390e7a515baec723b1b`. Its named-struct field order remains explicit and local; no protobuf/spec upgrade or DB-wide ordered-JSON abstraction is introduced. The raw Type::Object payload is now IndexMap (a Rust source API adjustment); legacy conversion helpers are retained. The fork also supports Cargo tagless checkout of the exact pinned spec release.

### GreptimeDB integration and regression coverage

- Restore primitive `Signature::uniform` contracts for clamp/min/max, modulo, rate and anomaly functions. Decimal inputs follow the legacy Float64 coercion path instead of reaching unsupported native-decimal execution. Actual Decimal128 coercion/execution regressions and standalone/distributed SQL column queries cover clamp and modulo. This does not add native Decimal arithmetic support.
- Restore scalar `count_hash` typed-NULL behavior, including its historical zero hash contribution. Preserve fixed seeds, raw floating-point hashing and state encoding; a persisted `[0]` state regression checks state/result agreement with fresh typed-NULL input. The pre-existing scalar/group NULL inconsistency is deliberately not corrected in this upgrade.
- Remove broad SortExec/topology masking. Separate standalone/distributed TQL expectations now expose their real sort keys and distinct plan shapes. PostgreSQL catalog checks explicitly assert `attnum`/column-name/type associations in ordinal order.
- Share only identical geo integer-signature lists in the existing helper module, and simplify reviewed type imports. Numeric and integer-only lists remain separate.
- RegionScan expression traversal includes pushed dynamic filters, cloning their Arcs before releasing the mutex and invoking callbacks; discovery preserves the same filter identity. PartSort exposes its existing produced filter, and RangeSelect visits its aggregate arguments, ORDER BY expressions and grouping roots. These repair DF55 interface omissions; they are not claims of reproduced SQL-result or serialization failures.
- PostgreSQL string-backed OID aliases retain text output, including COPY text, while binary output encodes four-byte OIDs. Name resolution is limited to static REGPROC catalog entries; numeric strings support all 12 advertised aliases. Unknown/ambiguous names remain valid text but fail binary encoding. Actual DataRow/COPY tests cover NULL, zero, mixed formats and u32 bounds; the real extended-protocol `SELECT typreceive FROM pg_catalog.pg_type WHERE oid = 16` checks REGPROC/OID 2436 and text `boolrecv`.
- Restore stale external-table, Parquet Map, native-histogram and OTLP histogram JSON expectations to the main baseline after isolating ordered serialization. Generate empty-result headers through SQLness; do not change returned rows or NULL semantics to fit snapshots.
- Fix the Kafka WAL helper Deployment/Pod creation race by waiting for Deployment rollout rather than querying a Pod selector before Pods exist.

### Validation

- [Rust CI](https://github.com/GreptimeTeam/greptimedb/actions/runs/34855679661) and [Integration CI](https://github.com/GreptimeTeam/greptimedb/actions/runs/34855680225) passed, including all five SQLness configurations, recent-release compatibility and fuzz tests.
- Local review-fix validation passed: table **87 tests**, servers **563 tests**, focused query **61 tests**, plus strict Clippy for all three packages.
- [Performance comparison](https://github.com/GreptimeTeam/greptimedb/actions/runs/34846025165) against main: 8 cases, 27 comparisons, no result-validation errors; all 22 configured thresholds passed (5 comparisons have no threshold). Median changes ranged from **−10.7% to +10.5%**. Integer-counter selection slowed from **453.58 to 501.10 ms**; the integer-counter threshold is 75%, so passing this gate does not establish acceptable performance or explain the observed slowdown. Its cause and repeatability remain unverified. The final commit changes only the CI readiness wait, not the benchmarked production code.

#### Targeted evidence

These checks were run as the corresponding patches were introduced; they supplement, rather than replace, the current-head CI above.

- Fork limits/distribution: native before/after regressions retain LIMIT 1 and LIMIT 0 through repeated optimization and preserve ordered Top-K `[10,20]` rather than returning four rows. Dictionary grouping tests cover cross-batch deduplication, NULLs, forced hash collisions, typed output and full-drain reuse. Fork [`579d2c9b9a`](https://github.com/discord9/datafusion/commit/579d2c9b9a18b30fd0ccb025b74648c4eecef312) validation passed 569 physical-optimizer, 82 group-value, 57 Substrait unit, 202 Substrait integration (6 ignored), 3 doc and 31 coercion tests, plus strict affected-crate Clippy; normal and forced-collision dictionary runs each passed 79 tests.
- Timestamp restoration ([`605effa378`](https://github.com/discord9/datafusion/commit/605effa378c7ec6bb934263b66a273023f4aa295)): 9 expr-common, 12 logical-optimizer and 58 physical-simplifier tests passed, with strict affected-crate Clippy. Standalone/distributed SQLness verify bare-millisecond EQ/IN scan predicates, normal-range results and the explicitly retained overflow policy.
- Aggregate-filter backport: complete `push_down_filter_regression.slt` passed, including mixed-aggregate numeric results and no-filter assertions. The DB SQL23 HTTP reproduction returned `[1000000, 1700000019999]` in all 20 repetitions without an incomplete aggregate dynamic filter in EXPLAIN ANALYZE.
- NULL/predicate simplification ([`ad5c4be23`](https://github.com/discord9/datafusion/commit/ad5c4be23b614a9cb7f9e56dbadda8a474371c2b)): 759 optimizer tests and complete `scalar.slt`, `simplify_expr.slt` and `simplify_predicates.slt` passed, plus fork all-target/all-feature strict Clippy.
- Numeric coercion and hashing: `cargo nextest run --locked -p common-function` passed 343 tests, including Decimal128 execution and persisted `[0]` count_hash restoration. Real Decimal-column clamp/min/max/modulo queries passed in standalone and distributed modes.
- JSON-order isolation: the locked production `cmd + vector_index` normal/build dependency graph has no `serde_json/preserve_order`. Physical pgjson tests passed with either feature state; Substrait passed 59 unit and 3 doc tests. The targeted DB validation passed 787 tests (2 skipped), with 44 SQLness executions across JSON, SST, SQL23 and applicable COPY cases. A separate Parquet/JSON suite passed 152 tests and both file-storage histogram regressions passed; the broader local S3 histogram run was blocked by unavailable S3 and is not claimed as passing.
- Sorting and metadata: standalone/distributed SQLness assert real sort keys and topology-specific plans without broad masking; PostgreSQL catalog checks assert `attnum`, column-name and type associations in ordinal order. Final-path fixtures and default/Flat-format NULL-result fixtures were rerun; `.result` files were generated, not hand-edited.

### Remaining issues

PostgreSQL Struct payload loss (#9150), incompatible List<Struct> JSON-array elements (#9151), and sub-millisecond `irate` infinity (#9152) reproduce on both the tested main and upgrade candidate. They remain bugs, are not fixed here, and have separate reproduction reports. The original nanosecond `irate` input returns identical `inf` on main and candidate in standalone/distributed mode, before and after flush. A separate populated probe with changed sample spacing returned `4e-6`, with PerSeries ordering and identity timestamp projection in ANALYZE; that different fixture does not explain or validate the original input.

An earlier logical-pruning fuzz run also encountered a candidate datanode health-probe timeout/restart; its retry passed, but the underlying stall was not established.

A [historical remote-WAL migration fuzz timeout](https://github.com/GreptimeTeam/greptimedb/actions/runs/34821819268/job/103909017295) remains unexplained: its exact input (`0x13`) is preserved but has not been replayed in the original environment. Subsequent passing random runs do not resolve that failure.

## PR Checklist

The compatibility boxes remain unchecked because of the output-type and generated-name changes above and incomplete validation of fixed-schema consumers; the passing compatibility suite is not a blanket compatibility guarantee.


sunng87:
orc and pg-catalog are our managed libraries, make sure it merges to upstream before merging this.

sunng87:
I'm working on the pg-catalog upgrade for datafusion 54. will commit to this branch.

discord9:
Thanks — I see `193726d60a` updates pg-catalog to 0.18 and adds the related planner/session compatibility changes. The PR DCO check is currently failing because this commit has no `Signed-off-by` trailer. Could you please add the sign-off using the DCO app’s accepted flow? I’ll retain and validate this commit while reconciling the latest `main`; I won’t overwrite the branch update.

sunng87:
DCO is a non-blocking check, just set it passed.

discord9:
Thanks, DCO is green now. One remaining managed-dependency setup: `GreptimeTeam/datafusion` still has `main` on 53.1, and my direct attempt to create a 54 branch returned HTTP 403 (`push=false`). Could you create an organization target branch such as `greptimedb-54.0.0` from official DataFusion `54.0.0@45d943dfb8699dc9cb9ef2320e955b73e3e6c03b`? The validated thin-fork head is `discord9/datafusion:thin-fork/2026-07-16-datafusion-54@027358559b74632d3231eea46a34fcc7f7c99fe9`; once the target exists I can open the PR and move #8555 to the organization-controlled SHA.

sunng87:
datafusion-orc is now on 0.9

discord9:
The pg-catalog 0.18 integration uncovered one remaining upstream compatibility gap during real PostgreSQL-protocol SQLness validation: OID aliases are represented as `Int32 + pg.oid_alias` metadata, but reverse casts and wire mapping currently expose numeric values (`25::regtype::text -> 25`, `typinput -> 46`) instead of PostgreSQL names/types. I filed [datafusion-postgres#384](https://github.com/datafusion-contrib/datafusion-postgres/issues/384) with library-level reproductions. I am fixing the Greptime-owned wire metadata mapping, `pg_attribute.attnum` ordering, and unstable OID assertions locally; I will not accept the numeric snapshot as the baseline. The symbolic alias-to-text fix belongs upstream and remains a Draft gate.

discord9:
Updated locally to the managed `datafusion-orc 0.9.0` crates.io release. The temporary personal git source is removed; the locked `common-datasource` and `mito2` checks pass. This will be included in the pending latest-main merge commit.

discord9:
Could you confirm the intended ownership for the DataFusion thin-fork pin itself? The existing dependency used `GreptimeTeam/datafusion`, but the validated DF54 head is currently `discord9/datafusion@027358559b...`; I cannot create the requested GreptimeTeam 54 target branch (`push=false`, direct push returns 403). Is moving this SHA into `GreptimeTeam/datafusion` a hard pre-merge requirement, or is the immutable personal-fork SHA acceptable for this upgrade? I am treating it as a conservative dependency-ownership gate until clarified.

discord9:
The managed DataFusion 54 path is now available. I created `GreptimeTeam/datafusion:greptimedb-54.0.0` directly from the exact official `54.0.0@45d943dfb...` release commit and opened [GreptimeTeam/datafusion#20](https://github.com/GreptimeTeam/datafusion/pull/20) from the validated thin-fork head `027358559b...`. PR #20 is CLEAN and review has been requested from @sunng87. After it merges, #8555 can replace all personal DataFusion URLs with the organization-controlled immutable SHA.

sunng87:
datafusion 54.1 is out.
I also have a new release for datafusion-pg-catalog 0.18.2 with some bugfix. Consider sending the quote patch to upstream.

src/common/function/src/scalars/geo/h3.rs
```diff
@@ -42,11 +41,22 @@ static CELL_TYPES: LazyLock<Vec<DataType>> =
 static COORDINATE_TYPES: LazyLock<Vec<DataType>> =
     LazyLock::new(|| vec![DataType::Float32, DataType::Float64]);
 
-static RESOLUTION_TYPES: &[DataType] = INTEGERS;
+const INTEGER_TYPES: &[DataType] = &[
```
sunng87, line 44: Extract this to avoid duplications

discord9:
Addressed: `eaf98f1a53` extracts the duplicated coordinate×resolution signature construction, coordinate/resolution argument parsing, and per-row resolution conversion into shared helpers (`signature_of_coordinates_and_resolution`, `extract_coordinate_resolution_args`, `resolution_at`) and drops the redundant `RESOLUTION_TYPES`/`DISTANCE_TYPES`/`POSITION_TYPES` alias statics in `src/common/function/src/scalars/geo/h3.rs`. Behavior, signature order, error messages, and visibility are unchanged; `common-function` h3 tests pass (1/1) and clippy/fmt are clean.

src/catalog/src/system_schema/pg_catalog.rs
```diff
@@ -224,6 +226,46 @@ impl CatalogInfo for CatalogManagerWrapper {
     }
 }
 
+/// Restores user-visible declaration order from column IDs when table metadata
+/// provides a complete, unique set of IDs. IDs are monotonically allocated by
+/// `TableMeta::alloc_new_column`, so sorting them preserves declaration order
+/// across the physical key/value/time-index layout and tolerates dropped-column
+/// gaps. Older or incomplete metadata falls back to its physical schema order.
+fn schema_in_column_id_order(
```
fengjiachun: Sorting by column id restores declaration order for logical tables, but with `ALTER TABLE ADD COLUMN ... FIRST/AFTER` the newest (highest) id sits mid-schema, so this reports that column last in `attnum` order — diverging from `SELECT *` order. Clients that rebuild schemas from `attnum` (e.g. ADBC) would see a different column order than the query results.

This matches real Postgres (which can only append), so it may be an acceptable trade-off — but the doc comment overstates the guarantee. Worth a note here, and ideally a test covering the positional-add case.

src/servers/src/postgres/handler.rs
```diff
@@ -424,7 +461,36 @@ impl ExtendedQueryHandler for PostgresServerHandlerInner {
             .with_label_values(&[crate::metrics::METRIC_POSTGRES_EXTENDED_QUERY, db.as_str()])
             .start_timer();
 
-        let pg_sql_plan = &portal.statement.statement;
+        let refreshed_plan = if let PgSqlPlan {
+            plan: SqlPlan::Plan(_, stmt),
+            copy_to_stdout_format,
+            planning_context,
+        } = &portal.statement.statement
+            && !planning_context.matches(&query_ctx)
```
fengjiachun: Once `search_path` changes, `planning_context.matches()` stays false for the lifetime of the statement, so every subsequent `Execute` replans from scratch (`do_describe` = full planning) plus contract validation. A client that switches schema once and then runs a prepared statement in a loop silently pays per-execute planning cost.

If the statement store allows it, consider caching the refreshed `PgSqlPlan` back (keyed by planning context); otherwise a comment documenting the cost would help.

fengjiachun (commented):
Two substantive comments inline. Also a couple of nits:

- `src/servers/src/postgres/types.rs`: the `pg_oid_alias_type` doc comment says "Version 0.18.1" but the dependency is 0.18.3.
- `src/servers/src/postgres/handler.rs`: when the refreshed plan is no longer describable (`SqlPlan::Statement`), the error still says "cached plan must not change parameter type" — a generic "cached plan is no longer valid" would be more accurate for that branch.

discord9:
@fengjiachun thanks for the review — both inline comments and the nits are addressed in `4354daab`:

1. **pg_catalog.rs attnum ordering**: doc comment corrected to state column-id order matches `SELECT *` only for plain appends and diverges for positional adds (`ADD COLUMN ... FIRST/AFTER`), consistent with real PostgreSQL's append-only semantics; added `test_schema_in_column_id_order_places_positional_add_last` covering the positional case.
2. **handler.rs replan cost**: documented that the statement store is immutable after parse (`&Portal`/`Arc<StoredStatement>`), so a changed `search_path`/`USE` makes every subsequent `Execute` re-plan (describe + logical-plan step, no param binding/execution); accepted over adding mutable cached state.
3. **types.rs doc**: version 0.18.1 → 0.18.3.
4. **handler.rs error message**: undescribable-refreshed-plan error generalized to "cached plan is no longer valid".

discord9:
The `test` job failures on this head are pre-existing flaky tests unrelated to this upgrade (neither file is touched by this branch): `wal_provider::tests::test_provider_with_kafka_allocate_wal_options` panics on transient Kafka topic-deletion errors (`topic_creator.rs:188`), and `mito2 engine::basic_test::test_all_index_metas_list_all_types` asserts a timing-dependent index-metadata ordering (`basic_test.rs:1234`). Both are environment/timing sensitive and also fail intermittently on `main`. The job has been re-run; everything else on this head passes (35 checks green incl. check/clippy/rustfmt/build/sqlness/fuzz).

fengjiachun:
One more DataFusion pin follow-up: the currently pinned 54.1 thin-fork head, [`d75a5582a`](https://github.com/GreptimeTeam/datafusion/commit/d75a5582a7112e11bf2f36449f47fd60ac820a05), still uses raw `usize` multiplication in `estimate_inner_join_cardinality`:

```rust
(left_num_rows * right_num_rows) / value
```

[GreptimeTeam/datafusion#24](https://github.com/GreptimeTeam/datafusion/pull/24) changes this to checked multiplication and returns unknown statistics on overflow. This matters for nested joins on constant columns: a safe pairwise estimate can become the input to a parent join and then panic in debug builds or wrap in release builds.

Because #24 targets the 53.1 maintenance branch, #8555 should not pin its eventual merge commit directly. Please port/cherry-pick the equivalent fix onto `thin-fork/2026-07-22-datafusion-54.1`, advance that branch, and then update all DataFusion revisions in this PR to the new 54.1 thin-fork SHA.

discord9:
@fengjiachun done — `b751bbd37` ports #24 (checked multiplication for the join-cardinality cartesian product, unknown statistics on overflow, plus the nested-join regression test) onto `thin-fork/2026-07-22-datafusion-54.1`; the org branch has been advanced and #8555 now pins `b751bbd37` in all 15 `[patch.crates-io]` DataFusion entries (Cargo.lock updated, `--locked` check passes).

src/common/function/src/aggrs/count_hash.rs
```diff
@@ -111,12 +103,7 @@ impl AggregateUDFImpl for CountHash {
 
         Ok(Box::new(CountHashAccumulator {
             values: HashSet::default(),
-            random_state: RandomState::with_seeds(
-                RANDOM_SEED_0,
-                RANDOM_SEED_1,
-                RANDOM_SEED_2,
-                RANDOM_SEED_3,
-            ),
+            random_state: FixedState::with_seed(RANDOM_SEED_0),
```
fengjiachun: [P1] Preserve compatibility of `CountHash` partial states

This switches the hash algorithm from the DF53 `ahash::RandomState` (with four fixed seeds) to DF54's foldhash-backed `RandomState`, while the serialized aggregate state remains an unversioned `List<UInt64>`. During a rolling upgrade, partial states produced by old and new nodes can therefore be merged: the same logical value has different hashes, so the final `HashSet` treats it as two distinct values and silently overcounts. Declaring cross-version state mixing unsupported does not prevent the distributed aggregation path from producing it or fail fast when it happens. Please either retain the v1 hash for this state schema, or version/reject incompatible states, and add a DF53-partial -> DF54-final regression test.

fengjiachun (commented):
One compatibility issue found in `CountHash`.

tests/cases/standalone/common/tql/general_table.sql
```diff
@@ -18,7 +18,7 @@ WITH(
 -- SQLNESS REPLACE (\s\s+) _
 -- SQLNESS REPLACE (cpu_usage\.ts_range) ts_range
 -- SQLNESS REPLACE (cpu_usage\.ts) ts
--- SQLNESS REPLACE (?m)^\|_\|_\|_SortExec:.*\n\|_\|_\|_RepartitionExec:.*\n
+-- SQLNESS REPLACE (?m)^\|_\|_\|_SortExec:.*\n
```
fengjiachun, line 22: This normalization change hides a regression rather than DF55 formatting noise. I built both revisions and ran the same query on the same data (20k rows / 20 series): on main `PromSeriesDivideExec` reads `SeriesScan` directly, while this branch inserts `RepartitionExec: partitioning=Hash([k@2], 10), input_partitions=1, maintains_sort_order=true` over a `SortPreservingMergeExec` — the scan's parallel partitions are merged down to one stream and hash-shuffled back out, so every PromQL query serializes its datanode scan and pays a full shuffle.

Could you root-cause why `RegionScanExec` (PerSeries) no longer satisfies `SeriesDivideExec`'s distribution/ordering requirements under `EnsureRequirements` before merging? For what it's worth I ruled out the deprecated `HashPartitioned` variant (DF55 `satisfaction()` accepts both), the `input_distribution_requirements()` shim, `hash_necessary` alignment, and the statistics path — the remaining suspect is the merged `EnforceDistribution`+`EnforceSorting` pass.

src/query/src/planner.rs
```diff
@@ -78,6 +82,35 @@ pub struct DfLogicalPlanner {
     session_state: SessionState,
 }
 
+fn range_query_has_nested_range(stmt: &SqlStatement) -> bool {
```
fengjiachun: This runs a full `visit_expressions` walk, with a `function.name.to_string()` allocation per function node, on every SQL statement — while the original check at `range_select/plan_rewrite.rs:424` is still in place. It also only inspects `range_fn`'s first argument, so a nested range in any later argument slips through. Can the walk be gated on the statement actually containing a range query?

src/common/function/src/aggrs/count_hash.rs
```diff
@@ -275,11 +259,36 @@ impl GroupsAccumulator for CountHashGroupAccumulator {
...
+        let array = &values[0];
+        let mut hashes = vec![0; array.len()];
+        if array.data_type() != &DataType::Null {
+            create_hashes(&[ArrayRef::clone(array)], &self.random_state, &mut hashes)?;
+        }
+
+        let mut builder = ListBuilder::new(UInt64Builder::with_capacity(array.len()))
+            .with_field(Arc::new(Field::new_list_field(DataType::UInt64, true)));
+        for row in 0..array.len() {
+            let included = array.is_valid(row)
```
fengjiachun: This filters with `array.is_valid(row)` (physical validity) while `update_batch` uses `array.logical_nulls()`. The two diverge for dictionary-encoded input, which GreptimeDB does produce — tag columns show up as `Dictionary(UInt32, Utf8)` in physical plans at runtime. Materializing `logical_nulls()` once and using it in both paths would keep the skip-partial-aggregation result consistent with the normal one.

tests/cases/standalone/common/information_schema/ssts.sql
```diff
@@ -36,7 +36,37 @@ SELECT * FROM information_schema.ssts_manifest order by file_path;
 -- SQLNESS REPLACE (/public/\d+/\d+_\d+/index/<UUID>\.puffin) /public/<TABLE_ID>/<REGION_ID>_<REGION_NUMBER>/index/<UUID>.puffin
 -- SQLNESS REPLACE (/public/\d+/\d+_\d+) /public/<TABLE_ID>/<REGION_ID>_<REGION_NUMBER>
 -- SQLNESS REPLACE (/public/\d+) /public/<TABLE_ID>
-SELECT * FROM information_schema.ssts_index_meta ORDER BY meta_json;
+-- SQLNESS REPLACE (\{"(bloom|fulltext|inverted)":.*\}) <META_JSON>
```
fengjiachun: This 60-line rewrite looks unrelated to the DataFusion upgrade, and the added `SQLNESS REPLACE` masks `meta_json` entirely — the column this case exists to verify. Could it move to its own PR with the motivation stated?

src/common/function/src/aggrs/count_hash.rs
fengjiachun: Still unaddressed on the DF55 head: the hash is now `foldhash::fast::FixedState::with_seed(RANDOM_SEED_0)` while the serialized state is still an unversioned `List<UInt64>`, so a rolling upgrade can still merge old- and new-hash partial states and silently overcount. Also worth considering `hash_utils::QualityRandomState` here — DF added it for exactly this hash-as-value case (it backs `HLL_RANDOM_STATE`), whereas the `fast` tier targets hash tables that resolve collisions by key comparison.

src/query/src/planner.rs
discord9: Fixed in `964cbc552e5`: nested RANGE detection now uses one allocation-free, depth-aware AST visitor over all function arguments (including DML query contexts), while preserving sibling-range and wrapper behavior.

src/common/function/src/aggrs/count_hash.rs
discord9: Fixed in `964cbc552e5`: `convert_to_state` now computes `logical_nulls()` once and uses it consistently. Added a regression with valid dictionary keys referencing NULL dictionary values across scalar, grouped, and converted-state paths.

tests/cases/standalone/common/information_schema/ssts.sql
discord9: Fixed in `964cbc552e5`: removed the unrelated query rewrite and whole-column `<META_JSON>` masking. Restored the original query and regenerated the result with SQLness so `meta_json` is verified directly.

src/common/function/src/aggrs/count_hash.rs
discord9: Fixed in `964cbc552e5` with DataFusion fork `dec6af133d6`: CountHash retains the historical four-seed ahash state, and the custom-hasher path preserves raw float representations recursively (including nested signed zero). Added frozen vectors and scalar/group/state-merge coverage; `QualityRandomState` was not used because it would create another incompatible state boundary.

tests/cases/standalone/common/tql/general_table.sql
discord9: Fixed by restoring the standalone sorting-enforcement pass after GreptimeDB’s custom scan/distribution rules. Both standalone and per-region distributed TQL plans now directly consume the PerSeries scan again, without the SortPreservingMerge/Repartition pair; the SQLness snapshots are restored to the pre-DF55 plans.

sunng87:
Released datafusion-orc 0.10.0 for datafusion 55
We may still need some time for a datafusion-postgres release because of datafusion-python and geodatafusion.

sunng87:
We can use this commit for arrow-pg and datafusion-pg-catalog: `3c77e6c32b8db80635a0d2f4b318a36b31170bc3`

fengjiachun (approved):
LGTM

discord9:
Marking this PR as draft for now. We will merge it after the 1.2 release.

MichaelScofield:
hold the merge for easier 1.2 backport plz

killme2008:
This upgrade also fixes #9038 (projection pushdown into a recursive CTE prunes columns the recursive term still reads, causing `project index N out of bounds` or silently missing rows). Fork rev `dec6af1` contains apache/datafusion#22476, which landed in DataFusion 55.0.0. Worth adding a sqlness case in `tests/cases/standalone/common/cte/cte.sql` here — see #9038 for repros.

src/catalog/src/information_extension.rs
```diff
@@ -173,6 +169,17 @@ impl ExecutionPlan for DistributedInspectExec {
         vec![]
     }
 
+    fn apply_expressions(
+        &self,
+        _f: &mut dyn FnMut(
+            &Arc<dyn datafusion::physical_expr::PhysicalExpr>,
+        ) -> datafusion::common::Result<
+            datafusion::common::tree_node::TreeNodeRecursion,
```
killme2008: `use datafusion::common::tree_node::TreeNodeRecursion` instead.

src/common/function/src/aggrs/aggr_wrapper/tests.rs
```diff
@@ -109,6 +105,15 @@ impl ExecutionPlan for MockInputExec {
         vec![]
     }
 
+    fn apply_expressions(
+        &self,
+        _f: &mut dyn FnMut(
+            &Arc<dyn datafusion_physical_expr::PhysicalExpr>,
+        ) -> datafusion_common::Result<TreeNodeRecursion>,
```
killme2008: The same suggestion, I don't repeat it by myself.

src/common/function/src/scalars/geo/geohash.rs
```diff
@@ -22,14 +22,24 @@ use datafusion::arrow::array::{Array, AsArray, ListBuilder, StringViewBuilder};
...
 use datafusion_common::DataFusionError;
-use datafusion_expr::type_coercion::aggregates::INTEGERS;
 use datafusion_expr::{ScalarFunctionArgs, Signature, TypeSignature, Volatility};
 use geohash::Coord;
 use snafu::ResultExt;
 
 use crate::function::{Function, extract_args};
 use crate::scalars::geo::helpers;
 
+const INTEGER_TYPES: &[DataType] = &[
```
killme2008: It's repeated everywhere, could we define it in the common mod or some other palces?

killme2008: `src/common/function/src/helper.rs` would be a good place.

killme2008:
### User-visible behavior changes inherited from the upgrade

Most of these are not introduced by this PR's own code, but they ship with it and are visible to users. They should be listed in the release notes, and the PR checklist needs updating — `API changes are backward compatible` is currently checked.

| Change | Source | Evidence |
| --- | --- | --- |
| `median()` / `approx_median()` / `approx_percentile_cont()` return Float64 for integer input, and even-count medians now interpolate (`median(r)::VARCHAR` returns `4.5`, was `4`) | DataFusion 55 — `functions-aggregate/src/median.rs`: "Integer inputs are coerced to Float64 so the average of the two middle values is exact" | `aggregate/median.result`, `aggregate/approx_median.result`, `aggregate/approx_percentile_cont.result`, `range/calculate.result` |
| `REPLACE(s, '', x)` returns `s` unchanged instead of interleaving `x` between every character (now matches PostgreSQL/MySQL) | DataFusion 55 | `function/string/replace.result` |
| `unnest()` output columns lose the `__unnest_placeholder(...)` wrapper: `geo_path(lat,lon,ts).lat` instead of `__unnest_placeholder(geo_path(lat,lon,ts)).lat` | DataFusion 55 | `function/geo.result`, `select/unnest.result` |
| `approx_distinct()` returns different estimates for the same input (2000 -> 1991) | DataFusion 55 | `aggregate/approx_distinct.result` |
| Empty result sets render with a header row instead of `++` | arrow 59 pretty-print | 159 occurrences across 66 `.result` files |
| PostgreSQL wire types: OID-alias columns are now reported as `OID`/`REGCLASS`/`REGTYPE`/... instead of `INT4`/`VARCHAR` | this PR (`servers/src/postgres/types.rs`) | `postgres/types.rs` |

The `median` change is the one with real blast radius: it changes the column type of existing views and flow sink tables built on `median()`, not just the printed value.

tests/cases/standalone/common/filter/cast_preimage.result
```diff
@@ -358,27 +358,29 @@ EXPLAIN SELECT host, v FROM cast_preimage_ts_ms
...
++---------------+-----------------------------------------------------------------------------------------------------------------------------------------------+
+| plan_type     | plan                                                                                                                                          |
++---------------+-----------------------------------------------------------------------------------------------------------------------------------------------+
+| logical_plan  | MergeScan [is_placeholder=false, remote_input=[                                                                                               |
+|               | Sort: cast_preimage_ts_ms.host ASC NULLS LAST                                                                                                 |
+|               |   Projection: cast_preimage_ts_ms.host, cast_preimage_ts_ms.v                                                                                 |
+|               |     Filter: CAST(cast_preimage_ts_ms.ts AS Timestamp(ns)) = TimestampNanosecond(5000000000, None)                                             |
```
killme2008: This is a regression, not just a plan text change. On main the cast was rewritten away and `ts = TimestampMillisecond(5000)` reached the scan. Now the CAST stays in `partial_filters`, so the time index can't prune on this predicate.

The comment above this case in the .sql says the rewrite is exact at millisecond precision, which is why it was added in the first place.

The description says widening equality is intentionally not re-enabled, but the fork does re-enable Date32 -> Date64 widening in `85fc18b51` with a whole-day alignment check. Same kind of lossless widening, two different answers. Either restore ms -> ns with the same alignment check, or drop it and file an issue, so it doesn't just live on as a line in a .result file.

tests/cases/standalone/common/tql/general_table.sql
```diff
@@ -18,7 +18,8 @@ WITH(
 -- SQLNESS REPLACE (\s\s+) _
 -- SQLNESS REPLACE (cpu_usage\.ts_range) ts_range
 -- SQLNESS REPLACE (cpu_usage\.ts) ts
--- SQLNESS REPLACE (?m)^\|_\|_\|_SortExec:.*\n\|_\|_\|_RepartitionExec:.*\n
+-- SQLNESS REPLACE (?m)^(\|_\|_\|_PromSeriesDivideExec:[^\n]*\n)(?:\|_\|_\|_SortExec:[^\n]*\n)?\|_\|_\|_RepartitionExec:[^\n]*\n $1
+-- SQLNESS REPLACE (?m)^\|_\|_\|_SortExec:.*\n
```
killme2008: This deletes every SortExec line from the plan instead of updating the expectation. The .result loses `SortExec: expr=[job@0 ASC NULLS LAST, ts@1 ASC NULLS LAST]`.

PromSeriesDivideExec needs sorted input. If SortExec is filtered out of the expectation, a future change that drops the sort passes this test. Please update the expected plan instead.

tests/cases/standalone/common/system/pg_catalog.sql
```diff
@@ -277,7 +277,30 @@ CREATE table foo
 SELECT attname, atttypid FROM pg_catalog.pg_class AS cls INNER JOIN
 pg_catalog.pg_attribute AS attr ON cls.oid = attr.attrelid INNER JOIN
 pg_catalog.pg_type AS typ ON attr.atttypid = typ.oid WHERE attr.attnum >= 0 AND
-cls.oid = 'foo'::regclass::oid ORDER BY attr.attnum;
+cls.oid = 'foo'::regclass::oid ORDER BY attname;
```
killme2008: Why the change from `ORDER BY attr.attnum`? attnum is what drivers (ADBC, psql, the language clients) read to get column order, and ordering by attname means we no longer check it.

If attnum values or ordering changed in datafusion-pg-catalog 0.18, select attnum as well and assert on it, instead of switching the sort key.

tests/cases/standalone/common/information_schema/ssts.result
```diff
@@ -120,15 +120,15 @@ SELECT * FROM information_schema.ssts_index_meta ORDER BY meta_json;
...
+| data/greptime/public/<TABLE_ID>/ | data/greptime/public/<TABLE_ID>/<REGION_ID>_<REGION_NUMBER>/index/<UUID>.puffin |<NUM>|<NUM>|<NUM>|<NUM>|<NUM>| <UUID> |<NUM>| bloom_filter   | column      |<NUM>| {"column":1} |<NUM>| {"bloom":{"rows_per_segment":10240,"segment_count":1,"row_count":1,"bloom_filter_size":64}}                                                                                                                              |<NUM>|
+| data/greptime/public/<TABLE_ID>/ | data/greptime/public/<TABLE_ID>/<REGION_ID>_<REGION_NUMBER>/index/<UUID>.puffin |<NUM>|<NUM>|<NUM>|<NUM>|<NUM>| <UUID> |<NUM>| inverted       | column      |<NUM>| {"column":0} |<NUM>| {"inverted":{"bitmap_type":"Roaring","base_offset":0,"inverted_index_size":81,"relative_fst_offset":26,"fst_size":55,"relative_null_bitmap_offset":0,"null_bitmap_size":8,"segment_row_count":1024,"total_row_count":1}} |<NUM>|
```
killme2008: The meta_json keys went from sorted to source order here, and `puffin_index.rs` isn't touched in this PR.

The cause is substrait 0.62.2 -> 0.63.0. 0.63.0 declares `serde_json = { version = "1.0.145", features = ["preserve_order"] }`, and feature unification turns that on for the whole workspace, so `serde_json::Map` becomes an IndexMap everywhere. Cargo.lock confirms it: serde_json on main has no indexmap dependency, on this branch it does.

So this affects every JSON we build or round-trip through `serde_json::Value`, not just this column. JSON column values are fine, they go through `jsonb::parse_value`. Please call it out in the description, and check that nothing compares or hashes serialized JSON bytes.

src/common/function/src/scalars/math/clamp.rs
```diff
@@ -33,7 +32,7 @@ impl Default for ClampFunction {
     fn default() -> Self {
         Self {
             // input, min, max
-            signature: Signature::uniform(3, NUMERICS.to_vec(), Volatility::Immutable),
+            signature: Signature::numeric(3, Volatility::Immutable),
```
killme2008: `Signature::numeric` is not the same input domain as `Signature::uniform(n, NUMERICS)`. `TypeSignature::Numeric` goes through `NativeType::is_numeric()`, which is integer + float + decimal, and it coerces to a common type instead of matching exactly. The old NUMERICS was Int*/UInt*/Float16/32/64 only.

Same swap in `modulo.rs`, `rate.rs` and the three anomaly window functions. Do those handle Decimal input? If not, we've moved a planning-time type error to execution time.

killme2008 (commented):
A few things from reading the diff. Breaking changes are in a separate comment above.

One more that has no source line: the description says the current head is `e749334f0d4`, but the head of this PR is `3b8c0a85b5`. None of the validation results in the description refer to the commit being reviewed.

tests/cases/standalone/common/information_schema/ssts.result
discord9: Updated in bc7ecdce88f. We decided not to accept global serde_json ordering changes as part of this upgrade. The production feature graph revealed two enablers, not just Substrait: substrait 0.63 and datafusion-physical-plan.

Both are now scoped locally:
- GreptimeTeam/substrait-rs@91ec978 preserves named-struct field order/ordinals with an explicit IndexMap, without enabling serde_json/preserve_order. The existing spec/protobuf version is unchanged; legacy raw-type conversion helpers are retained.
- DataFusion fork b66dca326 preserves physical EXPLAIN JSON field and metric ordering through renderer-local serialization, including duplicate metric overwrite behavior, without the global feature.

With the published Git pins, the locked cmd + vector_index normal/build dependency graph has no serde_json/preserve_order. General DB JSON ordering returns to the pre-upgrade main behavior; SQLness regenerated the SST meta_json expectation accordingly. Published-pin build and 787 Substrait/datasource/query tests passed (2 skipped); equivalent local-source integration passed 44 SQLness executions. Physical pgjson tests pass with and without preserve_order. New-head CI/compatibility is still pending.

We also checked the targeted byte-sensitive paths: table-info/table-route/table-repartition CAS compares captured original KV bytes, and manifest checksums cover original payload bytes. That is a focused audit, not a claim that every JSON consumer was exhaustively audited. The PR description now documents the local-ordering decision and validation. Leaving this thread open for your review.

discord9: **Problem:** The upgrade enabled `serde_json/preserve_order` through both DataFusion physical-plan and Substrait. Cargo feature unification made this a DB-wide change: ordinary JSON objects switched from sorted-key to insertion-order traversal/serialization, affecting external-table column discovery and JSON output. We should not accept that behavior change just to preserve order in two dependency-specific formats.

**Fix:** Removed `preserve_order` from both forks and kept ordering local:
- **DataFusion PG JSON EXPLAIN:** serialize fields directly with `SerializeMap`; use local ordered storage for metrics/extras. Do not round-trip through `Value::Object`, which would lose the intended order without the feature.
- **Substrait raw Type objects:** use a local `IndexMap<String, Value>` to preserve named-field/ordinal order.
- **GreptimeDB:** restore the four existing expectations for external-table columns, Parquet Map JSON, Prometheus native histograms, and OTLP histograms—not change assertions to accept the regression.

This restores ordinary DB JSON behavior while preserving order only where the two formats require it. Pins: [DataFusion `b66dca326`](https://github.com/GreptimeTeam/datafusion/commit/b66dca3260c8e04f314c8036cdfbb08f1a564d95), [Substrait `91ec978`](https://github.com/GreptimeTeam/substrait-rs/commit/91ec978b0649417ad3da8390e7a515baec723b1b).

**Review focus:**
1. Both feature enablers must be removed; check the resolved feature graph, not just one manifest.
2. PG EXPLAIN must retain ordering through final serialization, including metric duplicate handling (first position, last value).
3. Substrait's `Type::Object` payload type changes to `IndexMap`: this is a Rust source-API change, even though the protobuf/spec version is unchanged.
4. External schema inference regains sorted-key traversal **within each record**; columns across records still accumulate by first discovery. This does not make the whole schema alphabetically sorted.

**Evidence:** The checked cmd + vector_index normal/build feature graph has no `serde_json/preserve_order`. Substrait order/roundtrip/error tests, PG JSON tests with the feature both off and on, and the affected DB assertions passed. [Compatibility CI](https://github.com/GreptimeTeam/greptimedb/actions/runs/34802788832/job/103851733981) passed at `f4b853ded75`; validation of the subsequent main sync is separate and still in progress.

src/servers/src/postgres/types.rs
```diff
@@ -78,6 +83,39 @@ pub(super) fn schema_to_pg(
...
+/// The catalog exposes static catalog aliases as `Utf8` and dynamic aliases
+/// as `Int32`, so this must run before the ordinary `INT4`/`VARCHAR` fallbacks.
+/// The catalog crate only exposes named constants for the aliases it uses in
+/// dynamic catalog tables; keep the remaining aliases here for compatibility
+/// with its public `OID_ALIAS_TYPE_NAMES` contract. See
+/// https://github.com/datafusion-contrib/datafusion-postgres/issues/384.
+fn pg_oid_alias_type(column: &datatypes::schema::ColumnSchema) -> Option<Type> {
+    if !matches!(
+        &column.data_type,
+        ConcreteDataType::Int32(_) | ConcreteDataType::String(_)
```
killme2008, line 101: String-backed aliases also need binary encoding support. For example, `SELECT typreceive FROM pg_catalog.pg_type WHERE oid = 16` now reports REGPROC, but with binary results arrow-pg still sends the 8 UTF-8 bytes of `boolrecv`, rather than a 4-byte OID. Clients decoding it as REGPROC will fail. Please encode the actual OID for binary results and add a test; the new schema test only covers text format.

src/table/src/table/scan.rs
```diff
@@ -407,6 +403,17 @@ impl ExecutionPlan for RegionScanExec {
         vec![]
     }
 
+    fn apply_expressions(
+        &self,
+        f: &mut dyn FnMut(&Arc<dyn PhysicalExpr>) -> datafusion_common::Result<TreeNodeRecursion>,
+    ) -> datafusion_common::Result<TreeNodeRecursion> {
+        self.output_ordering
```
killme2008: Please visit `pushed_dyn_filters` here too. DF55 uses `apply_expressions()` to find dynamic filter consumers. When RegionScan is the only consumer, omitting these expressions makes AggregateExec drop its MIN/MAX filter and HashJoinExec skip filter generation. The scan accepts the filter, but its producer stops updating it. The existing pushdown test only checks that the scanner received a filter, so it misses this regression.

killme2008:
FYI, there is a CI failure

https://github.com/GreptimeTeam/greptimedb/actions/runs/34826861352/job/103920913163?pr=8555

sunng87:
The ci failure is unrelated. Just ignore it for this patch

src/servers/src/postgres/types.rs
discord9: Fixed in `0984a41ead8`, with the protocol regression and its catalog fixture in `4932b1d9a6b` / `81babd86afc` (included in current head `9c063a0594f`).

**What was wrong:** We updated the advertised alias type but left the string-backed value on the ordinary string encoder path. The previous schema test did not verify binary payloads.

**What changed:** The server now uses a small alias value adapter. Its binary serializer resolves `regproc` names against the same embedded `pg_proc` dataset and writes a four-byte big-endian OID; its text serializer preserves the original name. Resolution happens inside binary serialization, not based on `FieldInfo` format, because COPY selects its own serializer. `-` maps to zero; numeric strings support the unsigned 32-bit range for all advertised aliases. Unknown or ambiguous names fail binary encoding rather than emitting string bytes or an arbitrary OID. The one-time catalog lookup is fallible, not a request-path `expect`.

**Verification:** The real PostgreSQL protocol test runs your exact query, `SELECT typreceive FROM pg_catalog.pg_type WHERE oid = 16`: prepared/binary output advertises REGPROC and decodes exactly four bytes as OID **2436** (`00 00 09 84`); simple/text output remains **`boolrecv`**. Encoder tests also cover mixed formats, NULL, zero, ordinary VARCHAR, numeric aliases, unknown/ambiguous names, and COPY text. All **563 servers tests** and strict Clippy passed locally. Current-head [Rust CI](https://github.com/GreptimeTeam/greptimedb/actions/runs/34855679661) and [Integration CI, including actual compatibility](https://github.com/GreptimeTeam/greptimedb/actions/runs/34855680225) passed.

**Review boundary:** Named binary lookup is limited to static REGPROC values; this is not a general PostgreSQL name resolver for every alias. Logical catalog strings and their text behavior are unchanged. Leaving the thread open for your confirmation.

src/table/src/table/scan.rs
discord9: Fixed in `a801bdae0ed` (included in current head `9c063a0594f`).

**What was wrong:** `RegionScanExec::apply_expressions()` exposed only ordering expressions. A filter could be accepted by the scanner yet remain invisible to DF55's expression-ID search, causing Aggregate/HashJoin to stop producing it. The old acceptance-only test did not cover that link.

**What changed:** Expression traversal now includes the existing `pushed_dyn_filters` wrappers as well as ordering roots. It snapshots their Arcs and releases the mutex before invoking callbacks; it does not copy the current predicate or create replacement filter state. The existing traversal helper preserves Stop/error behavior.

**Verification:** The new regression installs a filter through the actual pushdown path and verifies that traversal exposes the same dynamic wrapper and expression ID, that the mutex is not held during the callback, and that Stop propagates. Existing live-update/reset tests remain intact. All **87 table tests** and strict Clippy passed locally. This directly tests the missing discovery link; it is not a claim that the new unit test runs every aggregate/join optimizer path.

The adjacent contract check also found and fixed two omissions in `0f5ff87a919`: PartSort now reports its existing produced dynamic filter, and RangeSelect visits its owned aggregate arguments/ORDER BY and grouping roots. Their focused suite passed **61 tests**, with strict Clippy passing. No extra pushdown pass or reset mechanism was added.

Current-head [Rust CI](https://github.com/GreptimeTeam/greptimedb/actions/runs/34855679661) and [Integration CI](https://github.com/GreptimeTeam/greptimedb/actions/runs/34855680225) passed. Leaving the thread open for your confirmation.

PR #8566, opened 2026-07-20
feat: grant creators access to newly created databases

shuiyisong:

This pull request adds support for automatically granting authenticated users access to databases they create.

It securely propagates and validates creator context, atomically commits schema metadata with the creator grant, and makes retries and procedure recovery idempotent. Legacy clients remain compatible, while `SchemaNameValue` gains an optional, serde-defaulted procedure identifier.

## PR Checklist

### DIFF
--- a/src/cmd/src/standalone.rs
+++ b/src/cmd/src/standalone.rs
@@ -556,13 +556,11 @@ impl StartCommand {
             create_database_metadata_committer: None,
         };
 
-        let ddl_manager = DdlManager::try_new(
+        let ddl_manager = DdlManager::new(
             ddl_context,
             procedure_manager.clone(),
             Arc::new(StandaloneRepartitionProcedureFactory),
-            false,
-        )
-        .context(error::InitDdlManagerSnafu)?;
+        );
 
         let ddl_manager = if let Some(configurator) =
             plugins.get::<DdlManagerConfiguratorRef<DdlManagerConfigureContext>>()
--- a/src/cmd/src/standalone.rs
+++ b/src/cmd/src/standalone.rs
@@ -553,13 +553,14 @@ impl StartCommand {
             region_failure_detector_controller: Arc::new(NoopRegionFailureDetectorControl),
             soft_drop_enabled: false,
             soft_drop_retention: None,
+            create_database_metadata_committer: None,
         };
 
         let ddl_manager = DdlManager::try_new(
             ddl_context,
             procedure_manager.clone(),
             Arc::new(StandaloneRepartitionProcedureFactory),
-            true,
+            false,
         )
         .context(error::InitDdlManagerSnafu)?;
 
@@ -578,6 +579,9 @@ impl StartCommand {
         } else {
             ddl_manager
         };
+        ddl_manager
+            .register_loaders()
+            .context(error::InitDdlManagerSnafu)?;
 
         let procedure_executor = creator
             .procedure_executor_creator
--- a/src/common/meta/src/ddl/create_database.rs
+++ b/src/common/meta/src/ddl/create_database.rs
@@ -13,10 +13,12 @@
 // limitations under the License.
 
 use std::collections::HashMap;
+use std::sync::Arc;
 
 use async_trait::async_trait;
+use common_error::ext::{BoxedError, ErrorExt};
 use common_procedure::error::{FromJsonSnafu, Result as ProcedureResult, ToJsonSnafu};
-use common_procedure::{Context as ProcedureContext, LockKey, Procedure, Status};
+use common_procedure::{Context as ProcedureContext, LockKey, Procedure, ProcedureId, Status};
 use serde::{Deserialize, Serialize};
 use serde_with::{DefaultOnNull, serde_as};
 use snafu::{ResultExt, ensure};
@@ -25,8 +27,38 @@ use strum::AsRefStr;
 use crate::ddl::DdlContext;
 use crate::ddl::utils::map_to_procedure_error;
 use crate::error::{self, Result};
+use crate::instruction::{CacheIdent, UserCacheIdent};
 use crate::key::schema_name::{SchemaNameKey, SchemaNameValue};
 use crate::lock_key::{CatalogLock, SchemaLock};
+use crate::rpc::ddl::CreatorGrantIntent;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AtomicCreateAccess {
+    AddedExactAll,
+    AlreadyEffectiveAll,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AtomicCreateOutcome {
+    Created(AtomicCreateAccess),
+    AlreadyCommitted,
+    UserChanged,
+    UserMissing,
+    SchemaConflict,
+}
+
+#[async_trait]
+pub trait CreateDatabaseMetadataCommitter: Send + Sync {
+    async fn commit(
+        &self,
+        catalog: &str,
+        schema: &str,
+        value: &SchemaNameValue,
+        creator: &CreatorGrantIntent,
+    ) -> std::result::Result<AtomicCreateOutcome, BoxedError>;
+}
+
+pub type CreateDatabaseMetadataCommitterRef = Arc<dyn CreateDatabaseMetadataCommitter>;
 
 pub struct CreateDatabaseProcedure {
     pub context: DdlContext,
@@ -41,6 +73,7 @@ impl CreateDatabaseProcedure {
         schema: String,
         create_if_not_exists: bool,
         options: HashMap<String, String>,
+        creator: Option<CreatorGrantIntent>,
         context: DdlContext,
     ) -> Self {
         Self {
@@ -51,6 +84,7 @@ impl CreateDatabaseProcedure {
                 schema,
                 create_if_not_exists,
                 options,
+                creator,
             },
         }
     }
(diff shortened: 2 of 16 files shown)


src/common/meta/src/rpc/ddl.rs
```diff
@@ -398,9 +433,9 @@ impl TryFrom<SubmitDdlTaskRequest> for PbDdlTaskRequest {
 
         Ok(Self {
             header: None,
-            query_context: Some(request.query_context.into()),
-            timeout_secs: request.timeout.as_secs() as u32,
-            wait: request.wait,
+            query_context: Some(query_context.into()),
+            timeout_secs: timeout.as_secs() as u32,
```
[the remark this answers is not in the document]
shuiyisong: This is an existing issue cc @WenyXu please see if we need to fix this one in the PR

src/operator/src/statement/ddl.rs
```diff
@@ -2150,16 +2158,27 @@ impl StatementExecutor {
...
         } else {
-            error::SchemaExistsSnafu { name: database }.fail()
+            return error::SchemaExistsSnafu { name: database }.fail();
+        };
+
+        #[cfg(feature = "enterprise")]
+        if let Some(handler) = &self.create_database_handler {
+            handler
+                .refresh_current_user(&query_context)
+                .await
```
sunng87, line 2177: 🟠 **A successful commit should not fail the user-visible DDL.** If `refresh_current_user` errors here, the database (and grant) are already durably committed — the procedure returned and `output` is computed — yet the client receives an error for a CREATE that actually succeeded.

Since meta-srv already invalidates the creator cache via the `InvalidateCreatorCache` procedure state, this frontend refresh is a best-effort optimization; its failure shouldn't fail the DDL. Suggest logging + continuing:

```rust
if let Err(e) = handler.refresh_current_user(&query_context).await {
    warn!("refresh_current_user after create database failed: {e}");
}
```

src/common/meta/src/ddl/create_database.rs
```diff
@@ -85,8 +134,74 @@ impl CreateDatabaseProcedure {
...
+            // Distinguishes a replay of this procedure from a competing create.
+            value.create_procedure_id = Some(procedure_id.to_string());
+            let Some(committer) = self.context.create_database_metadata_committer.as_ref() else {
+                return error::UnexpectedSnafu {
+                    err_msg: "Create-database creator grant committer is not configured"
+                        .to_string(),
+                }
+                .fail();
+            };
+            let _timer = crate::metrics::METRIC_META_CREATE_SCHEMA.start_timer();
```
sunng87, line 150: 🟠 **Possible metric double-counting.** `SchemaManager::create` (`schema_name.rs`) also starts the `METRIC_META_CREATE_SCHEMA` histogram timer around its conditional put. If the enterprise committer delegates to `schema_manager.create` internally (likely, for the atomic metadata write), this histogram is recorded twice per create.

The counter is correctly gated to `Created` only, so only the histogram observation is affected. Worth verifying in the committer impl — if it reuses `schema_manager.create`, drop this outer timer (or move ownership of the metric fully into the committer).

src/common/meta/src/key/schema_name.rs
```diff
@@ -60,6 +60,9 @@ pub struct SchemaNameValue {
     pub ttl: Option<DatabaseTimeToLive>,
     #[serde(default)]
     pub extra_options: BTreeMap<String, String>,
+    /// Identifies the create-database procedure that wrote this value.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub create_procedure_id: Option<String>,
```
sunng87, line 65: 🟡 **New persisted field — add a read-compat test.** The serde `default` + absence of `deny_unknown_fields` makes this backward compatible for reads (old stored values → `None`), and the PR checklist currently leaves "Schema/data changes are backward compatible" unchecked.

Per the repo's architecture-invariants guidance, consider adding an explicit test deserializing an *old-shape* stored value (JSON without `create_procedure_id`) into `SchemaNameValue`, to lock in the read-compat guarantee. The existing `test_create_procedure_id_serialization` covers roundtrip but not old-data migration.

src/operator/src/statement/ddl.rs
```diff
@@ -2138,7 +2138,15 @@ impl StatementExecutor {
             }
         );
 
-        if !self
+        #[cfg(feature = "enterprise")]
+        let creator = match &self.create_database_handler {
```
sunng87, line 2142: 🟡 `creator` is resolved before the `schema_exists` check, so `handler.creator()` now runs even on the `IF NOT EXISTS` / already-exists branches. If `creator()` errors for an unauthenticated session it will block `CREATE DATABASE IF NOT EXISTS <existing>` that previously succeeded.

Probably acceptable (unauthenticated sessions arguably shouldn't run DDL), but flagging the behavior change — consider resolving `creator` only inside the actual create branch, or noting that auth is now required even on the no-op path.

src/common/meta/src/ddl/create_database.rs
```diff
@@ -85,8 +134,74 @@ impl CreateDatabaseProcedure {
...
+                AtomicCreateOutcome::UserMissing => {
+                    return error::UnexpectedSnafu {
+                        err_msg: format!(
+                            "Creator '{}' no longer exists in catalog '{}'",
+                            creator.username, self.data.catalog
+                        ),
+                    }
+                    .fail();
+                }
+                AtomicCreateOutcome::SchemaConflict if self.data.create_if_not_exists => {
```
sunng87, line 189: Nit: worth a one-line comment here. Returning `done()` without granting is correct (a grant is only owed when *this* creator actually created the DB), but a future reader could mistake "schema already exists, do nothing" for a missing-grant bug.

sunng87 (commented):
Reviewed independently. Strong design — the idempotency model (`create_procedure_id` → `AlreadyCommitted` vs `SchemaConflict`) and the dual-channel, role-checked creator transport are well thought out, and the recovery tests (`lost_commit_response_replays_as_already_committed`, `lost_state_transition_recovers_committed_business_state`) cover the two hard cases directly. CI is green incl. `fuzz_create_database`.

Inline comments below. The one item I'd ask to address before merge is the `refresh_current_user` hardening (a durably-committed DDL should not surface an error to the client); the rest are verification / nit items.

A few questions:
1. For an enterprise admin/root creating a DB, does `handler.creator()` return `Some` or `None`? (determines whether admins self-grant)
2. Does the enterprise committer wrap `SchemaManager::create` internally? (resolves the metric double-count question)
3. `UserMissing` fails permanently with no schema created — intended (deleted creator ⇒ hard-fail CREATE) vs. create-without-grant?

WenyXu:
@codex review

src/common/meta/src/rpc/ddl.rs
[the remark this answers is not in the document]
WenyXu: Good catch

src/operator/src/statement/ddl.rs
shuiyisong: Keeping the current behavior. `refresh_current_user` is part of same-session completion, not only cache optimization: later statements in the same request reuse `query_ctx.current_user()`, and the refresh also rejects a deleted/recreated account via `created_at_ns`. Swallowing every failure could let those statements continue with stale identity. A post-commit client false negative is an intentional tradeoff here; retrying `CREATE DATABASE IF NOT EXISTS` performs the refresh again.

src/common/meta/src/ddl/create_database.rs
shuiyisong: Verified against the companion enterprise implementation: `KvCreateDatabaseMetadataCommitter::commit_inner` writes the schema and user rows directly with `kv_backend.txn`; it does not call `SchemaManager::create`. The outer timer therefore records the atomic creator path once, while `SchemaManager` owns the legacy creator-less path. No change needed.

src/common/meta/src/key/schema_name.rs
shuiyisong: The requested coverage already exists in `test_backward_compatibility_with_old_format`: it deserializes `{"ttl":"10s"}` without `create_procedure_id` and asserts `create_procedure_id: None`. `cargo test -p common-meta test_backward_compatibility_with_old_format --lib` passes, so no additional test is needed.

src/operator/src/statement/ddl.rs
shuiyisong: Keeping the current order. The handler is installed only when the built-in EE user provider is active, so unauthenticated and non-EE paths have no handler to call. For built-in users, applying the creator checks to the `IF NOT EXISTS` no-op path keeps authorization consistent, and the unconditional post-success refresh is intentional recovery behavior. No code change needed.

src/common/meta/src/ddl/create_database.rs
shuiyisong: No change needed. `AtomicCreateOutcome::SchemaConflict` already documents that a different schema value exists, and the `if self.data.create_if_not_exists => Status::done()` match arm makes the no-op rule explicit. Another comment would duplicate the code.

src/common/meta/src/rpc/ddl.rs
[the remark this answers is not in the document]
shuiyisong: Fixed in https://github.com/GreptimeTeam/greptimedb/pull/8566/changes/8c0c472b248c4c3ce7a820df6d076b34b4333964

src/cmd/src/standalone.rs
```diff
@@ -554,13 +553,14 @@ impl StartCommand {
...
             soft_drop_retention: None,
+            create_database_metadata_committer: None,
         };
 
         let ddl_manager = DdlManager::try_new(
             ddl_context,
             procedure_manager.clone(),
             Arc::new(StandaloneRepartitionProcedureFactory),
-            true,
+            false,
```
WenyXu: Why change this

src/common/meta/src/ddl/create_database.rs
```diff
@@ -100,6 +215,33 @@ impl CreateDatabaseProcedure {
...
+            )),
+        };
+        if let Err(source) = self
+            .context
+            .cache_invalidator
+            .invalidate(&ctx, &[ident])
+            .await
+        {
+            return Err(error::Error::retry_later(source));
+        }
```
WenyXu, line 241: nit: we can ignore the failure of cache invalidation.

sunng87 (approved):

src/cmd/src/standalone.rs
shuiyisong: https://github.com/GreptimeTeam/greptimedb/pull/8566/changes/BASE..8c0c472b248c4c3ce7a820df6d076b34b4333964#diff-f07d053e8724626739765610f16f842addf529404ec219c5d2f63775159c92ecR583
Loader is registered later. The gap is used for injection of the downstream

src/common/meta/src/ddl/create_database.rs
shuiyisong: Keeping this retryable. The metadata and ACL transaction is already durable when this state runs, so a transient invalidation failure can be retried without repeating the commit. Ignoring it could leave other frontend/user caches stale for their TTL and deny the creator access from another session or node. `created_and_recovered_commits_retry_only_cache_invalidation` covers that only invalidation is retried.

PR #8576, opened 2026-07-21
fix: preserve leader cache configuration and record cache hits

WenyXu:

This PR fixes two leader-cache issues in MetaSrv:

- Preserve the `enable_leader_cache` argument when `become_leader` is called while the previous state is already `Leader`. Previously, the cloned leader state retained its old cache setting.
- Record `batch_get` cache hits before returning on a full cache hit. Previously, fully cached requests bypassed the cache-hit metric increment.

This change does not alter public APIs, persisted data, wire formats, or schemas.

## PR Checklist


WenyXu:
@codex review

evenyag (approved):

PR #8580, opened 2026-07-21
fix(mysql): fail closed on unrepresentable timestamps

discord9:

N/A

MySQL result encoding could silently represent an out-of-Chrono-range, non-NULL Arrow timestamp as SQL NULL. This changes timestamp result serialization to fail closed with a client-visible timestamp-overflow error instead.

Timestamp conversion and precision formatting are staged for every timestamp column before the first column of a row is written. This keeps result packets synchronized when conversion fails after an earlier projected column, while preserving exact text-protocol fractional precision and binary-protocol datetime encoding for representable values. Actual Arrow NULL values continue to encode as SQL NULL.

The integration coverage exercises ordinary text queries and real binary prepared-statement execution, exact in-range and NULL controls, same-connection recovery after an overflow error, and multiple timestamp slots across mixed NULL/non-NULL rows.

This does not change an API, persisted format, schema, or valid timestamp representation.

## PR Checklist

### DIFF
--- a/src/servers/src/mysql/writer.rs
+++ b/src/servers/src/mysql/writer.rs
@@ -23,6 +23,7 @@ use arrow::datatypes::{
     UInt16Type, UInt32Type, UInt64Type,
 };
 use arrow_schema::{DataType, IntervalUnit};
+use chrono::NaiveDateTime;
 use common_decimal::Decimal128;
 use common_error::ext::ErrorExt;
 use common_error::status_code::StatusCode;
@@ -46,7 +47,10 @@ use session::context::QueryContextRef;
 use snafu::prelude::*;
 use tokio::io::AsyncWrite;
 
-use crate::error::{self, ConvertSqlValueSnafu, DataFusionSnafu, NotSupportedSnafu, Result};
+use crate::error::{
+    self, ConvertSqlValueSnafu, DataFusionSnafu, InternalSnafu, NotSupportedSnafu, Result,
+    TimestampOverflowSnafu,
+};
 use crate::metrics::*;
 
 /// Try to write multiple output to the writer if possible.
@@ -174,6 +178,20 @@ struct PrecisionTimestamp<'a> {
     datetime: chrono::NaiveDateTime,
 }
 
+struct StagedTimestamp {
+    datetime: Option<NaiveDateTime>,
+    formatted: String,
+}
+
+impl StagedTimestamp {
+    fn new() -> Self {
+        Self {
+            datetime: None,
+            formatted: String::with_capacity(32),
+        }
+    }
+}
+
 impl<'a> ToMysqlValue for PrecisionTimestamp<'a> {
     fn to_mysql_text<W: std::io::Write>(&self, w: &mut W) -> io::Result<()> {
         self.formatted.to_mysql_text(w)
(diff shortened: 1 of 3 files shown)


src/servers/src/mysql/writer.rs
```diff
@@ -209,10 +227,59 @@ impl MysqlResultWriter {
...
+                        staged_timestamps.get_mut(slot).context(InternalSnafu {
+                            err_msg: "timestamp staging slot is missing",
+                        })?;
+                    staged_timestamp.datetime = None;
+                    staged_timestamp.formatted.clear();
+                    if !column.is_null(i) {
+                        let timestamp = datatypes::arrow_array::timestamp_array_value(column, i);
+                        let datetime = timestamp
+                            .to_chrono_datetime_with_timezone(Some(&query_context.timezone()))
+                            .context(TimestampOverflowSnafu {
```
v0y4g3r:
```suggestion
                            .with_context(|| TimestampOverflowSnafu {
```

src/servers/src/mysql/writer.rs
```diff
@@ -209,10 +227,59 @@ impl MysqlResultWriter {
...
+            .enumerate()
+            .filter_map(|(index, column)| {
+                matches!(column.data_type(), DataType::Timestamp(_, _)).then_some(index)
+            })
+            .collect::<Vec<_>>();
+        let mut staged_timestamps = Vec::with_capacity(timestamp_indexes.len());
+        staged_timestamps.resize_with(timestamp_indexes.len(), StagedTimestamp::new);
         for i in 0..record_batch.num_rows() {
+            if !timestamp_indexes.is_empty() {
+                for (slot, &column_index) in timestamp_indexes.iter().enumerate() {
```
v0y4g3r: Inside the row loop, `record_batch.columns().get(column_index)` re-resolves each timestamp column every row, and the write arm does `timestamp_indexes.binary_search(&j)` for every timestamp cell. Both are loop-invariant and can be hoisted/precomputed (column refs resolved once; a `Vec<Option<usize>>` slot map for O(1) lookup).

src/servers/src/mysql/writer.rs
```diff
@@ -209,10 +227,59 @@ impl MysqlResultWriter {
...
+        let timestamp_indexes = record_batch
+            .columns()
+            .iter()
+            .enumerate()
+            .filter_map(|(index, column)| {
+                matches!(column.data_type(), DataType::Timestamp(_, _)).then_some(index)
+            })
+            .collect::<Vec<_>>();
+        let mut staged_timestamps = Vec::with_capacity(timestamp_indexes.len());
+        staged_timestamps.resize_with(timestamp_indexes.len(), StagedTimestamp::new);
```
v0y4g3r: nit: 
```suggestion
        let mut staged_timestamps = (0..n).map(|_| StagedTimestamp::new()).collect())
```

src/servers/src/mysql/writer.rs
```diff
@@ -209,10 +227,59 @@ impl MysqlResultWriter {
...
+                    let staged_timestamp =
+                        staged_timestamps.get_mut(slot).context(InternalSnafu {
+                            err_msg: "timestamp staging slot is missing",
+                        })?;
+                    staged_timestamp.datetime = None;
+                    staged_timestamp.formatted.clear();
+                    if !column.is_null(i) {
+                        let timestamp = datatypes::arrow_array::timestamp_array_value(column, i);
+                        let datetime = timestamp
+                            .to_chrono_datetime_with_timezone(Some(&query_context.timezone()))
```
fengjiachun: This only converts a `None` into `TimestampOverflow`, but `to_chrono_datetime_with_timezone` can panic when the session offset pushes a boundary timestamp outside `NaiveDateTime`. For example, `Timestamp::MAX_SECOND` with `SET time_zone = "+08:00"` panics in Chrono’s `naive_local()`, bypassing `finish_error` and dropping the connection. Could we use a checked timezone conversion here and add a same-connection recovery test for this boundary?

src/servers/src/mysql/writer.rs
```diff
@@ -209,10 +227,59 @@ impl MysqlResultWriter {
...
+                            "{}",
+                            datetime.format("%Y-%m-%d %H:%M:%S%.f")
+                        )
+                        .map_err(|_| {
+                            InternalSnafu {
+                                err_msg: "timestamp formatting failed",
+                            }
+                            .build()
+                        })?;
+                        staged_timestamp.datetime = Some(datetime);
```
fengjiachun: The staged value is considered valid as soon as Chrono can represent it, but the binary path later calls `NaiveDateTime::to_mysql_bin`, whose opensrv implementation writes `self.year() as u16`. I reproduced `+100000-01-01` being returned as year `34464` instead of an error. Could we validate the MySQL binary/date range before staging and cover an out-of-range year in the prepared-statement test so this fix also fails closed there?

src/servers/src/mysql/writer.rs
```diff
@@ -209,10 +227,59 @@ impl MysqlResultWriter {
...
+            .iter()
+            .enumerate()
+            .filter_map(|(index, column)| {
+                matches!(column.data_type(), DataType::Timestamp(_, _)).then_some(index)
+            })
+            .collect::<Vec<_>>();
+        let mut staged_timestamps = Vec::with_capacity(timestamp_indexes.len());
+        staged_timestamps.resize_with(timestamp_indexes.len(), StagedTimestamp::new);
         for i in 0..record_batch.num_rows() {
+            if !timestamp_indexes.is_empty() {
```
v0y4g3r: nit: this check is redundant that only increases indent level

src/servers/src/mysql/writer.rs
```diff
@@ -209,10 +227,59 @@ impl MysqlResultWriter {
...
+                        .map_err(|_| {
+                            InternalSnafu {
+                                err_msg: "timestamp formatting failed",
+                            }
+                            .build()
+                        })?;
+                        staged_timestamp.datetime = Some(datetime);
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
discord9: Addressed in aaea339f: timestamp columns are resolved once per batch into borrowed staging entries, with a precomputed column-to-slot map. The row loop no longer re-resolves columns.

src/servers/src/mysql/writer.rs
```diff
@@ -286,28 +353,24 @@ impl MysqlResultWriter {
...
-                            row_writer.write_col(&*format_buf)?;
-                        }
+                        let datetime = staged_timestamp.datetime.context(InternalSnafu {
+                            err_msg: "timestamp staging value is missing",
+                        })?;
+                        row_writer.write_col(PrecisionTimestamp {
+                            formatted: staged_timestamp.formatted.as_str(),
+                            datetime,
+                        })?;
                     }
```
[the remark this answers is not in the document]
discord9: Addressed in aaea339f: timestamp writes now use a guarded O(1) slot-map lookup instead of a binary search.

src/servers/src/mysql/writer.rs
discord9: Addressed in aaea339f: timestamp conversion now uses lazy with_context error construction.

src/servers/src/mysql/writer.rs
discord9: Addressed in aaea339f: staging entries are now constructed directly in one iterator pass, removing the capacity-plus-resize sequence.

src/servers/src/mysql/writer.rs
discord9: Addressed in aaea339f: column references are hoisted into borrowed staging entries and the write path uses a Vec<Option<usize>> slot map for O(1) lookup.

src/servers/src/mysql/writer.rs
discord9: Addressed in aaea339f: shared staging rejects local years outside MySQL DATETIME range 1000..=9999 before any row write. A prepared-protocol year-100000 regression verifies fail-closed behavior and same-connection recovery.

src/servers/src/mysql/writer.rs
discord9: Addressed in aaea339f: the redundant outer check was removed; the row loop now directly iterates the precomputed timestamp staging entries.

src/servers/src/mysql/writer.rs
discord9: Addressed in aaea339f: timezone conversion now resolves the offset at the UTC instant and applies it with checked_add_offset, returning overflow instead of panicking. Added a +08:00 real-connection regression that verifies the same connection remains usable.

fengjiachun (approved):
LGTM

PR #8582, opened 2026-07-21
feat(flow): handle time_ranges in DirtyWindowRequest

v0y4g3r:
## Summary

- Bump greptime-proto to include the new `time_ranges` field on `DirtyWindowRequest` (GreptimeTeam/greptime-proto#330). Note: the proto PR is still open; the rev currently pins its head commit `a618b744` and should be updated once it merges.
- Handle `DirtyWindowRequest::time_ranges` in `BatchingEngine::handle_mark_dirty_time_window`: each `[start_inclusive, end_exclusive)` range is aligned to time window boundaries (via the now-`pub(crate)` `DirtyTimeWindows::align_time_window`) and inserted as a dirty window with an explicit end, alongside the existing per-timestamp dirty marking.
- Flows without a `time_window_expr` fall back to `set_dirty()`, same as the timestamp path; invalid ranges (`end <= start`) are warned and skipped.
- Fix the `DirtyWindowRequest` construction site in `operator/bulk_insert.rs` for the new field.

## Checklist

- [x] `cargo check --workspace`
- [x] `cargo clippy -p flow -p operator --all-targets` (no warnings)
- [x] `cargo nextest run -p flow` (226 passed)

### DIFF
--- a/src/flow/src/batching_mode/engine.rs
+++ b/src/flow/src/batching_mode/engine.rs
@@ -236,13 +236,26 @@ impl BatchingEngine {
                             continue;
                         };
                         for timestamp in timestamps {
-                            let align_start = expr
-                                .eval(common_time::Timestamp::new(*timestamp, *unit))?
-                                .0
-                                .context(UnexpectedSnafu {
-                                    reason: "Failed to eval start value",
-                                })?;
-                            all_dirty_windows.insert(align_start);
+                            // On alignment failure, fall back to marking the
+                            // whole task dirty so the dirty signal is never
+                            // lost (the caller treats this RPC as accepted and
+                            // will not retry).
+                            match expr.eval(common_time::Timestamp::new(*timestamp, *unit)) {
+                                Ok((Some(align_start), _)) => {
+                                    all_dirty_windows.insert(align_start);
+                                }
+                                res => {
+                                    warn!(
+                                        "Failed to align dirty timestamp {} for flow id = {:?}: {:?}, marking the task fully dirty",
+                                        timestamp,
+                                        task.config.flow_id,
+                                        res.err().map(|e| e.to_string()).unwrap_or_else(|| {
+                                            "missing window lower bound".to_string()
+                                        })
+                                    );
+                                    is_dirty = true;
+                                }
+                            }
                         }
                         for time_range in time_ranges {
                             if time_range.end_exclusive <= time_range.start_inclusive {
@@ -252,12 +265,25 @@ impl BatchingEngine {
                                 );
                                 continue;
                             }
-                            let (align_start, align_end) = DirtyTimeWindows::align_time_window(
+                            match DirtyTimeWindows::align_time_window(
                                 common_time::Timestamp::new(time_range.start_inclusive, *unit),
                                 Some(common_time::Timestamp::new(time_range.end_exclusive, *unit)),
                                 expr,
-                            )?;
-                            all_dirty_ranges.push((align_start, align_end));
+                            ) {
+                                Ok((align_start, align_end)) => {
+                                    all_dirty_ranges.push((align_start, align_end));
+                                }
+                                Err(err) => {
+                                    warn!(
+                                        "Failed to align dirty time range [{}, {}) for flow id = {:?}: {}, marking the task fully dirty",
+                                        time_range.start_inclusive,
+                                        time_range.end_exclusive,
+                                        task.config.flow_id,
+                                        err
+                                    );
+                                    is_dirty = true;
+                                }
+                            }
                         }
                     }
                 }
--- a/src/flow/src/batching_mode/engine.rs
+++ b/src/flow/src/batching_mode/engine.rs
@@ -152,6 +152,11 @@ impl BatchingEngine {
             .collect()
     }
 
+    /// Mark dirty time windows for batching flows.
+    ///
+    /// Both `timestamps` and `time_ranges` (`[start_inclusive, end_exclusive)`)
+    /// in each `DirtyWindowRequest` are bare `i64`s interpreted in the source
+    /// table's time index column native unit, resolved via table metadata.
     pub async fn handle_mark_dirty_time_window(
         &self,
         reqs: DirtyWindowRequests,
@@ -1063,6 +1068,7 @@ mod tests {
     use common_meta::key::table_route::TableRouteValue;
     use common_meta::key::test_utils::new_test_table_info_with_name;
     use common_meta::kv_backend::memory::MemoryKvBackend;
+    use common_time::timestamp::TimeUnit;
     use query::options::QueryOptions;
     use session::context::QueryContext;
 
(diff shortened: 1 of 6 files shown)


src/flow/src/batching_mode/engine.rs
```diff
@@ -191,7 +194,7 @@ impl BatchingEngine {
                     .as_timestamp()
                     .unwrap()
                     .unit();
-                Some((table_name, (timestamps, time_index_unit)))
+                Some((table_name, (timestamps, time_ranges, time_index_unit)))
```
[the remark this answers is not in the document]
v0y4g3r: `group_by_table_name` is already wrapped in an `Arc` (see `let group_by_table_name = Arc::new(group_by_table_name);` right after the collect), so the per-task `.clone()` is already a cheap pointer increment. No change needed.

src/flow/src/batching_mode/engine.rs
```diff
@@ -219,13 +222,15 @@ impl BatchingEngine {
 
             let group_by_table_name = group_by_table_name.clone();
             let task = task.clone();
-
             let handle: JoinHandle<Result<(), Error>> = tokio::spawn(async move {
                 let src_table_names = &task.config.source_table_names;
                 let mut all_dirty_windows = HashSet::new();
+                let mut all_dirty_ranges = Vec::new();
```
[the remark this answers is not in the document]
v0y4g3r: Not applying this: `all_dirty_ranges` is only pushed to and then consumed once — there are no membership checks and no O(N²) behavior. Duplicate aligned ranges are harmless because `DirtyTimeWindows::add_window` inserts into a `BTreeMap` keyed by window start and unions the ends (`add_or_merge_window`), which is idempotent for duplicates. A `HashSet` would only save a few redundant B-tree insertions in the rare case of duplicate ranges in a single batch, at the cost of extra hashing. Keeping the `Vec` here.

src/flow/src/batching_mode/engine.rs
```diff
@@ -239,6 +244,21 @@ impl BatchingEngine {
...
+                                    time_range.start_inclusive, time_range.end_exclusive
+                                );
+                                continue;
+                            }
+                            let (align_start, align_end) = DirtyTimeWindows::align_time_window(
+                                common_time::Timestamp::new(time_range.start_inclusive, *unit),
+                                Some(common_time::Timestamp::new(time_range.end_exclusive, *unit)),
+                                expr,
+                            )?;
+                            all_dirty_ranges.push((align_start, align_end));
```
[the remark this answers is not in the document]
v0y4g3r: Not applying — see the reply on the companion comment above: `all_dirty_ranges` stays a `Vec` since duplicates are already deduplicated by `DirtyTimeWindows::add_window`/`add_or_merge_window`, and there are no membership checks on this collection.

src/flow/src/batching_mode/state.rs
```diff
@@ -769,8 +769,9 @@ impl DirtyTimeWindows {
         Ok(ret)
     }
 
-    fn align_time_window(
-        &self,
+    /// Align a time range `[start, end)`(end is optional and exclusive) to
+    /// time window boundaries defined by the time window expr.
```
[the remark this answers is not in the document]
v0y4g3r: Fixed in a5b38e2b53 by adding the missing space in the doc comment.

src/flow/src/batching_mode/engine.rs
```diff
@@ -239,6 +244,21 @@ impl BatchingEngine {
...
+                        }
```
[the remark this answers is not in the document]
v0y4g3r: Added in 2f6face4e1: `test_handle_mark_dirty_time_window_with_time_ranges` covers (1) a valid `[3s, 11s)` range being aligned to window boundaries `[0s, 15s)` and stored with an explicit end, and (2) empty (`start == end`) and reversed (`start > end`) ranges being skipped.

src/flow/src/batching_mode/engine.rs
```diff
@@ -249,6 +269,9 @@ impl BatchingEngine {
                 for timestamp in all_dirty_windows {
                     state.dirty_time_windows.add_window(timestamp, None);
                 }
+                for (start, end) in all_dirty_ranges {
+                    state.dirty_time_windows.add_window(start, end);
```
discord9, line 280: **Blocking:** Adding explicit ranges exposes a shrinking-union bug in `merge_dirty_time_windows()`. For `[0s, 15s)` followed by contained `[5s, 10s)`, the overlap branch assigns `prev_tw.1 = Some(cur_upper)`, producing `[0s, 10s)` and permanently missing `[10s, 15s)`. Please merge with `max(prev_upper, cur_upper)` and add nested-range plus timestamp+range tests.

discord9, line 280: **Blocking:** A bounded range crossing `expire_lower_bound` is currently dropped wholesale downstream because `merge_dirty_time_windows()` discards every entry whose `start < expire_lower_bound`. For example, `[0s, 15s)` with an expiry lower bound of `10s` loses the still-live `[10s, 15s)` suffix. Please drop only ranges whose end is at/before expiry and clip crossing ranges, with boundary tests.

src/flow/src/batching_mode/engine.rs
```diff
@@ -239,6 +244,21 @@ impl BatchingEngine {
...
+                                    "Ignoring invalid dirty time range with start_inclusive={} >= end_exclusive={}",
+                                    time_range.start_inclusive, time_range.end_exclusive
+                                );
+                                continue;
+                            }
+                            let (align_start, align_end) = DirtyTimeWindows::align_time_window(
+                                common_time::Timestamp::new(time_range.start_inclusive, *unit),
+                                Some(common_time::Timestamp::new(time_range.end_exclusive, *unit)),
+                                expr,
+                            )?;
```
discord9, line 266: An alignment error here aborts this task before any valid ranges accumulated for it are committed, but the join loop only logs the error and the RPC still returns `Ok(())`. A producer will treat the dirty notification as accepted and will not retry. Please propagate task failures to the caller, or use a conservative fallback that cannot lose the affected range.

src/flow/src/batching_mode/engine.rs
```diff
@@ -239,6 +244,21 @@ impl BatchingEngine {
...
+                        for time_range in time_ranges {
+                            if time_range.end_exclusive <= time_range.start_inclusive {
+                                warn!(
+                                    "Ignoring invalid dirty time range with start_inclusive={} >= end_exclusive={}",
+                                    time_range.start_inclusive, time_range.end_exclusive
+                                );
+                                continue;
+                            }
+                            let (align_start, align_end) = DirtyTimeWindows::align_time_window(
+                                common_time::Timestamp::new(time_range.start_inclusive, *unit),
```
discord9, line 263: Please make the wire contract explicit that `start_inclusive`/`end_exclusive` use the source table time-index column’s native unit. The proto currently carries bare `int64`s and no unit documentation, while this code interprets values using table metadata. Also add coverage for second/microsecond/nanosecond time-index units, not only milliseconds.

discord9 (changes requested):
The range API is the right shape, but these correctness issues can silently leave dirty windows unrecomputed. Please address them before merging.

src/flow/src/batching_mode/engine.rs
v0y4g3r: Confirmed and fixed in e80ee4abd3: the merge branch now keeps `max(prev_upper, cur_upper)` so a contained window (bounded `[5s, 10s)` inside `[0s, 15s)`, or an unbounded dirty window inside a bounded range) no longer shrinks the merged window. Added `test_merge_dirty_time_windows_with_bounded_ranges` covering nested bounded ranges, timestamp+range mixes, disjoint ranges, and overlapping unions.

v0y4g3r: Fixed in 21acb5e7f8: bounded ranges are now dropped only when their end is at/before `expire_lower_bound`, and crossing ranges are clipped to the bound (the caller aligns the expire bound to the time window boundary via `time_window_expr.eval`, so the clipped start stays aligned). Unbounded (`None`-end) windows keep the existing start-based drop behavior. Boundary tests added (end at/before expiry dropped, crossing clipped, start-at-boundary kept, unbounded behavior unchanged).

src/flow/src/batching_mode/engine.rs
v0y4g3r: Fixed in b06865bf3a using the conservative-fallback option: on eval/alignment failure (both the timestamp path and the new time-range path) we now log a warning and mark the whole task dirty via `set_dirty()` instead of aborting the per-task closure, so the affected data is always recomputed and the dirty signal is never silently lost while the RPC returns Ok.

src/flow/src/batching_mode/engine.rs
v0y4g3r: Code side done in 04935f11be: `handle_mark_dirty_time_window` now documents that `timestamps`/`time_ranges` are bare i64s interpreted in the source table time-index native unit, and `test_handle_mark_dirty_time_window_time_index_units` covers second/millisecond/microsecond/nanosecond tables (the same physical [3s, 11s) range aligning to [0s, 15s) in each unit). For the wire contract itself, the field-level documentation belongs to GreptimeTeam/greptime-proto#330 — I will add unit semantics to the `TimeRange`/`time_ranges` field comments there.

src/flow/src/batching_mode/engine.rs
v0y4g3r: Updated in 690d910425: I replaced the `set_dirty()` fallback with error propagation. `set_dirty()` is not a true full-repair marker for time-windowed flows (it becomes one epoch-start window), so alignment/eval failures now return an error through the join loop and the RPC caller can retry. Added `test_handle_mark_dirty_time_window_returns_error_on_alignment_failure` covering an invalid timestamp alignment path.

discord9 (approved):

PR #8586, opened 2026-07-21
feat(mito): add candidate series scanner

evenyag:

This PR adds the candidate-series discovery stage for metric series scans.

The new scanner reads only encoded primary keys from native memtable and SST ranges, applies supported tag predicates to those keys, locally deduplicates them, and uses DataFusion's streaming merge to produce globally sorted, deduplicated `(table_id, tsid)` candidates in bounded batches. This avoids reading full field data while discovering which metric series need to participate in the later data scan.

Supporting changes:

- add a primary-key-only Parquet row-group reader while preserving binary/dictionary decoding;
- expose reusable encoded-primary-key filters from reader prefilter plans;
- separate candidate-series entries from normal data entries in the range-result cache;
- add sparse primary-key decoding helpers and error variants;
- add unit tests and a benchmark for filtering, sorting, deduplication, decoding, and cache-key isolation;
- reject unsupported metadata and extension-range layouts so callers can retain the legacy fallback.

The change is internal to Mito's read path and does not alter public APIs, persisted schemas, or on-disk formats.

Validation:

- `cargo nextest run --retries 0 -p mito2 series_candidate` was attempted, but compilation stopped because the local disk ran out of space while rustc was writing build artifacts.

## PR Checklist

### DIFF
--- a/src/mito2/src/read/series_candidate.rs
+++ b/src/mito2/src/read/series_candidate.rs
@@ -49,9 +49,7 @@ use crate::read::range_cache::{
     build_candidate_range_cache_key, cache_flat_range_stream, cached_flat_range_stream,
 };
 use crate::read::scan_region::StreamContext;
-use crate::read::scan_util::{
-    PartitionMetrics, maybe_scan_flat_other_ranges, new_filter_metrics, scan_flat_mem_ranges,
-};
+use crate::read::scan_util::{PartitionMetrics, new_filter_metrics, scan_flat_mem_ranges};
 use crate::sst::parquet::DEFAULT_READ_BATCH_SIZE;
 use crate::sst::parquet::format::PrimaryKeyArray;
 use crate::sst::parquet::prefilter::{
@@ -85,6 +83,10 @@ pub(crate) struct SeriesCandidateScanner {
 
 #[allow(dead_code)]
 impl SeriesCandidateScanner {
+    /// Creates a candidate-series scanner for native memtable and SST ranges.
+    ///
+    /// Callers must fall back to the legacy series-scan path when the scan contains
+    /// extension ranges. Candidate-series discovery does not support other range types.
     pub(crate) fn try_new(
         stream_ctx: Arc<StreamContext>,
         partitions: Vec<Vec<PartitionRange>>,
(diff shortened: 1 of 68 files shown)


evenyag:
@codex review

src/mito2/src/read/series_candidate.rs
```diff
@@ -0,0 +1,733 @@
...
+        })
+        .collect();
+    let ordering = LexOrdering::new([PhysicalSortExpr {
+        expr: Arc::new(Column::new(PRIMARY_KEY_COLUMN_NAME, 0)),
+        options: SortOptions {
+            descending: false,
+            nulls_first: false,
+        },
+    }])
+    .unwrap();
```
v0y4g3r, line 484: Maybe add safety note for this unwrap

v0y4g3r (commented):
**Minor: metrics partition index collision between range merges and the final merge** (src/mito2/src/read/series_candidate.rs:169)

The range-level merges use the flattened `range_idx` (`0..all_ranges.len()`) as their metrics partition, but the final merge uses `self.partitions.len()`. Since `all_ranges` is the flattened `partitions`, `all_ranges.len() >= partitions.len()` — so whenever any partition contains more than one range, `partitions.len()` falls inside `0..all_ranges.len()` and the final merge shares a metrics partition index with one of the range merges.

E.g. 1 partition with 3 ranges: range merges take 0/1/2, and the final merge also lands on 1, so its `output_rows`/`elapsed_compute` get mixed with range 1's merge under the same partition label in the shared `ExecutionPlanMetricsSet`.

Suggest capturing `all_ranges.len()` before it's consumed and using that (or a sentinel like `usize::MAX`) for the final merge instead.

src/mito2/src/read/series_candidate.rs
```diff
@@ -0,0 +1,733 @@
...
+}
+
+impl SeriesCandidateRangeBuilder {
+    async fn build_range_stream(
+        &self,
+        part_range: PartitionRange,
+        partition_pruner: Arc<PartitionPruner>,
+        merge_partition: usize,
+    ) -> Result<BoxedRecordBatchStream> {
+        let cache_key = build_candidate_range_cache_key(&self.stream_ctx, &part_range);
```
v0y4g3r, line 194: nit: Instead of reusing `RangeScanCacheKey`, maybe we can add a tailored cache key for two tage scan that includes only tag filters and time range filters.

v0y4g3r (approved):
Almost LGTM

src/mito2/src/read/series_candidate.rs
evenyag: Keep using the old key to simplify the code.

PR #8587, opened 2026-07-21
perf(servers): optimize PromQL read conversion

lyang24:

the goal
exec-f4ff9908-8ddf-426e-bad2-90ad69dfaf3d

Previously, the conversion created a complete owned label vector for every row,
including cloning every label name and value. These temporary labels were then
used as keys in a BTreeMap to group samples belonging to the same time series.
For remote-read results containing many samples, most rows belong to an already
known series, so many of these allocations and copies were unnecessary.

The new implementation avoids constructing owned labels until a new time series
is actually found:

- It borrows label values directly from Arrow Utf8, LargeUtf8, and Utf8View
  arrays.
- It also handles the Dictionary<UInt32, Utf8> representation produced by the
  PromQL read path by resolving dictionary keys and borrowing the underlying
  Utf8 values, instead of materializing an owned string for every row.
- It correctly handles both NULL dictionary keys and NULL dictionary values.
- It uses the previously matched series as a fast path, which benefits query
  results where samples from the same series are adjacent.
- For non-contiguous series, it uses a hash index to find an existing series.
- It performs a full label comparison after each hash lookup, so hash
  collisions cannot merge unrelated series.
- It sorts the resulting time series by labels to preserve the ordering
  produced by the previous BTreeMap implementation.
- It retains the existing owned-string conversion fallback for non-string label
  columns and dictionary layouts not covered by the borrowed path.
- It returns an explicit error if a label column cannot be converted to a
  string, instead of silently omitting that column.

The tests cover non-contiguous series grouping, the adjacent-series fast path,
the previous output ordering, Utf8, LargeUtf8, Utf8View, and
Dictionary<UInt32, Utf8> labels, NULL labels, NULL dictionary keys and values,
and the non-string conversion fallback.

The intention is to reduce allocations, string copies, and tree lookups on the
Prometheus remote-read conversion path while preserving its externally visible
behavior. In particular, the optimized path now covers the dictionary-encoded
string labels normally returned by PromQL queries.

## PR Checklist

- [] I have written the necessary rustdoc comments.

### DIFF
--- a/src/servers/src/prom_store.rs
+++ b/src/servers/src/prom_store.rs
@@ -22,8 +22,10 @@ use std::hash::{Hash, Hasher};
 use api::prom_store::remote::label_matcher::Type as MatcherType;
 use api::prom_store::remote::{Label, Query, ReadRequest, Sample, TimeSeries, WriteRequest};
 use api::v1::RowInsertRequests;
-use arrow::array::{Array, AsArray, LargeStringArray, StringArray, StringViewArray};
-use arrow::datatypes::{Float64Type, TimestampMillisecondType};
+use arrow::array::{
+    Array, AsArray, DictionaryArray, LargeStringArray, StringArray, StringViewArray,
+};
+use arrow::datatypes::{Float64Type, TimestampMillisecondType, UInt32Type};
 use common_grpc::precision::Precision;
 use common_query::prelude::{greptime_timestamp, greptime_value};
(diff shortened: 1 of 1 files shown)


killme2008:
@v0y4g3r  Please take a look.

v0y4g3r (approved):
LGTM, very solid work!

PR #8590, opened 2026-07-22
fix(query): use physical partition types for metric route pruning

discord9:

N/A

Metric logical tables reuse their physical table's region routes. Normally, logical-table creation injects the physical partition columns that exist at creation time. However, a later physical-table repartition can introduce a new partition key without adding that column to existing logical schemas.

In that state, a logical query may still have a useful predicate on a shared partition key, but distributed planning previously built the route datatype map only from the logical schema. When a physical route expression also referenced the newly introduced key, normalization could not type that expression and conservatively fanned out to every region.

This PR:

- extends `PartitionRuleManager` with a lookup that returns both the resolved physical table ID and its route, while preserving the existing route-only API;
- keeps logical-table partition columns as the predicate-eligibility boundary;
- loads the physical table's complete partition-column datatypes for route-expression normalization;
- preserves logical table IDs in the selected `RegionId`s so Metric Engine still applies the required logical-table identity filter;
- falls back to all routed regions when physical metadata is missing or inconsistent, datatypes disagree for a shared predicate column, or any route lacks a partition expression;
- adds planner-level regression tests for exact pruning and conservative fallback behavior.

The change does not add persisted or wire metadata, alter logical schemas, or implement a logical-table data-presence index. Queries without a usable partition predicate retain the existing all-region behavior.

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/planner.rs
+++ b/src/query/src/dist_plan/planner.rs
@@ -31,10 +31,12 @@ use datafusion_common::tree_node::{TreeNode, TreeNodeRecursion, TreeNodeVisitor}
 use datafusion_common::{DataFusionError, TableReference};
 use datafusion_expr::{LogicalPlan, UserDefinedLogicalNode};
 use datafusion_physical_expr::{LexOrdering, PhysicalSortExpr};
+use datatypes::prelude::ConcreteDataType;
 use partition::manager::{PartitionRuleManagerRef, create_partitions_from_region_routes};
 use session::context::QueryContext;
 use snafu::{OptionExt, ResultExt};
 use store_api::storage::RegionId;
+use table::metadata::TableInfo;
 pub use table::metadata::TableType;
 use table::table::adapter::DfTableProviderAdapter;
 use table::table_name::TableName;
@@ -242,9 +244,9 @@ impl DistExtensionPlanner {
             })?;
 
         let table_info = table.table_info();
-        let physical_table_route = self
+        let (physical_table_id, physical_table_route) = self
             .partition_rule_manager
-            .find_physical_table_route(table_info.table_id())
+            .find_physical_table_route_with_id(table_info.table_id())
             .await
             .context(PartitionRuleManagerSnafu)?;
         let all_regions = physical_table_route
@@ -252,9 +254,17 @@ impl DistExtensionPlanner {
             .iter()
             .map(|r| RegionId::new(table_info.table_id(), r.region.id.region_number()))
             .collect::<Vec<_>>();
-        // Extract partition columns
-        let partition_columns: Vec<String> =
-            table_info.meta.partition_column_names().cloned().collect();
+        let Some(logical_partition_columns) = partition_column_types(&table_info) else {
+            debug!(
+                "DistExtensionPlanner: invalid partition metadata for table {}, using all regions: {:?}",
+                table_name, all_regions
+            );
+            return Ok(all_regions);
+        };
+        let partition_columns = logical_partition_columns
+            .iter()
+            .map(|(name, _)| name.clone())
+            .collect::<Vec<_>>();
         debug!(
             "DistExtensionPlanner: loaded table partition metadata, table: {}, table_id: {}, partition_key_indices: {:?}, partition_columns: {:?}, all_regions: {:?}",
             table_name,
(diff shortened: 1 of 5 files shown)


src/query/src/dist_plan/planner.rs
```diff
@@ -302,6 +297,70 @@ impl DistExtensionPlanner {
...
+        } else {
+            match self
+                .catalog_manager
+                .table_info_by_id(physical_table_id)
+                .await
+            {
+                Ok(Some(table_info)) if table_info.table_id() == physical_table_id => table_info,
+                Ok(Some(table_info)) => {
+                    debug!(
+                        "DistExtensionPlanner: physical table info id mismatch for table {}: expected {}, got {}, using all regions: {:?}",
```
killme2008: When does it happen? It seems impossible. It's over designed.

src/query/src/dist_plan/planner.rs
```diff
@@ -360,6 +429,24 @@ impl DistExtensionPlanner {
...
+    let mut names =
+        std::collections::HashSet::with_capacity(table_info.meta.partition_key_indices.len());
+    table_info
+        .meta
+        .partition_key_indices
+        .iter()
+        .map(|index| {
+            let column = column_schemas.get(*index)?;
+            if !names.insert(column.name.clone()) {
+                return None;
```
killme2008: It's unreachable as well.
If it happens, the metadata is corrupted.

src/query/src/dist_plan/planner.rs
```diff
@@ -320,6 +379,16 @@ impl DistExtensionPlanner {
         if partitions.is_empty() {
             return Ok(all_regions);
         }
+        if partitions
+            .iter()
+            .any(|partition| partition.partition_expr.is_none())
```
killme2008: Good catch, but it should be moved into `prune_regions` to prevent it from being missed in other calls.

WenyXu:
@codex review

src/query/src/dist_plan/planner.rs
discord9: Agreed. Duplicate names or invalid partition indices imply corrupt `TableMeta`. In `0e9f151491`, `partition_column_types` now relies on the existing `TableMeta::partition_columns()` invariant and returns the column/type pairs directly. The valid stale logical `{host}` / physical `{host,rack}` case and the route-datatype fallback remain covered.

src/query/src/dist_plan/planner.rs
discord9: Agreed. Moved this invariant into `ConstraintPruner::prune_regions` in `0e9f151491`. The pruner now returns every input RegionId if any partition expression is missing, instead of filtering out the unknown partition. Added a mixed `Some`/`None` unit test and retained the planner-level logical-RegionId regression.

src/query/src/dist_plan/planner.rs
discord9: Agreed. A returned `TableInfo` with a different ID would violate the `table_info_by_id` contract. Removed that branch in `0e9f151491` while retaining conservative fallback for `None` and lookup errors. This is separate from the expected logical-table ID differing from its resolved physical-table ID.

evenyag:
@codex review

src/query/src/dist_plan/region_pruner.rs
```diff
@@ -38,29 +38,32 @@ impl ConstraintPruner {
...
-            }
-        }
-        if all_partition_expressions.is_empty() {
-            return Ok(partitions.iter().map(|p| p.id).collect());
-        }
+        let Some(all_partition_expressions) = partitions
+            .iter()
+            .map(|partition| partition.partition_expr.clone())
+            .collect::<Option<Vec<_>>>()
+        else {
```
evenyag, line 55: Is it possible that some of the partition expressions are None?

src/query/src/dist_plan/planner.rs
```diff
@@ -360,6 +428,14 @@ impl DistExtensionPlanner {
     }
 }
 
+fn partition_column_types(table_info: &TableInfo) -> Vec<(String, ConcreteDataType)> {
+    table_info
+        .meta
+        .partition_columns()
```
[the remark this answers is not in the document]
discord9: Thanks for checking this. The panic observation is correct for an already-corrupt `TableMeta`, but an out-of-range `partition_key_indices` entry violates the `TableMeta` invariant and is not produced by the normal create/alter/repartition paths. This exact defensive check was removed in `0e9f151491` following the earlier human review. If corrupt table metadata needs to be recoverable, validation should be handled consistently at the metadata-loading boundary rather than only in this planner helper. I have also removed the stale PR-body claim that invalid indices fall back to all regions.

evenyag (approved):

src/query/src/dist_plan/region_pruner.rs
discord9: Yes. An empty route expression decodes to `None` for an unpartitioned table, and legacy `PartitionBound::Value`/`MaxValue` metadata also decodes to `None`. More importantly, an unpartitioned-to-partitioned repartition can temporarily expose a mixed route vector: new target routes have expressions while the retained default source route is still empty. The previous code filtered out only the `None` entries, which could omit an unknown region and cause false-negative pruning. Returning all regions when any expression is missing is therefore the conservative behavior, and keeping that invariant in `ConstraintPruner` protects every caller. The mixed `Some`/`None` case is covered by `test_missing_partition_expression_returns_all_regions`.

PR #8591, opened 2026-07-22
fix: bind Prom remote read schema per query

discord9:

N/A

Prometheus remote-read requests may contain multiple queries with independent database/schema selectors. Previously, the HTTP handler extracted the first selector across the entire request and mutated one shared query context, causing every later query to read from the first query's schema.

This PR:

- extracts database selectors per remote-read query instead of per request;
- resolves one canonical catalog/schema/table target for each query;
- uses the same ordered targets for cross-catalog validation, table-aware permission checks, and query execution;
- preserves request-schema fallback for queries without an equality selector; and
- adds focused unit and HTTP integration coverage for explicit and mixed-schema query ordering.

There are no persisted-format, wire-format, or user-facing API changes.

Focused validation:

- `CARGO_PROFILE_TEST_DEBUG=0 cargo nextest run -p servers --lib test_extract_schema_from_query`
- `CARGO_PROFILE_TEST_DEBUG=0 cargo nextest run -p frontend --lib test_resolve_remote_query_target`
- focused frontend permission-target unit tests
- `cargo nextest run -p tests-integration --test main integration_http_file_test::test_prometheus_remote_schema_labels`

## PR Checklist

### DIFF
--- a/src/servers/src/prom_store.rs
+++ b/src/servers/src/prom_store.rs
@@ -19,7 +19,7 @@ use std::collections::BTreeMap;
 use std::hash::{Hash, Hasher};
 
 use api::prom_store::remote::label_matcher::Type as MatcherType;
-use api::prom_store::remote::{Label, Query, ReadRequest, Sample, TimeSeries, WriteRequest};
+use api::prom_store::remote::{Label, Query, Sample, TimeSeries, WriteRequest};
 use api::v1::RowInsertRequests;
 use arrow::array::{Array, AsArray};
 use arrow::datatypes::{Float64Type, TimestampMillisecondType};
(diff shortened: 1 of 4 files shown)


evenyag:
@codex review

evenyag:
@codex review

src/servers/src/prom_store.rs
```diff
@@ -121,19 +121,15 @@ pub fn table_name(q: &Query) -> Result<String> {
...
-    None
+/// Extract database selector from a remote read query.
+pub fn extract_schema_from_query(query: &Query) -> Option<String> {
+    query
+        .matchers
+        .iter()
+        .find(|matcher| {
+            is_database_selection_label(&matcher.name) && matcher.r#type == MatcherType::Eq as i32
+        })
+        .map(|matcher| matcher.value.clone())
```
evenyag, line 132: Will it contain multiple schemas?

discord9: A `ReadRequest` can contain multiple `Query` entries, and those queries may select different schemas—that is the case this change fixes. A single `Query` still resolves to one schema/table target; this helper extracts that query’s equality selector. If one query contains multiple database-selector aliases, it preserves the existing first-match behavior rather than spanning schemas. Rejecting conflicting aliases would be a separate behavior change.

evenyag (approved):

PR #8573, opened 2026-07-21
feat(query): add typed RangeSelect partial wire contract

discord9:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

Stacked on #8571 and depends on GreptimeTeam/greptime-proto#329. This is the third part extracted from #8557. Please review commit `ff3ef83194`; parent commits will disappear from this PR diff as #8570 and #8571 merge.

## What's changed and what's your intention?

- Add a typed `RangeSelectPartialV1` wire contract for serializing only local Partial RangeSelect plans.
- Materialize canonical time, BY, and post-coercion aggregate argument slots without deduplicating repeated arguments.
- Encode and strictly validate alignment, aggregate kind/range, slot indexes, field contracts, and aggregate-state ABI version.
- Integrate Partial nodes with the default Substrait serializer while rejecting Complete and Final modes.
- Preserve qualified/aliased sources, dictionary-encoded BY inputs, non-millisecond public timestamps, and executable Partial output across a real Substrait round trip.
- Add golden V1 bytes plus unknown-field, unknown-payload, malformed-index, stale-metadata, ABI, and aggregate mismatch coverage.

The V1 aggregate-state ABI is explicitly versioned as `1`; unsupported versions are rejected. The protobuf change is additive field 6 in GreptimeTeam/greptime-proto#329.

This PR deliberately contains no feature option, lowering analyzer, DistPlanner placement, frontend configuration, SQL `SET` policy, or SQLness activation. Those belong to the final stacked PR. It remains Draft until the proto PR merges and this dependency is repinned to the canonical commit on proto `main`.

## PR Checklist

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [x] API changes are backward compatible.
- [x] Schema or data changes are backward compatible.

PR #8571, opened 2026-07-20
refactor(query): add local RangeSelect partial and final stages

discord9:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

Stacked on #8570. This is the second part extracted from #8557. Please review commit `a126440ab8`; the #8570 commit will disappear from this PR diff after the parent merges.

## What's changed and what's your intention?

- Introduce explicit local `Complete`, `Partial`, and `Final` RangeSelect execution modes without activating distributed pushdown.
- Split supported aggregate expressions into state-producing partial expressions and merge/final expressions for MIN, MAX, SUM, COUNT(column), AVG, and compound AVG.
- Reject unsupported DISTINCT, FILTER, ORDER BY, null-treatment, and non-column COUNT forms at the split boundary.
- Define deterministic internal state schemas, argument materialization, distribution requirements, partitioning, and emission behavior.
- Preserve the public timestamp and BY-column schemas while normalizing dictionary inputs for internal row conversion.
- Add execution coverage proving Complete equals two unequal Partials plus an empty Partial merged by Final, including null states, non-millisecond timestamps, dictionary BY values, and `FILL PREV` across a missing bucket.

This PR deliberately contains no wire format, serializer/lowering, feature option, DistPlanner, or SQLness changes. Those belong to later stacked PRs. It does not change persisted data or public API compatibility.

## PR Checklist

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [x] API changes are backward compatible.
- [x] Schema or data changes are backward compatible.

PR #8557, opened 2026-07-17
feat(query): push down RangeSelect partial aggregation

discord9:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

- Discussion: https://github.com/GreptimeTeam/greptimedb/discussions/8525
- Wire contract dependency: https://github.com/GreptimeTeam/greptime-proto/pull/329

## What's changed and what's your intention?

This PR adds a default-off experimental distributed RangeSelect optimization. Eligible range aggregates are split into remote Partial state production and frontend Final state merge:

```text
RangeSelectExec(Final)
  CoalescePartitionsExec
    MergeScanExec
      remote RangeSelectExec(Partial)
        ProjectionExec(materialized aggregate arguments)
```

Main changes:

- prune RangeSelect scan inputs to time, BY, aggregate, and filter dependencies;
- add Complete/Partial/Final logical and physical contracts using StateWrapper/MergeWrapper;
- materialize coerced aggregate arguments in a deterministic Projection;
- serialize only typed Partial V1 metadata through the greptime-proto contract;
- strictly validate and rebuild decoded Partial plans without sending DataFusion Expr bytes;
- classify only Partial as remotely commutative in DistPlanner;
- add operator-only, startup-time feature gating with mixed-version safety documentation;
- add real two-region MergeScan state-merge coverage and a dedicated frontend-enabled SQLness CI profile.

The feature remains disabled by default. Enabling it asserts that all reachable datanodes support `RangeSelectPartialV1` and a compatible aggregate-state ABI. Mixed-version rolling upgrades are not supported while enabled.

Validation completed:

- common-function aggregate-wrapper tests: 10 passed;
- RangeSelect plan tests: 57 passed;
- RangeSelect lowering tests: 5 passed;
- default serializer tests: 4 passed;
- DistPlanner analyzer tests: 64 passed, 1 ignored;
- operator option rejection test: passed;
- sqlness-runner: 35 passed;
- query checks, Rust/TOML formatting, generated config docs, and diff checks passed;
- dedicated feature-off/on distributed SQLness produced identical snapshots before the final fixture hardening;
- real Substrait TableScan roundtrip and in-process two-region MergeScan execution pass after the final qualifier fix.

Remaining Draft follow-up:

- merge greptime-proto PR #329 and repin this PR to the canonical main commit;
[...]

PR #8593, opened 2026-07-22
feat(query): push RangeSelect partial stages to datanodes

discord9:

Stacked on #8573 and depends transitively on #8571, #8570, and GreptimeTeam/greptime-proto#329. This is the final part extracted from #8557. Please review commit `3168552bde`; parent commits will disappear from this PR diff as the stack merges.

- Add the startup-only `experimental_enable_range_select_pushdown` frontend query option, disabled by default and rejected through SQL `SET`.
- Lower eligible Complete RangeSelect plans into typed Partial/Final stages only after the real Substrait serializer accepts the Partial. Unsupported or pre-dispatch encoding failures retain the Complete plan.
- Place only Partial stages below `MergeScan`; Complete and Final remain on the frontend.
- Propagate analyzer ordering and keep flownode execution disabled.
- Warn that every datanode reachable through routing, read preference, retries, and failover must support `RangeSelectPartialV1` and the same state ABI. Mixed-version operation is unsupported.
- Document that fallback is pre-dispatch only; Partial decode, planning, or execution failures after dispatch fail the query without a Complete retry.
- Add focused lowering/placement/fallback tests and an isolated experimental distributed SQLness case proving state-only Partial execution on two remote partitions with correct MIN/MAX/SUM/COUNT/AVG, compound expressions, and fill semantics.

The plan-sensitive SQLness case lives outside normal default-off cases and runs only in a Basic CI step with the frontend startup environment variable enabled. No runner or config-template plumbing is added.

## PR Checklist

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -795,6 +795,11 @@ jobs:
         run: tar -xvf ./bins.tar.gz
       - name: Run sqlness
         run: RUST_BACKTRACE=1 ./bins/sqlness-runner bare ${{ matrix.mode.opts }} -c ./tests/cases --bins-dir ./bins --preserve-state
+      - if: matrix.mode.name == 'Basic'
+        name: Run experimental RangeSelect pushdown sqlness
+        env:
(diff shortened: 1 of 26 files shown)


discord9:
Closing temporarily while review focuses on the parent layers. The validated branch and experimental SQLness coverage are preserved for reopening or recreation later.

PR #8594, opened 2026-07-22
fix(mito2): preserve referenced parquet in fast GC

discord9:

Found while reviewing the broader publication protocol in #8537. This PR replaces that approach with a smaller GC policy fix.

This PR fixes three GC safety issues without adding publication gates, persisted state, route epochs, or wire-format changes.

1. **Fast GC protects referenced Parquet files by `FileId`.**
   A Parquet file's identity does not include the Puffin index version. A temporary reference to `(file_id, v1)` must therefore protect the same Parquet object even when its removed-file record carries `v2`. Puffin files remain protected by the exact `(FileId, IndexVersion)` pair.

2. **Dropped-region GC fails closed when route/reference discovery is incomplete.**
   Table-route lookup failures and missing related destination leaders now return retryable errors instead of silently omitting regions. This prevents dropped-region full-list GC from continuing without all cross-region references needed after repartition. Existing dropped-primary route overrides remain unchanged.

3. **Active/open regions no longer delete unknown files.**
   A newly written flush or compaction output is visible in object storage before it is published to the manifest. Treating an old unknown object as deletable can therefore race with publication and leave the manifest referencing a deleted object. Full-list GC now:
   - retains unknown Parquet and Puffin objects for active/open regions;
   - continues deleting manifest-known removed files using the existing lingering/reference checks;
   - continues deleting unknown objects for dropped regions.

The tradeoff is intentional: a genuine unknown orphan in an active region can remain until the region is dropped. This removes the publication race with a small policy change instead of #8537's publication gate, object validation, additional context state, and lock protocol.

`unknown_file_lingering_time` remains accepted with the same default for configuration compatibility, but no longer affects active/dropped unknown-file decisions. The example and generated configuration documentation are updated accordingly.

No persisted manifest/WAL format, wire format, schema, or public request representation changes.

Verification:

- focused `mito2` tests covering fast Parquet references, active unknown retention through `LocalGcWorker::run()`, and physical dropped unknown deletion;
- focused `meta-srv` tests covering missing related leaders, missing route discovery, and logical routes with missing physical routes;
- `cargo check -p mito2 --features test`;
- `cargo check -p meta-srv --features mock`;
- `cargo fmt --all --check`;
- `taplo check --no-auto-config config/datanode.example.toml`;
- `git diff --check`.

## PR Checklist

### DIFF
--- a/config/config.md
+++ b/config/config.md
@@ -654,7 +654,7 @@
 | `region_engine.mito.gc` | -- | -- | -- |
 | `region_engine.mito.gc.enable` | Bool | `false` | Whether GC is enabled. Need to be the same with metasrv's `gc.enable` or unexpected behavior will occur |
 | `region_engine.mito.gc.lingering_time` | String | `1h` | Lingering time before deleting files.<br/>Should be long enough to allow long running queries to finish.<br/>If set to None, then unused files will be deleted immediately. |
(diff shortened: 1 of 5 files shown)


discord9:

PR #8613, opened 2026-07-23
feat: invoke gc hook for offline region cleanup

sunng87:

This patch adds region hook to soft-drop cleanup phase so we can remove extension files managed by the hook.

some open questions:

- this cleanup logic sits in `handle_open.rs`, which is a little confusing
- can we merge this cleanup with our current GC concept to reduce the complexity?

## PR Checklist

### DIFF
--- a/src/mito2/src/engine/region_hook.rs
+++ b/src/mito2/src/engine/region_hook.rs
@@ -421,10 +421,30 @@ pub trait RegionHook: Send + Sync + Debug {
     /// same `removed_files`; treat live cleanup as opportunistic.
     ///
     /// `region_metadata` is `None` for dropped regions; use `region_id` +
-    /// `access_layer`. Idempotent; runs on the background GC task (the offline
-    /// cleanup path runs it inline on the worker loop and propagates `Err` so
-    /// the caller retries the cleanup — the whole offline-cleanup handler is
-    /// idempotent).
+    /// `access_layer`. Idempotent.
+    ///
+    /// # Execution context
+    ///
+    /// On the global-GC path this runs on the background GC task, outside the
+    /// region worker loop. The offline cleanup path
+    /// (`handle_offline_cleanup_request`, reached by soft-drop PURGE) instead
+    /// runs it **inline in the region worker loop** and propagates `Err` so the
+    /// caller retries the (idempotent) cleanup. So — like `on_region_closed` /
+    /// `on_region_dropped` — implementations **must be fast and must not block
+    /// indefinitely** while awaited there: while the callback is pending, every
+    /// subsequent DDL request for every region on that worker is blocked.
+    /// Extension authors who wrote their hook against the "background GC task"
+    /// contract must account for this second, latency-sensitive trigger.
+    ///
+    /// The offline cleanup path fires this callback once the region directory is
+    /// confirmed **absent** — including no-op purges where nothing was actually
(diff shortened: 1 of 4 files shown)


v0y4g3r (commented):
Two documentation suggestions for the new offline-cleanup trigger of `on_region_gc`. Both are nits on the hook contract docs; no code changes requested.

src/mito2/src/engine/region_hook.rs
```diff
@@ -412,7 +421,10 @@ pub trait RegionHook: Send + Sync + Debug {
     /// same `removed_files`; treat live cleanup as opportunistic.
     ///
     /// `region_metadata` is `None` for dropped regions; use `region_id` +
-    /// `access_layer`. Idempotent; runs on the background GC task.
+    /// `access_layer`. Idempotent; runs on the background GC task (the offline
+    /// cleanup path runs it inline on the worker loop and propagates `Err` so
+    /// the caller retries the cleanup — the whole offline-cleanup handler is
+    /// idempotent).
```
v0y4g3r: Previously `on_region_gc` only ran on the background GC task; this PR adds a second trigger where it runs **inline on the region worker loop** (`handle_offline_cleanup_request`). While the hook is awaited there, all subsequent DDL requests for every region on that worker are blocked. The updated doc here mentions the inline execution but, unlike `on_region_closed` / `on_region_dropped` ("Runs inline in the region worker loop; implementations should be fast"), it doesn't warn implementors that the callback must now also be fast and must not block indefinitely in this path. Extension authors (e.g. an Iceberg export doing object-store listings) wrote their hooks against the "background GC task" contract and may not be aware of this. Could we add a similar sentence here, e.g. "On the offline cleanup path this runs inline in the region worker loop; implementations should be fast and must not block indefinitely"?

src/mito2/src/worker/handle_open.rs
```diff
@@ -67,6 +69,49 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+        // per-region state.
+        //
+        // Propagate the hook's error so the caller retries the cleanup, mirroring
+        // how the global GC worker keeps a region for the next pass when
+        // `on_region_gc` fails. The whole handler is idempotent (the existing
+        // `test_engine_offline_cleanup_closed_region` already drives it twice in
+        // a row), so a retry re-runs the directory removal, WAL obsolete and the
+        // hook safely. Runs inline, consistent with the blocking directory
+        // removal above; offline cleanup is already a slow path.
+        if let Some(hook) = self.plugins.get::<RegionHookRef>() {
```
v0y4g3r, line 88: `remove_region_dir_for_full_drop` returns `Ok` even when the region directory doesn't exist or is empty (listing an absent prefix yields nothing), so `on_region_gc(is_region_dropped = true, full_file_listing = true)` also fires on a no-op purge — e.g. a retry where the directory was already removed, or a region that never had files on this datanode. This is harmless for idempotent hooks, but it's a "terminal" notification where nothing was actually deleted, which is asymmetric with the GC path (fired only when files were deleted or a full listing was done). Could we add a rustdoc note on `on_region_gc` that the offline cleanup path fires the callback once the directory is confirmed absent, even if nothing was removed, so extensions don't interpret it as "files were deleted in this call"?

v0y4g3r (approved):
LGTM

Issue #8609, opened 2026-07-23
Tracking: Entity relationships and graph query


killme2008:
### What problem does the new feature solve?

Telemetry tables contain entity identities and evidence of relationships, but GreptimeDB has no common surface for querying them as topology. Users must manually join traces, metrics, and logs, or maintain a separate catalog or graph store, to answer questions such as which services call each other, where a workload runs, and what the topology looked like during an incident.

The accepted design is described in RFC #8605.

### What does the feature do?

Expose entities and time-ranged relationships as ordinary SQL tables derived from existing telemetry. The graph remains a logical view rather than a separate graph store.

This issue tracks the implementation:

**M0 and initial M1**

- [x] #8614 — entity declarations, computed `semantic_entities` and `semantic_relationships` tables, the entity registry, and trace-derived `calls` edges with RED metrics.

**M1 follow-ups**

- [x] #8794 — preserve caller authorization, cancellation, and deadlines in derived source scans; honor explicit query time windows.
- [x] Support metadata-only `ALTER TABLE SET/UNSET` for `greptime.semantic.*` options so existing tables can add and update entity declarations without routing semantic metadata through region-option mutations. https://github.com/GreptimeTeam/greptimedb/pull/8880
- [x] #8854 — zero-configuration Prometheus/Kubernetes entity declarations: embedded conventions.yaml + descriptor-metric whitelist (RFC alignment: #8853); generic per-table label rules deferred until the enumerate path has a scale strategy.
- [x] #8797 — stamp Remote Write 2.0 per-series metadata as semantic table options (#4765).
- [x] #8836 — derive virtual nodes for uninstrumented peers.
- [x] #8836 — derive attribute-based edges (same-row co-declaration vocabulary); relationship templates pending.
- [x] #8794 — add the declared-edge table and union declared edges into `semantic_relationships`.
- [x] #8836 — derive agent relationships (`parent_agent calls agent`, same-row `uses`/`invokes`; `gen_ai.*` naming via #8853/#8854); session entities and parent-child span derivation of `uses`/`invokes` pending.
- [x] #8904 — synthesize the `greptime_otel_resource_info` descriptor metric at OTLP ingestion, so OTLP-only deployments derive infrastructure entities without kube-state-metrics.
[...]

PR #8605, opened 2026-07-22
docs: add entity relationships and graph query RFC

killme2008:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

Implementation series starts at #8440.

## What's changed and what's your intention?

The RFC for modeling, deriving, and querying entity relationships (service → service calls, runs_on, contains, ...) over telemetry tables — the follow-up promised in the table-semantic-layer RFC's Future Work.

Core decisions:

- GreptimeDB does not become a graph database: the graph is a logical view over ordinary tables, queried with SQL (and later the ISO SQL/PGQ `GRAPH_TABLE`/`MATCH` surface over a snapshot relation).
- The graph is derived at read time: `greptime_private.semantic_entities` / `semantic_relationships` are computed, read-only tables over the declaring telemetry tables; nothing is materialised.
- Entity identity follows the OTel entity model (identifying attributes; column name = attribute key); edges are time-ranged facts with a small typed vocabulary, provenance, and RED metrics (servicegraph semantics).

## PR Checklist

Please convert it to a draft if some of the following conditions are not met.

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [x] API changes are backward compatible.
- [x] Schema or data changes are backward compatible.

PR #8614, opened 2026-07-23
feat: read-time entity relationships graph over telemetry (M0+M1)

killme2008:

Supersedes #8440.

Tracking issue: #8609.

Design: #8605.

Derive entity relationships (service → service calls, etc.) from ordinary telemetry tables at read time and expose them as two computed tables under `greptime_private`. Nothing is materialised; it is a logical view over data already stored. A user ingesting OTLP traces gets a queryable service dependency graph without an extra declaration or pipeline.

### Entity declaration

- Add the open-ended `greptime.semantic.entity.<type>.{id|descriptive|scope}` table-option namespace.
- Validate entity declarations during DDL: referenced columns must exist and entity `id` columns must be tag columns.
- Auto-stamp `greptime.semantic.entity.service.id = service_name` on the main OTLP trace table. Declarations are visible through `information_schema.table_semantics`.

### Read-time graph

- Add `greptime_private.semantic_entities` for observed entities and `greptime_private.semantic_relationships` for observed edges. They live in `greptime_private`, not `information_schema`, because scanning them performs derivation over telemetry data.
- Declare `EntityGraphProvider` in `catalog` and implement it in `frontend` over the query engine. Late injection avoids a `catalog → query` dependency cycle.
- Build derivation as typed DataFusion plans rather than generated SQL. The internally constructed default-window predicates push down into source scans, and user-controlled identifiers remain values rather than SQL text.
- Group entity declarations by source table and expand them with arrays plus `UNNEST`, so a table is scanned once regardless of how many entity types it declares.
- Derive `calls` edges by pairing trace client/server spans and globally aggregating RED metrics across trace tables.
- Stream derivation results through the computed tables instead of collecting the complete graph in memory.
- Use `observed_at` as the time index of both computed tables.
- Expose JSON outputs (`entity_id_attrs`, `descriptive`, `source_tables`, and `attributes`) as JSONB.
- Protect both computed tables from DDL and DML on every write path.

### Limitations

- Derivation currently uses an internal `QueryContext`; preserving caller authorization, cancellation, and deadlines is tracked for a follow-up.
- Scan predicates do not yet define the derivation window; it defaults to the last hour.
- Complete `greptime_trace_v1` schema validation is deferred; malformed legacy trace tables may still fail a relationship scan.
- Only the entity registry and trace-derived `calls` edges are included. Other edge types and declared edges are follow-ups.
- `WITH RECURSIVE` over computed tables is limited by DataFusion recursive CTE support; bounded-hop joins work.
- This is the first PR in the series and is not documented or announced yet. It introduces additive table metadata but does not change existing storage or wire formats.

## PR Checklist


src/operator/src/statement/semantic_graph.rs
```diff
@@ -0,0 +1,1040 @@
...
+        .select(REGISTRY_COLUMNS.into_iter().map(ident).collect::<Vec<_>>())?
+        .distinct()
+}
+
+/// Builds the `semantic_entities` registry plan: one branch and source scan per
+/// declaring table, filtered to `window`, then `UNION ALL` across source tables.
+/// Returns `None` when nothing declared an entity, so the computed table streams
+/// empty.
+pub fn build_registry_plan(
+    sources: Vec<(Vec<EntityDeclaration>, DataFrame)>,
```
sunng87: Data Clumps (minor): `build_registry_plan` accepts `Vec<(Vec<EntityDeclaration>, DataFrame)>` and `build_calls_plan` accepts `Vec<(EntityDeclaration, DataFrame)>`. Meanwhile `entity_graph.rs:64` already defines `struct EntitySource { declarations: Vec<EntityDeclaration>, table: TableRef }` with exactly these fields. Consider accepting `&[EntitySource]` in these plan-builder signatures so the grouping is named and the caller's `EntitySource` flows through without destructuring.

src/frontend/src/instance/builder.rs
```diff
@@ -260,6 +263,22 @@ impl FrontendBuilder {
...
+            .catalog_manager
+            .as_any()
+            .downcast_ref::<KvBackendCatalogManager>()
+        {
+            let provider: EntityGraphProviderRef = Arc::new(EntityGraphProviderImpl::new(
+                query_engine.clone(),
+                Arc::downgrade(&self.catalog_manager),
+            ));
+            kv_catalog.set_entity_graph_provider(provider);
+        }
```
sunng87, line 288: Judgement call: `experimental_` gate (architecture-invariant 5). This is the first PR in the series and is not yet documented or announced. Without a config gate the feature auto-stamps entity declarations on every new OTLP trace table and exposes the computed graph tables to all `information_schema` queries immediately. A short-lived `experimental_entity_graph` flag would let it bake before removal.

killme2008: Discussed offline — we're not adding a config gate for this. The decision is to treat the entire entity-graph feature as experimental and state that in the docs and the release note, rather than encoding the status in config naming. The surface is additive and read-only (namespaced `greptime.semantic.entity.*` options plus computed `information_schema` tables), so nothing lands in the stable config surface that a prefixed key would protect; and a short-lived `experimental_entity_graph` key would itself become config churn when dropped at stabilization. If the bake period shows we need a hard off-switch, we can add one then.

sunng87 (approved):

PR #8615, opened 2026-07-23
fix(query): preserve remote dynamic filter target

discord9:

Remote dynamic filter registrations are attached to the datanode selected for a region query, but subsequent update and unregister controls previously resolved the region route again. This loses affinity when a route changes or when a `RegionQueryHandler` adapter selects a non-leader peer.

This PR preserves the dispatch target across the RDF lifecycle:

- `RegionQueryHandler::do_get` returns the record-batch stream together with an immutable `RegionQueryTarget`;
- MergeScan registers RDF subscribers only after the region query succeeds and associates each subscriber with the returned target;
- RDF update and unregister fanout reuse the stored target instead of resolving the route again;
- the OSS frontend still selects the region leader, so existing OSS routing behavior is unchanged.

The dispatch order is initial registration encoding → successful `do_get` → target-aware subscriber registration → fanout start → stream polling. Updates produced while `do_get` is pending are reconciled from the latest stable filter snapshot after the subscriber is registered. Failed dispatches leave no subscriber and never start fanout.

There are no protobuf, wire-format, schema, datanode registry, or user-facing API changes. The internal `RegionQueryHandler` adapter interface changes mechanically to return and consume the routed target. Same-region/same-target duplicate attempts retain the existing deduplication semantics; per-attempt refcounting is outside this change.

Validation:

- `cargo nextest run -p query` — 547 passed, 2 skipped
- `cargo nextest run -p frontend` — 90 passed
- `cargo nextest run -p query remote_dyn_filter` — 47 passed
- `cargo clippy -p query -p frontend --all-targets --all-features -- -D warnings`
- repository pre-push formatting, license, typo, and clippy hooks

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/dyn_filter_bridge.rs
+++ b/src/query/src/dist_plan/dyn_filter_bridge.rs
@@ -27,6 +27,7 @@ use store_api::storage::RegionId;
 
 use crate::dist_plan::filter_id::build_remote_dyn_filter_id;
 use crate::dist_plan::{FilterId, QueryDynFilterRegistry, RemoteDynFilterProducerId, Subscriber};
+use crate::region_query::RegionQueryTarget;
 
 #[derive(Debug, Clone)]
 pub(crate) struct CapturedDynFilter {
@@ -93,9 +94,8 @@ fn downcast_dynamic_filter(
         .ok()
 }
 
-pub(crate) fn register_dyn_filters_for_region(
+pub(crate) fn register_remote_dyn_filters(
     registry: &QueryDynFilterRegistry,
-    region_id: RegionId,
     captured_dyn_filters: &[CapturedDynFilter],
 ) {
     for captured_dyn_filter in captured_dyn_filters {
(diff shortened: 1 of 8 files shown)


src/query/src/dist_plan/dyn_filter_bridge.rs
```diff
@@ -93,18 +94,29 @@ fn downcast_dynamic_filter(
...
+}
+
+pub(crate) fn register_dyn_filter_subscribers_for_region(
+    registry: &QueryDynFilterRegistry,
+    region_id: RegionId,
+    target: RegionQueryTarget,
+    captured_dyn_filters: &[CapturedDynFilter],
+) {
+    for captured_dyn_filter in captured_dyn_filters {
+        let _ = registry.register_subscriber(
```
killme2008: Do we need to check the registration result?

killme2008 (approved):
LGTM

src/query/src/dist_plan/dyn_filter_bridge.rs
discord9: Good point. Checked explicitly in 8bcca13bdd. `Added` and `Duplicate` require no caller action; `MissingFilter` would violate the expected ordering because producer entries are registered before dispatch. We now log that case as an invariant warning while continuing fail-open so an RDF bookkeeping anomaly cannot fail the query.

PR #8618, opened 2026-07-23
fix(mito2): remove stale compaction status when next compaction is throttled

### LINKED ISSUE #8617: Region stops compacting when next compaction is throttled by min_compaction_interval
### What type of bug is this?

Performance issue

Storage Engine, Datanode

### Minimal reproduce step

**Precondition:** configure a non-zero `min_compaction_interval` in the mito engine config, e.g.:

```toml
[region_engine.mito]
min_compaction_interval = "5m"
```

**Steps:**

1. Write data to a table and trigger a flush so that a compaction is scheduled (via the post-flush / post-edit `schedule_compaction` path). The worker records `last_schedule_compaction_millis` when the task is scheduled.
2. Let the compaction finish quickly (within `min_compaction_interval`, which is the common case for small compactions).
3. Keep writing and flushing, and/or issue a manual compaction (e.g. `ADMIN COMPACT_TABLE`), and observe.

What happens internally:

- In `CompactionScheduler::on_compaction_finished` (`src/mito2/src/compaction.rs`), when there is no pending manual request and no pending DDL, the function returns without removing the region's `CompactionStatus`, leaving it with `active_compaction = None`.
- Back in `RegionWorkerLoop::handle_compaction_finished` (`src/mito2/src/worker/handle_compaction.rs`), `is_compacting()` is now `false`, but the follow-up `schedule_next_compaction` call is gated by `min_compaction_interval`. Since the compaction finished within the interval, the cleanup/scheduling is skipped and the stale status remains in `region_status` forever.
- From then on, every compaction trigger for the region reaches `CompactionScheduler::schedule_compaction`, which sees an existing status and swallows the request without checking whether a compaction is actually active:
  - `Options::Regular` requests are merged into `status.waiters` and only woken by `on_compaction_finished`/failure paths, so they hang until region close/drop;
  - `Options::StrictWindow` (manual) requests are stored into `status.pending_request`, which is only consumed by `on_compaction_finished` — so the manual request is never executed and its waiter never completes.

No other code path removes the stale status except compaction failure/cancellation or region close/drop/truncate, so the region never compacts again until restart.

After a compaction finishes (even when the next scheduling is throttled by `min_compaction_interval`), the region should remain schedulable: subsequent flush-triggered, periodic, and manual compaction requests should eventually run, and manual compaction waiters should complete.

### What did you see instead?

The region never compacts again until restart:

- Automatic (flush/edit-triggered) compactions are silently dropped.
- Manual `ADMIN COMPACT_TABLE` requests hang indefinitely (parked in `pending_request`).
- SST files keep accumulating for the region, degrading read performance.

Any (reproduced by code inspection on Linux x64)

main @ 598a412fbb (mito2 v1.2.0)

N/A — the failure mode is silent (requests are swallowed without errors). The only observable symptom is the absence of subsequent compaction logs for the affected region and growing SST counts.

v0y4g3r:

Fixes #8617

**Summary:** Fixes a mito2 compaction scheduler bug where a region permanently stops compacting after a finished compaction when `min_compaction_interval` is configured to a non-zero value.

**Root cause:**

1. `CompactionScheduler::on_compaction_finished` falls through without removing the region's `CompactionStatus` when there is no pending manual request and no pending DDL, leaving a status with `active_compaction = None` in `region_status`.
2. Back in `RegionWorkerLoop::handle_compaction_finished`, `is_compacting()` is now `false`, but the follow-up `schedule_next_compaction` (whose `Ok(None)` branch is the only normal path that removes the status) is gated by `min_compaction_interval`. Since the scheduling timestamp was already updated when the task was scheduled, a compaction that finishes within the interval skips this cleanup, and the stale status remains forever.
3. From then on, every compaction trigger hits the "status exists" branch of `CompactionScheduler::schedule_compaction`, which does not check whether a compaction is actually active: regular requests are merged into `status.waiters` (never woken until close/drop/failure) and manual `StrictWindow` requests are parked in `status.pending_request` (never executed). The region never compacts again until restart and SSTs keep accumulating.

Note this only manifests with `min_compaction_interval > 0`; with the default `0` the gate always passes and `schedule_next_compaction` cleans up the status.

**How this PR works:**

- Adds `CompactionScheduler::remove_inactive_status(region_id)`, which removes the status only when it has no active compaction. A `debug_assert!` guards the invariant that such a status has no waiters/pending requests/pending DDLs (they can only be enqueued while a compaction is active, and the worker loop is strictly sequential).
- Calls it from `handle_compaction_finished` when the `min_compaction_interval` gate skips `schedule_next_compaction`, restoring the invariant that a status in the map always has an active compaction. The throttling semantics are unchanged: the next trigger still has to wait for the interval to pass.

**Tests:**

- `test_remove_inactive_status_keeps_running_compaction`: status with an active compaction (or no status) is never removed.
- `test_compaction_reschedulable_when_next_compaction_throttled`: reproduces the reported flow — schedule, finish (stale status remains), remove it as the throttled worker path does, and verify the region can be scheduled for compaction again.

**Compatibility:** no API, persisted-format, or wire-format changes; internal scheduler behavior only.

## PR Checklist

### DIFF
--- a/src/mito2/src/compaction.rs
+++ b/src/mito2/src/compaction.rs
@@ -343,6 +343,32 @@ impl CompactionScheduler {
             .unwrap_or(false)
     }
 
+    /// Removes the compaction status of the region if it has no active compaction.
+    ///
+    /// The worker calls this when it decides not to schedule the next compaction
+    /// after a compaction finishes (e.g. throttled by `min_compaction_interval`).
+    /// Otherwise a stale status without an active compaction would remain in the
+    /// map and all subsequent compaction requests would be swallowed as if the
+    /// region were still compacting.
+    ///
+    /// Returns whether the status is removed.
+    pub(crate) fn remove_inactive_status(&mut self, region_id: RegionId) -> bool {
+        let Some(status) = self.region_status.get(&region_id) else {
+            return false;
+        };
+        if status.active_compaction.is_some() {
+            return false;
+        }
+        // A status without an active compaction must have been drained by
+        // `on_compaction_finished`: no waiters, pending requests or DDLs can be
+        // enqueued while no compaction is running.
+        debug_assert!(status.waiters.is_empty());
+        debug_assert!(status.pending_request.is_none());
+        debug_assert!(status.pending_ddl_requests.is_empty());
+        self.region_status.remove(&region_id);
+        true
+    }
+
     /// Schedules next compaction upon a finished compaction.
     /// Returns whether the compaction is scheduled.
     pub(crate) async fn schedule_next_compaction(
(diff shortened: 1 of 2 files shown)


v0y4g3r:
@codex review

evenyag (approved):

PR #8622, opened 2026-07-23
fix: fail closed on malformed password assignments

fengjiachun:

Follow-up to #8539.

## What has changed and what is your intention?

Fail closed when a non-URI connection string contains a standalone `password =` assignment that cannot be safely parsed as a keyword option.

Malformed keyword strings can be accepted by `tokio-postgres` with the assignment embedded in another field, causing `Config::Debug` to retain the value. The same strings are returned unchanged when `pg_kvbackend` is disabled.

The keyword scan now:

- continues after redacting recognized assignments, so a second malformed assignment cannot bypass the fail-closed check;
- replaces the whole string when it finds a standalone assignment outside a recognized keyword boundary;
- matches `password` case-insensitively while leaving embedded identifiers such as `notPassword` unchanged.

Regression tests cover both parser-feature configurations, mixed assignments, case variants, and malformed separators.

## PR Checklist

### DIFF
--- a/src/common/meta/src/kv_backend/util.rs
+++ b/src/common/meta/src/kv_backend/util.rs
@@ -17,8 +17,8 @@ use url::Url;
 /// Placeholder substituted for any password found in a connection string.
 const REDACTED: &str = "***";
 
-/// Substituted for a whole URI-shaped connection string that no parser accepted.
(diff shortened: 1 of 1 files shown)


killme2008 (approved):
LGTM

PR #8624, opened 2026-07-23
perf(mito2): make compaction picker asynchronous to avoid blocking the region worker

### LINKED ISSUE #8621: Partial truncate can miss data rewritten by concurrent compaction
### What type of bug is this?

Data corruption, Incorrect result, Locking issue

Storage Engine

### Minimal reproduce step

This is a concurrency issue in Mito2's partial truncate and compaction paths.

1. Create a region containing SST files `A` and `B` whose time ranges are fully covered by a partial truncate request.
2. Start a compaction that rewrites `A` and `B` into a new SST file `C`.
3. Delay processing of the compaction's `CompactionFinished` notification after its `RegionEdit` has been committed to the manifest. At this point, the manifest contains `C`, while the in-memory `VersionControl` still contains `A` and `B`.
4. Execute `RegionTruncateRequest::ByTimeRanges` before the delayed compaction completion is applied to `VersionControl`.
5. Let the truncate and compaction completion finish, then query the truncated time range.

The partial truncate collects concrete file IDs from the current in-memory version in `src/mito2/src/worker/handle_truncate.rs`. It therefore records only `A` and `B` in `TruncateKind::Partial.files_to_remove`.

However, compaction may already have replaced `A` and `B` with `C` in the manifest. Applying the partial truncate removes only the recorded file IDs, so it does not remove `C`. When the delayed compaction completion applies its edit to the in-memory version, `C` becomes visible there as well.

The resulting state is internally consistent, but the truncate is incomplete: `C` still contains data from the requested truncate range.

After a partial truncate succeeds, data stored in SST files fully covered by the requested time ranges should no longer be visible, even if compaction overlaps with the truncate.

Partial truncate and compaction should be serialized or otherwise coordinated so that compaction output derived from truncated input files cannot survive the truncate.

### What did you see instead?

The partial truncate reports success, but data from the truncated time range can remain visible through a concurrent compaction output file.

The truncate action removes only the old input file IDs (`A` and `B`). It does not identify the newly generated compaction output (`C`) as containing the same data, so `C` remains in the manifest and is subsequently visible in `VersionControl`.

This behavior predates the async compaction picker changes and is independent of stale compaction completion handling.

Any (identified by code inspection on Linux x64)

main @ d1ac9e2da7

N/A. This is a timing-dependent race and does not necessarily emit an error.

v0y4g3r:

Fixes https://github.com/GreptimeTeam/greptimedb/issues/8621

This PR revises the mito2 compaction scheduling lifecycle by moving picker planning off the region worker loop and fixing a set of lifecycle bugs around plan tracking, cancellation, and status cleanup.

**Background planning with plan tracking**
- Compaction picking now runs in background (`spawn_compact` for planning, `spawn_blocking_compact` for the picker itself) instead of blocking the region worker loop; scheduling methods become synchronous and only dispatch planning.
- Each plan carries a monotonic `plan_id` and the owning `VersionControlRef`; results are delivered back to the worker via a new `BackgroundNotify::CompactionPickFinished` notification and accepted only when the plan and region instance are still current.

```mermaid
sequenceDiagram
    participant Worker as Region Worker
    participant Planning as Background Planning Task
    participant Picker as Picker (blocking thread)

    Worker->>Worker: 1. schedule_compaction()
    Worker->>Worker: Set status = Picking(plan_id)
    Worker->>Planning: 2. spawn_compact(dispatch_compaction_planning)
    Note over Worker: Worker loop continues processing writes/DDL

    Planning->>Picker: 3. spawn_blocking_compact(picker.pick)
    Picker-->>Planning: Return PickerOutput (side-effect-free)
    Planning->>Worker: 4. Send BackgroundNotify::CompactionPickFinished

    Worker->>Worker: 5. accept_compaction_pick_finished()
    Worker->>Worker: Validate VersionControl / plan_id / cancellation
    Worker->>Worker: Reserve CompactingFiles atomically (RAII) & start execution
```

**Execution identity and stale-result protection**
- Introduce `CompactionExecution` (plan id + version control identity + local/remote kind + file reservation lease) that is threaded through `CompactionFinished`/`CompactionCancelled`/`CompactionFailed`, so stale completions from reopened/truncated/superseded regions are reported as `StaleCompactionExecution` instead of being applied or misreported as region-closed.
- SST `compacting` marks are no longer set as a picker side effect; they are atomically reserved at plan-accept time via `CompactingFiles` (RAII, rolled back on conflict) after refreshing the picker output against the current SST version.

**Lifecycle bug fixes**
- Always notify the worker when planning finishes, even if the planning task panics (`catch_unwind`), so regions never get stuck in the picking phase.
- Remove idle compaction status when chaining is skipped, fixing a zombie status that blocked all future compaction scheduling of the region.
- Serialize `Truncate` with compaction: truncate is queued behind an in-flight compaction (cancel issued when possible) instead of racing it. This fixes #8621: a partial truncate previously collected file IDs from the in-memory `VersionControl` while a concurrent compaction had already committed its output file to the manifest, so the compaction output survived the truncate and its data remained visible. Now the truncate is only dispatched after the compaction's terminal notification has applied the region edit to the in-memory version, so the compaction output derived from truncated inputs is also collected into `files_to_remove` and deleted.
- Decrement the inflight compaction gauge only once per remote execution (after a re-entrancy guard in the notifier).
- Keep an in-flight compaction plan when scheduling the next one instead of orphaning it.

No API, schema, or data compatibility changes; all state is in-memory.

## PR Checklist


src/mito2/src/worker/handle_compaction.rs
```diff
@@ -151,17 +196,22 @@ impl<S> RegionWorkerLoop<S> {
...
-                self.compaction_scheduler
-                    .on_compaction_cancelled(region_id)
-                    .await
-            }
-            None => Vec::new(),
-        };
+        let mut pending_ddls = self
+            .compaction_scheduler
+            .on_execution_cancelled(region_id, &execution)
+            .await;
```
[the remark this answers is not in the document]
v0y4g3r: Resolved in `eac507ecab`. The cancellation handler now calls `on_compaction_result_notified` whenever cooperative cancellation returns pending DDLs, before `handle_ddl_requests` dispatches them. The regression test `test_local_compaction_cancellation_notifies_before_pending_ddl_dispatch` covers this exact ordering.

src/mito2/src/compaction.rs
```diff
@@ -325,53 +383,83 @@ impl CompactionScheduler {
...
+            for waiter in status.pending_regular.take().unwrap_or_default() {
+                waiter.send(Ok(0));
+            }
             self.region_status.remove(&region_id);
             // If there are pending DDL requests, we should return them to the caller.
             // And skip try to schedule next compaction task.
             return pending_ddl_requests;
         }
+
+        if status.pending_regular.is_some() {
```
evenyag: pending_regular is a Vec, do we still need an option?

src/mito2/src/compaction.rs
```diff
@@ -463,16 +559,20 @@ impl CompactionScheduler {
...
-        status.pending_ddl_requests.push(request);
+        if let Some(status) = self.region_status.get_mut(&request.region_id) {
+            // The first queued DDL fences later regular triggers from creating more follow-ups.
+            status.pending_ddl_requests.push(request);
+        } else {
+            warn!(
+                "Failed to add pending DDL request for region: {}, ddl: {:?}, region is not compacting",
+                request.region_id, request.request
+            );
+        }
```
evenyag: Is it possible?

src/mito2/src/compaction.rs
```diff
@@ -562,30 +723,195 @@ impl CompactionScheduler {
...
             ttl: Some(ttl),
             max_parallelism,
-            plugins: self.plugins.clone(),
+            plugins,
         };
 
-        let picker_output = {
+        listener.on_compaction_pick_begin(region_id).await;
+        let picker_region = compaction_region.clone();
+        let picker_output = match common_runtime::spawn_blocking_compact(move || {
```
evenyag, line 793: Will it have the risk of creating too many blocking threads for pickers? It may cause high cpu usage.

src/mito2/src/compaction.rs
```diff
@@ -1291,6 +1786,40 @@ fn estimate_compaction_bytes(picker_output: &PickerOutput) -> u64 {
         .sum()
 }
 
+/// Rebuilds picker output with current SST handles while preserving the picker's grouping.
+fn refresh_picker_output(output: PickerOutput, current: &SstVersion) -> Option<PickerOutput> {
+    let refresh = |file: FileHandle| {
+        current
+            .file_for_compaction(&file)
+            .filter(|current| !current.is_deleted() && !current.compacting())
+            .cloned()
```
evenyag, line 2003: Why do we need to get the handle again?

src/mito2/src/engine/listener.rs
```diff
@@ -89,6 +101,234 @@ pub trait EventListener: Send + Sync {
 
 pub type EventListenerRef = Arc<dyn EventListener>;
 
+/// Test gate that blocks selected compaction lifecycle phases for one region.
+pub struct CompactionPlanningGate {
```
evenyag: We should put it in the test mod.

src/mito2/src/compaction.rs
v0y4g3r: > pending_regular is a Vec, do we still need an option?

Yes. `Some(empty_vec)` means a compaction task triggered by flush (thus no OutputTx added).

src/mito2/src/compaction.rs
v0y4g3r: We can use `compact_rt_size` config item to limit the total threads spawned by compaction runtime.

src/mito2/src/compaction.rs
v0y4g3r: No, not with the current callers. Both truncate and enter-staging first called `request_cancel` and only enqueued the DDL for a non-`NotRunning` result. There is no await or scheduler mutation between those calls, and the region worker is serialized, so the status could not disappear in between.

The fallback was therefore unreachable and would silently drop the DDL if that invariant ever changed. I have changed this locally to a single `try_cancel_and_add_ddl` operation: it looks up the status, requests cancellation, and queues the DDL under the same mutable borrow. If compaction is not running, it returns the sender and typed request unchanged for immediate processing. I also added tests for both the queued and not-running paths.

src/mito2/src/compaction.rs
v0y4g3r: Answered in 1e3e513be5 by expanding the doc comment on `refresh_picker_output`.

The handles must be re-resolved because picking now runs in background on a version snapshot that may be stale by the time the plan is accepted: a concurrent flush, compaction, or index rebuild can replace a selected file with a new handle carrying updated metadata (e.g. `index_version`), or remove the file entirely. Re-fetching from the current `SstVersion` both detects gone files (aborting the plan) and ensures the execution reads and reserves the up-to-date handle.

src/mito2/src/engine/listener.rs
v0y4g3r: Fixed in f04d4a16cf by moving `CompactionPlanningGate`, its guards, and the `EventListener` implementation out of the shared `engine/listener.rs` module and into `engine/compaction_test.rs`, where they are exclusively used. The moved helper is now private to the test module.

src/mito2/src/compaction.rs
evenyag: `compact_rt_size` only limits the worker thread. There could be 512 blocking threads.
https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.max_blocking_threads

src/mito2/src/sst/version.rs
```diff
@@ -45,6 +45,22 @@ impl SstVersion {
         &self.levels
     }
 
+    /// Returns the unique current handle matching the selected file's identity and level.
+    pub(crate) fn file_for_compaction(&self, selected: &FileHandle) -> Option<&FileHandle> {
+        let mut files = self
+            .levels
+            .iter()
+            .filter_map(|level| level.files.get(&selected.file_id().file_id()));
+        let current = files.next()?;
```
evenyag: We can get the level using the level of the selected FileHandle. We never change the level of a file handle.

src/mito2/src/compaction.rs
```diff
@@ -888,8 +1333,10 @@ struct CompactionStatus {
     pending_request: Option<PendingCompaction>,
     /// Pending DDL requests that should run when compaction is done.
     pending_ddl_requests: Vec<SenderDdlRequest>,
-    /// Active compaction state.
-    active_compaction: Option<ActiveCompaction>,
+    /// Current compaction phase.
+    phase: Option<CompactionPhase>,
+    /// Waiters owned by a retained regular follow-up; `Some(empty)` records an automatic trigger.
+    pending_regular: Option<Vec<OutputTx>>,
```
evenyag: What's the difference between waiters and pending_regular? We also have PendingCompaction (pending_request). What's the relationship between it and pending_regular?

How to use and maintain these pending fields correctly?

Is it possible to merge some fields or move more fields into the CompactionPhase enum?

v0y4g3r: - `waiters` is the waiters of current running compaction
- `pending_regular` is the waiters of follow-up compaction, and a `Some(empty)` represents a follow-up compaction triggered by flush.

We can extract `waiters` and `pending_regular` into a struct to maintain the invariants.

v0y4g3r: Refactored in 356e9e44a5 by introducing `ActiveCompaction`, which groups the current `CompactionPhase`, current-cycle `waiters`, and the retained `pending_regular` follow-up. `CompactionStatus` now keeps only region context plus the orthogonal pending manual request and pending DDL queue.

The distinction is now encoded by ownership: `active.waiters` are satisfied by the current cycle, while `active.pending_regular` forces one fresh regular picking after that cycle (`Some(empty)` still represents an automatic trigger without an explicit waiter). Manual and regular pending work remain separate because they can coexist and carry different data/priority.

I kept `active` optional to preserve the existing transient idle state and added a TODO to remove idle map entries and make it non-optional once chained scheduling can recreate status from region context.

src/mito2/src/sst/version.rs
v0y4g3r: Fixed in cd8111f691 by looking up the current handle directly from the immutable level recorded by the selected FileHandle. The full RegionFileId check is retained, and a focused regression test covers level-scoped lookup.

src/mito2/src/compaction.rs
v0y4g3r: > `compact_rt_size` only limits the worker thread. There could be 512 blocking threads. https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.max_blocking_threads

We may need to limit the blocking threads in compact runtime, but it's out-of-current PR's scope.

src/mito2/src/compaction.rs
```diff
@@ -300,18 +361,18 @@ impl CompactionScheduler {
...
         // and defer returning the pending DDL requests to the caller.
-        if self
-            .handle_pending_compaction_request(
-                region_id,
-                manifest_ctx,
-                schema_metadata_manager.clone(),
-            )
-            .await
-        {
+        status.active = Some(active);
```
evenyag: Here we take the active and put it back again. Looks like we can just check whether it is some.

src/mito2/src/compaction.rs
```diff
@@ -320,58 +381,97 @@ impl CompactionScheduler {
             // So we return empty DDL requests.
             return Vec::new();
         };
+        let Some(restored_active) = status.take_active() else {
+            return Vec::new();
+        };
+        active = restored_active;
```
evenyag: We can also override active.
```suggestion
        let active = restored_active;
```

src/mito2/src/compaction.rs
```diff
@@ -727,9 +1149,165 @@ pub(crate) struct LocalCompactionState {
...
+    Remote {
+        execution: CompactionExecution,
+    },
+}
+
+#[derive(Debug)]
+struct ActiveCompaction {
+    phase: CompactionPhase,
+    waiters: Vec<OutputTx>,
+    pending_regular: Option<Vec<OutputTx>>,
```
evenyag: Document waiters and pending_regular.

src/mito2/src/compaction.rs
```diff
@@ -177,62 +256,57 @@ impl CompactionScheduler {
...
-                    status.set_pending_request(PendingCompaction {
-                        options,
-                        waiter,
-                        max_parallelism,
-                    });
-                    info!(
-                        "Region {} is compacting, manually compaction will be re-scheduled.",
-                        region_id
-                    );
+                    if status.pending_ddl_requests.is_empty() {
```
evenyag: Why do we set pending compaction only when there is no pending ddl?

src/mito2/src/compaction.rs
```diff
@@ -727,9 +1149,165 @@ pub(crate) struct LocalCompactionState {
...
+    fn merge_regular_trigger(&mut self, mut waiter: OptionOutputTx, ddl_fenced: bool) {
+        if matches!(self.phase, CompactionPhase::Picking { .. }) && !ddl_fenced {
+            let pending_regular = self.pending_regular.get_or_insert_default();
+            if let Some(waiter) = waiter.take_inner() {
+                pending_regular.push(waiter);
+            }
+        } else {
+            self.merge_waiter(waiter);
+        }
+    }
```
evenyag, line 1263: Why can we merge waiter when ddl fenced?

src/mito2/src/compaction.rs
```diff
@@ -177,62 +256,57 @@ impl CompactionScheduler {
...
-                    info!(
-                        "Region {} is compacting, manually compaction will be re-scheduled.",
-                        region_id
-                    );
+                    if status.pending_ddl_requests.is_empty() {
+                        status.set_pending_request(PendingCompaction {
+                            options,
+                            waiter,
+                            max_parallelism,
+                        });
```
evenyag: Does PendingCompaction only contains swcs compaction now?

src/mito2/src/compaction.rs
v0y4g3r: Updated in 5a2df73cad. Currently pending_ddl_requests only contains Truncate and EnterStaging, and either request acts as a scheduling fence after requesting cancellation of the active compaction. The fence is now checked before matching Regular or StrictWindow: any explicit compaction request has a waiter and receives CompactionCancelled, while automatic triggers have no waiter and are ignored. This avoids inferring whether a request is manual from its compaction options and guarantees that no later compaction request or follow-up cycle is inserted ahead of the waiting DDL.

src/mito2/src/compaction.rs
v0y4g3r: The same reason as https://github.com/GreptimeTeam/greptimedb/pull/8624#discussion_r3663992983

v0y4g3r: Fixed in 5a2df73cad. The DDL fence is now checked before dispatching by compaction options. When a Truncate/EnterStaging is pending, explicit Regular and StrictWindow requests receive CompactionCancelled, while automatic triggers have no waiter and are ignored. Therefore merge_regular_trigger is no longer called in a DDL-fenced state, and I removed the ddl_fenced parameter so this method cannot merge a waiter under that condition.

src/mito2/src/compaction.rs
v0y4g3r: fixed in 95b3b565564f40379749021dffa641ca60613944

src/mito2/src/compaction.rs
v0y4g3r: fixed in 95b3b565564f40379749021dffa641ca60613944

src/mito2/src/compaction.rs
v0y4g3r: fixed in https://github.com/GreptimeTeam/greptimedb/commit/5a2df73cad3bf018b048518b4c53442e879da095#diff-f413376968d6fea8034a8259e2f933c1cdfc370f30fa0d49df8a63156bb9cd9bR1181-R1187

src/mito2/src/compaction.rs
v0y4g3r: Yes. regular followup compactions are indicated by `ActiveCompaction::regular_followup_waiters`

src/mito2/src/compaction.rs
```diff
@@ -882,14 +1474,18 @@ struct CompactionStatus {
...
     /// Access layer of the region.
     access_layer: AccessLayerRef,
-    /// Pending waiters for compaction.
-    waiters: Vec<OutputTx>,
+    /// Current compaction lifecycle. `None` is the existing transient idle state.
+    // TODO: Remove idle statuses and make ActiveCompaction non-optional once chained
+    // scheduling can recreate the status from region context.
+    active: Option<ActiveCompaction>,
     /// Pending compactions that are supposed to run as soon as current compaction task finished.
     pending_request: Option<PendingCompaction>,
```
evenyag, line 1488: Maybe we should mention it only has swcs compaction. Regular (manual) compaction still in ActiveCompaction.

evenyag (approved):

PR #7908, opened 2026-04-02
refactor: introduce the ProjectInput structure

fengys1996:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## What's changed and what's your intention?

This PR refactors scan projection input from a plain Vec<usize> into a structured ProjectionInput, as a preparation step for nested/sub-field projection support.

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [ ] Schema or data changes are backward compatible.

PR #8629, opened 2026-07-24
refactor: simplify scan projection to root column indices

fengys1996:

- Simplify ScanRequest projection to root column indices. 
- Remove ProjectionInput.
- Remove ReadColumns merging because merging may lose ordering and other source-specific semantics.
- Preserve projection order and append predicate-required columns afterward. 
- Keep the existing nested-path representation unchanged. 

### Tests

- Relevant crate checks 
- Store API request tests 
- Mito2 projection tests

## PR Checklist


src/mito2/src/read/scan_region.rs
```diff
@@ -630,6 +608,106 @@ impl ScanRegion {
...
+
+        let mut read_col_ids = Vec::new();
+        let mut seen = HashSet::new();
+        for idx in projection {
+            let col_id = metadata
+                .column_metadatas
+                .get(*idx)
+                .with_context(|| InvalidRequestSnafu {
+                    region_id: metadata.region_id,
+                    reason: format!("projection index {} is out of bound", idx),
```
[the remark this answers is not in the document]
fengys1996: Good catch.

src/mito2/src/read/scan_region.rs
```diff
@@ -630,6 +608,106 @@ impl ScanRegion {
...
+                })?
+                .column_id;
+            let inserted = seen.insert(col_id);
+            debug_assert!(
+                inserted,
+                "projection contains duplicate column id: {}",
+                col_id
+            );
+            // Keep the projection order.
+            read_col_ids.push(col_id);
```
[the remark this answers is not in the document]
fengys1996: DataFusion’s logical OptimizeProjections rule already deduplicates pushed-down projection indices via RequiredIndices::compact(). The debug_assert! validates this upstream invariant; duplicate output expressions are handled above the table scan.

killme2008:
@fengys1996  Why we do this refactor?

MichaelScofield (approved):

fengys1996:
> @fengys1996 Why we do this refactor?

The current design was over-engineered, and this change mostly restores the original, simpler flow.

1. `ReadColumns` is derived from the projection pushed down by DataFusion together with the JSON type hints. Introducing `ProjectionInput` here does not provide additional value and increases cognitive overhead.

2. `ReadColumns::merge()` has ambiguous semantics. In particular, it cannot reliably preserve the original projection order. `ReadColumns` should instead be finalized once all required inputs are available and should not be modified afterward.

killme2008:
I think @evenyag should review this—there are many changes in mito.

killme2008:
Nice cleanup. The direction looks right to me — I checked and `ProjectionInput.nested_paths` never had a producer outside tests since #7908, so nested projection is effectively driven by `json_type_hint` alone. Removing the speculative API is the correct call, and since `ScanRequest` is in-process only there's no compat concern.

Two things I'd like to see addressed before merging:

**1. Duplicate projection indices (`build_read_col_ids`)**

```rust
let inserted = seen.insert(col_id);
debug_assert!(inserted, ...);
read_col_ids.push(col_id);
```

The old code silently deduped; this version asserts in debug and pushes the duplicate in release. The `OptimizeProjections` argument only holds for the DataFusion pushdown path, but `ScanRequest` is a contract shared by every caller — metric engine rewrite, `MetadataRegion`, scanbench (projection comes from CLI args), enterprise extensions. Nothing enforces the invariant at the type level, and `FlatProjectionMapper` explicitly documents that it "supports duplicate and unsorted projection indices", which contradicts the assert. Also note fuzz CI runs debug builds, so this panic is reachable from external input. In release, the duplicate flows into `ReadColumns::from_deduped_column_ids`, which doesn't dedup despite the name — `compute_format_projection` happens to merge it downstream, but that's luck, not design.

Could you restore first-seen dedup (keeping the `debug_assert` inside the dedup branch if you want to observe the DF invariant), or return `InvalidRequest` instead?

**2. Lost unit tests**

This PR deletes ~10 unit tests (out-of-bounds, dedup, empty projection → time index, predicate column resolution) and the new `build_read_col_ids` has no direct coverage — the tests listed in the description are pre-existing ones. The function only needs `self.version.metadata`, so it could be a free function taking `(metadata, projection, predicate)`; that makes it unit-testable and the still-relevant deleted cases can be ported back.

Minor points:

- `ScanRequestFingerprint` embeds raw `ReadColumns`, which is order-sensitive. The old BTreeMap merge made `SELECT a,b` / `SELECT b,a` share a fingerprint; now they miss the range cache. Not a correctness issue, but the comment explaining cache-key stability was deleted with the old code — worth a line somewhere, or canonicalize when building the fingerprint.
- FWIW the "merging may lose ordering" rationale in the description doesn't hold — everything downstream canonicalizes through `compute_format_projection`, so the old order was fine too. The simplification stands on its own.
- `"out of bounds"` here vs `"out of bound"` in `FlatProjectionMapper` — worth unifying while at it.
- The `fuzz_alter_table` failure is on an INSERT during alter (write path) — likely unrelated, but please rerun to confirm.
- This needs a companion enterprise PR (`ScanRequest` is used by the extension range provider side).

src/store-api/src/storage/requests.rs
```diff
@@ -140,15 +140,6 @@ pub struct ScanRequest {
     pub preserve_pk_dictionary_encoding: bool,
 }
 
-impl ScanRequest {
-    /// Returns the top-level projected column indices.
-    pub fn projection_indices(&self) -> Option<&[usize]> {
-        self.projection_input
-            .as_ref()
-            .map(|projection_input| projection_input.projection.as_slice())
-    }
```
evenyag, line 149: non-blocking: We can keep this method to reduce the change diff.

src/mito2/src/read/scan_region.rs
```diff
@@ -416,55 +413,36 @@ impl ScanRegion {
...
-            narrow_read_columns_by_json_type_hint(
-                &mut read_cols,
-                &self.request.json_type_hint,
-                &self.version.metadata,
-            );
-        }
-        let read_col_ids = read_cols.column_ids();
+
+        let read_cols = if has_structured_json {
+            self.read_columns_with_json_type_hint(&read_col_ids)
```
evenyag, line 436: Does it mean that we will read all nested path of the column?

fengys1996: No, the detail see [here](https://github.com/GreptimeTeam/greptimedb/pull/8629/changes#diff-4d0287919162c56176bb9cabaf16c44c15924294d034e144ab649df0ff391095R695-R707).

evenyag (approved):
I'd recommend adding more docs for json projection later.

PR #8637, opened 2026-07-24
Support unquoted postgres intervalstyle

yimeng:
## What

Allow PostgreSQL `SET INTERVALSTYLE = postgres` statements to use an unquoted identifier value.

## Why

PostgreSQL clients and extensions may send configuration values as bare identifiers. In particular, `postgres_fdw` initializes remote sessions with:

```sql
SET intervalstyle = postgres
```

GreptimeDB already accepts the quoted form:

```sql
SET INTERVALSTYLE = 'postgres'
```

and already supports `postgres` as a valid `PGIntervalStyle`, but the statement handler only accepted `Expr::Value`. This caused `postgres_fdw` queries against GreptimeDB's PostgreSQL protocol endpoint to fail during connection setup with:

```text
Not supported: Set variable value must be a value
```

## How

- Accept `Expr::Identifier` in `set_intervalstyle`.
- Reuse the existing `PGIntervalStyle::try_from(&str)` validation path.
- Add an integration test for `SET INTERVALSTYLE = postgres`.

## Validation

- Ran `git diff --check`.
- Could not run Rust tests locally because the current environment does not have a Rust toolchain installed.

### DIFF
--- a/src/operator/src/statement/set.rs
+++ b/src/operator/src/statement/set.rs
@@ -267,7 +267,7 @@ pub fn set_intervalstyle(exprs: Vec<Expr>, ctx: QueryContextRef) -> Result<()> {
         }
         _ => {
             return NotSupportedSnafu {
-                feat: "Set variable value must be a value",
+                feat: "Set variable value must be a value or identifier",
             }
             .fail();
         }
--- a/src/operator/src/statement/set.rs
+++ b/src/operator/src/statement/set.rs
@@ -256,15 +256,24 @@ pub fn set_intervalstyle(exprs: Vec<Expr>, ctx: QueryContextRef) -> Result<()> {
         }
         .fail();
     };
-    let Expr::Value(value) = var_value else {
-        return NotSupportedSnafu {
-            feat: "Set variable value must be a value",
+    let intervalstyle = match var_value {
+        Expr::Identifier(Ident {
+            value,
+            quote_style: _,
+            span: _,
+        }) => PGIntervalStyle::try_from(value.as_str()).context(InvalidConfigValueSnafu)?,
+        Expr::Value(value) => {
+            PGIntervalStyle::try_from(&value.value).context(InvalidConfigValueSnafu)?
         }
-        .fail();
+        _ => {
+            return NotSupportedSnafu {
+                feat: "Set variable value must be a value",
+            }
+            .fail();
+        }
     };
-    ctx.configuration_parameter().set_pg_intervalstyle_format(
-        PGIntervalStyle::try_from(&value.value).context(InvalidConfigValueSnafu)?,
-    );
+    ctx.configuration_parameter()
+        .set_pg_intervalstyle_format(intervalstyle);
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


tests-integration/tests/sql.rs
```diff
@@ -1160,6 +1160,12 @@ pub async fn test_postgres_intervalstyle(store_type: StorageType) {
     let client = validate_intervalstyle(client, "postgres", true).await;
     let client = validate_intervalstyle(client, "postgres_verbose", true).await;
     let client = validate_intervalstyle(client, "invalid_style", false).await;
+    assert!(
+        client
+            .simple_query("SET INTERVALSTYLE = postgres")
+            .await
+            .is_ok()
+    );
```
sunng87, line 1168: Can you add an assertion that verify the interval output format after setting this?

src/operator/src/statement/set.rs
```diff
@@ -256,15 +256,24 @@ pub fn set_intervalstyle(exprs: Vec<Expr>, ctx: QueryContextRef) -> Result<()> {
...
+            value,
+            quote_style: _,
+            span: _,
+        }) => PGIntervalStyle::try_from(value.as_str()).context(InvalidConfigValueSnafu)?,
+        Expr::Value(value) => {
+            PGIntervalStyle::try_from(&value.value).context(InvalidConfigValueSnafu)?
+        }
+        _ => {
+            return NotSupportedSnafu {
+                feat: "Set variable value must be a value",
```
v0y4g3r: The fallback error message still says `"Set variable value must be a value"`, but identifiers are now accepted input as well. Since this branch only triggers for expressions that are neither `Identifier` nor `Value` (e.g. function calls, binary expressions), consider rewording it to something like `"Set variable value must be a value or identifier"` to give users complete guidance. Purely cosmetic; optional.

v0y4g3r (commented):
Nit on error message wording (inline comment). Static review only; no verdict intended.

tests-integration/tests/sql.rs
yimeng: Added an assertion that verifies the interval output is `1 day 02:03:00` after executing the unquoted `SET INTERVALSTYLE = postgres`. Thanks!

src/operator/src/statement/set.rs
yimeng: Updated the fallback error message to `Set variable value must be a value or identifier`. Thanks for the suggestion!

yimeng:
The latest CI runs require maintainer approval. Could someone please approve them when convenient? Thanks!

sunng87 (approved):

sunng87:
@yimeng thank you for the patch! Just curious are you using greptimedb as a postgres foreign data wrapper? I would like to learn about your case if you want to share.

yimeng:
Thanks for asking! In my case it's not exactly PostgreSQL FDW.

I'm using Home Assistant, and I keep the entity relationships / metadata on the PostgreSQL side, while smart-home telemetry and time-series data are stored in GreptimeDB. Then I query them together through the PostgreSQL-compatible interface so I can correlate HA entities with the corresponding GreptimeDB time-series data.

So this patch mainly helps with compatibility in that PostgreSQL-oriented workflow, where clients or tools may emit `SET INTERVALSTYLE = postgres` without quotes before running queries.

yimeng:
To clarify a bit: using GreptimeDB through PostgreSQL FDW is also a good fit for this case.

The idea is to keep Home Assistant entity relationships / metadata in PostgreSQL, store smart-home time-series data in GreptimeDB, and then join PostgreSQL local tables with GreptimeDB foreign tables for correlation queries.

sunng87:
@yimeng Awesome! Thank you very much for sharing. Are you going to publish this work in future? If you will write a blog post about this project, we are very happy to quote and repost.

PR #8645, opened 2026-07-27
fix(partition): avoid panic on missing route columns

### LINKED ISSUE #8636: Logical metric table inserts can panic after physical repartition adds a partition column
## What type of bug is this?

Crash, Unexpected error

Distributed Cluster, Table Engine, Frontend

## Minimal reproduce step

A supported metadata transition can leave a logical metric table with fewer partition columns than its physical table:

1. Create a physical metric table whose schema contains `host` and `rack`, initially partitioned only on `host`.
2. Create a logical metric table before the repartition. Creation injects the then-current physical partition column `host`, but not `rack`.
3. Repartition the physical table with expressions using both `host` and `rack`. Existing logical schemas are not updated.
4. Insert through the stale logical table with a `host` value that reaches a physical route expression referencing `rack`.

The following component-level setup reproduces the write path deterministically using real `TableMetadataManager`, route/partition caches, and `PartitionRuleManager::split_rows`:

```text
physical table 1024: columns [host, rack], partition keys [0, 1]
logical table 1025:  columns [host],       partition keys [0]
logical route: 1025 -> 1024
physical expressions:
  host < "m"
  host >= "m" AND rack < "n"
  host >= "m" AND rack >= "n"
input row: host = "z"
```

The regression expectation is a typed `partition::error::Error::UndefinedColumn { column: "rack" }`, not a panic. Run:

```bash
cargo nextest run -p operator test_logical_split_rows_rejects_physical_only_partition_column --no-capture
```

The test panicked deterministically on all four nextest attempts.

## What did you expect to see?

The metadata transition should either be rejected/propagated safely, or write routing should return a controlled missing-column error. It must not panic.

## What did you see instead?

`PartitionRuleManager` combines partition names from the logical `TableInfo` (`host`) with expressions from the physical route (`host`, `rack`). `MultiDimPartitionRule::evaluate_expr` then performs an unchecked lookup for `rack`:

```text
thread panicked at src/partition/src/multi_dim.rs:124:58:
called `Option::unwrap()` on a `None` value
```

Relevant stack:

```text
MultiDimPartitionRule::evaluate_expr
MultiDimPartitionRule::find_region
SplitReadRowHelper::split_to_regions
RowSplitter::split
PartitionRuleManager::split_rows
```

This was discovered while validating #8590, but it is a pre-existing write-path problem and is not caused by that PR. The affected `split_rows`/`multi_dim` path is unchanged from `main`.

A distributed SQLness attempt could not complete the preceding repartition because the test topology failed staging-region synchronization with `Manifest missing for region ...`; therefore the confirmed evidence here is the production component call chain plus the independently verified DDL metadata reachability, rather than a completed SQL end-to-end run.

## What operating system did you use?

Ubuntu 22.04, Linux 6.8, x86_64

## What version of GreptimeDB did you use?

Current `main` base `8afb960dda285f96eb9997ac2437a079f37c8f39`. The deterministic component test ran at rebased head `03d3ac366ef9946d0e9b78feb2b7b6aa75bcdbe7`; the relevant write-routing code is identical to the stated main base.

## Relevant log output and stack trace

```text
test tests::partition_manager::test_logical_split_rows_rejects_physical_only_partition_column ... FAILED
thread panicked at src/partition/src/multi_dim.rs:124:58:
called `Option::unwrap()` on a `None` value
Summary: 1 test run, 0 passed, 1 failed (reproduced on all 4 attempts)
```

discord9:

Closes #8636

A physical metric-table repartition can introduce a partition column that is absent from an existing logical table. Write routing then combines logical partition-column names with physical route expressions. When evaluation reaches the physical-only column, `MultiDimPartitionRule` previously unwrapped a missing column index and panicked.

This PR:

- returns the existing typed `UndefinedColumn` / `InvalidArguments` error from both partition-expression operand orientations instead of panicking;
- preserves behavior for valid partition rules and does not change public APIs, schemas, or persisted data;
- adds focused unit coverage for both operand orientations;
- adds a component regression using real logical-to-physical route metadata, with physical partition keys `{a, b}` and a stale logical schema/key set `{a}`.

Validation:

- `cargo nextest run -p partition test_find_region_rejects_undeclared_column --no-capture`
- `cargo nextest run -p operator test_logical_split_rows_rejects_physical_only_partition_column --no-capture`
- `cargo check -p partition -p operator`
- `cargo clippy -p partition -p operator --all-targets --all-features -- -D warnings`

## PR Checklist

### DIFF
--- a/src/operator/src/tests/partition_manager.rs
+++ b/src/operator/src/tests/partition_manager.rs
@@ -15,6 +15,9 @@
 use std::collections::{BTreeMap, HashMap};
 use std::sync::Arc;
 
+use api::v1::helper::tag_column_schema;
+use api::v1::value::ValueData;
+use api::v1::{ColumnDataType, Row, Rows};
 use common_base::hash::partition_expr_version;
 use common_meta::cache::{TableRouteCacheRef, new_table_route_cache};
 use common_meta::key::TableMetadataManager;
@@ -37,7 +40,11 @@ pub fn new_test_table_info(
     table_name: &str,
     _region_numbers: impl Iterator<Item = u32>,
 ) -> TableInfo {
-    let column_schemas = vec![
+    new_test_table_info_with_columns(table_id, table_name, test_column_schemas(true), vec![0])
+}
+
+fn test_column_schemas(include_b: bool) -> Vec<ColumnSchema> {
+    let mut column_schemas = vec![
         ColumnSchema::new("a", ConcreteDataType::int32_datatype(), true),
         ColumnSchema::new(
             "ts",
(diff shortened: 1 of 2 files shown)


WenyXu:
@codex review

evenyag (approved):

PR #8646, opened 2026-07-27
perf(promql): avoid repeated scans in sliding range evaluation

discord9:

This PR removes repeated scanning work from PromQL sliding range evaluation in two independent layers.

### 1. Two-pointer range boundary generation (`range_manipulate.rs`)

`RangeManipulateStream::calculate_range` used a stale cursor heuristic that rescanned each evaluation window — `O(E × samples-per-window)` — and could drop valid samples after sparse gaps or trailing empty windows. It is replaced with monotonic `left`/`right` pointers, reducing boundary generation to `O(N + E)` while preserving exact `(curr-range, curr]` semantics, start/end shortening, and canonical empty-window output. Regression tests cover the gap/tail sample-loss cases (including positive time shifts), duplicates, range-zero, and shortening.

### 2. Adaptive edge counting for `changes()`/`resets()` (`functions/`)

The generic `#[range_fn]` macro slices, downcasts, and independently rescans every overlapping window. For `changes` and `resets`, this PR replaces the macro path with hand-written UDF wrappers backed by one shared private kernel: a direct raw-offset scan when the requested number of internal edges is small, otherwise a single global `u64` edge prefix over the backing array so each window is answered by a prefix difference (or a short direct scan).

Behavior is preserved bit-for-bit, verified by differential and deterministic unit tests: raw null-buffer values (null slots participate via their physical buffer, matching the existing macro behavior), NaN equivalence in `changes`, NaN false-comparisons in `resets`, signed zero, infinities, empty/singleton windows, independent timestamp/value offsets, arbitrary/reordered/repeated window layouts, and the exact DataFusion error messages for shape mismatches. The shared proc macro, planner, serializer, and all other range functions are untouched.

### Benchmarks (release binaries, same pinned source, fixed CPU affinity, ABBA ordering)

Boundary (commit 1):

- Public `RangeManipulate` wall time: **~28%** faster at 1m/15s, **~66%** at 5m/15s, **~96%** at 1h/15s (micro benchmark).
- Warmed distributed TQL ANALYZE 1h queries: **~17–21%** faster end to end; shorter windows stayed within run-order noise.

Edge counting (commit 2):

- Dense sliding windows (k=4/20/240): **91.7–95.6%** less public UDF wall time (12–23×).
- Low-coverage fallback (N=4096 backing, only 8 windows): **73.9–74.4%** faster — the cost gate avoids global-prefix overhead.
- Warmed distributed TQL ANALYZE 5m/1h `changes`/`resets`: **12.1–19.7%** client and **12.0–20.9%** server latency improvement; unrelated controls moved within ABBA drift (−1% to +3.7%).

The large micro numbers reflect removing per-window `O(k)` rescans; end-to-end gains are smaller because storage scan and sort dominate query time. These are controlled single-host measurements, not fleet-wide guarantees.

### Limitations and compatibility

- No API, schema, wire, or persisted-format changes.
- Only `changes`/`resets` get a specialized kernel; `count`, `sum`, `min`/`max`, etc. still use the generic macro path (follow-up candidates).
- The existing raw null-buffer semantics are intentionally preserved rather than "fixed"; changing null behavior would be a separate, behavior-visible decision.

### CI coverage

This PR also wires the new `promql_range_boundary` case into the query-regression workflow's `DEFAULT_CASES` (`.github/scripts/query-regression-run.py`). A historical audit found range queries had zero coverage in query-regression runs (208/208 PromQL `ANALYZE` samples were bare selectors), which is how the range-path issues this PR fixes went unnoticed. The case costs about a minute of execution on top of the workflow's existing release builds, so label-triggered runs now permanently cover this path.

## PR Checklist

### DIFF
--- a/src/promql/src/functions/changes.rs
+++ b/src/promql/src/functions/changes.rs
@@ -15,23 +15,9 @@
 //! Implementation of [`changes`](https://prometheus.io/docs/prometheus/latest/querying/functions/#changes) in PromQL. Refer to the [original
 //! implementation](https://github.com/prometheus/prometheus/blob/main/promql/functions.go#L1023-L1040).
 
-#[cfg(test)]
-use std::sync::Arc;
-
-use datafusion::arrow::array::{Float64Array, TimestampMillisecondArray};
-#[cfg(test)]
-use datafusion::common::DataFusionError;
 use datafusion::logical_expr::ScalarUDF;
-#[cfg(test)]
-use datafusion::physical_plan::ColumnarValue;
-#[cfg(test)]
-use datatypes::arrow::array::Array;
-#[cfg(test)]
-use datatypes::arrow::datatypes::DataType;
 
 use crate::functions::edge_count::{self, EdgeKind};
-#[cfg(test)]
-use crate::range_array::RangeArray;
 
 /// used to count the number of value changes that occur within a specific time range
 #[derive(Debug)]
@@ -47,33 +33,21 @@ impl Changes {
     }
 }
 
-#[allow(dead_code)]
-fn changes(_: &TimestampMillisecondArray, values: &Float64Array) -> Option<f64> {
-    if values.is_empty() {
-        None
-    } else {
-        let (first, rest) = values.values().split_first().unwrap();
-        let mut num_changes = 0;
-        let mut prev_element = first;
-        for cur_element in rest {
-            if cur_element != prev_element && !(cur_element.is_nan() && prev_element.is_nan()) {
-                num_changes += 1;
-            }
-            prev_element = cur_element;
-        }
-        Some(num_changes as f64)
-    }
-}
-
 #[cfg(test)]
 mod test {
+    use std::sync::Arc;
+
+    use datafusion::arrow::array::{Array, Float64Array, TimestampMillisecondArray};
     use datafusion::arrow::buffer::NullBuffer;
+    use datafusion::common::DataFusionError;
+    use datafusion::physical_plan::ColumnarValue;
     use datafusion_common::config::ConfigOptions;
     use datafusion_expr::ScalarFunctionArgs;
-    use datatypes::arrow::datatypes::Field;
+    use datatypes::arrow::datatypes::{DataType, Field};
 
     use super::*;
     use crate::functions::test_util::simple_range_udf_runner;
+    use crate::range_array::RangeArray;
 
     // build timestamp range and value range arrays for test
     fn build_test_range_arrays(
@@ -208,12 +182,10 @@ mod test {
         value_ranges: Vec<(u32, u32)>,
     ) -> Vec<Option<f64>> {
         assert_eq!(timestamp_ranges.len(), value_ranges.len());
-        assert!(
-            timestamp_ranges
-                .iter()
-                .zip(&value_ranges)
-                .all(|((_, timestamp_length), (_, value_length))| timestamp_length == value_length)
-        );
+        assert!(timestamp_ranges
+            .iter()
+            .zip(&value_ranges)
+            .all(|((_, timestamp_length), (_, value_length))| timestamp_length == value_length));
         assert_eq!(values.len(), raw_values.len());
         let nulls = values.iter().map(Option::is_none).collect::<Vec<_>>();
 
--- a/src/promql/src/functions/changes.rs
+++ b/src/promql/src/functions/changes.rs
@@ -182,10 +182,12 @@ mod test {
         value_ranges: Vec<(u32, u32)>,
     ) -> Vec<Option<f64>> {
         assert_eq!(timestamp_ranges.len(), value_ranges.len());
-        assert!(timestamp_ranges
-            .iter()
-            .zip(&value_ranges)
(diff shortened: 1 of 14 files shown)


evenyag:
@codex review

src/promql/src/functions/changes.rs
```diff
@@ -15,22 +15,40 @@
 //! Implementation of [`changes`](https://prometheus.io/docs/prometheus/latest/querying/functions/#changes) in PromQL. Refer to the [original
 //! implementation](https://github.com/prometheus/prometheus/blob/main/promql/functions.go#L1023-L1040).
 
+#[cfg(test)]
```
WenyXu: nit: Can we move it into test mod?

src/promql/src/functions/changes.rs
```diff
@@ -15,22 +15,40 @@
...
+        "prom_changes"
+    }
+
+    pub fn scalar_udf() -> ScalarUDF {
+        edge_count::scalar_udf(Self::name(), EdgeKind::Changes)
+    }
+}
+
+#[allow(dead_code)]
+fn changes(_: &TimestampMillisecondArray, values: &Float64Array) -> Option<f64> {
```
WenyXu: Nit: should we remove the dead code?

src/promql/src/functions/changes.rs
discord9: Moved all test-only imports into the test modules in both `changes.rs` and `resets.rs`. Verified with `cargo check -p promql --tests` and the focused changes/resets nextest suite (10/10).

src/promql/src/functions/changes.rs
discord9: Removed the unused pre-specialization `changes` helper and the corresponding `resets` helper. Production still delegates to `edge_count`; the independent test oracles are unchanged.

src/promql/src/functions/changes.rs
```diff
@@ -69,6 +66,36 @@ mod test {
         (ts_range_array, value_range_array)
     }
 
+    fn invoke_range_udf(
```
WenyXu: **Non-blocking:** Could we move the shared test scaffolding into `functions::test_util`?

`invoke_range_udf`, `assert_execution_error`, `TinyPrng`, and most of `run_oracle_ranges` here are duplicated almost verbatim in `resets.rs`:

```rust
fn invoke_range_udf(...) { ... }
fn assert_execution_error(...) { ... }
struct TinyPrng(u64);
fn run_oracle_ranges(...) { ... }
```

Keeping `changes_oracle` and `resets_oracle` independent is useful for differential testing, but the UDF invocation, range construction, error assertion, and PRNG plumbing can be shared. This would reduce the chance that the two suites drift while preserving separate behavior expectations.

tests/perf/query_cases/promql_range_boundary/case.toml
```diff
@@ -0,0 +1,166 @@
...
+series_count = 128
+start_unix_nanos = 1704067200000000000
+step_nanos = 15000000000
+time_range_layout = "non_overlapping_per_sst"
+series_layout = "timestamp_major"
+
+[[scenario.queries]]
+name = "plain_selector_control_2h"
+kind = "tql"
+query = "TQL ANALYZE VERBOSE (1704070800, 1704078000, '15s') promql_range_boundary{host=~'host.*'}"
```
WenyXu, line 62: **Non-blocking:** Could we also add a result-level regression test for the sparse-gap/tail cases fixed by this PR?

The queries in this case use `TQL ANALYZE VERBOSE` with only latency thresholds:

```toml
query = "TQL ANALYZE VERBOSE (...) count_over_time(...)"

[scenario.queries.thresholds]
max_candidate_latency_regression_pct = 20
```

This verifies that the public query path executes and does not regress in latency, but it would still pass if the produced range contents were incorrect. The current correctness tests exercise `RangeManipulateStream::calculate_range` directly. A small sqlness or integration test with sparse samples, an empty intermediate window, and a valid trailing sample would additionally protect the planner and execution path against the original sample-loss regression.

WenyXu (commented):
Two non-blocking testing suggestions.

src/promql/src/functions/changes.rs
discord9: Done in `958ef7c`. Shared UDF invocation, range construction, exact execution-error assertions, PRNG, and oracle-driving plumbing now live in `functions::test_util`; `changes_oracle` and `resets_oracle` remain independent in their respective modules. The focused changes/resets suites pass 10/10, and `cargo check -p promql --tests` passes.

tests/perf/query_cases/promql_range_boundary/case.toml
discord9: Added an end-to-end sqlness regression in `958ef7c` with sparse samples at 0s and 100s, 30s evaluation steps, and a 25s range. It verifies that the intermediate 30s/60s/90s windows produce no rows while the valid trailing 120s window is retained. The generated result contains only the expected 0s and 120s rows, and the case passes in both standalone and distributed environments.

WenyXu:
@codex review

WenyXu (approved):

PR #8647, opened 2026-07-27
feat: support old-stage datanode config overlays

discord9:

N/A

This PR adds a generic old-stage datanode configuration capability to the
compatibility runner. A case can opt in with a native TOML sidecar:

```toml
[old_config]
datanode = "old-datanode.overlay.toml"
```

The runner:

- loads and validates sidecars before creating state or starting services;
- merges TOML tables recursively while replacing all non-table values,
  including arrays, atomically;
- restores runner-owned topology, address, storage, and WAL fields;
- groups semantically equivalent overlays by a full SHA-256 profile key;
- runs baseline and overlay profiles sequentially with isolated state and etcd;
- applies overlays only to old-stage datanodes and keeps current-stage configs
  clean across restarts;
- uses unwind-safe process/state guards and checked etcd cleanup so failed
  starts and restarts do not leak resources.

The metadata format, merge/protection rules, and profile behavior are documented
under `tests/compatibility`. Existing cases without `[old_config]` continue to
run in the baseline profile. This does not change persisted data or wire formats.

Validation:

- `cargo fmt --check -p sqlness-runner`
- `cargo nextest run -p sqlness-runner` (64 passed)
- `cargo clippy -p sqlness-runner --tests -- -D warnings`
- `git diff --check`

A full Docker-backed old-version compatibility run was not executed locally.

## PR Checklist

### DIFF
--- a/tests/compatibility/AGENTS.md
+++ b/tests/compatibility/AGENTS.md
@@ -28,12 +28,42 @@ This file is intended for AI agents editing compat cases or the compat runner. F
 - All required fields must be non-empty: `name`, `reason`, `introduced_by`,
   `topologies`, `from_range`, `to_range`, `features`, `owner`.
 
+## Old-Stage Datanode Overlays
+
+- An old-stage datanode overlay is declared only as:
+
+  ```toml
+  [old_config]
+  datanode = "old-datanode.overlay.toml"
+  ```
+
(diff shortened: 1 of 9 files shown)


killme2008 (approved):
Good job!

killme2008:
Should we support standalone config overlays? @discord9

PR #8652, opened 2026-07-27
fix(auth): warn when credential load disables Postgres SCRAM or drops a line

killme2008:

The static and watch user providers degraded silently in two ways when loading credentials. This PR makes both cases observable. It is **logging only** — authentication behavior is unchanged.

**1. Silent Postgres SCRAM downgrade.** `postgres_auth_info_with_credential` selects one auth method for the whole endpoint: if *any* user's verifier can't back SCRAM, every Postgres login falls back to cleartext. This is deliberate (per-user method selection would leak verifier type / user existence, which the `mock_for_unknown_user` path exists to prevent). The problem is it happens with no signal — an operator who keeps one `mysql_native_password` account, or a legacy `pbkdf2_sha256` hash that predates SCRAM, silently serves cleartext passwords over Postgres while believing SCRAM is in effect.

Note both `mysql_native_password` *and* `pbkdf2_sha256` disqualify SCRAM, for different reasons: the former is a double-SHA1 digest unrelated to PBKDF2; the latter was derived without SASLprep and can't be safely reused as a SCRAM secret without the original password.

We now emit a warning at each credential load (startup + every watch reload) listing the offending users, so operators can enable TLS or regenerate those verifiers with `pg_scram_sha256`. The warning never logs secrets.

**2. Silently dropped credential lines.** A malformed line — most commonly a plaintext password containing `=`, which splits into more than two parts — was discarded by `filter_map` with no trace, so the user simply vanished from the map. We now warn with the **line number and file path** (never the line content) and note that passwords containing `=` are unsupported.

## PR Checklist

### DIFF
--- a/src/auth/src/user_provider.rs
+++ b/src/auth/src/user_provider.rs
@@ -328,8 +328,23 @@ fn load_credential_from_file(filepath: &str) -> Result<UserInfoMap> {
     let file = File::open(path).context(IoSnafu)?;
     let credential = io::BufReader::new(file)
         .lines()
-        .map_while(std::result::Result::ok)
         .enumerate()
+        .map_while(|(idx, line)| match line {
+            Ok(line) => Some((idx, line)),
+            Err(err) => {
+                // A read error (I/O failure or invalid UTF-8) ends the iterator,
+                // so every remaining credential is dropped. Warn instead of
+                // vanishing silently, matching the malformed-line handling below.
+                warn!(
+                    "Failed to read line {} of user provider file {}: {}; \
+                     all remaining credentials are ignored",
+                    idx + 1,
+                    filepath,
+                    err
+                );
+                None
+            }
+        })
         .filter_map(|(idx, line)| {
             // The line format is:
             // - `username=password` - Basic user with default permissions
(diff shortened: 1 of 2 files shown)


v0y4g3r (commented):
Static review of the credential-load warnings. One observation on a silent-drop path that remains next to the ones this PR fixes — inline below.

src/auth/src/user_provider.rs
```diff
@@ -328,7 +329,8 @@ fn load_credential_from_file(filepath: &str) -> Result<UserInfoMap> {
     let credential = io::BufReader::new(file)
         .lines()
         .map_while(std::result::Result::ok)
```
v0y4g3r: `map_while(Result::ok)` terminates the iterator permanently on the first `Err` from `lines()` (I/O error or invalid UTF-8): every credential after that point is silently discarded, and the new malformed-line warning can't fire because the line never reaches `parse_credential_line`. This is the same "user vanishes with no trace" class this PR fixes for parse failures, and `ensure!(!credential.is_empty())` doesn't catch it when earlier lines loaded fine.

Consider warning before truncating, e.g.:

```suggestion
        .map_while(|line| match line {
            Ok(line) => Some(line),
            Err(err) => {
                warn!(
                    ?err,
                    "Failed to read a line from user provider file {filepath}; \
                     all remaining credentials are ignored"
                );
                None
            }
        })
```

fengjiachun (approved):

PR #8655, opened 2026-07-27
ci(query-regression): include PromQL range boundary case in defaults

discord9:

Depends on #8646 (this PR's default-case entry references `tests/perf/query_cases/promql_range_boundary/case.toml`, which is introduced there).

Adds the PromQL range-boundary perf case to the query-regression workflow's `DEFAULT_CASES`, so label-triggered query regression runs actually measure the PromQL range path.

An audit of past query-regression runs (208 PromQL `ANALYZE` samples across four historical runs) found that range queries had **zero coverage** — every sample was a bare selector. This meant changes to range evaluation (e.g. `RangeManipulate`, range UDF kernels) could regress without CI noticing; indeed the range-boundary heuristic both rescanned overlapping windows (`O(E × samples-per-window)`) and silently dropped samples after sparse gaps, undetected. #8646 fixes those issues and introduces the `promql_range_boundary` case; this PR wires that case into the default set so the coverage gap stays closed for future PRs.

The case is intentionally cheap: a synthetic 99,840-row direct-SST fixture (generated in ~0.3s) with 10 TQL `ANALYZE` queries — a plain-selector control plus `count`/`sum`/`rate`/`changes`/`resets` over 1m/5m/1h windows at 15s cadence. A full base-vs-candidate pass takes about a minute locally, negligible next to the workflow's two release builds.

Verified locally: `split_cases("all")` returns 11 cases and every referenced case file exists; the case itself has been exercised end-to-end in controlled base/candidate runs.

## PR Checklist

### DIFF
--- a/.github/scripts/query-regression-run.py
+++ b/.github/scripts/query-regression-run.py
@@ -31,6 +31,7 @@ DEFAULT_CASES = [
     "tests/perf/query_cases/prom_remote_write_mixed_every/case.toml",
     "tests/perf/query_cases/prom_remote_write_integer_counter/case.toml",
     "tests/perf/query_cases/promql_pushdown_7913/case.toml",
+    "tests/perf/query_cases/promql_range_boundary/case.toml",
     "tests/perf/query_cases/analyze_verbose_many_files/case.toml",
     "tests/perf/query_cases/sql_topk_order_by/case.toml",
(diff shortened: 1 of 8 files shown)


discord9:
Folded into #8646 per author request: the DEFAULT_CASES entry now lives there as commit a7cb1fe359, since the case file is introduced by the same PR.

PR #8656, opened 2026-07-27
feat(common-query): add native histogram runtime model

shuiyisong:

This is the 4th PR of the 8 PRs of native histogram.

This pull request adds a shared runtime model for Prometheus native histograms in `common-query`.

It supports Arrow decoding and construction, histogram arithmetic and reconciliation, reset detection, quantiles, fractions, and variance calculations. It preserves the existing persisted field layout and prefix-aware internal column naming.

## PR Checklist

### DIFF
--- a/src/common/query/src/native_histogram.rs
+++ b/src/common/query/src/native_histogram.rs
@@ -34,6 +34,8 @@ use datatypes::data_type::{ConcreteDataType, DataType};
 use datatypes::types::{StructField, StructType};
 use once_cell::sync::Lazy;
 
+use crate::prelude::greptime_native_histogram;
+
 pub const NATIVE_HISTOGRAM_FIELD: &str = "greptime_native_histogram";
 pub const SCHEMA_FIELD: &str = "schema";
 pub const ZERO_THRESHOLD_FIELD: &str = "zero_threshold";
--- a/src/common/query/src/native_histogram.rs
+++ b/src/common/query/src/native_histogram.rs
@@ -16,6 +16,8 @@
 //!
 //! Prom remote-write v2 stores these names as children of one Struct field, while
 //! metric-engine uses the same contract to recognize native histogram tables.
+//! [`NativeHistogram`] is the query-time representation and therefore normalizes
+//! integer and floating-point payloads to absolute `f64` counts.
 
 use std::collections::BTreeMap;
 use std::sync::Arc;
@@ -26,7 +28,7 @@ use datafusion::arrow::array::{
 };
 use datafusion::arrow::buffer::NullBuffer;
 use datafusion::arrow::datatypes::{
-    ArrowPrimitiveType, DataType as ArrowDataType, Float64Type, Int32Type, Int64Type,
+    ArrowPrimitiveType, DataType as ArrowDataType, Field, Float64Type, Int32Type, Int64Type,
     TimestampMillisecondType, UInt32Type, UInt64Type,
 };
 use datafusion_common::{DataFusionError, Result as DfResult};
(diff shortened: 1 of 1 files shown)


src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
...
+    }
+
+    fn is_empty_payload(&self) -> bool {
+        self.count == 0.0
+            && self.zero_count == 0.0
+            && self.sum == 0.0
+            && self.positive_buckets.iter().all(|count| *count == 0.0)
+            && self.negative_buckets.iter().all(|count| *count == 0.0)
+    }
+}
```
sunng87, line 848: We should really consider make histogram a built-in datatype, then these functions can be member functions of the type.

src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
     data_type == native_histogram_value_type()
 }
 
+pub fn is_native_histogram_value_schema(name: &str, data_type: &ConcreteDataType) -> bool {
+    name == greptime_native_histogram() && is_native_histogram_value_type(data_type)
+}
+
 pub const CUSTOM_BUCKETS_SCHEMA: i32 = -53;
```
sunng87, line 136: **issue:** Missing doc comment on public constant (`docs/style-guide.md`: "Add comments for public functions and structs.").

The same applies to all five `pub const` sentinel values (lines 130–136): `CUSTOM_BUCKETS_SCHEMA`, `UNKNOWN_COUNTER_RESET_HINT`, `COUNTER_RESET_HINT`, `NOT_COUNTER_RESET_HINT`, `GAUGE_RESET_HINT`. Each deserves at least a one-line `///` explaining its role.

src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
...
+const MIN_EXPONENTIAL_SCHEMA: i32 = -4;
+const MAX_EXPONENTIAL_SCHEMA: i32 = 8;
+pub const UNKNOWN_COUNTER_RESET_HINT: i32 = 0;
+pub const COUNTER_RESET_HINT: i32 = 1;
+pub const NOT_COUNTER_RESET_HINT: i32 = 2;
+pub const GAUGE_RESET_HINT: i32 = 3;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Span {
```
sunng87, line 193: **issue:** Missing doc comment on public struct (`docs/style-guide.md`: "Add comments for public functions and structs.").

Same applies to `BoundaryRule` (line 145), `Bucket` (line 153), and `NativeHistogram` (line 161). These are the core public types of this module — they need `///` doc comments explaining their role in the native histogram model.

src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
...
+    pub positive_spans: Vec<Span>,
+    pub negative_spans: Vec<Span>,
+    pub count: f64,
+    pub zero_count: f64,
+    pub positive_buckets: Vec<f64>,
+    pub negative_buckets: Vec<f64>,
+}
+
+impl NativeHistogram {
+    pub fn uses_custom_buckets(&self) -> bool {
```
sunng87: **issue:** Missing doc comments on public methods (`docs/style-guide.md`: "Add comments for public functions and structs.").

All ~25 public methods on `NativeHistogram` (`uses_custom_buckets`, `compatible_with`, `add`/`sub`/`scale`/`divide_by`, `detect_reset`, `quantile`, `fraction`, etc.) are undocumented. Since this is "a shared runtime model" that callers across crates will depend on, each public method should carry at least a one-line `///` explaining what it computes and any invariants on its inputs.

src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
...
+        }
+    }
+    Some(zero_threshold)
+}
+
+fn ceil_div(value: i32, divisor: i32) -> i32 {
+    value.div_euclid(divisor) + i32::from(value.rem_euclid(divisor) != 0)
+}
+
+pub fn get_bound(idx: i32, schema: i32, custom_values: &[f64]) -> Option<f64> {
```
sunng87: **issue:** Missing doc comment on public function `get_bound` (`docs/style-guide.md`: "Add comments for public functions and structs.").

This function handles two very different code paths (exponential schema vs custom buckets) — a `///` comment explaining the returned bound semantics would help callers.

src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
...
+        })?;
+    if span_len != buckets {
+        return Err(DataFusionError::Execution(format!(
+            "native histogram {name} spans describe {span_len} buckets, found {buckets}"
+        )));
+    }
+    Ok(())
+}
+
+pub fn read_histogram(array: &StructArray, row: usize) -> DfResult<Option<NativeHistogram>> {
```
sunng87, line 1338: **issue:** Missing doc comment on public function `read_histogram` (`docs/style-guide.md`: "Add comments for public functions and structs.").

src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
...
+        positive_buckets,
+        negative_buckets,
+    }))
+}
+
+fn list_opt<T>(values: Vec<T>) -> Option<Vec<Option<T>>> {
+    Some(values.into_iter().map(Some).collect())
+}
+
+pub fn build_histogram_array(values: &[Option<NativeHistogram>]) -> ArrayRef {
```
sunng87, line 1412: **issue:** Missing doc comment on public function `build_histogram_array` (`docs/style-guide.md`: "Add comments for public functions and structs.").

Also worth noting: this always writes the f64 payload variant and leaves the u64/i64 variants null, so an integer-encoded histogram cannot round-trip back to integer form.

src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
...
+        match current_span_start {
+            None => {
+                spans.push(Span {
+                    offset: idx,
+                    length: 1,
+                });
+                current_span_start = Some(idx);
+            }
+            Some(_) if idx == previous_index + 1 => {
+                spans.last_mut().expect("span exists").length += 1;
```
sunng87: **issue:** `expect()` panic in non-test code (`.agents/architecture-invariants.md`: "In non-test code, return errors instead of `unwrap()` / `expect()` / `panic!()`.").

`suggestion:` Return `None` from `spans_from_indices_counts` when the Vec is empty (matching the `Option` convention already used by the surrounding free functions) rather than calling `expect()` on `last_mut()`.

src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
...
+}
+
+pub fn native_histogram_arrow_type() -> ArrowDataType {
+    native_histogram_value_type().as_arrow_type()
+}
+
+fn native_histogram_fields() -> datafusion::arrow::datatypes::Fields {
+    match native_histogram_arrow_type() {
+        ArrowDataType::Struct(fields) => fields,
+        _ => unreachable!("native histogram must be a struct"),
```
sunng87: **issue:** `unreachable!()` panic in non-test code (`.agents/architecture-invariants.md`: "In non-test code, return errors instead of `unwrap()` / `expect()` / `panic!()`.").

`suggestion:` Return a `DataFusionError` instead; the calling code already uses `DfResult`.

src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
...
+
+        assert_eq!(result.schema, CUSTOM_BUCKETS_SCHEMA);
+        assert_eq!(result.reset_hint, GAUGE_RESET_HINT);
+        assert_eq!(result.count, -2.0);
+        assert_eq!(result.sum, -2.0);
+        assert_eq!(result.positive_buckets, vec![-2.0]);
+    }
+}
 
 // ---------------------------------------------------------------------------
```
sunng87, line 1964: **issue:** `//` comment block where `///` is preferred (`docs/style-guide.md`: "Prefer document comment (`///`) over normal comment (`//`) for structs, fields, functions etc.").

This 16-line block documents the Parquet field-id scheme, which is part of the public API (`native_histogram_subfield_id`, `native_histogram_list_element_id`). It should use `///` so it appears in rendered documentation.

src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
...
+    pub count: f64,
+    pub boundary_rule: BoundaryRule,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct NativeHistogram {
+    pub schema: i32,
+    pub zero_threshold: f64,
+    pub sum: f64,
+    pub reset_hint: i32,
```
sunng87: **smell (judgment call):** Primitive Obsession — `reset_hint: i32` is paired with four `pub const` sentinel values (lines 133–136) that form a closed set. A dedicated enum would give exhaustiveness checking and self-documenting variants. Mitigation: the `i32` values follow the Prometheus native histogram spec; a newtype wrapping the enum might be a middle ground.

src/common/query/src/native_histogram.rs
```diff
@@ -110,7 +123,1652 @@ pub fn is_native_histogram_value_type(data_type: &ConcreteDataType) -> bool {
...
+        let mut result = self.clone();
+        result.count = 0.0;
+        result.zero_count = 0.0;
+        result.sum = 0.0;
+        result.positive_buckets.fill(0.0);
+        result.negative_buckets.fill(0.0);
+        result
+    }
+
+    pub fn add_exact(&self, other: &Self) -> Option<Self> {
```
sunng87: **smell (judgment call):** Duplicated Code — `add_exact` (L197) and `sub_exact` (L224) share an identical shape: `compatible_with` guard → clone → mutate count/zero_count/sum → `merge_side` ×2. Only the binary operator and `reset_hint` treatment differ. Consider extracting a private `fn combine_exact(self, other, reset_hint, op)` helper. Low blast radius, but worth revisiting if a third operation is added.

src/common/query/src/native_histogram.rs
shuiyisong: Addressed. `CUSTOM_BUCKETS_SCHEMA` and the four reset-hint aliases now have rustdoc. The reset-hint aliases are typed `CounterResetHint` values rather than raw `i32` sentinels.

src/common/query/src/native_histogram.rs
shuiyisong: Addressed. Added rustdoc for the public `CounterResetHint`, `Span`, and `NativeHistogram` APIs and their fields. `BoundaryRule` and `Bucket` are internal implementation details now, so they were made private.

src/common/query/src/native_histogram.rs
shuiyisong: Addressed. The remaining public `NativeHistogram` methods now document their behavior and failure semantics. Helpers with no cross-crate callers, including `uses_custom_buckets`, `compatible_with`, `add_exact`, `sub_exact`, `compact`, and bucket materialization helpers, were made private instead of expanding the public API.

src/common/query/src/native_histogram.rs
shuiyisong: Addressed by making `get_bound` private. It has no external callers in this stack, so keeping it out of the public API is smaller than documenting it as a supported contract.

src/common/query/src/native_histogram.rs
shuiyisong: Addressed. `read_histogram` now documents null handling, integer-to-`f64` normalization, and malformed field or span errors.

src/common/query/src/native_histogram.rs
shuiyisong: Addressed the actionable API-contract part. `build_histogram_array` now documents that query-time histograms are emitted through the canonical `f64` payload fields. Preserving the original integer family is intentionally not part of `NativeHistogram`; a regression test verifies `count_u64` is null, `count_f64` is populated, and the Arrow type remains exact.

src/common/query/src/native_histogram.rs
shuiyisong: Addressed without treating empty input as an error. `spans_from_indices_counts` now returns `Option`, removes `expect()`, and uses checked adjacency, offset, and span-length arithmetic. Tests cover empty input, adjacent and gapped buckets, and an unrepresentable extreme gap.

src/common/query/src/native_histogram.rs
shuiyisong: Addressed without widening the builder API to `DfResult`. The `unreachable!()` path was removed; the builder now derives Struct fields from the named arrays it constructs, and a regression test verifies the resulting Arrow type matches `native_histogram_arrow_type()`.

src/common/query/src/native_histogram.rs
shuiyisong: Addressed with a typed `CounterResetHint` enum. It includes `Unrecognized(i32)` and explicit `i32` conversions so persisted and protobuf values remain forward-compatible instead of being rejected or normalized. A round-trip test covers an unknown value.

src/common/query/src/native_histogram.rs
shuiyisong: Recorded. Let's first merge the NH PRs and then investigate the built-in data type plan.

src/common/query/src/native_histogram.rs
shuiyisong: Agreed after revisiting this. Addressed in `a2bb0577`: the two private wrappers are replaced by one `combine_exact(other, reset_hint, op)` helper, while `add` and `sub` pass their operation directly and preserve their distinct reset-hint behavior. This centralizes compatibility checks, scalar updates, and both bucket-side merges without adding another abstraction layer. All 47 `common-query` tests, Clippy, and the workspace check passed.

sunng87 (approved):

PR #8657, opened 2026-07-27
feat: add a dedicated http api server port

sunng87:

This patch adds a dedicated http api server port for our `/v1` and `/dashboard`.
The new port is opt-in, and can be configured via something like `enable_api_server`. 

This patch will reuse same axum `Router` for both servers, but adding a new middleware to block paths for api server.

## PR Checklist

### DIFF
--- a/src/servers/src/http.rs
+++ b/src/servers/src/http.rs
@@ -232,7 +232,7 @@ pub struct HttpOptions {
     /// only the `v1` interfaces plus the dashboard. It shares every other
     /// `[http]` option with the main server and only differs by its bound
     /// address (see `api_server_addr`).
-    pub api_server_enable: bool,
+    pub enable_api_server: bool,
     /// The address to bind the dedicated HTTP API server, in the same form as
     /// `addr` (e.g. `127.0.0.1:4006`).
     pub api_server_addr: String,
@@ -250,7 +250,7 @@ impl Default for HttpOptions {
             prom_validation_mode: PromValidationMode::Strict,
             experimental_enable_prometheus_native_histogram: false,
             experimental_enable_explain_analyze_stream: true,
-            api_server_enable: false,
+            enable_api_server: false,
             api_server_addr: format!("127.0.0.1:{}", DEFAULT_HTTP_API_ADDR_PORT),
         }
     }
@@ -839,7 +839,7 @@ impl HttpServerBuilder {
 
     /// Builds the full HTTP server (serves the `v1` interfaces **and** the
     /// internal interfaces such as health, status, metrics, config and debug).
-    /// When [`HttpOptions::api_server_enable`] is `true`, it also builds a
+    /// When [`HttpOptions::enable_api_server`] is `true`, it also builds a
     /// dedicated HTTP **API** server that serves only the `v1` interfaces plus
     /// the dashboard.
     ///
@@ -856,7 +856,7 @@ impl HttpServerBuilder {
     /// second server is bound to `api_server_addr` (e.g. `127.0.0.1:4006`). The
     /// API server inherits **every** option from `[http]` except the bound address.
     pub fn build_servers(self) -> (HttpServer, Option<HttpServer>) {
-        let api_enabled = self.options.api_server_enable;
+        let api_enabled = self.options.enable_api_server;
         let api_addr = self.options.api_server_addr.clone();
 
         let internal = HttpServer {
@@ -1629,7 +1629,7 @@ mod test {
         let instance = Arc::new(DummyInstance { _tx: tx });
         // Enable the dedicated API server so the split behavior can be exercised.
         let options = HttpOptions {
-            api_server_enable: true,
+            enable_api_server: true,
             ..HttpOptions::default()
         };
         HttpServerBuilder::new(options)
@@ -1642,7 +1642,7 @@ mod test {
     pub async fn test_http_api_options_defaults() {
         let opts = HttpOptions::default();
         // The API server is disabled by default.
-        assert!(!opts.api_server_enable);
+        assert!(!opts.enable_api_server);
         assert_eq!(opts.api_server_addr, "127.0.0.1:4006");
     }
 
@@ -1715,7 +1715,7 @@ mod test {
             timeout: Duration::from_secs(42),
             body_limit: ReadableSize::mb(128),
             cors_allowed_origins: vec!["https://example.com".to_string()],
-            api_server_enable: true,
+            enable_api_server: true,
             ..HttpOptions::default()
         };
         let (internal, api) = HttpServerBuilder::new(http_opts.clone()).build_servers();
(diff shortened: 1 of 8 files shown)


src/servers/src/http.rs
```diff
@@ -764,37 +804,96 @@ impl HttpServerBuilder {
...
+            Some(HttpServer {
+                options: api_options,
+                user_provider: self.user_provider.clone(),
+                shutdown_tx: Mutex::new(None),
+                router: StdMutex::new(self.router),
+                internal_router: StdMutex::new(self.internal_router),
+                bind_addr: None,
+                memory_limiter: self.memory_limiter,
+                api_only: true,
+            })
```
v0y4g3r, line 887: Splitting routes by moving `/metrics` and `/config` into `internal_router` assumes that everything accumulated in the main `router` is public API. That assumption holds in OSS, but breaks for downstream consumers that inject internal-only routes via `with_extra_router()` / `RouterConfigurator` — those routes are cloned here onto the API-only server and become reachable on the public-facing port, violating the "only the `v1` interfaces plus the dashboard" contract.

Concrete examples from the enterprise edition (GreptimeTeam/greptimedb-enterprise), which builds its own `HttpServerBuilder` with extra routers and passes it via `Services::with_http_server_builder`:

- `projects/enterprise/src/frontend.rs:326` — `GET /v1/region/metadata`: explicitly an internal, temporary interface for the enterprise Ingester (the code comment says it may be removed entirely once better options exist).
- `projects/enterprise/src/frontend.rs:322` — `POST /v1/sst/ingest`: an internal SST-ingestion endpoint paired with the above.
- `src/standalone-plugins/src/handler/dump_metadata.rs:152` — `POST /v1/admin/metadata/dump`: an admin operation that dumps metadata to object storage.
- `projects/enterprise/src/standalone/role_based_instance/http.rs:50` — `GET /status/standalone/role` and `GET /status/standalone/is_leader`: non-`/v1` paths registered through `RouterConfigurator`, so `need_auth` skips authentication for them; they would be served unauthenticated on the API port.

When `api_server_enable` is turned on, all of these would additionally be exposed on the API server. For deployments without auth enabled, the three `/v1` internal/admin endpoints become anonymously reachable on the public port.

Suggestion: give registrants an explicit choice of which server their routes belong to, e.g. keep `with_extra_router()` (public, both servers) and add `with_internal_extra_router()` merging into `internal_router` (internal server only), and split `RouterConfigurator` similarly. Defining this API surface before the enterprise edition bumps its dependency avoids a later route-ownership migration.

v0y4g3r (commented):
One design concern about the route split in `build_servers`, based on how the enterprise edition uses `HttpServerBuilder`. Posting a single inline comment with the details.

src/servers/src/http.rs
shuiyisong: After tracing this through the Enterprise integration, I think a cleaner architecture than adding `with_internal_extra_router()` and keeping an unscoped legacy path is to make the two listener roles explicit while keeping route registration unified.

### Proposed Full/API model

- Represent the closed listener modes as `HttpServerKind::{Full, Api}` rather than `api_only: bool`.
- Let `HttpServerBuilder` accumulate exactly one complete router, including built-in routes, `RouterConfigurator` routes, and other extra routers.
- The **Full** listener serves that complete router. This preserves the historical HTTP port, including `/v1`, for backward compatibility.
- The **API** listener wraps the finalized router with one outer surface guard. It accepts only exact `/v1`, `/v1/*`, exact `/dashboard`, and `/dashboard/*`; everything else returns `404` before reaching a handler.
- Apply the guard at the final build step, after plugin routes, debug routes, and fallbacks have been added. Putting it in `make_app()` would be weaker because callers can append routes before the later `build()` call.

This makes URL namespace the exposure contract instead of requiring every downstream registrant to select and maintain a second route bucket. Enterprise can continue using the existing `RouterConfigurator` and `with_extra_router()` integration unchanged. Its `/status/standalone/role` and `/status/standalone/is_leader` routes remain available on the Full listener and are automatically hidden from the API listener.

The corresponding invariant is that every `/v1` route is API-visible. Therefore the Enterprise routes currently described as internal (`/v1/sst/ingest`, `/v1/region/metadata`, and possibly `/v1/admin/metadata/dump`) need an explicit intent audit. If they are truly internal-only, they should move outside `/v1` in the downstream update, without retaining compatibility aliases. If those paths cannot move, then a namespace guard is insufficient and we really do need explicit per-route exposure metadata.

### Benefits

- One route registry and one plugin API; no duplicated `router`/`internal_router` state.
- One centralized boundary protects present and future plugin routes, including non-`/v1` routes that auth currently skips.
- No legacy unscoped registration API or route-ownership migration for downstream projects.
- The Full listener remains backward compatible while the API listener has a small, testable contract.
- Exact namespace matching rejects lookalikes such as `/v10` and `/dashboard-admin`.
- Reachability and authentication remain separate: the listener decides whether a path exists on that port, then the existing auth middleware handles allowed `/v1` requests.

The key tradeoff is deliberate: this design is the cleanest only if `/v1` genuinely means API-visible across OSS and Enterprise.

shuiyisong: See [the gist](https://gist.github.com/shuiyisong/855a14d862b4c4a5057fc29fd4affa3d) for the detailed design idea

sunng87: Sounds good. Will work on an update.

src/servers/src/http.rs
```diff
@@ -145,6 +164,43 @@ pub struct HttpServer {
...
+pub(crate) fn is_namespace(path: &str, root: &str) -> bool {
+    path == root
+        || path
+            .strip_prefix(root)
+            .is_some_and(|suffix| suffix.starts_with('/'))
+}
+
+/// Paths visible on the dedicated API listener: the `/v1` APIs and the dashboard.
+pub fn is_api_listener_path(path: &str) -> bool {
+    is_namespace(path, HTTP_API_PREFIX_WITHOUT_TRAILING_SLASH) || is_namespace(path, "/dashboard")
```
shuiyisong, line 183: The bundled dashboard served on this listener still depends on `/status`.

Reproducer: enable `http.enable_api_server`, open `http://<api_server_addr>/dashboard/` with fresh dashboard local storage, then inspect the navbar or open `/dashboard/#/dashboard/status`. The pinned dashboard v0.13.7 [requests `/status`](https://github.com/GreptimeTeam/dashboard/blob/v0.13.7/src/api/status.ts#L3-L7) and [infers its API host from the page origin](https://github.com/GreptimeTeam/dashboard/blob/v0.13.7/src/store/modules/app/index.ts#L169-L181); this guard returns `404`, so the version badge stays blank and the Status page always renders “unsupported”.

Could we route this dependency through an allowed/authenticated endpoint or adapt the dashboard for the API listener? Simply admitting the current `/status` handler may expose build and host details.

sunng87: this /status interface is just removed in latest dashboard release

shuiyisong (approved):

PR #8659, opened 2026-07-27
fix(prometheus): custom column remote reads

### LINKED ISSUE #8634: bug(prometheus): remote_read does not honor custom timestamp and value column names
### What type of bug is this?

Incorrect result, Unexpected error

Write Protocols, Frontend, Query Engine

### Minimal reproduce step

Prometheus remote write can accommodate an existing metric table whose timestamp and single Float64 value columns use names different from the process-wide defaults. However, remote read still looks up the columns through `greptime_timestamp()` and `greptime_value()` instead of resolving them from the table schema.

1. Start GreptimeDB without changing `default_column_prefix`.
2. Pre-create a physical metric table and a logical metric table with custom timestamp/value column names:

```sql
CREATE TABLE custom_phy (
    custom_ts TIMESTAMP TIME INDEX,
    custom_value DOUBLE
) ENGINE = metric
WITH ("physical_metric_table" = "");

CREATE TABLE custom_metric (
    custom_ts TIMESTAMP TIME INDEX,
    custom_value DOUBLE,
    job STRING PRIMARY KEY
) ENGINE = metric
WITH ("on_physical_table" = "custom_phy");
```

3. Send Prometheus remote-write samples for `custom_metric`, selecting `custom_phy` as the physical table. The existing-schema write alignment maps the incoming Prometheus timestamp/value columns to `custom_ts` and `custom_value`.
4. Send a Prometheus remote-read request for `__name__="custom_metric"`.

Relevant code paths:

- Existing-schema write alignment:
  - `src/operator/src/insert.rs`
  - `src/servers/src/prom_row_builder.rs`
- Remote-read fixed-name lookups:
  - `query_to_plan()` in `src/servers/src/prom_store.rs` builds the time filter with `greptime_timestamp()`.
  - `collect_timeseries_ids()` excludes only `greptime_timestamp()` and `greptime_value()`.
  - `recordbatch_to_timeseries()` retrieves columns only by `greptime_timestamp()` and `greptime_value()`.

Note that globally configured names through `default_column_prefix` already work because all of these call the same global getters. The problem is limited to arbitrary names obtained from an existing table schema.

A possible fix is to resolve the time-index column and the single Prometheus value field from the table/DataFrame schema, pass those names through query planning and result conversion, and add a remote-read regression test with `custom_ts`/`custom_value`.

Remote read should use the metric table schema to locate `custom_ts` and `custom_value`, and return the same samples that were accepted by remote write.

### What did you see instead?

Remote read references the process-wide default column names. Query planning can fail because `greptime_timestamp` is absent. If conversion is reached with differently named result columns, it reports that the `greptime_timestamp` or `greptime_value` column is missing.

N/A — identified through code inspection; the issue is platform-independent.

Current `main` checkout at `598a412fbb55`.

```bash
No runtime log captured. Static inspection shows fixed-name lookups in src/servers/src/prom_store.rs.
```

grezzko:

Fixes #8634 

Prometheus remote reads were using default values (greptime_timestamp() and greptime_value), instread of resolving them from table schema. Reads would fail if the table used custom names such as `custom_ts` and `custom_value`. Query planning could fail if `greptime_timestamp` did not exist. 

This change will make remote reads honour custom column names:
- Resolve the time-index and value-field column names from the table schema.
- Pass the resolved timestamp name into query planning for range filtering.
- Carry both resolved names through query execution and result conversion.
- Use the resolved columns when constructing Prometheus samples.
- Exclude the resolved timestamp and value columns from the generated label set.
- Add a remote-read regression test with `custom_ts`/ `custom_value`.

There are no wire-protocol, persisted-data, or user-facing schema changes.

## PR Checklist


src/frontend/src/instance.rs
```diff
@@ -2258,6 +2261,69 @@ mod tests {
...
+            }],
+            timeseries.labels
+        );
+        assert_eq!(
+            vec![Sample {
+                value: 2.0,
+                timestamp: 2000,
+            }],
+            timeseries.samples
+        );
```
v0y4g3r, line 2799: **The new regression test is missing its function ending and prevents compilation.**

After this assertion, `test_prom_remote_read_with_custom_timestamp_and_value_columns` does not return `Ok(())` or close its function body. Consequently, the following `test_event_recorder_is_exposed` declaration is parsed inside this function, and its original `#[tokio::test]` attribute has also been consumed by the new test.

Please add `Ok(())` and the closing brace here, and restore `#[tokio::test]` before `test_event_recorder_is_exposed`.

v0y4g3r (commented):
Static review of the custom-column Prometheus remote-read changes. I found two correctness issues that should be addressed before merge. No local build or test commands were run as part of this review.

src/frontend/src/instance/prom_store.rs
```diff
@@ -171,26 +188,50 @@ impl Instance {
...
+            .schema()
+            .timestamp_column()
+            .with_context(|| ColumnNotFoundSnafu {
+                msg: format!("time index in table '{table_name}'"),
+            })?
+            .name
+            .clone();
+        let value_column_name = table
+            .field_columns()
+            .next()
```
v0y4g3r: **Selecting the first field can return incorrect sample values for existing multi-field tables.**

`Table::field_columns()` returns every non-primary-key, non-time-index column in schema order; it does not identify a Prometheus value column. Before this change, remote read explicitly selected `greptime_value`. With this code, a table whose first field is another column and whose later field is `greptime_value` will silently use the wrong field as the sample value and expose the real `greptime_value` as a label.

The remote-write alignment logic in `src/operator/src/insert.rs` only maps a custom field name when the table has exactly one field, so there is no corresponding first-field convention for multi-field tables. Please use the sole field when there is exactly one; otherwise preserve `greptime_value` when present, and return an explicit ambiguity error if no value column can be uniquely identified. A regression test with an extra field before `greptime_value` would also protect the existing behavior.

src/frontend/src/instance.rs
grezzko: Fixed in 6e1f76d1f. The test now returns Ok(()), closes correctly, and #[tokio::test] is restored on the following test.

src/frontend/src/instance/prom_store.rs
grezzko: Fixed in 6e1f76d1f. 
Remote reads will now select the sole column when only one exists. Look for a default value when there are multiple fields, and return am ambiguous column error when the default field is not found.
For this an error type has been added. 
Added regression coverage for both multi-field cases: when there is an extra field before default value, and when the value column cannot be unambiguously identified, because of missing default value.
The three targeted tests pass.

grezzko:
Hey, thanks for the feedback! 
I have fixed the issues as directed.

v0y4g3r (approved):

PR #8661, opened 2026-07-28
fix(json2): standardize widening and projection cast semantics

fengys1996:

JSON2 requires different conversion semantics at different stages:

- Write-time, flush-time, and schema alignment must not lose data.
- Query-time projection may select only part of a JSON value and perform explicit casts.

Previously, these scenarios shared similar alignment and casting logic, which could silently discard fields during schema alignment or produce inconsistent query-time cast behavior.

This PR separates the two operations:

- `JsonArray::widen_to` provides lossless schema widening.
- `JsonArray::project_to` provides query-time projection with lossy casts.

### Lossless JSON2 widening

`JsonArray::widen_to` is used during write-time and flush-time alignment. It:

- recursively aligns structs and lists;
- fills fields missing from the source with typed null arrays;
- requires every source field to exist in the target schema;
- rejects conversions that could lose information;
- encodes concrete JSON values as JSONB when conflicting types are merged into `Variant`.

Conflicting JSON number categories, such as `UInt64` and `Int64`, or integer and floating-point values, are now merged into `Variant` instead of a common Arrow numeric type. This preserves large integers and avoids precision loss.

JSON2 write-time and flush-time alignment paths now use `widen_to`.

### Query-time JSON2 projection

`JsonArray::project_to` is used during query evaluation. It:

- discards source fields not requested by the target schema;
- fills missing fields with typed SQL `NULL`;
- converts incompatible values to SQL `NULL`;
- recursively projects structs and lists;
- follows existing GreptimeDB scalar cast semantics;
- uses Arrow casts as a fast path for compatible scalar types;
- decodes JSONB only when JSON-aware projection is required.

Read compatibility and projection paths now use `project_to`.

### Reimplement `json_get` through `JsonArray`

The JSON2 implementation of `json_get` now uses `JsonArray::project_to`, replacing its separate value builders and slow-path implementation.

When an intermediate field is stored as JSONB because of conflicting physical types, a nested projection type is constructed from the remaining path. For example, querying:

```sql
json_get(j, 'a.b')
```

may construct a projection type equivalent to:

```text
Struct<a: Struct<b: TargetType>>
```

The JSONB value is decoded and projected through `project_to`, after which field traversal continues normally.

The legacy JSON/JSONB implementation is not intentionally changed.

### Reject whole-column JSON2 queries

A JSON2 column may have an empty declaration-time struct schema. Projecting physical JSON2 values onto this empty struct would silently discard all runtime fields and return `{}`.

The optimizer now rejects whole-column JSON2 queries such as:

```sql
SELECT j FROM json2_table;
SELECT * FROM json2_table;
```

and recommends using `json_get` to select individual fields.

The validation checks the final logical-plan output schema, so it also handles plans where an identity projection has been removed or an `ORDER BY` node is present.

### Other changes

- Removes implicit JSON alignment from `RecordBatch`.
- Fixes `Int32Type::try_cast` returning values with incorrect variants after successful casts.
- Updates type inference expectations for mixed numeric JSON arrays.
- Removes obsolete JSON projection builders and intermediate abstractions.
- Adds unit tests for scalar, struct, list, nested JSONB, conflicting-type, and incompatible projections.
- Adds SQLness coverage for:
  - query-time scalar casts;
  - parent-object projection;
  - different physical JSON2 layouts across SSTs;
  - lossless write-time alignment;
  - unsupported whole-column JSON2 queries.

### Compatibility

This PR does not change persisted or wire formats.

JSON1/JSONB behavior is not intentionally changed. The behavior changes are limited to JSON2:

- write-time and flush-time alignment no longer silently loses data;
- query-time casts use unified projection semantics;
- querying a whole JSON2 column now returns an explicit error instead of an incorrect empty object.

## PR Checklist


fengys1996:
@codex review

fengys1996:
@codex review

src/query/src/optimizer/json_type_concretize.rs
```diff
@@ -72,6 +75,18 @@ impl OptimizerRule for JsonTypeConcretizeRule {
     }
 }
 
+fn ensure_no_whole_json2_output(plan: &LogicalPlan) -> Result<()> {
+    for field in plan.schema().fields() {
+        if is_structured_json_field(field) {
```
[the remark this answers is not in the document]
fengys1996: Good catch!

src/common/function/src/scalars/json/json_get.rs
```diff
@@ -681,8 +580,8 @@ mod tests {
             ("$.c", None),
             ("$.kind", None),
             ("$.payload.code", Some(404)),
-            ("$.payload.success", None),
-            ("$.payload.result.time_cost", None),
+            ("$.payload.success", Some(0)),
```
MichaelScofield, line 583: If "$.payload.success" is true, will it return `Some(1)`?

src/datatypes/src/types/primitive_type.rs
```diff
@@ -384,11 +384,11 @@ impl DataType for Int32Type {
             Value::Int8(v) => num::cast::cast(v).map(Value::Int32),
             Value::Int16(v) => num::cast::cast(v).map(Value::Int32),
             Value::Int32(v) => Some(Value::Int32(v)),
-            Value::Int64(v) => num::cast::cast(v).map(Value::Int64),
+            Value::Int64(v) => num::cast::cast(v).map(Value::Int32),
```
fengys1996, line 387: @evenyag ptal about this change. I'm not sure if this change is right, or if the previous design had some special considerations.

tests/cases/standalone/common/types/json/json2.result
```diff
@@ -8,30 +8,6 @@ create table json2_table (
 
 Affected Rows: 0
 
-insert into json2_table (ts, j) values (101, '[1, 2, 3]');
```
MichaelScofield, line 11: Why remove these test cases?

fengys1996: Already moved the case to `json2_limit.sql`.

src/common/function/src/scalars/json/json_get.rs
fengys1996: Yes. Boolean-to-numeric projection follows Arrow/JSONB cast semantics: true becomes 1 and false becomes 0. This fixture contains false, hence Some(0).

MichaelScofield: better document this somewhere, because other database like ck behaves different

src/datatypes/src/types/primitive_type.rs
evenyag: Looks like the fix is correct.

MichaelScofield (approved):

PR #8648, opened 2026-07-27
feat(event-recorder): configure lifecycle event recording

WenyXu:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

None.

## What's changed and what's your intention?

This PR records procedure lifecycle events in the events system table and adds `event_recorder.event_types` filtering for metasrv and standalone deployments.

- Centralize event-table schema helpers, nullable values, and test helpers in `common-event-recorder` while preserving the existing event-table and procedure-envelope layouts.
- Install the configured recorder in the procedure manager and expose its filter through `EventContext`, so procedure hooks skip event construction when the event type is disabled.
- Add `event_recorder.event_types` configuration:
  - Omitted: retain all current and future event types.
  - Empty list: disable procedure event recording.
  - Non-empty list: retain only listed event types.
- Expose the setting in metasrv and standalone configuration examples and regenerate the configuration reference. The currently available procedure event type is `region_migration`.
- Keep slow-query recording independent from procedure-event filtering: frontend/standalone create a dedicated unfiltered slow-query recorder with `slow_query.ttl`, while the procedure recorder uses `event_recorder.ttl` and `event_types`.
- Add configuration and integration snapshot coverage, including the serialized event-recorder section and default TTL behavior.

This change is additive and does not require a data migration. Schema or data compatibility is preserved.

Validation:
- `cargo fmt --all` passed.
- `git diff --check` passed.
- Added unit and integration coverage for event-table contracts, filter deserialization/filtering, procedure event rows, procedure-manager filter wiring, and configuration serialization.
- `cargo nextest run` and `make clippy` were attempted but did not complete because the build directory was locked; the PR remains draft pending successful checks.

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [x] This PR requires documentation updates.
- [ ] API changes are backward compatible.
- [x] Schema or data changes are backward compatible.

PR #8632, opened 2026-07-24
feat: add procedure events for Flow DDL

WenyXu:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

- Builds on #8648 and aligns with #8623.

## What's changed and what's your intention?

Add procedure-event recording for `CREATE FLOW` and `DROP FLOW`.

- Adds `create_flow` and `drop_flow` events with canonical Flow dimensions (`catalog_name`, `flow_name`, and `flow_id`) registered in the shared event-table schema.
- Records bounded submitted payloads and lightweight lifecycle rows; a successful create event contains only the resolved `flow_id`.
- Keeps sensitive or unbounded Flow definition details out of event payloads, including SQL text, comments, source/sink names, options, and query-context extensions.
- Organizes Flow DDL event construction under `ddl/event/flow.rs`, alongside the existing Database DDL event producer.
- Reuses a shared integration-test helper and matches the Database DDL event assertion style: submitted and lifecycle rows are correlated through `procedure_id`.
- Removes query-context `schema_name` from Flow event records because it is not part of Flow identity.
- Documents the new event types in standalone and metasrv recorder configuration, including regenerated configuration documentation.

Validation completed:

- `cargo fmt --all -- --check`
- `git diff --check`
- `make config-docs`
- `cargo check -p tests-integration --tests`
- `cargo clippy -p common-event-recorder -p common-meta --tests -- -D warnings`
- `cargo clippy -p tests-integration --test main -- -D warnings`
- `cargo nextest run -p common-meta 'ddl::tests::event::flow'`
- `cargo nextest run -p common-event-recorder event_table`

The File-storage Flow integration test was attempted but could not start because the local test cluster exhausted its available port pool (`No unused port found`); this is not reported as passing validation.

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [x] This PR requires documentation updates.
- [x] API changes are backward compatible.
- [x] Schema or data changes are backward compatible.

PR #8623, opened 2026-07-23
feat: add database DDL procedure events

WenyXu:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

- #8648

## What's changed and what's your intention?

Add procedure lifecycle events for `CREATE DATABASE`, `ALTER DATABASE`, and `DROP DATABASE` on top of the shared event-recorder infrastructure from #8648.

- Emit `create_database`, `alter_database`, and `drop_database` events from the corresponding MetaSrv procedures, respecting the recorder's shared event-type filter.
- Submitted events include a versioned DDL-intent payload plus catalog and schema locator columns; lifecycle events preserve the schema and emit null locators with a JSON-null payload.
- Organize DDL event sources under the singular `ddl::event` module, with database events in `event/database.rs`; this leaves a per-domain extension point for table and flow events.
- Reuse #8648's event-table definitions and `assert_event_contract` test utility for unit contract coverage.
- Add MetaSrv-backed and standalone integration coverage that queries the recorded rows in `greptime_private.events`, uses polling instead of a fixed wait, and disables unrelated SQL event recording in sqlness tests.
- Document all selectable event types in the MetaSrv and standalone configuration examples, then regenerate `config/config.md`.
- No user-facing API, persisted metadata, or wire-format compatibility changes. Documentation is intentionally limited to generated configuration reference updates.

Validation completed:

- `cargo fmt --all`
- `make fmt-toml`
- `make clippy`
- `make config-docs`

The relevant `cargo nextest` run is pending because the shared build directory was locked.

## PR Checklist
Please convert it to a draft if some of the following conditions are not met.

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [x] This PR requires documentation updates.
- [x] API changes are backward compatible.
- [x] Schema or data changes are backward compatible.

PR #8665, opened 2026-07-28
feat(metasrv): add repartition lifecycle events

WenyXu:

Builds on #8648 and follows the database and Flow DDL event conventions from #8623 and #8632.

- Add `repartition` and `repartition_group` lifecycle events and expose both through the existing metasrv event-type filter configuration.
- Record repartition intent with table identity, partition source/target information, and one submitted topology row per source-to-target region mapping. Queryable first-level fields include parent/group procedure IDs, table identity, source/target region IDs and numbers, and source/target partition expressions.
- Preserve the distinction between a parent's `ChildSubmitted` notification and the child's own `Submitted` lifecycle event.
- Build initial procedure events through a reusable Runner helper, then dispatch them only after `spawn_runner_task` accepts the root or child runner. This prevents phantom child Submitted/topology rows when shutdown makes the child submission return `SpawnFailed`.
- Add a deterministic shutdown-race regression test for `SpawnFailed`, while retaining parent/child lifecycle correlation coverage and repartition integration coverage.
- Centralize reusable event-table column definitions in `common-event-recorder`, and migrate region-migration events to use the shared definitions.
- Rename the metasrv event module to singular `event` and use domain-named children (`region_migration`, `repartition`). This intentionally changes the public Rust import path; downstream users should migrate from `meta_srv::events::*` to `meta_srv::event::*`.
- Preserve compatibility for persisted repartition-group contexts by defaulting newly added optional fields, and keep unsupported repartition and region-migration filter types out of standalone configuration documentation.

The procedure event dispatch is best effort and is not part of procedure execution or persistence. Persisted repartition-group contexts remain backward compatible through `#[serde(default)]` optional fields.

Validation completed successfully:

- `cargo clippy -p common-procedure --all-targets -- -D warnings`
- `cargo nextest run -p common-procedure --lib --retries 0` (78 tests)
- `cargo nextest run -p tests-integration test_repartition_event --retries 0` (1 test)

## PR Checklist


WenyXu:
@codex review

WenyXu:
@codex review

WenyXu:
@codex review

src/common/procedure/src/local/runner.rs
```diff
@@ -536,6 +536,8 @@ impl Runner {
             procedure_id,
         );
 
+        runner.record_event(EventTrigger::Submitted);
```
fengjiachun: Could we emit `Submitted` only after `spawn_runner_task` accepts the child? If shutdown happens after the initial `running()` check but before the locked spawn check, this event is recorded even though spawning returns `false`, the child metadata is removed, and the parent reports `SpawnFailed`, leaving a phantom submitted lifecycle/topology record. A negative test for this race would help.

src/meta-srv/src/lib.rs
```diff
@@ -20,7 +20,7 @@ pub mod cache_invalidator;
 pub mod cluster;
 pub mod discovery;
 pub mod error;
-pub mod events;
+pub mod event;
```
fengjiachun, line 23: Could we keep a compatibility re-export for `meta_srv::events`? Replacing this public module with `event` breaks downstream imports such as `meta_srv::events::EventHandlerImpl` and `meta_srv::events::region_migration_event::*`; updating in-repo callers does not preserve the exported API.

WenyXu: This rename is intentional, and I would prefer not to keep compatibility re-exports. meta-srv uses the singular event module, with child modules named after their domains (region_migration and repartition) rather than with a redundant _event suffix. I updated all in-repo imports accordingly. Adding pub use event as events and legacy child-module aliases would leave two public paths for the same internal implementation and preserve the naming convention this change is meant to remove. Downstream users should update their imports to meta_srv::event::EventHandlerImpl and meta_srv::event::region_migration::*.

src/common/procedure/src/local/runner.rs
WenyXu: Good catch, resolved in [cffb628](https://github.com/GreptimeTeam/greptimedb/pull/8665/commits/cffb628f9b29540f26cca0d1f04f99fd25808b57)

src/common/procedure/src/local/runner.rs
```diff
@@ -163,6 +163,8 @@ impl Runner {
             self.meta.id
         );
 
+        self.record_event(trigger);
```
fengjiachun: Could we record the initial lifecycle event synchronously in the accepted spawn path instead of inside `Runner::run`? `spawn_runner_task` can return `true`, but `LocalManager::stop()` may abort that `JoinHandle` before the future is first polled, so this line is never reached and an already accepted procedure loses its `Submitted`/`Recovered` event. For a child, that can also leave `ChildSubmitted(..., outcome=Accepted)` without the child’s `Submitted` row. The new test covers rejection before spawn, but not accepted-then-aborted.

WenyXu: [7ba3931](https://github.com/GreptimeTeam/greptimedb/pull/8665/commits/7ba39314e7c87c9f454b460143615d2aee5f69e1)

fengjiachun (approved):
LGTM

PR #8669, opened 2026-07-28
feat: support time range in manual compaction

v0y4g3r:

- Depends on https://github.com/GreptimeTeam/greptime-proto/pull/332

This PR allows `ADMIN COMPACT_TABLE` callers to constrain manual compaction to a time range through the existing options string:

```sql
ADMIN COMPACT_TABLE(
    't',
    'regular',
    'parallelism=2,start_time=2026-01-01T00:00:00Z,end_time=2026-02-01T00:00:00Z'
);
```

`start_time` and `end_time` must be provided together and define a half-open `[start_time, end_time)` range. Timestamps without an explicit timezone are interpreted in the query session timezone.

The range is propagated through the table request, the protobuf `CompactRequest`, the region request, and the Mito compaction scheduler to the compaction picker. Wire timestamps are normalized to seconds by rounding the start down and the end up, preserving every user-selected timestamp while matching region compaction window granularity.

Both manual compaction strategies are supported:

- `regular` keeps the existing TWCS trigger rules and filters candidate windows before applying the background-task limit.
- `strict_window` keeps forced-compaction semantics. If a selected SST spans multiple windows, the picker transitively expands the selected window set so every output that shares that input SST is retained and no out-of-range data is lost.

A ranged regular request receives its own pending manual-compaction slot when the region is already compacting, so the range is preserved when the request is rescheduled.

The protobuf change is additive and covered by forward/backward wire compatibility tests in the dependency PR. Existing callers that omit the range retain the previous behavior.

Verification performed:

- `cargo sqlness bare -t flush_compact_table`
- Targeted unit tests for the parser, protobuf conversion, region request conversion, scheduler, TWCS picker, and strict-window picker
- Clippy with `-D warnings` for the affected crates
- `make fmt`
- `git diff --check`

Current limitation: strict-window compaction may intentionally expand beyond the requested range when cross-window SST dependencies require it for data safety. Documentation updates should describe this behavior after the API is finalized.

## PR Checklist


src/operator/src/request.rs
```diff
@@ -145,13 +150,18 @@ impl Requester {
...
             .iter()
             .map(|partition| {
                 RegionRequestBody::Compact(CompactRequest {
                     region_id: partition.id.into(),
                     parallelism: request.parallelism,
                     options: Some(request.compact_options),
+                    time_range,
                 })
             })
             .collect();
```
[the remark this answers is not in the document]
v0y4g3r: No code change is needed here. Both `compact_request::Options` and `CompactionTimeRange` derive `Copy` in the generated `greptime-proto` bindings, so `Option<Options>` and `Option<CompactionTimeRange>` are copied into each region request rather than moved by the first iterator call. The current PR head also compiles successfully with all-target Clippy and `-D warnings`.

v0y4g3r:
@codex review

src/mito2/src/compaction/window.rs
```diff
@@ -123,6 +132,53 @@ impl Picker for WindowedCompactionPicker {
     }
 }
 
+fn filter_time_windows(
+    mut windows: BTreeMap<i64, (i64, Vec<FileHandle>)>,
+    time_range: Option<TimestampRange>,
+) -> BTreeMap<i64, (i64, Vec<FileHandle>)> {
```
evenyag, line 145: Can we filter the files by the time range before grouping them instead of filtering time windows by the time range?

It should be easier for users to understand and simplify the implementation.

src/common/function/src/admin/flush_compact_table.rs
```diff
@@ -187,118 +194,115 @@ fn parse_compact_request(
...
+            "start_time" => {
+                start_time = Some(Timestamp::from_str(value, Some(timezone)).map_err(|_| {
+                    InvalidFuncArgsSnafu {
+                        err_msg: format!("Invalid value for start_time: {}", value),
+                    }
+                    .build()
+                })?);
+            }
+            "end_time" => {
+                end_time = Some(Timestamp::from_str(value, Some(timezone)).map_err(|_| {
```
evenyag, line 260: Should we only accept timestamp to avoid the timezone issue?

src/mito2/src/compaction.rs
```diff
@@ -525,6 +557,7 @@ impl CompactionScheduler {
             plan_id,
             request,
             compact_request::Options::Regular(Default::default()),
+            None,
```
discord9: **Blocking:** This continuation drops the range after the first ranged execution. `handle_compaction_finished` calls `schedule_next_compaction` whenever the default zero `min_compaction_interval` has elapsed, and this path always dispatches `Regular` with `time_range: None`. A ranged regular request limited by `max_background_compactions` can therefore continue into unrelated windows; ranged strict-window also triggers an unrestricted regular pass. Please either stop chaining after a ranged manual request or retain the original range for every continuation, and cover the worker-level completion path with in-range and out-of-range windows.

src/mito2/src/compaction/window.rs
```diff
@@ -123,6 +132,53 @@ impl Picker for WindowedCompactionPicker {
...
+                .end()
+                .is_none_or(|range_end| window_start < range_end);
+            let ends_after_range_start = time_range
+                .start()
+                .is_none_or(|range_start| range_start < window_end);
+            (starts_before_range_end && ends_after_range_start).then_some(*lower_bound)
+        })
+        .collect::<HashSet<_>>();
+
+    loop {
```
discord9: **Non-blocking performance concern:** This fixed-point loop rebuilds `selected_files` and rescans every window on each round. With a chain of SSTs connecting adjacent windows, each round can add only one window, making planning quadratic in the number of windows. Please consider precomputing `FileId -> windows` adjacency and using a queue-based traversal; a chain-shaped test would also protect this case.

src/mito2/src/compaction/twcs.rs
```diff
@@ -381,6 +411,26 @@ fn assign_to_windows<'a>(
...
+fn time_window_intersects_range(
+    window_end: i64,
+    time_window_size: i64,
+    time_range: &TimestampRange,
+) -> bool {
+    let first_window = time_range.start().map_or(i64::MIN, |start| {
+        start
+            .convert_to(TimeUnit::Second)
+            .and_then(|timestamp| timestamp.value().align_to_ceil_by_bucket(time_window_size))
+            .unwrap_or(i64::MIN)
```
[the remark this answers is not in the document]
v0y4g3r: Fixed in 27235fc444. TWCS alignment failures now make the range non-matching instead of widening either endpoint to an unbounded value. The regression test covers a valid `[i64::MAX - 1, i64::MAX)` range whose 4-second bucket alignment overflows.

src/mito2/src/compaction/window.rs
v0y4g3r: No code change here. The intended API semantics select complete compaction windows that overlap `[start_time, end_time)`, rather than selecting individual SSTs. Filtering SSTs first would change that behavior and is unsafe for strict-window compaction when one cross-window SST is shared by multiple output windows; the picker therefore selects windows first and expands the dependency closure.

src/common/function/src/admin/flush_compact_table.rs
v0y4g3r: No code change here. The third argument is the existing string-based options interface, so `start_time` and `end_time` are parsed from that string. Inputs with an explicit offset keep that offset; inputs without one deliberately use the query session timezone, matching SQL timestamp interpretation. Restricting the option to an epoch value or requiring an offset would change the agreed API semantics.

src/mito2/src/compaction.rs
v0y4g3r: Fixed in 7f24a94741. `CompactionStatus` now retains the range for worker-triggered continuation planning and passes it to every subsequent picker cycle. An unrelated regular trigger still resets the continuation to an unrestricted regular cycle. The regression test simulates the completion/continuation path with in-range and out-of-range windows and verifies that the continuation never selects the latter.

src/mito2/src/compaction/window.rs
v0y4g3r: Fixed in 0defcb7f19. The fixed-point rescans were replaced with a precomputed `FileId -> windows` adjacency map and a queue traversal, so each selected window, file, and incidence is visited once. A 128-SST chain test covers transitive expansion across 129 windows.

evenyag:
@codex review

evenyag (approved):

src/mito2/src/compaction/window.rs
```diff
@@ -123,6 +132,59 @@ impl Picker for WindowedCompactionPicker {
...
+            if !visited_files.insert(file.file_id()) {
+                continue;
+            }
+            for dependent_window in &file_windows[&file.file_id()] {
+                if selected_windows.insert(*dependent_window) {
+                    pending_windows.push_back(*dependent_window);
+                }
+            }
+        }
+    }
```
evenyag, line 189: I think we can explain why we need this in a comment.

PR #8672, opened 2026-07-29
fix: enforce permissions for restricted HTTP endpoints

shuiyisong:

This pull request fixes missing authorization checks across the Jaeger, pipeline, and dashboard APIs.

It adds explicit query and management permission requests, applies target-aware authorization to Jaeger queries, and protects pipeline and dashboard operations according to their access mode. Pipeline dry-runs now check authorization before resolving or executing pipelines, including recursively dispatched pipelines, while normal ingestion remains unchanged.

This extends `PermissionReq` and `PipelineHandler`, so downstream exhaustive matches and handler implementations must adopt the new variants and method. The branch contains one commit across seven files, with a clean working tree and no untracked files.

## PR Checklist

### DIFF
--- a/src/auth/src/permission.rs
+++ b/src/auth/src/permission.rs
@@ -105,6 +105,11 @@ pub enum PermissionReq<'a> {
     PromStoreRead,
     Otlp,
     LogWrite,
+    JaegerQuery,
+    PipelineQuery,
+    PipelineManage,
+    DashboardQuery,
+    DashboardManage,
     BulkInsert {
         catalog: &'a str,
         schema: &'a str,
@@ -119,9 +124,12 @@ impl<'a> PermissionReq<'a> {
             PermissionReq::GrpcRequest(Request::Query(query_request)) => {
                 !matches!(query_request.query, Some(Query::InsertIntoPlan(_)))
             }
-            PermissionReq::PromQuery | PermissionReq::LogQuery | PermissionReq::PromStoreRead => {
-                true
-            }
+            PermissionReq::PromQuery
+            | PermissionReq::LogQuery
+            | PermissionReq::PromStoreRead
+            | PermissionReq::JaegerQuery
+            | PermissionReq::PipelineQuery
+            | PermissionReq::DashboardQuery => true,
             PermissionReq::SqlStatement(stmt) => stmt.is_readonly(),
 
             PermissionReq::GrpcRequest(_)
@@ -130,6 +138,8 @@ impl<'a> PermissionReq<'a> {
             | PermissionReq::PromStoreWrite
             | PermissionReq::Otlp
             | PermissionReq::LogWrite
+            | PermissionReq::PipelineManage
+            | PermissionReq::DashboardManage
             | PermissionReq::BulkInsert { .. } => false,
         }
     }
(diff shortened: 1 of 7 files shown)


src/frontend/src/instance.rs
```diff
@@ -1521,9 +1521,9 @@ pub fn check_permission(
         }
         // cursor operations are always allowed once it's created
         Statement::FetchCursor(_) | Statement::CloseCursor(_) => {}
-        // User can only kill process in their own catalog.
+        // KILL is scoped to the current catalog by the executor.
         Statement::Kill(_) => {}
-        // SHOW PROCESSLIST
+        // Process-control authorization is checked before statement validation.
```
sunng87: unrelated comment change

src/frontend/src/instance/log_handler.rs
sunng87, line 101: this function is not protected by permission check.

is it designed for users with pipeline query permission but can still use pipeline?

src/servers/src/http/event.rs
```diff
@@ -240,9 +240,10 @@ pub async fn query_pipeline_ddl(
     query_ctx.set_channel(Channel::Log);
     let query_ctx = Arc::new(query_ctx);
 
-    let pipeline = handler
-        .get_pipeline(&pipeline_name, version, query_ctx.clone())
+    let (pipeline, _) = handler
+        .get_pipeline_str(&pipeline_name, version, query_ctx.clone())
         .await?;
+    let pipeline = handler.build_pipeline(&pipeline)?;
```
sunng87: if we have get_pipeline protected by permission check, can we avoid this change?

src/auth/src/permission.rs
```diff
@@ -105,6 +105,11 @@ pub enum PermissionReq<'a> {
     PromStoreRead,
     Otlp,
     LogWrite,
+    JaegerQuery,
+    PipelineQuery,
+    PipelineManage,
+    DashboardQuery,
+    DashboardManage,
```
sunng87, line 112: can be postpone to next patch: this permission enum may not be able to enumerate all possible actions, for example, we may have additional ones from enterprise version. So what about using a text based action name for this, which will be more extensible.

src/frontend/src/instance/dashboard.rs
```diff
@@ -52,6 +53,19 @@ pub const DASHBOARD_TABLE_DEFINITION_COLUMN_NAME: &str = "definition";
...
+        ctx: &QueryContextRef,
+        req: PermissionReq<'_>,
+    ) -> servers::error::Result<()> {
+        self.plugins
+            .get::<PermissionCheckerRef>()
+            .as_ref()
+            .check_permission(ctx.current_user(), req)
+            .context(AuthSnafu)?;
+        Ok(())
+    }
```
sunng87: this implementation has no additional information about dashboard, can we inline it, or just name it as `check_permission` and lift it?

src/frontend/src/instance/log_handler.rs
```diff
@@ -34,6 +34,19 @@ use table::Table;
...
+        ctx: &QueryContextRef,
+        req: PermissionReq<'_>,
+    ) -> ServerResult<()> {
+        self.plugins
+            .get::<PermissionCheckerRef>()
+            .as_ref()
+            .check_permission(ctx.current_user(), req)
+            .context(AuthSnafu)?;
+        Ok(())
+    }
```
sunng87: similar to dashboard

src/frontend/src/instance/log_handler.rs
shuiyisong: Yes, this is intentional (assuming you meant users without pipeline-query permission). `get_pipeline` loads a compiled pipeline for ingestion execution. A user with write-only permission must be able to ingest through an existing pipeline; requiring `PipelineQuery` here would also require read permission for normal HTTP/OTLP/Splunk/Loki ingestion and recursive dispatch. Pipeline inspection and preview paths are protected separately: `get_pipeline_str`, DDL, and dry-run check `PipelineQuery`, while final writes still go through `LogWrite` and table-target permission checks. I will add a code comment to document this contract.

src/auth/src/permission.rs
shuiyisong: Good catch, we'll address this issue in a separate PR

sunng87 (approved):

PR #8673, opened 2026-07-29
feat(metasrv): add batch GC lifecycle events

WenyXu:

Follows the procedure event conventions established in https://github.com/GreptimeTeam/greptimedb/pull/8648, https://github.com/GreptimeTeam/greptimedb/pull/8623, https://github.com/GreptimeTeam/greptimedb/pull/8632, and https://github.com/GreptimeTeam/greptimedb/pull/8627.

- Add `batch_gc` procedure events for `BatchGcProcedure`, following the existing procedure-event contract.
- Record successful GC outcomes per Region only when files or indexes are deleted, or a retry is required. No-op and route-skipped Regions do not emit success rows.
- Store successful deletion/retry outcomes in JSONB `gc_report`; preserve the complete deleted-file list and omit empty deletion arrays.
- Record Batch GC configuration in procedure-level diagnostic events for retry and rollback transitions, and for failure or poison transitions without a meaningful GC report. Terminal failures with a report emit sparse region-level report rows instead. Suppress normal Submitted and Recovered events to keep the normal path low-noise.
- Add event-contract, filtering, procedure-envelope, and file-cluster integration coverage.
- Document `batch_gc` only in the MetaSrv event-recorder configuration; standalone does not schedule Batch GC.
- This change does not alter `GcReport` persistence/wire format or public `Event`/`Procedure` APIs.

Validation completed:

- `cargo fmt --all`
- `make config-docs`
- `cargo nextest run -p meta-srv batch_gc`
- `cargo clippy -p meta-srv --all-targets -- -D warnings`
- `cargo nextest run -p tests-integration --test main --retries 0 batch_gc_event`
- `cargo clippy -p tests-integration --test main -- -D warnings`

## PR Checklist


WenyXu:
@codex review

src/meta-srv/src/gc/procedure.rs
```diff
@@ -1024,4 +1025,35 @@ impl Procedure for BatchGcProcedure {
...
+
+    fn event(&self, ctx: &EventContext<'_>) -> Option<Box<dyn common_event_recorder::Event>> {
+        if !ctx.event_type_filter.allows(BATCH_GC_EVENT_TYPE) {
+            return None;
+        }
+
+        let event = match &ctx.trigger {
+            EventTrigger::Submitted
+            | EventTrigger::Recovered
+            | EventTrigger::ChildSubmitted { .. } => return None,
```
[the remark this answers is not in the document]
WenyXu: This is intentional. Batch GC suppresses Submitted/Recovered to keep the normal path limited to per-region Succeeded report rows. The request context (regions, full_file_listing, and timeout) is recorded only for retry/failure lifecycle events, where it is useful for diagnosis.

config/standalone.example.toml
```diff
@@ -907,7 +907,7 @@ ttl = "90d"
 ## `alter_database`, `drop_database`, `create_flow`, `drop_flow`,
 ## `create_table`, `create_logical_tables`, `alter_table`, `alter_logical_tables`,
 ## `drop_table`, `undrop_table`, `purge_dropped_table`, `truncate_table`,
-## `create_view`, `drop_view`.
+## `create_view`, `drop_view`, `batch_gc`.
```
fengjiachun: standalone never emits `batch_gc` events, so this line (and the corresponding standalone section in `config/config.md`) should be dropped — only the metasrv list needs updating.

Why: the only place that submits `BatchGcProcedure` is `DefaultGcSchedulerCtx`, and `GcScheduler` is only constructed at `src/meta-srv/src/metasrv/builder.rs:529` behind the `options.gc.enable` gate. `src/cmd/src/standalone.rs` never builds a Metasrv at all, and `config/standalone.example.toml` has no `[gc]` section.

These two lists are deliberately maintained per reachable-event-set — the absence of `region_migration` / `repartition` / `repartition_group` from the standalone list is the evidence.

src/meta-srv/src/event/gc.rs
```diff
@@ -0,0 +1,525 @@
...
+        .into_iter()
+        .flatten()
+        .map(|(file_id, index_version)| DeletedIndexPayload {
+            file_id: file_id.to_string(),
+            index_version: *index_version,
+        })
+        .collect::<Vec<_>>();
+
+    let need_retry = report.need_retry_regions.contains(&region_id);
+    if deleted_files.is_empty() && deleted_indexes.is_empty() && !need_retry {
```
fengjiachun, line 197: Returning `None` here writes `gc_report` as NULL, but two semantically very different situations land in this branch — and they happen to be exactly the two an operator needs to tell apart:

1. **Processed fine, nothing to delete** — `src/mito2/src/gc.rs:352` takes the else branch and does `deleted_files.insert(region, vec![])` + `processed_regions.insert(region)`, so all three conditions here hold → NULL.
2. **Never processed at all** — `discover_route_for_regions` (`src/meta-srv/src/gc/procedure.rs:498`) hits `continue` when `get_table_route` fails, silently skipping every region of that table. Those regions never enter `region_routes`, are never sent to a datanode, and never appear in the report — yet the procedure still **succeeds** (`first_error` stays None) → also NULL.

Case 2 is real, e.g. when a table gets dropped concurrently. I'd suggest emitting a full `{"processed": true, "need_retry": false}` for the no-op case and reserving NULL for "absent from the report" — or adding an explicit `outcome` enum that distinguishes processed / skipped / need_retry.

src/meta-srv/src/event/gc.rs
```diff
@@ -0,0 +1,525 @@
...
+        return None;
+    }
+
+    Some(BatchGcRegionReport {
+        deleted_files,
+        deleted_indexes,
+        processed: report.processed_regions.contains(&region_id),
```
fengjiachun: `processed` is currently equivalent to `!need_retry` — in `src/mito2/src/gc.rs` the `processed_regions.insert` and `deleted_files.insert` calls sit in the same else branch, so it's necessarily true whenever there are deletions and false whenever a retry is needed. On its own it carries no information; it only becomes meaningful alongside the NULL-ambiguity fix above.

src/meta-srv/src/gc/procedure.rs
```diff
@@ -1024,4 +1025,35 @@ impl Procedure for BatchGcProcedure {
...
         LockKey::new(lock_key)
     }
+
+    fn event(&self, ctx: &EventContext<'_>) -> Option<Box<dyn common_event_recorder::Event>> {
+        if !ctx.event_type_filter.allows(BATCH_GC_EVENT_TYPE) {
+            return None;
+        }
+
+        let event = match &ctx.trigger {
+            EventTrigger::Submitted
```
fengjiachun, line 1075: With `Submitted` returning `None`, `full_file_listing` and `timeout` are no longer recorded anywhere on the success path — only Retrying / Failed / Poisoned carry a payload.

`full_file_listing` is what determines whether orphan files were scanned in a given round, which matters a lot when reconstructing after the fact how a set of files got deleted. I get that cutting Submitted was about controlling fanout, but these two fields could go into the success event's payload instead — it's `Null` today, so adding them costs no extra rows.

Also worth updating the PR description: "Keep submitted Batch GC configuration in the standard payload" no longer holds.

src/meta-srv/src/event/gc.rs
```diff
@@ -0,0 +1,525 @@
...
+
+#[derive(Debug)]
+pub(crate) struct BatchGcEvent {
+    payload: Option<BatchGcPayload>,
+    // None emits a procedure-level lifecycle row with null Region dimensions.
+    regions: Option<Vec<BatchGcRegionRow>>,
+}
+
+impl BatchGcEvent {
+    pub(crate) fn lifecycle(
```
fengjiachun: This naming is inverted relative to `repartition.rs` in the same directory: there, `submitted()` carries the payload and `lifecycle()` returns an all-`None` empty event, whereas here `lifecycle()` is the one carrying the submitted config. Two same-named methods with opposite meanings in one directory is easy to misread — something like `with_config()` would be clearer.

src/meta-srv/src/event/gc.rs
```diff
@@ -0,0 +1,525 @@
...
+            &REGION_ID_COLUMN,
+            &TABLE_ID_COLUMN,
+            &REGION_NUMBER_COLUMN,
+            &GC_REPORT_COLUMN,
+        ])
+    }
+
+    fn extra_rows(&self) -> Result<Vec<Row>> {
+        let Some(regions) = &self.regions else {
+            return Ok(vec![null_row(4)]);
```
fengjiachun: The column count in `null_row(4)` is hardcoded, and the four-column list in `extra_schema()` is duplicated verbatim in the test helper `batch_gc_schema()` (line 573 of this file).

`repartition.rs` factors this into a `fn schema()` shared by production and tests, and writes `null_row(parent_schema().len())` in the tests. Worth doing the same here — otherwise adding a column later leaves the schema and row lengths out of sync, which only surfaces at insert time, and both test expectations need hand-editing, which makes it easy to miss one.

Also, `null_row` is only a test-only helper in `repartition.rs`, and this PR copies it into production code. Since both modules want it, it probably belongs in `common_event_recorder::event_table`.

tests-integration/tests/event_recorder_test_util.rs
```diff
@@ -102,13 +102,28 @@ async fn event_count_is_one(instance: &Arc<Instance>, query: &str) -> bool {
...
+            .next()
+            .flatten()
+            .map(ToString::to_string),
+        ArrowDataType::Utf8View => column
+            .as_string_view()
+            .iter()
+            .next()
+            .flatten()
+            .map(ToString::to_string),
+        _ => None,
```
fengjiachun: Silently returning `None` for an unexpected arrow type degrades into the caller polling for 15 seconds and then panicking with `timed out waiting for event query`, rather than telling you the column type was wrong. A `panic!` here that includes the actual `data_type` would be much easier to diagnose.

tests-integration/tests/gc_event.rs
```diff
@@ -0,0 +1,190 @@
...
+            instance,
+            &format!(
+                "INSERT INTO {TABLE_NAME} VALUES \
+                 ('2023-01-{day:02} 10:00:00', {day}.0, 'host{day}')"
+            ),
+        )
+        .await;
+        execute(instance, &format!("ADMIN FLUSH_TABLE('{TABLE_NAME}')")).await;
+    }
+    assert_sst_count(&cluster, 4).await;
```
fengjiachun, line 84: `deleted_file_ids` is captured after this line, so if mito's automatic compaction fires in between, the captured list is stale and the final `assert_eq!(actual_file_ids, deleted_file_ids)` fails for no obvious reason.

The failure is explicit rather than a silently wrong conclusion, so not a blocker — but disabling automatic compaction at table creation would make this test a lot more deterministic.

fengjiachun (commented):
The overall design lines up with the existing `repartition` / `region_migration` event modules, and I didn't find any functional bugs. I verified locally that `cargo check -p meta-srv --all-targets` passes and all 5 `event::gc::tests` pass. Nice touch that the `event()` match is exhaustive (no `_` arm) — adding a new `EventTrigger` variant will fail to compile rather than silently fall through.

Detailed comments are inline. Two things that don't map to a specific line:

**1. The PR description doesn't match what's here.** It says "Keep submitted Batch GC configuration in the standard payload", but 6a74611 changed `Submitted` to `return None`, so `full_file_listing` / `timeout` never actually land in the table on the success path. Separately, the first three bullets (repartition events, the `event` module refactor, child-procedure Submitted events) aren't in this PR's diff at all — they belong to the lower part of the stack. Worth trimming before merge, otherwise the merge commit claims changes it doesn't contain.

**2. Worth sizing the event volume up front.** One row per region on success. GC ticks every 5 minutes with one batch per datanode, and candidates are "top 20 per table with score > 0" — since `sst_count_weight = 0.5`, a region gets picked just for having SSTs, even with nothing to delete. That produces a lot of rows whose `gc_report` is NULL. Worth doing the math on `greptime_private.events` growth at your real table counts, given the 90d TTL.

The other end of this is that `deleted_files` is unbounded: a full file listing that cleans up orphans can turn up thousands of file IDs for a single region (36-byte UUID strings each), making for a very large single JSONB value. Consider truncating and adding a `deleted_files_count`.

WenyXu:
Thanks for the broader review. We will trim the stacked-PR content and update the description to match the current Submitted-suppression behavior. We will also make success reporting sparse, avoiding one NULL row for every selected region on every GC cycle. We intentionally keep the complete deleted-file list for auditability in this iteration and acknowledge the JSONB-size trade-off.

config/standalone.example.toml
WenyXu: Good catch. `BatchGcProcedure` is MetaSrv-only, so we will remove `batch_gc` from the standalone configuration docs and keep it only in the MetaSrv event-type list.

src/meta-srv/src/event/gc.rs
WenyXu: Agreed. We will make success reporting sparse: only regions with deleted files/indexes or `need_retry` will emit a success row. No-op and route-skipped regions will not produce a success event row.

src/meta-srv/src/event/gc.rs
WenyXu: Agreed. With sparse success rows, `processed` does not add independent information, so we will remove it from the report payload.

src/meta-srv/src/gc/procedure.rs
WenyXu: This is intentional: we do not want a normal-path Submitted row or to duplicate the configuration across per-region success rows. The request configuration remains diagnostic context for abnormal lifecycle events. We will update the PR description accordingly.

src/meta-srv/src/event/gc.rs
WenyXu: Agreed. We will rename this constructor to better reflect that it carries procedure configuration rather than an empty lifecycle event.

src/meta-srv/src/event/gc.rs
WenyXu: Agreed on removing the duplicated schema and hardcoded column count. We will share a local GC schema helper between production code and tests. We will keep `null_row` local for now, since there is not yet a production-level shared use case.

tests-integration/tests/event_recorder_test_util.rs
WenyXu: Agreed. We will fail immediately with the actual Arrow data type instead of polling until timeout.

tests-integration/tests/gc_event.rs
WenyXu: Thanks. This is a valid test-stability concern. We did not find an existing fixture-level switch for it, so we will keep this scoped out of the current event change rather than introduce new compaction-test infrastructure.

src/meta-srv/src/gc/procedure.rs
discord9, line 1077: **P2: Emit the initial lifecycle row**

`LocalManager::submit_root` records `Submitted` before execution, but this drops it; `Recovered` is dropped too. A normal first-attempt success therefore emits only `Succeeded`, whose payload is null, so `regions`, `full_file_listing`, and `timeout` are never recorded and the advertised lifecycle is incomplete. Please return `BatchGcEvent::lifecycle(...)` for `Submitted` (and preferably `Recovered`); it already emits one procedure-level row, so this does not restore per-region fanout.

src/meta-srv/src/gc/procedure.rs
```diff
@@ -1024,4 +1025,35 @@ impl Procedure for BatchGcProcedure {
...
+                } = ctx.lifecycle_state
+                else {
+                    return None;
+                };
+                let report = output.downcast_ref::<GcReport>()?;
+                BatchGcEvent::succeeded(&self.data.regions, report)
+            }
+            EventTrigger::Retrying { .. }
+            | EventTrigger::RollingBack
+            | EventTrigger::Failed
```
discord9: **P2: Preserve GC outcomes when post-GC cleanup fails**

The `Gcing` branch stores `self.data.gc_report = Some(report)` before entering `UpdateRepartition`. If `cleanup_region_repartition()` then fails and the runner eventually reaches `Failed`, this arm emits only the configuration payload and no `Succeeded` event will ever carry the report. Files may already have been irreversibly deleted, yet `deleted_files`, `deleted_indexes`, and `need_retry_regions` disappear from the event history. Please emit report-backed per-region rows for terminal failure when `gc_report` is present (while keeping retry events procedure-level if fanout is a concern), and cover a cleanup failure after a report has been produced.

WenyXu:
@codex review

src/meta-srv/src/gc/procedure.rs
WenyXu: This is intentional: we do not want a normal-path Submitted row or to duplicate the configuration across per-region success rows. The request configuration remains diagnostic context for abnormal lifecycle events. We will update the PR description accordingly.

src/meta-srv/src/gc/procedure.rs
WenyXu: Fixed in [a8dfa96](https://github.com/GreptimeTeam/greptimedb/pull/8673/commits/a8dfa96ebd86fd37362221ced00569f14f5b0c0b). Failed and Poisoned now emit the stored per-region GC report, with config-only fallback when the report is empty.

WenyXu:
@codex review

src/meta-srv/src/event/gc.rs
fengjiachun: Correction to my original comment: **case 2 was wrong.** A region whose table route can't be resolved does *not* lead to a successful procedure with a NULL row — it fails the procedure outright.

`discover_route_for_regions` does `continue` on a `get_table_route` error, but the regions it drops are then caught downstream: `get_file_references` (`src/meta-srv/src/gc/procedure.rs:587`) and `send_gc_instructions` (`:765`) both iterate `self.data.regions` and return `UnexpectedSnafu("region_routes: ... does not contain region_id: ...")` for any region missing from `region_routes`. So the Acquiring stage fails first, and `Succeeded` is never reached. Sorry for the noise.

That leaves only case 1 (processed, nothing to delete), which means NULL was unambiguous rather than overloaded — my framing was off. The change you made is still a clear improvement though, just for the volume reason rather than the correctness one: those no-op rows are pure noise, and dropping them plus returning `None` from `with_report` for an empty report is exactly the right call. `gc_report` is now non-NULL on every row that carries a `region_id`, which is a much cleaner contract.

src/meta-srv/src/gc/procedure.rs
```diff
@@ -1024,4 +1030,45 @@ impl Procedure for BatchGcProcedure {
...
+                };
+                let report = output.downcast_ref::<GcReport>()?;
+                BatchGcEvent::with_report(report)?
+            }
+            EventTrigger::Retrying { .. } | EventTrigger::RollingBack => BatchGcEvent::with_config(
+                &self.data.regions,
+                self.data.full_file_listing,
+                self.data.timeout,
+            ),
+            EventTrigger::Failed | EventTrigger::Poisoned => self
```
fengjiachun, line 1093: On a terminal failure this picks report *or* config, never both — and failure is exactly when you'd want both. When `gc_report` is present the emitted event has `payload == Null`, so `full_file_listing`, `timeout`, and the full requested region list are all lost; you only see the regions that happened to have deletions or retries.

The partial mitigation is that a preceding `Retrying` event carries the config under the same `procedure_id`, but a procedure that fails on a non-retryable error the first time through never emits one, and then the config is gone entirely.

`BatchGcEvent`'s `payload` and `regions` are independent fields, so the struct already supports carrying both — it's just that no constructor builds that combination. Something like a `with_report_and_config(...)` for the terminal arms would close this at basically no cost, and would also make the `Failed` event shape consistent with `Retrying` / `RollingBack`.

Minor, while you're here: `null_row()` now calls `schema()` just to take `.len()`, which allocates four `ColumnSchema` values per call. A `const` for the column count, or `schema().len()` hoisted, would avoid that — though it's off any hot path, so purely cosmetic.

fengjiachun (commented):
Re-reviewed at f809a3136a. All the previous points are addressed, and having `with_report` return `Option` so an empty report emits nothing is a better answer to the volume concern than what I suggested. All 6 `event::gc::tests` pass locally.

Two corrections on my side.

First, on the `region_report` thread: my "case 2" was wrong. A region with an unresolvable table route fails the procedure in `get_file_references` long before `Succeeded`, so NULL was never actually ambiguous. The change still holds, just for noise reduction rather than correctness.

Second, **withdrawing the unbounded `deleted_files` concern** from the last round — it doesn't hold. `RemovedFilesRecord::clear_deleted_files` drops files from the manifest's `removed_files` once GC actually deletes them, and the non-full-listing path only ever considers `recently_removed_files`, so each round handles just the delta since the previous GC. With a 5-minute ticker and a 5-minute cooldown that's a handful of files per region. The only larger case is a dropped region under full listing, and even there it's bounded by the region's actual file count. No truncation needed; sorry for the detour.

That leaves one new point, inline on the `Failed` / `Poisoned` arm.

WenyXu:
Thanks for calling this out. We considered the size trade-off and will keep the complete list for this initial audit event. In the normal fast path, GC removes manifest-tracked files, and the TWCS picker limits a single compaction to 32 input files. This is not a strict upper bound, though: removed-file backlogs can accumulate, and full-file-listing GC can reclaim orphan files independently of compaction. We should revisit a cap plus count based on operational evidence if this becomes a concern; for now, keeping the full list preserves auditability.

src/meta-srv/src/gc/procedure.rs
```diff
@@ -1024,4 +1030,45 @@ impl Procedure for BatchGcProcedure {
...
+                .data
```
discord9, line 1094: **P1: Keep the report until the framework commit succeeds**

The `UpdateRepartition` success path calls `self.data.gc_report.take()` at line 999 before returning `Status::Done`, but `Runner` only calls `commit_procedure()` afterward. If that commit fails, the runner marks this procedure `Retrying` and executes `UpdateRepartition` again; the second pass then fails deterministically with `GC report should be present`, and this terminal arm falls back to config because the report was already consumed. Thus a transient procedure-store failure turns completed GC into a failed procedure and loses its deletion report. Please clone/retain the report when building `done_with_output` and only let the procedure object drop it after a successful framework commit; a regression test with the first commit failing would cover the lifecycle.

src/meta-srv/src/gc/procedure.rs
```diff
@@ -1024,4 +1030,45 @@ impl Procedure for BatchGcProcedure {
...
+                .gc_report
+                .as_ref()
+                .and_then(BatchGcEvent::with_report)
```
discord9, line 1097: **P2: Preserve reports from peers that succeeded before a batch error**

`send_gc_instructions()` merges each successful peer reply into `all_report`, but if any other peer errors, lines 868-869 return only `first_error` and discard that accumulated report. The `Gcing` error arm therefore never stores it, so this terminal event takes the config-only fallback even though another datanode may already have irreversibly deleted files. This is reachable for a manual region list spanning multiple leaders. Please propagate/store the partial `GcReport` alongside the error so `Failed` can emit those completed outcomes, and test a two-peer batch with one successful report and one error.

discord9 (commented):
Re-reviewed the current head after the report-preservation update. Two remaining paths can still lose completed GC outcomes; details inline.

src/meta-srv/src/gc/procedure.rs
WenyXu: Fixed in 513109269c. `done_with_gc_report` now clones the retained report instead of consuming it, so an `UpdateRepartition` retry can build the same Done output again. Added a regression test that builds the Done output twice and verifies the stored report remains available.

src/meta-srv/src/gc/procedure.rs
WenyXu: Fixed in 513109269c. `send_gc_instructions` now merges all successful peer reports into procedure state before propagating the first error. Added a two-peer test with one successful reply and one failed peer, asserting that the successful peer report is retained.

WenyXu:
@codex review

src/meta-srv/src/gc/procedure.rs
```diff
@@ -284,6 +285,26 @@ impl BatchGcProcedure {
         })
     }
 
+    fn merge_gc_report(&mut self, report: GcReport) {
+        self.data.gc_report.get_or_insert_default().merge(report);
```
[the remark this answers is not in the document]
WenyXu: I don't think this path is reachable today. `Gcing` errors are returned as `ProcedureError::external`, while the runner retries only `Error::RetryLater`, so the procedure goes directly to `Failed`. `need_retry_regions` is for a later GC round, not a retry of the current `Gcing` state.

src/meta-srv/src/gc/procedure.rs
```diff
@@ -859,6 +879,8 @@ impl BatchGcProcedure {
             all_report.merge(report);
         }
 
+        self.merge_gc_report(all_report);
```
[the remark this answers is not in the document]
WenyXu: The runner does not retry this error. Partial reports are merged into the live procedure before it enters `Failed`, and the `Failed` event reads that report. A crash before event recording could still lose it, but that is a broader event-durability issue rather than this retry path.

src/meta-srv/src/gc/procedure.rs
fengjiachun: `GcReport::merge` ends with

```rust
self.need_retry_regions.retain(|region| !self.deleted_files.contains_key(region));
```

and `contains_key` is true even for an **empty** vec. `LocalGcWorker::run` does `deleted_files.insert(*region_id, data_files)` for every successfully processed region, including ones with nothing to delete — so a no-op region leaves a permanent empty key in the accumulator.

Once that key is there, every later merge silently strips that region from `need_retry_regions`. Repro:

```rust
let a = RegionId::new(1024, 1);
let mut acc = GcReport::default();

// Round 1: region A processed fine, zero files deleted.
acc.merge(GcReport {
    deleted_files: HashMap::from([(a, vec![])]),
    processed_regions: HashSet::from([a]),
    ..Default::default()
});

// Round 2 (procedure retry): region A now needs a retry.
acc.merge(GcReport {
    need_retry_regions: HashSet::from([a]),
    ..Default::default()
});

assert!(acc.need_retry_regions.contains(&a)); // fails — the set is empty
```

This wasn't reachable before this commit: the Gcing arm did `self.data.gc_report = Some(report)`, i.e. last-write-wins, so a given region never went through `merge` twice. Accumulating into `data.gc_report` is what exposes it, since every retry re-sends the full region list and merges the whole batch again.

The impact isn't limited to the `need_retry` field in the event. `cleanup_region_repartition` reads `gc_report.need_retry_regions` to decide whether to keep a region's tombstone; when the flag is lost it takes the `else` branch and does `src_to_dst.remove(&src_region)`, so the pending extension cleanup is never replayed on the next GC cycle.

Worth noting the other `merge` call sites (across peers in `send_gc_instructions`, across tables/datanodes in `handler.rs` / `scheduler.rs`) all combine **disjoint** region sets, so a region only ever appears in one report there. This new call site is the only one that merges the same regions repeatedly — which argues for fixing it here rather than changing the shared `GcReport::merge`.

Since each round re-runs the full batch, the round's own result is authoritative for the regions it covers, so letting the incoming `need_retry_regions` / `processed_regions` overwrite the accumulated ones (rather than union-ing them) would match the intent. The three new tests only exercise `deleted_files` accumulation, so this case slips through — worth a test with a no-op round followed by a retry round.

WenyXu: Thanks for the detailed repro. I confirmed that the current control flow does not re-enter `Gcing` on the same procedure, so this path is not reachable today. Still, I hardened the local accumulator defensively in `7ca7fd0498`: deleted objects remain cumulative, while `processed` and `need_retry` use the latest outcome for affected regions. I also added the no-op-success-to-retry regression test.

WenyXu:
@codex review

fengjiachun (approved):
LGTM

PR #8676, opened 2026-07-29
fix(mito2): make async index publication conditional

killme2008:

N/A

Async index publication had two races with compaction:

- Compaction could replace the source SST between the final source check and the manifest update.
- A delayed worker notification could apply metadata that a later manifest edit had already removed.

This PR fixes both publication stages:

- Commit index metadata only when the manifest still contains the full source `FileMeta` generation. The check and commit use the same manifest write lock.
- Before applying the result in memory, revalidate the committed `FileMeta` under the manifest read lock and keep that lock through `apply_edit`.
- Keep the logical region ID separate from `FileMeta.region_id`. The logical ID is used for scheduler state, worker routing, and manifest/version updates; the physical ID is used for SST/index paths.
- Choose the next artifact version from published index metadata, not the optional file-size hint. This prevents legacy manifests with a zero size from overwriting an already published version.
- Evict local caches and staging files for stale publications. Remote artifacts are left to reference-aware/full-listing GC.
- When a file is removed, record the index version currently stored in the manifest rather than the possibly stale version captured by compaction.

Close/reopen can otherwise schedule the same SST while a job from the previous region incarnation is still running. The scheduler keeps the active build lease after close, drop, or truncate. Reopened builds wait until the old active batch sends `IndexBuildStopped`; close does not cancel or wait for index work. Pending tasks carry their own `VersionControlRef`.

Deterministic tests cover both compaction orderings, cross-region files where the logical and physical region IDs differ, close/reopen overlap, repeated-row checks across reopen, legacy index metadata without a file-size hint, local cache cleanup, remote artifact retention for `ObjectStoreFilePurger`, and both purger implementations.

This does not change persisted formats, public APIs, or schemas.

## PR Checklist

### DIFF
--- a/src/mito2/src/manifest/action.rs
+++ b/src/mito2/src/manifest/action.rs
@@ -263,11 +263,14 @@ impl RegionManifestBuilder {
                 removed_files.push(RemovedFile::Index(old_file.file_id, old_index));
             }
         }
-        removed_files.extend(
-            edit.files_to_remove
-                .iter()
-                .map(|f| RemovedFile::File(f.file_id, f.index_version())),
-        );
+        removed_files.extend(edit.files_to_remove.iter().map(|file| {
+            let index_version = self
+                .files
+                .get(&file.file_id)
+                .and_then(FileMeta::index_version)
+                .or_else(|| file.index_version());
+            RemovedFile::File(file.file_id, index_version)
+        }));
         let at = edit
             .timestamp_ms
             .unwrap_or_else(|| Utc::now().timestamp_millis());
@@ -914,6 +917,54 @@ mod tests {
         let new: RegionManifest = serde_json::from_str(&json).unwrap();
 
         assert_eq!(manifest, new);
+    }
+
+    #[test]
+    fn test_remove_tracks_current_manifest_index_version() {
+        let file_id = FileId::random();
+        let current = FileMeta {
+            region_id: RegionId::new(1, 1),
+            file_id,
+            available_indexes: smallvec::smallvec![crate::sst::file::IndexType::InvertedIndex],
+            index_file_size: 1024,
+            index_version: 2,
+            ..Default::default()
+        };
+        let mut stale = current.clone();
+        stale.available_indexes.clear();
+        stale.index_file_size = 0;
+        stale.index_version = 0;
+
+        let mut builder = RegionManifestBuilder::default();
+        builder.apply_edit(
+            1,
+            RegionEdit {
+                files_to_add: vec![current],
+                files_to_remove: Vec::new(),
+                timestamp_ms: None,
+                compaction_time_window: None,
+                flushed_entry_id: None,
+                flushed_sequence: None,
+                committed_sequence: None,
+            },
+        );
+        builder.apply_edit(
+            2,
+            RegionEdit {
+                files_to_add: Vec::new(),
+                files_to_remove: vec![stale],
+                timestamp_ms: Some(42),
+                compaction_time_window: None,
+                flushed_entry_id: None,
+                flushed_sequence: None,
+                committed_sequence: None,
+            },
+        );
+
+        assert_eq!(
+            builder.removed_files.removed_files[0].files,
+            HashSet::from([RemovedFile::File(file_id, Some(2))])
+        );
     }
 
     /// Test if old version can still be deserialized then serialized to the new version.
--- a/src/mito2/src/region.rs
+++ b/src/mito2/src/region.rs
@@ -1229,7 +1229,7 @@ impl ManifestContext {
         source: &FileMeta,
         updated: FileMeta,
     ) -> Result<IndexPublication> {
-        let mut manager = self.manifest_manager.write().await;
+        let manager = self.manifest_manager.write().await;
         let manifest = manager.manifest();
         let region_id = manifest.metadata.region_id;
         let current_state = self.state.load();
@@ -1266,17 +1266,7 @@ impl ManifestContext {
             compaction_time_window: None,
         };
         let action_list = RegionMetaActionList::with_action(RegionMetaAction::Edit(edit));
-        let pending = self.update_locked(&mut manager, action_list, false).await?;
-        let manifest_version = pending.version();
-        drop(manager);
-
-        if self.state.load() == RegionRoleState::Follower {
-            warn!(
-                "Region {} becomes follower while publishing index metadata, manifest version: {}",
-                region_id, manifest_version
-            );
-        }
-        pending.fire().await;
+        let manifest_version = self.update_and_fire(manager, action_list, false).await?;
 
         Ok(IndexPublication::Committed {
             manifest_version,
@@ -1317,6 +1307,38 @@ impl ManifestContext {
         ))
     }
 
+    /// Updates the manifest using a caller-held write lock, releases the lock,
+    /// and then fires the manifest hook.
+    ///
+    /// Callers must perform state and applicability checks before calling this
+    /// method so the checks and update remain in the same lock critical section.
+    async fn update_and_fire(
+        &self,
+        mut manager: RwLockWriteGuard<'_, RegionManifestManager>,
+        action_list: RegionMetaActionList,
+        is_staging: bool,
+    ) -> Result<ManifestVersion> {
+        let region_id = manager.manifest().metadata.region_id;
+        let pending = self
+            .update_locked(&mut manager, action_list, is_staging)
+            .await?;
+        let version = pending.version();
+
+        // Hook implementations may read the manifest or send region requests
+        // that acquire this lock.
+        drop(manager);
+
+        if self.state.load() == RegionRoleState::Follower {
+            warn!(
+                "Region {} becomes follower while updating manifest which may cause inconsistency, manifest version: {version}",
+                region_id
+            );
+        }
+
+        pending.fire().await;
+        Ok(version)
+    }
+
     async fn update_manifest_with_state_check(
         &self,
         action_list: RegionMetaActionList,
@@ -1324,7 +1346,7 @@ impl ManifestContext {
         check_state: impl FnOnce(RegionRoleState, RegionId) -> Result<()>,
     ) -> Result<ManifestVersion> {
         // Acquires the write lock of the manifest manager.
-        let mut manager = self.manifest_manager.write().await;
+        let manager = self.manifest_manager.write().await;
         // Gets current manifest.
         let manifest = manager.manifest();
         // Checks state inside the lock. This is to ensure that we won't update the manifest
@@ -1382,28 +1404,7 @@ impl ManifestContext {
             }
         }
 
-        // `update_locked` returns a `PendingManifestHook` we fire after releasing the lock.
-        let region_id = manifest.metadata.region_id;
-        let pending = self
-            .update_locked(&mut manager, action_list, is_staging)
-            .await?;
-        let version = pending.version();
-
-        // Drop the write lock before invoking the hook. Hook implementations may
-        // read the manifest or send region requests that acquire this lock;
-        // holding it would deadlock. Slow hooks also must not block manifest updates.
-        drop(manager);
-
-        if self.state.load() == RegionRoleState::Follower {
-            warn!(
-                "Region {} becomes follower while updating manifest which may cause inconsistency, manifest version: {version}",
-                region_id
-            );
-        }
-
-        pending.fire().await;
-
-        Ok(version)
+        self.update_and_fire(manager, action_list, is_staging).await
     }
 
     /// Sets the [`RegionRole`].
--- a/src/mito2/src/region.rs
+++ b/src/mito2/src/region.rs
@@ -1019,6 +1019,18 @@ impl Drop for MitoRegion {
     }
 }
 
+/// Result of publishing rebuilt index metadata to the manifest.
+#[derive(Debug)]
+pub(crate) enum IndexPublication {
+    /// The index metadata was committed.
+    Committed {
+        manifest_version: ManifestVersion,
+        file_meta: FileMeta,
+    },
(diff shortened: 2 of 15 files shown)


src/mito2/src/sst/index.rs
```diff
@@ -851,18 +902,56 @@ impl IndexBuildTask {
...
                         region_id: self.file_meta.region_id,
                         notify: BackgroundNotify::IndexBuildFinished(index_build_finished),
                     }
                 }
+                Ok(IndexPublication::Stale) => {
+                    INDEX_PUBLICATION_STALE_TOTAL
+                        .with_label_values(&["manifest_commit"])
+                        .inc();
+                    let index_id = RegionIndexId::new(region_file_id, new_index_version);
+                    cleanup_stale_index_artifact(
```
v0y4g3r: The stale cleanup here can delete an index artifact that is still referenced by the manifest when two builds of the same file overlap.

`new_index_version` is derived solely from the snapshotted source `file_meta` (`index_version + 1`, or 0 for a first build). Two concurrent `IndexBuildTask`s created for the same file — e.g. an automatic flush-triggered rebuild overlapping with a manual `BuildIndex` request (whose empty `file_metas` scan keeps re-selecting the file until the worker applies the first build's edit) — snapshot the same source meta, compute the *same* `new_index_version`, and upload to the *same* artifact path. If build A commits first, build B's in-lock check fails (`Stale`) and this cleanup deletes `(file_id, new_index_version)` — which is exactly the artifact A just committed and the manifest now references.

Consequences: index reads on that file hit a missing puffin object, `is_index_consistent_with_region` still reports the index as built so no rebuild is triggered, and the dangling reference only heals when compaction replaces the file. Before this PR the loser would simply re-commit equivalent metadata, so this is a regression introduced by the cleanup.

The error-path cleanup below (after `update_manifest` fails) has the same shape.

Suggestion: before deleting the remote artifact, re-check under the manifest read lock that no current `manifest.files[file_id]` references this exact `index_version` (cache evictions are always safe; only the object-store delete is destructive). Alternatively/additionally, dedup in-flight index builds per file at the region level. A deterministic test with two same-version overlapping builds would cover this — the existing overlapping-publication unit test uses distinct index versions (1 vs 3), which sidesteps the shared artifact path.

v0y4g3r (commented):
There still can be race condition when concurrent index building happens:

src/mito2/src/sst/index.rs
fengjiachun: The per-worker `building_files` check does deduplicate duplicate requests while a region remains open, but close/reopen gives a concrete path around it. `remove_region()` stops the old manifest manager and drops the scheduler status without cancelling the running build. After reopen, a new build can commit the same `(file_id, index_version)`. When the old build resumes, `update_manifest_for_index()` returns `RegionStopped`, and the unconditional error-path cleanup below deletes the artifact now referenced by the reopened region manifest. Guarding the object-store delete against the current live manifest, plus a close/reopen same-version test, is therefore still necessary.

killme2008: Confirmed. The fix now keeps the scheduler lease for the active region batch across close/reopen instead of cancelling or waiting for it. Builds from the reopened incarnation stay pending until the old batch sends `IndexBuildStopped`, so two builds cannot write the same `(file_id, index_version)` artifact. Pending entries carry their own `VersionControlRef`.

Cleanup runs only for a definite stale publication. Generic manifest I/O errors no longer delete an artifact because the commit result may be indeterminate.

The deterministic close/reopen test blocks the old build, verifies close returns and the reopened build remains queued, then checks the final artifact and rows across another reopen.

src/mito2/src/request.rs
```diff
@@ -978,7 +978,8 @@ pub(crate) struct FlushFailed {
 pub(crate) struct IndexBuildFinished {
     #[allow(dead_code)]
     pub(crate) region_id: RegionId,
```
v0y4g3r: Can we remove this field?

killme2008: Yes. Removed it; the outer worker notification already carries `region_id`.

killme2008:
Addressed the cross-region cases in `bc0a630ec8`.

- `IndexBuildTask` now carries the logical region separately from the physical owner in `FileMeta.region_id`. Scheduler state, worker routing, and manifest/version updates use the logical region; index paths keep using the physical region.
- Stale publication no longer deletes the remote index object. It only evicts local cache/staging state. The object may already be referenced by another repartitioned manifest, so FilePurger/GC owns remote deletion.
- File removal bookkeeping now records the index version currently in the manifest, not the stale version captured by compaction.

Added a deterministic cross-region test with different logical and physical region IDs. The existing two compaction race tests now also verify that `ObjectStoreFilePurger` does not directly delete stale remote artifacts.

v0y4g3r:
Following up on https://github.com/GreptimeTeam/greptimedb/pull/8676#issuecomment-5117544319: the logical/physical region ID separation does not appear to reach the index writer yet.

`new_index_build_task()` constructs `IndexerBuilderImpl` with `version.metadata`, which belongs to the logical region. `IndexerBuilderImpl::build()` then sets `Indexer.region_id` from `self.metadata.region_id`, and the Puffin writer uses that ID to construct its `RegionIndexId`. However, the committed `FileMeta` retains the physical owner in `file_meta.region_id`.

As a result, without write cache the Puffin object is written under the logical region path while readers use the physical region path from `FileMeta`. With write cache enabled, the writer creates the logical-region cache key, but `maybe_upload_index_file()` looks up the physical-region cache key, so the upload will normally fail (or could pick up an unrelated stale entry).

Could we pass the physical `file_meta.region_id` explicitly into the indexer for Puffin/cache paths, while continuing to use the logical region ID only for scheduler state, worker routing, and manifest/version updates? The cross-region test should also verify the physical Puffin object or perform an actual indexed read; checking only `FileMeta::exists_index()` does not catch this mismatch.

killme2008:
Addressed in `d77d0fd329`.

`IndexerBuilder::build` now takes the physical `RegionFileId` explicitly. `Indexer` keeps the logical region for metadata and intermediate workspace, and uses the physical region only for Puffin and write-cache staging paths.

The cross-region test now enables write cache and verifies that the published Puffin object is readable from the physical path through the scanner.

Verified with `cargo check -p mito2 --all-targets --all-features`, `cargo nextest run -p mito2 index_build`, `cargo nextest run -p mito2 sst::index`, clippy, and fmt.

src/mito2/src/sst/index.rs
```diff
@@ -1031,10 +1136,18 @@ impl IndexBuildScheduler {
     ) -> Result<()> {
         let status = self
             .region_status
-            .entry(task.file_meta.region_id)
-            .or_insert_with(|| IndexBuildStatus::new(task.file_meta.region_id));
-
-        if status.building_files.contains(&task.file_meta.file_id) {
+            .entry(task.region_id)
```
v0y4g3r, line 1139: If an SST file is shared between two regions because of repartition, this `region_status` here can not properly deduplicate index building tasks, thus there can still be concurrent building tasks over the same SST file.

v0y4g3r (approved):
LGTM

src/mito2/src/sst/index.rs
killme2008: Simultaneous writes to the same artifact across regions

In the currently supported repartitioning flow:

* Two logical regions can reference the same physical SST.
* They belong to the same table, so their schemas and index definitions should be identical.
* Even if two tasks write the same `(physical_region_id, file_id, index_version)`, the artifacts have the same semantics; retaining either complete object is acceptable.
* If an index is corrupted or fails to load, the production query path is fail-open: it skips the index and reads the SST instead.

Therefore, this currently looks more like an architectural invariant that should be made explicit, rather than a confirmed bug.

It would become a real problem only if we later allow logical regions sharing the same physical SST to use different index definitions. In that case, two valid but different artifacts could overwrite each other. We would then need to incorporate the build generation into the artifact identity, or add coordination across regions/nodes. Introducing a global lease now for that hypothetical case would be overengineering.

src/mito2/src/region.rs
```diff
@@ -1207,6 +1219,57 @@ impl ManifestContext {
...
+        if !matches!(
+            current_state,
+            RegionRoleState::Leader(RegionLeaderState::Writable)
+                | RegionRoleState::Leader(RegionLeaderState::Downgrading)
+        ) || manager.is_stopped()
+        {
+            return Ok(IndexPublication::Stale);
+        }
+
+        if manifest.files.get(&source.file_id) != Some(source) {
```
discord9, line 1244: **[P1] Fence publication on the captured schema/index generation as well as `FileMeta`.** A build created under schema A can reach this check after `ALTER ... SET INDEX` has installed schema B, because `RegionChange` updates manifest metadata without changing `manifest.files`. The schema-B rebuild is then discarded as a same-file duplicate by `IndexBuildScheduler::schedule_build`, while the schema-A task still passes this equality check and publishes an index that is inconsistent with B. An idle SST can remain under-indexed until a manual rebuild or compaction. Please carry a schema version/index-definition fingerprint in the task, validate it under this manifest lock, and retain one latest-generation successor instead of aborting it as a duplicate. A gated ALTER-vs-in-flight-build test should verify `is_index_consistent_with_region()` before and after reopen.

src/mito2/src/manifest/action.rs
```diff
@@ -263,11 +263,14 @@ impl RegionManifestBuilder {
                 removed_files.push(RemovedFile::Index(old_file.file_id, old_index));
             }
         }
-        removed_files.extend(
-            edit.files_to_remove
-                .iter()
-                .map(|f| RemovedFile::File(f.file_id, f.index_version())),
-        );
+        removed_files.extend(edit.files_to_remove.iter().map(|file| {
```
discord9, line 266: **[P1] Apply this current-manifest index-version lookup to partial truncate too.** `TruncateKind::Partial` still records `f.index_version()` from the request's `FileMeta`. There is a deterministic window after an index build has committed N+1 to the manifest but before `IndexBuildFinished` applies it to `VersionControl`; partial truncate snapshots N, removes the manifest entry containing N+1, and records only N. The delayed worker notification is rejected because the file is gone, so N+1 is left untracked and LocalFilePurger/fast GC only delete through N. Please resolve the partial-truncate removal version from `self.files` as done here, ensure the committed generation is deleted when GC is disabled, and add an after-manifest-commit partial-truncate race test starting from an SST already at version N.

src/mito2/src/sst/index.rs
```diff
@@ -85,6 +87,48 @@ pub(crate) const TYPE_BLOOM_FILTER_INDEX: &str = "bloom_filter_index";
 #[cfg(feature = "vector_index")]
 pub(crate) const TYPE_VECTOR_INDEX: &str = "vector_index";
 
+/// Evicts local state for a stale index artifact.
+///
+/// The remote artifact may already be referenced by another region manifest
+/// because repartitioned regions share physical SST and index paths. Remote
+/// deletion therefore stays in the normal FilePurger/GC lifecycle; object-store
```
discord9, line 94: **[P2] This leaves stale N+1 objects orphaned while global GC is disabled.** If the source SST already has index version N, a rejected rebuild has uploaded N+1, but neither a manifest removal record nor a `FileHandle` references that artifact. `LocalFilePurger` later deletes only versions `0..=N`, and with `gc.enable = false` no full-listing GC discovers N+1, so it persists indefinitely for the live region. Please explicitly handle the GC-disabled case (rather than inferring it solely from the purger type) by safely deleting/registering the exact unpublished artifact. The regression test should start from an already-indexed SST; a first build at version 0 is already covered by the source-file purge range and does not expose this leak.

discord9 (commented):
I found three follow-up correctness issues in the async index publication paths. The PR is already merged, so these can be addressed in a follow-up.

src/mito2/src/region.rs
killme2008: Addressed in the follow-up. `IndexBuildTask` now carries an `IndexBuildSource` containing the authoritative manifest `FileMeta` and the builder schema version. Publication checks both under the same manifest write lock. The scheduler still uses a `BinaryHeap`, wrapped by a small per-SST coalescing abstraction: it keeps the latest schema generation, completes superseded result senders, and schedules the successor only after the active build stops. Deterministic tests cover consecutive `SetIndexes`, stale-generation rejection, manifest/VersionControl lag, and close/reopen retiring leases.

src/mito2/src/manifest/action.rs
killme2008: Addressed the metadata-accounting issue in the follow-up. `RegionEdit` and partial truncate now share one lookup that resolves the removal index version from the current manifest `FileMeta`, falling back to the request metadata only when the file is already absent. The regression test starts with manifest index version 2 and stale request version 0 and covers both edit and partial truncate. I intentionally did not add a GC-disabled exact-delete branch or a large worker-level race path: the after-manifest-commit/before-VersionControl window is extremely narrow and storage-only. Our object-store deployment policy enables GC, which collects the generation recorded here; we accept the possible orphan in local GC-disabled mode and leave a TODO documenting that tradeoff.

src/mito2/src/sst/index.rs
killme2008: After rechecking the lifecycle constraints, I am not adding remote deletion in this follow-up. #8678 makes GC a hard requirement for repartition: GC-disabled clusters reject new repartition, startup reconciles legacy mappings/unfinished procedures and rejects disabling GC, and metasrv/datanode GC mismatches fail startup. Therefore the shared-physical-path + GC-disabled combination cannot occur. The remaining non-repartition leak requires GC disabled, a successful N+1 upload, stale publication, and no successor publishing the same N+1; it is storage-only. Our object-store deployment policy enables GC, so we accept this edge in local GC-disabled mode and leave a focused TODO rather than adding a new purger/GC state path.

PR #8677, opened 2026-07-29
feat(metasrv): record WAL prune procedure events

WenyXu:

- Follows the procedure event conventions established in #8648, #8623, #8632, and #8627.

Add a sparse procedure event for remote WAL prune operations.

`WalPruneProcedure` records `Retrying`, `Failed`, and `Poisoned` transitions to retain failure diagnostics. It records `Succeeded` only when `WalPruneOutcome` confirms an actual physical or logical prune; `Submitted`, `Recovered`, and successful no-op runs emit no event. `RollingBack` and `ChildSubmitted` do not apply because WAL pruning neither supports rollback nor submits child procedures.

The event schema contains `topic_name`, `prunable_entry_id`, and nullable `latest_offset`. `prunable_entry_id` is the requested prune boundary; on non-`Succeeded` rows it is only an attempted boundary, so consumers must filter `procedure_trigger = 'Succeeded'` to identify persisted prune boundaries. `latest_offset` is the Kafka log-end offset, an exclusive upper bound. The versioned payload contains `logical_delete`; `WalPruneData` serialization remains unchanged.

Procedure event delivery is best effort rather than a durable audit log. A physical prune can complete before a metasrv restart and recover as a no-op without a `Succeeded` event.

The procedure honors the event-type filter. `wal_prune` is listed in metasrv configuration only: standalone can use Kafka WAL but does not start the WAL prune ticker or manager. The generated configuration reference is updated accordingly.

## Validation

- `make config-docs`
- `cargo nextest run -p common-event-recorder wal_prune_schema_preserves_names_types_semantics_and_order`
- `cargo nextest run -p meta-srv test_wal_prune_event`
- `cargo check -p tests-integration --test main`
- `git diff --check`

Kafka-backed procedure and standalone integration tests were not executed locally and require CI validation.

## PR Checklist


WenyXu:
@codex review

src/meta-srv/src/procedure/wal_prune.rs
```diff
@@ -158,7 +168,7 @@ impl WalPruneProcedure {
             "Successfully pruned WAL for topic: {}, prunable entry id: {}, latest offset: {}",
             self.data.topic, self.data.prunable_entry_id, latest_offset
         );
-        Ok(Status::done())
+        Ok(Status::done_with_output(WalPruneOutcome))
```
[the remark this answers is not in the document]
WenyXu: The scenario is valid, but it only affects observability—the prune and its metadata boundary have already been committed successfully. Procedure event delivery is explicitly best effort rather than a durable audit log, so persisting additional completion state solely to guarantee this Succeeded event would add recovery-state complexity without providing an end-to-end delivery guarantee. The failed framework commit is still visible through the Retrying event, including its error and prune context, and subsequent prune cycles are unaffected. I’d prefer to keep the no-op path silent as designed.

fengjiachun (commented):
The implementation follows the conventions from #8627 / #8665 and reads cleanly. `cargo check -p meta-srv --lib --tests` and `cargo nextest run -p common-event-recorder wal_prune` both pass locally; I couldn't run the Kafka-gated tests.

My main concern isn't the code — it's that the PR description documents the design of the first commit, not what's on the branch now. `490dfe43` ("expand WAL prune procedure events") changed the design and the description was never updated. Three concrete contradictions:

- It says "Submitted, recovered, retrying, failed, and poisoned lifecycle transitions are also omitted", but `event()` emits for `Retrying`, `Failed` and `Poisoned`.
- It says an event is emitted "only after an actual prune successfully advances the persisted boundary" — failed and retrying attempts emit too.
- It describes the columns as the topic, "the nullable previous prune boundary", and the completed boundary. The actual columns are `topic_name`, `prunable_entry_id` and `latest_offset`; there is no previous-boundary column, `previous_pruned_entry_id` was dropped in `490dfe43`.

`git show 3bbcf5de6e` matches the description exactly, which confirms it's just stale. Since the description is both the change record and the de-facto spec for event consumers, could you rewrite it against the current design? Rest of the comments are inline.

src/meta-srv/src/procedure/wal_prune.rs
```diff
@@ -199,12 +209,43 @@ impl Procedure for WalPruneProcedure {
...
+
+        if matches!(&ctx.trigger, EventTrigger::Succeeded) {
+            let ProcedureState::Done {
+                output: Some(output),
+            } = ctx.lifecycle_state
+            else {
+                return None;
+            };
+            output.downcast_ref::<WalPruneOutcome>()?;
+        } else if !matches!(
```
fengjiachun, line 232: This branch emits events for `Retrying`, `Failed` and `Poisoned`, which is the opposite of what the PR description says. Assuming the code is the intended design, two things worth confirming.

With `max_retry_times` defaulting to 3 and `num_topics` to 64 on a 30-minute prune interval, a sustained Kafka outage produces roughly 4 events x 64 topics per cycle — ~12k rows/day of pure failure noise. Survivable, but it's exactly the volume concern the description cites as the reason for *not* emitting these, so it'd help to state why the trade-off flipped.

Separately, `RollingBack` and `ChildSubmitted` fall through to `None` here. That's correct today (`rollback_supported()` is false and there are no children), but it's implicit — a short comment would keep it from reading as an oversight.

src/common/event-recorder/src/event_table.rs
```diff
@@ -124,6 +124,18 @@ pub const VIEW_NAME_COLUMN: EventTableColumn =
 /// The canonical View identifier dimension.
 pub const VIEW_ID_COLUMN: EventTableColumn =
     EventTableColumn::new("view_id", ColumnDataType::Uint32, SemanticType::Field);
+/// The canonical Kafka topic name dimension.
+pub const TOPIC_NAME_COLUMN: EventTableColumn =
+    EventTableColumn::new("topic_name", ColumnDataType::String, SemanticType::Field);
+/// The canonical WAL prune boundary.
```
fengjiachun: "The canonical WAL prune boundary" reads as "the boundary that was pruned", but on `Retrying` / `Failed` / `Poisoned` rows this column carries `data.prunable_entry_id` — the *attempted* boundary, which never reached the metadata CAS. The field name is defensible ("prunable" = candidate), but the doc invites the wrong reading.

This also diverges from `RepartitionEvent::lifecycle()`, which deliberately nulls out its dimension columns on non-`Submitted` triggers precisely so lifecycle rows can't be mistaken for completion records. Either follow that pattern, or expand this doc to say it's the attempted boundary and that consumers must filter on `procedure_trigger = 'Succeeded'`.

src/meta-srv/src/procedure/wal_prune.rs
fengjiachun, line 174: Successful prunes can be lost across a metasrv restart. If the process dies after `update_pruned_entry_id` commits but before `done()`, recovery re-runs `on_prune`; the records are already physically gone, so `prunable_entry_id <= earliest_offset` now holds and the procedure returns a bare `Status::done()` with no output. No `Succeeded` event is ever recorded for a prune that did happen. Logical delete isn't affected — the records remain, so it prunes and emits again.

Recording is best-effort by design, so I wouldn't engineer around this, but it's worth a line in the rustdoc so a missing event isn't later read as "no prune occurred".

config/standalone.example.toml
```diff
@@ -907,7 +907,7 @@ ttl = "90d"
 ## `alter_database`, `drop_database`, `create_flow`, `drop_flow`,
 ## `create_table`, `create_logical_tables`, `alter_table`, `alter_logical_tables`,
 ## `drop_table`, `undrop_table`, `purge_dropped_table`, `truncate_table`,
-## `create_view`, `drop_view`.
+## `create_view`, `drop_view`, `wal_prune`.
```
fengjiachun: The WAL prune ticker and manager are only wired in `metasrv/builder.rs`, so standalone never submits a `WalPruneProcedure` in production. This list omits `region_migration`, `repartition` and `repartition_group` for the same reason. The new integration test submits the procedure by hand, which doesn't change what a standalone user can actually observe.

Suggest dropping `wal_prune` here and from the standalone half of `config/config.md`.

src/meta-srv/src/procedure/wal_prune.rs
```diff
@@ -422,4 +469,126 @@ mod tests {
...
+        let status = procedure.on_prune().await.unwrap();
+
+        assert_matches!(status, Status::Done { output: None });
+        assert_eq!(procedure.observed_latest_offset, Some(0));
+        delete_topic(context.client, &topic_name).await;
+    }
+
+    #[tokio::test]
+    async fn test_wal_prune_event_trigger_selection() {
+        maybe_skip_kafka_integration_test!();
```
fengjiachun, line 499: This test exercises pure `event()` trigger-selection logic — Kafka is only needed to construct `Context`. Gating it behind `maybe_skip_kafka_integration_test!()` means none of the trigger-selection coverage runs in normal CI. Awkward to avoid given `Context` holds an `Arc<Client>`, so not asking for a change; just flagging that the coverage isn't where it looks like it is.

src/meta-srv/src/procedure/wal_prune/utils.rs
```diff
@@ -141,7 +142,7 @@ pub(crate) async fn get_offsets_for_topic(
 pub(crate) async fn update_pruned_entry_id(
     table_metadata_manager: &TableMetadataManagerRef,
     topic: &str,
-    pruned_entry_id: u64,
+    pruned_entry_id: EntryId,
```
fengjiachun, line 145: `EntryId` is a `u64` alias, so this is a no-op that only adds an import — leftover from the first commit, where the previous boundary was threaded back out of this function. Mildly consistent with `WalPruneData::prunable_entry_id` so it's fine to keep, but it is churn in the diff.

src/meta-srv/src/procedure/wal_prune.rs
WenyXu: This is intentional: `Submitted` and `Recovered` add little value, while `Retrying`, `Failed`, and `Poisoned` provide useful operational context. I will document this trade-off and add a short comment explaining why rollback and child-submission triggers do not apply here.

src/common/event-recorder/src/event_table.rs
WenyXu: Agreed. The value is the requested prune boundary; on non-`Succeeded` events it is an attempted, not persisted, boundary. I will clarify the column documentation and note that consumers should filter on `procedure_trigger = 'Succeeded'` for completed prunes.

src/meta-srv/src/procedure/wal_prune.rs
WenyXu: Agreed. Event delivery is best effort, so I will keep the no-op path silent and document that a missing `Succeeded` event does not prove that no physical prune occurred.

config/standalone.example.toml
WenyXu: Agreed. The standalone test submits the procedure manually, but standalone does not run the WAL prune ticker or manager in production. I will remove `wal_prune` from the standalone example and its generated configuration reference.

src/meta-srv/src/procedure/wal_prune.rs
WenyXu: Agreed. This coverage is currently Kafka-gated because constructing the procedure context requires a Kafka client. I will leave it as-is for now rather than add test-only indirection solely for this trigger-selection test.

src/meta-srv/src/procedure/wal_prune/utils.rs
WenyXu: Thanks. I will keep `EntryId` here because it preserves the WAL entry-id meaning at the helper boundary, even though it is currently a `u64` alias.

WenyXu:
Good catch. `490dfe43` expanded the event scope, but I did not update the PR description. I will rewrite it to match the current triggers, columns, and payload semantics.

WenyXu:
@codex review

fengjiachun (approved):
LGTM

PR #8679, opened 2026-07-29
refactor(mito2): add prerequisites for range-based series reads

evenyag:

Follow-up to #8586.

This PR prepares the internal Mito2 read path for a range-based metric-series reader while deliberately leaving the reader implementation and runtime wiring for a follow-up PR.

- Allow an opt-in pruner to retain already-cached file-range builders after the current range batch is consumed, so candidate and data phases can share them without changing existing scan behavior.
- Preserve the original cache-eligibility rule: late pruning results are never cached after all range references are consumed, while retained builders remain query-scoped until the pruner is dropped.
- Document that repeated `add_partition_ranges()` calls must register consistent ranges from the same logical scan; their reference counts accumulate in the pruner.
- Add distinct cache-key stages for candidate-series and assigned series-data reads.
- Add primary-key range/filter and row-selection primitives needed to read selected series from Parquet.
- Keep only the shared `SeriesRange` type in `series_reader.rs`; the actual reader, batch collection, filtering, and stream assembly are deferred.
- Keep existing `SeqScan`, `SeriesScan`, and `UnorderedScan` behavior unchanged; they continue to use the non-retaining `Pruner::new` path.
- Do not add series-v2 scanbench code or two-stage runtime wiring.

There are no public API, wire-format, schema, persisted-format, or runtime behavior changes. The internal APIs intended for the follow-up are narrowly marked `allow(dead_code)` until that reader consumes them.

Validation:

- `make fmt`
- `make check`
- `make clippy`
- `cargo nextest run -p mito2` (1,052 passed)
- `make check-udeps` completed compilation but reported the pre-existing `common-stat -> nix` unused-dependency warning; this PR changes no Cargo manifests or `common-stat` code.

## PR Checklist

### DIFF
--- a/src/mito2/src/read/range_cache.rs
+++ b/src/mito2/src/read/range_cache.rs
@@ -41,6 +41,7 @@ use crate::read::BoxedRecordBatchStream;
 use crate::read::read_columns::ReadColumns;
 use crate::read::scan_region::StreamContext;
 use crate::read::scan_util::PartitionMetrics;
+use crate::read::series_reader::SeriesRange;
 use crate::region::options::MergeMode;
 use crate::sst::parquet::DEFAULT_READ_BATCH_SIZE;
 
@@ -72,6 +73,7 @@ pub(crate) struct ScanRequestFingerprint {
 enum RangeScanStage {
     Data,
     CandidateSeries,
+    SeriesData(SeriesRange),
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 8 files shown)


evenyag:
@codex review

src/mito2/src/read/range_cache.rs
```diff
@@ -72,6 +73,7 @@ pub(crate) struct ScanRequestFingerprint {
 enum RangeScanStage {
     Data,
     CandidateSeries,
+    SeriesData(SeriesRange),
```
v0y4g3r, line 76: Do we need to take TableId into consideration when caching series ranges?

evenyag: All tables in one partition use the same series ranges. So we don't need to add every table id in the key.

v0y4g3r (approved):
LGTM

PR #8680, opened 2026-07-29
feat: expose MitoRegion::all_manifest_files for metadata rebuild

sunng87:

Adds a public read-only accessor that returns all live SST file metas and the current manifest version from the region manifest. Used by the downstream project admin path to enumerate the authoritative live file set without going through the worker loop.

## PR Checklist


src/mito2/src/region.rs
```diff
@@ -849,6 +849,17 @@ impl MitoRegion {
             .collect::<Vec<_>>()
     }
 
+    /// Returns all live SST file metas and the current manifest version from
+    /// the region manifest.
+    ///
+    /// Used by the Iceberg rebuild admin path to enumerate the authoritative
+    /// live file set without going through the worker loop.
+    pub async fn all_manifest_files(&self) -> (Vec<FileMeta>, ManifestVersion) {
```
v0y4g3r, line 861: **Should-fix: staged files are missed while the region is in staging mode**

This accessor only reads the normal manifest (`manifest_ctx.manifest()`) and skips `staging_manifest()`. While the region is in staging mode (e.g. during region copy/migration, `RegionLeaderState::Staging`), the authoritative live file set lives in the staging manifest, so the metadata rebuilt by the downstream admin path would silently miss those files.

The sibling accessor `manifest_sst_entries()` (same file, ~L771) explicitly merges `manifest().files` with `staging_manifest().files`, so this new method is inconsistent with the established semantics for enumerating the live file set.

Suggestion: either (a) merge the staging manifest's files here as well (and consider which manifest version to return in that state), or (b) if callers guarantee the region is never staging, document that precondition in the rustdoc and consider warning/asserting when the region is staging, instead of silently returning an incomplete set.

v0y4g3r (commented):
One should-fix finding on the new `MitoRegion::all_manifest_files` accessor: it does not account for the staging manifest, so the returned "authoritative live file set" can be incomplete while the region is staging. Details inline.

src/mito2/src/region.rs
sunng87: Do we really need to include manifests in staging mode if we want to build external metadata like iceberg manifest? Will the staging manifests cause some duplicates?

sunng87: @v0y4g3r updated, please check again

v0y4g3r (approved):
LGTM

PR #8682, opened 2026-07-29
fix(region-migration): fence promotion by manifest version

killme2008:

N/A

Region migration previously fenced candidate promotion only by WAL entry ID. That is insufficient because compaction and index metadata publication advance the manifest without advancing WAL. A candidate could therefore become writable with stale persisted region state.

This PR adds an explicit manifest-version fence:

- Graceful leader downgrade returns the final manifest version captured while holding the Mito manifest write lock. Metric regions return independent data and metadata fences.
- Migration instructions carry the target versions to the candidate. Catchup directly installs the requested manifest and requires both the manifest and WAL entry fences before promotion. Local WAL progress cannot satisfy a fence using the candidate's own latest entry ID.
- Metric catchup first validates both physical regions while they remain followers, then promotes them only after both fences pass.
- Missing manifest fields preserve the legacy path for rolling upgrades. The new JSON fields are optional and have compatibility tests.

The internal `SetRegionRoleStateSuccess` and `RegionCatchupRequest` Rust APIs now carry manifest versions. There is no schema or persisted-data format change.

Validation:

- `cargo nextest run -p common-meta instruction` (17 passed)
- `cargo nextest run -p mito2 catchup` (10 passed)
- `cargo nextest run -p meta-srv region_migration` (92 passed)
- `cargo nextest run -p datanode downgrade_region` (9 passed)
- `cargo nextest run -p datanode upgrade_region` (5 passed)
- `cargo nextest run -p metric-engine` (95 passed)
- `make fmt`
- `make clippy`

The Kafka-parameterized catchup cases require `GT_KAFKA_ENDPOINTS`; no external Kafka broker was configured for this local run.

## PR Checklist

### DIFF
--- a/src/common/meta/src/instruction.rs
+++ b/src/common/meta/src/instruction.rs
@@ -19,6 +19,7 @@ use std::time::Duration;
 use common_error::ext::{ErrorExt, RetryHint};
 use common_error::status_code::StatusCode;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use store_api::ManifestVersion;
 use store_api::region_engine::SyncRegionFromRequest;
 use store_api::region_request::{RegionFlushReason, RegionRequirements};
 use store_api::storage::{FileRefsManifest, GcReport, RegionId, RegionNumber};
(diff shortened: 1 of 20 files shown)


WenyXu:
@codex review

WenyXu:
The previous catchup path already calls has_update() and reopens the region when it finds a newer manifest.
test_catchup_with_manifest_update verifies that this reloads the follower from manifest v0 to v2.

The candidate upgrade starts only after the old leader finishes downgrade. Manifest updates and the follower transition
share the manifest write lock: an edit/compaction either commits before the follower snapshot, or is rejected after the
region becomes Follower.

Therefore, assuming committed manifests are visible to listing, the old has_update() + reopen path should also converge
before promotion. The new compaction test proves the new path works, but does not show that the old path fails.

PR #8683, opened 2026-07-29
fix: make select whole json2 column worked

MichaelScofield:

as title

## PR Checklist


MichaelScofield:
@codex review

fengys1996:
@codex review

src/common/recordbatch/src/lib.rs
```diff
@@ -429,6 +438,63 @@ impl RecordBatches {
...
+    ) -> std::result::Result<Option<ArrayFormatter<'a>>, ArrowError> {
+        if !array.data_type().is_binary() {
+            return Ok(None);
+        }
+
+        Ok(Some(ArrayFormatter::new(
+            Box::new(BinaryFormatter {
+                array,
+                is_json: field.is_some_and(is_json_extension_type),
+                default: ArrayFormatter::try_new(array, options)?,
```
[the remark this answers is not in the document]
MichaelScofield: ArrayFormatter::try_new directly constructs the default formatter through make_default_display_index and does not consult formatter_factory. The custom factory is only invoked by make_array_formatter, so this call intentionally does not recurse.

src/datatypes/src/data_type.rs
```diff
@@ -228,6 +228,11 @@ impl ConcreteDataType {
         matches!(self, ConcreteDataType::Json(_))
     }
 
+    /// Returns whether this is a JSON2 data type.
+    pub fn is_json2(&self) -> bool {
```
fengys1996, line 232: Nice abstraction.

src/datatypes/src/vectors/json/array.rs
```diff
@@ -120,8 +120,11 @@ impl JsonArray<'_> {
         let array_fields = struct_array.fields();
         let array_columns = struct_array.columns();
         let DataType::Struct(expect_fields) = expect else {
+            if expect.is_binary() {
```
fengys1996: ~~Could we update the `widen_to` documentation to explicitly mention the newly supported Struct-to-Binary JSONB encoding?~~

src/datatypes/src/vectors/json/array.rs
```diff
@@ -120,8 +120,11 @@ impl JsonArray<'_> {
...
+                return self.widen_scalar_to(expect);
+            }
             return AlignJsonArraySnafu {
-                reason: "expect struct datatype",
+                reason: "expect struct or binary datatype",
             }
             .fail();
         };
```
fengys1996: nit pick: more clear

```suggestion
        if expect == &DataType::Binary {
            return self.widen_scalar_to(expect);
        }

        let DataType::Struct(expect_fields) = expect else {
            return AlignJsonArraySnafu {
                reason: format!("expected Struct target datatype, got {expect}"),
            }
            .fail();
        };
```

src/common/function/src/scalars/json/json_get.rs
```diff
@@ -404,9 +406,15 @@ impl Function for JsonGetWithType {
                 let arg0 = compute::cast(&arg0, &DataType::BinaryView)?;
                 let jsons = arg0.as_binary_view();
 
-                let mut builder = result_builder(len, &with_type)?;
-                jsonb_get(jsons, path, builder.as_mut())?;
-                builder.build()
+                if args.arg_fields.first().is_some_and(is_json2_extension_type) {
```
fengys1996, line 409: nitpick: it would be helpful to add some comments as well.

src/common/recordbatch/src/lib.rs
```diff
@@ -429,6 +438,63 @@ impl RecordBatches {
     }
 }
 
+#[derive(Debug)]
+struct BinaryFormatterFactory;
```
fengys1996, line 442: nitpick: Should JSON-specific formatting live in the query/SQLness presentation layer instead? Adding it here makes the generic `RecordBatches::pretty_print()` understand JSON extension metadata and JSONB encoding.

src/datatypes/src/vectors/json/array.rs
fengys1996: After some testing, it seems we don’t need to extend widen_to() here, since it is intended for lossless alignment on the write path.

src/common/recordbatch/src/lib.rs
MichaelScofield: `pretty_print` is usually seen in tests, for human, placing the formatter here provides larger utility than placing it in sqlness. As to query layer, each protocol has its own formatting method, doesn't rely on this "pretty print".

fengys1996 (approved):

fengjiachun (commented):
Superseded by https://github.com/GreptimeTeam/greptimedb/pull/8683#pullrequestreview-4828433294 — same points, anchored inline.

src/common/function/src/scalars/json/json_get.rs
```diff
@@ -404,9 +406,18 @@ impl Function for JsonGetWithType {
...
```
fengjiachun, line 409: This can still drop the path, and I'd call it blocking. It's a different mechanism from the `SELECT j, j.a` case codex flagged, and it survives the `is_json2_extension_type` fix.

The comment says a binary JSON2 argument "is therefore an already-selected scalar or root value", but that doesn't hold when one query both reads the column whole and reads a path out of it:

1. `deduce_json_type` returns `JsonNativeType::String` for the whole-column read (json_type_concretize.rs:130)
2. `deduce_json_types` merges all hints for a column, and `merge(String, Object{..})` falls to `_ => Variant` (json_type.rs:119)
3. `as_arrow_type` maps `Variant` to `Binary` (json_type.rs:143)
4. so the *path* `json_get` now receives `Binary` + JSON2 extension, takes this branch, and `path` is discarded
5. `project_to(Utf8View)` on an Object hits the `to_type.is_string()` branch and returns the **whole object serialized as a string** for every row (array.rs:354). With an explicit `::bigint` it's `Value::Object(_) => Null` instead (array.rs:418)

Repro:

```sql
select json_get(j, ''), j.a.b from json2_table;
```

(Derived from reading the code, not executed — no build cache locally. Cheap to confirm by adding it to json2.sql.)

The fix is the predicate `ensure_no_whole_json2_read` already had:

```rust
let whole_column = path.trim_start_matches('$').split('.').all(str::is_empty);
if whole_column && args.arg_fields.first().is_some_and(is_json2_extension_type) {
```

src/common/function/src/scalars/json/json_get.rs
```diff
@@ -404,9 +406,18 @@ impl Function for JsonGetWithType {
                 let arg0 = compute::cast(&arg0, &DataType::BinaryView)?;
                 let jsons = arg0.as_binary_view();
```
fengjiachun, line 407: nit: `jsons` is now only used in the `else` branch — could move it in.

src/datatypes/src/extension/json.rs
```diff
@@ -103,8 +103,22 @@ impl ExtensionType for JsonExtensionType {
...
+    field.as_ref().extension_type_name() == Some(JsonExtensionType::NAME)
+}
+
+/// Check if this field is a JSON2 extension type.
+///
+/// Legacy JSONB and JSON2 share the same JSON extension name. The column schema construction
+/// invariant is that JSON2 always stores its settings as `Some`, including default settings,
+/// while legacy JSONB stores no JSON settings. Therefore, after checking the extension name,
+/// the presence of JSON settings distinguishes JSON2 from legacy JSONB.
+pub fn is_json2_extension_type<T: AsRef<Field>>(field: T) -> bool {
```
fengjiachun, line 116: "JSON2 always stores settings as `Some`, legacy JSONB stores none" is now a correctness-critical, cross-crate invariant held together by a doc comment. It holds in production paths today (I checked), but nothing enforces it.

Concretely, these fixtures build **JSON2** columns with `JsonMetadata::default()`, so under the new semantics they model legacy JSONB:

- `src/query/src/optimizer/json_type_concretize.rs:200` — `ConcreteDataType::json2(...)` + `JsonMetadata::default()`
- `src/mito2/src/sst/parquet/json_align/schema.rs:155` — the `json_struct_field` helper
- `src/mito2/src/memtable/bulk/json_align.rs:238`

Suggest switching those to `Some(JsonSettings::default())` and adding a `debug_assert` (or a constructor that can only produce valid JSON2 metadata) where JSON2 column schemas are built — otherwise the next test written on top of these fixtures bakes in the wrong assumption.

Also: the legacy-vs-JSON2 discrimination in `json_get` is currently guarded only by the compatibility suite. A direct unit test in json_get.rs would pin the regression much closer to the code.

Minor: `Field::try_extension_type` does a `serde_json::from_str` of the extension metadata on every call, and `invoke_with_args` runs once per RecordBatch. Small next to per-row JSONB decoding, but avoidable.

src/mito2/src/read/flat_projection.rs
```diff
@@ -365,7 +358,7 @@ impl FlatProjectionMapper {
             }
 
             let field = &self.output_schema.arrow_schema().fields()[output_idx];
-            if is_structured_json_field(field) {
+            if is_json_extension_type(field) {
```
fengjiachun, line 361: This is still the wide predicate, while json_get.rs:409 now uses the narrow `is_json2_extension_type`. Now that the narrow one exists, this should use it too.

As written it violates the invariant documented on `is_structured_json_field` ("legacy JSONB ... must not enter structured JSON alignment paths"), and only avoids breaking because `project_to` short-circuits when source and target types are identical. If a legacy JSONB column ever arrives as `BinaryView` against a `Binary` output field, `can_fast_cast_types` returns false and this silently degrades into a per-row jsonb decode/re-encode round-trip.

src/common/recordbatch/src/lib.rs
```diff
@@ -429,6 +438,63 @@ impl RecordBatches {
...
+        }
+
+        if self.array.is_null(idx) {
+            write!(f, "{}", self.null)?;
+        } else {
+            let bytes = match self.array.data_type() {
+                ArrowDataType::Binary => self.array.as_binary::<i32>().value(idx),
+                ArrowDataType::LargeBinary => self.array.as_binary::<i64>().value(idx),
+                ArrowDataType::BinaryView => self.array.as_binary_view().value(idx),
+                _ => unreachable!(),
```
fengjiachun, line 488: `DataType::is_binary()` includes `FixedSizeBinary`, and `is_json_extension_type` only compares the metadata name without validating against `supports_data_type`. So a mislabeled field panics the formatter rather than degrading. Falling back to `self.default` here costs nothing.

src/common/recordbatch/src/lib.rs
```diff
@@ -429,6 +438,63 @@ impl RecordBatches {
...
+        field: Option<&'a Field>,
+    ) -> std::result::Result<Option<ArrayFormatter<'a>>, ArrowError> {
+        if !array.data_type().is_binary() {
+            return Ok(None);
+        }
+
+        Ok(Some(ArrayFormatter::new(
+            Box::new(BinaryFormatter {
+                array,
+                is_json: field.is_some_and(is_json_extension_type),
```
fengjiachun, line 458: This uses the wide predicate, so **legacy JSONB columns now render as JSON text instead of binary** in `pretty_print`. Probably an improvement, but it's a user-visible behavior change — worth confirming it's intentional and noting it in the description.

src/query/src/optimizer/json_type_concretize.rs
```diff
@@ -227,7 +127,7 @@ fn deduce_json_type(expr: &Expr) -> Result<Option<(String, JsonNativeType)>> {
         JsonNativeType::try_from(&with_type).map_err(|e| plan_datafusion_err!("{e:?}"))?;
 
     let mut split = path.rsplit(".");
-    let Some(leaf) = split.next() else {
+    let Some(leaf) = split.next().filter(|&x| !x.is_empty() && x != "$") else {
```
fengjiachun, line 130: This only inspects the last segment, so `json_get(j, 'a.')` is treated as a whole-column read rather than a read of field `a`. Same predicate as the json_get.rs:409 comment — worth extracting a shared `is_whole_column_path(path)` and using it in both places.

fengjiachun (commented):
Re-posting my earlier comment anchored inline, so each point sits on the code it's about. Sorry for the duplication.

Two pieces of context worth recording:

- The `is_json2_extension_type` change fixes a **CI-confirmed regression**, not a hypothetical one. On the previous revision `Compatibility Test (recent releases)` failed on `legacy_jsonb` while passing on the merge base: `json_get_int(j, 'a') = 10` returned zero rows, because legacy JSONB columns carry the JSON extension metadata and so took the new branch with `path` dropped. Worth a line in the PR description — it's an upgrade-only silent-wrong-results path that sqlness cannot see.
- I checked the invariant the new predicate rests on: every `JsonMetadata::default()` (i.e. `json_settings: None`) construction site in the tree is inside `#[cfg(test)]`, and the only production site is `statements.rs:167`, which goes through `unwrap_or_default()` and is therefore always `Some`. So the discrimination does hold today.

Two more that don't anchor cleanly to a diff line:

- The json2_limit table is still named `json2_disable_whole_column_read`, which now means the opposite of what it tests, and the two new cases lost their explanatory comments.
- +1 to the earlier point that removing `InnerError::Datafusion` belongs in its own PR.

I'd treat the `json_get` comment as blocking; the `flat_projection` and `is_json2_extension_type` ones are cheap and close the same class of bug at its root.

PR #8692, opened 2026-07-30
refactor: replace endpoint permission variants with named actions

shuiyisong:

This pull request updates `PermissionReq` to retain structured SQL, gRPC, and bulk-insert requests while representing context-free protocol and endpoint permissions with closed, typed `PermissionAction` constants.

It centralizes stable action names and access modes in `ALL_ACTIONS`, prevents raw-string and mode mismatches, and preserves table-target and read/write authorization behavior. A companion enterprise change maps every declared action to `SqlSelect`, `SqlInsert`, or `SqlDelete`, with future unmapped actions requiring `Admin`.

This changes the public Rust `PermissionReq` API: downstream callers using the removed context-free variants must migrate to `PermissionReq::Action(...)`.

## PR Checklist

### DIFF
--- a/src/auth/src/permission.rs
+++ b/src/auth/src/permission.rs
@@ -93,6 +93,32 @@ impl PermissionTableTargets {
     }
 }
 
+/// The access mode of a named permission action.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AccessMode {
+    Read,
+    Write,
+}
+
+pub const JAEGER_QUERY: &str = "jaeger.query";
+pub const PIPELINE_QUERY: &str = "pipeline.query";
+pub const PIPELINE_INSERT: &str = "pipeline.insert";
+pub const PIPELINE_DELETE: &str = "pipeline.delete";
+pub const DASHBOARD_QUERY: &str = "dashboard.query";
+pub const DASHBOARD_SAVE: &str = "dashboard.save";
+pub const DASHBOARD_DELETE: &str = "dashboard.delete";
+
+/// All named permission actions and their access modes.
+pub const ALL_ACTIONS: &[(&str, AccessMode)] = &[
+    (JAEGER_QUERY, AccessMode::Read),
+    (PIPELINE_QUERY, AccessMode::Read),
+    (PIPELINE_INSERT, AccessMode::Write),
+    (PIPELINE_DELETE, AccessMode::Write),
+    (DASHBOARD_QUERY, AccessMode::Read),
+    (DASHBOARD_SAVE, AccessMode::Write),
+    (DASHBOARD_DELETE, AccessMode::Write),
+];
+
 #[derive(Debug, Clone)]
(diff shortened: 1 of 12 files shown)


src/auth/src/permission.rs
```diff
@@ -93,6 +93,32 @@ impl PermissionTableTargets {
     }
 }
 
+/// The access mode of a named permission action.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AccessMode {
+    Read,
+    Write,
```
sunng87, line 100: Suggest: ReadOnly, Full
because we won't have writeonly permission.

src/auth/src/permission.rs
```diff
@@ -105,11 +131,8 @@ pub enum PermissionReq<'a> {
     PromStoreRead,
     Otlp,
     LogWrite,
-    JaegerQuery,
-    PipelineQuery,
-    PipelineManage,
-    DashboardQuery,
-    DashboardManage,
+    ReadAction(&'static str),
+    WriteAction(&'static str),
```
sunng87: Is it possible to replace this enum completely with string based action?

src/auth/src/permission.rs
shuiyisong: Not really, the `write` here doesn't include read, it only indicates the request is a read request or a write one

src/auth/src/permission.rs
shuiyisong: For `PermissionReq`: 

Technically yes, but it would lose compile-time checking, turn typos or missing mappings into runtime authorization bugs, and require separate handling for read/write classification and structured resource context. I’d keep strings only for extension-specific actions.

shuiyisong: Wait do you mean `PermissionReq` or `AccessMode`?

sunng87: Yes, I mean `PermissionReq`. As we already introduced text based action name, the validation is required.
I can see some complex actions like `SqlStatement` or `GrpcRequest` cannot be simply replaced by text based action name, but otlp or promql query should be fine.

shuiyisong: Updated in https://github.com/GreptimeTeam/greptimedb/pull/8692/commits/461f6da052a95d7a5816f207919cf8b60694aea2, PTAL

src/auth/src/permission.rs
```diff
@@ -93,23 +93,86 @@ impl PermissionTableTargets {
...
+    OTLP_WRITE,
+    LOG_WRITE,
+    JAEGER_QUERY,
+    PIPELINE_QUERY,
+    PIPELINE_INSERT,
+    PIPELINE_DELETE,
+    DASHBOARD_QUERY,
+    DASHBOARD_SAVE,
+    DASHBOARD_DELETE,
+];
```
sunng87, line 174: make sure we can override or extend this in downstream projects

sunng87 (approved):

PR #8693, opened 2026-07-30
feat(query): add native histogram result plumbing

shuiyisong:

This is the 1/4 sub PR of the 6th PR of the 8 PRs of native histogram.

This pull request adds end-to-end result plumbing for Prometheus native histograms.

- Registers native-histogram functions for distributed plan decoding while keeping annotation-producing functions on the frontend.
- Preserves PromQL warnings and infos through query output metadata.
- Serializes native histogram vector and matrix samples, including mixed float/histogram results.

The optional `histogram`, `histograms`, and `infos` fields leave existing float-only responses unchanged.

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/commutativity.rs
+++ b/src/query/src/dist_plan/commutativity.rs
@@ -31,11 +31,34 @@ use datafusion_expr::{Expr, LogicalPlan, UserDefinedLogicalNode};
 use promql::extension_plan::{
     EmptyMetric, InstantManipulate, RangeManipulate, SeriesDivide, SeriesNormalize,
 };
+use promql::functions::{
+    NativeHistogramAdd, NativeHistogramAggAvg, NativeHistogramAggSum, NativeHistogramAvgOverTime,
+    NativeHistogramDelta, NativeHistogramIDelta, NativeHistogramIRate, NativeHistogramIncrease,
+    NativeHistogramRate, NativeHistogramSub, NativeHistogramSumOverTime,
+};
 use store_api::metric_engine_consts::DATA_SCHEMA_TSID_COLUMN_NAME;
 
 use crate::dist_plan::MergeScanLogicalPlan;
 use crate::dist_plan::analyzer::AliasMapping;
 use crate::dist_plan::merge_sort::{MergeSortLogicalPlan, merge_sort_transformer};
+
+// These functions capture the frontend annotation collector. Reconstructing them on a datanode
+// would execute correctly but silently lose their warnings and info messages.
+const ANNOTATING_NATIVE_HISTOGRAM_FUNCTIONS: &[&str] = &[
+    NativeHistogramAdd::name(),
+    NativeHistogramAggAvg::name(),
+    NativeHistogramAggSum::name(),
+    NativeHistogramAvgOverTime::name(),
+    NativeHistogramDelta::name(),
+    NativeHistogramIDelta::name(),
+    NativeHistogramIRate::name(),
+    NativeHistogramIncrease::name(),
+    NativeHistogramRate::name(),
+    NativeHistogramSub::name(),
+    NativeHistogramSumOverTime::name(),
+    "prom_native_histogram_drop_bool",
+    "prom_native_histogram_drop_float",
+];
 
 #[cfg(feature = "vector_index")]
 fn is_vector_sort(sort: &Sort) -> bool {
@@ -156,6 +179,13 @@ impl Categorizer {
             LogicalPlan::Filter(filter) => Self::check_expr(&filter.predicate),
             LogicalPlan::Window(_) => Commutativity::Unimplemented,
             LogicalPlan::Aggregate(aggr) => {
+                if aggr
+                    .aggr_expr
+                    .iter()
+                    .any(Self::contains_annotating_native_histogram_function)
+                {
+                    return Ok(Commutativity::NonCommutative);
+                }
                 let is_all_steppable = is_all_aggr_exprs_steppable(&aggr.aggr_expr);
                 let matches_partition = Self::check_partition(&aggr.group_expr, &partition_cols);
                 if !matches_partition && is_all_steppable {
@@ -297,6 +327,10 @@ impl Categorizer {
     }
 
     pub fn check_expr(expr: &Expr) -> Commutativity {
+        if Self::contains_annotating_native_histogram_function(expr) {
+            return Commutativity::NonCommutative;
+        }
+
         #[allow(deprecated)]
         match expr {
             Expr::Column(_)
@@ -339,6 +373,25 @@ impl Categorizer {
         }
     }
 
+    fn contains_annotating_native_histogram_function(expr: &Expr) -> bool {
+        let mut found = false;
+        expr.apply(|expr| {
+            let name = match expr {
+                Expr::ScalarFunction(func) => Some(func.name()),
+                Expr::AggregateFunction(func) => Some(func.func.name()),
+                _ => None,
+            };
+            if name.is_some_and(|name| ANNOTATING_NATIVE_HISTOGRAM_FUNCTIONS.contains(&name)) {
+                found = true;
+                Ok(TreeNodeRecursion::Stop)
+            } else {
+                Ok(TreeNodeRecursion::Continue)
+            }
+        })
+        .expect("expression traversal is infallible");
+        found
+    }
+
     /// Return true if the given expr and partition cols satisfied the rule.
     /// In this case the plan can be treated as fully commutative.
     ///
@@ -376,6 +429,8 @@ mod tests {
 
     use datafusion_common::Column;
     use datafusion_expr::LogicalPlanBuilder;
+    use datafusion_expr::expr::ScalarFunction;
+    use promql::functions::{NativeHistogramCount, NativeHistogramDrop, NativeHistogramRate};
 
     use super::*;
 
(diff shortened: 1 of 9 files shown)


src/servers/src/http/result/prometheus_resp.rs
```diff
@@ -36,13 +39,33 @@ use serde_json::Value;
...
+    values: Vec<(f64, String)>,
+    histograms: Vec<(f64, PromNativeHistogram)>,
+}
+
+fn prometheus_native_histogram(histogram: &NativeHistogram) -> Result<PromNativeHistogram> {
+    Ok(PromNativeHistogram {
+        count: histogram.count.to_string(),
+        sum: histogram.sum.to_string(),
+        buckets: histogram
+            .to_prometheus_buckets()
```
fengjiachun, line 62: Could we preserve the terminal exponential-bucket boundary used by Prometheus before exposing these buckets? `get_bound()` currently evaluates the last finite boundary as `2^1024`, which becomes `+Inf`, so the last finite bucket collapses with the actual overflow bucket. Prometheus special-cases that boundary to `MaxFloat64` and reserves the next bucket for +/-Inf observations: https://github.com/prometheus/prometheus/blob/main/model/histogram/generic.go. Please mirror that behavior and add a JSON regression covering both buckets.

src/query/src/dist_plan/commutativity.rs
```diff
@@ -156,6 +179,13 @@ impl Categorizer {
             LogicalPlan::Filter(filter) => Self::check_expr(&filter.predicate),
             LogicalPlan::Window(_) => Commutativity::Unimplemented,
             LogicalPlan::Aggregate(aggr) => {
+                if aggr
+                    .aggr_expr
+                    .iter()
+                    .any(Self::contains_annotating_native_histogram_function)
+                {
+                    return Ok(Commutativity::NonCommutative);
+                }
```
evenyag: This will prevent pushing down the native histogram functions and affect query performance.

I think we don't have to support annotations in our first implementation. We may extend our flight record batch stream to carry the annotations later.

src/servers/src/http/prometheus.rs
```diff
@@ -78,17 +78,31 @@ use crate::prometheus_handler::{
...
     pub values: Vec<(f64, String)>,
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
+    pub histograms: Vec<(f64, PromNativeHistogram)>,
+}
+
+#[derive(Debug, Default, Serialize, Deserialize, PartialEq)]
+pub struct PromNativeHistogram {
+    pub count: String,
+    pub sum: String,
+    pub buckets: Vec<(u8, String, String, String)>,
```
evenyag, line 113: Can we document these fields?

src/servers/src/http/result/prometheus_resp.rs
```diff
@@ -36,13 +39,33 @@ use serde_json::Value;
...
+            .context(UnexpectedResultSnafu {
+                reason: "native histogram cannot be converted to Prometheus buckets".to_string(),
+            })?,
```
evenyag, line 65:
```suggestion
            .context(UnexpectedResultSnafu {
                reason: "native histogram cannot be converted to Prometheus buckets",
            })?,
```

src/servers/src/http/result/prometheus_resp.rs
```diff
@@ -259,42 +304,74 @@ impl PrometheusJsonResponse {
...
+                        .then_some((timestamp_column.value(row_index), value))
+                });
+                let histogram = native_histogram_column
+                    .and_then(|column| {
+                        read_histogram(column, row_index)
+                            .context(DataFusionSnafu)
+                            .transpose()
+                    })
+                    .transpose()?
+                    .map(|histogram| {
```
discord9, line 331: Could we filter the native-histogram stale marker before serializing this sample? For float samples we call `is_prometheus_stale_nan`, but a histogram represents staleness with the same canonical NaN bit pattern in its `sum`. As written, a stale histogram is emitted as a real sample (for example, with `sum: "NaN"`) instead of being skipped. Please check `histogram.sum` here and add a regression test using `PROMETHEUS_STALE_NAN_BITS`; ordinary NaNs should remain distinct from the canonical stale marker.

src/servers/src/http/result/prometheus_resp.rs
discord9, line 62: This exposes an incorrect boundary rule for custom-bucket histograms. For schema `-53`, positive bucket index 0 has `-Inf` as its lower bound, and Prometheus treats that bound as inclusive, so the wire boundary rule should be `3` (closed on both sides). `side_buckets()` currently assigns `OpenLeft`/`0` to every positive bucket, describing this one as `(-Inf, upper]`. Could we special-case custom schema index 0 and add a JSON regression test that includes this bucket?

src/servers/src/http/result/prometheus_resp.rs
```diff
@@ -36,13 +39,33 @@ use serde_json::Value;
...
 
+#[derive(Default)]
+struct PromSeriesSamples {
+    values: Vec<(f64, String)>,
+    histograms: Vec<(f64, PromNativeHistogram)>,
+}
+
+fn prometheus_native_histogram(histogram: &NativeHistogram) -> Result<PromNativeHistogram> {
+    Ok(PromNativeHistogram {
+        count: histogram.count.to_string(),
```
discord9: Please use a Prometheus-compatible float formatter for histogram fields rather than Rust's `f64::to_string()`. Rust emits `inf`/`-inf`, while the Prometheus HTTP wire format uses `+Inf`/`-Inf` (and `NaN`). This affects `sum` and infinite bucket bounds, including custom overflow buckets. The same formatter should be applied to count, sum, and the bucket tuple fields, with tests for the special values. This is separate from the terminal exponential-boundary issue already noted above.

src/servers/src/http/result/prometheus_resp.rs
```diff
@@ -145,8 +173,16 @@ impl PrometheusJsonResponse {
                         "expected data result, but got affected rows",
                     ),
                 };
+            meta.collect_promql_annotations();
```
discord9: How are these annotations preserved when a metric-name matcher expands into multiple concurrent PromQL queries? The reducers in `http/prometheus.rs` currently append only `resp.data` and discard `warnings`/`infos` from every response after the first. Those subqueries also clone the same `QueryContext`, so collector snapshots can depend on completion order. Please either give each subquery an independent collector and merge/deduplicate the metadata, or snapshot a request-level collector only after all subqueries finish; an instant/range expansion regression test would help cover this path.

discord9 (commented):
I found a few correctness and compatibility gaps in the native-histogram response plumbing.

src/servers/src/http/result/prometheus_resp.rs
shuiyisong: Applied in 4944cd4a. The SNAFU context now uses the static string directly without allocating a String.

src/servers/src/http/result/prometheus_resp.rs
shuiyisong: Fixed in 4944cd4a. I added format_prometheus_float and use it for histogram count, sum, and every bucket bound/count field. Tests cover +Inf, -Inf, NaN, ordinary finite values, and the terminal overflow bucket JSON.

src/servers/src/http/result/prometheus_resp.rs
shuiyisong: Fixed in 4944cd4a. side_buckets now emits ClosedBoth for custom-schema bucket index 0, and the JSON regression asserts the wire tuple [3, "-Inf", "1", "1"].

shuiyisong: Fixed in 4944cd4a. get_bound now maps the terminal exponential boundary at exponent 1024 to f64::MAX while the following overflow bucket remains +Inf. I added both a schema-level regression and a JSON regression that verifies the two buckets stay distinct.

src/servers/src/http/result/prometheus_resp.rs
shuiyisong: Fixed in 4944cd4a. The response path now checks histogram.sum with is_prometheus_stale_nan immediately after decoding and skips only the canonical stale marker. The regression uses PROMETHEUS_STALE_NAN_BITS and verifies that an ordinary NaN is still serialized.

src/servers/src/http/prometheus.rs
shuiyisong: Documented in 4944cd4a. PromNativeHistogram and its fields now describe the string encoding, bucket tuple order, and the meaning of boundary rules 0 through 3.

src/query/src/dist_plan/commutativity.rs
shuiyisong: Agreed and updated in 4944cd4a. I removed the first-iteration annotation plumbing and the non-commutative special-casing, so native histogram functions can be pushed below MergeScan again. The native histogram UDFs and UDAFs, including the two synthetic drop UDFs, are now registered in DefaultPlanDecoder for remote decoding. Annotation transport is deferred to a later implementation.

src/servers/src/http/result/prometheus_resp.rs
shuiyisong: Agreed. In 4944cd4a I removed annotation propagation from this initial implementation, including the OutputMeta collector state and HTTP infos plumbing. Concurrent matcher expansion therefore no longer snapshots or merges annotation collectors. We can add annotations later with an explicit distributed transport and request-level merge design.

src/query/src/dist_plan/commutativity.rs
discord9: For the follow-up, I think `FlightData.app_metadata` would be a better transport than Arrow schema/RecordBatch metadata. The schema is emitted once before execution, while annotations are produced while consuming the stream; schema metadata also cannot represent an annotation-only empty result reliably. We already use `app_metadata` for metrics/affected rows, so we could add a trailing `QueryAnnotations { warnings, infos }` control message before EOF, decode it alongside metrics, and aggregate/deduplicate it across region streams before populating frontend `OutputMeta`. This keeps histogram functions pushdownable and also handles annotations produced after the last batch. The important missing seam is making stream-level completion metadata available through the current `SendableRecordBatchStream`, rather than attaching annotations to individual batches.

evenyag (approved):

evenyag:
@codex review

src/query/src/query_engine/default_serializer.rs
```diff
@@ -168,6 +177,48 @@ impl SubstraitPlanDecoder for DefaultPlanDecoder {
...
+            NativeHistogramAdd::scalar_udf(),
+            NativeHistogramAvg::scalar_udf(),
+            NativeHistogramAvgOverTime::scalar_udf(),
+            NativeHistogramChanges::scalar_udf(),
+            NativeHistogramCount::scalar_udf(),
+            NativeHistogramCountOverTime::scalar_udf(),
+            NativeHistogramDelta::scalar_udf(),
+            NativeHistogramDivScalar::scalar_udf(),
+            NativeHistogramDrop::bool_false_udf(String::new(), None),
+            NativeHistogramDrop::float_null_udf(String::new(), None),
```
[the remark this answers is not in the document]
shuiyisong: Proposed fix plan for review (no code changes yet):

1. Restore a recursive `contains_annotating_native_histogram_function` check in `dist_plan/commutativity.rs` and classify any projection, filter, or aggregate containing a collector-carrying native-histogram UDF as `NonCommutative`. Run the aggregate check before step splitting so nested UDFs cannot cross `MergeScan`. Apply this to the full annotation-producing UDF set, not only the two drop UDF names.
2. Remove the two `NativeHistogramDrop` registrations from `DefaultPlanDecoder`; if the locality guard regresses, decoding should fail instead of silently replacing the message and collector.
3. Reintroduce one request-scoped annotation collector shared by all expanded PromQL subqueries, snapshot it only after every result stream finishes, then populate the final HTTP `warnings` and `infos` once.
4. Add regressions for nested projection and aggregate locality, plus a final response containing the original annotation exactly once.

This is the minimal frontend-only fix and trades some pushdown for correctness. If preserving pushdown is required, the alternative is a larger follow-up that transports trailing annotations through Flight `app_metadata` and merges them across region streams.

Please review this direction before implementation.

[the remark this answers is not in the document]
shuiyisong: @evenyag @discord9 PTAL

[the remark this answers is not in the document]
evenyag: We can ignore this comment.

PR #8697, opened 2026-07-30
fix(mito2): fence async index builds by schema generation

killme2008:

Follow-up to #8676.

Async index publication was fenced only by `FileMeta`. An index build created under an old schema could still publish after a newer `SET INDEX`, while the newer task was discarded as a duplicate. This could leave an idle SST with indexes inconsistent with the latest schema.

This PR:

- captures the schema generation together with the source `FileMeta` in `IndexBuildSource`;
- validates both under the manifest write lock before publication;
- retries only when the SST still matches and the schema generation alone became stale, avoiding a permanently missing index after an index-unrelated schema change;
- routes retries through the existing per-SST scheduler, which keeps at most one active and one coalesced pending task, so repeated schema changes cannot create a retry storm;
- stores pending builds in a `HashMap<FileId, PendingIndexBuild>`, making the one-pending-task-per-SST invariant explicit;
- builds the index and captures its fence from the same `VersionRef`;
- completes result senders for coalesced, cancelled, and scheduler-rejected tasks;
- preserves close/reopen retiring leases and prevents concurrent builds for the same SST within a logical region;
- resolves removed index versions from current manifest metadata for edit and partial truncate paths.

No public API, persisted format, or data compatibility changes.

Tests cover consecutive `SetIndexes`, stale-generation publication, retry after an unrelated `ADD COLUMN`, manifest/VersionControl lag, latest-generation publication, sender completion, scheduler rejection, and close/reopen.

Validated with:

- `cargo nextest run -p mito2 index_build --retries 0`
- targeted scheduler and manifest nextest cases
- `make clippy`

## PR Checklist


src/mito2/src/sst/index.rs
```diff
@@ -922,11 +928,14 @@ impl IndexBuildTask {
                         notify: BackgroundNotify::IndexBuildFinished(index_build_finished),
                     }
                 }
-                Ok(IndexPublication::Stale) => {
+                Ok(IndexPublication::Stale(stale)) => {
                     INDEX_PUBLICATION_STALE_TOTAL
                         .with_label_values(&["manifest_commit"])
                         .inc();
```
discord9, line 934: Nitpick (non-blocking): by the time we reach this `Stale` branch, `maybe_upload_index_file` has already written the versioned Puffin object, while `cleanup_stale_index_caches` only evicts the local cache/write-cache/stager entry. The new schema-version fence makes this reachable when the SST `FileMeta` is still current but `schema_version` has advanced.

With GC disabled (the default), that remote object is not represented in manifest `removed_files`, so non-full-listing GC cannot discover it. A successful successor will normally reuse and overwrite the same uncommitted index-version path, so this should not leak one object per consecutive schema change; however, the final object can remain orphaned if the successor is cancelled by close/truncate, fails to schedule/build, or never runs. It can still be cleaned later by a manually triggered full-listing GC, so I think accepting this trade-off is reasonable.

If we want automatic cleanup, one option is to delete the exact `RegionIndexId` here when GC is disabled and the physical path cannot be shared by repartition; otherwise, durably register it for reference-aware GC. A test that lists the object store after a schema-generation-stale publication would also make the intended behavior explicit. Feel free to resolve if this orphan window is considered acceptable/unnecessary to handle here.

discord9 (commented):
One non-blocking cleanup nitpick; the schema-generation fence itself looks correct.

src/mito2/src/sst/index.rs
killme2008: Already noted below; this is acceptable.

When using object storage, GC will be enabled by default.

discord9 (approved):

PR #8699, opened 2026-07-30
fix(metric-engine): prevent stale metadata cache fills

shuiyisong:

## Description

This pull request fixes a race between metric-engine logical-column cache fills and concurrent metadata mutations.

### Reproduction

The race requires:

- No completed cache entry, such as after startup, expiration, eviction, or invalidation.
- A logical-column lookup concurrent with create, alter, drop, or region synchronization.

The deterministic regression case uses this ordering:

1. Start a cache fill and let `load_all` capture the old metadata snapshot.
2. Pause before `Cache::try_get_with` publishes that snapshot.
3. Concurrently call `add_logical_regions`, which persists new column metadata and invalidates the cache.
4. Before this fix, invalidation completes while the cache is still empty.
5. Resume the fill. It publishes the old snapshot after invalidation, so subsequent lookups return stale metadata.

The stale entry can survive until the five-minute TTL, eviction, or another invalidation. If `MetricEngineInner::load_logical_columns` copies it into the engine state cache, the stale schema can persist longer, potentially rejecting writes to a newly added field or omitting that field from read metadata and default projections.

The same ordering can reintroduce outdated metadata after delete or metadata transformation.

### Before

Read path:

`load_logical_columns` → engine-state miss → `MetadataRegion::logical_columns` → `get_all_with_prefix` → `Cache::try_get_with` → `load_all` → Mito scan → publish cache entry

Write path:

`add_logical_regions` / `delete` / `transform_logical_region_metadata` → Mito write → cache invalidation

These paths had no ordering guarantee. A write and invalidation could complete between the cache scan and publication of its result.

### After

Cached logical-column reads hold a shared lock across cache lookup, cache fill, and prefix extraction. Metadata puts and deletes now share `write_metadata`, which holds the exclusive lock across the Mito write and subsequent invalidation.

This guarantees either:

- The write completes first, and the later cache fill reads the new metadata.
- The cache fill completes first, after which the writer persists its change and invalidates that entry.

A stale fill can therefore no longer be published after the corresponding invalidation.

### Performance impact

- Cache hits gain one uncontended asynchronous read-lock acquisition. Concurrent readers still proceed together.
- Cache misses hold the read lock during the Mito scan, so concurrent metadata mutations may wait for a slow fill.
- Metadata mutations are serialized within the `MetadataRegion` instance, including mutations for unrelated region IDs, and hold the lock across Mito I/O and cache invalidation.
- Calls satisfied by the higher-level engine state cache, normal data-region reads, and data ingestion that does not reload logical metadata are unchanged.
- No storage format, wire protocol, or public API changes are introduced.

The expected impact is concentrated in concurrent DDL, recovery, and region-synchronization workloads. If that contention becomes measurable, the lock can be sharded by metadata region ID.

## PR Checklist

### DIFF
--- a/src/metric-engine/src/metadata_region.rs
+++ b/src/metric-engine/src/metadata_region.rs
@@ -124,6 +124,9 @@ impl CacheAccessLockLease {
     fn lock(&self) -> &RwLock<()> {
         self.lock
             .as_deref()
+            // Safety: `lock` is initialized when the lease is created and is taken only
+            // by `Drop`. `read` and `write` borrow `self`, so `Drop` cannot run while
+            // this reference is in use.
             .expect("cache access lock lease must hold a lock")
     }
 }
--- a/src/metric-engine/src/metadata_region.rs
+++ b/src/metric-engine/src/metadata_region.rs
@@ -40,7 +40,7 @@ use store_api::metric_engine_consts::{
     METADATA_SCHEMA_VALUE_COLUMN_NAME,
 };
 use store_api::region_engine::RegionEngine;
-use store_api::region_request::{RegionDeleteRequest, RegionPutRequest};
+use store_api::region_request::{RegionDeleteRequest, RegionPutRequest, RegionRequest};
 use store_api::storage::{RegionId, ScanRequest};
 use tokio::sync::{OwnedRwLockReadGuard, OwnedRwLockWriteGuard, RwLock};
 
@@ -73,6 +73,10 @@ pub struct MetadataRegion {
     /// The cache should be invalidated when any new values are put into the metadata region or any
     /// values are deleted from the metadata region.
     cache: Cache<RegionId, RegionMetadataCacheEntry>,
+    /// Serializes cache fills with metadata writes and invalidation.
+    ///
+    /// ponytail: This is global; shard it by metadata region if DDL contention becomes measurable.
+    cache_access_lock: RwLock<()>,
     /// Logical lock for operations that need to be serialized. Like update & read region columns.
     ///
     /// Region entry will be registered on creating and opening logical region, and deregistered on
@@ -104,6 +108,7 @@ impl MetadataRegion {
         Self {
             mito,
             cache,
+            cache_access_lock: RwLock::new(()),
             logical_region_lock: RwLock::new(HashMap::new()),
         }
     }
@@ -395,6 +400,7 @@ impl MetadataRegion {
         metadata_region_id: RegionId,
         prefix: &str,
     ) -> Result<HashMap<String, String>> {
+        let _cache_guard = self.cache_access_lock.read().await;
         let region_metadata = self
             .cache
             .try_get_with(metadata_region_id, self.load_all(metadata_region_id))
@@ -430,14 +436,20 @@ impl MetadataRegion {
     /// doesn't check if those keys exist or not.
     async fn delete(&self, metadata_region_id: RegionId, keys: &[String]) -> Result<()> {
         let delete_request = Self::build_delete_request(keys);
+        self.write_metadata(metadata_region_id, RegionRequest::Delete(delete_request))
+            .await
+    }
+
+    async fn write_metadata(
+        &self,
+        metadata_region_id: RegionId,
+        request: RegionRequest,
+    ) -> Result<()> {
+        let _cache_guard = self.cache_access_lock.write().await;
         self.mito
-            .handle_request(
-                metadata_region_id,
-                store_api::region_request::RegionRequest::Delete(delete_request),
-            )
+            .handle_request(metadata_region_id, request)
             .await
             .context(MitoWriteOperationSnafu)?;
-        // Invalidates the region metadata cache if any values are deleted from the metadata region.
         self.cache.invalidate(&metadata_region_id).await;
 
         Ok(())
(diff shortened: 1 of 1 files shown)


WenyXu:
@codex review

src/metric-engine/src/metadata_region.rs
```diff
@@ -73,6 +73,10 @@ pub struct MetadataRegion {
     /// The cache should be invalidated when any new values are put into the metadata region or any
     /// values are deleted from the metadata region.
     cache: Cache<RegionId, RegionMetadataCacheEntry>,
+    /// Serializes cache fills with metadata writes and invalidation.
+    ///
+    /// ponytail: This is global; shard it by metadata region if DDL contention becomes measurable.
+    cache_access_lock: RwLock<()>,
```
WenyXu: **[P1] Shard cache synchronization by `metadata_region_id`**

`metadata_region_id` is both the cache key and the consistency domain, but this lock serializes cache fills and metadata mutations for every physical region in this `MetadataRegion` instance.

A slow cache fill for region A can block a create/alter/drop/sync for unrelated region B. Once that writer queues, Tokio's fair `RwLock` can also block subsequent readers for region C. This is particularly costly for state-cache misses, cache expiration/eviction, recovery, and concurrent DDL across the many small regions handled by Metric Engine.

Please use a separate keyed lock map, e.g. `RwLock<HashMap<RegionId, Arc<RwLock<()>>>>`, keyed by `metadata_region_id`. `get_all_with_prefix()` and the metadata-write path should acquire the read/write guard for that specific key.

`logical_region_lock` cannot be reused: one metadata cache entry contains the metadata of all logical regions in a physical metadata region.

src/metric-engine/src/metadata_region.rs
```diff
@@ -882,4 +879,67 @@ mod test {
         assert_eq!(logical_columns.len(), 2);
         assert_eq!(column_metadatas, logical_columns);
     }
+
+    #[tokio::test]
+    async fn metadata_write_waits_for_cache_fill() {
```
WenyXu: **[P2] Cover the intended cross-region isolation after sharding**

This test correctly verifies that a write for the *same* metadata region waits for an in-flight cache fill. Once the lock is keyed by `metadata_region_id`, please also add the inverse case: pause a cache fill for metadata region A, then verify a metadata mutation for region B completes before A is released.

That guards the reason for replacing the instance-wide lock and prevents a later refactor from accidentally restoring cross-region head-of-line blocking.

src/metric-engine/src/metadata_region.rs
```diff
@@ -73,6 +73,10 @@ pub struct MetadataRegion {
     /// The cache should be invalidated when any new values are put into the metadata region or any
     /// values are deleted from the metadata region.
     cache: Cache<RegionId, RegionMetadataCacheEntry>,
+    /// Serializes cache fills with metadata writes and invalidation.
+    ///
+    /// ponytail: This is global; shard it by metadata region if DDL contention becomes measurable.
```
WenyXu: **nit:** `ponytail:` is not an established comment marker in this crate. Since the lock should be sharded now, please remove this comment. If it remains for another reason, use the existing `TODO:` / `TODO(name):` convention instead.

src/metric-engine/src/metadata_region.rs
```diff
@@ -430,14 +436,20 @@ impl MetadataRegion {
     /// doesn't check if those keys exist or not.
     async fn delete(&self, metadata_region_id: RegionId, keys: &[String]) -> Result<()> {
         let delete_request = Self::build_delete_request(keys);
+        self.write_metadata(metadata_region_id, RegionRequest::Delete(delete_request))
+            .await
+    }
+
+    async fn write_metadata(
```
WenyXu, line 534: **nit:** Could we add a short doc comment here? This helper centralizes a correctness invariant, not only a write:

```rust
/// Writes metadata and invalidates the corresponding cache entry.
```

`write_metadata` itself is a reasonable name; I would not rename it to `write_and_invalidate`.

WenyXu (commented):
The cache-fill/write ordering fix is correct. I found one required scalability issue: its synchronization must be scoped to the metadata cache key (`metadata_region_id`), rather than the entire `MetadataRegion` instance.

src/metric-engine/src/metadata_region.rs
shuiyisong: Fixed in `ad37720e44`. `cache_access_locks` is now keyed by `metadata_region_id`, and both `get_all_with_prefix` and `write_metadata` acquire the read/write guard for that key. Unrelated metadata regions no longer share the long-lived cache-fill/write lock.

src/metric-engine/src/metadata_region.rs
shuiyisong: Added `metadata_writes_are_synchronized_per_region` in `ad37720e44`. The test pauses a cache fill for region A, verifies a same-region write waits, and then verifies a write for region B completes before A is released. The full metric-engine suite passes (95 tests).

src/metric-engine/src/metadata_region.rs
shuiyisong: Removed the `ponytail:` marker in `ad37720e44`. The field comment now only documents that synchronization is scoped per metadata region.

src/metric-engine/src/metadata_region.rs
shuiyisong: Added the suggested doc comment in `ad37720e44`: `write_metadata` now explicitly documents that it writes metadata and invalidates the corresponding cache entry. The helper name remains unchanged.

WenyXu (commented):
One small documentation nit on the new lock-lease invariant.

src/metric-engine/src/metadata_region.rs
```diff
@@ -86,6 +95,63 @@ struct RegionMetadataCacheEntry {
...
+impl CacheAccessLockLease {
+    async fn read(&self) -> RwLockReadGuard<'_, ()> {
+        self.lock().read().await
+    }
+
+    async fn write(&self) -> RwLockWriteGuard<'_, ()> {
+        self.lock().write().await
+    }
+
+    fn lock(&self) -> &RwLock<()> {
```
WenyXu, line 124: **nit:** Please add a `Safety:` comment immediately above this `expect`, consistent with the repository convention:

```rust
// Safety: `lock` is initialized when the lease is created and is taken only
// by `Drop`. `read` and `write` borrow `self`, so `Drop` cannot run while
// this reference is in use.
```

WenyXu (approved):
Rest LGTM

PR #8700, opened 2026-07-30
refactor(event): store procedure trigger as JSONB

WenyXu:

None.

Store the procedure-event envelope field `procedure_trigger` as JSONB instead of a String.

`EventTrigger` now serializes as a tagged JSON object. Simple lifecycle triggers retain their existing names, for example `{"type":"Submitted"}`; child submission and retry triggers also retain their structured fields. The canonical event-table schema and emitted API value use the existing JSONB column/encoding helpers.

Update DDL, repartition, WAL-prune, and region-migration event tests to inspect the JSONB trigger type through `json_path_match` or `json_get_string`. Add serialization contracts for all trigger variants.

Compatibility: released v1.1.4 predates `procedure_trigger`, so no released String-to-JSONB conversion is required. `v1.2.0-beta.1` exists only as an unpublished tag and is outside the supported upgrade matrix. Existing event-table schema evolution adds the JSONB field for upgrades from released versions that predate it.

## PR Checklist

### DIFF
--- a/src/common/event-recorder/src/event_table.rs
+++ b/src/common/event-recorder/src/event_table.rs
@@ -101,9 +101,9 @@ pub const PROCEDURE_ERROR_COLUMN: EventTableColumn = EventTableColumn::new(
     SemanticType::Field,
 );
 /// The canonical procedure trigger envelope column.
-pub const PROCEDURE_TRIGGER_COLUMN: EventTableColumn = EventTableColumn::new(
+pub const PROCEDURE_TRIGGER_COLUMN: EventTableColumn = EventTableColumn::json_binary(
     "procedure_trigger",
-    ColumnDataType::String,
+    ColumnDataType::Binary,
     SemanticType::Field,
 );
 /// The canonical catalog name dimension.
@@ -281,6 +281,11 @@ pub fn nullable_json(value: Option<&serde_json::Value>) -> Value {
     nullable_value(value.map(|value| ValueData::BinaryValue(jsonb::Value::from(value).to_vec())))
 }
 
+/// Builds a JSONB API value.
+pub fn jsonb_value(value: &serde_json::Value) -> Value {
+    ValueData::BinaryValue(jsonb::Value::from(value).to_vec()).into()
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -319,18 +324,35 @@ mod tests {
     fn procedure_envelope_schema_preserves_names_types_semantics_and_order() {
         assert_eq!(
             procedure_event_column_schemas(),
-            [
-                "procedure_id",
-                "procedure_state",
-                "procedure_error",
-                "procedure_trigger",
+            vec![
+                ColumnSchema {
+                    column_name: "procedure_id".to_string(),
+                    datatype: ColumnDataType::String.into(),
+                    semantic_type: SemanticType::Field.into(),
+                    ..Default::default()
+                },
+                ColumnSchema {
+                    column_name: "procedure_state".to_string(),
+                    datatype: ColumnDataType::String.into(),
+                    semantic_type: SemanticType::Field.into(),
+                    ..Default::default()
+                },
+                ColumnSchema {
+                    column_name: "procedure_error".to_string(),
+                    datatype: ColumnDataType::String.into(),
+                    semantic_type: SemanticType::Field.into(),
+                    ..Default::default()
+                },
+                ColumnSchema {
+                    column_name: "procedure_trigger".to_string(),
+                    datatype: ColumnDataType::Binary.into(),
+                    semantic_type: SemanticType::Field.into(),
+                    datatype_extension: Some(ColumnDataTypeExtension {
+                        type_ext: Some(TypeExt::JsonType(JsonTypeExtension::JsonBinary.into())),
+                    }),
+                    ..Default::default()
+                },
             ]
-            .map(|column_name| ColumnSchema {
-                column_name: column_name.to_string(),
-                datatype: ColumnDataType::String.into(),
-                semantic_type: SemanticType::Field.into(),
-                ..Default::default()
-            })
         );
     }
 
--- a/src/common/meta/src/ddl/tests/event/database.rs
+++ b/src/common/meta/src/ddl/tests/event/database.rs
@@ -194,7 +194,7 @@ fn test_database_events_preserve_procedure_envelope_contract() {
         &submitted,
         CREATE_DATABASE_EVENT_TYPE,
         "Running",
-        EventTrigger::Submitted,
+        "Submitted",
         Some("greptime"),
         Some("metrics"),
     );
@@ -202,7 +202,7 @@ fn test_database_events_preserve_procedure_envelope_contract() {
         &lifecycle,
         CREATE_DATABASE_EVENT_TYPE,
         "Done",
-        EventTrigger::Succeeded,
+        "Succeeded",
         None,
         None,
     );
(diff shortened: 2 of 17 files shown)


MichaelScofield:
can it use json2?

WenyXu:
I’d prefer to keep using `json` here for consistency with the existing `payload` column in the event table. `json2` could be a good future option if we provide an `ALTER` path that makes migrating existing data straightforward.

src/common/event-recorder/src/event_table.rs
```diff
@@ -101,9 +101,9 @@ pub const PROCEDURE_ERROR_COLUMN: EventTableColumn = EventTableColumn::new(
     SemanticType::Field,
 );
 /// The canonical procedure trigger envelope column.
-pub const PROCEDURE_TRIGGER_COLUMN: EventTableColumn = EventTableColumn::new(
+pub const PROCEDURE_TRIGGER_COLUMN: EventTableColumn = EventTableColumn::json_binary(
     "procedure_trigger",
-    ColumnDataType::String,
+    ColumnDataType::Binary,
```
fengjiachun, line 106: The PR body says the betas predate this column, but `v1.2.0-beta.1` (`dde0ca11e`, 07-28) already has it as String — #8648 landed it on 07-27, a day earlier:

```bash
git show v1.2.0-beta.1:src/common/event-recorder/src/event_table.rs | grep -A3 PROCEDURE_TRIGGER_COLUMN
```

Auto-alter only adds *missing* columns (`expr_helper.rs:105`) and row-to-region does no cast (`row_to_region.rs:44`), so mito rejects the write: `column procedure_trigger expect type String(StringType), given: BINARY` (`request.rs:207`). On a cluster upgraded from beta.1 every procedure-event insert fails, silently — errors are only logged, and recovery means dropping the table by hand.

Either confirm beta.1 isn't a supported upgrade path (fix the PR body, note the drop in the release note), or pick a new column name.

Separately: 6 of the 9 variants are unit, so users lose `procedure_trigger = 'Submitted'` and the column can no longer be a Tag. A String tag column plus a nullable JSONB detail column for `ChildSubmitted` / `Retrying` would keep both — and sidestep the compat issue.

src/common/event-recorder/src/event_table.rs
```diff
@@ -273,6 +273,11 @@ where
     nullable_value(value.map(|value| ValueData::StringValue(value.as_ref().to_string())))
 }
 
+/// Builds a JSONB API value.
+pub fn jsonb_value(value: &serde_json::Value) -> Value {
```
fengjiachun, line 285: `recorder.rs:197` is the same expression — could be `jsonb_value(&event.json_payload()?)` now.

src/common/procedure/src/procedure.rs
```diff
@@ -255,7 +255,10 @@ pub struct EventContext<'a> {
 }
 
 /// Lifecycle action that causes the framework to invoke [`Procedure::event`].
-#[derive(Debug, Clone, PartialEq, Eq)]
+///
+/// It is recorded as a tagged JSON object in the `procedure_trigger` event envelope column.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+#[serde(tag = "type")]
```
fengjiachun, line 261: `Display for EventTrigger` (line 309) has no production caller left after this PR, only `test_event_trigger_display`; same for the `RetryPhase` / `ChildSubmissionOutcome` impls it uses. Drop them or note they're kept for logging. Public API, so check the enterprise side first.

tests-integration/tests/region_migration.rs
```diff
@@ -1344,7 +1341,14 @@ async fn check_region_migration_events_system_table(
     let query = Query::select()
         .column(RegionMigrationEvents::RegionMigrationTriggerReason)
         .column(RegionMigrationEvents::ProcedureState)
-        .column(RegionMigrationEvents::ProcedureTrigger)
+        .expr_as(
+            Expr::cust("json_path_match(procedure_trigger, '$.type == \"Submitted\"')"),
+            Alias::new("procedure_submitted"),
+        )
+        .expr_as(
+            Expr::cust("json_path_match(procedure_trigger, '$.type == \"Succeeded\"')"),
```
fengjiachun: Asserting two booleans is weaker than asserting the value. `json_get_string(procedure_trigger, 'type')` (`json_get.rs:95`) keeps the original single column with `Submitted` / `Succeeded`.

Also the "equivalent to" comment at line 1330 is stale.

fengjiachun (commented):
Checked out the branch and verified locally: `cargo +nightly fmt --check`, `cargo clippy -p common-event-recorder -p common-procedure --all-targets` (clean), `cargo nextest run -p common-procedure -p common-event-recorder` (99 passed), `common-meta ddl::tests::event` (26 passed), `cargo check -p tests-integration --tests` (ok). The change itself is clean.

One blocker inline on `event_table.rs`: the compatibility conclusion in the PR body does not hold for `v1.2.0-beta.1`, which already ships `procedure_trigger` as a String. The rest are optional.

src/common/meta/src/ddl/tests/event/database.rs
```diff
@@ -314,9 +314,7 @@ fn assert_procedure_event_contract(
                 Value {
                     value_data: Some(ValueData::StringValue(String::new())),
                 },
-                Value {
-                    value_data: Some(ValueData::StringValue(procedure_trigger.to_string())),
-                },
+                jsonb_value(&serde_json::to_value(procedure_trigger).unwrap()),
```
fengjiachun: flow.rs / view.rs / repartition.rs hardcode `json!({"type": ...})`; this one derives it through `serde_json::to_value`, so renaming the tag key wouldn't turn it red. Suggest hardcoding here too — `test_event_trigger_serialization` covers the derive side.

src/common/event-recorder/src/event_table.rs
WenyXu: Thanks for catching this. `v1.2.0-beta.1` was never published as a GitHub Release and is outside the supported upgrade matrix; released `v1.1.4` predates this column. We will clarify the PR description accordingly and keep the JSONB migration.

src/common/event-recorder/src/event_table.rs
WenyXu: Good catch. Updated to reuse `jsonb_value` here.

src/common/procedure/src/procedure.rs
WenyXu: Checked the enterprise side and found no callers. Removed these unused `Display` impls and their test.

tests-integration/tests/region_migration.rs
WenyXu: Updated to project `json_get_string(procedure_trigger, 'type')` and assert `Submitted` / `Succeeded` directly, and updated the equivalent SQL comment.

src/common/meta/src/ddl/tests/event/database.rs
WenyXu: Updated the expected JSON shape to be explicit here; serialization coverage remains in `test_event_trigger_serialization`.

fengjiachun (approved):
All five points addressed — rest LGTM.

Verified on `58a63a5`: `fmt --check`, `clippy` (common-procedure / common-event-recorder / common-meta, clean), unit tests 98 + 26 passed, `cargo check -p tests-integration --tests` ok. Also confirmed `extract_new_columns` propagates `datatype_extension` (`common/grpc-expr/src/util.rs:211`), so upgrades from v1.1.4 do get a real JSON column rather than a plain Binary one.

The integration tests were only type-checked here, not executed — CI green on tests-integration is the actual gate.

Issue #4954, opened 2024-11-06
Migration region chaos test failure


killme2008:
### What type of bug is this?

Unexpected error

### What subsystems are affected?

Other

### Minimal reproduce step

CI failure at migrating regions chaos test

https://github.com/GreptimeTeam/greptimedb/actions/runs/11704397141/job/32596953912?pr=4953

The error messages as below:

```text
2024-11-06T13:37:59.361626Z  INFO fuzz_migrate_mito_regions: Region: 4398046511104,  datanode: 0
thread '<unnamed>' panicked at /home/runner/work/greptimedb/greptimedb/tests-fuzz/src/utils/wait.rs:42:6:
called `Result::unwrap()` on an `Err` value: Elapsed(())
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
==23291== ERROR: libFuzzer: deadly signal
NOTE: libFuzzer has rudimentary signal handlers.
      Combine libFuzzer with AddressSanitizer or similar for better crash reports.
SUMMARY: libFuzzer: deadly signal
MS: 0 ; base unit: 0000000000000000000000000000000000000000


artifact_prefix='tests-fuzz/artifacts/fuzz_migrate_mito_regions/'; Test unit written to tests-fuzz/artifacts/fuzz_migrate_mito_regions/crash-da39a3ee5e6b4b0d3255bfef9560[189](https://github.com/GreptimeTeam/greptimedb/actions/runs/11704397141/job/32596953912?pr=4953#step:23:191)0afd80709
Base64: 

────────────────────────────────────────────────────────────────────────────────

Failing input:

	tests-fuzz/artifacts/fuzz_migrate_mito_regions/crash-da39a3ee5e6b4b0d3255bfef95601890afd80709

Output of `std::fmt::Debug`:

	FuzzInput {
	    seed: 0,
	    columns: 29,
	    partitions: 3,
	    rows: 899,
	    inserts: 3,
	}

Reproduce with:

	cargo fuzz run --fuzz-dir tests-fuzz -D --sanitizer=none fuzz_migrate_mito_regions tests-fuzz/artifacts/fuzz_migrate_mito_regions/crash-da39a3ee5e6b4b0d3255bfef95601890afd80709

Minimize test case with:

[...]

PR #8701, opened 2026-07-30
feat: add HDFS object storage backend

### LINKED ISSUE #7330: Feature Request: Add support for locally deployed HDFS in datanode object storage

I'd like to inquire about when the datanode's object storage will support locally deployed HDFS. Currently, the object storage system supports several backends including File, S3, Oss, Azblob, and Gcs, but HDFS support is missing.

### What does the feature do?

Many organizations already have locally deployed HDFS clusters for their big data infrastructure. Adding HDFS support to GreptimeDB's datanode would allow seamless integration with existing HDFS environments, enabling users to leverage their existing storage infrastructure without having to migrate to other object storage systems.

### Implementation challenges

I noticed that GreptimeDB uses the OpenDAL library for object storage abstraction, and OpenDAL already supports HDFS. This suggests that adding HDFS support should be feasible by extending the existing object storage configuration system.

Minghan2005:

Closes #7330.

This PR adds locally deployed HDFS as a datanode object-storage backend.

- Adds `type = "Hdfs"` to `ObjectStoreConfig`.
- Uses OpenDAL's `HdfsNative` service, backed by the pure-Rust `hdfs-native` client, so normal GreptimeDB builds do not gain a JVM or libhdfs runtime requirement.
- Supports a NameNode URI, object-store root, native client options, named providers, and the existing read-cache configuration.
- Wires the new backend through the object-store factory and provider/cache helpers.
- Adds serde, factory, layered-config, and example-config tests.
- Updates the datanode and standalone example TOMLs and regenerates `config/config.md` with `toml2docs v0.1.3`.

The backend follows the same configuration and factory flow as the existing remote object stores. The OpenDAL builder validates that `name_node` is present, normalizes the configured root, and forwards the options map to the native HDFS client.

Current limitation/assumption: I validated builder construction and all configuration paths locally, but I did not run end-to-end I/O against a live HDFS cluster. This initial version assumes OpenDAL's native HDFS capabilities are sufficient for GreptimeDB's object-store access pattern. If libhdfs-specific behavior such as a Kerberos ticket-cache setting is required for the first version, I can adjust the design.

[AI-assisted - Codex] I used Codex for implementation assistance. I reviewed the change end-to-end, inspected the OpenDAL `HdfsNative` implementation and capabilities, and ran the checks below.

Validation:

- `make fmt-check`
- `cargo test -p object-store --tests` (16 passed)
- `cargo test -p cmd --test load_config_test` (12 passed)
- `cargo clippy -p object-store --all-targets --all-features -- -D warnings`
- `cargo clippy -p cmd --test load_config_test -- -D warnings`
- regenerated configuration documentation with the official `toml2docs v0.1.3` binary and verified a second generation was byte-identical

## PR Checklist


Minghan2005:
The `pull_request` CI workflows are currently waiting for first-time contributor approval (they have no jobs yet). When convenient, could a maintainer please approve the workflow runs? DCO and the `pull_request_target` checks are already passing, and the local validation is listed in the PR description. Thank you.

v0y4g3r:
@codex review

Minghan2005:
Thank you for approving the workflows. The full CI run completed with 42 successful jobs and 2 failures that appear unrelated to the HDFS change:

- `Fuzz Test with Chaos (Distributed, Remote WAL, fuzz_migrate_metric_regions)` failed before its fuzz step: Kafka readiness could not fetch broker metadata after 14 attempts, then timed out. The fuzz step was skipped.
- `Fuzz Test with Chaos (Distributed, Remote WAL, fuzz_migrate_mito_regions)` completed earlier migration/query validation, then a later seed (`seed: 10`) timed out after 60 seconds while creating a Mito table under chaos. The repository has a prior report of the same migration-chaos target timing out (#4954).

All static checks, Clippy, unused-dependency checks, both binary builds, and the other completed test jobs passed. I attempted to rerun only the failed jobs, but GitHub requires repository admin permission. Could a maintainer please rerun the failed jobs only? If either failure reproduces, I will investigate it further.

sunng87:
Hi @Minghan2005 , could you add a feature flag for this backend. We should have mysql backend gated by a feature flag, it should be the same pattern.

src/object-store/src/factory.rs
```diff
@@ -40,11 +42,28 @@ pub async fn new_raw_object_store(
...
 
+/// Creates an object store backed by a native HDFS client.
+pub async fn new_hdfs_object_store(hdfs_config: &HdfsConfig) -> Result<ObjectStore> {
+    let root = util::normalize_dir(&hdfs_config.connection.root);
+    info!(
+        "The HDFS NameNode is: {}, root is: {}",
+        hdfs_config.connection.name_node, root
+    );
+
+    let builder = HdfsNative::from(&hdfs_config.connection);
```
v0y4g3r, line 65: **The HDFS writer does not provide the failure atomicity required by manifest writes.**

OpenDAL's `HdfsNative` writer writes directly to the final HDFS path, and its `abort()` operation returns `Unsupported`. GreptimeDB writes mito manifests directly to their final paths via `write_with` in `src/mito2/src/manifest/storage/utils.rs`. If a write fails midway or the process crashes, a truncated manifest can remain visible and may prevent the region from reopening during recovery.

Please provide atomic-write semantics for HDFS, for example by writing to a temporary file in the same directory and atomically renaming it after a successful close, with cleanup on failure. This should also have a failure/recovery test; validating builder construction alone does not cover this risk.

src/object-store/src/factory.rs
```diff
@@ -40,11 +42,28 @@ pub async fn new_raw_object_store(
...
+    let operator = ObjectStore::new(builder)
```
v0y4g3r, line 72: **`COPY_REGION` will deterministically fail on this backend.**

OpenDAL 0.57's `HdfsNative` capability does not support `copy`, and `CompleteLayer` delegates copy to the underlying accessor rather than emulating it. However, `RegionFileCopier` calls `object_store.copy(...)` directly in `src/mito2/src/region/utils.rs`, so a `COPY_REGION` request backed by HDFS will return `Unsupported`.

Please add a read/write copy fallback for backends without native copy support, either in the object-store integration or in `RegionFileCopier`, and cover the HDFS `COPY_REGION` path with an integration test.

v0y4g3r (commented):
Static review of the HDFS backend integration. I found two correctness issues related to storage semantics that should be addressed before treating HDFS as a fully supported datanode backend.

Minghan2005:
Implemented the requested HDFS feature gate in `6f40faa`. HDFS dependencies are now optional and forwarded through the `object-store`, `mito2`, and `cmd` crates; the default build/dependency tree does not include `hdfs-native` or `opendal-service-hdfs-native`. I also rebased onto the current `main` and reran formatting, focused tests, Clippy, and default-feature checks successfully.

src/object-store/src/factory.rs
Minghan2005: Addressed in `6f40faa`. HDFS non-append writes now use a unique same-directory temporary path and publish with HDFS Rename2 (`overwrite=true`) only after a successful close, preserving the previous destination until the atomic rename. Abort and close-failure paths clean up the temporary file. Tests cover abort preserving the old contents/removing the temp file and successful close replacing the destination. I also avoided OpenDAL's HDFS rename path because it deletes an existing destination before renaming.

src/object-store/src/factory.rs
Minghan2005: Addressed in `6f40faa`. The HDFS compatibility layer now advertises copy support and implements copy as a streaming read/write operation into the same atomic temporary-write path, so publication uses the same atomic rename semantics. A `RegionFileCopier` integration test exercises the actual `COPY_REGION` path through an FS-backed store wrapped by this layer, in addition to the object-store fallback test.

src/cmd/tests/load_config_test.rs
```diff
@@ -79,6 +80,43 @@ fn test_load_runtime_options_without_max_blocking_threads() {
     );
 }
 
+#[test]
+#[cfg(feature = "hdfs-object-store")]
```
v0y4g3r, line 84: nit: this test will not run in CI workflows except we add `hdfs-object-store` feature here: https://github.com/minghan2005/greptimedb/blob/6f40faa8ca659a23b4d8e5f794bb336871fdfb2f/.github/workflows/develop.yml#L1003

src/object-store/src/layers/hdfs.rs
```diff
@@ -0,0 +1,473 @@
...
+}
+
+async fn copy_via_read_write<A: Access>(
+    inner: Arc<A>,
+    renamer: AtomicRenamer,
+    source_path: &str,
+    target_path: &str,
+) -> Result<Metadata> {
+    let (_, mut reader) = inner.read(source_path, OpRead::new()).await?;
+    let temporary_path = temporary_path(target_path);
```
v0y4g3r, line 326: In case of process abortion, this temporary file may linger forever. This won't block this PR but I think we may add a todo here.

src/cmd/tests/load_config_test.rs
```diff
@@ -34,6 +34,7 @@ use meta_srv::metasrv::MetasrvOptions;
 use meta_srv::selector::SelectorType;
 use metric_engine::config::EngineConfig as MetricEngineConfig;
 use mito2::config::MitoConfig;
+use object_store::config::ObjectStoreConfig;
```
v0y4g3r, line 37: This import is only used by the test case gated by `hdfs-object-store` feature. We need to remove it.
```suggestion
```

src/cmd/tests/load_config_test.rs
```diff
@@ -79,6 +80,43 @@ fn test_load_runtime_options_without_max_blocking_threads() {
...
+            enable_read_cache = false
+            options = { "dfs.client.block.write.replace-datanode-on-failure.enable" = "true" }
+        "#,
+    )
+    .unwrap();
+
+    let options =
+        GreptimeOptions::<DatanodeOptions>::load_layered_options(config.path().to_str(), "")
+            .unwrap();
+    let ObjectStoreConfig::Hdfs(hdfs) = options.component.storage.store else {
```
v0y4g3r, line 104:
```suggestion
    let object_store::config::ObjectStoreConfig::Hdfs(hdfs) = options.component.storage.store else {
```

fengjiachun:
@Minghan2005 Thanks for the contribution! Since there hasn’t been any follow-up for a while, I’m closing this PR for now. Feel free to reopen it when you’re ready to continue.

PR #8703, opened 2026-07-31
feat(mito2): add range-based metric series reader

evenyag:

Follow-up to #8679.

This PR implements the internal range-based metric series reader on top of the prerequisites merged in #8679. Runtime wiring is deliberately deferred so the reader implementation can be reviewed independently.

- Assign candidate metric series to stable partitions across the complete TSID domain.
- Read only assigned series from memtables and Parquet row groups using sparse primary-key filtering.
- Prune files and row groups whose encoded primary-key bounds do not overlap the assigned TSID range.
- Support multiple metric tables by retaining the table ID in exact series matching.
- Apply supported non-tag predicates after merge and dedup.
- Reuse partition-range results through the series-data range cache.
- Keep existing `SeqScan`, `SeriesScan`, and `UnorderedScan` behavior unchanged.
- Do not add series-v2 scanbench code or wire the two-stage reader into `SeriesScan`.

This PR has no public API, wire-format, schema, persisted-format, or runtime behavior changes. The reader and collector remain internal and are marked as dead code until the follow-up runtime wiring uses them.

Validation:

- `cargo fmt --all`
- `cargo nextest run -p mito2 'read::series_reader::tests'` (6 passed)
- `cargo nextest run -p mito2` (1,070 passed)
- `make check`
- `make clippy`
- `make check-udeps` completed compilation but reported the pre-existing `common-stat -> nix` unused-dependency warning; this PR changes no Cargo manifests or `common-stat` code.

## PR Checklist

### DIFF
--- a/src/mito2/src/read/series_candidate.rs
+++ b/src/mito2/src/read/series_candidate.rs
@@ -110,9 +110,14 @@ impl SeriesCandidateScanner {
                 reason: "candidate-series scan does not support extension ranges; use the legacy series-scan path",
             }
         );
-        debug_assert!(
+        ensure!(
             !pruner.predicate_prefilter_enabled(),
-            "candidate-series scan requires a pruner without predicate prefiltering"
+            UnexpectedSnafu {
+                reason: format!(
+                    "candidate-series scan for region {} requires a pruner without predicate prefiltering",
+                    stream_ctx.input.region_metadata().region_id
+                ),
+            }
         );
         let all_ranges = partitions.iter().flatten().copied().collect::<Vec<_>>();
         pruner.add_partition_ranges(&all_ranges);
@@ -564,6 +569,8 @@ fn decode_metric_series(
 
 #[cfg(test)]
 mod tests {
+    use std::time::Instant;
+
     use datafusion::execution::memory_pool::UnboundedMemoryPool;
     use datafusion_expr::{col, lit};
     use datatypes::arrow::array::{ArrayRef, DictionaryArray, UInt32Array};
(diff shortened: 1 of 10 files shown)


evenyag:
@codex review

evenyag:
@codex review

src/mito2/src/read/series_candidate.rs
```diff
@@ -106,6 +110,10 @@ impl SeriesCandidateScanner {
                 reason: "candidate-series scan does not support extension ranges; use the legacy series-scan path",
             }
         );
+        debug_assert!(
```
discord9: The `!pruner.predicate_prefilter_enabled()` contract is only enforced by a `debug_assert!`, so a release-build wiring mistake (reusing a prefilter-enabled pruner here) would silently skip simple predicates — they stay in `remaining_simple_filters` but nothing applies them during the candidate phase, producing wrong results instead of a loud error. Consider upgrading this to an `ensure!(..., InvalidRequestSnafu)` so misuse fails at wiring time.

Related: the companion contract — phase 2 must call `precise_filter_flat(skip_tags=true)` (as `scan_series_file_ranges` does) — is not documented on `PrunerOptions::enable_predicate_prefilter`; a `skip_tags=false` caller would silently double-apply tag predicates. Worth a rustdoc note on the option and on `SeriesReader::try_new`.

src/mito2/src/read/series_reader.rs
```diff
@@ -35,28 +66,679 @@ impl SeriesRange {
...
+            table_id: min_table_id,
+            tsid: min_tsid,
+        };
+        let max = MetricSeriesId {
+            table_id: max_table_id,
+            tsid: max_tsid,
+        };
+        if min > max {
+            return None;
+        }
```
discord9, line 181: `overlaps_encoded_bounds` conservatively keeps every row group whose encoded bounds span multiple tables (`min_table_id != max_table_id` → keep). That is correct but weak: on multi-table physical regions (common for metric tables), row-group bounds almost always span tables, so this and the file-level prune below almost never prune. Fine as a first cut — a follow-up could intersect the per-table series ranges with the row-group bounds before deciding overlap.

src/mito2/src/read/series_reader.rs
```diff
@@ -35,28 +66,679 @@ impl SeriesRange {
...
+        assert_eq!(0, SeriesRange::partition_for(0, 3));
+        assert_eq!(2, SeriesRange::partition_for(u64::MAX, 3));
+
+        for (partition, range) in ranges.iter().enumerate() {
+            assert_eq!(partition, SeriesRange::partition_for(range.start as u64, 3));
+            if range.end < TSID_DOMAIN_END {
+                assert_eq!(
+                    partition + 1,
+                    SeriesRange::partition_for(range.end as u64, 3)
+                );
```
discord9, line 622: Test coverage: the unit tests exercise the `SeriesRange`/`MetricSeriesFilter` partition math well, but `build_stream` itself is not exercised against real SSTs + memtables — no I/O, range-cache, semaphore, or merge coverage. Before this reader is wired into the scan path, it would be good to add end-to-end tests for: empty assigned partitions, zero row groups after per-range pruning, memtable-only and mixed mem+file partition ranges, and concurrent permit exhaustion with multiple readers sharing one semaphore.

src/mito2/src/read/series_reader.rs
evenyag: This is expected. And usually there is only one table.

src/mito2/src/read/series_reader.rs
evenyag: This test needs to add a lot of code for the fixture, which could be avoided in an integration test for the scanner.
I'm going to add the test later.

discord9 (approved):
LGTM. The partition math is sound, the two-phase reader (candidate discovery + range-based data read) is well-tested, and the invariant fix (ensure! + rustdoc on the prefilter contract) closes the main correctness gap. Internal dead-code code with no behavior change; wiring + e2e tests to follow in a later PR.

PR #8705, opened 2026-07-31
feat(query): plan native histogram functions

shuiyisong:

This is the 2nd PR of the 4 PRs.

This pull request adds native-histogram-aware PromQL function planning on top of the result plumbing introduced by #8693.

- Routes supported histogram, rate/delta, and over-time functions to native histogram UDFs.
- Preserves mixed float/histogram behavior and emits PromQL annotations when unsupported functions drop histogram samples.
- Adds focused planner tests covering supported functions, dropped samples, quantiles, sorting, and scalar conversion.

This change does not modify APIs, schemas, persisted data, or wire formats.

## PR Checklist


evenyag:
@codex review

src/query/src/promql/planner.rs
```diff
@@ -2753,55 +2794,182 @@ impl PromPlanner {
...
+                    Arc::new(NativeHistogramIncrease::scalar_udf_with_collector(
+                        self.promql_annotations.clone(),
+                    )),
+                    self.ctx.range.context(ExpectRangeSelectorSnafu)?,
+                )
+            }
             "increase" => ScalarFunc::ExtrapolateUdf(
                 Arc::new(Increase::scalar_udf()),
                 self.ctx.range.context(ExpectRangeSelectorSnafu)?,
             ),
```
evenyag: 1. We can assign all_field_columns_are_native_histogram_ranges to a field.
2. Can we rewrite them into if-else blocks to avoid repeating the strings to match?

src/query/src/promql/planner.rs
```diff
@@ -4005,6 +4313,77 @@ impl PromPlanner {
...
+        if field_columns.len() != 2 {
+            return None;
+        }
+        let float = field_columns.iter().find(|field| {
+            field.starts_with(OR_FLOAT_FIELD_PREFIX)
+                && Self::field_column_type(schema, field) == Some(&ArrowDataType::Float64)
+        })?;
+        let histogram = field_columns.iter().find(|field| {
+            field.starts_with(OR_HISTOGRAM_FIELD_PREFIX)
+                && Self::field_column_is_native_histogram(schema, field)
```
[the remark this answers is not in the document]
shuiyisong: Partially addressed in `167d0adeb1`. The planner now recognizes the canonical `greptime_value`/`greptime_native_histogram` instant-field pair, uses an OR empty-value predicate for alternative lanes, and adds regression coverage for `histogram_count`, `scalar`, and lane-preserving functions. This fixes the empty-result and `MultiFieldsNotSupported` cases.

The range-selector part of this comment (`rate` and the other dictionary-wrapped range functions) is not addressed in this slice. It needs mixed-window semantics rather than independently applying the float and histogram UDFs, and is deferred to the 6.3 operator/range follow-up. I am leaving this thread open until that lands.

src/query/src/promql/planner.rs
```diff
@@ -3649,6 +3879,10 @@ impl PromPlanner {
         let input_plan = self.strip_tsid_column(input_plan)?;
         self.ctx.use_tsid = false;
 
+        if self.all_field_columns_are_native_histograms(input_plan.schema()) {
+            return self.create_native_histogram_quantile_plan(phi, input_plan);
+        }
```
evenyag, line 3958: What if we have both a native histogram and a normal value field?

shuiyisong: No, at least this is not supported for the ingestion path

src/query/src/promql/planner.rs
shuiyisong: Addressed in 4fb440d7a2. The native-histogram-range predicate is now evaluated once, and each function name has a single match arm with an if/else selecting the native-histogram or float implementation. The planner test module still passes (136 tests).

src/query/src/promql/planner.rs
```diff
@@ -3020,6 +3294,25 @@ impl PromPlanner {
                     .collect_vec();
                     exprs.push(DfExpr::ScalarFunction(ScalarFunction { func, args }))
                 }
+                ScalarFunc::NativeHistogramUdf(func) => {
+                    if value_is_histogram {
+                        let args = itertools::chain!(
+                            other_input_exprs.iter().take(field_column_pos).cloned(),
+                            std::iter::once(col_expr),
```
discord9, line 3301: Mixed float+histogram tables crash every range function. The instant arms above skip the histogram lane when `alternative_samples` is true (lines 3274/3286), but these `ScalarFunc::Udf`/`ExtrapolateUdf` arms have no per-field skip and only check `all_field_columns_are_native_histogram_ranges`, which is false for a mixed table (range lanes are `Dictionary(_, Float64)`/`Dictionary(_, Struct)`, so `field_column_is_native_histogram_range` at 4440 does not match the float lane). The float rate/delta/increase/extrapolate UDF is then applied to the histogram dictionary lane, and `extrapolate_rate.rs:186` does `.downcast_ref::<Float64Array>().expect(...)` on it → panic, not a clean error.

Trigger: a table with both `greptime_value` (Float64) and `greptime_native_histogram` (Struct) fields (the schema built by your own `build_test_mixed_native_histogram_table_provider`) queried with `rate(m[5m])`, `increase`, `delta`, `idelta`, `irate`, `deriv`, `resets`, `changes`, `*_over_time`, `predict_linear`, etc.

Even if mixed-window semantics are deferred to the 6.3 follow-up, this PR should at least route the histogram lane through `NativeHistogramDrop` (with annotation) or fail with a clear error instead of feeding the float UDF a Struct dictionary. Please also add an execution-level test, since the current plan-shape tests cannot catch this.

src/query/src/promql/planner.rs
```diff
@@ -3366,8 +3659,13 @@ impl PromPlanner {
         // This error context should be computed lazily: the planner may set `ctx.table_name` to
         // `None` for derived expressions (e.g. after projecting the LHS of a vector-vector
```
discord9, line 3660: `histogram_quantile(phi, m)` on a mixed float+histogram table silently returns empty. The native-quantile branch is gated on `all_field_columns_are_native_histograms` (pure histogram only); a mixed table falls through to the classic `le` path, and without an `le` tag the planner returns `EmptyRelation` (line 3954-3962) with no error and no annotation. Prometheus computes the quantile directly from the histogram lane in this scenario.

Suggest: when any field column is a histogram lane, route to `create_native_histogram_quantile_plan` (line 3997) or fail with a clear error, rather than silently returning an empty result. This is more dangerous than the explicit error case because it is invisible to users.

src/query/src/promql/planner.rs
```diff
@@ -3366,8 +3659,13 @@ impl PromPlanner {
...
         // comparison filter). Eagerly calling `table_ref()?` here can turn a valid plan into
-        // a `TableNameNotFound` error even when `conjunction(exprs)` succeeds.
-        conjunction(exprs).with_context(|| ValueNotFoundSnafu {
+        // a `TableNameNotFound` error even when predicate construction succeeds.
+        let predicate = if preserve_any_value {
+            disjunction(exprs)
+        } else {
+            conjunction(exprs)
+        };
+        predicate.with_context(|| ValueNotFoundSnafu {
```
discord9, line 3668: Aggregations are not handled for native-histogram lanes. `create_aggregate_exprs` (line 3691) applies the DataFusion `sum/avg/min/max/stddev/var/count/quantile` UDAFs to every field column, and their signatures reject `Struct`:

- `sum(m)`, `avg(m)`, `stddev(m)`, `stdvar(m)`, `quantile(0.9, m)` → DataFusion aggregation type error.
- `count(m)`, `min(m)`, `max(m)` → plan succeeds but emits one output column per lane; the Prometheus response serializer only takes `first_field_column_index` (prometheus_resp.rs:246-255), so the second column is silently dropped, and the Struct lane of min/max is even serialized as a histogram sample alongside the float value.

`NativeHistogramAggSum`/`NativeHistogramAggAvg` are registered in the serializer (default_serializer.rs:216-217) but never planned, so they are dead from the planner's perspective. The pure-histogram drop guard at line 728 requires `all_field_columns_are_native_histograms`, which is false for mixed tables, so the drop path is bypassed there too.

Suggest: route T_SUM/T_AVG to the native agg UDAFs, and for mixed tables route the unsupported aggregators (min/max/stddev/stdvar/count_values) through the same drop-with-annotation path used by topk/bottomk (line 728-747) or return a clear error. The ingestion path argument ("dual fields not supported") doesn't apply here: the mixed schema is exactly what your own planner support (histogram_count/scalar/sort/label_replace) models, so aggregation is in-scope for this PR. An execution-level `count(mixed_metric)` test would be valuable.

discord9 (commented):
Reviewed the planner routing. I found two P1 correctness gaps (mixed float+histogram range crash and unhandled aggregations) plus a P2 silent empty result for histogram_quantile on mixed tables.

src/query/src/promql/planner.rs
```diff
@@ -377,17 +387,30 @@ impl PromPlannerContext {
...
+        table_provider: DfTableSourceProvider,
+        stmt: &EvalStmt,
+        query_engine_state: &QueryEngineState,
+        promql_annotations: Option<PromqlAnnotationCollector>,
     ) -> Result<LogicalPlan> {
         let mut planner = Self {
             table_provider,
             ctx: PromPlannerContext::from_eval_stmt(stmt),
+            promql_annotations,
         };
```
discord9, line 415: Test coverage concern: all new planner tests (around lines 8606-8910) assert `display_indent_schema()`/plan text against `EmptyTable` providers; none execute a plan with real RecordBatches. That means the mixed-table range crash (float UDF on a histogram dictionary lane, `extrapolate_rate.rs:186`), the aggregation output routing issue, and the annotation paths are all invisible to CI.

Could we add at least one execution-level test: build a real StructArray batch + `histogram_count`/`rate` on a pure-histogram provider, and one `count(mixed_metric)` output-column check? A frontend encode → `DefaultPlanDecoder` decode round-trip covering a range UDF and the drop UDF would also help.

src/query/src/promql/planner.rs
discord9, line 3301: Minor: the `__promql_or_float_*`/`__promql_or_histogram_*` fallback in `alternative_sample_columns` (lines 4429-4437) is dead code — nothing in the codebase generates those synthetic names, and the `or` operator rejects mixed float/Struct operands outright (lines 4930-5011). The fallback suggests OR-combining mixed lanes is handled when it isn't. Either remove it or wire the intended producer in the 6.3 operator follow-up.

discord9 (commented):
Follow-up review: test coverage and a dead-code branch.

src/query/src/promql/planner.rs
shuiyisong: Confirmed. We are deferring the lane-aware range-function and mixed-window work to the 6.3 PromQL operators follow-up. The prepared task covers type-directed dispatch, correct mixed-window semantics and annotations, plus real RecordBatch execution coverage. Resolving this thread as tracked for 6.3, not fixed in PR 8705.

src/query/src/promql/planner.rs
shuiyisong: Confirmed. We are deferring native and mixed histogram aggregation planning to the 6.4 follow-up. The prepared task covers NativeHistogramAggSum and NativeHistogramAggAvg selection, count, group, count_values and unsupported-aggregation semantics, output-schema cleanup, plus execution and decoder coverage. Resolving this thread as tracked for 6.4, not fixed in PR 8705.

src/query/src/promql/planner.rs
shuiyisong: Tracked across the follow-ups: 6.3 will add real-batch execution tests for range and OR lanes plus decoder round trips; 6.4 will add aggregation and output execution tests. Resolving this coverage thread as deferred to those follow-ups, not satisfied by PR 8705.

src/query/src/promql/planner.rs
shuiyisong: Confirmed. The synthetic OR lanes belong to 6.3, where the OR operator will produce them collision-safely and downstream range and operator paths will preserve and test them. Resolving this thread as tracked for 6.3, not fixed in PR 8705.

src/query/src/promql/planner.rs
shuiyisong: Fixed in `877606db6d`. `histogram_quantile` now recognizes the canonical alternative float/native-histogram pair, narrows the planner context to the histogram member, and reuses the native quantile path; unrelated multi-field inputs remain rejected. Added a regression proving the mixed table no longer produces `EmptyRelation` and selects `prom_native_histogram_quantile`. Verified with all query tests (587 passed, 2 skipped), query Clippy, formatting, and diff checks.

evenyag (approved):
Let's go ahead.

PR #8707, opened 2026-07-31
fix(meta): preserve legacy WAL options compatibility

WenyXu:

None.

- Keep per-region WAL options typed internally, while serializing metadata and procedure state with the legacy outer JSON-string format required by v1.1.x.
- Continue accepting both legacy string values and already-persisted structured WAL option objects when deserializing.
- Add a current-to-v1.1.4 downgrade compatibility path to the compatibility runner and CI, covering both distributed and standalone topologies.
- Add the initial `downgrade_compatibility` case: create and flush a table with the current binary, restart with v1.1.4, then open and query the table.

This restores compatibility for tables written by the current binary to be reopened after downgrade without reverting the typed `RegionWalOptions` refactor. The full downgrade scenario is executed by CI because it requires Docker and a downloaded v1.1.4 binary.

## PR Checklist


WenyXu:
@codex review

.github/scripts/run-compat.py
```diff
@@ -157,14 +161,51 @@ def run_for_version(
...
+            str(runner),
+            "compat",
+            "--from-bins-dir",
+            str(current_bins_dir),
+            "--to-version",
+            to_version,
+            "--topology",
+            topology,
+            "--test-filter",
+            "^downgrade_compatibility$",
```
fengjiachun, line 197: nit (non-blocking): since this run passes an exact filter, it knows exactly one case must be selected. Consider making an empty selection a hard error (e.g. an `--expect-cases 1` flag on the runner, or failing on the "No compat cases" output here) — otherwise a future version-inference drift would silently skip the downgrade check while CI stays green.

.github/scripts/run-compat.py
```diff
@@ -157,14 +161,51 @@ def run_for_version(
         run_command(real_command, env=env)
 
 
+def run_downgrade_for_version(
```
fengjiachun, line 179: nit (non-blocking): this duplicates the preview/real/env pattern of `run_for_version`. A shared helper taking the base command and group title would keep the two paths from drifting.

tests/runner/src/env/bare.rs
```diff
@@ -645,37 +645,50 @@ impl Env {
...
-    async fn wait_frontend_ready(&self, db: &GreptimeDB) {
-        let frontend_mode = db
-            .ctx
-            .get_server_mode(SERVER_MODE_FRONTEND_IDX)
-            .cloned()
-            .unwrap();
-        if let Some(addr) = frontend_mode.check_addrs().first() {
-            println!("Waiting for frontend gRPC readiness at {addr}...");
+    /// Restart a standalone instance with a new binary directory.
+    pub(crate) async fn compat_restart_standalone(&self, db: &GreptimeDB, bins_dir: PathBuf) {
```
fengjiachun: nit (non-blocking): this is byte-identical to `compat_restart_distributed` above; the two could collapse into a single `compat_restart`, which would also remove the topology match at the call site.

src/common/meta/src/wal_provider.rs
```diff
@@ -144,23 +144,39 @@ where
         .collect()
 }
 
+fn serialize_region_wal_options<S>(
```
fengjiachun, line 147: nit (non-blocking): `serializer.collect_map(...)` over the iterator would avoid the intermediate `HashMap` allocation. Cosmetic — these maps are tiny.

tests/runner/src/cmd/compat.rs
```diff
@@ -57,6 +71,12 @@ pub struct CompatCommand {
     #[clap(long)]
     to_bins_dir: Option<PathBuf>,
 
+    /// Version of the "to" GreptimeDB binary (e.g. "v0.9.5") or "current".
```
fengjiachun: nit (non-blocking): a downgrade-flavored example (e.g. "v1.1.4") would better reflect this flag's purpose.

fengjiachun (approved):
LGTM

.github/scripts/run-compat.py
WenyXu: Addressed in 87d5734cef.

Added `--expect-cases` to the runner and pass `--expect-cases 1` for each downgrade invocation, so zero or multiple selected cases now fail in both dry-run and execution.

.github/scripts/run-compat.py
WenyXu: Addressed in 87d5734cef.

Extracted the shared preview / execution / environment setup into `run_compat`, used by both upgrade and downgrade paths.

tests/runner/src/env/bare.rs
WenyXu: Addressed in d56786f31f.

Collapsed the identical topology-specific restart helpers into `compat_restart`; readiness already selects the correct endpoint from `db.is_standalone`.

src/common/meta/src/wal_provider.rs
WenyXu: Thanks. I considered this, but kept the current implementation because `serde_json::to_string` is fallible and `Serializer::collect_map` cannot directly propagate the per-entry error without another buffer. Since these maps are tiny, keeping the current code makes the error handling simpler. Happy to switch to a manual `SerializeMap` implementation if you prefer.

tests/runner/src/cmd/compat.rs
WenyXu: Addressed in 87d5734cef.

Updated the example to `v1.1.4` to match the downgrade use case.

PR #8709, opened 2026-07-31
fix(mito2): fail open when Bloom IN predicate has non-literal or unencodable members

discord9:

Fixes QX-031: `BloomFilterIndexApplierBuilder::collect_in_list` silently dropped non-literal and encoding-failed members of an `IN` list and could build a partial hard-pruning predicate. With the Bloom filter enabled, scanning could then prune rows that actually match the query (false-negative results).

The fix makes extraction fail open:

- Any non-literal member of the `IN` list disables Bloom pruning for the whole `IN` expression.
- Any encoding failure (`encode_lit` error) also disables Bloom pruning for the whole `IN` expression instead of dropping that member; the error is logged and ignored, so queries never fail.
- Independent AND subpredicates such as `col = 42` are still extracted as before.

Tests added:

- Builder unit tests: pure-literal, pure-nonliteral, mixed literal+null, all-null, mixed-nonliteral-with-AND, and encoding-failure cases.
- Engine-level witness test `qx_031_bloom_mixed_in_index_on_off_equivalent`: asserts Bloom-on and Bloom-off scans return identical rows for a mixed IN filter.

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/index/bloom_filter/applier/builder.rs
+++ b/src/mito2/src/sst/index/bloom_filter/applier/builder.rs
@@ -419,8 +419,8 @@ mod tests {
         i.lit()
     }
 
-    fn qx_031_build_predicates(exprs: &[Expr]) -> Option<BTreeMap<ColumnId, Vec<InListPredicate>>> {
-        let (_d, factory) = PuffinManagerFactory::new_for_test_block("qx_031_bloom_builder_");
+    fn build_bloom_predicates(exprs: &[Expr]) -> Option<BTreeMap<ColumnId, Vec<InListPredicate>>> {
+        let (_d, factory) = PuffinManagerFactory::new_for_test_block("bloom_builder_");
         let metadata = test_region_metadata();
         BloomFilterIndexApplierBuilder::new(
             "test".to_string(),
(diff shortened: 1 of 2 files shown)


src/mito2/src/sst/index/bloom_filter/applier/builder.rs
```diff
@@ -204,21 +204,13 @@ impl<'a> BloomFilterIndexApplierBuilder<'a> {
...
-                Ok(p) => {
-                    valid_predicates.insert(p);
-                }
-                Err(e) => warn!(e; "Failed to convert value in InList"),
+        for expr in &in_list.list {
+            let Expr::Literal(lit, _) = expr else {
+                return Ok(());
+            };
+            if !lit.is_null() {
+                valid_predicates.insert(encode_lit(lit, data_type.clone())?);
```
evenyag, line 213: Will it abort the scan?

discord9: No, it does not abort the scan. The `?` here propagates the encode error up to `traverse_and_collect`, which catches it and only logs a warning (`warn!(err; "Failed to collect bloom filter predicates, ignore it. expr: {expr}")`) — it never propagates further. So an encoding failure (or a non-literal member) only disables Bloom pruning for that whole IN expression (fail open): if it was the only predicate, `build()` returns `Ok(None)` and the scan falls back to a full scan (correct, just no pruning); if other independent predicates exist, they still prune as before. The scan itself is never aborted and query results stay correct — which is the point of the fix: the old code silently dropped members and could build a partial hard-pruning predicate that wrongly excluded matching rows.

evenyag (approved):

PR #8719, opened 2026-08-01
feat(auth): support HTTP bearer-token authentication

### LINKED ISSUE #8718: Support JWT / OAuth2 Bearer token authentication for HTTP interfaces

GreptimeDB's HTTP auth layer only supports HTTP `Basic` (plus the influxdb/splunk variants). There is **no `Bearer` / JWT / OAuth2** path, so clients that authenticate with an access token (JWT, OAuth2 token from an external IdP / OIDC provider, short-lived rotating service tokens) cannot authenticate against **any** `/v1/` HTTP interface.

Concretely, `src/servers/src/http/authorize.rs`:

- `extract_username_and_password` / `AuthScheme` only ever produces `Basic(username, password)`.
- Any `Authorization: Bearer <token>` request is rejected with `UnsupportedAuthScheme("bearer")` -> **401**, before any handler runs.

And on the trait side, `auth::UserProvider` only authenticates via `authenticate(Identity, Password)` — there is no way to authenticate a request whose identity is carried *inside* an opaque token rather than as a username/password pair.

### What does the feature do?

Allow HTTP clients to authenticate with an `Authorization: Bearer <token>` header, across every `/v1/` interface (SQL, PromQL/prom-store, InfluxDB, OTLP, logs, Jaeger, OpenTSDB, dashboards, …). The server treats the token as **opaque** and hands it to a pluggable validator that derives the identity — keeping JWT/JWKS/OIDC dependencies and policy out of the core.

This unlocks OIDC / external-IdP integration, short-lived rotating tokens, and service-to-service auth for all HTTP ingress.

### Suggested implementation (minimal, backward-compatible)

**1. Add a token-auth hook to `auth::UserProvider` (`src/auth/src/user_provider.rs`):**

```rust
#[async_trait]
pub trait UserProvider: Send + Sync {
    // ... existing authenticate / authorize / auth ...

    /// Authenticate + authorize an opaque bearer token (JWT, OAuth2 access
    /// token, ...). The provider owns validation and identity derivation.
    ///
    /// Default rejects, so password-only providers keep today's behavior.
    async fn auth_token(
        &self,
        _token: &str,
        _catalog: &str,
        _schema: &str,
    ) -> Result<UserInfoRef> {
        UnauthorizedSnafu { reason: "token auth not supported" }.fail()
    }
}
```

This mirrors `auth()` (authenticate + authorize) but with no `Identity`/`Password` — the identity is derived *from* the token inside the provider. The default impl preserves today's behavior for every existing provider.

**2. Recognize `Bearer` and dispatch in `authorize.rs`:**

Generalize the credential extraction so the scheme drives the path (rather than always decoding Basic):

```rust
enum AuthCredential {
    Basic(Username, Password),
    Bearer(String), // opaque token -> UserProvider::auth_token
}

// in inner_auth:
match extract_credential(&req)? {
    AuthCredential::Basic(u, p) => {
        user_provider.auth(Identity::UserId(&u, None), Password::PlainText(p), &catalog, &schema).await
    }
    AuthCredential::Bearer(t) => {
        user_provider.auth_token(&t, &catalog, &schema).await
    }
}
```

The influxdb/splunk branches keep working as-is (they peel their own scheme and reduce to Basic). A `Bearer` header no longer hits `UnsupportedAuthScheme`; it is routed to `auth_token`, which a provider implements to validate the token (JWT signature/exp/claims, OAuth2 introspection, …), resolve the username, and authorize.

That is the whole core change: ~1 trait method + ~1 enum arm + the dispatch. Backward compatible, no new core dependencies, and `Basic` and `Bearer` coexist on the same server.

### Alternatives considered

- **Dedicated `TokenAuthenticator` trait** stored alongside `UserProviderRef` (e.g. in `AuthState`), with the server dispatching `Bearer` -> `TokenAuthenticator` and `Basic` -> `UserProvider`. Cleaner separation of password-auth vs token-auth, and allows stacking multiple token kinds (JWT, OAuth2 introspection) without growing `UserProvider`. Slightly more plumbing.
- **Generalize the credential model**: refactor to `enum Credential { Basic { .. }, Bearer { .. }, .. }` and make `authenticate` take a `Credential`. Most flexible, but touches every `authenticate` call site (gRPC, MySQL, Postgres) — heavier than warranted for a first cut.

### Optional follow-up (not required for this feature)

A generic OAuth2 token endpoint (e.g. `POST /v1/oauth/tokens`, client-credentials grant) to *issue* tokens. That is separate from *consuming* them and can land independently.

### References

- HTTP auth layer: `src/servers/src/http/authorize.rs` (`check_http_auth`, `AuthScheme`, `extract_username_and_password`).
- Trait + credential types: `src/auth/src/user_provider.rs` (`UserProvider`), `src/auth/src/common.rs` (`Identity`, `Password`).
- Wiring: `src/frontend/src/server.rs` (`with_user_provider`).

sunng87:
## Summary

Implements the bearer-token (JWT / OAuth2) HTTP authentication path proposed in #8718. Clients can now authenticate with `Authorization: Bearer <token>` against any `/v1/` HTTP interface; today such requests are rejected with `UnsupportedAuthScheme("bearer")` → 401 before any handler runs.

The token is treated as **opaque** by the server — validation and identity derivation stay in the `UserProvider`, so JWT/JWKS/OIDC policy remains pluggable and out of core.

## Changes

- **`auth::UserProvider`** gains `auth_token(token, catalog, schema) -> Result<UserInfoRef>` with a **default that rejects** (`Error::UnsupportedAuthMethod`), so password-only providers keep today's behavior. A provider that supports token auth overrides it to validate the token, resolve it to a user, and authorize the connection.
- **`auth::Error::UnsupportedAuthMethod`** for the default-reject case (mapped to the existing `UserPasswordMismatch` status code → 401 in the HTTP layer).
- **`servers::http::authorize::inner_auth`** extracts a bearer token (`extract_bearer_token`) and, when present, authenticates via `UserProvider::auth_token`; otherwise it falls through unchanged to the username/password path (Basic / influxdb / splunk). **Basic and bearer coexist** on the same server.

## Backward compatibility

The default `auth_token` impl preserves existing behavior, and non-bearer requests take the exact same code path as before (verified by the existing `authorize` tests still passing).

## Tests

- `extract_bearer_token` recognizes `Bearer` (on either the standard or `x-greptime-auth` header) and ignores `Basic` / influxdb `Token` / splunk `Splunk` / empty.
- `inner_auth` dispatches a bearer token to `auth_token` and populates the `QueryContext` user on success; rejects on failure.
- A password-only provider (`StaticUserProvider`, default `auth_token`) rejects bearer tokens.

## Checklist

- [x]  The trait method has a default impl, so existing `UserProvider` implementors need no changes.
- [x]  No new core dependencies (token validation stays in the provider).
- [ ]  Provider implementation (JWT validation/JWKS) is intentionally out of scope here — it belongs in a downstream provider.

Closes #8718.


shuiyisong (commented):
I found two correctness issues in the bearer-auth path:

1. `extract_bearer_token` only accepts `Bearer` and `bearer`, but HTTP authentication schemes are case-insensitive ([RFC 9110 §11.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-11.1)). A valid `Authorization: BEARER <token>` falls through to `UnsupportedAuthScheme`, so `auth_token` is never called. Please split the scheme from the credential and use `eq_ignore_ascii_case("bearer")`.

2. Bearer failures on `/v1/splunk/...` return the generic 401 `ErrorResponse` at `src/servers/src/http/authorize.rs:99`, bypassing the existing Splunk HEC 403 `{"text":"Invalid token","code":4}` response. Please route this branch through the same `is_splunk_request` / `splunk_hec_err` mapping so HEC clients retain the endpoint contract.

src/auth/src/user_provider.rs
```diff
@@ -66,6 +66,26 @@ pub trait UserProvider: Send + Sync {
...
+    /// the provider validates the token and *derives* the identity from it.
+    /// The token is opaque to the server, so JWT/JWKS/OIDC validation policy
+    /// stays pluggable and out of core.
+    ///
+    /// The default rejects token auth with
+    /// [`Error::UnsupportedAuthMethod`], so password-only providers keep
+    /// today's behavior. Providers that support token auth override this to
+    /// validate the token, resolve it to a user, and
+    /// [`authorize`](Self::authorize) the connection.
+    async fn auth_token(&self, _token: &str, _catalog: &str, _schema: &str) -> Result<UserInfoRef> {
```
killme2008: To be clear
```suggestion
    async fn auth_bearer_token(&self, _token: &str, _catalog: &str, _schema: &str) -> Result<UserInfoRef> {
```

src/servers/src/http/authorize.rs
```diff
@@ -307,6 +328,23 @@ impl From<AuthScheme> for api::v1::auth_header::AuthScheme {
...
+///
+/// Returns `None` for any other scheme (`Basic`, influxdb `Token`, splunk
+/// `Splunk`, …) so the caller can fall through to the username/password path.
+fn extract_bearer_token<B>(req: &Request<B>) -> Option<String> {
+    let header = req
+        .headers()
+        .get(AUTHORIZATION_HEADER)
+        .or_else(|| req.headers().get(http::header::AUTHORIZATION))?;
+    let value = header.to_str().ok()?;
+    let token = value
```
killme2008: Convert the value to lowercase and remove the `bearer` prefix?

src/servers/src/http/authorize.rs
```diff
@@ -307,6 +328,23 @@ impl From<AuthScheme> for api::v1::auth_header::AuthScheme {
...
+        .strip_prefix("Bearer ")
+        .or_else(|| value.strip_prefix("bearer "))?;
+    (!token.is_empty()).then(|| token.to_string())
```
killme2008: Can we return `Option<&str>` to avoid copying?

killme2008 (approved):
LGTM almostly

PR #8726, opened 2026-08-03
feat: make frontend heartbeat extensible and lifecycle-safe

fengjiachun:

## What this PR does

This PR adds a generic OSS frontend heartbeat extension SPI and hardens the heartbeat connection lifecycle. It does **not** implement any Bulk Load business behavior.

- Depends on GreptimeTeam/greptime-proto#333.
- The temporary dependency pin is the heartbeat-only commit `28b52f8d02b9f8fc5f036338c4e5ca46cdd57184` from GreptimeTeam/greptime-proto#333. After GreptimeTeam/greptime-proto#333 merges, this PR must be updated to its final merge/squash SHA and reviewed again at the new exact head before becoming ready.

- Add an ordered, duplicate-safe `HeartbeatExtensions` registry and a generic `HeartbeatExtension` trait for:
  - dynamically generated request extensions on every heartbeat;
  - response handlers integrated with the existing heartbeat handler system;
  - per-successful-connection generation notifications;
  - cancellation-aware shutdown callbacks.
- Preserve the existing OSS heartbeat behavior when no extensions are registered.
- Isolate request-provider failures and extension response-handler errors/`Done` results so base heartbeat fields, other extensions, suspend handling, and cache invalidation continue safely. Conflicting extension keys are rejected atomically instead of silently overwriting existing values.
- Replace competing reconnect paths with one cancellable supervisor on the common heartbeat runtime. Repeated/concurrent starts are idempotent, send/receive failures create only one next generation, and shutdown cancels and joins handshakes, loops, retries, in-flight response handling, and extension lifecycle.
- Add a generic async OSS plugin assembly hook after `Arc<Instance>` is available and before the heartbeat task is built.
- Shut down an already-started heartbeat when frontend server startup fails, and stop heartbeat before normal server shutdown.

The wire dependency change is additive: it consumes the new `HeartbeatResponse.extensions` map from greptime-proto #333. This PR does not modify protobuf schemas.

Validation performed on exact head `d523fe20eaa4e53783f46da90b7f03a31501ccf6`:

- `cargo nextest run -p frontend -p cmd` — 193 passed, 0 skipped
- `cargo check -p frontend -p cmd`
- `cargo clippy -p frontend -p cmd --all-targets -- -D warnings`
- `git diff --check upstream/main...HEAD`
- Review Bridge/CODEX_TASK SUCCESSOR review — Round 1 CLEAN, local gate passed

`make check-udeps` could not run because `cargo-udeps` is not installed in the development environment.

## PR Checklist

### DIFF
--- a/src/frontend/src/frontend.rs
+++ b/src/frontend/src/frontend.rs
@@ -200,7 +200,9 @@ mod tests {
     use tonic::{Request, Response, Status, Streaming};
 
     use super::*;
-    use crate::heartbeat::{HeartbeatExtension, HeartbeatExtensionResult, HeartbeatExtensions};
+    use crate::heartbeat::{
+        FrontendHeartbeatExtension, FrontendHeartbeatExtensionResult, FrontendHeartbeatExtensions,
+    };
     use crate::instance::builder::FrontendBuilder;
     use crate::server::Services;
 
@@ -232,12 +234,12 @@ mod tests {
     }
 
     #[async_trait]
-    impl HeartbeatExtension for ShutdownTrackingExtension {
+    impl FrontendHeartbeatExtension for ShutdownTrackingExtension {
         fn name(&self) -> &str {
             "shutdown-tracking"
         }
 
-        async fn shutdown(&self) -> HeartbeatExtensionResult<()> {
+        async fn shutdown(&self) -> FrontendHeartbeatExtensionResult<()> {
             self.shutdown_calls.fetch_add(1, Ordering::AcqRel);
             Ok(())
         }
@@ -480,7 +482,7 @@ mod tests {
         let extension = Arc::new(ShutdownTrackingExtension {
             shutdown_calls: AtomicUsize::new(0),
         });
-        let extensions = HeartbeatExtensions::default();
+        let extensions = FrontendHeartbeatExtensions::default();
         assert!(extensions.register(extension.clone()));
         let heartbeat_task = HeartbeatTask::new(
             instance.frontend_peer_addr().to_string(),
--- a/src/frontend/src/frontend.rs
+++ b/src/frontend/src/frontend.rs
@@ -130,8 +130,11 @@ pub struct Frontend {
 
 impl Frontend {
     pub async fn start(&mut self) -> Result<()> {
-        if let Some(t) = &self.heartbeat_task {
-            t.start().await?;
+        if let Some(t) = &self.heartbeat_task
+            && let Err(error) = t.start().await
+        {
+            t.shutdown().await;
+            return Err(error);
         }
 
         if let Err(source) = self.servers.start_all().await {
@@ -162,7 +165,7 @@ impl Frontend {
 mod tests {
     use std::any::Any;
     use std::net::SocketAddr;
-    use std::sync::atomic::{AtomicBool, Ordering};
+    use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
     use std::time::Duration;
 
     use api::v1::meta::heartbeat_server::HeartbeatServer;
@@ -197,6 +200,7 @@ mod tests {
     use tonic::{Request, Response, Status, Streaming};
 
     use super::*;
+    use crate::heartbeat::{HeartbeatExtension, HeartbeatExtensionResult, HeartbeatExtensions};
     use crate::instance::builder::FrontendBuilder;
     use crate::server::Services;
 
@@ -218,9 +222,26 @@ mod tests {
 
     struct SuspendableHeartbeatServer {
         suspend: Arc<AtomicBool>,
+        fail_heartbeat: bool,
     }
 
     struct FailingServer;
+
+    struct ShutdownTrackingExtension {
+        shutdown_calls: AtomicUsize,
+    }
+
+    #[async_trait]
+    impl HeartbeatExtension for ShutdownTrackingExtension {
+        fn name(&self) -> &str {
+            "shutdown-tracking"
+        }
+
+        async fn shutdown(&self) -> HeartbeatExtensionResult<()> {
+            self.shutdown_calls.fetch_add(1, Ordering::AcqRel);
+            Ok(())
+        }
+    }
 
     #[async_trait]
     impl Server for FailingServer {
@@ -251,6 +272,10 @@ mod tests {
             &self,
             request: Request<Streaming<HeartbeatRequest>>,
         ) -> std::result::Result<Response<Self::HeartbeatStream>, Status> {
+            if self.fail_heartbeat {
+                return Err(Status::unavailable("mock initial heartbeat failure"));
+            }
+
             let (tx, rx) = mpsc::channel(4);
 
             common_runtime::spawn_global({
@@ -402,6 +427,7 @@ mod tests {
         };
         let heartbeat_server = Arc::new(SuspendableHeartbeatServer {
             suspend: Arc::new(AtomicBool::new(false)),
+            fail_heartbeat: false,
         });
         let meta_client = create_meta_client(&meta_client_options, heartbeat_server).await;
         let instance = Arc::new(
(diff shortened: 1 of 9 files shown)


fengjiachun:
@codex review

When you finish, append exactly this marker to the review summary:

src/frontend/src/frontend.rs
```diff
@@ -134,13 +134,19 @@ impl Frontend {
             t.start().await?;
```
WenyXu: If the initial heartbeat connection fails, this `?` returns before `HeartbeatTask::shutdown()` runs. That leaves registered extensions without their shutdown callback, even though they may already own resources from setup. Could we shut down the heartbeat task before returning this error, and add a frontend-level regression test for this path?

WenyXu (commented):
I left one comment about cleaning up heartbeat extensions when initial startup fails.

src/frontend/src/heartbeat.rs
```diff
@@ -15,119 +15,478 @@
...
+    async fn report_heartbeats(
+        &self,
+        sender: Arc<dyn HeartbeatRequestSender>,
+        mut outgoing_rx: Receiver<OutgoingMessage>,
+        report_interval: Duration,
+    ) -> ConnectionEnd {
+        let total_cpu_millicores = self.resource_stat.get_total_cpu_millicores();
+        let total_memory_bytes = self.resource_stat.get_total_memory_bytes();
+        let mut extensions = HashMap::new();
+        self.env_vars.clone().into_extensions(&mut extensions);
```
WenyXu: `EnvVars::into_extensions()` only borrows `self`, so this clone copies every configured environment-variable key and value unnecessarily on each new connection. Could we call `self.env_vars.into_extensions(&mut extensions)` directly?

src/frontend/src/heartbeat.rs
```diff
@@ -15,119 +15,478 @@
...
+                }
+            };
+
+            if let Some(mut request) = request {
+                if !self.add_request_extensions(&mut request).await {
+                    return ConnectionEnd::Shutdown;
+                }
+                let result = tokio::select! {
+                    _ = self.cancellation.cancelled() => return ConnectionEnd::Shutdown,
+                    result = sender.send(request.clone()) => result,
```
WenyXu: This clones the full heartbeat request on every send only so it can be logged afterward. Dynamic extension values are part of that clone. Could we log the request before calling `sender.send(request)` and move it into `send` instead?

src/frontend/src/heartbeat.rs
```diff
@@ -15,119 +15,478 @@
...
+
+/// An extension to frontend heartbeat requests, responses, and lifecycle events.
+///
+/// [`HeartbeatExtension::request_extensions`] is called for every heartbeat. A failed call is
+/// isolated from the base heartbeat and from other extensions. [`HeartbeatExtension::connected`]
+/// is called once for every successfully established connection generation, including reconnects;
+/// implementations must therefore be idempotent. [`HeartbeatExtension::shutdown`] is called once
+/// during task shutdown after heartbeat I/O has stopped.
+#[async_trait]
+pub trait HeartbeatExtension: Send + Sync {
```
WenyXu: `HeartbeatExtension` is public, but its name does not indicate that it is frontend-specific. If this SPI is meant to be reused by datanode or flownode, could we move it to a shared crate instead of making a frontend-owned API the de facto common abstraction? Otherwise, could we name it `FrontendHeartbeatExtension` (and rename the related registry/result types) to make its scope explicit?

src/frontend/src/heartbeat.rs
```diff
@@ -15,119 +15,478 @@
...
+///
+/// Mailbox parsing always runs first, followed by extension handlers, suspension state handling,
+/// and cache invalidation.
+pub fn heartbeat_response_handler_executor(
+    extensions: &HeartbeatExtensions,
+    suspend_state: Arc<AtomicBool>,
+    cache_invalidator: CacheInvalidatorRef,
+) -> HeartbeatResponseHandlerExecutorRef {
+    let mut handlers: Vec<HeartbeatResponseHandlerRef> = vec![Arc::new(ParseMailboxMessageHandler)];
+    handlers.extend(extensions.response_handlers().into_iter().map(|handler| {
```
WenyXu, line 155: The extension handlers run after mailbox parsing but before the built-in suspend and cache-invalidation handlers. A response carrying `Suspend` or `InvalidateCaches` can therefore invoke arbitrary extension work while the frontend is still active and its cache is stale. Could we preserve the existing OSS order—parse, suspend, invalidate—and run extensions afterward? They can still read `incoming_message` populated by the parser.

src/frontend/src/heartbeat.rs
```diff
@@ -15,119 +15,478 @@
...
+            && let Err(error) = self.0.handle(ctx).await
+        {
+            error!(error; "Heartbeat extension response handler failed");
+        }
+        Ok(HandleControl::Continue)
+    }
+}
+
+#[async_trait]
+trait HeartbeatConnector: Send + Sync {
```
WenyXu, line 191: Nit: `HeartbeatConnector`, the sender/stream adapters, `HeartbeatConnection`, and `HeartbeatRunner` together form a client-side heartbeat transport and supervision abstraction. If FlowNode or DataNode is expected to adopt the same cancellable lifecycle, could we consider extracting that cohesive unit into a shared crate? If this remains frontend-only, the current placement is fine.

src/frontend/src/heartbeat/tests.rs
```diff
@@ -172,5 +197,943 @@ fn test_heartbeat_task_uses_resolved_peer_addr() {
...
+
+    task.start().await.unwrap();
+    connector.wait_for_calls(2).await;
+    second.wait_for_requests(1).await;
+    assert_eq!(task.generation(), 2);
+    task.shutdown().await;
+}
+
+#[tokio::test]
+async fn test_concurrent_send_receive_failure_creates_one_generation() {
```
WenyXu, line 1047: The concurrent failure test is useful for proving a single next generation, but it can pass through the send-failure path alone. Could we add a response-stream-error-only reconnect case, plus a `Ready → Fail → Ready` reconnect plan? That would verify both receive-side recovery and retries after an unsuccessful reconnect attempt.

WenyXu (commented):
I left a few comments around heartbeat API boundaries, handler ordering, avoidable copies, and reconnect coverage.

sunng87 (approved):
lgtm as initial step

src/frontend/src/heartbeat.rs
fengjiachun: Thanks, the concern about delaying mandatory OSS handling is valid. However, simply reordering the existing handler list would change behavior: SuspendHandler consumes incoming_message via take(), while InvalidateCacheHandler consumes it and returns HandleControl::Done; HandlerGroupExecutor stops on Done. Extensions placed afterward would therefore miss Suspend and be skipped entirely for InvalidateCaches, rather than still seeing the parsed incoming_message.

For this PR I kept the intended PR2 order: Parse -> extensions -> Suspend -> Invalidate. If we want mandatory handlers to run first, I think that should be a separate handler-contract change: retain a non-destructive parsed instruction, or split mandatory handling and extension observation into separate phases, so Done cannot suppress extension dispatch. I am leaving this thread open for your view.

src/frontend/src/heartbeat.rs
fengjiachun: Thanks. This PR remains frontend-only. The connector and stream adapters are private, while HeartbeatRunner also owns frontend-specific request construction, resource and environment reporting, extension lifecycle, and response-handler policy. Extracting the whole unit now would freeze a common contract before DataNode or FlowNode requirements are known. If another node adopts the same lifecycle, we can compare the semantics and extract only the minimal shared connection/session primitives, likely near meta-client, in a follow-up.

PR #8727, opened 2026-08-03
feat(function): add json_object_keys

qazxcdswe123:

https://github.com/GreptimeTeam/greptimedb/issues/4805

`json_object_keys` function that is semantically align with [databend](https://docs.databend.com/sql/sql-functions/semi-structured-functions/object/object-keys)

1. Add `JSON_OBJECT_KEYS` and `OBJECT_KEYS` as alias.
2. Only return the outer most object keys (top level keys)
3. Return `List<Utf8>`
4. Return `NULL` on  non-object, `[]` on empty object
5. `DataFusionError::Execution` on invalid jsonb
6. Algorithm: TODO, not the same as jsonb version is different

Potential next step: update jsonb lib?

## PR Checklist


MichaelScofield:
@qazxcdswe123 I notice there's an earlier pr that already added the same "json_object_keys" function, so I suggest you can abort this pr, and maybe look at upgrading the jsonb dependency, or other good first issues?

qazxcdswe123:
> @qazxcdswe123 I notice there's an earlier pr that already added the same "json_object_keys" function, so I suggest you can abort this pr, and maybe look at upgrading the jsonb dependency, or other good first issues?

I completely understand this decision, that said, I think it's also kinda rude for someone just file a PR, and get merged, despite I first find the issue and also find the potential performance impact, all without taking to me :(

something like datafusion's `take` could help on this.

cc @fengjiachun @WenyXu

MichaelScofield:
@qazxcdswe123 We highly appreciate your effort. You’re right to feel frustrated, and I’m sorry about how this was handled. Since you had already picked up the issue and shared useful investigation, especially the potential performance impact, we should have checked the ongoing work and coordinated with you before merging an overlapping PR.
This was a miss in our coordination, not a reflection of the value or quality of your contribution. Thank you for raising it directly.
Your suggestion around DataFusion’s take is still valuable. If you’re willing, I’d be happy to help preserve that context and coordinate a follow-up improvement with you. There’s no pressure to continue, though—I understand if you’d prefer to step back.

fengjiachun:
@qazxcdswe123 Sorry for the late reply, and more importantly, sorry for how this played out. You claimed the issue first and did solid homework on it — the overlapping PR getting merged was our coordination failure, plain and simple. I understand the frustration.

I'd hate to see your investigation go to waste. The jsonb upgrade you proposed (with the ~9x object_keys win from your benchmark) is a real improvement we want, and as @MichaelScofield said, we'd happily accept it if the compatibility issues check out. Consider that one yours if you're still interested — and feel free to file an issue for the DataFusion take idea as well, it's worth tracking.

Thanks for speaking up directly. That kind of feedback helps us do better.

Issue #8454, opened 2026-07-09
Follow-up: distributed-mode start_time, processed_rows, and last_errors for flow_statistics


onepizzateam:
### What problem does the new feature solve?

#8392 shipped the initial scope from #7987 `information_schema.flow_statistics` and
`SHOW FLOW STATUS`, covering flow_id, flow_name, start_time, last_execution_time,
uptime_seconds, and state_size.

Scoped down on purpose to keep it reviewable. Three pieces from the original proposal
got left out, tracking them here so they don't get lost in PR comments:

1. start_time/uptime_seconds return NULL in distributed mode, since the heartbeat wire
   type (api::v1::meta::FlowStat) doesn't carry start-time yet. Uptime isn't visible at
   all in distributed deployments, which is the main use case.
2. processed_rows was in the original proposal, deferred.
3. last_errors was in the original proposal, deferred.
4. Aggregate state_size and last_execution_time across flownodes instead of last-writer-wins. Currently each flownode heartbeat overwrites the shared state key, so distributed-mode values reflect only the most recent reporter.

Bundling all of this into #8392 would've meant touching the shared heartbeat protocol,
two engine instrumentation points, and the system table in one large change. Splitting
keeps each change reviewable.

### What does the feature do?

1. Add start_time to api::v1::meta::FlowStat, populate it flownode-side on heartbeat,
   read it frontend/metasrv-side when building flow_statistics. Fixes the NULL in
   distributed mode.
2. Extend FlowStat with processed_rows_map: BTreeMap<FlowId, u64>, incremented during
   data processing. Surface as processed_rows.
3. Extend FlowStat with error_map: BTreeMap<FlowId, Vec<String>> (capped at 3), captured
   at the existing error-handling path. Surface as last_errors, comma-separated.

Same map-based design and thread-safety already used for state_size/last_exec_time_map.

### Implementation challenges

- Heartbeat wire format change needs backward compat, similar to how start_time_map used
  #[serde(default)] in #8392.
- Right hook points for start-time, row counting, error capture, without adding overhead
  to the execution path.
- Concurrency needs to match existing thread-safe mechanisms.
- Error buffers capped at 3 per flow.
- sqlness golden results need regenerating for distributed-mode flow_status tests.

Left as TODO(#7987-followup) in #8392. Happy to pick this up, but fine for someone else
too, just wanted it tracked separately from #8392.

onepizzateam:
[...]

PR #8729, opened 2026-08-03
fix(flow): fix flow stats aggregation and df_plan_to_sql quoting

discord9:

Related follow-up: #8454 (distributed-mode flow_statistics aggregation, also affects `information_schema.flow_statistics` from #8392).

This PR fixes two flow issues:

### 1. Distributed-mode flow stats last-writer-wins overwrite

**Summary**: In distributed mode, `information_schema.flows` / `flow_statistics` `last_execution_time` was unreliable — each flownode heartbeat put its local flow state map into the single global `__flow/state` key, so reports from different nodes overwrote each other (last-writer-wins), leaving timestamps stale or NULL even though flows had executed.

**How it works**:
- `FlowStateManager::merge(node_id, incoming)` stores each flownode's report under a per-node key `__flow/state/node/{node_id}` in the in-memory KV, then aggregates across nodes into the global key on every heartbeat: `last_exec_time_map` takes the max timestamp per flow, `state_size` takes the max (a flow runs on one active flownode).
- Reports are replaced in arrival order per node (no epoch comparison), so a node restart with a rolled-back wall clock cannot permanently drop its later reports.
- Per-node keys are cleared automatically when the in-memory KV resets on leader change; no stale entries survive a leader switch.
- `FlowStateHandler` derives node identity from `HeartbeatRequest.header.member_id` (canonical flownode identity, falls back to `peer.id`), and ignores reports without an identity instead of clobbering the global state.
- Global key `__flow/state` format and the frontend `list_flow_stats()` interface are unchanged.

### 2. `df_plan_to_sql` unquoted special characters break flush / scheduled execution

**Summary**: The custom `ForceQuoteIdentifiers` dialect only quoted identifiers containing uppercase letters. Identifiers with special characters (e.g. `:` in Prometheus-style table names like `kube_pod_cpu_cores:sum`) were left unquoted, producing invalid SQL when the plan was unparsed and re-parsed during `ADMIN FLUSH_FLOW` / scheduled execution of eval-interval SQL flows without a time-window expression — failing with `SQL statement is not supported, keyword: :`.

**How it works**:
- `df_plan_to_sql` now quotes any identifier with non-`[a-z0-9_]` characters, using double quotes instead of backticks (dialect-neutral; `GreptimeDbDialect` accepts both).

**Limitations**:
- Dropped-flow entries are not proactively removed from the per-node state (user-invisible since the flows table joins live flow metadata).
- HA leader switch resets stats until flownodes re-report (pre-existing in-memory design).

**API / data compatibility**:
- No breaking changes: `FlowStateValue` wire format and global key layout unchanged; new per-node keys are internal.
- `df_plan_to_sql` output quoting is more conservative (more identifiers quoted), which only produces more correct SQL.

## PR Checklist

- [x] This PR requires documentation updates. (will follow up on docs site)

### DIFF
--- a/src/common/meta/src/key/flow/flow_state.rs
+++ b/src/common/meta/src/key/flow/flow_state.rs
@@ -16,15 +16,19 @@ use std::collections::BTreeMap;
 use std::sync::Arc;
 
 use serde::{Deserialize, Serialize};
+use tokio::sync::Mutex;
 
 use crate::error::{self, Result};
 use crate::key::flow::FlowScoped;
 use crate::key::{FlowId, MetadataKey, MetadataValue};
 use crate::kv_backend::KvBackendRef;
-use crate::rpc::store::PutRequest;
+use crate::rpc::store::{PutRequest, RangeRequest};
 
 /// The entire FlowId to Flow Size's Map is stored directly in the value part of the key.
 pub const FLOW_STATE_KEY: &str = "state";
+
+/// The inner prefix (under `state/`) of the per-flownode flow state keys.
+pub const FLOW_STATE_NODE_KEY_PREFIX: &str = "node";
 
 /// The key of flow state.
 #[derive(Debug, Clone, Copy, PartialEq)]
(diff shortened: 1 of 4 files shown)


src/common/meta/src/key/flow/flow_state.rs
```diff
@@ -150,6 +259,78 @@ impl FlowStateManager {
...
+        let _guard = self.merge_lock.lock().await;
+
+        // 1. Store this node's latest report under its per-node key.
+        let node_key = FlowStateNodeKey::new(node_id).to_bytes();
+        let value = incoming.try_as_raw_value()?;
+        let req = PutRequest::new().with_key(node_key).with_value(value);
+        self.in_memory.put(req).await?;
+
+        // 2. Read back every per-node report and aggregate them.
+        let req = RangeRequest::new().with_prefix(FlowStateNodeKey::prefix());
```
evenyag, line 299: These per-node snapshots are never expired: a node that is permanently removed or comes back with a different ID will never replace its old key, yet every later heartbeat scans and aggregates that key. On a long-lived leader, rolling replacements therefore accumulate stale statistics and make this scan grow without bound. Could we tie these entries to flownode lease expiry (or store a report timestamp and filter/delete inactive nodes) so departed nodes stop contributing?

src/flow/src/batching_mode/utils.rs
```diff
@@ -969,12 +969,22 @@ pub(crate) async fn gen_plan_with_matching_schema(
...
+    /// uppercase letters need quoting to preserve case. Special characters
+    /// (e.g. ':' in Prometheus-style table names like
+    /// `kube_pod_cpu_cores:sum`, '.', '-', spaces, keywords) would produce
+    /// invalid SQL if left unquoted, so they are quoted as well.
     struct ForceQuoteIdentifiers;
     impl datafusion::sql::unparser::dialect::Dialect for ForceQuoteIdentifiers {
         fn identifier_quote_style(&self, identifier: &str) -> Option<char> {
-            if identifier.to_lowercase() != identifier {
-                Some('`')
+            if identifier
```
evenyag: This predicate still treats reserved words and digit-leading names as safe because they contain only `[a-z0-9_]`. For example, plans referencing valid quoted tables named `"select"` or `"123metrics"` will be unparsed without quotes and fail when flush/scheduled execution reparses the SQL. The comment also says keywords are quoted, but the implementation does not check them. Could we retain the identifier-shape and `ALL_KEYWORDS` checks used by DataFusion's default dialect, and add cases for a keyword and a leading digit?

src/common/meta/src/key/flow/flow_state.rs
```diff
@@ -150,6 +259,78 @@ impl FlowStateManager {
...
+                    .and_modify(|v: &mut i64| *v = (*v).max(ts))
+                    .or_insert(ts);
+            }
+            for (flow_id, ts) in state.start_time_map {
+                start_time_map
+                    .entry(flow_id)
+                    .and_modify(|v: &mut i64| *v = (*v).max(ts))
+                    .or_insert(ts);
+            }
+        }
```
evenyag, line 324: Can we aggregate the global key periodically in background? Now it has to scan all nodes' states and update the global one multiple times.

src/meta-srv/src/handler/flow_state_handler.rs
```diff
@@ -60,11 +75,138 @@ impl HeartbeatHandler for FlowStateHandler {
...
+                        .await
+                        .context(FlowStateHandlerSnafu)?;
+                }
+                // No usable identity in the request: ignore the report instead
+                // of falling back to a whole-map replace, which would clobber
+                // other nodes' reports.
+                None => {
+                    warn!(
+                        "Ignore flow state report without node identity (no header.member_id and no peer.id): {value:?}"
+                    );
```
evenyag: Is it possible? Will it produce a large amount of warnings?

src/flow/src/batching_mode/utils.rs
discord9: Good catch on digit-leading names — fixed: `ForceQuoteIdentifiers` now also quotes identifiers starting with a digit (e.g. `123metrics`), with a regression test added.

On reserved words: I intentionally did not add an ALL_KEYWORDS check. Quoting every member of sqlparser`s ALL_KEYWORDS would also quote common non-reserved column names like `number` (NUMBER is in ALL_KEYWORDS but is not reserved), which changes the unparsed output for a lot of existing flows and broke several existing tests. The unparse-failure path already has an InsertIntoPlan fallback (task.rs:699-709), so a keyword-named identifier degrades gracefully rather than failing hard. If you think keyword quoting is worth the over-quoting cost, I can follow up separately.

src/meta-srv/src/handler/flow_state_handler.rs
discord9: Fixed — downgraded the identity-less report log from `warn!` to `debug!`. Normal flownodes always carry header.member_id/peer.id, so this path only fires for old or malformed clients; debug avoids an anomalous sender spamming warn every 3s heartbeat.

src/common/meta/src/key/flow/flow_state.rs
discord9: Agreed this is a real limitation. The impact is bounded in practice: dead-node keys are tiny (a JSON map per node) and are cleared on any leader change (the in-memory KV resets). The flows table is metadata-driven so dead-node flows that are dropped are invisible; only flow_statistics could briefly show a stale row until the flow is re-created on a live node. I initially planned to prune dead nodes against FlownodeLeaseKey inside merge(), but FlownodeLeaseKey lives in meta-srv while FlowStateManager is in common-meta, so that would invert the crate dependency. A timestamp-based prune in merge() (drop per-node entries not updated within N heartbeats) is a reasonable follow-up — I have noted it as a known limitation in the code comment for now. Happy to do it in a follow-up PR.

src/common/meta/src/key/flow/flow_state.rs
discord9: The per-heartbeat merge is intentionally kept synchronous: the range scan and both puts hit only the in-memory KV (no network, no disk), and the number of per-node keys is bounded by the number of flownodes (each node replaces its own key). A background aggregator would (a) add up to one aggregation-period of freshness lag to information_schema.flows.last_execution_time, a user-visible field, and (b) require leader-aware task lifecycle handling in metasrv. At realistic flownode counts the synchronous merge is a few local JSON decodes every 3s — negligible next to per-heartbeat lease/region-stat processing. Prefer keeping it simple; can revisit if profiling shows otherwise.

src/common/meta/src/key/flow/flow_state.rs
evenyag: Maybe we can create an issue to track it.

evenyag (approved):

PR #8730, opened 2026-08-03
feat: support enabling skip_wal with ALTER TABLE

fengjiachun:

This PR supports changing `skip_wal` from `false` to `true` with `ALTER TABLE`. The transition is intentionally one-way; unsetting it or changing it back to `false` is rejected.

The implementation:

- updates table metadata first, then applies the runtime option to every leader and follower replica;
- keeps follower handling in memory and separate from the leader path that may flush;
- serializes the operation with region migration, validates the captured physical route, and bounds route-change retries;
- supports Mito tables and Metric Engine physical tables while rejecting logical Metric Engine and unsupported physical engines;
- keeps the typed table option canonical, preserves an explicit table-level `skip_wal=false` override, and renders it correctly in `SHOW CREATE TABLE`.

No RFC, protobuf, manifest format, or schema format change is included.

Verification:

- focused unit tests for table options, ALTER procedure flow, and Mito skip-WAL behavior;
- `cargo clippy -p table -p operator -p query -p store-api -p common-meta -p mito2 --all-targets -- -D warnings`;
- standalone and distributed `skip_wal` and `create_database_opts` SQLness cases;
- independent Review Bridge review: `rb-2026-08-05T082419-828Z-4f8f483d` (`LOCAL_GATE_PASSED`).

## PR Checklist

### DIFF
--- a/src/common/meta/src/ddl/alter_table.rs
+++ b/src/common/meta/src/ddl/alter_table.rs
@@ -16,6 +16,7 @@ mod executor;
 mod metadata;
 mod region_request;
 
+use std::collections::HashSet;
 use std::vec;
 
 use api::region::RegionResponse;
@@ -73,6 +74,9 @@ pub struct AlterTableProcedure {
     /// The alter table executor.
     executor: AlterTableExecutor,
 }
+
+#[derive(Debug)]
+pub(crate) struct RegionRouteChanged;
 
 /// Builds the executor from the [`AlterTableData`].
 ///
@@ -170,6 +174,24 @@ impl AlterTableProcedure {
                     operation: "setting skip_wal on logical tables".to_string()
                 }
             );
+            let current_region_ids = physical_table_route
+                .region_routes
+                .iter()
+                .map(|route| route.region.id)
+                .collect::<HashSet<_>>();
+            let locked_region_ids = self
+                .data
+                .region_locks
+                .iter()
+                .copied()
+                .collect::<HashSet<_>>();
+            if physical_table_route.region_routes.len() != self.data.region_locks.len()
+                || current_region_ids != locked_region_ids
+            {
+                // Procedure lock keys are fixed at submission. Finish this attempt so the
+                // DDL manager can submit another procedure with locks for the current route.
+                return Ok(Status::done_with_output(RegionRouteChanged));
+            }
             self.data.region_distribution =
                 Some(region_distribution(&physical_table_route.region_routes));
             self.data.alter_regions_after_metadata = true;
--- a/src/common/meta/src/ddl/alter_table.rs
+++ b/src/common/meta/src/ddl/alter_table.rs
@@ -244,23 +244,47 @@ impl AlterTableProcedure {
         // Puts the poison before submitting alter region requests to datanodes.
         self.put_poison(ctx_provider, procedure_id).await?;
         let flow = self.data.flow();
-        let results = if flow == AlterTableFlow::MetadataFirst {
-            self.executor
+        if flow == AlterTableFlow::MetadataFirst {
+            let results = self
+                .executor
                 .on_alter_skip_wal_regions(
                     &self.context.node_manager,
                     &physical_table_route.region_routes,
                     alter_kind,
                 )
-                .await
-        } else {
-            self.executor
-                .on_alter_regions(
-                    &self.context.node_manager,
-                    &physical_table_route.region_routes,
-                    alter_kind,
-                )
-                .await
-        };
+                .await;
+
+            return match handle_multiple_results(results) {
+                MultipleResults::PartialRetryable(error) => Err(error),
+                MultipleResults::PartialNonRetryable(error)
+                | MultipleResults::AllNonRetryable(error) => {
+                    // The metadata already enables skip-WAL. Keep retrying until every
+                    // replica applies the runtime option instead of leaving the table in
+                    // a permanently inconsistent state.
+                    Err(BoxedError::new(error)).context(RetryLaterSnafu {
+                        clean_poisons: true,
+                    })
+                }
+                MultipleResults::AllRetryable(error) => {
+                    Err(BoxedError::new(error)).context(RetryLaterSnafu {
+                        clean_poisons: true,
+                    })
+                }
+                MultipleResults::Ok(results) => {
+                    self.handle_alter_region_response(results)?;
+                    Ok(Status::executing_with_clean_poisons(true))
+                }
+            };
+        }
+
+        let results = self
+            .executor
+            .on_alter_regions(
+                &self.context.node_manager,
+                &physical_table_route.region_routes,
+                alter_kind,
+            )
+            .await;
 
         match handle_multiple_results(results) {
             MultipleResults::PartialRetryable(error) => {
@@ -268,11 +292,6 @@ impl AlterTableProcedure {
                 Err(error)
             }
             MultipleResults::PartialNonRetryable(error) => {
-                if flow == AlterTableFlow::MetadataFirst {
-                    return Err(BoxedError::new(error)).context(RetryLaterSnafu {
-                        clean_poisons: true,
-                    });
-                }
                 error!(error; "Partial non-retryable errors occurred during alter table, table {}, table_id: {}", self.data.table_ref(), self.data.table_id());
                 // No retry will be done.
                 Ok(Status::poisoned(
@@ -288,19 +307,12 @@ impl AlterTableProcedure {
                 })
             }
             MultipleResults::Ok(results) => {
-                if flow != AlterTableFlow::MetadataFirst {
-                    self.submit_sync_region_requests(&results, &physical_table_route.region_routes)
-                        .await;
-                }
+                self.submit_sync_region_requests(&results, &physical_table_route.region_routes)
+                    .await;
                 self.handle_alter_region_response(results)?;
                 Ok(Status::executing_with_clean_poisons(true))
             }
             MultipleResults::AllNonRetryable(error) => {
-                if flow == AlterTableFlow::MetadataFirst {
-                    return Err(BoxedError::new(error)).context(RetryLaterSnafu {
-                        clean_poisons: true,
-                    });
-                }
                 error!(error; "All alter requests returned non-retryable errors for table {}, table_id: {}", self.data.table_ref(), self.data.table_id());
                 // It assumes the metadata on datanode is not changed.
                 // Case: The alter region request is sent but not applied. (e.g., InvalidArgument)
--- a/src/common/meta/src/ddl/alter_table.rs
+++ b/src/common/meta/src/ddl/alter_table.rs
@@ -194,13 +194,8 @@ impl AlterTableProcedure {
             }
             self.data.region_distribution =
                 Some(region_distribution(&physical_table_route.region_routes));
-            self.data.alter_regions_after_metadata = true;
-            self.data.state = AlterTableState::UpdateMetadata;
-        } else if is_metadata_only_alter(alter_kind) {
-            self.data.state = AlterTableState::UpdateMetadata;
-        } else {
-            self.data.state = AlterTableState::SubmitAlterRegionRequests;
-        };
+        }
+        self.data.state = self.data.flow().after_prepare();
         Ok(Status::executing(true))
     }
 
@@ -248,15 +243,24 @@ impl AlterTableProcedure {
         ensure!(!leaders.is_empty(), NoLeaderSnafu { table_id });
         // Puts the poison before submitting alter region requests to datanodes.
         self.put_poison(ctx_provider, procedure_id).await?;
-        let results = self
-            .executor
-            .on_alter_regions(
-                &self.context.node_manager,
-                &physical_table_route.region_routes,
-                alter_kind,
-                self.data.alter_regions_after_metadata,
-            )
-            .await;
+        let flow = self.data.flow();
+        let results = if flow == AlterTableFlow::MetadataFirst {
+            self.executor
+                .on_alter_skip_wal_regions(
+                    &self.context.node_manager,
+                    &physical_table_route.region_routes,
+                    alter_kind,
+                )
+                .await
+        } else {
+            self.executor
+                .on_alter_regions(
+                    &self.context.node_manager,
+                    &physical_table_route.region_routes,
+                    alter_kind,
+                )
+                .await
+        };
 
         match handle_multiple_results(results) {
             MultipleResults::PartialRetryable(error) => {
@@ -264,7 +268,7 @@ impl AlterTableProcedure {
                 Err(error)
             }
             MultipleResults::PartialNonRetryable(error) => {
-                if self.data.alter_regions_after_metadata {
+                if flow == AlterTableFlow::MetadataFirst {
                     return Err(BoxedError::new(error)).context(RetryLaterSnafu {
                         clean_poisons: true,
                     });
@@ -284,7 +288,7 @@ impl AlterTableProcedure {
                 })
             }
             MultipleResults::Ok(results) => {
-                if !self.data.alter_regions_after_metadata {
+                if flow != AlterTableFlow::MetadataFirst {
                     self.submit_sync_region_requests(&results, &physical_table_route.region_routes)
                         .await;
                 }
@@ -292,7 +296,7 @@ impl AlterTableProcedure {
                 Ok(Status::executing_with_clean_poisons(true))
             }
             MultipleResults::AllNonRetryable(error) => {
-                if self.data.alter_regions_after_metadata {
+                if flow == AlterTableFlow::MetadataFirst {
                     return Err(BoxedError::new(error)).context(RetryLaterSnafu {
                         clean_poisons: true,
                     });
@@ -319,11 +323,7 @@ impl AlterTableProcedure {
                 "altering table result doesn't contains extension key `{TABLE_COLUMN_METADATA_EXTENSION_KEY}`,leaving the table's column metadata unchanged"
             );
         }
-        self.data.state = if self.data.alter_regions_after_metadata {
-            AlterTableState::InvalidateTableCache
-        } else {
-            AlterTableState::UpdateMetadata
-        };
+        self.data.state = self.data.flow().after_regions();
         Ok(())
     }
 
@@ -355,7 +355,8 @@ impl AlterTableProcedure {
         let table_info_value = self.data.table_info_value.as_ref().unwrap();
         // Safety: Checked in `AlterTableProcedure::new`.
         let alter_kind = self.data.task.alter_table.kind.as_ref().unwrap();
-        let metadata_only_alter = is_metadata_only_alter(alter_kind);
+        let flow = self.data.flow();
+        let metadata_only_alter = flow == AlterTableFlow::MetadataOnly;
 
         // Gets the table info from the cache or builds it.
         let  new_info = match &self.new_table_info {
@@ -385,11 +386,7 @@ impl AlterTableProcedure {
         info!(
             "Updated table metadata for table {table_ref}, table_id: {table_id}, kind: {alter_kind:?}"
         );
-        self.data.state = if self.data.alter_regions_after_metadata {
-            AlterTableState::SubmitAlterRegionRequests
-        } else {
-            AlterTableState::InvalidateTableCache
-        };
+        self.data.state = flow.after_metadata();
         Ok(Status::executing(true))
     }
 
@@ -465,6 +462,46 @@ fn is_metadata_only_alter(alter_kind: &Kind) -> bool {
             keys.len() == 1 && keys[0].as_str() == REPARTITION_COLUMN_HINT_KEY
         }
         _ => false,
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum AlterTableFlow {
+    RegionFirst,
+    MetadataFirst,
+    MetadataOnly,
+}
+
+impl AlterTableFlow {
+    fn from_kind(kind: &Kind) -> Self {
+        if only_enables_skip_wal(kind) {
+            Self::MetadataFirst
+        } else if is_metadata_only_alter(kind) {
+            Self::MetadataOnly
+        } else {
+            Self::RegionFirst
+        }
+    }
+
+    fn after_prepare(self) -> AlterTableState {
+        match self {
+            Self::RegionFirst => AlterTableState::SubmitAlterRegionRequests,
+            Self::MetadataFirst | Self::MetadataOnly => AlterTableState::UpdateMetadata,
+        }
+    }
+
+    fn after_regions(self) -> AlterTableState {
+        match self {
+            Self::RegionFirst => AlterTableState::UpdateMetadata,
+            Self::MetadataFirst | Self::MetadataOnly => AlterTableState::InvalidateTableCache,
+        }
+    }
+
+    fn after_metadata(self) -> AlterTableState {
+        match self {
+            Self::MetadataFirst => AlterTableState::SubmitAlterRegionRequests,
+            Self::RegionFirst | Self::MetadataOnly => AlterTableState::InvalidateTableCache,
+        }
     }
 }
 
@@ -562,9 +599,6 @@ pub struct AlterTableData {
     table_info_value: Option<DeserializedValueWithBytes<TableInfoValue>>,
     /// Region distribution for table in case we need to update region options.
     region_distribution: Option<RegionDistribution>,
-    /// Whether this irreversible alter persists metadata before mutating regions.
-    #[serde(default)]
-    alter_regions_after_metadata: bool,
     /// Region locks held by irreversible region-option alters.
     #[serde(default)]
     region_locks: Vec<RegionId>,
@@ -579,7 +613,6 @@ impl AlterTableData {
             column_metadatas: vec![],
             table_info_value: None,
             region_distribution: None,
-            alter_regions_after_metadata: false,
             region_locks,
         }
     }
@@ -596,6 +629,11 @@ impl AlterTableData {
         self.table_info_value
             .as_ref()
             .map(|value| &value.table_info)
+    }
+
+    fn flow(&self) -> AlterTableFlow {
+        // Safety: Checked in `AlterTableProcedure::new`.
+        AlterTableFlow::from_kind(self.task.alter_table.kind.as_ref().unwrap())
     }
 
     #[cfg(test)]
--- a/src/common/meta/src/ddl/alter_table.rs
+++ b/src/common/meta/src/ddl/alter_table.rs
@@ -22,6 +22,7 @@ use api::region::RegionResponse;
 use api::v1::alter_table_expr::Kind;
 use api::v1::{RenameTable, SetTableOptions, UnsetTableOptions};
 use async_trait::async_trait;
+use common_catalog::consts::{METRIC_ENGINE, MITO_ENGINE};
 use common_error::ext::BoxedError;
 use common_procedure::error::{FromJsonSnafu, Result as ProcedureResult, ToJsonSnafu};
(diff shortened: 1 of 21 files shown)


fengjiachun:
@codex review

When you finish, append exactly this marker to the review summary:

fengjiachun:
@codex review

When you finish, append exactly this marker to the review summary:

fengjiachun:
@codex review

When you finish, append exactly this marker to the review summary:

fengjiachun:
@codex review

When you finish, append exactly this marker to the review summary:

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -130,7 +143,38 @@ impl AlterTableProcedure {
...
+                UnsupportedSnafu {
+                    operation: format!("setting skip_wal on {engine} engine tables")
+                }
+            );
+            // Persist the irreversible intent before any region stops writing WAL.
+            let (physical_table_id, physical_table_route) = self
+                .context
+                .table_metadata_manager
+                .table_route_manager()
+                .get_physical_table_route(self.data.table_id())
```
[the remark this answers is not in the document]
fengjiachun: This scenario is explicitly out of scope. The feature is operated only after all Datanodes run the same version; rolling-upgrade capability negotiation is intentionally not part of this change. No code change is planned for this finding.

fengjiachun:
@codex review

When you finish, append exactly this marker to the review summary:

fengjiachun:
@codex review

When you finish, append exactly this marker to the review summary:

src/common/meta/src/ddl_manager.rs
```diff
@@ -405,8 +405,33 @@ impl DdlManager {
                 .await;
         }
 
+        let region_locks = if alter_table_task
+            .alter_table
+            .kind
+            .as_ref()
```
WenyXu, line 413: The lock set is derived before the procedure is submitted, but `on_prepare` and `submit_alter_region_requests` later read the route again. If repartition completes in between, this procedure can hold locks for former regions while altering replacement regions without their locks, allowing a concurrent migration to interleave. Could we bind the route snapshot and lock set, or detect a mismatch and retry?

WenyXu (commented):
I left an inline comment about keeping the skip-WAL region locks bound to the route used for RPC delivery.

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -188,6 +235,14 @@ impl AlterTableProcedure {
...
                 &physical_table_route.region_routes,
                 alter_kind,
+                self.data.alter_regions_after_metadata.then(|| {
+                    self.data
+                        .datanode_capability_fence
+                        .iter()
+                        .map(|node| (node.node_id, node.start_time_ms))
+                        .collect()
+                }),
+                self.data.alter_regions_after_metadata,
```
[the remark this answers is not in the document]
fengjiachun: Fixed in dba01cd758: mixed table-option statements containing `skip_wal=true` are rejected during prepare, before metadata or region changes. Covered by `test_skip_wal_rejects_mixed_table_options`.

src/meta-srv/src/procedure/region_migration.rs
```diff
@@ -572,6 +611,129 @@ impl Context {
...
+    async fn bind_target_skip_wal_capability(&mut self) -> Result<()> {
+        let requires_capability = self.migration_requires_skip_wal_capability().await?;
+        if !requires_capability {
+            self.volatile_ctx.target_pusher_generation = None;
+            return Ok(());
+        }
+
+        let target = self.active_target_node_info().await?.filter(|(node, _)| {
+            node.last_activity_ts >= self.persistent_ctx.capability_fence_after_ms
+        });
```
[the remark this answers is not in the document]
fengjiachun: The capability-negotiation and migration-gating path that introduced this race was removed. The feature now has an explicit same-version deployment precondition, so this thread no longer applies to the current head.

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -240,10 +305,97 @@ impl AlterTableProcedure {
...
+                .supports_datanode_capability_checks()
+        {
+            return Ok(Vec::new());
+        }
+
+        let required_nodes = region_routes
+            .iter()
+            .flat_map(|route| route.leader_peer.iter().chain(route.follower_peers.iter()))
+            .map(|peer| (peer.id, peer.addr.clone()))
+            .collect::<BTreeMap<_, _>>();
```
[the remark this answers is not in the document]
fengjiachun: The capability gate referenced here was removed. With the explicit same-version deployment precondition, the partial capability invariant and this undrop mismatch no longer apply to the current head.

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -130,7 +132,25 @@ impl AlterTableProcedure {
 
         // Safety: Checked in `AlterTableProcedure::new`.
         let alter_kind = self.data.task.alter_table.kind.as_ref().unwrap();
-        if is_metadata_only_alter(alter_kind) {
+        if enables_skip_wal(alter_kind) {
```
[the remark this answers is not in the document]
fengjiachun: Fixed in dba01cd758: logical Metric Engine tables are rejected before metadata update by requiring the requested table ID to equal the physical route table ID. Covered by `test_skip_wal_rejects_logical_table`.

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -130,7 +133,38 @@ impl AlterTableProcedure {
...
+                .get_physical_table_route(self.data.table_id())
+                .await?;
+            ensure!(
+                physical_table_id == self.data.table_id(),
+                UnsupportedSnafu {
+                    operation: "setting skip_wal on logical tables".to_string()
+                }
+            );
+            self.data.region_distribution =
+                Some(region_distribution(&physical_table_route.region_routes));
```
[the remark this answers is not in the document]
fengjiachun: Fixed in dba01cd758: skip-WAL ALTER persists and acquires locks for the routed RegionIds, covered by `test_skip_wal_holds_region_locks`. The separate repartition/snapshot concern raised later remains open in its own thread.

src/common/meta/src/ddl_manager.rs
fengjiachun: Fixed in 61b45bb98f. `on_prepare` now compares the current physical RegionId set with the immutable region locks captured at submission. On mismatch it performs no metadata or region mutation, finishes that attempt, and `DdlManager` resubmits with a fresh route/lock set after the old locks are released. Added a deterministic stale-lock/current-route regression test; all 575 common-meta unit tests and the focused metric-engine physical-region test pass.

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -451,17 +562,25 @@ pub struct AlterTableData {
     table_info_value: Option<DeserializedValueWithBytes<TableInfoValue>>,
     /// Region distribution for table in case we need to update region options.
     region_distribution: Option<RegionDistribution>,
+    /// Whether this irreversible alter persists metadata before mutating regions.
+    #[serde(default)]
+    alter_regions_after_metadata: bool,
```
WenyXu: `alter_regions_after_metadata` is derived entirely from the persisted alter kind, but it is carried as an additional persisted flag and drives transitions in several handlers. Could we move the ordering decision into a small state machine instead? The stored procedure state would remain the existing `AlterTableState`; the flow is derived from `task.alter_table.kind` on each execution.

For example:

```rust
#[derive(Clone, Copy)]
enum AlterTableFlow {
    RegionsThenMetadata,
    MetadataThenRegions,
    MetadataOnly,
}

struct AlterTableStateMachine<'a> {
    state: &'a mut AlterTableState,
    flow: AlterTableFlow,
}

impl AlterTableStateMachine<'_> {
    fn on_prepared(&mut self) {
        *self.state = match self.flow {
            AlterTableFlow::RegionsThenMetadata => {
                AlterTableState::SubmitAlterRegionRequests
            }
            AlterTableFlow::MetadataThenRegions | AlterTableFlow::MetadataOnly => {
                AlterTableState::UpdateMetadata
            }
        };
    }

    fn on_metadata_updated(&mut self) {
        *self.state = match self.flow {
            AlterTableFlow::MetadataThenRegions => {
                AlterTableState::SubmitAlterRegionRequests
            }
            AlterTableFlow::RegionsThenMetadata | AlterTableFlow::MetadataOnly => {
                AlterTableState::InvalidateTableCache
            }
        };
    }

    fn on_regions_altered(&mut self) {
        *self.state = match self.flow {
            AlterTableFlow::RegionsThenMetadata => AlterTableState::UpdateMetadata,
            AlterTableFlow::MetadataThenRegions => {
                AlterTableState::InvalidateTableCache
            }
            AlterTableFlow::MetadataOnly => unreachable!(),
        };
    }
}
```

`skip_wal` would simply be the first alter kind mapped to `MetadataThenRegions`. This keeps the persisted state focused on recovery checkpoints and centralizes the transition rules instead of scattering checks for a derived flag.

src/common/meta/src/ddl/alter_table/executor.rs
```diff
@@ -169,26 +169,41 @@ impl AlterTableExecutor {
         node_manager: &NodeManagerRef,
         region_routes: &[RegionRoute],
         kind: Option<alter_request::Kind>,
+        alter_follower_regions: bool,
```
WenyXu: `alter_follower_regions` looks like a reusable ALTER capability, but this PR only defines all-replica semantics for `skip_wal`; other options still use the leader-only ALTER plus follower sync path. Could we keep this dispatch local to the skip-WAL flow for now, and generalize it once we define which options require follower-side runtime updates?

WenyXu (commented):
I left two inline comments about centralizing ALTER state transitions and avoiding premature generalization of follower-side runtime updates.

src/table/src/metadata.rs
```diff
@@ -401,6 +401,12 @@ impl TableMeta {
                         new_options.extra_options.remove(MAX_ROW_GROUP_ROW_COUNT);
                     }
                 }
+                SetRegionOption::SkipWal => {
+                    new_options.skip_wal = true;
+                    new_options
+                        .extra_options
+                        .insert(SKIP_WAL_KEY.to_string(), true.to_string());
+                }
```
evenyag, line 411: Why do we set another option key in extra_options after setting `new_options.skip_wal`? We already have a skip_wal field.

src/store-api/src/region_request.rs
```diff
@@ -1450,6 +1450,8 @@ pub enum SetRegionOption {
     AutoFlushInterval(Option<Duration>),
     // Modifying the max number of rows in a parquet row group.
     MaxRowGroupRowCount(Option<usize>),
+    // Stops writing new WAL entries. This operation is irreversible.
+    SkipWal,
```
evenyag, line 1458: Do we need to pass a bool for future extension?

src/mito2/src/worker/handle_alter.rs
```diff
@@ -50,7 +50,13 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
         sender: OptionOutputTx,
     ) {
-        let region = match self.regions.writable_non_staging_region(region_id) {
+        let region = match if only_enables_skip_wal(&request.kind) {
+            self.regions
+                .writable_non_staging_region(region_id)
+                .or_else(|_| self.regions.follower_region(region_id))
+        } else {
+            self.regions.writable_non_staging_region(region_id)
+        } {
```
evenyag: This may be risky in the future because the code below may trigger a flush.

Maybe we can implement a new branch for follower regions that can alter options that don't need flush.

src/common/meta/src/ddl_manager.rs
```diff
@@ -405,12 +405,47 @@ impl DdlManager {
...
+                    .get_physical_table_route(table_id)
+                    .await?;
+                route
+                    .region_routes
+                    .iter()
+                    .map(|route| route.region.id)
+                    .collect::<Vec<RegionId>>()
+            } else {
+                vec![]
+            };
```
evenyag: Why do we lock regions here?

evenyag (approved):

src/common/meta/src/ddl_manager.rs
```diff
@@ -405,12 +405,47 @@ impl DdlManager {
...
-        let procedure = AlterTableProcedure::new(table_id, alter_table_task, context)?;
-
-        let procedure_with_id = ProcedureWithId::with_random_id(Box::new(procedure));
+        let lock_regions = alter_table_task
+            .alter_table
+            .kind
+            .as_ref()
+            .is_some_and(only_enables_skip_wal);
+
+        loop {
```
WenyXu, line 417: The route-change retry is unbounded here. Under continuous region migration or repartition, every attempt can finish with RegionRouteChanged, causing this request to create procedures indefinitely while the caller cannot observe or control the retry. 

Could we bound this to a small number of attempts (for example, three) and then return a retryable route-change error for the caller to resubmit?

src/common/meta/src/ddl/alter_table.rs
fengjiachun: Fixed in a77197f370. The persisted alter_regions_after_metadata flag is removed. AlterTableFlow is now derived from the persisted alter kind on every execution, and the after_prepare/after_metadata/after_regions transitions are centralized there. Verified by the common-meta skip_wal tests.

src/common/meta/src/ddl/alter_table/executor.rs
fengjiachun: Fixed in a77197f370. The regular on_alter_regions path is leader-only again, while the all-replica behavior is exposed only through the explicitly named on_alter_skip_wal_regions path.

src/table/src/metadata.rs
fengjiachun: Fixed in a77197f370. TableOptions.skip_wal is now the canonical stored field and skip_wal is removed from extra_options. The TableOptions-to-region-options conversion now emits skip_wal from the typed field, preserving the runtime option propagation without duplicate metadata. All 60 table crate unit tests pass.

src/store-api/src/region_request.rs
fengjiachun: I kept SkipWal as a unit variant intentionally. This API only supports the irreversible false-to-true transition; carrying a bool would imply that false is a valid request. If reversible semantics are introduced later, they should come with an explicit new contract rather than being accepted accidentally here.

src/mito2/src/worker/handle_alter.rs
fengjiachun: Fixed in a77197f370. Follower skip-WAL handling now has an explicit branch that updates the in-memory option and returns before entering the leader ALTER path, so it cannot reach the memtable flush logic. All 12 mito2 skip_wal tests pass, including follower promotion.

src/common/meta/src/ddl_manager.rs
fengjiachun: The region locks serialize this operation with migration while we validate that the captured physical route is still the exact set of regions the procedure will mutate. Without the shared locks, a region could move after the snapshot and miss the irreversible runtime update. I added this rationale as a code comment in a77197f370.

src/common/meta/src/ddl_manager.rs
fengjiachun: Fixed in a77197f370. Route changes are now retried at most three times; if the route is still changing, the request returns a retryable error so the caller regains control.

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -182,21 +243,36 @@ impl AlterTableProcedure {
...
             MultipleResults::PartialRetryable(error) => {
                 // Just returns the error, and wait for the next try.
                 Err(error)
             }
             MultipleResults::PartialNonRetryable(error) => {
+                if flow == AlterTableFlow::MetadataFirst {
+                    return Err(BoxedError::new(error)).context(RetryLaterSnafu {
+                        clean_poisons: true,
+                    });
+                }
```
WenyXu: It's better to document this behavior

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -212,12 +288,19 @@ impl AlterTableProcedure {
...
                 self.handle_alter_region_response(results)?;
                 Ok(Status::executing_with_clean_poisons(true))
             }
             MultipleResults::AllNonRetryable(error) => {
+                if flow == AlterTableFlow::MetadataFirst {
+                    return Err(BoxedError::new(error)).context(RetryLaterSnafu {
+                        clean_poisons: true,
+                    });
+                }
```
WenyXu: Why does it need to retry here?

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -182,21 +243,36 @@ impl AlterTableProcedure {
...
+                )
+                .await
+        } else {
+            self.executor
+                .on_alter_regions(
+                    &self.context.node_manager,
+                    &physical_table_route.region_routes,
+                    alter_kind,
+                )
+                .await
```
WenyXu: nit(non-blocking): Can we consider a role-aware alter interface in a follow-up, rather than keeping this flow-specific dispatch? Leaders need to persist manifest changes, while followers usually only need the corresponding in-memory update, so the operations may need different handling while still being propagated to every replica.

src/store-api/src/region_request.rs
evenyag: There is a potential risk that region options derive Serialize/Deserialize. We may need to ensure compatibility in the future. Although we don't need to serialize the options currently. But we may need to document this to remind us.

src/table/src/requests.rs
```diff
@@ -232,6 +232,9 @@ impl fmt::Display for TableOptions {
         }
 
         for (k, v) in &self.extra_options {
+            if k == SKIP_WAL_KEY {
+                continue;
+            }
```
evenyag: Why do we still need to skip the SKIP_WAL_KEY key?

src/table/src/requests.rs
```diff
@@ -251,9 +254,13 @@ impl From<&TableOptions> for HashMap<String, String> {
         if let Some(ttl_str) = opts.ttl.map(|ttl| ttl.to_string()) {
             let _ = res.insert(TTL_KEY.to_string(), ttl_str);
         }
+        if opts.skip_wal {
+            let _ = res.insert(SKIP_WAL_KEY.to_string(), true.to_string());
+        }
         res.extend(
             opts.extra_options
                 .iter()
+                .filter(|(key, _)| key.as_str() != SKIP_WAL_KEY)
```
evenyag: Why do we need to filter this?

src/table/src/metadata.rs
```diff
@@ -401,6 +401,10 @@ impl TableMeta {
                         new_options.extra_options.remove(MAX_ROW_GROUP_ROW_COUNT);
                     }
                 }
+                SetRegionOption::SkipWal => {
+                    new_options.skip_wal = true;
+                    new_options.extra_options.remove(SKIP_WAL_KEY);
+                }
```
evenyag, line 411: Shouldn't we remove this line?
```suggestion
                SetRegionOption::SkipWal => {
                    new_options.skip_wal = true;
                }
```

src/query/src/sql/show_create_table.rs
```diff
@@ -66,11 +67,18 @@ fn create_sql_options(table_meta: &TableMeta, schema_options: Option<SchemaOptio
     {
         options.insert(TTL_KEY.to_string(), database_ttl);
     };
+    let schema_skip_wal = schema_options
+        .as_ref()
+        .and_then(|options| options.extra_options.get(SKIP_WAL_KEY))
+        .is_some_and(|value| value == "true");
+    if table_opts.skip_wal || schema_skip_wal {
+        options.insert(SKIP_WAL_KEY.to_string(), table_opts.skip_wal.to_string());
+    }
```
evenyag: I think we shouldn't get schema_skip_wal because updating the schema options won't change the skip_wal flag of an existing table. We only consider table_opts.

src/query/src/sql/show_create_table.rs
```diff
@@ -66,11 +67,18 @@ fn create_sql_options(table_meta: &TableMeta, schema_options: Option<SchemaOptio
...
 
     for (k, v) in table_opts
         .extra_options
         .iter()
-        .filter(|(k, _)| k != &FILE_TABLE_META_KEY)
+        .filter(|(k, _)| k != &FILE_TABLE_META_KEY && k != &SKIP_WAL_KEY)
```
evenyag: We don't put SKIP_WAL_KEY in extra_options now. Please check all usage.

src/operator/src/expr_helper.rs
```diff
@@ -215,12 +215,15 @@ pub fn create_to_expr(
...
-        &TableOptions::try_from_iter(create.options.to_str_map())
-            .context(UnrecognizedTableOptionSnafu)?,
+    let raw_table_options = create.options.to_str_map();
+    let mut table_options = HashMap::from(
+        &TableOptions::try_from_iter(&raw_table_options).context(UnrecognizedTableOptionSnafu)?,
     );
+    // Keep an explicitly specified false so it can override a schema-level true.
+    if let Some(skip_wal) = raw_table_options.get(SKIP_WAL_KEY) {
+        table_options.insert(SKIP_WAL_KEY.to_string(), skip_wal.to_string());
+    }
```
evenyag: So the problem is that TableOptions::skip_wal isn't optional, so we don't know whether the user set it.

Should we use a dedicated flag like COMPACTION_OVERRIDE?

src/mito2/src/worker/handle_alter.rs
evenyag: The whole block looks weird. Can we refactor it to make it clear?

src/common/meta/src/ddl_manager.rs
```diff
@@ -405,12 +407,60 @@ impl DdlManager {
...
+            .alter_table
+            .kind
+            .as_ref()
+            .is_some_and(only_enables_skip_wal);
+
+        let mut route_change_retries = 0;
+        loop {
+            // Hold the same physical region locks as migration while validating that
+            // this route snapshot is still the one the procedure will mutate.
+            let region_locks = if lock_regions {
```
evenyag: Is the variable name region_locks confusing? They are not locks.

src/common/meta/src/ddl/alter_table.rs
```diff
@@ -182,21 +243,36 @@ impl AlterTableProcedure {
...
-        let results = self
-            .executor
-            .on_alter_regions(
-                &self.context.node_manager,
-                &physical_table_route.region_routes,
-                alter_kind,
-            )
-            .await;
+        let flow = self.data.flow();
+        let results = if flow == AlterTableFlow::MetadataFirst {
```
evenyag: Can we rewrite this method into a large if-else block that checks `flow == AlterTableFlow::MetadataFirst`?

src/table/src/requests.rs
fengjiachun: The filter is still needed for persisted metadata created by the base implementation, which stored skip_wal in both the typed field and extra_options. The typed field must remain canonical if those values disagree. Added typed-true/legacy-false regression coverage.

src/operator/src/expr_helper.rs
fengjiachun: Implemented a dedicated internal skip_wal.override marker in 32eb2ce9a3. It records an explicitly specified table-level false without duplicating SKIP_WAL_KEY and preserves the schema-level override behavior.

src/common/meta/src/ddl/alter_table.rs
fengjiachun: Rewritten in 32eb2ce9a3. The MetadataFirst path is now one explicit top-level branch with its own result handling; the regular ALTER path follows afterward without repeated flow checks.

src/common/meta/src/ddl/alter_table.rs
fengjiachun: The retry is required because metadata has already committed skip_wal=true. Aborting on a replica error would leave metadata and runtime state permanently inconsistent; the reason is now documented next to the branch.

src/table/src/requests.rs
fengjiachun: The filter prevents a legacy extra_options["skip_wal"] value from overwriting the canonical typed field during map conversion. Base-created table metadata can contain both representations; the regression now verifies typed true wins over legacy false.

src/mito2/src/worker/handle_alter.rs
fengjiachun: Refactored in 32eb2ce9a3. The writable-region lookup is now a straightforward match with an explicit skip-WAL-only follower fallback and separate error returns.

src/common/meta/src/ddl/alter_table.rs
fengjiachun: Documented the metadata-first behavior in a213226b71: once metadata enables skip-WAL, replica-side failures must keep retrying until all replicas converge.

src/common/meta/src/ddl_manager.rs
fengjiachun: Renamed the local value to region_ids_to_lock in 32eb2ce9a3 to make clear that it contains RegionIds used to construct procedure lock keys.

src/common/meta/src/ddl/alter_table.rs
fengjiachun: Agreed as a follow-up. This PR keeps the all-replica dispatch explicitly skip-WAL-specific; a general role-aware ALTER interface should be designed separately.

src/query/src/sql/show_create_table.rs
fengjiachun: New metadata no longer stores SKIP_WAL_KEY in extra_options. The remaining filters only protect metadata persisted by the base implementation, which did store the duplicate key; typed skip_wal is now canonical at every output boundary.

src/table/src/metadata.rs
fengjiachun: Kept the cleanup intentionally. The base parser persisted skip_wal in extra_options as well as the typed field, so ALTER must remove that old entry when setting the canonical typed value to true. Added a legacy-metadata regression.

src/store-api/src/region_request.rs
fengjiachun: Documented that the serde representation of SetRegionOption must remain backward compatible. SkipWal stays a unit variant because this API intentionally supports only false-to-true.

src/query/src/sql/show_create_table.rs
fengjiachun: Fixed in 32eb2ce9a3. SHOW CREATE no longer reads the current schema skip_wal value; it uses the typed table value plus an internal marker for an explicitly specified table-level false.

src/table/src/metadata.rs
```diff
@@ -401,6 +401,12 @@ impl TableMeta {
                         new_options.extra_options.remove(MAX_ROW_GROUP_ROW_COUNT);
                     }
                 }
+                SetRegionOption::SkipWal => {
+                    new_options.skip_wal = true;
+                    // Older table metadata also stored skip_wal in extra_options.
+                    // Remove it so the typed field is canonical after ALTER.
+                    new_options.extra_options.remove(SKIP_WAL_KEY);
```
evenyag: The old table metadata shouldn't contain skip_wal_key.

fengjiachun: Fixed in 671a884c85. The override-marker approach was unnecessary. One detail is that the base TableOptions parser already retained an explicitly supplied skip_wal in extra_options because it filtered only write_buffer_size and ttl. This change restores that behavior, uses the existing key to preserve table-level explicitness, and keeps the typed value and extra option synchronized on ALTER. It removes skip_wal.override and the related special-case filters.

evenyag (approved):

PR #8736, opened 2026-08-04
feat(runtime): add weighted workload scheduler

discord9:

- [catio PR #1](https://github.com/GreptimeTeam/catio/pull/1) — weighted, work-conserving Tokio scheduler and soft CPU entitlements (merged)
- [catio PR #2](https://github.com/GreptimeTeam/catio/pull/2) — scheduler panic-test teardown fix (merged)
- [catio PR #3](https://github.com/GreptimeTeam/catio/pull/3) — atomic runtime class-weight updates (merged)

GreptimeDB pins catio PR #3's merge commit, `c75be64e7a58f620f6340c24b5b8d76f131af6aa`.

This PR adds an experimental, opt-in, process-local scheduler for sharing Tokio CPU poll execution time between query and write workloads. Its weights are relative, work-conserving entitlements under contention, not hard CPU caps: an active class may borrow all idle capacity.

### Runtime integration

- Adds `[runtime.experimental_workload_scheduler]` with:
  - `enable = false` by default;
  - positive `query_weight` and `write_weight` values, rejected during TOML/JSON deserialization when zero.
- Uses two fixed catio task classes for query and write submissions.
- Initializes runtimes by process role:
  - frontend, metasrv, and flownode use the global runtime without constructing a scheduler;
  - standalone shares its global runtime and constructs a scheduler with an internal admission bound equal to `global_rt_size`;
  - datanode uses dedicated query/ingest runtimes and constructs a scheduler with an internal checked bound of `query_rt_size + ingest_rt_size`.
- Keeps the admission bound internal; it is not configuration, HTTP status, or a Prometheus metric.
- Always constructs the scheduler for standalone and datanode, then applies the configured enabled state. This permits later runtime enablement even when startup configuration is disabled.
- Implements strict dynamic disable semantics through catio:
  - after disable returns, new query/write submissions spawn directly on Tokio without a `Scheduled` wrapper, admission, proxy waker, or scheduler accounting;
  - already managed tasks drain safely and remain on the sticky bypass path;
  - re-enable affects only later submissions.
- Updates query/write runtime weights through one atomic catio batch operation, so readers and dispatch never observe a sequentially mixed ratio.

### Internal debug API

The internal/full HTTP listener exposes:

- `GET /debug/workload_scheduler`
- `POST /debug/workload_scheduler/enabled` with a bare JSON boolean body
- `POST /debug/workload_scheduler/weights` with both required positive weights:

  ```json
  {"query": 3, "write": 7}
  ```

The GET response is intentionally small:

```json
{
  "enabled": true,
  "query": {"weight": 3, "polls": 42},
  "write": {"weight": 7, "polls": 81}
}
```

The debug routes remain excluded from the API-only listener. Detailed operational statistics are exported through Prometheus rather than duplicated in the JSON surface.

### Metrics

The existing runtime collector exports:

- gauges:
  - `greptime_workload_scheduler_enabled`
  - `greptime_workload_scheduler_active_polls`
  - `greptime_workload_scheduler_weight{workload}`
  - `greptime_workload_scheduler_queued_tasks{workload}`
- monotonic counters:
  - `greptime_workload_scheduler_tasks_total{workload}`
  - `greptime_workload_scheduler_wakes_total{workload}`
  - `greptime_workload_scheduler_polls_total{workload}`
  - `greptime_workload_scheduler_completed_total{workload}`
  - `greptime_workload_scheduler_cancelled_total{workload}`
  - `greptime_workload_scheduler_admission_wait_seconds_total{workload}`
  - `greptime_workload_scheduler_exec_duration_seconds_total{workload}`

Catio supplies absolute snapshots, so the collector serializes scrapes and advances Prometheus counters only by the snapshot delta. A focused test verifies metric names, types, labels, stable repeated scrapes, and later query/write deltas.

### Distributed integration coverage

A focused shell test reuses the existing Basic sqlness binary artifact instead of compiling a separate integration-test job. It starts real metasrv, scheduler-enabled datanode, and frontend processes, then verifies:

- SQL query and write traffic through the frontend;
- datanode scheduler status and atomic `1:1` to `3:7` weight update;
- representative Prometheus metrics;
- query/write poll growth while enabled;
- successful SQL with unchanged scheduler polls after runtime disable.

The previous large Rust process harness and dedicated CI recompilation job are removed.

### Limitations and compatibility

- The scheduler is experimental, disabled by default, and process-local.
- This phase has no tenant IDs, hierarchy, cluster quota, or hard CPU cap.
- Scheduling is cooperative at future poll boundaries and cannot preempt unbounded synchronous work that does not yield.
- GreptimeDB internally uses `sample_every_polls = 16` and does not expose it as configuration. In an isolated zero-work poll benchmark, this put clock overhead below measurement noise with about 4% fairness erosion; this is not a GreptimeDB workload benchmark.
- No schema, persisted metadata, wire format, or data-format change is introduced.

### Verification

- `cargo fmt --check`
- `cargo test -p common-runtime` — 22 passed, 0 failed
- `cargo check -p cmd -p servers`
- `make config-docs`
- `bash -n .github/scripts/workload-scheduler-e2e.sh`
- workflow YAML parse
- real local distributed scheduler E2E — passed with metasrv + datanode + frontend
- `git diff --check`
- pre-commit typos, format, clippy, and license-header hooks — passed

## PR Checklist


config/config.md
```diff
@@ -26,6 +26,11 @@
 | `runtime.global_rt_size` | Integer | `8` | The number of threads to execute the runtime for global read operations. |
 | `runtime.compact_rt_size` | Integer | `4` | The number of threads to execute compact operations. |
 | `runtime.compact_rt_max_blocking_threads` | Integer | `4` | The maximum number of blocking threads for compact operations.<br/>Defaults to max(num_cpus / 2, 1). |
+| `runtime.experimental_workload_scheduler` | -- | -- | Experimental weighted, work-conserving query/write task scheduler. |
```
killme2008, line 29: Do we need to briefly describe when it should be enabled?

src/common/runtime/src/global.rs
```diff
@@ -14,19 +14,52 @@
...
+#[serde(default)]
+pub struct WorkloadSchedulerOptions {
+    /// Enables policy-controlled query and write task spawning.
+    pub enable: bool,
+    /// Maximum polls admitted to Tokio at once. Zero uses four times
+    /// `global_rt_size` to keep worker queues fed without making them
+    /// effectively unbounded.
+    pub max_concurrent_polls: usize,
+    /// Relative share for query polls while writes are also backlogged.
+    pub query_weight: u32,
```
killme2008: Do we need to add a `validate `function to ensure that `query_weight` and `write_weight` are both positive and add up to 10?

src/common/runtime/src/global.rs
```diff
@@ -209,9 +287,50 @@ pub fn init_global_runtimes(options: &RuntimeOptions) {
...
+        c.workload_scheduler = create_workload_scheduler(options);
     });
 }
 
+fn create_workload_scheduler(options: &RuntimeOptions) -> Option<Scheduler> {
+    let scheduler_options = &options.experimental_workload_scheduler;
+    if !scheduler_options.enable {
+        return None;
+    }
+    if scheduler_options.query_weight == 0 || scheduler_options.write_weight == 0 {
```
killme2008: Got it. But we can introduce a `validate` function  for `scheduler_options` instead including validate `max_concurrent_polls` below.

src/servers/src/http.rs
```diff
@@ -1072,6 +1073,18 @@ impl HttpServer {
             Router::new()
                 // handler for changing log level dynamically
                 .route("/log_level", routing::post(dyn_log::dyn_log_handler))
+                .route(
```
killme2008, line 1081: We need to add an API to dynamically enable or disable this scheduler.

killme2008 (approved):
LGTM almostly!

Good job

config/config.md
discord9: Added a brief activation guideline: enable it when concurrent query and write workloads need weighted runtime sharing; it remains disabled by default. The standalone/datanode examples and generated config reference are synchronized in 9a06178872d.

src/servers/src/http.rs
discord9: Implemented dynamic control under the internal debug routes: `POST /debug/workload_scheduler/enabled` accepts a bare JSON boolean, and `POST /debug/workload_scheduler/weights` atomically updates the query/write pair. `GET /debug/workload_scheduler` reports enabled state and current weights. Disable uses the raw Tokio bypass for new submissions.

config/config.md
discord9: Follow-up: I replaced the vague wording with a concrete operator scenario in 351bd8ae00a: enable it when concurrent queries and writes interfere—for example, when long-running queries increase ingestion latency. The docs also explain that weights determine their relative runtime shares only while both are backlogged.

src/common/runtime/src/global.rs
discord9: Addressed in the current implementation: both weights use `NonZeroU32`, so zero is rejected during config deserialization and by the runtime weights DTO. We intentionally do not require them to sum to 10 because they are relative shares (for example, 1:4 and 2:8 are equivalent), not percentages.

src/common/runtime/src/global.rs
discord9: The current design no longer exposes `max_concurrent_polls` as user configuration: it is derived from the runtime worker counts (standalone global workers; datanode checked query+ingest workers) and asserted nonzero at scheduler construction. User inputs are represented by `NonZeroU32`/`NonZeroUsize`, so a separate validation layer would duplicate type-level validation.

PR #8739, opened 2026-08-04
fix(frontend): remove gRPC DDL panics for DropView and non-timestamp time index

discord9:

Fixes two panics in the direct gRPC DDL ingress (which bypasses the SQL parser):

**QX-152 — gRPC DDL DropView hit `todo!()`.** `DdlExpr::DropView` dispatched to `todo!("implemented in the following PR")` in `src/frontend/src/instance/grpc.rs`, so any authorized gRPC client (or unauthenticated deployment) could cause a repeatable request/task panic. It is now wired to the real drop-view implementation (`StatementExecutor::drop_view`, visibility widened from `pub(crate)` to `pub`), returning a structured error (e.g. `TableNotFound` for a missing view) instead of panicking.

**QX-153 — non-timestamp time index via direct gRPC DDL panicked.** A crafted `CreateTableExpr` whose `time_index` column is not a timestamp reached `Schema::new`'s unwrap (`ddl.rs:2346` → `schema.rs:114-119`). Fixed in two layers: `create_table_info` now uses `Schema::try_new` with `ConvertSchemaSnafu` context (`InvalidArguments`), and the direct gRPC `CreateTable` arm validates the request via `validate_create_expr` — which now also checks that the time-index column type is a timestamp — before any catalog work.

SQL/HTTP paths were already protected by the parser (e.g. `BIGINT TIME INDEX` is rejected at parse time); unchanged.

Tests: `qx_152_drop_view_via_grpc_ddl_returns_error_not_panic`, `qx_153_create_table_with_non_timestamp_time_index_via_grpc_returns_error` (asserts `InvalidArguments`), and `test_create_table_info_rejects_non_timestamp_time_index`.

## PR Checklist

### DIFF
--- a/src/frontend/src/instance.rs
+++ b/src/frontend/src/instance.rs
@@ -3267,4 +3267,89 @@ mod tests {
             assert_eq!(result.is_ok(), is_ok);
         }
     }
+
+    /// A `DropView` DDL sent through the direct gRPC ingress must return an error
+    /// (e.g. table not found) instead of panicking on `todo!()`.
+    #[tokio::test]
+    async fn qx_152_drop_view_via_grpc_ddl_returns_error_not_panic() -> TestResult<()> {
+        let instance =
+            test_instance_with_tables(test_table(1024, "source")?, test_table(1025, "target")?)
+                .await?;
+
+        let request = api::v1::greptime_request::Request::Ddl(api::v1::DdlRequest {
+            expr: Some(api::v1::ddl_request::Expr::DropView(
+                api::v1::DropViewExpr {
+                    catalog_name: String::new(),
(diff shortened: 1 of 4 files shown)


fengjiachun (approved):
LGTM. Verified the reused `drop_view` already guards non-view targets and `IF EXISTS`, the new time-index type check matches the SQL parser's rule, and both error variants map to `InvalidArguments`. Nice test harness addition with `MockProcedureExecutor`.

Optional follow-up: consider hoisting `validate_create_expr` into `create_table_inner` so the invariant holds for all callers (prom_store/dashboard paths) rather than each ingress.

PR #8742, opened 2026-08-04
chore(deps): bump opendal to 0.58.1

Xuanwo:

Bump Apache OpenDAL and the matching `object_store_opendal` integration so GreptimeDB stays on the current stable OpenDAL line.

- `opendal` `0.57` → `0.58` (lockfile: **0.58.1**)
- workspace `object_store_opendal` `0.57` → `0.58` (lockfile: **0.58.0**, latest matching line)

OpenDAL 0.58 changed operator construction and out-of-tree service/layer APIs. This PR only adjusts the dependency pins and the necessary fallout:

- `Operator::new` now returns a finished operator; remove `.finish()` at construction sites
- Replace custom HTTP client injection (`HttpClientLayer` / `raw::HttpClient`) with `OperationContext` + `HttpTransporter` (`ReqwestTransport`) so pool/timeout/SSL settings still apply
- Port `SecureFsBackend` and the testing `MockLayer` from the removed `Access` / `LayeredAccess` pipeline to `Service` + `Layer::apply_service`
- Adapt SecureFs reader/writer/lister to sync OIO factories and `StreamRead`

No intentional behavior changes beyond the dependency upgrade and API adaptation.

## PR Checklist

### DIFF
--- a/src/object-store/Cargo.toml
+++ b/src/object-store/Cargo.toml
@@ -27,7 +27,7 @@ derive_builder = { workspace = true, optional = true }
 futures.workspace = true
 humantime-serde.workspace = true
 lazy_static.workspace = true
-opendal = { version = "0.57", features = [
+opendal = { version = "0.58", features = [
     "layers-tracing",
     "layers-prometheus",
     "services-azblob",
@@ -37,6 +37,7 @@ opendal = { version = "0.57", features = [
     "services-oss",
     "services-s3",
 ] }
+opendal-http-transport-reqwest = "0.58"
 prometheus.workspace = true
 reqwest.workspace = true
 serde.workspace = true
--- a/src/object-store/src/secure_fs.rs
+++ b/src/object-store/src/secure_fs.rs
@@ -23,7 +23,8 @@ use cap_std::ambient_authority;
 use cap_std::fs::{Dir, DirEntry, OpenOptions, ReadDir};
 use opendal::raw::*;
 use opendal::{
-    Buffer, Capability, EntryMode, Error, ErrorKind, Metadata, Operator, OperatorBuilder, Result,
+    Buffer, BytesRange, Capability, EntryMode, Error, ErrorKind, Metadata, OperationContext,
+    Operator, Result,
 };
 use tokio::io::{AsyncReadExt, AsyncSeekExt, AsyncWriteExt};
 
@@ -102,7 +103,10 @@ impl SecureFsRoot {
 
     /// Builds an OpenDAL operator confined to this root.
     pub fn build_operator(&self) -> Operator {
-        OperatorBuilder::new(SecureFsBackend::new(self.clone())).finish()
+        Operator::from_parts(
+            OperationContext::default(),
+            Arc::new(SecureFsBackend::new(self.clone())) as Servicer,
+        )
     }
 }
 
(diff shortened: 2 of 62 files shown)


WenyXu:
@codex fix ci

WenyXu:
@codex review

sunng87:
hi @Xuanwo , we have a plan to contribute the secure fs backend to upstream. the goal of this backend is to prevent caller from reading or writing any files from the file system. We received multiple security reports about the previous implementation and we consider this can be helpful to a wide range of opendal users. For more information: https://github.com/GreptimeTeam/greptimedb/pull/8708 

If it sounds good to you, we can start to work on the upstream.

Xuanwo:
> hi @Xuanwo , we have a plan to contribute the secure fs backend to upstream. the goal of this backend is to prevent caller from reading or writing any files from the file system. We received multiple security reports about the previous implementation and we consider this can be helpful to a wide range of opendal users. For more information: #8708
> 
> If it sounds good to you, we can start to work on the upstream.

Seems like a cool idea. My intuition tells me that building a secure layer makes more sense.

Xuanwo:
btw, any ideas on this PR?

src/object-store/Cargo.toml
```diff
@@ -37,6 +37,7 @@ opendal = { version = "0.57", features = [
     "services-oss",
     "services-s3",
 ] }
+opendal-http-transport-reqwest = "0.58"
```
WenyXu: The new transport crate is declared with a crate-local version literal, but repository policy requires new dependencies to be added to the root `[workspace.dependencies]`. Keeping it local creates a separate version-management point from the rest of the OpenDAL stack. Could we declare it at the workspace root and use `workspace = true` here?

src/object-store/src/secure_fs.rs
```diff
@@ -102,7 +103,10 @@ impl SecureFsRoot {
 
     /// Builds an OpenDAL operator confined to this root.
     pub fn build_operator(&self) -> Operator {
-        OperatorBuilder::new(SecureFsBackend::new(self.clone())).finish()
+        Operator::from_parts(
```
WenyXu, line 107: `Operator::from_parts` starts without OpenDAL’s default layers, whereas the previous `OperatorBuilder::new(...).finish()` installed `CorrectnessCheckLayer` and the other standard wrappers. `SecureFsBackend` neither advertises nor implements options such as `if_match`, so those preconditions can now be silently ignored and a write may overwrite existing content. Could we restore the default layer stack after constructing the custom service?

src/object-store/src/secure_fs.rs
```diff
@@ -230,139 +243,118 @@ impl Access for SecureFsBackend {
...
+impl oio::StreamRead for SecureFsReader {
+    async fn open(&self, range: BytesRange) -> Result<(RpRead, Box<dyn oio::ReadStreamDyn>)> {
+        let path = self.path.clone();
+        let root = self.root.clone();
+        let file = common_runtime::spawn_blocking_global(move || root.dir.open(path))
+            .await
+            .map_err(new_task_join_error)?
+            .map_err(new_std_io_error)?;
+        let mut file = tokio::fs::File::from_std(file.into_std());
+        if range.offset() != 0 {
```
WenyXu, line 339: `BytesRange::Suffix(n)` has an offset of zero and a bounded size of `n`. This code therefore skips the seek and reads the first `n` bytes, not the final `n` bytes requested by a suffix range. Could we resolve suffix ranges using the file length before seeking, and add a regression test that reads the final bytes of known content?

src/object-store/src/secure_fs.rs
```diff
@@ -102,7 +103,10 @@ impl SecureFsRoot {
 
     /// Builds an OpenDAL operator confined to this root.
     pub fn build_operator(&self) -> Operator {
-        OperatorBuilder::new(SecureFsBackend::new(self.clone())).finish()
+        Operator::from_parts(
+            OperationContext::default(),
+            Arc::new(SecureFsBackend::new(self.clone())) as Servicer,
+        )
```
WenyXu, line 110: OpenDAL’s built-in Fs does not implement suffix reads natively; it relies on the default `SimulateLayer` to stat the object and normalize `BytesRange::Suffix` into an offset range. This custom operator is created with `from_parts`, so that layer is absent and `SecureFsReader` receives a suffix with offset `0`, returning the first `N` bytes. Could we append `SimulateLayer::default()` here and add a suffix-range regression test?

src/object-store/src/secure_fs.rs
WenyXu: Fixed in 09460467d0 by adding `SimulateLayer` to `build_operator()`. Suffix ranges are now resolved before reaching the backend, so the reader does not need native suffix support. The Operator-level regression test returned `012` before the fix and now correctly returns `789`.

src/object-store/src/secure_fs.rs
```diff
@@ -230,139 +243,118 @@ impl Access for SecureFsBackend {
...
+    async fn open(&self, range: BytesRange) -> Result<(RpRead, Box<dyn oio::ReadStreamDyn>)> {
+        let path = self.path.clone();
+        let root = self.root.clone();
+        let file = common_runtime::spawn_blocking_global(move || root.dir.open(path))
+            .await
+            .map_err(new_task_join_error)?
+            .map_err(new_std_io_error)?;
+        let mut file = tokio::fs::File::from_std(file.into_std());
+        if range.offset() != 0 {
+            file.seek(io::SeekFrom::Start(range.offset()))
```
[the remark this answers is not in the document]
WenyXu: Fixed in 09460467d0 by adding `SimulateLayer` to `build_operator()`. Suffix ranges are now resolved before reaching the backend, so the reader does not need native suffix support. The Operator-level regression test returned `012` before the fix and now correctly returns `789`.

src/object-store/src/secure_fs.rs
WenyXu: Added `SimulateLayer::default()` and an Operator-level suffix regression test in 09460467d0. All 8 `secure_fs` tests and object-store Clippy checks pass.

src/object-store/src/secure_fs.rs
WenyXu: This is still outstanding. 09460467d0 restores `SimulateLayer` only, not the correctness checks. The other default layers are private in OpenDAL 0.58.1, so they cannot be appended directly. We need to construct the backend through `Operator::new()` with a custom builder to restore the full stack.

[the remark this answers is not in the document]
WenyXu: This is still outstanding. 09460467d0 restores `SimulateLayer` only, not the correctness checks. The other default layers are private in OpenDAL 0.58.1, so they cannot be appended directly. We need to construct the backend through `Operator::new()` with a custom builder to restore the full stack.

src/object-store/src/secure_fs.rs
```diff
@@ -102,7 +104,11 @@ impl SecureFsRoot {
 
     /// Builds an OpenDAL operator confined to this root.
     pub fn build_operator(&self) -> Operator {
-        OperatorBuilder::new(SecureFsBackend::new(self.clone())).finish()
+        Operator::from_parts(
```
fengjiachun, line 107: This still bypasses OpenDAL’s default correctness checks. For example, `write_with("file", data).if_match("etag")` previously returned `Unsupported` on SecureFs; now the backend ignores the condition and overwrites the file. Adding `SimulateLayer` fixes suffix reads but does not restore those checks. Could we construct this through `Operator::new()` with a custom builder to restore the default layers, and add a regression test for unsupported conditional writes?

WenyXu: I initially scoped the fix to suffix reads because our current SecureFs call sites do not use if_match, and CorrectnessCheckLayer is private in OpenDAL 0.58.1, so it cannot be appended directly. That still leaves a regression in the Operator contract. Constructing it through Operator::new() with a custom builder and adding an unsupported conditional-write regression test would address that.

fengjiachun (approved):
Rest LGTM

PR #8752, opened 2026-08-05
fix(object-store): fix unused import on Windows after #8735

### LINKED ISSUE #8731: Workflow run 'Nightly CI' failed
@GreptimeTeam/db-approver
New failure: https://github.com/GreptimeTeam/greptimedb/actions/runs/30862258379

discord9:

Fixes nightly CI failure tracked in #8731; follows up #8735.

#8735 gated the whole `test_lister_skips_entry_removed_during_iteration` test with `#[cfg(not(windows))]`. On Windows that leaves `read_list_entry` unused in the test module, so `cargo test --no-run` fails with `-D warnings`:

```
error: unused import: `read_list_entry`
 --> src/object-store/src/secure_fs.rs:546:65
```

PR CI does not catch this because Windows jobs only run in the nightly workflow (which failed: nightly run 695).

This PR makes the test portable instead of excluding it: it keeps the cross-platform assertions (call succeeds, path is `"removed"`) on all platforms and gates only the Unix-specific strong assertion (`result.is_none()`), which also keeps the `read_list_entry` reference alive on Windows.

Verified on Linux: `cargo nextest run -p object-store --lib -E 'test(test_lister_skips_entry_removed_during_iteration)'` → 1 passed.

## PR Checklist

- [x] I have written the necessary rustdoc comments. (comment updated to describe the portable/Unix-specific split)
- [x] I have added the necessary unit tests and integration tests. (verified on Linux)

### DIFF
--- a/src/object-store/src/secure_fs.rs
+++ b/src/object-store/src/secure_fs.rs
@@ -625,11 +625,11 @@ mod tests {
     // On Windows, `std::fs::DirEntry` is a snapshot taken by
     // `FindFirstFileW`: `file_type()` and `metadata()` keep returning the
     // cached data even after the file is removed, so `read_list_entry`
-    // cannot observe the deletion and returns `Some` instead of `None`.
-    // This test only applies to platforms where metadata is fetched from
-    // the live filesystem (e.g. Unix `lstat` returns `ENOENT` after
-    // removal, which `read_list_entry` turns into `None`).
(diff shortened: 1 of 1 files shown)


sunng87:
there is a sqlness failure https://github.com/GreptimeTeam/greptimedb/actions/runs/30959961177/job/92161502056

evenyag (approved):

PR #8754, opened 2026-08-05
fix: support Utf8View labels in Prometheus response

### LINKED ISSUE #8732: Prometheus label_replace query panics when serializing Utf8View labels
### What type of bug is this?

Unexpected error

Query Engine

### Minimal reproduce step

1. Start GreptimeDB standalone:

     ```bash
     cargo run -- standalone start

  2. Create a table containing one label and one sample:

     curl -sS -X POST 'http://127.0.0.1:4000/v1/sql' \
       -H 'x-greptime-db-name: public' \
       --data-urlencode 'sql=CREATE TABLE label_replace_repro (
         ts TIMESTAMP TIME INDEX,
         host STRING PRIMARY KEY,
         value DOUBLE
       )'

     curl -sS -X POST 'http://127.0.0.1:4000/v1/sql' \
       -H 'x-greptime-db-name: public' \
       --data-urlencode "sql=INSERT INTO label_replace_repro
         VALUES ('2024-01-01 00:00:00', 'server-01', 1.0)"

  3. Execute a normal PromQL label_replace query through the Prometheus HTTP API:

     curl -v -G 'http://127.0.0.1:4000/v1/prometheus/api/v1/query' \
       -H 'x-greptime-db-name: public' \
       --data-urlencode 'query=label_replace(label_replace_repro, "host_copy", "$1", "host", "(.*)")' \
       --data-urlencode 'time=2024-01-01T00:00:00Z'

  The query execution produces the new label as an Arrow Utf8View array. While building the Prometheus JSON response, record_batches_to_data unconditionally downcasts every label column to StringArray/Utf8:

  batch.column(*i).as_string::<i32>()

  This causes an Arrow downcast panic for the Utf8View label. The client receives an empty response, commonly reported by curl as:

  curl: (52) Empty reply from server



promql return a valid response

### What did you see instead?

panic

all

master

```bash

```

evenyag:

Fixes #8732.

PromQL functions such as `label_replace` can produce Arrow `Utf8View` label columns. The Prometheus JSON response conversion assumed every label column was a regular `Utf8` array and downcast it with `as_string::<i32>()`, which panicked and caused the HTTP connection to close without a response.

This change reads label values through the generic Arrow string-array helper, supporting `Utf8`, `LargeUtf8`, and `Utf8View` while preserving null-label handling.

It also adds an isolated HTTP integration test that executes `label_replace`, verifies a successful Prometheus vector response, and checks the generated label.

This does not change the API, schema, persisted data, or wire format.

## PR Checklist

- [x] I have written the necessary rustdoc comments. (No public API was added.)

## Validation

- `cargo nextest run -p tests-integration integration_http_file_test::test_prometheus_label_replace_response`
- `cargo nextest run -p servers --features testing` (453 passed)
- `make check`
- `make fmt`

### DIFF
--- a/src/servers/src/http/result/prometheus_resp.rs
+++ b/src/servers/src/http/result/prometheus_resp.rs
@@ -30,6 +30,7 @@ use common_query::native_histogram::{
 use common_query::prometheus::{format_prometheus_float, is_prometheus_stale_nan};
 use common_query::{Output, OutputData};
 use common_recordbatch::RecordBatches;
+use datatypes::arrow_array::string_array_value_at_index;
 use datatypes::prelude::ConcreteDataType;
 use indexmap::IndexMap;
 use promql_parser::label::METRIC_NAME;
@@ -281,7 +282,7 @@ impl PrometheusJsonResponse {
             // prepare things...
             let tag_columns = tag_column_indices
                 .iter()
-                .map(|i| batch.column(*i).as_string::<i32>())
+                .map(|i| batch.column(*i))
                 .collect::<Vec<_>>();
             let tag_names = tag_column_indices
                 .iter()
@@ -344,9 +345,8 @@ impl PrometheusJsonResponse {
                     tags.push((METRIC_NAME, metric_name.as_str()));
                 }
                 for (tag_column, tag_name) in tag_columns.iter().zip(tag_names.iter()) {
-                    // TODO(ruihang): add test for NULL tag
-                    if tag_column.is_valid(row_index) {
-                        tags.push((tag_name, tag_column.value(row_index)));
+                    if let Some(tag_value) = string_array_value_at_index(tag_column, row_index) {
+                        tags.push((tag_name, tag_value));
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


src/servers/src/http/result/prometheus_resp.rs
```diff
@@ -344,9 +345,8 @@ impl PrometheusJsonResponse {
                     tags.push((METRIC_NAME, metric_name.as_str()));
                 }
                 for (tag_column, tag_name) in tag_columns.iter().zip(tag_names.iter()) {
-                    // TODO(ruihang): add test for NULL tag
-                    if tag_column.is_valid(row_index) {
-                        tags.push((tag_name, tag_column.value(row_index)));
+                    if let Some(tag_value) = string_array_value_at_index(tag_column, row_index) {
```
qazxcdswe123, line 348: I have the same fix but I worry that this may change the semantic

evenyag: Arrow has multiple different physical representations for string arrays. We can always treat the value as a string logically.

fengjiachun (approved):
LGTM.

PR #8758, opened 2026-08-05
feat(promql): define native histogram semantics

shuiyisong:

This is the 1st PR of (hopefully) the final 6 PRs of native histogram.

This pull request adds Prometheus-compatible semantics for native histogram samples across PromQL selection, arithmetic, comparison, and counter range functions. It handles histogram stale markers, layout reconciliation, sparse-layout equality, start-timestamp synthetic zeros for `rate()` and `increase()`, and overflow-safe bucket bounds. It also documents the storage, protocol, and query contract without changing the persisted Struct, protobuf dependencies, or public configuration.

## PR Checklist

### DIFF
--- a/docs/rfcs/2026-08-04-native-histograms.md
+++ b/docs/rfcs/2026-08-04-native-histograms.md
@@ -153,7 +153,7 @@ The rejected alternative is Prometheus's direct-sum/Kahan-compensation design
 with an overflow-triggered mode change. It offers closer rounding parity but
 requires additional state and substantially more merge logic.
 
-## Equality and changes
+## Equality, changes, and resets
 
 Equality follows Prometheus's represented-layout semantics. Histograms compare
 equal only when their represented bucket-index sequences and bucket value bit
@@ -165,6 +165,9 @@ and reset hints and start timestamps remain excluded.
 Custom bucket bounds use ordinary floating-point equality rather than bitwise
 equality, matching Prometheus's `CustomBucketBoundsMatch`. Histogram payload
 values continue to use bitwise equality so NaN payloads behave deterministically.
+
+`resets()` counts ordinary counter resets and, unlike Prometheus, also counts
+each transition between gauge and non-gauge native histograms as one reset.
 
 ## Annotations
 
--- a/docs/rfcs/2026-08-04-native-histograms.md
+++ b/docs/rfcs/2026-08-04-native-histograms.md
@@ -1,0 +1,225 @@
+---
+Feature Name: Native Histogram Support and Compatibility Decisions
+Tracking Issue: TBD
+Date: 2026-08-04
+Author: TBD
+---
+
+# Summary
+
+GreptimeDB stores a Prometheus native histogram in one Struct-valued field and
+evaluates it as a first-class PromQL sample. This document records the supported
+protocols, the storage invariant, and the compatibility decisions needed to make
+that path predictable.
+
+Native histograms are experimental. Prometheus Remote Write 2.0 is the only
+supported native-histogram ingestion protocol. Remote Write 1.0 histogram
+payloads are rejected instead of being acknowledged and dropped.
+Native-histogram Remote Read is deferred; the existing Remote Read path
+continues to return scalar samples only.
+
+# Goals
+
+1. Prevent silent native-histogram data loss at protocol boundaries.
+2. Keep one stable persisted representation for integer, float, exponential,
+   and custom-bucket native histograms.
+3. Match Prometheus query behavior where it is observable and practical.
+4. State intentional limitations explicitly so incomplete behavior is not
+   mistaken for support.
+
+# Non-Goals
+
+- Supporting native histograms in Remote Write 1.0.
+- Returning native histograms through Prometheus Remote Read.
+- Persisting Remote Write metric metadata.
+- Ingesting OTLP exponential histograms.
+- Persisting exemplars.
+- Removing mixed float/histogram handling from PromQL expressions.
+- Propagating PromQL annotations produced on datanodes back to the frontend.
+
+# Data Model and Invariants
+
+Each native histogram is stored in the configured native-histogram field
+(`greptime_native_histogram` by default), a Struct whose children preserve the
+wire-level schema, zero threshold, sum, reset hint, start timestamp, custom
+bounds, spans, and either the integer or float count family.
+Integer bucket deltas are converted to absolute integer counts for storage;
+float bucket counts are already absolute. No separate sample-kind discriminator
+is stored because the populated count family identifies it.
+
+Within one resolved catalog, schema, and physical-table routing context, a
+metric name has exactly one persisted sample kind:
+
+- a float metric uses `greptime_value`;
+- a native-histogram metric uses the configured native-histogram field.
+
+Labels do not change that choice. Remote Write 2.0 rejects a request when two
+label sets use different sample kinds for the same metric, and the existing
+table schema rejects kind changes across requests. Prometheus metadata type is
+not used for this decision: a classic histogram family is described as a
+histogram but is represented by float-valued `_bucket`, `_sum`, and `_count`
+series.
+
+The storage invariant does not remove the need for mixed-sample PromQL support.
+An expression can combine different metrics or branches, for example with
+`or`, and therefore produce a vector containing float samples and histogram
+samples from different series.
+
+# Protocol Decisions
+
+## Remote Write 1.0
+
+Remote Write 1.0 scalar samples remain supported. Any `TimeSeries.histograms`
+field causes the complete request to fail with an invalid-arguments response.
+Rejecting at protobuf decode time prevents both histogram-only data loss and
+partial ingestion of a request that mixes scalar and histogram series. Clients
+that send native histograms must use Remote Write 2.0.
+
+## Remote Write 2.0
+
+Remote Write 2.0 accepts integer and float native histograms while
+`http.experimental_enable_prometheus_native_histogram` is enabled. Supported
+exponential schemas are `-4` through `8`; schema `-53` represents native
+histograms with custom buckets.
+
+Exponential spans may end at the schema's overflow bucket but must not continue
+beyond it. The overflow index is `(1024 << schema) + 1` for non-negative schemas
+and `(1024 >> -schema) + 1` for negative schemas. Inputs beyond that index are
+rejected rather than clamped or merged into infinity.
+
+Exemplars and declared metric metadata are accepted on the wire but are not
+persisted by this feature.
+
+## Remote Read
+
+Native-histogram Remote Read is deferred. GreptimeDB's existing Remote Read path
+continues to negotiate `SAMPLES` responses and serialize scalar samples only.
+
+A follow-up can map the stored Struct into `TimeSeries.histograms`. It must
+preserve integer counts without passing through `f64`, retain the stored
+histogram fields, and decide sampled versus streamed response coverage. The
+sampled protobuf has no start-timestamp field, so that omission must remain
+explicit if sampled responses are implemented.
+
+## OTLP
+
+OTLP exponential histograms are not converted in this version. The ingestion
+branch intentionally remains deferred until it can map temporality, scale,
+reset behavior, attributes, and rejected-point reporting into the canonical
+native-histogram path. The code carries an explicit TODO rather than a partial
+encoder.
+
+# PromQL Compatibility Decisions
+
+## Staleness
+
+A native histogram is stale only when its `sum` has Prometheus's stale-NaN bit
+pattern. Ordinary NaN sums remain samples. Range selectors remove stale markers
+before range functions run. Instant selectors treat the marker as the end of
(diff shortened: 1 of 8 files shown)


discord9:
Follow-up notes that don't fit as inline comments (outside the diff hunks):

- `NativeHistogramAvg` divides `sum / count` without a zero guard (src/promql/src/functions/native_histogram.rs:692-694): an empty histogram yields NaN (matches Prometheus, but untested; `estimated_stdvar` guards `count == 0.0` explicitly — the inconsistency is worth a test).
- `NativeHistogramAnnotationUdf::invoke_with_args` records its info on every invocation (native_histogram.rs:148-156) and the UDF is `Volatile`, so a query with N output batches may record the same annotation N times — worth confirming `PromqlAnnotationCollector::record_info` dedupes. The same applies to the `record_warning` calls in `native_extrapolated_rate`.

Nits:
- `divide_by(0.0)` clears spans/buckets but leaves `count = ±Inf` (native_histogram.rs:454-460), a degenerate histogram that can flow into `promql_eq`/`changes` — untested (Prometheus's DivByZero keeps bucket structure).
- `push_histogram` calls `observe_reset_hints` before the compatibility check (native_histogram.rs:836-861), so dropped incompatible histograms still influence the contradiction warning — harmless but surprising ordering.

docs/rfcs/2026-08-04-native-histograms.md
```diff
@@ -0,0 +1,225 @@
...
+## Remote Write 2.0
+
+Remote Write 2.0 accepts integer and float native histograms while
+`http.experimental_enable_prometheus_native_histogram` is enabled. Supported
+exponential schemas are `-4` through `8`; schema `-53` represents native
+histograms with custom buckets.
+
+Exponential spans may end at the schema's overflow bucket but must not continue
+beyond it. The overflow index is `(1024 << schema) + 1` for non-negative schemas
+and `(1024 >> -schema) + 1` for negative schemas. Inputs beyond that index are
```
discord9, line 87: The RFC promises that exponential spans beyond the overflow bucket are rejected at ingestion, but `validate_native_histogram_spans` in `src/servers/src/prom_remote_write/v2.rs` only bounds bucket indices for the custom-bucket schema (`custom_max_index.is_some()`); exponential schemas only get negative-offset and i32-overflow checks. A span reaching index `(1024 << schema) + 5` is accepted and persisted, and the rejection happens at query time instead: `get_bound` returns `None` above the overflow index, so `side_buckets` propagates `None` -> `quantile`/`fraction`/`estimated_stdvar` return NaN and `add`/`sub` silently drop the sample via `merge_side`. That is exactly the silent-loss mode this sentence claims the protocol boundary prevents.

Suggest either adding the overflow-index check to `validate_native_histogram_spans` (small change), or softening this sentence to "stored but rejected at query time, deferred to a later PR". Since this PR positions the RFC as the authoritative contract, resolving it here would be ideal.

docs/rfcs/2026-08-04-native-histograms.md
```diff
@@ -0,0 +1,225 @@
...
+## Staleness
+
+A native histogram is stale only when its `sum` has Prometheus's stale-NaN bit
+pattern. Ordinary NaN sums remain samples. Range selectors remove stale markers
+before range functions run. Instant selectors treat the marker as the end of
+the series for lookback purposes and do not resurrect an older sample.
+
+## Start timestamps
+
+Reset detection already uses consecutive start timestamps in addition to
```
discord9, line 123: One sentence worth adding here: the `duration_to_zero` extrapolation correction in `native_extrapolated_rate` is gated behind `!synthetic_zero_start && histogram.count > 0.0` (src/promql/src/functions/native_histogram.rs:1682-1687). The `!synthetic_zero_start` part looks intentional - with a synthetic zero the counter start is known, so the "counter started inside window" heuristic is unnecessary - and `histogram.count > 0.0` mirrors Prometheus's `resultFloat > 0` condition. Recording that choice would help future readers and document the all-zero-counter edge where we intentionally diverge from Prometheus's more liberal correction.

docs/rfcs/2026-08-04-native-histograms.md
```diff
@@ -0,0 +1,225 @@
...
+nodes or versions. If distributed stepping is added later, its state must be
+versioned before rolling upgrades can safely mix implementations.
+
+The rejected alternative is Prometheus's direct-sum/Kahan-compensation design
+with an overflow-triggered mode change. It offers closer rounding parity but
+requires additional state and substantially more merge logic.
+
+## Equality and changes
+
+Equality follows Prometheus's represented-layout semantics. Histograms compare
```
discord9, line 158: This section documents `changes()`, but the `resets()` implementation (src/promql/src/functions/native_histogram.rs:1102-1110) also counts gauge<->counter flavor transitions - `(pair[0].reset_hint == GAUGE_RESET_HINT) != (pair[1].reset_hint == GAUGE_RESET_HINT) || pair[1].detect_counter_reset(...)` - an intentional extension over Prometheus's reset-only counting that is not documented here. One sentence would close the gap.

src/promql/src/extension_plan/instant_manipulate.rs
```diff
@@ -626,9 +625,8 @@ impl InstantManipulateStream {
                         aligned_ts.push(expected_ts);
                     }
                 }
-            } else if let Some(field_column) = &field_column
-                && field_column.is_valid(cursor)
-                && is_prometheus_stale_nan(field_column.value(cursor))
+            } else if stale_sample_column
```
discord9, line 629: When the last sample of a batch is a stale marker, every aligned eval timestamp - including ones strictly before the marker - falls into this "Do not use a stale marker as the newest value" branch and produces no row, whereas Prometheus would resurrect the pre-stale value for `eval_ts < stale_ts`. This is safe only under the (unstated) assumption that stale markers are terminal and batches are per-series (SeriesDivide), so timestamps before the marker are unreachable in practice. Suggest documenting that assumption in a comment and adding a test with a regular sample after the stale marker.

src/promql/src/functions/native_histogram.rs
```diff
@@ -1588,6 +1589,20 @@ fn native_extrapolated_rate<const IS_COUNTER: bool, const IS_RATE: bool>(
             continue;
         }
 
+        let first_ts = all_timestamps[offset];
+        let last_ts = all_timestamps[offset + length - 1];
+        let range_end = eval_ts.value(index);
+        let range_start = range_end - range_length;
+        let synthetic_zero_timestamp = IS_COUNTER
```
evenyag, line 1596: **[P2] Apply selector offsets to native-histogram start timestamps**

`SeriesNormalize` shifts sample timestamps by the selector offset, but the embedded `start_timestamp` remains unchanged. This comparison therefore mixes timelines: `range_start` and `first_ts` are offset-adjusted while `samples[0].start_timestamp` is not. For a query such as `rate(metric[5m] offset 1h)`, a valid start timestamp can appear outside the range, disabling the synthetic-zero path; a single-sample range then returns no result, and a multi-sample range gets incorrect extrapolation. Please either shift the start timestamp together with the sample timestamp or perform these comparisons in the original timeline.

docs/rfcs/2026-08-04-native-histograms.md
shuiyisong: Confirmed fixed in the prepared follow-up PR 2 branch (`split/nh-02-protocol-contract`, commit `ae85aed28a`). It now derives the bound with the same `exponential_overflow_bucket_index` helper used by query-time `get_bound`, then validates every represented positive and negative bucket against `i32::MIN..=overflow_index` before write batches are submitted. The overflow bucket remains valid; `(1024 << schema) + 5` and every later index are rejected at ingestion. The shared validator covers both integer-delta and float-count histograms. Verified with the focused boundary regression for schemas `-4`, `0`, and `8` on both sides, plus all 15 Remote Write v2 unit tests.

docs/rfcs/2026-08-04-native-histograms.md
shuiyisong: Documented in `a84157d75`. The RFC now explains that, without a usable start timestamp, a positive histogram-count increase may infer a zero point from the first count, while a synthetic zero uses its known timestamp instead of that heuristic. I also checked current Prometheus: its histogram path likewise requires `resultHistogram.Count > 0`, so the all-zero case is parity rather than an intentional divergence.

docs/rfcs/2026-08-04-native-histograms.md
shuiyisong: Documented this in f2146bb3d2. The RFC now explicitly states that, unlike Prometheus, `resets()` counts each gauge/non-gauge native-histogram transition as one reset. Thanks!

src/promql/src/extension_plan/instant_manipulate.rs
shuiyisong: Checked this against current Prometheus main (b6a1a10668ef): a future stale marker does not hide the preceding sample, and selection resumes after a later regular sample. GreptimeDB already follows the same cursor semantics. In fa10f22bfc8, I documented the sorted single-series input contract and strengthened the terminal-stale regression to assert that the pre-marker value is selected while evaluations at and after the marker are suppressed; the existing stale-marker test already covers resumption after a later regular sample. The stale-focused PromQL tests (7) and Clippy pass. Thanks!

src/query/src/promql/planner.rs
```diff
@@ -1516,42 +1517,77 @@ impl PromPlanner {
...
+                    table: self.ctx.table_name.clone().unwrap_or_default(),
+                })?;
 
-        let normalize = if timestamp_fn {
-            // If evaluating the PromQL `timestamp()` function, project the time index column as the value column
-            // before wrapping with [`InstantManipulate`], so the output matches PromQL's `timestamp()` semantics.
-            self.create_timestamp_func_plan(normalize)?
+        let (normalize, timestamp_value_column) = if timestamp_fn {
+            // Keep the original sample for stale-marker detection while carrying
+            // its timestamp through InstantManipulate in a private value column.
```
evenyag, line 1530: Do we have a test to assert the plan and cover this branch?

evenyag (commented):
I didn't go through the details of the histogram functions.

src/promql/src/functions/native_histogram.rs
shuiyisong: Fixed in 11265bf41b. SeriesNormalize now applies the selector offset to both sample timestamps and every non-zero native-histogram start timestamp, keeping rate and reset calculations on the same timeline. The zero unknown sentinel, histogram payload, and Struct null bitmap are preserved, and timestamp addition is checked for overflow.

The regression test offsets_native_histogram_timestamps_and_filters_stale_markers covers the adjusted start timestamp, zero sentinel, outer null, and stale-marker filtering. Checks passed: cargo nextest run -p promql (207 passed), cargo nextest run -p query (597 passed, 2 skipped), PromQL Clippy, formatting, and diff checks.

src/query/src/promql/planner.rs
shuiyisong: Added in 8e4fb9f7cd. The old expected-panic timestamp test is now the positive exact-plan test single_timestamp_plan_preserves_source_value. It covers this branch by asserting that the original sample field remains available through InstantManipulate for stale-marker detection, the private timestamp value travels alongside it, and only the upper projection exposes that timestamp as the output value.

Checks passed: the focused test, cargo nextest run -p query (597 passed, 2 skipped), Query Clippy, formatting, and diff checks.

evenyag (approved):

discord9:
Gentle ping on the follow-up items from my general comment above — happy to file them as separate issues if this PR is wrapping up:

- `NativeHistogramAvg` divides `sum / count` without a zero guard (native_histogram.rs:692-694); empty histogram -> NaN, matches Prometheus but untested (inconsistent with `estimated_stdvar`'s explicit `count == 0.0` guard).
- `NativeHistogramAnnotationUdf::invoke_with_args` records its info on every invocation and is `Volatile` (native_histogram.rs:148-156) — worth confirming `PromqlAnnotationCollector::record_info` dedupes so warnings don't repeat per batch.
- `divide_by(0.0)` clears spans/buckets but leaves `count = ±Inf` (native_histogram.rs:454-460) — degenerate histogram that can flow into `promql_eq`/`changes`, untested.
- `push_histogram` calls `observe_reset_hints` before the compatibility check (native_histogram.rs:836-861) — dropped incompatible histograms still affect the contradiction warning.

No blockers, just recording them so they aren't lost.

src/promql/src/extension_plan/instant_manipulate.rs
discord9: Thanks for checking against Prometheus main — I misread the cursor logic: for `eval_ts < stale_ts` the search resumes from an earlier sample, so the pre-marker value is selected and only evaluations at/after the marker are suppressed. My concern was unfounded; the strengthened terminal-stale regression in `fa10f22bfc8` covers it. Retracting that comment.

PR #8760, opened 2026-08-05
fix(servers): merge Prometheus series with the same labelset across RecordBatch boundaries

### LINKED ISSUE #8607: Prometheus remote read returns duplicate series across RecordBatch boundaries
### What type of bug is this?

Incorrect result

Query Engine

### Minimal reproduce step

```
#[test]
fn test_recordbatches_to_timeseries_merges_series_across_batches() {
    let schema = Arc::new(Schema::new(vec![
        ColumnSchema::new(
            greptime_timestamp(),
            ConcreteDataType::timestamp_millisecond_datatype(),
            true,
        ),
        ColumnSchema::new(
            greptime_value(),
            ConcreteDataType::float64_datatype(),
            true,
        ),
        ColumnSchema::new(
            "instance",
            ConcreteDataType::string_datatype(),
            true,
        ),
    ]));

    // Multiple scan partitions may reach CoalescePartitionsExec in any order. Simulate the
    // newer batch arriving before the older batch for the same time series.
    let recordbatches = RecordBatches::try_new(
        schema.clone(),
        vec![
            RecordBatch::new(
                schema.clone(),
                vec![
                    Arc::new(TimestampMillisecondVector::from_vec(vec![2000])) as _,
                    Arc::new(Float64Vector::from_vec(vec![2.0])) as _,
                    Arc::new(StringVector::from(vec!["host1"])) as _,
                ],
            )
            .unwrap(),
            RecordBatch::new(
                schema,
                vec![
                    Arc::new(TimestampMillisecondVector::from_vec(vec![1000])) as _,
                    Arc::new(Float64Vector::from_vec(vec![1.0])) as _,
                    Arc::new(StringVector::from(vec!["host1"])) as _,
                ],
            )
            .unwrap(),
        ],
    )
    .unwrap();

    let timeseries = recordbatches_to_timeseries("metric1", recordbatches).unwrap();

    assert_eq!(1, timeseries.len());
    assert_eq!(
        vec![
            Sample {
                value: 1.0,
                timestamp: 1000,
            },
            Sample {
                value: 2.0,
                timestamp: 2000,
            },
        ],
        timeseries[0].samples
    );
}
```

  The conversion should establish the Prometheus wire-format invariants
  independently of execution-plan ordering:

  1. Maintain one request-level grouping index across all RecordBatches.
  2. Return exactly one protobuf TimeSeries per labelset.
  3. Check whether each resulting sample vector is ordered by timestamp.
  4. Sort only series whose samples are not already ordered.
  5. Sort the final TimeSeries list by labels to preserve deterministic output.

  Correctness should not depend on the optimizer adding a particular ordering or
  distribution hint. A future ordered scan path may avoid the fallback sorting
  cost, but the remote-read converter must still produce a valid response when its
  input stream is unordered.

### What did you see instead?

  Prometheus remote-read conversion groups rows independently for every
  `RecordBatch` and then flattens the resulting `TimeSeries` vectors.

  If the same labelset appears in multiple batches, GreptimeDB emits multiple
  protobuf `TimeSeries` entries with identical labels instead of merging their
  samples.

  This is possible during normal execution because remote-read queries do not
  request an ordering, and multiple scan partitions are combined by
  `CoalescePartitionsExec`. Batches therefore may arrive in an arbitrary order.

  Prometheus does not merge duplicate `TimeSeries` entries returned in one
  `QueryResult`. This can cause PromQL to fail with:

  `vector cannot contain metrics with the same labelset`

  It can also produce out-of-order samples after the duplicate series are merged,
  because batches from newer time ranges may arrive before older batches.

all

master

## Evidence chain

  This is not limited to large series or batches with more than 8192 rows. The
  remote-read execution plan has no global ordering guarantee, so even a small
  series split across storage ranges can be returned in multiple, arbitrarily
  ordered `RecordBatch`es.

  ### 1. Remote read does not request an ordering

  `query_to_plan()` only applies timestamp and label filters and returns the
  resulting logical plan. It does not add an `ORDER BY` or `Sort` node:

  - [`query_to_plan()` builds only a filter](https://github.com/GreptimeTeam/greptimedb/blob/598a412fbb551bcb5022a7cca7ca5e7b130758e7/src/servers/src/prom_store.rs#L141-L192)

  ### 2. Without a Sort node, no per-series scan hint is installed

  `ScanHintVisitor` only captures an ordering requirement when it encounters a
  `LogicalPlan::Sort`:

  - [`ScanHintVisitor` extracts ordering from `LogicalPlan::Sort`](https://github.com/GreptimeTeam/greptimedb/blob/598a412fbb551bcb5022a7cca7ca5e7b130758e7/src/query/src/optimizer/scan_hint.rs#L242-L261)

  The scan hint is only installed when that captured ordering exists:

  - [`set_order_hint()` is conditional on `visitor.order_expr`](https://github.com/GreptimeTeam/greptimedb/blob/598a412fbb551bcb5022a7cca7ca5e7b130758e7/src/query/src/optimizer/scan_hint.rs#L85-L102)

  Because the remote-read plan has no `Sort`, its scan request keeps
  `distribution = None`.

  ### 3. A scan with no distribution has neither ordering nor known partitioning

  For `distribution = None`, `RegionScanExec` constructs empty
  `EquivalenceProperties`, meaning it declares no output ordering. It also uses
  `UnknownPartitioning(num_output_partition)`:

  - [`None` produces no ordering and unknown partitioning](https://github.com/GreptimeTeam/greptimedb/blob/598a412fbb551bcb5022a7cca7ca5e7b130758e7/src/table/src/table/scan.rs#L199-L213)

  Therefore downstream operators cannot assume rows are globally ordered by
  labelset or timestamp.

  ### 4. Scan ranges may be distributed across multiple execution partitions

  `ParallelizeScan` uses DataFusion's configured `target_partitions` and assigns
  the available storage ranges among those output partitions:

  - [`ParallelizeScan` distributes ranges across target partitions](https://github.com/GreptimeTeam/greptimedb/blob/598a412fbb551bcb5022a7cca7ca5e7b130758e7/src/query/src/optimizer/parallelize_scan.rs#L55-L78)

  A series can span multiple ranges, for example when older samples are in SST
  files while newer samples are still in a memtable. Consequently, different
  parts of the same series may be read by different execution partitions.

  ### 5. DataFusion coalesces partitions without restoring order

  GreptimeDB currently pins the DataFusion fork to revision
  `7a84c45104d35408f00bbf880268c47193b73b28`.

  When an execution plan has multiple output partitions, DataFusion wraps it in
  `CoalescePartitionsExec`:

  - [`execute_stream()` inserts `CoalescePartitionsExec`](https://github.com/GreptimeTeam/datafusion/blob/7a84c45104d35408f00bbf880268c47193b73b28/datafusion/physical-plan/src/execution_plan.rs#L1221-L1234)

  `CoalescePartitionsExec` starts an independent task for every input partition
  and sends their batches into a shared receiver stream:

  - [`CoalescePartitionsExec` runs input partitions independently](https://github.com/GreptimeTeam/datafusion/blob/7a84c45104d35408f00bbf880268c47193b73b28/datafusion/physical-plan/src/coalesce_partitions.rs#L202-L225)

  There is no sort or sort-preserving merge in this path. Therefore, batch arrival
  order depends on which partition produces data first. For example, a memtable
  partition containing newer samples may arrive before an SST partition containing
  older samples.

  ### 6. GreptimeDB preserves that arbitrary arrival order

  The frontend collects the execution stream before conversion:

  - [`to_query_result()` collects the stream into `RecordBatches`](https://github.com/GreptimeTeam/greptimedb/blob/598a412fbb551bcb5022a7cca7ca5e7b130758e7/src/frontend/src/instance/prom_store.rs#L123-L133)

  `RecordBatches::try_collect()` uses the stream's yielded order directly:

  - [`try_collect()` collects batches into a `Vec`](https://github.com/GreptimeTeam/greptimedb/blob/598a412fbb551bcb5022a7cca7ca5e7b130758e7/src/common/recordbatch/src/lib.rs#L370-L374)

  No ordering is restored between execution and Prometheus conversion.

  ### 7. Conversion groups each batch independently

  `recordbatches_to_timeseries()` calls `recordbatch_to_timeseries()` separately
  for every batch and then flattens the resulting vectors:

  - [`recordbatches_to_timeseries()` performs per-batch conversion](https://github.com/GreptimeTeam/greptimedb/blob/598a412fbb551bcb5022a7cca7ca5e7b130758e7/src/servers/src/prom_store.rs#L306-L318)

  The grouping map is local to `recordbatch_to_timeseries()`. If the same labelset
  appears in two batches, each invocation creates a separate protobuf
  `TimeSeries`.

  For example:

  
  batch 1: instance=host1, samples=[2000]
  batch 2: instance=host1, samples=[1000]

  Current output:

  TimeSeries{instance=host1, samples=[2000]}
  TimeSeries{instance=host1, samples=[1000]}

  Expected output:

  TimeSeries{instance=host1, samples=[1000, 2000]}

  ### 8. Prometheus does not merge duplicate series in one QueryResult

  Prometheus converts every protobuf TimeSeries into a separate
  concreteSeries. Sorting only changes their order; it does not deduplicate or
  merge identical labelsets:

  - FromQueryResult() preserves every protobuf TimeSeries (https://github.com/prometheus/prometheus/blob/f371367eedf94b61c2db222fd94fe8db42b0bef1/storage/remote/codec.go#L184-L204)
  - concreteSeriesSet returns those series individually (https://github.com/prometheus/prometheus/blob/f371367eedf94b61c2db222fd94fe8db42b0bef1/storage/remote/codec.go#L353-L372)

  Prometheus's generic merge layer cannot repair this. When there is only one
  SeriesSet, it returns that set unchanged:

  - Single-set merge fast path returns the original set (https://github.com/prometheus/prometheus/blob/f371367eedf94b61c2db222fd94fe8db42b0bef1/storage/merge.go#L329-L338)

  As a result, duplicate labelsets from one remote-read response remain separate
  all the way into the PromQL evaluator.

  ### 9. Duplicate labelsets can make PromQL fail

  The PromQL evaluator explicitly rejects vectors containing multiple metrics with
  the same labelset:

  - Instant-query duplicate-labelset check (https://github.com/prometheus/prometheus/blob/f371367eedf94b61c2db222fd94fe8db42b0bef1/promql/engine.go#L1538-L1547)
  - Matrix duplicate-labelset check (https://github.com/prometheus/prometheus/blob/f371367eedf94b61c2db222fd94fe8db42b0bef1/promql/engine.go#L2453-L2457)
  - Final vector duplicate-labelset check (https://github.com/prometheus/prometheus/blob/f371367eedf94b61c2db222fd94fe8db42b0bef1/promql/engine.go#L4265-L4273)

  The resulting user-visible error is:

  vector cannot contain metrics with the same labelset

  Because partition scheduling and I/O completion order affect batch arrival, the
  failure can be nondeterministic for the same logical query.

  ### 10. Merged samples must also be sorted

  The Prometheus remote-read protocol explicitly requires samples within each time
  series to be ordered by time:

  - Remote-read QueryResult ordering contract (https://github.com/prometheus/prometheus/blob/f371367eedf94b61c2db222fd94fe8db42b0bef1/prompb/remote.proto#L74-L77)

  Prometheus relies on this contract. Its remote series iterator uses binary
  search when seeking to a timestamp:

  - concreteSeriesIterator.Seek() performs binary search (https://github.com/prometheus/prometheus/blob/f371367eedf94b61c2db222fd94fe8db42b0bef1/storage/remote/codec.go#L424-L453)

  Therefore, globally grouping identical labelsets is not sufficient by itself.
  If batches arrive out of timestamp order, simply appending their samples would
  produce an invalid remote-read response and could cause silently incorrect query
  results.

discord9:

Fixes #8607: Prometheus remote read returned duplicate series across RecordBatch boundaries. `recordbatches_to_timeseries` converted each RecordBatch independently, so a series split across batch boundaries (same labelset, different samples) was emitted as multiple `TimeSeries` instead of one merged series.

Now samples are accumulated per labelset across all batches (`HashMap<labelset, index>` → merge into the existing `TimeSeries`) before emitting, preserving single-batch behavior.

Tests: `recordbatches_to_timeseries_merges_series_across_batches` (RED before: 2 series; GREEN after: 1 merged with both samples).

## PR Checklist

### DIFF
--- a/src/servers/src/prom_store.rs
+++ b/src/servers/src/prom_store.rs
@@ -392,9 +392,7 @@ pub fn recordbatches_to_timeseries(
     // order for equal timestamps.
     if multiple_batches {
         for series in &mut timeseries {
-            series
-                .samples
-                .sort_by(|left, right| left.timestamp.cmp(&right.timestamp));
+            series.samples.sort_by_key(|sample| sample.timestamp);
         }
     }
 
--- a/src/servers/src/prom_store.rs
+++ b/src/servers/src/prom_store.rs
@@ -356,13 +356,16 @@ pub fn recordbatches_to_timeseries(
     table_name: &str,
     recordbatches: RecordBatches,
 ) -> Result<Vec<TimeSeries>> {
+    let recordbatches = recordbatches.take();
+    let multiple_batches = recordbatches.len() > 1;
+
     let mut timeseries: Vec<TimeSeries> = Vec::new();
     // Maps a labelset to the index of its TimeSeries in `timeseries`, so that
     // series split across RecordBatch boundaries are merged into a single
     // TimeSeries instead of being emitted once per batch.
     let mut timeseries_by_labels: HashMap<Vec<Label>, usize> = HashMap::new();
 
-    for recordbatch in recordbatches.take() {
+    for recordbatch in recordbatches {
         let batch_timeseries = recordbatch_to_timeseries(table_name, recordbatch)?;
         for series in batch_timeseries {
             match timeseries_by_labels.entry(series.labels.clone()) {
@@ -381,14 +384,18 @@ pub fn recordbatches_to_timeseries(
     timeseries
         .sort_unstable_by(|left, right| compare_timeseries_labels(&left.labels, &right.labels));
 
-    // Samples accumulated from multiple RecordBatches (which may arrive in any
-    // order from parallel scan partitions) must be ordered by timestamp
-    // ascending, as required by the Prometheus remote read protocol. A stable
-    // sort keeps insertion order for equal timestamps.
-    for series in &mut timeseries {
-        series
-            .samples
-            .sort_by(|left, right| left.timestamp.cmp(&right.timestamp));
+    // When samples from multiple RecordBatches are merged, the batches may
+    // arrive in any order (parallel scan partitions), so the accumulated
+    // samples must be ordered by timestamp ascending, as required by the
+    // Prometheus remote read protocol. A single batch is already in row order
+    // within the batch, so no sorting is needed. A stable sort keeps insertion
+    // order for equal timestamps.
+    if multiple_batches {
+        for series in &mut timeseries {
+            series
+                .samples
+                .sort_by(|left, right| left.timestamp.cmp(&right.timestamp));
+        }
     }
 
     Ok(timeseries)
--- a/src/servers/src/prom_store.rs
+++ b/src/servers/src/prom_store.rs
@@ -380,6 +380,17 @@ pub fn recordbatches_to_timeseries(
 
     timeseries
         .sort_unstable_by(|left, right| compare_timeseries_labels(&left.labels, &right.labels));
+
+    // Samples accumulated from multiple RecordBatches (which may arrive in any
+    // order from parallel scan partitions) must be ordered by timestamp
+    // ascending, as required by the Prometheus remote read protocol. A stable
+    // sort keeps insertion order for equal timestamps.
+    for series in &mut timeseries {
+        series
+            .samples
+            .sort_by(|left, right| left.timestamp.cmp(&right.timestamp));
+    }
+
     Ok(timeseries)
 }
 
@@ -1377,6 +1388,66 @@ mod tests {
                 Sample {
                     value: 5.0,
                     timestamp: 3000,
+                },
+            ],
+            timeseries[0].samples
+        );
+    }
+
+    #[test]
+    fn recordbatches_to_timeseries_sorts_merged_samples_by_timestamp() {
+        let schema = Arc::new(Schema::new(vec![
+            ColumnSchema::new(
+                greptime_timestamp(),
+                ConcreteDataType::timestamp_millisecond_datatype(),
+                true,
+            ),
+            ColumnSchema::new(greptime_value(), ConcreteDataType::float64_datatype(), true),
+            ColumnSchema::new("instance", ConcreteDataType::string_datatype(), true),
+        ]));
+
+        // Parallel scan partitions may reach the merge point in any order: here
+        // the newer batch (ts = 2000) arrives before the older batch
+        // (ts = 1000). The merged TimeSeries must still list samples in
+        // ascending timestamp order as required by the remote read protocol.
+        let recordbatches = RecordBatches::try_new(
+            schema.clone(),
+            vec![
+                RecordBatch::new(
+                    schema.clone(),
+                    vec![
+                        Arc::new(TimestampMillisecondVector::from_vec(vec![2000])) as _,
+                        Arc::new(Float64Vector::from_vec(vec![4.0])) as _,
+                        Arc::new(StringVector::from(vec!["host1"])) as _,
+                    ],
+                )
+                .unwrap(),
+                RecordBatch::new(
+                    schema,
+                    vec![
+                        Arc::new(TimestampMillisecondVector::from_vec(vec![1000])) as _,
+                        Arc::new(Float64Vector::from_vec(vec![3.0])) as _,
+                        Arc::new(StringVector::from(vec!["host1"])) as _,
+                    ],
+                )
+                .unwrap(),
+            ],
+        )
+        .unwrap();
+
+        let timeseries = recordbatches_to_timeseries("metric1", recordbatches).unwrap();
+
+        // Merged into a single series with samples sorted by timestamp.
+        assert_eq!(1, timeseries.len());
+        assert_eq!(
+            vec![
+                Sample {
+                    value: 3.0,
+                    timestamp: 1000,
+                },
+                Sample {
+                    value: 4.0,
+                    timestamp: 2000,
                 },
             ],
             timeseries[0].samples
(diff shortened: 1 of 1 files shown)


src/servers/src/prom_store.rs
```diff
@@ -356,14 +356,42 @@ pub fn recordbatches_to_timeseries(
...
+                Entry::Occupied(entry) => {
+                    timeseries[*entry.get()].samples.extend(series.samples);
+                }
+                Entry::Vacant(entry) => {
+                    let index = timeseries.len();
+                    entry.insert(index);
+                    timeseries.push(series);
+                }
+            }
+        }
```
evenyag, line 381: I think we can record whether there are multiple batches and skip some sorts.

src/servers/src/prom_store.rs
```diff
@@ -356,14 +356,42 @@ pub fn recordbatches_to_timeseries(
...
+        .sort_unstable_by(|left, right| compare_timeseries_labels(&left.labels, &right.labels));
+
+    // Samples accumulated from multiple RecordBatches (which may arrive in any
+    // order from parallel scan partitions) must be ordered by timestamp
+    // ascending, as required by the Prometheus remote read protocol. A stable
+    // sort keeps insertion order for equal timestamps.
+    for series in &mut timeseries {
+        series
+            .samples
+            .sort_by(|left, right| left.timestamp.cmp(&right.timestamp));
```
evenyag: If there is only one record batch, then we don't have to sort it. Also, if the planner has already sorted the result, then we may not sort it.

PR #8761, opened 2026-08-05
fix(servers): honor custom timestamp and value column names in Prometheus remote read

### LINKED ISSUE #8634: bug(prometheus): remote_read does not honor custom timestamp and value column names

discord9:

Fixes #8634: Prometheus remote read looked up columns via the hardcoded `greptime_timestamp()` / `greptime_value()` names, so a table whose timestamp/value columns use custom names (which remote write already accommodates) failed with `SchemaError FieldNotFound`.

Adds `resolve_read_column_names`, which resolves the timestamp and single Float64 value columns from the table schema (mirroring the write path's `get_alter_table_expr_on_demand`), and threads the resolved names through `query_to_plan` and `recordbatch_to_timeseries`.

Tests: `remote_read_honors_custom_timestamp_and_value_column_names`; all existing prom_store suites (servers 13+10, frontend 7) pass.

## PR Checklist

### DIFF
--- a/src/frontend/src/instance/prom_store.rs
+++ b/src/frontend/src/instance/prom_store.rs
@@ -140,7 +140,12 @@ fn resolve_remote_query_target(
 }
 
 #[instrument(skip_all, fields(table_name))]
-async fn to_query_result(table_name: &str, output: Output) -> ServerResult<QueryResult> {
+async fn to_query_result(
+    table_name: &str,
+    ts_col_name: &str,
+    value_col_name: &str,
+    output: Output,
+) -> ServerResult<QueryResult> {
     let OutputData::Stream(stream) = output.data else {
         unreachable!()
     };
@@ -148,7 +153,12 @@ async fn to_query_result(table_name: &str, output: Output) -> ServerResult<Query
         .await
         .context(error::CollectRecordbatchSnafu)?;
     Ok(QueryResult {
-        timeseries: prom_store::recordbatches_to_timeseries(table_name, recordbatches)?,
+        timeseries: prom_store::recordbatches_to_timeseries(
+            table_name,
+            recordbatches,
+            ts_col_name,
+            value_col_name,
+        )?,
     })
 }
 
@@ -161,7 +171,7 @@ impl Instance {
         schema_name: &str,
         table_name: &str,
         query: &Query,
-    ) -> Result<Output> {
+    ) -> Result<(String, String, Output)> {
         let table = self
             .catalog_manager
             .table(catalog_name, schema_name, table_name, Some(ctx))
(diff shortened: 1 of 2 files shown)


killme2008:
Duplicated with https://github.com/GreptimeTeam/greptimedb/pull/8659

Issue #8575, opened 2026-07-21
Tracking: Improve compaction picker adaptiveness and scalability


v0y4g3r:
## Background

The current compaction picker works well for common workloads, but some metadata- and ingestion-heavy scenarios can expose performance limitations. In these cases, the picker may spend too much time analyzing a large number of files, or may choose compaction inputs that are locally reasonable but insufficient for reducing backlog quickly.

This issue tracks follow-up work to improve compaction picker behavior across these scenarios.

### High-ingestion workloads with many small files

Under high ingestion pressure, flushes may produce many small SST files. The picker can become expensive when it needs to analyze a large number of files before returning compaction tasks. Even when it does return tasks, the selected inputs may be too small to reduce the backlog efficiently.

The desired behavior is to keep picker latency predictable while still feeding compaction workers with sufficiently large and useful batches.

### Files shared by different regions after partition split

After partition split, existing files may be shared across different regions. This can make file layout more complex and may increase the cost of determining safe and useful compaction inputs.

The picker needs to account for this kind of file sharing pattern when evaluating candidate inputs, rather than assuming a simple file-to-region layout.

### Increased file count after partition merge

After partition merge, the number of files visible to a region may increase significantly, potentially even doubling in some cases. This can amplify picker cost and may temporarily create a large compaction backlog.

The picker should handle this transition efficiently and avoid spending excessive time on global analysis before producing useful compaction work.

### Compaction contention with foreground read and write traffic

Compaction consumes CPU, memory, local disk or cache I/O, object-store bandwidth, and network capacity. If tasks are scheduled only from file layout and backlog, they may start during foreground traffic peaks and amplify read/write tail latency. Periodic flushes or independently triggered compactions can also align across regions or datanodes, creating short resource bursts even when average utilization looks acceptable.

Compaction scheduling should therefore become traffic-aware. When foreground read or write pressure is high, the scheduler should reduce compaction admission or throughput and defer non-urgent work. When pressure drops, it should resume gradually. Task starts should also be staggered across regions, windows, and—where coordination is available—datanodes, avoiding synchronized compaction bursts while guaranteeing eventual backlog progress.

## Goal

Improve compaction picker behavior for workloads where file count, file layout, or partition topology changes make the current strategy inefficient.

The goal is not to introduce one hardcoded policy for all cases. Instead, the picker should evolve toward a framework where different strategies can be applied to different scenarios.

## Proposed Direction

The improvement can be approached in stages:

[...]

PR #8765, opened 2026-08-05
refactor(mito2): prioritize file count in TWCS picker

v0y4g3r:

- #8575

This PR rewrites the Mito2 TWCS single-window picker to prioritize reducing the physical SST file count while preserving time locality, and teaches the compaction scheduler to keep draining a region while compaction makes progress.

**Count-first picking (`compaction/twcs.rs`)**

- Replaces the overlap-first run reduction (`reduce_runs`) and the size-targeted merge (`merge_seq_files`) with `pick_count_first`: all files of a window are ordered globally by `(start asc, end desc)` and every contiguous interval of at most 32 physical SSTs is scored.
- An interval is eligible when it holds at least `trigger_file_num` files (and at least 2), and when no single file dominates it (`largest <= sum of the others`), so a large historical file is only rewritten once its peers contribute at least the same amount of data.
- Scoring prefers more files removed, then more files participating in an overlap (resolving overlaps is what reduces sorted runs), then smaller input bytes; exact ties keep the earliest interval.
- `trigger_file_num` now applies consistently to the physical SST count, including overlapping sets: an overlapping set smaller than the trigger is no longer compacted.

**FileGroup abstraction removed (`compaction/run.rs`)**

- The picker operates directly on `FileHandle`s. Files produced by a split compaction output (same sequence) are evaluated individually instead of being kept atomic in a `FileGroup`; deletion-marker safety for partial selections is enforced per file instead (see below).
- `reduce_runs`, `merge_seq_files` and their benchmarks are removed.

**Conservative tombstone filtering**

- `filter_deleted` no longer relies on the `found_runs <= 2` heuristic. It is only set when the region is not in append mode, no other window overlaps this one, and no unselected file in the window overlaps the selection (with a cheap time-span pre-filter before the precise overlap check).

**Drain compactable backlog (`compaction/scheduler.rs`, `worker/handle_compaction.rs`)**

- After a successful compaction that reduced the SST file count (`files_to_remove > files_to_add`), the scheduler chains an unrestricted automatic follow-up pick even without a latched trigger, keeping the region draining until the picker returns no plan. A rewrite that did not reduce the file count (e.g. its output was split into more files than its input) ends the chain, so a no-progress loop cannot form; the next flush trigger resumes compaction.
- This applies to any successful compaction, including manual time-ranged ones: their follow-up pick is unrestricted (regular options, no time range).

This change does not modify persisted SST/manifest metadata, wire formats, configuration, or public APIs. Primary-key boundary metadata changes discussed in the design are intentionally out of scope.

## PR Checklist

## Test Plan

- `cargo nextest run -p mito2 --no-fail-fast`
- `cargo clippy -p mito2 --all-targets -- -D warnings`
- `git diff --check`

## Benchmark monitoring

One-hour continuous write load followed by follow-up compaction, compared against the v1.1.4 baseline with the same CPU quota:

- Total compaction output bytes to reach the final state: **8.71GB → 5.78GB**
- Total merge duration: **3.66h → 3.05h**
- Peak alive SST file count: **954 → 756** (both fully drain afterwards)

Cumulative SST output bytes
Cumulative compaction total elapsed
Alive SST files trend

### DIFF
--- a/src/mito2/src/compaction/twcs.rs
+++ b/src/mito2/src/compaction/twcs.rs
@@ -168,17 +168,9 @@ impl TwcsPicker {
         };
         let found_runs = sorted_runs.len();
         let inputs = pick_count_first(sorted_runs, self.trigger_file_num);
-        let mut filter_deleted = !self.append_mode
+        let filter_deleted = !self.append_mode
             && !window_has_overlap(files, windows)
             && !selected_overlaps_unselected(&inputs, files);
-
-        // The inputs are only a subset of the window: `reduce_runs` and `merge_seq_files` both
-        // narrow the selection down and the file num limit above narrows it further. A deletion
-        // marker may only be dropped when every file that can hold a row it masks is compacted
-        // along with it, so re-check the final selection against what stays in the window.
-        if filter_deleted && overlaps_files_left_behind(&inputs, &files_to_merge) {
-            filter_deleted = false;
-        }
 
         if inputs.len() > 1 {
             // If we have more than one group to compact.
@@ -195,20 +187,6 @@ impl TwcsPicker {
         }
         (inputs, filter_deleted)
     }
-}
-
-/// Returns whether any file group of `window_files` that is not part of `inputs` overlaps
-/// `inputs`. Such a group keeps rows a deletion marker in `inputs` masks, so the compaction
-/// must not filter deleted rows out.
-///
-/// Overlapping is inclusive on both boundaries here: two files that only share a boundary
-/// timestamp can still hold the same row.
-fn overlaps_files_left_behind(inputs: &[FileGroup], window_files: &[FileGroup]) -> bool {
-    let picked: HashSet<_> = inputs.iter().flat_map(|fg| fg.file_ids()).collect();
-    window_files
-        .iter()
-        .filter(|fg| !fg.file_ids().iter().any(|id| picked.contains(id)))
-        .any(|fg| inputs.iter().any(|input| input.overlap_inclusive(fg)))
 }
 
 #[derive(Debug)]
--- a/src/mito2/src/compaction/twcs.rs
+++ b/src/mito2/src/compaction/twcs.rs
@@ -32,8 +32,8 @@ use crate::compaction::buckets::infer_time_bucket;
 use crate::compaction::compactor::CompactionRegion;
 use crate::compaction::picker::{Picker, PickerOutput, get_expired_ssts};
 use crate::compaction::run::{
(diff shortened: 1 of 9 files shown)


src/mito2/src/compaction/twcs.rs
```diff
@@ -32,17 +31,17 @@ use crate::compaction::buckets::infer_time_bucket;
...
 use crate::error::{JoinSnafu, Result};
 use crate::sst::file::{FileHandle, Level, overlaps};
 use crate::sst::version::LevelMeta;
 
 const LEVEL_COMPACTED: Level = 1;
 
-/// Default value for max compaction input file num.
-const DEFAULT_MAX_INPUT_FILE_NUM: usize = 32;
+/// Maximum number of input SST files in one compaction input.
+const MAX_INPUT_FILES: usize = 32;
```
evenyag: Do we need to expose a nodocs environment variable to change it?

src/mito2/src/compaction/twcs.rs
```diff
@@ -235,18 +193,227 @@ impl TwcsPicker {
...
+///   overlap between sorted runs.
+///
+/// The best interval wins by [`CandidateScore`]; ties keep the earliest interval.
+fn pick_count_first(
+    sorted_runs: Vec<SortedRun<FileHandle>>,
+    trigger_file_num: usize,
+    max_output_file_size: Option<u64>,
+) -> Vec<FileHandle> {
+    let files = ordered_files(&sorted_runs);
+    let min_files = trigger_file_num.max(2);
```
evenyag: `trigger_file_num` values above `MAX_INPUT_FILES` disable compaction permanently. Each candidate is capped at 32 files, but `min_files` remains the unrestricted configured threshold, so a valid setting such as 64 can never produce a candidate even after hundreds of SSTs accumulate. Previously the window threshold could exceed the per-task input cap. Please check the window’s total physical file count against the trigger and then select at most 32 inputs, or validate that the option cannot exceed 32.

v0y4g3r: Addressed in ab69b333ed. `trigger_file_num` now gates the window total physical SST count, while each compaction candidate remains independently capped by `MAX_INPUT_FILES`. Added a regression test with a trigger above the input cap.

src/mito2/src/compaction/twcs.rs
v0y4g3r: Addressed in 1499a0713b. Added the hidden `GREPTIME_TWCS_MAX_INPUT_FILES` override with a validated minimum of 2 while preserving 32 as the default.

evenyag (approved):

PR #8766, opened 2026-08-05
feat: add Parquet table name metadata

fengjiachun:
## Summary

- define the generic Parquet footer metadata key `greptime.table_name` with an opaque, unqualified table-name value
- write table identity for `COPY TABLE ... TO ... FORMAT PARQUET` while keeping generic writers, `COPY QUERY`, and non-Parquet exports unchanged
- add reusable single-file extraction and batch consistency validation for missing, invalid, incomplete, duplicate, and conflicting metadata

## Design goals

This change gives Parquet files exported by `COPY TABLE` an optional source-table identity without changing their schema, row data, or encoding.

- `COPY TABLE` writes the unqualified table name to the `greptime.table_name` footer metadata key.
- `COPY QUERY` does not write table identity because a query result may not correspond to a single source table.
- Generic Parquet writers retain their existing no-metadata behavior.
- Shared extraction helpers validate duplicate keys and multi-file consistency, returning actionable errors when table identities conflict or appear in only part of a file set.

The table name is intentionally treated as an opaque value. This change does not define catalog or schema resolution.

## Compatibility

- The metadata is additive and optional; it does not change the Parquet schema, columns, or encoded row data.
- Legacy Parquet files without `greptime.table_name` remain valid and continue through the unknown-table path.
- Existing Parquet readers can ignore the unknown footer key, while older GreptimeDB releases can still import files written with it.
- A bidirectional compatibility case covers current importing legacy Parquet without the key and an older release importing current Parquet with the key.
- JSON/JSONB encoding behavior is unchanged.

## Tests

- `cargo nextest run -p common-datasource -p operator -p sqlness-runner` (262 passed, 1 configured skip)
- `cargo check -p common-datasource -p operator -p sqlness-runner --all-targets`
- `cargo clippy -p common-datasource -p operator -p sqlness-runner --all-targets -- -D warnings`
- `cargo fmt --all --check`
- `python3 -m py_compile .github/scripts/run-compat.py`
- compatibility dry-run for `parquet_copy_footer_compatibility` from `v1.1.4` to `current` (exactly 1 case selected)
- `git diff upstream/main --check`


fengjiachun:
@codex review

src/common/datasource/src/file_format/parquet.rs
```diff
@@ -211,17 +283,42 @@ impl ArrowWriterCloser for ArrowWriter<SharedBuffer> {
...
+                .collect(),
+        )
+    };
     let write_props = column_wise_config(
         WriterProperties::builder()
             .set_compression(Compression::ZSTD(ZstdLevel::default()))
             .set_statistics_truncate_length(None)
-            .set_column_index_truncate_length(None),
+            .set_column_index_truncate_length(None)
+            .set_key_value_metadata(metadata),
```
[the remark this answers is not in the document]
fengjiachun: Fixed in a6a679aa686da91a10d941c3c00ee32a97db4e3b. Added a bidirectional standalone compatibility case for legacy Parquet without `greptime.table_name` and current Parquet with the footer key. CI now exercises v1.0.0/v1.1.0 -> current and current -> v1.1.4. Local self-compat passes; sqlness-runner: 37 passed; common-datasource/operator: 146 passed, 1 skipped; check/clippy/fmt all pass.

fengjiachun:
@codex review

killme2008:
What is the purpose of this design? I think it should be described in the PR @fengjiachun

src/common/datasource/src/file_format/parquet.rs
```diff
@@ -43,10 +43,82 @@ use tokio_util::compat::{Compat, FuturesAsyncReadCompatExt, FuturesAsyncWriteCom
...
+        };
+        if value.is_empty() {
+            return InvalidParquetTableNameMetadataSnafu {
+                reason: format!("metadata key '{PARQUET_TABLE_NAME_KEY}' has an empty value"),
+            }
+            .fail();
+        }
+        if table_name.is_some_and(|table_name| table_name != value) {
+            return InvalidParquetTableNameMetadataSnafu {
+                reason: format!("metadata key '{PARQUET_TABLE_NAME_KEY}' has conflicting values"),
```
killme2008: Add the  names into the error message

src/common/datasource/src/file_format/parquet.rs
```diff
@@ -43,10 +43,82 @@ use tokio_util::compat::{Compat, FuturesAsyncReadCompatExt, FuturesAsyncWriteCom
...
+pub fn parquet_table_name_from_metadata_batch<'a>(
+    metadata: impl IntoIterator<Item = Option<&'a [KeyValue]>>,
+) -> Result<Option<&'a str>> {
+    let mut table_name = None;
+    let mut saw_missing = false;
+    for metadata in metadata {
+        match parquet_table_name_from_metadata(metadata)? {
+            Some(value) => {
+                if saw_missing || table_name.is_some_and(|table_name| table_name != value) {
+                    return InvalidParquetTableNameMetadataSnafu {
```
killme2008, line 105: Add the  names into the error message

src/common/datasource/src/file_format/parquet.rs
```diff
@@ -43,10 +43,82 @@ use tokio_util::compat::{Compat, FuturesAsyncReadCompatExt, FuturesAsyncWriteCom
...
+                        reason: "Parquet files have inconsistent table names".to_string(),
+                    }
+                    .fail();
+                }
+                table_name = Some(value);
+            }
+            None => {
+                if table_name.is_some() {
+                    return InvalidParquetTableNameMetadataSnafu {
+                        reason: "Parquet files have inconsistent table names".to_string(),
```
killme2008: Add the  names into the error message

src/common/datasource/src/file_format/parquet.rs
```diff
@@ -211,17 +283,42 @@ impl ArrowWriterCloser for ArrowWriter<SharedBuffer> {
...
+/// Output the stream to a parquet file with custom key/value footer metadata.
+///
+/// Returns number of rows written.
+pub async fn stream_to_parquet_with_metadata(
     mut stream: SendableRecordBatchStream,
     schema: datatypes::schema::SchemaRef,
     store: ObjectStore,
     path: &str,
     concurrency: usize,
+    metadata: Vec<(String, String)>,
```
killme2008: Why not use a tuple `(String, String)`? A vector seems unnecessary, or you could define a type alias instead.

src/operator/src/statement/copy_table_to.rs
```diff
@@ -49,11 +51,21 @@ const WRITE_BUFFER_THRESHOLD: ReadableSize = ReadableSize::mb(8);
 /// Default number of concurrent write, it only works on object store backend(e.g., S3).
 const WRITE_CONCURRENCY: usize = 8;
 
+fn parquet_metadata(format: &Format, table_name: Option<&str>) -> Vec<(String, String)> {
```
killme2008: Use a tuple as the result type.

`Option<(String, String)>`.

src/common/datasource/src/file_format/parquet.rs
fengjiachun: Fixed in 5af1aa0104. The duplicate-key error now includes both conflicting table names (debug-escaped), with an exact-message regression test. Verified with 146 common-datasource/operator tests, all-targets check/clippy, fmt, and diff check.

src/common/datasource/src/file_format/parquet.rs
fengjiachun: Fixed in 5af1aa0104. Batch conflicts now report both observed table names, with an exact-message regression test. Verified with 146 common-datasource/operator tests, all-targets check/clippy, fmt, and diff check.

src/common/datasource/src/file_format/parquet.rs
fengjiachun: Fixed in 5af1aa0104. Mixed present/missing metadata now reports the observed table name, with tests covering both file orders and the full error text. Verified with 146 common-datasource/operator tests, all-targets check/clippy, fmt, and diff check.

src/common/datasource/src/file_format/parquet.rs
fengjiachun: Fixed in 5af1aa0104. The writer now accepts a single optional (String, String) metadata pair instead of a vector; the no-metadata path uses None. Verified with 146 common-datasource/operator tests, all-targets check/clippy, fmt, and diff check.

src/operator/src/statement/copy_table_to.rs
fengjiachun: Fixed in 5af1aa0104. parquet_metadata now returns Option<(String, String)>, and focused tests cover Parquet table export, Parquet query export, and non-Parquet export routing. Verified with 146 common-datasource/operator tests, all-targets check/clippy, fmt, and diff check.

PR #8767, opened 2026-08-06
perf(mito2): optimize dictionary primary key sorting

lyang24:

### Summarize your change

  Optimize Mito2 bulk record-batch sorting when the primary key is encoded as
  `Dictionary<UInt32, Binary>`.

  For eligible batches, replace the generic three-column Arrow lexsort with a
  specialized path that ranks dictionary values once, scatters rows into
  primary-key buckets, and sorts only buckets that are not already ordered by
  timestamp and sequence. Unsupported layouts continue to use the existing Arrow
  lexsort fallback.

  This change also adds correctness tests against Arrow lexsort and targeted
  Criterion benchmarks.

  ### How does this PR work?

  The required row order is:

  1. primary key ascending;
  2. timestamp ascending;
  3. sequence descending.

  The optimized path works as follows:

  1. Rank the dictionary values once. Equal dictionary values receive the same
     rank, even if they use different dictionary keys.
  2. Count rows for each rank and compute bucket offsets.
  3. Scatter row indices into primary-key buckets while preserving their input
     order.
  4. Check whether each bucket is already ordered by `(timestamp ASC, sequence
     DESC)`.
  5. Sort only the unordered buckets using direct typed comparisons.
  6. Apply the resulting indices with `take_record_batch`.

  The fast path avoids repeatedly dereferencing and comparing encoded primary-key
  bytes during an `O(n log n)` multi-column comparison sort.

  For the 8192-row microbenchmarks, the specialized path was:

  - about 17x faster when samples within each series were already ordered;
  - about 11x faster when samples within each series were shuffled;
  - about 1.58x faster for the high-cardinality case with one row per series.

  In the existing 10-million-row remote-write workload, CPU time per row decreased
  by approximately 12.8%, while throughput increased by approximately 13.6%.

  ### Describe clearly one logical change

  This PR contains one logical change: specializing Mito2's existing
  `(primary_key, timestamp, sequence)` bulk sorting operation for its known
  dictionary-encoded primary-key layout. The tests and benchmarks only validate
  that change.

  ### Limitations

  The specialized path currently applies only when:

  - the primary key is `Dictionary<UInt32, Binary>`;
  - the timestamp uses one of Arrow's supported timestamp units with an `i64`
    physical representation;
  - the sequence column is `UInt64`;
  - the three sorting columns contain no nulls;
  - the dictionary does not contain more values than rows.

  Batches outside these conditions fall back to Arrow's existing
  `lexsort_to_indices` implementation.

image

## PR Checklist

### DIFF
--- a/src/mito2/src/memtable/bulk/part.rs
+++ b/src/mito2/src/memtable/bulk/part.rs
@@ -30,11 +30,13 @@ use datafusion_common::pruning::PruningStatistics;
 use datafusion_expr::utils::expr_to_columns;
 use datatypes::arrow;
 use datatypes::arrow::array::{
-    Array, ArrayRef, BinaryArray, BooleanArray, StringDictionaryBuilder, UInt8Array, UInt64Array,
+    Array, ArrayRef, BinaryArray, BooleanArray, DictionaryArray, StringDictionaryBuilder,
+    TimestampMicrosecondArray, TimestampMillisecondArray, TimestampNanosecondArray,
+    TimestampSecondArray, UInt8Array, UInt32Array, UInt64Array,
 };
 use datatypes::arrow::compute::{SortColumn, SortOptions, concat_batches};
 use datatypes::arrow::datatypes::{
-    DataType as ArrowDataType, Field, Schema, SchemaRef, UInt32Type,
+    DataType as ArrowDataType, Field, Schema, SchemaRef, TimeUnit, UInt32Type,
 };
 use datatypes::data_type::DataType;
 use datatypes::extension::json::is_json2_extension_type;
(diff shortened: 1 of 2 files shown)


src/mito2/src/memtable/bulk/part.rs
```diff
@@ -765,6 +767,149 @@ fn new_primary_key_column_builders(
...
+
+    for (rank, count) in counts.into_iter().enumerate() {
+        let start = bucket_offsets[rank];
+        let end = start + count;
+        let bucket = &mut indices[start..end];
+        if bucket.windows(2).all(|rows| {
+            compare_time_sequence(timestamps, sequences.values(), rows[0], rows[1]).is_le()
+        }) {
+            continue;
+        }
```
evenyag: Maybe we can use is_sorted_by().
https://doc.rust-lang.org/stable/std/primitive.slice.html#method.is_sorted_by

evenyag:
@codex review

evenyag (approved):
This is impressive! I left one non-blocking comment.

src/mito2/src/memtable/bulk/part.rs
lyang24: thank you fixed along with other rust style formatting issues

PR #8768, opened 2026-08-06
feat: add admin function to discard unflushed data

### LINKED ISSUE #8464: Add an admin function to drop the unflushed data

In some case, we may encounter flush issue that caused by data inside the memtable. For example, large binary blobs that exceed array offset.

Once we failed to flush the memtable, we will fill up the write buffer and can't serve other write requests. Even we find out the table containing the data, we can't recover without deleting the whole WAL.

### What does the feature do?

Add a method to clear the memtable and drop the data that unable to flush.

Then we can recover from this error.

### Implementation challenges

We also need to remove the unflushed data from WAL. Otherwise, we will replay them after the datanode restarts.

evenyag:

- Proto change: https://github.com/GreptimeTeam/greptime-proto/pull/334
- Fixes https://github.com/GreptimeTeam/greptimedb/issues/8464

This PR adds the `ADMIN discard_unflushed_data(...)` function for explicitly dropping a Mito region's unflushed in-memory data.

It translates the admin request into the new protobuf truncate request variant and dispatches it to the region leader. The function accepts either a region ID for precise region-level administration or a table name to discard unflushed data from every region belonging to that table.

The existing Rust integration test covers the region-ID path. A sqlness case covers the table-name path in both standalone and distributed deployments, including preservation of flushed data and idempotent repeated calls.

## PR Checklist


src/operator/src/request.rs
```diff
@@ -206,6 +210,77 @@ impl Requester {
...
+            table_info.is_physical_table(),
+        )?;
+
+        let partitions = &self
+            .partition_manager
+            .find_physical_partition_info(table_info.ident.table_id)
+            .await
+            .with_context(|_| FindTablePartitionRuleSnafu {
+                table_name: full_table_name,
+            })?
```
v0y4g3r:
```suggestion
                table_name: table_name.to_string(),
            })?
```

src/operator/src/request.rs
```diff
@@ -206,6 +210,77 @@ impl Requester {
...
+        self.do_request(vec![request], None, &ctx).await
+    }
+
+    /// Discard all unflushed data from all regions of the table.
+    pub async fn handle_discard_unflushed_data_by_table(
+        &self,
+        table_name: TableName,
+        ctx: QueryContextRef,
+    ) -> Result<AffectedRows> {
+        let full_table_name = table_name.to_string();
```
v0y4g3r: nit: We can postpone the `to_string` call when needed

```suggestion
```

src/operator/src/request.rs
```diff
@@ -206,6 +210,77 @@ impl Requester {
...
+                &table_name.catalog_name,
+                &table_name.schema_name,
+                &table_name.table_name,
+                None,
+            )
+            .await
+            .context(CatalogSnafu)?;
+        let table = table.with_context(|| TableNotFoundSnafu {
+            table_name: full_table_name.clone(),
+        })?;
```
v0y4g3r:
```suggestion
            table_name: table_name.to_string(),
        })?;
```

v0y4g3r (approved):

v0y4g3r (approved):
LGTM

src/common/function/src/admin/discard_unflushed_data.rs
```diff
@@ -0,0 +1,232 @@
...
+use store_api::storage::RegionId;
+use table::table_name::TableName;
+
+use crate::handlers::TableMutationHandlerRef;
+use crate::helper::cast_u64;
+
+/// Discards all unflushed data from a region.
+#[admin_fn(
+    name = DiscardUnflushedDataFunction,
+    display_name = discard_unflushed_data,
```
killme2008: It's a poor name, it's too long and not like other SQL functions.

What about `discard_memtable` or `discard_unflushed`?

tests/cases/standalone/common/function/admin/discard_unflushed_data.sql
```diff
@@ -0,0 +1,46 @@
+CREATE TABLE discard_unflushed_data_test (
+    host STRING PRIMARY KEY,
+    val DOUBLE,
+    ts TIMESTAMP TIME INDEX
+) ENGINE = mito;
+
+INSERT INTO discard_unflushed_data_test VALUES ('persisted', 1, 1);
+
```
killme2008, line 8: We should test the function without any persisted data and ensure the table remains empty after discarding.

killme2008 (approved):
LGTM

But prefer to use another name.

src/common/function/src/admin/discard_unflushed_data.rs
evenyag: It not only discards the memtable, but it also needs to update the wal. So `discard_memtable` isn't a good name.

killme2008: Let's use `discard_unflushed`.

PR #8772, opened 2026-08-06
fix(metric-engine): handle Utf8View tag/label columns without panicking

### LINKED ISSUE #8732: Prometheus label_replace query panics when serializing Utf8View labels

discord9:

Fixes #8732: a PromQL `label_replace` query panicked when serializing Utf8View labels. `label_replace` is planned as DataFusion `regexp_replace`, whose signature coerces to Utf8View, so label columns materialize as `StringViewArray`; `build_tag_arrays`' `StringArray` downcast then panicked ("tag column must be utf8") — e.g. for OTLP/json2 ingest.

TSID computation, sparse-PK encoding and tag extraction now accept generic `ArrayRef` tag columns (Utf8/LargeUtf8/Utf8View/Dictionary) via `string_array_value_at_index`; `build_tag_arrays` errors instead of panicking on non-string columns. The mito2 time-series memtable string-field paths are hardened the same way.

Tests: `label_replace_with_utf8view_labels_does_not_panic`; prometheus_resp 6/6, batch_modifier 5/5, promql 150/150, memtable time_series 18/18.

## PR Checklist

### DIFF
--- a/src/metric-engine/src/batch_modifier.rs
+++ b/src/metric-engine/src/batch_modifier.rs
@@ -18,7 +18,7 @@ use std::sync::Arc;
 use datatypes::arrow::array::{Array, ArrayRef, BinaryBuilder, UInt64Array};
 use datatypes::arrow::datatypes::{DataType, Field, Schema as ArrowSchema};
 use datatypes::arrow::record_batch::RecordBatch;
-use datatypes::arrow_array::string_array_value_at_index;
+use datatypes::arrow_array::{is_string_null_at, string_array_value_at_index};
 use fxhash::FxHasher;
 use mito_codec::row_converter::SparsePrimaryKeyCodec;
 use snafu::ResultExt;
@@ -65,9 +65,7 @@ pub fn compute_tsid_array(
 
     let mut tsid_values = Vec::with_capacity(num_rows);
     for row in 0..num_rows {
-        let has_null = tag_arrays
-            .iter()
-            .any(|arr| string_array_value_at_index(arr, row).is_none());
+        let has_null = tag_arrays.iter().any(|arr| is_string_null_at(arr, row));
 
         let tsid = if !has_null {
             let mut hasher = FxHasher::default();
@@ -82,7 +80,7 @@ pub fn compute_tsid_array(
         } else {
             let mut name_hasher = FxHasher::default();
             for (tc, arr) in sorted_tag_columns.iter().zip(tag_arrays.iter()) {
-                if string_array_value_at_index(arr, row).is_some() {
+                if !is_string_null_at(arr, row) {
                     name_hasher.write(tc.name.as_bytes());
                     name_hasher.write_u8(0xff);
                 }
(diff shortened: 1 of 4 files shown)


src/metric-engine/src/batch_modifier.rs
```diff
@@ -64,20 +65,24 @@ pub fn compute_tsid_array(
 
     let mut tsid_values = Vec::with_capacity(num_rows);
     for row in 0..num_rows {
-        let has_null = tag_arrays.iter().any(|arr| arr.is_null(row));
+        let has_null = tag_arrays
+            .iter()
+            .any(|arr| string_array_value_at_index(arr, row).is_none());
```
evenyag: Should we provide another function to check whether it is null?

src/metric-engine/src/batch_modifier.rs
```diff
@@ -64,20 +65,24 @@ pub fn compute_tsid_array(
...
+                    hasher.write(value.as_bytes());
+                    hasher.write_u8(0xff);
+                }
             }
             hasher.finish()
         } else {
             let mut name_hasher = FxHasher::default();
             for (tc, arr) in sorted_tag_columns.iter().zip(tag_arrays.iter()) {
-                if !arr.is_null(row) {
+                if string_array_value_at_index(arr, row).is_some() {
```
evenyag: ditto.

evenyag (approved):

evenyag:
@codex review

PR #8776, opened 2026-08-06
feat: update opentelemetry family to 0.32 series

sunng87:

this patch updates our opentelemtry libraries to 0.32, for both inbound otlp and internal outbound opentelemetry sdk

## PR Checklist


evenyag:
There is a typo in the title: openetelemtry

sunng87:
> There is a typo in the title: openetelemtry

- [x] Human verification

src/servers/src/otlp/logs.rs
```diff
@@ -1031,6 +1032,16 @@ fn any_value_to_vrl_value(value: any_value::Value) -> VrlValue {
...
+        // `StringValueStrindex` references the Profiling signal's
+        // `ProfilesDictionary.string_table`, which is unavailable to logs/traces.
+        // Per the OTLP spec, non-Profiling receivers must treat it as a non-fatal
+        // issue and process the value as if it were absent.
+        any_value::Value::StringValueStrindex(_) => {
+            warn!(
+                "encountered a profiling-only `StringValueStrindex` value in a non-profiling signal; ignoring"
+            );
+            VrlValue::Null
+        }
```
shuiyisong, line 1044: Nit: could cause a lot of `warn`s if we have multiple profile records with in one request. But since profiling is not that popular, we can save it later.
Same as below

shuiyisong (approved):
LGTM

src/servers/src/otlp/logs.rs
sunng87: it's unlikely to see this type in trace/log data if client works correctly. Here we just deal with the match arms

PR #8778, opened 2026-08-06
perf: reduce ingestion and flat-merge overhead

shuiyisong:

This pull request improves ingestion and flat-merge performance by removing redundant hashing and allocations and avoiding per-row clock reads.

- **Column lookup:** Before, `TableData` used the standard `HashMap` hasher for frequent column-name lookups. After, its private index uses the existing AHash implementation.
- **PRW label validation:** Before, every time series allocated a new duplicate-label set. After, one request-scoped set is cleared and reused while keeping validation series-local.
- **PRW value-type validation:** Before, separate sample and histogram sets duplicated keys already stored in the output maps. After, conflicts are derived from those maps while preserving same-series and cross-series rejection.
- **Flat merge timing:** Before, every row advance read the clock and contributed to `fetch_cost`. After, timing occurs only when the last row triggers a next-batch fetch attempt.

The change does not modify public APIs, wire formats, persisted data, or ingestion results. `fetch_cost` now measures batch-fetch work rather than in-memory cursor advances.

## PR Checklist

### DIFF
--- a/src/servers/src/row_writer.rs
+++ b/src/servers/src/row_writer.rs
@@ -14,6 +14,7 @@
 
 use std::collections::HashMap;
 
+use ahash::{HashMap as AHashMap, HashMapExt};
 use api::v1::column_data_type_extension::TypeExt;
 use api::v1::helper::time_index_column_schema;
 use api::v1::value::ValueData;
@@ -37,7 +38,7 @@ use crate::error::{
 pub struct TableData {
     schema: Vec<ColumnSchema>,
     rows: Vec<Row>,
-    column_indexes: HashMap<String, usize>,
+    column_indexes: AHashMap<String, usize>,
 }
 
 impl TableData {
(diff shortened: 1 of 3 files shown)


src/servers/src/row_writer.rs
```diff
@@ -37,15 +38,15 @@ use crate::error::{
 pub struct TableData {
     schema: Vec<ColumnSchema>,
     rows: Vec<Row>,
-    column_indexes: HashMap<String, usize>,
+    column_indexes: AHashMap<String, usize>,
```
evenyag, line 41: Should we change the hash map here? @v0y4g3r

evenyag (approved):

src/servers/src/row_writer.rs
shuiyisong: The idea is to get some free performance gain. Is there any specific reason to use the std hashmap?

PR #8780, opened 2026-08-06
fix(function): stop CountHash convert_to_state merging raw values as hash-v1 state

discord9:
## What

Removes the custom `convert_to_state` override from `CountHashGroupAccumulator` so partial (hash-v1) state cannot be built by wrapping raw input values in a `List<UInt64>`.

## Why

The previous `convert_to_state` implementation wrapped the raw input array (any type) as a `List<UInt64>` state. When that state was later merged via `merge_batch`, the raw values were treated as hash-v1 values, which:

- double-counts a value after a normal→skip conversion (`convert_to_state` followed by merge),
- violates the non-UInt64 subtype contract and can panic,
- ignores the optional filter's validity.

The upstream fix removed the fallback; this ports it to GreptimeDB and freezes the hash-v1 wire format with a fixture test so the behavior cannot silently regress.

## Test

- `cargo test -p common-function count_hash` — 13 passed (incl. `verify_count_hash_hash_v1_fixture`)
- RED harness earlier confirmed the bug on main (repeat counts, false-filter, non-UInt64 panic) and PASS on the fixed worktree.

## Checklist

- [x] I have signed off my commits (git commit -s)
- [x] No internal issue numbers referenced in code

### DIFF
--- a/src/common/function/src/aggrs/count_hash.rs
+++ b/src/common/function/src/aggrs/count_hash.rs
@@ -35,7 +35,6 @@ use datafusion_expr::{
     Accumulator, AggregateUDF, AggregateUDFImpl, EmitTo, GroupsAccumulator, ReversedUDAF,
     SetMonotonicity, Signature, TypeSignature, Volatility,
 };
-use datafusion_functions_aggregate_common::aggregate::groups_accumulator::nulls::filtered_null_mask;
 use datatypes::arrow;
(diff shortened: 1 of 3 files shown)


killme2008:
This PR appears to cause a performance regression because it removes the fast path. Could we implement `convert_to_state` properly so that it works correctly? @discord9

PR #8784, opened 2026-08-06
feat(promql): support mixed sample ranges

shuiyisong:

This is the 3rd PR of the 6 PRs of native histogram.

This pull request adds coordinated PromQL evaluation for range vectors containing both float and native histogram samples. It selects compatible lanes per window, applies Prometheus-compatible warning and information policies, and reuses existing range UDFs.

It also supports nested range planning and distributed Substrait decoding. Native-histogram vector-operator integration remains scoped to the follow-up PR.

## PR Checklist


src/query/src/promql/planner.rs
```diff
@@ -8921,6 +9063,166 @@ mod test {
...
+        )
+        .await
+        .unwrap()
+        .display_indent_schema()
+        .to_string();
+        assert!(
+            plan.matches("prom_mixed_range_float").count() >= 2
+                && plan.matches("prom_mixed_range_histogram").count() >= 2,
+            "{plan}"
+        );
```
evenyag: I think we need an exact match for the output plan. It helps us understand what the final plan looks like.

shuiyisong: Addressed in https://github.com/GreptimeTeam/greptimedb/pull/8784/commits/6dfc1cacdc68bbc82f303b8deb9d93bb93340c4e, please take a look

src/promql/src/functions/native_histogram.rs
```diff
@@ -1264,6 +1269,772 @@ impl NativeHistogramResets {
...
+    output: MixedRangeOutput,
+    collector: Option<PromqlAnnotationCollector>,
+) -> DfResult<ColumnarValue> {
+    if args.args.len() < 4 {
+        return Err(DataFusionError::Plan(format!(
+            "{} function should have at least 4 inputs",
+            output.name()
+        )));
+    }
+    let function = MixedRangeFunction::parse(&args.args[0])?;
```
evenyag, line 1573: Why don't we parse and create the function in the planning phase?

src/promql/src/functions/native_histogram.rs
```diff
@@ -1264,6 +1269,772 @@ impl NativeHistogramResets {
...
+
+    let bounds = checked_window_bounds(keys, timestamps.len(), name)?;
+    let float_valid = (0..floats.len())
+        .map(|row| floats.is_valid(row))
+        .collect::<Vec<_>>();
+    let histogram_valid = (0..histograms.len())
+        .map(|row| histograms.is_valid(row))
+        .collect::<Vec<_>>();
+    let float_prefix = validity_prefix(&float_valid, name)?;
+    let histogram_prefix = validity_prefix(&histogram_valid, name)?;
```
evenyag, line 1636: Add some documents to the functions like checked_window_bounds, validity_prefix.

src/promql/src/functions/native_histogram.rs
```diff
@@ -1264,6 +1269,772 @@ impl NativeHistogramResets {
...
+    histogram_valid: &[bool],
+    float_prefix: &[usize],
+    histogram_prefix: &[usize],
+    collector: &Option<PromqlAnnotationCollector>,
+) -> Vec<Option<SampleLane>> {
+    bounds
+        .iter()
+        .map(|(offset, end)| {
+            let float_count = valid_count(float_prefix, *offset, *end);
+            let histogram_count = valid_count(histogram_prefix, *offset, *end);
```
evenyag, line 1778: Is valid_count the length of the element in the range vector? Can we get it from the range vector directly?

src/promql/src/functions/native_histogram.rs
```diff
@@ -1264,6 +1269,772 @@ impl NativeHistogramResets {
...
+                            ),
+                        );
+                        None
+                    }
+                    (true, false) => Some(SampleLane::Float),
+                    (false, true) => Some(SampleLane::Histogram),
+                    (false, false) => None,
+                },
+                MixedRangePolicy::FloatOnly => {
+                    if histogram_count > 0 {
```
evenyag: Prometheus emits the histogram-ignored info only for genuinely mixed windows. With the current condition, histogram-only windows also record the info even though Prometheus silently omits those ranges. Could we require both lanes here, i.e. `float_count > 0 && histogram_count > 0`?

src/promql/src/functions/native_histogram.rs
```diff
@@ -1264,6 +1269,772 @@ impl NativeHistogramResets {
...
+fn compact_lane_input(
+    timestamps: &TimestampMillisecondArray,
+    values: &dyn Array,
+    valid: &[bool],
+    prefix: &[usize],
+    bounds: &[(usize, usize)],
+    selections: &[Option<SampleLane>],
+    lane: SampleLane,
+    name: &str,
+) -> DfResult<Vec<ColumnarValue>> {
```
evenyag, line 1864: Why do we need compact?

src/promql/src/functions/native_histogram.rs
shuiyisong: Addressed in https://github.com/GreptimeTeam/greptimedb/pull/8784/commits/18cc673fd8660e4b62278b671584e2a893fd0046. Added documentation for the window-bound, validity-prefix/count, lane-selection, and lane-compaction invariants. Please take a look.

src/promql/src/functions/native_histogram.rs
shuiyisong: The function name is kept as an argument because Substrait only serializes the UDF name and arguments, not state captured inside `MixedRangeUdf`. Parsing it during planning would lose the function variant when the generic UDF is reconstructed remotely. Using separate UDF names per function could avoid this, but would significantly expand the decoder registry.

src/promql/src/functions/native_histogram.rs
shuiyisong: `valid_count` is the number of non-null samples in one lane, not the range’s raw length. Float and histogram ranges share the same bounds and use null placeholders for the other sample type, so their raw lengths are identical. The validity prefix also remaps offsets after compaction and gives O(1) counts for overlapping windows.

src/promql/src/functions/native_histogram.rs
shuiyisong: Good catch. I updated the condition to require both lanes (`float_count > 0 && histogram_count > 0`), so histogram-only windows are omitted silently. I also added a regression case to `mixed_ranges_follow_prometheus_sample_type_semantics` that verifies histogram-only `min_over_time` records no info while the existing mixed-window case still verifies the info is emitted. `cargo nextest run -p promql native_histogram` passes all 22 focused tests.

src/promql/src/functions/native_histogram.rs
shuiyisong: Mixed lanes are row-aligned and use null placeholders for samples of the other type, while the existing float and histogram range UDFs expect dense inputs. Compaction filters the selected lane and matching timestamps, remaps overlapping range offsets, and turns dropped/opposite-lane windows into empty ranges. Without it, functions such as `rate` and `idelta` would read null positions as physical values or emit results for a dropped mixed window.

src/promql/src/functions/native_histogram.rs
```diff
@@ -1264,6 +1269,778 @@ impl NativeHistogramResets {
...
+                            function,
+                            &mut previous,
+                            timestamps.value(row),
+                            MixedSample::Histogram(histogram),
+                        );
+                    }
+                }
+                result.append_value(count as f64);
+            }
+            _ => unreachable!(),
```
evenyag: Should we return an error to avoid panic if we change the policy and forget to update this?

evenyag (approved):

src/promql/src/functions/native_histogram.rs
shuiyisong: Fixed in https://github.com/GreptimeTeam/greptimedb/pull/8784/commits/2a283fec0edfa0682b66e3b4615d9ab02fe55a5b

PR #8794, opened 2026-08-07
feat: declared edges and the derivation contract for the entity graph

killme2008:

Tracking issue: #8609. Follow-up to #8614 and the [entity relationships RFC](https://github.com/GreptimeTeam/greptimedb/blob/main/docs/rfcs/2026-06-25-entity-relationships-and-graph-query.md). This PR finishes the remaining M1 work for the read-time entity graph.

### Run derivation as the caller

Graph derivation previously used a new default `QueryContext`, so it did not follow the outer query's source-table permissions.

It now uses the caller's context. Each source table is checked separately, and tables the caller cannot read are left out. The derived query also inherits the caller's cancellation and deadline.

### Use the query's time range

The `observed_at` filter now controls the derivation window:

- no filter uses the last hour;
- no upper bound means up to now;
- no lower bound or an unsupported filter returns an error;
- source scans cover complete 60-second buckets, so boundary buckets do not produce partial RED metrics.

### Add declared edges

This adds `greptime_private.semantic_relationships_declared` for relationships that cannot be derived from telemetry, such as a manually declared service dependency.

The table is created with a fixed schema on the first SQL or gRPC insert, even when automatic table creation is disabled. Its definition is owned by the system: CREATE, ALTER, RENAME, UNDROP, and a same-name view are rejected. DROP and TRUNCATE remain available for cleanup and recovery. INSERT and DELETE are allowed.

`semantic_relationships` now includes these rows. It selects the latest revision as of the query window and applies `valid_from` and `valid_until` before returning an edge.

### Current scope

- The entity-graph feature is still experimental, and its output schema may change.
- Attribute-derived edges, virtual nodes, and Prometheus Remote Write 2.0 metadata are not included in this PR.

## PR Checklist


src/operator/src/statement/semantic_graph.rs
```diff
@@ -406,21 +669,111 @@ const RELATIONSHIP_COLUMNS: [&str; 16] = [
...
+    // As-of the queried window: a revision recorded after its end, or whose
+    // validity starts after it, was not the edge's state inside the window and
+    // must not outrank (and thereby hide) the revision that was.
+    let as_of = ident(OBSERVED_AT_COLUMN)
+        .lt(window.observed_end())
+        .and(eff_valid_from.clone().lt(window.observed_end()));
+
+    let revision = row_number()
+        .partition_by(
+            DECLARED_PRIMARY_KEY_COLUMNS
```
fengjiachun: `scope` and `generation_id` are included in `DECLARED_PRIMARY_KEY_COLUMNS`, so two assertions of the same visible edge under different scopes/generations are ranked in separate partitions and both receive row number 1. The projection below then drops both fields, causing `semantic_relationships` to return duplicate, indistinguishable edges instead of the latest revision. For example, reasserting the same endpoints/provenance as generations `g1` and `g2` leaves both rows visible. Please either rank by the externally visible edge identity, or expose these fields and define them as part of the computed-table contract. A regression test with two generations would cover this.

fengjiachun (commented):
I found one correctness issue in the declared-edge revision selection.

src/operator/src/statement/semantic_graph.rs
killme2008: Good catch!
Fixed in https://github.com/GreptimeTeam/greptimedb/pull/8794/changes/4ac877adafe674c7707d78ca9454b72bfb18efb7

shuiyisong:
Revalidated against live head 5264b5de after checking all existing conversation comments, reviews, inline comments, replies, and resolved or outdated threads. These three findings are still present and are not already tracked.

### [P1] Reject non-contiguous or partially extracted graph windows

src/table/src/predicate.rs:258

extract_time_range_strict accepts the lenient extractor convex hull, but query_window later treats it as an exact derivation interval and declared_edges emits one synthetic timestamp. For example, an edge valid from t=15 queried with observed_at = t=10 OR observed_at = t=20 is emitted at t=15; the inexact outer system-table filter then drops it even though the edge is valid at t=20. The range implementation explicitly turns disjoint ranges into a hull.

Smallest fix: return Unsupported unless the filter is exactly representable as one contiguous literal range.

### [P1] Keep a future lower-only window empty

src/frontend/src/instance/entity_graph.rs:344

For observed_at >= now() + one day, the extracted start is later than the implicit end, now. An old open declaration still passes the as-of checks against that end, then greatest(valid_from, start) emits the future start; the original lower-only outer filter accepts it. This returns a false future edge even though the contract defines a missing upper bound as up to now.

Smallest fix: return an empty derivation when the resolved start >= end.

### [P2] Classify the canonically created table before routing its first gRPC write

src/operator/src/insert.rs:651

Every sibling existing or create path adds an instant-TTL table to instant_table_ids; this reserved-table branch does not. With GREPTIMEDB_DECLARED_RELATIONSHIPS_TTL=instant, the first request enters normal_requests and reports one affected row, while later requests are discarded normally. I reproduced this with:

    GREPTIMEDB_DECLARED_RELATIONSHIPS_TTL=instant cargo nextest run -p tests-integration test_grpc_first_write_creates_declared_relationships_table

The test passes its AffectedRows(1) assertion, whereas the established instant auto-create test expects zero.

Smallest fix: perform the same is_ttl_instant_table() insertion before storing table_info.

killme2008:
@shuiyisong 
All three fixed in 0851d606f0:

- OR/IN over observed_at are now rejected by the strict extractor instead of being hulled — they were never in the documented shapes anyway.
- A window whose resolved start >= end (future lower bound vs the implicit now) derives nothing.
- The reserved-table gRPC branch now does the same is_ttl_instant_table() classification as its siblings.

fengjiachun (approved):
Rest LGTM

Issue #4765, opened 2024-09-24
Prometheus remote write 2.0


sunng87:
### What problem does the new feature solve?

With Prometheus 3.0 beta release, a version of prometheus remote write is out

https://prometheus.io/docs/specs/remote_write_spec_2_0/

GreptimeDB will keep tracking this and keep our compatibility up to date.

### What does the feature do?

Implement PRW 2.0 while keeping 1.0 supported.

PR #8797, opened 2026-08-07
feat(servers): stamp prometheus remote write v2 metadata as semantic table options

killme2008:

#4765; the "Remote Write 2.0 metadata intake" M1 item of `docs/rfcs/2026-06-25-entity-relationships-and-graph-query.md`.

Remote write 2.0 carries per-series metadata (type, unit, help) on the wire; the v2 ingest decoded and dropped it, so tables kept the name-inferred quality. This PR stamps it into the tables' semantic options:

- `greptime.semantic.metric.type` from the declared metric type; conflicting declarations collapse to `mixed`.
- `greptime.semantic.metric.metadata_quality = declared` — only when an explicit type was declared; `UNSPECIFIED` series keep `inferred`.
- `greptime.semantic.metric.unit`, canonicalised from OpenMetrics words to UCUM (`seconds` → `s`); units outside the OpenMetrics base set are dropped. Help text is not persisted.

The OTLP per-table index machinery moves to a shared `servers::semantic` module and becomes schema-aware, keyed `{schema -> table -> options}` — one v2 request may write the same metric name into several schemas via the special labels. The index is parsed once per create-planning round on both auto-create paths (operator row-insert and the pending-rows batched create), and every metadata symbol reference is validated up front per the spec.

Options are stamped at auto-create only; updating existing tables from later metadata is a follow-up (metadata registry).

## PR Checklist


src/servers/src/prom_remote_write/v2.rs
```diff
@@ -166,9 +186,73 @@ pub(crate) fn into_write_requests(request: Request) -> Result<RemoteWriteV2Write
...
+    let unit = if series_metadata.unit_ref != 0 {
+        Some(symbol_ref(
+            symbols,
+            series_metadata.unit_ref,
+            "metadata unit",
+        )?)
+    } else {
+        None
+    };
+    let Some(metric_type) = metric_type_value(series_metadata.r#type) else {
```
shuiyisong: **[P2] Preserve a valid unit when the metric type is UNSPECIFIED**

This early return drops a valid mapped unit whenever metadata.type is UNSPECIFIED. OpenMetrics models TYPE and UNIT independently, with an omitted TYPE becoming Unknown ([spec](https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily-metadata)), and the Prometheus RW2 sender explicitly tests UNSPECIFIED metadata carrying units such as meters ([sender test](https://github.com/prometheus/prometheus/blob/main/storage/remote/queue_manager_test.go#L2357-L2428)).

Greptime accepts this series and keeps the request-level metadata quality as inferred, but the auto-create path receives no greptime.semantic.metric.unit. Since the unit is not stored in rows and later table-option updates are deferred, it cannot be recovered after creation.

Could we compute the metric type and mapped unit independently, gate only the type plus metadata_quality=declared stamps on a known type, and still record a mapped unit when present? The existing test_metadata_unspecified_leaves_no_stamp case can then assert unit=m while quality remains inferred.

killme2008: Good catch — fixed in 857725093b. Type and unit now stamp independently: an UNSPECIFIED series still records its mapped UCUM unit, while the type stamp and the metadata_quality=declared upgrade keep requiring an explicit type. The test now covers unit-only (unit=s, no type/quality stamp), UNSPECIFIED-without-unit, and no-metadata.

shuiyisong (approved):
LGTM 🚀

PR #8798, opened 2026-08-07
feat(promql): support native histogram vector operators

shuiyisong:

This is the 4th PR of the 6 PRs of native histogram

This pull request adds native histogram support to PromQL unary negation, arithmetic and comparison operators, and set operations. It preserves mixed float/native-histogram vectors through `or`, scalar broadcasting, filtering, and downstream operators while keeping annotation-producing UDFs on the frontend in distributed plans.

Unsupported histogram pairings are dropped with accurate info annotations only when matched samples are present, avoiding silent drops and false-positive annotations. The change does not alter public configuration, persisted data, or wire formats.

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/commutativity.rs
+++ b/src/query/src/dist_plan/commutativity.rs
@@ -31,34 +31,11 @@ use datafusion_expr::{Expr, LogicalPlan, UserDefinedLogicalNode};
 use promql::extension_plan::{
     EmptyMetric, InstantManipulate, RangeManipulate, SeriesDivide, SeriesNormalize,
 };
-use promql::functions::{
-    NativeHistogramAdd, NativeHistogramAggAvg, NativeHistogramAggSum, NativeHistogramAvgOverTime,
-    NativeHistogramDelta, NativeHistogramDrop, NativeHistogramIDelta, NativeHistogramIRate,
-    NativeHistogramIncrease, NativeHistogramRate, NativeHistogramSub, NativeHistogramSumOverTime,
-};
 use store_api::metric_engine_consts::DATA_SCHEMA_TSID_COLUMN_NAME;
 
 use crate::dist_plan::MergeScanLogicalPlan;
 use crate::dist_plan::analyzer::AliasMapping;
 use crate::dist_plan::merge_sort::{MergeSortLogicalPlan, merge_sort_transformer};
-
-// These functions capture the frontend annotation collector. Reconstructing them on a datanode
-// would execute correctly but silently lose their warnings and info messages.
-const ANNOTATING_NATIVE_HISTOGRAM_FUNCTIONS: &[&str] = &[
-    NativeHistogramAdd::name(),
-    NativeHistogramAggAvg::name(),
-    NativeHistogramAggSum::name(),
-    NativeHistogramAvgOverTime::name(),
-    NativeHistogramDelta::name(),
-    NativeHistogramIDelta::name(),
-    NativeHistogramIRate::name(),
-    NativeHistogramIncrease::name(),
-    NativeHistogramRate::name(),
-    NativeHistogramSub::name(),
-    NativeHistogramSumOverTime::name(),
-    NativeHistogramDrop::bool_false_name(),
-    NativeHistogramDrop::float_null_name(),
-];
 
 #[cfg(feature = "vector_index")]
 fn is_vector_sort(sort: &Sort) -> bool {
@@ -179,13 +156,6 @@ impl Categorizer {
             LogicalPlan::Filter(filter) => Self::check_expr(&filter.predicate),
             LogicalPlan::Window(_) => Commutativity::Unimplemented,
             LogicalPlan::Aggregate(aggr) => {
-                if aggr
-                    .aggr_expr
-                    .iter()
-                    .any(Self::contains_annotating_native_histogram_function)
-                {
-                    return Ok(Commutativity::NonCommutative);
-                }
                 let is_all_steppable = is_all_aggr_exprs_steppable(&aggr.aggr_expr);
                 let matches_partition = Self::check_partition(&aggr.group_expr, &partition_cols);
                 if !matches_partition && is_all_steppable {
@@ -327,10 +297,6 @@ impl Categorizer {
     }
 
     pub fn check_expr(expr: &Expr) -> Commutativity {
-        if Self::contains_annotating_native_histogram_function(expr) {
-            return Commutativity::NonCommutative;
-        }
-
         #[allow(deprecated)]
         match expr {
             Expr::Column(_)
@@ -373,25 +339,6 @@ impl Categorizer {
         }
     }
 
-    fn contains_annotating_native_histogram_function(expr: &Expr) -> bool {
-        let mut found = false;
-        expr.apply(|expr| {
-            let name = match expr {
-                Expr::ScalarFunction(func) => Some(func.name()),
-                Expr::AggregateFunction(func) => Some(func.func.name()),
-                _ => None,
-            };
-            if name.is_some_and(|name| ANNOTATING_NATIVE_HISTOGRAM_FUNCTIONS.contains(&name)) {
-                found = true;
-                Ok(TreeNodeRecursion::Stop)
-            } else {
-                Ok(TreeNodeRecursion::Continue)
-            }
-        })
-        .expect("expression traversal is infallible");
-        found
-    }
-
     /// Return true if the given expr and partition cols satisfied the rule.
     /// In this case the plan can be treated as fully commutative.
     ///
@@ -430,7 +377,7 @@ mod tests {
     use datafusion_common::Column;
     use datafusion_expr::LogicalPlanBuilder;
     use datafusion_expr::expr::ScalarFunction;
-    use promql::functions::{NativeHistogramCount, NativeHistogramDrop, NativeHistogramRate};
+    use promql::functions::NativeHistogramDrop;
 
     use super::*;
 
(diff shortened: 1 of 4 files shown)


src/query/src/dist_plan/commutativity.rs
```diff
@@ -156,6 +179,13 @@ impl Categorizer {
             LogicalPlan::Filter(filter) => Self::check_expr(&filter.predicate),
             LogicalPlan::Window(_) => Commutativity::Unimplemented,
             LogicalPlan::Aggregate(aggr) => {
+                if aggr
+                    .aggr_expr
+                    .iter()
+                    .any(Self::contains_annotating_native_histogram_function)
+                {
+                    return Ok(Commutativity::NonCommutative);
+                }
```
evenyag: This will block the aggregation pushdown. I don't want the less important annotation feature to affect the query performance. We should remove this check. cc @discord9

evenyag (commented):
I haven't finished reviewing the planner.

src/query/src/promql/planner.rs
```diff
@@ -4501,18 +4862,117 @@ impl PromPlanner {
...
+        op: TokenType,
+        left_is_scalar: bool,
+        right_is_scalar: bool,
+    ) -> (
+        Vec<(String, Vec<BinaryFieldPair<'a>>)>,
+        Vec<BinaryFieldPair<'a>>,
+    ) {
+        // Mixed vectors store mutually exclusive float and histogram samples in two columns.
+        // Retain each valid sample combination and group expressions by their output lane.
+        let left_alternative = Self::alternative_sample_columns(left_schema, left_field_columns);
```
evenyag, line 4874: Non-blocking follow-up: handling mixed float/native-histogram samples through two physical columns introduces substantial planner complexity around lane detection, alignment, filtering, projection, and generated names.

It may be worth introducing a logical abstraction such as:

```rust
enum Sample<T> {
    Float(T),
    Histogram(T),
    Mixed { float: T, histogram: T },
}

type SampleExpr = Sample<DfExpr>;
type SampleColumns = Sample<String>;
```

The planner could operate on one logical sample, while materializing it into the existing physical columns only at plan boundaries. This would centralize binary-operation dispatch and mixed-lane invariants without requiring changes to storage or Arrow representation.

I don’t think this needs to block the current PR, but it would be useful to discuss as a follow-up refactor before more native-histogram operations add similar lane-handling logic.

evenyag (commented):
Non-blocking refactor suggestion.

src/query/src/dist_plan/commutativity.rs
```diff
@@ -31,12 +31,35 @@ use datafusion_expr::{Expr, LogicalPlan, UserDefinedLogicalNode};
...
+    NativeHistogramAggSum::name(),
+    NativeHistogramAvgOverTime::name(),
+    NativeHistogramDelta::name(),
+    NativeHistogramIDelta::name(),
+    NativeHistogramIRate::name(),
+    NativeHistogramIncrease::name(),
+    NativeHistogramRate::name(),
+    NativeHistogramSub::name(),
+    NativeHistogramSumOverTime::name(),
+    "prom_native_histogram_drop_bool",
```
killme2008: why not use `bool_false_name` in `NativeHistogramDrop` insead?

Keep consistent with others.

src/query/src/dist_plan/commutativity.rs
```diff
@@ -31,12 +31,35 @@ use datafusion_expr::{Expr, LogicalPlan, UserDefinedLogicalNode};
...
+    "prom_native_histogram_drop_float",
```
killme2008: The same suggestion as above

src/query/src/dist_plan/commutativity.rs
killme2008: It looks like it checks whether it’s in `ANNOTATING_NATIVE_HISTOGRAM_FUNCTIONS` and, if so, marks it as `NonCommutative`.

This will not block other aggregations to pushdown.

src/query/src/promql/planner.rs
```diff
@@ -1240,10 +1314,47 @@ impl PromPlanner {
...
+                    .iter()
+                    .map(|col| {
+                        Self::binary_result_is_histogram(
+                            *op,
+                            Self::field_column_is_native_histogram(&input_schema, col),
+                            false,
+                        )
+                        .is_some()
+                    })
+                    .collect();
```
killme2008: What about refactoring `has_native_histogram` into a reusable function?

It looks like the same logic is duplicated in several places.

src/query/src/promql/planner.rs
```diff
@@ -1230,7 +1294,17 @@ impl PromPlanner {
...
+                        self.projection_for_each_field_column(input, bin_expr_builder)?;
+                    if has_native_histogram {
+                        self.filter_binary_projection(
+                            projected,
+                            preserve_any_value,
+                            retain_field_columns,
+                        )
+                    } else {
+                        Ok(projected)
+                    }
```
killme2008: Duplicated code, could we refactor into a function?

src/query/src/dist_plan/commutativity.rs
shuiyisong: Fixed in 61e0a93866. NativeHistogramDrop::bool_false_name() and ::float_null_name() are now public const accessors, and ANNOTATING_NATIVE_HISTOGRAM_FUNCTIONS uses them instead of duplicated strings. Validated with cargo nextest run -p promql -p query (845 passed, 2 skipped).

src/query/src/dist_plan/commutativity.rs
shuiyisong: Fixed alongside the preceding comment in 61e0a93866. Both drop-UDF entries now reference the associated name accessors on NativeHistogramDrop, so there are no duplicated name literals. Validated with cargo nextest run -p promql -p query (845 passed, 2 skipped).

src/query/src/promql/planner.rs
shuiyisong: Fixed in 61e0a93866. Added field_columns_contain_native_histogram(schema, field_columns) and reused it in the binary-island, literal/vector, and vector/vector paths. Validated with cargo nextest run -p promql -p query (845 passed, 2 skipped).

src/query/src/promql/planner.rs
shuiyisong: Fixed in 61e0a93866. filter_binary_projection now owns the no-histogram early return, so all three projection sites use the same helper without merging the operand-order branches. Validated with cargo nextest run -p promql -p query (845 passed, 2 skipped).

src/query/src/dist_plan/commutativity.rs
evenyag: Marking NonCommutative will block further pushdown. We have discussed this in the original PR for the native histogram planner.

We have another recommended way to implement the annotation.

discord9 (commented):
Overall the histogram operator semantics look solid and the float/histogram dual-lane design is reasonable. That said, I'm no longer able to manually review changes of this scale: `planner.rs` is past 10k lines and this PR adds ~1.5k more. I'm going to open a separate PR to split `planner.rs` into a `planner/` module directory (public API unchanged), and I'd suggest landing that split before the rest of the series (#5/#6) keeps building on the same file.

A few concrete findings from the review:

1. **timestamp(mixed_metric) stale-marker leak** — the planner only passes `field_columns.first()` to `InstantManipulate` for stale detection; a stale sample on the other mutually-exclusive lane is emitted as a valid result.
2. **Annotating whitelist misses MixedRange** — `ANNOTATING_NATIVE_HISTOGRAM_FUNCTIONS` in `dist_plan/commutativity.rs` doesn't list `prom_mixed_range_float`/`prom_mixed_range_histogram`, so their warnings/infos can be silently lost when the plan is pushed down in distributed mode.
3. **Two independent operator truth tables** — `native_histogram_binary_expr` and `binary_result_is_histogram` encode the same operator semantics separately (with `Option<bool>` encoding three states); adding operators in #5/#6 will require touching both.

Happy to help review the split once it's up.

src/query/src/dist_plan/commutativity.rs
shuiyisong: Fixed in bdcd37d866. Removed the annotation-based `NonCommutative` special-casing, so native histogram UDFs/UDAFs no longer stop distributed pushdown; annotation transport remains deferred to a stream-level solution as discussed in #8693. Added `annotations_do_not_block_aggregate_pushdown` to verify the drop UDF is commutative and a steppable aggregate is classified as `TransformedCommutative`. Validated with `cargo nextest run -p promql -p query` (845 passed, 2 skipped) and Clippy for both crates.

src/query/src/dist_plan/commutativity.rs
```diff
@@ -396,6 +398,38 @@ mod tests {
...
+            .unwrap();
+        let partition_cols = BTreeMap::from([(
+            "partition".to_string(),
+            BTreeSet::from([Column::from_name("partition")]),
+        )]);
+        assert!(matches!(
+            Categorizer::check_plan(&aggregate, Some(partition_cols)).unwrap(),
+            Commutativity::TransformedCommutative { .. }
+        ));
+    }
```
evenyag: I think we can remove this test.

evenyag (approved):

killme2008 (approved):
LGTM

I think we need to:
* Refactor the PromQL planner
* Add new sqlness tests for native hisgogram

src/query/src/dist_plan/commutativity.rs
shuiyisong: Removed in https://github.com/GreptimeTeam/greptimedb/pull/8798/commits/b7397373140d1102c1b18d6500fe451eafa7ccbb

PR #8808, opened 2026-08-09
fix(mito): re-encode bulk WAL entry after filling missing columns

### LINKED ISSUE #8806: Bulk insert with a stale schema diverges WAL from memtable: rows lost or region open hangs on replay
### What type of bug is this?

Data corruption, Crash

Storage Engine, Datanode

### Minimal reproduce step

1. Create a mito region, then add a column (region schema version 0 -> 1).
2. Send a bulk insert (`RegionBulkInsertsRequest`) whose batch and `raw_data` were built against schema version 0, i.e. `aligned_schema_version` doesn't match the region's schema version — like a frontend with a stale schema cache keeps writing right after the `ADD COLUMN`. The write succeeds: the worker fills the missing column via `BulkPart::fill_missing_columns` before writing to the memtable.
3. Restart the datanode before the region flushes, so the bulk entry is replayed from the WAL on region open.

Root cause chain (code refs at main `1f1c9270a`):

- `BulkPart::fill_missing_columns` (`src/mito2/src/memtable/bulk/part.rs`) replaces `self.batch` but keeps `self.raw_data` (the original Arrow Flight IPC bytes).
- `BulkWalEntry::try_from(&BulkPart)` prefers `raw_data`, so the WAL entry keeps the pre-fill bytes while the memtable receives the filled batch (`RegionWriteCtx::push_bulk`).
- On replay, `replay_memtable` (`src/mito2/src/region/opener.rs`) decodes the stale IPC and writes a batch that misses the added column into a memtable built from the current metadata.

Since an alter flushes non-empty memtables first, these stale bulk entries are the only replayable WAL entries that can miss columns — but they are enough to lose data on every restart that replays them.

A self-contained reproduction test (alter -> stale bulk insert -> reopen, for both formats) is included in the fix PR: `engine::alter_test::test_bulk_insert_stale_schema_wal_replay`.

The rows written by the bulk insert survive the restart: WAL replay restores exactly what the memtable contained before the crash.

### What did you see instead?

Two failure modes depending on the memtable kind:

- Bulk memtable (flat format): `convert_bulk_part` fails with `ColumnNotFound` for the added column, but during replay the error is swallowed by the no-op write notifier (`OptionOutputTx::none()`) — the region opens "successfully" and the rows are **silently lost**, with no log at all.
- Time series memtable: `BulkPart::to_mutation` builds rows shorter than the declared schema; the replay **panics** the region worker (index out of bounds in the `KeyValue` accessors) and the region open hangs.

Any (code-level issue, verified on macOS ARM)

main (`1f1c9270a`); affected since v1.1.0 (stale-schema bulk fill path from #8222; `BulkPart::fill_missing_columns` exists since v1.0.0 via #7293)

```bash
thread 'global-worker' panicked at src/mito-codec/src/key_values.rs:220:37:
index out of bounds: the len is 3 but the index is 3
```

(from the reproduction test, time series memtable variant; the flat-format variant produces no log output — the rows just disappear after replay)

fengjiachun:

Closes #8806

Fixes a WAL/memtable divergence for bulk inserts that carry a stale schema,
which silently loses the rows (bulk memtable) or hangs the region open
(time series memtable) on WAL replay. See #8806 for details.

**The bug**: when a bulk insert request's schema version is behind the
region's (e.g. a frontend with a cached schema keeps writing after an
`ADD COLUMN`), the write worker fills the missing columns via
`BulkPart::fill_missing_columns`. The method replaces `batch` but keeps
`raw_data` (the original Arrow Flight IPC bytes), while
`BulkWalEntry::try_from(&BulkPart)` prefers `raw_data`. So the memtable gets
the filled batch but the WAL records the pre-fill bytes.

On replay (region open / catchup), the decoded batch misses the filled
columns:

- Bulk memtable (flat format): `convert_bulk_part` fails with
  `ColumnNotFound`; the error is swallowed by the no-op write notifier
  (`OptionOutputTx::none()`), so the region opens "successfully" and the rows
  are **silently lost**.
- Time series memtable: `BulkPart::to_mutation` builds rows shorter than the
  declared schema, and replay **panics** the region worker
  (`key_values.rs` index out of bounds), hanging the region open.

Since an alter always flushes non-empty memtables first, these stale bulk
entries are the only replayable WAL entries that can miss columns.

**The fix**:

1. `fill_missing_columns` clears `raw_data` after replacing the batch, so the
   WAL entry is re-encoded from the filled batch and WAL/memtable stay
   consistent.
2. `replay_memtable` fills missing columns for replayed bulk parts
   (dense-encoding regions only), so WAL entries already written by affected
   releases replay correctly instead of losing rows. Sparse batches are
   skipped: their tag columns are encoded in `__primary_key` by design and
   `convert_bulk_part` doesn't look them up.

Tests:

- `test_fill_missing_columns_resets_raw_data`: unit test for the raw_data
  invariant and the WAL entry round trip.
- `test_bulk_insert_stale_schema_wal_replay`: end-to-end
  alter -> stale bulk insert -> reopen -> data survives, for both formats.
- `test_bulk_insert_stale_wal_entry_replay`: injects a pre-fix-format WAL
  entry directly into the log store and verifies replay fills it
  (compatibility with WALs written by affected versions), for both formats.

## PR Checklist

### DIFF
--- a/src/mito2/src/engine/alter_test.rs
+++ b/src/mito2/src/engine/alter_test.rs
@@ -17,21 +17,26 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Duration;
 
+use api::v1::bulk_wal_entry::Body;
 use api::v1::helper::{row, tag_column_schema};
 use api::v1::value::ValueData;
-use api::v1::{ColumnDataType, Row, Rows, SemanticType, Value};
+use api::v1::{ArrowIpc, BulkWalEntry, ColumnDataType, Row, Rows, SemanticType, Value, WalEntry};
 use common_error::ext::ErrorExt;
 use common_meta::ddl::utils::{parse_column_metadatas, parse_manifest_infos_from_extensions};
-use common_recordbatch::RecordBatches;
+use common_recordbatch::{DfRecordBatch, RecordBatches};
+use common_test_util::flight::encode_to_flight_data;
 use datafusion_expr::col;
+use datatypes::arrow::array::{ArrayRef, Float64Array, StringArray, TimestampMillisecondArray};
+use datatypes::arrow::datatypes::{DataType, Field, Schema, TimeUnit};
 use datatypes::prelude::ConcreteDataType;
 use datatypes::schema::{ColumnSchema, FulltextAnalyzer, FulltextBackend, FulltextOptions};
+use store_api::logstore::provider::Provider;
 use store_api::metadata::ColumnMetadata;
 use store_api::metric_engine_consts::TABLE_COLUMN_METADATA_EXTENSION_KEY;
 use store_api::region_engine::{RegionEngine, RegionManifestInfo, RegionRole};
 use store_api::region_request::{
-    AddColumn, AddColumnLocation, AlterKind, PathType, RegionAlterRequest, RegionOpenRequest,
-    RegionRequest, SetIndexOption, SetRegionOption,
+    AddColumn, AddColumnLocation, AlterKind, PathType, RegionAlterRequest,
+    RegionBulkInsertsRequest, RegionOpenRequest, RegionRequest, SetIndexOption, SetRegionOption,
 };
 use store_api::storage::{ColumnId, RegionId, ScanRequest};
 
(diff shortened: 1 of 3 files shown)


v0y4g3r (approved):
LGTM

PR #8810, opened 2026-08-09
fix(mito2): avoid region worker panic when building a WAL entry fails

fengjiachun:

None

Fixes a latent panic in the region worker's write path.

In `handle_write_requests`, when `RegionWriteCtx::add_wal_entry` fails (i.e. `LogStore::entry` returns an error), the worker only calls `set_error` on the context, which stays in `region_ctxs`. The failed region's entry is not added to the batch, so when `write_to_wal` succeeds, `response.last_entry_ids` has no entry for that region and the success branch panics at `response.last_entry_ids.get(region_id).unwrap()`, killing the whole region worker (all regions on that worker stop serving writes). When the failed region is the only region in the batch, the batch is empty and `append_batch` returns an empty response by contract, so the panic is guaranteed. The existing `// Safety:` comment only holds for regions whose entries actually made it into the batch.

No in-tree log store can fail to build an entry at runtime today (`provider_from_wal_options` validates the provider/log store combination when the region opens; the raft-engine and noop `entry()` implementations are infallible), so this cannot be triggered from user input on current main. But `LogStore::entry` is a fallible public API, and any implementation that starts returning errors turns this into a guaranteed worker crash.

Changes:

- Skip contexts already marked as failed when updating next entry ids after a successful WAL write; their waiters are already notified with the error, and other regions in the batch are unaffected.
- Extract the WAL phase of `handle_write_requests` into a `write_wal` function so it can be unit tested (behavior preserved).
- Add unit tests with a mock log store covering: one region fails to build its entry while another succeeds, all regions fail (empty batch), and the whole batch write fails. The first two panic without this fix.

## PR Checklist

### DIFF
--- a/src/mito2/src/region_write_ctx.rs
+++ b/src/mito2/src/region_write_ctx.rs
@@ -218,9 +218,20 @@ impl RegionWriteCtx {
         self.failed = true;
     }
 
+    /// Returns whether the write operation is already marked as failed.
+    pub(crate) fn is_failed(&self) -> bool {
+        self.failed
+    }
+
     /// Updates next entry id.
     pub(crate) fn set_next_entry_id(&mut self, next_entry_id: EntryId) {
         self.next_entry_id = next_entry_id
+    }
+
+    /// Returns the next entry id to write.
+    #[cfg(test)]
+    pub(crate) fn next_entry_id(&self) -> EntryId {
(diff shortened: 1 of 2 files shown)


evenyag:
@codex review

evenyag (approved):

PR #8812, opened 2026-08-09
perf: cache bulk-memtable unordered-part sort (-54~-80% on unflushed reads)

discord9:
## Summary

Cache the sorted result of the bulk-memtable unordered part so repeated scans with no intervening writes skip the per-scan concat+sort.

`BulkMemtable::ranges()` re-ran `concat_and_sort` on the unordered part on **every scan**, even when no new rows had arrived. Now the sorted result is cached on `UnorderedPart` and invalidated on `push`/`clear`, with `max/min timestamp` + `max sequence` stats cached alongside. Unit test covers cache hit, push invalidation, and sorted output.

### Perf (case `memtable_unflushed_read`, 1.98M rows pre-flush, base vs candidate)

| Query | Base median | Candidate median | Delta |
|---|---|---|---|
| count_all | 157.8ms | 68.2ms | **-57%** |
| count_time_filtered | 114.0ms | 52.7ms | **-54%** |
| avg_by_host | 1016.8ms | 469.2ms | **-54%** |
| min_over_time | 7443ms | 1481ms | **-80%** |
| avg_over_time | 2199ms | 1478ms | **-33%** |
| seed_memtable_fragments | 876ms | 311ms | **-65%** |

All 6 queries improve consistently. (This case has run-to-run variance — a same-binary control also moved — but the direction is uniform and the magnitude large.)

## Files
- `src/mito2/src/memtable/bulk.rs` (+test)
- `src/mito2/src/memtable/bulk/part.rs`

## Testing
- `cargo nextest run -p mito2`: 171/171 passed

## Checklist
- [ ] CLA signed
- [ ] Perf evidence attached (see table above)


discord9:
Superseded by #8821 (consolidated scan read-path optimizations PR).

PR #8814, opened 2026-08-09
perf: parallel file scanning in UnorderedScan (1.97x micro) + bound pruner worker pool

discord9:
## Summary

Parallelize file scanning in UnorderedScan (append-mode scans) and bound the pruner worker pool.

### SC10 — Parallel file scanning in UnorderedScan

UnorderedScan scanned file ranges **strictly sequentially** per part-range. Now per-file sources are built in parallel behind a bounded semaphore (mirroring SeqScan's existing pattern) and re-ordered into `ordered_sources`, preserving row-group order. Append mode has no dedup, so row-group order is all that matters — output is identical. Memory stays bounded via the existing channel-size logic (clamp 2..64).

**Micro-bench** (in `unordered_scan` test, 16 files x 4096 rows):

| | sequential | parallel | speedup |
|---|---|---|---|
| run 1 | 40.15ms | 21.3ms | **1.88x** |
| run 2 | 40.15ms | 20.40ms | **1.97x** |

**End-to-end** (query-regression harness, `sql_limit_unordered_scan` case — 128 SSTs append-mode):

| Query | run 1 | run 2 (swapped control) |
|---|---|---|
| select_ts_limit_100 | **-47.1%** | **-38.1%** |
| min_max_ts | -18.8% | +7.5% (noise) |
| select_ts_scan_all | +19.2% | -27.9% (run-order bias; geomean ~ -6%) |

### LC1 — Bound pruner worker pool

`Pruner::new` spawned `get_total_cpu_cores()` worker tasks per scan. Now clamped to `min(num_workers, num_files, MAX_PRUNE_WORKERS=16)` and skipped entirely when `num_files == 0`. Workers are stateless w.r.t. files and dispatch is `file_id_hash % num_workers`, so clamping preserves same-worker serialization for concurrent prunes of the same file. Cuts per-scan task churn on high-core machines.

## Why this is safe despite scan-parallelism history

Parallel scan was **tried before and removed** (see below). This change deliberately avoids the failure modes of that attempt:

- `#3992` (848bd7e553) first implemented a row-group-level parallel unordered scanner.
- `#5090` (2fcb95f50a, `fix!`) fixed a regression caused by **unbalanced partitions and splitting ranges** under that parallelism; the fix included `c3af2b1f90 "remove scan_parallelism"` (2024-12), which removed the scan-parallelism config entirely and left UnorderedScan **sequential** while SeqScan kept bounded parallel sources.
- This change reuses the **exact pattern SeqScan already ships** (`create_parallel_flat_sources` + `file_scan_semaphore` + `ordered_sources` re-ordering), rather than reintroducing a new parallelism knob:
  - concurrency is **bounded** by a semaphore (no unbounded fan-out → no partition imbalance storm),
  - output is **re-ordered into row-group order** (append mode has no dedup, so row-group order is the full contract),
  - only the **UnorderedScan (append-only, no ordering guarantee)** path is touched — SeqScan/SeriesScan semantics are unchanged.

## Files
- `src/mito2/src/read/unordered_scan.rs` (+ micro-bench test)
- `src/mito2/src/read/pruner.rs`

## Testing
- mito2 full suite: 1107 passed (1 flaky retried green)

## Checklist
- [ ] CLA signed
- [ ] Perf evidence attached (micro 1.97x + end-to-end limit -38~-47%)

### DIFF
--- a/src/mito2/src/read/pruner.rs
+++ b/src/mito2/src/read/pruner.rs
@@ -38,6 +38,15 @@ use crate::sst::parquet::reader::ReaderMetrics;
 /// Number of files to pre-fetch ahead of the current position.
 const PREFETCH_COUNT: usize = 8;
 
+/// Maximum number of pruner worker tasks spawned per scan.
+///
+/// Callers pass the total CPU core count, which would spawn one task per core
+/// even for scans touching a handful of files. The worker count only affects
+/// the per-scan task churn and the `file_id_hash % num_workers` dispatch
+/// granularity, so a modest cap is enough while keeping the file -> worker
+/// mapping consistent for the pruner's lifetime.
+const MAX_PRUNE_WORKERS: usize = 16;
+
 /// Local pruner in a partition that supports prefetching files to prune.
 pub struct PartitionPruner {
     pruner: Arc<Pruner>,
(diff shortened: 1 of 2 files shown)


discord9:
Superseded by #8821 (consolidated scan read-path optimizations PR, includes the scan-parallelism history note).

Issue #8805, opened 2026-08-08
Speed up Prometheus JSON response building: ryu float formatting and per-series entry reuse


lyang24:
### What type of enhancement is this?

Performance

### What does the enhancement do?

Reduces CPU spent building Prometheus HTTP API JSON responses
(`PrometheusJsonResponse::record_batches_to_data` in
`src/servers/src/http/result/prometheus_resp.rs`).

In a 60s CPU profile of concurrent Prometheus `query_range` workloads
(4 workers, 800k series, ~1.76 GB of JSON returned across 91 requests,
individual responses up to ~50 MB), `record_batches_to_data` accounts for
**5.04%** of total CPU. Almost all of it comes from two lines:

1. **`f64::to_string()` per sample — 47% of the function (2.4% of total CPU).**
   Prometheus API sample values are JSON strings, so every data point is
   formatted through Rust std's `Display` path (Grisu + `fmt::Formatter`
   machinery). Replacing it with the `ryu` crate (Ryū: shortest round-trip
   float formatting, same algorithm class as ClickHouse's dragonbox/zmij)
   typically gives 3-5x on this operation.

2. **`IndexMap::entry(Vec<(&str, &str)>)` per row — 32% of the function
   (1.6% of total CPU).** Every row builds a label vector and hashes it with
   the default SipHash hasher to find its series entry. Query output is
   almost always clustered by series, so consecutive rows hit the same entry:
   remember the previous row's tags and entry index and reuse it when equal,
   falling back to the map lookup on change (worst case adds one Vec
   comparison). Optionally switch the map to a faster hasher (e.g. ahash).

Both changes are on the serial per-request path, so they also reduce tail
latency for large responses.

ClickHouse formats floats in all text/JSON output through a shortest
round-trip algorithm plus integer fast paths, instead of general-purpose
formatting machinery:

- Entry point `writeFloatText`:
  https://github.com/ClickHouse/ClickHouse/blob/a8a51ccbadbca7d5b8dd8e71ceae0d735e73baf5/src/IO/WriteHelpers.h#L143
- Fast-path dispatch (`writeFloatTextFastPath`): exact integers via `itoa`,
  everything else via `zmij` (`zmij::detail::write` at line 334):
  https://github.com/ClickHouse/ClickHouse/blob/a8a51ccbadbca7d5b8dd8e71ceae0d735e73baf5/src/IO/WriteHelpers.cpp#L314-L341
[...]

PR #8815, opened 2026-08-10
perf(servers)!: speed up Prometheus JSON response building with ryu and per-series entry reuse

discord9:

- https://github.com/GreptimeTeam/greptimedb/issues/8805

This PR reduces the cost of building Prometheus HTTP API JSON responses.

Finite sample values now use ryu's shortest-roundtrip formatting. PromQL range output is normally clustered by series, so `PrometheusJsonResponse::record_batches_to_data` also reuses the preceding `IndexMap` entry when consecutive rows have the same labels, avoiding repeated hashing and lookup. Series transitions continue to use the normal map-entry path.

**BREAKING CHANGE:** finite Prometheus sample strings may change textually, including integral `.0`, scientific-notation thresholds, and shortest-roundtrip tie cases. Parsing the new text as `f64` yields the identical value, including the sign of `-0.0`. `NaN`, `inf`, and `-inf` retain their previous strings.

The PR also:

- adds a `prom_http` query-regression kind for the real `/v1/prometheus/api/v1/query_range` response path;
- sends the database through the endpoint's `db` URL query parameter and covers it with a focused test;
- documents the fixture DSL and its end-to-end timing scope;
- uses 256 distinct series in the benchmark fixture;
- aligns direct-SST fixture `CREATE TABLE` index declarations with generated region metadata.

### Performance

The formal A/B run used nightly-profile binaries with 256 distinct series, 481 evaluation timestamps per series, 3 warmups, and 9 measured iterations. Both sides use ryu; the comparison isolates per-series entry reuse. Latency includes the endpoint request, query execution, Prometheus JSON response construction, body read, and JSON parsing.

| Query | Base median | Candidate median | Delta |
| --- | ---: | ---: | ---: |
| `prom_range_2h` | 68.793 ms | 55.29 ms | **-19.62%** |
| `tql_range_2h_control` | 66.83 ms | 67.07 ms | +0.36% |

Both configured 10% maximum-regression thresholds passed with zero validation errors on each target.

### Verification

- Targeted `servers` Prometheus response tests: 12/12 passed.
- Targeted `cmd` Prometheus database URL test: 1/1 passed.
- Cargo checks passed for `servers`, `cmd`, and `tests-integration`.
- The normalized fixture plan completed successfully.
- The formal query-regression run completed successfully and passed both thresholds.

There are no schema or persisted-data compatibility changes. The finite Prometheus sample textual representation is intentionally breaking as described above.

## PR Checklist

### DIFF
--- a/src/cmd/src/bin/query_regression_runner/measure.rs
+++ b/src/cmd/src/bin/query_regression_runner/measure.rs
@@ -199,7 +199,9 @@ async fn run_target(
             let mut sample = post_query(client, port, query, db).await;
             let execution_time = sample
                 .get("response")
-                .and_then(extract_execution_time)
+                .and_then(|response| {
+                    extract_execution_time_for_kind(query.kind.as_deref(), response)
+                })
                 .cloned()
                 .unwrap_or(Value::Null);
             sample
@@ -263,6 +265,14 @@ fn response_text(body: &Value) -> String {
     body.as_str()
         .map(ToOwned::to_owned)
         .unwrap_or_else(|| serde_json::to_string(body).unwrap_or_default())
+}
+
+fn extract_execution_time_for_kind<'a>(kind: Option<&str>, body: &'a Value) -> Option<&'a Value> {
+    if kind == Some("prom_http") {
+        None
+    } else {
+        extract_execution_time(body)
+    }
 }
 
 fn extract_execution_time(body: &Value) -> Option<&Value> {
@@ -399,6 +409,24 @@ mod tests {
         assert_eq!(
             extract_execution_time(&json!({"output": [{"elapsed": 3}]})),
             Some(&json!(3))
+        );
+    }
+
+    #[test]
+    fn prom_http_does_not_extract_execution_time_from_response() {
+        let response = json!({
+            "data": {
+                "result": [{"metric": {"job": "api", "elapsed": "label"}}],
+                "elapsed": 42
+            }
+        });
+        assert_eq!(
+            extract_execution_time_for_kind(Some("prom_http"), &response),
+            None
+        );
+        assert_eq!(
+            extract_execution_time_for_kind(Some("sql"), &response),
+            Some(&json!(42))
         );
     }
 
--- a/src/cmd/src/bin/query_regression_runner/measure.rs
+++ b/src/cmd/src/bin/query_regression_runner/measure.rs
@@ -21,7 +21,7 @@ use serde_json::{Map, Value, json};
 
 use crate::query_regression_runner::model::{Measurement, Query, QueryResult, Scenario, Table};
 use crate::query_regression_runner::plan::{load_plan, normalize_scenario};
-use crate::query_regression_runner::sql::{http_post_sql, sql_ident};
+use crate::query_regression_runner::sql::{http_post_prom_range_query, http_post_sql, sql_ident};
 use crate::query_regression_runner::{MeasureArgs, Result};
 
 pub(super) async fn run_measure(args: MeasureArgs) -> Result<()> {
@@ -106,6 +106,26 @@ fn target_report(name: &str, http_port: u16, result: QueryResult) -> Value {
     })
 }
 
+/// Routes a configured query to the right endpoint: `prom_http` queries hit
+/// the Prometheus HTTP range API (which exercises the Prometheus JSON response
+/// builder), everything else goes through `/v1/sql` (including `TQL ANALYZE`).
+async fn post_query(client: &Client, port: u16, query: &Query, db: &str) -> Value {
+    if query.kind.as_deref() == Some("prom_http") {
+        http_post_prom_range_query(
+            client,
+            port,
+            &query.query,
+            query.start.as_deref(),
+            query.end.as_deref(),
+            query.step.as_deref(),
+            db,
+        )
+        .await
+    } else {
+        http_post_sql(client, port, &query.query, db).await
+    }
+}
+
 async fn run_target(
     port: u16,
     tables: &[Table],
@@ -118,6 +138,9 @@ async fn run_target(
             name: Some("count_all".to_string()),
             kind: Some("sql".to_string()),
             query: format!("SELECT count(*) FROM {}", sql_ident(&tables[0].name)),
+            start: None,
+            end: None,
+            step: None,
             warmup: 0,
             iterations: 1,
             thresholds: Map::new(),
@@ -147,7 +170,7 @@ async fn run_target(
         }
         validation.push(sample);
     }
-    let first = http_post_sql(client, port, &queries[0].query, db).await;
+    let first = post_query(client, port, &queries[0], db).await;
     if !first["ok"].as_bool().unwrap_or(false) {
         validation_errors.push(json!({
             "sql": queries[0].query,
@@ -160,7 +183,7 @@ async fn run_target(
     let mut measurements = Vec::with_capacity(queries.len());
     for query in &queries {
         for _ in 0..query.warmup {
-            let warmup = http_post_sql(client, port, &query.query, db).await;
+            let warmup = post_query(client, port, query, db).await;
             if !warmup["ok"].as_bool().unwrap_or(false) {
                 validation_errors.push(json!({
                     "sql": query.query,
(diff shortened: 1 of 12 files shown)


tests-integration/tests/grpc.rs
```diff
@@ -874,7 +874,7 @@ pub async fn test_prom_gateway_query(store_type: StorageType) {
                 ]
                 .into_iter()
                 .collect(),
-                value: Some((5.0, "1".to_string())),
+                value: Some((5.0, "1.0".to_string())),
```
fengjiachun, line 877: These four expectations (877, 887, 939, 949) still carry the `"1.0"` / `"2.0"` values introduced in 2cba61c, but the follow-up fix a7f7c7f (strip ryu's `.0` suffix) only touched `prometheus_resp.rs` and never reverted them.

`format_prometheus_sample_value(1.0)` now returns `"1"` — as this PR's own `format_prometheus_sample_value_matches_prometheus_wire_format` test asserts — so `test_prom_gateway_query` fails on all four. They should go back to `"1"` / `"2"`.

tests-integration/tests/http.rs
```diff
@@ -907,7 +907,7 @@ pub async fn test_prom_http_api(store_type: StorageType) {
     assert_eq!(
         body.data,
         serde_json::from_value::<PrometheusResponse>(
-            json!({"resultType":"scalar","result":[1.0,"2"]})
+            json!({"resultType":"scalar","result":[1.0,"2.0"]})
```
fengjiachun, line 935: Same stale expectation as `grpc.rs`. The scalar branch of `record_batches_to_data` pops from `samples.values`, which are formatted by `format_prometheus_sample_value(2.0)` → `"2"`, so this assertion fails. Should be reverted to `[1.0,"2"]`.

tests/perf/query_cases/prom_json_response/case.toml
```diff
@@ -0,0 +1,97 @@
...
+name = "host"
+type = "STRING"
+semantic = "tag"
+distribution = { kind = "cardinality", values = 16, prefix = "host" }
+
+[[scenario.tables.columns]]
+name = "instance"
+type = "STRING"
+semantic = "tag"
+distribution = { kind = "cardinality", values = 16, prefix = "instance" }
```
fengjiachun: `tag_value` in `query_perf_fixture::direct_sst` computes `format!("{prefix}{}", series % values)` from the *same* series index for every tag column. With `values = 16` on both `host` and `instance`, the two tags are perfectly correlated — slots `s` and `s + 16` produce identical labels.

So `series_count = 256` yields only **16** distinct series. The case measures ~16 × 481 ≈ 7.7k samples per request, not the ~123k stated in the header comment, and the reported −26.8% was obtained on that 16×-smaller fixture.

`promql_range_boundary` avoids this by giving one tag a cardinality equal to `series_count` (128). Setting `values = 256` on `instance` here would do the same.

src/servers/src/http/result/prometheus_resp.rs
```diff
@@ -70,6 +70,31 @@ fn prometheus_native_histogram(histogram: &NativeHistogram) -> Result<PromNative
...
+/// significantly faster than `f64::to_string()` and is the hot path when
+/// building large responses. ryu prints integral values with a trailing ".0"
+/// ("1.0"), but the Prometheus wire format (like std `f64::to_string()`)
+/// expects "1", so the ".0" suffix is stripped. ryu only emits ".0" for
+/// integral values (its scientific notation never ends in ".0"), so
+/// non-integral values, negative zero (`-0.0` -> `"-0"`) and extreme values
+/// (`1e30`, `f64::MAX`) are unaffected. Non-finite values (`NaN`, `+Inf`,
+/// `-Inf`) are not supported by `ryu::Buffer::format_finite`, so they keep the
+/// previous `f64::to_string()` output (`"NaN"`, `"inf"`, `"-inf"`).
+fn format_prometheus_sample_value(value: f64) -> String {
```
fengjiachun, line 79: ryu switches to scientific notation at different thresholds than `f64::to_string()`, which never used it at all — so this changes the wire format for extreme magnitudes, not just the `.0` suffix. Verified against ryu 1.0.23:

| value | before | after | Prometheus |
|---|---|---|---|
| `1.7e18` | `1700000000000000000` | `1.7e18` | `1700000000000000000` |
| `1e16` | `10000000000000000` | `1e16` | `10000000000000000` |
| `2.3e-6` | `0.0000023` | `2.3e-6` | `0.0000023` |

ryu's cutoffs are `|v| >= 1e16` and `|v| < 1e-5`; upstream Prometheus (`jsonutil.MarshalFloat`) uses `1e21` and `1e-6`. Numeric parsers are unaffected, but the doc comment above and the PR description both state the output is unchanged, and this widens the gap with Prometheus rather than closing it.

Worth either documenting the divergence explicitly, or falling back to `to_string()` outside ryu's decimal range so the hot path (ordinary sample values) still benefits.

src/cmd/src/bin/query_regression_runner/sql.rs
```diff
@@ -117,10 +117,54 @@ pub(super) fn value_f64(value: Option<&Value>) -> Option<f64> {
...
+) -> Value {
+    let mut sample = post_form(
+        client,
+        format!("http://127.0.0.1:{port}/v1/prometheus/api/v1/query_range"),
+        &[
+            ("query", query),
+            ("start", start.unwrap_or_default()),
+            ("end", end.unwrap_or_default()),
+            ("step", step.unwrap_or_default()),
+            ("db", db),
```
fengjiachun: `range_query` reads `db` only from `Query(params)` (`src/servers/src/http/prometheus.rs` L534) — it has no `.or(form_params.db)` fallback, unlike `/v1/sql` which does `query_params.db.or(form_params.db)`. The auth middleware (`extract_catalog_and_schema`) only looks at the `X-Greptime-DB-Name` header and the URI query string.

So this form field is silently dropped and the request runs against `public`. It happens to work here because the case sets `database = "public"`, but any other schema would benchmark the wrong table or fail with table-not-found. Passing `db` in the URL query string instead would fix it.

killme2008:
It's the idea from @lyang24 in #8805 

Could you help to review this PR? Thanks

src/servers/src/http/result/prometheus_resp.rs
```diff
@@ -380,15 +415,33 @@ impl PrometheusJsonResponse {
...
+                    // from `buffer`, so the captured index stays valid.
+                    let index = last_entry_index
+                        .expect("last entry index is always set together with last tags");
+                    &mut buffer
+                        .get_index_mut(index)
+                        .expect("reused series entry must exist")
+                        .1
+                } else {
+                    // Tags changed: fall back to the map lookup.
+                    let entry = buffer.entry(tags.clone());
```
lyang24: I think we can avoid the expect and clone here
```
  let reuse = last_entry_index
      .filter(|index| {
          buffer
              .get_index(*index)
              .is_some_and(|(key, _)| key == &tags)
      });

  let samples = if let Some(index) = reuse {
      buffer
          .get_index_mut(index)
          .map(|(_, samples)| samples)
          .with_context(|| UnexpectedResultSnafu {
              reason: format!("cached series entry index {index} is missing"),
          })?
  } else {
      let entry = buffer.entry(tags);
      last_entry_index = Some(entry.index());
      entry.or_default()
  };
```

lyang24: any non-clustered output—the fast path never hits so every row pays for both `tags.clone()` and a failed comparison. The original `buffer.entry(tags)` path did not have this additional allocation

lyang24 (approved):
lgtm

src/cmd/src/bin/query_regression_runner/measure.rs
```diff
@@ -173,7 +196,7 @@ async fn run_target(
         let mut samples = Vec::with_capacity(query.iterations);
         let mut good_latencies = Vec::with_capacity(query.iterations);
         for _ in 0..query.iterations {
-            let mut sample = http_post_sql(client, port, &query.query, db).await;
+            let mut sample = post_query(client, port, query, db).await;
```
lyang24, line 199: Can we skip extract_execution_time for prom_http queries? Prometheus API responses do not contain these SQL/TQL timing fields, so recursively searching the response is unnecessary and could accidentally match a metric label named elapsed.

PR #8816, opened 2026-08-10
feat: otlp duration_nano and trace_flag signed integer coercion

### LINKED ISSUE #8793: Transition built-in data models from unsigned to signed integers (Postgres/JDBC/Spark-Parquet compatibility)

Refactor, Tech debt reduction

Transition the **built-in user-visible data models** (the predefined schemas GreptimeDB auto-creates when ingesting from observability protocols — OTLP traces, OTLP logs, native histograms, and the Prometheus/metric-engine fields users query) from **unsigned** to **signed** integer types, while keeping existing on-disk user data readable.

This is scoped to the user-visible built-in schemas only; metric-engine *internal* columns (`__tsid`, `__table_id`) are explicitly out of scope, and user-authored `BIGINT UNSIGNED` columns are unaffected.

#### Why signed is the right default for built-in models

The unsigned types GreptimeDB chose for its built-in models create friction at nearly every interop boundary the product cares about, for little benefit (these are non-negative counts/durations that fit comfortably in signed ranges):

- **PostgreSQL has no unsigned integer types.** `smallint` / `integer` / `bigint` are all signed, and there is no `BIGINT UNSIGNED`. Since GreptimeDB targets Postgres-compatibility (wire protocol, SQL surface, tooling/ORMs), unsigned built-in columns are a compat gap — e.g. a user querying the auto-created trace table over the Postgres protocol gets a type their Postgres-native tooling doesn't expect.
- **Java / the JVM ecosystem has no unsigned primitives.** `byte`, `short`, `int`, `long` are all two's-complement signed, and `java.sql.Types` has no `*_UNSIGNED` variants (only `INTEGER`, `BIGINT`, `SMALLINT`, `TINYINT`). So any JDBC client — JDBC drivers, Hibernate, MyBatis, jOOQ, Spark-on-JVM, Flink, Kafka-Connect JDBC sinks — reading a built-in `BIGINT UNSIGNED` column either errors, falls back to `BigDecimal`/`String`, or silently misinterprets the high bit.
- **Spark / Parquet logical types are signed-only.** Parquet's logical-type annotations are `INT(64)`, `INT(32)`, … — all signed. Unsigned is only expressible via the legacy `ConvertedType` (`UINT_64`) or logical-type metadata, which Spark's Parquet reader largely ignores: Catalyst's `LongType`/`IntegerType` are signed, so unsigned Parquet columns are read as signed and clamp/misorder. The same applies to most Parquet/Iceberg/Delta consumers (DuckDB handles it; many others don't). When users export GreptimeDB data to Spark/Iceberg, unsigned built-in columns silently break.
- **The wider SQL/analytics ecosystem is signed.** The SQL standard defines only signed `INTEGER`/`BIGINT`; Snowflake, BigQuery, Redshift, SingleStore all use signed `NUMBER`/`INT64`. MySQL is the notable exception (it has unsigned ints), which is partly why GreptimeDB supports them — but MySQL-compat is a client concern, not a reason to make the *built-in observability* models unsigned.
- **Protobuf/OTLP do carry `uint64`/`uint32` on the wire** — so ingestion legitimately receives unsigned values — but that is a *wire* concern decoupled from storage. Storing them as signed is lossless for all realistic values and removes the downstream friction above.

Net: choosing signed for built-in models aligns GreptimeDB with the ecosystems its data flows into, without giving up anything the observability use case needs.

#### Inventory of unsigned integers in built-in models today

**A. Value columns (safe to convert — not part of any primary key):**

| Model | Field | Type | Source |
|---|---|---|---|
| OTLP traces (v0 + v1) | `duration_nano` | `Uint64` | `src/servers/src/otlp/trace/v0.rs`, `src/servers/src/otlp/trace/v1.rs` (`field(DURATION_NANO_COLUMN, ColumnDataType::Uint64)`) |
| OTLP logs | `trace_flags` | `Uint32` | `src/servers/src/otlp/logs.rs` (`build_otlp_logs_identity_schema`) |
| OTLP logs | dynamic attribute `UInt64` path | `Uint64` | `src/servers/src/otlp/logs.rs` (`decide_column_schema_and_convert_value`, `jsonb_value_to_log_value`) |
| Native histogram struct | `count_u64` | `UInt64` | `src/common/query/src/native_histogram.rs` (`native_histogram_field_type`) |
| Native histogram struct | `zero_count_u64` | `UInt64` | same |
| Native histogram struct | `positive_span_lengths` | `List<UInt32>` | same |
| Native histogram struct | `negative_span_lengths` | `List<UInt32>` | same |

(OTLP trace semconv fixed columns — ports, `http.response.status_code`, `k8s.container.restart_count`, `db.operation.batch_size` — are *already* `Int64`, so no change needed there. OTLP metrics has no fixed unsigned columns.)

#### Mapping-layer audit (external-value → GreptimeDB-type boundaries)

For completeness, the other ingestion mapping layers were audited; most are already signed, so they need no change:

| Mapping | Behavior | Verdict |
|---|---|---|
| **OTLP value → type** | OTLP `AnyValue` has only `IntValue` (int64); mapped to `Number::Int64` (`src/servers/src/otlp/utils.rs`) and `ColumnDataType::Int64` | Already signed — no work |
| OTLP trace/span/resource attributes | `trace_value_datatype` maps `I64Value → Int64` (`src/servers/src/otlp/coerce.rs`) | Already signed — no work |
| **JSON → jsonb (pipeline)** | `VrlValue::Integer(i) => Number::Int64` (`src/pipeline/src/etl/transform/transformer/greptime.rs`); VRL `Integer` is `i64` | Already signed — no work |
| `jsonb Number::UInt64` origin | Magnitude-based — only from JSON numbers `> i64::MAX` or re-ingested uint data; surfaces only in the OTLP logs dynamic-attribute path above | Already in inventory (data-dependent, rare) |
| **InfluxDB line protocol** | `FieldValue::U64 → ColumnDataType::Uint64` (`src/servers/src/influxdb.rs`); preserves the source's explicit `u` suffix | **Out of scope** — user-chosen/source-preserving, like a user-authored `BIGINT UNSIGNED` |

There is **no auto-inference of uint from a value's magnitude anywhere** — a column's type is always declared in a schema or carried verbatim from the source format. Every uint producer is explicit and enumerable.

> **Supporting evidence for the rationale:** GreptimeDB already degrades `UInt64` columns to Postgres **`NUMERIC`** (`src/servers/src/postgres/types.rs`) because Postgres has no unsigned integer type — the exact interop friction this issue exists to remove, happening today for any uint column queried over the Postgres protocol.

**B. Internal key columns — OUT OF SCOPE (for reference only):**

`__tsid` (`uint64`) and `__table_id` (`uint32`) are metric-engine *internal* columns (`is_metric_engine_internal_column` in `src/store-api/src/metric_engine_consts.rs`) that are never exposed through any user-facing API. **They are explicitly excluded from this effort** — changing them is a persisted-format migration (see below) that buys nothing for the Postgres/JDBC/Spark interop this issue is about.

| Model | Field | Type | Source |
|---|---|---|---|
| Prometheus / metric engine | `__tsid` | `uint64` | `src/metric-engine/src/engine/create.rs` (`tsid_col()`); byte-encoded via `SortField` in `src/mito-codec/src/row_converter/sparse.rs` |
| Prometheus / metric engine | `__table_id` | `uint32` | `table_id_col()`; same codec |

### Implementation challenges

The core finding from the investigation: **"silent, read-compatible conversion" works for value columns but not for key columns.** The deciding factor is whether the column participates in sort ordering. Reads are manifest-driven (`src/mito2/src/sst.rs` builds the Arrow schema from `metadata.schema.arrow_schema()`; there is no global read-time type-normalization hook), and Parquet stores `UInt64`/`Int64` as the same physical `INT64` bytes — so value columns reinterpret correctly for free, while key columns cannot.

#### A. Value columns — silent read-compat is free, and NO `ALTER` is needed

For `duration_nano`, `trace_flags`, and the logs dynamic `UInt64` path:

- **Read path costs nothing.** Reads are manifest-driven (`src/mito2/src/sst.rs`), so old UInt64 tables read back as UInt64 and new Int64 tables read back as Int64. The two never collide because type is per-table.
- **No `ALTER TABLE` on upgrade.** The OTLP ingest path already reconciles every incoming column against the *existing* table schema and treats it as authoritative (`choose_trace_reconcile_decision` / `choose_fixed_trace_reconcile_decision` in `src/frontend/src/instance/otlp/trace_types.rs`). When the existing type can absorb the request via coercion, it returns `TraceReconcileDecision::UseExisting(existing_type)`, and `requires_alter()` is only ever true for the single Int64→Float64 widening case. So adding one coercion arm makes existing UInt64 tables keep receiving data and staying UInt64, byte-for-byte, forever:
  - `is_supported_trace_coercion(Int64, UInt64)` → `true` (in `src/servers/src/otlp/coerce.rs`)
  - `coerce_non_null_value(UInt64, Int64, I64Value(n))` → `U64Value(*n as u64)` (lossless — durations/flags are non-negative)
  - The logs path (`coerce_log_value_data`) gets the equivalent numeric identity handling instead of its stringify-on-mismatch fallback.
- **The cost of avoiding ALTER is type-tolerant readers**, not a data rewrite. Because old tables return UInt64 and new tables return Int64 for the same logical column, any code that hardcodes one type breaks on the other — e.g. the Jaeger handler's `duration_nano` `.as_u64()` (`src/servers/src/http/jaeger.rs`) must also accept `.as_i64()`. This is a small, bounded, permanent set of call sites, far cheaper than rewriting every existing trace/log table on upgrade. The fleet becomes heterogeneous (some tables Int64, some UInt64, same column name) — acceptable because every table is self-describing in its manifest.

#### B. Native histogram — left unchanged in this effort

The native-histogram struct is a value column, so the byte/sort argument doesn't block it. But "no `ALTER`, leave old data as-is" does *not* extend cleanly to it, because:
1. Recognition is by **exact struct equality** (`is_native_histogram_value_type`), not a per-table flag — an existing v1 (UInt64-subfield) table and a new v2 (Int64-subfield) ingest don't even agree they're the same kind of column.
2. There is **no subfield-level write coercion** today — the struct is handled atomically.

Converting it would require a dual-recognition shim (accept v1 *or* v2 across the ~8 call sites) plus subfield coercion, and the only value-preserving concern is that `count_u64`/`zero_count_u64` are monotonic counters that can in principle exceed `2^63`. **Decision: leave native histograms emitting their current v1 (UInt64-subfield) schema.** They sit behind `experimental_enable_prometheus_n = true` (experimental), so excluding them costs nothing real; revisit when the feature stabilizes and the struct-recognition question is worth solving on its own.

#### C. Internal key columns — out of scope (kept as unsigned)

For context: `__tsid` and `__table_id` are primary-key columns byte-encoded via `SortField` (`sparse.rs`). `SortField` serializes each type with an **order-preserving byte encoding** — big-endian for unsigned, "flip the sign bit then big-endian" for signed — **a different byte sequence** for any value with the high bit set (~half of all TSIDs, since TSID is a random hash). This breaks a hard engine invariant: every SST and the memtable must be individually sorted by the same collation so merges are correct; silently reinterpreting old unsigned-encoded files as signed would produce a mis-ordered stream on merge → duplicate/missing rows, and would require a persisted-format version bump + rewrite.

**Decision: leave them unsigned.** They are internal (`is_metric_engine_internal_column`), not exposed through any API, and carry no interop cost — none of the Postgres/JDBC/Spark/Parquet concerns motivating this issue reach them. Noting the technical reason here only so a future reader understands *why* the sweep stops at the built-in user-visible models.

#### Proposed transition (backward-compatible, no `ALTER`)

Guiding principle: **new tables write signed; existing tables are never mutated and stay exactly as written; reads are manifest-driven so they're already compatible.** No global "read all uint as int" shim — that would break user-authored `BIGINT UNSIGNED` tables and change SQL semantics. No `ALTER TABLE` runs automatically — old tables keep their UInt64 columns forever. **No config flag**: the flip is silent and transparent to data because every table carries its own type in the manifest; a flag would only add rollout/rollback control for an auto-created internal column, which isn't worth the complexity.

- **Phase 0 — Prerequisites (no behavior change).** Add the `Int64 → UInt64` coercion arm to `src/servers/src/otlp/coerce.rs` (`is_supported_trace_coercion` + `coerce_non_null_value`); add the equivalent numeric identity path to the logs `coerce_log_value_data`. This is the safety net that lets existing UInt64 tables keep accepting new Int64-ingested data without error and without alteration.
- **Phase 1 — Flip new tables to signed (silent, no flag).** `duration_nano`, `trace_flags`, and logs dynamic numerics emit signed types for *new* tables. Existing tables are untouched and keep their unsigned columns (handled by Phase 0). No config flag is needed: the change is transparent to data because every table is self-describing (old tables read/write UInt64, new tables read/write Int64, no cross-table conflict). The reader audit (Phase 2) must ship in the same release.
- **Phase 2 — Reader audit (ships with the flip).** Verify the bounded set of consumers of these columns tolerate both UInt64 (old tables) and Int64 (new tables). This is likely near-trivial: serde_json-based readers (e.g. Jaeger's `duration_nano` `.as_u64()`) are naturally tolerant because serde_json does not preserve signedness for in-range non-negative integers — only a *type-strict* Arrow/`ScalarValue` downcast to `UInt64Array`/`UInt64` would break, and those should be enumerated and fixed.
- **No Phase 3.** There is no flag to flip or deprecate. Existing tables remain UInt64 indefinitely (no forced migration); users may `ALTER` manually if they want uniformity. New tables are signed from the release that ships Phase 1+2.
- **Native histograms: not changed.** Keep emitting the current v1 (UInt64-subfield) schema (see section B).
- **`__tsid` / `__table_id`: not changed.** Internal, out of scope (see section C).

Because no persisted format is mutated and no `ALTER` runs, the persisted-format compatibility-test framework (`docs/rfcs/2025-07-04-compatibility-test-framework.md`) does not gate this effort — existing files are read exactly as written.

#### Suggested follow-ups / discussion

- Is the heterogeneous-fleet tradeoff (old tables UInt64, new tables Int64, indefinitely, for the same column) acceptable, or do we eventually want a *manual*, opt-in `ALTER` helper for users who want uniformity?
- The reader audit (Phase 2) should enumerate type-strict consumers up front. Most readers (serde-json-based like Jaeger) are naturally tolerant; the risk is only Arrow/`ScalarValue` downcasts that assume `UInt64`.
- Distributed caveat: frontends perform the OTLP reconciliation, so during a frontend rolling upgrade a brand-new table created by a new frontend (Int64) can't be written by an old frontend (which lacks Phase 0). Existing tables are unaffected. Handled by upgrading frontends as a tier, not by a flag.

sunng87:

Part of #8793 

this patch switches built-in opentelemetry log and trace table from unsigned integer to signed, adds a compatibility layer to make sure existing table will continue to work.

## PR Checklist


src/servers/src/otlp/trace/v0.rs
```diff
@@ -110,9 +110,9 @@ pub fn write_span_to_row(writer: &mut TableData, span: TraceSpan) -> Result<()>
         ),
         make_column_data(
             DURATION_NANO_COLUMN,
-            ColumnDataType::Uint64,
-            Some(ValueData::U64Value(
-                span.end_in_nanosecond - span.start_in_nanosecond,
+            ColumnDataType::Int64,
```
fengjiachun: v0 also switches to `Int64`, but the compat layer only runs for v1: `insert_trace_requests` gates reconciliation on `is_trace_v1_model` (`trace_ingest.rs:1284`), and v0 falls through to `handle_log_inserts`, which never retypes a column. Existing v0 tables then fail every write in mito's `check_schema`:

```
column duration_nano expect type UInt64, given: INT64
```

Either route v0 through the same coercion, or keep `v0.rs` on `Uint64` since v0 is frozen. `test_otlp_traces_v0` misses this because it always starts from a fresh table.

src/servers/src/otlp/coerce.rs
```diff
@@ -88,6 +103,18 @@ pub fn coerce_non_null_value(
         (ColumnDataType::String, ColumnDataType::Boolean, ValueData::StringValue(s)) => {
             s.parse::<bool>().ok().map(ValueData::BoolValue)
         }
+        // Checked signed -> unsigned cast for built-in fields moving to signed
+        // types (durations, counts are always non-negative). Negative values
+        // are rejected instead of wrapping so the coercion stays lossless.
+        (ColumnDataType::Int64, ColumnDataType::Uint64, ValueData::I64Value(n)) => {
```
fengjiachun, line 125: `end - start` is a `u64` subtraction, so a span with `end < start` wraps to a negative `i64`. New tables (`BIGINT`) accept it; old tables (`BIGINT UNSIGNED`) hit this arm, and the error propagates through `reconcile_trace_column_types` as a whole-request failure — the per-batch exclusion path only runs when `pre_alter.requires_ddl()`, which is false here. So one bad span kills the entire chunk, and only on pre-upgrade tables. Previously it was stored as a large `u64`.

sunng87:
Both review comments are addressed in c3dde541ed:

**v0 gap** — kept v0 on `Uint64` (the frozen-model option): the signed ingest compatibility layer only runs on the v1 path, so flipping v0 would have broken writes into every pre-existing v0 table at mito's schema check. The schema is now pinned unsigned by a unit test (`test_v0_duration_nano_stays_uint64`) and an integration check on `information_schema.columns` in `test_otlp_traces_v0`.

**negative duration asymmetry** — `v1.rs` now computes the duration with `checked_sub` + `i64::try_from` and rejects the span with a clear `invalid span duration` error at conversion time, before any table interaction. New Int64 tables and existing UInt64 tables now fail identically instead of (a) storing a negative duration on new tables, which broke the Jaeger query API's `.as_u64()`, or (b) aborting the whole chunk mid-coercion on old tables only.

Also folded in from review elsewhere: extracted `is_supported_signed_to_unsigned_coercion` so the trace and log paths share one supported-pair predicate.

sunng87:
Follow-up in 29302512bc: negative span durations are no longer an error — a span whose end precedes its start records `duration_nano = 0` (saturating at `i64::MAX` for oversized durations), so malformed spans never fail the request on either table generation.

sunng87:
## Test report: OTLP signed-integer coercion (duration_nano / trace_flags)

Verified locally on a standalone (File backend), PR build `feat/otlp-signed-integer-coercion` @ `29302512` vs `main` @ `8fac7128`. OTLP traffic was hand-encoded protobuf so exact edge values could be driven through the real ingest paths: span durations 7ns / 1.5ms / 250ms / 1h / `i64::MAX` (start=0, end=2262 edge) / end<start; log `trace_flags` = `[0, 1, 256, 0x80000000]`. Query surface covered: HTTP SQL, MySQL, PostgreSQL, and the Jaeger query API.

### 1. New version, fresh tables (signed) — ✅

- Schema: `duration_nano` → `Int64`, `trace_flags` → `Int32` (as intended)
- Values round-trip exactly across OTLP ingest and all query interfaces (incl. `i64::MAX` at the 2262 edge)
- end<start span is clamped to `0` (no wrap), flags `0x80000000` → `-2147483648` (bit pattern preserved)
- SELECT / WHERE / ORDER BY / GROUP BY / casts / arithmetic and count/min/max/sum/avg all fine
- Repeated OTLP batches into the same table fine
- SQL INSERT: `-42` accepted and reads back; `9223372036854775808` rejected with a clean cast error
- Jaeger API returns correct services/traces/durations

### 2. Upgrade: historical **unsigned** tables → new version — ✅

- Tables created by `main` (`bigint unsigned` / `int unsigned`), node restarted with the PR build: starts cleanly, **no ALTER / no migration**
- Schema stays unsigned; all historical rows read exactly over HTTP/MySQL/PG (incl. `2147483648` flags and a SQL-inserted `18446744073709551615` duration)
- New OTLP ingest accepted via the in-place coercion path (Int64→UInt64 checked; Int32→UInt32 bit-preserving): reversed→0, `0x80000000`→`2147483648`, `i64::MAX` intact — and the schema remains unsigned afterwards
- SQL INSERT bounds unchanged: `-42` → clean cast error; full u64 range still accepted
- Survives flush + restart (persisted SSTs) with schema and values intact; Jaeger fine

### 3. Users creating new tables get signed — ✅ (same as 1)

### 4. Downgrade: signed tables → old version — reads ✅, OTLP writes ❌

- Node starts healthy; DESC / information_schema / SHOW CREATE / SELECT / WHERE / MySQL / PG / Jaeger all work; values (incl. `-42`, `-2147483648`) intact; SQL INSERTs still work
- **OTLP traces ingest fails**: `400 incompatible logical types for trace column 'duration_nano'` (chunk fallback, 0 spans accepted)
- **OTLP logs ingest fails**: `400 failed to align log column 'trace_flags' from Uint32 to Int32`
- Failures are clean per-request rejections, no data corruption

➡️ Rollback is read-compatible but OTLP-write-incompatible — worth a release note.

### 5. Downgrade + `ALTER ... UNSIGNED` to restore OTLP ingestion — ✅ (one caveat)

```sql
ALTER TABLE opentelemetry_traces MODIFY COLUMN duration_nano BIGINT UNSIGNED;
ALTER TABLE opentelemetry_logs    MODIFY COLUMN trace_flags   INT UNSIGNED;
```

- Both ALTERs succeed on `main` (including on the partitioned trace table)
- OTLP logs + traces ingestion resume (200), rows land correctly
- Non-negative values preserved exactly
- ⚠️ **Negative values silently become NULL**: a stored `-2147483648` `trace_flags` row turned NULL after the ALTER (no error, no wrap)
- Re-upgrading to the PR build afterwards also works: unsigned schema kept, coercion path active, Jaeger fine

### Pre-existing issues found while testing (NOT caused by this PR)

1. `min()`/`max()` alone (or `count+min`, `min+max`) on the partitioned `opentelemetry_traces` fails with `Schema error: MergeScan advertised remote stream schema field mismatch` — reproduces identically on `main` with unsigned columns, any column type; the full `count+min+max+sum+avg` shape takes a different plan and works. Probably worth a separate issue.
2. `main` panics its write worker on spans with end<start (`attempt to subtract with overflow`, `v1.rs:371`); the request is dropped. This PR's `saturating_sub` clamp incidentally fixes it. 🎉

### Behavioral notes

- `sum(duration_nano)` now wraps at `i64::MAX` overflow instead of u64 — irrelevant for real-world durations, but a semantic delta
- flags `0x80000000` reads back as `-2147483648` on signed tables vs `2147483648` on kept-unsigned tables — both preserve the bit pattern; only the 8 W3C LSBs are meaningful per spec anyway

**Verdict:** the signed flip + unsigned-compat coercion layer works as designed (scenarios 1–3 pass); downgrade is a one-way door for OTLP pipelines unless users `ALTER` back to unsigned, at the cost of nulling any negative values.

shuiyisong (approved):
LGTM

PR #8818, opened 2026-08-10
fix(query): ignore field metadata in MergeScan remote schema validation

discord9:

- Introduced by #8579 (validate merge scan remote schema)

`validate_remote_schema` (added in #8579) compared remote scan schemas with Arrow's default `Field` equality, which **includes field metadata**. In production, a column's SST region metadata often carries `greptime:skipping_index` (and/or `greptime:inverted_index`) while the frontend table schema does not declare those index options. The two Fields were identical in name, data type, and nullability, yet the validation rejected them as a mismatch — surfacing as HTTP 500 "MergeScan advertised remote stream schema field mismatch" on ordinary queries over tables whose SSTs carry index metadata.

Field metadata is auxiliary (index/encoding/compression hints). It is not part of field semantics and must not participate in schema comparison.

### Changes

1. **`fields_semantically_equal`** — the primary comparison is now exactly:
   - `name` equal
   - `data_type` equal
   - `is_nullable` equal
   
   No field metadata is compared, for any field type.

2. **JSON exemption kept** (`json_fields_compatible`) — JSON columns are the one intentional exception to strict `data_type` comparison: on the wire they are binary-encoded (`Binary` + `ARROW:extension:name=greptime.json` / `greptime:type=Json`), and after decoding they are a structured form (`Struct(...)`/`List(...)`) carrying the same extension metadata. Both forms describe the same column, so the raw physical arrow type is not compared for JSON fields; the JSON identity markers must still match. This exemption is scoped strictly to JSON fields (`is_json_field` on both sides) and never applies to non-JSON columns.

3. **Timestamp timezone difference kept as an accepted exception** — two fields whose `data_type` differs only in the Timestamp timezone string (same unit, e.g. `+08:00` vs `UTC`) are still accepted when name and nullability match. This mirrors Arrow's own field-equality behavior minus timezone, and was already intended by #8579; the difference is that this exception now no longer inspects metadata either.

### Why this is safe

- The validation's purpose (from #8579) is to catch **real** schema divergence — wrong column order, mismatched type, different nullability, different field count — all of which still fail.
- Skipping index metadata describes how a column is indexed on disk; it does not change the column's logical identity and legitimately differs between the table schema and the region manifest depending on how the table was created vs. how the SST was built.
- No API, wire, or persisted-format changes.

## PR Checklist

### DIFF
--- a/src/query/src/dist_plan/merge_scan.rs
+++ b/src/query/src/dist_plan/merge_scan.rs
@@ -129,25 +129,13 @@ fn is_json_field(field: &Field) -> bool {
 /// name, same data type, and same nullability.
 ///
 /// Arrow `Field` metadata is auxiliary information (indexing, encoding,
-/// compression hints, ...) and does not participate in the comparison.
-///
-/// JSON columns are the only exception: their JSON identity keys
-/// (`greptime:type`, `ARROW:extension:metadata`) are semantic, so a JSON
-/// column is never semantically equal to a non-JSON column of the same
-/// physical type and two JSON columns must agree on those keys. The
-/// wire/decoded physical-type difference of JSON columns is exempted
+/// compression hints, ...) and does not participate in the comparison. This
+/// applies to every field, JSON columns included: JSON identity keys
+/// (`greptime:type`, `ARROW:extension:metadata`) are never compared here.
+/// The wire/decoded physical-type difference of JSON columns is exempted
 /// separately by [`json_fields_compatible`].
 fn fields_semantically_equal(expected_field: &Field, actual_field: &Field) -> bool {
-    is_json_field(expected_field) == is_json_field(actual_field)
-        && expected_field.metadata().get(datatypes::schema::TYPE_KEY)
-            == actual_field.metadata().get(datatypes::schema::TYPE_KEY)
-        && expected_field
-            .metadata()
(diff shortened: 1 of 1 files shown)


src/query/src/dist_plan/merge_scan.rs
```diff
@@ -197,7 +219,6 @@ fn validate_remote_schema(
                 && expected_timezone != actual_timezone
                 && expected_field.name() == actual_field.name()
                 && expected_field.is_nullable() == actual_field.is_nullable()
```
evenyag, line 221: Are these checks duplicated with fields_semantically_equal()?

discord9: Not duplicated - they are a necessary guard inside the timezone-only exception. fields_semantically_equal() (name + data_type + nullable) runs first and continues on equal; a timezone-only difference means data_type is NOT equal, so we only reach the timezone branch after the first check already failed. At that point name/nullable equality is not guaranteed, and the exception must not widen into accepting fields that differ in name or nullability too. The test merge_scan_remote_schema_identity_rejects_timestamp_timezone_plus_field_mismatches pins exactly this: same-unit/different-tz + different nullable must be rejected. I briefly tried dropping those two checks and the test failed, confirming they are load-bearing.

evenyag (approved):

PR #8821, opened 2026-08-10
perf: cache bulk-memtable unordered-part sort (18-1818x on repeated reads)

discord9:
## Summary

Cache the sorted result of the bulk-memtable unordered part so repeated scans with no intervening writes skip the per-scan concat+sort.

`BulkMemtable::ranges()` re-ran `concat_and_sort` on the unordered part on **every scan**, even when no new rows arrived. The sorted result is now cached on `UnorderedPart` and invalidated on `push`/`clear`, with `max/min timestamp` + `max sequence` stats cached alongside. Unit test covers cache hit, push invalidation, and sorted output.

**Micro-bench** (`bench_bulk_memtable_unordered_part_sort_cache` in bulk.rs tests, 20x `ranges()` with no intervening writes, medians of 3 interleaved opt/base runs):

| parts (rows) | baseline per-call | optimized per-call | speedup |
|---|---|---|---|
| 100 | 393.4 µs | 21.6 µs | **18x** |
| 1,000 | 3.23 ms | 21.2 µs | **152x** |
| 5,000 | 19.9 ms | 23.9 µs | **833x** |
| 10,000 | 43.4 ms | 23.9 µs | **1,818x** |

Optimized repeated-read cost is flat (~21-25 µs, cache-hit only) while baseline grows linearly with part count. Write-interleaved scenario (cache invalidated every write): 264.5 vs 263.9 µs/iter — **no regression** (0.2%, noise).

**Scenario**: write-then-query repeatedly (e.g. remote-write then dashboard queries) — every scan previously re-sorted the whole unordered part.

## Files
- `src/mito2/src/memtable/bulk.rs` (+ unit test + micro-bench)
- `src/mito2/src/memtable/bulk/part.rs`

## Testing
- mito2 full suite: 1105-1108 passed (one flaky retried green)

## Checklist
- [ ] CLA signed
- [ ] Perf evidence attached (micro-benchmark table above)

### DIFF
--- a/src/mito2/src/memtable/bulk/part.rs
+++ b/src/mito2/src/memtable/bulk/part.rs
@@ -15,7 +15,7 @@
 //! Bulk part encoder/decoder.
 
 use std::collections::{HashMap, HashSet};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 
 use api::helper::{ColumnDataTypeWrapper, to_grpc_value};
@@ -351,6 +351,11 @@ pub struct UnorderedPart {
     max_timestamp: i64,
     /// Maximum sequence number across all parts.
     max_sequence: u64,
+    /// Cached sorted [BulkPart] built by `to_bulk_part`, invalidated on
+    /// `push`/`clear`. Avoids re-running concat+sort on every scan when no
+    /// new parts have arrived. Guarded by a mutex so `to_bulk_part(&self)`
+    /// can lazily fill it while `ranges()` only holds a read lock.
+    sorted_cache: Mutex<Option<BulkPart>>,
     /// Row count threshold for accepting parts (default: 1024).
     threshold: usize,
     /// Row count threshold for compacting (default: 4096).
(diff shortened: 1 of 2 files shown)


src/mito2/src/memtable/bulk/part.rs
```diff
@@ -345,6 +345,11 @@ pub struct UnorderedPart {
     max_timestamp: i64,
     /// Maximum sequence number across all parts.
     max_sequence: u64,
+    /// Cached sorted [BulkPart] built by `to_bulk_part`, invalidated on
+    /// `push`/`clear`. Avoids re-running concat+sort on every scan when no
+    /// new parts have arrived. Guarded by a mutex so `to_bulk_part(&self)`
+    /// can lazily fill it while `ranges()` only holds a read lock.
+    sorted_cache: Mutex<Option<BulkPart>>,
```
evenyag, line 358: Since we concat the sorted part by parts in UnorderedPart, I think we can replace the input parts instead of caching the sorted one.

evenyag:
@discord9 Please use our PR templates for these PRs

PR #8823, opened 2026-08-10
fix(meta): validate leaders before enabling skip_wal

fengjiachun:

Follow-up to #8730. The leaderless case was found while reviewing the v1.2 backport in #8817.

Enabling `skip_wal` uses a metadata-first flow because changing a replica to skip WAL is irreversible. Previously, a table with no region leaders could commit `skip_wal = true` in table metadata and then fail the following region step at its existing `NoLeader` check.

This PR validates that the locked physical route has at least one leader during the prepare step, before the procedure advances to update metadata. It also adds a regression test asserting that the leaderless error leaves `skip_wal` disabled in table metadata.

The replica request remains idempotent and uses the existing bounded procedure retry behavior. This PR does not change the common procedure retry policy.

## PR Checklist

Validation:

- `cargo nextest run -p common-meta skip_wal`
- `cargo nextest run -p mito2 test_alter_skip_wal_stops_wal_and_flushes_on_close`

### DIFF
--- a/src/common/meta/src/ddl/alter_table.rs
+++ b/src/common/meta/src/ddl/alter_table.rs
@@ -198,6 +198,12 @@ impl AlterTableProcedure {
                 // DDL manager can submit another procedure with locks for the current route.
                 return Ok(Status::done_with_output(RegionRouteChanged));
             }
+            ensure!(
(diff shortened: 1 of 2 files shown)


fengjiachun:
@codex review

WenyXu (approved):

PR #8825, opened 2026-08-10
refactor: share query channel definition

WenyXu:

Move the query `Channel` definition into `common-base::protocol`, preserving its wire values and manual `Display`/`AsRef<str>` behavior. Re-export it from `session::context` for source compatibility, keep SQL-dialect selection in `session`, and let metadata derive the event protocol name directly from the shared enum.

## PR Checklist


sunng87 (approved):

src/session/Cargo.toml
```diff
@@ -11,6 +11,7 @@ testing = []
 workspace = true
 
 [dependencies]
+common-base.workspace = true
```
sunng87: sort dependency by alphabet

PR #8833, opened 2026-08-10
refactor(json2): push down json2 type hints to parquet reads

fengys1996:

This PR pushes JSON2 target type handling deeper into the Mito2 read path. After this change, JSON2 values are concretized in `NestedSchemaAligner`; the compat layer no longer handles query-driven JSON2 projection and only continues to handle SST schema compatibility and type evolution for normal columns.

Previously, JSON2 type hints deduced by the query layer were first converted into parquet nested paths. The read path then built an intermediate schema from those paths, and the compat layer projected JSON2 values again to match the query-required target type. This split JSON2 type handling across multiple stages and made the scan output schema contract less clear.

After this change, the read path keeps two pieces of information separately:

- Root column projection: which root columns the scan needs.
- JSON2 target types: the expected output shape of JSON2 columns for this query.

Mito2 uses JSON2 target types to build the scan output schema and derives the actual parquet nested paths internally. `NestedSchemaAligner` aligns SST data directly to this target schema. Normal columns are not affected by JSON2 hints; they still use Region Metadata as the target type and rely on the compat layer for compatibility across different SST schemas.

Main changes:

- Keep JSON2 target types in `ReadColumns` / `FlatReadFormat` as part of scan semantics.
- Build parquet output schema and compat schema from JSON2 target types.
- Derive parquet nested paths internally from JSON2 target types.
- Remove query-driven JSON2 projection from the compat layer.
- Include JSON2 target types in file-level cache compatibility checks.
- Simplify JSON2 hint pushdown in the query optimizer and retain hints only for actual JSON2 fields.
- Fix projection index lookup for primary-key encoded SSTs before flat conversion.

Known limitations:

- Mixing whole JSON2 column reads with path projections is still a known limitation and will be fixed in a follow-up PR.
- JSON2 hints are still matched by unqualified column name, so same-name JSON2 columns in joins can still be ambiguous.

This PR does not change public APIs or persisted data formats.

## PR Checklist


fengys1996:
@codex review

src/query/src/optimizer/json_type_concretize.rs
```diff
@@ -56,22 +58,53 @@ impl OptimizerRule for JsonTypeConcretizeRule {
...
+    json_types: &HashMap<String, JsonNativeType>,
+) -> bool {
+    let schema = provider.schema();
+    let json_types = json_types
+        .iter()
+        .filter(|(column, _)| {
+            schema
+                .fields()
+                .iter()
+                .any(|field| field.name() == *column && is_json2_extension_type(field))
```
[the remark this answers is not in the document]
fengys1996: This will be fixed in a follow-up PR.

fengys1996:
@codex review

src/query/src/optimizer/json_type_concretize.rs
```diff
@@ -56,22 +58,56 @@ impl OptimizerRule for JsonTypeConcretizeRule {
...
+        })
+        .map(|(column, json_type)| (column.clone(), json_type.clone()))
```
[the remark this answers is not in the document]
fengys1996: This is a known limitation, not a regression from this PR.

  `json2_limit.sql` already covers this with expected errors:
  `select j, j.a ...` and `select j ... where j.a.b = 1`.

  The current behavior is a schema mismatch error (`expected Binary but found Struct(...)`), not silent truncation. It will be fixed in another PR.

src/query/src/optimizer/json_type_concretize.rs
```diff
@@ -56,22 +58,56 @@ impl OptimizerRule for JsonTypeConcretizeRule {
...
+        return false;
+    }
+
+    if let Some(adapter) = provider.as_any().downcast_ref::<DummyTableProvider>() {
+        adapter.with_json_type_hint(json_types);
+        return true;
+    }
+
+    if let Some(adapter) = provider.as_any().downcast_ref::<DfTableProviderAdapter>() {
+        adapter.with_json_type_hint(json_types);
```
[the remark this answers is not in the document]
fengys1996: I checked this locally and could not reproduce the reported failure.

  The existing root-path cases in `json2.sql` already cover `json_get(j, '')`, `json_get(j, '$')`, `json_get(j, '.')`, and `json_get(j, '$.')`, and the `json` sqlness target passes.

  Although the deduced native hint is `JsonNativeType::String`, scan wraps it as `ConcreteDataType::json2(String)`. `JsonFormat::Json2(...).as_arrow_type()` lifts a root scalar to `Variant`, so the materialized Arrow type remains `Binary`, not `Utf8View`.

src/mito2/src/read/flat_projection.rs
```diff
@@ -133,21 +136,8 @@ impl FlatProjectionMapper {
...
-                    })
-                    .map(ConcreteDataType::json2)
-                {
-                    *data_type = concretized;
-                } else if is_empty_json2_type(json_type) {
-                    // see `merge_scan::maybe_amend_json2_field`
-                    *data_type = ConcreteDataType::json2(JsonNativeType::Variant);
-                }
+            if let Some(updated) = json2_read_datatype(*column_id, data_type, &read_cols) {
+                *data_type = updated;
```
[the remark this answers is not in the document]
fengys1996: I tested an SST + memtable merge case locally and it works as expected.

  Memtable batches are normalized by `FlatProjectionMapper::convert()` before being returned from the scan stream. For JSON2 columns, it uses `JsonArray::project_to(field.data_type())` to project the array to the mapper's output schema, so the final schema is unified there.

  So I don't think this is an issue in the current read path.

src/mito2/src/read/compat.rs
```diff
@@ -270,11 +261,13 @@ impl FlatCompatBatch {
                     let old_column = batch.column(*pos);
 
                     if let Some(ty) = cast_type {
-                        let casted = if let Some(json_type) = ty.as_json()
-                            && json_type.is_json2()
-                        {
+                        let casted = if ty.is_json2() {
+                            // JSON2 query-time projection should have been handled before
+                            // flat compat. Keep a release-mode fallback for unexpected
```
MichaelScofield: unexpected is bug, so I think we should remove this if

src/query/src/optimizer/json_type_concretize.rs
```diff
@@ -56,22 +58,56 @@ impl OptimizerRule for JsonTypeConcretizeRule {
...
+                } else {
+                    Ok(Transformed::no(plan))
+                }
             }
             _ => Ok(Transformed::no(plan)),
         })
     }
 }
 
+// FIXME: `json_types` is keyed only by unqualified column name. In joins with
```
MichaelScofield, line 72: we can add some sqlness tests regarding these kinds of joins on json2 columns first

MichaelScofield (approved):
lgtm but we still need approve from mito master @evenyag

src/mito2/src/read/read_columns.rs
```diff
@@ -12,119 +12,69 @@
...
     where
         I: IntoIterator<Item = ColumnId>,
     {
-        let cols = column_ids
-            .into_iter()
-            .map(|col_id| ReadColumn::new(col_id, vec![]))
-            .collect();
-        ReadColumns { cols }
+        Self {
+            col_ids: col_ids.into_iter().collect(),
```
evenyag, line 44: Do we require that column ids are deduped? There is no document for it.

src/mito2/src/read/read_columns.rs
```diff
@@ -12,119 +12,69 @@
...
 
+use std::collections::BTreeMap;
 use std::mem;
+use std::sync::Arc;
 
-use store_api::storage::{ColumnId, NestedPath};
+use datatypes::prelude::ConcreteDataType;
+use store_api::storage::ColumnId;
+
+pub(crate) type JsonTargetTypes = Arc<BTreeMap<ColumnId, ConcreteDataType>>;
```
evenyag: What does JsonTargetTypes store? Now it looks like ConcreteDataType is always json.

evenyag (commented):
I only checked mito parts.

src/mito2/src/read/read_columns.rs
fengys1996: > What does JsonTargetTypes store?

It stores the mapping converted from JSON type hints pushed down by the query layer: `ColumnId -> JSON2 target type`.

> Now it looks like ConcreteDataType is always json.

Yes, the `ConcreteDataType` here is expected to always be a JSON2 type. It seems using `JsonNativeType` would be more precise here?

src/mito2/src/read/read_columns.rs
fengys1996: Current callers pass deduplicated column ids, but I'm not sure this API should require it as a strict contract.

src/mito2/src/read/read_columns.rs
fengys1996: Changed to `JsonNativeType` in https://github.com/GreptimeTeam/greptimedb/pull/8833/changes/7b9c8f2e95b22d8f2adb76d6d1b333900890e7bf.

src/mito2/src/read/read_columns.rs
fengys1996: Added doc in https://github.com/GreptimeTeam/greptimedb/pull/8833/changes/0c10ad34c090ac79909bbff8bb61ed7963ef4201

fengys1996:
@evenyag ptal

evenyag (approved):

PR #8836, opened 2026-08-10
feat: complete the derived-edge vocabulary of the entity graph

### LINKED ISSUE #8793: Transition built-in data models from unsigned to signed integers (Postgres/JDBC/Spark-Parquet compatibility)

killme2008:

#8609. Follows #8841 (mechanical module split, merged); each commit is one derivation.

Completes the derived-edge vocabulary of the read-time entity graph (#8794), per the [entity-relationships RFC](https://github.com/GreptimeTeam/greptimedb/blob/main/docs/rfcs/2026-06-25-entity-relationships-and-graph-query.md). Four derivations, all inside the existing 16-column `semantic_relationships` contract and `GraphQueryWindow` semantics:

- **Cross-table `calls` pairing.** The service-graph join previously self-joined each trace table, so a client span and its server span stored in different tables (`x-greptime-trace-table-name` routing) never paired. The derivation now projects every trace table's client and server spans to a normalized shape (with each table's own `service` identity expression), unions them, and joins once. A single table degenerates to the previous self-join.
- **Virtual-node edges.** A client span with no matching server span becomes a `calls` edge to a peer named by span attributes (`service.peer.name` / `db.namespace` first, the deprecated `peer.service` / `db.name` kept for existing telemetry, `server.address` last; empty values fall through), with `confidence = 0.5` and `attributes.connection_type` (`database` when named by `db.name`, else `virtual_node`). Real pairs win: when a window's edge key has any pair, the edge reports only the pair population (the RFC pins RED metrics to observed span pairs) and unmatched clients are suppressed, so one `(window, edge)` never yields two rows. Implemented as one left join with filtered aggregates — no extra scans over the trace tables.
- **Same-row co-declared edges.** A table declaring both entity types of a built-in vocabulary pair witnesses the edge on every row carrying both identities; the vocabulary fixes the direction: `service.instance|process -> host` and `k8s.pod -> k8s.node` (`runs_on`), `k8s.pod -> k8s.container` (`contains`), `service.instance -> service` and `k8s.pod -> k8s.workload` (`part_of`), with `provenance = 'attribute'`. The agent rules `agent -> model` (`uses`) and `agent -> tool` (`invokes`) apply only to trace sources — the RFC derives them from span structure, so a non-trace table co-declaring these types must not fabricate invocations — and carry `provenance = 'trace'`.
- **`parent_agent calls agent`.** Trace tables declaring an `agent` entity pair each span with its child span (no span-kind filter), keep pairs whose agent identities differ, and aggregate RED metrics, anchored on the parent span like the service derivation anchors on the client.

Also: `enumerate` now validates the fixed trace-v1 columns (names + ingest types) and derives around a malformed trace-model table with a warning instead of failing every relationships scan (closes the TODO left by #8794; the RFC requires schema drift never to poison a scan).

Limitations:

- A whole trace duplicated into several tables (fan-out ingestion) inflates pair counts quadratically, since the same spans appear on both join sides. Span identity for dedup (and its conflict rules) is deliberately not defined here; a single table keeps exactly the previous semantics.
- When the caller may read a client table but not the table holding the matching server spans, the unmatched clients can surface as virtual-node edges. These derive only from the readable table and leak nothing from the denied one.

## PR Checklist

### DIFF
--- a/src/frontend/src/instance/entity_graph.rs
+++ b/src/frontend/src/instance/entity_graph.rs
@@ -34,10 +34,8 @@ use auth::{
 use catalog::CatalogManager;
 use catalog::system_schema::semantic_graph::EntityGraphProvider;
 use common_catalog::consts::{
-    DEFAULT_PRIVATE_SCHEMA_NAME, DEFAULT_SCHEMA_NAME, DURATION_NANO_COLUMN,
-    INFORMATION_SCHEMA_NAME, PARENT_SPAN_ID_COLUMN, PG_CATALOG_NAME,
-    SEMANTIC_RELATIONSHIPS_DECLARED_TABLE_NAME, SERVICE_NAME_COLUMN, SPAN_ID_COLUMN,
-    SPAN_KIND_COLUMN, SPAN_STATUS_CODE_COLUMN, TRACE_ID_COLUMN, TRACE_TIMESTAMP_COLUMN,
+    DEFAULT_PRIVATE_SCHEMA_NAME, DEFAULT_SCHEMA_NAME, INFORMATION_SCHEMA_NAME, PG_CATALOG_NAME,
+    SEMANTIC_RELATIONSHIPS_DECLARED_TABLE_NAME, SERVICE_NAME_COLUMN,
 };
 use common_error::ext::{BoxedError, ErrorExt};
 use common_error::status_code::StatusCode;
@@ -83,39 +81,6 @@ struct TraceSource {
     service: Option<EntityDeclaration>,
     agent: Option<EntityDeclaration>,
     table: TableRef,
-}
-
-/// Whether a trace-model table still carries the fixed `greptime_trace_v1`
-/// columns (with their ingest types) the calls/agent derivations reference.
-/// `table_data_model` is only an option: a stale or hand-made table missing
-/// them would otherwise fail plan building and poison every relationships
-/// scan.
-fn trace_v1_schema_matches(table_info: &TableInfo) -> bool {
-    use datatypes::prelude::ConcreteDataType;
-    let schema = &table_info.meta.schema;
-    let column_type = |name: &str| {
-        schema
-            .column_schema_by_name(name)
-            .map(|column| column.data_type.clone())
-    };
-    // The derivations filter and bucket by this column; if it is not the time
-    // index, the window predicate cannot prune the scan and the buckets
-    // diverge from the table's time semantics.
-    schema
-        .timestamp_column()
-        .is_some_and(|column| column.name == TRACE_TIMESTAMP_COLUMN)
-        && column_type(TRACE_TIMESTAMP_COLUMN)
-            == Some(ConcreteDataType::timestamp_nanosecond_datatype())
-        && [
-            TRACE_ID_COLUMN,
-            SPAN_ID_COLUMN,
-            PARENT_SPAN_ID_COLUMN,
-            SPAN_KIND_COLUMN,
-            SPAN_STATUS_CODE_COLUMN,
-        ]
-        .iter()
-        .all(|column| column_type(column) == Some(ConcreteDataType::string_datatype()))
-        && column_type(DURATION_NANO_COLUMN) == Some(ConcreteDataType::uint64_datatype())
 }
 
 impl EntityGraphProviderImpl {
@@ -309,20 +274,7 @@ impl EntityGraphProviderImpl {
             while let Some(table) = tables.try_next().await.map_err(BoxedError::new)? {
                 let table_info = table.table_info();
                 let table_declarations = Self::declarations_for(&table_info);
-                let is_trace = match is_trace_v1_table(&table_info) {
-                    true if trace_v1_schema_matches(&table_info) => true,
-                    true => {
-                        // Schema drift never poisons a scan: derive around the
-                        // malformed table instead of failing plan building.
-                        warn!(
-                            "Trace table `{}` does not match the fixed trace-v1 schema; \
-                             skipping calls derivation",
-                            table_info.name
-                        );
-                        false
-                    }
-                    false => false,
-                };
+                let is_trace = is_trace_v1_table(&table_info);
                 // Authorize only tables that would contribute rows.
                 if per_table_auth
                     && (is_trace || !table_declarations.is_empty())
@@ -694,79 +646,6 @@ mod tests {
             ],
         );
         assert!(EntityGraphProviderImpl::declarations_for(&info).is_empty());
-    }
-
-    /// The canonical fixed trace-v1 columns, with an injection point for the
-    /// aspects `trace_v1_schema_matches` must reject.
-    fn trace_v1_info(mutate: impl FnOnce(&mut Vec<ColumnSchema>)) -> TableInfo {
-        let string =
-            |name: &str| ColumnSchema::new(name, ConcreteDataType::string_datatype(), true);
-        let mut columns = vec![
-            ColumnSchema::new(
-                TRACE_TIMESTAMP_COLUMN,
-                ConcreteDataType::timestamp_nanosecond_datatype(),
-                false,
-            )
-            .with_time_index(true),
-            string(TRACE_ID_COLUMN),
-            string(SPAN_ID_COLUMN),
-            string(PARENT_SPAN_ID_COLUMN),
-            string(SPAN_KIND_COLUMN),
-            string(SPAN_STATUS_CODE_COLUMN),
-            ColumnSchema::new(
-                DURATION_NANO_COLUMN,
-                ConcreteDataType::uint64_datatype(),
-                true,
-            ),
-        ];
-        mutate(&mut columns);
-        assemble_table_info(columns, &[(TABLE_DATA_MODEL, TABLE_DATA_MODEL_TRACE_V1)])
-    }
-
-    #[test]
-    fn malformed_trace_v1_schema_is_rejected() {
-        assert!(trace_v1_schema_matches(&trace_v1_info(|_| {})));
-
-        let missing_duration = trace_v1_info(|columns| {
-            columns.retain(|c| c.name != DURATION_NANO_COLUMN);
-        });
-        assert!(!trace_v1_schema_matches(&missing_duration));
-
-        let wrong_kind_type = trace_v1_info(|columns| {
-            columns
-                .iter_mut()
-                .find(|c| c.name == SPAN_KIND_COLUMN)
-                .unwrap()
-                .data_type = ConcreteDataType::int64_datatype();
-        });
-        assert!(!trace_v1_schema_matches(&wrong_kind_type));
-
-        // A millisecond time index is not the trace-v1 ingest schema.
-        let ms_timestamp = trace_v1_info(|columns| {
-            columns[0] = ColumnSchema::new(
-                TRACE_TIMESTAMP_COLUMN,
-                ConcreteDataType::timestamp_millisecond_datatype(),
-                false,
-            )
-            .with_time_index(true);
-        });
-        assert!(!trace_v1_schema_matches(&ms_timestamp));
-
-        // "timestamp" present with the right type, but the time index is a
-        // different column: the derivations would bucket by a non-indexed
-        // column and the window predicate could not prune the scan.
-        let time_index_elsewhere = trace_v1_info(|columns| {
-            columns[0] = columns[0].clone().with_time_index(false);
-            columns.push(
-                ColumnSchema::new(
-                    "ts",
-                    ConcreteDataType::timestamp_nanosecond_datatype(),
-                    false,
-                )
-                .with_time_index(true),
-            );
-        });
-        assert!(!trace_v1_schema_matches(&time_index_elsewhere));
     }
 
     #[test]
--- a/src/frontend/src/instance/entity_graph.rs
+++ b/src/frontend/src/instance/entity_graph.rs
@@ -34,8 +34,10 @@ use auth::{
 use catalog::CatalogManager;
 use catalog::system_schema::semantic_graph::EntityGraphProvider;
 use common_catalog::consts::{
-    DEFAULT_PRIVATE_SCHEMA_NAME, DEFAULT_SCHEMA_NAME, INFORMATION_SCHEMA_NAME, PG_CATALOG_NAME,
-    SEMANTIC_RELATIONSHIPS_DECLARED_TABLE_NAME, SERVICE_NAME_COLUMN,
+    DEFAULT_PRIVATE_SCHEMA_NAME, DEFAULT_SCHEMA_NAME, DURATION_NANO_COLUMN,
+    INFORMATION_SCHEMA_NAME, PARENT_SPAN_ID_COLUMN, PG_CATALOG_NAME,
+    SEMANTIC_RELATIONSHIPS_DECLARED_TABLE_NAME, SERVICE_NAME_COLUMN, SPAN_ID_COLUMN,
+    SPAN_KIND_COLUMN, SPAN_STATUS_CODE_COLUMN, TRACE_ID_COLUMN, TRACE_TIMESTAMP_COLUMN,
 };
 use common_error::ext::{BoxedError, ErrorExt};
 use common_error::status_code::StatusCode;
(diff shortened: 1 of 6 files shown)


shuiyisong (approved):
LGTM

src/frontend/src/instance/entity_graph.rs
```diff
@@ -73,9 +75,49 @@ pub struct EntityGraphProviderImpl {
...
+            TRACE_ID_COLUMN,
+            SPAN_ID_COLUMN,
+            PARENT_SPAN_ID_COLUMN,
+            SPAN_KIND_COLUMN,
+            SPAN_STATUS_CODE_COLUMN,
+        ]
+        .iter()
+        .all(|column| column_type(column) == Some(ConcreteDataType::string_datatype()))
+        && column_type(DURATION_NANO_COLUMN) == Some(ConcreteDataType::uint64_datatype())
+}
```
sunng87: We should just check the table_option for trace table instead of this tricky and expensive (both in terms of checking and maintaining) column pattern matching.

For example, in #8816 I may change duration_nano to int64 type and we can hardly track dependency like this.

With a special column named as duration_nano already means it's greptime_trace_v1 table.

src/frontend/src/instance/entity_graph.rs
```diff
@@ -613,6 +696,79 @@ mod tests {
...
+            string(SPAN_STATUS_CODE_COLUMN),
+            ColumnSchema::new(
+                DURATION_NANO_COLUMN,
+                ConcreteDataType::uint64_datatype(),
+                true,
+            ),
+        ];
+        mutate(&mut columns);
+        assemble_table_info(columns, &[(TABLE_DATA_MODEL, TABLE_DATA_MODEL_TRACE_V1)])
+    }
```
sunng87: Consider putting this into somewhere near trace/v1.rs in servers, for ease of maintenance.

src/operator/src/statement/semantic_graph/relationships.rs
```diff
@@ -315,9 +714,29 @@ mod tests {
...
+    type Span<'a> = (
+        i64,
+        &'a str,
+        &'a str,
+        Option<&'a str>,
+        &'a str,
+        &'a str,
+        &'a str,
+        u64,
+    );
```
sunng87, line 729: consider moving the definition to somewhere near trace/v1.rs

src/frontend/src/instance/entity_graph.rs
killme2008: Adopted — the guard is gone in b9f1fde; enumerate now trusts the table option alone. Since the option is only stamped by the ingest path, the fixed columns hold by construction. One heads-up for #8816: if a type change lands while older tables keep uint64, the cross-table span union will need a cast in the normalized projection (one line in relationships.rs) — happy to ride that in your PR.

src/frontend/src/instance/entity_graph.rs
killme2008: This one is a test-local mock (arrow MemTable, built directly against the fixture rows). The column names come from the same common_catalog consts trace/v1.rs uses, and with the schema guard removed (thread above) no canonical trace-v1 schema definition remains in production code here. A single shared canonical definition would be nice but needs a home below servers (operator cannot depend on it) — best done together with #8816 if you want it.

src/operator/src/statement/semantic_graph/relationships.rs
killme2008: Same as the sibling thread: a test-local mock, deliberately minimal and sometimes malformed on purpose; operator has no dependency path to servers, so it cannot reuse a definition living there. The shared source of truth for the column names is common_catalog::consts, which trace/v1.rs also consumes.

PR #8840, opened 2026-08-11
fix(tests): make two Windows CI failures deterministic (Nightly CI #8837)

### LINKED ISSUE #8837: Workflow run 'Nightly CI' failed
@GreptimeTeam/db-approver
New failure: https://github.com/GreptimeTeam/greptimedb/actions/runs/31441541251

discord9:

Closes #8837

Nightly CI's "Run tests on Windows" job has been failing repeatedly. This PR fixes both failures from run 31441541251:

**1. `cmd::datanode_overlay::tests::rejects_directories_and_parse_errors` (deterministic, 4/4 retries failed)**

Introduced by #8647. `DatanodeOverlay::load()` opened the target with `File::open` before checking `is_file()`. On Unix, opening a directory succeeds and the loader rejects it with "must be a regular file"; on Windows, `File::open` on a directory fails up front with "Access is denied", so the type check was never reached and the assertion on "regular file" always failed.

Fix: check `std::fs::metadata` before `File::open`. `metadata` succeeds on directories on both platforms, so the "must be a regular file" error is now identical everywhere and the test assertion holds on Windows too.

**2. `heartbeat::handler::upgrade_region::tests::test_region_error` (flaky, TRY 1 failed / TRY 2 passed)**

The second phase raced a 100ms mock handle delay against a 200ms `replay_timeout`. On a busy Windows runner the mock error could land after the timeout fired, so `reply.error` was still `None`.

Fix: the second phase now uses a mock handle that returns the error on its first poll with no delay — the catchup future completes before `replay_timeout` can ever fire, so the test no longer depends on wall-clock scheduling.

## PR Checklist

### DIFF
--- a/src/datanode/src/heartbeat/handler/upgrade_region.rs
+++ b/src/datanode/src/heartbeat/handler/upgrade_region.rs
@@ -408,9 +408,12 @@ mod tests {
     #[tokio::test]
     async fn test_region_error() {
         common_telemetry::init_default_ut_logging();
-        let mock_region_server = mock_region_server();
         let region_id = RegionId::new(1024, 1);
 
+        // With the default (zero) `replay_timeout`, the handler times out before the
+        // mock handle's 100ms delay elapses: it didn't wait for handle returns; it
+        // had no idea about the error.
+        let region_server = mock_region_server();
         let (mock_engine, _) =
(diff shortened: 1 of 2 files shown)


evenyag (approved):

evenyag:
@codex review

PR #8847, opened 2026-08-11
fix(mito2): fence checkpoints during region transitions

### LINKED ISSUE #8625: Manifest checkpoint cleanup races with repartition manifest replay
## What type of bug is this?

Unexpected error

- Distributed Cluster
- Storage Engine
- Metasrv
- Datanode

## Minimal reproduce step

Observed in CI job:
https://github.com/GreptimeTeam/greptimedb/actions/runs/30004985300/job/89202460122

- Fuzz target: `fuzz_repartition_metric_table`
- Seed: `0`; actions: `125`
- Tested SHA: `52ba89bb10d3919e6e5191a120a590a98287230b`

A deterministic regression test can use a blocking object store:

1. Create a manifest with deltas through version 10 and no checkpoint.
2. Start manifest recovery; let it observe no checkpoint and list delta 0.
3. Pause before recovery GETs delta 0.
4. Complete checkpoint creation, including publishing `_last_checkpoint` and deleting obsolete deltas.
5. Resume recovery.

Current behavior fails with `NotFound`; fixed behavior should restart from the new checkpoint and recover successfully.

## What did you expect to see?

Repartition should read a consistent source manifest and complete successfully. Checkpoint cleanup must not invalidate deltas already selected by an in-flight reader.

## What did you see instead?

During a split of source region `4398046511164(1024, 60)`:

1. An asynchronous checkpoint started for versions `[0, 10]`.
2. Repartition remap observed no checkpoint and began replay from delta 0.
3. Checkpoint completed and deleted obsolete deltas.
4. Remap then failed to read delta 0 with persistent `404 NoSuchKey`.
5. MinIO returned `x-amz-delete-marker: true`.
6. Metasrv failed the repartition; the fuzz target panicked because the DDL returned an unexpected error.

This strongly indicates a TOCTOU race: recovery lists deltas before fetching them, while checkpoint cleanup can delete those same objects after publishing a checkpoint.

## What operating system did you use?

Ubuntu 24.04.4 LTS x86_64, GitHub Actions Kind/Kubernetes CI

## What version of GreptimeDB did you use?

Source build at `52ba89bb10d3919e6e5191a120a590a98287230b`

## Relevant log output and stack trace

```text
2026-07-23T12:37:47.018106Z INFO Start doing checkpoint for region
4398046511164(1024, 60), compacted version: [0, 10]

2026-07-23T12:37:47.049512Z INFO Checkpoint not found ... build manifest from scratch
2026-07-23T12:37:47.049636Z INFO Replaying region manifest ... from version 0

2026-07-23T12:37:47.105760Z INFO Checkpoint ... success, version: 10

2026-07-23T12:37:47.113143Z ERROR Remap manifests failed:
NotFound (persistent) at read
S3Error { code: "NoSuchKey", resource: ".../manifest/00000000000000000000.json" }
x-amz-delete-marker: "true"
```

CI job: https://github.com/GreptimeTeam/greptimedb/actions/runs/30004985300/job/89202460122

WenyXu:

Fixes #8625.

Fence asynchronous manifest checkpoint cleanup during graceful region migration and repartition transitions.

A checkpoint task can publish `_last_checkpoint` and then asynchronously delete covered manifest deltas. Previously, migration or repartition could start reading those deltas before the cleanup task finished, so a delta returned by `LIST` could disappear before `GET`.

This change:

- keeps the pending checkpoint `JoinHandle` in the checkpointer and provides a cancellation-safe wait that also reaps completed tasks;
- waits for the pending checkpoint before `EnteringStaging` or `DowngradingLeader` transitions complete, including idempotent downgrade retries;
- prevents normal manifest updates during `Downgrading` from scheduling another checkpoint;
- classifies only a listed manifest delta that disappears before `GET` as retryable, and preserves that retry hint through datanode batch-open replies so the existing migration/repartition procedures retry;
- adds deterministic coverage for checkpoint publication and cleanup barriers, waiter cancellation, handle reaping, staging/downgrade transitions, retry propagation, and transient region-state fallback.

Validation completed:

- `cargo check -p mito2 -p datanode -p meta-srv --tests`
- `cargo clippy -p mito2 -p datanode -p meta-srv --tests -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

Test binaries were not run locally.

No public API, schema, or persisted-data format is changed.

## PR Checklist

### DIFF
--- a/src/meta-srv/src/procedure/repartition/group/sync_region.rs
+++ b/src/meta-srv/src/procedure/repartition/group/sync_region.rs
@@ -350,6 +350,9 @@ impl SyncRegion {
 mod tests {
     use std::assert_matches;
 
+    use common_error::ext::RetryHint;
+    use common_error::status_code::StatusCode;
+    use common_meta::instruction::InstructionError;
     use common_meta::peer::Peer;
     use common_meta::rpc::router::{Region, RegionRoute};
     use store_api::region_engine::SyncRegionFromRequest;
@@ -359,7 +362,9 @@ mod tests {
     use crate::procedure::repartition::group::GroupPrepareResult;
     use crate::procedure::repartition::group::sync_region::SyncRegion;
     use crate::procedure::repartition::test_util::{TestingEnv, new_persistent_context};
-    use crate::procedure::test_util::{new_sync_region_reply, send_mock_reply};
+    use crate::procedure::test_util::{
+        new_sync_region_reply, new_sync_region_reply_with_error, send_mock_reply,
+    };
     use crate::service::mailbox::Channel;
 
     #[test]
@@ -451,6 +456,48 @@ mod tests {
             ..Default::default()
         }];
         let sync_region = SyncRegion { region_routes };
+
+        let err = sync_region.sync_regions(&mut ctx).await.unwrap_err();
+        assert_matches!(err, Error::RetryLater { .. });
+    }
+
+    #[tokio::test]
+    async fn test_sync_regions_retryable_instruction_error() {
+        let mut env = TestingEnv::new();
+        let table_id = 1024;
+        let region_id = RegionId::new(table_id, 3);
+        let mut persistent_context = new_persistent_context(table_id, vec![], vec![]);
+        persistent_context.group_prepare_result = Some(test_prepare_result(table_id));
+
+        let (tx, rx) = tokio::sync::mpsc::channel(1);
+        env.mailbox_ctx
+            .insert_heartbeat_response_receiver(Channel::Datanode(1), tx)
+            .await;
+        send_mock_reply(env.mailbox_ctx.mailbox().clone(), rx, move |id| {
+            Ok(new_sync_region_reply_with_error(
+                id,
+                region_id,
+                false,
+                true,
+                Some(InstructionError {
+                    code: StatusCode::StorageUnavailable,
+                    message: "manifest delta disappeared".to_string(),
+                    retry_hint: RetryHint::Retryable,
+                }),
+            ))
+        });
+
+        let mut ctx = env.create_context(persistent_context);
+        let sync_region = SyncRegion {
+            region_routes: vec![RegionRoute {
+                region: Region {
+                    id: region_id,
+                    ..Default::default()
+                },
+                leader_peer: Some(Peer::empty(1)),
+                ..Default::default()
+            }],
+        };
 
         let err = sync_region.sync_regions(&mut ctx).await.unwrap_err();
         assert_matches!(err, Error::RetryLater { .. });
--- a/src/mito2/src/engine/set_role_state_test.rs
+++ b/src/mito2/src/engine/set_role_state_test.rs
@@ -295,34 +295,8 @@ async fn test_downgrading_waits_for_checkpoint_and_stops_new_checkpoints() {
         "downgrade returned before checkpoint cleanup finished"
     );
 
-    // Cancel the first transition while it is waiting. The checkpoint handle
-    // must remain owned by the manager so an idempotent retry can join it.
-    downgrade.abort();
-    assert!(downgrade.await.unwrap_err().is_cancelled());
-    let retry_wait_started = region
-        .manifest_ctx
-        .manifest_manager
-        .read()
-        .await
-        .checkpointer()
-        .pending_checkpoint_wait_started();
-    let retry_wait_started = retry_wait_started.notified();
-    let cloned_engine = engine.clone();
-    let retry_downgrade = tokio::spawn(async move {
-        cloned_engine
-            .set_region_role_state_gracefully(region_id, SettableRegionRoleState::DowngradingLeader)
-            .await
-    });
-    tokio::time::timeout(Duration::from_secs(5), retry_wait_started)
-        .await
-        .expect("idempotent downgrade retry did not start waiting for the checkpoint");
-    assert!(
-        !retry_downgrade.is_finished(),
-        "idempotent downgrade retry did not wait for checkpoint cleanup"
-    );
-
     blocker.release();
-    assert_success_response(&retry_downgrade.await.unwrap().unwrap(), 2);
+    assert_success_response(&downgrade.await.unwrap().unwrap(), 2);
 
     // Final flush publishes a normal delta, but Downgrading must not start a
     // checkpoint after the barrier.
@@ -388,6 +362,75 @@ async fn test_downgrading_waits_for_checkpoint_and_stops_new_checkpoints() {
         .await
         .wait_for_pending_checkpoint()
         .await;
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_idempotent_downgrading_waits_for_pending_checkpoint() {
+    let (blocker, mock_layer) = CheckpointTaskBlocker::block_cleanup();
+    let mut env = TestEnv::new().await.with_mock_layer(mock_layer);
+    let engine = env
+        .create_engine(MitoConfig {
+            manifest_checkpoint_distance: 1,
+            ..Default::default()
+        })
+        .await;
+    let region_id = RegionId::new(1, 1);
+    let request = CreateRequestBuilder::new().build();
+    let column_schemas = rows_schema(&request);
+    engine
+        .handle_request(region_id, RegionRequest::Create(request))
+        .await
+        .unwrap();
+    put_rows(
+        &engine,
+        region_id,
+        Rows {
+            schema: column_schemas,
+            rows: build_rows(0, 1),
+        },
+    )
+    .await;
+    engine
+        .handle_request(
+            region_id,
+            RegionRequest::Flush(RegionFlushRequest::default()),
+        )
+        .await
+        .unwrap();
+    tokio::time::timeout(Duration::from_secs(5), blocker.wait_until_blocked())
+        .await
+        .expect("checkpoint cleanup did not start");
+
+    // Arrange the state observed by a retried request after the first request
+    // entered Downgrading but its response was lost.
+    engine
+        .set_region_role(region_id, RegionRole::DowngradingLeader)
+        .unwrap();
+    let region = engine.get_region(region_id).unwrap();
+    let wait_started = region
+        .manifest_ctx
+        .manifest_manager
+        .read()
+        .await
+        .checkpointer()
+        .pending_checkpoint_wait_started();
+    let wait_started = wait_started.notified();
+    let cloned_engine = engine.clone();
+    let retry_downgrade = tokio::spawn(async move {
+        cloned_engine
+            .set_region_role_state_gracefully(region_id, SettableRegionRoleState::DowngradingLeader)
+            .await
+    });
+    tokio::time::timeout(Duration::from_secs(5), wait_started)
+        .await
+        .expect("idempotent downgrade did not start waiting for the checkpoint");
+    assert!(
+        !retry_downgrade.is_finished(),
+        "idempotent downgrade did not wait for checkpoint cleanup"
+    );
+
+    blocker.release();
+    assert_success_response(&retry_downgrade.await.unwrap().unwrap(), 1);
 }
 
 #[tokio::test]
--- a/src/mito2/src/engine/set_role_state_test.rs
+++ b/src/mito2/src/engine/set_role_state_test.rs
@@ -365,7 +365,7 @@ async fn test_downgrading_waits_for_checkpoint_and_stops_new_checkpoints() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
-async fn test_idempotent_downgrading_waits_for_pending_checkpoint() {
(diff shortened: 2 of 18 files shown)


WenyXu:
@codex review

src/mito2/src/manifest/recovery.rs
```diff
@@ -0,0 +1,555 @@
...
+            .is_some_and(|checkpoint| previous.is_none_or(|version| checkpoint.version > version)))
+    }
+
+    /// Loads the latest checkpoint and verifies its redundant outer and embedded versions.
+    async fn load_last_checkpoint(&mut self) -> Result<Option<LoadedCheckpoint>> {
+        let Some((_, bytes)) = self.store.load_last_checkpoint_without_tracking().await? else {
+            return Ok(None);
+        };
+        let checkpoint = RegionCheckpoint::decode(&bytes)?;
+        let version = checkpoint.last_version;
```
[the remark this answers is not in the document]
WenyXu: Fixed in `90f60b6397`. Recovery now validates metadata, checkpoint, and embedded manifest versions. Added mismatch and follower rollback tests.

src/mito2/src/manifest/recovery.rs
```diff
@@ -0,0 +1,555 @@
...
+        let manifest =
+            Arc::new(RegionManifestBuilder::with_checkpoint(checkpoint.checkpoint).try_build()?);
+        Ok(Some(LoadedCheckpoint {
+            version,
+            manifest,
+            size: bytes.len() as u64,
```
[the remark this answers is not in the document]
WenyXu: Agreed. This requires coordinated checkpoint, delta, compression, and cache changes, so I’ll address it separately.

src/mito2/src/manifest/recovery.rs
```diff
@@ -0,0 +1,555 @@
...
+                OpenAttempt::Retry => {
+                    if self.retry_state.can_retry() {
+                        continue;
+                    }
+                    return Err(self.retry_state.terminal_error(
+                        UnexpectedSnafu {
+                            reason: "checkpoint kept advancing during manifest recovery"
+                                .to_string(),
+                        }
+                        .build(),
```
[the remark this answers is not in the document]
WenyXu: Fixed in `90f60b6397`. Recovery exhaustion now returns a retryable `RegionNotReady` error, with test coverage.

src/mito2/src/manifest/recovery.rs
```diff
@@ -0,0 +1,555 @@
...
+    }
+
+    /// Recovers a manifest for a fresh reader.
+    pub(crate) async fn recover_open(&mut self) -> Result<Option<RecoveryCandidate>> {
+        loop {
+            self.retry_state.attempts += 1;
+            let attempt = self
+                .open_attempt()
+                .await
+                .map_err(|error| self.retry_state.terminal_error(error))?;
```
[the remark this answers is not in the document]
WenyXu: Fixed in `90f60b6397`. Checkpoint read/decode errors now propagate directly; the original delta `NotFound` is retained only when no newer checkpoint is available. Added regression coverage.

WenyXu:
@codex review

WenyXu:
@codex review

src/mito2/src/manifest/recovery.rs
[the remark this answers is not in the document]
WenyXu: Update: this was fixed in `81befad426`. Checkpoint recovery now carries the stored `content_length` through `CheckpointLoad`, including cache hits, and publishes that value instead of the decompressed payload length. Added gzip reopen and cache-hit regression tests.

src/mito2/src/manifest/recovery.rs
```diff
@@ -0,0 +1,632 @@
...
+    async fn install_attempt(
+        &mut self,
+        base: Arc<RegionManifest>,
+        start: ManifestVersion,
+        target: ManifestVersion,
+    ) -> Result<InstallAttempt> {
+        // Repartition writes intermediate versions to staging storage and persists their merged
+        // transition as a later regular delta. The regular delta range therefore need not start
+        // at `start`; reaching `target` and passing the checkpoint fence validate the attempt.
+        let fetch = self.fetch_deltas(start, next_version(target)?).await?;
```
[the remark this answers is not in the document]
WenyXu: Thanks. We checked this against the existing manifest-storage invariants. A covered normal delta is deleted only after `_last_checkpoint` is published, so if cleanup removes the leading delta while a later delta remains, the post-fetch checkpoint fence observes the newer checkpoint and retries from that base. Repartition gaps are different: skipped versions exist only in staging, and the later normal delta contains the merged transition. Requiring the first listed delta to equal `start` would reject that valid path. An incomplete LIST that omits an existing normal delta would violate the same storage assumption already used by open recovery, so we don't add a separate check here. The cleanup-before-LIST and LIST-to-GET race tests cover the reachable cases.

src/mito2/src/manifest/checkpointer.rs
```diff
@@ -122,6 +122,16 @@ impl Checkpointer {
...
 
+    /// Advances the latest durable checkpoint observed by this manager.
+    ///
+    /// A follower can install a checkpoint written by another manager. Remembering that version
+    /// avoids checkpointing the same manifest range if the follower is later promoted to leader.
+    pub(crate) fn advance_last_checkpoint_version(&self, version: ManifestVersion) {
+        self.inner
+            .last_checkpoint_version
+            .fetch_max(version, Ordering::Relaxed);
+    }
```
evenyag: Can we provide a way to stop/abort the checkpointer so we can ensure there is no checkpointer running?

Then we don't need to mix recovery with checkpointing.

WenyXu:
@codex review

src/mito2/src/region.rs
```diff
@@ -588,6 +592,13 @@ impl MitoRegion {
             }
         };
 
+        // The state is changed before waiting, so no new writable-leader work
+        // can race with the barrier. Keep the manager lock while joining to
+        // serialize the barrier with checkpoint scheduling.
+        if wait_for_checkpoint {
+            manager.wait_for_pending_checkpoint().await;
```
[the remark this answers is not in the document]
WenyXu: Fixed in `b926eee68b`. Graceful transitions from any leader state to `Follower` now join the pending checkpoint after changing the role, so no new leader work can race with the barrier. Added a deterministic cleanup-gate test that verifies the transition does not return before checkpoint cleanup finishes. `cargo nextest run -p mito2`: 1140 passed.

WenyXu:
@codex review

src/mito2/src/error.rs
```diff
@@ -1706,4 +1724,19 @@ mod tests {
         let error = RegionMetadataNotFoundSnafu {}.build();
         assert!(error.may_have_persisted_manifest_update());
     }
+
+    #[test]
+    fn test_manifest_delta_not_found_retry_hint() {
```
evenyag: Don't need to add a test for it.

src/mito2/src/manifest/checkpointer.rs
```diff
@@ -31,22 +34,20 @@ use crate::metrics::MANIFEST_OP_ELAPSED;
 pub(crate) struct Checkpointer {
     manifest_options: RegionManifestOptions,
     inner: Arc<Inner>,
+    pending_checkpoint: Option<JoinHandle<()>>,
```
evenyag: Looks like this is the current running checkpoint task, not a pending task.

src/mito2/src/manifest/checkpointer.rs
```diff
@@ -181,17 +182,44 @@ impl Checkpointer {
         self.do_checkpoint(checkpoint);
     }
 
-    fn do_checkpoint(&self, checkpoint: RegionCheckpoint) {
-        self.inner.set_doing_checkpoint();
-
+    fn do_checkpoint(&mut self, checkpoint: RegionCheckpoint) {
```
evenyag, line 185: non-blocking: Do we need an assert on `self.pending_checkpoint`?

src/mito2/src/manifest/checkpointer.rs
```diff
@@ -181,17 +182,44 @@ impl Checkpointer {
...
+            return;
+        };
+
+        #[cfg(test)]
+        self.pending_checkpoint_wait_started.notify_one();
+
+        let result = (&mut *handle).await;
+        // There is no cancellation point between observing completion and
+        // clearing the handle.
+        self.pending_checkpoint = None;
```
evenyag: Can we just take the handle from pending_checkpoint?

src/mito2/src/manifest/tests/checkpoint.rs
evenyag, line 1: Is it possible to merge some tests or reduce some tests?

src/mito2/src/engine/set_role_state_test.rs
evenyag, line 1: Can we simplify the tests?

src/meta-srv/src/procedure/repartition/group/sync_region.rs
```diff
@@ -455,4 +460,46 @@ mod tests {
...
         assert_matches!(err, Error::RetryLater { .. });
     }
+
+    #[tokio::test]
+    async fn test_sync_regions_retryable_instruction_error() {
+        let mut env = TestingEnv::new();
+        let table_id = 1024;
+        let region_id = RegionId::new(table_id, 3);
+        let mut persistent_context = new_persistent_context(table_id, vec![], vec![]);
+        persistent_context.group_prepare_result = Some(test_prepare_result(table_id));
```
evenyag, line 470: Is it necessary to test this? Can we merge it with other existing tests?

src/mito2/src/manifest/checkpointer.rs
WenyXu: Rather than aborting the checkpointer, this change uses its `JoinHandle` as the lifecycle fence. `abort()` is non-blocking, and a task may already have published `_last_checkpoint` while cleanup is still pending; the caller would still need to await the handle to establish a safe boundary. Waiting preserves both publication and cleanup ordering.

src/mito2/src/error.rs
WenyXu: Removed in `230605fffa`. The manifest recovery regression test already exercises `ManifestDeltaNotFound` through the actual LIST-to-GET path and asserts both `StorageUnavailable` and `Retryable`, so the direct `ErrorExt` mapping test was redundant.

src/mito2/src/manifest/checkpointer.rs
WenyXu: Renamed the field to `checkpoint_task` in `230605fffa`. The stored handle may be executing or finished but not yet reaped, so `running_checkpoint` would also be imprecise.

src/mito2/src/manifest/checkpointer.rs
WenyXu: I left this without an extra assertion. `do_checkpoint` is private, and `maybe_do_checkpoint` either returns for a running handle or reaps a completed handle before calling it, so an assertion would not enforce an additional invariant or add coverage.

src/mito2/src/manifest/checkpointer.rs
WenyXu: We cannot take the handle before awaiting it. A cancelled waiter would then drop its local Tokio `JoinHandle`, detaching the checkpoint task and losing the manager's only handle; a retried transition could no longer join the same cleanup. Borrowing the handle keeps it owned by `Checkpointer` until the await finishes.

src/mito2/src/manifest/tests/checkpoint.rs
WenyXu: Reduced the overlapping coverage in `cd71ba55c9`: removed the staging-only retry-classification test and collapsed the EnterStaging matrix to `UpdatePartitionExpr` cases for checkpoint publication and cleanup. Normal open/install recovery and each checkpoint-task lifecycle branch remain covered.

src/mito2/src/engine/set_role_state_test.rs
WenyXu: Also removed the out-of-scope Alter/Downgrade handler test and its listener test seam in `cd71ba55c9`. The remaining set-role tests cover distinct state rules: downgrading stops and later resumes checkpoint scheduling, direct follower transitions wait, and a cancelled downgrade retry waits the retained handle. Combining them would obscure which transition regressed.

src/meta-srv/src/procedure/repartition/group/sync_region.rs
WenyXu: I kept this test separate. `test_sync_regions_retryable` covers a mailbox timeout; this test covers a retryable `InstructionError` carried in a `SyncRegionsReply`, including JSON reply parsing and `handle_sync_region_reply`'s error branch. The OpenCandidate test exercises the shared conversion helper but not this reply shape.

evenyag (approved):

PR #8848, opened 2026-08-11
feat(promql): support native histogram aggregations

shuiyisong:

This is the 5th PR of the 6 PRs for native histogram.

This pull request adds PromQL aggregation support for native histograms.
- Supports histogram-aware `sum`, `avg`, `count`, `group`, and `count_values` planning for histogram-only and mixed inputs.
- Drops incompatible mixed results or unsupported histogram samples with appropriate warnings and info annotations, while filtering empty groups.
- Registers histogram aggregate functions for distributed plan decoding and preserves overflow-safe, correctly weighted averages across partial states.

## PR Checklist

### DIFF
--- a/src/promql/src/functions/native_histogram.rs
+++ b/src/promql/src/functions/native_histogram.rs
@@ -405,10 +405,8 @@ fn histogram_pair_udf_with_collector(
                     match (read_histogram(lhs, row)?, read_histogram(rhs, row)?) {
                         (Some(lhs), Some(rhs)) => {
                             record_custom_reconciliation(&collector, name, &lhs, &rhs);
+                            record_counter_reset_contradiction(&collector, name, &lhs, &rhs);
                             let result = op(&lhs, &rhs);
-                            if result.is_some() {
-                                record_counter_reset_contradiction(&collector, name, &lhs, &rhs);
-                            }
                             if result.is_none() {
                                 record_warning(
                                     &collector,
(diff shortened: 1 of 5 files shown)


src/promql/src/functions/native_histogram.rs
```diff
@@ -845,19 +925,38 @@ impl NativeHistogramAggregateAccumulator {
         }
     }
 
-    fn push_histogram(&mut self, histogram: NativeHistogram, count: u64) {
+    fn push_histogram(&mut self, histogram: NativeHistogram, count: u64) -> DfResult<()> {
         if self.dropped_incompatible {
```
evenyag: Once dropped_incompatible is set, this returns before observe_reset_hints, so later CounterReset/NotCounterReset samples cannot trigger the required contradiction warning. This makes annotations input-order dependent: for example, two incompatible same-hint histograms followed by an opposing hint emit only the schema warning. The same ordering issue exists in merge_batch, where hint flags from dropped partial states are skipped. Could we merge/observe reset flags before the early exits and cover both sequential and partial-state cases? The Prometheus native histogram specification says directly contradictory hints should warn in any case: https://prometheus.io/docs/specs/native_histograms/#aggregation-operators

shuiyisong: Fixed in f520b0ce6a. push_histogram now observes reset hints before the dropped-state early return, and merge_batch observes every partial state flag before honoring dropped. range_fold_histograms checks the complete input for contradictory hints before schema folding, and binary add/sub record the contradiction before reconciliation can drop the result.

Added regression coverage for sequential inputs and dropped partial states in both merge orders for sum/avg, plus incompatible sum_over_time/avg_over_time. Local validation passed: cargo nextest run -p promql (212 tests) and the targeted real RecordBatch query test.

src/query/src/promql/planner.rs
evenyag, line 1: The histogram-related plans are unclear, and we don't have enough comments to describe what the plan looks like.

However, the other parts of the planner also have this issue.

evenyag (approved):

PR #8852, opened 2026-08-11
fix: harden permission checks and process visibility

shuiyisong:

This pull request fixes authorization gaps across static credentials and SQL execution. It rejects invalid permission modes instead of silently granting read-write access, checks gRPC SQL permissions after parsing to cover dependencies such as `INSERT ... SELECT`, registers the `iceberg.read` action, and limits non-admin `SHOW [FULL] PROCESSLIST` results to the current catalog while preserving global visibility for admins.

Malformed permission-mode credentials previously accepted as read-write are now rejected or ignored.

## PR Checklist


src/auth/src/permission.rs
```diff
@@ -139,6 +139,7 @@ pub const PROMQL_QUERY: PermissionAction = PermissionAction::read("promql.query"
 pub const LOG_QUERY: PermissionAction = PermissionAction::read("log.query");
 pub const OPENTSDB_WRITE: PermissionAction = PermissionAction::write("opentsdb.write");
 pub const INFLUXDB_WRITE: PermissionAction = PermissionAction::write("influxdb.write");
+pub const ICEBERG_READ: PermissionAction = PermissionAction::read("iceberg.read");
```
killme2008, line 142: Need to feature gate by `enterprise`?

src/frontend/src/instance/grpc.rs
```diff
@@ -71,15 +71,21 @@ impl GrpcQueryHandler for Instance {
...
-            self.plugins
-                .get::<PermissionCheckerRef>()
-                .as_ref()
-                .check_permission_with_context(
-                    ctx.current_user(),
-                    PermissionReq::GrpcRequest(&request),
-                    Some(&ctx.current_schema()),
-                )
-                .context(PermissionSnafu)?;
+            if !matches!(
```
killme2008, line 74: Why change this behavior?

Only check permission for query requests.

src/auth/src/permission.rs
shuiyisong: The iceberg feature is not implemented in the OSS codebase; the OSS user cannot use the iceberg feature anyway. Also there's no feature gate on the OSS iceberg code. I don't think we need to gate the code here either.

src/frontend/src/instance/grpc.rs
shuiyisong: We need to distinguish write requests and DDLs as well. Permissions like `write_only` or disable certain DDL would require to distinguish the request and check detailed permission situations.

killme2008 (approved):

PR #8853, opened 2026-08-11
docs: align the entity-graph agent vocabulary with GenAI semantic conventions

killme2008:
I hereby agree to the terms of the GreptimeDB CLA.

#8609

RFC-only revision of the entity-graph RFC, from studying real agent telemetry (Claude Code / Codex / OpenClaw via OTLP) and spec-compliant GenAI semconv traces ingested through `greptime_trace_v1`:

- Agent-telemetry entity types follow the GenAI semantic-convention namespace as written: `gen_ai.agent` / `gen_ai.model` / `gen_ai.tool` (declarations `entity.gen_ai.agent.id`, ...). One sentence notes that most agent telemetry today carries ad-hoc attributes, which explicit declarations map onto these types.
- The relationship vocabulary gains the `uses` / `invokes` rows (inverse `used_by` / `invoked_by`); the agent-edge wording is aligned (`invoked` -> `invokes`, matching #8836).
- Entity columns are no longer required to be tags — only column existence is validated. The flattened `gen_ai.*` columns on trace tables are fields, so the tag rule locked real trace tables out of declaring entities while buying no correctness. The implementation lands with the convention-pack PR.
- Two future-work entries: parent-child span derivation of agent edges (the conventions put tool/model identity and agent identity on different spans, so same-row co-occurrence cannot witness them), and `ALTER TABLE` support for `greptime.semantic.*` options so declarations can be retrofitted onto existing tables.

## PR Checklist

- [x] Please confirm that the relevant document has been updated or no document update is required.
- [x] This PR requires no breaking change note.

### DIFF
--- a/docs/rfcs/2026-06-25-entity-relationships-and-graph-query.md
+++ b/docs/rfcs/2026-06-25-entity-relationships-and-graph-query.md
@@ -90,7 +90,7 @@ greptime.semantic.entity.<entity_type>.descriptive = comma-separated column name
 greptime.semantic.entity.<entity_type>.scope       = comma-separated column names   (optional)
 ```
 
-Values are column names, and **`id` columns must be tag/primary-key columns** so identity stays indexable and joinable; identity buried in a JSON attribute bag must be projected to a column first via a pipeline. A latency metric table that describes both a `service` and the `host` it runs on:
+Values are column names — tag or field: trace pipelines flatten identifying attributes (e.g. `span_attributes.gen_ai.agent.id`) into field columns, and the read-time derivation works on any column. Identity buried in a JSON attribute bag must be projected to a column first via a pipeline. A latency metric table that describes both a `service` and the `host` it runs on:
 
 ```sql
 CREATE TABLE app_request_latency (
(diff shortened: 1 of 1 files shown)


docs/rfcs/2026-06-25-entity-relationships-and-graph-query.md
```diff
@@ -90,7 +90,7 @@ greptime.semantic.entity.<entity_type>.descriptive = comma-separated column name
 greptime.semantic.entity.<entity_type>.scope       = comma-separated column names   (optional)
 ```
 
-Values are column names, and **`id` columns must be tag/primary-key columns** so identity stays indexable and joinable; identity buried in a JSON attribute bag must be projected to a column first via a pipeline. A latency metric table that describes both a `service` and the `host` it runs on:
+Values are column names — tag or field: trace pipelines flatten identifying attributes (e.g. `span_attributes.gen_ai.agent.id`) into field columns, and the read-time derivation works on any column. Identity buried in a JSON attribute bag must be projected to a column first via a pipeline. A latency metric table that describes both a `service` and the `host` it runs on:
```
shuiyisong, line 93: [P3] Match the implemented stable-string constraint

`any column` is broader than the DDL contract already shipped in #8854. `validate_entity_semantic_options` rejects Binary, Json, Vector, List, Struct, Dictionary, and Null because every entity role must have a stable string representation. As written, the RFC can lead users to declare a field that CREATE or ALTER rejects. Could this say “any column with a stable string representation” and update M0 to mention that validation?

docs/rfcs/2026-06-25-entity-relationships-and-graph-query.md
```diff
@@ -160,10 +160,12 @@ The relationship vocabulary is small, typed, and inverse-paired; stored directio
...
+| `invokes` | `gen_ai.agent` invokes `gen_ai.tool` | trace | `invoked_by` |
 | `depends_on` | logical/declared dependency | declared | `dependency_of` |
 | `owns` | team/service owns dst | declared | `owned_by` |
 
-`calls` lives at the *logical* `service` layer; `runs_on`/`contains` live at the *runtime* `service.instance`/`process`/`k8s.pod` layer, with `part_of` linking the two; this avoids "one logical service runs_on five hosts". A custom `rel_type` is just a string; only derivation rules and inverse names are built-in. The trace-derived `calls` edge is qualified by attributes such as `connection_type` (`database`, `messaging_system`, or `virtual_node`) rather than exploded into many edge types.
+`calls` lives at the *logical* `service` layer; `runs_on`/`contains` live at the *runtime* `service.instance`/`process`/`k8s.pod` layer, with `part_of` linking the two; this avoids "one logical service runs_on five hosts". A custom `rel_type` is just a string; only derivation rules and inverse names are built-in. The trace-derived `calls` edge is qualified by attributes such as `connection_type` (`database`, `messaging`, or `virtual_node`) rather than exploded into many edge types.
```
shuiyisong, line 168: [P3] Keep the established `messaging_system` value

Tempo and the OpenTelemetry service-graph connector use `messaging_system` alongside `database` and `virtual_node`; changing only this RFC to `messaging` defines a different contract for the deferred messaging implementation and downstream consumers. Since there is no runtime migration here, could we retain `messaging_system`? References: [Tempo service graphs](https://grafana.com/docs/tempo/latest/metrics-from-traces/service_graphs/) and [OTel servicegraph connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/servicegraphconnector/README.md).

PR #8854, opened 2026-08-11
feat: embedded convention pack for the entity graph (prom/k8s, gen_ai naming)

killme2008:
I hereby agree to the terms of the GreptimeDB CLA.

#8609. RFC counterpart: #8853.

The built-in conventions of the entity graph move into an embedded `conventions.yaml`, and the graph now connects traces and Kubernetes metrics without any user declaration:

```
service ──calls──> service
   ↑ part_of
service.instance ──runs_on──> k8s.pod ──runs_on──> k8s.node
                                ├──part_of──> k8s.workload
                                └──contains─> k8s.container
```

- `conventions.yaml` (embedded, not operator-editable): co-declared edge rules, virtual-destination candidates, implicit entity declarations for trace-v1 tables and for whitelisted kube-state-metrics / `target_info` tables. Parsed once, validated against the vocabulary constants; a broken file propagates as a plan error.
- Pod identity is the UID. Every whitelisted KSM metric carries it, and trace-v1 tables flatten `k8s.pod.uid`, so both sides land on one entity. Names stay descriptive.
- Trace-v1 tables get implicit `service` / `service.instance` / `k8s.pod` declarations per column actually present; the hardcoded service fallback is gone. No name-based guessing when a column is absent.
- The agent vocabulary follows GenAI semantic conventions: `gen_ai.agent` / `gen_ai.model` / `gen_ai.tool`.
- Entity columns are no longer required to be tags; instead their types must render as strings, validated on CREATE and ALTER.
- Empty identity components identify nothing (KSM emits empty `node` / `owner_*` for unscheduled/ownerless pods).
- Graph column names are single-sourced in common-catalog; the derivation code anchors on vocabulary constants.

Deferred with reasons: generic per-table label rules (plan size grows linearly with metric-table count), attribute alias for cross-signal `service.instance` merging (composite ids embed physical column names), messaging virtual destinations (need composite identity).

Known limitations: `k8s.node`/`k8s.workload` are name-scoped, so multiple clusters in one database merge same-named nodes/workloads; `kube_pod_info`'s `runs_on` means "scheduled on".

## PR Checklist

- [x] This PR requires no breaking change note.
- [x] Please confirm that the relevant document has been updated or no document update is required (#8853).


shuiyisong:
Follow-up review against live head `16cdcfcb84b0435c9d8c04fd67c351f6d4f40594`:

1. **Nit — legacy `agent` declarations no longer feed agent edges.** [`entity_graph.rs:396`](https://github.com/GreptimeTeam/greptimedb/blob/16cdcfcb84b0435c9d8c04fd67c351f6d4f40594/src/frontend/src/instance/entity_graph.rs#L396) searches only for `gen_ai.agent`, whereas `main` searches for `agent`. Existing tables using `greptime.semantic.entity.agent.*` therefore still appear in the registry but stop contributing agent-derived edges. If this unreleased feature has no compatibility contract, no change is required; recording this as a non-blocking nit.

2. **Resolved — `k8s.service` identity.** Commit `16cdcfcb` changes `kube_service_info` to identify services by `uid`, preventing recreation and cross-cluster name collisions.

3. **[P1] Add a cluster discriminator to KSM node and workload identities.** [`conventions.yaml:72-91`](https://github.com/GreptimeTeam/greptimedb/blob/16cdcfcb84b0435c9d8c04fd67c351f6d4f40594/src/operator/src/statement/semantic_graph/conventions.yaml#L72-L91) still identifies nodes as `[node]` and workloads as `[namespace, owner_kind, owner_name]`. Kubernetes names are unique only within a cluster: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/.

A concrete same-table remote-write case is:

- cluster A: pod UID `pod-a`, node `worker-0`, owner `default/ReplicaSet/api`
- cluster B: pod UID `pod-b`, node `worker-0`, owner `default/ReplicaSet/api`

The derivation emits both `pod-a runs_on worker-0` and `pod-b runs_on worker-0`, plus the same workload ID for both owners. `extend_with_implicit_entities` leaves scope empty, and both registry IDs and co-declared edge endpoints are rendered only from these `id` columns, so an external `cluster` label is ignored. Traversal therefore treats two real nodes/workloads as one and can produce false cross-cluster topology.

The cluster component must be consistent across `kube_pod_info`, `kube_node_info`, and `kube_pod_owner`; changing only the `kube_node_info` identity to `system_uuid` would leave pod `runs_on` endpoints unmatched. A shared conventional remote-write cluster label or an explicit single-cluster limitation plus required overrides would avoid silently wrong identities.

killme2008:
Thanks for the careful follow-up.  @shuiyisong 

1. Correct, but nothing released carries the old `agent` naming — #8836 merged after v1.2.0-beta.1 was cut, so there's no compat contract to keep. Staying with the clean rename.

3. The problem is there's no standardized cluster label to build on: kube-state-metrics doesn't emit one, it comes from `external_labels` / relabeling and the name varies per deployment (`cluster`, `k8s_cluster_name`, ...). Making identity conditional on whichever label happens to exist would fork entity identity across deployments and across the three tables — worse than the collision. Our recommendation is one database per cluster, which is why this is listed under Known limitations; explicit declarations are the override, and making that practical on auto-created tables is exactly the ALTER follow-up in #8609.

sunng87 (approved):

PR #3175, opened 2024-01-16
feat: make query be aware of timezone setting

killme2008:
I hereby agree to the terms of the [GreptimeDB CLA](https://gist.github.com/xtang/6378857777706e568c1949c7578592cc)

## What's changed and what's your intention?

Try to aware of timezone setting for `TypeConversionRule`, changes:

* Added `ExtensionAnalyzerRule` that extends DataFusion `AnalyzerRule` to accept `QueryEngineContext` as the second parameter.
* Implement `ExtensionAnalyzerRule` for `TypeConversionRule`.
* Optimize by extension analyzer rules at first in `DfLogicalPlanner` and `LogicalOptimizer`.
* Fixed MySQL `show variables time_zone` can't work.
* Adds timezone sqlness test.
* Make `Date`, `Timestamp` and `DateTime` parsing from string value aware of timezone.
* Make UDF functions aware of timezone.

## Checklist

- [x]  I have written the necessary rustdoc comments.
- [x]  I have added the necessary unit tests and integration tests.
- [ ]  This PR does not require documentation updates.

## Refer to a related PR or issue link (optional)

#2907

PR #8859, opened 2026-08-12
fix(query): respect query timezone in timestamp casts

### LINKED ISSUE #8857: Incorrectly parsed time string literal when "now()" is also in the insert sql
### What type of bug is this?

Incorrect result

Query Engine

### Minimal reproduce step

Start GreptimeDB with `GREPTIMEDB_STANDALONE__DEFAULT_TIMEZONE=Asia/Shanghai`, then:

```text
mysql> select timezone();
+---------------+
| timezone()    |
+---------------+
| Asia/Shanghai |
+---------------+
1 row in set (0.00 sec)

mysql> create table foo (ts timestamp time index, st timestamp);
Query OK, 0 rows affected (0.01 sec)

mysql> insert into foo (ts) values ('2026-08-01 12:00:00.001');
Query OK, 1 row affected (0.00 sec)

mysql> insert into foo (ts, st) values ('2026-08-02 12:00:00.001', now());
Query OK, 1 row affected (0.00 sec)

mysql> insert into foo (ts, st) values ('2026-08-03 12:00:00.001', '2026-08-12 14:00:00.001');
Query OK, 1 row affected (0.01 sec)

mysql> select * from foo order by ts;
+-------------------------+-------------------------+
| ts                      | st                      |
+-------------------------+-------------------------+
| 2026-08-01 12:00:00.001 | NULL                    |
| 2026-08-02 20:00:00.001 | 2026-08-12 15:07:00.717 |
| 2026-08-03 12:00:00.001 | 2026-08-12 14:00:00.001 |
+-------------------------+-------------------------+
3 rows in set (0.00 sec)
```

Note that the second insert uses "now()", then the query result of "ts" is wrong. Clearly the "ts" is parsed as UTC timezone instead of UTC+8(Asia/Shanghai) timezone.

now() does not affect other time string literals' parsing

### What did you see instead?

the opposite

ubuntu

greptime-registry.cn-hangzhou.cr.aliyuncs.com/greptime/greptimedb-nightly:nightly-20260812-943eee852

```bash

```

fzlzjerry:

Closes #8857.

Timestamp strings in a literal-only `INSERT` are parsed by the statement converter with the session timezone. Adding an expression such as `now()` routes the row through DataFusion, where destination-column coercion can leave the string as `CAST(Utf8 AS Timestamp)`. Arrow then interprets that cast as UTC, so the same input changes meaning depending on the neighboring expressions.

This change handles string-literal timestamp casts at the generic `TypeConverter` boundary and parses them with `Timestamp::from_str(..., QueryContext::timezone())`. Timestamp comparison literals and explicit casts share one helper that converts the parsed value to the target timestamp precision.

For `INSERT ... SELECT`, DataFusion represents destination coercion as a parent `CAST(column AS Timestamp)` over a child projection containing the literal. The analyzer therefore resolves only direct or aliased literal outputs from that child projection by schema index; it does not inspect the enclosing DML shape or evaluate non-literal expressions.

Coverage includes:

- literal-only `INSERT` and `VALUES + now()` under `Asia/Shanghai`;
- `INSERT ... SELECT` with a timestamp string and `now()`;
- `TIMESTAMP(9)` input preserving nanosecond precision;
- planner and type-conversion regressions for the timezone and target precision.

There are no API, schema, or stored-data format changes.

Validation:

- `cargo nextest run -p query` (648 passed, 2 skipped)
- `cargo sqlness bare -t insert_default_timezone` (standalone and distributed passed)
- `cargo clippy -p query --all-targets -- -D warnings`
- `python3 scripts/check-snafu.py`
- `python3 scripts/check-super-imports.py`

## PR Checklist


killme2008:
The expected result in `tests/cases/standalone/common/insert/insert_default_timezone.result` is correct: `2026-08-02 12:00:00.001` in Asia/Shanghai represents `2026-08-02T04:00:00.001` UTC.

I think the implementation should be reworked before merge. The new DML arm in `src/query/src/optimizer/type_conversion.rs` searches the INSERT input for `Values` nodes and then rewrites one exact expression shape. This duplicates `TypeConverter::cast_scalar_value()` in the same file, depends on DataFusion currently placing the coercion inside `Values`, and misses the equivalent query path:

```sql
INSERT INTO timestamps (ts, st)
SELECT <timestamp string>, now();
```

That plan uses a projection rather than a `Values` node, so the string still reaches Arrow as `CAST(Utf8 AS Timestamp)` and is interpreted as UTC.

A smaller and more general change would be:

1. Remove the INSERT-specific DML traversal and `InsertValueConverter` from `src/query/src/optimizer/type_conversion.rs`.
2. Add handling for `Cast(Utf8 literal AS Timestamp)` to `TypeConverter::f_up()` in the same file.
3. Extract the timestamp parsing and target-precision cast into one helper shared with `TypeConverter::cast_scalar_value()`.
4. Keep `Timestamp::from_str(..., QueryContext::timezone())` from `src/common/time/src/timestamp.rs` as the source of parsing semantics.

The existing `transform_up_with_subqueries` traversal in `TypeConversionRule::analyze()` already visits expressions in `Values`, projections, joins, and subqueries, so this avoids depending on the enclosing DML shape.

This would make explicit string-to-timestamp casts follow the same session-timezone rule as INSERT literals and timestamp comparisons. I believe that is the consistent behavior intended by #3175. If explicit `CAST` is meant to retain different semantics, `src/query/src/planner.rs` needs to preserve whether a cast came from destination-column coercion; after SQL-to-rel, inferring that from `Dml -> Values` is fragile.

For coverage, I suggest keeping the planner regression in `src/query/src/planner.rs` and extending `tests/cases/standalone/common/insert/insert_default_timezone.sql` with:

- literal-only INSERT and `VALUES + now()` side by side under Asia/Shanghai;
- `INSERT ... SELECT` with the same timestamp;
- one `TIMESTAMP(9)` value to ensure the rewrite preserves target precision.

Thanks for picking this up. The reproduction and the current sqlness expectation are both useful; the main issue is where the conversion is applied.

fzlzjerry:
Reworked this in `925f9b128` around the generic type-conversion boundary:

- removed the INSERT-specific DML/`Values` traversal and `InsertValueConverter`;
- taught `TypeConverter::f_up()` to normalize string-literal timestamp casts with the `QueryContext` timezone;
- shared the timestamp parsing and target-precision conversion helper with comparison-literal conversion;
- kept `Timestamp::from_str` as the source of timezone parsing semantics.

The `INSERT ... SELECT` plan has one additional wrinkle: DataFusion expresses the destination cast as `CAST(column AS Timestamp)` in a parent projection, while the string literal is emitted by the child projection. The analyzer now resolves only direct or aliased literal child outputs by schema index. It remains independent of the enclosing DML shape and does not evaluate functions or other expressions.

The SQLness coverage now includes literal-only INSERT, `VALUES + now()`, `INSERT ... SELECT`, and `TIMESTAMP(9)`. Final validation passed:

- `cargo nextest run -p query` — 648 passed, 2 skipped
- `cargo sqlness bare -t insert_default_timezone` — standalone and distributed passed
- `cargo clippy -p query --all-targets -- -D warnings`
- both repository import/style scripts

killme2008:
Thanks for reworking this and adding the extra coverage. The new `insert_default_timezone` results look correct.

There are still two blocking regressions in the full sqlness run:

- Handling every `Cast(Utf8 literal AS Timestamp)` in `TypeConverter::f_up()` also changes explicit casts and typed timestamp literals in normal queries. This breaks existing parsing behavior and even causes duplicate output field names in `timestamp.sql`.
- Reusing `cast_string_to_timestamp()` for comparisons casts the literal to the column precision. For a `TIMESTAMP(3)` column, `.000000195` is truncated to `.000`, which changes the result of `>=` in `ts_precision_comparison.sql`.

I think `TypeConversionRule` is still a reasonable place for the fix, but the new conversion should be scoped to assignment casts under `LogicalPlan::Dml(WriteOp::Insert)`. Predicate literals should keep their original precision, and normal explicit casts should stay on the existing DataFusion path.

fzlzjerry:
Reworked this in `9fe364cad` to scope timezone-aware folding to INSERT assignment casts under `LogicalPlan::Dml(WriteOp::Insert)`:

- only the outer assignment projection and its direct `Values` source are considered;
- source-query expressions and ordinary explicit timestamp casts remain on DataFusion's existing path;
- comparison literals retain their parsed precision instead of being truncated to the column precision;
- unsupported or out-of-range timestamp syntax falls back to DataFusion's parser.

The existing coverage for literal-only `VALUES`, `VALUES + now()`, `INSERT ... SELECT`, and `TIMESTAMP(9)` remains in place. Validation passed:

- `cargo nextest run -p query` — 650 passed, 2 skipped
- affected timestamp SQLness cases — all passed in standalone and distributed
- `cargo sqlness bare -t '^standalone:insert_default_timezone$'`
- `cargo clippy -p query --all-targets -- -D warnings`
- both repository import/style scripts

The branch is also synced with the current `main`.

killme2008:
@fzlzjerry Please make sure the tests(including sqlness tests) passed in your machine before asking review. Thanks

fzlzjerry:
Confirmed on the current head (`901ca2376`) with the full, unfiltered local suites:

- `cargo nextest run --build-jobs 1 --test-threads 4` — 7,650 passed, 15 skipped
- `cargo sqlness bare` — the complete standalone and distributed SQLness run passed (`All sqlness tests passed!`)

The worktree remained clean after both runs. Thanks for the reminder.

src/query/src/optimizer/type_conversion.rs
```diff
@@ -107,10 +109,8 @@ impl ExtensionAnalyzerRule for TypeConversionRule {
                 let Ok(schema) = join.left.schema().join(join.right.schema()) else {
                     return Ok(Transformed::no(LogicalPlan::Join(join)));
                 };
-                let mut converter = TypeConverter {
-                    schema: std::sync::Arc::new(schema),
-                    query_ctx: ctx.query_ctx(),
-                };
+                let mut converter =
+                    TypeConverter::new(std::sync::Arc::new(schema), ctx.query_ctx());
```
killme2008: Perfer to `use std::sync::Arc;`

Than change to `Arc::new(schema).

The same suggestion in other places in this file.

killme2008 (commented):
I think this is a good scoped fix for #8857 and is safe to merge. Thanks.

I found one limitation worth tracking separately. The rewrite currently handles direct VALUES and direct child projections, but wrapped INSERT ... SELECT plans can still leave the assignment as `CAST(Utf8 AS Timestamp)`. For example:

```sql
  INSERT INTO t
  SELECT '2026-08-02 12:00:00.001', now()
  LIMIT 1;
```

The same applies to `UNION ALL` and subquery wrappers. These cases were already broken before this PR, so I don't think they should block the current fix. We should open a follow-up to make the rewrite follow output-column lineage instead of depending on the immediate plan shape.

src/query/src/optimizer/type_conversion.rs
fzlzjerry: Updated in 3ed6a7c63: `Arc` is now imported at module scope, and the two fully qualified uses in this file now use `Arc::new(...)`.

Validation passed:
- `cargo clippy -p query --all-targets -- -D warnings`
- `cargo nextest run -p query` (650 passed, 2 skipped)

killme2008 (approved):
LGTM

Thanks

PR #8862, opened 2026-08-13
fix(mito2): publish committed sequence only after rows are installed

discord9:

PR 1/5 of the incremental Flow data-plane series: #8862 → #8865 → #8866 (with #8863, #8864 in parallel).

### Problem

The region's committed-sequence watermark was published inside `write_memtable()`, before bulk parts were installed into the memtable. A scan opening a snapshot could bind H to sequences whose rows were not yet physically visible: the checkpoint advances past them and the rows are permanently missed.

### Change

- Publish the watermark once, strictly after both ordinary and bulk memtable writes complete, on all three paths: the single-region fast path, the multi-region spawned tasks, and WAL replay.
- Failed write contexts (WAL entry could not be built) never publish.
- A `#[cfg(test)]` watch-channel barrier makes the race deterministic (region-targeted, empty-bulk early return hoisted); a worker-level regression test pins the ordering.

### Verification

- `cargo nextest run -p mito2`: 1113 passed (base 1112 + 1 new race test)
- `make fmt` / `git diff --check` clean

## PR Checklist

### DIFF
--- a/src/mito2/src/region/opener.rs
+++ b/src/mito2/src/region/opener.rs
@@ -908,6 +908,9 @@ where
         region_write_ctx.set_next_entry_id(last_entry_id + 1);
         region_write_ctx.write_memtable().await;
         region_write_ctx.write_bulk().await;
+        // Publish the replayed sequences only after all rows (including bulk
+        // parts) are installed, matching the write path ordering.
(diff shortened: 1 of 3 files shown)


evenyag:
@codex review

evenyag (approved):

PR #8865, opened 2026-08-13
feat: preserve row sequences and support exact sequence-range reads

discord9:

- Depends on #8862; the Flow caller that enables `exact_sequence_range=true` is intentionally delivered in a later stacked PR.

This PR adds the OSS storage support needed for Flow sequence-range reads. It introduces the append-only `preserve_row_sequence` option, which keeps each row's physical `__sequence` through flush and compaction, and persists a per-SST capability marker so the storage layer can safely answer an exact `(C, H]` sequence-range request.

The exact-read path derives the time-selected SST set once and uses that same set for capability validation and scanning. It applies row-level sequence filtering only when the region and every selected file can prove that their sequence domains are compatible; legacy, unmarked, mixed, copied, foreign-region, or otherwise untrusted files fail closed with `SEQUENCE_RANGE_UNSUPPORTED` rather than returning an approximation. Existing non-exact scans retain their normal behavior, and incompatible pruning hints are rejected.

The change also:

- Supports `ALTER TABLE ... SET/UNSET 'preserve_row_sequence'`, including metadata persistence, `SHOW CREATE TABLE`, restart behavior, and rejection when `append_mode` is not enabled.
- Preserves the option through both normal and write-cache SST creation and through compaction only when all inputs are trusted. Compaction rewrites untrusted inputs without laundering their sequence capability; copied files likewise clear the marker.
- Adds the required request, region-option, file-metadata, scan, cache-key, and compatibility serialization coverage.
- Adds Mito tests for flat and primary-key formats, flush/compaction rewrites, legacy and mixed-file fail-closed behavior, copy-region handling, `LastRow`, bulk publication ordering, and exact-range filtering, plus the `alter_preserve_row_sequence` SQLness case.
- Includes final diff cleanup: a single exact-read SST selection derivation, hardened compaction rewrite coverage, consolidated serde coverage, removal of obsolete no-op bulk-compaction claims, ALTER comment cleanup, and the SQLness result trailing-newline fix.

The persisted option and file metadata remain backward compatible. This PR provides the storage/data-plane capability; it does not yet make Flow issue `exact_sequence_range=true`, which is handled by the later stacked caller change.

## PR Checklist

### DIFF
--- a/src/mito2/src/compaction/compactor.rs
+++ b/src/mito2/src/compaction/compactor.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::num::{NonZero, NonZeroU64};
+use std::num::NonZeroU64;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -455,17 +455,49 @@ impl SstMerger for DefaultSstMerger {
             .iter()
             .map(|f| f.file_id().to_string())
             .join(",");
-        let max_sequence = known_max_input_sequence(&output.inputs);
+        let input_max_sequence = known_max_input_sequence(&output.inputs);
         // The compaction output can only preserve per-row sequences when the
         // region option is enabled AND every input file carries the
-        // preserved-sequence marker. Otherwise a legacy (unmarked) input would
-        // be laundered into a trusted output and exact sequence-range scans
-        // would silently skip or wrongly filter its normalized sequences.
+        // preserved-sequence marker. Otherwise, rewrite the untrusted input as
+        // a sequence-less file. `committed_sequence + 1` is the region-local
+        // admission barrier for this compacted SST: once Flow's cursor reaches
+        // it, the input rows (which were all flushed in this snapshot) have
+        // already been consumed.
         let output_preserves_sequence = compaction_region.region_options.preserve_row_sequence
             && output
                 .inputs
                 .iter()
                 .all(|f| f.meta_ref().preserve_row_sequence);
+        let output_sequence = if output_preserves_sequence {
+            input_max_sequence
+        } else {
+            // The manifest records the region's latest accepted sequence for
+            // edit paths and its flushed frontier otherwise. Compaction only
+            // rewrites the immutable SST snapshot, so this is the same
+            // region-local admission barrier used when accepting files.
+            let manifest = compaction_region
+                .manifest_ctx
+                .manifest_manager
+                .read()
+                .await
+                .manifest();
+            NonZeroU64::new(
+                manifest
+                    .committed_sequence
+                    .unwrap_or(manifest.flushed_sequence)
+                    + 1,
+            )
+        };
+        // The flat/primary-key parquet formats require their three internal
+        // columns. For a sequence-less output, write the sequence column as
+        // zero instead of its file barrier. The reader recognizes all-zero
+        // sequence columns and overrides them from FileMeta for legacy reads;
+        // exact reads skip this file at file level before row filtering.
+        let write_max_sequence = if output_preserves_sequence {
+            input_max_sequence.map(NonZeroU64::get)
+        } else {
+            Some(0)
+        };
         let builder = CompactionSstReaderBuilder {
             metadata: compaction_region.region_metadata.clone(),
             sst_layer: compaction_region.access_layer.clone(),
@@ -489,7 +521,7 @@ impl SstMerger for DefaultSstMerger {
                     source,
                     cache_manager: compaction_region.cache_manager.clone(),
                     storage,
-                    max_sequence: max_sequence.map(NonZero::get),
+                    max_sequence: write_max_sequence,
                     sst_write_format: if flat_format {
                         FormatType::Flat
                     } else {
@@ -544,7 +576,7 @@ impl SstMerger for DefaultSstMerger {
                     index_version: 0,
                     num_rows: sst_info.num_rows as u64,
                     num_row_groups: sst_info.num_row_groups,
-                    sequence: max_sequence,
+                    sequence: output_sequence,
                     partition_expr: partition_expr.clone(),
                     num_series: sst_info.num_series,
                     primary_key_min,
--- a/src/mito2/src/compaction/compactor.rs
+++ b/src/mito2/src/compaction/compactor.rs
@@ -404,13 +404,10 @@ pub trait SstMerger: Send + Sync + 'static {
 #[derive(Clone)]
 pub struct DefaultSstMerger;
 
-/// Computes the maximum sequence bound of the output of merging `inputs`.
-///
-/// Returns `None` for empty inputs or if any input's sequence bound is unknown:
-/// a partial max over the known inputs would launder the unknown file's rows into a
-/// trusted sequence bound, letting exact sequence-range scans silently skip or
-/// wrongly filter rows from that file. Only when every input bound is known is
-/// the `max` computed.
+/// Computes the maximum target-domain sequence bound of the output of merging
+/// `inputs`. Foreign files are described by their target-local barrier; local
+/// trusted files use their physical sequence bound. Unknown bounds never become
+/// trusted through a partial maximum.
 fn known_max_input_sequence(inputs: &[FileHandle]) -> Option<NonZeroU64> {
     let mut max: Option<NonZeroU64> = None;
     for input in inputs {
@@ -456,18 +453,14 @@ impl SstMerger for DefaultSstMerger {
             .map(|f| f.file_id().to_string())
             .join(",");
         let input_max_sequence = known_max_input_sequence(&output.inputs);
-        // The compaction output can only preserve per-row sequences when the
-        // region option is enabled AND every input file carries the
-        // preserved-sequence marker. Otherwise, rewrite the untrusted input as
-        // a sequence-less file. `committed_sequence + 1` is the region-local
-        // admission barrier for this compacted SST: once Flow's cursor reaches
-        // it, the input rows (which were all flushed in this snapshot) have
-        // already been consumed.
+        // The output is trusted only when every input is trusted in the target
+        // domain. Foreign marker values describe the source and are ignored;
+        // their present FileMeta.sequence is the target-local barrier.
         let output_preserves_sequence = compaction_region.region_options.preserve_row_sequence
             && output
                 .inputs
                 .iter()
-                .all(|f| f.meta_ref().preserve_row_sequence);
+                .all(|f| f.is_effective_target_sequence_trusted(region_id));
         let output_sequence = if output_preserves_sequence {
             input_max_sequence
         } else {
@@ -488,16 +481,11 @@ impl SstMerger for DefaultSstMerger {
                     + 1,
             )
         };
-        // The flat/primary-key parquet formats require their three internal
-        // columns. For a sequence-less output, write the sequence column as
-        // zero instead of its file barrier. The reader recognizes all-zero
-        // sequence columns and overrides them from FileMeta for legacy reads;
-        // exact reads skip this file at file level before row filtering.
-        let write_max_sequence = if output_preserves_sequence {
-            input_max_sequence.map(NonZeroU64::get)
-        } else {
-            Some(0)
-        };
+        // For untrusted output, keep the physical sequence override compatible
+        // with the main write path: use the known maximum sequence of the
+        // inputs. The FileMeta sequence remains the admission barrier, but it
+        // must not be written into the rows (or replaced with zero).
+        let write_max_sequence = input_max_sequence.map(NonZeroU64::get);
         let builder = CompactionSstReaderBuilder {
             metadata: compaction_region.region_metadata.clone(),
             sst_layer: compaction_region.access_layer.clone(),
@@ -844,6 +832,36 @@ mod tests {
 
     fn new_file_handle(meta: FileMeta) -> FileHandle {
         FileHandle::new(meta, Arc::new(NoopFilePurger))
+    }
+
+    #[test]
+    fn test_effective_target_trust_and_max_sequence() {
+        let target = RegionId::new(1, 1);
+        let local_trusted = new_file_handle(FileMeta {
+            region_id: target,
+            sequence: NonZeroU64::new(3),
+            preserve_row_sequence: true,
+            ..dummy_file_meta()
+        });
+        let foreign_untrusted_marker = new_file_handle(FileMeta {
+            region_id: RegionId::new(1, 2),
+            sequence: NonZeroU64::new(9),
+            preserve_row_sequence: false,
+            ..dummy_file_meta()
+        });
+        let local_untrusted = new_file_handle(FileMeta {
+            region_id: target,
+            sequence: NonZeroU64::new(11),
+            preserve_row_sequence: false,
+            ..dummy_file_meta()
+        });
+        assert!(local_trusted.is_effective_target_sequence_trusted(target));
+        assert!(foreign_untrusted_marker.is_effective_target_sequence_trusted(target));
+        assert!(!local_untrusted.is_effective_target_sequence_trusted(target));
+        assert_eq!(
+            NonZeroU64::new(9),
+            known_max_input_sequence(&[local_trusted, foreign_untrusted_marker])
+        );
     }
 
     #[test]
--- a/src/mito2/src/compaction/compactor.rs
+++ b/src/mito2/src/compaction/compactor.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::num::NonZero;
+use std::num::{NonZero, NonZeroU64};
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -52,7 +52,7 @@ use crate::region::version::VersionRef;
 use crate::region::{ManifestContext, RegionLeaderState, RegionRoleState};
 use crate::schedule::scheduler::LocalScheduler;
 use crate::sst::FormatType;
-use crate::sst::file::{FileMeta, UncommittedSsts};
+use crate::sst::file::{FileHandle, FileMeta, UncommittedSsts};
 use crate::sst::file_purger::LocalFilePurger;
 use crate::sst::index::intermediate::IntermediateManager;
 use crate::sst::index::puffin_manager::PuffinManagerFactory;
@@ -404,6 +404,22 @@ pub trait SstMerger: Send + Sync + 'static {
 #[derive(Clone)]
 pub struct DefaultSstMerger;
 
+/// Computes the maximum sequence bound of the output of merging `inputs`.
+///
+/// Returns `None` if any input's sequence bound is unknown (`None`): a partial
+/// max over the known inputs would launder the unknown file's rows into a
+/// trusted sequence bound, letting exact sequence-range scans silently skip or
+/// wrongly filter rows from that file. Only when every input bound is known is
+/// the `max` computed.
+fn max_input_sequence(inputs: &[FileHandle]) -> Option<NonZeroU64> {
+    let mut max: Option<NonZeroU64> = None;
+    for input in inputs {
+        let sequence = input.meta_ref().sequence?;
+        max = Some(max.map_or(sequence, |current| current.max(sequence)));
+    }
+    max
+}
+
 #[async_trait::async_trait]
 impl SstMerger for DefaultSstMerger {
     async fn merge_single_output(
@@ -439,12 +455,17 @@ impl SstMerger for DefaultSstMerger {
             .iter()
             .map(|f| f.file_id().to_string())
             .join(",");
-        let max_sequence = output
-            .inputs
-            .iter()
-            .map(|f| f.meta_ref().sequence)
-            .max()
-            .flatten();
+        let max_sequence = max_input_sequence(&output.inputs);
+        // The compaction output can only preserve per-row sequences when the
+        // region option is enabled AND every input file carries the
+        // preserved-sequence marker. Otherwise a legacy (unmarked) input would
+        // be laundered into a trusted output and exact sequence-range scans
+        // would silently skip or wrongly filter its normalized sequences.
+        let output_preserves_sequence = compaction_region.region_options.preserve_row_sequence
+            && output
+                .inputs
+                .iter()
+                .all(|f| f.meta_ref().preserve_row_sequence);
         let builder = CompactionSstReaderBuilder {
             metadata: compaction_region.region_metadata.clone(),
             sst_layer: compaction_region.access_layer.clone(),
@@ -474,6 +495,7 @@ impl SstMerger for DefaultSstMerger {
                     } else {
                         FormatType::PrimaryKey
                     },
+                    preserve_row_sequence: output_preserves_sequence,
                     index_options,
                     index_config,
                     inverted_index_config,
@@ -527,6 +549,7 @@ impl SstMerger for DefaultSstMerger {
                     num_series: sst_info.num_series,
                     primary_key_min,
                     primary_key_max,
+                    preserve_row_sequence: output_preserves_sequence,
                 }
             })
             .collect::<Vec<_>>();
@@ -789,6 +812,33 @@ mod tests {
 
     fn new_file_handle(meta: FileMeta) -> FileHandle {
         FileHandle::new(meta, Arc::new(NoopFilePurger))
+    }
+
+    #[test]
+    fn test_max_input_sequence() {
+        let meta_with = |sequence| FileMeta {
+            sequence,
+            ..dummy_file_meta()
+        };
+
+        // All inputs with known bounds yield their max.
+        let inputs = vec![
+            new_file_handle(meta_with(NonZeroU64::new(3))),
+            new_file_handle(meta_with(NonZeroU64::new(9))),
+            new_file_handle(meta_with(NonZeroU64::new(5))),
+        ];
+        assert_eq!(NonZeroU64::new(9), max_input_sequence(&inputs));
+
+        // A single unknown bound poisons the whole output: the known inputs'
+        // max must not be laundered onto the merged output.
+        let inputs = vec![
+            new_file_handle(meta_with(NonZeroU64::new(9))),
+            new_file_handle(meta_with(None)),
+        ];
+        assert_eq!(None, max_input_sequence(&inputs));
+
+        // No inputs at all yield `None`.
+        assert_eq!(None, max_input_sequence(&[]));
     }
 
     /// Build a minimal [`CompactionRegion`] suitable for tests where the
--- a/src/mito2/src/compaction/compactor.rs
+++ b/src/mito2/src/compaction/compactor.rs
@@ -860,14 +860,12 @@ mod tests {
         ];
         assert_eq!(NonZeroU64::new(9), known_max_input_sequence(&inputs));
 
-        // Unknown input fails closed.
         let inputs = vec![
             new_file_handle(meta_with(NonZeroU64::new(9))),
             new_file_handle(meta_with(None)),
         ];
         assert_eq!(None, known_max_input_sequence(&inputs));
 
-        // Empty input has no bound.
         assert_eq!(None, known_max_input_sequence(&[]));
     }
 
--- a/src/mito2/src/compaction/compactor.rs
+++ b/src/mito2/src/compaction/compactor.rs
@@ -406,12 +406,12 @@ pub struct DefaultSstMerger;
 
 /// Computes the maximum sequence bound of the output of merging `inputs`.
 ///
-/// Returns `None` if any input's sequence bound is unknown (`None`): a partial
-/// max over the known inputs would launder the unknown file's rows into a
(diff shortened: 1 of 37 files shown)


fengjiachun (commented):
Found two exactness gaps.

src/mito2/src/worker/handle_copy_region.rs
```diff
@@ -173,6 +173,16 @@ impl<S> RegionWorkerLoop<S> {
...
+                // The target region has an independent sequence domain: the
+                // physical per-row sequences in the copied file belong to the
+                // source region, so they must not be trusted for exact
+                // sequence-range reads on the target. Clear the
+                // `preserve_row_sequence` marker to fail closed until the scan
+                // provably cannot intersect the copied rows (see
+                // `files_allow_exact_sequence_range`); otherwise an exact
+                // request would replay source-domain rows as if they were
+                // target sequences.
+                new_file_meta.preserve_row_sequence = false;
```
fengjiachun: Clearing only the marker still leaves the cloned source-domain `FileMeta::sequence`. `files_allow_exact_sequence_range()` later treats an unmarked file with `sequence <= C` as proven disjoint, but that comparison is invalid in the target's independent sequence domain. For example, if the target checkpoint is already above the source file's max sequence, an exact scan after the copy silently skips the copied file instead of failing closed. Please also clear/translate the file sequence into the target domain (and publish the corresponding target sequence if the copy is meant to be an atomic target update), or otherwise keep this file unconditionally unsupported.

src/mito2/src/read/scan_region.rs
```diff
@@ -1608,6 +1664,51 @@ fn pre_filter_mode(append_mode: bool, merge_mode: MergeMode) -> PreFilterMode {
...
+/// - every unmarked SST carries a known max sequence not exceeding the lower
+///   bound `min`, so it cannot contribute rows to `(min, max]`.
+///
+/// Returns `None` otherwise. Callers must keep main's fences/fallback semantics
+/// and never silently approximate.
+pub(crate) fn exact_sequence_range(
+    request: &ScanRequest,
+    version: &crate::region::version::Version,
+) -> Option<SequenceRange> {
+    if !request.exact_sequence_range
```
fengjiachun: This capability check only covers `SstVersion`, but an enterprise follower can attach extension ranges later in `scan_input()`. Normal SST streams get the new row-level sequence filter, while `scan_flat_extension_range()` returns the extension reader's stream directly with no trust check or generic sequence filter. Thus an exact request can succeed and still emit out-of-range extension rows. Please either reject exact mode when an extension provider can contribute ranges, or add a preservation capability and apply the same `(C, H]` filtering to trusted extension ranges.

src/mito2/src/read/scan_region.rs
```diff
@@ -437,6 +440,23 @@ impl ScanRegion {
...
+        // with max = 8, losing sequences 6..8), so the conflicting combination
+        // is rejected explicitly instead of being silently approximated.
+        if sst_min_sequence.is_some() && self.exact_sequence_range().is_some() {
+            return SequenceRangeUnsupportedSnafu {
+                region_id: self.region_id(),
+                min_seq: self.request.memtable_min_sequence.unwrap_or_default(),
+                max_seq: self.request.memtable_max_sequence.unwrap_or_default(),
+                reason:
+                    "sst_min_sequence pruning hint is incompatible with exact sequence-range reads"
+                        .to_string(),
```
killme2008:
```suggestion
                reason:
                    "sst_min_sequence pruning hint is incompatible with exact sequence-range reads",
```

src/mito2/src/read/scan_region.rs
```diff
@@ -458,10 +478,27 @@ impl ScanRegion {
...
+        };
         let mut files = Vec::new();
         if !self.request.skip_sst_files {
             for level in ssts.levels() {
                 for file in level.files.values() {
+                    if let Some(min) = exact_min_sequence
+                        && let Some(file_sequence) = file.meta_ref().sequence
+                        && file_sequence.get() <= min
+                    {
+                        continue;
```
killme2008: Better to add a debug log

src/mito2/src/read/scan_region.rs
```diff
@@ -437,6 +440,23 @@ impl ScanRegion {
     async fn scan_input(self) -> Result<ScanInput> {
         let metadata = &self.version.metadata;
         let sst_min_sequence = self.request.sst_min_sequence.and_then(NonZeroU64::new);
+        // Exact `sequence_range` mode filters rows row-level on every
+        // time-matching SST. The legacy `sst_min_sequence` file-pruning hint
+        // would silently skip preserved files whose max sequence does not
+        // exceed it (e.g. `(5, 10]` with `sst_min_sequence = 8` drops a file
+        // with max = 8, losing sequences 6..8), so the conflicting combination
+        // is rejected explicitly instead of being silently approximated.
+        if sst_min_sequence.is_some() && self.exact_sequence_range().is_some() {
```
killme2008: Use `ensure!` instead.

src/mito2/src/read/scan_region.rs
```diff
@@ -589,6 +627,22 @@ impl ScanRegion {
         let input = if !self.request.skip_sst_files
             && let Some(provider) = self.extension_range_provider
         {
+            if sequence_range.is_some() {
```
killme2008, line 598: Use `ensure!` instead.

The comment seems unnecessary, since the error message already documents the behavior.

src/mito2/src/read/scan_region.rs
```diff
@@ -1608,6 +1680,51 @@ fn pre_filter_mode(append_mode: bool, merge_mode: MergeMode) -> PreFilterMode {
...
+/// and never silently approximate.
+pub(crate) fn exact_sequence_range(
+    request: &ScanRequest,
+    version: &crate::region::version::Version,
+) -> Option<SequenceRange> {
+    if !request.exact_sequence_range
+        || request.skip_sst_files
+        || !version.options.preserve_row_sequence
+    {
+        return None;
```
killme2008: Add debug logs for these ealry returns.

src/mito2/src/worker/handle_alter.rs
```diff
@@ -300,6 +304,19 @@ impl<S: LogStore> RegionWorkerLoop<S> {
                         all_options_altered = false;
                     }
                 }
+                SetRegionOption::PreserveRowSequence(new_preserve) => {
+                    // Toggling preserve_row_sequence only affects in-memory
+                    // RegionOptions and how future flushes/compactions mark SST
+                    // metadata; it does not require an empty memtable, so it
+                    // never blocks the fast path.
```
killme2008: Remove these unnecessary comments.

src/mito2/src/worker/handle_alter.rs
```diff
@@ -308,15 +325,33 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+            // options directly without flushing.
+            region.version_control.alter_options(candidate);
             Ok(None)
         } else {
-            let kind = AlterKind::SetRegionOptions { options };
-            Ok(new_region_options_on_empty_memtable(
-                &current_options,
-                &kind,
-            ))
+            // Some option independently requires an empty memtable (e.g.
```
killme2008: The same suggestion  as above.

src/mito2/src/engine.rs
```diff
@@ -1059,36 +1059,125 @@ impl EngineInner {
...
 
-        if let Some(given_seq) = request.memtable_min_sequence {
-            let min_readable_seq = version.flushed_sequence;
-            ensure!(
-                given_seq >= min_readable_seq,
-                IncrementalQueryStaleSnafu {
-                    region_id,
-                    given_seq,
-                    min_readable_seq,
+        // Shared capability check used identically by the engine (fence
```
killme2008: Useless comments

src/mito2/src/engine.rs
```diff
@@ -1059,36 +1059,125 @@ impl EngineInner {
...
+        // relaxation) and the reader (row-level filtering) — see
+        // `read::scan_region::exact_sequence_range`. Only Flow's explicit
+        // `sequence_range` mode requests exactness; historical `memtable_only`
+        // scans never set `exact_sequence_range` and therefore keep every fence
+        // below regardless of region options.
+        let exact_sequence_range = exact_sequence_range(&request, &version);
```
killme2008: The value will be overridden when the enterprise feature is enabled.

It would be better to add an exact_sequence_range function that handles the enterprise-enabled and non-enterprise cases separately.

killme2008: The code would be more clear

src/mito2/src/engine.rs
```diff
@@ -1059,36 +1059,125 @@ impl EngineInner {
...
+                // return rows outside (C, H], so fail instead.
+                return SequenceRangeUnsupportedSnafu {
                     region_id,
-                    given_seq,
-                    min_enforceable_seq,
+                    min_seq: request.memtable_min_sequence.unwrap_or_default(),
+                    max_seq: request.memtable_max_sequence.unwrap_or_default(),
+                    reason: {
+                        #[cfg(feature = "enterprise")]
+                        let extension_reason = extension_provider_blocks_exact.then(|| {
```
killme2008: Extract a function `get_extension_reason` that handles the enterprise-enabled and non-enterprise cases separately.

src/mito2/src/engine.rs
```diff
@@ -1059,36 +1059,125 @@ impl EngineInner {
...
+                let min_enforceable_seq = version.flushed_sequence;
+                ensure!(
+                    given_seq >= min_enforceable_seq,
+                    SnapshotFenceStaleSnafu {
+                        region_id,
+                        given_seq,
+                        min_enforceable_seq,
+                    }
+                );
+            }
```
killme2008: I recommend introducing a `validate_region_scan_request` function to encapsulate the validation logic above.

`scan_region` is getting a bit long and hard to read.

src/mito2/src/engine/scan_test.rs
```diff
@@ -1259,3 +1272,1288 @@ async fn test_range_cache_separates_or_equality_time_filters() {
...
+    );
+    assert!(scan.contains("| 3 "), "second bulk part missing:\n{scan}");
+
+    test_util::flush_region(&engine, region_id, None).await;
+    assert_eq!(scan_exact(Some(2), Some(100)).await, scan);
+
+    engine
+        .handle_request(
+            region_id,
+            RegionRequest::Compact(RegionCompactRequest::default()),
```
evenyag: This compaction step appears to be a no-op: the test creates two bulk parts but flushes them into a single SST, so Compact has no second input to rewrite. The post-flush assertion already covers the persisted bulk-part behavior, while real compaction is covered by the dedicated multi-input tests below. Could we remove this block and update the test comment to say before and after flush?

src/mito2/src/engine/scan_test.rs
```diff
@@ -1259,3 +1272,1288 @@ async fn test_range_cache_separates_or_equality_time_filters() {
...
+        1,
+        outputs.len(),
+        "two inputs should rewrite into one output file"
+    );
+    assert!(
+        !outputs[0].meta_ref().preserve_row_sequence,
+        "legacy input must not be laundered into a marked output"
+    );
+
+    // The compacted output is unmarked. The region-edit handler filled its max
```
evenyag: The unique assertion in this test is that compaction output remains unmarked (above). The empty disjoint scan and overlapping unsupported scan repeat the eligibility behavior already covered by test_exact_sequence_range_unsupported_when_legacy_file plus test_files_allow_exact_sequence_range. Could we end this test after the marker assertion?

src/mito2/src/region/options.rs
```diff
@@ -1069,10 +1119,29 @@ mod tests {
             max_row_group_row_count: None,
             primary_key_encoding: None,
             write_buffer_size: None,
+            preserve_row_sequence: false,
         };
         assert_eq!(options, got);
     }
 
+    #[test]
+    fn test_region_options_serde_preserve_row_sequence_roundtrip() {
```
evenyag: Could this coverage be folded into the existing test_region_options_serde test by enabling append_mode and preserve_row_sequence in that fixture, then asserting the old JSON defaults the new field to false? That keeps round-trip and backward-compatibility coverage without another test function.

src/mito2/src/engine/scan_test.rs
```diff
@@ -1259,3 +1272,1288 @@ async fn test_range_cache_separates_or_equality_time_filters() {
...
+        !pretty.contains("| 0 ") && !pretty.contains("| 5 "),
+        "rows outside (1, 5] leaked:\n{pretty}"
+    );
+}
+
+/// Exact sequence-range reads through the active production `PerSeries` series
+/// scan path: row-level filtering must hold across the flushed SST and the
+/// memtable regardless of the requested time-series distribution.
+#[tokio::test]
+async fn test_exact_sequence_read_series_scan_per_series() {
```
evenyag, line 3021: This repeats the same setup and expected rows as test_exact_sequence_read_partial_sst_partial_memtable: sequences 1..3 are flushed, 4..6 stay in memory, and (2, 6] returns rows 2..5. Could the earlier test loop over distribution = None and PerSeries, then remove this duplicate setup?

src/mito2/src/sst/file.rs
```diff
@@ -282,6 +282,14 @@ pub struct FileMeta {
         deserialize_with = "deserialize_bytes_option"
     )]
     pub primary_key_max: Option<Bytes>,
+    /// Whether the file preserves per-row sequence numbers usable for exact
+    /// row-level sequence filtering.
+    #[serde(default, skip_serializing_if = "is_false")]
+    pub preserve_row_sequence: bool,
```
evenyag, line 288: Should we store the optional min sequence in the FileMeta? If it is Some, the file preserves row sequence.

Then we have both min/max stats.

src/mito2/src/read/scan_region.rs
evenyag: exact_sequence_range() is costly as it iterates over all the files. We should avoid calling them multiple times.

src/mito2/src/read/scan_region.rs
```diff
@@ -458,10 +478,27 @@ impl ScanRegion {
...
         let mut files = Vec::new();
         if !self.request.skip_sst_files {
             for level in ssts.levels() {
                 for file in level.files.values() {
+                    if let Some(min) = exact_min_sequence
+                        && let Some(file_sequence) = file.meta_ref().sequence
+                        && file_sequence.get() <= min
+                    {
+                        continue;
+                    }
```
evenyag: Can we update the exact sequence range during the iteration? So we can avoid iterating files twice.

src/mito2/src/worker/handle_alter.rs
```diff
@@ -308,15 +325,31 @@ impl<S: LogStore> RegionWorkerLoop<S> {
...
+            store_api::metadata::InvalidRegionRequestSnafu {
+                region_id: region.region_id,
+                err: e.to_string(),
+            }
+            .build()
+        })?;
         if all_options_altered {
-            region.version_control.alter_options(current_options);
+            // preserve_row_sequence toggles apply in memory and do not require
+            // an empty memtable.
```
evenyag: Maybe we can remove this comment.

src/mito2/src/engine.rs
```diff
@@ -1059,37 +1059,39 @@ impl EngineInner {
...
-                }
-            );
-        }
+        // Shared capability check used identically by the engine (fence
+        // relaxation) and the reader (row-level filtering) — see
+        // `read::scan_region::exact_sequence_range`. Only Flow's explicit
+        // `sequence_range` mode requests exactness; historical `memtable_only`
+        // scans never set `exact_sequence_range` and therefore keep every fence
+        // below regardless of region options.
+        let exact_sequence_range = exact_sequence_range(&request, &version);
```
evenyag: Now we have to compute exact_sequence_range() multiple times.

src/mito2/src/worker/handle_copy_region.rs
discord9: Fixed in `remap_copied_file_meta`: copied files now clear both `preserve_row_sequence` and the source-domain `FileMeta::sequence`, so the target cannot use that stale value for exact-range disjointness. The copy-region regression now verifies that the sequence hint is absent and that the target fails closed with `SequenceRangeUnsupported`.

src/mito2/src/read/scan_region.rs
discord9: Fixed by treating exact sequence-range capability as unavailable whenever a follower has an extension range provider, so the engine rejects the request instead of admitting unfiltered extension rows. `ScanRegion::scan_input` also has a reader-side fail-closed check before extension ranges are attached.

src/mito2/src/sst/file.rs
discord9: Thanks. I don’t think an optional minimum is required for correctness here: files with max/barrier `<= C` are already skipped, every remaining marked SST is filtered row-by-row against `(C, H]`, and an overlapping unmarked file fails closed. Without a min we may read a file whose rows are all above `H`, but cannot return out-of-range rows. A trustworthy min would need propagation through normal and split SST writes, bulk parts, compaction, copy/edit trust clearing, remote-compaction serialization, and manifest compatibility. I suggest keeping `preserve_row_sequence` as the explicit capability marker here and adding a benchmarked optional min pruning optimization separately; existing marked files will not have that new statistic, so `None` must not be interpreted as unpreserved.

src/mito2/src/read/scan_region.rs
discord9: Thanks. I kept the explicit `.to_string()` because `SequenceRangeUnsupported.reason` is an owned `String`; leaving the conversion at the construction site makes that ownership explicit.

src/mito2/src/read/scan_region.rs
discord9: Added a debug summary for exact sequence-range admission showing whether exact mode was requested, whether it is available, and the selected SST count.

src/mito2/src/read/scan_region.rs
discord9: I kept the explicit `.fail()` here because this is the dedicated unsupported branch after exact capability has been computed; changing it to `ensure!` would only invert the condition without simplifying the control flow.

src/mito2/src/read/scan_region.rs
discord9: I kept this explicit fail-closed branch and its comment. It is reader-side defense in depth after engine validation, and the comment documents the non-obvious risk that unfiltered extension streams could bypass the row-level sequence filter.

src/mito2/src/read/scan_region.rs
discord9: Rather than logging every normal fail-closed return in the capability helper, I added one debug summary at scan admission with the requested/available state and selected-file count. This keeps the diagnostic state without repetitive logs for each capability condition.

src/mito2/src/worker/handle_alter.rs
discord9: Removed the explanatory comments from the `PreserveRowSequence` match arm; it now only logs and applies the changed value.

src/mito2/src/worker/handle_alter.rs
discord9: I removed the branch-level comments that only restated the code, but retained the staged-validation comment because it records the order-independent combined-ALTER invariant and why validation is required before the fast path.

src/mito2/src/engine.rs
discord9: Removed the broad comment that restated the capability call. The remaining comments are limited to why exact requests preselect files and why enterprise extension ranges must disable exactness.

src/mito2/src/engine.rs
discord9: I did not split `exact_sequence_range` into enterprise and non-enterprise implementations because the SST capability calculation is identical in both builds. The enterprise-only follower/provider condition is applied separately, avoiding duplicated core logic.

src/mito2/src/engine.rs
discord9: Extracted the unsupported-reason selection into `sequence_range_unsupported_reason`, which handles the extension-provider, unpreserved-file, and option-disabled cases in one place.

src/mito2/src/engine.rs
discord9: Extracted the sequence-fence request validation into `validate_sequence_fences`; scanner construction now computes capability, logs it, invokes the helper, and continues with scan setup.

src/mito2/src/engine/scan_test.rs
discord9: Removed the no-op bulk compaction step and narrowed the test to verify the same folded bulk-part result before and after flush. Real multi-input compaction remains covered by the dedicated compaction tests.

src/mito2/src/engine/scan_test.rs
discord9: I retained the post-marker assertions because they cover compaction-specific behavior not established by generic eligibility tests: the admission barrier, physically zeroed sequence column, and fail-closed behavior below and at the barrier.

src/mito2/src/region/options.rs
discord9: Folded the coverage into `test_region_options_serde`: the existing fixture now round-trips `preserve_row_sequence = true`, and old JSON without the field is asserted to default it to `false`. The separate test function was removed.

src/mito2/src/engine/scan_test.rs
discord9: I kept this as a separate test because it intentionally enables flat format and exercises the production `PerSeries` series-scan path, while the earlier test covers the default distribution. Separate setup makes a regression select the failing scan implementation directly.

src/mito2/src/read/scan_region.rs
discord9: The engine now computes the selected exact-read SST set once and passes that vector into `ScanRegion`; the reader reuses it instead of walking the SST version again.

src/mito2/src/read/scan_region.rs
discord9: Capability validation and scan construction now share the same selected-file vector. This removes the second full SST iteration while ensuring both stages reason about the identical read set.

src/mito2/src/worker/handle_alter.rs
discord9: Removed that comment. The fast-path branch is clear from the condition and now applies the validated candidate directly.

src/mito2/src/engine.rs
discord9: I did not remove the engine/reader double check entirely. The engine needs capability before relaxing sequence fences, while the reader revalidates the same preselected set before attaching the row-level filter as a fail-closed boundary; file discovery is shared, so neither layer rebuilds the SST set.

src/mito2/src/compaction/compactor.rs
```diff
@@ -439,12 +470,50 @@ impl SstMerger for DefaultSstMerger {
...
+        };
+        // The flat/primary-key parquet formats require their three internal
+        // columns. For a sequence-less output, write the sequence column as
+        // zero instead of its file barrier. The reader recognizes all-zero
+        // sequence columns and overrides them from FileMeta for legacy reads;
+        // exact reads skip this file at file level before row filtering.
+        let write_max_sequence = if output_preserves_sequence {
+            input_max_sequence.map(NonZeroU64::get)
+        } else {
+            Some(0)
```
killme2008: For non-preserve outputs — which is every table today — writing the sequence column as all zeros does not leave the file sequence-less at read time. It gives its rows a *higher* sequence than they had on main.

`need_override_sequence()` (`sst/parquet/format.rs:830`) returns true for an all-zero column, so `reader.rs:474` replaces the whole column with `FileMeta.sequence`, which this PR now sets to `committed_sequence.unwrap_or(flushed_sequence) + 1`. On main the same rows read back as `max(input FileMeta.sequence)`.

That value feeds dedup: `flat_merge.rs:562` orders by `(pk, ts, sequence desc)` and dedup keeps the first row per key. Compaction may leave overlapping files behind — `twcs.rs:217` `overlaps_files_left_behind()` exists for exactly that case. So:

- file A (older flush, sequence 10) and file B (newer flush, sequence 50) both hold `(k, ts)`;
- compaction picks A only, and its output rows read back as `flushed_sequence + 1` >= 51;
- a `last_row` / `last_non_null` query returns A's stale value instead of B's.

Memtable rows whose sequence is exactly `flushed_sequence + 1` tie with the output, and the tie-break order is not defined either.

The exact-read path never reads an unmarked file: it is pruned by `barrier <= C` or the scan fails closed. So the physical column does not need rewriting here. Keeping main's `input_max_sequence.map(NonZeroU64::get)` and putting the barrier only in `FileMeta.sequence` preserves the merge order and still gives you the admission barrier. If the zero-write is deliberate, could you explain why dedup is unaffected and add a case that compacts a subset while an overlapping newer file stays behind? No test currently pins the merge order of a non-preserve table across compaction.

One consequence in the same direction: the inflated `FileMeta.sequence` makes compaction outputs practically un-prunable by `sst_min_sequence` (the barrier is always newer than the caller's checkpoint), so existing non-exact Flow incremental reads will re-scan them after every compaction. Not a correctness issue, but a behavior change worth confirming.

src/mito2/src/read/scan_region.rs
```diff
@@ -457,33 +469,25 @@ impl ScanRegion {
...
-                    // and the batch will be filtered out by tree reader anyway.
-                }
+        let files = if self.request.skip_sst_files {
+            Vec::new()
+        } else if let Some(files) = self.selected_files.take() {
+            files
+        } else {
+            exact_sequence_range_files(&self.request, &self.version)
+        };
+        let sequence_range = self.exact_sequence_range(&files)?;
```
killme2008: Following up on @evenyag's point about calling this twice: the engine already computed the same value at `engine.rs:1068`, over the same `version` snapshot and the same file list it then hands over via `with_selected_files()`. Recomputing it here re-walks every selected file for a result that cannot differ — and if it ever did differ, that divergence would be a bug the `Result` here silently re-derives instead of surfacing.

Passing the range down alongside the files would drop the second walk and make the "single source of truth" property structural rather than a convention two call sites have to keep:

```rust
scan_region.with_exact_sequence_range(files, sequence_range)
```

with `self.selected_files.take()` yielding both, and the local computation kept only for the path where the engine did not pre-compute one (`exact_sequence_range` false, where it returns `None` before touching any file).

There is a second walk in the same path worth folding in: `exact_sequence_range_files()` iterates the SSTs to select them, then `files_allow_exact_sequence_range()` iterates the selection again. Both decide on `meta_ref()` of the same files, so the barrier check can be made during selection and the two functions collapsed into one pass returning `(files, allows_exact)`.

killme2008 (approved):
Leave two comments, please take a look @discord9 

After addressing them, I think it's ready to go cc @evenyag

discord9:
We are going to split the existing repartition/copy-region sequence-domain issue into a separate prerequisite PR.

The issue already exists on `main`: an imported SST may retain physical row sequences from its source region, but those sequence numbers have no ordering meaning in the target region. The target region should instead assign the imported files a target-local virtual sequence in its own manifest (`FileMeta.sequence`) without rewriting the potentially shared SST bytes. Reads and compaction must then use that target-local manifest sequence for ordering.

The plan is:

1. Submit an independent PR that fixes the repartition/copy-region sequence-domain handling on `main`, including pre- and post-compaction regression coverage.
2. Rebase this PR (#8865) on top of that fix and make it a stacked PR.
3. Keep #8865 focused on preserving trusted per-row sequences. Imported/shared files will continue to fall back to the target-local import barrier rather than carrying source-region sequence semantics into the target region.
4. Once the prerequisite PR is merged, rebase #8865 onto `main` again.

The two compaction follow-up commits previously added to #8865 have already been removed from this branch so the concerns can be reviewed separately and in the correct dependency order.

src/mito2/src/engine.rs
```diff
@@ -1108,6 +1121,11 @@ impl EngineInner {
         .with_ignore_fulltext_index(self.config.fulltext_index.apply_on_query.disabled())
         .with_ignore_bloom_filter(self.config.bloom_filter_index.apply_on_query.disabled())
         .with_start_time(query_start);
+        let scan_region = if let Some(files) = selected_files {
+            scan_region.with_selected_files(files)
+        } else {
+            scan_region
+        };
```
evenyag: I think we can set exact_sequence_range and selected files as a tuple at the same time.

src/mito2/src/engine.rs
```diff
@@ -1059,37 +1061,48 @@ impl EngineInner {
...
-        }
+        // Only exact requests need the capability check and its selected file
+        // list. Other requests keep the regular scan pruning path unchanged.
+        let selected_files = (request.exact_sequence_range && !request.skip_sst_files)
+            .then(|| exact_sequence_range_files(&request, &version));
+        let exact_sequence_range = exact_sequence_range(
+            &request,
+            &version,
+            selected_files.as_deref().unwrap_or_default(),
+        )?;
```
evenyag: Maybe we can add debug logs for all related sequences and context for exact_sequence_range() to help find sequence issues.

evenyag (approved):

PR #8870, opened 2026-08-13
feat: add json_object function and use it in the entity-graph derivation

killme2008:

#8609

The entity-graph derivation built `entity_id_attrs` and `descriptive` by concatenating a JSON text and re-parsing it, escaping only `\` and `"` in runtime values. A label containing a control character (e.g. a newline) made the text unparseable and failed the whole `semantic_entities` scan instead of skipping one attribute.

This PR adds a general `json_object` scalar function (MySQL `JSON_OBJECT` semantics: interleaved key/value arguments, `json_object()` returns `{}`, keys are non-NULL strings, values are strings/integers/floats/booleans/NULL) that assembles the JSONB binary directly from the value columns — no JSON text round-trip, so hostile characters need no escaping. The derivation now calls it instead of the text pipeline; the function is registered globally, which is also required for the plan to decode on datanodes.

`NULL -> ""` for absent descriptive values remains derivation-side policy, expressed with `coalesce` at the call site; the function itself renders NULL values as JSON null.

Limitation: value types outside strings/integers/floats/booleans/NULL are rejected with a hint to cast. Decimals are rejected rather than cast to `Float64` — JSONB numbers (i64/u64/f64) cannot represent them exactly, and silent precision loss is worse than an explicit cast. Nested JSON values can be added later.

## PR Checklist

### DIFF
--- a/src/common/function/src/scalars/json/json_object.rs
+++ b/src/common/function/src/scalars/json/json_object.rs
@@ -1,0 +1,275 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fmt::{self, Display};
+use std::sync::Arc;
+
+use datafusion_common::DataFusionError;
+use datafusion_common::arrow::array::{Array, ArrayRef, AsArray, BinaryViewBuilder};
+use datafusion_common::arrow::compute;
+use datafusion_common::arrow::datatypes::{DataType, Float64Type, Int64Type, UInt64Type};
+use datafusion_expr::{ColumnarValue, ScalarFunctionArgs, Signature, Volatility};
+
+use crate::function::Function;
+
+const NAME: &str = "json_object";
+
+/// Builds a `JSONB` object from interleaved `(key, value, key, value, ...)`
+/// arguments, like MySQL's `JSON_OBJECT`. Values are written into the binary
+/// directly, so they need no JSON text escaping. Keys must be non-NULL strings;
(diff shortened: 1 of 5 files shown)


src/common/function/src/scalars/json/json_object.rs
```diff
@@ -0,0 +1,267 @@
...
+            .map(|i| jsonb::from_slice(array.value(i)).unwrap().to_string())
+            .collect())
+    }
+
+    fn key(name: &str) -> ColumnarValue {
+        ColumnarValue::Scalar(datafusion_common::ScalarValue::Utf8(Some(name.to_string())))
+    }
+
+    #[test]
+    fn builds_objects_from_mixed_types_without_escaping() {
```
WenyXu: nit: The neighboring JSON scalar modules consistently prefix test names with `test_` (for example, `test_json_object_keys_function`). Could we follow the same convention for the three tests in this module to keep test naming consistent?

src/common/function/src/scalars/json/json_object.rs
```diff
@@ -0,0 +1,267 @@
...
+        if arrays.len() % 2 != 0 {
+            return Err(DataFusionError::Execution(format!(
+                "{NAME} expects (key, value) argument pairs, got {} arguments",
+                arrays.len()
+            )));
+        }
+        let pairs = arrays
+            .chunks(2)
+            .map(|pair| {
+                let keys = compute::cast(&pair[0], &DataType::Utf8View).map_err(|_| {
```
WenyXu, line 139: `compute::cast` accepts primitive keys such as integers and converts them to strings, while the function documentation says keys must be strings. Could we clarify the intended contract here? If convertible keys are intentional, please update the documentation/error wording and add a numeric-key test to make that behavior explicit.

src/common/function/src/scalars/json/json_object.rs
killme2008: Renamed with the `test_` prefix in 195992ed1c.

src/common/function/src/scalars/json/json_object.rs
killme2008: Convertible keys are intentional — MySQL's `JSON_OBJECT` converts keys to strings, and this function follows its semantics. Documented in the rustdoc (`a numeric key like 1 becomes "1"`), reworded the cast-failure message to `cannot convert keys of type ... to string`, and added `test_numeric_key_converts_to_string`. 195992ed1c.

WenyXu (approved):

PR #8872, opened 2026-08-13
fix(mito2): keep deletion markers when compacting part of a window

### LINKED ISSUE #8871: Deleted rows can become visible again after TWCS compacts part of a time window
### What type of bug is this?

Data corruption, Incorrect result

Storage Engine

### Minimal reproduce step

`TwcsPicker::find_inputs` decides whether a compaction may drop deletion markers from the shape of the **whole** time window:

```rust
// src/mito2/src/compaction/twcs.rs
let filter_deleted = found_runs <= 2 && !self.append_mode && !window_has_overlap(files, windows);
```

That reasoning only holds if the compaction then takes every file the shape covers. It does not: the selection is narrowed afterwards, and `filter_deleted` reaches the compactor unchanged (`CompactionSstReaderBuilder { filter_deleted: output.filter_deleted, .. }` in `compaction/compactor.rs`). When a deletion marker ends up inside the compacted set while the file holding the row it masks is left behind, the marker is dropped from the output and the old row becomes visible again. No later compaction can repair it — the marker is physically gone.

There are two ways to get there.

**1. The input file num limit.** `find_inputs` caps a compaction at `DEFAULT_MAX_INPUT_FILE_NUM = 32` files by sorting file groups by size and taking the smallest ones, without revisiting `filter_deleted`:

```
create table with compaction.type = twcs, compaction.twcs.time_window = '1h'
put key 'a' at ts 0s..3000s, flush          -- one large file spanning the window
delete key 'a' at ts 10s..16s, flush        -- the marker, in a tiny file
31 more flushes, each putting key 'a' over a 6s range that does not overlap
the others (20s..26s, 30s..36s, ... 320s..326s)
compact
```

The window holds 33 file groups in 2 runs, so `filter_deleted` is `true`. The limit then picks the 32 tiny groups (marker included) and leaves the large file behind. Rows `10s..16s` are visible again after the compaction.

**2. Compacting part of a single run.** With one run, `merge_seq_files` picks a subset of it. That is sound as long as a run really holds no overlapping files. Run detection uses `Ranged::overlap`, which compares time ranges *exclusively*:

```rust
// src/mito2/src/compaction/run.rs
if lhs_start.max(rhs_start) >= lhs_end.min(rhs_end) {
    return false;
}
```

So a file covering a single timestamp — what a flush of one delete produces — never overlaps anything and lands in the same run as the file it deletes rows from. `find_overlapping_items` uses `overlap_inclusive` for exactly this reason; `find_sorted_runs` does not. Same setup as above, but with the delete covering a single timestamp (`1s..2s`) and the other 31 flushes each writing one row of a different key: one run, `merge_seq_files` picks 32 of the 33 groups, and ts `1s` is back.

Both are covered by regression tests in the linked PR (`test_compaction_input_limit_keeps_rows_deleted` and `test_compaction_of_part_of_a_run_keeps_rows_deleted` in `src/mito2/src/engine/compaction_test.rs`), plus a picker-level test asserting `filter_deleted` after the file num limit truncates.

Rows deleted before the compaction stay deleted after it.

### What did you see instead?

The deleted rows are returned by scans again, because the compaction output no longer carries their deletion markers while the files holding the old values were never part of that compaction.

macOS (Darwin 25.5.0) arm64

`main` at 546625c45 (1.3.0 dev)

```bash
Region (...) compaction pick result: current window: 1970-01-01 01:00:00+0000, found runs: 2,
file num: 33, max output file size: None, filter deleted: true, input files: [...]
Compaction for region ... enforces max input file num limit: 32, current total: 33, input: [...]
```

fengjiachun:

Closes #8871

`TwcsPicker::find_inputs` decides whether a compaction may drop deletion markers from the shape of the **whole** time window:

```rust
let filter_deleted = found_runs <= 2 && !self.append_mode && !window_has_overlap(files, windows);
```

That reasoning only holds if the compaction then takes every file the shape covers, and it doesn't. The selection is narrowed afterwards — `reduce_runs` and `merge_seq_files` both return a subset, and the `DEFAULT_MAX_INPUT_FILE_NUM = 32` limit narrows it further by sorting file groups by size and keeping the smallest — while `filter_deleted` travels to the compactor unchanged. When a deletion marker ends up inside the compacted set while the file holding the row it masks is left behind, the marker is dropped from the output and the old row becomes visible again. Nothing can repair it later: the marker is physically gone.

This PR re-checks the final selection against the rest of the window and turns `filter_deleted` off when anything left behind still overlaps the inputs:

```rust
if filter_deleted && overlaps_files_left_behind(&inputs, &files_to_merge) {
    filter_deleted = false;
}
```

The check compares ranges **inclusively**, which matters beyond the file num limit. Run detection (`find_sorted_runs`) uses `Ranged::overlap`, which is exclusive, so two files that share only a boundary timestamp — for instance a flush of a single delete, whose file covers one timestamp — count as non-overlapping and land in the same run as the file they delete rows from. `merge_seq_files` may then compact one without the other. `find_overlapping_items` already uses `overlap_inclusive` for the same reason; this check follows it.

The picking itself is untouched — the same files are compacted as before, only the deletion-filtering decision changes — so this stays conservative in exactly the cases that were unsafe and keeps filtering deletions everywhere else. In particular a window whose file groups do not overlap at all (the "merge many small files" case) still filters deleted rows after the file num limit truncates its input, which `test_limit_max_input_files_still_filters_without_overlap` pins down.

### Tests

Three tests that fail on `main` and pass here:

- `compaction::twcs::tests::test_limit_max_input_files_keeps_deletion_markers` — a window of 33 file groups in 2 runs where the file num limit drops the one large group; asserts `filter_deleted` is off.
- `engine::compaction_test::test_compaction_input_limit_keeps_rows_deleted` — the same layout end to end, with real deletes; asserts the deleted rows do not come back after the compaction.
- `engine::compaction_test::test_compaction_of_part_of_a_run_keeps_rows_deleted` — the single-run variant, where the delete file covers one timestamp and `merge_seq_files` compacts 32 of the 33 groups.

Plus `test_limit_max_input_files_still_filters_without_overlap` as a guard against making the check too conservative.

`cargo nextest run -p mito2` is green (1116 tests).

## PR Checklist

### DIFF
--- a/src/mito2/src/compaction/twcs.rs
+++ b/src/mito2/src/compaction/twcs.rs
@@ -169,7 +169,7 @@ impl TwcsPicker {
         let found_runs = sorted_runs.len();
         // We only remove deletion markers if we found less than 2 runs and not in append mode.
         // because after compaction there will be no overlapping files.
-        let filter_deleted =
+        let mut filter_deleted =
             found_runs <= 2 && !self.append_mode && !window_has_overlap(files, windows);
         if found_runs == 0 {
             return (vec![], filter_deleted);
@@ -210,6 +210,14 @@ impl TwcsPicker {
             );
         }
 
+        // The inputs are only a subset of the window: `reduce_runs` and `merge_seq_files` both
+        // narrow the selection down and the file num limit above narrows it further. A deletion
+        // marker may only be dropped when every file that can hold a row it masks is compacted
+        // along with it, so re-check the final selection against what stays in the window.
+        if filter_deleted && overlaps_files_left_behind(&inputs, &files_to_merge) {
+            filter_deleted = false;
+        }
+
         if inputs.len() > 1 {
             // If we have more than one group to compact.
             log_pick_result(
@@ -225,6 +233,20 @@ impl TwcsPicker {
         }
         (inputs, filter_deleted)
     }
+}
+
+/// Returns whether any file group of `window_files` that is not part of `inputs` overlaps
+/// `inputs`. Such a group keeps rows a deletion marker in `inputs` masks, so the compaction
+/// must not filter deleted rows out.
+///
+/// Overlapping is inclusive on both boundaries here: two files that only share a boundary
+/// timestamp can still hold the same row.
+fn overlaps_files_left_behind(inputs: &[FileGroup], window_files: &[FileGroup]) -> bool {
+    let picked: HashSet<_> = inputs.iter().flat_map(|fg| fg.file_ids()).collect();
+    window_files
+        .iter()
+        .filter(|fg| !fg.file_ids().iter().any(|id| picked.contains(id)))
+        .any(|fg| inputs.iter().any(|input| input.overlap_inclusive(fg)))
 }
 
 #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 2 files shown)


fengjiachun:
@codex review

v0y4g3r (approved):
LGTM

PR #8873, opened 2026-08-13
perf: optimize Prometheus remote write v2 decoding

shuiyisong:

This pull request improves Prometheus Remote Write v2 ingestion by replacing full-request Prost materialization with a borrowed envelope and fused per-series decoder that builds sample and native-histogram rows directly. It reuses sample row templates, borrows symbol-table label names, and avoids repeated fixed-column allocations, allowing v2 to match or outperform the tuned v1 path in the included workloads.

It also adds an equivalent v1/v2 Criterion benchmark matrix and documents the updated decoder and codec-timing boundaries. The PRW v2 wire contract and output schema remain unchanged; compression fallback, malformed-wire validation, metadata handling, and experimental native-histogram gating are preserved.

### Benchmark results

`cargo bench -p servers --bench bench_prom --features testing -- prom_rw_`

Criterion point estimates from the same local run; lower is better. Full-path cases compare equivalent protobuf-to-row boundaries.

#### Payload size

| Workload | v1 | v2 | Bytes saved |
|---|---:|---:|---:|
| fixture | 189,265 B | 50,569 B | 138,696 B (73.3%) |
| 1k series × 1 sample | 221,563 B | 48,313 B | 173,250 B (78.2%) |
| 1k series × 10 samples | 383,563 B | 211,313 B | 172,250 B (44.9%) |
| 5k series × 1 sample | 1,112,313 B | 244,313 B | 868,000 B (78.0%) |

#### Raw Prost decode

| Workload | v1 | v2 | Time saved |
|---|---:|---:|---:|
| fixture | 506.64 µs | 154.43 µs | 352.21 µs (69.5%) |
| 1k × 1 | 735.07 µs | 172.90 µs | 562.17 µs (76.5%) |
| 1k × 10 | 927.13 µs | 356.57 µs | 570.56 µs (61.5%) |
| 5k × 1 | 3,739.10 µs | 855.50 µs | 2,883.60 µs (77.1%) |

#### Full decode-to-rows path

| Workload | v1 custom | v1 Prost | v2 manual | v2 vs v1 custom |
|---|---:|---:|---:|---:|
| fixture | 550.31 µs | 955.32 µs | 504.68 µs | 45.63 µs faster (8.3%) |
| 1k × 1 | 622.81 µs | 1,230.90 µs | 511.65 µs | 111.16 µs faster (17.8%) |
| 1k × 10 | 2,474.40 µs | 5,142.80 µs | 2,297.70 µs | 176.70 µs faster (7.1%) |
| 5k × 1 | 3,064.70 µs | 6,292.20 µs | 2,439.10 µs | 625.60 µs faster (20.4%) |

The optimized v2 path is smaller and faster to decode while also outperforming the tuned v1 custom full path across all four workloads.

## PR Checklist

### DIFF
--- a/src/servers/src/prom_remote_write/v2.rs
+++ b/src/servers/src/prom_remote_write/v2.rs
@@ -22,8 +22,11 @@ use api::greptime_proto::io::prometheus::write::v2::{
 #[cfg(test)]
 use api::greptime_proto::io::prometheus::write::v2::{Request, TimeSeries};
 use api::helper::ColumnDataTypeWrapper;
+use api::v1::helper::{field_column_schema, time_index_column_schema};
 use api::v1::value::ValueData;
-use api::v1::{ColumnSchema, ListValue, RowInsertRequest, Rows, SemanticType, Value};
+use api::v1::{
+    ColumnDataType, ColumnSchema, ListValue, RowInsertRequest, Rows, SemanticType, Value,
+};
 use bytes::{Buf, Bytes};
 use common_grpc::precision::Precision;
 use common_query::native_histogram::*;
@@ -390,6 +393,8 @@ fn decode_series_leaves(
     mut rows_remaining: usize,
     scratch: &mut LeafScratch,
 ) -> Result<()> {
+    let mut sample_row_template = None;
+
     while buf.has_remaining() {
         let (tag, wire_type) = decode_key(&mut buf).context(error::DecodePromRemoteRequestSnafu)?;
         match tag {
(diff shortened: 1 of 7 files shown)


src/servers/src/prom_remote_write/v2.rs
```diff
@@ -258,6 +316,250 @@ fn metric_type_value(wire_type: i32) -> Option<&'static str> {
...
+    metadata: &mut Metadata,
+) -> std::result::Result<SeriesCounts, DecodeError> {
+    labels_refs.clear();
+    metadata.clear();
+    let mut counts = SeriesCounts::default();
+
+    while buf.has_remaining() {
+        let (tag, wire_type) = decode_key(&mut buf)?;
+        match tag {
+            1 => uint32::merge_repeated(wire_type, labels_refs, &mut buf, DecodeContext::default())
```
sunng87: can we define constants for this 1/2/3/4/5

src/servers/src/row_writer.rs
```diff
@@ -326,25 +349,29 @@ pub(crate) fn write_by_schema(
     Ok(())
 }
 
-fn write_by_semantic_type(
+fn write_by_semantic_type<K>(
     table_data: &mut TableData,
     semantic_type: SemanticType,
-    ktv_iter: impl Iterator<Item = (String, ColumnDataType, Option<ValueData>)>,
+    ktv_iter: impl Iterator<Item = (K, ColumnDataType, Option<ValueData>)>,
```
sunng87, line 355: Just wonder is there any chance to make this IntoIterator so we can take the String key without name.into() which is a copy

sunng87 (approved):

src/servers/src/prom_remote_write/v2.rs
shuiyisong: Fixed in 17890d2534. I added named TIME_SERIES_*_TAG constants for fields 1 through 5 and reused them in both decoder passes (scan_series and decode_series_leaves).

src/servers/src/row_writer.rs
shuiyisong: Changing this parameter to IntoIterator would not remove the conversion; it would only let callers pass a collection directly. The current Iterator already yields K by value, so when K is String, name.into() is the identity conversion and moves the String without copying its buffer. In the v2 path K is &str because label names borrow from the request symbol table; converting to String is necessary only when a new schema column must own the name, while existing columns use the borrowed lookup without converting. I have therefore left this unchanged.

v0y4g3r (approved):
LGTM

PR #8874, opened 2026-08-13
feat: support quantile and fraction queries on mixed histograms

shuiyisong:

This pull request adds unified PromQL `histogram_quantile` and `histogram_fraction` evaluation for classic, native, and mixed histogram inputs. Previously, `histogram_fraction` only used the native-histogram UDF, while mixed tables could not evaluate classic buckets and native samples through one helper plan.

- Generalizes `HistogramFold` with explicit quantile and fraction operations.
- Evaluates mixed inputs in one scan while preserving native `le` labels and valid native results when classic buckets are malformed.
- Emits Prometheus-compatible information annotations for NaN observations and warnings for classic/native collisions.
- Keeps collector-backed helpers frontend-local during distributed planning.
- Accepts constant bound expressions while preserving the existing quantile Substrait encoding; mixed and fraction folds remain frontend-only.

## PR Checklist


src/query/src/promql/planner.rs
```diff
@@ -4683,18 +4687,28 @@ impl PromPlanner {
     }
 
     /// Try to build a [f64] from [PromExpr].
-    #[deprecated(
-        note = "use `Self::get_param_as_literal_expr` instead. This is only for `create_histogram_plan`"
-    )]
     fn try_build_float_literal(expr: &PromExpr) -> Option<f64> {
```
evenyag, line 4689: I think we shouldn't remove the deprecated note. Instead we should remove the usage of this method.

src/query/src/dist_plan/commutativity.rs
```diff
@@ -297,6 +298,21 @@ impl Categorizer {
...
+                        if udf.name() == NativeHistogramQuantile::name()
+                            || udf.name() == NativeHistogramFraction::name()
+                            || NativeHistogramDrop::is_annotation_udf(udf.name())
+                ))
+            })
+            .unwrap_or(true);
+        if has_annotated_histogram_helper {
+            return Commutativity::NonCommutative;
+        }
+
```
evenyag: This PR adds the NonCommutative back. **DO NOT** add this back. We need to add an explicit comment to mention that we don't want the annotation implementation block pushdown. We need to adjust the implementation.

evenyag (commented):
I only checked planner related changes.

src/query/src/promql/planner.rs
evenyag: Mentioned in https://github.com/GrepTimeTeam/greptimedb/pull/6297

shuiyisong: Fixed in 54312f4688. I removed `try_build_float_literal` entirely and routed histogram arguments through `get_param_as_literal_expr` plus DataFusion coercion/simplification, so we no longer remove its deprecation note or keep using the old helper.

src/query/src/dist_plan/commutativity.rs
shuiyisong: Fixed in 53a39c0ff6. I removed the annotation-specific `NonCommutative` check and the now-unused `is_annotation_udf` helper. Scalar UDFs remain pushdownable, with an explicit comment that distributed annotations require a separate return path from datanodes; the regression test now covers quantile, fraction, and drop helpers.

evenyag:
@codex review

src/query/src/dist_plan/commutativity.rs
```diff
@@ -315,7 +315,9 @@ impl Categorizer {
             | Expr::Exists(_)
             | Expr::InList(_)
             | Expr::Case(_) => Commutativity::Commutative,
-            Expr::ScalarFunction(_udf) => Commutativity::Commutative,
+            // Annotation collection must not affect distribution. Keep scalar UDFs pushdownable;
+            // distributed annotations require a separate return path from datanodes.
+            Expr::ScalarFunction(_) => Commutativity::Commutative,
```
evenyag, line 321: Maybe leave a todo that carries the annotation in our flight stream.

src/promql/src/extension_plan/histogram_fold.rs
```diff
@@ -1055,47 +1239,63 @@ impl HistogramFoldStream {
...
+                        fixed.push(val);
+                        prev = val;
                     }
-                    fixed.push(val);
-                    prev = val;
+                    Cow::Owned(fixed)
                 }
-                Cow::Owned(fixed)
             }
+            HistogramFoldOperation::Fraction { .. } => Cow::Borrowed(counter),
```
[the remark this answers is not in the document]
shuiyisong: Rechecked current Prometheus `main` at `7a73c9ae`: [`BucketFraction`](https://github.com/prometheus/prometheus/blob/7a73c9ae654485b92ff3000bbb7e7e03b6a6d1f2/promql/quantile.go#L520) still sorts/coalesces and interpolates raw cumulative counters, while `BucketQuantile` alone applies the monotonicity repair. This PR intentionally preserves Prometheus `histogram_fraction` behavior for non-monotonic inputs. If upstream changes that behavior, we can follow it separately.

src/common/query/src/native_histogram.rs
```diff
@@ -668,42 +684,66 @@ impl NativeHistogram {
             } else if self.uses_custom_buckets() {
                 if bucket.lower == f64::NEG_INFINITY {
                     if bucket.upper <= 0.0 {
-                        return bucket.upper;
+                        return (bucket.upper, None);
                     }
                     bucket.lower = 0.0;
                 } else if bucket.upper == f64::INFINITY {
-                    return bucket.lower;
+                    return (bucket.lower, None);
```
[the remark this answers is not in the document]
shuiyisong: Fixed in 1278a50fd6. `quantile_with_info` now computes the existing `NaNSkew` annotation before the custom-bucket boundary branches and returns it from both the leading `-Inf` and trailing `+Inf` early returns. The existing quantile NaN test now covers both cases. Validation passed with all 58 `common-query` tests, the focused PromQL collector test, formatting, and clippy.

src/query/src/dist_plan/commutativity.rs
shuiyisong: Added in `4035ab692f`. The TODO now tracks preserving datanode PromQL annotations through plan decoding, returning them in Flight stream metadata, and merging them on the frontend.

src/query/src/promql/planner.rs
```diff
@@ -4821,6 +4866,68 @@ impl PromPlanner {
...
+        self.ctx.field_columns = vec![output_field.clone()];
+        LogicalPlanBuilder::from(LogicalPlan::Extension(Extension {
+            node: Arc::new(folded),
+        }))
+        .filter(keep)
+        .context(DataFusionPlanningSnafu)?
+        .project(
+            std::iter::once(col(&time_index_column))
+                .chain(std::iter::once(value.alias(output_field)))
+                .chain(tag_columns.iter().map(col)),
```
discord9, line 4924: Please remove `le` from both this projection and `self.ctx.tag_columns`. Mixed `HistogramFold` preserves its input schema, so this currently leaks `le` into the helper result. A nested `histogram_quantile`/`histogram_fraction` then sees `has_le_tag()` and folds the already-computed result again as classic buckets; native rows may contribute their `le` string, while the normal null-filled case can fail the outer fold's `+Inf` check. Histogram helper outputs should not retain the classic bucket label.

src/promql/src/extension_plan/histogram_fold.rs
```diff
@@ -343,18 +429,28 @@ impl HistogramFold {
         )?))
     }
 
-    pub fn serialize(&self) -> Vec<u8> {
+    pub fn serialize(&self) -> DataFusionResult<Vec<u8>> {
+        if self.histogram_column.is_some() {
```
discord9, line 433: Please add a distributed-planner regression test for both guards here: build a mixed `HistogramFold` and a fraction `HistogramFold` over a table scan, run `DistPlannerAnalyzer`, and assert that `HistogramFold` remains outside `MergeScan.remote_input` while the remote input still encodes successfully. The serialize-only fraction assertion does not exercise the encode-probe boundary, and mixed serialization is not covered at all; without this test, a planner-ordering regression could expose this `NotImplemented` error to users or place frontend-only semantics on a datanode.

src/query/src/promql/planner.rs
shuiyisong: Addressed in `1305a38601` at the fold boundary rather than by dropping `le`. Mixed/native helper output must preserve native sample labels; Prometheus likewise preserves those labels and skips float samples whose `le` is missing or malformed. Classic `HistogramFold` now uses a tolerant numeric sort key and ignores null or unparseable bounds in both optimistic and safe paths. `nested_histogram_helpers_ignore_unparseable_bucket_labels` covers nested quantile and fraction with both null and nonnumeric native `le` values.

src/promql/src/extension_plan/histogram_fold.rs
shuiyisong: Added in `d84fb4d810`. `frontend_only_histogram_folds_stay_above_merge_scan` covers mixed quantile and fraction folds over a table scan. It verifies that each full fold fails Substrait encoding, `DistPlannerAnalyzer` keeps the fold above `MergeScan`, the remote input contains no fold, and that remote input encodes successfully.

src/query/src/promql/planner.rs
discord9: Thanks for adding the tolerance and the nested NULL/non-numeric cases. There is still a parseable-label hole: after this projection, a native row has a computed float and no histogram column, but retains its original `le`. With `native_le="+Inf"`, the outer helper selects the classic path and accepts it as a valid one-bucket classic histogram; nested `histogram_fraction(-Inf, +Inf, ...)` can then emit `1.0` rather than ignoring the already-folded native result. Please add this case to the nested regression and keep a discriminator so helper output is not classified as classic solely from a retained `le` label.

evenyag (approved):
I think we can merge this first.

PR #8878, opened 2026-08-13
feat(flow): support eval schedule offsets

discord9:

N/A

This PR adds `EVAL OFFSET` to batching Flow schedules so users can select a stable phase within an `EVAL INTERVAL`:

```sql
EVAL INTERVAL '1 hour'
EVAL OFFSET '17 minutes'
```

The schedule follows the UTC Unix-epoch lattice `offset + k * interval`, so this example evaluates at UTC `:17` every hour without being anchored to creation time or drifting after retries and recovery.

### Diff size

The diff is test-heavy. Of the **1,667 added lines** at the current head:

- **642 lines are production code** (a conservative count that also includes production comments);
- **794 lines are Rust tests**, including inline `#[cfg(test)]` modules;
- **86 lines are SQLness test inputs**;
- **145 lines are generated SQLness `.result` expectations**.

In other words, **1,025 added lines (~61%) are tests or generated expectations**, while the core implementation is **642 added lines across the parser, metadata transport/persistence, scheduler, execution context, and display paths**.

### Behavior

- `EVAL OFFSET` requires `EVAL INTERVAL`.
- The offset must be an integral number of seconds satisfying `0 <= offset < interval`.
- An omitted or zero offset uses the epoch phase; zero is omitted from rendered SQL.
- Newly created scheduled flows use batching mode so the interval and offset are honored.
- Sources with `ttl=instant` are rejected when a scheduled flow requires batching.
- The resolved typed schedule is persisted so retries and recovery preserve its phase.
- `now()` and `current_timestamp()` use the logical scheduled evaluation time.
- `CREATE OR REPLACE` preserves an unchanged schedule, recomputes it when the interval or offset changes, and resets an omitted offset to zero.
- `SHOW CREATE FLOW` and `information_schema.flows` expose non-zero offsets.
- Schedule arithmetic returns explicit errors at unrepresentable boundaries rather than saturating to an incorrect phase.
- Existing bounded catch-up limits remain in effect.

The schedule is intentionally based on UTC epoch seconds. Timezone, DST, calendar/month, and cron semantics are not part of this change. Deprecated streaming-engine compatibility and migration are also outside this PR; this change stays focused on batching Flow scheduling.

### Verification

After rebasing the current head onto `main`, the following passed with an external Cargo target:

```text
cargo fmt --all -- --check
cargo check -p common-meta -p flow -p operator -p sql --tests
```

Focused test results:

```text
common-meta create-flow/schedule tests: 42 passed
flow schedule/offset tests:            17 passed
operator flow-type tests:               4 passed
SQL parser offset test:                 1 passed
```

The trailing blank line in generated SQLness `.result` files is the generator's repository-wide output format and was not hand-edited.

## PR Checklist

User-facing SQL documentation can be updated separately.


src/flow/src/adapter/flownode_impl.rs
```diff
@@ -387,7 +387,9 @@ impl FlowDualEngine {
                         comment: Some(info.comment().clone()),
                         sql: info.raw_sql().clone(),
                         flow_options: info.options().clone(),
-                        eval_schedule: effective_eval_schedule_from_flow_info(&info),
+                        eval_schedule: effective_eval_schedule_from_flow_info(&info)
```
fengjiachun, line 390: Legacy flows created before this change can already persist both `flow_type=streaming` and an `eval_interval`. On recovery this path now attaches an `eval_schedule`, but `FlowDualEngine::create_flow` still routes solely by `flow_type`, and `StreamingEngine::create_flow_inner` discards both `eval_interval` and `eval_schedule`. After an upgrade and restart, such a flow is therefore recreated successfully while continuing to ignore its schedule, rather than being explicitly rejected as described in this PR. Please reject this legacy shape before dispatch and cover it with a compatibility fixture.

src/operator/src/statement/ddl.rs
```diff
@@ -844,10 +950,20 @@ impl StatementExecutor {
...
         )? {
             return Ok(flow_type);
         }
 
+        // A flow with `EVAL INTERVAL` (and therefore possibly `EVAL OFFSET`)
+        // always uses the batching scheduler: the fixed epoch-phase schedule is
+        // only honored by the batching scheduler, regardless of the SQL shape.
+        // Non-aggregate SQL is supported as an explicit full-query flow (the
+        // batching engine requires a time-window expression or EVAL INTERVAL).
+        if has_eval_interval {
```
fengjiachun, line 923: Forcing `EVAL INTERVAL` onto batching only covers newly created flows. Flows created before this change with `EVAL INTERVAL` + non-aggregate SQL (or an instant-TTL source) were persisted as `flow_type=streaming` *with* an `eval_schedule`, and `StreamingEngine::create_flow_inner` drops it (`eval_interval: _`), so after an upgrade they keep running as streaming and silently ignore the schedule.

The PR description says "historical persisted streaming flows carrying a schedule are explicitly rejected during recovery" and "a compatibility fixture covers old scheduled-streaming metadata", but neither is in the diff any more (`tests/compatibility/` is untouched). Could you either restore that handling or drop those two bullets so the description matches the code?

src/operator/src/statement/ddl.rs
```diff
@@ -259,6 +312,19 @@ fn determine_flow_type_for_source_state(
     }
 
     if has_instant_ttl_source_table {
+        if force_batching {
```
fengjiachun: This is unreachable when `defer_on_missing_source=true`: the missing-source branch above returns `Ok(Some(FlowType::Batching))` before `has_instant_ttl_source_table` is ever inspected. So `EVAL INTERVAL` + `defer_on_missing_source=true` + an existing instant-TTL source still yields Batching — exactly the case this check is meant to prevent. The test asserting that combination stays Batching has no comment explaining the intent.

src/operator/src/statement/ddl.rs
```diff
@@ -232,11 +236,60 @@ fn validate_and_normalize_flow_options(
...
 
+/// Rejects a raw `CreateFlowExpr` that smuggles the internal transient
+/// transport keys (`__greptime_internal_eval_offset_secs` /
+/// `__greptime_internal_eval_schedule`) through `flow_options`.
+///
+/// These keys are trusted only when inserted by the SQL path *after* user
+/// option validation (see `create_flow_procedure`). A key present on an expr
+/// arriving at a direct gRPC boundary is always a spoof attempt and must be
+/// rejected before any trusted extraction could happen.
+fn reject_spoofed_internal_flow_transport_keys(
```
fengjiachun: This looks redundant. Both entry points funnel into `create_flow_procedure`, which calls `validate_and_normalize_flow_options` — that already rejects `INTERNAL_EVAL_OFFSET_KEY` and `INTERNAL_EVAL_SCHEDULE_KEY` with the identical `is reserved for internal use` message. The function, its test, and the `create_flow_inner` call site can all go.

src/operator/src/statement/ddl.rs
```diff
@@ -783,6 +873,16 @@ impl StatementExecutor {
         expr.flow_options
             .insert(FlowType::FLOW_TYPE_KEY.to_string(), flow_type.to_string());
 
+        // Trusted insertion AFTER user option validation: this is the only
+        // place the transient offset key may be added, and it is added from the
+        // typed SQL AST value, not from anything the user supplied in options.
+        if let Some(offset_secs) = eval_offset_secs {
+            expr.flow_options.insert(
+                INTERNAL_EVAL_OFFSET_KEY.to_string(),
```
fengjiachun: The offset is serialized into `expr.flow_options` here and parsed straight back out by `CreateFlowTask::try_from` a few lines below — a purely local round-trip. `let mut task = CreateFlowTask::try_from(...)?; task.eval_offset_secs = eval_offset_secs;` is equivalent, and the actual frontend→metasrv hop is already covered by `From<CreateFlowTask> for PbCreateFlowTask`. That would also remove the need for the spoof guard above.

src/sql/src/parsers/create_parser.rs
```diff
@@ -452,6 +497,47 @@ impl<'a> ParserContext<'a> {
         )
     }
 
+    /// Parses an interval that must not contain months and returns the total
+    /// whole seconds together with whether the interval contains a sub-second
+    /// fraction. Whole seconds are computed by truncating the nanosecond part;
+    /// callers that require exact whole-second precision (e.g. `EVAL OFFSET`)
+    /// must reject `has_fractional_secs`.
+    fn parse_interval_no_month_whole_secs(&mut self, context: &str) -> Result<(i64, bool)> {
```
fengjiachun, line 493: This is byte-identical to `parse_interval_no_month` apart from the extra bool. Since that one now has a single caller (`EXPIRE AFTER`), it could just delegate: `Ok(self.parse_interval_no_month_whole_secs(context)?.0)`.

Also, the `interval.months as i64 * 60 * 60 * 24 * 3044 / 1000` term is dead in both copies — `months != 0` is rejected a few lines above.

src/operator/src/statement/ddl.rs
```diff
@@ -741,23 +807,39 @@ impl StatementExecutor {
         query_context: QueryContextRef,
     ) -> Result<Output> {
         // TODO(ruihang): do some verification
-        let expr = expr_helper::to_create_flow_task_expr(stmt, &query_context)?;
+        let expr = expr_helper::to_create_flow_task_expr(stmt.clone(), &query_context)?;
```
fengjiachun: `stmt.clone()` copies the whole `CreateFlow` AST (query included) just to read `stmt.eval_offset`. Capturing `let eval_offset = stmt.eval_offset;` before moving `stmt` avoids the clone.

src/flow/src/adapter/flownode_impl.rs
discord9: Confirmed: historical metadata with `flow_type=streaming` plus a schedule can recover through the deprecated streaming engine and ignore that schedule. This PR no longer claims to reject that legacy shape and explicitly excludes deprecated streaming-flow compatibility and mixed-version rolling upgrades. I am keeping that compatibility policy out of this EVAL OFFSET change rather than reintroducing the removed compatibility fixture and recovery branch.

src/operator/src/statement/ddl.rs
discord9: Agreed that the removed recovery rejection and compatibility fixture are not present. The PR description has been corrected accordingly: deprecated streaming-flow compatibility and mixed-version rolling upgrades are explicit non-goals. This patch therefore does not claim support for that historical persisted shape.

src/operator/src/statement/ddl.rs
discord9: Fixed in `88a8077d57a`: the `force_batching && has_instant_ttl_source_table` rejection now runs before missing-source deferral. A deferred flow with no forced schedule remains pending batching, while `EVAL INTERVAL` plus any known instant-TTL source is rejected even if another source is missing. The focused source-state test now covers this combination.

src/operator/src/statement/ddl.rs
discord9: Removed the duplicate boundary guard, its call site, and its dedicated test in `88a8077d57a`. Both SQL and direct gRPC paths still enter `create_flow_procedure`, where `validate_and_normalize_flow_options` rejects the reserved internal keys; the centralized reserved-key tests remain.

src/operator/src/statement/ddl.rs
discord9: Simplified in `88a8077d57a`: the operator now converts the expression normally and assigns the validated typed `eval_offset_secs` directly on `CreateFlowTask`, removing the local option-map serialization/reparse round-trip. The existing `From<CreateFlowTask> for PbCreateFlowTask` conversion still carries the offset across the actual frontend-to-metasrv boundary.

src/sql/src/parsers/create_parser.rs
discord9: Deduplicated in `88a8077d57a`: `parse_interval_no_month` now delegates to `parse_interval_no_month_whole_secs`, and the unreachable month arithmetic is removed after the nonzero-month rejection. Existing CREATE FLOW interval and offset parser tests pass.

src/operator/src/statement/ddl.rs
discord9: Fixed in `88a8077d57a`: `stmt.eval_offset` is captured before moving `stmt` into `to_create_flow_task_expr`, so the full CREATE FLOW AST/query is no longer cloned.

src/common/meta/src/rpc/ddl.rs
```diff
@@ -1296,6 +1296,13 @@ pub struct CreateFlowTask {
     /// Duration in seconds. Data older than this duration will not be used.
     pub expire_after: Option<i64>,
     pub eval_interval_secs: Option<i64>,
+    /// Phase offset of the evaluation schedule within `eval_interval_secs`,
+    /// in seconds. Must be in `[0, eval_interval_secs)`. `None` means a zero
+    /// offset (epoch-anchored schedule).
+    /// Transported through the transient option map (no proto field), see
+    /// `INTERNAL_EVAL_OFFSET_KEY`.
+    #[serde(default)]
```
WenyXu, line 1304: This adds `eval_offset_secs` to the persisted `CreateFlowTask`, and that field affects schedule reconstruction in `on_prepare`. The serde unit tests cover direct decoding, but the repository invariant requires a `tests/compatibility` case for persisted-format changes. Could we add or extend a Flow compatibility fixture that exercises recovery of old metadata and verifies that an offset schedule retains its phase after persistence/recovery?

fengjiachun (approved):
LGTM

src/common/meta/src/rpc/ddl.rs
discord9: Added `flow_eval_offset_persistence` in `711789f2684`. The distributed compatibility run now creates a v1.3+ flow with `EVAL INTERVAL '10s'` and non-zero `EVAL OFFSET '3s'`, restarts on preserved metadata, and verifies the recovered `SHOW CREATE FLOW` output contains `EVAL OFFSET '3 s'`. The generated snapshot was reviewed, the internal transport option is absent, and the accepted snapshot passed on rerun.

PR #8880, opened 2026-08-13
feat: manage semantic table options via ALTER TABLE SET/UNSET

killme2008:

#8609

Semantic declarations can now be managed on existing tables (previously `ALTER TABLE SET` rejected all `greptime.semantic.*` keys):

```sql
ALTER TABLE opentelemetry_traces SET 'greptime.semantic.entity.host.id' = 'resource_attributes.host.id';
ALTER TABLE opentelemetry_traces UNSET 'greptime.semantic.entity.host.id';
```

Changes:

- `AnnotationFamily` (table crate) classifies marker-style options that no region consumes — `Semantic` and `RepartitionHint`. Classification, per-family logical-table eligibility, batch rules, and validation live there; the previous hand-rolled special cases for `repartition.column.hint` (converter branches, `is_metadata_only_alter` check, dedicated `SetRepartitionColumnHint`/`UnsetRepartitionColumnHint` AlterKinds) are deleted. Existing hint behavior and messages are unchanged.
- New `AlterKind::SetAnnotations` / `UnsetAnnotations` go through the existing metadata-only procedure flow — no region dispatch, no proto / store-api changes. Mixing annotation keys with regular options in one statement is rejected, and the classifier propagates that error everywhere (a mixed batch on a logical table reports it instead of `UnexpectedLogicalRouteTable`).
- CREATE and ALTER validate through one core (`check_annotation`) with thin adapters keeping each entry point's existing error variants and status codes: ALTER missing column stays `4002 TableColumnNotFound`, CREATE stays `1004 InvalidArguments`. SET is strict; UNSET only checks the namespace. `MODIFY COLUMN` on a declared entity column rejects types that don't render as strings.
- Works on logical metric tables (regular alter-table task, locks only the logical table); the repartition hint stays physical-only via `allows_logical_tables`.
- First commit fixes a latent bug: `AlterLogicalTablesProcedure` never actually acquired its logical table locks (`table_info_values` is empty until `Prepare`, locks are fixed at submission). Ids are now resolved at submission and persisted with `#[serde(default)]`; post-`Prepare` dumps from older versions recover locks from the persisted table snapshots.

Tested by unit tests in the four touched crates plus sqlness `table_semantics` / `semantic_graph` / `show_create`: declare-after-ingest, UNSET, logical tables, error paths, and unchanged repartition-hint behavior.

## PR Checklist

### DIFF
--- a/src/common/grpc-expr/src/alter.rs
+++ b/src/common/grpc-expr/src/alter.rs
@@ -29,7 +29,7 @@ use snafu::{OptionExt, ResultExt, ensure};
 use store_api::region_request::{SetRegionOption, UnsetRegionOption};
 use table::metadata::{TableId, TableMeta};
 use table::requests::{
-    AddColumnRequest, AlterKind, AlterTableRequest, ModifyColumnTypeRequest,
+    AddColumnRequest, AlterKind, AlterTableRequest, AnnotationFamily, ModifyColumnTypeRequest,
     REPARTITION_COLUMN_HINT_KEY, SetDefaultRequest, SetIndexOption, UnsetIndexOption,
 };
 
@@ -43,6 +43,52 @@ use crate::error::{
 
 const LOCATION_TYPE_FIRST: i32 = LocationType::First as i32;
 const LOCATION_TYPE_AFTER: i32 = LocationType::After as i32;
+
+/// Classifies a SET/UNSET key batch: `Ok(Some(family))` when every key belongs
+/// to the same annotation family, `Ok(None)` when none does, and an error on a
+/// mixed batch — annotation alters skip region dispatch, so they cannot share
+/// a statement with options that regions must see.
+fn annotation_family_of_keys<'a>(
+    mut keys: impl Iterator<Item = &'a str>,
+) -> Result<Option<AnnotationFamily>> {
+    let Some(first) = keys.next() else {
+        return Ok(None);
+    };
+    let family = AnnotationFamily::of_key(first);
+    for key in keys {
+        let this = AnnotationFamily::of_key(key);
+        if this != family {
+            let prefix = family.or(this).unwrap().prefix();
+            return error::InvalidTableOptionRequestSnafu {
+                err_msg: format!(
+                    "`{prefix}*` options must be altered separately from other table options"
+                ),
+            }
+            .fail();
+        }
+    }
+    Ok(family)
+}
+
+/// Returns the annotation family when `kind` is a SET/UNSET whose keys all
+/// belong to one family — the alters that only rewrite table metadata and skip
+/// region dispatch. Mixed batches answer `None`; [`alter_expr_to_request`]
+/// rejects them with a precise error.
+pub fn annotation_alter_family(kind: &Kind) -> Option<AnnotationFamily> {
+    match kind {
+        Kind::SetTableOptions(api::v1::SetTableOptions { table_options }) => {
+            annotation_family_of_keys(table_options.iter().map(|option| option.key.as_str()))
+                .ok()
+                .flatten()
+        }
+        Kind::UnsetTableOptions(api::v1::UnsetTableOptions { keys }) => {
+            annotation_family_of_keys(keys.iter().map(|key| key.as_str()))
+                .ok()
+                .flatten()
+        }
+        _ => None,
+    }
+}
 
 fn set_index_option_from_proto(set_index: api::v1::SetIndex) -> Result<SetIndexOption> {
     let options = set_index.options.context(MissingAlterIndexOptionSnafu)?;
@@ -179,6 +225,16 @@ pub fn alter_expr_to_request(
                 AlterKind::SetRepartitionColumnHint {
                     column_name: option.value.clone(),
                 }
+            } else if let Some(family) = annotation_family_of_keys(
+                table_options.iter().map(|option| option.key.as_str()),
+            )? {
+                AlterKind::SetAnnotations {
+                    family,
+                    options: table_options
+                        .iter()
+                        .map(|option| (option.key.clone(), option.value.clone()))
+                        .collect(),
+                }
             } else {
                 AlterKind::SetTableOptions {
                     options: table_options
@@ -203,6 +259,13 @@ pub fn alter_expr_to_request(
                     }
                 );
                 AlterKind::UnsetRepartitionColumnHint
+            } else if let Some(family) =
+                annotation_family_of_keys(keys.iter().map(|key| key.as_str()))?
+            {
+                AlterKind::UnsetAnnotations {
+                    family,
+                    keys: keys.clone(),
+                }
             } else {
                 AlterKind::UnsetTableOptions {
                     keys: keys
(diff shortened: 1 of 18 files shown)


killme2008:
Fixed a bug: `AlterLogicalTablesProcedure` never actually acquired its logical table locks (`table_info_values` is empty until Prepare, locks are fixed at submission). Ids are now resolved at submission and persisted with `#[serde(default)]` for old procedure state. Please take a look @WenyXu

src/common/grpc-expr/src/alter.rs
```diff
@@ -44,6 +44,52 @@ use crate::error::{
...
+}
+
+/// Returns the annotation family when `kind` is a SET/UNSET whose keys all
+/// belong to one family — the alters that only rewrite table metadata and skip
+/// region dispatch. Mixed batches answer `None`; [`alter_expr_to_request`]
+/// rejects them with a precise error.
+pub fn annotation_alter_family(kind: &Kind) -> Option<AnnotationFamily> {
+    match kind {
+        Kind::SetTableOptions(api::v1::SetTableOptions { table_options }) => {
+            annotation_family_of_keys(table_options.iter().map(|option| option.key.as_str()))
```
WenyXu, line 89: Converting the classification error to `None` changes the error returned for directly submitted logical-table DDL. A mixed semantic/regular batch is treated as a non-annotation alter, so `handle_alter_table_task()` returns `UnexpectedLogicalRouteTable` instead of `InvalidTableOptionRequest`.

Could we return `Result<Option<AnnotationFamily>>` and propagate the error before checking the table route? The existing test should assert the mixed-batch error rather than make `None` part of the contract.

src/common/meta/src/ddl_manager.rs
```diff
@@ -1053,8 +1072,18 @@ async fn handle_alter_table_task(
         .get(table_id)
         .await?
         .context(TableRouteNotFoundSnafu { table_id })?;
+    // Semantic annotation alters only rewrite the logical table's own
+    // metadata (no region dispatch), so they may target logical tables.
+    let semantic_annotation_alter =
```
WenyXu: This logical-table exception is not directly covered by the current common-meta tests; the standalone SQL case does not exercise this routing branch.

Could we add a test using `TableRouteValue::Logical` that verifies a semantic ALTER updates only the logical table metadata, sends no datanode request, and returns `InvalidTableOptionRequest` for a mixed semantic/regular batch?

src/table/src/requests.rs
```diff
@@ -301,6 +301,27 @@ pub struct ModifyColumnTypeRequest {
     pub target_type: ConcreteDataType,
 }
 
+/// A family of annotation table options: pure metadata markers that no region
+/// consumes. Setting or unsetting them only rewrites the table's
+/// `extra_options`, so the alter skips region dispatch entirely.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum AnnotationFamily {
```
WenyXu, line 308: `AnnotationFamily` currently abstracts only `Semantic`, while the behavior is not actually generic: `set_annotations()` immediately destructures it, the operator and procedure treat every family as metadata-only, and the metasrv logical-table path permits only `Semantic`. A future variant could therefore be routed inconsistently.

Could we use `SetSemanticOptions` and `UnsetSemanticOptions` for now? If the generic abstraction is intentional, its metadata-only behavior, logical-table eligibility, and validation rules should be centralized in one exhaustive classification API.

src/common/grpc-expr/src/alter.rs
```diff
@@ -179,6 +225,16 @@ pub fn alter_expr_to_request(
                 AlterKind::SetRepartitionColumnHint {
                     column_name: option.value.clone(),
                 }
+            } else if let Some(family) = annotation_family_of_keys(
+                table_options.iter().map(|option| option.key.as_str()),
+            )? {
+                AlterKind::SetAnnotations {
+                    family,
+                    options: table_options
```
WenyXu, line 223: nit: These match arms own `table_options` and `keys`, so the payload does not need to be cloned after classification. Could we consume `table_options` with `into_iter()` and move `keys` directly into `UnsetAnnotations`?

src/common/grpc-expr/src/alter.rs
killme2008: Fixed in 6c1d6d2 — `annotation_alter_family` now returns `Result<Option<AnnotationFamily>>` and the DDL manager propagates it before the route guard (via `ConvertAlterTableRequest`), so a mixed batch on a logical table reports "must be altered separately" instead of `UnexpectedLogicalRouteTable`. The flow classifiers (`is_metadata_only_alter`, `AlterTableFlow::from_kind`) also return `Result` now — no caller interprets the error as "not an annotation alter". The test asserts the mixed-batch error as you suggested.

src/common/meta/src/ddl_manager.rs
killme2008: Added `test_logical_table_annotation_alter_routing` (9b43e86) at the DDL manager level: semantic SET on a `TableRouteValue::Logical` table succeeds with zero datanode requests and only the logical table's metadata updated; a mixed batch reports its own error rather than the route guard's; the repartition hint stays rejected on logical routes.

src/table/src/requests.rs
killme2008: Took the second option and made the abstraction real (6c1d6d2): `repartition.column.hint` is now the second family — its converter special case, `is_metadata_only_alter` special case, and the dedicated `SetRepartitionColumnHint`/`UnsetRepartitionColumnHint` AlterKind pair are deleted. Classification, per-family logical-table eligibility (`allows_logical_tables`), batch cardinality, and validation all live on `AnnotationFamily` in the table crate; CREATE and ALTER validate through the same core with thin adapters that keep each entry point's existing error variants and status codes (ALTER missing column stays 4002, CREATE stays 1004). Existing hint behavior and messages are unchanged — the CREATE-side hint tests and `show_create` sqlness pass as-is.

src/common/grpc-expr/src/alter.rs
killme2008: Done in 6c1d6d2 — the arms consume `table_options`/`keys` with `into_iter()` now, no clones.

src/table/src/metadata.rs
```diff
@@ -959,6 +984,23 @@ impl TableMeta {
         for col_to_change in requests {
             let change_column_name = &col_to_change.column_name;
 
+            // A column referenced by an entity declaration may be dropped (the
+            // read-time derivation skips the stale declaration), but must not
+            // change to a type without a stable string form.
+            if !has_stable_string_form(&col_to_change.target_type)
```
fengjiachun, line 1021: This narrows the check to `ModifyColumnTypes`, but the loop it replaces in `verify_alter` ran over `new_meta.schema` — the schema *after* the alter — so it also covered `AddColumns`.

Since dropping a declared column is deliberately allowed, the declaration can be re-satisfied by a later `ADD COLUMN`:

```sql
CREATE TABLE t (ts TIMESTAMP TIME INDEX, svc STRING);
ALTER TABLE t SET 'greptime.semantic.entity.service.id' = 'svc';
ALTER TABLE t DROP COLUMN svc;       -- allowed, declaration goes stale
ALTER TABLE t ADD COLUMN svc BINARY; -- rejected before this PR, accepted now
```

The table now carries `greptime.semantic.entity.service.id = 'svc'` pointing at a Binary column — exactly the state `has_stable_string_form` exists to reject at DDL time, deferred to graph-scan time instead. Ingestion auto-alter can hit the same path when a dropped attribute comes back as a binary/JSON column.

Covering it in `add_columns` too (or validating the post-alter schema once in `builder_with_alter_kind`) would restore the previous guarantee.

src/table/src/metadata.rs
```diff
@@ -959,6 +984,23 @@ impl TableMeta {
...
+                    err: format!(
+                        "column `{change_column_name}` is referenced by entity option \
+                         `{key}` and must keep a type that renders as a string, got `{}`",
+                        col_to_change.target_type
+                    ),
+                }
+                .fail();
+            }
+
             let index = table_schema
```
fengjiachun: Minor: the entity check above runs before this existence lookup, so a MODIFY on a dropped-but-still-declared column reports the entity error instead of `ColumnNotExists`.

With `greptime.semantic.entity.service.id = 'svc'` set and `svc` already dropped, `ALTER TABLE t MODIFY COLUMN svc BINARY` returns `1004 InvalidArguments` rather than the `4002 TableColumnNotFound` every other MODIFY on a missing column returns, so a client can't tell "column gone" from "type rejected". Moving the check below this lookup fixes it.

src/table/src/metadata.rs
killme2008: Good catch — the old loop scanned the post-alter schema, so it covered this; the mutation-layer move only kept the MODIFY path. Fixed in e236e63: `add_columns` now runs the same guard (which also covers ingestion auto-alter re-adding a dropped attribute), with a test for the DROP → re-add sequence in both directions (binary rejected, string re-satisfies).

src/table/src/metadata.rs
killme2008: Fixed in e236e63 — the drift check moved below the existence lookup, so MODIFY on a missing column reports `ColumnNotExists` (4002) regardless of stale declarations; pinned with an explicit status-code assertion.

src/table/src/requests.rs
```diff
@@ -301,6 +301,211 @@ pub struct ModifyColumnTypeRequest {
...
+                "cannot set {REPARTITION_COLUMN_HINT_KEY} to the time index column"
+            ),
+        }
+    }
+}
+
+/// Validates one annotation option and returns the value to store — the
+/// repartition hint is trimmed to the bare column name, semantic values pass
+/// through unchanged.
+pub(crate) fn check_annotation(
```
WenyXu: nit: These helpers do more than check: `check_annotation()` validates and returns a normalized value, while `check_annotation_options()` validates and writes normalized values back. The nearby `validate_and_normalize_flow_options()` naming describes this contract more precisely. Could we use `validate_and_normalize_annotation*` here and rename the wrapper, `AnnotationCheckError`, and `test_check_annotations` consistently—for example, `AnnotationValidationError` and `test_validate_and_normalize_annotations`?

src/table/src/requests.rs
```diff
@@ -301,6 +301,211 @@ pub struct ModifyColumnTypeRequest {
...
+            ),
+            Self::RepartitionHint => {
+                format!("{REPARTITION_COLUMN_HINT_KEY} must be altered separately")
+            }
+        }
+    }
+}
+
+/// Table shape an annotation option is validated against.
+pub struct AnnotationCx<'a> {
```
WenyXu: nit: `Cx` is not commonly used for domain context types in this repository; nearby types generally spell out `Context`. Could we rename `AnnotationCx` to `AnnotationContext` for consistency?

src/table/src/metadata.rs
```diff
@@ -966,6 +1009,25 @@ impl TableMeta {
                     table_name,
                 })?;
 
+            // A column referenced by an entity declaration may be dropped (the
+            // read-time derivation skips the stale declaration), but must not
+            // change to a type without a stable string form. Checked after the
+            // existence lookup so a missing column keeps reporting
+            // `ColumnNotExists` regardless of stale declarations.
+            if !has_stable_string_form(&col_to_change.target_type)
```
WenyXu, line 1021: nit: This moves the entity-reference check ahead of the existing primary-key, time-index, castability, and nullability guards, changing which error is returned when a `MODIFY` violates multiple constraints. Could we preserve the previous validation priority by moving this check after the existing guards, while keeping the column lookup first so missing columns still return 4002?

killme2008: I think it's fine here.

src/table/src/requests.rs
killme2008: Good idea

killme2008: Renamed in the follow-up commit: `validate_and_normalize_annotation` / `validate_and_normalize_annotation_options` / `AnnotationValidationError`, plus the operator adapter and its test.

src/table/src/requests.rs
killme2008: Renamed to `AnnotationContext`.

WenyXu (approved):

PR #8884, opened 2026-08-14
feat(query): add experimental DataFusion spill-to-disk controls

discord9:

Adds experimental spill-to-disk controls for DataFusion-backed queries in the open-source edition, matching the capability previously available only in the enterprise edition.

- Adds `QuerySpillMode {default, custom, disabled}`, `QuerySpillCompression {uncompressed, lz4_frame, zstd}`, and `QueryMemoryPoolPolicy {greedy, fair}` configuration enums.
- Adds five `experimental_*` keys under `[query]`: `experimental_spill_mode`, `experimental_spill_path`, `experimental_spill_max_temp_directory_size` (default 1GiB), `experimental_spill_compression`, and `experimental_memory_pool_policy`.
- Uses the existing `QueryRuntimeProvider` seam: `default` preserves DataFusion behavior; `custom` configures spill path, quota, and compression and logs the effective settings once; `disabled` installs a disabled disk manager.
- Makes `MetricsMemoryPool` policy-selectable while retaining `TrackConsumersPool` for top-consumer error context. `greedy` is first-come-first-served; `fair` divides memory remaining after unspillable reservations among spillable reservations and may spill earlier.
- Covers parsing/defaults, runtime construction, memory-pool behavior, a controlled SortExec disk-spill path with direct spill-metric assertions, and serialized defaults through the config API integration test.
- Updates all query example TOMLs and regenerates `config/config.md`.

## PR Checklist

### DIFF
--- a/config/config.md
+++ b/config/config.md
@@ -125,6 +125,11 @@
 | `query` | -- | -- | The query engine options. |
 | `query.parallelism` | Integer | `0` | Parallelism of the query engine.<br/>Default to 0, which means the number of CPU cores. |
 | `query.memory_pool_size` | String | `50%` | Memory pool size for query execution operators (aggregation, sorting, join).<br/>Supports absolute size (e.g., "2GB", "4GB") or percentage of system memory (e.g., "20%").<br/>Setting it to 0 disables the limit (unbounded, default behavior).<br/>When this limit is reached, queries will fail with ResourceExhausted error.<br/>NOTE: This does NOT limit memory used by table scans. |
+| `query.experimental_memory_pool_policy` | String | `greedy` | Experimental memory pool allocation policy:<br/>- "greedy" (default): preserves current greedy allocation behavior.<br/>- "fair": shares memory evenly among spillable operators.<br/>Only effective when `memory_pool_size` is bounded (>0). |
+| `query.experimental_spill_mode` | String | `default` | Spill mode:<br/>- "default": preserve DataFusion built-in OS temp directory (default).<br/>- "custom": explicitly configure spill path, quota, and compression.<br/>- "disabled": explicitly disable disk spilling.<br/>Set this to "custom" before using the path/quota/compression keys below. |
+| `query.experimental_spill_path` | String | Unset | Spill directory path. Ignored unless mode is "custom". |
+| `query.experimental_spill_max_temp_directory_size` | String | `100GiB` | Maximum total size of spill directory (default: "100GiB").<br/>Ignored unless mode is "custom". |
+| `query.experimental_spill_compression` | String | `uncompressed` | Compression for spilled data files: "uncompressed" (default), "lz4_frame", "zstd".<br/>Ignored unless mode is "custom". |
 | `storage` | -- | -- | The data storage options. |
 | `storage.data_home` | String | `./greptimedb_data` | The working home directory. |
 | `storage.copy_root` | String | `./greptimedb_data/copy` | Root directory for standalone SQL access to local files.<br/>Relative SQL paths are resolved below this directory. Absolute paths are accepted only when<br/>they are inside this directory. Defaults to `<data_home>/copy`.<br/>Distributed deployments always reject SQL access to local files.<br/>Upgrade note: COPY commands and existing external tables that reference paths outside this<br/>directory will fail. Move those files below the copy root, set this option to a dedicated<br/>directory containing them, or migrate the files to object storage before upgrading. |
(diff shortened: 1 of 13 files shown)


config/config.md
```diff
@@ -123,6 +123,11 @@
...
+| `query.experimental_memory_pool_policy` | String | `greedy` | Experimental memory pool allocation policy:<br/>- "greedy" (default): preserves current greedy allocation behavior.<br/>- "fair": shares memory evenly among spillable operators.<br/>Only effective when `memory_pool_size` is bounded (>0). |
+| `query.experimental_spill_mode` | String | `default` | Spill mode:<br/>- "default": preserve DataFusion built-in OS temp directory (default).<br/>- "custom": explicitly configure spill path, quota, and compression.<br/>- "disabled": explicitly disable disk spilling.<br/>Set this to "custom" before using the path/quota/compression keys below. |
+| `query.experimental_spill_path` | String | Unset | Spill directory path. Ignored unless mode is "custom". |
+| `query.experimental_spill_max_temp_directory_size` | String | `100GiB` | Maximum total size of spill directory (default: "100GiB").<br/>Ignored unless mode is "custom". |
```
killme2008: The default value seems too high compared with those of the other disk quota parameters.

config/datanode.example.toml
```diff
@@ -278,6 +278,30 @@ parallelism = 0
 ## NOTE: This does NOT limit memory used by table scans.
 memory_pool_size = "50%"
 
+## Experimental memory pool allocation policy:
+## - "greedy" (default): preserves current greedy allocation behavior.
```
killme2008: Need to explain what's the `greedy` behavior

src/query/src/query_engine/runtime.rs
```diff
@@ -67,14 +69,67 @@ pub struct DefaultQueryRuntimeProvider;
...
-            RuntimeEnvBuilder::new().with_memory_pool(Arc::new(MetricsMemoryPool::new(
+            builder = builder.with_memory_pool(Arc::new(MetricsMemoryPool::new(
                 ctx.resolved_memory_pool_size,
-            )))
-        } else {
-            RuntimeEnvBuilder::new()
+                ctx.query_options.experimental_memory_pool_policy,
+            )));
+        }
+
```
killme2008, line 83: Add a info log when enabling spill mode below.

src/query/src/query_engine/state.rs
```diff
@@ -868,4 +882,367 @@ mod tests {
...
+        assert!(env.memory_pool.reserved() == 0);
+        // Disk manager remains default (OS temp dir)
+        assert!(env.disk_manager.tmp_files_enabled());
+    }
+
+    /// Custom mode with explicit path: disk manager is configured with the
+    /// custom directory and the directory actually gets created.
+    #[test]
+    fn test_build_runtime_env_custom_mode_with_path() {
+        // Use a temp directory managed manually so we don't need the `tempfile` crate.
```
killme2008, line 913: Please remove all the trivial comments! They are too many and meanless.

killme2008 (approved):
Good job!

A few comments attached, please take a look @discord9 

FYI, could we add an integration test that using the disk spill while querying?

config/config.md
discord9: Reduced the default maximum spill directory size from 100GiB to 1GiB and updated the option default, tests, example TOMLs, config API expectation, and generated config reference.

config/datanode.example.toml
discord9: `greedy` is now documented as first-come-first-served. The docs also contrast `fair`: after unspillable reservations are deducted, available memory is divided among spillable reservations, so it can trigger spilling earlier.

src/query/src/query_engine/runtime.rs
discord9: Added an info log when constructing the custom spill runtime. It reports the effective path, maximum temporary-directory size, and compression; default and disabled modes do not emit the enabled message.

src/query/src/query_engine/state.rs
discord9: Removed the trivial section banners and comments that repeated the code or assertions. Kept only comments that explain invariants or non-obvious test setup.

discord9:
Regarding the full-server spill integration test: I tried exercising spill through the HTTP query path, but the public query configuration does not expose DataFusion's sort spill reservation and in-place-sort thresholds, so a server-level workload could not trigger disk spill deterministically. Making it deterministic would require adding production configuration solely for the test, which I did not retain.

The existing controlled `test_sort_spill_smoke_with_custom_runtime` test uses the real `SortExec` and custom disk manager with a 512KiB memory pool, a 64KiB spill reservation, in-place sorting disabled, and 200k input rows. It verifies sorted output and directly requires `spill_count > 0`, `spilled_rows > 0`, and `spilled_bytes > 0`, along with spill-directory initialization and memory-reservation release. The config API remains covered by the File-backed integration test.

PR #8889, opened 2026-08-14
fix(query): preserve timestamp literal semantics in inserts

killme2008:

Follow-up to #8859.

## What’s changed and what’s your intention?

Handle timestamp string literals behind common `INSERT` source plans, including subqueries, `LIMIT`, filters, sorting, `DISTINCT`, and `UNION ALL`.

Fold constant literals at the assignment boundary so the source plan keeps its original types. Explicit timestamp casts keep the existing DataFusion behavior.

## PR Checklist


MichaelScofield (approved):
I originally intuitively thought that timezone in the `query_context` can either be passed to datafusion's session state then be used in datafusion's (or arrow's) cast; or be passed to datafusion's `ScalarValue::Timestamp*`'s second argument, then be handled by datafusion all the way down to itself. All in all, timezone parameter is not new, so it shouldn't be "corrected/parsed" outside the datafusion. But this fix looks legit, resolve the problem in some ways, so lgtm.

src/query/src/optimizer/type_conversion/insert_assignment.rs
```diff
@@ -0,0 +1,315 @@
...
+        // column is an INSERT-assigned literal in every branch.
+        for input in &union.inputs {
+            let Some(rewritten) = self.rewrite_output_column(input, output_idx, target_type)?
+            else {
+                return Ok(None);
+            };
+            inputs.push(Arc::new(rewritten));
+        }
+
+        Union::try_new(inputs).map(LogicalPlan::Union).map(Some)
```
fengjiachun: `try_new` is the **strict** constructor — it requires every column to have the same type across branches. But this rule runs before DataFusion's `TypeCoercion` (see `planner.rs`), and the SQL planner deliberately builds unions with `try_new_with_loose_types` precisely because coercion happens later. So mismatched branch types are a legal intermediate state here, and rebuilding with `try_new` rejects them.

The catch: `try_new` checks *all* columns, while this rule only rewrites the timestamp one. So it fails on columns it never touched:

```sql
INSERT INTO t (ts, st) SELECT '2026-08-06 12:00:00.001', now()
UNION ALL SELECT '2026-08-07 12:00:00.001', NULL;
-- UNION field 1 have different type in inputs: left has Timestamp(ns) whereas right has Null
```

`st` is `Timestamp(ns)` vs `Null` — TypeCoercion would have reconciled it, and this plans fine on main.

`test_insert_mixed_union_keeps_source_coercion` misses this because there the mismatch is on `ts`, so `rewrite_union` bails before reaching the rebuild.

`try_new_with_loose_types` is a drop-in fix — the two differ only in that check. Verified locally, `-p query` stays green.

src/query/src/optimizer/type_conversion/insert_assignment.rs
```diff
@@ -0,0 +1,315 @@
...
+            // them even though the Values schema already has the target type.
+            LogicalPlan::Values(values) => self.rewrite_values(values, output_idx, target_type),
+            // Once the source query has produced the target type, its casts and
+            // coercions belong to the source query rather than INSERT assignment.
+            _ if plan.schema().field(output_idx).data_type() == target_type => Ok(None),
+            LogicalPlan::Projection(projection) => {
+                self.rewrite_projection(projection, output_idx, target_type)
+            }
+            LogicalPlan::Union(union) => self.rewrite_union(union, output_idx, target_type),
+            // Filter, Sort and Distinct are excluded here: retyping a column
```
fengjiachun: Side effect: `UNION` (distinct) and `GROUP BY` still fall back to UTC parsing, so `UNION` vs `UNION ALL` store different instants. Worth a sqlness case.

src/query/src/optimizer/type_conversion/insert_assignment.rs
```diff
@@ -0,0 +1,315 @@
...
+        let target_type = assignment.schema.field(output_idx).data_type();
+        if !matches!(target_type, DataType::Timestamp(_, _)) {
+            continue;
+        }
+
+        let Some(input_idx) = assignment_input_index(expr, input.schema()) else {
+            continue;
+        };
+
+        let literal = lineage_literal(&input, input_idx).cloned();
```
fengjiachun, line 55: No single-consumer check here, unlike `rewrite_projection`. Fine for SQL INSERT (each target maps to a distinct source position), but flow builds `DmlStatement` directly. Worth noting the assumption.

src/query/src/optimizer/type_conversion/insert_assignment.rs
killme2008: Good catch!

src/query/src/optimizer/type_conversion/insert_assignment.rs
killme2008: Yes, this is a known limitation and is documented.

If we want to support them completely, it's complex and out of this PR.

src/query/src/optimizer/type_conversion/insert_assignment.rs
killme2008: Fixed in abd5c92 — `rewrite_union` now rebuilds with `try_new_with_loose_types`, matching how the SQL planner constructs pre-coercion unions. Your repro is pinned as `test_insert_union_tolerates_uncoerced_untouched_column`. Good catch on the untouched-column angle; the mixed-union test indeed bails before the rebuild.

src/query/src/optimizer/type_conversion/insert_assignment.rs
killme2008: UNION (distinct) is fixed rather than documented (abd5c92): `Distinct::All` joined the rewrite passthrough. It holds no expressions, and every surviving row feeds the same assignment cast; the one observable tradeoff — dedup keys on parsed instants instead of raw strings, so different spellings of the same instant collapse — is commented at the match arm. The sqlness case now shows UNION and UNION ALL storing the same instants.

GROUP BY turns out to be covered already by the fold path: the literal sits in the projection above the `Aggregate`, so `lineage_literal` resolves it before reaching the aggregate node:

```
Projection: TimestampMillisecond(1786852800001, None) AS ts, max(timestamps.st) AS st, ...
  Projection: Utf8("2026-08-16 12:00:00.001"), max(timestamps.st)
    Aggregate: groupBy=[[timestamps.note]], aggr=[[max(timestamps.st)]]
```

Both shapes are pinned in `test_insert_timestamp_literals_use_query_timezone`.

src/query/src/optimizer/type_conversion/insert_assignment.rs
killme2008: Went one step further than noting it (abd5c92): the top-level rewrite path now runs the same single-consumer guard as `rewrite_projection` before mutating the input, so a hand-built DML plan sharing one source column between targets falls back instead of retyping every reader. The fold path needs no guard — it never touches the source plan.

PR #8894, opened 2026-08-17
feat: allow widening the time index column's timestamp unit via ALTER TABLE, mito2 table only

sunng87:

The rational for this because some of the external systems only allows micro-second timestamp, so user will need to recreate table if they want to adopt.

This change allows user to widening timestamp unit, for example, from millis to micros, on non-metric-engine table. If it works well, I can follow up with a patch for metric-engine table.

## PR Checklist


src/mito2/src/series_index/searcher.rs
evenyag, line 283: We may need to change the index schema to use a fixed timestamp unit or the unit of the time index.

src/mito2/src/series_index/writer.rs
```diff
@@ -616,6 +616,12 @@ fn is_reserved_column(column_id: ColumnId) -> bool {
     column_id == ReservedColumnId::table_id() || column_id == ReservedColumnId::tsid()
 }
 
+/// Extracts raw i64 timestamp values for the `__series_min_ts`/`__series_max_ts`
+/// columns. NOTE: the unit is dropped; the values must always be interpreted in
+/// the unit of the region metadata the index is written with. Once this index
+/// is wired into scans, files written before/after a time index unit widening
+/// would carry mixed units — the index schema or the searcher must record the
+/// unit per file, or the index must be rebuilt on such an alter.
 fn timestamp_values(array: &ArrayRef) -> Result<Int64Array> {
```
evenyag, line 528: Handle timestamps unit here.

src/mito2/src/worker/handle_alter.rs
```diff
@@ -162,6 +162,10 @@ impl<S: LogStore> RegionWorkerLoop<S> {
             "Try to alter region {}, version.metadata: {:?}, version.options: {:?}, request: {:?}",
             region_id, version.metadata, version.options, request,
         );
+        if let Err(e) = check_time_index_widening_overflow(region.region_id, &version, &request) {
```
evenyag: Overflow validation happens independently after each region reaches its alter handler. The DDL procedure dispatches region alters concurrently, so for a partitioned table where one region fits and another overflows, successful regions can persist the new schema while the failing region returns a non-retryable InvalidArguments error. The procedure is then poisoned before table metadata is updated, leaving inconsistent region schemas; narrowing is forbidden, so the successful regions cannot be rolled back. Please preflight overflow across all regions before mutating any manifest, or use a two-phase validate/commit flow.

src/common/recordbatch/src/filter.rs
```diff
@@ -339,6 +340,171 @@ impl SimpleFilterEvaluator {
...
+            }));
+        }
+
+        let scalar = ScalarValue::try_from_array(self.literal.get().0, 0).ok()?;
+        let (value, unit) = timestamp_scalar_parts(&scalar)?;
+        let Some(value) = value else {
+            // A NULL literal never compares true (arrow comparison
+            // semantics: null results are filtered out).
+            return Some(TimestampUnitCast::Pruned);
+        };
```
evenyag, line 401: We can add a fast path if the scalar has the same unit as the target. We can return the filter directly.

src/mito2/src/series_index/writer.rs
sunng87: let's cover the series index in next patch before this one goes too large.

src/mito2/src/worker/handle_alter.rs
sunng87: updated. please check if this `check_time_index_widening_overflow` implementation is practical for addressing the issue. I think we should try to avoid it from opening too much files just for checking the time index range. This implementation should only visit in-memory state.

src/mito2/src/worker/handle_alter.rs
```diff
@@ -489,19 +489,22 @@ fn check_time_index_widening_overflow(
...
-                            start.to_iso8601_string(),
-                            end.to_iso8601_string(),
-                        ),
-                    }
-                    .fail();
+                        file.file_id(),
+                        err
+                    );
+                    return store_api::metadata::InvalidRegionRequestSnafu { region_id, err }
+                        .fail();
```
evenyag: Maybe we can allow this error.

sunng87: Sounds good. Fixed.

src/mito2/src/series_index/writer.rs
sunng87: tracked in #8996

evenyag (approved):

PR #8900, opened 2026-08-18
feat(otlp): support cumulative exponential histograms

shuiyisong:

This pull request adds opt-in ingestion for cumulative OTLP `ExponentialHistogram` metrics by normalizing them into GreptimeDB’s existing Prometheus-compatible native-histogram Struct.

### Design idea

Use OTLP only as a transport adapter and keep one canonical persisted representation. OTLP and Prometheus Remote Write 2.0 share the same validation and encoding path, avoiding a second storage format or protocol-specific query behavior.

### Implementation outline

- Extract the native-histogram validator and Struct encoder from Remote Write 2.0 into a shared server module.
- Convert OTLP dense buckets into Prometheus spans and integer deltas, including the required bucket-index shift.
- Support scales `-4..=8` directly and downscale higher resolutions to schema `8` with checked bucket merging; reject lower scales, invalid counts, timestamps, thresholds, and arithmetic overflow.
- Preserve zero buckets, attributes, timestamps, start timestamps, legacy/non-legacy naming, and `NoRecordedValue` stale markers.
- Add `MetricsIngestOutcome` to track accepted and rejected points. Return partial success for mixed requests and `InvalidArgument` when every point is rejected.
- Add the disabled-by-default `otlp.experimental_enable_exponential_histogram` option and wire it through frontend, standalone, and OTLP/HTTP context configuration.
- Explicitly reject OTel Arrow exponential histograms because its current wire representation omits `zero_threshold`.
- Enforce float/native-histogram sample-kind exclusivity and emit semantic metadata only when a metric produces an accepted row.
- Update configuration examples, generated documentation, the native-histogram RFC, and focused unit and integration coverage.

### Compatibility and scope

The change adds no protobuf dependency or persisted-format change, and the new option defaults to `false`. Disabled or invalid exponential histogram points are now reported instead of being silently acknowledged.

Delta accumulation, min/max, exemplars, and OTel Arrow exponential-histogram ingestion remain deferred.

## PR Checklist


src/servers/src/otel_arrow.rs
```diff
@@ -42,6 +49,29 @@ impl<T> OtelArrowServiceHandler<T> {
...
+fn batch_status(batch_id: i64, outcome: MetricsIngestOutcome) -> BatchStatus {
+    let status_code = if outcome.accepted_data_points == 0 && outcome.rejected_data_points > 0 {
+        ArrowStatusCode::InvalidArgument
+    } else {
+        ArrowStatusCode::Ok
+    };
+    let status_message = outcome
+        .error_message
+        .map(|message| {
+            if message.contains("otlp.experimental_enable_exponential_histogram") {
```
sunng87: This substring match couples the Arrow handler to the exact wording of a message produced in `src/servers/src/otlp/metrics.rs` (`reject_data_points`). If that message is ever reworded — e.g. dropping the `experimental_` prefix when the option is stabilized, per the repo's config convention — this check silently stops matching and Arrow clients get the raw gate message instead of the accurate Arrow-format explanation.

Consider carrying a structured rejection reason on `MetricsIngestOutcome` (e.g. a small enum or a `gate_disabled: bool` field) so this decision is type-checked rather than string-matched. Only comment in this review I'd ask to address before merge.

src/servers/src/native_histogram.rs
```diff
@@ -0,0 +1,530 @@
...
+    }
+
+    Ok(ValueData::StructValue(api::v1::StructValue { items }))
+}
+
+fn validate_native_histogram(histogram: &Histogram, uses_float_counts: bool) -> Result<()> {
+    let exponential_overflow_index = validate_native_histogram_schema(histogram.schema)?;
+    validate_native_histogram_custom_values(histogram)?;
+    ensure!(
+        histogram.zero_threshold.is_finite() && histogram.zero_threshold >= 0.0,
```
sunng87: Worth flagging explicitly: this check is **new** relative to the old `prom_remote_write/v2.rs` code — the extracted `validate_native_histogram` did not previously validate `zero_threshold`. So as a side effect of the move, Remote Write 2.0 payloads with NaN or negative `zero_threshold` that used to be accepted are now rejected.

The behavior matches Prometheus and I think it's right, but since it's a tightening of an existing protocol hidden inside an extraction refactor, it deserves a line in the PR description (and a quick confirm that no existing RW2 users depend on it).

src/operator/src/insert.rs
```diff
@@ -1018,8 +1018,7 @@ impl Inserter {
             return Ok(None);
         };
 
-        // If accommodate_existing_schema is true, update request schema for Timestamp/Field columns
-        if accommodate_existing_schema {
+        if is_single_value {
```
sunng87, line 1021: Widening this guard from `accommodate_existing_schema` to `is_single_value` also changes the OpenTSDB and Prom-store paths (`handle_row_inserts(requests, ctx, true, true)` in `instance/opentsdb.rs` / `instance/prom_store.rs`), not just OTLP. Coherent with the RFC's one-sample-kind invariant, just confirming the broader blast radius is intended given the PR title scopes to OTLP.

src/servers/src/otlp/metrics.rs
```diff
@@ -368,16 +424,391 @@ fn encode_metrics(
...
+                outcome,
+            )?,
         }
+    } else {
+        false
+    };
+
+    if emitted {
+        // Stamp semantic metadata only after at least one row was accepted.
+        record_metric_semantics(semantic_index, metric, &name, metric_ctx);
```
sunng87, line 446: Gating `record_metric_semantics` on "produced an accepted row" changes behavior for **all** metric kinds, not just exponential histograms — e.g. a gauge with zero data points no longer stamps semantic metadata where it previously did. Seems like the right call (metadata should follow accepted data) and it's stated in the PR body, just noting it's observable beyond exp-histogram metrics.

src/servers/src/otlp/metrics.rs
```diff
@@ -368,16 +424,391 @@ fn encode_metrics(
...
+        )?;
+        table.add_row(row);
+        add_accepted_data_points(outcome, 1)?;
+        emitted = true;
+    }
+
+    Ok(emitted)
+}
+
+fn exponential_histogram_value(
```
sunng87, line 613: Minor: `exponential_histogram_value` / `convert_bucket_range` return `Result<_, String>`, while the shared module a file over (`native_histogram.rs`) established the typed-error pattern in this same PR. Any reason not to reuse a typed error here? Consistency nit, not blocking.

src/servers/src/otlp/metrics.rs
```diff
@@ -368,16 +424,391 @@ fn encode_metrics(
...
+            Err(reason) => {
+                reject_data_points(
+                    outcome,
+                    1,
+                    &format!("metric `{name}` data point {index}: {reason}"),
+                )?;
+                continue;
+            }
+        };
+        let column_schema = native_histogram_column_schema().map_err(|error| {
```
sunng87: Nit: `native_histogram_column_schema()` is recomputed for every data point inside this loop, but it's a fixed value — could be hoisted above the `for` loop.

src/servers/src/otlp/metrics.rs
```diff
@@ -123,13 +141,46 @@ pub fn to_grpc_insert_requests(
...
+            continue;
+        };
+        let field_count = rows
+            .schema
+            .iter()
+            .filter(|column| column.semantic_type == SemanticType::Field as i32)
+            .count();
+        let has_native_histogram = rows.schema.iter().any(|column| {
+            column.semantic_type == SemanticType::Field as i32
+                && api::helper::is_column_type_value_eq(
```
sunng87, line 167: Nit: fully-qualified `api::helper::is_column_type_value_eq` inline vs. a `use` import like the surrounding code. Fine either way.

sunng87 (commented):
Overall this looks solid — faithful to the updated RFC, well-tested (unit + HTTP + Arrow gRPC integration), and the trickiest math (OTLP→Prometheus bucket-index shift, downscale merging, overflow handling) is correct and covered. I verified the +1 index shift against the OTLP proto semantics (index = lower-bound exponent vs. Prometheus upper-bound), the signed floor division for negative indices, and the pinned otel-arrow schema — it indeed omits `zero_threshold`, so the Arrow rejection is justified.

Inline comments: one coupling concern worth addressing before merge (otel_arrow.rs message sniffing), three behavior-scope flags that deserve a line in the PR description (RW2 `zero_threshold` tightening, `record_metric_semantics` gating, `is_single_value` widening), and a few minor nits.

PR-hygiene nit: the checklist boxes for rustdoc and tests are unchecked although the work appears done.

src/servers/src/otel_arrow.rs
shuiyisong: Fixed in 9bb46a03. The Arrow handler now derives whether the decoded request contains exponential-histogram data points and uses that request fact when producing the Arrow-specific status. It no longer matches OTLP error or config message text, and no extra rejection state was added to MetricsIngestOutcome.

src/servers/src/native_histogram.rs
shuiyisong: Fixed in 9bb46a03. I removed the finite and non-negative zero_threshold check from the shared native-histogram validator, so Remote Write 2.0 keeps its existing behavior; OTLP still validates this at its conversion boundary. I also added an RW2 regression covering negative and NaN thresholds. Prometheus Histogram.Validate and FloatHistogram.Validate do not enforce this constraint for exponential schemas, so keeping it OTLP-specific avoids an unintended RW2 compatibility change.

src/servers/src/otlp/metrics.rs
shuiyisong: Fixed in 8ae0b740. `native_histogram_column_schema()` is now constructed once after the feature and temporality guards, and the prepared schema is reused for each accepted data point instead of repeating the type conversion inside the loop. All six focused exponential-histogram tests pass.

src/operator/src/insert.rs
shuiyisong: Verified the caller matrix. OpenTSDB and non-metric Prom-store pass `(accommodate_existing_schema, is_single_value) = (true, true)` on both the base and this head, so the old and new conditions both run the sample-kind guard for them. The newly covered path is OTLP `(false, true)`. Influx is `(true, false)`, but its line-protocol conversion only emits scalar field types, so a native-histogram/float mismatch is not reachable there. No broader OpenTSDB or Prom-store behavior change is introduced.

src/servers/src/otlp/metrics.rs
shuiyisong: Kept the fully qualified call. The helper is used only once in this file, so adding an import would not reduce repetition or clarify the surrounding validation. No behavioral change is needed.

src/servers/src/otlp/metrics.rs
shuiyisong: Confirmed; this is intentional and already called out in the PR description. An empty-only request returns before any insert, so it cannot persist semantic metadata by itself. The observable difference is limited to preventing an empty metric declaration from influencing the merged metadata of another accepted metric that resolves to the same table.

src/servers/src/otlp/metrics.rs
shuiyisong: Kept `Result<_, String>` intentionally. These helpers are private to the OTLP conversion path, and every error is immediately folded into the per-point partial-success diagnostic; no caller matches on the error type or needs recovery/source classification. Reusing `NativeHistogramError` would couple OTLP-specific validation to the shared encoder without adding behavior.

src/servers/src/otel_arrow.rs
```diff
@@ -42,6 +50,30 @@ impl<T> OtelArrowServiceHandler<T> {
...
+    outcome: MetricsIngestOutcome,
+    has_exponential_histogram_data_points: bool,
+) -> BatchStatus {
+    let status_code = if outcome.accepted_data_points == 0 && outcome.rejected_data_points > 0 {
+        ArrowStatusCode::InvalidArgument
+    } else {
+        ArrowStatusCode::Ok
+    };
+    let status_message = match outcome.error_message {
+        Some(_) if has_exponential_histogram_data_points => {
```
sunng87, line 65: Thanks — the structural check is the right fix and the coupling is gone.

One residual edge, not blocking: when `has_exponential_histogram_data_points` is true, *any* error message is replaced wholesale with `EXPONENTIAL_HISTOGRAM_UNSUPPORTED`. That's always accurate today because the Arrow path hardcodes `OtlpMetricCtx::default()` (gate off, so every exp point is rejected by the gate). But if the gate is ever wired through to Arrow, a point rejected for a different reason (scale, temporality) would be mislabeled as an Arrow-format limitation. A one-line comment capturing that assumption would help future readers.

src/servers/src/otlp/metrics.rs
```diff
@@ -668,9 +1092,14 @@ fn encode_histogram(
         }
 
         if let Some(sum) = data_point.sum {
```
sunng87, line 1094: Minor behavior note from the fragment-merging change: `sum_table` is now only created when at least one data point carries a `sum`. Previously an all-`None`-sums metric still emitted an (empty) `..._sum` insert. Dropping it looks strictly better — just flagging in case any downstream consumer assumed the 3-table fan-out shape unconditionally.

sunng87 (commented):
Reviewed the four new commits (`671bf475`..`b6349e3`) on top of the rebased base, and re-verified the rebase itself was content-faithful (only hunk offsets moved). All previous findings are addressed, and I confirmed each fix in the code:

- **otel_arrow coupling** — replaced with a structural request-derived flag; no more message sniffing. ✓
- **RW2 `zero_threshold` tightening** — resolved by *removing* the shared check rather than documenting it, restoring exact RW2 behavior; OTLP still validates at its conversion boundary (`exponential_histogram_value`), and the new RW2 regression covering `-1.0`/NaN pass-through pins the contract. The Prometheus `Validate` reference makes the case convincingly. ✓
- **schema hoist** — done, cloned per row. ✓
- **lazy rejection messages** — closure is only evaluated when the cap allows; laziness is tested. ✓
- **`fix: merge repeated OTLP histogram fragments`** — this one is a bonus: the old local-`TableData` + `add_table_data` pattern (pre-existing on main) made a second same-name explicit histogram silently *replace* the first one's rows via `HashMap::insert` — real silent data loss across same-name metrics in one request. The `get_or_default_table_data` merge plus the new cross-resource merge test fix that properly.

One correction to my own earlier review: the `is_single_value` widening **does not** change OpenTSDB or Prom-store behavior — both pass `(true, true)`, so the guard already ran under the old `accommodate_existing_schema` condition. The only widened path is OTLP itself, and the narrowed Influx path is unreachable for native histograms. Thanks for the caller matrix.

No remaining blockers from my side. Two minor non-blocking notes inline (Arrow message-replacement assumption, sum-table fan-out shape).

src/servers/src/native_histogram.rs
```diff
@@ -0,0 +1,524 @@
+// Copyright 2023 Greptime Team
```
sunng87, line 1: We can consider putting this module to a lower-level crate. If we have native histogram as a built-in data type, it should be perfect in data type, but we may still find this a new home right now. Our servers module is becoming bloated.

sunng87: We can address this in a dedicated pr.

sunng87 (approved):

src/servers/src/otel_arrow.rs
shuiyisong: Addressed in 1b2a0f7a18. Added an inline comment documenting that the Arrow path keeps the exponential-histogram feature gate disabled, so these points are rejected before per-point validation. If Arrow support is wired later, the assumption is now visible beside the status-message override.

src/servers/src/otlp/metrics.rs
shuiyisong: Confirmed the downstream path. An OTLP `sum: None` previously produced only an empty `<metric>_sum` row-insert request; both physical and metric-engine insert paths remove zero-row requests before table creation, alteration, or storage writes. Its pre-filter effect was an unnecessary `<metric>_sum` permission target. Lazy creation therefore preserves persisted behavior and models the optional sum more accurately, so no further code change is needed.

PR #8903, opened 2026-08-18
feat: add database ingest rate limit hooks

v0y4g3r:

Enterprise-side feature: database-level ingest rate limiting (design doc in the enterprise repo).

This PR adds the OSS-side hooks needed by the enterprise **database-level write rate limiting** feature (`ingest_rows_rate_limit`, a cluster-wide rows/sec quota per database, enforced by enterprise plugins). The enforcement itself lives in the enterprise repository; OSS only gains:

1. **A new database option key** `ingest_rows_rate_limit` added to `VALID_DB_OPT_KEYS` (`src/table/src/requests.rs`), so it can be set via `CREATE DATABASE ... WITH('ingest_rows_rate_limit'='1000')` and updated dynamically via `ALTER DATABASE ... SET/UNSET`. Like other database options, it is persisted in `SchemaNameValue.extra_options` and propagated through the existing schema-cache invalidation broadcast. In pure OSS builds the option is stored but inert (no enforcement), same as other enterprise-consumed options.

2. **An `InsertLimitInterceptor` hook on `Inserter`** (`src/operator/src/insert.rs`):
   - `Inserter::do_request` is the single choke point for nearly all write paths (SQL INSERT, gRPC row/column inserts, InfluxDB, OTLP, Prom remote write, OpenTSDB, logs, traces). The interceptor is invoked there with the total row count before dispatch to datanodes.
   - `Inserter::handle_bulk_insert` (Flight DoPut) dispatches directly and bypasses `do_request`, so it gets its own interception call.
   - The charge target is derived from the **target table's** `catalog_name`/`schema_name` (falling back to the session context), so fully-qualified `INSERT INTO other_db.t`, DataFusion DML, and `COPY FROM` cannot bypass a limit by writing cross-database from an unlimited session.
   - Interceptor errors propagate through `Error::External`, so an enterprise-side `StatusCode::RateLimited` maps cleanly to gRPC `ResourceExhausted` / HTTP 429 / MySQL / PostgreSQL error codes.
   - The interceptor defaults to `None` (zero overhead in OSS) and is registered via `Plugins`; the frontend builder wires it with `plugins.get::<InsertLimitInterceptorRef>()`. The `Inserter::new` signature is unchanged, so the flownode call site is unaffected (flow sink writes are intentionally out of scope).

Notes / limitations:
- Instant-TTL rows (mirrored to flownodes only) are counted, as they represent real frontend ingest load.
- Auto-create/alter happens before the check, so a rejected first write may still have created/altered the table.
- Deletes are out of scope (they never pass through `Inserter`).

## PR Checklist

### DIFF
--- a/src/operator/src/insert.rs
+++ b/src/operator/src/insert.rs
@@ -33,6 +33,7 @@ use common_catalog::consts::{
     TRACE_TABLE_NAME_SESSION_KEY, default_engine, is_ddl_reserved_table,
     trace_operations_table_name, trace_services_table_name,
 };
+use common_error::ext::BoxedError;
 use common_grpc_expr::util::ColumnExpr;
 use common_meta::cache::TableFlownodeSetCacheRef;
 use common_meta::node_manager::{AffectedRows, NodeManagerRef};
@@ -82,6 +83,23 @@ use crate::req_convert::insert::{
 };
 use crate::statement::StatementExecutor;
 
+/// Intercepts row inserts right before they are dispatched to datanodes,
+/// e.g. to enforce database-level ingest rate limits.
+/// Registered via `Plugins`; `None` means no interception (OSS default).
+#[async_trait::async_trait]
+pub trait InsertLimitInterceptor: Send + Sync {
+    /// Checks whether `rows` rows may be ingested into `catalog.schema`.
+    /// Returning `Err` rejects the whole request.
+    async fn check_ingest(
+        &self,
+        catalog: &str,
+        schema: &str,
+        rows: u64,
+    ) -> std::result::Result<(), BoxedError>;
+}
+
+pub type InsertLimitInterceptorRef = Arc<dyn InsertLimitInterceptor>;
+
 pub struct Inserter {
     catalog_manager: CatalogManagerRef,
     pub(crate) partition_manager: PartitionRuleManagerRef,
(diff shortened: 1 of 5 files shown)


src/operator/src/insert.rs
```diff
@@ -82,6 +83,24 @@ use crate::req_convert::insert::{
...
+pub trait InsertLimitInterceptor: Send + Sync {
+    /// Checks whether `rows` rows may be ingested into `catalog.schema`.
+    /// Returning `Err` rejects the whole request.
+    async fn check_ingest(
+        &self,
+        catalog: &str,
+        schema: &str,
+        rows: u64,
+    ) -> std::result::Result<(), BoxedError>;
+}
```
sunng87, line 99: we should not simply count by rows because rows may vary on size. 
this should already be measured with meter module as wcu, we should reuse the wcu concept.

shuiyisong:
We can move the calculation part to the existing meter system, the rest can stay unchanges.
Some implementation migration proposal: https://gist.github.com/shuiyisong/a85367e9d01930f0fcf4bbca8b22c5d8

shuiyisong:
Done in https://github.com/GreptimeTeam/greptimedb/pull/9239, should we close this PR?

PR #8904, opened 2026-08-18
feat: synthesize OTLP resource descriptor for the semantic entity graph

killme2008:

Follows up #8880.

Metrics-only services never show up in the semantic entity graph. OTLP resource attributes get filtered by the promote list and scattered as tags across hundreds of logical metric tables, so entities only ever come from trace tables. On an Astronomy Shop run: 380 metric tables, zero entities from any of them.

So each OTLP metrics request now also writes one small table, `greptime_otel_resource_info`, holding the identity-relevant resource attributes under their raw OTel keys plus derived `job`/`instance` columns, one row per resource per 60s graph window. A new conventions entry declares `service`, `service.instance`, `host`, `container` and `k8s.pod` from it, so a single row derives `part_of` and the `service.instance -> container -> host` chain.

**Off by default** behind `otlp.experimental_enable_resource_info` — it creates and writes a table the user never sent. With it off a request costs what it did before: nothing projected, no table, no extra write.

Along the way:

- `job` now folds in `service.namespace` (`<namespace>/<name>`) per the OTel Prometheus compatibility spec. **Behavior change**, not gated: without it the descriptor and `target_info` can never agree on service identity.
- `host`/`container` identities are the stable ids only (`host.id`, `container.id`); names are descriptive, since `host.name` was `orbstack` in one place and a container hostname in another. Trace tables get host/container too, and no `service.instance` is invented when `service.instance.id` is missing.
- Composite entity ids no longer encode their column names. A trace table names its columns `service_name` and `resource_attributes.service.instance.id` where a metric table names them `job` and `instance`, so one instance became two nodes with two parallel edge sets. Ids are now the values in declared order, escaped and joined; `entity_id_attrs` still carries the structured form.
- `StandaloneOptions` had no `otlp` field, so the whole `[otlp]` section was silently ignored in standalone mode (`trace_ingest_chunk_size` included).

The descriptor is derived data: it is written after the main insert commits and never fails the request. A conflicting table, `auto_create_table=false` or a permission denial degrades to an OTLP `partial_success` warning with `rejected_data_points = 0`.

Limitations: descriptor-derived k8s coverage stops at `k8s.pod`; `instance` is omitted rather than written empty when `service.instance.id` is absent, which PromQL treats identically and keeps a permanently empty column out of the primary key.

## PR Checklist


killme2008:
The final piece of Semantic Graph stage 1—please review it. @shuiyisong @sunng87

sunng87 (commented):
Follow-up nits and one spec-level concern from my review (https://github.com/GreptimeTeam/greptimedb/pull/8904). Everything else checked out: gating, layering, degrade path, and the cross-signal identity tests all verified locally.

src/operator/src/statement/semantic_graph.rs
```diff
@@ -417,30 +423,88 @@ fn cast_string_or_empty(column: &str) -> Expr {
...
+    let qualifier = escaped_id_value(
+        core_fns::coalesce().call(vec![cast(col(qualifier), DataType::Utf8), lit("")]),
+    );
+    when(qualifier.clone().eq(lit("")), value.clone())
+        .otherwise(concat_expr(vec![
+            qualifier,
+            lit(ID_QUALIFIER_SEPARATOR),
+            value,
+        ]))
+        .expect("CASE with a single WHEN and an ELSE is well-formed")
```
sunng87: This is the only `unwrap`/`expect` family call added on a non-test path in the PR. The repo invariant (`.agents/architecture-invariants.md` §4) is "return errors instead of `unwrap()` / `expect()` / `panic!()` in non-test code", and these get flagged on sight even when unreachable.

It does assert a static property — a `when().otherwise()` with an ELSE can't fail — so no runtime input can trip it. Still, would you consider `unwrap_or_else(|e| ...)` with a debug log, or a one-line comment tying it to the invariant so future readers don't re-derive why it's safe? Judgement call, not a blocker.

src/servers/src/otlp/metrics.rs
```diff
@@ -232,30 +290,54 @@ fn from_metric_type(data: &metric::Data) -> MetricType {
...
-            KEY_SERVICE_INSTANCE_ID => {
-                tmp.push(KeyValue {
-                    key: INSTANCE_KEY.to_string(),
-                    value: kv.value.clone(),
-                });
-            }
+/// Prometheus-style `(job, instance)` identity. Per the OTel Prometheus
+/// compatibility spec `job` folds in `service.namespace` when present; a
+/// resource without `service.name` gets no job rather than a fabricated one.
+pub(crate) fn service_identity(attrs: &[KeyValue]) -> (Option<String>, Option<String>) {
```
sunng87: Two call sites destructure this as `let (job, instance) = ...` (`process_resource_attrs` here and `ResourceInfoData::observe`). Both fields are `Option<String>`, so a swapped `(instance, job)` would type-check and only surface as wrong emitted columns — and this is now identity data feeding the entity graph.

A tiny `struct ServiceIdentity { job, instance }` (or named accessors) would make the destructuring sites self-documenting. Cosmetic.

src/servers/src/otlp/metrics/resource_info.rs
```diff
@@ -0,0 +1,350 @@
...
+/// the newest data-point time seen in that window, which is what the row for
+/// that window is stamped with.
+///
+/// Windows are an inner map so the attributes are stored, and moved, once per
+/// resource. They are keyed separately because one request may carry data for
+/// several of them: a single row per resource would describe only the newest
+/// window, leaving the earlier ones with metric rows but no entities.
+#[derive(Debug, Default)]
+pub struct ResourceInfoData {
+    rows: BTreeMap<Vec<(String, String)>, BTreeMap<i64, i64>>,
```
sunng87, line 82: The double-nested map takes a moment to decode (attribute-set → window → newest ts). The field docs carry it and there are only two touch points, so this is minor — but a small named type (e.g. `struct ResourceRow { tags: Vec<(String, String)>, windows: BTreeMap<i64, i64> }`) or a per-level type alias would let `observe` / `into_row_insert_requests` read without the comment. Optional polish.

src/operator/src/statement/semantic_graph.rs
```diff
@@ -417,30 +423,88 @@ fn cast_string_or_empty(column: &str) -> Expr {
...
+///
+/// The identifying *column names* are deliberately absent: the same identity
+/// reaches us under different names per source — a trace table's
+/// `service_name` against a metric table's `job` — and encoding them would
+/// split one entity into one per signal. `entity_id_attrs` keeps the
+/// structured form.
+///
+/// `col` constructs the column reference (unqualified for registry branches,
+/// join-side-qualified for the calls derivation).
+fn entity_id_expr(
```
sunng87, line 469: With ids now rendered in declared order (instead of the previous sorted form), column order is semantically significant — but nothing cross-validates it. The new conventions validation catches arity drift (`[job]` vs `[job, instance]`), yet two sources declaring the same entity type as `[job, instance]` vs `[instance, job]` render `shop/api,inst-1` vs `inst-1,shop/api` and silently split the entity, which is exactly the failure mode this PR fixes across signals. The shipped conventions.yaml is consistent today, so this is forward-looking, but two things seem worth doing:

1. Since "broad to narrow" is already the documented invariant, validation could pin a canonical column order per entity type that both whitelists must match (the arity map in `conventions.rs::validate` is a natural place).
2. For **user-declared** tables this is a quiet behavior regression: the old rendering sorted `id_columns` before composing the id, so two tables declaring the same entity with columns in different orders matched; after this change they no longer do, with no warning. At minimum the `greptime.semantic.entity.*` option docs should state that id-column order is part of identity.

src/operator/src/statement/semantic_graph.rs
killme2008: Gone — the CASE is built through `Case::new` directly instead of the `when().otherwise()` builder, which is where the `Result` (and the `expect`) came from. Left a line saying why, so nobody switches it back to the builder.

https://github.com/GreptimeTeam/greptimedb/pull/8904/changes/e1c9816178343c58812f5f0c8fea90324ed3ad15

src/servers/src/otlp/metrics.rs
killme2008: Done, `ServiceIdentity { job, instance }`. A swap fails to compile now.

https://github.com/GreptimeTeam/greptimedb/pull/8904/changes/e1c9816178343c58812f5f0c8fea90324ed3ad15

src/servers/src/otlp/metrics/resource_info.rs
killme2008: Leaving this one. With both touch points sitting right next to the field docs, an alias or wrapper moves the explanation rather than removes the need for it. Say the word if you disagree and I will add it.

src/operator/src/statement/semantic_graph.rs
killme2008: Point 2 is a fair catch and I had missed that it was a silent change: the old rendering sorted the columns, so two tables declaring the same entity with the same column names in different orders used to match, and now they do not. Nothing is released so no one is affected, but it is a footgun. Documented on `EntityRole::Id` (where users writing the option land) and at the top of `conventions.yaml` (where the next editor of that file lands).

Point 1 I do not think is reachable yet. The id column names differ per source by design — `job` here, `service_name` there — so there is nothing to compare orders against. Checking it needs the identity expressed as semantic slots that each source maps its columns onto, which is a conventions-format change rather than an extra validation. That is RFC Open Question 2; happy to file an issue if you want it tracked rather than sitting in the RFC.

https://github.com/GreptimeTeam/greptimedb/pull/8904/changes/e1c9816178343c58812f5f0c8fea90324ed3ad15

src/servers/src/otlp/metrics/resource_info.rs
```diff
@@ -0,0 +1,363 @@
...
+                Some(metric::Data::Sum(s)) => {
+                    visit_all(&mut s.data_points.iter().map(|p| p.time_unix_nano))
+                }
+                Some(metric::Data::Histogram(h)) => {
+                    visit_all(&mut h.data_points.iter().map(|p| p.time_unix_nano))
+                }
+                Some(metric::Data::Summary(s)) => {
+                    visit_all(&mut s.data_points.iter().map(|p| p.time_unix_nano))
+                }
+                Some(metric::Data::ExponentialHistogram(h)) if exponential_histograms => {
```
sunng87: The `exponential_histograms` flag mirrors the experimental gate, but `encode_exponential_histogram` has two more accept conditions this visitor doesn't see:

1. **Temporality** — the encoder rejects the whole metric on delta or unspecified temporality (`reject_data_points` + `return Ok(false)`), but this arm visits those times whenever the flag is on.
2. **Per-point validation** — points failing `exponential_histogram_value` (`start_time_unix_nano > time_unix_nano`, i64-overflowing timestamps) are dropped by the encode loop but still counted here.

Consequence: with both experimental flags on, a resource whose only metric is a delta exponential histogram still gets descriptor rows — an entity with no measurements, exactly what `observe_ignores_data_the_encoder_drops` exists to prevent (it currently pins only the flag-off case; there's no delta-temporality coverage). Severity is contained — it needs both flags on and only pollutes derived enrichment, not user data or stability.

This is the second hand-sync this mirror has needed (the flag was threaded through during the merge; the temporality gate didn't come with it), which suggests the structure rather than another manual sync. The dispatch in `encode_metrics` and this visitor are two copies of "which OTLP metrics does GreptimeDB accept"; one shared decision would make drift impossible:

- Extract `classify_metric(metric, ctx) -> Accepted(shape ref) | Rejected { points, reason } | None` in `metrics.rs`, owning the flag gate, the temporality gate, and the rejection message strings.
- `encode_metrics` matches on `Accepted` to pick its encode fn and reports `Rejected` — the match is non-exhaustive over the shapes, so adding a shape or a gate fails to compile until the encoder handles it.
- `for_each_encoded_time` asks the same `classify_metric` and walks an `visit_written_times()` on the accepted shape, filtering exponential points through `exponential_histogram_value(..).is_ok()` — the per-point rule stays next to its encode loop.

sunng87 (commented):
Found while re-reviewing the #8900 merge adaptation — this arm now mirrors only one of the encoder's three accept conditions for the shape, and I think it's worth fixing structurally rather than syncing by hand again.

src/frontend/src/instance/otlp.rs
```diff
@@ -165,12 +171,43 @@ impl OpenTelemetryProtocolHandler for Instance {
...
+        // failing here would make the client retry data the server already
+        // accepted, so every failure degrades to a warning instead.
+        if let Some(resource_info) = resource_info {
+            let written = match self.check_row_insert_permission(
+                &resource_info,
+                &ctx,
+                PermissionReq::Action(OTLP_WRITE),
+            ) {
+                Ok(_) => self
+                    .handle_row_inserts(resource_info, ctx, false, false)
```
shuiyisong, line 191: **[P2] Reject persisted descriptor-name collisions before this write**

The per-request guard in `to_grpc_insert_requests` does not protect a collision created by an earlier request. I reproduced this on this head with resource-info enabled:

1. Export a gauge named `greptime_otel_resource_info`; it succeeds and creates a metric-engine logical table stamped `metric.type=gauge`.
2. Export an ordinary gauge in a later request; this generic insert path finds the existing table, auto-adds the descriptor columns, and writes the synthesized row.

Both requests return full success. The colliding table then contains the user's `7.0` sample plus the synthetic descriptor `1.0` sample, while it remains `metric.type=gauge`. The entity graph therefore excludes it because the OTel descriptor convention requires `metric.type=info`: user data is polluted and descriptor-derived entities silently disappear.

Semantic options are applied only when a table is created, so a compatible existing table never acquires descriptor ownership. Please restore an engine/semantic ownership check before the derived write (or reserve/create this table canonically); a persisted collision should take the documented partial-success path instead of succeeding.

src/servers/src/otlp/metrics/resource_info.rs
killme2008: Fixed. Both whole-metric rules — the experimental gate and cumulative-only temporality — now live in `exponential_histogram_gate`, which the encoder and the descriptor call; the encoder formats the rejection message from the returned variant, so the three message strings are unchanged. The descriptor takes its timestamps through `exponential_histogram_value` as well, so points the encode loop drops no longer contribute.

I kept it to this shape rather than `classify_metric`. Exponential histograms are the only shape with accept conditions — gauge/sum/summary/histogram take every point — and both matches are exhaustive over `metric::Data` already, so a new shape breaks the build either way. What actually drifted was the gates, and those are now in one function.

`observe_ignores_data_the_encoder_drops` covers the delta case with the flag on.

https://github.com/GreptimeTeam/greptimedb/pull/8904/commits/703a635761

src/frontend/src/instance/otlp.rs
killme2008: Thanks for the repro — the analysis is right, and this was a deliberate removal, not an oversight. The ownership check existed earlier in the branch and went away when the table was renamed to `greptime_otel_resource_info`.

Not restoring it. `greptime_` marks a table as engine-managed across this codebase; a user metric claiming that exact name is not a case we are willing to spend a per-request catalog lookup on, on the write path, for every OTLP metrics request with the feature enabled. The per-request guard stays because it is free — the table names are already in hand.

If it does happen, the failure is loud enough to diagnose: the descriptor rows land in a table that `information_schema.tables` shows as `metric.type=gauge`, and the entity graph reports nothing for it. Renaming the offending metric is the fix.

Worth revisiting if we ever reserve the `greptime_` metric namespace at ingest — then a colliding user metric gets rejected at the front door and the descriptor needs no guard at all.

sunng87 (approved):

PR #8909, opened 2026-08-19
refactor(json2): add bounded auto-expansion to the JSON2 vector builder

MichaelScofield:

as title

## PR Checklist

### DIFF
--- a/src/datatypes/src/vectors/json/builder.rs
+++ b/src/datatypes/src/vectors/json/builder.rs
@@ -39,6 +39,11 @@ use crate::vectors::{Helper, MutableVector, StructVectorBuilder};
 type JsonObjectValue = BTreeMap<String, JsonVariant>;
 
 /// Builds JSON2 vectors from object values.
+///
+/// Legacy mode merges all observed paths into the explicit Struct schema.
+/// Auto-expanding mode always materializes type-hinted paths, selects up to
+/// `max_auto_expanded_paths` compatible unhinted leaf paths by frequency, and stores
+/// conflicting or unselected paths in the Variant remainder field.
 #[derive(Clone)]
 pub(crate) struct JsonVectorBuilder {
     state: JsonVectorBuilderState,
@@ -469,10 +474,8 @@ fn split_to_explicit(
                 explicit.insert(name.clone(), value);
                 continue;
             }
-            let (value, child_remainder) = split_to_explicit(value, data_type)?;
-            if !value.is_empty() {
-                explicit.insert(name.clone(), JsonVariant::Object(value));
-            }
+            let (child_explicit, child_remainder) = split_to_explicit(value, data_type)?;
+            explicit.insert(name.clone(), JsonVariant::Object(child_explicit));
             if !child_remainder.is_empty() {
                 remainder.insert(name.clone(), JsonVariant::Object(child_remainder));
             }
@@ -522,6 +525,9 @@ fn json_variant_into_struct_value(
 fn json_variant_into_value(value: JsonVariant, expected_type: &ConcreteDataType) -> Result<Value> {
     let value = match (value, expected_type) {
         (JsonVariant::Null, _) | (_, ConcreteDataType::Null(_)) => Value::Null,
+        (JsonVariant::Object(object), ConcreteDataType::Struct(struct_type)) => {
+            Value::Struct(json_variant_into_struct_value(object, struct_type.clone())?)
+        }
         (JsonVariant::Object(object), _) if object.is_empty() => Value::Null,
         (JsonVariant::Bool(x), ConcreteDataType::Boolean(_)) => Value::Boolean(x),
         (JsonVariant::Number(x), ConcreteDataType::UInt64(_)) => {
@@ -564,9 +570,6 @@ fn json_variant_into_value(value: JsonVariant, expected_type: &ConcreteDataType)
                 .map(|v| json_variant_into_value(v, &item_type))
                 .collect::<Result<Vec<_>>>()?;
             Value::List(ListValue::new(values, Arc::new(item_type)))
-        }
-        (JsonVariant::Object(value), ConcreteDataType::Struct(struct_type)) => {
-            Value::Struct(json_variant_into_struct_value(value, struct_type.clone())?)
         }
         (value, ConcreteDataType::Binary(_)) => Value::from(encode_json_variant(value)?),
         (value, expected_type) => {
@@ -932,6 +935,40 @@ mod tests {
                 "tie_b": null
             }),
             decode_json_variant(reconstructed.value(2)).unwrap()
+        );
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_reconstruct_nested_remainder_only_value() -> Result<()> {
+        let settings = JsonSettings::try_new(vec![], Some(1))?;
+        let values = [
+            json!({"a": {"hot": 1}}),
+            json!({"a": {"hot": 2}}),
+            json!({"a": {"cold": 3}}),
+        ];
+        let mut builder = JsonVectorBuilder::with_settings(&settings, values.len());
+        for value in values.clone() {
+            let value = settings.encode(value)?;
+            builder.try_push_value_ref(&value.as_value_ref())?;
+        }
+
+        let array = builder.to_vector().to_arrow_array();
+        let field = Field::new("data", array.data_type().clone(), true).with_extension_type(
+            Json2ExtensionType::new(Arc::new(JsonMetadata::new_v2(settings))),
+        );
+        let reconstructed = JsonArray::from(&array).project_to_v2(&field, &DataType::Binary)?;
+        let reconstructed = reconstructed.as_binary::<i32>();
+        assert_eq!(
+            vec![
+                values[0].clone(),
+                values[1].clone(),
+                json!({"a": {"cold": 3, "hot": null}}),
+            ],
+            (0..reconstructed.len())
+                .map(|i| decode_json_variant(reconstructed.value(i)).unwrap())
+                .collect::<Vec<_>>()
         );
 
         Ok(())
--- a/src/datatypes/src/vectors/json/builder.rs
+++ b/src/datatypes/src/vectors/json/builder.rs
@@ -178,7 +178,7 @@ impl JsonVectorBuilder {
 }
 
 fn build_legacy(values: &mut Vec<JsonVariant>, merged_type: &JsonNativeType) -> Result<VectorRef> {
-    build_explicit(values, merged_type, |value| match value {
+    build_explicit(values, merged_type, false, |value| match value {
         JsonVariant::Null => Ok(None),
         JsonVariant::Object(value) => Ok(Some(value)),
         _ => TryFromValueSnafu {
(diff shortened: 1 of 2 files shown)


MichaelScofield:
@codex review

fengys1996:
@codex review

src/datatypes/src/vectors/json/builder.rs
```diff
@@ -13,72 +13,480 @@
...
 use crate::types::StructType;
 use crate::types::json_type::{JsonNativeType, is_include};
 use crate::value::{ListValue, StructValue, StructValueRef, Value};
-use crate::vectors::{MutableVector, StructVectorBuilder};
+use crate::vectors::json::variant::{append_json_variant, variant_field};
+use crate::vectors::{Helper, MutableVector, StructVectorBuilder};
 
+type JsonObjectValue = BTreeMap<String, JsonVariant>;
+
+/// Builds JSON2 vectors from object values.
```
fengys1996, line 41: We should add more detailed documentation about this behavior.

sunng87:
**Cross-batch layout divergence can make rows unreadable at read time**

The auto-expanding layout is chosen per builder instance — i.e. per record batch. I traced what happens when two batches over the same column pick different expanded sets: the rows of one of them become **unreadable**, while writes succeed silently.

**1. The expanded set is a pure function of the current batch.** `with_settings` (builder.rs:161) receives only `JsonSettings` (type hints + budget). `infer_expanded_type` (builder.rs:259) ranks paths by frequency within the buffered values of *this* builder. Nothing seeds it with the layout the column already has on disk / in previous batches.

**2. Two batches can legitimately diverge.** Two easy triggers:
- *Conflict appears mid-stream*: batch 1 has `{"a": 1}, {"a": 2}` → `a` is stable → expanded as `a: Int64`. Batch 2 contains `{"a": 3}, {"a": "str"}` → `a` conflicts → loses expansion eligibility entirely → no `a` field; every `a` value (including the int) goes to the remainder.
- *Frequency drift*: with a finite budget, a borderline path can be in the top-N for one batch and out of it for the next — no type conflict needed at all.

**3. Reconciliation fills typed nulls.** At flush / part merge, `Json2Aligner` merges the concrete types (`merged.merge`, `src/mito2/src/memtable/bulk/json_align.rs:91`) — the merged schema keeps `a: Int64` from batch 1 — and aligns batch 2 via `widen_to` (`json_align.rs:139`), whose contract is "fields missing from the source are filled with typed null arrays" (`array.rs:165`). After alignment, batch 2's rows have an explicit `a = null` while the actual `a` values still sit in each row's remainder blob. (The same reconciliation happens in the compaction reader, `compaction/reader.rs:109-115`.)

**4. Read fails on the whole row.** `project_to_v2` → `json2_values` (`array.rs:118`) merges explicit fields into the remainder via `merge_explicit_value` (`array.rs:395`), which only accepts the both-objects case. `null` (explicit) vs scalar (remainder) → `"cannot merge 'a' in explicit fields and remainder: not both objects"` (`array.rs:414`) — the **entire row** read fails, not just the `a` field.

What makes this worse than a normal bug:
- **Silent at write time**: each batch is individually self-consistent; inserts, flush, and alignment all succeed. The failure only surfaces at read time.
- **Data- and batching-dependent**: whether it triggers depends on how rows are batched and on frequency ordering across batches — hard to repro, intermittent in production.
- **Row-granular failure**: one diverged path poisons the whole row for reads.

I wrote two regression tests (below). Both **currently fail** with the error above; they assert only that rows round-trip after flush-time alignment, which is independent of which selection policy is preferred — the gap exists for *any* policy where the expanded set can differ between two batches of the same column.

```rust
/// Merges the concrete JSON2 types of several built batches the way
/// `Json2Aligner` does at flush time, and restores the canonical remainder
/// field so the result can be used as a column schema for reads.
fn merged_json2_type(arrays: &[ArrayRef]) -> Result<DataType> {
    let mut merged = JsonNativeType::try_from(arrays[0].data_type())?;
    for array in &arrays[1..] {
        merged.merge(&JsonNativeType::try_from(array.data_type())?);
    }
    let DataType::Struct(fields) = merged.as_arrow_type() else {
        unreachable!("merged JSON2 layout must be a struct");
    };
    let fields = fields
        .iter()
        .map(|field| {
            if field.name() == JSON2_REMAINDER_FIELD_NAME {
                Arc::new(variant_field(JSON2_REMAINDER_FIELD_NAME, true))
            } else {
                field.clone()
            }
        })
        .collect::<Vec<_>>();
    Ok(DataType::Struct(Fields::from(fields)))
}

/// Decodes a physical JSON2 array back into logical JSON values through the
/// v2 read path, as readers do.
fn decode_round_trip(array: &ArrayRef, field: &Field) -> Result<Vec<serde_json::Value>> {
    let projected = JsonArray::from(array).project_to_v2(field, &DataType::Binary)?;
    let binary = projected.as_binary::<i32>();
    Ok((0..binary.len())
        .map(|i| decode_json_variant(binary.value(i)).unwrap())
        .collect())
}

/// Two batches of the same column may pick different expanded layouts
/// after a mid-stream type conflict, and rows from the diverged batch
/// must still round-trip after flush-time alignment.
///
/// Batch 1 sees a stable path and expands it into a typed field; batch 2
/// sees conflicting types on the same path, which loses expansion
/// eligibility entirely, so the path has no field and all its values
/// (including well-typed ones) land in the remainder. Flush-time alignment
/// (`Json2Aligner` -> `widen_to`) keeps the merged schema's field and fills
/// it with typed nulls for batch 2. Reading the aligned batch back must
/// still return the original values: the remainder's real value has to win
/// over the null-padded explicit field.
///
/// This test currently FAILS: `merge_explicit_value` rejects the
/// explicit-null-vs-scalar merge with "not both objects", which makes the
/// WHOLE batch unreadable even though writes and alignment succeeded.
#[test]
fn test_cross_batch_conflict_divergence_breaks_reads() -> Result<()> {
    let settings = JsonSettings::try_new(vec![], Some(1))?;

    // Batch 1: "a" is stable across rows, so it takes the single slot.
    let mut builder = JsonVectorBuilder::with_settings(&settings, 2);
    for value in [json!({"a": 1}), json!({"a": 2})] {
        let value = settings.encode(value)?;
        builder.try_push_value_ref(&value.as_value_ref())?;
    }
    let batch1 = builder.to_vector().to_arrow_array();
    assert!(batch1.as_struct().column_by_name("a").is_some());

    // Batch 2: "a" conflicts, so it is not expanded at all.
    let mut builder = JsonVectorBuilder::with_settings(&settings, 2);
    for value in [json!({"a": 3}), json!({"a": "conflict"})] {
        let value = settings.encode(value)?;
        builder.try_push_value_ref(&value.as_value_ref())?;
    }
    let batch2 = builder.to_vector().to_arrow_array();
    assert!(batch2.as_struct().column_by_name("a").is_none());

    // Flush-time alignment: the merged schema keeps "a" (from batch 1),
    // and widen_to fills batch 2's missing "a" with typed nulls while its
    // real values stay in the remainder blob.
    let merged = merged_json2_type(&[batch1.clone(), batch2.clone()])?;
    let aligned1 = JsonArray::from(&batch1).widen_to(&merged)?;
    let aligned2 = JsonArray::from(&batch2).widen_to(&merged)?;

    let field = Field::new("data", merged, true).with_extension_type(Json2ExtensionType::new(
        Arc::new(JsonMetadata::new_v2(settings)),
    ));
    assert_eq!(
        vec![json!({"a": 1}), json!({"a": 2})],
        decode_round_trip(&aligned1, &field)?
    );
    assert_eq!(
        vec![json!({"a": 3}), json!({"a": "conflict"})],
        decode_round_trip(&aligned2, &field)?
    );

    Ok(())
}

/// Same requirement without any type conflict: with a finite budget,
/// frequency drift alone flips which paths are expanded across batches,
/// and rows from both batches must round-trip after alignment.
///
/// Batch 1 expands the more frequent "b" and keeps "a" in the remainder;
/// batch 2 only contains "a", so it expands "a". The merged schema keeps
/// both fields, and aligning batch 1 leaves its remainder value `{"a": 1}`
/// facing a null-padded explicit "a" — the remainder value must win.
///
/// This test currently FAILS with the same "not both objects" error.
#[test]
fn test_cross_batch_budget_drift_divergence_breaks_reads() -> Result<()> {
    let settings = JsonSettings::try_new(vec![], Some(1))?;

    // Batch 1: "b" is seen twice, "a" once -> only "b" is expanded.
    let mut builder = JsonVectorBuilder::with_settings(&settings, 2);
    for value in [json!({"a": 1, "b": 2}), json!({"b": 3})] {
        let value = settings.encode(value)?;
        builder.try_push_value_ref(&value.as_value_ref())?;
    }
    let batch1 = builder.to_vector().to_arrow_array();
    let structs = batch1.as_struct();
    assert!(structs.column_by_name("b").is_some());
    assert!(structs.column_by_name("a").is_none());

    // Batch 2: only "a" occurs -> "a" is expanded this time.
    let mut builder = JsonVectorBuilder::with_settings(&settings, 2);
    for value in [json!({"a": 4}), json!({"a": 5})] {
        let value = settings.encode(value)?;
        builder.try_push_value_ref(&value.as_value_ref())?;
    }
    let batch2 = builder.to_vector().to_arrow_array();
    assert!(batch2.as_struct().column_by_name("a").is_some());

    let merged = merged_json2_type(&[batch1.clone(), batch2.clone()])?;
    let aligned1 = JsonArray::from(&batch1).widen_to(&merged)?;
    let aligned2 = JsonArray::from(&batch2).widen_to(&merged)?;

    let field = Field::new("data", merged, true).with_extension_type(Json2ExtensionType::new(
        Arc::new(JsonMetadata::new_v2(settings)),
    ));
    assert_eq!(
        vec![json!({"a": 1, "b": 2}), json!({"a": null, "b": 3})],
        decode_round_trip(&aligned1, &field)?
    );
    assert_eq!(
        vec![json!({"a": 4, "b": null}), json!({"a": 5, "b": null})],
        decode_round_trip(&aligned2, &field)?
    );

    Ok(())
}
```

The only import change needed in the test module is `use arrow_schema::{Field, Fields};`. Expected-vs-actual for the failing tests:

```
Error: Invalid JSON: cannot merge 'a' in explicit fields and remainder: not both objects
at src/datatypes/src/vectors/json/array.rs:414:10
```

(The second test matters as much as the first: it shows **frequency drift alone** — with zero type conflicts anywhere — is enough to trigger the failure, so this isn't just about the conflict policy.)

Note the bounded path isn't wired into `create_mutable_vector` yet (builder.rs:160 still `#[expect(dead_code)]`), so today it's latent — but it will materialize as soon as it's wired.

MichaelScofield:
@sunng87 `widen_to`, or rather, the current flush/compaction method of json2, will not be used in the final v2 storage layout impl. the merging of json2 batches will be guided by a fixed-target rewrite.

sunng87:
@MichaelScofield does it mean the `merged_json2_type` strategy from test case is no longer valid in upcoming changes? if so, it should be fine.

MichaelScofield:
@sunng87 json types will be merged in a slightly different way and the json batches, too

sunng87 (approved):

PR #8910, opened 2026-08-19
fix(meta): avoid blocking runtime on stats cache lock

WenyXu:

None.

This PR fixes a potential executor-level deadlock in heartbeat statistics collection. `CollectStatsHandler` previously held a synchronous `DashMap` entry guard while awaiting node-address and statistics KV writes. On a single-worker runtime, another heartbeat targeting the same map shard could synchronously block the only worker, preventing the lock holder from being polled again.

The statistics cache now uses `DashMap<DatanodeStatKey, Arc<tokio::sync::Mutex<EpochStats>>>`. The `DashMap` guard is held only long enough to obtain the per-Datanode state handle. The async mutex then serializes epoch handling, node-address rewrites, and statistics flushes for the same Datanode without blocking a runtime worker. Different Datanodes retain independent state locks.

The tests cover current-thread runtime liveness using an external wall-clock timeout and verify that concurrent flushes persist every statistic exactly once while allowing at most one in-flight put for the same Datanode.

This change does not modify public APIs, persisted schemas, or data formats.

Validation performed:

- `cargo fmt --check`
- `cargo test -p meta-srv collect_stats_handler` (3 passed)
- `cargo clippy -p meta-srv --tests -- -D warnings`

## PR Checklist

### DIFF
--- a/src/meta-srv/src/handler/collect_stats_handler.rs
+++ b/src/meta-srv/src/handler/collect_stats_handler.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use std::cmp::Ordering;
+use std::sync::Arc;
 
 use api::v1::meta::{HeartbeatRequest, Role};
 use common_meta::datanode::{DatanodeStatKey, DatanodeStatValue, Stat};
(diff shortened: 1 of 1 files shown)


WenyXu:
@codex review

fengjiachun (approved):
LGTM

PR #8911, opened 2026-08-19
fix(query): keep INSERT timestamp conversion out of the source query

killme2008:

Follow-up to #8889.

#8889 made `INSERT` interpret string timestamps with the session timezone by pushing the conversion **down into the source query**. Rewriting the source query changes what it means, and it produced two bugs:

**Rows silently dropped.** Pushing through a `UNION`'s `DISTINCT` moves the dedup key from the raw strings to parsed instants, so two rows spelling the same instant differently collapse into one:

```sql
CREATE TABLE t (ts TIMESTAMP TIME INDEX) WITH ('append_mode'='true');
-- the source query returns 2 rows, the insert stored 1
INSERT INTO t (ts) SELECT '2026-08-06 04:00:00' UNION SELECT '2026-08-06 04:00:00.000';
```

**One branch cancelled the whole column.** A `UNION` branch needing no conversion (a `NULL`, or an explicit cast) made the entire column give up, so sibling branches stayed on UTC while the rest of the row used the session timezone — the same row ended up 8 hours apart:

```sql
INSERT INTO t (ts, v)
SELECT '2026-08-07 12:00:00.001' AS a, '2026-08-07 12:00:00.001' AS b
UNION ALL SELECT '2026-08-07 13:00:00.001', NULL;
```

### How it works now

Convert at the assignment instead of in the source query. The assignment's cast is routed through a timezone-carrying timestamp type and back: arrow applies the timezone when a cast target carries one, and stripping it afterwards only drops metadata. The source query is never touched, so both bugs disappear rather than being patched, and the whole tree-walking rewrite (~160 lines across five functions) is deleted.

Literals still fold through `common_time`'s parser, the same one plain `INSERT ... VALUES` uses, so a given string means the same thing however it reaches a column.

The rewrite reads source types, so it now runs `TypeCoercion` first. Before coercion a `UNION` still carries its loose per-branch schema, and retargeting a cast whose input later becomes a timestamp would shift the value instead of reinterpreting it. This also keeps explicit casts recognisable — the SQL layer turns `CAST(x AS TIMESTAMP)` into an `arrow_cast` call, which only becomes an `Expr::Cast` later in the optimizer.

### Limitation

The emitted cast keeps its timezone only where it is evaluated locally. Substrait drops the timezone name on push-down (it encodes any zoned timestamp as `PrecisionTimestampTz` and decodes it back as UTC), so a source reading from a table falls back to UTC — which is the behaviour it already had before this PR, not a regression. Literals, `VALUES` and `UNION`s of them never leave the node and keep the session timezone; that is what an INSERT's timestamp assignment is in practice. Documented on the rule; the substrait gap is worth its own issue since it affects anything relying on zoned timestamps across nodes.

## PR Checklist


MichaelScofield:
@codex review

src/query/src/optimizer/insert_assignment.rs
```diff
@@ -0,0 +1,277 @@
...
+
+fn session_timezone(query_ctx: &QueryContextRef) -> Option<SessionTimezone> {
+    let parsed = query_ctx.timezone();
+
+    // A UTC session already gets UTC semantics from the plain assignment cast.
+    // Only an exact UTC match is skipped: a named zone sitting at +00:00 today
+    // may not have been there for the timestamp being inserted.
+    let is_utc = match &parsed {
+        Timezone::Offset(offset) => offset.local_minus_utc() == 0,
+        Timezone::Named(tz) => tz.name() == "UTC",
```
MichaelScofield: Do we need to use the `eq_ignore_ascii_case` here? If not, are there any consts in the `Timezone` crate we can use for the "UTC"?

src/query/src/optimizer/insert_assignment.rs
```diff
@@ -0,0 +1,277 @@
...
+        && let Some(rewritten) = rewrite_values(values, timezone)?
+    {
+        input = LogicalPlan::Values(rewritten);
+        changed = true;
+    }
+
+    if !changed {
+        return Ok(LogicalPlan::Projection(assignment));
+    }
+    Projection::try_new(exprs, Arc::new(input)).map(LogicalPlan::Projection)
```
MichaelScofield: Seems you wasted a maybe heavy (not an `Arc`) clone if the `assignment.input` is not a `LogicalPlan::Values`, in line 133.

How about this:

```suggestion
    // The planner types `VALUES` against the target table, so the assignment
    // cast lands inside the `Values` rows instead of on the projection above.
    let mut input = assignment.input.clone();
    if let LogicalPlan::Values(values) = &input
        && let Some(rewritten) = rewrite_values(values, timezone)?
    {
        input = Arc::new(LogicalPlan::Values(rewritten));
        changed = true;
    }

    if !changed {
        return Ok(LogicalPlan::Projection(assignment));
    }
    Projection::try_new(exprs, input).map(LogicalPlan::Projection)
```

src/query/src/optimizer/insert_assignment.rs
```diff
@@ -0,0 +1,277 @@
...
+    let unit = *unit;
+
+    if !matches!(
+        source.get_type(schema)?,
+        DataType::Utf8 | DataType::LargeUtf8 | DataType::Utf8View
+    ) {
+        return Ok(false);
+    }
+
+    let source = source.as_ref().clone();
```
MichaelScofield: Seems this clone can be postponed after the "if" below?

MichaelScofield (approved):
No blocking comments, so LGTM.

src/query/src/optimizer/insert_assignment.rs
killme2008: Not needed — `tz.name()` returns the canonical name, which is always uppercase `UTC`.

There was no existing constant to use: `Timezone` did not expose one, and `query` does not depend on `chrono-tz` so `Tz::UTC` was not reachable either. Added `Timezone::is_utc()` instead — the check belongs to the type anyway, and it drops the bare string compare. 0aeed2141b

src/query/src/optimizer/insert_assignment.rs
killme2008: Good catch, that clone was pure waste — `input` is almost never `Values`. Applied in 0aeed2141b: clone the `Arc`, and only rebuild it when a `Values` row actually changes.

One tweak to your suggestion: `if let LogicalPlan::Values(values) = &input` does not compile against `&Arc<LogicalPlan>` (no auto-deref in match), so it reads `assignment.input.as_ref()`.

src/query/src/optimizer/insert_assignment.rs
killme2008: Yes. The clone was there to end the borrow before assigning to `*expr`, but it is wasted whenever the literal-folding path returns.

0aeed2141b computes `folded` through an immutable borrow first, so the borrow ends before the assignment, and the clone only happens on the path that actually builds the cast.

src/query/src/optimizer/insert_assignment.rs
```diff
@@ -0,0 +1,277 @@
...
+    else {
+        return Ok(false);
+    };
+    let unit = *unit;
+
+    if !matches!(
+        source.get_type(schema)?,
+        DataType::Utf8 | DataType::LargeUtf8 | DataType::Utf8View
+    ) {
+        return Ok(false);
```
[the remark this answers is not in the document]
killme2008: The scenario is real — `planner.rs` does handle `Cast(Placeholder, ..)` for prepared statements — but the failure mode is not: `get_type` does not error on an untyped placeholder, it returns `Ok(DataType::Null)`.

```rust
// datafusion/expr/src/expr_schema.rs
Expr::Placeholder(Placeholder { field, .. }) => {
    if let Some(field) = field {
        Ok(field.data_type().clone())
    } else {
        // If the placeholder's type hasn't been specified, treat it as
        // null (unspecified placeholders generate an error during planning)
        Ok(DataType::Null)
    }
}
```

`Null` does not match the `Utf8 | LargeUtf8 | Utf8View` guard, so the rewrite returns early and leaves the cast for parameter substitution. No planning error, and no rewrite.

Added a regression test in 0aeed2141b so the guard keeps covering this if the type check is ever widened.

killme2008:
Pushed 0aeed2141b for the review comments: `Arc` clone instead of a subtree copy, cast-source clone deferred past literal folding, UTC check moved onto `Timezone::is_utc()`, plus a regression test for prepared `INSERT ... VALUES (?)`.

Two of Copilot's suppressed comments deserve an answer since they are about deliberate choices rather than oversights.

**Ambiguous local times parse differently for literals and non-literals.** Intentional. Plain `INSERT ... VALUES (...)` never reaches DataFusion — it goes through `sql_value_to_value`, which parses with `common_time::Timestamp::from_str` and resolves an ambiguous local time to the first instant. Keeping literals on that same parser is what makes a given string mean the same thing on every path. Switching everything to Arrow's parser instead would mean `INSERT INTO t VALUES ('2026-11-01 01:30:00')` succeeds while `INSERT INTO t VALUES ('2026-11-01 01:30:00', now())` fails, because the second form takes the DataFusion path — same shape of SQL, different behaviour. That inconsistency seemed worse than the one being flagged. Non-literal sources got no timezone conversion at all before this PR, so Arrow's stricter parsing there is not a regression.

**Prepared plans bake in the prepare-time timezone.** Real, but pre-existing: the previous implementation folded literals into timestamp constants, which embeds the prepare-time zone just the same. Fixing it means changing how prepared plans are cached or invalidated, which is out of scope here.

Worth recording separately: the emitted cast only keeps its timezone where it is evaluated locally. Substrait discards the timezone name on push-down — it encodes any zoned timestamp as `PrecisionTimestampTz` and decodes it back as UTC — so an INSERT sourced from a table scan falls back to UTC, which is what it did before this PR. Literals, `VALUES`, and `UNION`s of them never leave the node and keep the session timezone, and that is what an INSERT timestamp assignment is in practice. Documented on the rule; I will open an issue for the substrait gap since it affects anything relying on zoned timestamps across nodes.

PR #8916, opened 2026-08-20
fix(mito2): deliver complete WAL entries without waiting for next input

discord9:

Both WAL entry adapters used a look-ahead buffer: a received entry was only decoded and delivered after the **next** entry arrived, or when the stream/channel closed. Entry completeness is fully determined by the entry itself via `Entry::is_complete()` (naive entries are always complete; multipart entries require both First and Last parts), so waiting for the next input is unnecessary.

Concretely:

- `LogStoreEntryReader` (`entry_reader.rs`): buffered each entry and only yielded it when the next entry was polled or on EOF. A complete entry followed by a long-pending upstream stream would be withheld indefinitely.
- `WalEntryReceiver` (`entry_distributor.rs`): same pattern over a Tokio mpsc channel — a complete entry was held until the next `recv()` or channel close.

This changes both to decode and yield a complete entry immediately upon receipt; incomplete entries are still skipped one by one as before. Replay/replication ordering, entry IDs, and tail-incomplete semantics are unchanged.

Regression tests construct an upstream that delivers one complete entry and then stays alive but pending, and assert the entry is delivered within a short timeout (the old implementation would block waiting for a second entry).

## PR Checklist


WenyXu:
@codex review

WenyXu (approved):

PR #8918, opened 2026-08-20
fix(client): yield Flight batches and affected rows without waiting for next message

discord9:

### Summary

The gRPC Flight client used to deliver a `RecordBatch` (or an `AffectedRows` result) only after polling the **next** Flight message, so that trailing terminal `Metrics` could be attached first. In practice the next message often never arrives promptly (e.g. `Batch → Pending` on a plain query), so already-received batches/affected-row results were withheld and queries timed out (~100ms+ stalls; flow batching tasks saw repeated watermark timeouts). This PR makes results flow immediately and handles terminal metrics independently.

### How it works

- **RecordBatch lookahead removed** (`src/client/src/region.rs`, `src/client/src/database.rs`): each received batch is yielded immediately via a per-message state machine; `Metrics` messages update terminal metrics when polled, never blocking batch delivery.
- **AffectedRows lookahead removed** (`src/client/src/database.rs`): new servers send inline metrics in the `AffectedRows` message, which are authoritative and make the output final synchronously — no trailing message is consumed. For legacy servers that put `Metrics` in a separate trailing message, the result is returned immediately while a bounded (100ms timeout) background read collects the trailing metrics; completion is signaled via `Notify` (`OutputMetrics::wait_ready()`), and the spawned task is cancelled through an `AbortHandle` when the `OutputMetrics` handle is dropped.
- **Hardened metrics parsing** (`src/client/src/database.rs`): malformed terminal metrics JSON no longer silently vanishes — parse failures are surfaced as errors (stream error on the query path, `completion_error()` on the affected-rows path).
- **Flow** (`src/flow/src/batching_mode/task.rs`): awaits `wait_ready()` on the terminal metrics before reading watermarks, so per-region watermarks are computed only from final metrics.
- **Tests**: fixed false positives in the Flight metrics tests — poll counting replaces a sentinel flag, and a oneshot confirms the legacy-trailing completion task is cancelled before the 100ms timeout fires.

### Limitations

- Legacy (pre-inline-metrics) servers still rely on the bounded 100ms trailing-metrics read; if a legacy server sends metrics later than that, affected-row terminal metrics are absent and a `completion_error` is logged instead of blocking the caller.
- `OutputMetrics` API additions (`wait_ready`, `completion_error`, `is_ready`) are additive; no existing API is removed or changed in behavior for well-formed servers.

### Compatibility

No API or data compatibility break: the Flight wire protocol is unchanged — inline metrics remain preferred and the legacy trailing `Metrics` message is still accepted. Client-side behavior only changes in that results are delivered earlier instead of being withheld.

### Verification

- `cargo nextest run -p client`: 35 passed
- `cargo nextest run -p flow`: 234 passed
- Red-light confirmation: before the fix the affected tests failed 4/4 runs with 100ms timeouts; they pass after the fix.

## PR Checklist

### DIFF
--- a/src/client/src/database.rs
+++ b/src/client/src/database.rs
@@ -16,7 +16,7 @@ use std::collections::HashMap;
 use std::pin::Pin;
 use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex, RwLock};
 use std::task::{Context, Poll};
 use std::time::Duration;
 
@@ -55,6 +55,7 @@ use futures::future;
 use futures_util::{Stream, StreamExt, TryStreamExt};
 use prost::Message;
 use snafu::{IntoError, ResultExt};
+use tokio::sync::Notify;
 use tonic::metadata::{AsciiMetadataKey, AsciiMetadataValue, MetadataMap, MetadataValue};
 use tonic::transport::Channel;
 
@@ -70,20 +71,47 @@ type FlightDataStream = Pin<Box<dyn Stream<Item = FlightData> + Send>>;
 type DoPutResponseStream = Pin<Box<dyn Stream<Item = Result<DoPutResponse>>>>;
 
 const HINTS_METADATA_KEY: &str = "x-greptime-hints";
+/// Maximum time to wait for the optional trailing metrics message after
+/// affected rows have already been delivered.
+const FLIGHT_TRAILING_METRICS_TIMEOUT: Duration = Duration::from_secs(5);
 
 /// Terminal metrics associated with a query output.
 ///
 /// For streaming outputs, metrics are only final after the stream is fully
-/// drained and [`Self::is_ready`] returns `true`.
+/// drained and [`Self::is_ready`] returns `true`. Affected-row outputs may
+/// briefly await a compatibility trailing metrics message.
 #[derive(Debug, Clone, Default)]
 pub struct OutputMetrics {
     inner: Arc<OutputMetricsInner>,
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 struct OutputMetricsInner {
     metrics: RwLock<Option<RecordBatchMetrics>>,
+    completion_error: RwLock<Option<String>>,
     ready: AtomicBool,
+    ready_notify: Notify,
+    compatibility_task: Mutex<Option<tokio::task::AbortHandle>>,
+}
+
+impl Default for OutputMetricsInner {
+    fn default() -> Self {
+        Self {
+            metrics: RwLock::new(None),
+            completion_error: RwLock::new(None),
+            ready: AtomicBool::new(false),
+            ready_notify: Notify::new(),
+            compatibility_task: Mutex::new(None),
+        }
+    }
+}
+
+impl Drop for OutputMetricsInner {
+    fn drop(&mut self) {
+        if let Some(handle) = self.compatibility_task.get_mut().unwrap().take() {
+            handle.abort();
+        }
+    }
 }
 
 impl OutputMetrics {
@@ -98,10 +126,45 @@ impl OutputMetrics {
 
     /// Marks the terminal metrics as final for this output.
     pub fn mark_ready(&self) {
-        let _ = self
+        if self
             .inner
             .ready
-            .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire);
+            .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
+            .is_ok()
+        {
+            self.inner.ready_notify.notify_waiters();
+        }
+    }
+
+    /// Waits until terminal metrics are final.
+    pub async fn wait_ready(&self) {
+        loop {
+            let notified = self.inner.ready_notify.notified();
+            if self.is_ready() {
+                return;
+            }
+            notified.await;
+        }
+    }
+
+    /// Returns an error encountered while completing the output, if any.
+    pub fn completion_error(&self) -> Option<String> {
+        self.inner.completion_error.read().unwrap().clone()
+    }
+
+    fn set_completion_error(&self, error: impl Into<String>) {
+        *self.inner.completion_error.write().unwrap() = Some(error.into());
+    }
+
+    fn set_compatibility_task(&self, handle: tokio::task::AbortHandle) {
+        let mut task = self.inner.compatibility_task.lock().unwrap();
+        if !self.is_ready() {
+            *task = Some(handle);
+        }
+    }
+
+    fn take_compatibility_task(&self) -> Option<tokio::task::AbortHandle> {
+        self.inner.compatibility_task.lock().unwrap().take()
     }
 
     /// Returns whether terminal metrics are final.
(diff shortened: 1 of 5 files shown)


killme2008:
@discord9 There are conflicts

src/client/src/database.rs
```diff
@@ -275,29 +316,67 @@ where
...
-                            "'AffectedRows' Flight message can only be followed by a Metrics message, got {other:?}"
-                        ),
+                terminal_metrics.mark_ready();
+            } else {
+                // Old servers may put Metrics in a separate message. The
+                // AffectedRows result can be returned immediately, but its
+                // metrics are not final until this bounded compatibility read
+                // has completed.
+                let weak_metrics = Arc::downgrade(&terminal_metrics.inner);
+                let task = common_runtime::spawn_global(async move {
```
killme2008, line 357: Should we use `read` runtime here?

src/client/src/database.rs
```diff
@@ -178,6 +217,8 @@ impl OutputWithMetrics {
     }
 }
 
+const LEGACY_TRAILING_METRICS_TIMEOUT: Duration = Duration::from_millis(100);
```
killme2008: Is it a standard measured value, or based purely on experience?

src/flow/src/batching_mode/task.rs
```diff
@@ -770,6 +770,12 @@ impl BatchingTask {
         }
 
         let res = res?;
+        res.metrics.wait_ready().await;
```
killme2008: This hangs forever if `res.output.data` is a `Stream`.

`StreamWithMetrics::poll_next` only calls `mark_ready()` on `Poll::Ready(None)`, so metrics become ready only after the stream is drained. Here we just call `extract_rows_and_cost()`, which returns `(0, cost)` for anything that isn't `AffectedRows` and never touches the stream — so nothing will ever mark it ready.

Stream output isn't hypothetical: `wrap_standalone_output_with_terminal_metrics` explicitly branches on `OutputData::Stream(_)`, and the existing test at `frontend_client.rs` builds exactly that case. The reachable path is `breakup_insert_plan` returning `None` above, where we send the raw logical plan and get a SELECT back.

Suggest gating on `matches!(res.output.data, OutputData::AffectedRows(_))`, or wrapping in a timeout. Either way `wait_ready`'s rustdoc should say that stream outputs must be drained first — that's an easy trap for the next caller.

killme2008 (commented):
Removing the lookahead is the right fix — the server only emits `Metrics` after the batch loop ends (`servers/src/grpc/flight/stream.rs`), so the old code held every batch until the *next* batch was produced. Three comments inline, plus: this needs a rebase, and #8943 rewrote the same block in `region.rs` to wrap stream errors with `flight_stream_error(&stream_addr, e)` — please keep that when resolving.

src/client/src/region.rs
```diff
@@ -226,70 +226,39 @@ where
...
+                    ));
+                }
+                FlightMessage::Metrics(s) => match serde_json::from_str(&s) {
+                    Ok(m) => {
+                        // Metrics are updated when the message is polled after a batch.
+                        metrics_ref.swap(Some(Arc::new(m)));
+                    }
+                    Err(e) => {
+                        yield IllegalFlightMessagesSnafu {
+                            reason: format!("Invalid terminal metrics message: {e}"),
```
killme2008: This turns a metrics parse failure into a query failure on the frontend→datanode path, where the two sides are upgraded independently. `RecordBatchMetrics::{elapsed_compute, memory_usage, plan_metrics}` have no `#[serde(default)]`, so any future change to their shape breaks queries during a rolling upgrade instead of just dropping a diagnostic field.

`database.rs` already behaves this way, but that's client→frontend. I'd keep this one non-fatal: warn + a counter, and leave the stream alone.

src/client/src/database.rs
```diff
@@ -275,29 +316,67 @@ where
...
+                        Ok(Some(Ok(FlightMessage::Metrics(metrics)))) => {
+                            match parse_terminal_metrics(&metrics) {
+                                Ok(metrics) => Some((Some(metrics), None)),
+                                Err(error) => Some((
+                                    None,
+                                    Some(format!("Invalid trailing terminal metrics: {error}")),
+                                )),
+                            }
+                        }
+                        Ok(Some(Ok(message))) => Some((
```
killme2008: Worth calling out: this used to be a hard error, and dropping `test_affected_rows_inline_metrics_rejects_trailing_metrics` removes the only check on "a server must not send both inline and trailing metrics". All protocol violations now collapse into a `completion_error()` string that only flow logs — every other caller silently ignores it.

Fine by me given `AffectedRows` is already terminal, but let's be explicit that the invariant is now unenforced.

src/client/src/database.rs
killme2008: Adding to this: the cost of getting it wrong isn't symmetric. Missing watermarks make `apply_query_result_to_state` fall back to `FullSnapshot` — safe, but a full recompute. So a tight bound trades a rare stall for a full snapshot rebuild, which argues for erring large, not small.

Also worth being clear about what it actually bounds. The terminal message already arrived; if an old server sends no trailing `Metrics`, EOF comes with the trailers anyway and we return immediately. The only case left is a stalled connection, which is network-scale — closer to `batch_opts.grpc_conn_timeout` (5s) than to 100ms.

Reusing `batch_opts` directly isn't possible as written: the const lives in `src/client`, and flow depends on client, not the reverse. It'd mean putting the timeout on `Database` and having flow pass it down from `BatchingModeOptions`. That seems worth doing if we want a knob at all; otherwise pick the value from the cost asymmetry above and say so in a comment, so the next person doesn't have to re-derive it.

src/client/src/database.rs
discord9: The compatibility reader now uses `common_runtime::spawn_global`. There is no `read` runtime or `spawn_read` entry point in the runtime API, and this is asynchronous Flight-stream I/O rather than blocking work.

src/client/src/database.rs
discord9: The compatibility bound is now 5 seconds rather than 100 ms. It only bounds the optional legacy trailing-metrics read after affected rows have already been delivered.

src/flow/src/batching_mode/task.rs
discord9: Addressed by gating `wait_ready()` to `OutputData::AffectedRows`. Flow never waits on an undrained stream; stream metrics become ready only after EOF.

src/client/src/region.rs
discord9: Addressed. The region path now warns and continues on malformed metrics, while actual Flight transport errors still retain peer context through `flight_stream_error(&stream_addr, error)`.

src/client/src/database.rs
discord9: This is now explicit and tested: inline affected-row metrics are authoritative, and that path completes without polling or consuming a trailing message.

discord9:
Rebased and rebuilt at `b43c51c58db65f76b52b0283df701cf5db544fb4`. The retained-Flight error context from #8943 remains in the region stream path, and batches/affected rows are now delivered without lookahead. Post-rebase validation passed: client 42/42, Flow 234/234, focused distributed Flight integration 1/1, fmt/check/clippy.

src/client/src/database.rs
```diff
@@ -280,32 +347,55 @@ where
...
+                            let error = flight_stream_error(&remote_addr, error);
+                            metrics.set_completion_error(error.to_string());
+                            warn!("{}", error);
+                        }
+                        Err(_) => {
+                            let error = "Timed out waiting for trailing Flight metrics";
+                            metrics.set_completion_error(error);
+                            warn!("{} from {}", error, remote_addr);
+                        }
                     }
```
WenyXu: nit: Can we move this logic into a function?

discord9: Extracted the background task into `spawn_affected_rows_trailing_metrics_task` in 32e26ba13c696e49a3142befe54967fcfef5fb7f. The affected-rows branch now calls the helper; timeout, weak ownership, cancellation, and metrics completion behavior are unchanged. Client tests, all-targets check, formatting, and Clippy with `-D warnings` passed.

Issue #8879, opened 2026-08-13
MySQL server-side prepared statements ignore session timezone for timestamp parameters


evenyag:
### What type of bug is this?

Incorrect result

### What subsystems are affected?

Frontend, Write Protocols, Query Engine

### Minimal reproduce step

Connect to GreptimeDB through MySQL Connector/J with server-side prepared statements enabled and a non-UTC connection/session timezone:

```text
jdbc:mysql://127.0.0.1:4002/public?useServerPrepStmts=true&connectionTimeZone=Asia/Shanghai&forceConnectionTimeZoneToSession=true
```

Create a table and verify the session timezone:

```sql
CREATE TABLE timezone_prepared_test (
  ts TIMESTAMP TIME INDEX,
  value INT
);

SELECT @@session.time_zone;
-- Asia/Shanghai
```

Insert a timestamp using a server-side prepared statement:

```java
try (PreparedStatement stmt = conn.prepareStatement(
        "INSERT INTO timezone_prepared_test (ts, value) VALUES (?, ?)")) {
    stmt.setTimestamp(1, Timestamp.valueOf("2026-08-13 08:00:00"));
    stmt.setInt(2, 1);
    stmt.executeUpdate();
}
```

The same problem affects timestamp parameters in prepared query predicates:

```java
try (PreparedStatement stmt = conn.prepareStatement(
        "SELECT * FROM timezone_prepared_test WHERE ts = ?")) {
    stmt.setTimestamp(1, Timestamp.valueOf("2026-08-13 08:00:00"));
    stmt.executeQuery();
}
```

For comparison, execute an equivalent text-protocol statement after setting the session timezone:

```sql
SET time_zone = 'Asia/Shanghai';
INSERT INTO timezone_prepared_test VALUES ('2026-08-13 08:00:00', 2);
```

The two paths interpret the timezone-less timestamp differently.

The server-side prepared parameter path converts a MySQL binary `DATETIME` to a naive datetime and then unconditionally calls `.and_utc().timestamp_millis()` in `src/servers/src/mysql/helper.rs`. This conversion does not receive the query/session timezone. Byte parameters targeting timestamp columns are also parsed with `Timestamp::from_str_utc`.

### What did you expect to see?

Timestamp parameters in MySQL server-side prepared statements should have timezone semantics consistent with ordinary SQL timestamp literals and the MySQL session.

[...]

PR #8922, opened 2026-08-20
fix(mysql): respect session timezone for prepared timestamps

DebadityaHait:

## Refer to a related PR or issue link

Refs #8879

MySQL server-side prepared timestamp parameters were interpreted as UTC even when the connection session used another timezone. This caused binary prepared values to differ from equivalent text timestamp literals.

This change passes the query session timezone into MySQL prepared-parameter conversion. Binary datetime values and byte-form timestamp parameters now use the session timezone while preserving explicit timestamp offsets. Storage remains UTC.

A regression test compares text and binary prepared timestamp predicates under `+08:00`.

Validation:
- `cargo +1.97.1 fmt --all -- --check`
- `git diff --check`

## PR Checklist

### DIFF
--- a/src/servers/src/mysql/handler.rs
+++ b/src/servers/src/mysql/handler.rs
@@ -25,6 +25,7 @@ use common_catalog::parse_optional_catalog_and_schema_from_db_string;
 use common_error::ext::ErrorExt;
 use common_query::Output;
 use common_telemetry::{debug, error, tracing, warn};
+use common_time::Timezone;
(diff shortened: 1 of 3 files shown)


sunng87:
Hello @DebadityaHait ,thank you for sending this. However, there are several patches working on the same issue. After comparing we decided to move on with #8923 . I'm closing this. You are welcomed to work on other open issues of greptimedb.

PR #8923, opened 2026-08-20
fix(mysql): interpret prepared statement datetime params in session timezone

### LINKED ISSUE #8879: MySQL server-side prepared statements ignore session timezone for timestamp parameters

wy471x:
Binary DATETIME parameters of server-side prepared statements were converted as if UTC, ignoring the session timezone set via SET time_zone. Convert them with the session timezone and add an integration test covering prepared inserts and predicates under Asia/Shanghai.

Fixes: [#8879](https://github.com/GreptimeTeam/greptimedb/issues/8879)

 Fix:                                                                                                                                                                               
  - src/servers/src/mysql/handler.rs — pass the session timezone from query_ctx into parameter conversion (replace_params_with_values / replace_params_with_exprs).                  
  - src/servers/src/mysql/helper.rs — convert timezone-less datetimes with the session timezone (datetime_to_timestamp_millis), handle DST gaps/ambiguity, and parse byte-encoded timestamps with the timezone (convert_bytes_to_timestamp). Added unit tests.                                                                                                       
                                                                                                                                                                                     
  Test:                                                                                                                                                                              
  - tests-integration/tests/sql.rs — new test_mysql_prepare_stmt_timezone: under Asia/Shanghai, a prepared insert of 2026-08-13 08:00:00 stores 00:00:00 UTC, and prepared predicates match the same instant as text literals.
<!--    

-->

## PR Checklist

- [X] I have written the necessary rustdoc comments.
- [X] I have added the necessary unit tests and integration tests.
- [X] This PR requires documentation updates.
- [X] API changes are backward compatible.
- [X] Schema or data changes are backward compatible.


sunng87 (commented):
A few inline suggestions (all non-blocking). Thanks for the fix!

src/servers/src/mysql/helper.rs
```diff
@@ -308,8 +319,26 @@ pub fn convert_expr_to_scalar_value(param: &Expr, t: &ConcreteDataType) -> Resul
     }
 }
 
-fn convert_bytes_to_timestamp(bytes: &[u8], ts_type: &TimestampType) -> Result<ScalarValue> {
-    let ts = Timestamp::from_str_utc(&String::from_utf8_lossy(bytes))
+/// Interprets a timezone-less datetime in the given timezone and returns the
+/// corresponding epoch timestamp in milliseconds.
+fn datetime_to_timestamp_millis(datetime: NaiveDateTime, timezone: &Timezone) -> Result<i64> {
```
sunng87, line 322: Nice to see the DST gap/ambiguity handled explicitly. One thought: this `LocalResult` match duplicates the policy in common-time's private `naive_datetime_to_timestamp` (`src/common/time/src/timestamp.rs`), which is what `Timestamp::from_str(s, Some(tz))` uses for the text protocol (gap -> error, ambiguous -> earlier instant). Today the two agree so text/binary parity holds, but if common-time ever changes that policy, this path would silently diverge from the text protocol — the same class of bug this PR fixes.

Would you consider exposing something like `Timestamp::from_naive_datetime(NaiveDateTime, &Timezone)` in common-time (or moving this helper there) so the policy lives in one place? Not a blocker either way.

src/servers/src/mysql/helper.rs
```diff
@@ -15,9 +15,11 @@
 use std::ops::ControlFlow;
 use std::time::Duration;
 
-use chrono::NaiveDate;
+use chrono::{LocalResult, NaiveDate, NaiveDateTime};
 use common_query::prelude::ScalarValue;
 use common_sql::convert::sql_value_to_value;
+use common_time::timezone::Timezone;
```
sunng87: Minor: `Timezone` is re-exported at the crate root and the rest of `servers` uses the short form (e.g. `use common_time::Timezone;` in `src/servers/src/postgres/auth_handler.rs`). Mind using that here and in `handler.rs` for consistency?

src/servers/src/mysql/helper.rs
```diff
@@ -234,15 +242,13 @@ pub fn convert_value(param: &ParamValue, t: &ConcreteDataType) -> Result<ScalarV
...
-                })?
-                .and_utc()
-                .timestamp_millis();
+            let datetime = to_naive_datetime(param.value).map_err(|e| {
+                error::MysqlValueConversionSnafu {
+                    err_msg: e.to_string(),
+                }
+                .build()
+            })?;
+            let timestamp_millis = datetime_to_timestamp_millis(datetime, timezone)?;
```
sunng87: Nit: `t` is only checked after the datetime->millis conversion. Matching `t` first (and moving the conversion inside the `Timestamp` arm) would surface the `PreparedStmtTypeMismatch` error before doing any work and skip a wasted conversion on the mismatch path. Cosmetic.

wy471x:
> A few inline suggestions (all non-blocking). Thanks for the fix!

Hi, these issues has been fixed! Please take a look when you have time. @sunng87

sunng87 (approved):
The test failure is unrelated to this patch. Will update the branch once it's fixed.

PR #8924, opened 2026-08-20
fix: respect session timezone in MySQL prepared statements

### LINKED ISSUE #8879: MySQL server-side prepared statements ignore session timezone for timestamp parameters

changshenhan:
## Problem

In `src/servers/src/mysql/helper.rs`, prepared-statement parameters are converted to timestamps without considering the session timezone:

- `ValueInner::Datetime` calls `.and_utc()` on the naive wall-clock datetime the client sent, hard-coding UTC;
- `convert_bytes_to_timestamp` parses bytes strings via `Timestamp::from_str_utc`.

For any session whose timezone is not UTC, the wall-clock values are interpreted at the wrong offset, so the stored instant is off by the session offset.

## Fix

Thread the session timezone through the parameter-conversion chain:

- `do_execute` reads `query_ctx.timezone()` and passes it to `replace_params_with_values` / `replace_params_with_exprs`;
- `helper::convert_value` and `helper::convert_expr_to_scalar_value` take a `&Timezone`;
- the `Datetime` path converts with `datetime_to_utc` (the same `from_local_datetime`-based conversion `Timestamp::from_str` already uses) instead of `.and_utc()`, returning a clear error for wall-clock times that don't exist in the session timezone;
- `convert_bytes_to_timestamp` parses with `Timestamp::from_str(s, Some(&timezone))`;
- the CLI path `convert_expr_to_scalar_value` now forwards the timezone to `sql_value_to_value` (previously `None`).

## Tests

- `test_convert_bytes_to_timestamp_with_timezone`: a `+08:00` session shifts a wall-clock string to the matching UTC instant, while strings with explicit offsets keep their own offset.

Closes #8879

### DIFF
--- a/src/servers/src/mysql/handler.rs
+++ b/src/servers/src/mysql/handler.rs
@@ -25,6 +25,7 @@ use common_catalog::parse_optional_catalog_and_schema_from_db_string;
 use common_error::ext::ErrorExt;
 use common_query::Output;
 use common_telemetry::{debug, error, tracing, warn};
+use common_time::Timezone;
 use datafusion_common::ParamValues;
 use datafusion_expr::LogicalPlan;
 use datatypes::prelude::ConcreteDataType;
@@ -289,12 +290,13 @@ impl MysqlInstanceShim {
                     .fail();
                 }
 
+                let timezone = query_ctx.timezone();
                 let replaced_plan = match params {
                     Params::ProtocolParams(params) => {
-                        replace_params_with_values(&plan, param_types, &params)
+                        replace_params_with_values(&plan, param_types, &params, &timezone)
                     }
                     Params::CliParams(params) => {
-                        replace_params_with_exprs(&plan, param_types, &params)
+                        replace_params_with_exprs(&plan, param_types, &params, &timezone)
                     }
                 }?;
 
@@ -807,6 +809,7 @@ fn replace_params_with_values(
     plan: &LogicalPlan,
     param_types: HashMap<String, Option<ConcreteDataType>>,
     params: &[ParamValue],
+    timezone: &Timezone,
 ) -> Result<LogicalPlan> {
     debug_assert_eq!(param_types.len(), params.len());
 
@@ -824,7 +827,7 @@ fn replace_params_with_values(
 
     for (i, param) in params.iter().enumerate() {
         if let Some(Some(t)) = param_types.get(&format_placeholder(i + 1)) {
-            let value = helper::convert_value(param, t)?;
+            let value = helper::convert_value(param, t, timezone)?;
 
             values.push(value.into());
         }
@@ -839,6 +842,7 @@ fn replace_params_with_exprs(
     plan: &LogicalPlan,
     param_types: HashMap<String, Option<ConcreteDataType>>,
     params: &[sql::ast::Expr],
+    timezone: &Timezone,
 ) -> Result<LogicalPlan> {
     debug_assert_eq!(param_types.len(), params.len());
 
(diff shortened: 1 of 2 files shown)


sunng87:
Hello @changshenhan  ,thank you for sending this. However, there are several patches working on the same issue. After comparing we decided to move on with #8923 . I'm closing this. You are welcomed to work on other open issues of greptimedb.

PR #8925, opened 2026-08-20
fix(mito2): fail region open when WAL replay cannot apply an entry

### LINKED ISSUE #8809: mito: memtable apply failures during WAL replay are silently swallowed; partial write failures leave stats inconsistent
### What type of bug is this?

Incorrect result, Other

Storage Engine, Datanode

### Minimal reproduce step

Code refs at main `943eee852`. Four related defects:

**1. WAL replay is fail-open: memtable apply failures are swallowed silently.**
`replay_memtable` pushes every mutation with `OptionOutputTx::none()` (`src/mito2/src/region/opener.rs:874`, bulk: `:896`), so an apply failure during replay produces no error, no log and no metric — the region opens "successfully" with the rows missing. #8806 showed this path is load-bearing: its flat-format variant silently lost acknowledged rows on every restart. #8808 fixed that specific trigger (and the replay-side `fill_missing_columns` at `opener.rs:889` now heals stale bulk entries), but the backstop itself is unchanged: whatever makes a memtable apply fail at replay next time will again be dropped in silence. Note the sparse write path still skips `check_schema` entirely (`src/mito2/src/request.rs:676`), so WAL admission is not fully gated by validation.

**2. `rows_replayed` counts rows that were not applied.**
The counter is bumped when the mutation is pushed (`opener.rs:865`, `:891`), not when it is applied, so the recovery log reports dropped rows as replayed.

**3. Partial application leaves memtable contents and stats inconsistent.**
`TimeSeriesMemtable::write` early-returns mid-loop (`src/mito2/src/memtable/time_series.rs:229`; the TODO at `:235-237` already notes it), so a failed mutation can leave an applied prefix behind — and `update_stats` is skipped on error, so those rows are invisible to the stats. `BulkMemtable::write_bulk` has the same shape: the fragment is buffered before the fallible `to_bulk_part()?` compaction of accumulated parts (`src/mito2/src/memtable/bulk.rs:460-478`), and the `?` skips `update_stats` (`:488-492`) whose comment states `ranges()` relies on exact stats to avoid re-pruning by time range. Rows present but unaccounted can therefore be mis-pruned at read time — an incorrect-result path that needs no replay at all.

**4. Post-WAL apply failures surface as definitive rejections while the entry stays durable.**
When the memtable apply fails after the WAL write, the raw memtable error is returned to the caller — including `StatusCode::InvalidArguments`-class errors (e.g. `PrimaryKeyLengthMismatch`, `src/mito2/src/error.rs:1482`) — while the mutation remains in the WAL and `write_memtable`/`write_bulk` still advance the committed sequence over it (`src/mito2/src/region_write_ctx.rs:292`, `:373`). An error that cannot guarantee "not written" must not use a definitive-rejection class; `InvalidArguments` tells the client the data was rejected, yet the row may still exist after a restart (or be gone permanently if a flush truncates the WAL first — the outcome is decided by a race the client cannot observe).

I have four unit tests (verified at `d90cca4b7`, pre-#8808; the row-mutation paths above are unchanged at `943eee852`) asserting the current behavior — happy to send them as a PR together with the fixes. The central one:

<details>
<summary>The replay silent-drop test (drop-in for <code>src/mito2/src/region/opener.rs</code> tests)</summary>

```rust
/// Builds a WAL entry reader that yields the given entries.
fn mock_wal_reader(
    provider: &Provider,
    region_id: RegionId,
    entries: Vec<(u64, WalEntry)>,
) -> Box<LogStoreEntryReader<MockRawEntryStream>> {
    Box::new(LogStoreEntryReader::new(MockRawEntryStream {
        entries: entries
            .into_iter()
            .map(|(entry_id, wal_entry)| {
                Entry::Naive(NaiveEntry {
                    provider: provider.clone(),
                    region_id,
                    entry_id,
                    data: wal_entry.encode_to_vec(),
                })
            })
            .collect(),
    }))
}

/// A mutation whose primary key value cannot be encoded: the schema declares
/// `tag_0` as a string but the row carries an i64. `WriteRequest::new` rejects
/// such rows today, but nothing re-validates WAL contents on replay.
fn poison_mutation(sequence: u64) -> Mutation {
    Mutation {
        op_type: OpType::Put as i32,
        sequence,
        rows: Some(Rows {
            schema: vec![
                time_index_column_schema("ts", ColumnDataType::TimestampMillisecond),
                tag_column_schema("tag_0", ColumnDataType::String),
            ],
            rows: vec![Row {
                values: vec![
                    ts_ms_value(1),
                    v1::Value {
                        value_data: Some(ValueData::I64Value(42)),
                    },
                ],
            }],
        }),
        write_hint: None,
    }
}

#[tokio::test]
async fn test_replay_silently_drops_mutation_failing_memtable_write() {
    let mut builder = VersionControlBuilder::new();
    builder.set_memtable_builder(Arc::new(TimeSeriesMemtableBuilder::new(
        None,
        true,
        MergeMode::LastRow,
    )));
    let region_id = builder.region_id();
    let version_control = Arc::new(builder.build());
    let provider = Provider::raft_engine_provider(region_id.as_u64());

    let wal_entry = WalEntry {
        mutations: vec![poison_mutation(1)],
        bulk_entries: vec![],
    };
    let reader = mock_wal_reader(&provider, region_id, vec![(1, wal_entry)]);

    let last_entry_id = replay_memtable(
        &provider,
        reader,
        region_id,
        0,
        &version_control,
        false,
        |_, _, _| async { Ok(()) }.boxed(),
    )
    .await
    .unwrap();

    // Replay reports success, so the region opens normally.
    assert_eq!(1, last_entry_id);
    // The committed sequence covers the durable row.
    assert_eq!(1, version_control.committed_sequence());
    // But the row from the WAL vanished silently.
    assert!(
        version_control
            .current()
            .version
            .memtables
            .mutable
            .is_empty()
    );
}
```

</details>

1. A memtable apply failure during replay/catchup fails the region open **loudly** instead of opening with silently missing rows. Flipping fail-open to fail-closed is a real tradeoff — a deterministic poison entry would then block the open — so it should come with the existing `skip_wal_replay` documented as the operator escape hatch; the tradeoff should be a conscious decision, not a side effect.
2. `rows_replayed` counts only successfully applied rows.
3. Stats always agree with memtable contents: either count the applied prefix on failure, or make per-mutation apply all-or-nothing.
4. A post-WAL apply failure maps to an indeterminate/internal error class ("outcome unknown"), never a definitive-rejection class — or is prevented outright by lifting the fallible work (primary key encoding, type checks, bulk conversion) ahead of the WAL write, which is the direction the TODO in `time_series.rs:235-237` already points to. The sparse path needs an equivalent admission gate since it skips `check_schema`.

### What did you see instead?

- `replay_memtable` returns `Ok` and the region opens with rows missing and no trace; `rows_replayed` includes them.
- A failed mutation can leave an applied, readable prefix whose rows are absent from the stats that `ranges()` uses for pruning.
- Clients receive `InvalidArguments`-class rejections for mutations that remain durable in the WAL, with the final outcome decided by a flush/restart race.

Any (code-level issue, verified on macOS ARM)

main (`943eee852`); long-standing behavior — #8808 removed the known production trigger of #8806, this issue tracks the backstop behavior and accounting it exposed.

```bash
# The point is the absence of output: the silent-drop scenario produces no
# log line at all. Four repro tests (at d90cca4b7) pass, asserting the
# current behavior:
Summary [ 0.049s] 4 tests run: 4 passed, 1104 skipped
```

changshenhan:
## Summary

During region open, a WAL entry that could not be applied to the memtable was silently skipped: the rows stayed acknowledged in the WAL but never made it into memory, and `rows_replayed` counted them anyway. This PR makes replay fail-closed and fixes the two stats inconsistencies that let such failures go unnoticed.

### Replay fails closed (`src/mito2/src/region/opener.rs`)

`replay_memtable` previously pushed each replayed mutation/bulk with `OptionOutputTx::none()`, discarding the apply result. Now each entry gets a real oneshot channel and, after dropping the write context to flush the notifiers, any apply error or lost result fails the region open. `rows_replayed` counts only entries that applied successfully. Recovery is either fixing the cause or opening the region with `skip_wal_replay`.

### Stats stay consistent with memtable contents

- `TimeSeriesMemtable::write` (`memtable/time_series.rs`): on a mid-iteration failure it now counts the already-applied prefix, so `ranges()` cannot mis-prune rows that are present.
- `BulkMemtable::write_bulk` (`memtable/bulk.rs`): stats are updated before a compaction error propagates, since the fragment is already buffered in `unordered_part`.

### Apply errors after the WAL write map to Internal (`region_write_ctx.rs`, `error.rs`)

An error raised while applying an already-durable mutation cannot guarantee the write is rejected — the row may still exist after a restart. `map_apply_error` wraps `InvalidArguments` in the new `MemtableApply` variant (`StatusCode::Internal`), so callers never mistake an unknown-durability write for a definitive rejection.

## Tests

- `test_replay_fails_closed_on_mutation_failing_memtable_write`: a WAL entry whose row cannot be encoded into the schema fails the region open (previously it opened "successfully" with the row missing).
- `test_replay_applies_valid_mutation`: a valid entry still replays, publishes the sequence, and lands in the mutable memtable.
- Full `mito2` library suite: 1131 passed, 0 failed.

Closes #8809

### DIFF
--- a/src/mito2/src/region/opener.rs
+++ b/src/mito2/src/region/opener.rs
@@ -43,7 +43,7 @@ use store_api::metadata::{
 use store_api::region_engine::RegionRole;
 use store_api::region_request::{PathType, RegionRequirements};
 use store_api::storage::{ColumnId, RegionId};
-use tokio::sync::Semaphore;
+use tokio::sync::{Semaphore, oneshot};
 
 use crate::access_layer::AccessLayer;
 use crate::cache::file_cache::{FileCache, FileType, IndexKey};
@@ -69,7 +69,7 @@ use crate::region::{
     RegionStats,
 };
 use crate::region_write_ctx::RegionWriteCtx;
-use crate::request::OptionOutputTx;
+use crate::request::{OptionOutputTx, OutputTx};
 use crate::schedule::scheduler::SchedulerRef;
 use crate::series_index::{IndexFilePurger, load_version_control};
 use crate::sst::FormatType;
@@ -940,17 +940,26 @@ where
             // For WAL replay, we don't need to track the write bytes rate.
             None,
         );
+        // Replay must be fail-closed: an entry that cannot be applied has to
+        // fail the region open instead of silently dropping the rows. The
+        // write context only surfaces apply failures through its output
+        // senders (the notifiers deliver on drop), so hand each replayed
+        // mutation a real oneshot channel and collect the results afterwards.
+        let mut entry_rows = 0usize;
+        let mut replay_results = Vec::new();
         for mutation in entry.mutations {
-            rows_replayed += mutation
+            entry_rows += mutation
                 .rows
                 .as_ref()
                 .map(|rows| rows.rows.len())
                 .unwrap_or(0);
+            let (tx, rx) = oneshot::channel();
+            replay_results.push(rx);
             region_write_ctx.push_mutation(
                 mutation.op_type,
                 mutation.rows,
                 mutation.write_hint,
-                OptionOutputTx::none(),
+                OptionOutputTx::new(Some(OutputTx::new(tx))),
                 // We should respect the sequence in WAL during replay.
                 Some(mutation.sequence),
                 false,
@@ -968,12 +977,14 @@ where
             if region_metadata.primary_key_encoding != PrimaryKeyEncoding::Sparse {
                 part.fill_missing_columns(&region_metadata)?;
             }
-            rows_replayed += part.num_rows();
+            entry_rows += part.num_rows();
             // During replay, we should adopt the sequence from WAL.
             let bulk_sequence_from_wal = part.sequence;
+            let (tx, rx) = oneshot::channel();
+            replay_results.push(rx);
             ensure!(
                 region_write_ctx.push_bulk(
-                    OptionOutputTx::none(),
+                    OptionOutputTx::new(Some(OutputTx::new(tx))),
                     part,
                     Some(bulk_sequence_from_wal)
                 ),
@@ -991,6 +1002,42 @@ where
         // Publish the replayed sequences only after all rows (including bulk
         // parts) are installed, matching the write path ordering.
         region_write_ctx.publish_sequence_and_entry_id();
+
+        // Flush the notifiers by dropping the write context, then collect the
+        // apply results. Any failure fails the region open (fail-closed): the
+        // WAL entry stays durable, so opening "successfully" with the rows
+        // missing would silently lose acknowledged data. The operator must fix
+        // the underlying cause, or open the region with `skip_wal_replay`.
+        drop(region_write_ctx);
+        for rx in replay_results {
+            match rx.await {
+                Ok(Ok(_)) => {}
+                Ok(Err(err)) => {
+                    error!(
+                        "Failed to apply a replayed WAL entry (entry id: {}, region id: {}) to the memtable: {}. Failing the region open to avoid silently losing acknowledged rows; recover by fixing the cause or opening the region with `skip_wal_replay`.",
+                        entry_id, region_id, err
+                    );
+                    return Err(err);
+                }
+                // The notifier always sends before drop; a closed channel means
+                // the result is unknown, which must also fail the open.
+                Err(_) => {
+                    error!(
+                        "Lost the apply result for a replayed WAL entry (entry id: {}, region id: {}). Failing the region open; recover by opening the region with `skip_wal_replay`.",
+                        entry_id, region_id
+                    );
+                    return RegionCorruptedSnafu {
+                        region_id,
+                        reason: format!(
+                            "apply result channel closed for replayed entry {entry_id}"
+                        ),
+                    }
+                    .fail();
+                }
+            }
+        }
+        // Only rows from entries that applied successfully are counted.
+        rows_replayed += entry_rows;
     }
 
     // TODO(weny): We need to update `flushed_entry_id` in the region manifest
@@ -1399,6 +1446,9 @@ mod tests {
     use std::collections::HashMap;
     use std::sync::Arc;
 
+    use api::v1::helper::{tag_column_schema, time_index_column_schema};
+    use api::v1::value::ValueData;
+    use api::v1::{self, ColumnDataType, Mutation, OpType, Row, Rows, WalEntry};
     use arrow_schema::extension::ExtensionType;
     use common_base::readable_size::ReadableSize;
     use common_error::ext::WhateverResult;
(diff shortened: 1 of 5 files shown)


changshenhan:
Fixed all four CI failures flagged on the previous commit:

- **DCO** — added `Signed-off-by` to the commit.
- **typos** — fixed a message typo in `memtable/bulk.rs` ("mis-prune" → "prune them away").
- **clippy** — replaced a manual loop counter with `.enumerate()` in `memtable/time_series.rs`; `MemtableApplySnafu` is now constructed via `IntoError::into_error` in `region_write_ctx.rs`.
- **rustfmt** — fixed import ordering.

Verified locally before pushing: `cargo check -p mito2 --all-targets` clean, `cargo fmt -p mito2 --check` clean, and the full `mito2` test suite passes (1131 passed, 0 failed).

The workflows for the re-pushed commit appear to be waiting on approval — could you approve and run them? Thanks!

evenyag (commented):
Thanks for working on this. The fail-closed direction looks right. I left a few inline comments, mainly about observing apply errors directly and checking them before publishing the replay watermark.

src/mito2/src/region/opener.rs
```diff
@@ -861,17 +861,26 @@ where
...
+        let mut entry_rows = 0usize;
+        let mut replay_results = Vec::new();
         for mutation in entry.mutations {
-            rows_replayed += mutation
+            entry_rows += mutation
                 .rows
                 .as_ref()
                 .map(|rows| rows.rows.len())
                 .unwrap_or(0);
+            let (tx, rx) = oneshot::channel();
```
evenyag, line 877: Do we need channels here? Replay has a single synchronous consumer, while `OutputTx` is primarily client-notification plumbing. Could `write_memtable()` and `write_bulk()` return the first apply error directly while still recording individual errors in their notifiers? Replay could then continue using `OptionOutputTx::none()` and avoid per-mutation channels and correctness depending on `WriteNotify::drop()`.

src/mito2/src/region/opener.rs
```diff
@@ -911,6 +922,42 @@ where
         // Publish the replayed sequences only after all rows (including bulk
         // parts) are installed, matching the write path ordering.
         region_write_ctx.publish_sequence_and_entry_id();
```
evenyag, line 924: I think the apply result must be checked before publishing the replay watermark. `replay_memtable()` is also used for catch-up with an existing `VersionControl`. If application fails before inserting any rows, this advances `last_entry_id` while leaving the memtable empty. A subsequent catch-up may not reopen the region and could start after the failed entry, skipping it. Could we publish only after both mutation and bulk application succeed? It would also be useful for the failure test to assert that `committed_sequence` and `last_entry_id` remain unchanged.

src/mito2/src/region_write_ctx.rs
```diff
@@ -209,6 +211,19 @@ impl RegionWriteCtx {
         self.provider == Provider::Noop || self.version.options.skip_wal
     }
 
+    /// Maps an error raised while applying a mutation to the memtable after the
+    /// WAL write. Such an error cannot guarantee the mutation is not durable, so
+    /// it must never be surfaced as a definitive rejection (e.g.
+    /// `InvalidArguments`): the row may still exist after a restart, or be gone
+    /// after a flush. The caller sees an internal error, i.e. the outcome of the
+    /// write is unknown.
+    fn map_apply_error(err: Error) -> Error {
```
evenyag, line 220: I do not think we need this error remapping in this PR. A memtable `InvalidArguments` error after admission should be treated as an invariant violation rather than remapped here. `RegionWriteCtx` also handles `skip_wal` and `Provider::Noop`, where the assumption that the mutation was already written to WAL is incorrect. Could we remove `MemtableApply` and `map_apply_error` and keep this change focused on fail-closed replay?

PR #8926, opened 2026-08-21
feat(mito2): add series index searcher

evenyag:

This PR adds a series-index searcher for metric regions so callers can discover matching `(table_id, tsid)` pairs directly from aggregate series-index Parquet files.

The searcher:

- translates supported tag predicates and query time ranges into index filters;
- prunes row groups using Parquet statistics;
- projects only the series identifiers and predicate columns;
- incrementally fetches byte ranges through the Parquet push decoder;
- returns sorted, deduplicated series IDs in bounded batches.

It also shares the metric-series identifier types and tag-filter extraction with the existing candidate-series path, and exposes reusable Parquet statistics helpers.

The series-index file schema and persisted data layout are unchanged.

Validation: `cargo nextest run -p mito2` (1,132 passed).

## PR Checklist

### DIFF
--- a/src/mito2/src/series_index/searcher.rs
+++ b/src/mito2/src/series_index/searcher.rs
@@ -13,11 +13,12 @@
 // limitations under the License.
 
 use std::collections::HashSet;
+use std::ops::Range;
 use std::sync::Arc;
 
 use api::v1::SemanticType;
 use async_stream::try_stream;
-use common_datasource::file_format::parquet::LazyParquetFileReader;
+use bytes::Bytes;
 use common_recordbatch::filter::SimpleFilterEvaluator;
 use common_time::range::TimestampRange;
 use datafusion_common::pruning::PruningStatistics;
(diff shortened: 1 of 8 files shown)


src/mito2/src/series_index/searcher.rs
```diff
@@ -0,0 +1,643 @@
...
+            })?;
+    let unit = timestamp_type.unit();
+    let mut exprs = Vec::with_capacity(2);
+    if let Some(start) = time_range
+        .start()
+        .and_then(|start| start.convert_to_ceil(unit))
+    {
+        exprs.push(col(MAX_TS_COLUMN).gt_eq(lit(start.value())));
+    }
+    if let Some(end) = time_range.end().and_then(|end| end.convert_to_ceil(unit)) {
```
v0y4g3r, line 272: We should use `convert_to` instead of `convert_to` on time range end to avoid precision loss.

src/mito2/src/series_index/searcher.rs
```diff
@@ -0,0 +1,643 @@
...
+    filters: &[SimpleFilterEvaluator],
+) -> Result<ProjectionMask> {
+    let mut indices = HashSet::new();
+    for name in [TABLE_ID_COLUMN, TSID_COLUMN] {
+        let index = arrow_schema
+            .index_of(name)
+            .ok()
+            .context(InvalidRecordBatchSnafu {
+                reason: format!("series index is missing internal column {name}"),
+            })?;
```
v0y4g3r:
```suggestion
            .with_context(|| InvalidRecordBatchSnafu {
                reason: format!("series index is missing internal column {name}"),
            })?;
```

src/mito2/src/series_index/searcher.rs
```diff
@@ -0,0 +1,643 @@
...
+        indices.insert(index);
+    }
+    for filter in filters {
+        let index =
```
v0y4g3r: Looks like `arrow_schema` comes from parquet file, what if `arrow_schema` lacks of some newly added primary key columns in filters? `projection_mask` will return an error and fail current search? Does this break the query?

src/mito2/src/series_index/searcher.rs
```diff
@@ -0,0 +1,643 @@
...
+            arrow_schema
+                .index_of(filter.column_name())
+                .ok()
+                .context(InvalidRecordBatchSnafu {
```
v0y4g3r: Use `with_context`

src/mito2/src/series_index/searcher.rs
```diff
@@ -0,0 +1,643 @@
...
+
+fn column<'a>(
+    batch: &'a datatypes::arrow::record_batch::RecordBatch,
+    name: &str,
+) -> Result<&'a ArrayRef> {
+    let index = batch
+        .schema()
+        .index_of(name)
+        .ok()
+        .context(InvalidRecordBatchSnafu {
```
v0y4g3r: Use `with_context`

src/mito2/src/series_index/searcher.rs
evenyag: The current code intentionally uses `convert_to_ceil` for both bounds. For a half-open query range `[start, end)`, overlap requires `series_max >= ceil(start)` and `series_min < ceil(end)` in the index timestamp unit. Using floor-style `convert_to` for the end can cause a false negative: with millisecond index timestamps and an end of 30.001 ms, flooring to 30 ms would exclude a series whose minimum is 30 ms, while ceiling to 31 ms keeps it. I added comments explaining both conversions and a finer-to-coarser precision regression test. (The two method names in the original comment are identical, so please let me know if you intended a different change.)

v0y4g3r (approved):
LGTM

PR #8927, opened 2026-08-21
fix(servers): anchor pending-rows flush to batch creation time

### LINKED ISSUE #8641: prom_store pending-rows batcher: flush tick misaligns with batch creation, waits up to 2x flush_interval
## Problem

In `start_worker` (`src/servers/src/pending_rows_batcher.rs`), the flush trigger combines:

- `tokio::time::interval(flush_interval)` — ticks are aligned to **worker creation time** (and the first tick fires immediately, always wasted on an empty batch);
- an age check `batch.created_at.elapsed() >= flush_interval` — `created_at` is recorded at the **first Submit** processed by the worker.

Because the tick schedule and batch creation time are decoupled, a batch created at `t ∈ (n·fi, (n+1)·fi)` always fails the age check at tick `(n+1)·fi` and is only flushed at tick `(n+2)·fi`. So **every batch waits (flush_interval, 2·flush_interval) before its flush even starts** — not just the first batch of a newly spawned worker.

## Impact

With synchronous batching (the default, `PENDING_ROWS_BATCH_SYNC` defaults to true), `submit()` blocks on the flush result, so HTTP remote-write request latency is:

```
pre-processing (DDL/align) + up to ~2×flush_interval queueing + flush duration
```

Any HTTP timeout budget sized as `flush_interval + margin` (e.g. the `flush_interval + 1s` fallback introduced in #8639) can therefore expire before the flush starts whenever `flush_interval > margin`, causing avoidable 504s and client retries on `/v1/prometheus/write`. High-throughput workloads are partially masked by the `max_batch_rows` early flush; low-traffic periods are the most exposed.

## Possible fixes

1. Schedule the flush relative to batch creation instead of the worker's fixed tick — e.g. arm `sleep_until(batch.created_at + flush_interval)` when a batch is first created. This makes the queueing delay a stable `flush_interval`.
2. Alternatively, keep the tick scheme but size dependent timeout budgets for the real worst case: `2×flush_interval + flush margin`.

## Context

Surfaced by review discussion on #8639: https://github.com/GreptimeTeam/greptimedb/pull/8639#discussion_r3651982483

changshenhan:
## Summary

The pending-rows worker flushes a batch on a **worker-aligned** `tokio::time::interval(flush_interval)` tick, but the age check reads `created_at` from the **first `Submit`**. Because the tick schedule and the batch creation time are decoupled, a batch created at `t ∈ (n·fi, (n+1)·fi)` always fails the age check at tick `(n+1)·fi` and is only flushed at `(n+2)·fi`. The queueing delay is therefore `(flush_interval, 2·flush_interval)`, not `flush_interval`.

With synchronous batching (`PENDING_ROWS_BATCH_SYNC`, the default), `submit()` blocks on the flush result, so `/v1/prometheus/write` latency includes up to ~2× the interval of pure queueing. A timeout budget sized as `flush_interval + margin` (e.g. the `flush_interval + 1s` fallback from #8639) can expire before the flush even starts, causing avoidable 504s and client retries. Low-traffic periods — where `max_batch_rows` never triggers an early flush — are the most exposed.

### Fix (option 1 from the issue)

Replace the worker-aligned `interval` with a deadline anchored to `batch.created_at + flush_interval`, recomputed at the top of each loop iteration from the batch's **original** creation time. Two properties follow:

- The queueing delay becomes a stable `flush_interval` regardless of when between ticks the batch was created.
- The deadline does **not** slide as new submissions arrive (it is derived from `created_at`, not `now()`), preserving the "flush `flush_interval` after the batch was first seen" semantics.

The change is local to `start_worker` (`src/servers/src/pending_rows_batcher.rs`), +31/−15.

## Testing

Ran the `pending_rows_batcher` module suite: **39 passed, 0 failed**. No regression to the existing flush/notification/partition tests.

A dedicated wall-clock timing test (asserting the flush fires at `fi` rather than `2·fi`) would need to drive the worker with mocked `PartitionRuleManager`/`CatalogManager` under `tokio::time::pause`, which this crate has no existing scaffolding for (the flush dependencies are concrete manager refs, not the testable `PhysicalFlush*Provider` traits). Given the change is a single local transformation with clear semantics, I kept the PR to the implementation plus the existing suite — happy to add the timing test if you'd prefer, e.g. as a follow-up that threads the flush dependencies through the existing trait adapters.

Closes #8641

### DIFF
--- a/src/servers/src/pending_rows_batcher.rs
+++ b/src/servers/src/pending_rows_batcher.rs
@@ -941,14 +941,25 @@ fn start_worker(
     flush_semaphore: Arc<Semaphore>,
 ) {
     tokio::spawn(async move {
-        let mut batch = None;
-        let mut interval = tokio::time::interval(flush_interval);
+        let mut batch: Option<PendingBatch> = None;
         let mut shutdown_rx = shutdown.subscribe();
         let idle_deadline = tokio::time::Instant::now() + worker_idle_timeout;
         let idle_timer = tokio::time::sleep_until(idle_deadline);
         tokio::pin!(idle_timer);
 
         loop {
+            // Anchor the flush deadline to the batch's creation time rather than
+            // to a worker-aligned tick. A batch created at any point between two
+            // ticks used to wait for the next tick — up to almost 2× the flush
+            // interval after its creation — before its flush even started (#8641).
+            // Anchoring on `created_at` makes the queueing delay a stable
+            // `flush_interval` and, being recomputed every iteration from the
+            // batch's original creation time, does not slide as new submissions
+            // arrive.
+            let flush_deadline = batch
+                .as_ref()
+                .map(|batch| tokio::time::Instant::from(batch.created_at) + flush_interval);
+
             tokio::select! {
                 cmd = rx.recv() => {
                     match cmd {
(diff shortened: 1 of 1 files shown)


killme2008:
@v0y4g3r Please take a look

I am not sure if it's a valid issue here.

changshenhan:
Closing as superseded: #8802 (merged 2026-09-08) implements the same fix for #8641 — anchoring the pending-rows flush deadline to the batch's first submission instead of the worker-aligned tick. The approach there (resetting a pinned `sleep` on first `Submit` under `if batch.is_some()`) is functionally equivalent to this PR's `created_at + flush_interval` deadline recomputation, and the added `Ack`-based test synchronization is a nice determinism improvement. Nothing further to add — thanks @v0y4g3r.

PR #8928, opened 2026-08-21
refactor(json2): optimize JSON2 building without auto-expanded paths

MichaelScofield:

for `max_auto_expanded_paths = 0`

no auto expanding, no buffered values, build array directly to accelerate flushing

## PR Checklist

### DIFF
--- a/src/datatypes/src/vectors/json/builder.rs
+++ b/src/datatypes/src/vectors/json/builder.rs
@@ -24,16 +24,17 @@ use snafu::{ResultExt, ensure};
 
 use crate::data_type::ConcreteDataType;
 use crate::error::{
-    ArrowComputeSnafu, Result, TryFromValueSnafu, UnexpectedSnafu, UnsupportedOperationSnafu,
+    ArrowComputeSnafu, Result, TryFromValueSnafu, UnexpectedSnafu, UnimplementedSnafu,
+    UnsupportedOperationSnafu,
 };
 use crate::extension::json::JSON2_REMAINDER_FIELD_NAME;
-use crate::json::value::{JsonNumber, JsonVariant, encode_json_variant};
+use crate::json::value::{JsonNumber, JsonVariant, JsonVariantRef, encode_json_variant};
 use crate::json::{JSON2_MAX_STRUCTURED_DEPTH, JsonSettings};
 use crate::prelude::{ValueRef, Vector, VectorRef};
 use crate::types::StructType;
 use crate::types::json_type::{JsonNativeType, is_include};
 use crate::value::{ListValue, StructValue, StructValueRef, Value};
-use crate::vectors::json::variant::{append_json_variant, variant_field};
+use crate::vectors::json::variant::{append_json_variant, append_json_variant_ref, variant_field};
 use crate::vectors::{Helper, MutableVector, StructVectorBuilder};
 
 type JsonObjectValue = BTreeMap<String, JsonVariant>;
@@ -44,16 +45,24 @@ type JsonObjectValue = BTreeMap<String, JsonVariant>;
 /// Auto-expanding mode always materializes type-hinted paths, selects up to
 /// `max_auto_expanded_paths` compatible unhinted leaf paths by frequency, and stores
 /// conflicting or unselected paths in the Variant remainder field.
-#[derive(Clone)]
 pub(crate) struct JsonVectorBuilder {
     state: JsonVectorBuilderState,
 }
 
-#[derive(Clone)]
 enum JsonVectorBuilderState {
     Legacy {
         merged_type: JsonNativeType,
         values: Vec<JsonVariant>,
+    },
+    ExplicitOnly {
+        /// Paths declared by type hints and stored as dedicated Struct fields.
+        explicit_type: JsonNativeType,
+        /// Concrete Struct type used to append explicit values without buffering rows.
+        struct_type: StructType,
+        /// Builder for values selected by the explicit type hints.
+        explicit: StructVectorBuilder,
+        /// Builder for all values outside the explicit type hints.
+        remainder: VariantArrayBuilder,
     },
     AutoExpanding {
         /// Paths declared by type hints and always stored as dedicated Struct fields.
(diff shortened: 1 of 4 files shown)


MichaelScofield:
@codex review

src/datatypes/src/vectors/json/builder.rs
```diff
@@ -615,7 +834,9 @@ impl MutableVector for JsonVectorBuilder {
     }
 
     fn to_vector_cloned(&self) -> VectorRef {
-        self.clone().to_vector()
+        self.state
+            .try_build_cloned()
+            .unwrap_or_else(|e| panic!("{:?}", e))
```
sunng87: > [`.agents/architecture-invariants.md` §4](../blob/main/.agents/architecture-invariants.md) asks for "no panics in non-test code" — this converts the `Unimplemented` error from `try_build_cloned` into a `panic!` for any `ExplicitOnly` builder.

I traced the callers to check how reachable this is today:

- `TimeSeriesMemtable`'s `FieldBuilder::finish_cloned` (`src/mito2/src/memtable/builder.rs:85`) does call `to_vector_cloned()` generically, but its JSON2 builders come from `create_mutable_vector` → `JsonVectorBuilder::new`, i.e. the `Legacy` state — which clones fine.
- flow's `to_vector_cloned` usage is `#[cfg(test)]` only.

So currently this is unreachable in production. But once `with_settings` is wired into `JsonType::create_mutable_vector` (the follow-up this PR is staged for), a `max_auto_expanded_paths = 0` column flowing through `TimeSeriesMemtable` will hit this panic at snapshot. Two asks:

1. Link the follow-up/tracking issue from the comment in `try_build_cloned` above, so the constraint is greppable from the code.
2. In the wiring PR, either give `VariantArrayBuilder` non-consuming snapshot support, or make sure the zero-budget mode cannot be constructed for builders that reach a `finish_cloned`/`to_vector_cloned` path.

src/datatypes/src/vectors/json/variant.rs
```diff
@@ -110,6 +110,52 @@ pub(super) fn append_json_variant(
     Ok(())
 }
 
+pub(super) fn append_json_variant_ref(
```
sunng87, line 113: This duplicates all nine arms of `append_json_variant` above (modulo `String(String)` vs `String(&str)`), and the same owned/ref twin pattern repeats in `builder.rs` (`json_variant_ref_into_value_ref` vs `json_variant_into_value`, `remainder_ref` vs `split_to_explicit`).

The subtle cases — the `i64::try_from` large-u64 fallback and the non-finite-float-as-`"NaN"` handling — now live in two places each and can drift apart.

One note before reaching for the obvious fix: delegating the owned version through `JsonVariant::as_ref()` is *not* the answer — `as_ref()` allocates a fresh `Vec`/`BTreeMap` for arrays and objects, which would defeat the point of this zero-copy path. A macro that expands both enums' arms from one template (the serde/arrow approach to owned/borrowed twins), or a small trait abstracting the two structural arms, would keep a single source of truth without the allocation.

(Separately: `json_variant_ref_into_value_ref` omitting the `List`/`Binary` arms looks unreachable by construction since `validate_type_hints` only admits scalars — a one-line comment there would spare the next reader the same check.)

src/datatypes/src/vectors/json/builder.rs
```diff
@@ -125,6 +176,22 @@ impl JsonVectorBuilderState {
...
+                struct_type,
+                explicit,
+                remainder,
+            } => {
+                if value.is_null() {
+                    explicit.push_null();
+                    remainder.append_null();
+                } else {
+                    let (value, rest) =
+                        split_to_explicit_ref(value.variant(), explicit_type, struct_type.clone())?;
```
sunng87: Minor API-shape nit — I checked the cost before flagging it: `StructType` is `Arc`-backed (`fields: Arc<Vec<StructField>>`), so this per-row `clone()` — and the one per nested object in `json_object_ref_into_struct_value_ref` — is just a refcount bump, not a schema deep-copy. No perf concern.

If you'd like the hot path to read that way: `split_to_explicit_ref` / `json_object_ref_into_struct_value_ref` could take `&StructType` and clone exactly once at the single point where ownership is actually required (the `StructValueRef::RefList { fields }` construction).

sunng87 (commented):
Standards-focused pass (style guide + architecture invariants + code-health smells), all findings verified against the tree: `cargo nextest run -p datatypes json` 63/63 green, `cargo clippy -p datatypes --all-targets --all-features -D warnings` and `cargo fmt --check` clean.

Three inline notes: one invariant-adjacent (the `panic!` conversion), two code-health nits (owned/ref duplication, `&StructType` vs by-value).

src/datatypes/src/vectors/json/builder.rs
MichaelScofield: I'm planning to add clone to upstream `VariantArrayBuilder`. until then, it's ok here, at least the old memtable is not used by default

src/datatypes/src/vectors/json/variant.rs
MichaelScofield: the repetition just show twice, don't worth the complication that brought by macro I think

sunng87: Fair enough — owned/borrowed twin macros are genuinely messy, and twice is borderline. If the drift risk is worth anything, the cheap middle ground is an equivalence test: run `append_json_variant` and `append_json_variant_ref` over the same corpus (including the `PosInt > i64::MAX` and non-finite-float cases, which are the subtle ones) and assert byte-identical output. No abstraction, just a lock on the invariant that the two stay in sync. No objection if you'd rather leave it as-is, though.

(Also fine to note the same owned/ref pairing appears twice more in `builder.rs` — `json_variant_ref_into_value_ref` / `json_variant_into_value` and `remainder_ref` / `split_to_explicit` — so the count is five functions, not two. Same conclusion either way.)

src/datatypes/src/vectors/json/builder.rs
sunng87: We should at least void `panic!` in production code.

MichaelScofield: `VariantArrayBuilder` can't clone, so here we can only make it noop

MichaelScofield:
@sunng87 PTAL

sunng87 (approved):

PR #8896, opened 2026-08-17
feat(meta): record catalog and database reconciliation events

dhruvxvaishnav:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

Part of #8861.

This is the remaining catalog and database stage after #8935 and #8941.

## What's changed and what's your intention?

This PR adds two independently filterable procedure event types:

- `reconcile_catalog`
- `reconcile_database`

Both use the shared five-column locator contract: `catalog_name`, `schema_name`, `table_name`, `table_id`, and `physical_table_id`. Catalog events populate only the catalog name. Database events populate the catalog and schema names. The unused locator columns remain null.

Submitted events contain a versioned, bounded description of the submitted intent: resolve strategy, fail-fast behavior, parallelism, and, for database reconciliation, whether it is a subprocedure. Terminal events keep result fields in the versioned JSON payload and report bounded summaries from the existing reconciliation metrics. Full resolved column definitions are not stored. Other lifecycle triggers use a null reconciliation payload while retaining the standard procedure trigger envelope.

The result metrics are best-effort observations for the current process and reset after recovery, consistent with the merged logical-table event implementation. This PR does not change procedure persistence, scheduling, recovery behavior, or reconciliation execution.

The existing event envelope provides parent and child correlation. The integration test verifies that a catalog `ChildSubmitted` event references the database procedure whose own `Submitted` and terminal events are recorded.

Configuration examples now list the two event types, and `config/config.md` was regenerated from those sources.

Coverage includes:

- the exact shared locator schema and nullable locator rows;
- versioned submitted and terminal payload contracts;
- catalog and database event filtering;
- lifecycle and partial terminal payload behavior;
- root and child database submitted intent;
- recovery-compatible loading with volatile metrics reset;
- catalog-to-database parent and child correlation in the integration test.

This adds no public API, wire-protocol, schema, or persisted-state compatibility change.

Validation completed locally:

- `cargo nextest run --retries 0 -p common-meta reconciliation::event::tests` (3 passed)
- `cargo nextest run --retries 0 -p common-meta reconciliation::reconcile_catalog::tests` (5 passed)
- `cargo nextest run --retries 0 -p common-meta reconciliation::reconcile_database::tests` (5 passed)
- `cargo nextest run --retries 0 -p tests-integration --test main reconciliation_event::test_catalog_and_database_reconciliation_events` (1 passed)
[...]

Issue #8861, opened 2026-08-12
Record lifecycle events for reconciliation procedures


WenyXu:
### What type of enhancement is this?

- Tech debt reduction

### What does the enhancement do?

Reconciliation procedures currently do not emit lifecycle events.

`ReconcileCatalogProcedure`, `ReconcileDatabaseProcedure`, `ReconcileTableProcedure`, and `ReconcileLogicalTablesProcedure` do not override `Procedure::event()`, so the default implementation returns `None`.

Add reconciliation procedure events following the existing procedure-event conventions. Events should expose reconciliation activity in the events table while reusing the framework-owned lifecycle envelope (`procedure_id`, `procedure_state`, `procedure_error`, and `procedure_trigger`).

### Implementation challenges

- Define a stable reconciliation event type and schema, following existing procedure event patterns.
- Include applicable canonical locators: catalog, schema, table name, table ID, and physical table ID; keep unavailable dimensions nullable.
- Implement and filter `Procedure::event()` for catalog, database, table, and logical-table reconciliation procedures.
- Use a bounded, versioned payload for reconciliation-specific metadata.
- Add event-contract tests for schema, rows, filtering, and lifecycle events; cover parent and child reconciliation procedures.
- Update event-recorder configuration documentation if it lists supported event types explicitly.

dhruvxvaishnav:
Hi @WenyXu, I’d like to work on this if it is available. I reviewed the four reconciliation procedures and confirmed that they currently use the default Procedure::event() implementation, so no reconciliation lifecycle events are produced.
My proposed approach is to add a stable, versioned reconciliation event payload, implement filtered events for the catalog, database, table, and logical-table procedures using nullable canonical locators, and add event-contract tests covering lifecycle states and parent/child procedures.
Would this scope and direction be appropriate?

WenyXu:
Hi @dhruvxvaishnav, do you have a detailed plan for this issue?

dhruvxvaishnav:
Hi @WenyXu, yes, I’m preparing a more detailed plan based on the existing procedure-event patterns and the reconciliation procedure contexts. I’ll post the proposed event schema, per-procedure mappings, filtering behaviour, and test strategy here shortly.

dhruvxvaishnav:
@WenyXu, here is my proposed implementation plan.

### 1. Event representation

I plan to add a reconciliation event module under `common-meta`, following the existing database/table DDL event patterns.

I propose four independently filterable event types:

[...]

PR #8935, opened 2026-08-21
feat(meta): record physical table reconciliation events

dhruvxvaishnav:

Part of #8861.

This is now the first PR in the maintainer-requested child-first order: physical table (#8935), logical tables (#8941), then catalog/database (#8896). The branch is based directly on current `main` and contains no catalog/database or logical-table procedure instrumentation.

The configuration examples and generated `config/config.md` are updated. The generated file was produced with the repository's exact pinned `make config-docs` command using Docker inside Ubuntu WSL2; it was not hand-edited.

This PR adds the independently filterable `reconcile_table` lifecycle event for physical-table reconciliation procedures and introduces the shared five-column nullable locator schema needed by the approved four-event design:

- `catalog_name`
- `schema_name`
- `table_name`
- `table_id`
- `physical_table_id`

For a physical table, its identity is stored in `table_id`; `physical_table_id` is null because that column is reserved for logical-to-physical correlation in #8941.

Submitted payload version 1 records only bounded intent:

- `resolve_strategy`
- `is_subprocedure`

Terminal payload version 1 records bounded complete or partial results:

- metadata consistency state
- resolution strategy actually applied
- resolved column count
- scanned and updated region counts
- whether MetaSrv table information was updated
- last completed phase

No resolved column definitions, `TableMeta`, routes, or other unbounded internal metadata are retained. Framework-owned fields continue to carry procedure state, errors, triggers, and correlation data.

A bounded result summary is persisted with `serde(default)` so older procedure JSON remains readable. Recovery clears only the phases that will be replayed. Partial failure summaries record work that completed before a later fallible operation without incorrectly advancing `last_completed_phase`.

The integration test invokes table reconciliation directly instead of depending on catalog/database event assertions. Root and subprocedure payload variants, filtering, lifecycle triggers, recovery resets, partial results, and old persisted-state compatibility are covered by focused unit tests.

### Compatibility

This is additive and does not change a public API or wire protocol. Existing procedure type names and loaders are unchanged. The new persisted summary remains backward compatible through `serde(default)`.

## Validation

Passed locally after rebuilding this vertical diff on current `main`:

- `make config-docs` (Docker inside Ubuntu WSL2)
- `cargo fmt --all`
- `cargo nextest run --retries 0 -p common-meta reconciliation::event::tests`
- `cargo nextest run --retries 0 -p common-meta reconciliation::reconcile_table`
- `cargo nextest run --retries 0 -p tests-integration --test main reconciliation_event::test_table_reconciliation_events`
- `cargo clippy -p common-meta --all-targets -- -D warnings`
- `cargo clippy -p tests-integration --test main --no-deps -- -D warnings`
- `git diff --check`

The integration test initially encountered a transient Windows random-port bind failure before the cluster started; rerunning selected a new port and passed.

Not completed locally:

- Workspace-wide tests, workspace-wide Clippy, and `make check-udeps` were not run locally.
- No process-kill/restart compatibility integration scenario was run.

### AI assistance

OpenAI Codex assisted with codebase exploration, implementation support, test preparation, validation analysis, and submission preparation. I reviewed the change and remain responsible for understanding and maintaining it.

## PR Checklist


WenyXu:
@codex review

src/common/meta/src/reconciliation/reconcile_table/resolve_column_metadata.rs
```diff
@@ -126,6 +132,12 @@ impl State for ResolveColumnMetadata {
                 let region_ids =
                     resolve_column_metadatas_with_metasrv(&column_metadata, &self.region_metadata)?;
 
+                ctx.persistent_ctx.result_summary.record_resolved_columns(
+                    TableMetadataState::Inconsistent,
+                    Some(self.strategy),
+                    Some(column_metadata.len()),
+                );
```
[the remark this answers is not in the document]
dhruvxvaishnav: Fixed in `7fe7cd67b`. The summary now records `metadata_state = inconsistent` immediately after the consistency check fails, before either fallible resolver runs. Successful resolution still records the applied strategy, resolved-column count, and phase completion; resolver errors retain the known inconsistency while leaving `last_completed_phase` at `start`. The abort path records its applied strategy without marking resolution complete. Added focused partial terminal-payload coverage, and the common-meta reconciliation-table tests passed locally.

src/common/meta/src/reconciliation/reconcile_table/reconcile_regions.rs
```diff
@@ -126,6 +126,7 @@ impl State for ReconcileRegions {
             }
         }
 
+        let updated_region_count = sync_column_tsks.len();
```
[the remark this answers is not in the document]
dhruvxvaishnav: Fixed in `7fe7cd67b`. After `join_all`, the procedure now counts successful RPC results and stores that partial count before collecting the batch result, so a later RPC error does not erase completed side effects. The phase-completion marker remains after successful response validation. Focused unit and reconciliation integration tests passed locally.

WenyXu:
@codex review

src/common/meta/src/reconciliation/reconcile_table.rs
```diff
@@ -123,6 +247,8 @@ pub(crate) struct PersistentContext {
     pub(crate) physical_table_route: Option<PhysicalTableRouteValue>,
     // Whether the procedure is a subprocedure.
     pub(crate) is_subprocedure: bool,
+    #[serde(default)]
```
WenyXu: Adding `result_summary` changes the JSON persisted for an in-flight `ReconcileTableProcedure` through `dump()` / `from_json()`. The unit test removes the field from JSON generated by the current binary, so it does not cover a procedure persisted by the previous release and recovered after an upgrade. The repository requires compatibility-suite coverage for persisted-format changes. Could we add an upgrade case that leaves an in-flight reconciliation procedure on the old binary and verifies recovery on the new binary?

src/common/meta/src/reconciliation/reconcile_table/reconcile_regions.rs
```diff
@@ -126,10 +126,12 @@ impl State for ReconcileRegions {
             }
         }
 
-        let mut results = future::join_all(sync_column_tsks)
-            .await
-            .into_iter()
-            .collect::<Result<Vec<_>>>()?;
+        let results = future::join_all(sync_column_tsks).await;
```
WenyXu, line 127: The partial-summary test seeds `ReconcileTableResultSummary` directly, while the integration test only covers a successful single-region reconciliation. Neither executes the mixed SyncColumns outcome handled here: one region succeeds, another fails, and the Failed event must retain the successful `updated_region_count` without advancing `last_completed_phase`. Could we add a focused test for that mixed-outcome path?

src/common/meta/src/reconciliation/reconcile_table/reconciliation_start.rs
```diff
@@ -46,7 +46,7 @@ impl State for ReconciliationStart {
         procedure_ctx: &ProcedureContext,
     ) -> Result<(Box<dyn State>, Status)> {
         let table_id = ctx.table_id();
-        let table_name = ctx.table_name();
+        let table_name = ctx.table_name().clone();
```
WenyXu: This clone copies the catalog, schema, and table strings only to keep the name available across the new summary mutation. The name is subsequently used only for diagnostics, so short-lived `ctx.table_name()` reborrows avoid the allocation cleanly. The same applies to the clone in `reconcile_regions.rs:68`. Could we keep both as borrows?

src/common/meta/src/reconciliation/reconcile_table.rs
```diff
@@ -275,3 +469,372 @@ pub(crate) trait State: Sync + Send + Debug {
...
+    use crate::key::table_info::TableInfoValue;
+    use crate::key::test_utils::new_test_table_info;
+    use crate::reconciliation::reconcile_table::reconcile_regions::ReconcileRegions;
+    use crate::reconciliation::reconcile_table::reconciliation_end::ReconciliationEnd;
+    use crate::reconciliation::reconcile_table::resolve_column_metadata::ResolveColumnMetadata;
+    use crate::reconciliation::reconcile_table::update_table_info::UpdateTableInfo;
+    use crate::test_util::{MockDatanodeManager, new_ddl_context};
+
+    #[test]
+    fn table_event_hook_covers_root_child_filters_and_lifecycle_payloads() {
```
WenyXu: Could we split this test by event contract and keep a small `TableEventHarness` private to this test module? The current name is long because the body verifies four independent behaviors: root/child submitted payloads, lifecycle payloads, terminal payloads, and event filtering. A local harness can own the repeated default `EventContext` setup, while focused tests make failures easier to localize. I would avoid moving this helper into production because `EventContext` construction is owned by the procedure runner.

src/common/meta/src/reconciliation/event.rs
```diff
@@ -0,0 +1,346 @@
...
+    CATALOG_NAME_COLUMN, PHYSICAL_TABLE_ID_COLUMN, SCHEMA_NAME_COLUMN, TABLE_ID_COLUMN,
+    TABLE_NAME_COLUMN, column_schemas, nullable_string, nullable_value,
+};
+use serde::Serialize;
+use snafu::ResultExt;
+use store_api::storage::TableId;
+
+use crate::reconciliation::ResolveStrategy;
+
+pub(crate) const RECONCILE_TABLE_EVENT_TYPE: &str = "reconcile_table";
```
WenyXu, line 32: Could we add short Rustdoc for the crate-visible event constant and the locator/event constructors below? `ddl/event/table.rs` documents the analogous locator and event-builder contract, which makes the stable payload and lifecycle semantics easier to reuse from the upcoming reconciliation procedure types.

WenyXu (changes requested):
I left comments on persisted-procedure upgrade coverage, partial-failure event accounting, and keeping the reconciliation event tests focused. I also noted two small cleanup items in the new reporting code.

src/common/meta/src/reconciliation/reconcile_table/reconcile_regions.rs
dhruvxvaishnav: Addressed in bbd3a56. The focused test now executes ReconcileRegions with two SyncColumns requests through PartialSuccessDatanodeHandler, where one region succeeds and one fails. It asserts that the Failed event retains updated_region_count = 1 and keeps last_completed_phase at resolve_column_metadata. The focused common-meta reconciliation suite passed all 9 tests locally.

src/common/meta/src/reconciliation/reconcile_table.rs
dhruvxvaishnav: Addressed in bbd3a56. I added a compatibility-suite upgrade case for v1.0.2 and v1.1.4. The runner watches the old stage in etcd, captures the real persisted root ReconcileTable step, clones that exact old-binary message under a fresh procedure ID, and verifies that the current binary emits Recovered and Succeeded while preserving the empty legacy result summary. This avoids timing-based process termination. The new runner unit tests, runner formatting, and runner clippy passed locally. The Docker-backed upgrade execution is left to CI because the local WSL GreptimeDB link did not complete in a reasonable time.

src/common/meta/src/reconciliation/event.rs
dhruvxvaishnav: Addressed in bbd3a56. Added short Rustdoc for RECONCILE_TABLE_EVENT_TYPE, the physical-table locator constructor, and the submitted, terminal-result, and lifecycle event constructors. The comments describe the stable event name and bounded payload semantics.

src/common/meta/src/reconciliation/reconcile_table/reconciliation_start.rs
dhruvxvaishnav: Addressed in bbd3a56. Both owned TableName clones were removed. The diagnostic sites now use short-lived ctx.table_name() borrows after the summary mutations, so no allocation is needed and the borrow scopes remain compatible with mutable context access.

src/common/meta/src/reconciliation/reconcile_table.rs
dhruvxvaishnav: Addressed in bbd3a56. I added a private TableEventHarness in the test module and split the previous combined test into focused submitted-intent, non-terminal lifecycle, terminal-result, and event-filtering tests. The helper remains test-only and keeps EventContext construction with the procedure-runner contract.

src/common/meta/src/reconciliation/reconcile_table.rs
```diff
@@ -123,6 +265,8 @@ pub(crate) struct PersistentContext {
     pub(crate) physical_table_route: Option<PhysicalTableRouteValue>,
     // Whether the procedure is a subprocedure.
     pub(crate) is_subprocedure: bool,
+    #[serde(default)]
+    result_summary: ReconcileTableResultSummary,
```
WenyXu: Do we need this summary to survive recovery? It only enriches lifecycle events; procedure execution and retry decisions do not read it. Treating it as persisted state creates a durable JSON contract, recovery-specific reset logic, and the new old-procedure compatibility framework solely to test it. Could we move it into `VolatileContext` instead? After recovery, the terminal event can report only work performed since recovery, which seems acceptable for best-effort observability and removes this persistence-specific complexity.

dhruvxvaishnav: Addressed in `9ec1dc769`. `ReconcileTableResultSummary` now lives in `VolatileContext` and is excluded from `dump()` and `from_json()`. Recovery therefore starts with an empty best-effort summary, and terminal events report only work observed after recovery.

I also removed the recovery reset logic, the old-procedure compatibility runner extension, the upgrade case, and the extra runner dependencies that were needed only for persisted summary compatibility. The mixed `SyncColumns` partial-success accounting and focused event-contract coverage remain.

Locally passed:

- `cargo nextest run --retries 0 -p common-meta reconciliation::reconcile_table`
- `cargo nextest run --retries 0 -p tests-integration --test main reconciliation_event::test_table_reconciliation_events`
- `cargo clippy -p common-meta --all-targets -- -D warnings`

src/common/meta/src/reconciliation/reconcile_table.rs
```diff
@@ -147,6 +249,7 @@ impl PersistentContext {
 pub(crate) struct VolatileContext {
     pub(crate) table_meta: Option<TableMeta>,
     pub(crate) metrics: ReconcileTableMetrics,
+    result_summary: ReconcileTableResultSummary,
```
WenyXu, line 251: Now that `result_summary` is volatile, the recovery override that downcast table states is gone. `State::as_any` and the related `Any` imports/implementations in the table reconciliation states have no remaining caller. Could we remove this table-specific downcasting plumbing as part of the cutover? It is dead code now and does not need to be retained for the other reconciliation state traits.

dhruvxvaishnav: Addressed in `395f77252`. Removed `State::as_any`, the table trait’s `Any` import, and the corresponding imports/implementations from every table reconciliation state. The other reconciliation state traits remain unchanged.

I also checked the rest of the PR-scoped diff and resolved review threads; no additional cleanup was needed without expanding the scope.

Locally passed:

- `cargo build -p common-meta`
- `cargo nextest run --retries 0 -p common-meta` (562 passed)
- `cargo nextest run --retries 0 -p tests-integration --test main reconciliation_event::test_table_reconciliation_events` (1 passed)
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `git diff --check`

WenyXu (approved):
LGTM

PR #8936, opened 2026-08-23
feat: report what the graph derives and fix two duplicate-node bugs

killme2008:

Follows #8904. Tracked by #8609.

Four changes to the identity the graph derives and what it reports about it.

### 1. `table_semantics` reports the derived entity declarations

`information_schema.table_semantics` only read table options, so the
declarations the conventions derive — trace tables, whitelisted Prometheus and
OTel descriptor metrics — were invisible. Answering "why is my table not in the
graph" took debug logs.

The new `entity_declarations` column lists the entities a table contributes:
identity columns, whether the declaration came from an option or from the
conventions, and any row-level condition. A missing entity is the answer — name
not whitelisted, wrong source stamp, absent id column.

```
> SELECT table_name, entity_declarations FROM information_schema.table_semantics
    WHERE table_name = 'kube_pod_info';

kube_pod_info | [{"entity_type":"k8s.node","id":["node"],"origin":"convention"}]
```

A table with no semantic option but convention-derived entities now gets a row
too, since it is in the graph.

`EntityGraphProvider` gains a `table_declarations` method reading `TableInfo`
only, so `catalog` still does not depend on `operator` and the
`information_schema` path stays metadata-only.

### 2. A single-column identity keeps its structured form

`entity_id_attrs` was NULL whenever the identity came from one column, leaving
`host` = `a3f2...` with nothing saying which column produced it, and no way
back to the source table. Most identities are single-column — `host`,
`k8s.pod`, `k8s.node`, `service` — so the common case was the opaque one.

```
 entity_type      | entity_id | entity_id_attrs
------------------+-----------+-------------------------
 host             | h1        | {"host":"h1"}
 service.instance | cart-0    | {"instance":"cart-0"}
```

Entity equality still reads `entity_id` alone, so nothing merges differently.
`entity_id` deliberately omits the column names — one identity arrives under
different names per source, a trace table's `service_name` against a metric
table's `job` — and `entity_id_attrs` now carries them beside it for every
entity rather than only composites.

### 3. The generic `container` yields to `k8s.container`

A container in a pod reached the graph twice: as `k8s.container`
(`[pod uid, container name]`, from kube-state-metrics) and as the generic
`container` (`container.id`, from OTel resource attributes).

```yaml
- entity: k8s.container
  id: [k8s.pod.uid, k8s.container.name]
- entity: container
  id: [container.id]
  superseded_by: k8s.container
```

`superseded_by` names the superseding type rather than a trigger column, and
withdraws only on rows carrying that type's full identity: a container can
change type but cannot disappear. Keying it on "row has a pod UID" would delete
the node on OTLP-only deployments, because `k8s.container` used to come only
from kube-state-metrics. Both OTel sources now declare `k8s.container` under the
kube-state-metrics identity, so the two sources name one node;
`k8s.container.name` joins the descriptor's projected attributes.

The rule is per row — one descriptor table holds pod rows and bare-runtime rows
— and lives in the shared `declaration_predicate`, so an entity's edges follow
it too. Yielding costs the row nothing: the runtime container id
and name stay descriptive on `k8s.container` (as `kube_pod_container_info`
already keeps them), and the vocabulary gains
`service.instance -runs_on-> k8s.container` and `k8s.container -runs_on-> host`
to mirror the generic type's edges.

`k8s.container` keeps `[uid, name]` rather than moving to `container_id`: that
label carries a runtime prefix (`containerd://<hash>`), changes on every
restart, and is empty while a container starts.

### 4. `unmatched_count` and `duration_max` on `semantic_relationships`

`real_wins` pins a `calls` edge's RED metrics to the observed span pairs: if a
window's edge key holds any pair, the unmatched client spans are suppressed so
one (window, edge) yields one row. "The callee stopped responding" and "the
traffic stopped arriving" then look identical — both are just a lower
`request_count`.

`unmatched_count` counts the client spans with no server span, deliberately
outside `real_wins`, on the same row. NULL for agent calls (inner join, nothing
left unmatched) and for declared edges.

`duration_max` is the longest single request, inside `real_wins`: a pair is
timed by the server span, an unmatched client by its own span including network
wait, so mixing them would make the max describe a different population than
`duration_sum`/`duration_count`. Agent calls compute it from the child spans
they already aggregate. The projection contract goes from 16 to 18 columns.

### Out of scope

The `entity.<type>.qualifier` role, which was in an earlier revision of this PR, has been pulled out and moved to its own change (#8609). It fixed a real problem — a namespaced `service` reaches the graph as two entities, one per signal — but qualifying the identity broke the string equality that unmatched client spans rely on to find their peer, and the compensating fix needed a second pass over the span data plus a new conventions concept. That is identity design, not a patch to land beside four unrelated changes.

`k8s.workload`'s conditional mapping is not addressed here and is not queued
behind this PR. It needs literal id components
(`[namespace, <const Deployment>, deployment.name]`) and several declarations of
one type per table, neither of which the row-level condition above provides;
both reshape `id_columns` across the derivation.

## PR Checklist


fengjiachun (approved):
LGTM

shuiyisong:
One correctness finding before merge:

### [P1] Keep unmatched virtual service IDs in the qualified identity space

For a namespaced trace-v1 service, this PR makes the real endpoints canonical IDs such as `shop/frontend` and `shop/cart`. However, `client_spans` still forwards `peer.service`/`service.peer.name` unchanged as `virtual_dst`, and `calls_branch` coalesces and groups that raw value directly with qualified server IDs.

I reproduced one paired and one unmatched call to the same peer in the same window using valid OTel values (`service.namespace=shop`, client `service.name=frontend`, remote `service.name=cart`, `peer.service=cart`). The resulting rows are:

```text
shop/frontend -> cart       confidence=0.5 unmatched_count=1
shop/frontend -> shop/cart  confidence=1.0 unmatched_count=0
```

This defeats the new `unmatched_count` contract: the unanswered client does not contribute to the real `shop/frontend -> shop/cart` edge, and instead creates a second unqualified virtual node. OTel defines `peer.service` as the remote service's bare `service.name`, so telemetry is not expected to carry `shop/cart` here.

Relevant paths: [the trace-v1 qualifier](https://github.com/GreptimeTeam/greptimedb/blob/e6cffa11130603bc60ec6337bfe3415ef1166ce1/src/operator/src/statement/semantic_graph/conventions.yaml#L46-L48), [raw virtual destination](https://github.com/GreptimeTeam/greptimedb/blob/e6cffa11130603bc60ec6337bfe3415ef1166ce1/src/operator/src/statement/semantic_graph/relationships.rs#L568-L596), and [direct coalesce/group key](https://github.com/GreptimeTeam/greptimedb/blob/e6cffa11130603bc60ec6337bfe3415ef1166ce1/src/operator/src/statement/semantic_graph/relationships.rs#L367-L370).

Please canonicalize service virtual destinations using a remote qualifier when it can be determined, and define an explicit fallback when it cannot. Blindly applying the caller's namespace would misidentify cross-namespace calls. A regression test should cover one paired plus one unmatched call under a qualified service declaration.

sunng87 (approved):

killme2008:
Confirmed, and thank you — I reproduced it exactly as described:

```
shop/frontend -> shop/cart  confidence=1.0  unmatched_count=0
shop/frontend -> cart       confidence=0.5  unmatched_count=1
```

You are also right that applying the caller's namespace is not a fix: `peer.service` holds the remote's bare `service.name` and a client span carries nothing about the remote's namespace, so that heuristic would merge `other/cart` into `shop/cart` — the same class of error this work is trying to remove.

I have taken the `entity.<type>.qualifier` role out of this PR rather than patch it. The reasoning: `entity_id` is a rendered string, and the calls derivation matched unmatched clients to real peers only because that string happened to equal the bare `peer.service`. Qualifying the identity breaks the coincidence, and the fix I had working — resolve the peer name against the services the window observed serving, keep it a virtual node when the name is absent or ambiguous — needs an extra pass over the server-side spans and a new `names_service` flag in the conventions to keep `server.address` out of it. That is identity design; it does not belong beside four unrelated changes.

What remains in this PR is the four changes that never depended on qualification, and your scenario no longer reproduces on it.

One thing worth flagging separately: **this is latent on main**, not only here. `qualified_by` on the trace-v1 `service` convention landed in #8904. Today the ingest-side `entity.service.id` stamp suppresses that convention, so the qualified id is normally never produced — but a trace-v1 table without the stamp (hand-created, or created before the stamp existed) that has a `resource_attributes.service.namespace` column will hit exactly the split you found. Registered on the tracking issue with your repro so the follow-up starts from it.

PR #8863, opened 2026-08-13
feat(flow): mergeable UDDSketch state for incremental aggregation

discord9:
I hereby agree to the terms of the [GreptimeDB CLA](https://github.com/GreptimeTeam/.github/blob/main/CLA.md).

## Refer to a related PR or issue link (optional)

PR 4/5 of the incremental Flow data-plane series. Depends on #8867 (UDDSketch implementation, upstream); will rebase and re-verify once it lands.

### Problem

Per-window UDDSketch percentiles needed a mergeable state so incremental flow can maintain group state instead of rescanning full source history.

### Change

- new scalar `uddsketch_merge_state` (binary state merge) with fallible parameter validation (bucket size, error rate, percentile range) and empty/null/corrupt hardening
- Flow `IncrementalAggregateMergeOp::UddSketchState` merges `uddsketch_state`/`uddsketch_merge` deltas with existing sink state
- exact error-rate comparison before merging; dependency assert panics on malformed state converted to errors
- flow's direct bincode/uddsketch dev-deps removed

### Verification

- targeted uddsketch scalar + flow merge-op tests (utils/test.rs)
- `make fmt` / `git diff --check` clean

## PR Checklist

- [x] I have written the necessary rustdoc comments.
- [x] I have added the necessary unit tests and integration tests.
- [ ] This PR requires documentation updates.
- [x] API changes are backward compatible.
- [x] Schema or data changes are backward compatible.

PR #8938, opened 2026-08-24
feat(flow): add generic delta merge for incremental aggregates

discord9:

Supersedes #8863's per-type scalar merge approach for opaque aggregate states.

### Problem

Incremental Flow execution needs to combine a newly computed aggregate state with the state already persisted in the sink. Opaque state aggregates such as HLL and UDDSketch already provide merge UDAFs, but those UDAFs consume one state column at a time. The incremental join produces two state columns: the delta state and the persisted state.

Adding a separate public scalar function for every state representation would duplicate the existing merge implementation and couple Flow to each binary state format.

### Changes

This PR adds the small `common-function` core needed to reuse the existing merge UDAFs:

- Add an internal Exact-signature delta-merge adapter. It extends an existing merge UDAF from `P..., State` to `P..., delta_state, persisted_state` and feeds both opaque states to the existing accumulator without decoding their representation.
- Explicitly register the two currently supported derived UDAFs:
  - `__hll_delta_merge(Binary, Binary) -> Binary`
  - `__uddsketch_state_delta_merge(Int64, Float64, Binary, Binary) -> Binary`
- Add aggregate-function lookup by internal name.
- Test real, non-null HLL and UDDSketch state merges and verify their decoded semantics.

The adapter is internal and Exact-signature-only. This PR does not introduce a generic aggregate-family registration framework or automatic support for arbitrary aggregates.

### Scope and follow-up

This PR contains no Flow rewrite or Flow integration tests. A follow-up PR will resolve these explicitly registered internal UDAFs by name when rewriting incremental aggregate plans.

Existing ordinary Flow aggregates remain on their scalar merge-expression path: `sum`, `count`, `min`, `max`, `bool_and`, `bool_or`, `bit_and`, `bit_or`, and `bit_xor`. The opaque-state path added here is additive and does not replace that support.

There are no public SQL functions, configuration changes, persisted-format changes, or wire-format changes.

## PR Checklist


src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -685,6 +691,239 @@ impl Accumulator for StateAccum {
...
+    ///
+    /// The caller supplies the known merge signature, so construction cannot
+    /// fail while inspecting an arbitrary UDAF signature.
+    pub(crate) fn new(
+        inner: AggregateUDF,
+        state_name: &str,
+        inner_types: Vec<DataType>,
+        state_type: DataType,
+    ) -> Self {
+        let mut wrapper_types = inner_types.clone();
```
killme2008, line 720: Why cloning `inner_types`? We can make it `mut` and consume it directly.

src/common/function/src/aggrs/aggr_wrapper.rs
```diff
@@ -685,6 +691,239 @@ impl Accumulator for StateAccum {
...
+        if input_fields.len() != self.inner_types.len() + 1 {
+            return Err(datafusion_common::DataFusionError::Plan(
+                "delta merge requires parameters, delta state, and persisted state".to_string(),
+            ));
+        }
+        for (field, expected_type) in input_fields[..self.inner_types.len()]
+            .iter()
+            .zip(&self.inner_types)
+        {
+            if field.data_type() != expected_type {
```
killme2008, line 744: I am not sure if we should consider the type coercion here.

killme2008 (approved):
LGTM

Question: need to add sqlness cases?

src/common/function/src/aggrs/aggr_wrapper.rs
discord9: `inner_types` is the original `P..., State` merge-family signature. The wrapper needs to keep that list for `resolve_inner_args` and for projecting the parameter/state fields when it constructs the inner accumulator, while the registered `Signature::exact` needs a separate expanded `P..., delta_state, persisted_state` list. Consuming `inner_types` to build `wrapper_types` would therefore force another allocation/copy anyway; keeping one clone here is intentional.

src/common/function/src/aggrs/aggr_wrapper.rs
discord9: This wrapper uses an exact internal signature. The latest head removes the previous `coerce_types` passthrough because DataFusion does not call a custom coercion hook for `Signature::exact`; keeping it was dead and misleading. Delta merge is intentionally internal and generated with exact parameter/state types, so this PR does not add a custom coercion policy.

PR #8940, opened 2026-08-24
refactor(json2): support querying v2 storage layout

MichaelScofield:

This PR adds query-side support for the JSON2 v2 storage layout.

JSON2 v2 stores explicitly materialized fields alongside a sibling Variant remainder containing non-materialized or conflicting values. The existing Parquet nested projection only matched explicit paths or legacy Variant parents, so paths stored in the v2 remainder could be treated as absent.

This PR:

- Makes Parquet projection planning aware of the JSON2 physical layout version.
- Reads the v2 remainder leaves when a requested path has no explicit physical field, while retaining explicit-field-only pruning when the path is materialized.
- Reconstructs logical JSON2 values from explicit fields and the remainder during flat projection.
- Preserves complete root JSON2 values when predicates access nested paths.
- Handles JSON2 root and nested paths correctly in `json_get`.
- Returns JSON representations for nested objects and arrays cast to strings.
- Adds unit coverage for explicit fields, remainder fallback, and legacy layout behavior.
- Moves supported read cases out of `json2_limit` and adds a dedicated `json2_join` sqlness case.

The remainder lookup is performed at most once per JSON2 root when building a Parquet projection plan.

This PR does not introduce another persisted format change. Legacy JSON2 reads and existing storage layouts remain supported.

## PR Checklist


MichaelScofield:
@codex review

src/mito2/src/sst/parquet/read_columns.rs
fengys1996: Should update the comment.

src/mito2/src/sst/parquet/read_columns.rs
```diff
@@ -287,6 +315,56 @@ fn build_parquet_leaves_indices(
...
+    }
+    matches!(field.data_type(), DataType::Struct(_))
+}
+
+/// Finds the Parquet leaves backing a JSON2 v2 remainder field.
+///
+/// The remainder is a sibling of explicitly materialized fields, so prefix matching a
+/// requested path cannot find it. These leaves are needed when an explicit path is absent
+/// or an explicitly materialized object may have additional children in the remainder.
+fn find_v2_remainder_leaves(
```
fengys1996: nit: better name `find_remainder_leaves`

src/mito2/src/sst/parquet/read_columns.rs
```diff
@@ -210,12 +216,16 @@ pub fn build_projection_plan(
...
+/// `arrow_schema` identifies JSON2 layouts when a requested path needs
+/// to fall back to a Variant parent or the v2 remainder field.
+///
 /// Returns `(matched_leaves, matched_roots)`:
 /// - `matched_leaves`: matched leaf-column indices in the current parquet file schema.
 /// - `matched_roots`: root-field indices read from the current parquet file schema.
 fn build_parquet_leaves_indices(
     parquet_schema_desc: &SchemaDescriptor,
     projection: &ParquetReadColumns,
+    arrow_schema: &Schema,
```
fengys1996: Do we have to pass `arrow_schema` here? For example, could `path_points_to_struct` determine this based on the `Parquet schema description` instead?

src/mito2/src/sst/parquet/read_columns.rs
MichaelScofield: I kept the v2 qualifier because this helper only applies after verifying the JSON2 v2 physical layout. find_remainder_leaves would make the scope less explicit.

fengys1996 (approved):

PR #8941, opened 2026-08-24
feat(meta): record logical-table reconciliation events

dhruvxvaishnav:

Part of #8861. This is the logical-table stage of the agreed order: #8935 (merged) → #8941 → #8896 (draft).

Adds the independently filterable `reconcile_logical_tables` lifecycle event. Each logical table has its own locator row containing catalog, schema, name, table ID, and physical table ID. Physical and logical reconciliation have separate `Event` implementations; only the locator schema and row conversion are shared.

Submitted payload version 1 contains `logical_table_count` and `is_subprocedure`. Other non-terminal triggers have a null payload.

Following the review, terminal payload version 1 reads only existing procedure state and metrics:

- `complete`: true for `Succeeded`, false for `Failed` and `Poisoned`.
- `processed_table_count`: the number of persistent logical-table IDs in the request, including tables whose reconciliation may not have completed.
- `metadata_consistent_table_count`, `metadata_inconsistent_table_count`, `create_table_count`, and `update_table_info_count`: the existing `ReconcileLogicalTableMetrics` counters.

These are best-effort metrics from the current process. `create_table_count` records tables identified for creation during resolution, rather than completed region creates. Metrics reset after recovery and do not account for every partial RPC or transaction side effect. The framework continues to provide lifecycle state, trigger, and error.

The event-only summary, custom phase, missing-region tracking/rescans, individual region RPCs, retry accounting, and batch-update progress callback have been removed. The affected reconciliation execution and helper files match the upstream base. Persisted procedure contexts, serialization, state transitions, batching, and retry/recovery behavior are unchanged from that base.

Coverage includes exact payloads, root/child intent, filtering, multiple locator rows, metrics reset on deserialization, and multi-table event recording through the integration cluster. Configuration examples and their generated reference include the new event type.

### Validation

Passed locally with the repository's pinned nightly toolchain:

- `cargo nextest run --retries 0 -p common-meta` — 568 tests passed
- `cargo nextest run --retries 0 -p tests-integration --test main reconciliation_event::` — physical- and logical-table event tests passed
- `cargo clippy -p common-meta --all-targets -- -D warnings`
- `cargo clippy -p tests-integration --test main --no-deps -- -D warnings`
- `git diff --check`

## PR Checklist


WenyXu:
@codex review

WenyXu:
@dhruvxvaishnav Would you mind splitting these changes vertically, so each PR contains only the commits relevant to its own scope? The later PRs can still depend on the earlier changes if needed.

dhruvxvaishnav:
Thanks, @WenyXu. I agree that each final PR should show only the commits for its own vertical scope.

I checked whether I could retarget the dependent PRs to the preceding branches. GitHub's stacked-PR support requires all stack branches to be in the same repository; cross-fork stacks are not supported: https://docs.github.com/en/pull-requests/reference/stacked-pull-requests. These branches are in my fork, while the PR base branches must be in `GreptimeTeam/greptimedb`, and I do not have permission to create dependency base branches in the upstream repository. That is why #8935 currently also shows #8896's commits and this PR also shows the earlier two stages, even though each stage has its own scope-specific commit.

The two ways I can make the upstream diffs genuinely vertical are:

1. close #8935 and #8941 now and reopen them sequentially after their prerequisite PRs merge; or
2. keep them open, then rebase #8935 onto `main` immediately after #8896 merges and rebase #8941 after #8935 merges.

Would you prefer that I close the two dependent PRs now, or keep the existing review history and rebase them as their prerequisites land?

Separately, I am addressing the physical-table partial-result findings from the Codex reviews in the #8935 branch and validating those fixes now. I will keep those changes in PR 2's scope and propagate only the corrected dependency into PR 3.

src/common/meta/src/reconciliation/reconcile_table/reconciliation_start.rs
```diff
@@ -114,6 +114,10 @@ impl State for ReconciliationStart {
             }
         });
```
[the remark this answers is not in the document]
dhruvxvaishnav: Fixed in the PR 2 dependency commit `7fe7cd67b` and propagated into this PR's new head. The scan count is now recorded immediately after `list()` returns, before the missing/open-region checks; failures keep the attempted count without marking the start phase complete.

src/common/meta/src/reconciliation/reconcile_table/reconcile_regions.rs
```diff
@@ -143,6 +144,10 @@ impl State for ReconcileRegions {
                 },
             )?;
 
+        ctx.persistent_ctx
+            .result_summary
+            .mark_regions_reconciled(updated_region_count);
```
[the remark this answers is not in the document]
dhruvxvaishnav: Fixed in the PR 2 dependency commit `7fe7cd67b` and propagated into this PR's new head. Successful `SyncColumns` responses are counted before batch-error propagation and before response-extension parsing, while `last_completed_phase` advances only after validation succeeds.

WenyXu:
@dhruvxvaishnav I wouldn’t recommend making these stacked PRs. Ideally, the three PRs should be independent and each contain only its own vertical scope.

If there is a real dependency, I’d prefer keeping the dependent PR as a draft and waiting for the prerequisite PR to merge before rebasing/updating it.

dhruvxvaishnav:
Thanks, @WenyXu — understood. I have marked #8935 and this PR as drafts.

I will stop treating them as active stacked PRs. Once #8896 merges, I will rebase each draft separately onto the then-current `main`, remove the inherited commits, revalidate the resulting vertical diff, and mark it ready only when it contains its own scope. If this logical-table PR still has a genuine dependency on #8935 after that rebase, I will keep it as a draft until #8935 merges rather than present stacked changes for review.

WenyXu:
> Thanks, @WenyXu — understood. I have marked #8935 and this PR as drafts.
> 
> I will stop treating them as active stacked PRs. Once #8896 merges, I will rebase each draft separately onto the then-current `main`, remove the inherited commits, revalidate the resulting vertical diff, and mark it ready only when it contains its own scope. If this logical-table PR still has a genuine dependency on #8935 after that rebase, I will keep it as a draft until #8935 merges rather than present stacked changes for review.

I think it might be better to merge #8935 and this PR first. Since the database/catalog reconciliation procedure will spawn sub-procedures, merging it first would mean we’d have to keep updating the test cases as #8935 and this PR evolve.
What do you think? Should we start with table-level reconciliation instead？

dhruvxvaishnav:
Yes, I agree, starting with table-level reconciliation is cleaner. It stabilizes the child event contracts before the catalog/database parent integration assertions are finalized, so we avoid repeatedly updating the parent tests.

I have also marked #8896 as a draft. I am rebuilding #8935 directly on the current `main` with the shared five-column locator/event infrastructure plus only the physical-table vertical scope, including a direct table-level integration test rather than catalog/database event assertions. After that vertical diff is regenerated and revalidated, I will update #8935 and mark it ready. Then I will rebase/update this logical-table PR, and handle #8896 last.

So the intended review/merge order is now: #8935 → #8941 → #8896.

dhruvxvaishnav:
Update: I have implemented the child-first restructuring we discussed.

#8935 has been rebuilt directly on the current `main` and force-pushed with lease at `3b470abcd`. Its diff now contains the shared five-column event infrastructure plus only the physical-table vertical scope; there are no catalog/database or logical-table procedure changes. The integration test now invokes table reconciliation directly, so it does not depend on parent event assertions. All focused local tests, formatting, Clippy, and `git diff --check` passed.

I am keeping #8935 as a draft for one explicit remaining item: Docker is unavailable locally, so the generated `config/config.md` has not yet been regenerated from the updated example TOMLs, and I did not hand-edit it.

I also checked the linked full-workspace CI failure on the previous head. It was unrelated to reconciliation: `test_create_etcd_client_with_mtls_and_ca` failed on every retry because the CI etcd certificate had expired. A separate Kafka cleanup test passed on retry.

dhruvxvaishnav:
Final update on the reordered first stage: the generated configuration-documentation gap is resolved. I found the working Docker daemon inside Ubuntu WSL2, ran the repository's exact pinned `make config-docs` target, verified that it added only the two expected `reconcile_table` documentation entries, and pushed the signed-off generated-doc commit `b6604c1ac`.

#8935 is now a clean vertical physical-table PR based directly on `main`, all focused local checks have passed, and it has been marked ready for review. #8941 and #8896 remain drafts for the next two stages.

dhruvxvaishnav:
@WenyXu, #8941 has been rebased onto the current `main` after #8935 merged and is ready for review. The diff now contains only the logical-table reconciliation event scope, with volatile best-effort result summaries aligned with #8935 and focused unit/integration coverage.

WenyXu:
@codex review

src/common/meta/src/reconciliation/event.rs
```diff
@@ -104,11 +150,78 @@ struct TableResultPayload {
 #[derive(Debug)]
 pub(crate) struct ReconciliationEvent {
```
WenyXu: This type now owns both physical-table and logical-table payloads and lifecycle constructors. The logical-table addition also changes the physical-table representation from one locator to `Vec<ReconciliationLocator>` solely to support its multi-row output. A future reconciliation event will need to extend the same enums and constructors again. Could we give each event type its own `Event` implementation and keep only the locator schema/row conversion shared?

src/common/meta/src/reconciliation/reconcile_logical_tables/update_table_infos.rs
```diff
@@ -94,11 +98,12 @@ impl State for UpdateTableInfos {
             table_info_values_to_update.push((table_info_value, new_table_info));
         }
         let table_id = ctx.table_id();
-        let table_name = ctx.table_name();
-
         let updated_table_info_num = table_info_values_to_update.len();
         batch_update_table_info_values(&ctx.table_metadata_manager, table_info_values_to_update)
             .await?;
```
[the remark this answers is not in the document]
dhruvxvaishnav: Addressed in `ba4bd43be`. `batch_update_table_info_values_with_progress` now reports each successfully committed chunk, and the logical reconciliation summary accumulates those counts before any later chunk error is propagated. The focused summary test covers accumulation without advancing the completed phase.

src/common/meta/src/reconciliation/reconcile_logical_tables/reconcile_regions.rs
```diff
@@ -59,27 +65,36 @@ impl State for ReconcileRegions {
...
 
-        future::join_all(create_table_tasks)
-            .await
-            .into_iter()
-            .collect::<Result<Vec<_>>>()?;
+        let results = future::join_all(create_table_tasks).await;
+        let created_region_count = results
+            .iter()
+            .filter_map(|result| result.as_ref().ok())
+            .sum();
```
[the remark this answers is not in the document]
dhruvxvaishnav: Addressed in `ba4bd43be`. Region reconciliation now sends one logical-region create per RPC while retaining per-datanode sequencing, records every successful `(table_id, region_number)` before propagating the first error, and removes successful entries from the volatile missing-region state. The mixed-outcome test asserts one create per request and verifies the partial terminal counts.

src/common/meta/src/reconciliation/reconcile_logical_tables/reconcile_regions.rs
```diff
@@ -59,27 +65,36 @@ impl State for ReconcileRegions {
...
+        let created_region_table_count = ctx.persistent_ctx.create_tables.len();
         let mut create_table_tasks = Vec::with_capacity(leaders.len());
         for (datanode_id, region_role_set) in region_distribution {
             if region_role_set.leader_regions.is_empty() {
                 continue;
             }
             // Safety: It contains all leaders in the region routes.
             let peer = leaders.get(&datanode_id).unwrap().clone();
+            let created_region_count =
+                region_role_set.leader_regions.len() * created_region_table_count;
```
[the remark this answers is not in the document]
dhruvxvaishnav: Addressed in `ba4bd43be`. The metadata lister now preserves each requested `RegionId`; resolution stores only entries whose metadata is `None`, and reconciliation schedules and counts only those missing logical regions. The missing-region map is volatile, so recovered procedures relist actual state instead of trusting stale progress. Focused tests cover partial presence and recovery.

src/common/meta/src/reconciliation/event.rs
dhruvxvaishnav: Addressed in `ba4bd43be`. `ReconcileTableEvent` and `ReconcileLogicalTablesEvent` now have separate payload enums, lifecycle constructors, and `Event` implementations. The physical-table event again owns one locator, the logical-table event owns its multi-row locator vector, and only `ReconciliationLocator::schema` and `ReconciliationLocator::row` are shared.

dhruvxvaishnav:
@WenyXu, I addressed all current review feedback in `ba4bd43be`.

- Physical-table and logical-table events now have separate payload types, constructors, and `Event` implementations. Only the shared five-column locator schema and row conversion remain common.
- Logical-region reconciliation now preserves the exact missing region numbers, schedules only missing regions, and rebuilds that volatile state from datanodes after procedure recovery.
- Each logical-region create is sent separately while preserving per-datanode sequencing, so successful side effects are counted before a later failure is returned.
- Table-info progress is recorded after each committed KV transaction chunk, so a later chunk failure retains the completed count.
- Focused tests cover exact locator rows, partial region outcomes, recovered-state reconstruction, and accumulated table-info progress.

The final local checks all passed:

- `cargo nextest run --retries 0 -p common-meta reconciliation::reconcile_logical_tables`
- `cargo nextest run --retries 0 -p tests-integration --test main reconciliation_event::test_logical_table_reconciliation_events`
- `cargo clippy -p common-meta --all-targets -- -D warnings`
- `git diff --check`

I replied to and resolved the addressed review threads. This is ready for another look.

src/common/meta/src/reconciliation/reconcile_logical_tables/reconcile_regions.rs
```diff
@@ -42,20 +43,38 @@ impl State for ReconcileRegions {
         ctx: &mut ReconcileLogicalTablesContext,
         _procedure_ctx: &ProcedureContext,
     ) -> Result<(Box<dyn State>, Status)> {
+        ctx.volatile_ctx
```
WenyXu: missing_regions_by_table retains and prunes the regions created before a retryable error, so a retry only creates the remaining regions. However, this resets the creation counters on every execution, and the later updates overwrite them with the current attempt's values. As a result, the succeeded event under-reports regions and tables created before the retry. Could we preserve/accumulate these counters across retries and add a partial-success → retry-success regression test?

src/common/meta/src/reconciliation/reconcile_logical_tables/reconcile_regions.rs
```diff
@@ -66,20 +85,74 @@ impl State for ReconcileRegions {
...
             let peer = leaders.get(&datanode_id).unwrap().clone();
-            let request = self.make_request(&region_role_set.leader_regions, ctx)?;
+            let mut requests = Vec::new();
+            for region_number in region_role_set.leader_regions {
+                let Some(table_ids) = missing_tables_by_region.get(&region_number) else {
+                    continue;
+                };
+                for table_id in table_ids {
+                    requests.push((
+                        self.make_request(region_number, *table_id, ctx)?,
```
WenyXu: This path now builds one RPC per missing `(table_id, region_number)`, but each call rescans `create_tables`, recreates the `CreateRequestBuilder`, and rebuilds the partition-expression map. Those inputs are fixed for a logical table during this state. Could we prepare them once per table while keeping the individual RPCs for partial-progress tracking?

src/common/meta/src/reconciliation/reconcile_logical_tables.rs
```diff
@@ -53,6 +57,113 @@ pub struct ReconcileLogicalTablesContext {
     pub volatile_ctx: VolatileContext,
```
WenyXu, line 56: This introduces a substantial amount of instrumentation and behavioral change to `ReconcileLogicalTablesProcedure` solely for event reporting. Could we step back and reconsider which result fields are actually necessary, using `ReconcileTableProcedure` as a reference? We should prefer fields already available from the procedure state or existing metrics, rather than changing reconciliation execution or retry behavior to populate the event.

dhruvxvaishnav: Thanks, I agree. I think the right fix is to remove the event-only progress instrumentation rather than make reconciliation execution track more retry state.

I propose that the logical-table terminal payload contain only bounded fields already available without changing procedure behavior:

- `version` and `complete`
- `processed_table_count`, from the persistent logical-table IDs
- `metadata_consistent_table_count`
- `metadata_inconsistent_table_count`
- `create_table_count`
- `update_table_info_count`

The last four come directly from the existing `ReconcileLogicalTableMetrics`. The framework-owned trigger, procedure state, and error already describe lifecycle position and failure, so I would remove the custom phase field as well.

This would remove the missing-region rescan/pruning state, region scan and creation counters, per-chunk table-info progress callback, and the retry-specific counter handling. A recovered procedure would therefore emit the metrics from its current process rather than a stale accumulated summary, matching the existing best-effort metrics behavior and avoiding any persisted procedure-format change.

Does this narrower payload match the direction you want? If so, I will simplify the PR along these lines and add event-contract coverage for the resulting payload.

src/common/meta/src/reconciliation/reconcile_logical_tables/reconcile_regions.rs
dhruvxvaishnav: Addressed in `cd8d1f5f2`. `begin_region_reconciliation` now preserves prior creation totals, each attempt adds only its newly successful table/region counts, and `mark_regions_reconciled` advances only the phase without overwriting accumulated progress. Because successful regions are pruned from `missing_regions_by_table`, retries do not count them twice.

The new `logical_table_region_progress_accumulates_across_retry` regression test records 1 completed table and 2 created regions in the partial failed event, retries only the remaining region, then verifies the succeeded event reports the full 2 tables and 3 regions. The exact test passed locally, as did the broader focused `common-meta` reconciliation suite and logical-table integration event test.

This addresses the current implementation while the separate narrower-payload discussion remains open.

src/common/meta/src/reconciliation/reconcile_logical_tables/reconcile_regions.rs
dhruvxvaishnav: Addressed in `cd8d1f5f2`. `prepare_create_requests` now walks `create_tables` once per state execution and builds one `CreateRequestBuilder` plus one partition-expression map per logical table. The shared storage path is also computed once. Individual `(table_id, region_number)` RPC construction now performs a map lookup and reuses those prepared inputs, preserving the partial-progress behavior without repeated table scans or builder reconstruction.

Focused `common-meta` tests, the logical-table integration event test, formatting, targeted Clippy, and `git diff --check` all passed locally. This addresses the current implementation while the separate narrower-payload discussion remains open.

src/common/meta/src/reconciliation/reconcile_logical_tables.rs
WenyXu: This still expands reconciliation behavior solely for event reporting: retry/recovery state, per-region progress tracking, and the batch-update callback exist only to populate the terminal payload. Please reduce the payload to fields already available from persistent state or existing metrics, and remove this event-only instrumentation.

dhruvxvaishnav: @WenyXu, addressed in `ccd53a5a0`.

The logical-table terminal payload now contains only `version`, `complete`, the number of logical-table IDs in the persistent request, and the four existing `ReconcileLogicalTableMetrics` counters: metadata-consistent tables, metadata-inconsistent tables, tables identified for creation, and table-info updates.

I removed the event-only result summary and phase tracking, missing-region rescan and pruning state, per-region RPC restructuring, retry accumulation, and the batch-update progress callback. The affected reconciliation execution and helper paths now match the base behavior, and the persisted context, serialization, state transitions, batching, and retry/recovery behavior are unchanged. After recovery, the volatile metrics start from zero and the terminal event reports only best-effort observations from the current process.

The event tests now cover the reduced payload directly, including a failure before resolution, existing metrics, and recovery. The root/child intent, filtering, locator rows, and physical/logical integration coverage remain. The full `common-meta` suite passed all 568 tests, both reconciliation event integration tests passed, and the targeted Clippy, formatting, and diff checks passed.

tests-integration/tests/reconciliation_event.rs
```diff
@@ -135,6 +135,130 @@ async fn test_table_reconciliation_events() {
...
+             AND json_get_int(payload, 'update_table_info_count') = 0"
+        ),
+        "\
++---------+
+| matches |
++---------+
+| true    |
++---------+",
+    )
+    .await;
```
WenyXu, line 272: Could we list the actual values in the expected output table instead of just matches = true?

tests-integration/tests/reconciliation_event.rs
```diff
@@ -135,6 +135,130 @@ async fn test_table_reconciliation_events() {
...
+             AND json_get_int(payload, 'version') = 1 \
+             AND json_get_int(payload, 'logical_table_count') = 2 \
+             AND json_get_bool(payload, 'is_subprocedure') = true"
+        ),
+        "\
++---------+
+| matches |
++---------+
+| true    |
++---------+",
```
WenyXu: ditto

WenyXu (approved):
Rest LGTM

tests-integration/tests/reconciliation_event.rs
dhruvxvaishnav: Updated the succeeded-event assertion to show the locator counts and all result payload fields as concrete expected values, including `complete` and each result counter. The query groups by those payload values, so inconsistent event rows appear separately and fail the expected output. The focused logical-table integration test passes. Fixed in `1fd856452`.

tests-integration/tests/reconciliation_event.rs
dhruvxvaishnav: Updated the submitted-event assertion as well. It now returns the event count, distinct logical and physical table ID counts, payload version, logical-table count, and subprocedure flag as explicit expected values. Grouping by the payload fields also verifies that both locator rows carry the same payload. Fixed in `1fd856452`.

Issue #8933, opened 2026-08-21
Query hangs when a region remains routed to a scaled-in datanode


WenyXu:
## What type of bug is this?
- Unexpected error

## What subsystems are affected?
- Distributed Cluster
- Query Engine
- Frontend
- Datanode

## Minimal reproduce step
1. Start a distributed GreptimeDB cluster with multiple datanodes.
2. Create a table and write data so that one of its regions is hosted by a specific datanode.
3. Run a query whose scan range includes that region to establish the frontend-to-datanode channel.
4. Scale in the datanode hosting the region.
5. Before the region is recovered or rerouted, run the same query again.
6. Observe that the second query does not return a result or an error.
7. Restart the frontend and run the query again. The query then returns a TCP connection error.

## What did you expect to see?
The query must not wait indefinitely after its routed datanode becomes unavailable.

If a new connection to the datanode cannot be established, the query should return within the configured `[datanode.client] connect_timeout` with a diagnosable transport error. Depending on the network path, this may be `connection refused`, connection timeout, `no route to host`, `Unavailable`, or another equivalent error.

If the connection already exists but the remote request or Flight response stream stops making progress, the query must still have an effective cancellation or deadline path. This should not impose a fixed frontend-to-datanode request timeout on valid long-running queries.

## What did you see instead?
The first query establishes a frontend-to-datanode channel. After the datanode is scaled in, the second query can hang indefinitely without returning either a result or an error message.

After restarting the frontend, the same query returns a TCP connection error instead of hanging.

## Root cause analysis
`src/frontend/src/instance/region_query.rs` resolves the region leader from route metadata and sends the query to that datanode.

`NodeClients` caches a `Client` by peer, and `ChannelManager` caches a lazily connected HTTP/2 channel by address. The second query therefore reuses the channel created by the first query instead of creating a new TCP connection.

Frontend intentionally disables the frontend-to-datanode gRPC request timeout in `src/cmd/src/frontend.rs`:

```rust
// frontend to datanode need not timeout.
// Some queries are expected to take long time.
channel_config.timeout = None;
```

The TCP `connect_timeout` remains configured separately and defaults to one second unless overridden by `[datanode.client] connect_timeout`. It only applies while establishing a new connection, so it does not protect the reused channel.

[...]

PR #8943, opened 2026-08-24
fix(flight): bound DoGet response wait

WenyXu:

Related to #8933.

- Apply a 10-second timeout while waiting for frontend-to-datanode Flight `DoGet` response headers.
- This bounds a suspected hang where a datanode endpoint, or an intervening listener, accepts the HTTP/2/Flight connection but never returns the `DoGet` response headers. Without a request deadline, frontend waits indefinitely even though the connection-level transport remains usable.
- Return `DoGet` response headers before `handle_remote_read` initializes the query. The request timeout ends when the headers return; deferred query initialization and the subsequent Flight stream remain unbounded.
- Keep the shared frontend-to-node Channel timeout disabled so unary datanode and flownode RPCs retain their existing timeout behavior.
- Preserve cancellation: dropping the Flight stream aborts its producer task, including a pending deferred initializer.
- Add unit and integration coverage for deferred initialization errors and for a `DoGet` response that succeeds within its timeout while its Flight data is delayed beyond that timeout.

## PR Checklist

### DIFF
--- a/src/client/src/region.rs
+++ b/src/client/src/region.rs
@@ -112,7 +112,8 @@ impl RegionRequester {
         let mut flight_client = self
             .client
             .make_flight_client(self.send_compression, self.accept_compression)?;
-        // Limit Flight establishment without limiting query stream execution.
+        // Limit Flight DoGet response time without limiting query stream execution.
+        let addr = flight_client.addr().to_string();
         let mut request = tonic::Request::new(ticket);
         request.set_timeout(FLIGHT_DO_GET_TIMEOUT);
         let response = flight_client
@@ -124,11 +125,11 @@ impl RegionRequester {
                 let e: error::Error = e.into();
                 error!(
                     e; "Failed to do Flight get, addr: {}, code: {}",
-                    flight_client.addr(),
+                    addr,
                     tonic_code
                 );
                 Err(BoxedError::new(e)).with_context(|_| FlightGetSnafu {
-                    addr: flight_client.addr().to_string(),
+                    addr: addr.clone(),
                     tonic_code,
                 })
             })?;
@@ -139,7 +140,7 @@ impl RegionRequester {
         let flight_message_stream = flight_data_stream
             .filter_map(move |flight_data| decode_flight_data(&mut decoder, flight_data));
 
-        recordbatches_from_flight_message_stream(flight_message_stream).await
+        recordbatches_from_flight_message_stream(addr, flight_message_stream).await
     }
 
     async fn handle_inner(&self, request: RegionRequest) -> Result<RegionResponse> {
@@ -201,6 +202,7 @@ impl RegionRequester {
 }
 
 async fn recordbatches_from_flight_message_stream<S>(
+    addr: String,
     mut flight_message_stream: S,
 ) -> Result<SendableRecordBatchStream>
 where
@@ -212,7 +214,9 @@ where
         }
         .fail();
     };
-    let FlightMessage::Schema(schema) = first_flight_message? else {
+    let FlightMessage::Schema(schema) =
+        first_flight_message.map_err(|e| flight_stream_error(&addr, e))?
+    else {
         return IllegalFlightMessagesSnafu {
             reason: "Expect schema to be the first flight message",
         }
@@ -227,6 +231,7 @@ where
     let schema =
         Arc::new(datatypes::schema::Schema::try_from(schema).context(error::ConvertSchemaSnafu)?);
     let schema_cloned = schema.clone();
+    let stream_addr = addr.clone();
     let stream = Box::pin(stream!({
         let _span = tracing_context.attach(common_telemetry::tracing::info_span!(
             "poll_flight_data_stream"
@@ -246,7 +251,8 @@ where
             let flight_message = match flight_message_item {
                 Some(Ok(message)) => message,
                 Some(Err(e)) => {
-                    yield Err(BoxedError::new(e)).context(ExternalSnafu);
+                    yield Err(BoxedError::new(flight_stream_error(&stream_addr, e)))
+                        .context(ExternalSnafu);
                     break;
                 }
                 None => break,
@@ -279,7 +285,8 @@ where
                                 break;
                             }
                             Err(e) => {
-                                yield Err(BoxedError::new(e)).context(ExternalSnafu);
+                                yield Err(BoxedError::new(flight_stream_error(&stream_addr, e)))
+                                    .context(ExternalSnafu);
                                 break;
                             }
                         }
@@ -318,6 +325,21 @@ where
     Ok(Box::pin(record_batch_stream))
 }
 
+fn flight_stream_error(addr: &str, error: error::Error) -> error::Error {
+    let tonic_code = error.tonic_code().unwrap_or(tonic::Code::Unknown);
+    error!(
+        error; "Failed to receive Flight data, addr: {}, code: {}",
+        addr,
+        tonic_code
+    );
+
+    error::Error::FlightGet {
+        addr: addr.to_string(),
+        tonic_code,
+        source: BoxedError::new(error),
+    }
+}
+
 pub fn build_remote_dyn_filter_update_request(
     query_id: impl Into<String>,
     update: RemoteDynFilterUpdate,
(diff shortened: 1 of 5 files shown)


fengjiachun (commented):
Direction looks right: splitting establishment from execution and bounding only the former is the correct shape, and putting the timeout on `RegionRequester` covers both callers, since flownode's embedded frontend shares the same `timeout: None` `NodeClients` (`src/cmd/src/flownode.rs:325`).

Two follow-ups, neither blocking:

**Unary region requests on the same channel are still unbounded.** `timeout: None` applies to the whole `NodeClients`, so `RegionRequester::handle_inner` (inserts, deletes, region DDL, `list_metadata`, dyn-filter updates) has no deadline either. The same "peer accepts but never responds" failure hangs the write path just as easily, and unary requests have no reason to be exempt: the `// Some queries are expected to take long time` rationale only holds for Flight. That case is also simpler than DoGet, since no establishment/execution split is needed.

**Please don't copy the deferred initializer to `GreptimeRequestHandler::do_get`.** It still runs the query before returning headers, and both of its callers depend on that: flownode->frontend sets `timeout = batch_opts.query_timeout` (600s), and metasrv->frontend (`logical_plan`) inherits `ChannelConfig::default()`'s 10s. Today those bound the whole query. Deferring initialization there would silently reduce them to establishment-only and remove the only query-level cap those two paths have.

src/client/src/region.rs
```diff
@@ -109,9 +112,12 @@ impl RegionRequester {
         let mut flight_client = self
             .client
             .make_flight_client(self.send_compression, self.accept_compression)?;
+        // Limit Flight establishment without limiting query stream execution.
+        let mut request = tonic::Request::new(ticket);
+        request.set_timeout(FLIGHT_DO_GET_TIMEOUT);
```
fengjiachun, line 118: `set_timeout` emits the standard `grpc-timeout: 10S` header, which per the gRPC protocol is the deadline for the **entire** RPC. The establishment-only behavior relies on tonic's `GrpcTimeout` layer happening to wrap just the response-headers future (`transport/service/grpc_timeout.rs:79-90`) - that is not what the header means on the wire.

This matters here specifically because the failure being chased involves an intervening listener. If that hop is an L7 gRPC proxy, it will apply the deadline to the whole stream (Envoy's router reads `grpc-timeout` and uses it as the route timeout), so every Flight stream running longer than 10s gets reset mid-query. The read path has no retry, so the query just fails.

A local timeout has identical semantics without leaking the deadline to the peer or to any hop in between:

```rust
let response = tokio::time::timeout(
    FLIGHT_DO_GET_TIMEOUT,
    flight_client.mut_inner().do_get(ticket),
)
.await
```

`do_get()`'s future also resolves as soon as response headers arrive, so this still bounds establishment only, and it no longer depends on a tonic implementation detail.

What is the suspected hop in #8933? If it is L4, this concern goes away.

src/client/src/region.rs
```diff
@@ -109,9 +112,12 @@ impl RegionRequester {
...
         let response = flight_client
             .mut_inner()
-            .do_get(ticket)
+            .do_get(request)
             .await
             .or_else(|e| {
```
fengjiachun, line 123: When this timeout fires, tonic maps it to `Status::cancelled("Timeout expired")`, not `DeadlineExceeded` (`status.rs:642`). `define_from_tonic_status!` then turns `Code::Cancelled` into `StatusCode::Cancelled` - the same code used for a user-initiated cancel and for `attach_timeout`'s `StreamTimeout`.

So the hang this PR is meant to surface reaches operators as an SSE `"canceled"` event (`servers/src/http/handler.rs:312`) and PG `57000 query_canceled`, indistinguishable from a client cancelling its own query. Owning the timeout locally (see the suggestion above) would let you pick an error type that says what actually happened.

src/datanode/src/region_server.rs
```diff
@@ -976,13 +978,19 @@ impl FlightCraft for RegionServer {
...
-            .handle_remote_read(request, query_ctx.clone())
-            .trace(tracing_context.attach(info_span!("RegionServer::handle_read")))
-            .await?;
+        let region_server = self.clone();
+        let initializer_query_ctx = query_ctx.clone();
+        let initializer_tracing_context = tracing_context.clone();
+        let initializer = async move {
+            region_server
+                .handle_remote_read(request, initializer_query_ctx)
+                .trace(initializer_tracing_context.attach(info_span!("RegionServer::handle_read")))
```
fengjiachun, line 987: Deferring initialization moves every `handle_remote_read` failure out of the DoGet status and into the Flight stream. It now surfaces from `recordbatches_from_flight_message_stream` and bypasses the `or_else` in `RegionRequester::do_get_inner` entirely, including its `error!(..., "addr: {}, code: {}", ...)` and the `FlightGetSnafu { addr, .. }` context.

So when one datanode out of many starts failing region reads, the frontend no longer logs which one, and the returned error carries no address. Worth attaching the peer address on the first-message path.

WenyXu:
Thanks for the review. We will keep this PR scoped to frontend-to-datanode Flight `DoGet`. Unary region request deadlines need a separate timeout and write-semantics discussion, and `GreptimeRequestHandler::do_get` remains unchanged to preserve flownode and metasrv query-timeout behavior.

src/client/src/region.rs
WenyXu: Thanks for the protocol-level distinction. The frontend connects to datanode Pod IPs directly in this deployment, so no L7 gRPC proxy is expected. We intentionally retain `Request::set_timeout` here to bound the `DoGet` request itself. If an L7 gRPC hop is introduced later, this needs revisiting because `grpc-timeout` would then apply to the complete stream.

src/client/src/region.rs
WenyXu: Thanks. We considered a dedicated error, but intentionally keep the existing `Cancelled` mapping: the goal here is to prevent an indefinite wait, and callers already handle cancellation. A richer failure classification can be added independently if operators need to distinguish client cancellation from a `DoGet` timeout.

src/datanode/src/region_server.rs
WenyXu: Addressed in ce15fab874. `recordbatches_from_flight_message_stream` now receives the peer address and wraps errors received before Schema or later in the Flight stream with the existing `FlightGet` context. This restores address and tonic-code logging; client coverage verifies the peer address is preserved.

src/client/src/region.rs
```diff
@@ -206,7 +214,9 @@ where
         }
         .fail();
     };
-    let FlightMessage::Schema(schema) = first_flight_message? else {
+    let FlightMessage::Schema(schema) =
```
fengjiachun, line 218: Both error paths now carry `addr`, but the empty-stream branch just above still does not, and that is the one deferred initialization newly makes reachable: if the spawned initializer task panics, `tx` is dropped, `rx` ends immediately, and the client sees a clean `grpc-status: OK` stream with zero messages. Before this PR the datanode either failed the RPC outright or already held the stream and went on to send a schema, so an empty stream was effectively unreachable.

It is also the least diagnosable of the three: no status code, no peer, just `Expect the response not to be empty`. Worth threading `addr` through here as well.

src/client/src/region.rs
```diff
@@ -312,6 +325,21 @@ where
     Ok(Box::pin(record_batch_stream))
 }
 
+fn flight_stream_error(addr: &str, error: error::Error) -> error::Error {
```
fengjiachun, line 330: This logs unconditionally, whereas the server-side equivalent gates on `should_log_error()` (`src/servers/src/grpc/flight/stream.rs:310`).

The distinction matters more on this path than in the `or_else` above. `or_else` only ever sees establishment failures (`Unavailable`, `DeadlineExceeded`), which all deserve an error line. Deferred initialization routes *server-side business errors* through here instead, and `RegionNotFound` / `RegionNotReady` / `RegionBusy` all return `should_log_error() == false` precisely because they are expected transients during region migration or a datanode restart. They also map to `Code::Unavailable`, so `should_retry()` is true and they are meant to be retried. Each retry would now cost a frontend error line.

```rust
if error.status_code().should_log_error() {
    error!(error; "Failed to receive Flight data, addr: {}, code: {}", addr, tonic_code);
}
```

src/client/src/region.rs
```diff
@@ -322,6 +350,28 @@ pub fn build_remote_dyn_filter_update_request(
     )
 }
 
+#[cfg(test)]
```
fengjiachun: This `mod tests` splits `build_remote_dyn_filter_update_request` from `build_remote_dyn_filter_unregister_request`, and the file already has a `mod test` at line 427. Folding the new test in there would keep one test module per file.

fengjiachun (approved):
Rest LGTM

src/client/src/region.rs
WenyXu: Addressed in 0bd254f270. The empty-stream branch keeps the existing `.fail()` construction and wraps that error in `FlightGet` context, so it now carries the peer address and an `Unknown` tonic code. Added coverage for an empty Flight stream.

src/client/src/region.rs
WenyXu: Addressed in 0bd254f270. Flight stream errors now log only when `error.status_code().should_log_error()`, matching the server-side behavior. The returned `FlightGet` context still retains the peer address and tonic code when logging is suppressed.

src/client/src/region.rs
WenyXu: Addressed in 0bd254f270. Moved the peer-context test into the existing `region::test` module so the file has a single test module.

PR #8944, opened 2026-08-24
fix(json2): collect root-aware JSON type hints

fengys1996:

Related to JSON2 root/path projection type inference.

This PR fixes JSON2 queries that read the root column and nested paths in the same scan, such as:

- `SELECT j, j.a FROM t`
- `SELECT j FROM t WHERE j.a.b = 1`

When a JSON2 root column is selected, its read hint is upgraded to `Variant` so the scan keeps the full JSON value instead of narrowing the output schema to a nested struct.

It also updates `json_get` to extract paths from JSON2 `Variant` root values, and moves the regression cases from `json2_limit` to `json2_nested`.

## PR Checklist


fengys1996:
duplicate with https://github.com/GreptimeTeam/greptimedb/pull/8940

PR #8946, opened 2026-08-25
fix(mito2): use target sequence for foreign SSTs

discord9:

Prerequisite for #8865. Supersedes closed #8942, whose GitHub PR head remained stale after the fork branch advanced.

SSTs installed by repartition's ApplyStaging path retain the creator region in their `FileHandle`, while the target manifest assigns them a target-local admission sequence in `FileMeta.sequence`. Their physical row sequences belong to the source region and have no ordering meaning in the target region. Previously those non-zero source sequences participated in target-region deduplication and compaction.

This change uses the existing file-handle provenance to apply the target-local sequence only to foreign SSTs:

- a file is foreign when `FileHandle::region_id()` differs from the current target region metadata;
- foreign files use the target manifest's `FileMeta.sequence` as their logical sequence;
- local SSTs keep their physical row sequences unchanged;
- the existing all-zero sequence compatibility override remains unchanged;
- shared physical SST bytes are never rewritten.

Normal scans and compaction already supply target region metadata to the Parquet reader. Index rebuild is also region-scoped and can read foreign handles, so this PR wires its current target metadata into the same reader path.

Regression coverage includes:

- local all-zero, local non-zero, absent metadata sequence, and foreign-file reader behavior in both SST formats;
- ApplyStaging assigning a target-local barrier to a source-owned handle, followed by a target write and a real StrictWindow compaction;
- compaction replacing both inputs with a target-owned output while preserving the target write as the winner;
- index rebuild applying target partition metadata to a foreign SST. The focused test is mutation-checked: removing the target metadata wiring changes the indexed row count from 100 to 200.

This PR does not change persisted or wire formats. `copy_region_from` deliberately remaps file metadata to the target region and is outside this focused foreign-handle fix.

## PR Checklist

- [x] I have written the necessary rustdoc comments. (No new public API.)

### DIFF
--- a/src/mito2/src/engine/apply_staging_manifest_test.rs
+++ b/src/mito2/src/engine/apply_staging_manifest_test.rs
@@ -12,20 +12,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashSet;
 use std::sync::Arc;
 use std::{assert_matches, fs};
 
 use api::v1::Rows;
+use api::v1::region::{StrictWindow, compact_request};
 use common_function::utils::partition_expr_version;
 use common_recordbatch::RecordBatches;
+use datatypes::arrow::array::AsArray;
+use datatypes::arrow::datatypes::Float64Type;
 use datatypes::value::Value;
 use partition::expr::{PartitionExpr, col};
 use store_api::region_engine::{
     RegionEngine, RegionRole, RemapManifestsRequest, SettableRegionRoleState,
 };
 use store_api::region_request::{
-    ApplyStagingManifestRequest, EnterStagingRequest, RegionFlushRequest, RegionPutRequest,
-    RegionRequest, StagingPartitionDirective,
+    ApplyStagingManifestRequest, EnterStagingRequest, RegionCompactRequest, RegionFlushRequest,
+    RegionPutRequest, RegionRequest, StagingPartitionDirective,
 };
 use store_api::storage::{FileId, RegionId};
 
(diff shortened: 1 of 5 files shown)


src/mito2/src/sst/index.rs
```diff
@@ -706,6 +706,12 @@ pub struct IndexBuildTask {
     pub region_id: RegionId,
     /// The SST file handle to build index for.
     pub file: FileHandle,
+    /// The target region metadata used to decode rows from the SST.
+    ///
+    /// An SST may have been created by another region while being visible in
```
killme2008: When does this happen?

src/mito2/src/sst/parquet/reader.rs
```diff
@@ -471,7 +471,16 @@ impl ParquetReaderBuilder {
...
-        if need_override_sequence(&parquet_meta) {
+        // `region_meta` comes from the Parquet/source file and must not be used as the
+        // target identity. When the caller has no current metadata, the handle is the
+        // only local identity available and therefore denotes a local read.
+        let expected_region_id = self
+            .expected_metadata
+            .as_ref()
+            .map(|metadata| metadata.region_id)
+            .unwrap_or(self.file_handle.region_id());
+        let is_foreign = self.file_handle.region_id() != expected_region_id;
```
killme2008, line 482: We need to add a debug log when `is_foreign = true`?

src/mito2/src/sst/index.rs
killme2008: I see it's explained in the PR decritipn, should we document it here too?

src/mito2/src/engine/apply_staging_manifest_test.rs
evenyag, line 1: Do we need to add tests for staging manifest for this change?

v0y4g3r (commented):
One minor test-organization note from a static review. Nothing blocking.

src/mito2/src/sst/index.rs
```diff
@@ -2123,13 +2135,31 @@ mod tests {
     async fn test_index_build_task_increments_legacy_index_version() {
```
v0y4g3r: Nit: this test was repurposed from covering the local-file legacy index version increment into a foreign-file scenario (source metadata writes the SST, target metadata builds the index, plus the partition-filter mutation check of 100 vs 200 rows). The name still only mentions the legacy version increment, so it now conflates two concerns, and the local-file version-increment case is only covered indirectly by other task tests. Consider renaming it to reflect the foreign-file scenario (e.g. `test_index_build_task_foreign_file_uses_target_metadata`), or splitting it into two tests: keep the original local-file version-increment case and add a dedicated foreign-file case.

v0y4g3r (approved):
LGTM

src/mito2/src/sst/index.rs
discord9: Documented the staging/imported SST scenario and the later rebuild triggers in `694db34132a`.

src/mito2/src/sst/parquet/reader.rs
discord9: Added a foreign-SST-only debug log with the file, source region, and expected target region in `694db34132a`.

src/mito2/src/engine/apply_staging_manifest_test.rs
discord9: Covered by `test_apply_staging_manifest_sequence_domain`, which exercises the real apply-staging path for ordinary and flat SST formats; reran it successfully.

src/mito2/src/sst/index.rs
discord9: Renamed it to `test_index_build_task_foreign_file_uses_target_metadata` while retaining the legacy index-version assertion in `694db34132a`.

PR #8954, opened 2026-08-26
feat(mito2): add SST range index writer

evenyag:

N/A

## What changed and what is the intention?

This PR adds a per-SST range index writer that records the row range for each `(row_group_id, table_id, tsid)` tuple. The index lets readers locate the contiguous rows belonging to a series within an SST row group without scanning unrelated rows.

The writer:

- accepts binary and dictionary-encoded sparse primary keys;
- validates the required `(__table_id, __tsid)` primary-key prefix, sorted keys, and contiguous source row groups;
- emits Parquet rows containing the source row-group ID, table ID, TSID, and half-open `[start, end)` offsets;
- tracks aggregation, write, finish, cleanup, and output-size metrics;
- removes incomplete output when writing fails or is aborted.

This PR also extracts a shared `ParquetIndexWriter` used by both the existing series-index writer and the new range-index writer, keeping Parquet writer setup, byte accounting, finish, and cleanup behavior in one place.

Unit tests cover range aggregation across batches and row groups, binary and dictionary primary keys, output metrics, invalid row-group sequences, unsorted input, and cleanup after failure.

## PR Checklist

### DIFF
--- a/src/mito2/src/sst/parquet/index_writer.rs
+++ b/src/mito2/src/sst/parquet/index_writer.rs
@@ -1,0 +1,163 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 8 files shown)


src/mito2/src/sst/parquet/index_writer.rs
```diff
@@ -0,0 +1,163 @@
...
+            name,
+            object_store,
+            file_name,
+            writer: Some(writer),
+        })
+    }
+
+    /// Writes one batch to the index file.
+    pub(crate) async fn write(&mut self, batch: &RecordBatch) -> Result<()> {
+        self.writer
```
WenyXu, line 126: `format!` is evaluated before `context()` is called, so every successful write and finish allocates this error message. Could we use `with_context(|| UnexpectedSnafu { reason: format!(...) })` here and in `finish()` to construct it only on the error path?

WenyXu (approved):

PR #8959, opened 2026-08-26
fix(query): add global fetch above multi-partition aggregate soft limit

### LINKED ISSUE #8958: DISTINCT + LIMIT on information_schema.ssts_manifest returns more rows than the limit
### What type of bug is this?

Incorrect result

### What did you do?

```sql
select distinct(region_id) from information_schema.ssts_manifest limit 1;
```

At most 1 row.

### What did you see instead?

One row **per target partition** (e.g. 3 rows with `target_partitions = 3`).

The physical plan shows that no global limit operator survives planning — the limit only exists as the per-partition soft limit `lim=[1]` on both `AggregateExec` nodes:

```
AggregateExec: mode=FinalPartitioned, gby=[region_id@0 as region_id], aggr=[], lim=[1]
  RepartitionExec: partitioning=Hash([region_id@0], 3), input_partitions=3
    AggregateExec: mode=Partial, gby=[region_id@0 as region_id], aggr=[], lim=[1]
      RepartitionExec: partitioning=RoundRobinBatch(3), input_partitions=1
        DistributedInspectExec: kind=SstManifest, scan=ScanRequest { projection: [1] }
```

### Root cause analysis

The `lim=[1]` hint pushed down by DataFusion's `LimitedDistinctAggregation` physical rule is only a **per-partition soft limit** (enforced via `group_values_soft_limit` in `GroupedHashAggregateStream`). Correctness relies on a global limit node (`GlobalLimitExec` / `CoalescePartitionsExec: fetch=1`) above the final aggregate.

For `information_schema` inspect tables, the `MergeScan` local-execution fallback (`DistExtensionPlanner::plan_extension`) plans the remote input with `planner.create_physical_plan()`, which already runs the full physical optimizer pipeline once; the outer query planner then runs it **a second time**. In the second pass, upstream DataFusion's `EnforceDistribution::remove_dist_changing_operators` unconditionally strips the fetch-carrying `CoalescePartitionsExec` as an "unnecessary" shuffle (it assumes such operators will be re-added if needed — but the `fetch` semantics are lost and nothing re-adds them). The result is a plan where the only limit enforcement is per-partition.

This is reproducible with vanilla DataFusion by running the physical optimizer pipeline twice over the plan of `select distinct a from t limit 1`: the first pass produces `CoalescePartitionsExec: fetch=1` on top (correct), the second pass drops it (incorrect).

A workaround is to disable the soft-limit pushdown:

```sql
SET datafusion.optimizer.enable_distinct_aggregation_soft_limit = false;
```

main (commit 4ac3423261)

### Relevant log output / more information

The existing sqlness case `tests/cases/distributed/information_schema/ssts_limit.sql` only covers plain `LIMIT` (which goes through the `FilterExec` fetch path guarded by `EnsureGlobalLimitForFetch`), not `DISTINCT + LIMIT`.

v0y4g3r:

Closes #8958.

**Summary**: `SELECT DISTINCT ... FROM information_schema.ssts_manifest LIMIT 1` could return up to `limit * target_partitions` rows instead of `limit` rows. This PR extends the `EnsureGlobalLimitForFetch` physical optimizer rule to restore the missing global limit.

**Root cause**: the `lim` hint that DataFusion's `LimitedDistinctAggregation` rule pushes into `AggregateExec` is only a *per-partition* soft limit (`group_values_soft_limit` in `GroupedHashAggregateStream`). Correctness depends on a global limit node above the final aggregate (`GlobalLimitExec` / `CoalescePartitionsExec: fetch=N`). For information_schema inspect tables, the `MergeScan` local-execution fallback (`DistExtensionPlanner::plan_extension`) plans the remote input with `planner.create_physical_plan()` — which runs the whole physical optimizer pipeline — and the outer planner then optimizes the same plan **a second time**. In that second pass, upstream DataFusion's `EnforceDistribution::remove_dist_changing_operators` strips the fetch-carrying `CoalescePartitionsExec` as an "unnecessary" shuffle, losing the global limit. (Reproducible with vanilla DataFusion 53.1.0 by running the physical optimizer pipeline twice; a follow-up fix to the fork's `remove_dist_changing_operators` is planned.)

**How this PR works**: `EnsureGlobalLimitForFetch` already guards the per-partition `fetch` of `FilterExec` by adding a `CoalescePartitionsExec` with the fetch on top when the output has multiple partitions and no parent provides a global fetch. This PR makes it also treat the soft limit of a `FinalPartitioned` `AggregateExec` as a fetch:

- multi-partition final aggregate with `lim` and no parent global fetch → wrap with `CoalescePartitionsExec::with_fetch(lim)`;
- idempotent: no-op when a parent already provides an equal-or-tighter global fetch (verified by test);
- `Partial` aggregates and single-partition outputs are left untouched.

**Compatibility**: no API or data format changes; only query plans gain a correct global limit node.

**Tests**:
- 4 new unit tests in `src/query/src/optimizer/global_limit.rs`;
- new sqlness regression case in `tests/cases/standalone/common/information_schema/ssts.sql` (verified end-to-end: returns 3 rows without the fix, 1 row with it);
- `cargo nextest run -p query` (660 passed), `cargo clippy -p query --all-targets --all-features -- -D warnings`, `make fmt`.

## PR Checklist

### DIFF
--- a/src/query/src/optimizer/global_limit.rs
+++ b/src/query/src/optimizer/global_limit.rs
@@ -128,17 +128,32 @@ impl EnsureGlobalLimitForFetch {
     }
 }
 
-/// Returns the soft limit of a `FinalPartitioned` [`AggregateExec`], if any.
+/// Returns the soft limit of a multi-partition-capable [`AggregateExec`], if any.
 ///
 /// The `lim` hint pushed down by DataFusion's `LimitedDistinctAggregation` rule
 /// is only enforced per partition (see `group_values_soft_limit` in
 /// `GroupedHashAggregateStream`), so a multi-partition final aggregate needs a
 /// global fetch on top, just like a multi-partition [`FilterExec`].
+///
+/// Only unordered soft limits (`FinalPartitioned` or `SinglePartitioned` mode)
+/// qualify. Ordered limits come from `TopKAggregation`, where each partition
+/// keeps its local top-N candidates and the global top-N is chosen by the sort
+/// machinery above; truncating the merged candidates with a coalesce fetch
+/// would drop true top-N rows.
 fn aggregate_soft_limit(plan: &Arc<dyn ExecutionPlan>) -> Option<usize> {
     plan.as_any()
         .downcast_ref::<AggregateExec>()
-        .filter(|agg| matches!(agg.mode(), AggregateMode::FinalPartitioned))
-        .and_then(|agg| agg.limit_options().map(|options| options.limit()))
+        .filter(|agg| {
+            matches!(
+                agg.mode(),
+                AggregateMode::FinalPartitioned | AggregateMode::SinglePartitioned
+            )
+        })
+        .and_then(|agg| {
+            agg.limit_options()
+                .filter(|options| options.descending().is_none())
+                .map(|options| options.limit())
+        })
 }
 
 #[derive(Clone)]
@@ -330,6 +345,43 @@ mod tests {
             EnsureGlobalLimitForFetch::optimize_plan(agg, ParentContext::default()).unwrap();
 
         assert!(optimized.as_any().is::<AggregateExec>());
+    }
+
+    #[test]
+    fn ignores_topk_ordered_aggregate_limit() {
+        // Ordered limit options come from TopKAggregation: each partition only
+        // keeps its local top-N candidates, and the global top-N is chosen by
+        // the sort machinery above. Truncating the merged candidates with a
+        // coalesce fetch would drop true top-N rows.
+        let agg = agg_with_limit_options(
+            unordered_input(),
+            AggregateMode::FinalPartitioned,
+            LimitOptions::new_with_order(1, true),
+        );
+
+        let optimized =
+            EnsureGlobalLimitForFetch::optimize_plan(agg, ParentContext::default()).unwrap();
+
+        assert!(optimized.as_any().is::<AggregateExec>());
+    }
+
+    #[test]
+    fn adds_global_limit_for_single_partitioned_aggregate_soft_limit() {
+        // CombinePartialFinalAggregate rewrites FinalPartitioned into
+        // SinglePartitioned while preserving limit_options; its output can
+        // still have multiple partitions.
+        let agg = agg_with_limit_options(
+            unordered_input(),
+            AggregateMode::SinglePartitioned,
+            LimitOptions::new(1),
+        );
+
+        let optimized =
+            EnsureGlobalLimitForFetch::optimize_plan(agg, ParentContext::default()).unwrap();
+
+        assert!(optimized.as_any().is::<CoalescePartitionsExec>());
+        assert_eq!(optimized.fetch(), Some(1));
+        assert_eq!(optimized.output_partitioning().partition_count(), 1);
     }
 
     #[test]
@@ -649,6 +701,14 @@ mod tests {
         mode: AggregateMode,
         limit: usize,
     ) -> Arc<dyn ExecutionPlan> {
+        agg_with_limit_options(input, mode, LimitOptions::new(limit))
+    }
+
+    fn agg_with_limit_options(
+        input: Arc<dyn ExecutionPlan>,
+        mode: AggregateMode,
+        limit_options: LimitOptions,
+    ) -> Arc<dyn ExecutionPlan> {
         let schema = input.schema();
         let group_by = PhysicalGroupBy::new_single(vec![(
             col("a", schema.as_ref()).unwrap(),
@@ -657,7 +717,7 @@ mod tests {
         Arc::new(
             AggregateExec::try_new(mode, group_by, vec![], vec![], input, schema)
                 .unwrap()
-                .with_limit_options(Some(LimitOptions::new(limit))),
+                .with_limit_options(Some(limit_options)),
         )
     }
 
--- a/src/query/src/optimizer/global_limit.rs
+++ b/src/query/src/optimizer/global_limit.rs
@@ -285,6 +285,7 @@ mod tests {
     use datafusion::arrow::record_batch::RecordBatch;
     use datafusion::execution::TaskContext;
     use datafusion::physical_expr::expressions::{col, lit};
+    use datafusion::physical_optimizer::combine_partial_final_agg::CombinePartialFinalAggregate;
     use datafusion::physical_plan::aggregates::{
         AggregateExec, AggregateMode, LimitOptions, PhysicalGroupBy,
     };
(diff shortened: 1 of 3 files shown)


src/query/src/optimizer/global_limit.rs
```diff
@@ -126,6 +128,19 @@ impl EnsureGlobalLimitForFetch {
...
+///
+/// The `lim` hint pushed down by DataFusion's `LimitedDistinctAggregation` rule
+/// is only enforced per partition (see `group_values_soft_limit` in
+/// `GroupedHashAggregateStream`), so a multi-partition final aggregate needs a
+/// global fetch on top, just like a multi-partition [`FilterExec`].
+fn aggregate_soft_limit(plan: &Arc<dyn ExecutionPlan>) -> Option<usize> {
+    plan.as_any()
+        .downcast_ref::<AggregateExec>()
+        .filter(|agg| matches!(agg.mode(), AggregateMode::FinalPartitioned))
+        .and_then(|agg| agg.limit_options().map(|options| options.limit()))
```
discord9: `LimitOptions` is also used by `TopKAggregation`, which installs ordered options via `LimitOptions::new_with_order`. Treating every option as the unordered distinct soft limit can insert `CoalescePartitionsExec(fetch=N)` between a partitioned TopK aggregate and its global sort, discarding candidates from other partitions before the global TopK is chosen. Could this restrict the workaround to unordered options (for example, `options.descending().is_none()`) and add a multi-partition TopK execution regression?

src/query/src/optimizer/global_limit.rs
```diff
@@ -126,6 +128,19 @@ impl EnsureGlobalLimitForFetch {
...
+/// Returns the soft limit of a `FinalPartitioned` [`AggregateExec`], if any.
+///
+/// The `lim` hint pushed down by DataFusion's `LimitedDistinctAggregation` rule
+/// is only enforced per partition (see `group_values_soft_limit` in
+/// `GroupedHashAggregateStream`), so a multi-partition final aggregate needs a
+/// global fetch on top, just like a multi-partition [`FilterExec`].
+fn aggregate_soft_limit(plan: &Arc<dyn ExecutionPlan>) -> Option<usize> {
+    plan.as_any()
+        .downcast_ref::<AggregateExec>()
+        .filter(|agg| matches!(agg.mode(), AggregateMode::FinalPartitioned))
```
discord9: This also appears to miss `AggregateMode::SinglePartitioned`. `CombinePartialFinalAggregate` can turn `FinalPartitioned` into `SinglePartitioned` while preserving `limit_options()`, and its output can still have multiple partitions. An unordered distinct soft limit would then have the same per-partition semantics but would not get a global boundary here. Should this match both `FinalPartitioned | SinglePartitioned` (while continuing to exclude `Partial` / `PartialReduce`) and add a corresponding test?

src/query/src/optimizer/global_limit.rs
v0y4g3r: Fixed in c73bf61475. Good catch — the guard now only matches unordered soft limits (`limit_options().descending().is_none()`), i.e. only the `LimitedDistinctAggregation` case. Ordered options from `TopKAggregation` are left alone: each partition keeps its local top-N candidates and the global top-N is chosen by the sort machinery above, so inserting a coalesce fetch below the sort could indeed drop true top-N rows. Added a regression test `ignores_topk_ordered_aggregate_limit` (fails before the fix, passes after).

src/query/src/optimizer/global_limit.rs
v0y4g3r: Fixed in c73bf61475. The guard now matches both `AggregateMode::FinalPartitioned` and `AggregateMode::SinglePartitioned`, since `CombinePartialFinalAggregate` preserves `limit_options` when rewriting FinalPartitioned into SinglePartitioned, whose output can still be multi-partitioned. (`Single`/`Final` modes require single-partition input and are excluded by the existing partition-count guard.) Added a regression test `adds_global_limit_for_single_partitioned_aggregate_soft_limit` (fails before the fix, passes after).

src/query/src/optimizer/global_limit.rs
```diff
@@ -265,6 +298,92 @@ mod tests {
...
+        let agg = agg_with_limit_options(
+            unordered_input(),
+            AggregateMode::FinalPartitioned,
+            LimitOptions::new_with_order(1, true),
+        );
+
+        let optimized =
+            EnsureGlobalLimitForFetch::optimize_plan(agg, ParentContext::default()).unwrap();
+
+        assert!(optimized.as_any().is::<AggregateExec>());
```
discord9: Could this be an execution-level regression for the complete TopK path rather than only a node-shape assertion? The wrong-result risk occurs across `SortExec(fetch=1) -> ordered AggregateExec -> multi-partition input`: each aggregate partition emits local candidates and the global sort must see all of them. For example, placing `100` in one partition and the winning ascending candidate `1` in another would prove the optimized plan still returns `1`; the previous implementation could insert `CoalescePartitionsExec(fetch=1)` before the global sort and discard that winning partition.

src/query/src/optimizer/global_limit.rs
```diff
@@ -265,6 +298,92 @@ mod tests {
...
+        let agg = agg_with_limit_options(
+            unordered_input(),
+            AggregateMode::SinglePartitioned,
+            LimitOptions::new(1),
+        );
+
+        let optimized =
+            EnsureGlobalLimitForFetch::optimize_plan(agg, ParentContext::default()).unwrap();
+
+        assert!(optimized.as_any().is::<CoalescePartitionsExec>());
```
discord9, line 471: Could this exercise the production-like `SinglePartitioned` path and execute it? The current synthetic input reports `UnknownPartitioning(3)`, while `SinglePartitioned` requires hash-partitioned input and normally comes from `CombinePartialFinalAggregate` preserving `limit_options()`. A hash-partitioned multi-partition case (ideally produced through that rewrite) should collect the optimized plan and assert the unordered distinct limit returns one row globally, not one per partition.

src/query/src/optimizer/global_limit.rs
v0y4g3r: Done in 21cc61434e — the test is now an execution-level regression for the complete TopK path. It builds the real post-EnforceDistribution shape `SortExec(fetch=1) -> CoalescePartitionsExec -> ordered AggregateExec` over two partitions, with a losing candidate (100) in one partition and the global winner (1) in the other. Verified both directions: with the guard removed the inserted coalesce fetch makes the executed plan return 100 (the winner is discarded exactly as you described), and with the guard narrowed it returns 1. Also added a shape assertion that no candidate-truncating node sits below the merge.

src/query/src/optimizer/global_limit.rs
v0y4g3r: Done in af866b3b1e — the SinglePartitioned test now goes through the production path and executes. It builds FinalPartitioned + Partial over a RepartitionExec(Hash) input, runs the real CombinePartialFinalAggregate rewrite (asserting the resulting mode is SinglePartitioned with limit_options preserved and 3 output partitions), then executes the optimized plan and asserts the unordered distinct limit returns exactly one row globally instead of one per partition. A duplicate group key across raw partitions exercises the hash merge. Verified to fail (no global fetch inserted) when the mode guard is narrowed back to FinalPartitioned only.

discord9:
I investigated whether we can fix this in DataFusion rather than keep extending the aggregate-specific compensation in GreptimeDB.

The local end-to-end experiment now passes with that compensation removed. There were two fetch-loss paths in the pinned DF fork:

- `EnforceDistribution` removes distribution-changing operators without preserving a fetched merge's global limit. A fetched sort-preserving merge also needs to retain its ordering, not just its numeric fetch.
- `EnforceSorting` can remove an inner fetched `CoalescePartitionsExec` during sort parallelization, including below a partial COUNT. The fix makes fetched coalesces stop the removal connection and makes the deletion site respect that connection flag (the latter is necessary for a Union with differently connected branches).

The second path mattered in the actual SQL regression: the distribution fix alone passed the initial native tests, but `COUNT(*) FROM (SELECT DISTINCT ... LIMIT 1)` still returned **3 instead of 1**. A new full-optimizer regression reproduces this even with distinct-aggregation soft limits disabled. With both fixes, the actual standalone and distributed SQL cases return **1**.

Validation of the local fork + Greptime integration:
- DF core integration: **919 passed**; physical optimizer library: **20 passed**.
- DF formatting and Rust 1.93 default-feature, all-targets strict Clippy: passed.
- Greptime query tests: **661 passed, 2 skipped**; check and binary build: passed.
- Actual SQLness: standalone `ssts`, distributed `ssts`, and distributed `ssts_limit` all executed and passed. The common DISTINCT count expectation remains **1**. The only regenerated snapshot change is an EXPLAIN repartition above the retained global fetch; query results are unchanged.

This lets us remove the aggregate-specific workaround (41 net production lines) while leaving the original Filter handling unchanged. The changes are still local: the intended submission is a minimal fix PR to **our DataFusion fork**, followed by a **GreptimeDB PR** using an available fork revision and removing the aggregate compensation. The experimental absolute-path dependency wiring will not be submitted.

I suggest using that root-fix route instead of adding more aggregate-specific compensation here. No changes have been pushed to this PR by this investigation.

PR #8960, opened 2026-08-26
feat: Add built-in daemon mode for standalone service

### LINKED ISSUE #7314: Feature Request: Add built-in daemon mode to keep GreptimeDB running independently of shell sessions

API improvement

# Description
When running GreptimeDB in standalone mode on environments like ARM64 boards or edge devices, the process currently depends on the shell session that started it. If the user disconnects (e.g., network cable removed or SSH session closed), the GreptimeDB process exits.

To improve resilience and usability in embedded or self-hosted setups, it would be helpful to have an optional built-in daemon mode, allowing GreptimeDB to continue running in the background independently of the user session — similar to the daemonize yes option in Redis.

The idea comes from #7295

# Motivation
Users often deploy GreptimeDB on ARM-based or edge environments where systemd may not be used.
Running via nohup or systemd works but adds operational complexity.
A native --daemon or configuration flag would make standalone and test deployments more robust and user-friendly.
Proposed Solution
Add a configuration option or CLI flag such as:

```bash
greptimedb standalone --daemon
```

or `-d` for short.

This would run the server as a background process detached from the shell, handling its own stdout/stderr redirection and PID management.

# Alternatives Considered
* Using nohup or creating a custom systemd service file works, but it’s less convenient for quick setups and embedded use cases.
* Relying on containerization helps, but is not always practical on lightweight ARM systems.

# Environment Example
 
* Platform: ARM64 development board (Linux)
* Mode: Standalone

# Use case: Local testing and edge data collection

Benefits
* More stable background execution
* Better user experience for standalone and edge deployments
* Consistent behavior regardless of interactive shell sessions

### Implementation challenges

_No response_

tian1220A:
## What does this PR do

Add a built-in daemon mode to `greptimedb standalone start`, allowing the server to run in the background independently of the shell session — similar to Redis `daemonize yes`.

Closes #7314

## Changes

- Add `--daemon` / `-d` CLI flag to `standalone start`.
- Use the `daemonize` crate (Unix only) to fork/setsid and detach from the controlling terminal before the Tokio runtime is built.
- stdin/stdout/stderr are redirected to `/dev/null`; logs continue to go to `data_home/logs/` via the existing tracing appender.
- On non-Unix platforms, `--daemon` is a no-op with a warning (runs in foreground).

## Testing

- Verified `--daemon` and `-d` are both parsed and trigger daemonization.
- Verified the daemon process has PPID=1, its own session, and no controlling terminal.
- Verified HTTP health check returns 200 and all four ports listen.
- Verified the process survives an SSH disconnect (control-group process dies, daemon survives).


src/cmd/src/standalone.rs
```diff
@@ -291,6 +302,9 @@ pub struct StartCommand {
     /// The working home directory of this standalone instance.
     #[clap(long)]
     data_home: Option<String>,
+    /// Run in the background as a daemon (Unix only; no-op on Windows).
+    #[clap(short, long)]
+    daemon: bool,
```
sunng87, line 308: Can we just gate this with `#[cfg(unix)]` so there is no such option on non-unix platforms?

src/cmd/Cargo.toml
```diff
@@ -118,6 +118,7 @@ tonic.workspace = true
 tracing-appender.workspace = true
 
 [target.'cfg(unix)'.dependencies]
+daemonize.workspace = true
```
killme2008: I think we can move this crate directly into `cmd`. It isn’t used in any other sub-crates.

killme2008 (approved):
LGTM

killme2008:
Thanks. Congrats for your first contribution! @tian1220A

